PR #460, opened 2015-12-29
There is no need to have pacman force a refresh of all package lists.

SushiDude:
https://wiki.archlinux.org/index.php/Pacman#Upgrading_packages


felixonmars:
I would say +65536 here xD

PR #462, opened 2015-12-29
vim: add :w, :wq, i

shoeletz:
Added vim commands for save, save and exit, and insert mode


waldyrious:
Nice. But could you change the verb tense (exiting --> exit, etc.)?

sethwoodworth:
`[esc]` in vim takes you to Normal mode, not command mode.  Pressing `:` is what takes you into command mode.

PR #181, opened 2014-04-04
wall: add page

Gubolin:

PR #464, opened 2015-12-29
Move some previously linux-only pages to common.

hayd:
These are available on OSX too.

As mentioned in #181.


hayd:
Note: The following are in both linux and osx:

    du.md head.md hostname.md locate.md netstat.md shutdown.md top.md

Should these all be brought into common?

sonph:
Not sure about the others but I'm pretty sure that OS X has a default implementation of `du` that is different from the GNU `du` available on Linux.

igorshubovych:
@hayd have you tried them both in Linux and OSX to ensure that keys are the same?

just FYI, there are [Linux version of du](https://github.com/tldr-pages/tldr/blob/master/pages/linux/du.md) and [OSX version of du](https://github.com/tldr-pages/tldr/blob/master/pages/osx/du.md), and they are different.

I always try my commands before I submit a pull request.

hayd:
Ah! There are a couple of small differences in a couple. I'll update (which will involve moving back to linux & creating an osx) shortly.

igorshubovych:
If you are moving pages, please ensure that every example works in the other environment too.

igorshubovych:
`base64`, `iostat`, `wall` - have different options in OSX and Linux. The pages cannot be moved
There is no `watch` on OSX.

rubenvereecken:
@hayd still up for that update or should we close this?

igorshubovych:
i'd rather close it it. the only remaining page which remains is `xargs`, and i rather do it carefully. we need to double check the options there.

PR #467, opened 2015-12-29
paste: add page

chuanconggao:


pages/common/paste.md
```diff
@@ -0,0 +1,23 @@
+# paste
+
+> Merge lines of files.
+
+- join all the lines into one, using TAB as delimiter to separate the columns.
```
waldyrious: I think the "to separate the columns" is unnecessary and possibly even confusing. How about removing that part?

pages/common/paste.md
```diff
@@ -0,0 +1,23 @@
...
+
+`paste -s {{file}}`
+
+- join all the lines into one, using the specified delimiter to separate the columns.
```
waldyrious: same as above

pages/common/paste.md
```diff
@@ -0,0 +1,23 @@
...
+
+`paste -s -d {{delimiter}} {{file}}`
+
+- merge two files side by side, using TAB as delimiter to separate the columns.
```
waldyrious: here too

chuanconggao:
@waldyrious Thanks for the suggestions. I have made some changes.

pages/common/paste.md
```diff
@@ -0,0 +1,23 @@
...
+
+- join all the lines into one, using TAB as delimiter.
+
+`paste -s {{file}}`
+
+- join all the lines into one, using the specified delimiter.
+
+`paste -s -d {{delimiter}} {{file}}`
+
+- merge two files side by side, using TAB as delimiter. Lines in the first file and the second file form the first column and the second column, respectively.
```
igorshubovych: Think about how to make the description shorter.

E.g.
- Merge two files side by side, each in its column, using TAB as delimiter.

Or propose your example

pages/common/paste.md
```diff
@@ -0,0 +1,23 @@
...
+
+`paste {{file1 file2}}`
```
igorshubovych: Should be 
`paste {{file1}} {{file2}}`

pages/common/paste.md
waldyrious: Brilliant conciseness skills, @igorshubovych!

waldyrious:
It looks much better now :) One question though: does -s stand for "single line"? If so, the example description should use that expression, as it makes it easier to understand what the option does and to memorize its short name.

chuanconggao:
@waldyrious In its manpage, `-s` actually stands for `--serial`. I think using that may cause some confusion and need more explanation.

waldyrious:
Well, I think we can take the liberty of calling it "single line" as it is helpful for beginners, a good mnenomic, and technically it isn't incorrect. What do you think? "join all the lines into a single line, using TAB as delimiter."

chuanconggao:
@waldyrious Sorry, I misunderstood your before. Your suggestion does sound more clear. I have made the change.

waldyrious:
Ok, I'm very happy with how this page turned out, pretty simple and intuitive as all TLDRs should be! LGTM :+1:

igorshubovych:
LGTM. But please, squash the commits.

chuanconggao:
@igorshubovych It is my first time using rebase, and it took me quite a while. Now it is clean.

felixonmars:
LGTM, thanks :)

waldyrious:
@chuanconggao while it's fresh on your mind, take a look at the candidate `tldr git-rebase` (#473) in case it could be improved to make it easier for others :)

PR #472, opened 2015-12-29
Partially fix #468: warns that Ruby and Bundler must be installed

### LINKED ISSUE #468: add clear error message when Ruby is missing
As someone wanting to contribute some pages, this was confusing:

    me@laptop: ~/code/tldr$ make setup
    Git pre-commit hook installed.
    make: bundle: Command not found
    Makefile:21: recipe for target 'deps' failed
    make: *** [deps] Error 127

In #428 it appears the solution (on Debian) is:

    sudo apt-get install ruby
    gem install bundler

Make setup should say something more clear about what the issue is.

igorshubovych:
- adds prerequisites task to Makefile
- it warns about installation of Ruby and Bundler.
- It does not actually check anything, it just warns

Probably this should fix #468, as I think people were mostly frustrated by lack of messaging.

Actual checks that software is not installed is more complicated, and I might do it later.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,5 @@
 default: lint
 
-all: setup index
-
 index:
 	@echo "WARNING!"
 	@echo "Index rebuilding is deprecated."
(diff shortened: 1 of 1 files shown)


ostera:
Good enough for now and very much needed.

PR #474, opened 2015-12-29
adds emacs docs to quit and open file

sethwoodworth:


waldyrious:
Isn't it a little awkward to start with the command to exit emacs? I'd swap the order of the examples. Other than that, LGTM.

rubenvereecken:
I'd second @waldyrious

pages/common/emacs.md
```diff
@@ -0,0 +1,11 @@
+# emacs
+
+> A command-line text editor (and operating system)
+
+- Exit emacs
+
+`CTRL`+`x`, `CTRL`+`c`
```
ostera: I think the notation for this is `C-x C-c`

sethwoodworth:
@waldyrious As a non-emacs user, I've found myself in more situations where I've been trying to quit emacs than start it.  Happy to sort it by 'sane' order rather than my pass at most-frequently asked.
@ostera Will-do

sethwoodworth:
@waldyrious Reordered, and merged with #491

pages/common/emacs.md
waldyrious: Here the 2/3 LGTM policy would have helped; I would argue that this notation is an emacs idiom and we should probably use something a little more generic, aimed at beginners. You made this point yourself recently in another PR regarding `cr` vs. `enter`.

PR #443, opened 2015-12-29
Add python3.md

rselwyn:
I pretty much took the python.md and added a similar python3.md.  Just `python` will call up a python 2 interpreter, but `python3` will call a python 3 interpreter.  Because they are different, I feel having a different page would be appropriate.

PR #353, opened 2015-12-09
Adds -m syntax to python page

ostera:
I'm wondering if it'd make sense to include samples of the actual purposes for running the `-m` syntax, like running a webserver, or parsing json output in the terminal, with examples for those.

```md
* Parse JSON output
`cat raw.json | python -m json.tool`

* Serve the current directory at `localhost:8000`

`python -m SimpleHTTPServer`
```

Any ideas?

PR #328, opened 2015-11-22
Add python

igorshubovych:
Added pages/common/python.md Python interpreter (similar to node).

PR #475, opened 2015-12-29
adds python instructions to serve current folder over http

sethwoodworth:


waldyrious:
I believe @ostera was involved in a recent discussion about how to best demonstrate python modules in TLDR, but I may be misremebering.

sethwoodworth:
Some previous conversations were #353, #328, and (kinda) #443.

 @ostera I too have python's httpserver set to an alias.  It is important to note that in python3 the module has been renamed to `-m http.server`!

igorshubovych:
i don't think we should merge this.

waldyrious:
@igorshubovych do elaborate?

ostera:
@sethwoodworth while I think knowing which python modules you can just use comes in very very handy all the time (as our aliases point out), we have to adhere with our vision and only strictly provide help for the command you're looking for, and in this case you really be looking for a web server in general, and a python _module_ in particular – not really the python executable.

It has been mentioned a few times that some of this pages should have _see also's_ or _notes_, but we haven't reached an agreement there yet, but if we did, it could be considered adding the SimpleHTTPServer module as a result in a search for anything web-server related. Of course that's out of the scope of this pull request. But feel free to join us on [gitter](https://gitter.im/tldr-pages/tldr) to talk about it. Thanks again for taking the time to contribute!

igorshubovych:
Don't get me wrong. It is a convenient command.

However it does not help to learn the syntax of Python CLI. It is rather a trick, suitable for [cheat](https://github.com/chrisallenlane/cheat), then part of simplified man-page.

Additionally Python 2 and 3 examples are different in this case, which makes us add 2 examples.

sethwoodworth:
:+1:!

PR #476, opened 2015-12-29
Added dnf and yum to linux.

cdminigun:
Added support for Fedora, CentOS and RHEL.

### DIFF
--- a/pages/linux/dnf.md
+++ b/pages/linux/dnf.md
@@ -1,0 +1,24 @@
+# dnf
+
+>Package management utility for RHEL, Feodra, and CentOS (replaces yum)
+
+- Synchronize list of packages and versions available. This should be run first, before running subsequent dnf commands.
+
+`dnf update`
+
+- Install a new package
+
+`dnf install {{package}}`
+
+- Install a new package and assume yes to all questions
+ 
+`dnf -y install {{package}}`
+
+- Remove a package
+
+`dnf remove {{package}}`
+
+- Upgrade installed packages to newest available versions
+
+`dnf upgrade`
+
--- a/pages/linux/yum.md
+++ b/pages/linux/yum.md
@@ -1,0 +1,24 @@
+# yum
+
+>Package management utility for RHEL, Feodra, and CentOS (for older versions)
+
+- Synchronize list of packages and versions available. This should be run first, before running subsequent yum commands.
+
+`yum update`
+
+- Install a new package
+
+`yum install {{package}}`
+
+- Install a new package and assume yes to all questions (Also works with update, great for automated updates)
+ 
+`yum -y install {{package}}`
+
+- Remove a package
+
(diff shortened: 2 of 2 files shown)


pages/linux/dnf.md
```diff
@@ -0,0 +1,24 @@
+# dnf
+
+> RHEL, Feodra, and CentOS package management utility (Replaces yum)
```
waldyrious: Reordering this sentence makes more sense IMO: "Package management utility for RHEL, Feodra, and CentOS (replaces yum)"

pages/linux/yum.md
```diff
@@ -0,0 +1,24 @@
+# yum
+
+> RHEL, Feodra, and CentOS package management utility (For older versions)
```
waldyrious: Same as above.

waldyrious:
Just a minor nitpick, which I've added as inline comments. Other than that, looks great :+1:

cdminigun:
Thanks for the recommendations! I agree that your wording makes more sense and I have updated the files.

waldyrious:
Looks good to me :)

While we wait for another approval (we have implemented a policy of only merging a PR after at least 2 thumbs-up from the maintainers), feel free to fix the capitalization of "For older versions" (should be lowercase).

cdminigun:
Great policy change! Will definitely help validate commits. 

I've gone ahead and changed the capitalization as per your request.

pages/linux/dnf.md
```diff
@@ -0,0 +1,24 @@
+# dnf
+
+>Package management utility for RHEL, Feodra, and CentOS (replaces yum)
+
+- Synchronize list of packages and versions available. This should be run first, before running subsequent dnf commands.
+
+`dnf update`
+
+- install a new package
```
felixonmars: I think the first letter of a sentence should be consistently capitalized, so this line should become "Install a new package", and same for the two examples below.

cdminigun:
Changed were added per your request. Thanks for catching that, @felixonmars.

pages/linux/yum.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Synchronize list of packages and versions available. This should be run first, before running subsequent yum commands.
+
+`yum update`
+
+- Install a new package
+
+`yum install {{package}}`
+
+ - Install a new package and assume yes to all questions (Also works with update, great for automated updates)
```
felixonmars: And now it has different spacing here between the two files. Sorry to be nitpicking :P

cdminigun:
I believe I fixed what you were talking about. Noticed that a few lines had spaces before text. 

If that wasn't the issue, would you please clarify? Also, no problem. I'm more than happy to help make it as perfect as possible.

felixonmars:
All fixed now, thanks :)

PR #480, opened 2015-12-30
Add npm -g option

sonph:
Add the `npm install -g` example

### DIFF
--- a/pages/common/npm.md
+++ b/pages/common/npm.md
@@ -10,6 +10,10 @@
 
 `npm install`
 
+- Download and install a module globally
+
+`npm install -g {{module_name}}`
+
 - Download a given dependency, and add it to the package.json
 
 `npm install {{module_name}}@{{version}} --save`


pages/common/npm.md
```diff
@@ -10,6 +10,10 @@
 
 `npm install`
 
+- Download and install a module globally so it's always available in the shell
```
ostera: This isn't always true, you could spawn a node shell with a different NODE_LIB path that knows nothing about the global npm modules. Also it could be interpreted as available from the terminal shell (bash, csh, zsh) and thus might be misleading – most packages can be installed globally and don't provide any binaries.

igorshubovych: Let's just write 
"Download and install a module globally"

sonph: Yup I agree that wasn't technically correct. Added a commit.

igorshubovych:
Please squash commits. Overwise looks good.

sonph:
Squashed :)

ostera:
:+1:

PR #482, opened 2015-12-30
add virtualenv page

sonph:
Examples include creating, sourcing and deactivating a virtual environment.

### DIFF
--- a/pages/common/virtualenv.md
+++ b/pages/common/virtualenv.md
@@ -1,0 +1,16 @@
+# virtualenv
+
+> Create virtual isolated Python environments
+
+- Create a new environment
+
+`virtualenv {{path/to/venv}}`
+
+- Start (select) the environment
+
+`source {{path/to/venv}}/bin/activate`
+
+- Stop the environment
+
+`deactivate`
+


ostera:
I wanted to make annotations on the diffs, but I feel it's better to address things at the PR level.

It feels you're writing this PR aimed at someone who knows what a `venv` is. Most people might, and likely will, not know. A slightly different approach could be better. Since virtualenv is mostly used within projects, why not refer to `venv` as `path/to/project` ? That makes much more sense for a first time reader and still gives an experienced virtualenv-er the knowledge they need without sacrificing too much terseness.

Just a thought, I've seen this pattern in a bunch of PR's and I think it should be addressed rather soon.

waldyrious:
I agree with @ostera, the suggested change improves beginner-friendliness without sacrificing conciseness neither making it harder for experts.

felixonmars:
I think `path/to/project` is a bit misleading, since he may not want to put the venv right inside his project (which introduces dirs like bin, include, lib). `path/to/venv` may be better :)

sonph:
Personally I think `/path/to/env` works better.

> It feels you're writing this PR aimed at someone who knows what a `venv` is. Most people might, and likely will, not know.

Correct me if I'm wrong, but if someone looks for the `virtualenv` TLDR page then he/she should already know what a virtual env is and how it works, right?

ostera:
@sonph if they did then good for them, but if you were starting out and I sent you a link to it you'll probably ask me what that `venv` is. Either way, as long as it's clear that it's a path it's fine by me.

sonph:
Ok changed to `path/to/venv` to indicate that it's a path.

ostera:
Thanks @sonph!

PR #489, opened 2015-12-30
Update psql.md

shybovycha:
Added password prompt parameter


igorshubovych:
LGTM :)
Even though i would remove those asterisks. But don't worry. We will do in a separate PR.

PR #490, opened 2015-12-30
Update npm.md

shybovycha:
Added cleanup command

### DIFF
--- a/pages/common/npm.md
+++ b/pages/common/npm.md
@@ -21,3 +21,7 @@
 - Publish the current project
 
 `npm publish`
+
+- Cleanup packages (removes packages, which are installed but are not listed in `package.json`)
+
+`npm prune`


ostera:
:+1:

felixonmars:
:shipit:

igorshubovych:
You hurried up to merge it, but i actually disagree with it.

This is the list of NPM commands
npm_commands

Pretty long. We will not be able to cover them all.

Out of those commands, `npm prune` is really something artificial, used by hackers and advanced users. I am using `npm` for several years and run it a few times.

There are better candidates to include to this page `npm list`, `npm uninstall`, `npm run`, `npm test`, `npm help`, `npm init` and more. Adding all of these commands will make page look bloated already.

I will create a PR to remove it.

PR #491, opened 2015-12-30
Started tldr for emacs

shybovycha:


ostera:
:+1: for a start. Would be nice to see some more commands!

shybovycha:
As with VIM, more cool things are behind keyboard shortcuts and in-app commands :simple_smile:

waldyrious:
Note that there's already #474 so we should probably coordinate both PRs. /cc @sethwoodworth

sethwoodworth:
@shybovycha I like your description better, so I rebased my PR on this one.

PR #493, opened 2015-12-30
Added a page for the Linux lsattr command and extended grep, find and curl pages"

Cvetomird91:


rubenvereecken:
I don't like being that guy, but would you please consider splitting this PR into multiple PRs, one per affected page? I see your commits aren't cleanly separated either, so cherry picking isn't the obvious answer. You should be able to do something along the lines of checking out a file at a time and then committing it. If you want I'll guide you through, it's just that some people prefer copy/pasting. 

The reason we prefer separate PRs is that it makes it less messy if we have comments on a specific page that don't concern the others.

Cvetomird91:
Thank you for the quick feedback! I fully understand your concern, rubenvereecken. 

Yes, this will be taken into consideration the next time I commit content to the TLDR pages. 

I will fix my commits shortly. :) 

Best Regards,
Tsvetomir

PR #500, opened 2015-12-30
Fixed up cat.md

rselwyn:
I fixed it to conform with the file bracketing (using {{ }}) requirement.  I also added appending.

### DIFF
--- a/pages/common/cat.md
+++ b/pages/common/cat.md
@@ -2,10 +2,14 @@
 
 > Print and concatenate files.
 
-- Print the contents of *file1* to the standard output
+- Print the contents of a file to the standard output
 
-`cat {{file1}}`
+`cat {{file}}`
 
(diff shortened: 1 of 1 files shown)


ostera:
This was just fine as it was. Adding `.ext` makes no difference.

rselwyn:
I can remove the .ext and just stick with the append @ostera

ostera:
:+1:

waldyrious:
As we discussed recently (/cc @rprieto, @igorshubovych), the `>` and `>>` are not strictly `cat` features, but they are quite closely related to it, so I suppose we can open an exception as we did with `ls` and the wildcards (I don't have the PR reference handy, sorry). So my verdict is tentative merge :+1:

rprieto:
Looks great! Agreed with what @waldyrious said, so :+1: from me.

PR #503, opened 2015-12-30
make: add page

Kaligule:
added tldr page for make


ostera:
Sorry, this is just too vague and doesn't provide any helpful commands. Please, they aren't called "parts", they are called `targets`. And running `make` will run the first target (which by convention is named `all`).

And it still wouldn't provide any value then.

PR #505, opened 2015-12-30
defaults:  add page

sfarzy:
OS X command to read or modify user defaults.

### DIFF
--- a/pages/osx/defaults.md
+++ b/pages/osx/defaults.md
@@ -1,0 +1,19 @@
+# defaults
+
+> Access OS X user defaults
+
+- read all defaults
+
+`defaults read -app {{app-name}}`
+
+- read specific defaults
+
+`defaults read -app {{app-name}} {{key}}`
+
+- write key value
+
+`defaults write -app {{app-name}} {{key}} {{-type}} {{value}}`
+
+- delete all defaults
+
+`defaults write -app {{app-name}}`
--- a/pages/osx/defaults.md
+++ b/pages/osx/defaults.md
@@ -16,4 +16,4 @@
 
 - delete all defaults
 
-`defaults write -app {{app-name}}`
+`defaults delete -app {{app-name}}`
--- a/pages/osx/defaults.md
+++ b/pages/osx/defaults.md
@@ -2,18 +2,22 @@
 
 > Access OS X user defaults
 
-- read all defaults
+- read system default value
 
-`defaults read -app {{app-name}}`
+`defaults read {{domain}} {{key}}`
 
-- read specific defaults
+- read default values of applications
 
-`defaults read -app {{app-name}} {{key}}`
+`defaults read -app {{app_name}} {{key}}`
 
 - write key value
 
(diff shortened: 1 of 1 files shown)


pages/osx/defaults.md
```diff
@@ -0,0 +1,19 @@
...
+
+`defaults read -app {{app-name}} {{key}}`
+
+- write key value
+
+`defaults write -app {{app-name}} {{key}} {{-type}} {{value}}`
+
+- delete all defaults
+
+`defaults write -app {{app-name}}`
```
ostera: was this supposed to be `defaults delete -app {{app-name}}` ?

sfarzy: Yes, Missed that error. Thanks for pointing out.
Corrected now and update has been committed.

waldyrious:
Thanks, looks great already :) Just two minor nitpicks:
1. It would probably be better to write {{app_name}} as that's the format we're trying to standardize on in tldr, for separating words in tokens (unless the app names in OS X for this command indeed are specified in lowercase, hyphen-separated form).
2. How about adding a small note to the command that uses `{{-type}}` to explain what it is, e.g. by providing one or two example types?

pages/osx/defaults.md
```diff
@@ -0,0 +1,23 @@
+# defaults
+
+> Access OS X user defaults
+
+- read system default value
+
+`defaults read {{domain}} {{key}}`
+
+- read default values of applications
```
igorshubovych: should it be application, not applicationS, considering the fact that you are talking about some particular {{app_name}}?

waldyrious: I think this is just a language subtlety. The description is written in a generic sense, not literally referring to a concrete set of multiple applications. That said, it certainly doesn't hurt to be explicit and change the phrasing to refer to a single app, which is what the command actually does.

pages/osx/defaults.md
```diff
@@ -0,0 +1,23 @@
...
+
+`defaults read -app {{app_name}} {{key}}`
+
+- write key value
+
+`defaults write {{domain}} {{key}} {{-type}} {{value}}`
+
+- Speed up Mission Control animations
```
waldyrious: This is a very specific example which I feel doesn't fit with the purpose of tldr and with the rest of the examples on this page. I'd suggest dropping it.

igorshubovych:
@sfarzy do you need help with completing this?

sfarzy:
@igorshubovych Help is very much appreciated.
I was looking for a simple example.

igorshubovych:
@sfarzy here you can find a several minutes video about how to squash commits:
https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#submitting-a-pull-request (number 9)

pages/osx/defaults.md
igorshubovych: @waldyrious do we still need to remove it. i feel this example might be ok. in this case, maybe we just can merge it.

PR #506, opened 2015-12-30
nano: started tldr for nano

jlems:
Starting up tldr for nano.

### DIFF
--- a/pages/common/nano.md
+++ b/pages/common/nano.md
@@ -4,12 +4,12 @@
 
 - Start nano in terminal with {filename}
 
-`nano {filename}`
+`nano {{filename}}`
 
-- Save and then exit nano
+- Enable smooth scrolling.
 
-`ctrl-x`
+`nano -S {{filename}}`
 
-- Save file without exitting
+- Indent new lines to the previous lines' indentation
 
-`ctrl-o`
+`nano -i {{filename}}`
--- a/pages/common/nano.md
+++ b/pages/common/nano.md
@@ -1,0 +1,15 @@
+# nano
+
+> Simple, easy to use editor. An enhanced, free Pico clone.
+
+- Start nano in terminal with {filename}
+
+`nano {filename}`
+
+- Save and then exit nano
+
+`ctrl-x`
+
+- Save file without exitting
+
+`ctrl-o`
--- a/pages/common/nano.md
+++ b/pages/common/nano.md
@@ -10,6 +10,6 @@
 
 `nano -S {{filename}}`
 
-- Indent new lines to the previous lines' indentation
+- Indent new lines to the previous lines' indentation.
 
 `nano -i {{filename}}`


ostera:
:+1:

igorshubovych:
Finally an editor for humans.

But actually Ctrl+o writes it to disk: http://take.ms/0IHDU

Frankly speaking I don't think we should include key bindings on this page. Vi and Emacs pages need them, because there is no visual helper. Nano has it, so they are visible. I would rather add some command line options.

jlems:
@igorshubovych 
That's a good point. I can update it with some command line options. 

Thanks!

waldyrious:
I'd say we should still mention that the common commands are at the bottom of the screen. Btw if you want a CLI editor for *real* human beings, try [dit](https://github.com/hishamhm/dit), [sanos](http://www.jbox.dk/sanos/editor.htm) or [slap](https://github.com/slap-editor/slap) :smile:

igorshubovych:
so what is the resolution here?

waldyrious:
I propose mentioning in the main description that the commands commands are listed at the bottom of the screen.

notpeter:
I don't think we should be mentioning any keys used in nano, nor that there's help at the bottom of the screen.  Nano devotes two whole lines of your display in-app to that at all times. I vote for keeping it short and sweet.

85pando:
Acttually @notpeter I thought so too, so i recommended nano to my non-cli accustomed friend for work on the server and he was not even able to save the file because he could not understand the message about which line endings to use (which i never consciously noticed).

So if someone looks for info I'd rather say there should be a comment here about this…

@waldyrious What about vim (yes, vim is usable and is therefore for _real humans_, some even say emacs is an editor…).

waldyrious:
@85pando vim is only usable once you learn the basics (how to enter insert mode, how to exit). Before that it's [terrible UX](https://twitter.com/waldyrious/status/681839033245380608) IMO. Nano does way better than it, as do the editors I suggested above, except for more advanced stuff, obviously.

rubenvereecken:
@85pando raises a valid point for a newbie that might `tldr nano` to figure out how to save files. But then again.. it's a text editor. Everyone should have _a_ text editor they can at least open and save files in (nano was my first). It's not quite up to us to teach someone how to approach whatever text editor. Just tldr.

waldyrious:
Something simple like "Common commands are displayed in the bottom" seems mostly harmless IMO.

ostera:
I think this is good as is. If we still feel we have to add common commands, we can revisit. Merging now.

waldyrious:
Just for the record, I'm against cutting discussions off like this without consensus. No big deal, especially since the project does gain a new page which could remain in bikeshedding for a while, but I don't like the way this was resolved either. At least we could have agreed to disagree before a final decision was made.

igorshubovych:
I agree with @waldyrious, let's in future wait till we have no disagreement.

PR #509, opened 2015-12-30
Fixed tldr for gem

phalle5:
Specific version instruction missed gem name and mixed up -v and --version


igorshubovych:
Good one! Thanks!

PR #513, opened 2015-12-30
fix zpool

lizhuoli1126:
zpool: fix for `one description one command`

### DIFF
--- a/pages/common/zpool.md
+++ b/pages/common/zpool.md
@@ -14,9 +14,8 @@
 
 `zpool import`
 
-- Import a zpool, optionally specifying a new name
+- Import a zpool(optionally specifying a new name)
 
-`zpool import {{poolname}}`
 `zpool import {{poolname}} {{newpoolname}}`
 
 - Export a zpool (unmount all filesystems)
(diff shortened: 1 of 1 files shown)


waldyrious:
Nice one, thanks! Before merging, I'd suggest splitting the "Import a zpool, optionally specifying a new name" example into two.

lizhuoli1126:
@waldyrious I agree with you, but we can just use a bracket with `optionally specifying a new name` to show that it's an optional arg, I think it don't need one more separate line for this.

waldyrious:
The problem is that the example uses two parameters, so the description should match. The tldr project exists precisely to remove the burden of reasoning over ambiguous documentation, and provide simple, straightforward examples. @igorshubovych or @rprieto what are your thoughts?

rubenvereecken:
I'm quite sorry @lizhuoli1126 but this page has been the subject of quite some debate and I've recently merged a PR by @notpeter that affected this file in a way that just can't trivially be merged with your changes. I'm closing this commit for now because there's really nothing we can easily do - the best would be to reopen if things need changing.

PR #514, opened 2015-12-30
add axel

lizhuoli1126:
axel: add

### DIFF
--- a/pages/common/axel.md
+++ b/pages/common/axel.md
@@ -1,0 +1,24 @@
+# axel
+
+> Download accelerator
+> Supports HTTP, HTTPS, and FTP
+
+- Download a URL to a file
+
+`axel {{url}}`
+
+- Download and specify filename
+
+`axel {{url}} -o {{filename}}`
+
+- Download with multiple connections
+
+`axel -n {{connections_num}} {{url}}`
+
+- Search for mirrors
+
+`axel -S {{mirrors_num}} {{url}}`
+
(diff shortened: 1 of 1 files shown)


pages/common/axel.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Download and specify filename
+
+`axel {{url}} -o {{filename}}`
+
+- Download with multiple connections
+
+`axel -n {{connections_num}} {{url}}`
+
+- Search for mirrors (can specify numbers)
```
felixonmars: I think "(can specify numbers)" can be omitted, since there's `{{mirrors_num}}` in the example already.

G570972680893:
😎

pages/common/axel.md
```diff
@@ -0,0 +1,24 @@
...
+
+`axel -S {{mirrors_num}} {{url}}`
+
+- Limit download speed (bytes per second)
+
+`axel -s {{speed}} {{url}}`
```
G570972680893: 😎

pages/common/axel.md
lizhuoli1126: I means the default mirrors_num is 1 if you just use 'axel -n {{ur}}l'..But I think it may cause misunderstanding

rubenvereecken:
I'm merging this one in but the last example doesn't seem very necessary. If we ever need a different example, that one's the first to go :P.

Thanks!

Issue #478, opened 2015-12-30
Contribution guidelines: files and paths


rprieto:
It would be nice to come up with a standard across TLDR to have a consistent approach to `{{file}}`, `{{file.ext}}`, `{{path/to/file}}`, `{{/path/dir/file}}`, {{path_to_file}}`.

Personally I'd say `{{file}}` is enough for  80% of cases, and is generic enough that anyone with basic Unix knowledge knows you can specify a file name, relative path, absolute path.... or even the `<(command)` operator in bash. I can't think of a command that would specifically **need** a relative or absolute path.

We'd also need guidelines for acceptable exceptions, for example
- specifying the extension is acceptable if it matters to the command, e.g. `open {{file.ext}}`
- using other words like `source` or `target` if it clarifies the example, e.g. `tar cfz {{target.tar.gz}}`
- using `file1`, `file2`... if there are multiple files in the command, e.g. `tar cf {{target.tar}} {{file1}} {{file2}}`

igorshubovych:
Cc @rprieto

rprieto:
Why the cc?

waldyrious:
I mostly agree w/ @rprieto, but I'd add that making the paths explicit is useful at least for commands dealing with two locations, e.g, `cp`, `mv`, `scp`, etc, or with remote locations.

igorshubovych:
oh, sorry Romain. i thought somebody else opened it :)

rubenvereecken:
I wouldn't mind writing a draft is no one else is eager to, and then we can iterate on that. I followed most of the discussions related to this one so I've got a fairly recent overview, now it's about wording it concisely.

igorshubovych:
I would add that path can be used for specific folders:

``virtualenv {{path/to/venv}}``
`usermod -m -d {{/path/to/home}} {{user}}`

rprieto:
Very good point differentiating files/folders. To avoid the questions about absolute/relative, whether to start with a `/` etc... could we use `{{file}}` and `{{dir}}`, or sometimes being more specific like `{{home_dir}}`? It also allows us to make it very clear when a command needs a trailing slash, e.g. `somecommand {{source_dir}}/`

waldyrious:
What about commands that take either files or directories? We've discussed about the (ugly IMO) `{{file_or_dir}}` somewhere recently, and decided against it IIRC.

agnivade:
[...]

PR #516, opened 2015-12-30
ar: add examples and clean-up

eddieantonio:
Does notation like `{{libfoo.a}}` and `{{foo.o}}` seem cromulent? Or should it be something like `{{archive.a}}` `{{source1.o}}...`?

### DIFF
--- a/pages/common/ar.md
+++ b/pages/common/ar.md
@@ -2,10 +2,22 @@
 
 > Create, modify, and extract from archives (.a .so .o)
 
-- Extract members of the archive
+- Extract all members from an archive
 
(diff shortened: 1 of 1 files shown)


eddieantonio:
I'll be sure to update this as soon as more work on #478 is done.

waldyrious:
Thanks! I believe using `{{libfoo.a}}` and `{{foo.o}}` is more intuitive. I'd say we can merge this now and update later along with other pages depending on what is decided on #478, so :+1: from me.

igorshubovych:
:+1: Thank you!

PR #517, opened 2015-12-30
ulimit: added to Linux section

Cvetomird91:
Added a page for the ulimit shell builtin for getting and setting various user and shell limits.

### DIFF
--- a/pages/linux/ulimit.md
+++ b/pages/linux/ulimit.md
@@ -1,0 +1,19 @@
+# ulimit
+
+> Get and set user limits
+
+- get the properties of all the user limits
+
+`ulimit -a`
+
+- get hard limit for the number of simultaneously opened files 
+
+`ulimit -H -n`
+
+- get soft limit for the number of simultaneously opened files 
+
+`ulimit -S -n` 
+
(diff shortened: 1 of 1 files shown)


waldyrious:
Would it be possible to use words in the example descriptions that match the letters of the options? That helps memorizing and undestanding the commands.

Cvetomird91:
I have edited those descriptions accordingly and added some new options.

waldyrious:
I think you swapped "soft" and "hard" in the descriptions relative to the corresponding commands :)

Also, I believe "per-user" (with a hyphen) is the correct form.

waldyrious:
Btw, you may consider splitting `-Sn` and `-Hn` into `-S -n` and `-H -n` just to make it clear that they are separate options.

Cvetomird91:
Thank you for your feedback. This has been taken care of.

igorshubovych:
LGTM.
Actually all these commands work in OSX.

waldyrious:
If that's the case, then this page should be in the common folder. Other than that, this looks good to go :+1:

igorshubovych:
Thank you for contribution!

PR #519, opened 2015-12-30
last: added the command in common section

Cvetomird91:
A command to check the last logged in users.

### DIFF
--- a/pages/common/last.md
+++ b/pages/common/last.md
@@ -1,0 +1,27 @@
+# last 
+
+> View the last logged in users
+
+- view last logins, their duration  and other information as read from /var/log/wtmp
+
+`last`
+
+- specify how many of the last logins to show
+
+`last -n {{login_count}}`
+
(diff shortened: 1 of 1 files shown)


waldyrious:
The only example provided seems a bit at odds with that the main purpose of the command is. Could you provide a more typical example before that?

Also, please capitalize the descriptions (sentence case).

Cvetomird91:
I have added more examples and capitalized the description.

waldyrious:
Much nicer, thanks! As for capitalization, I meant that the first letter of each example should be uppercase, but [looking again at the contributing guide](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#markdown-format), it looks like I misremembered, so this looks :+1: to me as-is.

Cvetomird91:
Thank you for your feedback.

igorshubovych:
Thank you, @Cvetomird91 !

PR #520, opened 2015-12-30
removes no-longer-there diskutil verb

sdaitzman:
As of El Capitan, `sudo diskutil repairPermissions /` returns `diskutil: did not recognize verb "repairPermissions"; type "diskutil" for a list` since repairPermissions has been removed.

In the interest of keeping this tldr-compatible I've removed it. In the interest of not leaving things out, there could instead be a "this no longer works" notice.

### DIFF
--- a/pages/osx/diskutil.md
+++ b/pages/osx/diskutil.md
@@ -6,10 +6,6 @@
 
 `diskutil list`
 
-- Repair permissions on a volume
-
-`diskutil repairPermissions {{/Volumes/Name}}`
-
 - Repair the file system data structures of a volume
 
 `diskutil repairVolume {{/dev/diskX}}`


waldyrious:
:+1: for removing rather than adding a deprecation notice. Out of curiosity, what do you mean by "tldr-compatible"?

sdaitzman:
If it has only critical info, it’s tldr-compatible!

(which is a concept I made up halfway through the PR)

igorshubovych:
LGTM!

rubenvereecken:
:+1: for keeping our pages up to date!

felixonmars:
:+1:

waldyrious:
@sdaitzman oh, I see what you mean -- compatible with the ethos of tldr: conciseness. :)

PR #524, opened 2015-12-31
Shutdown at a specific time

ElectricSquid:
Added section to indicate how to shutdown at a specific time rather than shutting down X minutes from now


igorshubovych:
I like it. LGTM.

felixonmars:
:+1: Thanks :)

PR #525, opened 2015-12-31
Added the "dd" command to common

Parth-Mehrotra:


Parth-Mehrotra:
I just realized that `dd` exists in linux, but it should be in common, I'm going to remove my entry from common, and just move the one from linux to common.

PR #526, opened 2015-12-31
Add Dart client to README

hterkelsen:
I wrote a client for Dart and would love for it to be included in the README. Thanks!


waldyrious:
Thanks @hterkelsen :D

Btw, we're currently discussing internally about providing a set of specs for clients to follow, in order to preserve minimal functionality and compatibility; so we'll reach out to you as soon as that's settled.

hterkelsen:
Thanks for merging. I'll be on the lookout for the client specs. For what it's worth, I tried to make my client mimic the screenshot in the README

igorshubovych:
You don't need to mimic it. You can freely develop it, maybe overpassing reference client by performance, features, simplicity etc.

We basically wanted to warn that using Github API to retrieve the pages can cause denial of service (if your client become popular overnight). You should download static assets from [static site](http://tldr-pages.github.io/) instead.

PR #528, opened 2015-12-31
Fix interpunction, mention scripts in -A.

d33tah:


igorshubovych:
We don't put colons at end of command description.

d33tah:
@igorshubovych: this wasn't by your CI then. Changed to dots - wanted that to be uniform.

pages/common/nmap.md
```diff
@@ -11,18 +11,18 @@
...
 
 `nmap -A {{address_or_addresses}}`
 
-- Assume good network connection and speed up execution:
+- Assume good network connection and speed up execution.
 
 `nmap -T4 {{address_or_addresses}}`
 
-- Scan a specific list of ports (use -p- for all ports 1-65535)
+- Scan a specific list of ports (use -p- for all ports 1-65535).
```
igorshubovych: Don't use this trick to have +1 code example in parens.
If the example is common and useful add it separately.

pages/common/nmap.md
```diff
@@ -11,18 +11,18 @@
...
 
 `nmap -p {{port1,port2,...,portN}} {{address_or_addresses}}`
 
-- Perform TCP and UDP scanning (use -sU for UDP only, -sZ for SCTP, -sO for IP):
+- Perform TCP and UDP scanning (use -sU for UDP only, -sZ for SCTP, -sO for IP).
```
igorshubovych: Don't use this trick to have +1 code example in parens.
If the example is common and useful add it separately.

igorshubovych: It is better to break 5-6 examples rules, then write something like this.

rubenvereecken:
While @igorshubovych's comments are justified, the changes in question are actually [two days old](https://github.com/tldr-pages/tldr/commits/master/pages/common/nmap.md). I suggest we merge this in as it is and patch in fixes for those comments in a separate PR.

igorshubovych:
@rubenvereecken go for it

PR #530, opened 2015-12-31
add enca to common page

joelhy:
add file converting utility to common page.

### DIFF
--- a/pages/common/enca.md
+++ b/pages/common/enca.md
@@ -1,0 +1,19 @@
+# enca
+
+> Detect and convert encoding of text files
+
+- detect file(s) encoding according to your system's locale
+
+`enca {{file(s)}}`
+
+- detect file(s) encoding; -L option tells enca the current language; language is in the POSIX/C locale format, e.g. zh_CN, en_US etc.
+
+`enca -L {{language}} {{file(s)}}`
+
+- convert file(s) to specified encoding
+
+`enca -L {{language}} -x {{to_encoding}} {{file(s)}}`
+
+- save original_file as new_file and convert new_file to specified encoding
+
+`enca -L {{language}} -x {{to_encoding}} < {{original_file}} > {{new_file}}`


pages/common/enca.md
```diff
@@ -0,0 +1,19 @@
+# enca
+
+> Detect and convert encoding of text files
+
+- detect file encoding; -L option tell enca the current language; LANGUAGE as zh_CN, en_US etc.
```
waldyrious: - "tell" --> "tells"
- "as" --> "is in the POSIX/C locale format, e.g."

pages/common/enca.md
```diff
@@ -0,0 +1,19 @@
...
+
+`enca -L {{LANGUAGE}} file`
```
waldyrious: file --> {{file}}

pages/common/enca.md
```diff
@@ -0,0 +1,19 @@
...
+
+- convert file to specified encoding
+
+`enca -L {{LANGUAGE}} -x {{to_encoding}} file`
```
waldyrious: file --> {{file}}

waldyrious:
Nice one! Besides the inline comments, I have two general questions:
1. Couldn't {{LANGUAGE}} be lowercase?
2. Wouldn't it be useful to provide at least one example that doesn't use the -L option?

pages/common/enca.md
igorshubovych: I don't mind if encoding example remains.

waldyrious: Just to be clear, I didn't suggest to remove the examples, the only part I was suggesting to replace was the word "as" :)

joelhy: fixed :-)

igorshubovych:
If commits are squashed, then it LGTM. And it is good to go.

pages/common/enca.md
waldyrious: @joelhy only the "tell" --> "tells" was changed. Did you overlook the second suggestion, or do you disagree with the proposed wording?

joelhy:
Commits are squashed.

felixonmars:
Looks like it's one more commit instead of a squash :P

G570972680893:
😎

igorshubovych:
These commits seems to be from other branch:
http://take.ms/6nuin

joelhy:
Not quite familar with squash operation.
I just delete the enca-page branch and recreate it.
Is it feasible?

pages/common/enca.md
```diff
@@ -0,0 +1,23 @@
...
+
+`enca -L {{language}} {{file}}`
+
+- convert file to specified encoding
+
+`enca -L {{language}} -x {{to_encoding}} {{file}}`
+
+- convert multiple files to specified encoding
+
+`enca -L {{language}} -x {{to_encoding}} *`
```
felixonmars: The two examples "convert file to specified encoding" and "convert multiple files to specified encoding" could be merged into one, just say something like "convert file(s) to specified encoding" etc. `*` is a shell behavior and should not be in the examples, also "multiple files" doesn't always mean "all files (except hidden) in the current folder" :)

pages/common/enca.md
```diff
@@ -0,0 +1,23 @@
...
+
+- convert file to specified encoding and save to new file
+
+`enca -L {{language}} -x {{to_encoding}} {{original_file}} {{new_file}}`
```
felixonmars: This one is actually incorrect. I got `enca: Cannot open file `new_file' in mode r+b: No such file or directory` when there is only `original_file` in the current folder. Also the syntax described here is ambiguous with the last two examples.

pages/common/enca.md
joelhy: fixed by using your phrase  "convert file(s) to specified encoding"

pages/common/enca.md
joelhy: Fixed by adding `<` (less than sign) and `>` (greater  than sign).

joelhy:
Please note that 

    `enca -L {{language}} -x {{to_encoding}} <<> {{original_file}} <>> {{new_file}}`

will be output as 

`enca -L language -x to_encoding < original_file > new_file`

in shell (tested in bash and zsh shell) using [Node.js client](https://github.com/tldr-pages/tldr-node-client).

joelhy:
For the record,  `\<`, `&lt;`, `&#60;` does not render as `<`. So I hacked around, and found out that `<<>` work like charm.

felixonmars:
I think this is something that should be fixed in the client. What do you think @igorshubovych ?

pages/common/enca.md
```diff
@@ -0,0 +1,19 @@
...
+
+- detect file(s) encoding; -L option tells enca the current language; `language` is zh_CN, en_US etc.
+
+`enca -L {{language}} {{file(s)}}`
+
+- convert file(s) to specified encoding
+
+`enca -L {{language}} -x {{to_encoding}} {{file(s)}}`
+
+- save `original_file` as `new_file` and convert `new_file` to specified encoding
```
igorshubovych: Do not use backticks in descriptions (``) unless you mention command. It is ok to use it for `git`, `ls`, `cp`, but not for arguments.

pages/common/enca.md
```diff
@@ -0,0 +1,19 @@
...
+
+`enca -L {{language}} -x {{to_encoding}} <<> {{original_file}} <>> {{new_file}}`
```
igorshubovych: Do not optimize for client. Optimize for correctness.

pages/common/enca.md
waldyrious: @igorshubovych I tweaked the markdown in your comment, since two unescaped backticks were messing up the formatting of the rest of the sentence.

igorshubovych: @waldyrious thanks

joelhy:
Backticks in descriptions removed.

pages/common/enca.md
joelhy: Corrected. I overlooked, sorry for that.

joelhy:
Last commit change:
"is" => "is in the POSIX/C locale format, e.g.".

joelhy:
@waldyrious , @igorshubovych 
As stated in https://github.com/tldr-pages/tldr/pull/590#discussion_r48820085 : "Also maybe wrap "part_size" in code tags `` or the token syntax {{}} to make it explicit that it refers to the command parameter below."

I added code tags to `language` accroding to this, may be we can carify code tags usage in descriptions, and add that to [contributing guidelines](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md).

waldyrious:
@joelhy you're right, my bad. I've add that concern to the current discussion about formatting of the pages in #622.

As for this PR, LGTM :+1:

igorshubovych:
LGTM. Thanks!

PR #531, opened 2015-12-31
add convmv to common page

joelhy:
add filename converting utility convmv to common page


convmv.md
```diff
@@ -0,0 +1,7 @@
+# convmv
+
+> Converts filenames from one encoding to another
+
+- convert encoding without dry run
```
waldyrious: "without dry" run is confusing. The default invocation does a dry run in addition to the conversion, or instead of it? Maybe add an example above this one without the `--notest`, to better clarify the command's behavior.

joelhy: dry run example added

pages/common/convmv.md
```diff
@@ -0,0 +1,11 @@
+# convmv
+
+> Converts filenames from one encoding to another
+
+- convert encoding(dry run)
```
waldyrious: Needs a space before "(dry run)"

pages/common/convmv.md
```diff
@@ -0,0 +1,11 @@
...
+
+`convmv -f {{from_encoding}} -t {{to_encoding}} {{input_file}}`
+
+- convert encoding(not dry run)
```
waldyrious: Needs space here too

pages/common/convmv.md
joelhy: Space added. Without `--notest` the command just shows what it wants to do, since the operation could be destructive.

pages/common/convmv.md
joelhy: Space added.

waldyrious:
I think we can merge but I'll probably propose a different wording afterward, as the current one still strikes me as a bit awkward.

joelhy:
```bash
joelhy@archlinux: /home/joelhy/workspace/demo master ⚡ $ convmv -f cp936 -t utf-8 config.php  
Starting a dry run without changes...
No changes to your files done. Would have converted 0 files in 0 seconds.
Use --notest to finally rename the files.
```
This is an example command output.
I'll twist the comment a bit accroding this.

joelhy:
Committed, does this look better?

waldyrious:
Huh, weird. Does this convert the encoding *and* rename the files? Why?

joelhy:
Only change the filename encoding, **not** file content encoding.

convmv.md
G570972680893: 😎
1 янв. 2016 г. 21:23 пользователь "Waldir Pimenta" <notifications@github.com>
написал:

joelhy:
Different locales might encode filenames in different encodings, this cause some trouble when copying files. And this command fix the issue.

waldyrious:
Oooh, I missed that completely! I thought it dealt with file contents indeed. Maybe we could make that clearer in the main description, to prevent the same confusion in others :)

waldyrious:
Sorry for the delay, LGTM!

PR #535, opened 2015-12-31
afl-fuzz

d33tah:
In this example I show how to prepare a fuzzable binary and fuzz it.

### DIFF
--- a/pages/common/afl-fuzz.md
+++ b/pages/common/afl-fuzz.md
@@ -1,0 +1,16 @@
+# afl-fuzz
+
+> American fuzzy lop is a fuzzer that uses instrumentation and genetic algorithms.
+> After fuzzing, have a look at the "crashes" directory in {{output_directory}}.
+
+- Compile and instrument a "hello world" that crashes if you write "1".
+
+`echo {{"main(){ if (getchar() == '1') abort(); }"}} > {{hello.c}} ; afl-gcc {{hello.c}} -o {{executable}}`
+
+- Create a directory with a sample test case - one byte saying "2".
+
+`mkdir {{input_cases}} && echo -n 2 >> {{input_cases}}/{{example}}`
+
+- Perform fuzzing of the "hello world" from first step using the sample test case.
+
+`AFL_EXIT_WHEN_DONE=1 afl-fuzz -i {{input_cases}} -o {{output_directory}} -- {{./executable}}`


rubenvereecken:
While this file has certainly taught me something interesting I wouldn't otherwise know, it doesn't really adhere to the tldr spirit. I see how it shows a whole `afl-fuzz` workflow, but this isn't really what tldr aims to achieve. Rather, tldr is there as a really quick reference because you can't remember exactly how to invoke a tool. In that respect, the third example is what we're looking for, the fourth is irrelevant and the first two near a border case because they're really part of how `afl-fuzz` works. Still, they bloat the file while people might just be looking for a quick way to fire up `afl-fuzz`.

I know it feels bare, but I suggest you keep this file limited to only `afl-fuzz` commands. Even then, aim at the tldr structure of using something like

`AFL_EXIT_WHEN_DONE=1 afl-fuzz -i {{input_files}} -o {{output_directory}} -- {{executable}}`
instead of
`AFL_EXIT_WHEN_DONE=1 afl-fuzz -i input_cases -o output_directory -- ./hello`

We're currently working to refine guidelines on how to name those 'variables', but that's a start.

d33tah:
@rubenvereecken: I removed the final example and added parameters. Is it passable now?

waldyrious:
While I love the concept of afl-fuzz, I'm really torn on this one, in that the examples aren't really independent, but the whole page is a sequential manual for preparing an input for afl-fuzz and then actually run it.

I am not sure how we can address that. Could these two minimal examples perhaps be added to the [README](http://lcamtuf.coredump.cx/afl/README.txt), the [Quick start guide](http://lcamtuf.coredump.cx/afl/QuickStartGuide.txt) or even the [Wikipedia page for the tool](https://en.wikipedia.org/wiki/American_fuzzy_lop_(fuzzer)) (although the latter option [may be questionable](https://en.wikipedia.org/wiki/WP:NOTHOWTO)), and then linked to from this page?

d33tah:
@waldyrious reference: https://groups.google.com/d/msg/afl-users/YrZk5ri4nnA/H0FIJBM6BwAJ

waldyrious:
Yeah, that's definitely something that should be provided in the quick start guide IMO :) but for tldr it kinda deviates from the goal / format (collections of independent usage examples of the command in question, not a getting started guide). Do you know if the project is receptive to contributions in this form (i.e. adding this info to the gettign started guide)?

rubenvereecken:
Closing for now, sorry lads.

PR #536, opened 2015-12-31
Add several mkfs tools

85pando:
Tools to create files systems from the command line.

Im not sure with `mke2fs.md` because thats actually the tool, but you normally use the explicit versions like `mkfs.ext2` to actually create a filesystem (much like `git-commit`). Some feedback here would be welcome. The alternative would be to split it up into three single files with each virtually the same description.

### DIFF
--- a/pages/linux/mke2fs.md
+++ b/pages/linux/mke2fs.md
@@ -1,0 +1,15 @@
+# mke2fs
+
+> Creates a Linux filesystem inside a partition.
+
+- Create an ext2 filesystem in partition 1 of device b (`sdb1`):
+
+`mkfs.ext2 {{/dev/sdb1}}`
+
+- Create an ext3 filesystem in partition 1 of device b (`sdb1`):
+
+`mkfs.ext3 {{/dev/sdb1}}`
+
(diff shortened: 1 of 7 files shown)


sbrl:
Suggestion: mkfs.vfat

85pando:
(I will merge the changes when I get the go for this merge)

85pando:
Add colons instead of full stops to make `lint` happy.

Now `lint` is unhappy, because of full stops in title. The thing is that the tools do have full stops in their name/command.

igorshubovych:
@85pando we will work on this.

igorshubovych:
Linter is currently unhappy because we did not realize that name of command can contain dots.

85pando:
Yeah, probably a corner case, as not many tools will have this…

rubenvereecken:
Thanks for the heads up, will be incorporated very soon. I constricted the original case because more often than not something containg a period ends in one, you'd get titles like `Git.`

rubenvereecken:
Fixed in current master. Additionally, tldr will now use tldr-lint's master branch for Travis.

PR #321, opened 2015-11-08
print command

Qtrain:
fixed and deleted line in shell script that was giving me problems.  Let me know if you need changes.  thx.

PR #538, opened 2015-12-31
nm: add page

eddieantonio:
I did not use long options for compatibility between Apple and GNU `nm`. I limited examples to options that are available on both platforms.

### DIFF
--- a/pages/linux/nm.md
+++ b/pages/linux/nm.md
@@ -1,0 +1,19 @@
+# nm
+
+> List symbol names in object files
+
+- List global (extern) functions in a file (prefixed with T)
+
+`nm -g {{file.o}}`
+
+- Demangle C++ symbols (make them readable)
+
+`nm --demangle {{file.o}}`
+
+- List only undefined symbols in a file
+
+`nm -u {{file.o}}`
+
+- List all symbols, even debugging symbols
+
+`nm -a {{file.o}}`
--- a/pages/osx/nm.md
+++ b/pages/osx/nm.md
@@ -1,0 +1,15 @@
+# nm
+
+> List symbol names in object files (see c++filt)
+
+- List global (extern) functions in a file (prefixed with T)
+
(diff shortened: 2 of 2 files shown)


pages/common/nm.md
```diff
@@ -0,0 +1,19 @@
...
+
+> List symbol names in object files
+
+- List global (extern) functions in a file (prefixed with T)
+
+`nm -g {{file.o}}`
+
+- Demangle C++ symbols (make them readable)
+
+`nm {{file.o}} | c++filt`
```
igorshubovych: It is uncommon for us to have example of one command piped to another command.

eddieantonio: Unfortunately, this is the only cross-platform way (between GNU, BSD) of doing C++ name demangling (GNU does have the [`--demangle` option](http://man7.org/linux/man-pages/man1/nm.1.html#OPTIONS)). Since `nm` output is less than helpful for C++ object files without name demangling (and, personally, the whole reason I'd use a tldr page for `nm` anyway), `c++filt` should be mentioned in here somewhere. Should it be put in a "see also" in the description?

waldyrious: We do have a linux folder and are considering one for bsd (I suppose it would be ok if you created one yourself, /cc @rprieto), so the distinction could be done in that way instead of using the lowest common denominator in a common/ page.

waldyrious: Or you could add a small note in the description with the workaround, although that would probably be frowned upon by some maintainers (cf. #321)

eddieantonio:
I created two pages: one for Linux and OS X. In the OS X version, I added a little "see c++filt"; for GNU/Linux, I added `--demangle`. I stress that for the tl;dr page, listing how to demangle C++ symbols is invaluable to have in nm's page.

igorshubovych:
LGTM. Thanks.

PR #541, opened 2016-01-01
pkill: add page

kmoe:
There is already an example of `pkill` on the `pgrep` page but I assume we want users to be able to look things up directly wherever possible. For that reason I have created a standalone page for `pkill` with an additional example.

### DIFF
--- a/pages/common/pkill.md
+++ b/pages/common/pkill.md
@@ -1,0 +1,12 @@
+# pkill
+
+> Signal process by name
+> Mostly used for stopping processes
+
+- kill all processes which match
+
+`pkill -9 {{process_name}}`
+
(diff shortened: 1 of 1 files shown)


igorshubovych:
And there should not be `pkill` example in `pgrep` page

waldyrious:
:+1: for `pkill -9 {{process_name}}`

kmoe:
Amended. I can remove the `pkill` example from `pgrep` in this PR as well if you like, or open another one?

igorshubovych:
No, I guess, this is good to go now. Thanks! LGTM.

Please open new one on `pgrep`

waldyrious:
LGTM as well, thanks for the contribution!

PR #543, opened 2016-01-01
timedatectl: add page

maonx:
Add timedatectl page


waldyrious:
For future reference, this PR was followed-up by #543 (by the same author).

PR #545, opened 2016-01-01
ufw: add page

Torild:


waldyrious:
Thanks for your contribution! A suggestion: how about adding [the meaning of ufw](https://en.wikipedia.org/wiki/Uncomplicated_Firewall) in the main description? That would help people memorize the command.

Torild:
Sure thing. I changed the description a bit

pages/linux/ufw.md
```diff
@@ -0,0 +1,23 @@
+## ufw
+
+> Ufw (Uncomplicated Firewall) is a frontend for iptables aiming to make configuration of a firewall easier
```
igorshubovych: Can be simplified

```markdown
> Uncomplicated Firewall
> frontend for iptables aiming to make configuration of a firewall easier
```

pages/linux/ufw.md
```diff
@@ -0,0 +1,23 @@
...
+
+`ufw enable`
+
+- Disable ufw
+
+`ufw disable`
+
+- Add ufw allow rule
+
+`ufw allow {{port or service name}}`
```
igorshubovych: Maybe split {{port or service name}}` into 2 examples.

waldyrious:
Are the last two examples technically correct? The first version of this PR suggested that the parameter could be either a port or a service name, not both. The new version suggests both can be passed. So what's the correct invocation?

Torild:
It can be either, not both

G570972680893:
А люди где?
2 янв. 2016 г. 16:30 пользователь "Torild" <notifications@github.com>
написал:

waldyrious:
In that case I suppose the example should be split it two to prevent confusion.

pages/linux/ufw.md
```diff
@@ -0,0 +1,24 @@
...
+
+`ufw disable`
+
+- Add ufw allow rule
+
+`ufw allow {{port}} {{service name}}`
+
+- Add ufw deny rule
+
+`ufw deny {{port}} {{service name}}`
```
igorshubovych: We earlier agreed to have argument names without spaces, so it should be
{{service_name}}

igorshubovych:
Check out linter. It says:

pages/linux/ufw.md:1: MD002 First header should be a h1 header
pages/linux/ufw.md:3: MD009 Trailing spaces

igorshubovych:
Also please squash commits.

waldyrious:
Don't forget to split the last two examples to make sure people don't get the impression they can pass both parameters simultaneously.

igorshubovych:
@Torild the page already looks good.
Could you please squash these 6 commits into one?
http://take.ms/uihhF

Here is the instruction on how to do it:
https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#submitting-a-pull-request

Torild:
I squashed the commits and pushed them to my fork, but I'm not quite sure how to get them to show up here?

igorshubovych:
You have to do 

`git push --force`.

**IMPORTANT**
But be careful, check log (`git log`) of your branch, it should contain only one commit on top of checkout commit (since you did `git checkout -b`)

rubenvereecken:
Good stuff, thanks!

waldyrious:
@rubenvereecken how do we address the `{{port}}` vs `{{service_name}}` issue raised above, without exceeding the line count / max. number of examples? Use one for allow and another for deny? Would that be clear enough?

rubenvereecken:
You're right for raising the issue again @waldyrious, annoying I merged that in. We can do like you suggest, it's a bit confusing but it'll work. Too bad we don't use concrete examples anymore, otherwise we could have used "80" and "http" or something, well-known ports that make it easy to grasp for anyone.

waldyrious:
Perhaps if the descriptions mention the extra option, people can infer from the similarity how it can be composed otherwise.

PR #547, opened 2016-01-02
Add ssh-copy-id page

jdecool:


igorshubovych:
LGTM!

waldyrious:
So many cool and convenient commands out there! Thanks, LGTM as well :)

igorshubovych:
@waldyrious it is only Linux. But I installed this one with brew, which contains OpenBSD version:

@jdecool thank you for your contribution!

PR #550, opened 2016-01-02
Add imagemagick montage command to common

sbrl:
I'm always using imagemagick's [montage](https://www.imagemagick.org/Usage/montage/) and I keep forgetting how to use it. This pull request adds the montage command to tldr, along with a set of commands that I think are the most common / useful.

Have I created the 'feature-branch' correctly by the way? I had a ton of trouble with it.

### DIFF
--- a/pages/common/montage.md
+++ b/pages/common/montage.md
@@ -1,0 +1,24 @@
+# montage
+
+> Imagemagick image montage tool.
+> Tiles images into a customisable grid.
+
+- Tile images into a grid, automatically resizing images larger than the grid cell size:
+
+`montage {{image1.png}} {{image2.jpg}} {{imageN.png}} montage.jpg`
+
+- Tile images into a grid, automatically calculating the grid cell size from the largest image:
+
+`montage {{image1.png}} {{image2.jpg}} {{imageN.png}} -geometry +0+0 montage.jpg`
+
+- Set the grid cell size and resize images to fit before tiling:
+
+`montage {{image1.png}} {{image2.jpg}} {{imageN.png}} -geometry 640x480+0+0 montage.jpg`
+
+- Limit the number of rows and columns in the grid, causing input images to overflow into multiple output montages:
+
+`montage {{image1.png}} {{image2.jpg}} {{imageN.png}} -geometry +0+0 -tile 2x3 montage_%d.jpg`
+
+- Resize and crop images to completely fill their grid cells before tiling:
+
+`montage {{image1.png}} {{image2.jpg}} {{imageN.png}} -geometry +0+0 -resize 640x480^ -gravity center -crop 640x480+0+0 montage.jpg`


pages/common/montage.md
```diff
@@ -0,0 +1,20 @@
+# montage
+
+> Imagemagick image montage tool
+> Tiles images into a customisable grid
+
+- Tile images into grid without resizing
+
+`montage {{image1.png}} {{image2.jpg}} {{imagen.png}} -geometry +0+0 montage.jpg`
```
igorshubovych: What are the parameters of this grid?

sbrl:
@igorshubovych I don't understand the question. That command simply takes the input images, and tiles them into a grid whose size should be *loosely* square shaped, but there's no way to tell.

On another note, I realise now that I forgot to add an example which uses `-geometry` to add a border around each image. I don't know if it should be included though.

igorshubovych:
I just did not understand what +0+0 means.
Now after reading these links, I think I understand it:
- http://www.imagemagick.org/script/montage.php
- http://www.imagemagick.org/Usage/montage/#montage

It is a very complicated command. Too many options.

sbrl:
Do you mean the first one?

Unfortunately Imagemagick's montage command is inherently complicated - that's why I wrote this tldr page.

The `+0+0` simply prevents Imagemagick from resizing any images. A full geometry spec would be `640x480+0+0`, which means something like this:

> A width of 640, a height of 480, and no white space around the edges.

I can't see any way of simplifying the command. If you remove the `-geometry` bit, it resizes images to some unknown default size, which isn't very helpful.

Suggestions are welcome.

igorshubovych:
@sbrl i also found this command very complicated

sbrl:
@igorshubovych How can I simplify it?

waldyrious:
I don't think we can simplify it much further. I can only suggest you to add extra details to the command descriptions -- e.g. explain that in the first couple examples the grid subdivisions are automatically derived, or that the `-geometry` is what does the resizing. In fact, it could help to provide an example without the `-geometry` and then explain the difference in the one with it. Also, I'd prefer if you used "imageN" rather than "imagen", to make it clear the N isn't part of the word.

sbrl:
@waldyrious Done. Is that ok?

I was wondering about adding an example about borders that explains what the `+0+0` is for, but I'm not sure about it.

igorshubovych:
It looks ok. Could you plz squash the commits?

sbrl:
@igorshubovych I have a little bit of a problem - I accidentally hit the 'merge master branch' button on the github website, and now I have a bunch of extra commits on my history. Unless I can remove them, I don't think that I'll be able to squash :(

igorshubovych:
It is relatively simple to remove those commits from branch at all. Check out 7 mins video from this link: 
https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#submitting-a-pull-request

sbrl:
@igorshubovych Done! I've learnt something (several things actually) doing that. Apparently if you delete a line whilst rebasing it removes that commit.

Anyway, I've squashed my commits down so it's all ready for merging.

I've also tweaked it slightly to make it pass all the tests.

pages/common/montage.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Tile images into a grid, automatically calculating the grid size from the largest image:
+
+`montage {{image1.png}} {{image2.jpg}} {{imageN.png}} -geometry +0+0 montage.jpg`
+
+- Set the grid size and resize images to fit before tiling:
+
+`montage {{image1.png}} {{image2.jpg}} {{imageN.png}} -geometry 640x480+0+0 montage.jpg`
+
+- Limit size of grid, causing input images to overflow into multiple output images:
```
waldyrious: Ok, if I'm understanding this correctly, so far you seem to have used "size" to mean the dimensions of images or the grid; but now you're talking about the number of rows and columns, right? Or am I reading this wrong? Either way, I believe we should remove this ambiguity before merging. I'm sorry not to have caught this earlier :/

sbrl:
@waldyrious Thanks for the comments. I've tried to reword it again. Does that aid clarity?

pages/common/montage.md
```diff
@@ -0,0 +1,24 @@
+# montage
+
+> Imagemagick image montage tool.
+> Tiles images into a customisable grid.
+
+- Tile images into a grid, automatically resizing images larger than the grid size:
```
waldyrious: "than the grid cell size", to match the rest of the descriptions?

pages/common/montage.md
```diff
@@ -0,0 +1,24 @@
...
+
+`montage {{image1.png}} {{image2.jpg}} {{imageN.png}} montage.jpg`
+
+- Tile images into a grid, automatically calculating the grid cell size from the largest image:
+
+`montage {{image1.png}} {{image2.jpg}} {{imageN.png}} -geometry +0+0 montage.jpg`
+
+- Set the grid cell size and resize images to fit inside their grid cells before tiling:
```
waldyrious: "to fit it" would be shorter and still understandable, I believe.

pages/common/montage.md
```diff
@@ -0,0 +1,24 @@
...
+
+`montage {{image1.png}} {{image2.jpg}} {{imageN.png}} -geometry 640x480+0+0 montage.jpg`
+
+- Limit the number of row and columns in the grid, causing input images to overflow into multiple output montages:
```
waldyrious: "rows" (to match "columns")

pages/common/montage.md
```diff
@@ -0,0 +1,24 @@
...
+
+`montage {{image1.png}} {{image2.jpg}} {{imageN.png}} -geometry +0+0 -tile 2x3 montage_%d.jpg`
+
+- Resize and crop images to completely fill their grid cells before tiling:
```
waldyrious: I don't think this could be much clearer :) We could try to explain what `-gravity` means here, but the description would get too verbose.

igorshubovych:
@sbrl you can make changes suggested by @waldyrious, and the do 
```sh
git commit -a --amend
git push --force
```

sbrl:
@waldyrious @igorshubovych Done.

igorshubovych:
Thank you!

sbrl:
@igorshubovych No problem! Thanks for all your help too!

waldyrious:
:+1: excellent work @sbrl

sbrl:
@waldyrious Thanks!

PR #553, opened 2016-01-02
useradd, userdel, usermod: move from common to linux directory

kmoe:
It's my understanding that `useradd`, `userdel`, and `usermod` aren't available on OSX, so I've moved them to the linux directory.


igorshubovych:
You are right! Thanks for catching!
http://superuser.com/questions/320649/groupadd-useradd-not-found-on-mac-osx

PR #554, opened 2016-01-02
head's args aren't in bytes

Parth-Mehrotra:


pages/linux/head.md
```diff
@@ -8,7 +8,7 @@
 
 - Output the first few bytes of a file
 
-`head -c {{size in kilobytes}} {{filename}}`
+`head -c {{size in bytes}} {{filename}}`
```
igorshubovych: Interesting. I would rather put something like
{{bytes_size}}

waldyrious: That wouldn't be as intuitive, IMO. But the token should indeed use underscores as word separators, @Parth-Mehrotra, so I'd suggest `{{size_in_bytes}}`.

Parth-Mehrotra:
Okay @waldyrious @igorshubovych, just replaced all the spaces with underscores

igorshubovych:
Please squash commits.

In future you don't need to squash commits if you do
```shell
git add .
git commit --amend
git push --force
```
This way you are updating your commit and overwriting it in published branch.

Parth-Mehrotra:
Squashed, and updated @igorshubovych

felixonmars:
There are still 2 commits atm, and it should be one.

Parth-Mehrotra:
Good? Or do I need to update this branch as well? @felixonmars

waldyrious:
I now see only one commit, but the diff shows no underscores...

Parth-Mehrotra:
Okay, sorry, never needed to squash commits before and I'm apparently really bad at it haha

felixonmars:
It looks good now, however it shows "This branch is out-of-date with the base branch" and I cannot merge.

waldyrious:
I also see a merge conflict message, but apparently @igorshubovych sees something different in some cases (still a mystery to me how that can happen), so hopefully he can merge.

PR #400, opened 2015-12-27
add dd.md

WANG-lp:
add `dd.md` to osx and linux

PR #555, opened 2016-01-02
Moved dd.md from linux into common

Parth-Mehrotra:
these two files are identical, the only difference being the `bs=4m` vs `bs=4M`, which are essentially the same thing, moved into common


waldyrious:
I did [pose this question](https://github.com/tldr-pages/tldr/pull/400#issuecomment-167420374) when these two pages were added, but indeed didn't get a satisfying answer. Let's hope this issue gets settled better in this discussion. Thanks for raising the topic :+1:

felixonmars:
I don't have an OS X to test with. But I only agree to the merge if someone can confirm that the capital "4M" works also on OS X, since I've tested with the lowercase one and it doesn't work here on Linux.

waldyrious:
Hm, according to [this page](http://www.ubuntu.com/download/desktop/create-a-usb-stick-on-mac-osx), the casing is indeed significant:

> Execute `sudo dd if=/path/to/downloaded.img of=/dev/rdiskN bs=1m`
> (...)
> If you see the error `dd: Invalid number '1m'`, you are using GNU dd.
> Use the same command but **replace bs=1m with bs=1M**

(emphasis mine)

Parth-Mehrotra:
Oh you guys are correct, the uppercase `M` does not work, I'll close this pull request

PR #557, opened 2016-01-02
cowsay: add page

fluxw42:


pages/common/cowsay.md
```diff
@@ -0,0 +1,15 @@
...
+
+> Generate an ASCII character like a cow or sheep saying or thinking something
+
+- Print an ASCII cow saying "Hello world!"
+
+`cowsay {{"Hello world!"}}`
+
+- Print an ASCII dragon telling you what date it is
+
+`date | cowsay -f {{dragon}}`
```
igorshubovych: Just 

```markdown
`date | cowsay -f dragon`
```

igorshubovych: We also agreed that we try to avoid piping other commands except the most basic ones (`echo` and `ls`). So in your case it should be

```md
`echo "Hello" | cowsay -f dragon`
```

pages/common/cowsay.md
```diff
@@ -0,0 +1,15 @@
+# cowsay
+
+> Generate an ASCII character like a cow or sheep saying or thinking something
+
+- Print an ASCII cow saying "Hello world!"
+
+`cowsay {{"Hello world!"}}`
```
igorshubovych: In this case it should be one of two:

```md
`cowsay {{message}}`
```
or
```md
`cowsay "Hello world!"
```

pages/common/cowsay.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Print an ASCII dragon saying "Hello!"
+
+`echo "Hello!" | cowsay -f dragon`
+
+- Print a stoned thinking ASCII cow
+
+`cowthink -s {{"I'm just a cow, not a great thinker ..."}}`
```
igorshubovych: Same here.
We use {{}} - for token (named arguments).

waldyrious: Maybe we should call them placeholders rather than tokens (it's more self-explanatory and intuitive for non-native English speakers)

igorshubovych: Yes, tokens are not very suitable word.
cc @rprieto

waldyrious:
Well, I would have preferred the generic `{{message}}` rather than the specific "Hello world!", but since we were the ones suggesting either option was fine, I'm OK with merging this.

igorshubovych:
Thank you for your contribution!

Issue #558, opened 2016-01-02
Looks a lot like bro pages...


michaeleisel:
And yet doesn't give any credit. I'm not associated in any way with bro pages, but this is unfair. Why does this repo exist?

waldyrious:
According to the commit history tldr was created earlier ([dec 2013](https://github.com/tldr-pages/tldr/graphs/contributors)) than bropages ([jan 2014](https://github.com/hubsmoke/bro/graphs/contributors)). So if anything, bropages "should" give credit to tldr (by your logic), although I don't believe that's how attribution works -- they seem to have been two [independent inventions](https://en.wikipedia.org/wiki/Multiple_discovery).

In any case, there have been calls for collaboration between the two initiatives (#266), but they have been met with little to no reaction from the bropages side (see https://github.com/hubsmoke/bro/issues/57). So I'd suggest closing this issue, either as misguided/incorrect, or as a duplicate of #266.

michaeleisel:
Eh, fair enough

Issue #488, opened 2015-12-30
They are not man pages.


RalphCorderoy:
"Simplified and community-driven man pages" is misleading.  They are not man pages in many ways;  they're not roff source, man(1) cannot format them, "man -k" doesn't search them, etc.  bro says their content is a supplement to man pages.  tdlr seem to be usage examples;  perhaps that phrase can be used instead of man pages.

sdaitzman:
I think that phrasing is okay. The project should not undersell itself. It's simpler than manpages (therefore simplified), it's community-driven (accepts PRs) and it's a manual system that lets you learn things (therefore man pages).

ostera:
@sdaitzman I couldn't have said it better. Thanks anyways @RalphCorderoy for taking the time to open this issue!

Issue #266, opened 2015-03-02
Collaborate with bropages/cheat?


waldyrious:
**(Note for context: this issue was created with the title "Collaborate with bropages?")**

It's quite unfortunate that there are two projects splitting the mindshare of easily approachable, community-maintained quick reference documents for console commands: tldr and [bropages](http://bropages.org).

I didn't find signs of any past discussion, so I'd thought I'd mention the possibility and call the maintainers @rprieto and @hubsmoke (and all interested parties, of course) to go over the possibilities.

To kick things off, it might be worth discussing:
- the merits of the friendlier submission mechanism of bropages and the possibilities of tldr adopting such a workflow
- the many-competing-versions model of bropages versus the more wiki-like single-canonical-document model of tldr
- ~~the possibility of developing a Ruby-based [tldr client](https://github.com/tldr-pages/tldr#clients)~~ (done since #288)
- whether a web-based client like [tldr.js](https://github.com/leostera/tldr.js) connecting to both databases could be a nice first step towards greater integration

What do you guys think?

leostera:
**Warning: Off the top of my head.**

> the merits of the friendlier submission mechanism of bropages and the possibilities of tldr adopting such a workflow

While the cli is friendlier than submitting PR's and allows everyone to collaborate easily, `tldr` wouldn't be able to guarantee that the pages are actually useful. I fear this: https://github.com/hubsmoke/bro/issues/47 and this https://github.com/hubsmoke/bro/issues/23 – this are features that would need to be developed that otherwise are available already in form of PR's.

And a bit of trouble goes a long way in keeping trolls away.

*Idea: Perhaps we could anonymously track the commands that are being searched, and thus easily know which ones people want but are not yet available? This way we could suggest which pages people could collaborate with, and even have them sorted by priority.*

> whether a web-based client like tldr.js connecting to both databases could be a nice first step towards greater integration

This I could, but I'd rather do not. Mostly because of the format of their documents. Sample: http://bropages.org/man.json – I'm not saying it can't be worked with, I'm saying it's heavily redundant.

But if tldr-pages was to entertain the idea of one API that would make client-side page-caching easy, one API to rebuild the pages index by a mere GitHub Webhook, one API to rule all clients and to HTTP bind them...

Then I'd take that API to Mordor.

waldyrious:
[...]

PR #561, opened 2016-01-03
Mention cheat and bro as similar projects in README.md

igorshubovych:
We are often asked to mention or collaborate with bro pages or cheat:
- #266
- #558
- #488 
- and more

Let's at least mention those projects in README.md as it is done in cheat.


felixonmars:
:+1:

PR #562, opened 2016-01-03
lsof: add page

mumumu:


ostera:
There already seems to be a page for lsof [here](https://github.com/tldr-pages/tldr/blob/55a3c44ea8be9aeb17fca01a798d78180ef46c9f/pages/common/lsof.md), why not merge your changes with that one?

mumumu:
> There already seems to be a page for lsof here

Oh, thank you for your notification. I'll create another PR to merge that page.

PR #564, opened 2016-01-03
lsof: add -u, -c example and note about privileges

mumumu:


ostera:
Thanks! While you are at it, would you mind replacing the `{{8080}}` for `{{port}}` and removing the pipe to `kill -9` ? So the page conforms to the format and conventions as you've followed in your additions.

pages/common/lsof.md
```diff
@@ -1,6 +1,7 @@
 # lsof
 
 > Lists open files and the corresponding processes
+> Note: In most case, you need root privilege (or sudo) because you want to list files opened by other than you.
```
waldyrious: Here it should be "In most case**s**".

pages/common/lsof.md
```diff
@@ -13,3 +14,11 @@
 - only output the process PID (e.g. to pipe into kill)
 
 `lsof -t {{/path/to/file}} | xargs kill -9`
+
+- list files opened by user
```
waldyrious: "by user" --> "by the given user"

pages/common/lsof.md
```diff
@@ -13,3 +14,11 @@
...
+
+`lsof -u {{username}}`
+
+- list files opened by command (or process) name (e.x nginx)
```
waldyrious: Suggestion: "list files opened by the given command or process"

mumumu:
> Thanks! While you are at it, would you mind replacing the {{8080}} for {{port}} and removing the pipe to kill -9 ? 

OK. fixed.

pages/common/lsof.md
```diff
@@ -8,8 +9,16 @@
...
 - only output the process PID (e.g. to pipe into kill)
 
-`lsof -t {{/path/to/file}} | xargs kill -9`
+`lsof -t {{/path/to/file}}`
+
+- list files opened by the given user
+
+`lsof -u {{username}}`
+
+- list files opened by the given command or process (e.x nginx)
```
waldyrious: I don't think the "(e.x nginx)" is necessary here.

pages/common/lsof.md
```diff
@@ -8,8 +9,16 @@
 
 - find the process that opened a local internet port
 
-`lsof -i :{{8080}}`
+`lsof -i :{{port}}`
 
 - only output the process PID (e.g. to pipe into kill)
 
-`lsof -t {{/path/to/file}} | xargs kill -9`
+`lsof -t {{/path/to/file}}`
```
ostera: While you forgot to remove the comment in line 14, `(e.g. to pipe into kill)`, it looks quite natural. I'd say it can stay but I'll let some of the other @maintainers decide.

mumumu:
> I don't think the "(e.x nginx)" is necessary here.
> ...
> While you forgot to remove the comment in line 14, (e.g. to pipe into kill)

OK. removed.

waldyrious:
LGTM :)

ostera:
Agh, my bad. We should have squashed this before merging :(

igorshubovych:
@ostera in the similar case in future, if you notice it in the several minutes after it, you can do a revert.
http://take.ms/zxsw7

waldyrious:
Doesn't the revert functionality just create another commit with the inverse diff?

ostera:
@waldir: yes that's why I didn't revert it!

PR #479, opened 2015-12-30
arp: add page

Misch-:
First time I've contributed to open source, starting small ;)

Not sure about putting the ip link command in since it isn't an arp command, but it's nice for flushing the table and I believe it's pretty standard. Windows and OSX seem to have different syntax for their arp command.

PR #565, opened 2016-01-03
arp: add page

nevsan:


igorshubovych:
See #479

rubenvereecken:
Sup @nevsan, we just merged #479 which introduces `arp` first. Do you want to have a look if you can merge your results? You can always ask if we can help you out.

rubenvereecken:
I see some value in this page (for example, clearing the cache) but for the sake of not dragging this on for too long, I'll close this for now since we can't trivially merge this with the page we already have.

I'd be very happy to review this merged with the other page.

PR #566, opened 2016-01-03
Add another Rust client to the list

dbrgn:
@rilut beat me to it :) Here's another Rust client that aims to support
offline caching, the way the NodeJS client does it.

Maybe we can merge the two clients somehow (rilut/rust-tldr#1),
otherwise they can coexist as they have a different focus: One fetches
the pages from the net on each command invocation, the other does local
caching.


waldyrious:
It would be much preferred if the efforts could be merged, indeed, especially since now a third Rust client has emerged per the issue linked.

dbrgn:
@waldyrious we're working on that :)

waldyrious:
@dbrgn awesome, looking forward to hear about the results :+1:

PR #567, opened 2016-01-04
ssh: update ssh.md

antoniom13:
ssh: connect via proxy

ssh: copy ssh key for password-less login without using ssh-copy-id

ssh: execute list of commands from txt file

### DIFF
--- a/pages/common/ssh.md
+++ b/pages/common/ssh.md
@@ -30,3 +30,15 @@
 - ssh enable agent forward
 
 `ssh -A {{username}}@{{remote_host}}`
+
+- use a jump box as a proxy to log in to a server (not accessible from a public subnet)
+
+`ssh -t {{proxy_host}} ssh {{username}}@{{remote_host}}`
+
+- copy your ssh public key to a server for password-less login without ssh-copy-id
+
+`cat ~/.ssh/id_rsa.pub | ssh {{username}}@{{remote_host}} "mkdir ~/.ssh; cat >> ~/.ssh/authorized_keys"`
+
+- execute list of cmds in a cmd.txt file over ssh
+
+```ssh {{username}}@{{remote_host}} {{user}} "`cat cmd.txt`"```


igorshubovych:
All of these commands look more tips-n-tricks from cheat, than common use cases in tldr.

waldyrious:
I have to agree with @igorshubovych, but I'm open to consider arguments supporting these cases.

antoniom13:
Connecting via proxy is a common way to connect to boxes that do not have a public interface.  More commonly when connecting to self-hosted database servers, or when connecting to a server that is on a private subnet (i.e. AWS VPC, Private Corporate or Educational Networks, etc...  ).  Unless you feel or believe all resources have (or should have) a public interfaces :)

`ssh-copy-id` is not distributed as a standard binary in some cases.  This is a common way to copy your ssh public key with older distros like FreeBSD. I'd also like to point out that `ssh-copy-id` is not part of the standard OS X distribution, so if you don't have `brew` installed or install `ssh-copy-id` from source, this is a common way of adding it.

I would agree the execute list is more on the side of tips-n-tricks.

igorshubovych:
cc @rprieto

notpeter:
Although jump hosts are quite common, `ssh -t {{proxy_host}} ssh {{username}}@{{remote_host}}` doesn't support public key.  For this to work, the second host has to support password authentication (or both your local host and proxy_host have agent forwarding enabled). It also introduces a flag `-t` with no description.   The commonly accepted successor to this style of jump host used [ProxyCommand and netcat](https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Proxies_and_Jump_Hosts#ProxyCommand_with_Netcat) and works great with systems that only support public key (no password-interactive). 

I totally appreciate the utility a pre-req free example for appending a public key into a remote .ssh/authorized_keys, but this example is actually insufficient for it to work with default umasks. Specifically the  .ssh dir needs to be 700 and .ssh/authorized_keys needs to be 600 or else modern SSH versions will exit citing unsafe permissions.  Even if it had the missing `chmod`s I think it's doing too much for a simple example.

> Execute list of cmds in a cmd.txt file over ssh":

```ssh {{username}}@{{remote_host}} {{user}} "`cat cmd.txt`"```

This is a crafty solution to a common problem (run two commands remotely via ssh) but we're trying to shy away from examples which are merely just shellisms (e.g. cat in backticks).  Additionally, this is not a self contained example (requires user to create a textfile with a list of commands).  And even then, I'm not sure what that spurious `{{user}}` is doing there.

File: pages/common/ssh.md [hunk 1]
```diff
@@ -30,3 +30,15 @@
...
 
 `ssh -A {{username}}@{{remote_host}}`
+
+- use a jump box as a proxy to log in to a server (not accessible from a public subnet)
+
+`ssh -t {{proxy_host}} ssh {{username}}@{{remote_host}}`
+
+- copy your ssh public key to a server for password-less login without ssh-copy-id
+
+`cat ~/.ssh/id_rsa.pub | ssh {{username}}@{{remote_host}} "mkdir ~/.ssh; cat >> ~/.ssh/authorized_keys"`
```
ostera: this looks way too specific, I'd take it out.

igorshubovych:
@antoniom13 

After some thinking I tend to agree with @notpeter and @waldyrious. The existing `curl` page is already quite long 6 examples, and none of its examples is less common those added by you.

Anyway we thank you for your contribution. May the force be with you.

PR #570, opened 2016-01-04
mailx: add page

marksylee:
mailx is a Unix utility program for sending and receiving mail


waldyrious:
For future reference: this PR was recreated by the author as #572

PR #432, opened 2015-12-28
Added command for testing server configuration without affecting the …

joearcher:
…running server, invaluable for anyone new to using Nginx, when making configuration changes on a "live" server

PR #573, opened 2016-01-04
nginx: add -t example

joearcher:
Added the syntax for testing nginx configuration without affecting the running server, invaluable for people new to nginx when making configuration changes on a 'live' server.


igorshubovych:
As this was already discussed in #432 and was previously approved, I am merging this.

@joearcher thank you for taking an initiative to complete it.

Issue #574, opened 2016-01-04
Remove the index.json


lord63:
In issue #336, we talked about whether we need it or not. Finally, we removed the build. The latest index.json is in https://tldr-pages.github.io/assets/index.json
In issue #487, @ostera [said](https://github.com/tldr-pages/tldr/issues/487#issuecomment-168048513) that we may remove the index.json.

I'm also trying to make [tldr.py](https://github.com/lord63/tldr.py) support rebuilding the index, but one more thing here. The old index.json is still here, when we want to update the pages via `git pull` after we rebuilding the index.json(add some custom new pages, the index.json has changed), it fails because of the uncommited changes.

Since it is out of time and may cause some problems, it's time that we should remove the index.json now?

lord63:
Note there may be some compatible problems for those clients who rely on the `index.json`.

For example, [tldr.py](https://github.com/lord63/tldr.py) use local file and fetch latest pages via `git pull`. If we remove it from the main repo, users who fetch the latest page will can't use `tldr.py` since it can't find the `index.json`. Although I can release a new version to support reindex, some old users that installed it before will run into this problem if they update the pages.

And the update problem for `tlrdr.py` can be resolved by commit the changed `index.json`. But you have to commit it before you run the update command every time.

I still think the `index.json` is a good idea. But I can't come out a good way to solve the problem: make it fresh while easy for collaborating with others.

ostera:
@lord63 we have an up-to-date index located at http://tldr-pages.github.io/assets/index.json

We should remove the index from this repository as soon as the biggest clients migrate over to the new one. 

@igorshubovych perhaps can make a list of the clients here and we can start checking them off as they update, and then use this issue to submit a PR to remove the index.

lord63:
@ostera  Yes, I know, as I've point this out in my comment. And I agree that we should hold on for a while before we make the final decision to remove it. We need discuss more on it.

hidroh:
My client depends on the old `index.json` to discover commands. I'm quite late to the conversation so I may get things wrong, so we should point to http://tldr-pages.github.io/assets/index.json now, or even this `index.json` would soon be removed?

ostera:
@hidroh the new index will be hosted at the tldr-pages URL. You only need to change that url.

hidroh:
Got it thanks.

hidroh:
[...]

Issue #487, opened 2015-12-30
find a way to rebuild automatically the index


Shir0kamii:
Hi,

I've just discovered tldr, and I think it's a great tool.

One annoying thing though, is that the index wasn't up to date when I forked. It's normal for us developers to find out what wrong when we see a `vim.md` but cant do `tldr find vim`. But I think normal people wouldn't find help about vim in the first place if tldr don't find it.

For information purpose, I use tldr.py 0.2.0 and if a page isn't in the index, the client won't find it.

Should we add the index updating to the pre-commit hook ? I guess it's the best way to keep it up to date, regardless of the client. What do you guys think about this suggestion ?

lord63:
> Should we add the index updating to the pre-commit hook ?

In fact, it was in our pre-commit hook and then we removed it.  You can take a look at this issue #336 . Currently we don't have a clean solution to solve this problem.

As for the [tldr.py](https://github.com/lord63/tldr.py) client for building the index, I've opened an issue here before: lord63/tldr.py#8

ostera:
@Shir0kamii, we discussed and have a working solution here: https://github.com/tldr-pages/tldr/issues/336

Other than that, closing as this is related to a client not breaking their index/cache.

Shir0kamii:
Okay, thank you two for the fast replies.

Please tell me if I misunderstood, but from what I read, the index.json in this repositories is not guarranted to be up to date ? If so, why not remove it ?

ostera:
The updated index can be found on the tldr-pages.github.io repo, and you're
right, it should be removed from this one.

dbrgn:
Same problem here, I relied on the `index.json` file to be up to date. But it actually contains entries that aren't present in the directroy structure.

Where exactly can I find an up to date index on https://tldr-pages.github.io/?

igorshubovych:
tldr-pages.github.io/assets/index.json

PR #575, opened 2016-01-04
Removing pages/index.json from source tree

igorshubovych:
We were asked several times to remove the index (#574, #487).

It seems the time has come.

No index, no cry!
Bob Marley


.gitignore
```diff
@@ -1,10 +1,6 @@
...
-# the 'global ignore file'.
-# See: http://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration
-#    : search for "global .gitignore"
-
 # bundler specific
 .gem
 .bundle
+
+# Pages index is not staged
+pages/index.json
```
lord63: I think there is no need to remove this line, cause there will be an index.json. It may be annoying to see this file when we run `git status`.

igorshubovych: I did not remove it from `.gitignore`. I added it to `.gitignore`.

So you will never see it in `git status`.

felixonmars:
:+1:

G570972680893:
😎

.gitignore
lord63: ooooooooooooooooooooooooooooooh no, totally my bad. >_<

lord63:
By the way, I don't know whether other clients rely on the `index.json`. We can hold this PR for a while, in case there may be some other side effects.

igorshubovych:
@lord63, yes, i think it is good idea to wait a little bit.

ostera:
:+1:

igorshubovych:
Do you think guys that the time has come?

felixonmars:
:+1: It has been so long :P

ostera:
Closing as Github didn't automatically close with the commit message reference (https://github.com/tldr-pages/tldr/commit/b223ebe01f64f57cba019cee87a7b9bca5baddf9).

PR #578, opened 2016-01-04
Fixed notify-send

rakeshdas1:


pages/linux/notify-send.md
```diff
@@ -0,0 +1,15 @@
+# Notify-send
+
+Uses the current desktop environment's notification system to create a notification. Currently tested and working on the XFCE, Unity, Cinnamon, Enlightenment, GNOME, KDE Plasma, and MATE desktop enviroments. Requires the dependency `libnotify`.
```
igorshubovych: Too long; didn't read.
Can you make this text shorter?

pages/linux/notify-send.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Sends out a notification with the title "Test" and the content "This is a test"
+
+`notify-send "Test" "This is a test`
```
igorshubovych: It's better to use tokens as it is said in [Contributing Guidelines](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md).

PR #581, opened 2016-01-04
add ssh-keygen

antoniom13:


waldyrious:
Nice one, thanks. I think it would be useful to start with simpler examples and build up the complexity / number of options gradually. That way the defaults are also exposed.

antoniom13:
Sounds good ! Will push changes and see what you think.

antoniom13:
@waldyrious what do you think?

waldyrious:
Better, but I'd suggest making a distinction in the example descriptions making it clear which of the parameters are defaults. The reader can deduce this, but we're just slowing them down by not providing this information upfront :)

denis-sokolov:
@antoniom13, consider not documenting so many details. See #604 for additional inspiration.

waldyrious:
@denis-sokolov I actually think these examples are quite useful (at least the first and the third, I never had to use the second one). 3 examples is still a pretty low number so we're fine in that respect. Btw, would you like to add the examples you included in your PR?

denis-sokolov:
I might not have been clear enough in my previous comment. I find that the descriptions for the commands ("Generate a 2048 bit RSA key in ~/.ssh/id_rsa directory", for an example) are incorrectly detailed. There is no guarantee about the key length, type by default. The path will be confirmed interactively. Thus the more correct description for the command with no options is "Generate a key interactively". Then people who care about the type or length, should have these options described to them separately.

As for my examples, I closed my PR, because antoniom13 was first and I wanted to preserve "ownership" over the PR. :)

Here's my PR to this PR:

```md
# ssh-keygen

> Generate ssh keys for authenticating

- generate a key interactively

`ssh-keygen`

- generate a 4096 bit DSA key with your email as a comment

`ssh-keygen -b 4096 -t DSA -C "{{email}}"`

- Retrieve a cached fingerprint for a host to compare on a new machine.

`ssh-keygen -l -F {{remote_host}}`
```

waldyrious:
Good points @denis-sokolov, and I agree with much of what you say, but I don't think it's a good idea to introduce 3 new options simultaneously, as happens in the third example in your proposal. I would split that up into two examples to make the ramp-up gentler.

denis-sokolov:
I think the combination of options allows for much space saving, and when the options are simple, compatible, and can be visually compared to the description, it works fine:

```
generate a   4096 bit  DSA key  with your email as a comment
ssh-keygen  -b 4096  -t DSA     -C "{{email}}"
```

But this is a judgement call and I am not one to make it as I am not the decision maker in the project! It's alright if you choose to split it.

waldyrious:
Sure, but in this case we aren't exactly pressed for space, so we can afford the extra beginner-friendliness ;)

antoniom13:
@denis-sokolov very good points.  I'm trying to balance between tl;dr guidelines and what is "enough" information for someone new to understand.  Things get a little muddy when you start talking about encryption keys, defaults, and available options.  I always think of Einstein's Quote:

_"Everything should be made as simple as possible, but not simpler."_

I added the DSA option to show that you can change the type of encryption.  Probably more advanced than most people might need starting off, but shows the correct flag to do so.

@waldyrious do you want to add:

```
- Retrieve a cached fingerprint for a host to compare on a new machine.

`ssh-keygen -l -F {{remote_host}}`
```

waldyrious:
@antoniom13 I haven't used that and the description makes it sound a bit advanced, so I wouldn't insist on including it, but I'll let you be the judge :)

denis-sokolov:
You know how when you SSH the first time to a machine, it asks you to confirm a fingerprint?

> The authenticity of host 'example.com (1.2.3.4)' can't be established.
> RSA key fingerprint is 25:77:b1:58:0f:63:f7:3b:fb:8a:de:13:9e:ef:f5:fd.
> Are you sure you want to continue connecting (yes/no)?
```

This is to ensure you are connecting to the machine in question and not to an attacker. How do you validate this fingerprint? The easiest for me is to take another machine that I already have the fingerprint stored and see what the fingerprint was there. That's what `ssh-keygen -l -F {{remote_host}}` does and what you should be running, too, unless you have a better way to validate the fingerprint.

waldyrious:
Hmm, I wonder if we can include some of that information in the command description. Maybe "Retrieve the RSA key fingerprint for a host (useful for confirming the authenticity of the host when first connecting to it via SSH)"?

denis-sokolov:
Sure! I'm don't yet fully understand the logistics of long descriptions in the TLDR project.

waldyrious:
They're not set in stone, but hopefully will become clearer as we grow :) @antoniom13 feel free to update the PR with my suggested description or your proposal.

antoniom13:
Updated @waldyrious  @denis-sokolov

waldyrious:
LGTM :+1:

denis-sokolov:
:ship:

felixonmars:
:+1: Please squash the commits, thanks :)

PR #582, opened 2016-01-04
Create Xnest

hhirsch:


pages/linux/Xnest
```diff
@@ -0,0 +1,9 @@
+# Xnest
+
+> Start a nested x-server
+
+- Start nested x-server as :1 and start a terminal inside
+
+`Xnest :1 -ac &`
+
+`xterm -display :1`
```
igorshubovych: Why is this command?

hhirsch:
@igorshubovych Why is this command what?

pages/linux/Xnest
igorshubovych: You have `Xnest` command example, and then `xterm` example without explanation.

G570972680893:
😎💜

hhirsch:
@igorshubovych You can use the terminal to start any number of other processes inside Xnest.

felixonmars:
I'm still not convinced that `xterm -display :1` belongs here, since it's an example of using an application-specific option to choose an alternative display, which is not limited to Xnest either.

hhirsch:
@felixonmars Yeah. Whatevs.

PR #586, opened 2016-01-04
Add Sass

aarinsmith:


pages/common/sass.md
```diff
@@ -0,0 +1,12 @@
+# Sass
+
+> Converts SCSS or Sass files to CSS
+> .scss and .sass can be used interchangeably in below examples depending on your preference
+
+- Immediately one time convert SCSS or Sass file to CSS
+
+`sass {{inputfile}}.scss:{{outputfile}}.css`
```
igorshubovych: As far as I understood from documentation, there should be space, not a colon:

`sass [options] [INPUT] [OUTPUT]`

aarinsmith: This is true however the sass command accepts either syntax. For sass --watch, colon is used to distinctly name the output file. Kept with colon syntax to keep consistency between commands.

pages/common/sass.md
```diff
@@ -0,0 +1,20 @@
...
+
+`sass {{inputfile}}.scss {{outputfile}}.css`
+
+- Watch SCSS or Sass file for changes and output or update CSS file with same filename
+
+`sass --watch {{inputfile}}.scss`
+
+- Watch SCSS or Sass file for changes and output or update CSS file with specified filename
+
+`sass --watch {{inputfile}}.scss:{{outputfile}}.css`
```
igorshubovych: Isn't it the same syntax for `--watch` key?

aarinsmith: No, --watch will not allow you to use a space when specifying an output file. The single sass command will let you use a space OR a colon to delimit the input and output file but the watch command only allows for the colon delimiter. I used colon in the first commit as I felt it kept some consistency between the commands but it is less documented and is does not seem to be the standard so reverted to space delimiter.

image

pages/common/sass.md
```diff
@@ -0,0 +1,20 @@
+# Sass
+
+> Converts SCSS or Sass files to CSS
+> .scss and .sass can be used interchangeably in below examples depending on your preference
```
rubenvereecken: We shouldn't be 'picking sides', but I don't think it's useful to add a note about the conventional file extensions. Especially because they're not required I believe?

pages/common/sass.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Immediately convert scss file and output css with same filename
+
+`sass {{inputfile}}.scss`
```
rubenvereecken: I don't see why, but I don't get this behavior. My css is just written to stdout, no file to be seen.

rubenvereecken:
We tend to put file extensions inside brackets because usually they are hints as to the use, not requirements. From experimentation I learned that the sass compiler defaults to the sass syntax, but compiles scss if it sees the `scss` extension. Is this what you had in mind when you put the extensions outside brackets?

pages/common/sass.md
aarinsmith: Oh interesting, my mistake. I think I got mixed up with the functionality of the `--watch` command

aarinsmith:
This is what I had in mind. As it is your choice which syntax you are using I intended this to mean it is your choice. I see that is a bit misleading. I have committed a revision trying a different format let me know what you think of this.

rubenvereecken:
OK, good call, I like the new version where it's made obvious two formats are supported. My bad for mistaking the two interchangeable. If we go with this one I think you can safely remove the line about scss vs. sass in the main description.

aarinsmith:
I agree, removed the line in description. Glad you find the new version clear. I was unsure whether that would be straight forward enough to be understood.

rubenvereecken:
Thanks for clearing that up! Looking at it now it looks very concise, short and snappy in the tldr mindset.

Thanks for a great contribution.

aarinsmith:
Awesome! Great project, glad I could contribute!

PR #587, opened 2016-01-04
bash: add page

Ethiraric:
The description may not be the best.
Feel free to suggest another one.

I'm actually surprised nobody has made this page already. Is there a particular reason ?
Looking through the Pull Request history, I couldn't find any one adding the page.

### DIFF
--- a/pages/common/bash.md
+++ b/pages/common/bash.md
@@ -1,0 +1,20 @@
+# bash
+
+> Bourne-Again SHell
+> `sh`-compatible command line interpreter.
+
+- Start interactive command line interpreter
+
+`bash`
+
+- Execute command passed as parameter
+
+`bash -c {{command}}`
+
+- Run commands from file (script)
+
+`bash {{file}}`
+
+- Run commands from file and print them as they are executed
+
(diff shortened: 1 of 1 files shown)


Ethiraric:
I mostly wanted to differ from the manual page.

waldyrious:
Looks nice :) One suggestion: maybe instead of "run", "start" could make it more explicit that the interpreter will be used interactively. The same kind of rephrasing could make the -x example description clearer as well.

Ethiraric:
I've changed locally the run to start, but I don't seem to understand what you expect me to change for the -x example description. This option may be used in both interactive and non-interactive modes :/

waldyrious:
Yeah, but the specific example (just bash -x) does start an interactive session, right? In that case I believe the description should match. Besides, a one-off usage of e.g. bash -c -x would be a bit artificial since the -x wouldn't be justified in most cases.

Ethiraric:
Should I then make the -x example non-interactive ?

waldyrious:
Looks great, thanks! :+1:

igorshubovych:
Thank you!

PR #588, opened 2016-01-04
sort: add -h example

Ethiraric:


pages/common/sort.md
```diff
@@ -13,3 +13,7 @@
 - Sort passwd file by the 3rd field
 
 `sort -t: -k 3n /etc/passwd`
+
+- Sort file containing sizes in human readable form
+
+`sort -h {{filename}}`
```
igorshubovych: File containing sizes? This is something artificial.
Maybe (I do not insist) in this case it is better to use pipe:

```md
`ls -lh | sort -h`
```

igorshubovych: Also, `-h` is for Linux only. It does not work in OSX. In this case you can copy this page in linux and osx folder, removes the one in common, and add this line only to linux page.

Ethiraric:
I've read several times you do not like to have cross references between manual pages.
I personally use it only with pipes (most of my aliases output to human readable form).
I did not know this command was only on Linux.
I'll do what you suggest tomorrow (copying to separate folders).

waldyrious:
I agree in this case the use of ls is warranted and actually more natural than a hypothetical file containing sizes. Piping is generally avoided, but simple examples are ok to demonstrate common usages.

igorshubovych:
@Ethiraric 
> I've read several times you do not like to have cross references between manual pages.
> I did not know this command was only on Linux.

Please checkout [du on Linux](https://github.com/tldr-pages/tldr/blob/master/pages/linux/du.md) and [du on OSX](https://github.com/tldr-pages/tldr/blob/master/pages/osx/du.md). They are almost the same, but have slightly differences in details.

In such cases we create copy-paste things in common and add details to each page.

waldyrious:
@igorshubovych I believe by cross-references he meant using commands other than the one being covered (i.e. using piping in the examples).

igorshubovych:
Oh. Ok. Never mind then. Regarding using ls, I am not sure. I would love to hear others' opinion.

rubenvereecken:
"file containing sizes" is indeed too artificial. 

I've noticed a tendency where pipe or stdin examples are needed to use rudimentary commands to pipe that into the command in question. Just today I saw something pass by that catted a file and pipped it to the command. I feel it's justified, as long as these commands piping in data are easy to understand, not an extra effort, and should belong to common use cases.

In this particular case, is `sort -h` common? I didn't know about it, useful indeed. `ls -lh` won't work though, as the file sizes are not the first column. I recall being able to choose the sorting column and indeed I got the required result with `ls -lh | sort -h -k5` (5th column), but then that becomes a non-trivial use of `ls`.

Ethiraric:
When running `du -h`, the first column is the size, and this can easly be piped to `sort -h`.

However, `du` is not one of the most common commands.

rubenvereecken:
@Ethiraric I'd also thought of `du -h` because it's one of the only places where I use `-h`, but I found it awkward because by default it's recursive and it gets you a lot of output. But it might be just good enough to demonstrate the `-h` option for `sort`. I'm up for using that one instead of `ls`, but I'm eager to hear back from @igorshubovych.

igorshubovych:
To me `du` is far more advanced command than `ls`. I would prefer to use only `ls` and `echo` in piping.

@rprieto what do you think about using `du` in this context?

rprieto:
How does `sort -h` work on Linux? Can you both `sort -h {{glob}}` and `command_that_outputs_files | sort -h`?

igorshubovych:
@rprieto 
oh, it does not. Should be in Linux version. 
To run it on OSX I did

```sh
brew install coreutils
# And then
gsort -h
```

rubenvereecken:
I'm all up for the tougher example `ls -lh | sort -h -k5`, introducing two options at the same time. Basically a merge.

waldyrious:
@Ethiraric please create a linux/sort.md page as a copy of this one, with the proposed change, per @rprieto and @igorshubovych's comments above. If you prefer, I can do that myself -- let me know.

Ethiraric:
Wow sorry, this completely got out of my mind.
I don't have that much time at the moment, you'll probably do that faster and better.
If you don't mind, I'd prefer you do it.

waldyrious:
No problem @Ethiraric, we'll take it from here :)

waldyrious:
Superseded by #859. Thanks @Ethiraric!

PR #486, opened 2015-12-30
Add command 7za, unrar and git-svn command

joelhy:
add 7zip command line utility 7za and unrar command support.

PR #590, opened 2016-01-05
7za: add page

joelhy:
add 7zip command line utility 7za to common page

### DIFF
--- a/pages/common/7za.md
+++ b/pages/common/7za.md
@@ -1,0 +1,19 @@
+# 7za
+
+> A file archiver with high compression ratio
+
+- compress directory or file
+
+`7za a {{compressed.7z}} {{directory_or_file_to_compress}}`
+
+- decompress an existing 7z file with original directory structure
+
+`7za x {{compressed.7z}}`
+
+- compress to zip format
+
+`7za a -tzip {{compressed.zip}} {{directory_or_file_to_compress}}`
+
+- create multipart 7zip file; `part_size` specifies part size in Bytes, Kilobytes, Megabytes or Gigabytes.
+
+`7za -v{{part_size}}{{[b|k|m|g]}} {{compressed.7z}} {{directory_or_file_to_compress}}`


joelhy:
origin from https://github.com/tldr-pages/tldr/pull/486

pages/common/7za.md
```diff
@@ -0,0 +1,19 @@
+# 7za
+
+> A file archiver with highest compression ratio
```
waldyrious: "highest" is a bit unclear here -- highest than anything else currently available? highest possible? I would suggest replacing with just "high".

pages/common/7za.md
```diff
@@ -0,0 +1,19 @@
...
+
+- package and compress directory or file
```
waldyrious: **a** directory

pages/common/7za.md
```diff
@@ -0,0 +1,19 @@
...
+
+- decompress an existing 7z file
+
+`7za x {{compressed.7z}}`
+
+- compress to zip format
+
+`7za a -tzip {{compressed.zip}} {{directory_or_file_to_compress}}`
+
+- create multipart 7zip file; part_size: specifies part size in Bytes, Kilobytes, Megabytes or Gigabytes.
```
waldyrious: the ":" is unnecessary here. Also maybe wrap "part_size" in code tags `` or the token syntax {{}} to make it explicit that it refers to the command parameter below.

pages/common/7za.md
joelhy: Fixed. Thanks a lot.

pages/common/7za.md
joelhy: Wording changed and code tags `` added.

rubenvereecken:
I've got a paranoia with compression tools about them keeping directory structure. I think it started in the Windows days with zip and the like where I often ended up with everything dumped in the current directory (extract zipped directory to this directory) or, in an effort to battle this behavior, ended up with a two-layered directory `unzipped/unzipped/...`.

Would it be useful to add "with original directory structure" to the decompress example, [like unrar did](https://github.com/tldr-pages/tldr/pull/592/files#diff-569a49e2a10656da462089e7bb76e82cR9)?

joelhy:
"with original directory structure" added to decompress comment.

rubenvereecken:
I guess it doesn't hurt. Thanks for caring for my paranoia. You have my :+1:

waldyrious:
:+1:

PR #593, opened 2016-01-05
linux/pacman: add more useful commands

imrehg:


felixonmars:
:+1: as they are really useful use cases.

pages/linux/pacman.md
```diff
@@ -25,3 +25,11 @@
...
 
 `pacman -Qe`
+
+- find which package owns a certain file
+
+`pacman -Qo filename`
+
+- empty package cache to free up space
+
+`pacman -Scc`
```
felixonmars: There may be arguments between `-Sc` and `-Scc` here, though.

imrehg: Yeah, of course :) I'm simplifying and I guess curating with this choice a bit. Personally I always use `-Scc` that's why I included that. Looking at [the wiki page](https://wiki.archlinux.org/index.php/Pacman#Cleaning_the_package_cache), it says about `-Scc` that "_It should be avoided unless there is an immediate need for disk space._" I'm only running this command when I see that "oh, I'm out of space, let's see what we can get rid of", so in that sense it does exactly what the label says. So anyways...

felixonmars: OK, then let's keep it :)

imrehg: I'm happy to see any other useful commands or later changing these variations if needed :)

rubenvereecken:
This page now clocks in at 8 examples. I suggest we trim it down a bit. I've only rarely used pacman but from a glance I'd say `-Qe` and `-Scc` are the less common/useful ones here.

imrehg:
Well, I haven't really used `-Qe`, but all the other ones are pretty common for me.

If need to cut by 2, I'd remove `-Qe` and `-Qo`, and replace `-Q` with `-Qi package-name`. Just a thought.

rubenvereecken:
Eh, tough nut. The 5-6 recommendation was chosen a bit arbitrarily because that's about what a person would suffer to skim through to find the options she needs. Pacman is a large and powerful command though, so I find it hard to contain it within that recommendation.

@imrehg, those two removals you suggest are already enough of a cutdown. Does `-Qi` list files contained in a package or something? Don't have pacman here.

imrehg:
No, it does not, that's `-Ql package-name` to list all the files. An example of the `-Qi package-name` is this:

```
$pacman -Qi chromium
Name           : chromium
Version        : 47.0.2526.106-2
Description    : The open-source project behind Google Chrome, an attempt at creating a safer, faster, and more stable browser
Architecture   : x86_64
URL            : http://www.chromium.org/
Licenses       : BSD
Groups         : None
Provides       : None
Depends On     : gtk2  nss  alsa-lib  xdg-utils  bzip2  libevent  libxss  icu  libexif  libgcrypt  ttf-font  systemd  dbus
                 flac  snappy  speech-dispatcher  pciutils  libpulse  harfbuzz  libsecret  libvpx  perl  perl-file-basedir
                 desktop-file-utils  hicolor-icon-theme
Optional Deps  : kdebase-kdialog: needed for file dialogs in KDE
                 gnome-keyring: for storing passwords in GNOME keyring [installed]
                 kwallet: for storing passwords in KWallet
Required By    : chromium-pepper-flash-stable
Optional For   : lastpass
Conflicts With : None
Replaces       : None
Installed Size : 166.38 MiB
Packager       : Evangelos Foutras <evangelos@foutrelis.com>
Build Date     : Fri 18 Dec 2015 03:36:41 AM CST
Install Date   : Fri 18 Dec 2015 04:39:10 PM CST
Install Reason : Explicitly installed
Install Script : Yes
Validated By   : Signature
```

kuanyui:
>This page now clocks in at 8 examples. I suggest we trim it down a bit. I've only rarely used pacman but from a glance I'd say -Qe and -Scc are the less common/useful ones here.

`-Scc` is useful for me at least because my disk free space is less. But I always need to Google for it every time I use.

rubenvereecken:
Taking #594 as a precedent for large tldr files, I'll merge this one in since it's such a big tool. I've used all of the commands present except for `-Qe`, plus the ones added, so I don't feel good snipping them away. If anyone else feels like trimming out the less useful commands, I'd be happy to take a look at it.

Note to self or other contributors: Powerful commands are annoying because they often have a lot of common use cases. Maybe add a `--find` in clients, but that gets dangerously bloated quickly.

PR #594, opened 2016-01-05
grep: add line number

antoniom13:
Option to print line number when a match is found.  If you are a VIM user, it's common to grep -> show line number -> open file and go directly to the line number in the file instead of having to grep for the matches once again once in vim.

### DIFF
--- a/pages/common/grep.md
+++ b/pages/common/grep.md
@@ -23,6 +23,10 @@
 
 `grep -c {{something}} {{file_path}}`
 
+- print line number for each match
+
+`grep -n {{something}} {{file_path}}`
+
 - use the standard input instead of a file
 
 `cat {{file_path}} | grep {{something}}`


igorshubovych:
I like your example. It is quite common and useful. But the page looks long: 8 examples instead of recommended 5-6.

rubenvereecken:
I'd say it might already be more useful than for example the last one.

Shir0kamii:
I agree with @antoniom13 about the vim note. The line number example would be much more useful to me than the invert match example.

rubenvereecken:
I'd be for swapping in your example for the last one for example, for the sake of brevity. However, that's stating my preference so I'd like to hear someone else on this as well.

igorshubovych:
Let me then merge it. If anybody of you feel that some example is less usable, please submit a new PR.

My personal preference is to eliminate "print the count of matches instead of the matching text".

igorshubovych:
@antoniom13 Thanks for contribution!

PR #373, opened 2015-12-23
fortune: add page

igorshubovych:
And finally the most useful Unix command arrived!

PR #595, opened 2016-01-05
fortune: reword description of -f option, reorder

waldyrious:
As I've commented in #373, I think the description of the -f example should use the word "file" because it matches with the short option name.
I've also updated the `fortune {{database}}` example to match the terminology, and reordered the examples in an order that I believe is more intuitive.

### DIFF
--- a/pages/common/fortune.md
+++ b/pages/common/fortune.md
@@ -6,14 +6,6 @@
 
 `fortune`
 
-- Print a quotation from a given database:
-
-`fortune {{database}}`
-
-- Print a list of quotation databases:
-
-`fortune -f`
-
 - Print an offensive quotation:
 
 `fortune -o`
@@ -25,3 +17,11 @@
 - Print a short quotation:
 
 `fortune -s`
+
+- List the available quotation database files:
+
+`fortune -f`
+
+- Print a quotation from one of the database files listed by `fortune -f`:
+
+`fortune {{filename}}`


felixonmars:
I think "database" somehow implies "database file" for single-file databases, just like we say "a sqlite 3.x database" instead of "a sqlite 3.x database file" when the context is not related to the inner structure of the file.

igorshubovych:
@rprieto we need your opinion here

waldyrious:
@felixonmars I don't think sqlite is a good reference for this discussion, since it was specifically built as a single-file database, while in general the term "database" tends to be used interchangeably with DBMSs, e.g. MySQL. [Wikipedia corroborates this](https://en.wikipedia.org/wiki/Database):

> Sometimes a DBMS is loosely referred to as a 'database'.

igorshubovych:
@waldyrious 
I understand you and partially agree with you,
But I still don't like the word `{{file}}`.

1. It is factually wrong. There are 2 files for each dictionary/database. One file without extension and one with `.dat` extension.

2. These files has be in certain folder, you cannot provide files from anywhere. And when you list them with `fortune -f` it is just list the names. So overall it is quite similar to operation of listing databases in MySQL, PostgreSQL, and not similar to listing of files.

3. It is files of certain format. Not every file can be used in this context.
You can do `cp`, `rm`, `git commit` any file. But you cannot `fortune {{any_file}}`.

So if you really don't like `{{database}}`, I would propose to use `{{dictionary}}`.

waldyrious:
It's not that I don't like database, I just think it gives the wrong idea about what people are dealing with. I'll investigate some more and get back to you guys in a bit.

waldyrious:
Ok so even though there are other files in the fortunes directory (.dat, and in my case .u8 as well), the fact is that fortune {{whatever}} expects "whatever" to be one of the entries returned by fortune -f. Also note that the [Wikipedia article for the fortune command](https://en.wikipedia.org/wiki/Fortune_%28Unix%29#Fortune_files) does call these "fortune files", and describes the -f option as
> Print out a list of all fortune files that would have been searched, but do not print a fortune.

Furthermore, "database" is a generic term and we can't exactly pass any database (e.g. sqlite) as a parameter to the command either.

Given all the above, I would suggest using {{fortune_file}}, the same way we would use, e.g., {{photoshop_file}} or {{html_file}} (referring to a specific content format, which isn't guaranteed to be provided by an extension).

Would that option work for you guys?

igorshubovych:
Wikipedia does it by mistake. If I have more time, i will go and fix it there.

Regarding other file types: picture.psd, home_page.html, module.py, HelloWorld.java, my_resume.doc - they all look so natural and do not need further explanation. Unlike {{fortune_file}}. Is a full path to fortune file? No. The full path make it fail here. Is it a .dat file? No. It should be just a name.

waldyrious:
Ok, so how about changing the description to "Print a quotation from one of the quotation database files listed by fortune -f"? I really think we should keep the -f here otherwise it looks like it comes out of the blue.

rprieto:
A lot of good arguments here, we probably need to choose one and move on. IMHO:
- it's a name, not a path, so the token should probably be `something_name`
- `file` is maybe too ambiguous for that reason
- I actually like `database_name`, I don't think anyone will think they can provide a random database they have lying around (just like `say -v {{voice}}`, no one will think they can pick a random voice like `veryHappy`)

waldyrious:
@rprieto What you say makes a lot of sense, but it pains me to remove the connection -f --> file. Perhaps we could get the best of both worlds by using "filename"? :D

rprieto:
Good point about `f` for `file`. Although they're different examples (`fortunes -f` and `fortunes {{database_name}}`), so the first one could say `List all fortune database files`.

Otherwise I'm happy with `filename` just to get this resolved and move on :)

waldyrious:
@felixonmars @igorshubovych @rprieto please re-review :)

igorshubovych:
LGTM!

igorshubovych:
Finally we agreed to something! Good job!

waldyrious:
I knew we could do it ;)

PR #597, opened 2016-01-05
Create lsblk.md

gautamkrishnar:
lsblk man

### DIFF
--- a/pages/linux/lsblk.md
+++ b/pages/linux/lsblk.md
@@ -1,0 +1,27 @@
+# lsblk
+
+> lists information about devices
+
+- list all storage devices in a tree-like format.
+
+`lsblk`
+
+- Also list empty devices.
+
+`lsblk −a`
+
+- Print the SIZE column in bytes rather than in a human-readable format.
+
+`lsblk −b`
+
+- Output info about filesystems.
+
+`lsblk −f`
+
+- Use ASCII characters for tree formatting.
+
+`lsblk −i`
+
+- Output info about block-device topology.
+
+`lsblk −t`


rubenvereecken:
You managed to upset Travis a bit, check out the details to find out where the linter got annoyed. The html complaints are erroneous, you can ignore those.

On a more serious note, while I see some really good examples listed the page is simply way too long. In our [contributing guidelines](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines) we suggest 5-6 examples, though we've gone a bit over that for large commands. This is not such a one I'm afraid. In fact, I see you listed more examples than my MAN page list usages, like the exotic `-S` option.

Our goal with tldr pages is to have common usage patterns quickly accessible, without having to skim through too much. Pages like that would deter users, and while they're really good for extensive examples, they're pretty much what tldr tries to solve. 

There's some great stuff in here, so I suggest you think over what commands would be used most often, or at least by newbies. For example, I've only ever used the `-a` option. 

I'd be more than happy to review a slimmed-down version of this file.

gautamkrishnar:
@rubenvereecken Ok :+1:

gautamkrishnar:
@rubenvereecken is it ok now :smile:

pages/linux/lsblk.md
```diff
@@ -0,0 +1,27 @@
+# lsblk
+
+> lsblk lists information about all available or the specified storage devices.
```
rubenvereecken: Could we shorten this into "lists information about storage devices"? That sounds snappier while still in essence containing the above.

pages/linux/lsblk.md
```diff
@@ -0,0 +1,27 @@
...
+
+- General syntax
```
rubenvereecken: We don't do general syntax, just examples. Syntax is a bit contained in those.

pages/linux/lsblk.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Output info about filesystems.
+
+`lsblk −f`
+
+- Use ASCII characters for tree formatting.
+
+`lsblk −i`
+
+- Produce output in the form of a list.
```
rubenvereecken: So just out of curiosity: this option gives me exactly the same as without. Is the list form the default one?

rubenvereecken:
@gautamkrishnar thanks for swiftly trimming it down. Let's just go over the inline comments and all should be good :).

gautamkrishnar:
@rubenvereecken done :+1: .

rubenvereecken:
We don't have guidelines for line formatting yet (capitalization/end with punctuation) but you're being inconsistent there with 2/6 having no periods at the end. If you can just patch those up you have my heartfelt :+1:

gautamkrishnar:
@rubenvereecken :+1:

gautamkrishnar:
@rubenvereecken squashed

rubenvereecken:
Great, thanks. We'll fix the syntax with the formatter.

Thanks for the page!

PR #600, opened 2016-01-05
Add mocha.md - The JavaScript test runner CLI

kkamperschroer:
Add the three most common cases I've encountered when using mocha.

Mocha: https://mochajs.org/

### DIFF
--- a/pages/common/mocha.md
+++ b/pages/common/mocha.md
@@ -1,0 +1,23 @@
+# mocha
+
+> Execute Mocha JavaScript test runner
+
+- Run tests with default configuration or as configured in `mocha.opts`
+
+`mocha`
+
+- Run tests contained at a specific location
+
+`mocha {{folder/with/tests}}`
+
+- Run tests that match a specific grep pattern
+
+`mocha --grep {{^regex$}}`
+
+- Run tests on changes to JavaScript files in the current directory and once initially
+
+`mocha --watch`
+
+- Run tests with a specific reporter
+
+`mocha --reporter {{reporter}}`


pages/common/mocha.md
```diff
@@ -0,0 +1,15 @@
+# mocha
+
+> Execute Mocha JavaScript test runner
+
+- Run tests with default configuration, or as configured in `mocha.opts`
```
rubenvereecken: You can safely drop the comma I think.

pages/common/mocha.md
```diff
@@ -0,0 +1,15 @@
...
+
+`mocha`
+
+- Run tests that match a specific grep pattern
+
+`mocha --grep {{^regex$}}`
+
+- Run tests every time files change until you kill it
```
rubenvereecken: How about something like "Run tests every time a file in the current directory changes", so that it gives more information? Or are the watched files configured through `mocha.opts` as well?

pages/common/mocha.md
kkamperschroer: Whoops. Thanks. Will fix in followup commit.

rubenvereecken: Cool, thanks. I know it's nitpicky, but commas are hard to sneak out again afterwards and the world already has so many of them. Could you squash your commits together once you've gone over the comments?

pages/common/mocha.md
kkamperschroer: According to the mocha docs it's: "Executes tests on changes to JavaScript in the CWD, and once initially."

So how about something like: "Run tests on changes to JavaScript files in the current directory and once initially"?

I'll commit that for now and if you'd like it changed just let me know.

Thanks.

kkamperschroer:
Alright, just squashed the commits back into one. Let me know if you see anything else you would like changed. Thanks for your time looking at this. I appreciate it!

igorshubovych:
IMO `mocha --grep` example is not very common.

I used a lot
```md
`mocha --reporter=min` # Reporters are always in mocha.opts, but it is useful to overwrite in CLI
`mocha {{path}}`
`mocha {{file1}} {{file2}} ...`
```

rubenvereecken:
I haven't used mocha manually myself but I have used other test runners where I did use the equivalent of `--grep`, even though just passing directories is way more common.

kkamperschroer:
Ok, so if I'm understanding this correctly, you would like to see something along the lines of `mocha {{path1}} {{path2}} ...` in there before the `--grep` option? Or no grep option at all?

According to the contributing guidelines, 5 or 6 most common cases is ok, so I could still add the `--reporter` option in there as well. Where would you prefer that placement as far as commonality?

The reason I created this is because I couldn't remember what the flag was for grep and saw `tldr` had nothing for mocha. I understand every person's most common usage of tools like mocha is going to be different. I was just going off the common uses for myself and those I work with, so the feedback is greatly appreciated.

rubenvereecken:
My own comment wasn't really going anywhere to be honest, not far anyway, just voicing support for `--grep`. I'd say, yes, bump it down a bit, definitely add a case for testing specific files/directories as @igorshubovych suggested, and maybe consider that reporter. I remember I've at least overridden the reporter using the CLI a couple of times myself, so at least there are other people using that.

kkamperschroer:
Thanks for all the input. I've made the suggested updates. Let me know what you think.

pages/common/mocha.md
```diff
@@ -0,0 +1,23 @@
...
+
+> Execute Mocha JavaScript test runner
+
+- Run tests with default configuration or as configured in `mocha.opts`
+
+`mocha`
+
+- Run tests contained at specific paths (files or directories)
+
+`mocha {{path-1}} {{path-2}} {{path-N}}`
```
igorshubovych: this way we start looking like `man`, you described it too formal.

In 95% of cases it is folder ./test or ./spec. So maybe it is good to mention that.
```
`mocha {{path/to/tests}}`
```

igorshubovych: Or even
```md
`mocha {{folder/with/tests}}`
```

kkamperschroer:
@igorshubovych Thanks for the feedback. Hard to find that balance I guess. :-)

I pushed another update with your suggestions. Let me know if there is anything else.

rubenvereecken:
I'm happy as it is :+1:

igorshubovych:
I think it is good enough at this stage! Thanks!

PR #601, opened 2016-01-05
ssh: remove quotes around remote command

denis-sokolov:


rubenvereecken:
Tested and confirmed, don't see it doing any harm. Thanks for the entry, I'll merge it in.

PR #602, opened 2016-01-05
ssh: use -N on port forwards

denis-sokolov:
A prompt is confusing when the terminal is meant for forwarding.
A "stuck" shell realistically represents something being "busy".

### DIFF
--- a/pages/common/ssh.md
+++ b/pages/common/ssh.md
@@ -21,11 +21,11 @@
 
 - ssh tunneling: dynamic port forwarding (SOCKS proxy on localhost:9999)
 
-`ssh -D {{9999}} -C {{username}}@{{remote_host}}`
+`ssh -D {{9999}} -C {{username}}@{{remote_host}} -N`
 
 - ssh tunneling: forward a specific port (localhost:9999 to slashdot.org:80)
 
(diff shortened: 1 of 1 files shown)


igorshubovych:
If you add an option, you have to explain the meaning in description.

denis-sokolov:
The command already includes `-C` with no explanation. A few common sense options that are good by default are nice. This is TL;DR. A user that needs to get things done has his recipe. A user that wants to know what `-N` does, only needs to `man ssh`, no?

(`-C`, in particular, can go away, because it is better served with in the SSH config. `-N` can not be added to the config)

waldyrious:
Nope, the goal of tldr is not to just provide quick refreshers, but to help people learn the commands and hopefully memorize them so they don't need tldr anymore (which is the ultimate tldr: didn't actually read anything at all!).

One of the ways this is achieved is by providing descriptions that use words matching the short option names, in order to provide useful mnemonics. So by all means, do describe `-N` and `-C` as well if you're feeling generous :)

denis-sokolov:
In light of trying to make the ssh page lighter in #603, I no longer see any benefit in adding `-N`. It does not provide enough value to offset the cost of additional visual complexity.

Issue #273, opened 2015-03-25
Command search


rubenvereecken:
I'd like to raise an idea I've played with for the last couple of days, especially when I noticed @rprieto raised it in #267 as well. 

The next step after (or for) tldr is searching for commands by descriptions in natural language. When I thought of it I didn't think of it necessarily as part of tldr but it could be built on top of it. The idea is simple: we can already invoke tldr to remember those pesky command line options for a certain command. Now take a step back into novicehood and you'll want to find commands that do something in the first place. That's where this idea comes in.

A very naive version would consist of tagging commands (again, not necessarily inside tldr) and search through those when a user tries something along the lines of `(tldr) search unpack tar`. This can actually be built straight into #tldr using the index proposed at #267 because then the client could search the tags that would be included in the index.

Take a step further and you need an actual server to do language processing. I'd actually be interested in trying this out but it'd be a considerably bigger project and it couldn't live inside tldr anymore.

Thoughts on this?

rprieto:
I like this idea :+1:

I agree on a simple version being built into `tldr`, that'd be very handy. And if it's a bigger service, could definitely be built on top. Not sure if it would have to be processing anything server side though, I thought simplicity was one of the strengths of `tldr` compared to other tools for example, there's no need for servers, very little maintenance, etc.

A while ago I thought of adding it to the `node` client, where searching is easy because it caches all the files locally, but the index file in #267 is great because it would enable it for everyone.

Without getting into complex NLP, I imagine we could still have a text/keywords search that's quite powerful. Keywords would be nice but would also be redundant... otherwise could we just parse & process the text on each page? Thoughts?

If we just use the text on each page, I imagine we could concat all examples, then remove any articles, pronouns etc... For the current `tar` page, this would give us:

> create archive file gzipped extract target folder current directory bzipped compressed suffix determine compression program contents tar

or once passed through [stemmer](https://www.npmjs.com/package/stemmer)

> creat archiv file gzip extract target folder current directori bzip compress suffix determin compress program content tar

then it should be easy to find all commands that match

```
$ tldr --search compressed bzip
```

Sadly it wouldn't return anything for `--search unpack` because that wasn't used in the page text.

rprieto:
Update: it seems that https://www.npmjs.com/package/stm also takes care of removing articles & extra stop words.

ostera:
:+1: on the search.

If the clients implement caching, then full-text search is an obvious answer for the cached pages. Not so for the ones that are not available tho.

[...]

PR #603, opened 2016-01-05
ssh: add remote tunneling

denis-sokolov:


igorshubovych:
The page overall looks bloated. With your example added, it has 8 examples.

denis-sokolov:
I agree. I already have plans on cleaning it up and shortening it by combining some options if you approve the PR.

That said, the reason is `ssh` is a very powerful command that does LOTSATHINGS.

waldyrious:
I would prefer you to do all your proposed changes in this PR then.

denis-sokolov:
Major changes to the PR proposed.

1. Reduced mental load where possible. In particular, replaced `username@remote_host` with `remote_host` everywhere except the first example. Removed a non-important and unexplained `-C`.

2. Combined `-i` and `-p` option into one due to the lack of space. The options are clearly separate and should cause no confusion.

3. Removed `-A` option. It is both simple and dangerous and anyone willing to use it should already know what they are doing.

4. Added a combined note on tunneling. The description is a bit long, but I could not make it shorter.

igorshubovych:
1. I like it.
2. I would not do it. I would rather split it anyway. But i am open to discuss it.
3. I like it.
4. Not sure. Do we really need a tunneling example at all? Is it for newbies?

denis-sokolov:
@igorshubovych, please excuse my boldness, I but only am excited about the project and want it to be awesome.

I think you need to make some decisions and decide what you want the tool to achieve. I see two very conflicting use cases the application tries to cover - newbies and more advanced users. This is reflected in your comments and in the readme.

1. Command line newbies. These people have heard that `ssh` is useful or they need to use it for work, but they are intimidated by the command line, `man`, and searching the internet. This use case is reflected in the following quote from the Readme: "New to the command-line world?" It is also reflected in your comment "Is it for newbies?".

2. People that use command line periodically and need a reference on the order of flags to `tar` and how exactly do you set up a tunnel for `ssh`. These people know how options work, they do not very close hand-holding, but they do need to remember how to use one thing or another. This is reflected in quotes from Readme: "Or just a little rusty? Or perhaps you can't always remember the arguments to lsof". Notably, `lsof` is in no way a newbies tool.

A tool that covers both these use cases is awesome, of course, but sometimes the goals of covering the two use cases conflict. In the case of tldr page for `ssh`, for an example, newbies need hand holding and separate introduction of `-i` and `-p` options, while users that use tldr for reference can be satisfied with a quick one-liner reminding them of 3 options, and then 2 more examples with tunneling. (Show something on a local computer for a remote client? `ssh -R` Access a restricted web service? `ssh -L`) Because we have limited space, we need to choose - gradual option introduction or tunneling examples.

Because we need to choose it would be best to have a clear value vision what the project aims to achieve first. In other words, which group of users is the more important one to the project. From the readme I did not get a clear value jugement like this, thus the current misunderstanding with my proposals.

I am clearly biased, as I want the tool to be useful for me and people like me. That said, it seems to be that newbies have plenty better alternatives for introduction for command tools. "ssh for beginners" or "howto ssh" can often bring awesome tutorials on the internet. In addition, it is unlikely many newbies can be expected to start with installing tldr and actually referencing it when in trouble. At the same time the more advanced users waste time and nerves for something that should be fairly trivial. We can already go to Google, and it's sometimes faster than digging through `man`, but typing something in the shell is even better.

In my biased opinion, I'd sacrifice the needs of newer users when the interests collide. The final call, of course, is up to you. But if possible, please make it and communicate it so that my future pull requests would be more valuable.

:rocket:

ostera:
personally, I think one future of tldr-pages lies in full-text search of a big number of useful commands for different tools.

@denis-sokolov, you're spot on on the contradiction between being for newbies and having expert commands there as well, and not limiting the number of commands we have per page would make it very easy to have both very simple and very complex examples living together.

You'd run `> tldr ssh tunnel` and you'll get all ssh tunneling examples, you'll run just `> tldr ssh` and you'll get the page in a similar way you get a man page, but filled with useful examples instead sorted from newbie to expert.

denis-sokolov:
>  not limiting the number of commands we have per page

We already have `man`. :) I think an important value of tldr is not to flood the user with information.

> You'd run > tldr ssh tunnel and you'll get all ssh tunneling examples,

I would hate to remember that. If, though, `tldr ssh` gave a few simple examples and then suggested "See also tldr ssh tunnels, tldr ssh agent and tldr ssh-keygen", I could dig that, indeed.

igorshubovych:
@denis-sokolov check out cheat:
https://github.com/chrisallenlane/cheat

ostera:
@denis-sokolov you misinterpreted me. `tldr ssh tunnel` isn't a fixed command but rather `tldr page query`. The query could be used to parse the text, could be used to match from a given list of tags generated at a client level, the possibilities are endless. We have discussed similar functionality in the past: https://github.com/tldr-pages/tldr/issues/273 – besides, you're already doing this when you google it, so no reason to hate anything :smile: 

Having an `ssh-tunnels.md` page will be inconsistent, since all of our pages have a 1-to-1 mapping to an actual executable. `ssh-keygen` or other `ssh-*` commands are very welcome.

And on the case of `man vs tldr`, it's not only length, but denseness. Man is super dense. Having a simple way to search for very specific keywords on very practical examples, however many there are, will likely yield a limited number of results. Not limiting the number of commands doesn't mean we will allow repetition: DRYness is something we strive for on every PR review.

@igorshubovych +1

denis-sokolov:
@igorshubovych, I saw it in the TLDR readme, when I looked at it. What of it? I have a more comfy tool of my own notes - I wanted to have a shared library of notes to use and contribute to, not a notepad. Thanks for pointing me in that direction, of course, but it's not useful to me.

> tldr ssh tunnel isn't a fixed command but rather tldr page query

That could be great, but do note the considerably higher cost of that compared to simple static pages. You discussed that in #273 and where is it? Still in theory. Adding multiple pages for a single command is already possible today.

> Having an ssh-tunnels.md page will be inconsistent, since all of our pages have a 1-to-1 mapping to an actual executable.

Does this mapping have an inherent value? I don't think so. Besides, this is not true already. `alias` has `unalias`, `bashmarks` have `l`, `s`, `g`, `p`, `d`; `cal` has `ncal`, `git-*` are not standalone executables, `mysqldump` has `mysql`, `sshfs` has `fusermount`, and so forth.

We can provide real value to real users by finally giving them a convenient reference on how to set up a damn tunnel and whether -L or -R is right for them, but we won’t?

rubenvereecken:
As a disclaimer, I only read half of the discussion because I'm short on time atm, but I'm planning on filling in the important gaps later. I'm addressing random statements that caught my eye for the moment.

I regularly think about the `tldr page query` functionality but I wanted to try it out in my own client. That one is pushed a bit farther off since I started on `tldr-lint`, alas. It would be decent, but it would be best if we don't limit the examples. I don't want to do that however; we need that limit to keep an individual page readable and tldr-befitting. The chances for such a query failing are big and the user will resort to the page anyway. 
On @denis-sokolov's note, the cost is only in the clients, so it's very, very bearable.

On the mappings, they should be as consistent as possible. The `git-*` for example are a nice exception because they fit well in the git world where it's common to refer to `git add` as `git-add`. This still doesn't mean we should shy away from referencing other pages, that's where a lot of strength lies.

That brings us to the difficult topic... who is the audience? I have to ask myself this question often and like to think it's just me, that's easiest. Not a newbie, not a sysadmin, but a very regular user that wants some more control over their system. A command line newbie that actually has tldr would still never do `tldr dpkg-query` to find out how to list files contained by a specific package. That's what StackOverflow is for. I'm that user that already landed at that post twice and from now on has a quick glance at tldr. 

I think this sides with @denis-sokolov's opinions, but I always liked the idea of combining some options to have _useful commands_ that do _useful, common_ actions. We're by no means consistent. Often pages are designed for newbies but will never be found by newbies. 

I'd forego simplicity in some cases in order to offer actual value, offering some tougher examples. In moderation, of course.

igorshubovych:
@denis-sokolov 

TLDR pages is available under MIT License. So you are free to go whatever you want with it.
Obviously several of us have the similar thoughts and ideas and we are open to do something more with it.

However the existing TLDR pages are for newbies, it is in the motto and in Contributing Guidelines. Therefore I would say no to SSH tunnelling examples.

denis-sokolov:
Thank you for the clarification, @igorshubovych, this helps out a lot!

waldyrious:
Without commenting on specific examples, I would say that in general we should never favor experts over newbies, since the experts can much more easily get around their problems if tldr fails them. That doesn't prevent us from catering to experts whenever we can, but if we have to *choose*, I'd be in favor of allowing tldr to be beginner-friendly even if that prevents us from covering one or two expert usages of a command.

But leaving the theoretical discussion aside, I think we can always attempt a compromise when such a conflict arises: for example maybe we can introduce two options at a time to make room for an extra example, if in that specific case we collectively agree the end result is a net positive. But it would be in a case.by-case basis -- the default stance should be IMO newbies first, and accomodating experts as well as we can. @denis-sokolov does that make sense to you?

denis-sokolov:
Yes, @waldyrious, thank you. This is in line with @igorshubovych's opinion and a bit unlike @rubenvereecken's.

waldyrious:
@denis-sokolov I would then suggest keeping this open as we settle that down. I'm sorry we ended up hijacking your PR for this discussion, but in a way it makes sense as it keeps the discussion framed in context.

waldyrious:
(Especially since I would hate for tldr to fail you as a way to make your (clearly comprehensive) personal library of notes publicly available and helpful to others)

igorshubovych:
@denis-sokolov @waldyrious 
Should we instead close it and at the same time open issue/task where somebody will summarize the discussion and link to this PR?

waldyrious:
I don't think so, @igorshubovych. We don't seem that far from reaching a consensus. Let's see what @rubenvereecken says.

PR #604, opened 2016-01-05
ssh-keygen: add

denis-sokolov:


waldyrious:
See #581

denis-sokolov:
Apologies for a duplicate.

PR #605, opened 2016-01-05
for/if/while: add

denis-sokolov:


pages/common/if.md
```diff
@@ -0,0 +1,11 @@
+# if
+
+> Simple shell conditional
+
+- Perform a command depending on another command's success.
+
+`{{cmd}} && echo "success" || echo "failure"``
```
igorshubovych: This is not if.

BTW, check the [page for test](https://github.com/tldr-pages/tldr/blob/master/pages/common/test.md).

igorshubovych:
Overall LGTM.

It's great how simply you explained them!

pages/common/if.md
denis-sokolov: I know. But do we want to stick to precise definition of what is a command? Because shell built-ins are not commands.
Point is, if I want to do a conditional, I will look for `if` (or `test`, as you suggested).
A `&& ||` pattern is a very good substitute for an if, isn't it?

denis-sokolov:
_(Update: added missing `sleep 1` to the first `while` example)_

pages/common/if.md
waldyrious: It could be even better to make it clear that "another comand" refers to {{cmd}}, since the order is reversed in the description and the actual example. And in any case, there's no need to compact "command" to "cmd" here. Long gone are the times that got us [umount instead of unmount](http://unix.stackexchange.com/questions/9832/why-is-umount-not-spelled-unmount) ;)

pages/common/while.md
```diff
@@ -0,0 +1,11 @@
+# while
+
+> Simple shell loop
+
+- Execute a command forever once every second.
```
waldyrious: It would be gentler to start with a terminating while example before this one. "Forever" is scary for beginners :)

pages/common/for.md
```diff
@@ -0,0 +1,11 @@
+# for
+
+> Shell loop over parameters
+
+- Perform an action with different arguments.
```
waldyrious: Why use "argument" in the command description but "parameter" in the actual command? I'd suggest unifying the nomenclature.

waldyrious:
Very useful and needed! I've added a few inline comments to ensure these are at their peak usefulness when merged ;)

denis-sokolov:
_(Update: taken into account the comments received)_

pages/common/for.md
```diff
@@ -0,0 +1,11 @@
...
+
+`for argument in 1 2 3; do {{cmd $argument}}; done`
+
+- Perform an action in every directory.
+
+`for d in *; do (cd $d; make test); done`
```
igorshubovych: @waldyrious 
do you think `make test` is appropriate here?

denis-sokolov: Yeah, the example was poor. Improved!

waldyrious: Good catch. Perhaps a simpler command that makes it clear this is just an example command for demonstration, and not part of the `for` loop, would work better. I believe `echo` would be a fine choice for this.

pages/common/if.md
```diff
@@ -0,0 +1,11 @@
+# if
+
+> Simple shell conditional
+
+- Echo a different thing depending on a command's success.
+
+`{{command}} && echo "success" || echo "failure"``
```
waldyrious: You forgot an extra backtick at the end of the line :)

pages/common/for.md
waldyrious: Ah, this was changed before my comment above, please disregard. @denis-sokolov's improvement was a better one than my suggestion :D

pages/common/while.md
```diff
@@ -0,0 +1,11 @@
+# while
+
+> Simple shell loop
+
+- Read stdin and perform an action on every line.
+
+`while read line; do {{echo "$line"}}; done`
```
waldyrious: Since this is a concrete example, I don't think echo "$line" should use placeholder syntax. But this is an interesting use case indeed -- should we distinguish concrete examples like we distinguish placeholders? Perhaps highlight them to differentiate from the command's own structure, but in a different way than generic tokens? Calling @igorshubovych @ostera @rprieto @felixonmars @rubenvereecken

pages/common/if.md
denis-sokolov: Thank you! My editor added it. Fixed.

pages/common/while.md
```diff
@@ -0,0 +1,11 @@
...
+
+- Execute a command forever once every second.
+
+`while :; do {{cmd}}; sleep 1; done`
```
waldyrious: Please expand {{cmd}} to {{command}} here too.

denis-sokolov: Done.

pages/common/while.md
igorshubovych: I would not overcomplicate it now. It should be just
```md
`while read line; do echo "$line"; done`
```

denis-sokolov: Done.

igorshubovych:
LGTM

pages/common/while.md
rprieto: FWIW, I wouldn't introduce a new syntax to differentiate between examples and tokens, I feel it would over-complicate things. I'm quite happy with either
- `while read line; do {{echo "$line"}}; done`
- `while read line; do echo "$line"; done`
- `while read line; do {{command using $line}}; done`  to be more generic

rprieto:
Looks great, I really like the examples :+1:

PR #606, opened 2016-01-05
Add notify-send

rakeshdas1:


pages/linux/notify-send.md
```diff
@@ -0,0 +1,15 @@
+# Notify-send
+
+Uses the current desktop environment's notification system to create a notification. This should work on many modern desktop environments such as XFCE, GNOME, and Cinnamon.
```
igorshubovych: IMO, sentence "This should work on many modern desktop environments such as XFCE, GNOME, and Cinnamon." should not be here. If users are looking for it, then they have this command.

pages/linux/notify-send.md
```diff
@@ -0,0 +1,15 @@
+# Notify-send
```
igorshubovych: Should be lowercase

igorshubovych:
Plz squash commits.

rubenvereecken:
Alright @rakeshdas1 could you please squash your commits? Too many of them just say the same.

rubenvereecken:
Okay so apparently it's really not easy to pull in just a single commit and I still want to give you credit so let's try this once more @rakeshdas1, I'll be nice :). Just copy the page as it currently is and commit it onto a clean branch. Just start fresh from master, paste it there. I don't even care if it's squashed.

Since it's too hard to pull in just that commit I want we'll have to submit our own version of this :( And this command is really useful.

PR #608, opened 2016-01-05
Add tldr client for Emacs.

kuanyui:


igorshubovych:
Thank you! This might be really helpful for those who live in Emacs Operating System.

kuanyui:
XDDDDDDD

PR #609, opened 2016-01-05
ln: remove note about hard links for directories

denis-sokolov:
Hard links are generally not allowed for directories.


igorshubovych:
True.
http://askubuntu.com/questions/210741/why-are-hard-links-not-allowed-for-directories

LGTM!

waldyrious:
Thanks for another great contribution! :+1:

PR #610, opened 2016-01-05
ln: remove overwrite option

denis-sokolov:
TL;DR project is meant to contain shortest possible lists of common operations.
`-f` is a corner-case that has a trivial workaround: `unlink && ln`

### DIFF
--- a/pages/common/ln.md
+++ b/pages/common/ln.md
@@ -6,14 +6,6 @@
 
 `ln -s {{path/to/original/file}} {{path/to/link}}`
 
-- overwrite a symbolic link to a file
-
-`ln -sf {{path/to/new/original/file}} {{path/to/file/link}}`
-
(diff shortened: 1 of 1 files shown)


waldyrious:
I think 4 examples can hardly be considered a lot, and I don't think `-f` is such a corner case. But let's wait for others to comment :)

denis-sokolov:
No problem. Opinion poll time!

igorshubovych:
@denis-sokolov 

I love the fact that somebody finally started removing the stuff, not adding.

But I would have to agree with @waldyrious, to me `-f` option is natural. The same option with the similar meaning (`--force`) appears in `rm`, `cp`, `mv`, `git push`. And this simplifies the remembering.

denis-sokolov:
Great! I'm happy to see other people do use `-f`. Alright, let's move on to #612, then.

PR #289, opened 2015-08-10
reword to prevent ambiguity

waldyrious:
"original" can be associated with the origin of the link and lead to confusion if not read carefully. Using "target" instead prevents this ambiguity.

Btw, the third example seems to have been copied from the prior example but not fully adapted: "original/file" should have said (if I'm reading this correctly) "original/folder".

Also, I removed `/file` from the "file or folder" examples, to make it more generic.

PR #612, opened 2016-01-05
ln: remove overwrite option for directories

denis-sokolov:
-T is not present on my system, and the behavior itself is completely unintuitive.
How exactly do you overwrite a directory? Do you recursively delete its children?

### DIFF
--- a/pages/common/ln.md
+++ b/pages/common/ln.md
@@ -10,10 +10,6 @@
 
 `ln -sf {{path/to/new/original/file}} {{path/to/file/link}}`
 
-- overwrite a symbolic link to a folder
-
-`ln -sfT {{path/to/new/original/file}} {{path/to/folder/link}}`
-
 - create a hard link to a file or folder
 
 `ln {{path/to/original/file}} {{path/to/link}}`


igorshubovych:
That's interesting `-F` is present in OSX. Do you use Linux?

To me it does not look useful. So I agree with you here. I would love to hear others' opinions.

denis-sokolov:
`-F`? It's `-T` in our current code. So it's a typo, then. I do have `-F`.

rubenvereecken:
I have `-T` on my Linux and I had to use it a couple of days ago because my symlink wouldn't overwrite. Apparently that was because it was a symlink to a _directory_, not a file. The `-T` fixed it for me. But to be honest, I'd never find it given just this description and even now I had to think hard about what it does.

igorshubovych:
Sorry, `-T`. Yes, I do not have `-T` option on OSX.

igorshubovych:
Thanks

waldyrious:
After merging #289, I'm thinking this one was too hastily merged. The new description with that PR would have been:

> Overwrite an existing symbolic to point to a different folder:
> `ln -sfT {{path/to/new/folder}} {{path/to/folder/symlink}}`

Isn't that clear enough, and useful enough, to warrant its presence? In what systems doesn't this option exist? @rubenvereecken

denis-sokolov:
@waldyrious, that description is still confusing. How is overwriting a symlink by pointing it to a file different from overwriting a symlink by pointing it to a directory?

If that makes sense and I miss it, how useful is this command, how popular is such a use case?

Also, two comments above you can see `-T` option is missing on OSX.

waldyrious:
@denis-sokolov it's no different, hence the reasoning that if one is considered a valid application of the command, the other should too. A single example would suffice if the folder usage merely implied changing the first parameter from `path/to/new/file` to `path/to/new/folder`, but if the -T is needed, I would say it's useful to have it as a separate example. I hope that makes sense.

As for `-T` not existing on OSX (btw, did I get it right that -F has the same effect on OS X?), then we should probably duplicate this page under linux/ and add the -T example, and also on osx/ adding the -F example (even though that creates the duplication/syncing problem we've discussed elsewhere, /cc @rprieto).

denis-sokolov:
Either I misunderstand the application, or it does not work that way. `ln` creates a symlink _inside_ the linked directory:

```
$ mkdir test1 test2
$ ln -s test1 mylink
$ ln -sfF test2 mylink
$ ls -l mylink
lrwxr-xr-x mylink -> test1
$ ls mylink/
test2
$ ls test1/
test2
```

waldyrious:
@rubenvereecken as the only one who's mentioned having used ln in this particular way, can you chip in?

rubenvereecken:
I was trying to avoid this particular conversation, but alas I am uncovered.

Before looking at your discussion, a description of the combination `-sfT` would be "overwrite a symlink to a folder", because that was what I queried Google for and finally found my answer on StackOverflow to. 

I don't know what `-F` does.

rubenvereecken:
Oh wait that was the original page. @denis-sokolov I think the confusion here is caused by a discrepancy between the platform versions? I suggest we recreate this for linux.

PR #613, opened 2016-01-05
added a page for the read builtin

Cvetomird91:
Added a page for the read builtin of BASH shell. I mentioned in the description that this is the read for this specific shell, as the options of read in zsh and fish differ from those in BASH.

### DIFF
--- a/pages/common/read.md
+++ b/pages/common/read.md
@@ -2,7 +2,6 @@
 
 > BASH builtin for retrieving data from standard input
 
-<<<<<<< HEAD
 - Store data that you type from the keyboard:
 
 `read {{variable}}`
--- a/pages/common/read.md
+++ b/pages/common/read.md
@@ -20,6 +20,4 @@
 
 - Use a specific character as a delimiter instead of a new line:
 
-- Store data that you type from the keyboard:
-
-`read -d {{delimiter}} {{variable}}`
+`read -d {{new_delimiter}} {{variable}}`
--- a/pages/common/read.md
+++ b/pages/common/read.md
@@ -1,6 +1,6 @@
 # read
 
-> BASH builtin for retrieving data from standard input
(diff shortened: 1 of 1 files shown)


igorshubovych:
Thanks. It works in Zsh too, with some of these options working.
http://zsh.sourceforge.net/Doc/Release/Shell-Builtin-Commands.html

Maybe it makes sense to combine.

Cvetomird91:
Should I add a notice at the end of the descriptions of the BASH specific options? How should I proceed?

igorshubovych:
@Cvetomird91 

Did you use all of these options on daily basics. I would suggest to cut some options to those working in both Bash and Zsh. The fact that we accept 5-6 examples does not mean the page has to have 5-6 examples. Less is better in this case.

igorshubovych:
@Cvetomird91 also could you please tune syntax according to recent changes in our [contributing guidelines](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#markdown-format). Notice colons and capital letters.

rubenvereecken:
@Cvetomird91 you can use our automatic formatting to fix your linting woes, try `npm install tldr-lint` and then `tldr tldrl` to figure out how to go from there.

Cvetomird91:
Thank you for creating the linter for the project. I will run it and update this PR accordingly shortly today.

igorshubovych:
@Cvetomird91 let us know if you have any problem with it

Cvetomird91:
Thank you for your patience. 

I will resolve this shortly. 

I checked and all of these options for read under zsh appear not to be present.

I decided to leave those. I will squash the commits and update you today.

igorshubovych:
Hey. No need to squash commits anymore. Thank you! for your contribution!

PR #450, opened 2015-12-29
adds commands to xargs and moves to pages/common

audibleblink:
- xargs is also native to OS X, moved to pages/common

PR #615, opened 2016-01-05
xargs: add page

cole-brokamp:


waldyrious:
See #165 and #450

cole-brokamp:
Sorry about that! Didn't see it because it was listed in the linux folder when its actually a common command to linux and unix.  I've moved it to the right place and added two more examples.

waldyrious:
Ok, thanks. I'd suggest waiting until #450 is merged and then updating this as needed. Meanwhile, you're welcome in the discussion at #450 :)

igorshubovych:
These are more examples about `find`, not `xargs`. So they can be on `find`'s page, but should not be added here.

aicioara:
@igorshubovych If this is a reject, let's close the PR and / or look into https://github.com/tldr-pages/tldr/pull/656 as well.

Thank you

igorshubovych:
Does anybody else disagree with me? Opinions?
@waldyrious do you agree with me on that?

waldyrious:
xargs will always require other commands to be properly explain, but yeah, we would probably want to use simpler commands like ls, echo, etc., if at all possible.

igorshubovych:
@cole-brokamp 
Thanks for your contribution. At the moment we cannot merge this PR, because it is different variants of `find`, not `xargs`.

PR #616, opened 2016-01-06
change signal explanation

lanzhiwang:
add common kill signal

### DIFF
--- a/pages/common/kill.md
+++ b/pages/common/kill.md
@@ -15,7 +15,7 @@
 
 `kill [-1][-SIGHUP] {{process_id}}`
 
-- [ctrl]-c
+- interrupt, program can handle SIGINT
 
 `kill [-2][-SIGINT] {{process_id}}`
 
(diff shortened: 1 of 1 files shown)


igorshubovych:
I like it. But I think it deserve a place at the bottom of the page, not at the top.

markusthoemmes:
I'ld suggest adding some explanation for CTRL-C (interrupt, program can handle SIGINT) and CTRL-D (signaling end of file)

waldyrious:
@markusthoemmes :+1:

rubenvereecken:
Alright we'll have to figure something out here because the page has simply become too long.

ostera:
I don't think we have to explain signals. It's good enough to show how to send a custom signal, but it's definitely out of the scope to explain what signals are, which ones are available, and what do they do. I'd rather close and only add `kill -{{signal_number}} {{process_id}}` instead.

waldyrious:
@ostera I agree, but it would be nice to provide a reference to e.g. the man page of this (or another) command that explains signals; what do you think?

PR #618, opened 2016-01-06
ssh-copy-id: add page

forsigner:
Add ssh-copy-id support.

### DIFF
--- a/pages/common/ssh-copy-id.md
+++ b/pages/common/ssh-copy-id.md
@@ -1,0 +1,11 @@
+# ssh-copy-id
+
+> Use locally available keys to authorise logins on a remote machine
+
+- Copy local key to remote machine
+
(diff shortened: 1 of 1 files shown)


igorshubovych:
https://github.com/tldr-pages/tldr/blob/master/pages/linux/ssh-copy-id.md

forsigner:
But there is no Mac support

igorshubovych:
Because it is not available on OSX.
You can install it on OSX though in brew.

waldyrious:
@forsigner feel free to improve the linux page with the parts of your proposal that you feel are better, though :)

forsigner:
@waldyrious  the page of ssh-copy-id.md in linux is nice,  but I think it should be moved from linux to common.

ostera:
This is a duplicate! It's about the same as the linux one. Closing.

PR #621, opened 2016-01-06
curl: fix example according to style guide.

jamapag:
According user contributing guidelines: User-provided values should use the {{token}} syntax.


igorshubovych:
Great! Thanks!

PR #625, opened 2016-01-06
Bmaptool

85pando:
Bmaptool is an improvement over dd as it tries not to copy as many zeros.

### DIFF
--- a/pages/common/bmaptool.md
+++ b/pages/common/bmaptool.md
@@ -1,0 +1,19 @@
+# bmaptool
+
+> Create or Copy blockmaps intelligently (and therefore faster than `cp` or `dd`).
+
+- Create a blockmap from image file.
+
+`bmaptool create -o {{blockmap.bmap}} {{source.img}}`
+
+- Copy an image file into sdb.
+
+`bmaptool copy --bmap {{blockmap.bmap}} {{source.img}} {{/dev/sdb}}`
+
+- Copy a compressed image file into sdb.
+
+`bmaptool copy --bmap {{blockmap.bmap}} {{source.img.gz}} {{/dev/sdb}}`
+
+- Copy an image file into sdb without using a blockmap.
+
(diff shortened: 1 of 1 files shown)


pages/common/bmaptool.md
```diff
@@ -0,0 +1,15 @@
+# bmaptool
+
+> Create or Copy blockmaps intelligently (and therefore faster than `cp` or `dd`).
+
+- Create a blockmap into sdb1.
+
+`bmaptool create -o {{/dev/sdb1}}`
```
felixonmars: I didn't use the tool itself, but as the example on its documentation `bmaptool create -o image.bmap image.raw`, I don't think the command here is correct (missing the image itself). Also I don't understand why one would want to create a blockmap *into* a single partition?

```
$ bmaptool create -o /dev/null
usage: bmaptool create [-h] [-o OUTPUT] [--no-checksum] image
bmaptool create: error: too few arguments
```

pages/common/bmaptool.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Copy an image file into sdb1.
+
+`bmaptool copy --nobmap {{source.img}} {{/dev/sdv1}}`
+
+- Copy a compressed image file into sdb1.
+
+`bmaptool copy --nobmap {{source.img.gz}} {{/dev/sdb1}}`
```
felixonmars: Both of the two `copy` examples here have `--nobmap` specified, which would be somewhat defeating the purpose of the tool. Since you already provided an example for generating the bmap file, how about defaulting to specify it, and optionally add `--nobmap` as another example?

85pando:
Yeah, you are right, I was probably quite tired, when I did this…

felixonmars:
:+1: Looks good now.

rubenvereecken:
Looking good, thanks for another great contribution @85pando!

PR #626, opened 2016-01-06
tcpdump: fixed syntax error and added better params

agnivade:
- Fixed the syntax error in 'dest'.
- Changed the port filter to differentiate between source and destination ports. This gives a better idea to the user if he/she wants to filter on a particular source or destination port.


igorshubovych:
LGTM

felixonmars:
:+1:

PR #627, opened 2016-01-06
Add pass command

naps62:
this is a tool I use to safely store & retrieve passwords & other sensitive data, while keeping them shared across my team via a GitHub repository

http://www.passwordstore.org/

I'm not sure if this makes sense here, since it's not a standard command, but rather a tool. But I use this a lot, so I figured I'd give it a shot

### DIFF
--- a/pages/common/pass.md
+++ b/pages/common/pass.md
@@ -1,0 +1,28 @@
+# pass
+
+> safely store and read passwords or other sensitive data easily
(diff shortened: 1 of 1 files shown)


igorshubovych:
Interesting tool. LGTM.

waldyrious:
Ohai @naps62 :)

LGTM -- even though it isn't a standard Unix utility, it's available in several package managers so that's fine by me.

PR #291, opened 2015-08-18
Add curl example for printing response headers

jeremyruppel:
This is probably the most useful curl command I can never remember. Thanks for the great project!

PR #631, opened 2016-01-07
curl: add --header example

jamescarlos:
Add curl header example.

### DIFF
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -23,6 +23,10 @@
 
 `curl --head {{http://localhost}}`
 
+- include an extra header
+
+`curl --header "{{X-MyHeader: 123}}" {{http://localhost}}`
+
 - pass a user name and password for server authentication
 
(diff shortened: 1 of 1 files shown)


igorshubovych:
See #291

felixonmars:
I would say +1 for this one, since `-H, --header` is a very useful (and often used, for me) option, and it's different from `--head` :)

However I would suggest to rephrase the description a bit. How about `include an extra header` or `include a custom header`?

igorshubovych:
Agree with @felixonmars that it is good add yet one header to show the syntax in this case.

jamescarlos:
Thanks all, I have changed the description to `include an extra header`

igorshubovych:
@jamescarlos could you plz squash commits? don't worry about merge conflict, I will merge it. But plz squash commits.

jamescarlos:
@igorshubovych commits have been squashed. thanks!

igorshubovych:
Thanks!

PR #634, opened 2016-01-07
Edited CONTRIBUTING.md to reflect final convention.

rubenvereecken:
I changed the tldr page example in `CONTRIBUTING.md`, the starting point for most of our contributors. This way we should have some nicely new-looking pages in no time.

I also took ahold of the opportunity to add examples that demonstrate the naming convention that we've been using, then explained it afterwards. Happy to debate this bit, but I feel we've already done a fair bit of discussing on that. If we do, I'll split this PR and merge in the first bit. It's urgent.


igorshubovych:
LGTM!

ostera:
:+1:

PR #639, opened 2016-01-08
Add AutoSSH

notpeter:
Add autoSSH.

This one is perhaps a little overwhelming, especially since we're mostly documenting SSH flags for port forwarding. But as someone who supports hundreds of customer AutoSSH tunnels at work, the final command is (crazily enough) the minimal sane autossh command for a persistent tunnel.

### DIFF
--- a/pages/common/autossh.md
+++ b/pages/common/autossh.md
@@ -1,0 +1,28 @@
+# autossh
+
+> Runs, monitors and restarts SSH connections.
+> Auto-reconnects to keep port forwarding tunnels up. Accepts all ssh flags.
+
+- Open an SSH session, restarting when a monitoring port fails return data.
+
+`autossh -M {{monitor_port}} {{ssh_command}}`
+
+- Open an SSH session which forwards a local port to a remote one, restarting if necessary.
+
+`autossh -M {{monitor_port}} -L {{local_port}}:localhost:{{remote_port}} {{user}}@{{host}}`
+
+- Fork before executing ssh (runs in the background) and don't open a remote shell.
+
+`autossh -f -M {{monitor_port}} -N {{ssh_command}}`
+
+- Run autossh in the background, with no monitoring port instead relying on SSH keep-alives every 10secs to detect failure.
+
+`autossh -f -M 0 -N -o "ServerAliveInterval 10" -o "ServerAliveCountMax 3"  {{ssh_command}}`
+
+- Run autossh in the background, with no monitoring port, no remote shell, exiting if the port forward fails.
+
+`autossh -f -M 0 -N -o "ServerAliveInterval 10" -o "ServerAliveCountMax 3" -o ExitOnForwardFailure=yes -L {{local_port}}:localhost:{{remote_port}} {{user}}@{{host}}`
+
+- Run autossh in the background with debug output logged to a file, ssh verbose output logged to a second file.
+
+`AUTOSSH_DEBUG=1 AUTOSSH_LOGFILE={{log_file}} autossh -f -M {{monitor_port}} -v -E {{ssh_logfile}} {{ssh_command}}`


waldyrious:
Any chance you can tweak the descriptions to use words corresponding to the option names? It's ok if you don't use the "official" words from which the letters were derived.

notpeter:
I rewrote the descriptions so they match the words which the flags are derived from, it definitely helps the readability of long commands.  Also added verbosity examples.

pages/common/autossh.md
```diff
@@ -0,0 +1,21 @@
...
+`autossh -M {{monitor_port}} {{ssh_command}}`
+`autossh -M {{monitor_port}} -L {{local_port}}:localhost:{{remote_port}} {{user}}@{{host}}`
+
+- Fork before executing ssh (runs in the background), no remote shell.
+
+`autossh -f -M {{monitor_port}} -N {{ssh_command}}`
+
+- Log at maximum verbosity to a file from autossh running in the background.
+
+`AUTOSSH_LOGLEVEL=7 AUTOSSH_LOGFILE={{log_file}} -f -M {{monitor_port}} -N {{ssh_command}}
```
waldyrious: Missing the closing backtick here

notpeter:
Fixed the missing backtick.  Might be helpful to add a [markdown linter](https://github.com/mivok/markdownlint) to travis.

pages/common/autossh.md
```diff
@@ -0,0 +1,21 @@
+# autossh
+
+> Runs, monitors and restarts SSH connections.
+> Auto-reconnects to keep port forwarding tunnels up. Accepts all ssh flags.
```
waldyrious: maybe port-forwarding (with a hyphen)?

pages/common/autossh.md
```diff
@@ -0,0 +1,21 @@
...
+
+- Fork before executing ssh (runs in the background), no remote shell.
+
+`autossh -f -M {{monitor_port}} -N {{ssh_command}}`
+
+- Log at maximum verbosity to a file from autossh running in the background.
+
+`AUTOSSH_LOGLEVEL=7 AUTOSSH_LOGFILE={{log_file}} -f -M {{monitor_port}} -N {{ssh_command}}`
+
+- No monitor port, vebose ssh output, no remote shell, require a successful local port forward, send keep-alives every 10 secs and disconnecting if 3 in a row fail.
```
waldyrious: You can split this example if it would still make sense. 5-6 examples is quite acceptable, and it's preferable to introduce options gradually to reduce cognitive load and make reading tldr pages a simple and straightforward experience.

pages/common/autossh.md
notpeter: SSH docs never hyphenate it.

pages/common/autossh.md
notpeter: I can, but my intention was not to introduce log-level and the logfile individually, but to suggest a command sufficient for troubleshooting.  The default behavior logs only to syslog but in the unix tradition is basically silent unless there's actually an error so if you're troubleshooting you'll likely need both.  If I were to introduce a sixth example I would sooner include yet another SSH flag for writing SSH to a logfile instead of stderr but was trying really hard to highlight autossh options, not ssh ones.

pages/common/autossh.md
```diff
@@ -0,0 +1,21 @@
+# autossh
+
+> Runs, monitors and restarts SSH connections.
+> Auto-reconnects to keep port forwarding tunnels up. Accepts all ssh flags.
+
+- Open an SSH connection with port forward monitoring and forward a local port to a remote one.
+
+`autossh -M {{monitor_port}} {{ssh_command}}`
+`autossh -M {{monitor_port}} -L {{local_port}}:localhost:{{remote_port}} {{user}}@{{host}}`
```
igorshubovych: I would split it.

igorshubovych:
@waldyrious @notpeter we are using `mdl` for linting Markdown already.
You can check it out in [Makefile](https://github.com/tldr-pages/tldr/blob/master/Makefile#L34)

pages/common/autossh.md
waldyrious: Fair enough.

waldyrious:
@igorshubovych shouldn't it have caught the missing backtick error then?

pages/common/autossh.md
waldyrious: Ok, I suspected this example needed to be this complex to be useful. @igorshubovych @rprieto any thought about the complexity of this example?

igorshubovych:
@waldyrious no. because it is a valid Markdown. @rubenvereecken's linter should do more.

pages/common/autossh.md
igorshubovych: @notpeter 
Is there any way to split it?

I understand that it will not be perfect according to our recommendations (1 new option per example), but you are introducing 3 new options (`-v`, `-N`, `-o`).

pages/common/autossh.md
notpeter: I can split it, especially if the double syntax isn't supported.

My logic was that I wanted to not introduce both autoSSH and SSH flags on the same command line.  Specifically, I wanted to make sure the commands paired with the first example actually let you do something.  If I split it to a second entry, the first example requires intricate knowledge of SSH flags before its useful.

pages/common/autossh.md
notpeter: I'll reorder the -v so the autossh flags are grouped together ("-M 0, -N"), but -N is present in the two earlier commands.

The reason I introduced -v here is so that I only ever introduced new autossh OR ssh flags on a given command line.

notpeter:
I replied to your [line level comments](https://github.com/tldr-pages/tldr/pull/639/files#diff-18628552838370c355f61dce6e62564b) inline, but wanted to link there for when I push new changes.

So:
1) What's the hard cap on number of commands, 6?
2) If two command examples aren't supported I can certainly reword my non-compliant first example.  Contributing.md should probably also mention this. At one point, I believe there were double examples in some pages, but I could be mistaken.

notpeter:
Alright, hopefully this will do. For what it's worth, this discussion has definitely led to a better page (I think) and so I appreciate both the feedback and the commitment to ensure tldr examples as clear, concise and consistent as possible.

If you've got more notes, let er rip. ;)

waldyrious:
1) It's not a hard cap, more of a guideline, but yes, 5-6 is the recommended per the contribution guidelines. More is ok if the command is exceptionally complex, otherwise it's better to combine or remove examples.
2) There were indeed pages with more than one commands per example -- we're trying to get rid of them. Unfortunately we hadn't settled down on precise guidelines due to the relatively low activity level of the project. The recent flurry of edits required us to do so, in order to allow contributors to get consistent feedback, and have a place to look for the expected format for new contributions. Sorry you were one of those caught in the middle of this growth period :)

pages/common/autossh.md
waldyrious: As I said above, contriving the examples for the sake of simplicity to the point they're no longer useful is probably not a good idea. We should strive for a compromise between usefulness and approachability. Unfortunately I'm not familiar with this tool in order to provide more useful (specific) feedback :/

waldyrious:
By the way, feel free to submit PRs updating the contributing guidelines in ways you feel makes some of the ad-hoc guidelines more explicit and visible :)

igorshubovych:
Thanks. I like it much more. LGTM!

waldyrious:
Thanks for your patience @notpeter :)

PR #641, opened 2016-01-08
glances: add page

chuanconggao:


pages/common/glances.md
```diff
@@ -0,0 +1,23 @@
...
+
+- run in terminal.
+
+`glances`
+
+- run in web server mode.
+
+`glances -w`
+
+- run in server mode.
```
waldyrious: It's unclear what's the difference between this option and the above. Do you think you could clarify it?

igorshubovych:
Here is the link to site: https://github.com/nicolargo/glances

igorshubovych:
LGTM

PR #645, opened 2016-01-08
Add ltrace

TerkaSlan:


waldyrious:
Thanks for the contribution! Please use the token syntax {{}} for the placeholders in the examples.

pages/linux/ltrace.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Count library calls. Print a handy summary at the bottom.
+
+`ltrace -c {{/usr/bin/who}}`
+
+- Trace calls to malloc and free, omit those done by libc
+
+`ltrace -e malloc+free-@libc.so* {{/usr/bin/who}}`
+
+- Write to `{{file}}` instead of terminal.
```
igorshubovych: What kind of file is this? Can it be called something like {{output.trace}}?

pages/linux/ltrace.md
```diff
@@ -0,0 +1,19 @@
+# ltrace
+
+> Display dynamic library calls of a process
+
+- Print (trace) library calls of a binary file `{{program}}`.
+
+`ltrace {{./program}}`
```
waldyrious: The ./ isn't part of the placeholder, since it will be present for any program you'd want to run this way.

waldyrious:
Perhaps instead of {{/usr/bin/who}} you could use {{/path/to/program}} which makes it clearer what the parameter is.

TerkaSlan:
Apologies for my fuzzy formating and naming, will repair !

waldyrious:
On the contrary, sorry if we're being too bothersome :)

By the way, I'm not sure @igorshubovych was suggesting to name the parameter output.trace. The way I read his comment was he was asking whether that file could be anything (i.e. a text file) or if it needed to be in a specific format. @igorshubovych can you clarify? I would personally prefer the original parameter name, {{file}}, since that's more generic, unless there's a reason to prefer output.trace (i.e. that's a standard name for such files).

igorshubovych:
@waldyrious @TerkaSlan 
Exactly. I was just asking. I don't know `strace`. So I was curious, is there a way to help user identify the output result by looking at example.
Check out our [Contributing Guidelines about tokens](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#token-syntax).

TerkaSlan:
@igorshubovych Oh, sorry for the misunderstanding, `file` is just a plain text file where the result of ltrace is dumped instead of terminal.

waldyrious:
Looks :ok_hand: to me!

felixonmars:
:+1: Please squash the commits, thanks :)

igorshubovych:
@TerkaSlan see https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#submitting-a-pull-request, number 9.

igorshubovych:
@TerkaSlan can I offer my support with finishing this?

igorshubovych:
I committed it manually.

Thank you, @TerkaSlan! Keep up the good work!

igorshubovych:
Here is the commit in case you wonder:
https://github.com/tldr-pages/tldr/commit/5b3660214930332a37d485b4e3ac83d911d70635

waldyrious:
Good thing authorship was preserved. Yay git! :)

igorshubovych:
BTW, congratulations!
https://twitter.com/firstprbot/status/685409859873992704

PR #651, opened 2016-01-10
nginx:add stop,quit,reopen,reload

WALL-E:
Send a signal to the Nginx master process, stop,quit,reopen,reload.

### DIFF
--- a/pages/common/nginx.md
+++ b/pages/common/nginx.md
@@ -17,3 +17,19 @@
 - Test configuration without affecting the running server:
 
 `nginx -t`
+
+- Reload configuration, gracefully shut down old worker processes
+
+`nginx -s reload`
+
+- Shut down quickly
+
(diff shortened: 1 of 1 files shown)


waldyrious:
Do you think you could shorten the page to have 5-6 examples?

notpeter:
Is sending signals to nginx using `nginx -s` a common use? Don't most people just send signals directly or use init scripts?

ostera:
@notpeter, @waldyrious, as a heavy nginx user I've never had the need to directly use -s for anything other than the very very sporadic testing of a configuration reload.

I've merged and left that sample since is the most useful, the rest can be achieved by using the service scripts. Changes here: https://github.com/tldr-pages/tldr/commit/b7bca90124e8a3938452659fffa8c75fbe7b38b7

notpeter:
Looks great!

PR #652, opened 2016-01-10
Adds not-found page

ostera:
In order to provide a single, consistent not-found message across all clients, we could leverage a not-found page.

There's mainly two things to discuss here:

- [ ] the content of such page
- [ ] the location 

If it was part of the index it'd be very easy to implement since the fallback command would be the not found one. Semantically would be a bit wrong perhaps? But it should fit right in with the current pipeline that packages the rest of the pages.

### DIFF
--- a/pages/not-found.md
+++ b/pages/not-found.md
@@ -1,0 +1,11 @@
+
+# Oops! :(
+
+## Command Not Found
+
+It seems the command you were looking for is not available yet.
+
+Contributing is simple! All done from within Github. Please take a look at our [Contributing Guidelines](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md); then [Add this page](https://github.com/tldr-pages/tldr/new/master/pages).
+
+Questions? [Chat with us on gitter](https://gitter.im/tldr-pages/tldr)!
+


igorshubovych:
Did you try to `--render` this file with existing clients?

muteness:
:thumbsup:

ostera:
@igorshubovych yes, I just did and I found out that it didn't work properly. Filed issue addressing it on the tldr-node repo.

Other than that, any thoughts? @waldyrious @rubenvereecken @felixonmars @rprieto

felixonmars:
How about "guidelines" => "Guidelines"?

Other than that, looks nice to me :+1:

waldyrious:
Maybe we could put the link to the guidelines before the link to add the page? Otherwise it may be overlooked.

rubenvereecken:
It doesn't fit with my linter, hence I don't like you.

ostera:
@rubenvereecken it seems to have failed because ruby couldn't be installed? (for some reason)

rubenvereecken:
Oh don't look at it just yet. And that's _really_ weird.

On a more serious note though, the page would be great. At first I thought you chucked it in with the rest of the pages, or even referred to the `not-found` command (which has been reported missing on my system before?).

The only annoying bit is that we can't do this anymore:
`find pages/* -name ".md" -exec linter {} \;`

ostera:
Hm, aren't we doing this instead? https://github.com/tldr-pages/tldr/blob/master/Makefile#L34

rprieto:
I like the idea of a nice "not found" message, and do get the value of having some consistency between the clients, for example making sure they all mention the Github URL to add pages. I wonder though if it's important to make **everything** exactly the same on all clients, and if we need to couple all clients via a single Markdown file.

For example on the web it makes sense to have a nice formatted page with `<h1>` and `<h2>` headings, then some text, a few clickable URLs... And on the command line we might want something different like a much shorter message (without all the headings) and obviously we'd need to find a way to render Markdown links. For example how do we render `Please take a look at our [Contributing Guidelines](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md); then [Add this page](https://github.com/tldr-pages/tldr/new/master/pages).`? Maybe something like `Please take a look at our Contributing Guidelines (https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md); then Add this page (https://github.com/tldr-pages/tldr/new/master/pages).` It feels though that this might need custom Markdown processing that will look good with the current exact `not-found.md`, but might look weird if we make changes to the page.

Personally I'm in favour of letting each client do its thing. If we keep the number of "official" clients down it should be easy to keep an eye once in a while and submit PRs to suggest "please add the URL for people to add pages", etc.

waldyrious:
I'm inclined to agree with @rprieto.

hidroh:
Instead of putting this as another command page, maybe adding a pseudo not found page as a reference for all clients and mention it in CONTRIBUTING.md or README.md would be more flexible? After all it's a static informative page, and not meant to be interpreted as another command.

Of course the drawback would be to inform all clients to update their not found page if something changes, says the contributing link.

ostera:
We could still use the base and any client could override it as they see fit (or format it in whichever way it's best for their platform's and communities style of output).

I'd like to get this merged so all clients can fallback to it easily.

igorshubovych:
@ostera, i would rather agree with @rprieto.

I think we can just add this requirement to https://github.com/tldr-pages/tldr/wiki/Minimum-specifications-for-TLDR-command-line-clients

waldyrious:
:+1: @igorshubovych. Ideally we would be able to have the whole spec for a client in the tldr page for tldr itself (see #744), but it seems some items will have to be specified elsewhere.

ostera:
Okay then! This seems to not be going to happen, so I might as well close the PR.

PR #653, opened 2016-01-11
uniq: add page

muteness:
added page for uniq command

PR #654, opened 2016-01-12
uniq: add page

muteness:
added a new page for **uniq** command.

See discussion in #653.

### DIFF
--- a/pages/common/uniq.md
+++ b/pages/common/uniq.md
@@ -1,0 +1,19 @@
+# uniq
+
+> Report or omit repeated lines.
+
+- Display each line once:
+
+`uniq {{file}}`
+
+- Display only unique lines:
+
+`uniq -u {{file}}`
+
+- Display only duplicate lines:
+
+`uniq -d {{file}}`
+
+- Display number of occurences of each line along with that line:
+
+`uniq -c {{file}}`


igorshubovych:
This is related to #653

muteness:
That PR is closed because of some git issues. I opened this new PR instead.

igorshubovych:
@muteness yes, i know. I just let other reviewers to know.

pages/common/uniq.md
```diff
@@ -0,0 +1,19 @@
+# uniq
+
+> Report or omit repeated lines.
+
+- Display each line once.
+
+`uniq {{file}}`
+
+- Display occurence of each line along with that line.
```
rubenvereecken: occurrence

rubenvereecken:
Just got the one inline comment. That and we recently switched to a different syntax for example descriptions. Really sentences just end in colons instead of periods. Have a look at [any random page](https://github.com/tldr-pages/tldr/blob/master/pages/common/ack.md)

pages/common/uniq.md
waldyrious: :+1: @rubenvereecken, I was thinking the same thing.

muteness:
Just fixed the two issues.

igorshubovych:
LGreatTM

waldyrious:
Doesn't `-c` imply `-u`? Maybe it should appear after it.

muteness:
I don't quite understand your logic here.

And I think they are used for different purposes.

rubenvereecken:
Alright I'll go with @waldyrious on this one: the very first use case for `uniq` would be to `uniqueify` a file or other, then for counting occurrences. It'd be great if you could switch 1st and 3rd example.

muteness:
**uniq {{file}} ** displays each line once which means if one line repeatedly appears in a file, the output just shows that line once.

waldyrious:
Oh yeah, I was mixing it up with uniq -u, nevermind. LGTM, thanks for the contribution!

PR #655, opened 2016-01-12
Add pages for latexmk and pdflatex

85pando:
Pdflatex and latexmk are ywo tools to create Output documents from LaTeX source files.

### DIFF
--- a/pages/common/latexmk.md
+++ b/pages/common/latexmk.md
@@ -1,0 +1,28 @@
+# latexmk
+
+> Compile LaTeX source files into finished documents.
+> Automatically does multiple runs when needed.
+
+- Compile a dvi (DeVice Independent file) document from every source:
+
+`latexmk`
+
+- Compile a dvi document from a specific source file:
+
+`latexmk {{source.tex}}`
+
+- Compile a pdf document:
+
+`latexmk -pdf {{source.tex}}`
+
+- Clean up all temporary tex files in the folder:
+
+`latexmk -c`
+
+- Clean up temporary tex files created for a specific tex file:
+
+`latexmk -c {{source.tex}}`
+
+- Clean up temporary tex and output files:
+
+`latexmk -C`
--- a/pages/common/pdflatex.md
+++ b/pages/common/pdflatex.md
@@ -1,0 +1,11 @@
+# pdflatex
+
+> Compile a pdf document from LaTeX source files.
+
+- Compile a pdf document:
+
+`pdflatex {{source.tex}}`
+
+- Compile a pdf document, halting on each error:
+
+`pdflatex -halt-on-error {{source.tex}}`


pages/common/latexmk.md
```diff
@@ -0,0 +1,24 @@
+# latexmk
+
+> Compile LaTeX source files into finished documents. Does multiple runs when
+> the output will change.
```
waldyrious: 1) Put each sentence into its own line
2) The second sentence isn't clear. Can you rephrase it?

pages/common/latexmk.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Compile a dvi document.
```
waldyrious: Spell out what "dvi" stands for, in parenthesis

pages/common/latexmk.md
```diff
@@ -0,0 +1,24 @@
...
+
+`latexmk {{source.tex}}`
+
+- Compile a pdf document.
+
+`latexmk -pdf {{source.tex}}`
+
+- Clean up temporary tex files created for a single tex file.
```
waldyrious: suggestion: single --> specific

85pando:
* Tried to rephrase the sentence about multiple runs.
* Elongated dvi in parentheses.
* reordered commands.

    * Should I put the `-C` flag also before? I ordered it like this, because in my experience it's not that often that you want to delete your outputfiles.

pages/common/pdflatex.md
```diff
@@ -0,0 +1,11 @@
+# pdflatex
+
+> Compile pdf document from LaTeX source files.
+
+- Compile pdf document:
```
waldyrious: "Compile *a* pdf document" (as in the other command)

pages/common/pdflatex.md
```diff
@@ -0,0 +1,11 @@
...
+
+`pdflatex {{source.tex}}`
+
+- Compile pdf document, halt on each error:
```
waldyrious: "Compile **a** pdf document, halt**ing** on each error"

waldyrious:
Thanks for the improvements, much better now :+1: 

> Should I put the -C flag also before?

No, I think you made a good call :)

Before I can endorse this, could you address the inline comments I left? Also the one about "single" --> "specific", if it was an overlook -- otherwise please explain why you disagree.

85pando:
* Changed the descriptions of the pdf version as suggested.
* Added an additional command with no options (did not know it also does this ☻ )

For the "single" --> "specific" comment, I understood it as a comment on my ordering of the commands of `latexmk -c {{source.tex}}` coming before `latexmk -c` which I then changed.
If adding the command without options was what you meant, all is fine now, otherwise please advise.

waldyrious:
Sorry if I wasn't clear. I meant changing the description from "Clean up temporary tex files created for a **single** tex file:" to "Clean up temporary tex files created for a **specific** tex file:" (although "single" works, I believe "specific" gets the idea across better)

85pando:
Allgright, got you now ☺

This should be it then.

waldyrious:
Looks great! :+1:

rubenvereecken:
Live and learn, thanks a lot @85pand, I could actually use this one :+1:

85pando:
Isn't that the essence of tldr? ☻

PR #658, opened 2016-01-13
Fix all of the syntaxes

rubenvereecken:
Cleaned up what needed cleaning.


igorshubovych:
Thanks, @rubenvereecken
I guess nobody will complain if I merge it, it is just syntax check.

rubenvereecken:
Right I was about to merge this, but you guys beat me every time.

PR #661, opened 2016-01-13
Fix nm syntax

rubenvereecken:


waldyrious:
@rubenvereecken this seems to have eaten the last character of each string and replaced it with a colon.

rubenvereecken:
I'll enjoy unfucking my mess :).

PR #664, opened 2016-01-13
crontab: add page

andrewhavens:


rubenvereecken:
So you went around to writing the crontab page, good on you!

I like the based-on-examples approach, instead of generic commands. In fact, I think we can leverage it to squash the introductory description, since you reuse those symbols you introduce there in the examples. What you could replace it with is something like the default crontab format `m h dom mon dow command`, and even that is up for debate. I'm saying this because it's a tad too long.

I think your examples are good. Try to incorporate months (`feb`, `mar`) in there as well, and they should cover the whole introduction.

igorshubovych:
I think this page should have commands like
- `crontab -e`
- `crontab -l`
- `crontab -r`
- `crontab -u`

You can describe format but in a single line without much details.

igorshubovych:
Any progress here? @andrewhavens could you please reply?

igorshubovych:
@andrewhavens 

You prepared a good explanation. However this page does not follow our [guidelines and recommendations](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md).

I think your text is more suitable for [cheat project](https://github.com/chrisallenlane/cheat), so you can try to put it there.

Thank you for your contribution. Keep up the good work.

PR #665, opened 2016-01-13
pkill: Add example with -f flag

AMeng:
Its useful to be able to kill processes by their full command. The pgrep tldr page already has a -f flag example.


igorshubovych:
LGTM

waldyrious:
Thank you!

PR #671, opened 2016-01-14
`ls` only directory

steelywing:


pages/common/ls.md
```diff
@@ -23,6 +23,10 @@
 `ls {{prefix}}*`
 `ls *{{suffix}}`
 
+- List only directory:
```
waldyrious: Shouldn't this say "directories"?

rubenvereecken:
We recently cut down the `ls` page, it's now at 5 examples, closing in on our cap at 6. I noticed recently we keep reorganizing examples so I'd like to think twice before adding this one. @steelywing if you really think it's useful enough, could you do a rebase on master to fit in with the new page? This can be done by updating your master and then trying `git rebase master` on your `patch-1` branch.

waldyrious:
@rubenvereecken I think we should give PR authors at least a week to react. Or did @steelywing react elsewhere that isn't noted on this thread? Just reading this looks like we simply gave up on him after 4 days.

PR #672, opened 2016-01-14
`rm -rf`

steelywing:


pages/common/rm.md
```diff
@@ -10,6 +10,10 @@
 
 `rm -r {{/path/to/folder}}`
 
+- Remove file/directory without prompt:
```
waldyrious: This shouldn't say "file/directory", since -r is only valid for directories.

steelywing: Um... `-r` can use with file, but doesn't make sense when only with file, but I always use `-rf` for file or directory, because I only need to remember 1 command, or when I use `*`.

waldyrious: Yes but that's providing incorrect information to users. We should encourage people to memorize command options by helping them understand what each option means, not by providing formulas to be recalled mechanically and applied blindly as if they were magic incantations.

igorshubovych:
LGTM

waldyrious:
Thanks!

PR #673, opened 2016-01-14
`netstat -n` and `-r`

steelywing:


pages/linux/netstat.md
```diff
@@ -21,3 +21,7 @@
 - List information continuously:
 
 `netstat -c`
+
+- List routes (-r) without resolve ip to hostname (-n):
```
waldyrious: "without resolving"

pages/linux/netstat.md
```diff
@@ -21,3 +21,7 @@
...
+
+`netstat -nr`
```
waldyrious: How about matching the order that the options are mentioned in the description? I.e. "-rn".

steelywing:
Sorry for my bad english :sweat_smile:

rubenvereecken:
I usually don't like mentioning parameters explicit in the description, rather have it deductible from context. If you follow @waldyrious' last comment I think you can safely drop the `(-r)` and the other one.

waldyrious:
Maybe if it was written "and do not reslolve ip..." even the -n could be removed :smile: 

By the way, IP should be capitalized.

rubenvereecken:
That looks like intuitive enough, thanks for taking care of our comments, and thanks for a great addition to tldr pages!

waldyrious:
:+1:

PR #675, opened 2016-01-15
Add iOS Client for TLDR Man Page

freesuraj:


igorshubovych:
Thank you!

igorshubovych:
I added it also here:
https://github.com/tldr-pages/tldr/wiki/TLDR-clients#mobile-clients

Issue #527, opened 2015-12-31
Simple shell client shipping with the tldr-pages repo


sitaramc:
It would be nice to have a very simple shell client (nothing fancy; just 'less' the md file if found, else print a list of pages that start with the argument supplied).

At present I use https://gist.github.com/sitaramc/142c9e3d730583b9271c

And though I hesitate to offer that in a pull request, if someone thinks it looks decent or requires only minor changes to be decent I can do that.

waldyrious:
+:100: from me, I love the simplicity.

igorshubovych:
If you want, feel free to add it here [in Wiki](https://github.com/tldr-pages/tldr/wiki/TLDR-clients). There is already one bash client.
Or you can create a PR to add it to README.md.

waldyrious:
@igorshubovych the proposal was to include this with the repo itself, i.e. as you clone the tldr-pages/tldr repo, you'd have a simple way to query and display the existing pages. It's a different thing than installing a separate, full-fledged client. I think that's a reasonable proposal and would support its inclusion.

igorshubovych:
Oh, but then I doubt we should do it.

We have some Windows users at the moment, and it will not work for them. They seems like are ok with NodeJS at the moment (seems like Rust client will beat it soon).

If we consider using Bash, i would rather use this one, as it has more features:
https://github.com/raylee/tldr

waldyrious:
Yeah but that's a full-fledged client and wouldn't make sense to be included in this repo, contrary to a 11 SLOC bash script. The point about Windows is a good one, though. @sitaramc what would you suggest?

igorshubovych:
what's the difference between 11 SLOC and 250 SLOC?

Are you trying to save some bytes?

igorshubovych:
Additionally NodeJS client can be added with 1 line in package.json

waldyrious:
It's subjective, sure, but I don't see the bash script as an actual client, so the separation of implementation (the client repos) and content (tldr-pages) would largely remain in place. It would just be a helper script for convenience instead of `cd`ing and `less`ing by hand.

sitaramc:
On 29/01/16 23:36, Waldir Pimenta wrote:
> It's subjective, sure, but I don't see the bash script as an actual client, so the separation of implementation (the client repos) and content (tldr-pages) would largely remain in place. It would just be a helper script for convenience instead of |cd|ing and |less|ing by hand.

That is *precisely* what I meant it as.

[...]

PR #679, opened 2016-01-15
Adding link to Clients wiki page in README.md

igorshubovych:


waldyrious:
Before we merge this, I think we should wrap the discussion about where to place the list of clients. Below are a few relevant quotes from a discussion that took place on gitter a few days ago:

@waldyrious:
> Hey guys, I had an idea regarding the clients that I think may help deal with the proliferation and reliability of clients.
> How about we list in the readme the client *and* the author? In fact, rather than aliasing the link to the repo, we could display it in the form username/repo as is standard on github.
> That would make maintenance responsibility clearer, and give due credit to the client authors 
what do you think?

@igorshubovych:
> I think all clients should be mentioned here:
https://github.com/tldr-pages/tldr/wiki/TLDR-clients
> With or without authors (actually i like your idea about authors)
> And on README - only a few recommended clients

@waldyrious:
> I really think we don't have real-estate problems in the README, and tucking them away like that does a disservice to the community's thriving
> I mean, what's the reason not to keep them in the readme?

@igorshubovych:
> it is not real estate problem, it is an issue of support the code we don’t own.

@waldyrious:
> exactly my point :smile:  if we clearly indicate who the authors/maintainers of the clients are, we clear ourselves of that responsibility 

@igorshubovych:
> @waldyrious can you plz prepare short list of the (maybe for 3-4 clients) of how you see it

@waldyrious:
> Sure 

@ostera:
> I think this is important because it helps to think of how to get people to pick a client faster – shoving N+1 clients in a poorly readable format (be it a bad table or just very unstructured plain text) is probably not a good idea

@ostera:
> There's currently 3 lists of clients:
> https://github.com/tldr-pages/tldr/wiki/TLDR-clients
> https://github.com/tldr-pages/tldr#clients
> https://github.com/tldr-pages/tldr-pages.github.io/blob/master/index.md#cli-installation
> can we DRY this up please?

@igorshubovych:
> well it seems like i will soon remove Wiki page
> but i don’t know what to do with table on site
> maybe link to README

@waldyrious:
> Ok my suggestion is: we can use a list like this in the readme: https://github.com/waldyrious/tldr/wiki/Clients (either as text or as a table)
> and in the website, we list the officially supported ones and point to the readme for more
> 
> Additionally, we can define minimal specs to have a client featured in the readme, such as being published for install using a distribution tool (if it's written in a language that's connected to one, e.g. node.js --> npm, pthon --> pip, etc.)
> clients that require git cloning / compile from source should not be featured IMO

igorshubovych:
@waldyrious,
it is just adding a link, we are not removing clients from README.md. I do not see a purpose in having this discussion now. I think we can just add link, and see how it goes.
IMO some of the clients will die naturally disappear like current Go client and Elixir client.

waldyrious:
I don't think it's wise to keep postponing the discussion. Adding the link to the readme only promotes further confusion as we'd be providing two lists with roughly the same content with no indication of the (intended) difference between them.

igorshubovych:
I think that full, comprehensive list of clients should be in Wiki.

The README.md IMO should have:
- link to clients list in Wiki
- link and short installation instruction of the reference CLI client
- link to reference web-client
- link to reference Android, iOS clients
- link to integrations with other apps like Dash (either to link to Wiki or direct links)

Reasons:
- 95% of users just use TL;DR and do not contribute, they need dead simple client that just works without compilation issues
- it is better to keep README.md short to make it more user-friendly
- there are some almost dead clients (like Go and Elixir clients), and I would like to avoid the situation when newcomers start with them.

I understand that some contributors may go emotional about it. I am maintaining (together with @rprieto) Node.js client. I do NOT mind if it is not a reference client anymore, even though it was historically first. To me this is not that important as growing the amount of users and contributors of the tldr pages database. I recently found that Crystal client has the same features and code is better, unfortunately it cannot be compiled on Windows yet. I also like the Bash client, it is dead simple and does the job.

I think on this stage we can just include link to Wiki page and leave it as is, so we will not hurt anybody’s feelings. In future we can select a single or a few reference clients to mention.

README.md
```diff
@@ -70,6 +70,8 @@ You can access these pages on your computer using one of the following clients:
 
 Let us know if you are building one and we can add it to this list!
 
+There is also a comprehensive [list of clients in Wiki](https://github.com/tldr-pages/tldr/wiki/TLDR-clients).
```
ostera: Redundant with the previous line, might read better as "we can add it to our list of clients in Wiki". The latter part being a link.

Otherwise people will think that they can suggest and be added to the README, whereas in reality they'd be added to the wiki page.

igorshubovych: Do you think I should also remove the previous line?

ostera: Yeah, that's what I meant. I might have not expressed myself correctly :smile:

igorshubovych:
@waldyrious 
It seems like everybody else agrees with this one. Do you still see need to continue discussing this? Are you satisfied with my previous answer?

waldyrious:
I'm not sure what you mean. Who has expressed agreement that this is how we should move forward?

By the way, I understand your reasons, but you didn't address the points I quoted above. Can you do that? I'll address the reasons you posed:

> 95% of users just use TL;DR and do not contribute, they need dead simple client that just works without compilation issues

I don't see how providing a list of clients, with the officially supported ones clearly highlighted, would be incompatible with the statement above. Do you disagree?

> it is better to keep README.md short to make it more user-friendly

I don't think it is too long or that it would become unbearably long with a list of (working) clients. Do you?

> there are some almost dead clients (like Go and Elixir clients), and I would like to avoid the situation when newcomers start with them.

If any client doesn't work well, I'm fully onboard with removing them from the readme until they're restored to a working state. In fact I even proposed that we define a set of minimal functionality that clients should offer, so we can make that assessment objectively. So that's not an issue, we're in agreement here.

igorshubovych:
I really don't understand why we cannot just postpone this discussion. I do not want to hurt anybody's feelings, so I do not remove anything, I just add a link.

You previously wrote.
> Adding the link to the readme only promotes further confusion as we'd be providing two lists with roughly the same content with no indication of the (intended) difference between them.

The wiki page is comprehensive (more clients) and structured. Therefore I am adding link to it. The clients section in README looks like trash now, and it will not become worse if we add this link.

IMO it would be good for end-users to have a few clients: a single CLI client, a single web client, mobile clients for each platform and a link to Wiki page. However IMO we are not ready for this move and here is why:
- we need to choose recommended clients
- they have to be good (none of them is perfect at the moment)
- we have to prepare a set of recommendations for clients
- we need to explain this move to community

Therefore I think we should just add this link to Wiki page.

waldyrious:
> I really don't understand why we cannot just postpone this discussion.

Why should we? If we can't resolve this issue now, what makes us think we will later?

> I do not want to hurt anybody's feelings, so I do not remove anything, I just add a link.

I don't think anyone else is making an emotional argument here. I certainly am not. We shouldn't keep stuff on the readme just to please others. I support keeping them in the readme mostly because it encourages people to use (and contribute to) clients based on environments they're already familiar with (e.g. I'd rather use the node.js client if I already have npm on my machine). Of course that doesn't mean we should keep sub-optimal clients on the readme, so again, we're in agreement in that regard.

Adding a link is not a neutral action; I believe it's detrimental, and @ostera agrees, due to the non-DRY-ness of the result (three separate lists maintained in different places). I would certainly support editing the README list to be more comprehensive and structured, and presented a proposal to do just that **[here](https://github.com/waldyrious/tldr/wiki/Clients)**. Besides making the readme longer, do you have any other objections to doing that?

igorshubovych:
@waldyrious 

> Besides making the readme longer, do you have any other objections to doing that?

I already explained that for the most of the users (and I am one of them), it does not matter the technology behind the CLI client. I do not care if it is written in Python, Ruby, NodeJS, C++, any major programming language you name it. If it supports all the features, if it is good, if it is maintained, then I am sold, I will use it. Somewhen in the future I would like to come to the situation when we have a single or just a few clients for any type of distribution.

> We shouldn't keep stuff on the readme just to please others. I support keeping them in the readme mostly because it encourages people to use (and contribute to) clients based on environments they're already familiar with (e.g. I'd rather use the node.js client if I already have npm on my machine). Of course that doesn't mean we should keep sub-optimal clients on the readme, so again, we're in agreement in that regard -- as long as we clearly identify what's wrong with those clients and what it would take for them to be accepted back on the readme.

Ok. Seems like we have an agreement here. Can you speak more about this? Maybe in form of PR?

waldyrious:
> Ok. Seems like we have an agreement here. Can you speak more about this? Maybe in form of PR?

Sure, but I haven't tried all the clients on the readme. You've mentioned the Elixir one for instance, but what exactly does it do wrong? Same for the others. My thinking is that if we don't have specific complaints about a client, then we shouldn't remove it just because it has been stagnant for a while.

waldyrious:
Responding to your other point (although just for completeness -- not arguing here :)

> I already explained that for the most of the users (and I am one of them), it does not matter the technology behind the CLI client.

That may be true, but even installation ease depends on the technology used. I may not care what is used to implement a client, but if I have to git clone, configure / make / make install, versus just npm install -g, then I'll probably pick the latter if I already have node on my machine. Same for python, rust, go, R, etc. Besides that, there's the small fraction of users who may be encouraged to contribute to a client that uses a language they're familiar with, and in my view attracting this group of people is quite beneficial to the project, especially considering that we can still highlight a couple "main" clients in the same list (bold, larger font, different color, whatever) for those who don't care. So it's win-win!

rubenvereecken:
> That may be true, but even installation ease depends on the technology used.

However little I'm joining this discussion, this is very much true for me at least.

hidroh:
Why not require each client to have a link to its Github project? Installation instructions, setup requirements, issue tracking are responsibilities of each Github project. Each project can optionally provide a ready-to-use link (e.g. website link, app store).

Projects can be simply sorted alphabetically, grouped by platform. We can add a Github star badge next to each project, as a way to show how popular it is. No recommendations needed, let the stars speak for the project.

Our only job now is to do regular check with maintainers on potentially deserted projects before removing them. Where to put this list doesn't really matter, Github pages, readme or wiki. Once will be enough as long as it can be easily found IMO.

rprieto:
This seems like a conversation that should be taken to Gitter. I definitely agree with the following

> That may be true, but even installation ease depends on the technology used. I may not care what is used to implement a client, but if I have to git clone, configure / make / make install, versus just npm install -g, then I'll probably pick the latter if I already have node on my machine. 

However I don't believe having 15 officially supported clients is the way to go. Most widly-distributed projects seem to choose either of the following
1. pick a runtime, and assume consumers will be OK with it. This is the `npm install -g` option, and projects that choose this obviously take an ecosystem with a solid package manager (Node, Python, Ruby...). For example, the official [aws cli](http://docs.aws.amazon.com/cli/latest/userguide/installing.html) is written in Python. Yes, you do need `python` to run it, and yes it's recommended you use `pip` to download it... although you could get it from source too but not sure why you would.
2. write the tool in a compiled language and provide platform binaries. I fully agree that no one should have to `git clone, configure / make / make install` if they don't want to, and platform binaries solve just that. They have the advantage that they don't force the user to install Node or any large ecosystem, can literally be a "right-click save-as" operation for users who want that, and can typically run on any system with no dependencies (e.g. no issue installing on a server).

That being said, I agree having multiple clients can be a good thing. It encourages people to try new things, add new features, try their hand at something new (e.g. `hey tldr is cool, and I want to learn Foobar, so I could try making a client`) etc... However if should be very clear which is the official client and which are related-projects.

For example, imagine if AWS provided a version of their CLI tools in Python, Node, Ruby, R, Elixir, Go, Rust. This would be impossible to maintain, as
- every tool would eventually have different features
- bug reports & feature request would get very confusing (we're already seeing this with tldr when people ask for color themes, local caching, ...)
- people would be confused which tool to get when they're getting started and just want something that works (imagine posts on StackOverflow saying `if you want to auto-scale with option X, use the Python client, but since you also want you want to load config from a JSON file you might have to use the Rust client).
- it would make changing to the API incredibly difficult for them

About contributing, I personally don't see this as an issue. I don't know Python, so I don't contribute to the AWS CLI. If I really wanted to, I'd learn python (yay for learning :smile:), and otherwise there's plenty of other projects I can contribute to.

All of that to say, I'm in favour of keeping a list of all available clients as "other clients / related projects" in the README. I'd just recommend that
- we make it clear there's one main client (officially supported)
- have a best effort to make it easy to install & require minimal dependencies. Sadly Node might not be the best choice in hindsight - I get why people don't necessarily want Node on their servers if they have nothing else that requires it.
- keep an eye on unofficial clients for ideas, features, and how to improve `tldr` as a whole

igorshubovych:
Initial version of recommendations:
https://github.com/tldr-pages/tldr/wiki/Implementation-of-TL;DR-client

waldyrious:
> This seems like a conversation that should be taken to Gitter.

It did happen on Gitter, in fact it took place in (at least) two different occasions there, but kept being interspersed with other topics, and arguments kept being lost and not responded to. That's why I moved took the time to dig relevant quotes to add to this thread, where we can at least attempt to keep the conversation on topic.

> I don't believe having 15 officially supported clients is the way to go.

Agreed, hence my suggestion to keep the full list in the readme (however formatted) *and* highlight the officially supported ones -- with everything else having their maintainers clearly indicated -- again, [my proposal](https://github.com/waldyrious/tldr/wiki/Clients). I can tell you agree with this when you say

> "it should be clear which is the official client and which are related-projects."

So the only thing missing is a set of recommendations for clients. Thanks @igorshubovych for getting that started! I wonder if we should start a separate issue to discuss the recommendations.

ostera:
Not trying to go over anyone here, but the sooner we get this on the front-page of the repo, the less people will be trying to PR client additions. I'm going out on a limb and merging this, we can continue discussion for recommendations on a separate issue to keep the conversation focused.

waldyrious:
I feel this is kinda sweeping problems under the rug, since a proper spec would allow us to list clients in the readme, forgo the wiki list, and have clear reasons to refuse PRs that attempt to add clients which don't conform to our standards. So now all those clients will instead be dumped on the wiki, the list in the readme will be a subset of it with unclear criterion, and the discussion about specs will have less pressure to move on. That said, it's no big deal, I just ask that we keep working on the specs discussion (on #744 or elsewhere) and finalize that issue.

igorshubovych:
@waldyrious 

They might still wanna list it, like the one in #527. So I guess it is ok to have Wiki page.

waldyrious:
Well, that's a different beast ;) if we decide to ship that (or a variant of it) with the repo, all other clients should in principle conform to the specs.

PR #680, opened 2016-01-16
Add usage for getting a time-slice

cincodenada:
Not sure how best to elaborate on -t.  There's also -sseof to start offset from the end of the file, but that seems beyond scope, and -ss should be enough to find that in the man page.

### DIFF
--- a/pages/common/ffmpeg.md
+++ b/pages/common/ffmpeg.md
@@ -18,6 +18,6 @@
 
 `ffmpeg -i {{in.avi}} -acodec libfaac -ab 128k -vcodec mpeg4 -b 1250K {{out.mp4}}`
 
-- Extract part of a video between given timestamps. `-t` for length instead of end time.  Omit to copy through end.
+- Extract part of a video between given timestamps. Replace `-to` with `-t` for length instead of end time.  Omit `-to` to copy through end.
 
 `ffmpeg -i {{in.mp4}} -ss {{1:30}} -to {{2:45}} -acodec copy {{out.mp4}}`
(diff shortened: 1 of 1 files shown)


rubenvereecken:
Could you have a look at Travis' details so you can repair the syntax?

pages/common/ffmpeg.md
```diff
@@ -17,3 +17,7 @@
 - Convert AVI video to MP4. AAC Audio @ 128kbit, Video @ 1250Kbit:
 
 `ffmpeg -i {{in.avi}} -acodec libfaac -ab 128k -vcodec mpeg4 -b 1250K {{out.mp4}}`
+
+- Extract part of a video between given timestamps. Replace `-to` with `-t` for length instead of end time.  Omit `-to` to copy through end.
```
rubenvereecken: While it's a good example, you're merging two conflicting parameters in your description (`to` and `t`). I'm never a fan of making examples more complicated by squeezing more logic in there. I know it's a tough one and I can see uses for both, but can you try to opt for either option and discard the other?

rubenvereecken:
By the way I just tried this command because the option you added is the main reason I'd use it for (the others are a bit arcane). I noticed the output quality was crap and found out the default output settings are really low. Could you be bothered to add something for quality or is that just not a common scenario?

rubenvereecken:
@cincodenada could you have a look at the inline comment again please?

waldyrious:
I disagree with closing like this, especially without comment. The user hasn't been active on github recently, and who's to say their availability isn't reduced? I don't think we have anything to lose by keeping the PR open -- even if we get conflicts further down the road, we can either allow them to rebase or start a new PR. But closing the door like this is likely to discourage further contributions.

PR #681, opened 2016-01-16
SSH File Transfer (SCP, SFTP)

notpeter:
Reworked SCP:
 * No double examples
 * Replace IP address with hostname place holders.
 * change word use to avoid upload & download
 * alter two remote hosts example to not require agent forwarding.

Added SFTP.

### DIFF
--- a/pages/common/scp.md
+++ b/pages/common/scp.md
@@ -1,30 +1,28 @@
 # scp
 
-> Copies files between hosts on a network.
-> Works over a secure connection (SSH).
+> Secure copy.
+> Copy files between hosts using Secure Copy Protocol over SSH.
 
-- Upload a file, or upload and rename a file:
+- Copy a local file to a remote host:
 
-`scp {{/local/file.txt}} {{10.0.0.1}}:{{/remote/path/}}`
-`scp {{/local/file.txt}} {{10.0.0.1}}:{{/remote/path/newname.txt}}`
+`scp {{local_file}} {{remote_host}}:{{/path/remote_file}}`
 
-- Download a file:
+- Copy a file from a remote host to a local folder:
 
-`scp {{10.0.0.1}}:{{/remote/path/file.txt}} {{/local/folder}}`
+`scp {{remote_host}}:{{/path/remote_file}} {{/path/local_dir}}`
 
-- Upload or download a directory:
+- Recursively copy the contents of a directory on a remote host to a local directory:
 
-`scp -r {{/local/folder}} {{10.0.0.1}}:{{/remote/path/}}`
-`scp -r {{10.0.0.1}}:{{/remote/path}} {{/local/folder}}`
+`scp -r {{remote_host}}:{{/path/remote_dir}} {{/path/local_dir}}`
 
-- Specify username on host:
+- Copy a file between two remote hosts transferring through the local host:
 
-`scp {{/local/file.txt}} {{my_user}}@{{10.0.0.1}}:{{/remote/path}}`
+`scp -3 {{host1}}:{{/path/remote_file.ext}} {{host2}}:{{/path/remote_dir}}`
 
-- Copy a file from one host to another:
+- Use a specific username when connecting to the remote host:
 
-`scp {{10.0.0.1}}:{{/remote/path/file.txt}} {{20.0.0.2}}:{{/other/remote/path}}`
+`scp {{local_file}} {{remote_username}}@{{remote_host}}:{{/remote/path}}`
 
-- Download a file with ssh key:
+- Use a specific ssh private key for authentication with the remote host:
 
-`scp -i {{/local/key}} {{10.0.0.1}}:{{/remote/path/file.txt}} {{/local/folder}}`
+`scp -i {{~/.ssh/id_rsa}} {{local_file}} {{remote_host}}:{{/path/remote_file}}`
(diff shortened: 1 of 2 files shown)


pages/common/scp.md
```diff
@@ -1,30 +1,28 @@
 # scp
 
-> Copies files between hosts on a network.
-> Works over a secure connection (SSH).
+> Secure copy.
+> Copy files between hosts using Secure Copy Protocol over SSH.
 
-- Upload a file, or upload and rename a file:
+- Copy a loocal file to a remote host:
```
rubenvereecken: local

pages/common/scp.md
```diff
@@ -1,30 +1,28 @@
...
 
-`scp {{/local/file.txt}} {{10.0.0.1}}:{{/remote/path/}}`
-`scp {{/local/file.txt}} {{10.0.0.1}}:{{/remote/path/newname.txt}}`
+`scp {{local_file}} {{remote_host}}:{{/path/remote_file}}`
```
rubenvereecken: I agree on removing the suffixes here, they were unnecessary. I did rather like the 'to directory' example though... one could argue that that case is intuitively deductible, but not all commands support that

pages/common/scp.md
```diff
@@ -1,30 +1,28 @@
...
 
-- Download a file:
+- Copy a file from a remote host to a local folder:
 
-`scp {{10.0.0.1}}:{{/remote/path/file.txt}} {{/local/folder}}`
+`scp {{remote_host}}:{{/path/remote_file}} {{/path/local_dir}}`
```
rubenvereecken: You're being a tad inconsistent here, with the upload you do file to file, here it's file to folder.

pages/common/scp.md
```diff
@@ -1,30 +1,28 @@
...
 
-- Upload or download a directory:
+- Recursively copy the contents of a directory on a remote host to a local directory:
 
-`scp -r {{/local/folder}} {{10.0.0.1}}:{{/remote/path/}}`
-`scp -r {{10.0.0.1}}:{{/remote/path}} {{/local/folder}}`
+`scp -r {{remote_host}}:{{/path/remote_dir}} {{/path/local_dir}}`
 
-- Specify username on host:
+- Copies a file between two remote hosts transferring through the local host:
```
rubenvereecken: Prefer infinitive tense, "copy", over "copies".

I guess this replaces the copy from one host to another example below? This adds `-3`, I looked it up and it does what you describe. Then what's the default? Since the previous author left out the parameter.

pages/common/scp.md
```diff
@@ -1,30 +1,28 @@
...
+- Use a specific username when connecting to the remote host:
 
-`scp {{10.0.0.1}}:{{/remote/path/file.txt}} {{20.0.0.2}}:{{/other/remote/path}}`
+`scp {{local_file}} {{remote_username}}@{{remote_host}}:{{/remote/path}}`
 
-- Download a file with ssh key:
+- Use a specific ssh private key for authentication with the remote host:
 
-`scp -i {{/local/key}} {{10.0.0.1}}:{{/remote/path/file.txt}} {{/local/folder}}`
+`scp -i {{~/.ssh/id_rsa}} {{local_file}} {{remote_host}}:{{/path/remote_file}}`
```
rubenvereecken: Great example of how to suggest file names!

pages/common/sftp.md
```diff
@@ -0,0 +1,20 @@
+# sftp
+
+> Secure file transfer program.
+> Copy files between hosts using Secure File Transfer Protocol over SSH.
```
rubenvereecken: Shouldn't the page somehow state how `sftp` starts an interactive session? I usually use sftp instead of scp for just this reason.

pages/common/scp.md
notpeter: The default is equivalent to running `ssh host1 "scp /path/filename user@host2:/path/filename" and so requires host1 to allow outgoing SSH and host2 to accept connections from host1.

Which might be sane with passwords on a simple network, but with SSH keys it requires a Agent Forwarding (dangerous) to function.

rubenvereecken:
I think this is a great endeavour for some tricky but extremely usefools tools by the way.

pages/common/sftp.md
notpeter: Modified the first example to explicitly say "and enter an interactive command mode" which I think is better than trying to stuff "interactive" into the program description.

pages/common/scp.md
notpeter: That was actually intentional.  I wanted to offer the three styles of copy (file->file, file->dir, recursive dir->dir) in only three examples.  But if you think "local file to remote file", "remote file to local file" and "dir to dir" are better without any mention that the filename is optional on the right side, I'd be fine with that too.

pages/common/sftp.md
rubenvereecken: You're right, it's not interactive in all cases. The second one is though, but I don't mind implicitly building on the first example instead of being explicit about the whole behavior in every example description.

pages/common/scp.md
rubenvereecken: I tried reading it for my own personal use and it's clear enough although it doesn't per se aid interpretation. I guess I've just come to expect certain behavior and scp nicely fits my expectations. Good enough for me.

rubenvereecken:
This one has my :+1:, someone else have a look at this so we can get this merged.

igorshubovych:
Thanks, @notpeter!

PR #494, opened 2015-12-30
Add tldr for openssl

Duologic:
An excerpt from https://www.sslshopper.com/article-most-common-openssl-commands.html

PR #682, opened 2016-01-16
openssl is the worst

notpeter:


rubenvereecken:
Hi there... Peter? 

We've already got a PR for openssl at #494, though it's not merged because the commits aren't squashes _yet_. I'll hunt after that one first. Do you think yours could fit together with that page? If so I suggest you create a page off of that one (because it already has a green light) so we can have a look at it after we merge that one. You can even rewrite most of the page but that'll require some more motivation/persuasion on your part.

Feel up to it :)?

notpeter:
Yeah, I'll create a new PR once that's merged.

PR #683, opened 2016-01-16
Zcat and zless.

notpeter:


pages/common/zcat.md
```diff
@@ -0,0 +1,11 @@
+# zcat
+
+> Print data from gzip compressed files.
+
+- Print the uncompressed contents of a gzipped file to the standard output:
+
+`zcat {{file.txt.gz}}`
+
+- Grep the contents of a compressed file:
```
waldyrious: Maybe don't use "grep" as a verb. How about "Filter (with `grep`)"?

pages/common/zless.md
```diff
@@ -0,0 +1,7 @@
+# zless
+
+> View compressed files.
+
+- Page through a compressed archive with less:
```
waldyrious: Format "less" using backticks to make it clearer that it's a command.

waldyrious:
Nice one! I've added two minor comments inline.

notpeter:
Fixed and Squashed.

pages/common/zcat.md
```diff
@@ -0,0 +1,11 @@
...
+
+> Print data from gzip compressed files.
+
+- Print the uncompressed contents of a gzipped file to the standard output:
+
+`zcat {{file.txt.gz}}`
+
+- Filter the contents of a compressed file with `grep`:
+
+`zcat {{file.txt.gz}} | grep {{pattern}}`
```
igorshubovych: This example should not be here. It is against our guidelines:
https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines

notpeter: Then there's no practical example of why zcat is useful: feeding a pipeline from a compress file.

waldyrious:
Without the examples these pages seem kinda purposeless. @notpeter any options that could be added? @igorshubovych maybe we could open an exception here? The guidelines are just that, not strict rules.

notpeter:
> This example should not be here. It is against our guidelines
> https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines

Your comment is not constructive. It fails to mention *why* (you think) it's incorrect, offers no path to resolution and then positions the PR submitter in as outside the bounds of the community to which they have offered their time.   Also, you top it off with a [guidelines link](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines) which makes no mention of shell redirections or pipelines. Perhaps you read this line differently than I do. 
> Be specific: avoid explaining general UNIX concepts that could apply to any command (ex: relative/absolute paths, brace expansion, character escaping...).

waldyrious:
Just so things don't get too heated up here, I am fully aware that we've been a bit blunt at times -- after repeating the same lines (Thanks for the contribution, etc.) many times in the past few weeks due to the recent surge of PRs, we sometimes leave the pleasantries aside and it may sound like we're imposing hurdles on volunteer contributions to a project we're supposed to nurture and curate.

That said, Igor's comment, albeit failing in the manner I mentioned above, *did* suggest, and not that implicitly, a resolution path (remove the example). Unfortunately this isn't (IMO) a sensible choice for this particular example, and the guidelines indeed don't mention piping commands. I do agree that we ought to be more careful with the feedback we provide in order not to alienate contributors. Thanks for helping us address this issue @notpeter.

igorshubovych:
@notpeter we try to avoid using other commands in the pages. Even if it is useful. Because it is man pages, helper for options, not the list of useful commands. 

However there are exceptions from this rule, i.e. you cannot explain `xargs` without pipes.

There is another project, which goal is to collect useful code examples, which can be reused. It is called [cheat](https://github.com/chrisallenlane/cheat), and your example suits them.

I think the page still makes sense with no options. Some client may soon implement search or related pages, and it will be helpful information.

@rprieto, @ostera, @rubenvereecken, @felixonmars what do you think regarding this case?

ostera:
:-1: no piping, people can figure out what to do with the output

notpeter:
It honestly doesn't matter to me, there just isn't a second example without using pipes.

I think it's worth having pages with only a single example, if only because a reasonable text description may suggest use that isn't entirely apparent looking at the man pages.  Zless and zcat in particular don't have their own man pages at all (aliased to `gzip`) and so simple description/suggestion from tldr is still more useful than what may be available.

If other folks agree, I can also add something to the contributing.md explicitly discouraging use of pipes or secondary commands as well as explicitly OKing pages with a single example.

waldyrious:
> a reasonable text description may suggest use that isn't entirely apparent looking at the man pages

Good point.

>  I can also add something to the contributing.md explicitly discouraging use of pipes or secondary commands as well as explicitly OKing pages with a single example.

That'd be much appreciated, although IIRC simple commands like `ls` or `echo` are OK for demonstrating specific features of commands (which is why I am in favor of an exception here, as long as we keep the examples [as simple as reasonable, but no simpler](https://en.wikiquote.org/wiki/Albert_Einstein#simplicity_limit)).

notpeter:
Pulled the second example with grep. Should be clean to merge.

igorshubovych:
LGTM

waldyrious:
Thanks, you've been exceedingly patient.

igorshubovych:
Thank you, @notpeter!

PR #504, opened 2015-12-30
find: added options for multiple search criteria and empty files

Cvetomird91:
Extended the find page.

PR #684, opened 2016-01-16
Minor cleanup: ping, zpool, zfs

notpeter:
ZFS/Zpool:  snake_case and removed doubled examples. 

Ping: Improve awkward english.

### DIFF
--- a/pages/common/zfs.md
+++ b/pages/common/zfs.md
@@ -8,20 +8,20 @@
 
 - Create a new ZFS filesystem:
 
-`zfs create poolname/newfsname`
+`zfs create {{pool_name/filesystem_name}}`
 
 - Delete a ZFS filesystem:
 
-`zfs destroy {{poolname/newfsname}}`
+`zfs destroy {{pool_name/filesystem_name}}`
 
 - Create a Snapshot of a ZFS filesystem:
 
-`zfs snapshot {{poolname/filesystem@snapshot-name}}`
+`zfs snapshot {{pool_name/filesystem_name}}@{{snapshot_name}}`
 
 - Enable compression on a filesystem:
 
-`zfs set compression=on {{poolname/fileystem}}`
+`zfs set compression=on {{pool_name/filesystem_name}}`
 
 - Change mountpoint for a filesytem:
 
-`zfs set mountpoint={{/my/mount/path}} {{poolname/filesystem}}`
+`zfs set mountpoint={{/my/mount/path}} {{pool_name/filesystem_name}}`
--- a/pages/common/zpool.md
+++ b/pages/common/zpool.md
@@ -4,30 +4,28 @@
 
 - Show the configuration and status of all ZFS zpools:
 
-`zpool status [{{poolname}}]`
+`zpool status`
 
 - Check a ZFS pool for errors (verifies the checksum of EVERY block). Very CPU and disk intensive:
 
-`zpool scrub {{poolname}}`
+`zpool scrub {{pool_name}}`
 
 - List zpools available for import:
 
 `zpool import`
 
-- Import a zpool, optionally specifying a new name:
+- Import a zpool:
 
-`zpool import {{poolname}}`
-`zpool import {{poolname}} {{newpoolname}}`
+`zpool import {{pool_name}}`
 
 - Export a zpool (unmount all filesystems):
 
-`zpool export {{poolname}}`
+`zpool export {{pool_name}}`
 
 - Show the history of all pool operations:
 
-`zpool histrory {{poolname}}`
+`zpool histrory {{pool_name}}`
 
 - Create a mirrored pool:
 
-`zpool create {{poolname}} mirror {{disk1}} {{disk2}}`
-`zpool create {{poolname}} mirror {{disk1}} {{disk2}} mirror {{disk3}} {{disk4}}`
+`zpool create {{pool_name}} mirror {{disk1}} {{disk2}} mirror {{disk3}} {{disk4}}`
--- a/pages/common/ping.md
+++ b/pages/common/ping.md
@@ -6,13 +6,13 @@
 
 `ping {{host}}`
 
-- Ping host limiting the number of packages to be send to four:
+- Ping a host only a specific number of times:
 
-`ping -c 4 {{host}}`
+`ping -c {{count}} {{host}}`
 
-- Ping host, waiting for 0.5 s between each request (default is 1 s):
+- Ping host, specifying the interval in seconds between requests (default is 1 second):
 
-`ping -i 0.5 {{host}}`
+`ping -i {{seconds}} {{host}}`
 
 - Ping host without trying to lookup symbolic names for addresses:
 


pages/common/zfs.md
```diff
@@ -8,20 +8,20 @@
 
 - Create a new ZFS filesystem:
 
-`zfs create poolname/newfsname`
+`zfs create {{pool_name}}/{{filesystem_name}}`
 
 - Delete a ZFS filesystem:
 
-`zfs destroy {{poolname/newfsname}}`
+`zfs destroy {{pool_name}}/{{filesystem_name}}`
```
rubenvereecken: Are these really worth splitting up?

rubenvereecken:
I'm actually really happy with these changes, it's part of our effort to harmonize our token format as well as some overall better English.

pages/common/zpool.md
```diff
@@ -4,30 +4,28 @@
...
 - List zpools available for import:
 
 `zpool import`
 
-- Import a zpool, optionally specifying a new name:
+- Import a zpool (optionally renaming it):
 
-`zpool import {{poolname}}`
-`zpool import {{poolname}} {{newpoolname}}`
+`zpool import {{pool_name}} {{[new_pool_name]}}`
```
waldyrious: I feel this is kinda cheating. There shouldn't be two commands in this example anyway, so this isn't your fault @notpeter, but I'd rather we come to an agreement on this. I don't think we should include examples with optional parameters, as that could open a can of worms, but I'm open to reconsider.

pages/common/zpool.md
```diff
@@ -4,30 +4,28 @@
...
 - Show the history of all pool operations:
 
-`zpool histrory {{poolname}}`
+`zpool histrory {{pool_name}}`
 
 - Create a mirrored pool:
 
-`zpool create {{poolname}} mirror {{disk1}} {{disk2}}`
-`zpool create {{poolname}} mirror {{disk1}} {{disk2}} mirror {{disk3}} {{disk4}}`
+`zpool create {{pool_name}} mirror {{disk1}} {{disk2}} mirror {{disk3}} {{disk4}}`
```
waldyrious: Not sure leaving only the longer example is a good choice, but I'll leave it to those who actually know this command to decide.

waldyrious:
Excellent work on the Ping page, @notpeter. The other two may need a little discussion but overall great start :+1:

igorshubovych:
Also could you plz squash commits?

waldyrious:
@igorshubovych shouldn't we wait until the discussion is finalized?...

pages/common/zpool.md
notpeter: Yeah, it's tough, because you can be much more expressive without requiring adding a whole extra example.  It also breaks copy paste (both cause you have to remove brackets and decide whether you want that option).

If we decide to allow this I'd imagine a lot of examples would have a `[-v]` which wouldn't be terrible (again a compact way to express things) but would certainly require it's on style guide mention: "if possible put all optional flags at the end so as to simplify copy/paste for the common use"

pages/common/zfs.md
notpeter: Nope. I've switched it back.

igorshubovych:
@waldyrious it can be done simultaneously

rubenvereecken:
This PR gave me a nice edge case to update the linter with, but you've already repaired it :(

notpeter:
And that's why squashing commits in PRs is evil. You get a cleaner commit history on merge, but at the cost of:
* PR history is lost unless someone previously commented on that earlier commit 
* implies force pushing is acceptable, nay required on collaborative projects
* significantly raising the barrier to submitting a PR

It's not your fault, but `git rebase -i REF` is just a horrible workflow. IMO making it a pre-req for contributors (unless their changes are magically perfect on first go) is a hostile project policy.

waldyrious:
I agree with @notpeter, the squashing policy has caused lots of headaches in PRs lately and makes the project less welcoming to contributors. Especially in this case where it was requested even before discussion was finalized, likely requiring further squashing down the road. I suggest we remove that requirement (which was never formalized anyway).

pages/common/zpool.md
waldyrious: I think there isn't a strong justification for keeping the optional argument in this particular case, but we can reconsider this policy if an example eventually appears where it makes sense.

rubenvereecken:
> significantly raising the barrier to submitting a PR

> I agree with @notpeter, the squashing policy has caused lots of headaches in PRs lately and makes the project less welcoming to contributors.

That's actually a problem but one that we may be able to work around if we automate it. 

> implies force pushing is acceptable, nay required on collaborative projects

Force pushing never happens to the collaborative effort, just personal branches, which makes all the difference here I feel.

I like how it ensures the cleanliness of the repository. Roughly a page or change will be equal to a commit. No one gets in 7 commits because they can't get the syntax right. No one would _want_ 7 commits that just keep adding colons and adding backticks. Squashing has never been agreed on among contributors because of the recent grow in the team, we've always insisted on squashing. And it is mentioned in the contributing guidelines.

I know that "to squash or not to squash" is a topic that people always enjoy making a problem of. The big problem here that you guys mentioned is first-time users, possibly through the web interface. I'll look into automated squashing.

rubenvereecken:
Actually, rewriting is rather evil. I shouldn't go there.

igorshubovych:
That's an interesting discussion.

In general I agree with @notpeter for any other project. But tldr is almost like Wiki with some nuances, and therefore it is different from project with code.

Check out these PRs: #606, #583, #531. There were more examples like this. I saw once a PR with 13 commits, and most of them were annotated with messages like "Update somepage.md" (I do not rembember exactly which page it was). The readability of overwritten PR and 13 commits with the same message is the same. However a single commit does not mess the history.

However there are examples, where good PR was spoiled by attempt to squash (#504). And so @notpeter is correct that breaking this rule may help us lower barrier for contributors.

Therefore I have  a question to you guys. What if we ask to squash commits only when:
- there are more than 5 commits (exact number is debatable)
- commits have the same or meaningless messages

igorshubovych:
I forget to add that due to @rubenvereecken's linter we don't need to explain contributing guidelines in every PR, so it means that we have less PRs with 13 commits from now.

waldyrious:
I agree with a threshold for asking for squashing. 5 sounds good to me. As for the commit messages, I would agree only if that's an additional criteria (i.e. we'd ask for squashing if there are +5 commits *and* no meaningful messages), not a stand-alone one (i.e. we'd ask for squashing if there are 3 commits without proper messages).

igorshubovych:
We discussion in Gitter and we agreed not to restrict by any number of commits as commit messages are meaningful.

In this case, if there are no other objections, IMO this PR is almost good to go.

@waldyrious is your feedback addressed?

igorshubovych:
By the way, @notpeter I have to thank you for raising this discussion. It will allow us to move faster.

waldyrious:
> not to restrict by any number of commits as commit messages are meaningful.

Do you mean *if* commit messages are meaningful? If so, then yes, my concerns are addressed.

> It will allow us to move faster.

And more importantly, to provide a more welcoming environment for contributors :)

> IMO this PR is almost good to go.

What's missing? (beside by inline comment)?

igorshubovych:
> What's missing? (beside by inline comment)?
Inline comment.

notpeter:
Setting the squash/no-squash discussion aside does anyone have any concerns about this PR?

My reasoning for choosing the four disk mirror over the more simplistic two disk one, is that I personally have accidentally created a four way mirror (`zpool create pool_name mirror disk1 disk2 disk3 disk4`) when I intended to create a striped pair of two way mirrors (`zpool create pool_name mirror disk1 disk2 mirror disk3 disk4`) and the simple case (`zpool create pool_name mirror disk1 disk2`) is reasonably inferred from the more complex example.

waldyrious:
@notpeter I'll defer to your expertise as someone who has actually used the tool. I just ask that you keep in mind people who are learning to use the tool -- I'm not sure who you have in mind when you say "reasonably inferred".

Apart from that, I think this PR is ready for merging.

rubenvereecken:
I'd agree with the last discussion on zpool and will give my second :+1: for this PR, thereby gaining yet another step to our goal of unduplicating all examples :tada: .

PR #685, opened 2016-01-16
Add devfsadm for solaris.

notpeter:


pages/sunos/devfsadm.md
```diff
@@ -0,0 +1,15 @@
+# devfsadm
+
+> Administration command for /dev. Maintains the /dev namespace.
```
waldyrious: I'd say format "/dev" using backticks, to highlight that it is a path. @igorshubovych any thoughts?

pages/sunos/devfsadm.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Scan for new disks:
+
+`devfsadm -c disk`
+
+- Cleanup any dangling /dev links and scan for new device:
+
+`devfsadm -C -v`
+
+- Dry-run, output what would be changed but make no modifications:
```
waldyrious: instead of a comma, perhaps a dash would be more appropriate here.

notpeter:
Agreed. I think backticking /dev is an odd corner case but helpful to highlight it's not just an arbitrary folder.

igorshubovych:
LGTM

waldyrious:
Thanks :)

PR #636, opened 2016-01-08
yaourt: add page

dylanrees:

PR #690, opened 2016-01-18
Create yaourt.md

dylanrees:


waldyrious:
@dylanrees thanks for redoing this (following #636). Don't worry about the error: you can edit this PR to fix the lint error pointed by Travis (missing `#` before the command name), since we've relaxed the squashing requirements and now only ask if there is an excessive (5+) number of commits in a single PR.

rubenvereecken:
Never thought anyone would miss that hash :P. Added the case to tldr-lint for better feedback.

waldyrious:
@rubenvereecken suggestion: "should be started with" --> "should start with"

rubenvereecken:
@waldyrious definitely, that would conform with the other errors. I've actually reworded the first ones I wrote a couple of times, switching between "this is how it should be" and "this is what you did wrong". I opted for the former everywhere where possible.

waldyrious:
excellent call :+1:

rubenvereecken:
@dylanrees could you just patch up the linter errors please? You can even do this automatically by installing tldr-lint through `npm install tldr-lint` and then running `tldr -fi pages`.

dylanrees:
Ok, good?

rubenvereecken:
Hmm something tells me you didn't run the formatter, because then it'd be correctly formatted. Or did it crash somewhere? Just run the formatter and commit, everything should be fine :). I'll merge right away when you do so.

waldyrious:
@rubenvereecken have you considered he could be using the github web interface? The commit message certainly matches what editing a file on github produces by default. I've said it before and repeat, I think out workflow should always accommodate users of the web interface, and never assume a local clone made using git, with pre-commit hooks, build scripts and all those bells and whistles. Our pickiness in reviewing contributions is troublesome enough, no need to add extra hurdles.

rubenvereecken:
I keep forgetting. Too bad that, really. @waldyrious let's make our own interface :).

dylanrees:
To clarify, I did use the web interface on this commit.  I've had issues getting the CLI method set up properly on my machine, but I'll attempt it again if it's an issue.

Bit of a newcomer to Github :)

waldyrious:
@dylanrees the burden is on us to provide a simple contribution workflow :) but if you want to tae the opportunity to hone your git skills, I can only encourage that! Otherwise we can take it from here. Let us know what you prefer.

rubenvereecken:
I'm totally hijacking this one. If you feel up for a challenge, your pages are welcome anytime :).

PR #691, opened 2016-01-18
add base64

andyxning:
add base64 command usage

### DIFF
--- a/pages/osx/base64.md
+++ b/pages/osx/base64.md
@@ -1,0 +1,19 @@
+# base64
+
+> Encode and decode using Base64 representation.
+
+- Encode a file:
+
+`base64 -i {{plain_file}}`
+
+- Decode a file:
+
+`base64 -D -i {{base64_file}}`
+
+- Encode from stdin:
+
+`echo {{plain_text}} | base64`
+
+- Decode from stdin:
+
+`echo {{base64_text}} | base64 -D`


pages/common/base64.md
```diff
@@ -0,0 +1,11 @@
+# base64
+
+> Encode and decode using Base64 representation
+
+- Encode using Base64:
+
+`echo {{any content}} | base64 `
```
waldyrious: I suppose this would always be text, in order for you to type it, right? In that case I'd suggest `{{plain_text}}` and `{{base64_text}}`.

andyxning: that's right. i will fix it. :)

igorshubovych:
Please check this out:
https://github.com/tldr-pages/tldr/blob/master/pages/linux/base64.md

andyxning:
@igorshubovych , i have not checked `base64` for linux. That is my fault. However, after digging into it, i find `base64` command on `GNU/Linux` is not the same as `OS X`. So i will update it and it think we can put a new `base64` under `osx`. :)

igorshubovych:
Now it looks good to me.

andyxning:
@igorshubovych , if then, i think you may merge it. :)

rubenvereecken:
I'm happy myself :+1:

PR #692, opened 2016-01-18
update cat command

andyxning:
* fix typo
* add output line numbers

### DIFF
--- a/pages/common/cat.md
+++ b/pages/common/cat.md
@@ -10,6 +10,10 @@
 
 `cat {{file1}} {{file2}} > {{target-file}}`
 
-- Append serveral files into the target file:
+- Append several files into the target file:
 
 `cat {{file1}} {{file2}} >> {{target-file}}`
+
+- Number all output lines:
+
+`cat -n {{file}}`


pages/common/cat.md
```diff
@@ -10,6 +10,14 @@
 
 `cat {{file1}} {{file2}} > {{target-file}}`
 
-- Append serveral files into the target file:
+- Append several files into the target file:
 
 `cat {{file1}} {{file2}} >> {{target-file}}`
+
+- Number the non-blank lines:
```
rubenvereecken: Sounds like specialized behavior of the one below, wouldn't switching make more sense? Also I think you can drop "the".

andyxning: both `linux` and `osx` have the `-b` and `-n` options for `cat`.

rubenvereecken: I don't mean that, I just thought `-n` is more common than `-b` (a specialized version of `-n` you could say).

andyxning: you mean that we may only need `-n`?

rubenvereecken:
Alright, I'm happy with this one :+1:

waldyrious:
LGTM, thanks!

PR #695, opened 2016-01-19
tldrl: add page

rubenvereecken:
Contributing guidelines in #694 refer to this page, but currently do so using `tldr tldr-lint`. I'm trying to get `tldrl` to become the pedantic way of referring to it.

Let's get this tool used and save us some work like it's supposed to be. It's rather sad Travis is the only audience.


igorshubovych:
LGTM!

felixonmars:
:+1:

PR #696, opened 2016-01-19
-H instead of --head for cURL, adding verbose output

bobstrecansky:


bobstrecansky:
Using -I instead of --head for curl saves a couple keystrokes; added verbose output (has saved me many times)

waldyrious:
Thanks for the contribution! I'm ok with `--header` --> `-H`, since the description uses the word "header" which allows people to see what it means. But swapping `--head` for `-I` would require the description to use a word starting with I (hopefully the word from which the "I" was derived), so that people *understand* the commands, not just *memorize* them. What would you suggest?

bobstrecansky:
The reason I chose -I is that it's much shorter and I've seen it used in the wild more often. Perhaps we could include both?

rubenvereecken:
I'm talking besides the option discussion: The page is too long with the additions. If you feel the additions are absolutely needed, replace older examples. Otherwise, please remove them.

igorshubovych:
I like `-H`.
I don't like like `-I`, agree with @waldyrious.
I don't think `-v` and `-vvv` are more important than existing examples.

From [Contributing Guidelines](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines):
1. Focus on the 5 or 6 most common usages. It's OK if the page doesn't cover everything; that's what `man` is for.

waldyrious:
> The reason I chose -I is that it's much shorter and I've seen it used in the wild more often.

I don't think those are good reasons to use it. `--head` isn't that much harder to type, and it's much clearer and memorable.

> Perhaps we could include both?

We don't provide multiple ways to use the same option in any other page (that I know of). Short option names are OK if they are explained in the description in a way that makes them logical and memorable (particularly, by using the word they were derived from, or a simpler synonym that starts with the same letter). `-I` could be used if we manage to come up with a description that explains what it stands for. But if it doesn't stand for anything, then I'd suggest we keep `--head`.

igorshubovych:
@bobstrecansky 
It is ok that some of the options that you are using frequently are not included.

I am in the same situation. There are options which I use daily in `git`, `npm`. But then I think if they are good for newbies or not, think about how common they are. And, honestly, not all of them are common.

bobstrecansky:
Okay, it sounds like none of these proposed changes are relevant?  If so, i'll close the PR.

igorshubovych:
well, it seems like everybody here liked `-H` example.

bobstrecansky:
Okay, i'll just merge the -H example.  @waldyrious - i'm not sure why -I is used for head, so we can omit that if it is unclear for the end user.

rubenvereecken:
That's settled then. Sorry this took so long but thanks for bearing with us @bobstrecansky. Happy to have this.

waldyrious:
@bobstrecansky thanks for understanding. Maybe if someone comes up with a nice explanation for the -I we can switch to using it.

PR #697, opened 2016-01-19
Use NodeJS NPM package markdownlint instead of Ruby gem mdl

igorshubovych:
Inside:
- Remove Gemfile and Gemfile.lock
- Remove mention about Ruby from CONTRIBUTING.md, mention NodeJS
- Remove Makefile, all build scripts are in NodeJS
- Use markdownlint and markdownlint-cli
- Create package.json
- Rewrite indexing script in NodeJS

Motivation:
- [markdownlint](https://github.com/DavidAnson/markdownlint) is more actively developed than [mdl](https://github.com/mivok/markdownlint), which is last released in April 2015 and has some rare commits after that.
- Unification of the linting stack: tldr-lint is already based on NodeJS, and so the full stack will be on NodeJS
- Standardization of linting stack: our current system is tricky for novice, package.json is a documented standard
- Performance: 1sec vs 2 sec. But more important is the fact that due to standardization of Travis container (no need to setup both Ruby and Node.JS) the check of pull-request can take 20-30 sec comparing to 40-50 sec

Installation instructions:
```
rm .git/hooks/pre-commit
npm install
```


ostera:
:+1:

waldyrious:
+:100: :smiley:

rubenvereecken:
Happy you took the incentive, we did talk about this before. It's great to have everything Nodeified even though that might be somewhat biased. 

I'm actually wondering if we'll ever see a markdown linter error again since tldrl catches about the same errors except for inline markdown. Great work though!

I'll merge this in if you don't mind, eager to see this in the works.

PR #633, opened 2016-01-07
Formatted all pages according to guidelines.

rubenvereecken:
Alright, don't merge this in yet. I will after due discussion.

All pages are now formatted according to discussion in #622. It's not final, as not everyone has had their say yet.

It's important that once this gets merged in we update `CONTRIBUTING.md`, and be very vigilant when merging in new changes. At least until I get time to incorporate my linter. It's best to have it start with a clean slate though, so Travis can play with it after every PR.

Summary of changes: Whitespace, top descriptions are full sentences and example description are sentences ending in a colon. The last is still under debate.

Hurray [`tldr-lint`](https://github.com/rubenvereecken/tldr-lint) and its new brother, `tldr-format`. Soon available through `npm`.

PR #698, opened 2016-01-19
Update docker description

lukeyeager:
That period just doesn't belong.
```
$ tldr docker
# docker                                                                                              
                                                                                                      
  Docker allows you to package an application with all of its.                                        
  Dependencies into a standardized unit for software development.
```

### DIFF
--- a/pages/common/docker.md
+++ b/pages/common/docker.md
@@ -1,7 +1,6 @@
 # docker
 
-> Docker allows you to package an application with all of its.
-> Dependencies into a standardized unit for software development.
+> Manage Docker containers and images.
 
 - List of running docker containers:
 


igorshubovych:
I actually don't like this description at all. I mean I don't like both variants. IMO we should not try to explain what Docker is in this short description. Because it impossible, or possible but useless.

In my opinion it is better to say something like

```md
> Command line interface for managing Docker containers 
```

Or something like this.

lukeyeager:
Sounds good to me. Updated.

igorshubovych:
Put period (.) at the end.

Also Google suggests it should be "command-line" with dash, not "command line".

waldyrious:
Thanks @lukeyeager for the attention to detail! But I'm afraid Igor's suggestion goes to the other extreme, to be redundant and not informative at all. How about something in the middle? Here's a suggestion:

> **docker**
> Command line interface for managing the deployment of applications inside standardized software containers

waldyrious:
By the way, pinging @rubenvereecken because the awkward formatting was due to [#633](https://github.com/tldr-pages/tldr/pull/633/files#diff-35)

igorshubovych:
@waldyrious 
I am briefly familiar with Docker, so please correct me if I am wrong.
Docker is a format of containers. There are other packagers/containers: Rocket (rkt), Odin and [others](https://en.wikipedia.org/wiki/Operating-system-level_virtualization). And `docker` is just a CLI to manage only Docker containers. Hence it is incorrect to say "applications inside standardized software containers". It is only about Docker containers.

https://en.wikipedia.org/wiki/Docker_(software)

lukeyeager:
I'm happy to update the PR with whatever y'all decide. Also happy to let you close this PR and fix it in your own way if you like.

waldyrious:
@igorshubovych good point. The Docker format is a standardized one, but it's better to be explicit to avoid ambiguity. I'd propose then changing "standardized" to "Docker":

> **docker**
> Command line interface for managing the deployment of applications inside Docker software containers

igorshubovych:
I checked that `docker`’s help is even shorter:

> A self-sufficient runtime for containers.

So it would be weird if tldr’s version is longer than official help :)

To me, it should be something like

> Manage Docker containers and images

or just

> Manage Docker containers

waldyrious:
I like "Manage Docker containers and images" :+1:

lukeyeager:
> I like "Manage Docker containers and images"

Updated to this.

igorshubovych:
Thanks!

Issue #687, opened 2016-01-17
Pages with multiple commands per example


rubenvereecken:
Related to tldr-pages/tldr-node-client#64 and somewhat to #685. 

The current linter at `tldr-pages/tldr-lint#multiple-examples` (npm install path) detects multiple commands per example. It also comes with a tabular output formatter. That enabled me to do the following: 

`tldrl pages/ -t 2>&1 | cut -f1 | uniq | grep "pages.*" -o`

Resulting in this beauty.
```
pages/common/cal.md
pages/common/ls.md
pages/common/redis-cli.md
pages/common/rename.md
pages/common/salt-key.md
pages/common/scp.md
pages/common/srm.md
pages/common/touch.md
pages/common/which.md
pages/common/zpool.md
pages/linux/md5sum.md
pages/linux/sha1sum.md
pages/linux/sha224sum.md
pages/linux/sha256sum.md
pages/linux/sha384sum.md
pages/linux/sha512sum.md
pages/linux/ss.md
pages/linux/wpa_cli.md
pages/osx/md5.md
pages/osx/sysctl.md
pages/osx/systemsetup.md
pages/osx/xed.md
```

These are all files with multiple commands per example. 

EDIT: @notpeter had a go at them, we're almost through.

- [x] pages/common/scp.md
- [x] pages/common/zpool.md
- [x] pages/linux/ss.md
- [x] pages/linux/wpa_cli.md

igorshubovych:
cc @rprieto

notpeter:
I already have PRs waiting for:
pages/common/zpool.md (#684)
pages/common/scp.md (#681)

I think I got all the rest except two, because I'm not familiar.
pages/linux/ss.md
pages/linux/wpa_cli.md

rubenvereecken:
Now merged in tldr-lint's master branch.

PR #699, opened 2016-01-20
Single command per example

notpeter:
Gets almost all of #687.

### DIFF
--- a/pages/common/cal.md
+++ b/pages/common/cal.md
@@ -2,18 +2,18 @@
 
 > Prints calendar information.
 
-- Display a calendar for the current month or specified month:
+- Display a calendar for the current month:
 
 `cal`
-`cal -m {{12}}`
-`cal -m {{Dec}}`
 
-- Display a calendar for the current year or a specified year:
+- Display a calendar for a specific month:
 
-`cal -y`
+`cal -m {{month_number}}`
+
+- Display all 12 months of calendar for a specific year:
+
 `cal 2013`
 
 - Display date of Easter (western churches):
 
-`ncal -e`
-`ncal -e 2013`
+`ncal -e {{year}}`
--- a/pages/common/cal.md
+++ b/pages/common/cal.md
@@ -10,9 +10,13 @@
 
 `cal -m {{month_number}}`
 
-- Display all 12 months of calendar for a specific year:
+- Display a 12 month calendar for the current year:
 
-`cal 2013`
+`cal -y`
+
+- Display a 12 month calendar for a specific year:
+
+`cal 2016`
 
 - Display date of Easter (western churches):
 
--- a/pages/common/ls.md
+++ b/pages/common/ls.md
@@ -14,7 +14,7 @@
 
 `ls -la`
 
-- Long format list with size displayed using human readable units (KB,MB,GB):
+- Long format list with size displayed using human readable units (KB, MB, GB):
 
 `ls -lh`
 
--- a/pages/common/ls.md
+++ b/pages/common/ls.md
@@ -2,33 +2,22 @@
 
 > List directory contents.
 
-- List all files, even hidden:
+- List files one per line:
+
+`ls -1`
+
+- List all files, including hidden files:
 
 `ls -a`
 
-- List all file names (no extra info):
+- Long format list (permissions, ownership, size and modification date) of all files:
 
-`ls -A1`
+`ls -la`
 
-- List all files with their rights, groups, owner:
-
-`ls -l`
-
-- List all files and display the file size in a human readable format:
+- Long format list with size displayed using human readable units (KB,MB,GB):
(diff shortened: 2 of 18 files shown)


rubenvereecken:
First off before I dig in, I love you for this!

notpeter:
I'm actually at fault for almost all of these multi-liners so cleaning up my mess is the least I could do.

igorshubovych:
Wow. That's an amazing work. I always wanted to do this, but was always too lazy for it.

However in future it is better to split it by page. Imagine somebody will have ideas on how to improve it, you will have long discussion and we cannot merge the whole nice PR, just because of unfinished discussion about minor things.

igorshubovych:
Actually how about we merge it sooner and then we can open PRs on individual pages in case of need?

notpeter:
Well, I certainly wasn't going to create a dozen individual branches and pull requests. 

If there's issues I'm happy to cherry pick contentious commits into their own branches/PRs for discussion and rebase them off this branch.

pages/common/cal.md
```diff
@@ -2,18 +2,18 @@
...
 
-- Display a calendar for the current year or a specified year:
+- Display a calendar for a specific month:
+
+`cal -m {{month_number}}`
+
+- Display all 12 months of calendar for a specific year:
 
-`cal -y`
 `cal 2013`
```
waldyrious: While we're at it, if you don't mind, please change 2013 to {{year}}. And perhaps include a `cal -y` example before this one.

pages/common/ls.md
```diff
@@ -2,33 +2,22 @@
...
+`ls -a`
 
-- List all files with their rights, groups, owner:
+- Long format list (permissions, ownership, size and modification date) of all files:
 
-`ls -l`
+`ls -la`
 
-- List all files and display the file size in a human readable format:
+- Long format list with size displayed using human readable units (KB,MB,GB):
```
waldyrious: Spaces after commas?

pages/common/rename.md
```diff
@@ -2,19 +2,18 @@
...
 > Renames multiple files.
 
-- Change foo to bar in matching filenames:
+- Rename files using a Perl Common Regular Expression (substitute 'foo' with 'bar' wherever found):
 
-`rename {{'s/foo/bar/'}} {{*.txt}}`
+`rename {{'s/foo/bar/'}} {{\*}}`
 
-- Convert to lower case:
+- Dry-run, display which renames would occur without performing them:
```
waldyrious: I'd use a dash rather than a comma here.

pages/common/srm.md
```diff
@@ -1,17 +1,19 @@
 # srm
 
 > Securely remove files or directories.
-> Overwrites the existing data one or multiple. Drop in replacement for rm.
+> Overwrites the existing data one or multiple times. Drop in replacement for rm.
 
-- Remove a file after overwriting (single pass, 7 pass, 35 pass):
+- Remove a file after a single pass overwriting with random data:
```
waldyrious: I'd suggest either "a single-pass overwriting" or "a single pass of overwriting"

pages/common/srm.md
```diff
@@ -1,17 +1,19 @@
...
 
 `srm -s {{/path/to/file}}`
+
+- Remove a file after seven passes overwriting it with random data:
```
waldyrious: Same as above.

pages/common/srm.md
```diff
@@ -1,17 +1,19 @@
...
+
 `srm -m {{/path/to/file}}`
-`srm {{/path/to/file}}`
 
-- Scurely remove recursively a directory and all it's subdirectories:
+- Recursively remove a directory and contents overwriting each file with a single pass of random data:
```
waldyrious: Suggestion: "and its contents,"

waldyrious:
I've done a detailed review of the first few diffs, and added minor comments inline, which I assume are uncontroversial. I'll stop there, assuming the rest of the changes are in the same vein, to avoid cluttering this PR with too much nitpicking. You made very good calls IMO in choosing what examples to keep and rephrasing some descriptions.

igorshubovych:
@waldyrious, it seems like @notpeter will reply only tomorrow. What if we merge it now, and you will create another PR with your changes?

waldyrious:
@igorshubovych let's see what @notpeter has to say. IMO we can wait until tomorrow.

notpeter:
@waldyrious All you suggestions improved clarity so I implemented them all.   Keep the comments coming.

waldyrious:
Thanks @notpeter. I won't push it, though -- from what I've seen, I trust your judgment and am :+1: with merging this ASAP.

igorshubovych:
@notpeter great thanks on taking responsibility and doing this! well done!

PR #700, opened 2016-01-20
OpenSSL condensing.

notpeter:
Shorten OpenSSL to 24 lines. Split keygen and csr to distinct examples. Standardize pathing. Remove commands which are just md5 verifying csr/key/csr.

The </dev/null are there because otherwise openssl waits for more input (batch mode) like a jerk until you send it an EOF.

Also switch to signing the CSR with SHA256 because SHA1 cert signing is deprecated for most browsers.

### DIFF
--- a/pages/common/openssl.md
+++ b/pages/common/openssl.md
@@ -10,9 +10,9 @@
 
 `openssl req -new -sha256 -key {{filename.key}} -out {{filename.csr}}`
 
-- Read contents of a signed certificate:
+- Generate a self-signed certificate from a certificate signing request valid for some number of days:
 
-`openssl x509 -text -noout -in {{certificate.crt}}`
+`openssl x509 -req -days {{days}} -in {{filename.csr}} -signkey {{filename.key}} -out {{filename.crt}}`
 
 - Display the certificate presented by an SSL/TLS server:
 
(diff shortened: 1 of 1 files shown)


rubenvereecken:
Quite the rework. I approve though. Thanks for explaining the need for consulting `/dev/null`.

:+1:

notpeter:
I actually replaced the viewing a csr example with an example for creating a self-signed certificate (a very common use-case for openssl) because I had to do it at work today and was bummed my previous revision was missing, so I amended my PR. ;)

igorshubovych:
Hey @notpeter, could you plz reply to my private message in Gitter?

pages/common/openssl.md
```diff
@@ -1,35 +1,23 @@
...
-`openssl x509 -noout -modulus -in {{path/to/certificate.crt}} | openssl md5`
-
-- Check MD5 hash of a private key:
-
-`openssl rsa -noout -modulus -in {{path/to/pivate.key}} | openssl md5`
-
-- Check MD5 hash of a CSR file:
-
-`openssl req -noout -modulus -in {{path/to/CSR.csr}} | openssl md5`
+`openssl s_client -connect {{google.com}}:{{443}} -showcerts </dev/null`
```
plessbd: should it be consistent and use {{host}}:{{port}}  or {{google.com}}:{{443}} for both instead of switching?

ostera:
Looking good! Merging before it gets stale/conflict-ish. Thanks @notpeter :)

PR #704, opened 2016-01-21
Added more notes on tldr-lint

rubenvereecken:
Added a note on our linter/formatter right after mentioning our syntax. It's also somewhere in the step-by-step instructions but it looks like people hardly read those. This way it's near the bits that people _probably_ read if they read the page at all. No guarantees.


waldyrious:
:+1: :smile:

felixonmars:
:+1:

PR #705, opened 2016-01-21
Add mmv

oxguy3:
[mmv](http://ss64.com/bash/mmv.html) is a cross-platform tool for renaming/moving/copying lots of files using wildcard patterns. It's a bit of a complex, nuanced tool, and I couldn't really come up with many good generic examples. I think I have a good mix of examples that demonstrate what it can do; they're just a bit specific to odd cases. I also was kind of unsure where to put the curly braces in some places.

### DIFF
--- a/pages/common/mmv.md
+++ b/pages/common/mmv.md
@@ -1,0 +1,19 @@
+# mmv
+
+> Move and rename files in bulk.
+
+- Rename all files with a certain extension to a different extension:
+
+`mmv "*{{.old_extension}}" "#1{{.new_extension}}"`
+
+- Copy report6part4.txt to ./french/rapport6partie4.txt along with all similarly named files:
+
+`mmv -c {{"report*part*.txt"}} {{"./french/rapport#1partie#2.txt"}}`
+
+- Append all .txt files into one file:
+
+`mmv -a "*.txt" {{"all.txt"}}`
+
+- Convert dates in filenames from "M-D-Y" format to "D-M-Y" format:
+
+`mmv "[0-1][0-9]-[0-3][0-9]-[0-9][0-9][0-9][0-9].txt" "#3#4-#1#2-#5#6#7#8.txt"`


pages/common/mmv.md
```diff
@@ -0,0 +1,19 @@
+# mmv
+
+> Move and rename files in bulk.
+
+- Rename all files with a certain extension to a different extension:
+
+`mmv "*{{.old_extension}}" "#1{{.new_extension}}"`
+
+- Copy report6part4.txt to ./french/rapport6partie4.txt along with all similarly named files:
```
rubenvereecken: I feel we can do better with the proposed example. This stuff is _vague_. Not an easy comment to work with, I know. I got so distracted by the example that I didn't see immediately that the keyword here is "copy".

pages/common/mmv.md
```diff
@@ -0,0 +1,19 @@
...
+
+`mmv -c {{"report*part*.txt"}} {{"./french/rapport#1partie#2.txt"}}`
+
+- Append all .txt files into one file:
+
+`mmv -a "*.txt" {{"all.txt"}}`
+
+- Convert dates in filenames from "M-D-Y" format to "D-M-Y" format:
+
+`mmv "[0-1][0-9]-[0-3][0-9]-[0-9][0-9][0-9][0-9].txt" "#3#4-#1#2-#5#6#7#8.txt"`
```
rubenvereecken: I think we should put double braces around these.

jazzonmymind:
seems interesting i hadn't heard of it, it reminds me of mustache.js, like maybe a more simple variant could be written using mustache(comes in a few flavors i think) and shell piping.

rubenvereecken:
I wanted to use this today but I had no tldr page to aid me :(

rubenvereecken:
Yay I've managed. It didn't support non-greedy wildcards though (`*?`), looks like it's non-greedy by default.

ostera:
@oxguy3 thanks a lot man! Fixed the last small details here: https://github.com/tldr-pages/tldr/commit/5f21c67ead37289ef8e868ea43a3a7225efcad30

PR #706, opened 2016-01-22
Add Redshift

oxguy3:
[Redshift](http://jonls.dk/redshift/) is a tool for adjusting your computer screen's color temperature to match your surroundings. (Same concept as [f.lux](https://justgetflux.com/), if you've heard of that, but it has more config options and you run it from the command-line.)

I used a lot of specific examples rather than generic examples, because it seemed like it'd be harder to succinctly get the point across otherwise (given how odd the formatting is on a lot of redshift's params).

### DIFF
--- a/pages/common/redshift.md
+++ b/pages/common/redshift.md
@@ -1,0 +1,27 @@
+# redshift
+
+> Adjust the color temperature of your screen according to your surroundings.
+
+- Turn on Redshift with 5700K temperature during day and 3600K at night:
+
+`redshift -t {{5700}}:{{3600}}`
+
+- Turn on Redshift with a manually-specified custom location:
+
+`redshift -l {{latitude}}:{{longitude}}`
+
+- Turn on Redshift with 70% screen brightness during day and 40% brightness at night:
+
+`redshift -b {{0.7}}:{{0.4}}`
+
+- Turn on Redshift with custom gamma levels (between 0 and 1):
+
+`redshift -g {{red}}:{{green}}:{{blue}}`
+
+- Turn on Redshift with temperature transitions disabled:
+
+`redshift -r`
+
+- Turn on Redshift with a constant unchanging color temperature:
+
+`redshift -O {{temperature}}`
--- a/pages/common/redshift.md
+++ b/pages/common/redshift.md
@@ -18,10 +18,6 @@
 
 `redshift -g {{red}}:{{green}}:{{blue}}`
 
-- Turn on Redshift with temperature transitions disabled:
-
-`redshift -r`
-
 - Turn on Redshift with a constant unchanging color temperature:
 
 `redshift -O {{temperature}}`


pages/common/redshift.md
```diff
@@ -0,0 +1,27 @@
...
+
+`redshift -b {{0.7}}:{{0.4}}`
+
+- Turn on Redshift with custom gamma levels (between 0 and 1):
+
+`redshift -g {{red}}:{{green}}:{{blue}}`
+
+- Turn on Redshift with temperature transitions disabled:
+
+`redshift -r`
```
rubenvereecken: Wouldn't `-t` imply `-r`, or did I get temperature wrong? Curious flux user here.

oxguy3: The `-r` flag just makes it so that Redshift changes the color temperature instantly, instead of gradually fading to the proper temperature. It's not a terribly important flag; it could probably be dropped from this page, actually.

oxguy3: And to be clear, I listed all of the flags in separate commands, but you can actually combine all of them (well, using the `-O` flag while also having separate night/day settings would be a bit pointless). And also, pretty much every flag that accepts multiple colon-delimited parameters, will also accept a single parameter (i.e. instead of `-b 0.7:0.4` for night/day brightness, you could do `-b 0.7` for persistent brightness. same basic deal for `-g` and `-t`) -- perhaps this should be conveyed in the tldr page somehow (couldn't think of a good way when I was writing it).

rubenvereecken:
Is there a way to adjust transition time length? I like smooth transitions.

pages/common/redshift.md
rubenvereecken: You say every parameter accepts a single constant value. What's the difference between `-t` and `-O` then? The latter looks like the constant version of the former.

oxguy3:
Not that I know of, seems to just be a boolean. But that might just be cuz I'm on OS X, where I have the "quartz" transition method instead of the "randr" or "vidmode" methods; maybe those methods have options for that.

rubenvereecken:
You raised a valid point in your inline comment but all you could do is have an example with a constant, but that's losing in terms of expressiveness, in terms of power of the tool. I say we leave it like it is for now. Just address that inline comment and I'll happily merge this.

oxguy3:
What inline comment?

rubenvereecken:
Ugh didn't see it was on an outdated diff. I'll merge it, but I'm still curious. I'll quote the question:

> You say every parameter accepts a single constant value. What's the difference between -t and -O then? The latter looks like the constant version of the former. Shouldn't we just be able to use -t with a single temperature value?

oxguy3:
The `-t` flag means you're setting the night and day (or the night+day) temperatures for Redshift to be in constant shifting mode. In addition to shifting temperature, Redshift will also be shifting gamma and other stuff. With the `-O` flag, it's just a pure constant, and it shifts instantly (which means it's also the same as passing the "transitions disabled" flag). I think Redshift might also run in the background when you don't use the `-O` flag maybe?? (I pretty much always use the `-O` flag so I can't recall -- currently away from normal computer) Whereas using the `-O` flag runs in the current terminal, allowing you to Ctrl+C to instantly kill it.

Basically, at the very least, `-O` is a very useful shortcut to replace two flags with one, and I think it might do more beyond that. I'll have to play around with this when I'm at my normal computer.

PR #707, opened 2016-01-22
Add watch

oxguy3:
[watch](http://linux.die.net/man/1/watch) is a simple tool for monitoring stuff by repeatedly running a command and showing the output. Just three simple examples to cover the most useful stuff.

### DIFF
--- a/pages/common/watch.md
+++ b/pages/common/watch.md
@@ -1,0 +1,15 @@
+# watch
+
+> Execute a program periodically, showing output fullscreen.
+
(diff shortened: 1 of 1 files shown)


waldyrious:
Nice one! In the last example, maybe use "differences" rather than "changes", so people will better memorize the -d option?

oxguy3:
(This PR is actually fine, never mind Travis. An bad page in `master` is causing the build to fail; PR #708 has the fix.)

felixonmars:
:+1: LGTM

PR #457, opened 2015-12-29
add pages: seq, uniq; Add -d option to ls; Add -i option to w

raamana:
1. add page: seq 
2. add page: uniq
3. add -d option to ls, 
4. an example to list hidden files & folders

PR #710, opened 2016-01-22
w: add -i example, uniq: additions, seq: add page

notpeter:
Squashed, rebased version of #457 which fixes the outstanding merge conflicts and drops the edits to `ls` because they were shell globbing examples.

### DIFF
--- a/pages/common/uniq.md
+++ b/pages/common/uniq.md
@@ -1,6 +1,6 @@
 # uniq
 
-> Report or omit repeated lines.
+> Output the unique lines from the given input or file.
 
 - Display each line once:
 
--- a/pages/common/seq.md
+++ b/pages/common/seq.md
@@ -1,0 +1,15 @@
+# seq
+
+> Output a sequence of numbers to stdout.
+
+- Sequence from 1 to 10:
+
+`seq 10`
+
+- Every 3rd number from 5 to 20:
+
+`seq 5 3 20`
+
+- Separate the output with a space instead of a newline:
+
+`seq -s " " 5 3 20`
--- a/pages/osx/w.md
+++ b/pages/osx/w.md
@@ -1,0 +1,16 @@
+# w
+
+> Show who is logged on and what they are doing.
+> Print user login, TTY, remote host, login time, idle time, current process.
+
+- Show logged-in users info:
+
+`w`
+
+- Show logged-in users info without a header:
+
+`w -h`
+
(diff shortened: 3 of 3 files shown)


notpeter:
I also added the punctuation which caused test to fail, squashed and repushed.

raamana:
thanks.

rubenvereecken:
You can _do_ this? Multiple authors? What is this magick.

pages/common/w.md
```diff
@@ -10,3 +10,7 @@
 - Show logged-in users info without a header:
 
 `w -h`
+
+- Show info about logged-in users, sorted their idle time:
```
rubenvereecken: sorted _by_

igorshubovych:
`-i` has different meaning for OSX and Linux versions

notpeter:
Put `w -i` into new osx/w.md and changed it to "sorted by", left common/w.md unchanged.

igorshubovych:
it is actually interesting. should we move page from common to linux?

notpeter:
The old example truly is "common" I double checked and Linux/Solaris both support those same limited flags so I think we should leave it as is.

pages/common/uniq.md
```diff
@@ -1,6 +1,7 @@
 # uniq
 
-> Report or omit repeated lines.
+> Output the unique lines from the given input or file.
+> Compare the adjacent files in input (file or standard input) and outputs unique lines.
```
igorshubovych: Does it really makes sense to have such long description?
I like the original description or just a first line of your description. I think that 2nd line of your description does not add too much and hence I'd rather remove it.

raamana: I agree with @igorshubovych

raamana:
Great work and discussion guys., Lets get this merged already. Lets not defeat the purpose of this project with too long a discussion..

rubenvereecken:
@raamana we've got an agreement among contributors to only merge after two :+1:s, which can create some delays in times of less activity. Anyway, here goes mine after @igorshubovych's concerns are addressed: :+1:

pages/common/uniq.md
waldyrious: :+1: for using the first line of the new description.

notpeter: Done.

rubenvereecken:
Et voila, finalement :tada: 

Thanks @notpeter!

PR #711, opened 2016-01-22
Add license

oxguy3:
[license](https://github.com/nishanths/license) is a tool for quickly/easily adding a license file to an open source project.

In the last example, you might wonder why I say "explicitly-set name and year". By default, license pulls your name from your git/mercurial config files (or the environment variable `LICENSE_FULL_NAME`), and just uses the current year. Thus, you only need to use those command flags if you specifically don't want your git/mercurial name or the current year. I can't really think of a succinct way to explain that in the tldr page, unfortunately.

### DIFF
--- a/pages/common/license.md
+++ b/pages/common/license.md
@@ -16,4 +16,4 @@
 
 - Create a license with explicitly-set name and year:
 
-`license --name {{custom_name}} --year {{custom_year}} {{license_name}}`
+`license --name {{author}} --year {{release_year}} {{license_name}}`
--- a/pages/common/license.md
+++ b/pages/common/license.md
@@ -1,0 +1,19 @@
+# license
+
+> Create license files for open-source projects.
+
+- Create a license:
+
+`license {{license_name}}`
+
+- Create a license with custom filename:
+
+`license -o {{filename.txt}} {{license_name}}`
+
+- List all locally available licenses:
+
+`license ls`
+
+- Create a license with explicitly-set name and year:
+
+`license --name {{custom_name}} --year {{custom_year}} {{license_name}}`


pages/common/license.md
```diff
@@ -0,0 +1,19 @@
...
+
+`license -o {{filename.txt}} {{license_name}}`
+
+- List all locally available licenses:
+
+`license ls`
+
+- Create a license with explicitly-set name and year:
+
+`license --name {{custom_name}} --year {{custom_year}} {{license_name}}`
```
igorshubovych: I would rather call tokens something like {{author_name}} and {{release_year}}

igorshubovych:
Good one!

pages/common/license.md
rubenvereecken: Make that just {{author}}?

igorshubovych: :+1:

rubenvereecken:
:tada:

pages/common/license.md
```diff
@@ -0,0 +1,19 @@
+# license
+
+> Create license files for open-source projects.
+
+- Create a license:
```
waldyrious: @oxguy3 what's the default filename?

waldyrious: Also, implicitness of author and year should probably be noted here. Does this work only for mercurial repos? What else does it support?

igorshubovych: @waldyrious 

> @oxguy3 what's the default filename? Or is this output to stdout?

It outputs to stdout

waldyrious: @igorshubovych then we should probably mention that :)

rubenvereecken: Isn't that the usual default behavior though? Especially with commands that have got a `-o` parameter.

igorshubovych: @rubenvereecken that's right. but probably it makes sense to mention this [algorithm](https://github.com/nishanths/license#customize-name-and-year-on-the-license) somehow, if it is possible to do it shortly.

waldyrious: @igorshubovych I would have expected that, but the description of the example following this suggests otherwise. So it's best to be explicit, IMO.

rubenvereecken: That algorithm is lengthy and not really easy to condense. That environmental variable for example I wouldn't mention, that's behavior that's set up once by a person who knows this tool and figures she needs it more frequently. That sounds beyond the tldr experience. The part about the git config could be mentioned because that's easy to get the author extracted.

@waldyrious was that directed at my comment?  I don't see how any following examples contradict? If I get my output to stdout instead of a file I just stream it into a file, no harm done. Of course I'm not really against `and writes to stdout` or something but it's already an often-made assumption.

waldyrious: @rubenvereecken you're right, I @-mentioned the wrong person :) What I meant was that the next example, "create a license with **custom** filename" (emphasis mine) suggests that the vanilla one uses a default filename, so the assumption of output to stdout is shaken. If the description instead said something like "create a license and it save to a file", then we'd reinforce the assumption that the plain calling of the command outputs directly to stoud. It's not a big deal, of course, but we should strive to reduce confusion wherever possible.

rubenvereecken: Well explained!

PR #712, opened 2016-01-22
Add archey

oxguy3:
Archey is a tool that displays some basic system information alongside an ASCII version of your OS's logo. The tool exists on both OS X and Linux, but they're separate codebases that work slightly differently, so they get separate pages.

The OS X version has a couple of flags for tweaking the output, while the Linux version has absolutely no options (and I would have never bothered adding it if I wasn't also simultaneously adding the OS X one). I used the long forms of the flags for the OS X page, even though every flag does have a shortcut, simply because I don't think anyone uses archey frequently enough to memorize the flags.

### DIFF
--- a/pages/osx/archey.md
+++ b/pages/osx/archey.md
@@ -1,0 +1,19 @@
+# archey
+
+> Simple tool for stylishly displaying system information.
+
+- Show system information in stylish output:
+
+`archey`
+
+- Show system information without colored output:
+
+`archey --nocolor`
+
+- Show system information, using MacPorts instead of Homebrew:
+
+`archey --macports`
+
+- Show system information without IP address check:
+
+`archey --offline`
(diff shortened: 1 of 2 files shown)


rubenvereecken:
It's good you think of intuitiveness of the flags. Good call.

:+1:

pages/osx/archey.md
```diff
@@ -0,0 +1,19 @@
+# archey
+
+> Simple tool for stylishly displaying system information.
+
+- Show system information in stylish output:
```
rubenvereecken: You imply the linux archey is not very stylish :P?

oxguy3: Hehe, no, it'd probably make sense to remove that "in stylish output" bit from the OS X one too.

rubenvereecken:
Ah, you updated. Great stuff

PR #713, opened 2016-01-22
Add tabula-extractor

oxguy3:
[Tabula](https://github.com/tabulapdf/tabula-extractor) is a tool for attempting to extract tabular data from PDF files.

### DIFF
--- a/pages/common/tabula.md
+++ b/pages/common/tabula.md
@@ -4,7 +4,11 @@
 
 - Extract all tables from a PDF to a CSV file:
 
-`tabula {{file.pdf}} -o {{file.csv}}`
+`tabula -o {{file.csv}} {{file.pdf}}`
+
+- Extract all tables from a PDF to a JSON file:
+
+`tabula --format JSON -o {{file.json}} {{file.pdf}}`
 
 - Extract tables from pages 1, 2, 3, and 6 of a PDF:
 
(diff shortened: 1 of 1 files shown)


oxguy3:
Six examples may seem excessive; however, in practice, all of these options are frequently used. Extracting tables from PDFs is very difficult, and Tabula often gives imperfect output. Very frequently, you'll have to try it several different ways before you get a satisfactory result.

pages/common/tabula.md
```diff
@@ -0,0 +1,27 @@
...
+
+`tabula --format JSON -o {{file.json}} {{file.pdf}}`
+
+- Extract tables from pages 1, 2, 3, and 6 of a PDF:
+
+`tabula --pages {{1-3,6}} {{file.pdf}}`
+
+- Extract tables from page 1 of a PDF, guessing which portion of the page to examine:
+
+`tabula --guess --pages {{1}} {{file.pdf}}`
```
rubenvereecken: Not sure about the difference with the one above here. Won't it guess in the above example as well? Will it guess more loosely now?

oxguy3: If you don't specify `--guess` or `--area`, it'll just do the entire page. What's generally preferrable is to do a rectangular area within the page. If every page of your PDF has a header, and you don't use one of those flags, that header will probably show up over and over in your CSV, which is generally undesirable.

rubenvereecken: So if you extract all tables from a PDF it'll expect the contents to exist solely out of tables?

oxguy3: If you don't provide an area or tell it to guess, then yeah basically. Extracting tables from PDFs is hard :stuck_out_tongue:

rubenvereecken:
:+1: fancy

PR #714, opened 2016-01-22
Add atom

oxguy3:
There's a nice little command-line tool `atom` for GitHub's [Atom](https://atom.io) text editor. The only things a common user would need from it is "open thing" and "open thing in new window", so that's all I've documented. (There are a lot of other options available for `atom`, but most of them are only really useful to Atom's developers.)

### DIFF
--- a/pages/common/atom.md
+++ b/pages/common/atom.md
@@ -1,0 +1,11 @@
+# atom
+
+> A hackable text editor for the 21st century.
+
+- Open a file or folder:
+
+`atom {{path/to/file/or/folder}}`
+
+- Open a file or folder in a new window:
+
+`atom -n {{path/to/file/or/folder}}`
--- a/pages/common/atom.md
+++ b/pages/common/atom.md
@@ -1,6 +1,6 @@
 # atom
 
-> A hackable text editor for the 21st century.
+> A cross-platform text editor.
 
 - Open a file or folder:
 


pages/common/atom.md
```diff
@@ -0,0 +1,11 @@
+# atom
+
+> A hackable text editor for the 21st century.
```
igorshubovych: This is not a place for their motto. It is a place for description of command.
So it should be rather 
```md
> Atom text editor
```

waldyrious: Maybe a little more descriptive? I.e. explain what makes it unique (configurability, based on x, aimed at devs of kind y...)

oxguy3: How about "A free cross-platform text editor."

rubenvereecken: Most tools here are free... or do you mean free in the ideological sense?

oxguy3: Uh, I guess I meant both, but yeah that is silly. How about just "A cross-platform text editor."

rubenvereecken: Boring and correct :+1:

oxguy3:
(this PR is fine; something else is wrong with master causing the build fail)

rubenvereecken:
Node just changed version I see, down from 5.x to 0.10. @igorshubovych is this something weird and temporary or will this require fixing? We should add a `.nvmrc` file if this might occur again.

pages/common/atom.md
waldyrious: I still think it would be nice to highlight its main appeal in relation to other cross-platform text editors, but this isn't a strong opinion.

igorshubovych:
I think it was error on Travis because new Node was released.

igorshubovych:
Thank you.

PR #715, opened 2016-01-22
Add x_x

oxguy3:
[x_x](https://github.com/kristianperkins/x_x) is a command-line tool for viewing Excel and CSV files in the command-line (it builds an ASCII table and pipes to less). Here's a tldr page for it with three simple examples that cover just about everything x_x can do.


rubenvereecken:
I like it! :+1:

igorshubovych:
Thanks!

PR #716, opened 2016-01-22
less: add -S option

chuanconggao:


igorshubovych:
It is not listed in OSX's man. So I assume it does not work on OSX. In this case if you want it to be added you should move this page to pages/linux/ folder and copy it without the option in OSX.

But is this option really common, useful for everybody? I lived 10 years on Linux and OSX without it. Again, this does not tell it is not, it might be useful, I just don't know it.

rubenvereecken:
Thanks for editing although you forgot to remove the common page. I'm closing this PR as the example doesn't add enough value to warrant making the page longer.

PR #719, opened 2016-01-24
Reworked linux ss page

### LINKED ISSUE #687: Pages with multiple commands per example

rubenvereecken:
An effort to close #687. This page was in a pretty bad condition. On the bright side, I learned a whole new command today.

I usually veer towards practical examples. This command especially calls for them because it has its own query language. Like tcpdump has one, kinda similar. I went out of my way and tried to come up with a few practical examples but my so-called expertise is limited to networks, I don't know much about Unix sockets so haven't managed to dredge up an example on those. 

This is somewhat more practically oriented than the average page. Lay it on me but be gentle, I didn't want to rewrite aaaall of the page just yet. Even though I nearly did.

### DIFF
--- a/pages/linux/ss.md
+++ b/pages/linux/ss.md
@@ -10,7 +10,7 @@
 
 `ss {{state/exclude}} {{bucket/big/connected/synchronized/...}}`
 
-- Show all TCP sockets connected to the local HTTPS port:
+- Show all TCP sockets connected to the local HTTPS port (443):
 
 `ss -t src :{{443}}`
 
@@ -18,10 +18,10 @@
 
 `ss -pt dst :{{ssh}}`
 
-- Show all UDP sockets connected on a local port within a range:
+- Show all UDP sockets connected on specific source and destination ports:
 
-`ss -u 'sport >= :{{lowest_port}} and sport <= :{{highest_port}}'`
+`ss -u 'sport == :{{source_port}} and dport == :{{destination_port}}'`
 
 - Show all TCP IPv4 sockets locally connected on the subnet 192.168.0.0/16:
 
-`ss -4t {{192.168/16}}`
+`ss -4t src {{192.168/16}}`
--- a/pages/linux/ss.md
+++ b/pages/linux/ss.md
@@ -4,7 +4,7 @@
 
 - Show all TCP/UDP/RAW/UNIX sockets:
 
-`ss -a {{-t/-u/-w/-x}}`
+`ss -a -{{t|u|w|x}}`
 
 - Filter TCP sockets by states, only/exclude:
 
--- a/pages/linux/ss.md
+++ b/pages/linux/ss.md
@@ -4,7 +4,7 @@
 
 - Show all TCP/UDP/RAW/UNIX sockets:
 
-`ss -a -{{t|u|w|x}}`
+`ss -a {{-t|-u|-w|-x}}`
 
 - Filter TCP sockets by states, only/exclude:
 
--- a/pages/linux/ss.md
+++ b/pages/linux/ss.md
@@ -2,26 +2,26 @@
 
 > Utility to investigate sockets.
 
-- Dump all TCP/UDP/RAW/UNIX sockets:
+- Show all TCP/UDP/RAW/UNIX sockets:
 
 `ss -a {{-t/-u/-w/-x}}`
-
-- Show process(es) that using socket:
-
-`ss -p`
 
 - Filter TCP sockets by states, only/exclude:
 
 `ss {{state/exclude}} {{bucket/big/connected/synchronized/...}}`
 
-- Filter sockets by address and/or port:
+- Show all TCP sockets connected to the local HTTPS port:
 
-`ss -t dst 1.2.3.4:80`
-`ss -u src 127/8`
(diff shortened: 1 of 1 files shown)


pages/linux/ss.md
```diff
@@ -2,26 +2,26 @@
 
 > Utility to investigate sockets.
 
-- Dump all TCP/UDP/RAW/UNIX sockets:
+- Show all TCP/UDP/RAW/UNIX sockets:
 
 `ss -a {{-t/-u/-w/-x}}`
```
waldyrious: We need to decide if we are to allow this, what character to use (`/`, `!`, `,`), and where this mechanism is admissible. In terms of formatting, I'd suggest we adhere to the [docopt](http://docopt.org/) standard as much as possible, i.e. use `|` for separating mutually exclusive options.

pages/linux/ss.md
```diff
@@ -2,26 +2,26 @@
...
-- Show process(es) that using socket:
-
-`ss -p`
-
 - Filter TCP sockets by states, only/exclude:
 
 `ss {{state/exclude}} {{bucket/big/connected/synchronized/...}}`
 
-- Filter sockets by address and/or port:
+- Show all TCP sockets connected to the local HTTPS port:
```
waldyrious: I'd suggest mentioning 443 in the description, e.g. "to the local HTTPS port (443):", to make it explicit.

pages/linux/ss.md
```diff
@@ -2,26 +2,26 @@
...
+`ss -pt dst :{{ssh}}`
+
+- Show all UDP sockets connected on a local port within a range:
 
-`ss -t dst 1.2.3.4:80`
-`ss -u src 127/8`
-`ss -t 'dport >= :1024'`
-`ss -x "src /tmp/.X11-unix/*"`
-`ss -t state established '( dport = :ssh or sport = :ssh )'`
+`ss -u 'sport >= :{{lowest_port}} and sport <= :{{highest_port}}'`
```
waldyrious: what does the `s` in `sport` stand for? We should indicate that in the description if possible.

rubenvereecken: I'll change this example to use both sport and dport,  should be very intuitive that way

pages/linux/ss.md
rubenvereecken: :+1:

pages/linux/ss.md
rubenvereecken: This whole example bothers me tbh. I'd have preferred to describe 3 or so of these in other examples that also introduce filters, but I can't come up with a good Unix example so all I can manage is tcp and udp

waldyrious: I don't think we should outright ban this construction; in cases like this where the commands are identical except for the option, it seems justifiable. But we need to agree that it's acceptable and when, so we don't impose restrictions on others that we don't follow universally.

pages/linux/ss.md
waldyrious: Good call :+1:

pages/linux/ss.md
rubenvereecken: The ipv4/6 example I originally fixed by having `ss -{{4|6}}`. The dash can be inside the braces even, but I'd insist on having `|` for separators we've had a couple of times.

waldyrious: I'm not sure I understand your wording, but it seems we are in agreement. Including the dash next to each option may be more intuitive, though.

rubenvereecken:
@waldyrious I reworked that first example. Next to a discussion about a convention, would this be good to go?

waldyrious:
@rubenvereecken yes, but as I said in the inline thread: "Including the dash next to each option [i.e. inside the `{{}}`] may be more intuitive". Not a dealbreaker, so if you prefer the way it is, I'm ok with merging.

rubenvereecken:
I've been in doubt about this myself. I'll change to the more intuitive.

PR #720, opened 2016-01-24
Reworked wpa_cli to adhere to one-example rule

rubenvereecken:
Less invasive than the `ss.md` page, but here you go. The last one needed for #687.


waldyrious:
:+1: LGTM, thanks for tackling this!

PR #721, opened 2016-01-24
ssh reverse tunneling added

abshammeri:
I've added an example to do a reverse tunneling through ssh. Reverse tunneling used to access a personal machine from a remote server. It common to use this feature to bypass corporate firewalls so the user can access its machine from remote server. This example will complete the tunneling options list:
* -L
* -D
* -R (the added item)

Thanks,

### DIFF
--- a/pages/common/ssh.md
+++ b/pages/common/ssh.md
@@ -30,3 +30,7 @@
 - SSH enable agent forward:
 
 `ssh -A {{username}}@{{remote_host}}`
+
+- SSH reverse tunneling: Open a listening port on the remote server to forward it back to localhost (remote-server.com:22 to your-personal-machine:2222)
+
+`ssh -R {{2222}}:localhost:22 {{username}}@{{remote_host}}`


waldyrious:
Many thanks for your contribution, but this page is already quite long, and adding more commands makes it longer still. In the spirit of tl;dr, what changes do you think could make sense to *shorten* the page, even if we add -R? Maybe combine two examples, or remove one that's less common?

rubenvereecken:
Pretty cool example. There are some formatting issues (besides syntax) but there's a bigger problem here: We restrict our pages to have about 6 examples. We're even removing the "about" part of that statement and this page was already slightly beyond its limit. The example you propose is one too many. If examples are absolutely necessary, we replace an old one. I don't see which one we could drop.

I'd refuse this PR with a sore heart, but I'll have other contributors get their say in first. I'm sorry @abshammeri.

abshammeri:
It's fine guys, and sorry for the failed test, I did not notice that there is CI here, if you might add the automated testing step to the contribution page.

rubenvereecken:
Yeah, we should do that somewhere. I've been meaning to add that. Cheers @abshammeri

PR #723, opened 2016-01-25
Added multiple directories creating with mkdir

shybovycha:
As it says:

`$ mkdir -p app/{routes/{users,admins},views,helpers}`:

results with

```
app
├── helpers
├── routes
│   ├── admins
│   └── users
└── views
```

I believe directory names in the manpage could be more descriptive, thus will be glad for any recommendations.


shybovycha:
test this please

pages/common/mkdir.md
```diff
@@ -9,3 +9,7 @@
 - Create directories recursively (useful for creating nested dirs):
 
 `mkdir -p {{path}}`
+
+- Create multiple directories recursively (note the curly braces, marking directory set):
+
+`mkdir -p dir1/dir2/{dir3/{dir4,dir5},dir6,dir7}`
```
igorshubovych: It is the same as the previous example

waldyrious: Perhaps the previous example should include something resembling a real path with mutiple components, e.g. `{{path/to/directory}}`

rubenvereecken: With a new description, these are not the same. `Create a complicated directory hierarchy recursively` or something.

waldyrious: Oh, so that's an actual tree with multiple branches? Somehow I feel that's a bit out of scope for tldr...

shybovycha: @igorshubovych the option is the same; but the argument and the effect is different.

@waldyrious why you treat this as out of scope? I suppose this is quite a good alternative to _create multiple directories with one command_

waldyrious: I meant it feels a bit too complex for a project that aims to provide short lists of the most common uses of a command, not an exhaustive listing of what can be done with it. But I'm not opposed to adding this, just raising the question so that we make that decision explicitly. I'll let others chime in :)

igorshubovych:
I agree with @waldyrious.

This is a command-line trick which suits very well for a project like [cheat](https://github.com/chrisallenlane/cheat).

I am using CLI for approx. 10 years and I've never used anything like this. Knowing this command I would still prefer running a few `mkdir` commands in a row or I will write a simple script.

I do not think this command is for new users of CLI.

Additionally your description does not explain it very well. But that's not your fault, this is one of those cases which are very hard to formulate.

Also {} (curly braces) are used for tokens in tldr pages, so it adds additional confusion.

PR #725, opened 2016-01-26
Add pv

d33tah:


pages/common/pv.md
```diff
@@ -0,0 +1,27 @@
...
+
+`pv {{file}}`
+
+- Measure the speed and amount of data flow between pipes (`-s` is optional):
+
+`command1 | pv -s {{expected_amount_of_data_for_eta}} | command2`
+
+- Recompress a file, also see the uncompressed and recompressed file size:
+
+`pv -cN gz file.gz | gzcat | pv -cN uncompressed | xz | pv -cN xz > out.xz`
```
igorshubovych: What does `-cN` do? Does it make sense to have 2 examples instead for `-c` and `-N`?

pages/common/pv.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Attach to an already running process and see its file reading progress:
+
+`pv -d {{PID}}`
+
+- Read an erroneous file, skip errors as `dd conv=sync,noerror` would:
+
+`pv -EE {{path_to_faulty_media}} > image.img`
```
igorshubovych: Why you double `-E` option? I do not see this in their help.

d33tah:
@igorshubovych: Hi, thanks for reviewing my patch! According to `man pv`, `Specify -E twice to only report a read error once per file, instead of reporting each byte range skipped.` as for `-cN`, `-c` only makes little sense in use cases other than using multiple progress bars. `-N` is used for naming those, so that you can keep track of separate progress bars.

igorshubovych:
The whole example 
```md
`pv -cN gz file.gz | gzcat | pv -cN uncompressed | xz | pv -cN xz > out.xz`
```
is too complicated for tldr.

Could you think about more simple example?

d33tah:
@igorshubovych: How about:

`pv -cN compressed file.tar.gz | gzip -dc | pv -cN uncompressed > file.tar`

?

d33tah:
@igorshubovych: updated with a hopefully simpler example.

ostera:
@d33tah, still pretty convoluted. Do you think we could do something simpler?

`cat {{big_text_file}} | pv | grep {{pattern}}`

should be easy enough for anyone to grasp?

d33tah:
@ostera: the problem is that grep output would overwrite pv progress bar. How about `pv {{big_text_file}} | grep {{pattern}} > {{filtered_file}}`?

ostera:
Sounds about right!

d33tah:
@ostera: I just realized that this doesn't really address the point I was trying to make, which was double usage of pv in the same line. How about `pv -cN in {{big_text_file}} | grep {{pattern}} | pv -cN out > {{filtered_file}}`?

ostera:
@d33tah I'm on-board as long as we use simple commands everyone can understand without having to further look into other pages. That is, `grep ls cat | >` and similar sorts.

d33tah:
@ostera: so, `pv -cN in {{big_text_file}} | grep {{pattern}} | pv -cN out > {{filtered_file}}` is ok?

ostera:
Looks good to me. Also according to http://linux.die.net/man/1/pv there's no `-EE` flag, if that's a typo get it fixed and I'll merge it asap :)

d33tah:
@ostera: this is not the latest manpage.

ostera:
@d33tah link to the latest manpage?

ostera:
Nevermind, found it. I was wrong! D: Merging this now :+1:

ostera:
For further reference: http://man7.org/linux/man-pages/man1/pv.1.html

PR #726, opened 2016-01-26
Adds git-add -p instructions

sethwoodworth:


pages/common/git-add.md
```diff
@@ -17,3 +17,7 @@
 - Also add ignored files:
 
 `git add -f`
+
+- Add parts of a file interactively
+
+`git add -p {{PATHSPEC}}`
```
waldyrious: Can we use something other than "PATHSPEC"? It doesn't have to be 100% technically correct, just cover the most common cases, while being intelligible to beginners.

ostera: `path/to/file` should be enough.

igorshubovych: Maybe better `{{path/in/repo}}` or `{{path/in/project}}`.

waldyrious: I like @ostera's version more, I don't expect many people would try adding files from outside the project directory in order to justify @igorshubovych's caveat.

sethwoodworth:
Agreed with everyone above.
Changed to `path/to/file` which definitely reads better, fixed a small lint error, squashed and updated.

waldyrious:
Great :+1: LGTM!

rubenvereecken:
Cool, thanks.

PR #727, opened 2016-01-26
Adds nano editor instructions

sethwoodworth:
I'm pretty sure that these aren't the format you would like for CTRL + o.  Happy to rework if incorrect.


igorshubovych:
We already have discussion about this #506

PR #728, opened 2016-01-26
pwgen: add page

sylvainbannier:
`pwgen` is a linux password generator. 
`pwgen` defaults generate pronouceable passwords. I think this would be usefull to provide quick access to options generating a more secure password.

### DIFF
--- a/pages/linux/pwgen.md
+++ b/pages/linux/pwgen.md
@@ -1,0 +1,7 @@
+# pwgen
+
+> Generate passwords.
+
+- Generate 8 characters long random passwords with symbols :
+
+`pwgen -sy`


pages/linux/pwgen.md
```diff
@@ -0,0 +1,7 @@
+# pwgen
+
+> Generate passwords.
+
+- Generate 8 characters long random passwords with symbols :
+
+`pwgen -sy`
```
ostera: As per: http://linux.die.net/man/1/pwgen, you're missing the length (8). So a few samples could be:

```md
- Generate random password with symbols:

`pwgen -y {{length}}`

- Generate hard-to-memorize passwords:

`pwgen -s {{length}}`
```

pages/linux/pwgen.md
```diff
@@ -0,0 +1,7 @@
+# pwgen
+
+> Generate passwords.
```
ostera: According to their webpage, they are a tool to `generate pronounceable passwords`

rubenvereecken:
I tried the default and some of these are far from pronounceable, I feel cheated :(.

ostera:
Thanks @sylvainbannier for your time! All the fixes are available here: https://github.com/tldr-pages/tldr/commit/dc3ba725f9b706d12a8baec971ac63fcb567080b

PR #729, opened 2016-01-27
xclip: add page

Kjwon15:
xclip is very simple and useful.

### DIFF
--- a/pages/linux/xclip.md
+++ b/pages/linux/xclip.md
@@ -1,0 +1,11 @@
+# xclip
+
+> Copy STDIN to clipboard or print clipboard to STDOUT.
+
+- Copy output to clipboard:
+
+`echo 123 | xclip -i`
+
+- Paste clipboard:
+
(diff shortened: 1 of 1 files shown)


pages/linux/xclip.md
```diff
@@ -0,0 +1,11 @@
+# xclip
+
+> Copy STDIN to clipboard or print clipboard to STDOUT.
+
+- Copy output to clipboard:
+
+`echo 123 | xclip -i`
+
+- Paste clipboard:
```
ostera: This doesn't exactly paste the clipboard but rather output it, and then you can save it somewhere. So maybe the text should read `Print clipboard's content` ? Just a thought, ultimately you know what this command does best than I do (I'm more of an `pboard` user)

ostera:
This hasn't gotten any feedback yet, and it really is just a nitpick from me. We should be on the clear to merge.

Issue #686, opened 2016-01-16
Propose capping length of rendered output at 24 rows


notpeter:
I recently ran across a couple commands which I think have too many examples.  To combat this, I propose cap on rendered markdown output at 24 lines. Then in a standard height (25 rows) terminal we never scroll the first example off the top of the screen.

This leaves **plenty** of space for examples.  6 examples with a two line command description is 23 lines rendered. It'll even support 7 commands, if the command description is a single line and two of the example commands are nested under a common description.

Commands which would need to compacted to comply:
- grep (29 lines, 8 examples)
- ssh (26 lines, 7 examples)
- tar (26 lines, 7 examples)

This also makes the "only 5-6 examples" recommendation easy to enforce with Travis.

igorshubovych:
Interesting idea. I like it.

rubenvereecken:
Ugh sooo much repairing old pages nowadays. I should know.

24 feels a bit arbitrary, even though you've thought it out well. We render on more screens than just the 25-line terminals though. I doubt I've seen one lately. `ls` certainly renders nicely on mine. Then there's web clients and the like.

notpeter:
Default window size for PuTTY, cmd.exe and OSX Terminal are all still 80x25 and this is a project targeting inexperienced command line users.

waldyrious:
I agree with a cap, especially since it's a clear way to ensure we're fulfilling the most essential element of tl;dr, which is to be brief. I'll just add that two commands under the same example is a no-no (although case-by-case reasoning is always possible), but 6 examples are certainly enough to cover the basics of most commands.

rubenvereecken:
Capping at 6 is a good thing either way. We'll get rid of the multiple examples thing. I'll generate a list with the offending pages soon. 

This should be in the linter.

waldyrious:
Uh... I must've miscounted earlier, but

> 6 examples with a two line command description is 23 lines rendered.

doesn't seem to hold out (anymore? Did the recent formatting changes include more newlines? @rubenvereecken)

rubenvereecken:
I noticed the same thing. Nothing changed.

waldyrious:
@notpeter please enlighten us :)

notpeter:
It turns out there's a little bit of variation between how clients render the markdown. My count previously was based on the [tldr-cpp-client](https://github.com/tldr-pages/tldr-cpp-client) but things are slightly different with the node based client.

[...]

PR #730, opened 2016-01-27
Adding long format list sorted by reverse mod date

CBrowne:
This a pretty useful and common use of ls to display all files with the full details showing the last modified as the last file.

### DIFF
--- a/pages/common/ls.md
+++ b/pages/common/ls.md
@@ -21,3 +21,7 @@
 - Long format list sorted by size (descending):
 
 `ls -lS`
+
+- Long format list of all files, sorted by modification date (oldest first):
+ 
+`ls -ltr`
--- a/pages/common/ls.md
+++ b/pages/common/ls.md
@@ -23,5 +23,5 @@
 `ls -lS`
 
 - Long format list of all files, sorted by modification date (oldest first):
- 
+
 `ls -ltr`


CBrowne:
Does anything need changed for this to be pulled?

waldyrious:
Hi @CBrowne, thanks for the contribution :) I'm just wondering, is the use case for this example common enough to warrant expanding the page to 6 examples and almost 30 lines? Note that we're discussing the implementation of a 5-example, 25-line limit (#686).

CBrowne:
Ah @waldyrious I wasn't aware of the proposed limit. It's up to you guys! Thanks for making this tool, it is very handy :+1:

rubenvereecken:
Keep this PR here for a bit as the discussion isn't finished yet.

igorshubovych:
I personally don't like that you introduce to options here: `-t` and `-r`.
It is against guidelines.

waldyrious:
@igorshubovych as I've said elsewhere, guidelines are just that: general hints that we can deviate from if justified. We shouldn't be applying them blindly as strict rules. In this case I would say that introducing two options may make sense if the example is common, we need compactness *and* both options are explained in the text using the first letter so that it's clear what they refer to (rather than just presenting a cryptic command that one has to memorize). The last criterion isn't applied here, but I'd say we first need to decide whether we want to expand the page in the first place. Hopefully the discussion at #686 is finalized soon.

CBrowne:
Any update on this?

Thanks

ostera:
I think it doesn't hurt and we are still not going over the 6-commands soft-limit. I'll merge for usefulness and maybe we can revisit if someone else proposes a new tldr for this page? Thanks @CBrowne for the contribution!

PR #732, opened 2016-01-27
grep: add -i option

keune:


waldyrious:
:+1:

felixonmars:
LGTM, thanks!

igorshubovych:
9 examples? too much...

waldyrious:
Actually 8, but good point. We definitely need to implement the line count limit (#686) so we can start dealing with this issue.

PR #735, opened 2016-01-28
Client certificate in cURL example

imjacobclark:
Passing client certificates to access protected resources is something I find myself commonly doing in cURL and is something difficult to decipher solely from the `man` pages.

### DIFF
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -30,3 +30,7 @@
 - Pass a user name and password for server authentication:
 
 `curl -u myusername:mypassword {{http://localhost}}`
+
+- Pass client certificate and key for a secure resource:
+
+`curl -v –key {{key.pem}} –cacert {{ca.pem}} –cert {{client.pem}} -k {{https://localhost}}`


rubenvereecken:
This definitely is a good example so don't get me wrong here but it feels too advanced. Curl is a powerful tool with a lot of options so we're having trouble keeping pages like this small, a goal we're very much dedicated to.

I know this is probably useful for some people but the other examples we already have are simpler and probably more useful to the newbie public. I propose we don't merge this, with a sore heart.

igorshubovych:
@rubenvereecken +1

We thank for your time and contribution.

These are part of our [guidelines](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines):
1. Focus on the 5 or 6 most common usages. It's OK if the page doesn't cover everything; that's what man is for.
2. When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness.

IMO your example is useful, but not for new command-line users.

igorshubovych:
@rubenvereecken actually i think that using certificate is quite basic and common example, so maybe we can make an exception just for curl (same as we did for `ls`)

imjacobclark:
I believe it is quite a basic example also, especially into todays world of migration to HTTPS by default.

ostera:
Ah! certificates! Very much like this here, but I would be mindful of further samples.

igorshubovych:
@rubenvereecken do you still think it should not be merged?

igorshubovych:
@rubenvereecken ping

imjacobclark:
@rubenvereecken @igorshubovych ping

igorshubovych:
I think we should merge it. It is quite common and important example. We broke the rules for `ls`, we should break it now.

@rubenvereecken, if you think this (or other example from that page) should be deleted, please create new PR.

@imjacobclark 
Thank you for contribution and taking an initiative to complete it!
bike-velo-velib-3osxYp4vALaiUR2252

PR #737, opened 2016-01-28
airport: added sniff and status info commands

oxguy3:
Just noticed the two commands I find most useful in `airport` are both excluded from `airport`'s tldr page! Here's a quick revision adding these commands.

### DIFF
--- a/pages/osx/airport.md
+++ b/pages/osx/airport.md
@@ -6,6 +6,14 @@
 
 `sudo ln -s /System/Library/PrivateFrameworks/Apple80211.framework/Versions/Current/Resources/airport /usr/sbin/airport`
 
+- Show current wireless status information:
+
+`airport -I`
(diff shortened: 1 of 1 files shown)


waldyrious:
Nice, thanks! How about also changing the command description? The current one is redundant to the point of being nearly information-free.

oxguy3:
Haha wow, didn't even notice that silly description. Wrote a better one now.

igorshubovych:
LGTM. Although I would remove the 1st example, not added by you.

rubenvereecken:
Thanks for ridding us of that weird symlink example as well.

oxguy3:
It's very useful to run that command, since OS X doesn't put `airport` in your $PATH for some reason. But yeah, doesn't quite fit in on a tldr page.

PR #739, opened 2016-01-28
mitmdump: add page

oxguy3:
[mitmdump](http://docs.mitmproxy.org/en/stable/mitmdump.html) is a tool for monitoring, recording, and transforming HTTP traffic on your local machine.

### DIFF
--- a/pages/common/mitmdump.md
+++ b/pages/common/mitmdump.md
@@ -1,0 +1,16 @@
+# mitmdump
+
+> View, record, and programmatically transform HTTP traffic.
+> The command-line counterpart to mitmproxy.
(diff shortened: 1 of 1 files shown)


rubenvereecken:
Hehe never thought to see that tool here. I met the author at a hackathon once, we used it for our project. 

I'm happy with this one. So, where's mitmproxy :D.

oxguy3:
Oh neat!

Didn't think to add mitmproxy because usually you just run it `mitmproxy`, though I guess it does have quite the cavalcade of optional arguments now that i'm looking at the `--help` list.

PR #740, opened 2016-01-28
xz: add page

oxguy3:
[xz](http://tukaani.org/xz/) is a compression tool.

### DIFF
--- a/pages/common/xz.md
+++ b/pages/common/xz.md
@@ -18,10 +18,10 @@
 
 `xz -k {{file}}`
 
-- Compress a file using fastest compression:
+- Compress a file using the fastest compression:
 
 `xz -0 {{file}}`
 
-- Compress a file using best compression:
+- Compress a file using the best compression:
 
 `xz -9 {{file}}`
--- a/pages/common/xz.md
+++ b/pages/common/xz.md
@@ -18,6 +18,10 @@
 
 `xz -k {{file}}`
 
-- Compress a file and attempt to improve compression ratio by using more CPU time:
+- Compress a file using fastest compression:
 
-`xz -e {{file}}`
+`xz -0 {{file}}`
+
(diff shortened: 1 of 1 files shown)


pages/common/xz.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Decompress a file and write to stdout:
+
+`xz -dc {{file.xz}}`
+
+- Compress a file, but don't delete the original:
+
+`xz -k {{file}}`
+
+- Compress a file and attempt to improve compression ratio by using more CPU time:
```
rubenvereecken: You have a lot of different ways of saying "better but slower" among your many, many compression pages. Might want to stick to one ;).

oxguy3: I tend to sort of copy the wording the original man page uses. :stuck_out_tongue: Actually, I'm realizing that the `-e` flag doesn't make sense without a `-0`...`-9` flag as well. I'm gonna drop this example and use those instead (I'm typically too lazy to use anything but default compression ratio, but I imagine anyone interested enough in compression to not just use the more-popular gzip might care about compression ratios).

pages/common/xz.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Decompress a file and write to stdout:
+
+`xz -dc {{file.xz}}`
+
+- Compress a file, but don't delete the original:
+
+`xz -k {{file}}`
+
+- Compress a file using fastest compression:
```
igorshubovych: *the* fastest

pages/common/xz.md
```diff
@@ -0,0 +1,27 @@
...
+
+`xz -0 {{file}}`
+
+- Compress a file using best compression:
```
igorshubovych: *the* best

igorshubovych:
LGTM. Thanks!

PR #742, opened 2016-01-28
optipng: add page

oxguy3:
[OptiPNG](http://optipng.sourceforge.net/) is a popular PNG optimization tool.

I'm a bit curious what you guys think of the parenthetical notes for examples 2 and 3. They perhaps make the commands easier to understand, but they're also a bit clunky and I'm not entirely sure they're necessary. My hope is that someone is able to read those two examples, and easily deduce that there are also flags `-o1`, `-o2`, ... `-o6` for optimization in between the extremes of `-o0` and `-o7`.

### DIFF
--- a/pages/common/optipng.md
+++ b/pages/common/optipng.md
@@ -1,0 +1,23 @@
+# optipng
+
+> PNG image file optimization utility.
+
+- Compress a PNG with default settings:
+
+`optipng {{file.png}}`
+
+- Compress a PNG as optimally (but slowly) as possible:
+
+`optipng -o{{7}} {{file.png}}`
+
+- Compress a PNG as quickly (but non-optimally) as possible:
+
+`optipng -o{{0}} {{file.png}}`
+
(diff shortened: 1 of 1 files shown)


rubenvereecken:
I recently merged `pigz` and fiddled a bit with it to my own liking:
```
  - Compress a file using the best compression method:
    pigz -9 filename

  - Compress a file using no compression and 4 processors:
    pigz -0 -p4 filename
```

I didn't put braces around the 0 and 9, but you could (or should). I don't require the parantheses describing how slower gets you better results, but I don't mind, either.

oxguy3:
I like your "no compression" description -- that's pretty much what `-o0` does as well, so I'll change it to that. And if neither of us thinks the parentheses are necessary then I'm gonna drop 'em.

rubenvereecken:
:+1: great :D

igorshubovych:
LGTM! Thanks.

PR #743, opened 2016-01-28
mitmproxy: add page

oxguy3:
To match mitmdump in #739, here's [mitmproxy](https://mitmproxy.org/)! mitmproxy is "an interactive console program that allows traffic flows to be intercepted, inspected, modified and replayed".

### DIFF
--- a/pages/common/mitmproxy.md
+++ b/pages/common/mitmproxy.md
@@ -6,6 +6,6 @@
 
 `mitmproxy`
 
-- Start mitmproxy binded to custom address and port:
+- Start mitmproxy bound to custom address and port:
 
 `mitmproxy -b {{ip_address}} -p {{port}}`
--- a/pages/common/mitmproxy.md
+++ b/pages/common/mitmproxy.md
@@ -1,0 +1,11 @@
+# mitmproxy
+
+> An interactive man-in-the-middle HTTP proxy.
+
+- Start mitmproxy with default settings:
+
+`mitmproxy`
+
+- Start mitmproxy binded to custom address and port:
+
+`mitmproxy -b {{ip_address}} -p {{port}}`


pages/common/mitmproxy.md
```diff
@@ -0,0 +1,11 @@
+# mitmproxy
+
+> An interactive man-in-the-middle HTTP proxy.
```
rubenvereecken: It's a good description but I think proxies are always man-in-the-middle. That's being annoying because you're literally just spelling out the name. I think they threw in the MITM part because it's used especially for MITM attacks? Or, well, mitm analysis.

Please correct me if I'm wrong.

pages/common/mitmproxy.md
```diff
@@ -0,0 +1,11 @@
...
+
+- Start mitmproxy with default settings:
+
+`mitmproxy`
+
+- Start mitmproxy binded to custom address and port:
```
rubenvereecken: I'm not a native English speaker but is "binded" correct?

rubenvereecken:
I like your style. You've got a nice rhythm going!

pages/common/mitmproxy.md
oxguy3: Errrrr, I guess not :/ Whoops!

pages/common/mitmproxy.md
oxguy3: I think you're right, but I think people tend not to think of a straightforward proxy as a "man-in-the-middle" because usually when you say "man-in-the-middle", you mean "man-in-the-middle attack". A simple proxy that just passes requests through blindly? Not really a MITM attack. But a proxy that strips out the SSL/TLS on HTTPS requests? That sounds more like an attack. But I dunno, that's just like, my opinion, man.

In any event, I think it's probably a good idea if our description spells out "man-in-the-middle" so people can figure out what the name means.

oxguy3:
Hey, thanks! This is actually the first page I wrote today -- all my other PRs are just old ones left over from that massive PR that I hadn't gotten around to making individual PRs for until now. :stuck_out_tongue:

pages/common/mitmproxy.md
rubenvereecken: That was a good explanation. I wasn't all too sure but you confirmed that mitm really only is used in context of a mitm attack. Viewed in that light, it really makes sense to have it spelled out like you have it here.

rubenvereecken:
:tada: 

@mhils look at that beauty.

mhils:
:+1: :smiley:

rubenvereecken:
There, now you've got author approval as well

oxguy3:
Oh neat, hi there! :smile:

PR #744, opened 2016-01-28
tldr: added `--update` example

oxguy3:
The `--update` flag in the tldr command is super useful, and I was really surprised it wasn't in the tldr page when I was trying to remember how to update the other day. Thus, I've added it here!

### DIFF
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -5,3 +5,7 @@
 - Get typical usages of a command (hint: this is how you got here!):
 
 `tldr {{command}}`
+
+- Update the local cache of tldr pages:
+
+`tldr --update`


igorshubovych:
Well, it turned to be there are tons of [tldr CLI clients](https://github.com/tldr-pages/tldr/wiki/TLDR-clients#console-clients)

And not all of them have this flag :(

cc @waldyrious, @rubenvereecken

oxguy3:
Mannn, why'd they all have to call themselves `tldr`? :stuck_out_tongue:

rubenvereecken:
This is rather too bad but we never standardized clients, only this repository and how to interpret its contents. I don't even know which ones can be considered official and I've never used anything else except the Node.js client (official if any is).

I wouldn't add it.

oxguy3:
I think the NodeJS one could be considered "official", as that's the one that tldr-pages' website tells you to install: <https://tldr-pages.github.io/>.

rubenvereecken:
The reason I'm not deciding here is because I abstained from the whole discussion on clients and which to consider official. I'd love to just consider that one official and have that flag in otherwise.

waldyrious:
This is clearly another reason for us to decide on a standard set of features we'd expect tldr (command line) clients to provide. @igorshubovych already started a page at https://github.com/tldr-pages/tldr/wiki/Minimum-specifications-for-TL;DR-command-line-clients, but it needs some love. Shall we open an issue to track that discussion?

ostera:
This actually is a good way of enforcing a spec for a client, not only because it outlines the features we require, but also because it's in itself a tldr page.

waldyrious:
Great point, @ostera -- the tldr page for tldr should indeed work as the minimum spec for tldr clients! So looking at it from this point of view, what other options would we want to be universal to all clients?

igorshubovych:
@ostera @waldyrious agree with you about `--update`, it seems it makes sense.

However, `--help` becomes standard de-facto. Majority of commands have it, so I do not think it makes sense to list it separately. Additionally, NodeJS client has the same output for `$ tldr --help` and for just `$ tldr`.

waldyrious:
Well, if we start reaching the maximum limit of examples (#686), we can remove --help, but for now it would be a good way to enforce it for all clients. It seems to fit better here (and in a more explicit way) than in the wiki page, which currently says "it should list supported features".

The fact that just `tldr` does the same is either a bonus for the node client, or something we would want for all clients (and thus we should add here too).

igorshubovych:
Thank you!

PR #745, opened 2016-01-28
say: lots of cleanup

oxguy3:
The tldr page for `say` seemed like it needed some revision. I corrected the audio file command (it creates an AIFF file, not an M4A file). I added an example of how to use a custom voice and speech rate (it told you how to list the voices, but it didn't tell you how to actually use them!). Then there are just a couple of other formatting things and clarifications I made.

### DIFF
--- a/pages/osx/say.md
+++ b/pages/osx/say.md
@@ -1,19 +1,23 @@
 # say
 
-> Uses text-to-speech to speak through the default sound device.
+> Converts text to speech.
 
 - Speak a phrase aloud:
 
-`say "I like to ride my bike."`
+`say {{"I like to ride my bike."}}`
 
 - Speak a file aloud:
 
 `say -f {{filename}}`
 
-- Create an AAC compressed audio file with the spoken text:
+- Speak a phrase with a custom voice and speech rate:
 
-`say -o {{filename.m4a}} "Everyone loves iTunes"`
+`say -v {{voice}} -r {{words_per_minute}} {{"I'm sorry Dave, I can't let you do that."}}`
 
 - List the available voices:
 
-`say -v '?'`
+`say -v ?`
+
+- Create an audio file of the spoken text:
+
+`say -o {{filename.aiff}} {{"Everyone loves iTunes."}}`
--- a/pages/osx/say.md
+++ b/pages/osx/say.md
@@ -8,7 +8,7 @@
 
 - Speak a file aloud:
 
-`say -f {{filename}}`
+`say -f {{filename.txt}}`
 
 - Speak a phrase with a custom voice and speech rate:
 


pages/osx/say.md
```diff
@@ -1,19 +1,23 @@
...
 - Speak a file aloud:
 
-`say -f {{filename}}`
+`say -f {{filename.txt}}`
 
-- Create an AAC compressed audio file with the spoken text:
+- Speak a phrase with a custom voice and speech rate:
 
-`say -o {{filename.m4a}} "Everyone loves iTunes"`
+`say -v {{voice}} -r {{words_per_minute}} {{"I'm sorry Dave, I can't let you do that."}}`
```
rubenvereecken: Hahaha

pages/osx/say.md
```diff
@@ -1,19 +1,23 @@
...
 
 - List the available voices:
 
-`say -v '?'`
+`say -v ?`
+
+- Create an audio file of the spoken text:
+
+`say -o {{filename.aiff}} {{"Everyone loves iTunes."}}`
```
rubenvereecken: Hmm.

pages/osx/say.md
```diff
@@ -1,19 +1,23 @@
...
-`say "I like to ride my bike."`
+`say {{"I like to ride my bike."}}`
 
 - Speak a file aloud:
 
-`say -f {{filename}}`
+`say -f {{filename.txt}}`
 
-- Create an AAC compressed audio file with the spoken text:
+- Speak a phrase with a custom voice and speech rate:
```
rubenvereecken: I know you didn't introduce this but I think the correct phrase is "to say a phrase"?

pages/osx/say.md
waldyrious: Brilliant!

pages/osx/say.md
waldyrious: :+1: @rubenvereecken

pages/osx/say.md
waldyrious: Not a fan either.

pages/osx/say.md
igorshubovych: When we use tokens {{}}, we put there named parameter.
So it should be one of those:
```md
`say "I am sorry, Dave..."`

`say {{message}}`
```

rubenvereecken: Oh, we do? Never got the hang of this myself. It's incorrect in a few then.

waldyrious: That's the default, but in some cases IIRC it's been decided that an actual example is clearer. It boils down to what's best for the user, rather than following a guideline strictly. In this case I sadly agree that {{message}} is the most appropriate choice.

pages/osx/say.md
oxguy3: Yeah I thought that one was weird -- that one is from the existing version of this page (that's definitely not something I'd write).

pages/osx/say.md
oxguy3: I was about to change everything to "say", but it sounds really awkward on some of them ("Say a file aloud" is awkward and "Say the contents of a file aloud" is wordy and still a bit awkward).

To my ear, it tends to sound wrong if the object of the verb "say" is anything but a literal quotation; i.e. "say 'hello'" is fine, but "say a greeting" sounds really weird, "speak a phrase" sounds okay, but "say a phrase" is a bit uneasy. There are exceptions ("say nice things about me" and "say what you need to say" both sound less weird), but for the most part, I think a verb like "speak" or "utter" sounds more natural.

Another approach would be "Read a message aloud", "Read a message aloud with a custom voice and speech rate", etc. That wording sounds very natural and comfortable to me (though maybe others might not realize I mean "read" like "read a book" and not "read a file into memory"). What do you guys think?

pages/osx/say.md
ostera: I think that leaving a funny message like this doesn't hurt the clarity of the command, and will put a smile on some :) – that's never a bad thing. It has my :+1: to stay

ostera:
Hey @oxguy3, before this got stale, I merged it in and did some changes that were apparently of consensus ("say" vs "speak") here: https://github.com/tldr-pages/tldr/commit/6f4901d836f9e67ac6463b89574e4962699909ff

Thanks a ton for taking the time to contribute! :)

waldyrious:
Doing god's work @ostera :+1:

PR #747, opened 2016-01-29
Add common and useful netstat use case

eidge:
This is one of the most useful netstat options I've found.

I thought it was worth sharing.


ostera:
Looking good! Thanks for the contribution :)

PR #748, opened 2016-01-29
arp: Generalize to common and add cache clearing.

nevsan:
Merging my previous arp PR with the initial one in #479

### DIFF
--- a/pages/common/arp.md
+++ b/pages/common/arp.md
@@ -4,9 +4,13 @@
 
 - Show current arp table:
 
-`arp`
+`arp -a`
 
-- Delete an entry:
+- Clear the entire cache:
+
+`sudo arp -a -d`
+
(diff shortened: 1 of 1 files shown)


rubenvereecken:
Thanks for merging the efforts!

pages/common/arp.md
```diff
@@ -2,11 +2,15 @@
 
 > Show and manipulate your system's ARP cache.
 
-- Show current arp table:
+- Show current arp table [as numeric address]:
```
rubenvereecken: We don't do "optionals" or parameter describing here even though it's an easy way to introduce two options at the same time. Drop the brackets, with the added danger that the options below aren't clear individually.

ostera:
@nevsan thanks for the time!

PR #752, opened 2016-01-30
Update aptitude.md: purge command + better desc

85pando:
* `aptitude install {{package}}` also installs dependencies.
* `aptitude remove {{package}}` also removes packages depending on `{{package}}`
* `aptitude purge {{package}}` removes packages and also removes config files.

If someone thinks the "and also dependencies" part is too much, I'm totally ok, but I thing purge should at least be added.

### DIFF
--- a/pages/linux/aptitude.md
+++ b/pages/linux/aptitude.md
@@ -6,7 +6,7 @@
 
 `aptitude update`
 
-- Install a new package:
+- Install a new package and its dependencies:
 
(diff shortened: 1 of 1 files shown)


waldyrious:
Nice one! :+1: Please fix the `.:` though.

85pando:
And there I was, thinking I'd get it right on the first try this time!

igorshubovych:
LGTM. Thanks!

PR #753, opened 2016-01-30
npm: removed some examples and added others

igorshubovych:
- 2 lines description, mentioned that it is not only Node.js package manager, but also JavaScript package manager (see official docs)
- Moved `npm init` to the bottom, it is the rarest command in the list. Change its description to more obvious one.
- Removed `npm prune`, this command is for hackers, not for regular users
- Removed `npm version`, it is useless
- Removed `npm publish`. It is for those who have npm modules published. Well in fact, even them do not do it every day.
- Added `npm uninstall` example
- Added `npm list` example

### DIFF
--- a/pages/common/npm.md
+++ b/pages/common/npm.md
@@ -1,10 +1,7 @@
 # npm
 
-> Node package manager, to manage Node.js projects and install module dependencies.
-
-- Create a new project in the current folder:
-
-`npm init`
+> JavaScript and Node.js package manager.
+> Manage Node.js projects and their module dependencies.
 
 - Download all dependencies referenced in package.json:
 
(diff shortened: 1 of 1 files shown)


pages/common/npm.md
```diff
@@ -1,10 +1,7 @@
...
 
-> Node package manager, to manage Node.js projects and install module dependencies.
-
-- Create a new project in the current folder:
-
-`npm init`
+> JavaScript and Node.js package manager.
+> Manage Node.js projects and their module dependencies.
 
 - Download all dependencies referenced in package.json:
```
waldyrious: I don't think this should be the first example. npm install. It could make sense to come after npm init, given the latter's description.

igorshubovych: In reality you rarely do `npm init`. But you often do `npm install`. Honestly if we decide to remove any example, then `npm init` should be 1st candidate. All it does, it just helps you creating package.json. In many cases you just do it yourself, or clone repo, or sometimes even don't use package.json at all, as it was on tldr some.

In contrast, `npm install` is the most common, the most popular command. That's why `npm` is created - to do `npm install`.

waldyrious: I feel like many more people use npm just as package consumers than as package creators, hence my suggestion to have install [-g] first. But I'll defer to you guys who are more familiar with the tool to make the decisions :)

igorshubovych: That's fair. I will update it.

ostera:
Looking good :) Thanks @igorshubovych

PR #755, opened 2016-01-31
sftp: add navigation and transfer examples

slash3b:
Just added a few commands to sftp page.

### DIFF
--- a/pages/common/sftp.md
+++ b/pages/common/sftp.md
@@ -1,7 +1,8 @@
 # sftp
 
-> Secure file transfer program.
-> Copy files between hosts using Secure File Transfer Protocol over SSH.
+> Secure File Transfer Program.
+> Interactive program to copy files between hosts over SSH.
+> For non-interactive file transfers, see `scp` or `rsync`.
 
 - Connect to a remote server and enter an interactive command mode:
 
@@ -9,12 +10,20 @@
 
 - Connect using an alternate port:
 
-`sftp -P {{remote_port}} {{remote_host}}`
+`sftp -P {{remote_port}} {{remote_user}}@{{remote_host}}`
 
-- Copy a local file to a remote host:
+- Transfer remote file to the local system:
 
-`sftp {{local_file}} {{remote_host}}:{{/path/remote_file}}`
+`get {{/path/remote_file}}`
 
-- Copy a remote file to the local host:
+- Transfer local file to the remote system:
 
-`sftp {{remote_host}}:{{/path/remote_file}} {{local_file}}`
+`put {{/path/local_file}}`
+
+- Get list of files on local machine:
+
+`lls`
+
+- Get list of files on remote machine:
+
+`ls`


pages/common/sftp.md
```diff
@@ -11,6 +11,10 @@
 
 `sftp -P {{remote_port}} {{remote_host}}`
 
+- Navigate through the LOCAL system file hierarchy:
+
+`lpwd, lls, lcd`
```
85pando: While I am totally for including those three commands, it should probably rather be split in three separate commands with an explanation each

85pando:
I know this was in here before you edited it, but I just want to bring this up.

The commands "Copy local file to remote" and "Copy remote file to local" don't work for me and I have never seen this with `sftp` before (and I don't find it in the manpages). This seems to me like `scp`-syntax. Seing as this was added via a "Rework scp" PR it might come from here.
I think they should both be removed. This would also make more space for `cd`, `ls`, `cd` which are also missing from the page (maybe also `du [-h]`).

pages/common/sftp.md
slash3b: I think these commands are pretty easy to catch up with. User who intended to use `sftp` will guess that `lpwd` stands for _local present working directory_.  I'm just trying to keep thing concise. Please let me know if I have to split them.

slash3b:
@85pando thanks. I've checked `get` and `put` commands on Ubuntu 14.04 and Fedora 23, please take a look at this screenshot ([sftp example  - connecting from Ubuntu 14.04 to Fedora 23 ](http://i.imgur.com/pB5JapK.png) ). On the left side there is man page for `sftp` and on the right side I tested the commands.

85pando:
I think you mistook me. I'm talking about those two:

* Copy a local file to a remote host:

`sftp {{local_file}} {{remote_host}}:{{/path/remote_file}}`

* Copy a remote file to the local host:

`sftp {{remote_host}}:{{/path/remote_file}} {{local_file}}`

slash3b:
@85pando oh yes, I did misunderstand you, sorry  : )
Could you please review the changes I made ?

85pando:
Looking good. You might want to also add a "navigate through remote file system".

Still not sure about multiple commands in one line though, maybe @waldyrious @rubenvereecken @igorshubovych want to say something to this?

waldyrious:
I am not sure I like the comma-separated command list (@rprieto), but I'll just comment on the phrasing: The command descriptions, for consistency, ought to be phrased as declarative statements, or instructions, e.g "do this". In this case, that would mean: "Temporary drop to local shell" --> "Temporarily drop to local shell" and "After dropping to local shell get back to `sftp` with:" --> "Get back to `sftp` after dropping to local shell". Does that make sense?

igorshubovych:
I do not like comma in commands. There should be one command per description according to guidelines.

Also the page is already too long.

slash3b:
Thank you for the comments. 
I removed "Temporarily drop to local shell" command for the list to be shorter, I left only `lls` and `ls`
 commands.

waldyrious:
@slash3b shoulddn't there be at least one non-interactive version? The simplest use cases should be listed -- note that tldr doesn't aim to be a complete reference for the commands it covers, just provide a well-balanced intro/refresher.

85pando:
`sftp` is totally interactive. For non-interactive file copying to remote machines there is `scp` (and later `rsync`).

waldyrious:
How about editing the main description to reflect that? Currently it seems to indicate quite the contrary :)

We could include a suggestion to use scp for single files or rsync for single directories (which I believe is a reasonable recommendation, but happy to stand corrected).

waldyrious:
Sorry, clicked the wrong button

slash3b:
@85pando, thank you.
@waldyrious, how about this description:

> Secure and interactive file transfer program. Performs all operations over an encrypted ssh transport.

waldyrious:
@slash3b I'd include the expansion of "SFTP" intact in the description, and start with "interactive" to make that more evident. How about something along these lines?

> Interactive program to copy files between hosts using the Secure File Transfer Protocol over SSH.
> For non-interactive file transfers, see `scp` or `rsync`.

slash3b:
Is it okay now ? I edited slightly description to make it shorter.

rprieto:
:+1: on making it clear it's interactive. However unless I'm missing something obvious, I wouldn't mention `scp` or `rsync` since they won't help you get/put files on an FTP server.

85pando:
`sftp` and `scp` both only need an `sshd` on the server side. `rsync` needs to be installed on both sides of the communication. So `scp` actually has the same dependencies as `sftp`.

rubenvereecken:
@rprieto maybe some confusion arose because of the name? There's `sftp` for file transfer (not ftp) over ssh and `ftps` (sometimes also called `sftp`) for FTP over SSL.

waldyrious:
Let's just mention scp then? Also, are we ok with a 3-line description?

rprieto:
> sftp and scp both only need an sshd on the server side

I didn't know that, thanks for explaining! Sounds good to me then :+1:

ostera:
Many :+1:s! Seems like it's merging time :) before it gets stale. Thanks @slash3b and @85pando for your contributions!

slash3b:
Thank you guys! 
Couldn't reply in time, my home machine is broken.
Will add some new command for sure.

PR #756, opened 2016-02-03
Add route.md

nelsonchung:
I want to add route.md file for provide the information about route command.

### DIFF
--- a/pages/common/route.md
+++ b/pages/common/route.md
@@ -1,0 +1,17 @@
+# route 
+
+> Use route cmd to set the route table .
+
+- Display the information of route table:
+
+`route -n`
+
+- Add route rule:
+
+`sudo route add -net {{ip_address}} netmask {{netmask_address}} gw {{gw_address}}`
+
+- Delete route rule:
+
+`sudo route del -net {{ip_address}} netmask {{netmask_address}} dev {{gw_address}}`
+
+


ostera:
@nelsonchung looks good, except for 2 things: `pages/common/route.md:16: TLDR008 File should contain no trailing whitespace`, and the `sudo`.

If you can get the `sudo` out and fix that linting error (and any other that may arise), I'll make sure to get this merged :)

Thanks for the time to contribute!

waldyrious:
@ostera if the sudo is required, it should at least be mentioned in the description. Removing it altogether would be a disservice if it's needed for the usual user permissions setup.

ostera:
@waldyrious makes sense, but it isn't part of the command – It's part of the permission settings for the user running it, so I agree that it should be mentioned but still think it shouldn't be part of the command.

waldyrious:
Sure, sudo isn't part of any commands except itself, but I think it's reasonable that if a command requires root privileges *by default*, it is only a disservice to the user to make it less convenient to use the command with sudo (e.g. by copy-pasting from the tldr output).

In fact, since we already allow usage of other commands when justified (ls, cat, etc), I see it as a reasonable extension of that reasoning to also allow sudo in that list, don't you?

ostera:
Fair enough! I'd rather have this merged than debate whether `sudo` is a vital part of it.

ostera:
@nelsonchung I merged it in and fixed the trailing whitespace for you, thanks a lot for the contribution! :)

PR #757, opened 2016-02-04
awk: add /search/ example

solarsailer:
Searching for values in lines is a common operation with awk. I didn't specify that "something" is a regexp because it's not really useful at first.

### DIFF
--- a/pages/common/awk.md
+++ b/pages/common/awk.md
@@ -6,6 +6,10 @@
 
 `awk '{print $5}' {{filename}}`
 
+- Print the second column of the lines containing "something" in a space separated file:
+
+`awk '/{{something}}/ {print $2}' {{filename}}`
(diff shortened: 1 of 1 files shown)


igorshubovych:
LGTM

waldyrious:
I'd at least reorder the description sentence to match the order that the parts of the command appear. That helps in identifying what does what, especially since the // aren't mentioned explicitly.

solarsailer:
I wanted to keep the same structure as the other examples. And I do refer to the value inside `//`, with "something" => "/{{something}}/". ;)

waldyrious:
In the end, it's a judgment call. I don't think your choice is bad, I just think mine is more intuitive to beginners. That said, I won't oppose merging this as-is.

PR #758, opened 2016-02-04
Switch from `` code blocks to 4 spaces

jedahan:
This makes it a lot easier to copy paste with the mouse in the commandline.

I am testing now with a few implementations to make sure it works fine.

### DIFF
--- a/pages/common/7za.md
+++ b/pages/common/7za.md
@@ -4,16 +4,16 @@
 
 - Compress directory or file:
 
-`7za a {{compressed.7z}} {{directory_or_file_to_compress}}`
+    7za a {{compressed.7z}} {{directory_or_file_to_compress}}
 
 - Decompress an existing 7z file with original directory structure:
 
-`7za x {{compressed.7z}}`
+    7za x {{compressed.7z}}
 
 - Compress to zip format:
 
-`7za a -tzip {{compressed.zip}} {{directory_or_file_to_compress}}`
+    7za a -tzip {{compressed.zip}} {{directory_or_file_to_compress}}
 
 - Create multipart 7zip file; `part_size` specifies part size in Bytes, Kilobytes, Megabytes or Gigabytes:
 
-`7za -v{{part_size}}{{[b|k|m|g]}} {{compressed.7z}} {{directory_or_file_to_compress}}`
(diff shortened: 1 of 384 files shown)


igorshubovych:
Oh, man.
You did so much work without asking ahead. Please tell me that you did some script for it and you did not do it manually.

In cases like this it is better to discuss it in tickets first, before doing any job. This will be most likely rejected, because linter and the most of the clients depends on the backticks when parsing. Usually clients parse the format and present in the nice way.

E.g. NodeJS client present it this way
1 igor igors-macbook-air projects tldr zsh 2016-02-04 21-53-27

It is quite easy to do copy-pasting.

jedahan:
It was a one line script with sed, so not a lot of work. This was kinda the introduction to discuss the change. I would have to change https://github.com/tldr-pages/tldr-lint/blob/master/lib/tldr-parser.js#L736 and https://github.com/tldr-pages/tldr-lint/blob/master/lib/tldr-parser.js#L758 to allow for 8 spaces, etc.

When using a client, they are formatted in an easy way to copy paste, but if you open the markdown files themselves it is a bit tougher. Of course most people should be using a client, but I thought it looks nicer with 8 spaces (and matches the output of most tools that parse it).

Would people be vehemently against changing the tldr-linter to look for 8 spaces instead of backticks?

jedahan:
nevermind, saw a bunch of implementations, they rely on '`'

waldyrious:
I would still like this to be discussed. See https://github.com/tldr-pages/tldr/issues/686#issuecomment-173939607

rubenvereecken:
Poor man actually checked out the _generated_ parser file.

jedahan:
Well, never worked with any lexer or parser, so didnt even look at the .yy or .l files. Still don't know if these are the right places to change things, but:

https://github.com/tldr-pages/tldr-lint/blob/master/tldr.yy#L53-L55 and https://github.com/tldr-pages/tldr-lint/blob/master/tldr.l#L138-L141 probably relevant.

Somewhat related, maybe generated files shouldn't be checked into the source repo? But I don't see how the generated files are made. Maybe a makefile or just that jison thing?

rubenvereecken:
I had that problem as well, but I needed the generated files to be present upon `npm install`. A work around would be to use jison upon tldr-lint startup to generate the needed source code, but that's not a very pretty way to go about it.

Anyway, we'll continue the conversation

PR #759, opened 2016-02-04
fzf: add page

loliee:
# ❯ fzf

An awesome command-line fuzzy finder, here is the [homepage of FZF](https://github.com/junegunn/fzf).

Can fuzzy find quite everything, the [wiki](https://github.com/junegunn/fzf/wiki/examples) is a great place to discover this. 
It's also a great alternative to `ctrl-p` or `command-t` for vim users.

### DIFF
--- a/pages/common/fzf.md
+++ b/pages/common/fzf.md
@@ -1,0 +1,27 @@
+# fzf
+
+> Command line fuzzy finder.
+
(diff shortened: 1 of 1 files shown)


ostera:
I use this on a daily basis, part of my flow within vim as well. Looks very good!

PR #760, opened 2016-02-04
Change 'ls' example for xargs page to 'find -name'

TiredSounds:
The use of `ls` for piping is not recommended because of it's tendency to format into columns. It is more user friendly than machine friendly.

See http://stackoverflow.com/a/938052

### DIFF
--- a/pages/common/xargs.md
+++ b/pages/common/xargs.md
@@ -12,4 +12,4 @@
 
 - Delete all files that start with 'M':
 
-`ls M* | xargs rm`
+`find -name 'M*' | xargs rm`
--- a/pages/common/xargs.md
+++ b/pages/common/xargs.md
@@ -12,4 +12,4 @@
 
 - Delete all files that start with 'M':
 
-`find -name 'M*' | xargs rm`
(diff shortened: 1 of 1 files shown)


igorshubovych:
Maybe just `ls -1 M*`?

TiredSounds:
There are other reasons to not use `ls` in general for anything other than user interaction, which are detailed in that SO answer I linked.

But if you'd still prefer to use `ls -l` I'll make another commit

waldyrious:
The find example is simple and intuitive enough to make sense here, IMO.

ostera:
I agree with @waldyrious, besides `find` will yield more appropriate results given the text says. Other than that, it needs a path, so it should be `find . -name "M*"`

TiredSounds:
Find assumes `.` if no path is given, though I suppose for demonstration that would make sense.

ostera:
I pointed that out because it didn't work on OS X, maybe I'm not using the GNU version of it?

TiredSounds:
Ah I see, I haven't used OS X in a long time but yeah, many of the unix tools have subtle differences on there.

PR #761, opened 2016-02-05
phpize: add page

oxguy3:
phpize is a little script you have to run in the process of compiling a PHP extension from source. Not a very frequently used or complex tool, but hey, it was one I knew how to use that wasn't already in the tldr pages.


igorshubovych:
LGTM

felixonmars:
:+1: Thanks :)

PR #764, opened 2016-02-05
base64: move page to common (and add base32)

oxguy3:
For some reason, the GNU coreutils utility "base64" was tucked away in the linux folder; however, you can get GNU coreutils on any OS (I have it on OS X via `brew install coreutils`). I moved it to the correct directory.

While I was at it, I also made a page for the [base32](https://www.gnu.org/software/coreutils/manual/html_node/base32-invocation.html#base32-invocation) tool. The base32 tool works 100% exactly the same as the base64 tool, so I just copied that page and changed all instances of "32" to "64".


oxguy3:
Ah, hmm, figured out why base64 was in the linux folder. I guess OS X ships a different version of base64 than the coreutils version. For someone like me, who replaces all the OS X garbage with GNU stuff, the fact that typing `tldr base64` gives me the OS X page is bound to cause serious confusion. Perhaps there should be some other way of sorting these tools?

rubenvereecken:
I know the Node.js client has a way of requesting a specific platform, it also goes in its config... That'd be what you want right?

oxguy3:
Well, I don't want to change my OS, because I am running OS X and using OS X tools often. I just want it so that, if a common page and an osx page exist, I have the ability to specify which version my client should retrieve. I'd imagine the default would be to retrieve the OS X version, but then my preference would be to retrieve the common version.

This would allow the GNU coreutils files (which are 100% cross-platform) to be placed in the common directory where they belong.

rubenvereecken:
That's a fair though maybe niche request, could you file it with a specific client?

igorshubovych:
`base64` is available on OSX and Linux.
`base32` is available only on Linux.

oxguy3:
They're both cross-platform via GNU coreutils, which are available on any platform (for OS X, just do `brew install coreutils`). It's just that OS X also happens to ship with its own, incompatible version of base64. The GNU coreutils versions should probably be in /common, and then the OS X special alternative version of base64 should probably be in /osx

igorshubovych:
> It's just that OS X also happens to ship with its own, incompatible version of base64.
If it ships with the system, then 99% will use this version. And we have to describe that version on a page.

`coreutils` by default install all utilities prefixed with `g`. You need to make a hack to replace standard OSX utility. So this can be the solution, to the problem you can have pages `osx\gbase64.md` and `osx\gbase32.md`.

ostera:
I think it wouldn't hurt to have them in common since they can be made available on all systems.

igorshubovych:
@ostera 99% of people will use other versions of these tools, and they will wonder why `tldr` doc is not valid.

ostera:
But that's why we have the `osx` and other os's versions of them, whereas the `common` ones are the ones that are available on all systems.

Hmm...so if I don't have `coreutils` brewed into my OS X, I will use `osx/base64`, and if I do (and only if I brewed it to replace the defaults) then the it will still give me the `osx/base64` page will might not be the same. I guess it'd be up to the client to determine if `coreutils` is installed and return one page or the other.

That way the organization of the pages is still sound, and the actual logic for which page to show is delegated to the client, who has access to the underlying platform running it.

What do you guys think? @oxguy3  @rubenvereecken  @igorshubovych  @waldyrious

oxguy3:
That's exactly what I was thinking @ostera -- hit the nail on the head! :smile:

Issue #736, opened 2016-01-28
README.md is too long.


kindraywind:
Could you add the TL;DR section at the top that only have a short description, one screenshot and the quickest way to install (e.g., `npm install -g tldr`)

:D

igorshubovych:
cc @waldyrious :)

waldyrious:
@kindraywind has a good point. It's not that the Readme is too long, but poorly organized. In honor of the spirit of tldr, we should definitely make the changes suggested here. I'll take a stab at this soon (hopefully this weekend).

ostera:
I think we can close this now since the #765 was merged

PR #765, opened 2016-02-06
Add tldr for installing or trying right on top.

ostera:
Also moves what is TL;DR to the bottom, since it's more of a miscellanea.

Refers to #736


igorshubovych:
I like it, but let's hear other people's opinions. I am curious what @waldyrious thinks.

waldyrious:
Even though I believe we can improve the client list further (e.g. by clearly indicating the authors/maintainers, etc.), that is orthogonal to this particular PR. I agree with @ostera's changes :+1:.

PR #766, opened 2016-02-10
Added the dirs command

sbrl:
Another simple tldr page.

### DIFF
--- a/pages/common/dirs.md
+++ b/pages/common/dirs.md
@@ -1,0 +1,20 @@
+# dirs
+
+> Displays or manipulates the directory stack.
+> The directory stack is a list of recently visited directories that can be manipulated with the `pushd` and `popd` commands.
+
+- Display the directory stack with a space between each entry:
+
+`dirs`
+
+- Display the directory stack with one entry per line:
+
+`dirs -p`
+
+- Display only the nth entry in the directory stack, starting at 0:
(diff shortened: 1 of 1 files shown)


waldyrious:
Nice :) but the term "stack" should be defined somewhere, as it's not 100% intuitive what it means.

sbrl:
@waldyrious Good point. I've added a definition, and fixed the errors in the build.

I'm also thinking that a 'see also' section would be really useful here to point readers at the page for `pushd` / `popd`. What's the protocol on that?

waldyrious:
It's ok to mention them in the main description, as in, for example, [uname.md](../blob/master/pages/common/uname.md).

sbrl:
@waldyrious Right. I've added another sentence to the description. Once this page is given the ok, I'll squash all my commits down.

waldyrious:
LGTM :+1:

sbrl:
@waldyrious Awesome, squashing now

waldyrious:
@sbrl I'll let someone else merge the PR, so there's a second set of eyes looking at this, but if a few days go by without news, I'll merge.

sbrl:
@waldyrious All squashed down into one nice neat package.

felixonmars:
:100: Thanks!

PR #768, opened 2016-02-12
rsync: add example with exclude file

akoenig:
Add an example where `excludes` are defined in a `exclude list`.

### DIFF
--- a/pages/common/rsync.md
+++ b/pages/common/rsync.md
@@ -27,3 +27,8 @@
 - Transfer file over SSH and show progress:
 
 `rsync -e ssh --progress {{remote_host_name}}:{{remote_file}} {{local_file}}`
+
+- Transfer a directory and all its children from a remote to local and respect all excludes defined in a file:
+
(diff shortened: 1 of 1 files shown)


waldyrious:
Thanks for the contribution! The page is starting to get a bit too long, though -- can you see any way to compact it? Maybe one or two of the existing examples could be removed? tl;dr is meant to only provide the 5-6 most common usages, not an exhaustive reference for each command.

igorshubovych:
I do not think this example is quite common to be included. I have personally never used it personally. I think example with `--exclude` will be more common (the last one listed [here](https://en.wikipedia.org/wiki/Rsync#Examples)).

The page is already quite long: 6 examples and it will have 7 examples with your added.

ostera:
I'm with @igorshubovych. :-1:

igorshubovych:
As there are no feedback, I am closing this. Feel free to reopen if you think it should be there.

Thanks anyway! Keep going!

PR #769, opened 2016-02-14
hub command

brpaz:
hub is a command-line wrapper for git that provides some nice extra commands for interacting with Github. see https://hub.github.com/ for more details.

In this PR I have added the basic hub commands.

### DIFF
--- a/pages/common/hub.md
+++ b/pages/common/hub.md
@@ -1,0 +1,16 @@
+# hub
+
+> A command-line wrapper for git that makes you better at GitHub.
+> The commands can also be used using "git" instead of "hub".
+
+- Clone a repository you own:
+
+`hub clone {{ repo_name }}`
+
+- Clone another user repository:
+
+`hub clone {{ github_username }}/{{ repo_name }}`
+
+- Open the current project's issues page:
+
+`hub browse -- issues`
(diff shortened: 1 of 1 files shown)


pages/common/hub.md
```diff
@@ -0,0 +1,15 @@
+# Hub
```
igorshubovych: should be lowercase, same as command name

igorshubovych:
I would probably mention that the same commands are available under `git` too.

brpaz:
OK. I have changed to command to lowercase and also add a note about using "git" instead of "hub"

pages/common/hub.md
```diff
@@ -0,0 +1,16 @@
+# hub
+
+> A command-line wrapper for git that makes you better at GitHub.
+> The commands can also be used using "git" instead of "hub".
+
+- Clone a repository you own:
+
+`hub clone {{ repo_name }}`
```
igorshubovych: Yet one thing. We usually do not add spaces between brackets.

brpaz:
ok. removed the spaces.

ostera:
Looking good! Thanks for the contribution @brpaz.

PR #770, opened 2016-02-15
crontab: add page

kocsenc:
Add a page to `common` for `crontab` command.

I found the most useful use cases were to:
* __edit__ crontab
* __list__ cron jobs
* __remove__ cron jobs

Tested on Ubuntu and OSX.

### DIFF
--- a/pages/common/crontab.md
+++ b/pages/common/crontab.md
@@ -1,0 +1,15 @@
+# crontab
+
+> Schedule cron jobs to run on a time interval for the current user.
+
+- Edit the crontab file for the current user:
+
+`crontab -e`
+
+- View a list of existing cron jobs for current user:
+
+`crontab -l`
+
+- Remove all cron jobs for the current user:
+
+`crontab -r`


igorshubovych:
Good one!
Do you think it is worth mentioning that jobs will be run on behalf of user and every user has his own list of cron jobs?

kocsenc:
@igorshubovych I thought about it - let's do it!

Do you think it should be on the heading description?
(i.e. `> Schedule current user cron jobs to run on a time interval.`)

OR

on each command label?:
`- Edit crontab file for current user:`
`-XXX for current user`

igorshubovych:
Both variants are good.

ostera:
I'm :+1: for the on each user. Less DRY but likely people will head straight to the commands and see it more often. I might be wrong thou!

waldyrious:
I'm with @ostera, that seems like the safest option.

kocsenc:
I've done both approaches. :+1:

igorshubovych:
Thanks!

PR #772, opened 2016-02-16
New page chattr

rozuur:
Change attributes of files or folders


igorshubovych:
Why closed? It is a good page.

PR #775, opened 2016-02-17
Additional java jhat example

notpeter:
Guess who spent yesterday debugging java heap dumps too big for machines with only 8Gb of ram.


ostera:
Sounds not fun at all!

PR #779, opened 2016-02-19
visudo and getent were added

slash3b:


igorshubovych:
Please in future create one pull request per page.

igorshubovych:
Also `getent` is not available under OSX.

waldyrious:
@igorshubovych about the inclusion of sudo, please see the discussion at #756 (/cc @ostera). Let's keep in mind this tool should always strive to support CLI beginners.

slash3b:
@igorshubovych thank you for the comments, I fixed what you mentioned and double checked the files. Next time only one commit!

@waldyrious thanks, indeed, in `visudo` case it makes sense to keep it prefixed with `sudo`.

igorshubovych:
Thanks, LGTM!

waldyrious:
LGTM too, thanks!

PR #786, opened 2016-02-22
sleep: add page

dawidd6:
Added sleep page

### DIFF
--- a/pages/common/sleep.md
+++ b/pages/common/sleep.md
@@ -1,0 +1,15 @@
+# sleep
+
+> Delay for a specified amount of time
+
+- Delay in seconds:
+
+`sleep {{seconds}}`
+
+- Delay in minutes:
+
+`sleep {{minutes}}m`
+
+- Delay in hours:
+
+`sleep {{hours}}h`


oxguy3:
LGTM!

I will note though, the minutes and hours commands only work with the [GNU coreutils version](http://www.gnu.org/software/coreutils/sleep). The [POSIX spec for sleep](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/sleep.html) only requires that the sleep command support an integer number of seconds, and from [the looks of OS X man pages](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/sleep.1.html), I think OS X's version only support the POSIX bare minimum. Thus, might be a good idea to have an osx version of this page in addition to the common version.

igorshubovych:
Thanks you!

oxguy3:
Uhh, could someone with contribution privileges please fix this? Build of master branch is failing b/c of errors in this PR (would have helped if Travis CI didn't just arbitrarily ignore this PR and not show build status on it...)

igorshubovych:
Fixed.

PR #788, opened 2016-02-22
Added missing p7zip tools (7z and 7zr)

oxguy3:
There's already a tldr page for `7za`, the 7zip archiving tool from the p7zip package. But actually, the main executable from p7zip is `7z`. `7z` uses plugins to support lots of archive formats, while `7za` is a standalone version that only supports a limited set of archive formats. There's also `7zr`, which is another standalone version that exclusively supports the .7z formats. (I got all this info from [the ever-indispensible Arch Linux wiki](https://wiki.archlinux.org/index.php/p7zip#Differences_between_7z.2C_7za_and_7zr_binaries))

This PR adds a note to 7za.md explaining the relationship to `7z`, and also adds pages for `7z` and `7zr` (both pages are essentially copies of the original 7za.md page).

### DIFF
--- a/pages/common/7z.md
+++ b/pages/common/7z.md
@@ -12,7 +12,11 @@
 
 - Archive using a specific archive type:
 
-`7z a -t{{7z|zip|gzip|bzip2|tar}} {{compressed}} {{path/to/file}}`
+`7z a -t{{zip|gzip|bzip2|tar|...}} {{compressed}} {{path/to/file}}`
+
+- List available archive types:
+
+`7z i`
 
 - List the contents of an archive file:
 
--- a/pages/common/7z.md
+++ b/pages/common/7z.md
@@ -1,0 +1,19 @@
+# 7z
+
+> A file archiver with high compression ratio.
+
+- Compress directory or file:
+
+`7z a {{compressed.7z}} {{directory_or_file_to_compress}}`
+
+- Decompress an existing 7z file with original directory structure:
+
+`7z x {{compressed.7z}}`
+
+- Compress to zip format:
+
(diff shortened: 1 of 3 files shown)


waldyrious:
Would it make sense to use "archive" and "extract" (even if in parenthesis) to make it clearer what the `a` and `x` stand for?

oxguy3:
Might not be a bad idea; I sorta just blindly copied the way 7za was written, but might as well make it better if we're modifying it. :smile:

oxguy3:
Okay, I went ahead and changed the descriptions of the commands to archive and extract (7zip is capable of working with non-compressed archive formats like .tar files, so compress/decompress were actually misleading/incorrect terminology). I also went ahead and did a bunch of miscellaneous cleanup. In particular, the multipart archive example didn't seem particularly useful (I don't think I've ever created a multipart archive in my life), so I replaced it with the much more helpful list command.

oxguy3:
I realized that there are a lot more archive formats available than the small list I copied from the man page, so I added "..." to the list of archive formats in the `-t` example, and then added the `7z i` command as an example of how to see the full list of available formats.

waldyrious:
If you moved away from "compress"/"decompress", then shouldn't the tokens be changed as well?

oxguy3:
Duhhh, yes, good catch; completely slipped my mind. Will fix that when I'm
at computer.

igorshubovych:
@oxguy3 any updates?

waldyrious:
Since the missing fix is quite simple and not problematic, we'll merge this and make the fix afterwards.

PR #790, opened 2016-02-22
c99: add page

oxguy3:
c99 is the POSIX standard compiler for C that is meant to strictly meet the ISO C standard. This tldr page is based on [this POSIX spec](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/c99.html).

### DIFF
--- a/pages/common/c99.md
+++ b/pages/common/c99.md
@@ -2,18 +2,18 @@
 
 > Compiles C programs according to the ISO C standard.
 
-- Compile source files and create an executable:
+- Compile source file(s) and create an executable:
 
 `c99 {{file.c}}`
 
-- Compile source files and create an executable with a custom name:
+- Compile source file(s) and create an executable with a custom name:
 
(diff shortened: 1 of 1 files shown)


pages/common/c99.md
```diff
@@ -0,0 +1,19 @@
+# C99
+
+> Compiles C programs according to the ISO C standard.
+
+- Compile source files and create an executable:
```
waldyrious: Using "files" may be confusing as the example only uses a single file.

oxguy3: Perhaps I should do "file(s)"? Or I could follow the gcc page's example, and do "{{source1.c}} {{source2.c}}"

waldyrious: "file(s)" sounds good to me, it makes it clear without making the example longer than necessary.

ostera:
Looking good! Thanks @oxguy3

PR #791, opened 2016-02-22
fg: add page

oxguy3:
[fg](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/fg.html) is the POSIX standardized utility for bringing suspended/background jobs to the foreground. If you run a job in the background with `somecommand &`, or if you suspend a job by pressing ^Z, you use fg to resume it.

### DIFF
--- a/pages/common/fg.md
+++ b/pages/common/fg.md
@@ -1,0 +1,11 @@
+# fg
+
+> Run jobs in foreground.
+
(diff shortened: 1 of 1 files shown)


oxguy3:
build failures occurring because #786 was merged with syntax errors -- my page is fine

waldyrious:
Man, you're on a roll! LGTM :+1:

PR #792, opened 2016-02-22
jobs: add page

oxguy3:
[jobs](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/jobs.html) is the POSIX utility for viewing status info on jobs running in the current shell environment. I pretty much only included arguments specified in the POSIX standard, but [Bash has some additional arguments](https://www.gnu.org/software/bash/manual/html_node/Job-Control-Builtins.html#index-jobs).

### DIFF
--- a/pages/common/jobs.md
+++ b/pages/common/jobs.md
@@ -10,6 +10,10 @@
 
 `jobs {{job_id}}`
 
-- Print process IDs of all jobs:
+- Show status and process IDs of all jobs:
+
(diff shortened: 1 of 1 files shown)


oxguy3:
build failures occurring because #786 was merged with syntax errors -- my page is fine

waldyrious:
:+1: :smile:

igorshubovych:
:trophy:

PR #794, opened 2016-02-23
Add sh

oxguy3:
Shocked to find that [sh](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html), the classic Bourne shell, didn't already have a tldr page! Created one based on bash's page, but I tweaked the language a little bit, then modified bash.md to match sh.md.


igorshubovych:
:+1:

PR #795, opened 2016-02-23
tty: add page

oxguy3:
[tty](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/tty.html) is one of the most utterly simple commands ever, but I figure everything ought to have a tldr page.


igorshubovych:
LGTM

PR #797, opened 2016-02-23
basename: add page

oxguy3:
Another fairly straightforward POSIX utility.

Links:
* [POSIX spec](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/basename.html)
* [GNU man page](http://www.gnu.org/software/coreutils/basename)
* [BSD man page](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/basename.1.html)

### DIFF
--- a/pages/common/basename.md
+++ b/pages/common/basename.md
@@ -8,4 +8,4 @@
 
 - Show only the file name from a path, with a suffix removed:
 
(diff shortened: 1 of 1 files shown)


igorshubovych:
LGTM

PR #798, opened 2016-02-23
pwd: fixed minor typo

oxguy3:
Okay this is super trivial but bear with me. The acronym "e.g." means "for example, ", but in this case, an example wasn't being given, so the correct acronym to use is "i.e." (meaning "that is, ").

### DIFF
--- a/pages/common/pwd.md
+++ b/pages/common/pwd.md
@@ -6,6 +6,6 @@
 
 `pwd`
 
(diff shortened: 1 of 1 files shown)


igorshubovych:
Thanks!

waldyrious:
Definitely appreciated, @oxguy3; however trivial, that was the right thing to do.

PR #537, opened 2015-12-31
Moved scripts from linux/osx into common

Parth-Mehrotra:

PR #799, opened 2016-02-23
Move mis-sorted commands

oxguy3:
Okay, this is kind of a big PR, but it would have been a nightmare to do all this in smaller PRs, and it's just the same thing done for a _lot_ of commands.

There are a lot of cross-platform tools that were in the Linux folder, that really belonged in the common folder. Primarily, these are GNU tools which were put in linux/ because they don't ship with OS X. However, they're available on every single platform, including OS X. In a recent PR that moved the base64 tool, we established that the GNU tools should go in common/, and if OS X ships with some weird version of the tool, we put it in osx/, and it'll override the common version.

Here's a breakdown of every tool I moved or otherwise messed with:
## GNU stuff (coreutils/binutils/etc)
- **chroot**: Moved from linux/ to common/. I created osx/chroot.md, which is just a copy of common/chroot.md with the one incompatible example removed (GNU chroot and OSX chroot are mostly compatible).
- **du**: Moved from linux/ to common/. A separate osx/du.md file already existed.
- **head**: Moved from linux/ to common/. A separate osx/head.md file already existed.
- **nm**: Moved from linux/ to common/. A separate osx/nm.md file already existed.
- **dd**: Moved from osx/ to common/, and deleted the linux/ version. Both files were identical (well, save for one character, but I checked the GNU man page, and that character didn't matter). Since the GNU and OSX versions are so compatible, there's no need for an OS X page.
- **shuf**: Moved from linux/ to common/. OS X doesn't ship with _any_ version of this tool, so there's no need for an OS X page.
- **shaXXsum**: Moved from linux/ to common/. OS X does not ship with _any_ version of this tool, so no need for an OS X page. (Grouping sha1sum, sha224sum, sha256sum, sha284sum, and sha512sum into one list item.)
- **hostname**: Moved the osx/ version to common/, and left the linux/ version where it was. The [GNU version](https://www.gnu.org/software/coreutils/hostname) and the [OS X version](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/hostname.1.html) of `hostname` are functionally identical. However, most/all Linux distros ship with a different, more-complex version of `hostname` instead of the GNU version (specifically checked the [Debian version](http://manpages.debian.org/cgi-bin/man.cgi?sektion=1&query=hostname) and the [RH(EL)/CentOS/Fedora version](https://www.freebsd.org/cgi/man.cgi?query=hostname&apropos=0&sektion=1&manpath=Red+Hat+Linux%2Fi386+9&format=html)).
## Miscellaneous
- **pwgen**: Moved from linux/ to common/. This is just [some random Sourceforge project](https://sourceforge.net/projects/pwgen/). It's available on all platforms (all I had to do was `brew install pwgen` on OS X), so I'm not sure how it ended up in linux/.
- **watch**: The linux/ version was just redundant with the common/ version, so I deleted the Linux version. There are no platform-specific versions of this tool; there's just the one global version.
- **xclip**: Moved from linux/ to common/. Even though you usually wouldn't need this tool outside of Linux, it does work with XQuartz on Mac; just `brew install xclip`.
- **wall**: Moved from linux/ to common/. One of the examples used a `-t` flag for adding a timeout to the message. This flag doesn't consistently exist even within Linux (Ubuntu has it, but CentOS doesn't), and it's not a particularly useful flag anyway, so I removed it from the tldr page.
- **screenfetch**: Moved from linux/ to common/. This is just [some random GitHub project](https://github.com/KittyKatt/screenFetch). It's available on all platforms (just got it with `brew install screenfetch`), so I'm not sure how it ended up in linux/.


waldyrious:
First of all, wow. Amazing work.

One comment (I may have to look at the rest of the changes more attentively): regarding dd, are you aware of the distinction discussed in #555?

@igorshubovych has been commenting in several PRs about small differences in OSX versions of commands, so he probably will have something to say here.

oxguy3:
Huh, I must have misread the gnu man page; mea culpa. I'll revert that when
I'm at my computer.

igorshubovych:
Please also read these threads #464, #537, #553.

ostera:
It would be great to merge this even if it's not in it's best shape. We can fix as we need afterwards.

waldyrious:
@ostera agreed, but whoever merges this should at least ensure that the issues raised above are fixed right away. The one I pointed out is fairly easy to fix. Maybe @igorshubovych can be more specific about the issues he linked to, so we also have a clear action path on what needs to be done.

waldyrious:
Reopening as the "previous prompt" wasn't provided by @probot-stale. Please see #1350 and 1360 for details.

sbrl:
Ping @igorshubovych, @ostera, and @oxguy3 - what's the status of this?

waldyrious:
For future reference, @morrme [mentioned](https://gitter.im/tldr-pages/tldr?at=591fcece33e9ee771cc5e748) he might want to tackle this, in the manner @sbrl suggested ("we need to draw up a list of commands that need moving in a new issue, and then create & merge a PR on it"). Is doing that still in your plans, @morrme?

morrme:
If you or someone on your team determines what needs to be moved, I am happy to knock it out.

PR #800, opened 2016-02-23
hn: add page

oxguy3:
[hn](https://github.com/rafaelrinaldi/hn-cli) is a CLI for browsing Hacker News.


igorshubovych:
How we lived without it? LGTM!

PR #801, opened 2016-02-23
xcv: add page

oxguy3:
[xcv](https://github.com/busterc/xcv) is a tool for cut/copy/paste on the command-line (unlike mv/cp, you don't have to specify a destination when cutting/copying, or a source when pasting -- it works just like Ctrl+X/C/V on your keyboard).


igorshubovych:
LGTM.
Oh man, you are doing so much good work! Thanks you! Thank you! Thank you!

PR #805, opened 2016-02-26
look: add page

igorshubovych:
It is rarely used command, but still worth having it.
http://linux.die.net/man/1/look


ostera:
Looking good! Thanks :)

PR #806, opened 2016-02-27
Use a alternative separators

minhajuddin:


85pando:
Good work for pointing that out, this is one very important thing that is missing from the current page.

- - - -

There are multiple code lines for one example. Maybe with a little rephrasing you could do something like

- Replace separator `/` by any other character not used in your string, e.g., `#`:

and then only one example. This should make it clear that this is possible.

waldyrious:
Nice. I'd remove the "potential chars" list, though, I think the text is descriptive enough.

waldyrious:
Sorry we let this one slip for a while. Thanks, this is a very useful contribution (I use this all the time myself).

PR #807, opened 2016-02-27
Sed: adds replace before matching sample

ostera:
Something I find myself doing more and more is to match on a line instead of trying to match all I want with `{{find}}`.

So in a file like:

```erl
{ riak_core, [ {
    {riak_ip, "127.0.0.1"},
    %% ...
}, 
   %% ...
] }
```

Instead of replacing that ip by doing

```zsh
sed 's/riak_ip,"127\.0\.0\.1/riak_ip,"$NEW_IP/'
```

I do

 ```zsh
sed '/riak_ip/s/127\.0\.0\.1/$NEW_IP/'
```

Which has the added benefit of working regardless of whether I use single or double quotes on the file, or  if there's any space between `riak_ip` and the `,` and the actual ip.

### DIFF
--- a/pages/common/sed.md
+++ b/pages/common/sed.md
@@ -6,6 +6,10 @@
 
 `sed 's/{{find}}/{{replace}}/' {{filename}}`
 
+- Replace the first occurrence on all lines matching a pattern:
+
+`sed '/{{line_pattern}}/s/{{find}}/{{replace}}/'`
+
 - Replace all occurrences of a string in a file, overwriting the file (i.e. in-place):
 
 `sed -i 's/{{find}}/{{replace}}/g' {{filename}}`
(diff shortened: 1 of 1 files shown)


waldyrious:
Awesome, TIL :) But isn't the "the first occurrence on all lines" a little confusing? I found myself re-reading that to make sure I got it right.

ostera:
What about `Replace only on lines matching the first pattern`?

waldyrious:
Why "first pattern"? Isn't the token named "line pattern"? Or am I missing something here?

ostera:
Yeah, but it also happens to be the first pattern you specify. `Replace only on lines matching the line pattern` works as well, I'll change it to that :+1:

waldyrious:
Looks clear to me now, nice one :+1:

PR #808, opened 2016-02-28
http: add page

jkbrzt:
https://github.com/jkbrzt/httpie

### DIFF
--- a/pages/common/http.md
+++ b/pages/common/http.md
@@ -25,7 +25,3 @@
 - Pass a user name and password for server authentication:
 
 `http -a {{username:password}} {{example.org}}`
-
-- Specify raw request body via stdin:
-
-`cat data.txt | http PUT example.org`
--- a/pages/common/http.md
+++ b/pages/common/http.md
@@ -29,11 +29,3 @@
 - Specify raw request body via stdin:
 
 `cat data.txt | http PUT example.org`
-
-- View both the response and the request being sent:
-
-`http -v {{example.org}}`
-
-- Send JSON object with non-string fields:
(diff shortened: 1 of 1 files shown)


igorshubovych:
9 examples are too much. Plz think what can be omitted.

jkbrzt:
Seven — better?

igorshubovych:
Better, but I will probably omit the last one too. The recommended amount of commands is 5-6.

jkbrzt:
Okay, I've deleted the last one.

ostera:
Honestly this 7 examples are 99.9% of the use cases ever needed. I'd perhaps remove the first one, since it might be obvious from the output what to do with it (maybe `http url > file`?).

Maybe the authentication and extra headers can be merged? After all, you can authenticate via headers.

ostera:
Nooo, the last one was the most useful one haha (imo, since it shows you can actually use this in a `|` chain)

jkbrzt:
That's the thing :smile:  The original nine examples covered everything important and removing any of them is a compromise. As for `http url > file` vs. `http -d url`: they are actually different and `-d` is more interesting for everyday use.

jkbrzt:
I leave it up to you guys. Feel free to remove/include what you deem important. I might be biased.

igorshubovych:
After reading the thread I agreed with @ostera. Let's have 7 commands. So i merged it manually with all 7 commands.

Thank you for contribution!

ostera:
Yays!

jkbrzt:
👍👍👍

PR #812, opened 2016-03-01
Ssh keygen filepath

jedahan:
i always forget its lowercase -f

### DIFF
--- a/pages/common/ssh-keygen.md
+++ b/pages/common/ssh-keygen.md
@@ -6,6 +6,10 @@
 
 `ssh-keygen`
 
+- Specify file in which to save the key:
+
+`ssh-keygen -f ~/.ssh/{{filename}}`
+
 - Generate a DSA 2048 bit (default) key:
 
 `ssh-keygen -t dsa`


ostera:
Could you make sure not to include other commits in your PR?

jedahan:
whoops, fixed

waldyrious:
A couple comments:

- it would be nice to use the term "file" or "filename" in the example description, to make the meaning of -f even more obvious.
- if we're going to use "location" (i.e. a path, as I see it) rather than just "filename", the "~/.ssh/" should be part of the token.

Ideally we'd choose one of the two options above. I'm more inclined towards the first, even if the second is technically correct, considering that tldr aims to be beginner-friendly.

jedahan:
great feedback @waldyrious - I am definitely having trouble figuring out a concise wording that also fits with that real-world example that is also easy to parse TT

jedahan:
ssh prompt is
```
Generating public/private rsa key pair.
Enter file in which to save the key (/Users/jedahan/.ssh/id_rsa):
``

waldyrious:
Sounds good @jedahan

jedahan:
OK, try number 2

ostera:
Thanks!

PR #815, opened 2016-03-02
added readlink

Cvetomird91:
Added readlink for getting symlink info.

I also added the -f option for getting the absolute path to a file, which proved to me to be very useful in 
shell scripting.

### DIFF
--- a/pages/common/readlink.md
+++ b/pages/common/readlink.md
@@ -1,0 +1,11 @@
+# readlink
+
+> Follow symlinks and get symlink information.
+
+- Get the actual file to which the symlink points:
+
+`readlink {{filename}}`
+
+- Get the absolute path to a file:
+
+`readlink -f {{filename}}`


igorshubovych:
-f option is not available in OSX.

Cvetomird91:
I moved the page to the Linux section.

igorshubovych:
Well, `readlink` is available on OSX. The 1st example works correctly. But it does not have `-f` option.

Cvetomird91:
I have a FreeBSD box, I will test it there. Excuse me, but I don't have a Mac OS X instance at my disposal.

igorshubovych:
NP. That's why I am telling you.

Well, I think you can move it to common anyway. And I (or somebody else) can just add the one to osx folder. It is available on OSX anyway, if you install coreutils.

Sorry, can you please revert your last commit back? It is ok, if it is in common.

Cvetomird91:
The page has been moved to the common section and I just squashed the commits.

ostera:
Thanks @Cvetomird91!

PR #819, opened 2016-03-08
Fix example for grep for using standard input instead of a file

kartikanand:
Currently the example for using `grep` with standard input instead of a file seems wrong. 
The current example is 

`cat {{file_path}} | grep {{something}}`

The example itself mentions {{file_path}}

The following are better examples:

`cat - | grep {{something}}`

or even

`cat | grep {{something}}`

and even a shorter form

`grep {{something}}`

I've included only the first one as it feels best for a new user.

### DIFF
--- a/pages/common/grep.md
+++ b/pages/common/grep.md
@@ -33,7 +33,7 @@
 
 - Use the standard input instead of a file:
 
-`cat {{file_path}} | grep {{something}}`
+`cat - | grep {{something}}`
 
 - Invert match for excluding specific strings:
 


rubenvereecken:
While it's true that this new example uses actual user-generated input, I don't feel like it adds value. Neither example is useful in practice, they're both designed to demonstrate the exact same thing.

I'll wait on what the others have to say.

kartikanand:
I agree with you. The thing is people don't really use grep with standard input a lot. The thing is that the old example is wrong. So either we should remove it, or atleast add the right example.

rubenvereecken:
It's a perfectly valid stdin example, it's just not the way you'll want to use grep. That's why it's at the bottom.

kartikanand:
Yeah but new users can get confused with terms. The example says how to use grep with standard input instead of file, and the example itself is passing a file. I guess we can change the example to something like "how to pipe input into grep" ?

sethwoodworth:
I for one, only use grep in pipelines or on stdin. I use other tools like
ack and ag to find strings in files.

PR #821, opened 2016-03-10
pv -d does not exist

aicioara:
Not sure what was the original intention, but the -d option does not exist. 

[Source](http://linux.die.net/man/1/pv)

@d33tah?

### DIFF
--- a/pages/common/pv.md
+++ b/pages/common/pv.md
@@ -14,10 +14,6 @@
 
 `pv -cN in {{big_text_file}} | grep {{pattern}} | pv -cN out > {{filtered_file}}`
 
-- Attach to an already running process and see its file reading progress:
-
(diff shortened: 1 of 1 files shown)


igorshubovych:
From `pv --help`:

```
-d, --watchfd PID[:FD]   watch file FD opened by process PID
```

I have the following error:
```
pv: -d: not available on systems without /proc/self/fdinfo
```

aicioara:
Maybe I'm using an outdated version.

`pv 1.2.0` at the moment

igorshubovych:
pv 1.6.0

aicioara:
Happy to drop the PR, didn't realize pv was under active development.

igorshubovych:
Ok. So I am closing it.
Thanks anyway.

PR #822, opened 2016-03-14
crontab: add format reference

kkk777kkk:


igorshubovych:
pages/common/crontab.md:19: TLDR103 Command example is missing its closing backtick

igorshubovych:
well, this looks cool.

but it will break the majority of the parsers.

kkk777kkk:
Not expert in markdown, but Github preview shows it smoothly. Is there any other way to do multiline?

igorshubovych:
This is ok for Markdown. It violates our project format though:
https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#markdown-format

ostera:
:-1: I think we can come up with a better way of explaining cron syntax than this.

This comes to mind:

```
Entry   = Minutes Hours Day Month Weekday
Minutes = [0..60]
Hours   = [0..24]
Day     = [0..31]
Month   = [0..12]
Weekday = [0..7]
```

kkk777kkk:
As for me you proposition bring even more confusion and not bringing complete understanding how final entry should look like, my (wikipedia) variant will work better for somebody who never seen how entry looks like.
But anyway it still has more than one line so not fits into https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#markdown-format

ostera:
That was the tersest I'd go, but here's a slightly more verbose attempt:

```md
- Job definition in crontab:

`{{schedule}} {{command}}`

- Schedule definition (value in range or * for all):

`(minutes: 0..60) (hours: 0..24) (days: 0..31) (months: 0..12) (weekday: 0..6)`
```

That's more `tldr`-format friendly, and still doesn't recur to ASCII art to point out which values are which.

igorshubovych:
@ostera 
It is still complicated. It is better to do a few concrete examples.

waldyrious:
Concrete examples would probably be the easiest way to get an idea of the pattern. I'm not opposed to @ostera's proposal though, even if the wording can be a little awkward as it is (sorry, I can't think of anything better atm)

PR #826, opened 2016-03-19
Wrong order of arguments in command

mamigot:
The order of arguments in `scp` was [wrong](http://serverfault.com/a/264598/283749). Thank you, though –this is awesome!


igorshubovych:
Good catch! Thanks!

PR #828, opened 2016-03-19
shorten history -c explanation using suggestion by @waldyrious

matkoniecz:


waldyrious:
LGTM, thanks for doing this. /cc @igorshubovych

igorshubovych:
Thanks!
Both @matkoniecz and @waldyrious!

Issue #823, opened 2016-03-15
Pages for gulp and bower


emileber:
I want to contribute by adding pages for gulp and bower but I'm not sure if it's a good fit since tldr is mostly about UNIX / Linux / OSX / SunOS. I searched the issues (both open and closed) and PRs and have not found any either positive nor negative about gulp or bower.

I want to add pages for those because I found myself trying `tldr bower` with no avail.

igorshubovych:
Do it! Create `common/gulp.md` and start editing.

emileber:
Both commands have been merged. I'm closing this.

PR #831, opened 2016-03-24
Added gulp to common pages.

emileber:
Related to issue #823 that I recently opened.

### DIFF
--- a/pages/common/gulp.md
+++ b/pages/common/gulp.md
@@ -1,0 +1,12 @@
+# gulp
+
+> JavaScript task runner and streaming build system.
+> Tasks are defined within gulpfile.js at the project root.
+
+- Run the default task:
+
+`gulp`
+
+- Run individual tasks:
+
+`gulp {{task}} {{othertask}}`


pages/common/gulp.md
```diff
@@ -0,0 +1,11 @@
+# gulp
+
+> JavaScript task runner and streaming build system.
+
+- Run the default task defined inside gulpfile.js:
+
+`gulp`
+
+- Run specific task:
```
waldyrious: Shouldn't this say "tasks" given the actual example command below?

pages/common/gulp.md
```diff
@@ -0,0 +1,11 @@
+# gulp
+
+> JavaScript task runner and streaming build system.
```
rubenvereecken: It so happens I started using Gulp for the first time yesterday. To slightly touch @waldyrious's comment, I'd say we compromise and have a line here saying that tasks are defined in a Gulpfile. That way we don't really have to explain what tasks are. Whaddayathink @waldyrious?

waldyrious: I was going to suggest mentioning the gulpfile in the description of the second comment, but here works too :)

pages/common/gulp.md
emileber: Yes, it should say "tasks". I feel like explaining what are tasks is out of scope for a tldr page. It's provided in the first example that they are defined in a `gulpfile.js`, the user should seek further information himself by checking the gulpfile or the gulp documentation.

pages/common/gulp.md
emileber: Since gulp is so simple, I wasn't sure it deserved a 2-line description. But I could move the gulpfile mention up into the description.

pages/common/gulp.md
waldyrious: We don't need to write a lot about gulp, just a friendly hint that can save users some time. How about changing the command description to "Run specific tasks from the gulpfile:"?

igorshubovych:
I like it as is. LGTM.

waldyrious:
I'll :+1: this if @emileber makes the change I suggested above :)

emileber:
Yes, i'll do them when I have time, probably tomorrow.

emileber:
Fixed it.

waldyrious:
@emileber the change I was referring to was changing the second example description to "Run specific tasks from the gulpfile:". Instead, it seems that you changed the *first* command description from "Run the default task defined inside gulpfile.js:" to "Run the default task:". Care to explain why you chose to do it that way? That reduces information rather than adding...

pages/common/gulp.md
```diff
@@ -0,0 +1,12 @@
...
+> Tasks are defined within gulpfile.js at the project root.
```
emileber: @waldyrious I moved the gulpfile bit up into the description and shortened the commands description. I made that change according to @rubenvereecken note and you agreeing with him. Is that correct for you?

waldyrious: Weird, I didn't see it there (I looked). I think it works, yes, and is more in the DRY spirit :+1: Thanks!

PR #834, opened 2016-03-27
umask: add page

slash3b:
New page for `umask` command.

### DIFF
--- a/pages/linux/umask.md
+++ b/pages/linux/umask.md
@@ -1,0 +1,15 @@
+# umask
+
+> Display file creation mask or set file creation mask to octal value.
+
+- Display current value as octal:
+
+`umask`
+
+- Display current value symbolically:
+
+`umask -S`
+
+- Turn off permission bits with octal value 'nnnn':
+
+`umask {{nnnn}}`
--- a/pages/linux/umask.md
+++ b/pages/linux/umask.md
@@ -1,19 +1,19 @@
 # umask
 
-> Set permission for newly created files for current user.
+> Manage the read/write/execute permissions that are masked out (i.e. restricted) for newly created files by the user.
 
-- Display mask value as octal:
+- Display the current mask in octal notation:
 
 `umask`
 
-- Set the mask using octal notation:
-
-`umask {{nnnn}}`
-
-- Display mask value in human-readable mode:
+- Display the current mask in symbolic (human-readable) mode:
 
 `umask -S`
 
-- Set the mask using symbolic notation:
+- Change the mask symbolically to allow read permission for all users (the rest of the mask bits are unchanged):
 
-`umask {{ u - user | g - groups | o - others | a - all }}{{ + enable | - disable | = enable specified }}{{ r - read | w -write | x- execute }}`
+`umask {{a+r}}`
+
+- Set the mask (using octal) to restrict no permissions for the file's owner, and restrict all permissions for everyone else:
+
+`umask {{077}}`
--- a/pages/linux/umask.md
+++ b/pages/linux/umask.md
@@ -1,15 +1,19 @@
 # umask
 
-> Display file creation mask or set file creation mask to octal value.
+> Set permission for newly created files for current user.
 
-- Display current value as octal:
+- Display mask value as octal:
 
 `umask`
 
-- Display current value symbolically:
+- Set the mask using octal notation:
+
+`umask {{nnnn}}`
+
+- Display mask value in human-readable mode:
 
 `umask -S`
 
-- Turn off permission bits with octal value 'nnnn':
+- Set the mask using symbolic notation:
 
-`umask {{nnnn}}`
+`umask {{ u - user | g - groups | o - others | a - all }}{{ + enable | - disable | = enable specified }}{{ r - read | w -write | x- execute }}`


pages/linux/umask.md
```diff
@@ -0,0 +1,15 @@
+# umask
+
+> Display file creation mask or set file creation mask to octal value.
```
waldyrious: This description wasn't entirely clear for me. Wikipedia puts it better: "determines the settings of a **mask** that controls how file permissions are set for [the current **user**'s] newly created files" (emphasis mine). Shall we rewrite the command description to use something closer to this?

pages/linux/umask.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Display current value as octal:
+
+`umask`
+
+- Display current value symbolically:
```
waldyrious: "symbolically" is too vague IMO. How about "using human-readable symbols"?

pages/linux/umask.md
```diff
@@ -0,0 +1,15 @@
...
+
+`umask -S`
+
+- Turn off permission bits with octal value 'nnnn':
```
waldyrious: Not sure this is clear for beginners... any ideas how to reword the description?

pages/linux/umask.md
slash3b: I agree. What is better "Set file permissions for newly created files." or something like "Set file mode creation mask." ?

pages/linux/umask.md
slash3b: Ok, I like "human-readable", haven't thought about that. Will make the change.

pages/linux/umask.md
slash3b: I tried to describe exactly what command does, here are a few more variants:
```
- Set permissions to be turned off with sequence of 3 or 4 numbers from 0 to 7:
`umask {{nnnn}}`
- Turn off permissions providing permission rule with number sequence 'nnnn' where 'n' >= 0 or <= 7:
`umask {{nnnn}}`
```

pages/linux/umask.md
waldyrious: I would prefer "user" and "mask" to remain in the description somehow.

waldyrious:
I'm still confused tbh, especially with the "turn off" part. Here's what Wikipedia says, which seems straightforward to me:

`umask 0077        # set the mask to 0077`

pages/linux/umask.md
waldyrious: I'm still confused tbh, especially with the "turn off" part. Here's what Wikipedia says, which seems straightforward to me:

`umask 0077        # set the mask to 0077`

slash3b: Ok, I'm trying to explain it. Set the mask for ... may be confusing for beginners, I googled it. A lot of people do not know what the mask is and how it works in this case.

If you insist I will change it to `- Set the mask to {{nnnn}}`

waldyrious: I'm not advocating for using "mask" in this example description, just in the global description (in order to explain where the command name comes from). What confuses me here is the "turn off" expression -- what do you mean by that? Apart from that, I'm perfectly fine with phrasing the example description with "set permissions".

slash3b: By "turn off" I mean the following example:
```
[root@latitude /]# umask
0022
[root@latitude /]# mkdir test
drwxr-xr-x    2 root root  4096 Apr 13 21:13 test
[root@latitude /]# umask 0777
[root@latitude /]# mkdir test2
d---------    2 root root  4096 Apr 13 21:14 test2

```

Please find updated commit with corrections.

waldyrious: Huh. I didn't get this at first, but re-reading [Wikipedia](https://en.wikipedia.org/wiki/Umask) (emphasis mine):

> The mask **restricts** permission settings. If the mask has a bit set to "1", it means the corresponding initial file permission will be **disabled**. A bit set to "0" in the mask means that the corresponding permission will be determined by the program and the system. In other words, the mask acts as a last-stage filter that **strips away permissions** as a file is created; **each bit that is set to a "1" strips away its corresponding permission**.

This isn't clear in the command description. We should find a way to make that clear, without making it too long.

Also, perhaps we should add at least one example using the [symbolic notation](https://en.wikipedia.org/wiki/Umask#Setting_the_mask_using_symbolic_notation), which is more intuitive.

slash3b: Sure, I'll add example with symbolic notation.
Do you have any suggestions for  `umask {{nnnn}}` ?

pages/linux/umask.md
```diff
@@ -1,15 +1,19 @@
...
+
+- Display mask value in human-readable mode:
 
 `umask -S`
 
-- Turn off permission bits with octal value 'nnnn':
+- Set the mask using symbolic notation:
 
-`umask {{nnnn}}`
+`umask {{ u - user | g - groups | o - others | a - all }}{{ + enable | - disable | = enable specified }}{{ r - read | w -write | x- execute }}`
```
slash3b: @waldyrious, please check my last commit. I'm not sure about symbolic notation example I added. It seems wrong to not show all options and and the same time the line seems cumbersome.

waldyrious: Will do later today and give you feedback :) sorry for the delay.

waldyrious:
Here's my suggestion. What do you think?

```
# umask

> Manage the read/write/execute permissions that are masked out (i.e. restricted) for newly created files by the user.

- Display the current mask in octal notation:

`umask`

- Display the current mask in symbolic (human-readable) mode:

`umask -S`

- Change the mask symbolically to allow read permission for all users (the rest of the mask bits are unchanged):

`umask {{a+r}}`

- Set the mask (using octal) to restrict no permissions for the file's owner, and restrict all permissions for everyone else:

`umask {{077}}`
```

slash3b:
thank you @waldyrious, you really managed to soften the rough edges. 
I am going to commit your version.

waldyrious:
Glad you like it :) I'll merge then. Thanks for the patience!

slash3b:
thank you for guidance.

PR #836, opened 2016-03-29
xclip:Add copy to system clipboard

jezenielzapanta:


ostera:
Looks alright to me, but I feel it's not very clear what's the difference between the first and second?

Besides that, needs to fix this: https://travis-ci.org/tldr-pages/tldr/builds/119129502#L163

jezenielzapanta:
@ostera sorry if it is too vague. The first example is like selecting the text and pasting it using `middle mouse` button. While the second one is like using `ctrl+c` and can be paste using `ctrl+v`.

waldyrious:
@jezenielzapanta is you explain what the "sel" means in the command description, the meaning may become clearer :) where does the word come from? I'd assume from "selection", but please confirm.

jezenielzapanta:
@waldyrious Yes, it means `-selection clipboard`, there are 3 selections for X windows system, `XA_PRIMARY`, `XA_SECONDARY` and `XA_CLIPBOARD`. If no `-selection` is passed it will default to `XA_PRIMARY`. What's your opinion in writing the difference of the 3? or the first and second examples?

waldyrious:
I suspect we can't explain the differences between the three succinctly enough in a tldr page. Perhaps we could add a link to the description pointing to a good (concise and beginner-friendly) reference on the subject?

ostera:
I think we don't need to explain those differences here. If the default behavior is copying it so that you can paste it right after with `Ctrl+V`, then it should be good enough. Merging as is, we can change if we need later.

waldyrious:
I'm planning to take a stab at this later.

PR #837, opened 2016-03-29
git-submodule: Add tldr page for the git-submodule command.

kazmasaurus:
I think this pretty well covers the basics. I'm not including removing a submodule because you use `git-rm` instead of `git-submodule` for that, but I would be happy to add it in if it's something that makes sense here.


igorshubovych:
I like it! LGTM!

felixonmars:
:+1:

PR #838, opened 2016-03-30
added bower to common pages

emileber:
Still related to issue #823. This command page is a lot based on the `npm` common page.

Link to the [bower homepage](http://bower.io/).

### DIFF
--- a/pages/common/bower.md
+++ b/pages/common/bower.md
@@ -1,0 +1,32 @@
+# bower
+
+> A package manager optimized for front-end web development.
+> A package can be a GitHub user/repo shorthand, a Git endpoint, a URL or a registered package.
+
+- Install a project's dependencies, listed in its bower.json:
+
+`bower install`
+
+- Install one or more packages to the bower_components directory:
+
+`bower install {{package}} {{package}}`
+
+- Uninstall packages locally from the bower_components directory:
+
+`bower uninstall {{package}} {{package}}`
+
+- List local packages and possible updates:
+
+`bower list`
+
+- Display help information about a bower command:
+
+`bower help {{command}}`
+
+- Create a bower.json file for your package:
+
+`bower init`
+
+- Install a specific dependency version, and add it to bower.json:
+
+`bower install {{local_name}}={{package}}#{{version}} --save`


pages/common/bower.md
```diff
@@ -0,0 +1,32 @@
+# bower
+
+> A package manager optimized for the front-end web development.
```
waldyrious: I would remove "the" from "the front-end web development".

pages/common/bower.md
```diff
@@ -0,0 +1,32 @@
...
+> A package can be a GitHub shorthand, a Git endpoint, a URL, a registered package.
```
waldyrious: suggestion: "or a registered package". Also, "github shorthand" may not be entirely clear at first. I wonder if we can come up with a more self-evident, yet still concise, expression to mean the same. Any ideas?

pages/common/bower.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Install packages to bower_components directory:
```
waldyrious: suggestion: "to *the* bower_components directory"

pages/common/bower.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Install a specific dependency version, and add it to the bower.json:
+
+`bower install {{local_name}}={{endpoint}}#{{version}} --save`
+
+- List local packages and possible updates:
+
+`bower list`
+
+- Uninstall packages locally from your bower_components directory:
```
waldyrious: Suggestion: "your" --> "the" (for consistency)

waldyrious:
Very nice, thanks! But apart from the inline comments, I'd suggest changing the order of the commands to match the "simple to complex" philosophy described in CONTRIBUTING.md. In fact, I'd pretty much turn the current command order upside down, but as a user of the tool yourself, I'll leave it to you to propose a specific ordering in light of that principle.

pages/common/bower.md
igorshubovych: Just "front-end development" or "web development". "Front-end web development" is a tautology.

emileber: I'll go with "front-end development" as it's the key word with bower. Web development includes the backend which bower doesn't focus on.

ostera: I think `front-end web development` is accurate. From my experience working with web dev, `front-end` is user facing, but the user might be a `back-end` service.

Sure, it's common for recruiters to refer to `front end devs` to people who can juggle some css/js, but being engineers, the semantics of front and back end mean nothing but what is exposed.

Interestingly, by that definition, `front end dev` and `API dev` would be fairly similar :stuck_out_tongue: since an API is the front end of some service.

emileber: I was going to disagree with @igorshubovych but then I tried to find a reference somewhere of front-end development referring to something else than the web (html/css/js) but I couldn't. 

So, could it really be that confusing? What's our final choice?

pages/common/bower.md
emileber: I think I could change the first command example to reflect all the possible cases like so:

```
bower install {{package_name}} {{endpoint}} {{user/package}} {{url}}
```

But then, it almost more confusing. I think I should just remove the "Github shorthand" part since it adds little.

emileber:
For the ordering, I followed the npm command order and it makes sense since it's from the (probably) most used command to the least used/expected to be used and they're almost all as simple. Except for the specific dependency version one, which I could move down to the bottom of the list.

Most of the time, you want to install a new package, install the dependencies of a project or uninstall a package. I could swap list with uninstall.

What do you think?

pages/common/bower.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Display help information about a bower command:
+
+`bower help {{command}}`
+
+- Create a bower.json file for your package:
+
+`bower init`
+
+- Install a specific dependency version, and add it to the bower.json:
```
waldyrious: Either "to bower.json" or "to the bower.json file".

pages/common/bower.md
```diff
@@ -0,0 +1,32 @@
+# bower
+
+> A package manager optimized for front-end web development.
+> A package can be a GitHub shorthand, a Git endpoint, a URL or a registered package.
```
waldyrious: How about "Github user/repo shorthand"?

pages/common/bower.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Install the project dependencies listed in bower.json:
```
waldyrious: I'd suggest "Install a project dependencies listed in its bower.json".

pages/common/bower.md
```diff
@@ -0,0 +1,32 @@
...
+
+`bower install`
+
+- Install packages to the bower_components directory:
```
waldyrious: Could this be clearer as "one or more packages"?

pages/common/bower.md
```diff
@@ -0,0 +1,32 @@
...
+
+`bower install {{endpoint}} {{endpoint}}`
```
waldyrious: "endpoint" is slightly unclear here. Any chance we could use a different token name?

pages/common/bower.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Uninstall packages locally from the bower_components directory:
+
+`bower uninstall {{name}} {{name}}`
```
waldyrious: "name" is vague. Should this be "package_name"?

pages/common/bower.md
emileber: What about `{{package_name}}` or just `{{name}}` as there are to much possibilities to choose from.

emileber:
Ok, with consistency in mind, I'd suggest to use `{{package_name}}` for every command.

pages/common/bower.md
waldyrious: How about `{{package}}`, since we explain in the main description that a package can be specified in several different formats?

pages/common/bower.md
waldyrious: Or maybe {{package}}, see above inline comment.

emileber:
I made the changes.

waldyrious:
Looks great, thanks for bearing with us :)

PR #840, opened 2016-03-30
qemu.md

Cvetomird91:
I added a page for the QEMU emulator. 

It usually is operated though CLI and opens a dialog, native to the OS's GUI, in which virtual machine 
displays are shown.

It may be useful for users, who need to create virtual instances on architectures different than x86 and x64, as it supports PowerPC, MIPS, ARM and quite few other CPU architectures.

Update:
It comes with separate commands for each CPU architecture, e.g. qemu-system-ppc64, qemu-system-i386, qemu-system-arm. This is the reason examples in this markdown file are involving separate commands. With qemu-img you can create a virtual HDD device for any of your virtual machines, e.g. on which you can install a particular OS. 

All of those binaries are usually bundled with the standard QEMU package tha you can install on a Debian/Ubuntu system with for example apt-get.

### DIFF
--- a/pages/common/qemu.md
+++ b/pages/common/qemu.md
@@ -1,0 +1,28 @@
+# qemu
+
+> Generic machine emulator and virtualizer.
+> Supports a large variety of CPU architectures.
+
+- Create disk image with a specific size (in gigabytes):
+
+`qemu-img create {{image_name.img}} {{gigabites}}G`
+
+- Boot from image emulating i386 architecture:
+
+`qemu-system-i386 -hda {{image_name.img}}`
+
+- Boot from image emulating x64 architecture:
+
+`qemu-system-x86_64 -hda {{image_name.img}}`
+
+- Boot QEMU instance with a live ISO image:
+
+`qemu-system-i386 -hda {{image_name.img}} -cdrom {{os-image.iso}} -boot d`
+
+- Specify amount of RAM for instance:
+
+`qemu-system-i386 -m 256 -hda image_name.img -cdrom os-image.iso -boot d`
+
+- Boot from physical device (e.g. from USB to test bootable medium):
+
+`qemu-system-i386 -hda /dev/{{storage_device}}`


pages/common/qemu.md
```diff
@@ -0,0 +1,36 @@
+# qemu
+
+> Generic machine emulator and virtualizer.
+> Supports a large variety of CPU architectures.
+
+- Create disk image with a specific size (in gigabytes):
+
+`qemu-img create {{image_name.img}} {{gygabytes}}G`
```
waldyrious: there's a typo here: "gygabytes" --> "gigabytes"

waldyrious:
Looks great, but is there any chance it can be shortened a bit?

Cvetomird91:
I shortened it with two options. Now they are six. 

I can remove the option to boot from physical bootable storage device
(e.g. in case the user would like to test a bootable USB medium he just created).

I considered to be useful though. :)

waldyrious:
6 examples are ok, thanks! LGTM :+1:

igorshubovych:
I do not know too much about qemu. Why do you have several commands on a single page: `qemu-system-i386`, `qemu-im`, etc?

Cvetomird91:
I updated my initial message here.

igorshubovych:
They are different commands. I suggest they should be listed in different pages.

waldyrious:
I'm not so sure @igorshubovych. Unless each command is as complex and versatile as e.g. git subcommands, it would probably be a disservice to the user to force them to jump between tldr pages rather than presenting a unified summary. @Cvetomird91 what do you say?

igorshubovych:
Well, at least `qemu-img` is quite different from `qemu-system`.

Cvetomird91:
All qemu-system-* commands have almost identical options.

 I just made diffs by comparing the outputs of "qemu-system-mips -help", "qemu-system-i386 -help" and "qemu-system-x86_64 -help" and they have almost the same options. The only differences were options which are specific for the particular CPU architecture. 

I've selected those options because they are most commonly used tasks in the typical workflow with such a hypervisor and were the commands I most commonly looked up when getting into QEMU. 

Those commands include the following steps.

- create a virtual HDD
- fire up a virtual machine by booting installation media (e.g. live Linux USB)
- install a particular operating system on that HDD via the bootable media
- boot the newly installed OS on the virtual HDD

 Those of you that've used VirtualBox can agree that it is used most commonly for these purposes, but the steps are performed through a GUI when using VB, not through CLI. That's the part that can be confusing at the beginning when a person starts using QEMU. ;)

igorshubovych:
What about qemu-img command?

Cvetomird91:
I can create a separate page for it shortly, as system administrators most probably use it to manage Xen/KVM instance storage devices.

I think it's best to stick to the contributions guideline and make the page more suitable for beginners:

> When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness.

waldyrious:
That sounds like a good plan @Cvetomird91 and I agree with that reasoning. For me we can merge this as-is.

igorshubovych:
Thank you!

PR #842, opened 2016-03-31
replace − with dash(-)

pindexis:
fix some commands templates


igorshubovych:
Thanks for the fix!
11BAxHG7paxJcI

PR #843, opened 2016-04-01
Sftp

smevawala:
Added an example for pulling folders.

### DIFF
--- a/pages/common/sftp.md
+++ b/pages/common/sftp.md
@@ -20,6 +20,10 @@
 
 `put {{/path/local_file}}`
 
+- Transfer remote file to the local system recursively (works with `put` too):
+
+`get -R {{/path/remote_file}}`
+
 - Get list of files on local machine:
 
 `lls`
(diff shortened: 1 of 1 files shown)


waldyrious:
This page is getting rather large. We usually recommend 5-6 examples, and this change bumps the page to 7. Perhaps we could remove one of the existing examples in order to add this one? Remember, we only aim to list the most common usages, not provide an exhaustive usage reference.

smevawala:
True, maybe I could delete 'ls' and 'lls' and replace that with '?' which loads a list of sftp commands?  I feel as thought the recursive flag is less intuitively explained in that list.

waldyrious:
I don't regularly use sftp myself, so I'll let others comment on your proposal. Pinging @igorshubovych @ostera @rubenvereecken @rprieto @felixonmars

igorshubovych:
It is still 7 examples. I think it is useful and common. LGTM.

felixonmars:
:+1:

PR #844, opened 2016-04-05
Update systemctl.md;Add page of systemd-analyze

responsible:


pages/linux/systemctl.md
```diff
@@ -6,9 +6,9 @@
 
 `systemctl --failed`
 
-- Start/Stop/Restart a service:
+- Start/Stop/Restart/Reloadconfiguration a service:
```
waldyrious: "Reloadconfiguration a service" is not valid English. Can't we reword this?

pages/linux/systemctl.md
```diff
@@ -18,6 +18,10 @@
 
 `systemctl enable/disable {{unit}}`
 
+- Mask/Unmask a unit, make it impossible to start it:
```
waldyrious: I've never used systemctl, so this might be a stupid question, but why would one want to do this? It strikes me as odd, and possibly an unusual action (which wouldn't fit tldr's ethos).

pages/linux/systemd-analyze.md
```diff
@@ -0,0 +1,11 @@
+# systemd-analyze
+
+> Show timing details about the boot process.
```
waldyrious: Perhaps add a short description of what a "unit" is.

pages/linux/systemctl.md
responsible: It can prevent the service to be started on bootup in advance.For example, if you don't want tomcat start automaticlly, you can use `systemctl mask tomcat.service`(it will ln -s '/dev/null' '/etc/systemd/system/tomcat.service),then `systemctl enable tomcat.service` can not take effect. Sometimes, two services boot automaticlly both time will conflict, and we can mask one.

waldyrious: In that case, I'd suggest changing the description to make it clear this is about preventing automatic execution, rather than preventing execution altogether. Or did I get that wrong?

pages/linux/systemd-analyze.md
```diff
@@ -1,6 +1,6 @@
 # systemd-analyze
 
-> Show timing details about the boot process.
+> Show timing details about the boot process(services, mount points, devices, sockets).
```
waldyrious: I'd suggest actually using the word that is going to be used in the tokens, e.g. "unit". Also, this needs a space before the parenthesis :)

pages/linux/systemctl.md
responsible: Agree, it is more clearly.

waldyrious:
LGTM, thanks! :+1:

PR #847, opened 2016-04-14
id: add page

WALL-E:
id - print real and effective user and group IDs


waldyrious:
Thanks for your contribution! Would you care to clarify what "effective" means in this context? Also, I'd suggest using "identity" rather than ID in the description.

ostera:
I think he means `current`. At least that was the output I got when trying the command

waldyrious:
Thanks for the fixes and sorry for the delay!

PR #849, opened 2016-04-16
less: Add example for enabling ANSI color output

SShrike:
Example usage of the `-R` flag for enabling the interpretation of ANSI escape "colour" codes by less.

### DIFF
--- a/pages/common/less.md
+++ b/pages/common/less.md
@@ -26,7 +26,7 @@
 
 - Enable output of ANSI colors:
 
-`{{something}} | less -R`
+`git diff --color | less -R`
 
 - Exit:
 
--- a/pages/common/less.md
+++ b/pages/common/less.md
@@ -26,7 +26,7 @@
 
 - Enable output of ANSI colors:
 
-`less -R {{source_file}}`
+`{{something}} | less -R`
 
 - Exit:
 
--- a/pages/common/less.md
+++ b/pages/common/less.md
@@ -24,6 +24,10 @@
 
 `?{{something}}   then   n (next), N (previous)`
 
+- Enable output of ANSI colors:
+
+`less -R {{source_file}}`
+
 - Exit:
 
 `q`


waldyrious:
Interesting. I'm a `less` noob -- can you briefly explain where the colors come from / where they are specified?

SShrike:
To output colours on a UNIX terminal (e.g. Linux, BSD, OS X) you make use of [ANSI escape codes/sequences](https://en.wikipedia.org/wiki/ANSI_escape_code). For example if you wanted the following text to be red in a terminal: `Hello world!`, then you would wrap it in the following codes: `\x1b[31m<text>\x1b[0m`, resulting in: `\x1b[31mHello world!\x1b[0m`. `\x1b[31m` specifies the colour, and `\x1b[0m` resets the colour back to default so that anything printed after it is not red.

If you don't use the '-R' switch, less will print out garbled output when these escape codes are used. It is also worth noting that the '-R' switch is a variation of the '-r' switch, with the difference being that '-R' will only interpret ANSI escape codes that modify colours. The reason for this difference is that if you were to interpret all escape codes, some might break the pager as they can modify the cursor position and such.

Hopefully I managed to explain it somewhat. I'd recommend reading the [Wikipedia article](https://en.wikipedia.org/wiki/ANSI_escape_code) for a better explanation.

waldyrious:
I suppose I didn't make it clear what confused me, sorry about that. I knew about color escape codes, but so far I've only used them in scripts, to control the output in the command line. This particular example uses a source file -- is the intention that the source file will have those color codes embedded in it? Is that usual for some kind of file, or are we talking about custom-produced files?

If the latter's the case, maybe we could change the example to receive, via a pipe, the output from a well-known command that produces colored output.

SShrike:
Oh sorry, I did intend to showcase using a pipe but for some reason wrote down like that. I'll update the example in a second.

waldyrious:
Hmm, I'm afraid "something" is awfully vague. In this case I would suggest giving a concrete example to make it clear what kind of "something" we are referring to.

SShrike:
How about `{{command}} | less -R`?

waldyrious:
I'd rather use a specific, concrete command example. I think in this case it'd be more useful than a generic token.

SShrike:
How about `dmesg -L=always | less -R`?

The `-L=always` is in the dmesg command because for some reason when I pipe the output into less colours are automatically disabled.

I'm pretty terrible with coming up with examples.

waldyrious:
That could work. I would provide a suggestion myself if I could come up with one, but I've never used that feature before. My only worry is that dmesg may be a bit of an obscure command for an example. I wonder if we could use something like `ls`, `cat`, `grep` or even `git`, which are more familiar I suppose.

SShrike:
IMO it shouldn't really matter what the example is, it's not a complicated command. I'm under the impression that tldr is intended for quick reference pages, not a hand holding tutorial. 99% of the people browsing them should be able to substitute `{{command}}` *or* `dmesg -L=always` with the command they want to pipe into `less`.

waldyrious:
I don't think hand-holding should be seen as something bad. If we have the opportunity to be clearer without compromising on conciseness, I believe we should do it. We should not assume people know [the same stuff we do](https://xkcd.com/1053/) -- we've all started at the beginning. I'll defer to others' comments, though. /cc @igorshubovych @ostera @felixonmars @rprieto @rubenvereecken

waldyrious:
I just had the need to use just this functionality, where color is actually pretty useful:

    git show 944f724 --color | less -R

Maybe we could use that as a practical example. Thoughts?

waldyrious:
Another useful one: `grep <pattern> <path> --color=always | less -R`

SShrike:
I'm not against using `git diff --color | less -R` as the example.

waldyrious:
Great. If you make that change, I'll merge :)

SShrike:
Just pinging, I've made the change.

waldyrious:
Awesome, thanks!

SShrike:
:tada:

PR #855, opened 2016-04-22
various tweaks to grep

waldyrious:
- use "case-insensitive" in the description of the -i option to make the initialism clear
- use path/to/file instead of file_path
- mention extended regex syntax and expand first occurrence to spell "regular expression"
- clarify wording of the -C example description


ostera:
Looking good!

PR #858, opened 2016-04-27
comm: make the last examples more useful

waldyrious:
Output to file and sort files.


waldyrious:
@igorshubovych @felixonmars @rubenvereecken @ostera @rprieto any comments?

waldyrious:
Well, I'll assume there are no objections. In any case, fixes can always be made afterwards :)

PR #859, opened 2016-04-27
sort: add -h option

waldyrious:
per discussion in #588


waldyrious:
@Ethiraric @igorshubovych @rubenvereecken @rprieto as you were involved in the discussion at #588, how does this look to you?

waldyrious:
Well, if eventually there are any objections, we can always edit the page afterwards, so merging for now 🎉

PR #829, opened 2016-03-21
Add OSX version of sed -- see #281, #293 and #332

waldyrious:

PR #860, opened 2016-04-28
fix: remove a typo in osx sed page

VEBERArnaud:
## Description

Remove in typo in the osx sed page

### DIFF
--- a/pages/osx/sed.md
+++ b/pages/osx/sed.md
@@ -12,7 +12,7 @@
 
 - Replace all occurrences of a string in a file, overwriting the file (i.e. in-place):
 
-`sed -i '' 's/{{find}}/{{replace}}/g' {{filename}}`
+`sed -i 's/{{find}}/{{replace}}/g' {{filename}}`
 
 - Replace only on lines matching the line pattern:
 


waldyrious:
That's not a typo :) see #829 for references.

VEBERArnaud:
Thanks for the references but it's not working for me

```sed -i 's/is/are/g' example.md``` ✅ is working
```sed -i'' 's/is/are/g' example.md``` ✅ is working
but
```sed -i '' 's/is/are/g' example.md``` ❌ is not working (The tldr osx sed page version)


```
❯ sed -i '' 's/is/are/g' example.md
sed: can't read s/is/are/g: No such file or directory
```

**Comment by ostera** [comment]
Hi @VEBERArnaud, are you using GNU sed or BSD sed?

BSD sed comes by default with OS X, and requires `-i ''`. But if you brew your coreutils, then you are using GNU sed, and thus don't require it.

**Comment by VEBERArnaud** [comment]
My bad, I'm using GNU sed 
If BSD sed is the default with OS X, then the actual sed page is fine for me 😌

**Comment by waldyrious** [comment]
If `i''` works in both BSD and GNU sed, maybe we can use that and merge the two examples back into the common directory :) @VEBERArnaud can you test that?

**Comment by VEBERArnaud** [comment]
```/usr/local/bin/sed``` <- GNU
```/usr/bin/sed``` <- BSD

```
❯ /usr/local/bin/sed -i'' 's/is/are/g' example.md

❯ /usr/bin/sed -i'' 's/is/are/g' example.md
sed: 1: "example.md": invalid command code e
```

Nope the ```i''``` is not working with BSD

waldyrious:
Too bad. Thanks for confirming.

PR #861, opened 2016-04-30
Export to PDF

maxbeier:
Added code to export all current Markdown files as PDF with [pandoc](http://pandoc.org/).

To create the PDF file, go to `scripts` and run `./generate-pdf.sh` (It takes a bit to compile; about 30 seconds on my MacBook Air).

### DIFF
--- a/scripts/generate-pdf.sh
+++ b/scripts/generate-pdf.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env sh
+
+src=../pages
+target=../tldr-book.pdf
+template=template.tex
+
+function getContent {
+
+    cd $src
+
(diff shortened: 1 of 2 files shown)


ostera:
@maxbeier, nice and thanks!

I personally think that we shouldn't have any derived artifacts in this repository, thus the `tldr-book.pdf` file shouldn't be part of it – instead it'd be possible to host it in the `tldr-pages.github.io` repo.

maxbeier:
Ja, you're right, I shouldn't have put this in the pull request. Will change this when I know how to 😀

maxbeier:
Bumpy ride, but now it should be correct.

waldyrious:
Thanks @maxbeier!

PR #862, opened 2016-04-30
Update screen.md

Dahie:
sessionname is also the session id, so have the same wording

### DIFF
--- a/pages/common/screen.md
+++ b/pages/common/screen.md
@@ -8,7 +8,7 @@
 
 - Start a new named screen session:
 
-`screen -S {{session name}}`
+`screen -S {{session_name}}`
 
 - Show open screen sessions:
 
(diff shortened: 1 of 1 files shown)


waldyrious:
Maybe the token should be called "session name" rather than just "name", which may be ambiguous. What do you think?

Dahie:
I changed it accordingly. Thanks!

waldyrious:
Please use snake_case for the token names, per the guidelines in CONTRIBUTING.md. Other than that, LGTM :)

waldyrious:
@rubenvereecken should the linter detect non-snake-case tokens and react accordingly?

Dahie:
Snake-Casified it. 8)

waldyrious:
Awesome :) I'll merge directly since this PR has been open for over two weeks with no further comments. Thanks!

PR #864, opened 2016-05-02
cut: use different separator characters

waldyrious:
to exemplify more possibilities

### DIFF
--- a/pages/common/cut.md
+++ b/pages/common/cut.md
@@ -14,14 +14,14 @@
 
 `cut -c{{3-}}`
 
-- Cut out the fifth field, split on the colon characters of each line:
+- Cut out the fifth field of each line, using a colon as a field delimiter (default delimiter is tab):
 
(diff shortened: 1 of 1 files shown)


ostera:
Maybe we could generalize this? `split out in the specified character`? To make it very clear that it's any character that goes.

waldyrious:
Makes sense, I'll change.

waldyrious:
On second thought, after reading the page again, I don't think that would match well with the style of the examples, which uses specific tokens rather than generalized ones. Since this tool does a narrow set of operations, I believe the various examples using different parameters give a better idea of how to use it than a smaller set of generic examples would, since the latter would require the reader to perform some "variable substitution" by head, which would only be justified if we were pressed for space. Does that make sense to you?

ostera:
You're right, we're not pressed by space. We can revisit if/when more is added to the page. Merging! 👍

PR #866, opened 2016-05-04
Add npm install --save-dev 🐖

livioso:
Am I the only one who always forgets if it's --saveDev or --save-dev? My suggestions add --save-dev too. 💩

### DIFF
--- a/pages/common/npm.md
+++ b/pages/common/npm.md
@@ -11,11 +11,11 @@
 
 `npm install`
 
-- Download a given dependency, and add it to the package.json (as `dependencies`):
+- Download a given dependency required for the application to run, and add it to the package.json:
 
 `npm install {{module_name}}@{{version}} --save`
 
-- Download a given dependency, and add it to the package.json (as `devDependencies`):
+- Download a given dependency for development purposes, and add it to the package.json:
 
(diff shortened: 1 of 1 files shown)


waldyrious:
Thanks for the contribution! May I suggest explaining in the command descriptions what's the use-case difference for `--save` vs. `--save-dev`? Happy to brainstorm possible phrasings.

livioso:
For sure. 👍 How about this?

- Download a given dependency required for the application to run, and add it to the package.json:  
 `npm install module_name@version --save`

- Download a given dependency for development purpose, and add it to the package.json:  
`npm install module_name@version --save-dev`

↘️ Inspired by [this](http://stackoverflow.com/questions/22891211/what-is-difference-between-save-and-save-dev).

waldyrious:
That works. I would only change "purpose" to "purposes", but other than that I'd OK that proposal 👍

livioso:
@waldyrious 🐿 Changed the descriptions as discussed.

waldyrious:
LGTM. @igorshubovych?

igorshubovych:
Thanks!
redneckisland-redneck-island-season-5-episode-4-l4KhZIn63RpdUxFUQ

PR #868, opened 2016-05-05
ghc: add page

alx741:
The Glasgow **Haskell compiler**
https://www.haskell.org/ghc/

### DIFF
--- a/pages/common/ghc.md
+++ b/pages/common/ghc.md
@@ -1,0 +1,23 @@
+# ghc
+
+> The Glassgow Haskell Compiler, compiles and links Haskell source files.
+
+- Find and compile all modules in the current directory:
+
+`ghc Main`
+
+- Compile a single file:
+
+`ghc {{file.hs}}`
+
+- Compile using optimization, for faster code:
+
+`ghc -O {{file.hs}}`
+
+- Stop after generating object files:
+
+`ghc -c {{file.hs}}`
+
+- Run in interactive mode (REPL):
+
+`ghci`
--- a/pages/common/ghc.md
+++ b/pages/common/ghc.md
@@ -1,6 +1,7 @@
 # ghc
 
-> The Glassgow Haskell Compiler, compiles and links Haskell source files.
+> The Glasgow Haskell Compiler.
+> Compiles and links Haskell source files.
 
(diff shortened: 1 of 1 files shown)


pages/common/ghc.md
```diff
@@ -0,0 +1,23 @@
+# ghc
+
+> The Glassgow Haskell Compiler, compiles and links Haskell source files.
```
waldyrious: maybe use a period, dash, semicolon or colon (your preference) instead of a comma here.

pages/common/ghc.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Find and compile all modules in the current directory:
+
+`ghc Main`
+
+- Compile a single file:
+
+`ghc {{file.hs}}`
+
+- Compile using optimization, for faster code:
```
waldyrious: I would say the "for faster code" is unnecessary here, and doesn't match the similar entries for other compilers. If we decide to keep this, then the others should be edited to say the same.

pages/common/ghc.md
```diff
@@ -0,0 +1,23 @@
...
+
+`ghc -O {{file.hs}}`
+
+- Stop after generating object files:
```
waldyrious: Stop what? It would be nice to be a tad more descriptive here, for total beginners.

pages/common/ghc.md
```diff
@@ -0,0 +1,23 @@
...
+
+`ghc -c {{file.hs}}`
+
+- Run in interactive mode (REPL):
```
waldyrious: Maybe "run the Haskell interpreter" or something to that effect?

waldyrious:
Another great one, thanks. I've left a few inline comments.

pages/common/ghc.md
alx741: The reason is that the documentation calls it *"the interactive mode"* while the *"Haskell interpreter"* is actually a complete separate program. But I think i could try something else

alx741:
Thank you for the suggestions!
Let me know if there is something else to be improved

waldyrious:
No further issues, LGTM as it is! 👍 from me.

PR #870, opened 2016-05-05
rustc: add page

alx741:
The **Rust Compiler**
https://www.rust-lang.org/

### DIFF
--- a/pages/common/rustc.md
+++ b/pages/common/rustc.md
@@ -1,6 +1,7 @@
 # rustc
 
 > The Rust compiler.
+> Processes, compiles and links Rust language source files.
 
 - Compile a single file:
 
--- a/pages/common/rustc.md
+++ b/pages/common/rustc.md
@@ -14,7 +14,3 @@
 - Compile with debugging information:
 
 `rustc -g {{file.rs}}`
-
-- Build a test harness:
-
-`rustc --test {{file.rs}}`
--- a/pages/common/rustc.md
+++ b/pages/common/rustc.md
@@ -1,0 +1,19 @@
+# rustc
+
+> The Rust compiler.
+
+- Compile a single file:
+
+`rustc {{file.rs}}`
+
+- Compile with high optimization:
+
+`rustc -O {{file.rs}}`
+
+- Compile with debugging information:
+
+`rustc -g {{file.rs}}`
+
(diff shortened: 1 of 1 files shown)


pages/common/rustc.md
```diff
@@ -0,0 +1,19 @@
...
+
+`rustc {{file.rs}}`
+
+- Compile with high optimization:
+
+`rustc -O {{file.rs}}`
+
+- Compile with debugging information:
+
+`rustc -g {{file.rs}}`
```
waldyrious: Does the "g" stand for anything in particular?

pages/common/rustc.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Build a test harness:
```
waldyrious: Any chance we could use a more beginner-friendly description? Maybe a bit more descriptive would help.

waldyrious:
Thanks for the contribution! Please check the inline comments.

pages/common/rustc.md
alx741: Actually *rustc* takes the `-g` parameter as an alias of `-C debuginfo=2`  so the former is kind of a useful layer for ease of use, and there is no expanded option like `--debug` or anything like that. Should I use `-C debuginfo=2`  instead?

waldyrious: No, I don't think the extra complexity is worth it. I suppose one can hang on to the "g" in "debug" for a mnemonic, which should still be more memorable than `-C debuginfo=2`. Thanks for the explanation.

alx741:
This one here has corrections to be reviewed

waldyrious:
The description expansion looks good. Please check my comment about the "test harness" example.

pages/common/rustc.md
alx741: I was digging a bit more about it and the `--test` option is currently very poorly documented and is almost of none practical use for the user because **cargo** manages the testing. So If it is not practical i think it doesn't belong in the *tldr page*.

alx741:
As i said in the inline comments, it shouldn't actually be in the *tldr page* so it's gone now.

waldyrious:
👍, let's just wait for another maintainer to vet the PR, per our contribution policy. Thanks!

PR #871, opened 2016-05-05
Add cloc

sbrl:
Find out more about cloc here: https://github.com/AlDanial/cloc

### DIFF
--- a/pages/common/cloc.md
+++ b/pages/common/cloc.md
@@ -4,12 +4,12 @@
 
 - Count all the lines of code in a directory:
 
-`cloc {{directory}}`
+`cloc {{/path/to/directory}}`
 
 - Count all the lines of code in a directory, displaying a progress bar during the counting process:
 
-`cloc --progress=1 {{directory}}`
+`cloc --progress=1 {{/path/to/directory}}`
 
 - Compare 2 directory sturctures and count the differences between them:
 
-`cloc --diff {{directoryone}} {{directorytwo}}`
+`cloc --diff {{/path/to/directoryone}} {{/path/to/directorytwo}}`
--- a/pages/common/cloc.md
+++ b/pages/common/cloc.md
@@ -12,4 +12,4 @@
 
 - Compare 2 directory sturctures and count the differences between them:
 
-`cloc --diff {{/path/to/directoryone}} {{/path/to/directorytwo}}`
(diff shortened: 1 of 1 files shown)


waldyrious:
LGTM, but do you think the "base" is really necessary in the token paths? I would suggest it to be removed.

pages/common/cloc.md
```diff
@@ -0,0 +1,15 @@
...
+
+> Count, and compute differences of, lines of source code and comments.
+
+- Count all the lines of code in a directory:
+
+`cloc {{/path/to/base/directory}}`
+
+- Count all the lines of code in a directory, displaying a progress bar during the counting process:
+
+`cloc --progress 1 {{/path/to/base/directory}}`
```
waldyrious: what does the 1 mean here?

sbrl:
@waldyrious On second thought, removing the base path would probably be a good idea.

The `1` is the value of the `--progress` flag. It tells `cloc` that it should output it's progress on every file. Setting it to `2` would cause `cloc` to output it's progress every 2 files.

waldyrious:
Shouldn't it be `--progress-rate=N`, per the [documentation](https://github.com/AlDanial/cloc#options-)?

sbrl:
@waldyrious Oh, right! I didn't notice. It still seems to work either way though. I'll update the documentation anyway.

waldyrious:
I think the long version makes it clearer why the 1 is needed :) LGTM 👍

pages/common/cloc.md
```diff
@@ -0,0 +1,15 @@
+# cloc
+
+> Count, and compute differences of, lines of source code and comments.
+
+- Count all the lines of code in a directory:
+
+`cloc {{directory}}`
```
waldyrious: Sorry if I wasn't clear before. I meant we should leave this as just `{{/path/to/directory}}`.

sbrl: Oh right! Is that better?

waldyrious:
Almost there! I'd suggest either .../directory/one or .../first/directory (and the equivalent for the 2nd).

sbrl:
How about that?

waldyrious:
perfect 👍 LGTM

sbrl:
:D

PR #872, opened 2016-05-06
mpc: add page

alx741:
Add **mpc** (mpd client) page.
https://www.musicpd.org/clients/mpc/

### DIFF
--- a/pages/common/mpc.md
+++ b/pages/common/mpc.md
@@ -1,0 +1,28 @@
+# mpc
+
+> Music Player Client.
+> Program for controlling Music Player Daemon (MPD).
+
+- Toggle between play and pause:
+
+`mpc toggle`
+
+- Stop playing:
+
+`mpc stop`
+
+- Show information about currently playing song:
+
+`mpc status`
+
+- Play next song:
+
+`mpc next`
+
+- Play previous song:
+
+`mpc prev`
+
+- Seek playing song:
+
+`mpc [+-]{{seconds}}`
--- a/pages/common/mpc.md
+++ b/pages/common/mpc.md
@@ -1,9 +1,9 @@
 # mpc
 
 > Music Player Client.
-> Program for controlling Music Player Daemon (MPD).
+> Program for controlling the Music Player Daemon (MPD).
 
-- Toggle between play and pause:
+- Toggle play/pause:
 
 `mpc toggle`
 
@@ -11,7 +11,7 @@
 
 `mpc stop`
 
-- Show information about currently playing song:
+- Show information about the currently playing song:
 
 `mpc status`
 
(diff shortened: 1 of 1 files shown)


pages/common/mpc.md
```diff
@@ -0,0 +1,28 @@
+# mpc
+
+> Music Player Client.
+> Program for controlling Music Player Daemon (MPD).
```
waldyrious: This sentence construction looks a bit awkward to me. I understand why it's like this, but wouldn't "the Music Player Daemon" sound more natural? Just asking, I could be persuaded otherwise.

pages/common/mpc.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Toggle between play and pause:
```
waldyrious: I'd say just "Toggle play/pause", but this is just a suggestion -- feel free to ignore it if you prefer the current phrasing.

pages/common/mpc.md
```diff
@@ -0,0 +1,28 @@
...
+
+`mpc toggle`
+
+- Stop playing:
+
+`mpc stop`
+
+- Show information about currently playing song:
```
waldyrious: "about *the* currently playing song"?

pages/common/mpc.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Play next song:
+
+`mpc next`
+
+- Play previous song:
+
+`mpc prev`
+
+- Seek playing song:
```
waldyrious: Seek is a bit of a technical term, albeit the correct one of course. I wonder if "jump forward/backward within the playing song", or something to that effect ("move"?), be a better option. What do you think?

pages/common/mpc.md
alx741: much better! changing it...

pages/common/mpc.md
alx741: I'm not a native english speaker and wasn't sure using *forward* and *backward* would be correct, mainly because I was thinking *"rewind"* :) , definitely using that instead.

alx741: I just found this:
https://en.wikipedia.org/wiki/Media_controls

And it calls it *"rewind"* so maybe it's actually right? Will try it in a new commit...

waldyrious: I thought about "fast-forward" and "rewind", indeed, but I think that could be associated with fast-paced playback in an interactive way, where the user would end the operation once they recognized the part of the song they wanted. But they would work here too, IMO, so I'd say pick whatever you prefer :)

waldyrious:
I see you went with forward/rewind for the last example -- I agree, it's a good choice. LGTM 👍

PR #873, opened 2016-05-06
mdp: add page

alx741:
Mark down presentation (mdp)
https://github.com/visit1985/mdp

### DIFF
--- a/pages/common/mdp.md
+++ b/pages/common/mdp.md
@@ -1,8 +1,8 @@
 # mdp
 
-> A command-line based tool to make presentations from markdown formatted files.
+> A command-line, ncurses based tool to make presentations from markdown files.
 
-- Start a presentation from a markdown file:
+- Launch a presentation from a markdown file:
 
 `mdp {{presentation.md}}`
 
@@ -14,6 +14,6 @@
 
 `mdp --invert {{presentation.md}}`
 
-- Disable transparency in transparent terminal:
+- Disable transparency in transparent terminals:
 
 `mdp --notrans {{presentation.md}}`
--- a/pages/common/mdp.md
+++ b/pages/common/mdp.md
@@ -10,7 +10,7 @@
 
 `mdp --nofade {{presentation.md}}`
 
-- Invert font colors tu use within a light background terminal:
+- Invert font colors to use in terminals with light background:
 
 `mdp --invert {{presentation.md}}`
 
--- a/pages/common/mdp.md
+++ b/pages/common/mdp.md
@@ -1,8 +1,8 @@
 # mdp
 
-> A command-line, ncurses based tool to make presentations from markdown files.
+> A command-line based tool to make presentations from markdown files.
 
-- Launch a presentation from a markdown file:
+- Launch a presentation from a markdown file in the terminal:
 
 `mdp {{presentation.md}}`
 
@@ -10,7 +10,7 @@
 
 `mdp --nofade {{presentation.md}}`
 
-- Invert colors for light background:
+- Invert font colors tu use within a light background terminal:
 
 `mdp --invert {{presentation.md}}`
 
--- a/pages/common/mdp.md
+++ b/pages/common/mdp.md
@@ -1,0 +1,19 @@
+# mdp
+
+> A command-line based tool to make presentations from markdown formatted files.
+
+- Start a presentation from a markdown file:
+
+`mdp {{presentation.md}}`
+
+- Disable fading transitions:
+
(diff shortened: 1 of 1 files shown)


pages/common/mdp.md
```diff
@@ -0,0 +1,19 @@
+# mdp
+
+> A command-line based tool to make presentations from markdown formatted files.
```
waldyrious: Minor nitpick: I believe the correct way to phrase this is "markdown-formatted" (with a hyphen).

pages/common/mdp.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Start a presentation from a markdown file:
```
waldyrious: Does this make the presentation in the terminal? if so, that should be explicit in the description. "Start" is kinda vague.

pages/common/mdp.md
```diff
@@ -0,0 +1,19 @@
...
+
+`mdp {{presentation.md}}`
+
+- Disable fading transitions:
+
+`mdp --nofade {{presentation.md}}`
+
+- Invert colors for light background:
```
waldyrious: This assumes your terminal is set to a dark background, right? Or does it always use a dark background by default regardless of the terminal's color scheme?

pages/common/mdp.md
```diff
@@ -0,0 +1,19 @@
...
+
+`mdp --invert {{presentation.md}}`
+
+- Disable transparency in transparent terminal:
```
waldyrious: I'd write "terminals" (plural) here, but that's just a nitpick.

pages/common/mdp.md
alx741: **mpd** assumes dark background by default, so `--invert` swaps white and black so it can be used with light backgrounds.  What about change it to just *"Use dark font for light backgrounds"* , or something like that?

alx741:
I added **ncurses** in the description and terminal**s**, but is left to see if the `--invert` thing needs more clarification.

pages/common/mdp.md
```diff
@@ -0,0 +1,19 @@
+# mdp
+
+> A command-line, ncurses based tool to make presentations from markdown files.
+
+- Launch a presentation from a markdown file:
```
waldyrious: I don't think "Launch" is much clearer than "Start". The main potential point of confusion here is where/how the presentation is shown. If this happens directly in the terminal window, I think it'd be nice to explicitly say that. Otherwise one might wonder if, e.g., a GUI window appears, or maybe it would open in a browser, or something like that.

pages/common/mdp.md
waldyrious: I would expect such a tool to adapt to the current terminal's settings, so that if I had my terminal set to a light color, the presentations would appear in light background by default and `--invert` would show them in dark background. But if that's not the case, my suggestion would be something along the lines of: "Invert presentation colors to use a light background instead of the default dark background".

waldyrious:
I don't think mentioning ncurses is important here -- that's just the implementation backend, and users need not know this to use the tool or understand what it does. I'd suggest leaving the main description as it was before.

pages/common/mdp.md
alx741: Yeah I hoped ncurses would give that idea, but the description is now a bit technical so... trying something else...

pages/common/mdp.md
alx741: I'm afraid that would give the idea that **mdp** changes the background color (and inverts *every* color, including the background), but it doesn't, it only changes the text color. Maybe removing the *invert* thing from the description? like: "Use dark font color for light background terminals".

waldyrious: Ah, so that's what it does. Your last proposal sounds good, I'd just add a comma after "font color".

alx741:
I tried something else, what do you think?

pages/common/mdp.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Launch a presentation from a markdown file in the terminal:
+
+`mdp {{presentation.md}}`
+
+- Disable fading transitions:
+
+`mdp --nofade {{presentation.md}}`
+
+- Invert font colors tu use within a light background terminal:
```
waldyrious: typo: "tu". Also, may I suggest changing "within a light background terminal" --> "in terminals with light background"?

alx741:
>may I suggest changing "within a light background terminal" --> "in terminals with light background"?

Like that, applied.

waldyrious:
LGTM! 👌

PR #827, opened 2016-03-19
expr, bzip2: add page. Typo fixed in iptables

slash3b:
Two new pages: expr and bzip2.
Fixed type on iptables.md page.

PR #877, opened 2016-05-10
cpio: add page

FlyingJester:


waldyrious:
Thanks! Any chance you could integrate what each option short name stands for into the example descriptions?

FlyingJester:
Yes, I can.

FlyingJester:
Are there any good examples already in the pages of this I can use for reference?

waldyrious:
I believe this is present in most pages (or at least I'd hope so -- I didn't conduct an exhaustive survey) but for example you could look at [sed](https://github.com/tldr-pages/tldr/blob/master/pages/common/sed.md): note how `-r` is matched by "replace" in the description, and how `-i` was integrated into the description to make sure it was explained. Other good examples of discussions about this can be seen [here](https://github.com/tldr-pages/tldr/pull/870#discussion-diff-62154910) and [here](https://github.com/tldr-pages/tldr/pull/827#discussion-diff-56882383). Let me know if that helps.

FlyingJester:
I think my examples kind of already do that, though. `i` is for input, `o` is for output. I could remove or separately explain the `v` for verbose, if that would help. `cpio` doesn't have long options, except for a couple which are kind of unusual to actually use (not for the basic stuff like what is shown in these examples).

I can change the descriptions to explicitly say 'input' and 'output', and make a separate `v` example (or mention that it's for verbose mode).

waldyrious:
> I can change the descriptions to explicitly say 'input' and 'output', and make a separate `v` example (or mention that it's for verbose mode).

Yes, that'd be great!

pages/common/cpio.md
```diff
@@ -0,0 +1,15 @@
...
+
+> Archiving utility.
+
+- Create an archive from files:
+
+`echo "{{file1}} {{file2}} {{file3}}" | cpio -ov > {{archive.cpio}}`
+
+- Create an archive from a directory:
+
+`find {{directory}} | cpio -ov > {{archive.cpio}}`
```
agnivade: {{directory}} => {{path/to/directory}}

agnivade:
Hey @FlyingJester , any updates on this ?

PR #878, opened 2016-05-11
xargs man page - shorter

aicioara:
I wrote that command a while ago. When I read it today, I did not understand what I meant. I think it is now shorter, more clear and inline with the rest of the `tldr` style

### DIFF
--- a/pages/common/xargs.md
+++ b/pages/common/xargs.md
@@ -14,6 +14,6 @@
 
 `find . -name 'M*' | xargs rm`
 
-- Insert arguments at chosen position:
+- Insert arguments at chosen position, using '%' as the placeholder marker:
 
-`{{arguments}} | xargs -I piped_arguments {{command}} piped_arguments {{rest_of_arguments}}`
+`{{arguments}} | xargs -I % {{command}} % {{extra_arguments}}`


waldyrious:
Looks clearer indeed. And learning from our past selves is the best kind of learning 👍 

Just one question: is one meant to type an actual % in the command? I don't understand how this matches with the way the command was written previously.

aicioara:
Yeah, that is why the previous command was so verbose. You can say something like

```bash
ls | xargs -I % -n 1 cp % /tmp
```

or 

```bash
ls | xargs -I ^ -n 1 cp ^ /tmp
```

The symbol is up to the user, but I think giving one is enough for 99% of the cases (which is the purpose of tldr)

aicioara:
That being said, I am biased, since I wrote it and that's how I've been using it my entire life. Let me know if you have suggestions to make it clearer.

waldyrious:
Ah, thanks for the explanation. How about this, then?

    - Insert extra arguments at chosen position, using '%' as the placeholder marker for them:
    `{{arguments}} | xargs -I % {{command}} % {{extra_arguments}}`

By the way, what if the first `{{arguments}}` becomes `{{command}}`? Would it make it clearer?

aicioara:
@waldyrious 

I like the `{{extra_arguments}}` change.

I think `{{command}}` is not sensible, as the rest of the examples say `{{arguments}}` and it will be confused with the already existing `{{command}}`

I think the explanation is too verbose, how about
`Insert extra arguments arbitrarily, using '%' as the placeholder marker`

aicioara:
Edit: I don't like the `{{extra_arguments}}` change. Based on the definition, extra arguments are the `%`, those should be `{{other_arguments}}`

aicioara:
Edit:

`-Insert arguments arbitrarily, using '%' as the placeholder marker`

waldyrious:
Ah, I didn't think the "command" proposal through. Would `{{first_command}}` and `{{second_command}}` be acceptable?

Also, I didn't understand what you meant by "I don't like the `{{extra_arguments}}` change. Based on the definition, extra arguments are the `%`, those should be `{{other_arguments}}`". Care to clarify?

Finally, I agree with the verbosity reduction, but I don't think "at chosen position" --> "arbitrarily" offers a good tradeoff between conciseness and clarity. I'd keep the former (but I agree "as the placeholder marker" is sufficient).

aicioara:
I don't get the `command1` and `command2`

eg
```bash
ls | grep log | xargs -n 1 -I % cp % ~/logs
```

the `~/logs` bit is not a command, it is an argument

waldyrious:
Ok, let's see if I got this. Is the below technically correct? (not an actual proposal for the example, just trying to figure out if I understood this correctly)

`{{command1}} | xargs -I {{command1_output}} {{command2}} {{command1_output}} {{command2_arguments}}`

aicioara:
Somehow. Pedantically it is incorrect, but it is possible that you might have gotten the point.

waldyrious:
Great, we're getting somewhere :) I wonder if we could use `{{command1}} | xargs -I '%' {{command2}} % {{command2_arguments}}`, with the quote marks to make the distinction between the definition of the placeholder marker, and its actual usage. What do you think?

aicioara:
That works, except for `{{command2_arguments}}`. That is misleading, because `%` is also part of command 2 arguments. maybe `{{more_arguments}}`

Final:
```
{{command1}} | xargs -I '%' {{command2}} % {{more_arguments}}
```

waldyrious:
I suppose we should split these two concepts (using the output of one command as arguments to another command, and placing arguments in specific places). Something like this:

```
Insert arguments at chosen position, using '%' as the placeholder marker:
{{arguments}} | xargs -I '%' {{command}} % {{extra_arguments}}

Use the output of one command as arguments to another command:
{{command1}} | xargs {{command2}}
```

aicioara:
I like it. Merge?

waldyrious:
Sure, do you want to make the changes to match my latest comment, or should I merge and change the page myself?

aicioara:
Pushed

waldyrious:
Thanks for the contribution and the patience!

aicioara:
Thanks for working with me. Have a nice week!

PR #879, opened 2016-05-11
odps: add page

onesuper:
odps is the command line tool of a open data processing service.

homepage:
https://www.aliyun.com/product/odps/?lang=en 

yum/apt repo to odpscmd:
http://repo.aliyun.com/odpscmd/

### DIFF
--- a/pages/common/odps.md
+++ b/pages/common/odps.md
@@ -1,0 +1,39 @@
+# odps
+
+> Aliyun ODPS command line tool.
+
+- Start the command line:
+
+`odpscmd --config {{odps_config.ini}}`
+
+- Switch current project:
+
+`use {{project_name}};`
+
+- Show tables in the current project:
+
+`show tables;`
+
+- Show table partitions:
+
+`show partitions {{table_name}};`
+
+- Describe table/partition:
+
+`desc {{table_name}} partition ({{partition_spec}});`
+
+- Create a partitiond table with lifecycle:
+
+`create table {{name}} ({{col}} {{type}}) partitioned by ({{col}} {{type}}) lifecycle {{days}};`
+
+- Read the content of table:
+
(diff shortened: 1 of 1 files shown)


pages/common/odps.md
```diff
@@ -0,0 +1,39 @@
+# odps
+
+> Aliyun ODPS command line tool.
```
waldyrious: It would be nice to expand the ODPS acronym here.

pages/common/odps.md
```diff
@@ -0,0 +1,39 @@
...
+
+- Start the command line:
+
+`odpscmd --config {{odps_config.ini}}`
```
waldyrious: Is there a default config? Ideally we'd start with that, the simplest example, but I see this tool is quite versatile, so we can use --config here, but we should mention it in the command description -- e.g. "Start the command line with a custom configuration file".

waldyrious:
Nice, thanks. but the page is a bit too long. Ideally we'd only showcase the most common 5-6 usages -- which examples do you think could be removed to shorten this page a bit?

waldyrious:
(note that tldr doesn't aim to be comprehensive, just provide a quick overview/reminder)

onesuper:
Yep. This tool is quite complex to use. So I try to write a tl;dr for it. I will reduce the number of example to obey the project's style. : )
But this tool has many concepts and uses a lot of subcommands. Maybe I could add more pages to show the case of sub-commands like git, e.g. odps-resource, odps-table

waldyrious:
The updated page has adequate size. If you feel it doesn't cover the tool in a way that's useful, feel free to create multiple pages for each subcommand. For for now this is a great start. 👍 from me :)

onesuper:
Thanks :-)
Let's merge!

felixonmars:
:+1:

PR #881, opened 2016-05-12
odps-resource: add page

onesuper:
odps resource command(like the concept of subcommand) deals with a set of commands dedicated to resouce management.

Other subcommands like odps-function, odps-instance... will come in other pages.

### Note

e.g. all odps commands can be executed within console OR externally, like:

```
odpscmd -e 'list resources;'
```

### DIFF
--- a/pages/common/odps-resource.md
+++ b/pages/common/odps-resource.md
@@ -1,25 +1,26 @@
 # odps resource
 
-> Manage resources in odps.
+> Manage resources in ODPS (Open Data Processing Service).
 
 - Show resources in the current project:
 
 `list resources;`
 
-- Add local file as odps resource (overwrite if it already exists):
+- Add file or archive:
 
-`add file {{file}} as {{alias}} -f;`
-`add archive {{file}} as {{alias}} -f;`
-`add jar {{file.jar}} -f;`
-`add py {{file.py}} -f;`
+`add file {{file_name}} as {{alias}};`
 
-- Add odps table as odps resource  (overwrite if it already exists):
+- Add archive:
 
-`add table {{table_name}} partition ({{partition_spec}}) as {{as alias}} -f;`
+`add archive {{archive.tar.gz}} as {{alias}}`
 
-- Download odps resource to local file:
+- Add Jar package:
 
-`get resource {{project_name}}:{{resource_name}} {{file}};`
+`add jar {{package.jar}};`
+
+- Add Python script:
+
+`add py {{script.py}};`
 
 - Delete resource:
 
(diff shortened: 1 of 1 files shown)


pages/common/odps-resource.md
```diff
@@ -0,0 +1,26 @@
+# odps resource
+
+> Manage resources in odps.
```
waldyrious: Again, please expand the acronym.

pages/common/odps-resource.md
```diff
@@ -0,0 +1,26 @@
...
+- Show resources in the current project:
+
+`list resources;`
+
+- Add local file as odps resource (overwrite if it already exists):
+
+`add file {{file}} as {{alias}} -f;`
+`add archive {{file}} as {{alias}} -f;`
+`add jar {{file.jar}} -f;`
+`add py {{file.py}} -f;`
```
waldyrious: We typically don't allow multiple commands per example (see #687 and #699). You could either separe this as multiple examples, use a single example of these, or build the command using generic tokens rather than specific parameters.

pages/common/odps-resource.md
```diff
@@ -0,0 +1,26 @@
...
+
+- Add odps table as odps resource  (overwrite if it already exists):
```
waldyrious: Double space before the parenthesis

pages/common/odps-resource.md
```diff
@@ -0,0 +1,26 @@
...
+
+`add table {{table_name}} partition ({{partition_spec}}) as {{as alias}} -f;`
+
+- Download odps resource to local file:
+
+`get resource {{project_name}}:{{resource_name}} {{file}};`
```
waldyrious: Is the semicolon at the end required? what happens if one doesn't include it?

waldyrious:
Thanks :) please check the inline comments.

pages/common/odps-resource.md
onesuper: yep. odps support multi-line command.
If a semicolon is not given,  the console will ask for more input in a newline

waldyrious: Cool, thanks for confirming.

onesuper:
Thanks for suggestions. I've tried to make it simple. :)

waldyrious:
Looks great now, thanks! Just a minor nitpick: I'd use "Java package" rather than "Jar package", since jar is already used in the example itself, and this way "Java" is actually mentioned in this page (just like "Python" in the example below it).

onesuper:
@waldyrious I totally agree that Jar package is redundant.

I choose the terms .jar resource/.py resource, since they are referred in odps-function.

waldyrious:
Looks great now! 👍 to merging.

felixonmars:
:+1: Thanks!

PR #882, opened 2016-05-12
xsel: Add page

SShrike:
I've added a page for xsel, with is essentially the same as xclip, just with a different CLI.

Look over it and tell me what you think, there's some wording I wasn't sure of. I was also wondering if I should document the flags for operating on each of the three selections?

### DIFF
--- a/pages/linux/xsel.md
+++ b/pages/linux/xsel.md
@@ -1,7 +1,7 @@
 # xsel
 
 > Selection manipulation tool.
-> For example xsel can copy STDIN to clipboard or print clipboard to STDOUT.
+> For example, xsel can copy STDIN to clipboard, or print clipboard to STDOUT.
 
 - Copy output to clipboard:
 
--- a/pages/linux/xsel.md
+++ b/pages/linux/xsel.md
@@ -1,0 +1,16 @@
+# xsel
+
+> Selection manipulation tool.
+> For example xsel can copy STDIN to clipboard or print clipboard to STDOUT.
+
+- Copy output to clipboard:
+
+`echo 123 | xclip -bi`
+
+- Paste clipboard:
+
+`xsel -bo > file.txt`
+
+- Clear the clipboard:
+
+`xsel -bc`
--- a/pages/linux/xsel.md
+++ b/pages/linux/xsel.md
@@ -3,13 +3,33 @@
 > Selection manipulation tool.
 > For example, xsel can copy STDIN to clipboard, or print clipboard to STDOUT.
 
-- Copy output to clipboard:
+- Clipboard selection:
+
+`-b or --clipboard`
+
+- Primary selection:
+
+`-p or --primary`
+
+- Secndary selection:
+
+`-s or --secondary`
+
+- Copy output of a command into the clipboard:
 
 `echo 123 | xclip -bi`
 
-- Paste clipboard:
+- Copy contents of a file into the clipboard:
 
-`xsel -bo > file.txt`
+`cat {{file}} | xsel -bo`
+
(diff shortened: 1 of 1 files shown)


pages/linux/xsel.md
```diff
@@ -0,0 +1,16 @@
+# xsel
+
+> Selection manipulation tool.
+> For example xsel can copy STDIN to clipboard or print clipboard to STDOUT.
```
waldyrious: I'd write the second sentence as:

"For example, xsel can copy STDIN to the clipboard, or print clipboard contents to STDOUT."

pages/linux/xsel.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Copy output to clipboard:
```
waldyrious: Perhaps "Copy output of a command to the clipboard".

pages/linux/xsel.md
```diff
@@ -0,0 +1,16 @@
...
+
+`echo 123 | xclip -bi`
+
+- Paste clipboard:
```
waldyrious: Why not "Paste clipboard into a file"? We could have a separate example where it is piped into another command, if you want to convey the generality of this operation.

waldyrious:
Simple and straightforward, I like it :) As for the flags, I would attempt to include the words corresponding to "i", "o" and "b" in the command descriptions ("c" is already included as "clear").

pages/linux/xsel.md
SShrike: How about "Copy STDIN to the clipboard."

pages/linux/xsel.md
SShrike: How about "Print the clipboard to STDOUT"?

I actually might make some more examples to clarify things.

SShrike:
I'll make some edits once I'm home.

SShrike:
I've made an update, I added some more examples and tried to document each selection. I'm not quite sure I'm happy with how I've documented the selections though.

waldyrious:
Unfortunately, I don't think using the description/command format to document the flags is in line with the project's page formatting guidelines. Fortunately, I don't think we need to be as detailed in describing the flags :) we could simply integrate those words into the descriptions of the commands that use them (and try to change the flags across different examples in a pattern that's reasonably discernible). I can give that a try if you'd like.

> I could also update the xclip page in a separate PR, in order to bring them in line with each other as they both do the same thing, so making the pages as similar as possible might be useful.

👍

SShrike:
> I can give that a try if you'd like.

Sorry for taking ages, I forgot about this issue. Sounds good :+1:

waldyrious:
@SShrike I'm on it, will make a proposal soon. Just one clarification: are you sure the example `cat {{file}} | xsel -bo` uses `-bi` and not `-bi`?

SShrike:
Technically it should be `-bi` when reading from STDIN, but It doesn't seem to matter. I'd suggest changing it to `-bi` though.

waldyrious:
Thanks for confirming. Also, is the first example using `xclip` on purpose, or was that an oversight?

SShrike:
Pretty sure that wasn't on purpose, I probably wrote this at 2 AM or something and was half-asleep.

SShrike:
Ping @waldyrious, any updates?

waldyrious:
@SShrike sorry, I've been meanint to look over PRs to this repo but other stuff has been taking priority. I'll finish this now :)

waldyrious:
Unfortunately the notes I had written down are in a different computer I don't have access to at the moment. I'll merge this as-is and will make a separate PR later with proposed changes. Thanks for your patience!

PR #883, opened 2016-05-12
odps-func: add page

onesuper:
odps func command(like the concept of subcommand) deals with a set of commands dedicated to function management.

Other subcommands like odps-instance, odps-auth... will come in other pages.

### Note

e.g. all odps commands can be executed within console OR externally, like:

```
odpscmd -e 'list resources;'
```

### DIFF
--- a/pages/common/odps-func.md
+++ b/pages/common/odps-func.md
@@ -12,7 +12,7 @@
 
 - Create a Python function using a .py resource:
 
-`create function {{func_name}} as {{script.Func}} using 'script.py';`
+`create function {{func_name}} as {{script.Func}} using '{{script.py}}';`
 
 - Delete a function:
 
--- a/pages/common/odps-func.md
+++ b/pages/common/odps-func.md
@@ -6,11 +6,11 @@
 
 `list functions;`
 
-- Create a Java function using a .jar resource:
+- Create a function using .jar:
 
 `create function {{func_name}} as {{path.to.package.Func}} using '{{package.jar}}';`
 
-- Create a Python function using a .py resource:
+- Create a function using .py:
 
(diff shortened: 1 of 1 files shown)


pages/common/odps-func.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Show functions in the current project:
+
+`list functions;`
+
+- Create a Java function using a jar resouce:
+
+`create function {{func_name}} as {{path.to.package.Func}} using '{{package.jar}}';`
+
+- Create a Python function using a py resource:
```
waldyrious: "py resource" sounds awkward to me. Would ".py file" work? Or we could say ".py resource" (and ".jar resource" above, to match).

onesuper: sounds reasonable

waldyrious:
I don't think commit 2cb618c was an improvement. I'd suggest reverting that one.

onesuper:
Revert done.

pages/common/odps-func.md
```diff
@@ -0,0 +1,19 @@
...
+
+`list functions;`
+
+- Create a Java function using a .jar resource:
+
+`create function {{func_name}} as {{path.to.package.Func}} using '{{package.jar}}';`
+
+- Create a Python function using a .py resource:
+
+`create function {{func_name}} as {{script.Func}} using 'script.py';`
```
waldyrious: Sorry I missed this before, but shouldn't `script.py` be wrapped in `{{}}` just like `package.jar` above?

onesuper: @waldyrious thanks!

waldyrious:
LGTM 👍 :)

felixonmars:
:ok_hand: Thanks!

PR #884, opened 2016-05-12
Odps auth tldr

onesuper:
odps auth command(like the concept of subcommand) deals with a set of commands dedicated to authority management.

Other subcommands like odps-function, odps-instance... will come in other pages.

### DIFF
--- a/pages/common/odps-auth.md
+++ b/pages/common/odps-auth.md
@@ -4,15 +4,15 @@
 
 - Show authorities of a user:
 
-`show grants for {{account_name}};`
+`show grants for {{user_name}};`
 
 - Add user to current project:
 
-`add user {{account_name}};`
+`add user {{user_name}};`
 
 - Grant a set of authorities to a user:
 
-`grant {{action_list}} on {{object}} {{object_name}} to user {{account_name}};`
+`grant {{action_list}} on {{object}} {{object_name}} to user {{user_name}};`
 
 - Create a role:
 
--- a/pages/common/odps-auth.md
+++ b/pages/common/odps-auth.md
@@ -12,16 +12,16 @@
 
 - Grant a set of authorities to a user:
 
-`grant {{action_list}} on {{object}} {{object_name}} to user {{user_name}};`
+`grant {{action_list}} on {{object_type}} {{object_name}} to user {{user_name}};`
 
 - Create a user role:
 
 `create role {{role_name}};`
 
-- Grant a user role to a user:
+- Grant a role to a user:
 
 `grant {{role_name}} to {{user_name}};`
 
 - Grant a set of authorities to a user role:
 
-`grant {{action_list}} on {{object}} {{object_name}} to role {{role_name}};`
+`grant {{action_list}} on {{object_type}} {{object_name}} to role {{role_name}};`
--- a/pages/common/odps-auth.md
+++ b/pages/common/odps-auth.md
@@ -14,14 +14,14 @@
 
 `grant {{action_list}} on {{object}} {{object_name}} to user {{user_name}};`
 
-- Create a role:
+- Create a user role:
 
 `create role {{role_name}};`
 
-- Grant a set of authorities to a role:
-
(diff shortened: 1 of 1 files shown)


pages/common/odps-auth.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Grant a set of authorities to a user:
+
+`grant {{action_list}} on {{object}} {{object_name}} to user {{user_name}};`
+
+- Create a role:
+
+`create role {{role_name}};`
+
+- Grant a set of authorities to a role:
```
waldyrious: Perhaps "to a user role" would make this clearer -- let me know if that would be a technically correct way to phrase it.

pages/common/odps-auth.md
```diff
@@ -0,0 +1,27 @@
...
+
+`grant {{action_list}} on {{object}} {{object_name}} to role {{role_name}};`
+
+- Grant a role to a user:
```
waldyrious: Maybe this example should come before the previous one, to introduce the concept of a role.

pages/common/odps-auth.md
```diff
@@ -0,0 +1,27 @@
...
+
+`show grants for {{user_name}};`
+
+- Add user to current project:
+
+`add user {{user_name}};`
+
+- Grant a set of authorities to a user:
+
+`grant {{action_list}} on {{object}} {{object_name}} to user {{user_name}};`
```
waldyrious: what's the difference between "object" and "object_name"?

onesuper: e.g. object is table when object is a table_name.
object is function when object is a function_name.

waldyrious: then maybe `{{object_type}}` / `{{object_name}}`?

pages/common/odps-auth.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Create a user role:
+
+`create role {{role_name}};`
+
+- Grant a user role to a user:
```
waldyrious: Here "user role" is redundant, due to the phrasing. In this case it's ok to use only "role" :)

pages/common/odps-auth.md
onesuper: sounds good.

waldyrious:
LGTM :) 👍

felixonmars:
:ok_hand: Thanks!

waldyrious:
Hey @onesuper, one question:

I noticed that the first and 5th examples use the {{user_name}} token without a "user" prefix:
1) `show grants for {{user_name}};`
5) `grant ... to {{user_name}};`

...while examples 2, 3, 4 and 6 prefix {{user_name}} with "user" and {{role_name}} with "role":
2) `add user {{user_name}};`
3) `grant ... to user {{user_name}};`
4) `create role {{role_name}};`
6) `grant ... to role {{role_name}};`

Can you confirm that the examples 1 and 5 don't require the "user" prefix too?

onesuper:
Hi @waldyrious 

`show grants` command works for both role and user, e.g.

```show grants for {{user_name}}
show grants for {{role_name}}
```

I can confirm that. :)

waldyrious:
Thanks. And what about `grant {{role_name}} to {{user_name}};`? In fact, even if both `grant` and `show grants` work without prefixes, do they work if the prefixes are included? I.e. 

```
show grants for {{user_name}}
show grants for {{role_name}}
grant role {{role_name}} to user {{user_name}}
```

If that works, I wonder if we shouldn't consider including these versions instead, because the consistency would help one to memorize the pattern of the commands.

onesuper:
@waldyrious 
I have checked the manual book.  `show grants for {{role_name}}` does not exist. The command to check the auth of a role should be `desc role {{role_name}}`. Sorry for that. I will submit a merge request to fix that.

And `grant role {{role_name}} to user {{user_name}}` does not work. The standard way is `grant {{role_name}} to {{user_name}}`.

Thanks!

waldyrious:
Huh. I don't know where I got the `show grants for {{role_name}}` example from 😄 I must have miscopied. Sorry for the mess :P

PR #886, opened 2016-05-12
odps-inst: add page

onesuper:
similar to https://github.com/tldr-pages/tldr/pull/881

### DIFF
--- a/pages/common/odps-inst.md
+++ b/pages/common/odps-inst.md
@@ -14,7 +14,7 @@
 
 `status {{instance_id}};`
 
-- Wait on the termination of an instance:
+- Wait on the termination of an instance, printing log and progress information until then:
 
 `wait {{instance_id}};`
 
--- a/pages/common/odps-inst.md
+++ b/pages/common/odps-inst.md
@@ -1,0 +1,23 @@
+# odps inst
+
+> Manage instances in ODPS (Open Data Processing Service).
+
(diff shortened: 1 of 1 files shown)


waldyrious:
Is "logview" a technical term? Or could it be replaced by "log output" or something like that? Also, when is it printed -- after the instance is terminated, or while it's running?

onesuper:
@waldyrious 
Em...logview, which is a url link pointing to the log output (webpage) of an instance, is an term in odps.
I will double-quote it.

waldyrious:
It's ok not to mention it, but if the printing happened after the termination (please confirm), that would make sense as to what the wait command does... without that information, all we know is that it waits for the termination... and then what? It would sound like a no-op.

onesuper:
@waldyrious 
The 'wait' command will block the user input. An url to 'logview' will print. And some progress  information will be printed in the console, like:

```
> wait instance123456;
> logview: http://hostname:port/...
> 10%
> 20%
...
```

The user will retain the control of console until the instance is terminated.

waldyrious:
How about this, then?

`Wait on the termination of an instance, printing log and progress information until then:`

onesuper:
I have changed the description. Thanks!

waldyrious:
LGTM

felixonmars:
:+1: Thanks

PR #889, opened 2016-05-15
Update wget.md

Larry850806:


waldyrious:
Thanks, but don't you mean "number of retries"? Also, what's the default if this option isn't specified -- zero, I suppose?

Larry850806:
Yes, it doesn't retry in default. But `wget -t 0 {{url}}` means retry until finish download. Besides, I have modified the description to "number of retries".

waldyrious:
I asked about the default so the description could be made clearer :) how about "Retry a given number of times if the download doesn't succeed at first"? I' saying this because "set maximum number of retries" may imply that there is an unlimited number of retires by default, and this option just limits them.

Larry850806:
Oh, I see. Sorry for my poor English. I have modified the description.

waldyrious:
LGTM :)

felixonmars:
:+1:

PR #890, opened 2016-05-16
odps-auth: add desc role

onesuper:
Add a `desc role {{user_role}}` to reduce the confusion as mentioned in https://github.com/tldr-pages/tldr/pull/884

### DIFF
--- a/pages/common/odps-auth.md
+++ b/pages/common/odps-auth.md
@@ -1,6 +1,6 @@
 # odps auth
 
-> Manage authorities in ODPS (Open Data Processing Service).
+> User authorities in ODPS (Open Data Processing Service).
 
 - Show authorities of a user:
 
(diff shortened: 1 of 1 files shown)


onesuper:
@waldyrious Now the auth has seven commands. Is that OK?

waldyrious:
This is more structured, I support it even though it bumps the page to 7 example. Looking at the pattern, I would perhaps suggest the "show/describe authorities" examples to come after the "grant authorities" ones, since that would create a mental model of the workflow (create > grant > show) for both users and roles. Does that make sense to you?

onesuper:
Great advice! I have reordered them to create a better mental model. @waldyrious

waldyrious:
Thanks for all these great contributions :) LGTM 👍

onesuper:
Please merge

PR #891, opened 2016-05-16
xargs: move specific example up

waldyrious:
Also moved the concrete example to the end.
I wonder if we should keep it, as it doesn't match the pattern of the remaining ones.

/cc @aicioara

### DIFF
--- a/pages/common/xargs.md
+++ b/pages/common/xargs.md
@@ -10,10 +10,14 @@
 
 `{{arguments_null_terminated}} | xargs -0 {{command}}`
 
-- Delete all files that start with 'M':
-
-`find . -name 'M*' | xargs rm`
-
 - Insert arguments at chosen position, using '%' as the placeholder marker:
 
-`{{arguments}} | xargs -I % {{command}} % {{extra_arguments}}`
+`{{arguments}} | xargs -I '%' {{command}} % {{extra_arguments}}`
+
+- Use the output of one command as arguments to another command:
+
+`{{command1}} | xargs {{command2}}`
+
+- Specific example: delete all files that start with 'M':
+
+`find . -name 'M*' | xargs rm`
--- a/pages/common/xargs.md
+++ b/pages/common/xargs.md
@@ -6,6 +6,10 @@
 
 `{{arguments}} | xargs {{command}}`
 
+- Specific example: delete all files that start with 'M':
+
+`find . -name 'M*' | xargs rm`
+
 - Handle whitespace in arguments:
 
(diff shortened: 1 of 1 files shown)


pages/common/xargs.md
```diff
@@ -10,10 +10,14 @@
...
+- Insert arguments at chosen position, using '%' as the placeholder marker:
 
-`find . -name 'M*' | xargs rm`
+`{{arguments}} | xargs -I '%' {{command}} % {{extra_arguments}}`
 
-- Insert arguments at chosen position, using '%' as the placeholder marker:
+- Use the output of one command as arguments to another command:
 
-`{{arguments}} | xargs -I % {{command}} % {{extra_arguments}}`
+`{{command1}} | xargs {{command2}}`
```
aicioara: Duplication of line 7

waldyrious: Technically, yes, but not conceptually, especially for users unfamiliar with the tool. Would you support swapping all the examples to use command1 and command2 so that it is clearer how `{{arguments}}` come to be?

aicioara: I believe users unfamiliar with the tool should read the man pages. `tldr` in my opinion is for people who already know the tool, but are not using it often enough to be worth remembering it. At least that is how I use it, more like buffed cheatsheet.

Please correct me if I am wrong.

For the other point, I believe having `{{command1}}` instead of `{{arguments}}` is clearer.

waldyrious:
> tldr in my opinion is for people who already know the tool, but are not using it often enough to be worth remembering it. Please correct me if I am wrong.

From README.md: 
> New to the command-line world? Or just a little rusty?

From CONTRIBUTING.md's Guidelines section, item #2:
> When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness.

Notice that if I now pointed out that you should have read the documents above in advance, it would be somewhat rude. But it's ok that you didn't know that. As I see it, me saying that wouldn't be all that different from the position "users unfamiliar with the tool should read the man pages." As [xkcd wisely reminds us](https://xkcd.com/1053/), we are all beginners in *something*. The goal of tldr is to be inclusive and useful to both beginners and experts, and I believe we can remain open to the former without sacrificing utility for the latter.

> For the other point, I believe having {{command1}} instead of {{arguments}} is clearer.

Glad you agree :) I'll send a separate PR then making the conversion, and will leave this one for tweaking the concrete example using `find`and `rm`.

waldyrious:
Merging since there's been no feedback in over a month. In any case, this isn't written in stone -- we can always change it later :)

PR #893, opened 2016-05-17
ufw: Add another example

dbrgn:
There was no example involving a specific port and source ip.

### DIFF
--- a/pages/linux/ufw.md
+++ b/pages/linux/ufw.md
@@ -11,14 +11,18 @@
 
 `ufw disable`
 
+- Show ufw rules:
+
+`ufw status`
+
 - Add ufw allow rule:
 
 `ufw allow {{port}} {{service_name}}`
 
+- Example: Allow access to port 1337 from IP 1.2.3.4 using udp:
+
+`ufw allow from 1.2.3.4 to any port 1337 proto udp`
+
 - Add ufw deny rule:
 
 `ufw deny {{port}} {{service_name}}`
-
-- Show ufw rules:
-
-`ufw status`


waldyrious:
Shouldn't this example come right below the existing `ufw allow` example? I suppose it's just a special case of that one, right?

dbrgn:
It depends. I thought it's better to have the simple examples (allow/deny) first, and then the more complex ones.

waldyrious:
I understand and agree with that general principle, but in this case I feel like conceptually it's easier to present that example as a concrete implementation of the `ufw allow` one. Something like:

```
- Add ufw allow rule:
`ufw allow {{port}} {{service_name}}`
  		  
- Concrete example: allow access to port 1337 from IP 1.2.3.4 using udp:
`ufw allow from 1.2.3.4 to any port 1337 proto udp`
```

Or perhaps it could be made generic like the other ones, and use tokens rather than hardcoded values. What do you think?

dbrgn:
The thing is that `ufw` is very flexible and has a lot of different options. The full grammar would probably be quite complicated, so a few often used examples are better IMO.

But I'll reorder the lines as requested.

waldyrious:
> The thing is that `ufw` is very flexible and has a lot of different options. The full grammar would probably be quite complicated

That was what I feared. To be honest, the concrete example added here would leave me a bit unsure how to adapt it to do something slightly different if I wanted to. If there's a way it could be generalized with tokens, I'd support merging. As it is, and coupled with the fact that it differs in style from all the other examples, I'll remain neutral for now and wait for another maintainer to comment. Thanks for the contribution and discussion so far :)

dbrgn:
Isn't the whole idea of `tldr` that the pages contain a few common usage examples? If someone wants to see the whole syntax, you can always just type `man ufw`.

waldyrious:
You're correct, but note that the tldr project is more of an explainer than a cheatsheet -- in other words, the underlying mission is not to merely provide copy-pastable snippets (like, say, commandlinefu.com) but rather to allow readers to *understand* the language of the commands, so that the logic behind the options is evident and they can then derive new commands from the provided examples, or compose several simple examples into more complex ones that are useful to them. This is basically the same purpose of man pages, but executed in a more readable way (so we believe), .i.e. through a series of examples.

As I'm not a user of ufw myself, I can't judge properly whether the new example is common enough or generalizable enough to warrant inclusion above the other possible concrete examples we could include (and we're space-limited, so there's that factor too). That said, I'm not voting against the merging of this PR or anything -- I just want another set of eyes (hopefully more experienced than mine) to take a look.

dbrgn:
@Torild created the ufw page. What do you think?

agnivade:
Hey @waldyrious, I've had minor interactions with ufw in my previous company. Although I'm no ops guy, but I know tinkering with allow/deny rules and checking status are the most common usages of this command. 

I can volunteer to make this page better. For example, I noticed that the allow and deny rules are incorrect. Because you cannot put port and service name together. Its an either/or thing.

waldyrious:
@agnivade please do :)

PR #895, opened 2016-05-19
xattr: add page

vpilot:
Added a new page for osx xattr utility which deals with filesystem extended attributes.

http://xattr.sourceforge.net/
http://gotofritz.net/blog/geekery/os-x-extended-attibutes/
https://www.safaribooksonline.com/library/view/macintosh-terminal-pocket/9781449328962/re38.html

### DIFF
--- a/pages/osx/xattr.md
+++ b/pages/osx/xattr.md
@@ -1,0 +1,23 @@
+# xattr
+
+> Utility for use with extended filesystem attributes.
+
+- List key:value extended attributes for a given file:
+
+`xattr -l {{file}}`
+
+- Write an attribute for a given file:
+
+`xattr -w {{attribute_key}} {{attribute_value}} {{file}}`
+
+- Delete an attribute from a given file:
+
+`xattr -d {{attribute_key}} {{file}}`
+
+- Delete all extended attributes from a given file:
+
+`xattr -c {{file}}`
+
+- Recursively delete an attribute in a given directory:
+
+`xattr -rd {{attribute_key}} {{directory}}`


waldyrious:
Very nicely written, thanks!

waldyrious:
@vpilot are there any guidelines or standards for what attributes can be used, which are recognized by existing applications, etc?

[Wikipedia](https://en.wikipedia.org/wiki/Extended_file_attributes#OS_X) mentions:

> the name is a null-terminated Unicode string

so it appears that any string is acceptable to be used as a key; but then it mentions

> user space extended attributes

which suggests that there are certain predefined classes of attributes.

Indeed, the [linux manpage for xattr](http://man7.org/linux/man-pages/man7/xattr.7.html) says:

> The namespace mechanism is used to define different classes of extended attributes. (...)
> Currently, the security, system, trusted, and user extended attribute classes are defined as described below. Additional classes may be added in the future.

Does this notion exist in OS X as well?

vpilot:
@waldyrious Apple is doing a good job by making hard to find the list of predefined attributes. Some of the known to me are:

'com.apple.quarantine'
'com.apple.FinderInfo'
'com.apple.metadata'

Apple allows both string and binary data as values for defined namespaces. There is no mention of any namespace convention. From the man page:

> Extended attributes are arbitrary metadata stored with a file,
     but separate from the filesystem attributes (such as modification time or
     file size).  The metadata is often a null-terminated UTF-8 string, but
     can also be arbitrary binary data.

Let me know if you have more luck with finding the list of predefined attributes.

waldyrious:
I tried searching some more but didn't find anything useful. Hopefully someone else can chime in if they come across this discussion. Thanks for the quick response anyway!

PR #896, opened 2016-05-20
gcc: add -c option

liuderchi:
writing `Makefile` often involves `-c` option to create object files from source code.

### DIFF
--- a/pages/common/passwd.md
+++ b/pages/common/passwd.md
@@ -10,7 +10,7 @@
 
 `passwd {{username}} {{new password}}`
 
-- Get the current statuts of the user:
+- Get the current status of the user:
 
 `passwd -S`
 


liuderchi:
I just came up with another explanation: Compile source code into object file without linking.

ostera:
@liuderchi can you check your git config? make sure your name and email are properly set so your commit will be attributed to your github user.

Otherwise looks good to me!

liuderchi:
I did set my git config WRONG. Thanks for the notice.

1. Should I send another Pull Request to fix it? (turn this PR off?)
2. I want to discuss about the final description:

`- Compile source code without linking:`
or
`- Compile source code into object file without linking:`

which one is better?

waldyrious:
For future reference, this PR was followed-up by #897 (by the same author).

PR #899, opened 2016-05-23
Create halt.md

dylanrees:


ostera:
This actually doesn't seem to work on OS X.

According to `man halt`:

```
REBOOT(8)                 BSD System Manager's Manual                REBOOT(8)

NAME
     halt, reboot -- stopping and restarting the system

SYNOPSIS
     halt [-lnqu]
     reboot [-lnq]

DESCRIPTION
     The halt and reboot utilities flush the file system cache to disk, send all running processes a SIGTERM (and subsequently a SIGKILL) and, respectively, halt or restart the system.  The action is logged, including entering a shutdown record
     into the wtmp(5) file.
```

dylanrees:
Hmm...  so it works on BSD but not OSX?  Should I re-submit this under the Linux folder instead of common?

ostera:
Or force-push to this branch after fixing the commit :)

PR #902, opened 2016-05-27
Create htop.md

dylanrees:


waldyrious:
Nice one, thanks! But shouldn't the command descriptions say "htop" instead of "top"?

/cc @hishamhm

dylanrees:
Oops, you're right!  Just patched it.

waldyrious:
Looks good :)

PR #907, opened 2016-06-05
Add examples for exporting GPG keys

robyoung:
This chage adds examples for exporting public and private GPG keys.

This is particularly useful for backing up keys. In both cases they have
been output as ASCII armored text with the `--armor` flag to make the
output easier to work with.


waldyrious:
Great contribution, thanks!

PR #911, opened 2016-06-14
bg: add page

liuderchi:
There has been info for `fg` and I think there should also be one for `bg`.


waldyrious:
Nice one, thanks! :+1:

PR #912, opened 2016-06-14
screen: Start a new deamon and log the output

beshrns:


waldyrious:
Thank you for the contribution!

waldyrious:
By the way @beshrns, what does the "m" stand for in the options for this example?

PR #913, opened 2016-06-17
update grep.md

gwanii:
1.grep -A
2.grep -B
3.grep -l

### DIFF
--- a/pages/common/grep.md
+++ b/pages/common/grep.md
@@ -19,14 +19,6 @@
 
 `grep -e {{^regex$}} {{path/to/file}}`
 
-- Print 3 lines of trailing context after each match
-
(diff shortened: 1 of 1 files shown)


ostera:
If you take a look at the [build](https://travis-ci.org/tldr-pages/tldr/builds/138260159) you'll see there's some syntax errors :)

waldyrious:
We had discussed before about including -A and -B, and agreed that given grep's long list of options, it would be preferable to only mention -C (which is both -A and -B combined, and as a bonus stands for "context" as a mnemonic). Please remove the -A and -B examples (which incidentally are the lines generating syntax errors in Travis) and we can merge this PR :)

waldyrious:
Thanks for the quick reaction ;)

PR #920, opened 2016-06-21
equery: add page

fuerbringer:
Messed up the last PR :cry: Anyways, this one has better and more consistent descriptions.

### DIFF
--- a/pages/linux/equery.md
+++ b/pages/linux/equery.md
@@ -1,0 +1,23 @@
+# equery
+
+> View information about Portage packages.
+
+- List all installed packages:
+
+`equery list '*'`
+
+- Search for installed packages in the Portage tree and in overlays:
+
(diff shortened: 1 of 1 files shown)


waldyrious:
Looks great! Just one question: what are overlays?

fuerbringer:
Gentoo overlays are additional repositories and packages not directly provided by the Gentoo developers. Once a package from an overlay is installed, it appears as an installed package just like any other applications from the main portage tree. It's somewhat comparable to the Arch User Repository (AUR).

waldyrious:
Ah, understood. Would it be clearer to say "Search for installed packages in the Portage tree and in overlays", then? If so, I'll merge and make the change directly, to save you the trouble :)

waldyrious:
Thanks :+1:

PR #921, opened 2016-06-21
top: Add a better example

agnivade:
I feel the `-p` option is much more used than `-i`. Me personally, always use top to monitor a set of nginx/node processes.

### DIFF
--- a/pages/linux/top.md
+++ b/pages/linux/top.md
@@ -6,11 +6,11 @@
 
 `top`
 
-- Start top ignoring any idle or zombie processes:
+- Filter processes by PID. It takes a comma separated list. (Normally you wouldn't know PIDs off hand. This example picks the PIDs from the process name):
 
-`top -i`
+`top -p $(pgrep -d ',' {{nginx}})`
 
-- Start top displaying only processes owned by given user:
+- Filter processes owned by given user:
 
 `top -u {{user-name}}`
 
--- a/pages/linux/top.md
+++ b/pages/linux/top.md
@@ -6,13 +6,17 @@
 
 `top`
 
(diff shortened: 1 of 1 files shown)


waldyrious:
"Filter" is probably not a good word to use here, as it can be understood as "filter out". Can you think of an alternative phrasing that would be less prone to misunderstanding?

As for replacing -i with -p, I'll wait for others to chime in. I'm not a frequent user of the top command, but from the descriptions I would say the -i example is more useful (and simpler).

agnivade:
I see. How about "list" processes ? Or "match" processes ? I am even okay with "display" processes. I just didn't want the "start top" part, as its obvious.

About replacing `-i` with `-p`, I use top quite frequently and I have never used the `-i` option. Sure, let's see what others have to say. Frankly, top is rarely used with any flags. Most of the stuff happens after the command is fired. I agree that the description of the `-i` option looks simpler. But I think, we should provide examples which are more relevant to daily use and which users are more likely to look up rather than an option having simpler description. Maybe we can have both `-i` and `-p` ?

waldyrious:
"List" works very well indeed! We could use that. 👍 

As for `-i` vs. `-p`, my main argument wasn't that the description is simpler (that's a bonus), but indeed it seemed to me that what the command does would be more useful than what -p does. I mean, the few times I've used top or top-like tools (say, nethogs), the existence of idle entries in the list has been a nuisance that I'd rather get rid of, while showing top for one or a handful of processes seems a narrower use case.

But hey, as the little girl says, why not both? The page isn't overly long yet, so we can certainly add the `-p` example without removing the `-i` one :)

agnivade:
Done !

waldyrious:
Thanks :)

PR #922, opened 2016-06-22
ifup: add page

agnivade:


waldyrious:
"bring up interfaces" is a very vague description. Could we use something more specific / layman-friendly?

agnivade:
How about now ?

waldyrious:
Better, but "allowing it to transfer data" is redundant. If we decide to keep it, it should be changed to read "them" rather than "it".

Also, please reword "bring up" --> "enable" in the command example descriptions, too. For example: "Enable the eth0 interface".

One more thing: would it be a better idea to put quotes around "auto"? I feel that makes the sentence clearer.

agnivade:
Agree with all of them. Pushed.

I will follow the exact same format for `ifdown` also. Do you think we need to cover `ifquery` too ?

waldyrious:
Great. As for `ifquery`, I'd say the more commands the merrier :)

PR #928, opened 2016-06-25
simplify xsel.md, add example using X11 selection

waldyrious:
Also add equivalence to standard operations, to make it clearer what each example does.

### DIFF
--- a/pages/linux/xsel.md
+++ b/pages/linux/xsel.md
@@ -22,6 +22,6 @@
 
 `xsel -cb`
 
-- Output the X11 primary selection's contents into the terminal (equivalent to a mouse middle-click)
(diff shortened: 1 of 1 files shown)


waldyrious:
@ostera do you have thoughts about using the new squash-and-merge workflow that github provides? I've been using it lately and like it a lot. The only case where I preserve the separate commits is if they were deliberately made atomically to distinguish separate changes.

ostera:
@waldyrious I haven't seen the button yet, might have not been AB tested in my area, but i will absolutely give it a try when it's available here!

waldyrious:
that's weird, I was under the impression that they had [already rolled the feature out to everyone](https://github.com/blog/2141-squash-your-commits). It's the same button as the "merge" one, just click the dropdown and choose the squash option. It'll then look like this:

PR #929, opened 2016-06-25
at: add page

slash3b:
New page for 'at' command

### DIFF
--- a/pages/linux/at.md
+++ b/pages/linux/at.md
@@ -1,0 +1,15 @@
+# at
+
+> Executes commands at a specified time.
+
+- Open an `at` prompt to create a new set of scheduled commands, press Ctrl+D to save and exit:
+
+`at {{hh:mm:ss}}`
+
+- Execute the commands and email the result using a local mailing program such as sendmail:
+
+`at {{hh:mm:ss}} -m`
+
+- Execute a script at the given time:
+
+`at {{hh:mm:ss}} -f {{path/to/file}}`


pages/linux/at.md
```diff
@@ -0,0 +1,11 @@
+# at
+
+> Executes commands at a specified time.
+
+- Open 'at' prompt to execute commands at specific time. When you done press Ctrl+D to register job(s):
```
waldyrious: Suggestion: "Open \`at` prompt to type commands to be executed at the given time"

"When you done" --> "When you're done", and maybe also add a comma here, to make the reading flow more explicit.

Also, it's slightly unclear what "register job(s)" mean, since the terms aren't used before. How about "save the scheduling"?

pages/linux/at.md
```diff
@@ -0,0 +1,11 @@
...
+
+`at {{hh:mm:ss}}`
+
+- Execute file (-f) and mail result(-m):
```
waldyrious: I wouldn't include the -m option in this example. We have the space to include it as a separate example.

Also, what is file meant to be? An executable script? A list of commands?

pages/linux/at.md
```diff
@@ -0,0 +1,11 @@
...
+
+`at {{hh:mm:ss}} -m -f {{/path/to/the/file}}`
```
waldyrious: The convention for file tokens has been to use `{{path/to/file}}`.

slash3b: fixed

pages/linux/at.md
slash3b: thank you, I've made changes.

pages/linux/at.md
slash3b: Will add separate example for mailing option.

I meant to say that the file should be executable script.

ostera:
@slash3b don't forget to push the fixes!

slash3b:
I've pushed updated version, please check it.

pages/linux/at.md
```diff
@@ -0,0 +1,15 @@
...
+
+`at {{hh:mm:ss}}`
+
+- Execute commands and email result:
+
+`at {{hh:mm:ss}} -m`
+
+- Execute script:
+
+`at {{hh:mm:ss}} -f {{/path/to/file}}`
```
waldyrious: please remove the leading / in the path, since it doesn't have to be an absolute path (this is the convention used in other pages)

pages/linux/at.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Open `at` prompt to execute commands at specific time. When you're done, press Ctrl+D to save the schedule:
+
+`at {{hh:mm:ss}}`
+
+- Execute commands and email result:
+
+`at {{hh:mm:ss}} -m`
+
+- Execute script:
```
waldyrious: Just to be explicit, how about "Execute a script at the given time"?

pages/linux/at.md
```diff
@@ -0,0 +1,15 @@
+# at
+
+> Executes commands at a specified time.
+
+- Open `at` prompt to execute commands at specific time. When you're done, press Ctrl+D to save the schedule:
+
+`at {{hh:mm:ss}}`
+
+- Execute commands and email result:
```
waldyrious: It's not clear how the email feature works. Shouldn't an email address have to be passed in this command? Is it configured elsewhere?

pages/linux/at.md
```diff
@@ -0,0 +1,15 @@
+# at
+
+> Executes commands at a specified time.
+
+- Open `at` prompt to execute commands at specific time. When you're done, press Ctrl+D to save the schedule:
```
waldyrious: I still think "to insert commands to be executed" is clearer/more accurate than "to execute commands". @ostera any thoughts?

pages/linux/at.md
slash3b: oh, sorry, haven't thought about that. It is fixed.

pages/linux/at.md
slash3b: yeah, this is much better explanation.

pages/linux/at.md
slash3b: honestly I do not have much experience with MTA, 
but I found this useful answer [`at -m` explanation](http://unix.stackexchange.com/a/82845)

waldyrious: How about "Execute the commands and email the result using a local Mail Transfer Agent (MTA) such as `sendmail`"?

slash3b: thank you, I just cut "such as `sendmail`" to have a little bit shorter description.

pages/linux/at.md
slash3b: I slightly changed your description to "and type commands to be executed", how about that ?

waldyrious: It kinda works, but deviates slightly from the structure of other command descriptions; the initial verb (open) refers to the command, while the second (type) refers to the user, so the sentence technically has two subjects. But that's a minor point, so if you really prefer it like that, I won't object :) let me know what you think.

pages/linux/at.md
waldyrious: Iw we want to shorten the description, I'm wondering if we should cut out the expanded acronym instead, since I suspect more people will have heard of sendmail than of "Mail Transfer Agent". I.e. "using a local MTA such as sendmail". I'm not 100% sure this is better though.... perhaps drop the MTA altogether and say simply "using a local mailing program such as sendmail"?

pages/linux/at.md
ostera: I'd say

```markdown
- Open an `at` prompt to create a new schedule of commands; press Ctrl+D to save and exit:
```

Note the `an`. It's easy to read the `at` as part of the sentence, making it necessary to re-read until it clicks in your head that it's not.

I also think we can rely on the word `schedule` to signify that the commands will be specified `at a specific time`.

waldyrious: I like your proposal @ostera 👍 maybe I'd tweak it to say "create a new set of scheduled commands" or something like that.

pages/linux/at.md
slash3b: ok, last one is quite good, short and simple, thank you.

waldyrious: glad you like it :)

pages/linux/at.md
slash3b: I really like the combination -  "Open an `at` prompt to create a new set of scheduled commands, press Ctrl+D to save and exit:"
thank you guys for help, will push final version with all fixes.

waldyrious:
Looks great! 👍 LGTM. @ostera do the honors :)

slash3b:
looks great indeed, thanks @ostera  and @waldyrious  for help.

ostera:
👍 woo 🎆

PR #930, opened 2016-06-26
ngrep: add page

zhangda7:
Ngrep is an easy to use basic packet sniffing tool.
HomePage is http://ngrep.sourceforge.net/usage.html.


pages/common/ngrep.md
```diff
@@ -0,0 +1,15 @@
+# ngrep
+
+> Ngrep is a pcap-aware tool that will allow you to specify extended regular or hexadecimal expressions to match against data payloads of packets.
```
waldyrious: This description is too technical and jargon-y. Something closer to the description you used in the PR ("an easy-to-use basic packet sniffing tool.") would match the ethos of the tldr project better.

For example: is PCAP a critical qualifier to describe what this command does? It sounds like a technical feature that could be left to a more detailed description (e.g. in the manpage).

pages/common/ngrep.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Monitor all activity crossing source or destination port on network interface dev:
```
waldyrious: Perhaps "source or destination" could be removed, making the sentence easier to read without losing much information?

pages/common/ngrep.md
```diff
@@ -0,0 +1,15 @@
...
+
+`ngrep -d {{dev}} port {{port}}`
+
+- Monitor all activity crossing source or destination port on any network interface:
```
waldyrious: How about placing this example before the previous one? The convention is to go from simpler to more complex/detailed examples.

pages/common/ngrep.md
```diff
@@ -0,0 +1,15 @@
...
+
+`ngrep -d any port {{port}}`
+
+- Monitor all activity crossing source or destination port, show timestamp and format data in hex format:
+
+`ngrep -t -x -d any port {{port}}`
```
waldyrious: It's unclear what options match each part of the description. Maybe this command could be split in two examples?

waldyrious:
Thanks for the contribution! Please check the inline comments.

pages/common/ngrep.md
zhangda7: Yes, this description is too technical and not easy to read.

pages/common/ngrep.md
zhangda7: Agree.

pages/common/ngrep.md
zhangda7: Agree.

pages/common/ngrep.md
zhangda7: Yes, split it in two examples is more clear. 
Thanks for your suggested comments, I will reorganize it all.

pages/common/ngrep.md
```diff
@@ -0,0 +1,23 @@
+# ngrep
+
+> Grep traffic on a network.
+
+- Capture traffic of all interface:
```
agnivade: "interface" -> "interfaces"

pages/common/ngrep.md
```diff
@@ -0,0 +1,23 @@
...
+
+`ngrep -d any`
+
+- Capture traffic of a specific interface {{dev}}:
```
agnivade: I think it might be better to change `{{dev}}` to a concrete example like `eth0`. Users might be confused to other devices like '/dev/sda1' etc.

pages/common/ngrep.md
```diff
@@ -0,0 +1,23 @@
+# ngrep
+
+> Grep traffic on a network.
```
waldyrious: Instead of "Grep", which is a specialized/jargon term, I'd suggest something like "Filter network traffic packets using regular expressions." (or "Search" instead of "Filter", if you prefer)

pages/common/ngrep.md
zhangda7: Thx, i missed it.

pages/common/ngrep.md
zhangda7: ok, I will change them to specific value.

pages/common/ngrep.md
zhangda7: I use grep because ngrep is like "grep command used in network", but it will confuse for users who have not used grep. Your comment is more meaningful.

pages/common/ngrep.md
```diff
@@ -0,0 +1,23 @@
+# ngrep
+
+> Filter network traffic packets using regular expressions.
+
+- Capture traffic of all interfaces:
+
+`ngrep -d any`
```
agnivade: Envelop "any" with {{ and }}. Same with all other examples.

agnivade:
Please wrap the token words with {{ }}. Ref-  https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#token-syntax

waldyrious:
Just to clarify, @agnivade is talking about "eth0", "22", "www.example.com" (i.e. user-provided values) -- these should all be wrapped in `{{}}`.

pages/common/ngrep.md
zhangda7: I must misunderstood the token syntax, I will wrap "eth0", "22" with {{ }}, but "any" is a keywork in ngrep, thx.

waldyrious:
LGTM! 👍

PR #933, opened 2016-06-27
ifdown: add page

agnivade:


pages/linux/ifdown.md
```diff
@@ -0,0 +1,11 @@
+# ifdown
+
+> Tool used to disable network interfaces.
```
waldyrious: "Tool used to" is redundant, we can remove that :)

waldyrious:
Looks good overall, I just left minor comments inline.

pages/linux/ifdown.md
agnivade: Sure, will remove "tool used to".

I actually disagree on the "disable" part. This is exactly what it means and I feel this is easy enough for a layman to understand. If for `ifup`, we can use "enable", why not "disable" for `ifdown` ? "Turn off" is misleading because we are not exactly turning off anything. I would be still okay with "deactivate" but in a technical context, "disable" is the proper term. 

Also, for `ifdown -a`, the explanation is `disable all interfaces which are enabled`. If we replace "disable" with something else, the "enable" part doesn't fit nicely.

waldyrious: Yes, you make a good point -- I actually had the same reservations when I wrote my suggestion. Let's keep "disable" then.

Issue #785, opened 2016-02-22
Add "see also" section to tldr pages


oxguy3:
Very often, it seems like it'd be very likely that someone viewing page A might also find page B very useful. It could be very helpful if tldr pages were allowed to have a "See also" section at the bottom. For example, related or similar tools could link to each other (e.g. grep/ack, chgrp/chown, more/less, rm/shred, head/tail, etc). Another use case: sub commands and parent commands could all link to each other (e.g. git-whatever tools could link to git, and the `git` page could link to _all_ of them. ditto for csvkit and all of its tools).

A very simple implementation might look like:

```
# foo

> Most awesome tool ever.
...

## See also
- bar: another awesome tool
- baz: legacy version of foo
- qux: for reverting the effects of foo
```

It shouldn't be too hard for clients to implement this (command-line clients could probably get away with printing the raw markdown, more or less, and clients that already rely on a proper Markdown library won't need to do anything).

---

Issue #784 (which I've just submitted) goes hand-in-hand with this issue. If #784 and this issue were both implemented, then the See also section would definitely make use of the link syntax, and probably look more like this:

```
## See also
- [bar]: another awesome tool
- [common/baz]: cross-platform version of foo
- [qux]: for reverting the effects of foo
```

igorshubovych:
Just wonder, have you seen this?
https://github.com/tldr-pages/tldr-node-client/pull/81

oxguy3:
Dang, I had not seen that. Huh, well guess this is a bit redundant. >.<

igorshubovych:
Which one is redundant? Your variant or mine?

oxguy3:
Mine is redundant I guess, since yours was way earlier. That said, I do think it'd be nice if the see also list included descriptions of _why_ that particular tool was being linked. And also, being able to link to specific OS versions would be nice, like I showed below the horizontal rule in my original post.

waldyrious:
This is actually a duplicate of #481. I'll close this one and reopen the previous one :)

PR #934, opened 2016-06-28
Create git-imerge.md

waldyrious:
https://github.com/mhagger/git-imerge

/cc @mhagger

### DIFF
--- a/pages/common/git-imerge.md
+++ b/pages/common/git-imerge.md
@@ -5,7 +5,7 @@
 
 - Start imerge-based rebase (checkout the branch to be rebased, first):
 
-`git imerge rebase {{branch_to_rebase_into}}`
+`git imerge rebase {{branch_to_rebase_onto}}`
 
 - Start imerge-based merge (checkout the branch to merge into, first):
 
--- a/pages/common/git-imerge.md
+++ b/pages/common/git-imerge.md
@@ -3,19 +3,19 @@
 > Perform a merge or rebase between two git branches incrementally.
 > Conflicts between branches are tracked down to pairs of individual commits, to simplify conflict resolution.
 
-- Start imerge-based rebase (checkout the branch to rebase first):
+- Start imerge-based rebase (checkout the branch to be rebased, first):
 
-`git imerge rebase {{target_root_branch}}`
+`git imerge rebase {{branch_to_rebase_into}}`
 
 - Start imerge-based merge (checkout the branch to merge into, first):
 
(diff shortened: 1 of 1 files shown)


pages/common/git-imerge.md
```diff
@@ -0,0 +1,24 @@
+# git-imerge
+
+> Perform a merge or rebase between two git branches incrementally.
+> Conflicts between branches are tracked down to pairs of individual commits, to simplify conflict resolution.
+
+- Start imerge-based rebase (checkout the branch to be rebased, first):
+
+`git imerge rebase {{branch_to_rebase_into}}`
```
mhagger: I suggest changing the argument name to `branch_to_rebase_onto`, since your branch will end up on top of the other branch; the other branch is not changed.

waldyrious: Sure! as a non-native English speaker, these nuances sometimes elude me :) thanks for the review!

pages/common/git-imerge.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Start imerge-based merge (checkout the branch to merge into, first):
```
waldyrious: @mhagger should this also be changed to "onto"?

mhagger:
I didn't know about this tldr thing, but it seems useful.

I think you've captured the most important `git imerge` commands with good explanations.

Do tldr pages typically mention where commands can be installed from, or include cross-references to the main documentation? `git imerge` is not part of standard Git, so I would expect that to be a frequent question.

But anyway, :+1: thanks for this :sparkles:

pages/common/git-imerge.md
mhagger: No, in this case the branch you are on will be modified by the merge, so "into" is more accurate here.

waldyrious: Got it. I'll update the PR.

waldyrious:
> Do tldr pages typically mention where commands can be installed from, or include cross-references to the main documentation? `git imerge` is not part of standard Git, so I would expect that to be a frequent question.

Yeah, I thought about that too, but it hasn't been the case so far. I suppose the expectation is that once someone tldr's a command, it's because they already know about it, which seems fair. Still, a future "references", "links" or "notes" section is certainly something we should consider. c.f #785

waldyrious:
Given the blessing of the tool's author himself, I'll self-merge :)

ostera:
👍

PR #935, opened 2016-06-29
Adds command to show SSH key fingerprint

robyoung:
Adds an ssh-keygen command to show the fingerprint of an SSH key in the same format that is used by Github to list SSH keys that it knows about.

### DIFF
--- a/pages/common/ssh-keygen.md
+++ b/pages/common/ssh-keygen.md
@@ -22,6 +22,10 @@
 
 `ssh-keygen -l -F {{remote_host}}`
 
+- Retrieve the fingerprint of a key in MD5 Hex:
+
+`ssh-keygen -l -E md5 -f ~/.ssh/{{filename}}`
+
 - Change the password of a key:
 
 `ssh-keygen -p -f ~/.ssh/{{filename}}`


pages/common/ssh-keygen.md
```diff
@@ -22,6 +22,10 @@
 
 `ssh-keygen -l -F {{remote_host}}`
 
+- Retrieve the fingerprint of a key in MD5 Hex (useful for checking a key against the list in Github):
```
ostera: This looks very long. I'd drop the comment in parens.

waldyrious: The page itself is getting pretty long too, btw. Should we consider dropping one of the current examples to add this one (if we agree the tradeoff is worth it)?

ostera: I think this would be the last one before starting to swap new samples with more useful ones. We can deal with that separately as additions come.

waldyrious: Fair enough.

robyoung:
Sorry about the delay, I've updated the commit.

ostera:
Perfect, thanks!

PR #938, opened 2016-07-03
Update URL of Web Client

ostera:
Hey guys, I've rebuilt the web client in a nicer way, with nicer tools and a better design.

I've also hosted it elsewhere so it can use HTML5 PushState (now url's look like `tldr.ostera.io/linux/du` instead of the scrambled mess it was before).

Would be great to have it merged ASAP so I can get rid of the old one :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 
 A collection of simplified and community-driven man pages.
 
-Install it with `npm install -g tldr` or [try the web client](https://ostera.github.io/tldr.jsx).
+Install it with `npm install -g tldr` or [try the web client](http://tldr.ostera.io).
 
 ## What is tldr?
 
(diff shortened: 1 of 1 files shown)


rubenvereecken:
Ooh pretty.

rubenvereecken:
Was going to comment on Alfred Work flow change but seems you just moved it. Looks goodgood!

waldyrious:
Btw, slightly off-topic: @rubenvereecken, @ostera and I have been having some issues with fixing stuff every now and then because we're not members of the tldr org. Think you could take care of that?

waldyrious:
For future reference, after a brief discussion in Gitter (full disclosure: not in the public channel, although we've since decided to hold all project-related discussions in the project's public channel), @ostera and I have been made organization members on the same day (3 July).

PR #940, opened 2016-07-04
webpack: add page

agnivade:
This is what I feel is the most basic usage of webpack from the command line.

/cc @sokra / @bebraw (webpack authors) for further comments.


pages/common/webpack.md
```diff
@@ -0,0 +1,19 @@
...
+
+`webpack {{app.js}} {{bundle.js}}`
+
+- Load css files too from your js file. (This uses the css loader for .css files):
+
+`webpack {{app.js}} {{bundle.js}} --module-bind 'css=css'`
+
+- Pass a config file and show compilation progress along with colors:
+
+`webpack --config {{webpack.config.js}} --progress --colors`
```
bebraw: If I remember correctly, `--colors` might be redundant now. I think it might figure out the color support automatically. You would have to check this from the source, though.

agnivade: You are right ! It makes a call to the `supports-color` module automatically if the flag is not passed explicitly. Removed the flag now.

pages/common/webpack.md
```diff
@@ -0,0 +1,19 @@
+# webpack
+
+> Bundle your js files and other assets into a single output file.
```
waldyrious: "your" --> "a web project's"?

pages/common/webpack.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Create a single output file from an entry point file:
+
+`webpack {{app.js}} {{bundle.js}}`
+
+- Load css files too from your js file. (This uses the css loader for .css files):
+
+`webpack {{app.js}} {{bundle.js}} --module-bind 'css=css'`
```
waldyrious: How are css assets included in the bundle.js file? This isn't clear from this example.

pages/common/webpack.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Pass a config file and show compilation progress:
+
+`webpack --config {{webpack.config.js}} --progress`
```
waldyrious: What does the config file include? One or two examples of its possible (required/optional) contents would help clarify this.

pages/common/webpack.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Automatically recompile on file change:
+
+`webpack --watch {{app.js}} {{bundle.js}}`
```
waldyrious: Does this only watch changes in the app.js file, or in any of its dependencies?

pages/common/webpack.md
agnivade: Yea I know, it is kind of beyond the purview of a command line thing. Because you would be going into the contents of the js file. That's why, I did not know how to go about it. Where should I write that ?

pages/common/webpack.md
agnivade: The example contents is kind of pretty big. So I was not sure whether to include it or not. Here is one such example
```javascript
module.exports = {
  entry: './src/main.js',
  output: {
    path: './dist',
    filename: 'output.js'
  }
};
```
How should I include it ?

pages/common/webpack.md
agnivade: Any dependency in the graph.

pages/common/webpack.md
waldyrious: I'm just asking so I will know, in order to be able to give more informed commentary :)

pages/common/webpack.md
waldyrious: what's the bare minimum such a file must have to work? I would say just list the properties included in the file, e.g. the entry point and the output filename.

pages/common/webpack.md
waldyrious: In that case, I'd suggest "recompile on changes to the project files", or something like that.

pages/common/webpack.md
```diff
@@ -10,10 +10,10 @@
 
 `webpack {{app.js}} {{bundle.js}} --module-bind 'css=css'`
 
-- Pass a config file and show compilation progress:
+- Pass a config file and show compilation progress. (As an example config, you can export an object containing `entry` and `output.filename` properties, which correspond to the entry and output file names):
```
waldyrious: I was envisioning something simpler, say: "Pass a config file (with e.g. the entry script and output filename) and show compilation progress:". What do you think?

agnivade: While it does not give the exact syntax, you have some idea regarding what the config file should contain. I am okay with it.

waldyrious: Yes, since providing the exact syntax would make this too verbose, I think this is the right balance to strike.

pages/common/webpack.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Create a single output file from an entry point file:
+
+`webpack {{app.js}} {{bundle.js}}`
+
+- Load css files too from your js file. (This uses the css loader for .css files):
+
+`webpack {{app.js}} {{bundle.js}} --module-bind 'css=css'`
+
+- Pass a config file(with eg. the entry script and the output filename) and show compilation progress:
```
waldyrious: please add a space before the opening parenthesis.

pages/common/webpack.md
```diff
@@ -0,0 +1,19 @@
+# webpack
+
+> Bundle a web project's js files and other assets into a single output file.
+
+- Create a single output file from an entry point file:
+
+`webpack {{app.js}} {{bundle.js}}`
+
+- Load css files too from your js file. (This uses the css loader for .css files):
```
waldyrious: "your" --> "the". Also, remove the period before the parenthesis (since the whole sentence is punctuated with a colon in the end), and lowercase the word "this" inside the parenthesis,

pages/common/webpack.md
agnivade: done

pages/common/webpack.md
agnivade: done

waldyrious:
LGTM (sorry for all the nit-picking!), but unfortunately this seems to have diverged from the master branch. Could you rebase?

agnivade:
Done.

waldyrious:
Thanks :)

PR #941, opened 2016-07-04
Make extended regular expressions clearer

thalesmello:
By default, grep already uses regular expressions when searching.

The example `grep -e {{^regex$}} {{path/to/file}}` is the same as `grep {{^regex$}} {{path/to/file}}`.

However, because of the comment about extended regular expressions, I mistakenly assumed `-e` was the option to enable it.

I believe most people would refer to `tldr` in this use case looking for the `-E` extended regular expressions.

With this in mind, I believe that example would be better rephrased as this pull request makes it.

### DIFF
--- a/pages/common/grep.md
+++ b/pages/common/grep.md
@@ -15,9 +15,9 @@
 
 `grep -rI {{search_string}} .`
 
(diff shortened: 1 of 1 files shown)


ostera:
Great catch @thalesmello!

PR #944, opened 2016-07-07
add ed.md

waldyrious:
inspired by this very nice introductory blog post:
https://sanctum.geek.nz/arabesque/actually-using-ed/


rubenvereecken:
This is some arcane old school stuff, woah.

I always think commands that fire up an interactive session are a bit of a grey zone (also we can barely ever cover enough for the session to be usable as such), but I'm cool with this one.

:+1:

PR #945, opened 2016-07-09
uniq: add sorting example

waldyrious:
This involves another command, but it's such a natural extension of uniq's `-c` functionality that I feel it's warranted to show here.

We should probably add a sort to the -c example too, because uniq only deals with *sequential* line repetitions.

### DIFF
--- a/pages/common/uniq.md
+++ b/pages/common/uniq.md
@@ -17,3 +17,7 @@
 - Display number of occurences of each line along with that line:
 
 `uniq -c {{file}}`
+
+- Display number of occurences of each line, sorted by the most frequent:
+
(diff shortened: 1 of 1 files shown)


ostera:
I was about to comment on that, `uniq` only works with `sequential` repetitions, so it wouldn't work if the lines aren't sorted. 

On Line 7 we could do `sort | uniq`

waldyrious:
@ostera exactly. The example added in this PR would also benefit from a preliminary sort. I'm planning on proposing that change in a separate PR after this one is merged.

ostera:
@waldyriois, can't wait for the next one :)

PR #946, opened 2016-07-09
iostat: Added some flags

agnivade:
- Added basic command usage without any flags.
- Removed the '-h' flag as it seems to be added implicitly by the cmd.
- Added the interval parameter.
- Added the '-m' flag.

### DIFF
--- a/pages/linux/iostat.md
+++ b/pages/linux/iostat.md
@@ -1,19 +1,23 @@
 # iostat
 
-> Report stats for devices and partitions.
+> Report statistics for devices and partitions.
 
-- Display disk statistics with disk IDs in human readable format:
+- Display a report of CPU and disk statistics since system startup:
 
-`iostat -h`
-
-- Display disk statistics with disk names (including LVM) in human readable format:
-
-`iostat -Nh`
+`iostat`
 
 - Display CPU statistics:
 
 `iostat -c`
 
-- Display extended disk statistics with disk names:
+- Display disk statistics with disk names (including LVM):
 
-`iostat -xN`
+`iostat -N`
+
+- Display extended disk statistics with disk names for device "sda":
+
+`iostat -xN {{sda}}`
+
+- Display incremental reports of CPU and disk statistics every "interval" second(s) with units converted to megabytes:
+
+`iostat -m {{interval}}`
--- a/pages/linux/iostat.md
+++ b/pages/linux/iostat.md
@@ -6,6 +6,10 @@
 
 `iostat`
 
+- Display a report of CPU and disk statistics since system startup with units converted to megabytes:
+
+`iostat -m`
+
 - Display CPU statistics:
 
 `iostat -c`
@@ -18,6 +22,6 @@
 
 `iostat -xN {{sda}}`
 
-- Display incremental reports of CPU and disk statistics every "interval" second(s) with units converted to megabytes:
+- Display incremental reports of CPU and disk statistics every "2" second(s):
 
-`iostat -m {{interval}}`
+`iostat {{2}}`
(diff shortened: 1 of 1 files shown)


pages/linux/iostat.md
```diff
@@ -1,19 +1,23 @@
...
+`iostat -N`
+
+- Display extended disk statistics with disk names for device "sda":
+
+`iostat -xN {{sda}}`
+
+- Display incremental reports of CPU and disk statistics every "interval" second(s) with units converted to megabytes:
 
-`iostat -xN`
+`iostat -m {{interval}}`
```
waldyrious: Perhaps a specific example (like "sda" above) would make the description easier to read -- e.g., "every 5 seconds". 

Also, does the -m take the parameter or is it independent (i.e., for the megabytes thing)? If it's the latter, having these together may give out the wrong idea.

waldyrious:
It's strange that the -h option does nothing. But assuming you tested this, I really like this PR, it's clearly an improvement. I just added an inline comment to the last example, which I think could be clearer.

agnivade:
Yea, I checked the [source code](https://github.com/sysstat/sysstat/blob/master/iostat.c#L956). It just adds a newline for -x case, which I does not make any difference anyway.

pages/linux/iostat.md
agnivade: Will add the example of 'interval'.

About the `-m`, it does give the impression that adding `-m` leads to incremental reports. Do you think I should add `-m` separately as another example ? Or merge with some other ? There are 5 examples already.

waldyrious: 6 examples is ok -- it's more of a guideline than a hard limit. If -m isn't related to the incremental reports, I would say it's preferable to have an extra example, yes.

agnivade: Added `-m` separately.

pages/linux/iostat.md
```diff
@@ -18,6 +22,6 @@
 
 `iostat -xN {{sda}}`
 
-- Display incremental reports of CPU and disk statistics every "interval" second(s) with units converted to megabytes:
+- Display incremental reports of CPU and disk statistics every "2" second(s):
```
waldyrious: I'd remove the quote marks around "2" and the parenthesis around the s in second(s). These make sense for a generic example, but are awkward for a concrete one.

pages/linux/iostat.md
```diff
@@ -1,19 +1,27 @@
...
+> Report statistics for devices and partitions.
 
-- Display disk statistics with disk IDs in human readable format:
+- Display a report of CPU and disk statistics since system startup:
 
-`iostat -h`
+`iostat`
 
-- Display disk statistics with disk names (including LVM) in human readable format:
+- Display a report of CPU and disk statistics since system startup with units converted to megabytes:
```
waldyrious: Considering that this example comes right after the basic one with no flags, perhaps the description could dispense with repeating the "since system startup"? That would make it shorter and easier to read, IMO.

pages/linux/iostat.md
```diff
@@ -1,19 +1,27 @@
...
+- Display disk statistics with disk names (including LVM):
 
-`iostat -xN`
+`iostat -N`
+
+- Display extended disk statistics with disk names for device "sda":
+
+`iostat -xN {{sda}}`
+
+- Display incremental reports of CPU and disk statistics every "2" second(s):
```
waldyrious: I'd remove the quote marks around "2" and the parenthesis around the s in second(s). These make sense for a generic example, but are awkward for a concrete one.

agnivade: Agree. Done.

waldyrious:
I'd have left the quote marks around "sda", since that is an actual name, but this is fine too :) I'll merge as soon as the Travis build passes. Thanks!

PR #947, opened 2016-07-12
Applying the 'path/to/item' convention uniformly

agnivade:
All pages follow a "path/to/some_file" convention. 

Its better to follow a single convention throughout. So fixed a few pages, which were not following it.

### DIFF
--- a/pages/linux/cryptsetup.md
+++ b/pages/linux/cryptsetup.md
@@ -4,8 +4,8 @@
 
 - Format a drive (or a file) to make it a LUKS volume:
 
-`cryptsetup luksFormat {{path_to_luks_volume_to_be}}`
+`cryptsetup luksFormat {{path/to/luks_volume_to_be}}`
 
 - Open a volume so that you can access it decrypted on the fly at /dev/mapper/{{a}}:
 
-`cryptsetup luksOpen {{path_to_luks_volume_to_be}} {{a}}`
+`cryptsetup luksOpen {{path/to/luks_volume_to_be}} {{a}}`
--- a/pages/osx/open.md
+++ b/pages/osx/open.md
@@ -16,7 +16,7 @@
 
 - Reveal a file in finder:
 
-`open -R {{/path/to/file}}`
+`open -R {{path/to/file}}`
 
 - Open all the files of a given extension in the current directory with the associated application:
 
--- a/pages/osx/open.md
+++ b/pages/osx/open.md
@@ -16,7 +16,7 @@
 
 - Reveal a file in finder:
 
-`open -R /path/dir/{{file}}`
+`open -R {{/path/to/file}}`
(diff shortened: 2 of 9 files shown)


pages/osx/open.md
```diff
@@ -16,7 +16,7 @@
 
 - Reveal a file in finder:
 
-`open -R /path/dir/{{file}}`
+`open -R {{/path/to/file}}`
```
waldyrious: Maybe remove the leading slash here? I believe the convention is to omit it, but not sure if you aimed at fixing that too.

ostera:
@agnivade man, thanks a lot for this :) let's get it merged asap!

pages/linux/cryptsetup.md
```diff
@@ -4,8 +4,8 @@
 
 - Format a drive (or a file) to make it a LUKS volume:
 
-`cryptsetup luksFormat {{path_to_luks_volume_to_be}}`
+`cryptsetup luksFormat {{path/to/luks_volume_to_be}}`
```
waldyrious: If I understand this example correctly, perhaps "path/to/new_luks_volume" would be more intuitive. But I may be misinterpreting the meaning of the token; let me know what you think.

pages/linux/cryptsetup.md
```diff
@@ -4,8 +4,8 @@
...
 
 - Open a volume so that you can access it decrypted on the fly at /dev/mapper/{{a}}:
 
-`cryptsetup luksOpen {{path_to_luks_volume_to_be}} {{a}}`
+`cryptsetup luksOpen {{path/to/luks_volume_to_be}} {{a}}`
```
waldyrious: This is slightly odd. Is `{{a}}` the existing volume and `{{path/to/luks_volume_to_be}}` a new LUKS volume that will be created from it? We should probably clarify the token names. Any suggestions?

waldyrious:
Great PR, thanks a lot! This was sorely needed. We ought to include this guideline in CONTRIBUTING.md, by the way.

pages/osx/open.md
agnivade: oops, yes.

pages/linux/cryptsetup.md
agnivade: These comments seem to be about the `cryptsetup` command. I need to research this a bit. How about I push this commit and address these concerns in a separate commit ?

waldyrious: Sure, makes sense.

waldyrious:
LGTM. Since @ostera didn't voice any objections, I'll merge :)

pages/linux/cryptsetup.md
agnivade: Raised #953

PR #949, opened 2016-07-13
w3m: add page

rcorrie:
http://w3m.sourceforge.net/

### DIFF
--- a/pages/common/w3m.md
+++ b/pages/common/w3m.md
@@ -1,0 +1,11 @@
+# w3m
+
+> A text based Web browser and pager.
+
+- Open an URL:
+
+`w3m {{http://example.com}}`
+
+- Quit w3m:
+
+`'q' then 'y'`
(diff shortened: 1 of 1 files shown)


pages/common/w3m.md
```diff
@@ -0,0 +1,11 @@
+# w3m
+
+> A text based Web browser and pager.
```
waldyrious: I believe "text-based" is the correct form. Also, the "...and pager" is probably more puzzling than helpful to beginners. Is it really needed?

waldyrious:
That's a nice start, thanks :) please check out the inline comment.

pages/common/w3m.md
rcorrie: @waldyrious agreed on both counts. Will update shortly.

waldyrious:
Looks good 👍

PR #950, opened 2016-07-13
git-svn: add `fetch` subcommand

rcorrie:
The fetch command is as essential as the rebase command when working with remote SVN repos

### DIFF
--- a/pages/common/git-svn.md
+++ b/pages/common/git-svn.md
@@ -6,6 +6,10 @@
 
 `git svn clone {{http://example.com/my_subversion_repo}} {{local_dir}}`
 
+- Fetch new revisions from remote SVN repository:
+
+`git svn fetch`
+
 - Update local clone from the upstream SVN repository:
 
 `git svn rebase`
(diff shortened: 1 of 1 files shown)


waldyrious:
Thanks for the contribution! One question: does `git svn rebase` do the same as `git pull --rebase` for a git repo? If so, perhaps this example should specify "without changing the current HEAD" or something like that (and possibly be inserted after the git svn rebase example, which would be a more common operation).

rcorrie:
On your question: yes, correct. Although I wouldn't move it after the 'rebase' example since you would perform the fetch before the rebase.

```sh
git svn fetch
git svn rebase
# change something
git svn dcommit
# rinse, repeat
```

pages/common/git-svn.md
```diff
@@ -10,6 +10,10 @@
 
 `git svn rebase`
 
+- Fetch updates from remote SVN repository without changing the GIT HEAD:
```
waldyrious: does GIT have to be in uppercase? I thought the name of the variable was just "HEAD".

ostera:
btw @rcorrie, check your `.gitconfig` file because your commits are not being linked to your user

PR #951, opened 2016-07-14
git-remote: unicode quotes to ascii quotes

lmount:
`git-remote` is the only file with unicode characters (two of them), both of which can be replaced with their ascii counterparts.

Living in 2016 this shouldn't pose a problem, but then again many terminals are ascii only.


waldyrious:
Tkank you, makes sense :+1:

PR #953, opened 2016-07-14
cryptsetup: simplify page

agnivade:
- Cited actual examples rather than placeholder text
- Added the 'close' command

### DIFF
--- a/pages/linux/cryptsetup.md
+++ b/pages/linux/cryptsetup.md
@@ -1,15 +1,15 @@
 # cryptsetup
 
-> Manage plain dm-crypt and LUKS encrypted volumes.
+> Manage plain dm-crypt and LUKS(Linux Unified Key Setup) encrypted volumes.
 
 - Initialize a LUKS volume (overwrites all data on the partition):
 
 `cryptsetup luksFormat {{/dev/sda1}}`
 
-- Open a LUKS volume and create a mapping at/dev/mapper/{{backup}}:
+- Open a LUKS volume and create a decrypted mapping at /dev/mapper/{{target}}:
 
-`cryptsetup luksOpen {{/dev/sda1}} {{backup}}`
+`cryptsetup luksOpen {{/dev/sda1}} {{target}}`
 
 - Remove an existing mapping:
 
-`cryptsetup luksClose {{backup}}`
+`cryptsetup luksClose {{target}}`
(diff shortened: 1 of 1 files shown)


agnivade:
My only concern is the word "mapping" might be hard to understand. I couldn't find a better alternative. Let me know if there's anything better.

pages/linux/cryptsetup.md
```diff
@@ -2,10 +2,14 @@
 
 > Manage plain dm-crypt and LUKS encrypted volumes.
```
waldyrious: would be nice to expand the LUKS acronym here :)

pages/linux/cryptsetup.md
```diff
@@ -2,10 +2,14 @@
...
 
-- Format a drive (or a file) to make it a LUKS volume:
+- Initialize a LUKS volume (overwrites all data on the partition):
 
-`cryptsetup luksFormat {{path/to/luks_volume_to_be}}`
+`cryptsetup luksFormat {{/dev/sda1}}`
 
-- Open a volume so that you can access it decrypted on the fly at /dev/mapper/{{a}}:
+- Open a LUKS volume and create a mapping at/dev/mapper/{{backup}}:
```
waldyrious: using  "backup" may give out the wrong idea (it may not be clear that this is just an example, and people may instead interpret it as representing the purpose of this particular command). Would `{{decrypted}}` make sense as an alternative?

agnivade: `{{decrypted}}` sounds like a mouthful. How about `{{target}}` ?

(Nice catch on the missing space :+1: )

waldyrious: That could work, but if we go with that we should somehow mention that this decrypts the volume so it can be accessed normally; otherwise, it seems kinda purposeless to simply map a volume to another one.

agnivade: Agree. Will do.

pages/linux/cryptsetup.md
```diff
@@ -1,11 +1,15 @@
 # cryptsetup
 
-> Manage plain dm-crypt and LUKS encrypted volumes.
+> Manage plain dm-crypt and LUKS(Linux Unified Key Setup) encrypted volumes.
```
waldyrious: missing space before parenthesis

waldyrious:
The missing space is a minor detail, I'll fix after merging. Thanks! This page is much clearer now! 🎉 😄

PR #954, opened 2016-07-15
sort uniq first

alfredbez:
'uniq' does not detect repeated lines unless they are adjacent, so we need to sort them

### DIFF
--- a/pages/common/uniq.md
+++ b/pages/common/uniq.md
@@ -1,23 +1,24 @@
 # uniq
 
 > Output the unique lines from the given input or file.
+> Since it does not detect repeated lines unless they are adjacent, we need to sort them first.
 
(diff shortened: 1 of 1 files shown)


pages/common/uniq.md
```diff
@@ -1,23 +1,23 @@
 # uniq
 
-> Output the unique lines from the given input or file.
+> Output the unique lines from the given input or file. Since it does not detect repeated lines unless they are adjacent, so we need to sort them first.
```
waldyrious: Please add the second sentence in a new line, and remove the "so", which is redundant with "Since".

alfredbez:
fixed and rebased it

ostera:
Brilliant, thanks @alfredbez!

PR #960, opened 2016-07-19
rename.md: add -f flag

waldyrious:


pages/common/rename.md
```diff
@@ -10,7 +10,7 @@
 
 `rename -n {{'s/foo/bar/'}} {{\*}}`
 
-- Convert filenames to lower case:
+- Convert filenames to lower case (use `-f` in case-insensitive filesystems to prevent "already exists" errors):
```
agnivade: Just wondering - does "use -f to overwrite existing files" make sense ? Its just that reading the entire "in case-insensitive filesystems to prevent "already exists" errors" is trying to suggest that adding the '-f' flag prevents some sort of errors. Whereas the previous description says what its actually used for.

waldyrious: Yes, but then it wouldn't really apply to the case changing the case of a file in case-insensitive filesystems, since the target filename doesn't actually exist. We could add an example with -f before this one to ensure it gets a proper explanation -- would that work?

agnivade: Yep. I think that would be the proper way.

agnivade:
LGTM.

PR #962, opened 2016-07-20
CONTRIBUTING: greatly simplify the PR instructions

waldyrious:
plus some wording tweaks here and there

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,20 +1,20 @@
 # Contributing
 
-Contributions are most welcome! All `tldr` pages are stored in Markdown right here on GitHub. Just open an issue or send a pull request and we'll merge it as soon as possible.
+Contributions are most welcome! All `tldr` pages are stored in Markdown right here on GitHub. Just open an issue or send a pull request and we'll incorporate it as soon as possible.
 
-*Note*: when submitting a new command, don't forget to check if there's already a pull request in progress.
+*Note*: when submitting a new command, don't forget to check if there's already a pull request in progress for it.
 
 ## Guidelines
 
-Note that `tldr` is focused on concrete examples.
+The basic format of a `tldr` page is a set of concrete usage examples.
 Here are a few guidelines to get started:
 
 1. Focus on the 5 or 6 most common usages. It's OK if the page doesn't cover everything; that's what `man` is for.
 2. When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness.
-3. Try to incorporate the spelled-out version of single-letter options in the example's description.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -38,7 +38,8 @@ The format of each page should match the following:
 ```
 
 We actually have a linter/formatter that enforces our format.
-It even automatically cleans up your pages for you! Installing it is easy:
+It is run automatically on every pull request,
+but you may install it to test your contributions locally before submitting them:
 
 ```
 npm install
```
agnivade: Will it be better to write `npm test` instead of `tldr tldrl` ? The `npm test` was mentioned below but you removed it, which is fine. I think mentioning it here makes sense.

CONTRIBUTING.md
```diff
@@ -66,97 +67,32 @@ clients that need to extract a single description/example.
...
 ## Submitting a pull request
 
-TL;DR: fork, clone, `npm install`, feature branch, commit, push, pull request, check Travis.
+For submitting changes, you can use whatever workflow you're more comfortable with.
+The easiest way is to just edit the page directly on the Github interface.
+Check out the step-by-step instructions (with screenshots) on
+[Github Help](https://help.github.com/articles/editing-files-in-another-user-s-repository/).
 
-Detailed explanation:
+Alternatively, you can do it using the command line:
```
agnivade: I think adding a 1 line command example with each of these points will go a long way. For a newcomer, its extremely easy to just copy-paste the commands and immediately start working on a new command. :smile:

agnivade:
Lovely PR ! The page looks much neater now. Left some minor inline comments.

CONTRIBUTING.md
waldyrious: Yes, I thought about that. Even though for a newcomer the web-based workflow would be easier, it fits the spirit of this project to help people learn to use tools that may seem daunting at first :)

I'm thinking of doing something like this: https://www.mediawiki.org/wiki/Gerrit/Getting_started (simplified wherever possible/relevant). What do you think?

CONTRIBUTING.md
waldyrious: Sure -- nice catch. I'll change it.

CONTRIBUTING.md
agnivade: >   for a newcomer the web-based workflow would be easier

agree.

> I'm thinking of doing something like this: https://www.mediawiki.org/wiki/Gerrit/Getting_started

This is also okay. I like the section wise approach the article takes.
- Get the code
```
some git commands
```
- Modify the code
```
some git commands
```
- Send a pull request
```
more commands
```
Of course, a lot needs to be simplified. The git commands shown in gerrit are too deep for a newbie.

waldyrious: Yes. I'll give it a shot later today.

agnivade:
LGTM.

PR #963, opened 2016-07-21
echo: added -n, -e examples

AgamAgarwal:
Added examples for the options `-n` and `-e` for `echo`

### DIFF
--- a/pages/common/echo.md
+++ b/pages/common/echo.md
@@ -9,3 +9,11 @@
 - Print a message with environment variables:
 
 `echo {{"My path is $PATH"}}`
+
+- Print a message without the trailing newline:
+
+`echo -n {{"Hello World"}}`
+
+- Enable backslash escapes:
+
+`echo -e {{"Column 1\tColumn 2"}}`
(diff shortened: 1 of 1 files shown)


pages/common/echo.md
```diff
@@ -9,3 +9,11 @@
 - Print a message with environment variables:
 
 `echo {{"My path is $PATH"}}`
+
+- Print a message without the trailing newline:
+
+`echo -n {{"Hello World"}}`
+
+- Enable backslash escapes:
```
waldyrious: How about "Enable enhanced mode to include special characters (escaped with backslash)"? That way the -e maps to "enhanced"

AgamAgarwal: This sounds interesting.
Though, as per my knowledge, `-e` corresponds to "escapes".

waldyrious: You're right. In that case I'd suggest "Enable interpretation of backslash escapes (special characters)".

AgamAgarwal: Done

ostera:
LGTM

PR #964, opened 2016-07-21
[RFC] cp: major overhaul of the page

waldyrious:
- change "directory" to "folder" (more beginner-friendly)
- change "original" --> "copy" to "file"/"folder" --> "copy", to be more explicit (inspired by #289)
- use an extension to distinguish file paths from folder paths
- reword some of the descriptions for clarity
- include examples to clarify when the target argument defines the actual copy, and when it defines the folder where to put the copy
- add an example to copy contents of a folder into another (existing) folder

### DIFF
--- a/pages/common/cp.md
+++ b/pages/common/cp.md
@@ -4,24 +4,28 @@
 
 - Copy a file to another location:
 
-`cp {{/path/to/file.ext}} {{/path/to/copy.ext}}`
+`cp {{path/to/file.ext}} {{path/to/copy.ext}}`
 
 - Copy a file into another folder, keeping the filename:
 
-`cp {{/path/to/file.ext}} {{path/to/target/parent/folder}}`
+`cp {{path/to/file.ext}} {{path/to/target/parent/folder}}`
 
 - Copy a folder recursively to another location:
 
-`cp -r {{/path/to/folder}} {{/path/to/copy}}`
+`cp -r {{path/to/folder}} {{path/to/copy}}`
 
 - Copy a folder recursively into another folder, keeping the folder name:
 
-`cp -r {{/path/to/folder}} {{/path/to/target/parent/folder}}`
+`cp -r {{path/to/folder}} {{path/to/target/parent/folder}}`
 
 - Copy a folder recursively, in verbose mode (shows files as they are copied):
 
-`cp -vr {{/path/to/folder}} {{/path/to/copy}}`
+`cp -vr {{path/to/folder}} {{path/to/copy}}`
 
 - Copy the contents of a folder into another folder:
 
-`cp -r {{/path/to/source/folder/*}} {{/path/to/target/folder}}`
+`cp -r {{path/to/source/folder/*}} {{path/to/target/folder}}`
+
+- Interactively copy text files in the current directory to another location:
+
+`cp -i {{*.txt}} {{path/to/source/}}
--- a/pages/common/cp.md
+++ b/pages/common/cp.md
@@ -28,4 +28,4 @@
 
 - Copy text files to another location, in interactive mode (prompts user before overwriting):
 
-`cp -i {{*.txt}} {{path/to/source/}}
+`cp -i {{*.txt}} {{path/to/source/}}`
--- a/pages/common/cp.md
+++ b/pages/common/cp.md
@@ -1,27 +1,35 @@
 # cp
 
-> Copy files.
+> Copy files and folders.
 
-- Copy files in arbitrary locations:
+- Copy a file to another location:
 
-`cp {{/path/to/original}} {{/path/to/copy}}`
+`cp {{/path/to/file.ext}} {{/path/to/copy.ext}}`
 
-- Copy a file to a parent directory:
+- Copy a file into another folder, keeping the filename:
 
-`cp {{/path/to/original}} ../{{path/to/copy}}`
+`cp {{/path/to/file.ext}} {{path/to/target/parent/folder}}`
 
-- Copy directories recursive using the option -r:
+- Copy a folder recursively to another location:
 
-`cp -r {{/path/to/original}} {{/path/to/copy}}`
+`cp {{/path/to/folder}} {{/path/to/copy}}`
 
-- Show files as they are copied:
+- Copy a folder recursively into another folder, keeping the folder name:
 
-`cp -vr {{/path/to/original}} {{/path/to/copy}}`
+`cp -r {{/path/to/folder}} {{/path/to/target/parent/folder}}`
 
-- Make a copy of a file, adding an extension:
+- Copy a folder recursively, in verbose mode (shows files as they are copied):
+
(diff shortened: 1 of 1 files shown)


waldyrious:
I'm marking this as RFC because these changes make the page too long, so we might want to trim down some of the examples.

pages/common/cp.md
```diff
@@ -1,27 +1,35 @@
...
+> Copy files and folders.
 
-- Copy files in arbitrary locations:
+- Copy a file to another location:
 
-`cp {{/path/to/original}} {{/path/to/copy}}`
+`cp {{/path/to/file.ext}} {{/path/to/copy.ext}}`
 
-- Copy a file to a parent directory:
+- Copy a file into another folder, keeping the filename:
```
ostera: This and the first one are redundant IMO.

waldyrious: As I said in the opening comment, I've added both examples "to clarify when the target argument defines the actual copy, and when it defines the folder where to put the copy". Don't you think the second usage may not be clear if we don't explicitly mention it?

pages/common/cp.md
```diff
@@ -1,27 +1,35 @@
...
+`cp -r {{/path/to/folder}} {{/path/to/target/parent/folder}}`
 
-- Make a copy of a file, adding an extension:
+- Copy a folder recursively, in verbose mode (shows files as they are copied):
+
+`cp -vr {{/path/to/folder}} {{/path/to/copy}}`
+
+- Copy the contents of a folder into another folder:
+
+`cp -r {{/path/to/source/folder/*}} {{/path/to/target/folder}}`
```
ostera: I think this one can be scrapped, but I have no strong arguments against keeping it. It just feels a bit cumbersome to have 3 recursive-copy examples.

pages/common/cp.md
```diff
@@ -1,27 +1,35 @@
...
+
+- Copy of a file adding an extension:
```
ostera: This last two could be squashed into a single one.

pages/common/cp.md
waldyrious: I added because this was what prompted me to make this change :P but I'm ok with removing it, as it doesn't exactly cover a cp-specific functionality, despite its (subjective) usefulness.

pages/common/cp.md
waldyrious: Agreed. Any suggestions on what to change/keep?

pages/common/cp.md
```diff
@@ -1,27 +1,35 @@
...
+- Copy a file into another folder, keeping the filename:
 
-`cp {{/path/to/original}} ../{{path/to/copy}}`
+`cp {{/path/to/file.ext}} {{path/to/target/parent/folder}}`
 
-- Copy directories recursive using the option -r:
+- Copy a folder recursively to another location:
 
-`cp -r {{/path/to/original}} {{/path/to/copy}}`
+`cp {{/path/to/folder}} {{/path/to/copy}}`
```
waldyrious: Ha! Good catch :) I'll change.

pages/common/cp.md
ostera: I like the idea of using shell expansion, as I use them all the time (`!!:s/this/that/` is a life savior), but they are a bit specific to whatever the user is running.

I'd rather have `cp file.html file.html.backup` which is slightly more readable and covers both cases (since the backup could clearly be named anything), without digging into what `{{,.backup}}` actually implies.

pages/common/cp.md
ostera: I mean this could be covered by `cp -r path1 path2; mv path2 path3`, but I see what you mean. Could we remove the one that keeps the folder name maybe?

waldyrious: I'd like to keep that one for symmetry with the file-based examples, and for completeness :( I'm honestly conflicted as to what to remove (hence the RFC)

pages/common/cp.md
agnivade: Gave some thought to this. Have to agree with @ostera on this. IMO, the second usage is easily deducible from the first.

pages/common/cp.md
agnivade: 3 examples with `-r` does seem a bit too much :smile:. Honestly, I always get stumped over whether to add the trailing / at the end of a folder or not. I would just keep your first and last example(the one with /*) and be done with it.

agnivade:
Since we are on this page, can we add the `-i` option too ? Seems like a useful flag to add.

pages/common/cp.md
waldyrious: From my tests, the trailing slash makes no difference by itself -- in fact, adding a trailing slash to the source directory actually seems to cause the copying not to happen (tested on OS X, if it matters).

pages/common/cp.md
waldyrious: Fair enough, I can see your points -- I'll remove it.

pages/common/cp.md
waldyrious: @agnivade does that mean you'd remove the "Copy a folder recursively into another folder, keeping the folder name" example? I think the file version of this example is deducible, but this one is less intuitive. I may be biased though as I recently had to do it, so your comments are welcome. How about we keep that one and merge the -v with the `/*` example?

pages/common/cp.md
waldyrious: In that case we could simply remove the last two examples, since the first example covers this. I'd rather keep one of the -r examples as they're related to cp's functionality and may not be as easily deduced.

waldyrious:
We could add -i, sure, but we first need to resolve the existing space issues 😄

pages/common/cp.md
agnivade: >  in fact, adding a trailing slash to the source directory actually seems to cause the copying not to happen

Wow .. interesting.

>  How about we keep that one and merge the -v with the /* example?

I'm okay with it. Since we are removing it from the plain copy example, its okay to keep it here.

waldyrious:
Ok guys, I've removed the last 2 examples which were shell features rather than cp-specific features, and added the missing `-r` as @ostera suggested. Since that means we won't have the excess space issues, I think it could be merged as-is, i.e. including two examples for file copy and their duals for directory copy.

agnivade:
Can we add `-i` now ? :stuck_out_tongue_closed_eyes:

waldyrious:
@agnivade sure -- any suggestions for a description?

agnivade:
Something on the lines of "Interactively copy files or folders from target to destination"

waldyrious:
Hmm, maybe it would be nice to use a glob pattern for this one for a change, what do you think?

agnivade:
Sure.

waldyrious:
@agnivade PTAL :)

waldyrious:
Hurray! After almost an year, this PR will get merged :)

ps - I rebased the commits interactively in the command line, to clean up the branch a bit before merging.

PR #966, opened 2016-07-22
Removed useless use of cat

AgamAgarwal:
Removed a case of useless use of `cat`


waldyrious:
Nice one :) LGTM :+1:

rubenvereecken:
We have a few of these lying around, really just artifacts of clueless contributors like me ;)

felixonmars:
👍

PR #969, opened 2016-07-22
Removed 5 cases of useless use of cat

AgamAgarwal:
This removes some more cases of useless use of `cat`.

Note for `pages/common/grep.md`: Piping the contents of file to stdin defeats the purpose of the example demonstrating how to use the stdin instead of a file.

### DIFF
--- a/pages/common/grep.md
+++ b/pages/common/grep.md
@@ -37,7 +37,7 @@
 
 - Use the standard input instead of a file:
 
-`cat {{path/to/file}} | grep {{search_string}}`
+`grep {{search_string}}`
 
 - Invert match for excluding specific strings:
 
--- a/pages/common/http.md
+++ b/pages/common/http.md
@@ -28,4 +28,4 @@
 
 - Specify raw request body via stdin:
 
-`cat data.txt | http PUT example.org`
+`http PUT {{example.org}} < {{path/to/file}}`
--- a/pages/common/lwp-request.md
+++ b/pages/common/lwp-request.md
@@ -9,7 +9,7 @@
 
 - Upload a file with a POST request:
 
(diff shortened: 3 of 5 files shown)


waldyrious:
I'm not sure I understand what you mean regarding the grep example... care to clarify?

waldyrious:
Also, I wouldn't call every such usage of `cat` useless. The usage of `<` is less intuitive as the command can't be read sequentially, so there's certainly some pedagogical / memorization advantage to using `cat`, even though it's not strictly required.

AgamAgarwal:
`grep` - The example is about how to use stdin instead of a file. The original example pipes the contents of the file to `grep`. Thus, when you run this example, you are not typing the input in stdin but are essentially reading it from a file. Thus, this sort of defeats the purpose of the example.

The usage of `cat` in such cases is useless because:
1. You are calling 2 programs instead of one.
2. What you want is to "connect" the file to the stdin. That is exactly what the redirection operator `<` does.

Also, there are numerous articles which discuss this. You may want to take a look.

ostera:
@AgamAgarwal, while I agree that these look more idiomatic and have slightly less command overhead, they are clearly less clear (oh the wordplay).

I stand with @waldyrious, since the average un*x shell already has [plenty of obscure mannerisms](https://github.com/ostera/try#motivation), just as [I stated before](https://github.com/tldr-pages/tldr/pull/964#discussion_r71716109).

Ultimately, it defeats our purpose.

We want to ease people into using the shell more and more by providing digestible pieces of content. Showing the _one true way_ and scaring the living crap out of them with cryptic symbology that they may or may not need isn't doing anyone any favors.

Closing until further notice.

waldyrious:
Cross-referencing #966 where a similar proposal (albeit mixed with another change) was received differently.

waldyrious:
Also tangentially related: #819

PR #971, opened 2016-07-23
meteor: add page

agnivade:


agnivade:
Adding one of my favorite javascript frameworks :smile:

pages/common/meteor.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Run a meteor project from its root directory in development mode:
+
+`meteor`
+
+- Create a project under the given directory:
+
+`meteor create {{path/to/directory}}`
+
+- Display the list of packages your project is currently using:
```
waldyrious: I'd rather avoid the "you" (yet another thing that could be in the guidelines, although we must take care not to overdo it)

How about: "Display the list of packages the project is currently using" (or "a project"), "Display the list of packages the current project is using", or "Display the list of packages used in the current project"?

pages/common/meteor.md
```diff
@@ -0,0 +1,27 @@
...
+
+`meteor list`
+
+- Add a package to your project:
```
waldyrious: I'd remove the "your" here too.

pages/common/meteor.md
```diff
@@ -0,0 +1,27 @@
...
+
+`meteor add {{package_name}}`
+
+- Remove a package from your project:
```
waldyrious: ibid.

pages/common/meteor.md
```diff
@@ -0,0 +1,27 @@
...
+
+`meteor remove {{package_name}}`
+
+- Create a production build of your project under the given directory. (Omit the --directory option if you want to wrap the build inside a tarball instead):
```
waldyrious: idem (the second "your" could be removed with simply "if you want to" --> "to").

Also, the second part could be a separate example. It would make this page longer than we've been comfortable with so far, but since we haven't made a decision yet in #686, I can't really enforce it, just discourage it.

pages/common/meteor.md
agnivade: Ah, I make the same mistake again ! Will fix.

pages/common/meteor.md
agnivade: > Also, the second part could be a separate example. It would make this page longer than we've been comfortable with so far, 

Yea more than 6 examples is a bit too much. I will just remove the '--directory' option. Since no other commands have a repeat example, its better to keep things tidy.

waldyrious:
LGTM :+1:

agnivade:
@ostera - Can we have your blessings on this one ? :smile:

PR #973, opened 2016-07-24
Improved "Git" command description

gustavosotnas:


ostera:
While we are here, it'd be interesting to point out in this page _some_ of the most common git commands. I've noticed people are bailing after reaching this page and this might be the cause.

Think you're starting with Git and don't necessarily understand how it implements extensions. You'd likely just go with `git` as your search term. Only the brave will go the extra mile and paste part of their commands. Only the lucky will get what they need.

That already is a very small subset of our users, and I think we can do better :)

waldyrious:
A "see also" section (#785) would be most useful here. We need to decide on that ASAP.

PR #974, opened 2016-07-24
add subsections to "submit a PR" section

waldyrious:
this allows the reader to visually separate the instructions and makes the longer git-command section less prone to trigger a TLDR (heh) when reading this part.

also, line wrapping tweaks, split a paragraph into two, and reword a couple passages for clarity

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,7 @@
 # Contributing
 
-Contributions are most welcome! All `tldr` pages are stored in Markdown right here on GitHub. Just open an issue or send a pull request and we'll incorporate it as soon as possible.
+Contributions are most welcome! All `tldr` pages are stored in Markdown right here on GitHub.
+Just open an issue or send a pull request and we'll incorporate it as soon as possible.
 
 *Note*: when submitting a new command, don't forget to check if there's already a pull request in progress for it.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -59,43 +62,44 @@ Some examples:
 - `mysql {{database_name}}`
 - `unrar x {{compressed.rar}}`
 
-In short, make it as intuitive as possible for the user to figure out
-how to use the command and fill it in with values.
+In short, make it as intuitive as possible
```
agnivade: IMO this is more readable as a single line.

CONTRIBUTING.md
```diff
@@ -121,4 +125,8 @@ Examples:
 **IMPORTANT**: By submitting a patch, you agree to license your work
 under the same license as that used by the project.
 
-You're free to modify or redistribute the content. That being said, but why not contribute over here? :) Say if you wanted to have `tldr` pages in `groff` format, why not have a client that uses [pandoc](http://johnmacfarlane.net/pandoc/) and periodically updates straight from this repo?
+You're free to modify or redistribute the content.
+That being said, but why not contribute over here? :)
+Say if you wanted to have `tldr` pages in `groff` format,
```
agnivade: Same here, IMO the last sentence need not be split into multiple lines.

CONTRIBUTING.md
ostera: _if_ we are going for semantic-ish linebreaks then it should be split at the comma and the rest left as a single line.

That'd minimize diffs in the future, and it'd still be rendered just fine.

waldyrious: My interpretation of semantic linebreaks is that lines should be broken in semantically meaningful blocks -- ideally full clauses, but not necessarily.

In this case, the full clause would be "make it as intuitive as possible for the user to figure out how to use the command and fill it in with values", but since that's still too long, I've broken down the sentence to the basic form "make it as intuitive as possible / for the user to / X", and split it in the first slash.

The second most sensible alternative, IMO, would be in the second slash, rather than in the middle of X as it was previously.

Using the comma doesn't help much since we still need to wrap the line, and the clause before the comma is too short for it to be useful to set it apart.

In any case, as @ostera says, this is ultimately not very problematic as the rendered version looks the same regardless of where we break the sentence.

CONTRIBUTING.md
waldyrious: I take it that you haven't heard of [semantic linebreaks](rhodesmill.org/brandon/2012/one-sentence-per-line/)? :)

ostera: Fixed link: http://rhodesmill.org/brandon/2012/one-sentence-per-line/

agnivade: Indeed, I hadn't heard of it. :stuck_out_tongue_closed_eyes: 

Well, 
if that's the intention here. 
I am okay with it too.

ostera:
Alright then! Seems that all feedback has been covered. I'll merge this :)

PR #976, opened 2016-07-24
Create dd.md

dylanrees:


agnivade:
Hi there, 

The page for dd already exists under the linux and osx folders. Please check whether the command runs under sunos or not, and then depending on it - either edit those pages, or remove those pages and add it in the `common` folder.

Also, please follow the git commit message format as specified in [CONTRIBUTING.md](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md)

dylanrees:
Looks like it runs on Solaris as indicated here (it's a pretty old tool so should be broadly available across Unix systems):

http://www.unix.com/man-page/OpenSolaris/1m/dd/

I'll go ahead and make a new commit.

Issue #481, opened 2015-12-30
Are you guys interested in adding a "see also" where it's relevant?


Parth:
For example, the [git.md](https://github.com/tldr-pages/tldr/blob/master/pages/common/git.md) might be one of the first pages a user views, only to see that `git clone`, `git commit`, etc aren't there. It doesn't seem obvious that `tldr git-clone` would show me how to use it. Perhaps if at the foot of pages like that, we could have a bulleted list of _See Also_, and list any other obvious commands. Would this form of a change be welcome?

waldyrious:
Well we already do this in the main description, e.g. in [uname.md](https://github.com/tldr-pages/tldr/blob/master/pages/common/uname.md). Do you think a dedicated section would be better?

Parth-Mehrotra:
Yes, I think it would add more structure if users knew that the if they go to the bottom of the notes they can look at other commands that are closely related, and how they're related. What do you think?

Parth-Mehrotra:
I'd be happy to go through and add this to the ones I know about

igorshubovych:
We were discussing this from time to time.

Actually at the moment any change like this require the changes in the numerous clients. We should very carefully change the format of data and it should be approved by different parties. There are several different propositions to change the format, but this is a drastic change for the project and I'd rather do it very-very carefully.

Parth-Mehrotra:
How about just adding a

# See also
+ git-commit: for blah blah blah
+ git-add: for blah blah blah

On the relevant pages?

igorshubovych:
We have many problems at the moment:
- not enough pages, many even basic commands not covered
- buried by existing pull requests
- better documentation, format description, guidelines
- issues with clients
- organizational issues
and others

Bringing change to the such a basic format will distract us from the main goal.

"See Also" is in our pipeline. One day we will come back discussing it. But it will be not today and not the next week and probably not the next month.

However there is other great open-source project which look at the man problem from different point of view. It might be useful for your case, I mean git example. It is called [cheat](https://github.com/chrisallenlane/cheat).

Parth-Mehrotra:
Okay, I'll go ahead and close this then, and maybe reopen in the distant future, is there a list of missing commands? I'd love to contribute in a meaningful way.

igorshubovych:
There is a [list of missing commands](https://github.com/tldr-pages/tldr/wiki/Pages-plan), but even this one is outdated :)

Now you see how far we are from discussion about format change.

[...]

PR #977, opened 2016-07-25
Move dd.md to common

dylanrees:


pages/common/dd.md
```diff
@@ -1,6 +1,6 @@
 # dd
 
-> Convert and copy a file.
+> Convert and copy a file. Useful for working directly with storage devices and filesystems.
```
ostera: I think we could split this into a second `> {text}` statement right below.

pages/osx/dd.md
```diff
@@ -1,19 +0,0 @@
-# dd
```
ostera: What was the diff between the osx and linux page? Were they exactly the same?

agnivade: Yes, except that the osx page didn't have the `status=progress` flag.

waldyrious: That difference is justified: I've tested `status=progress` on OSX and it results in "dd: unknown operand status". We should either remove `status=progress` from the common page, or keep the separate osx page.

sethwoodworth: Also the units for bs=2m is capitalized on one and not the other.
The gnu and BSD implementations of dd are sufficiently different imo
On Jul 25, 2016 06:14, "Waldir Pimenta" <notifications@github.com> wrote:

In pages/osx/dd.md
<https://github.com/tldr-pages/tldr/pull/977#discussion_r72039508>:

> @@ -1,19 +0,0 @@
> -# dd

That difference is justified: I've tested status=progress on OSX and it
results in "dd: unknown operand status". We should either remove
status=progress from the common page, or keep the separate osx page.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/tldr-pages/tldr/pull/977/files/071da22125918034aba0c10592e2dc751f8c87f2#r72039508>,
or mute the thread

waldyrious: Yes, I recall that discussion (#555). Let's keep two pages then.

ps - this is the third time these small differences in dd raise confusion. A "notes" section (c.f. #481) would probably prevent this sort of common pitfalls.

dylanrees:
Thanks for the input!

PR #981, opened 2016-07-26
CONTRIBUTING.md: Add token conventions

agnivade:
xRef-  #478

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,9 +71,7 @@ Open all the files of a given extension in the current directory with the associ
 open {{*.ext}}
 ```
 
-Its clear that `{{.ext}}` helps to clarify the command. But here, using `{{file}}` makes more sense
-
-`wc -l {{file}}`
+Its clear that `{{.ext}}` helps to clarify the command. But here - `wc -l {{file}}`, using `{{file}}` makes more sense
 
 - Lastly, follow the `{{/path/to}}` convention when there is a path related command. Not when the file location is implicit. But of course, use proper judgement, keeping simplicity and user friendliness as the top priority.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,23 +53,59 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
 [`tldr tldrl`](https://github.com/tldr-pages/tldr/blob/master/pages/common/tldrl.md)
 
 ### Token syntax
-User-provided values should use the `{{token}}` syntax in order to allow clients
-to highlight them. 
+User-provided values should use the `{{token}}` syntax
+in order to allow clients to highlight them.
+Stick to [`snake_case`](https://en.wikipedia.org/wiki/Snake_case)
+when there are more than a word in the token.
 
-Some examples: 
-- `tar cf {{file}}`
-- `ln -s {{path/to/original/file}} {{path/to/link}}`
-- `mysql {{database_name}}`
-- `unrar x {{compressed.rar}}`
+Some conventions that are followed:
+
+- If the example is clearer with an actual value rather than a generic placeholder, use the actual value
+
+`iostat {{2}}` :white_check_mark:
+
+`iostat {{interval_in_secs}}` :x:
+
+- For any reference to a path or folder, use this
+
+`ln -s {{path/to/file}} {{path/to/symlink}}`
+
+- In case of a possible reference both to a file or a folder, use this
+
+`atom {{path/to/file_or_folder}}`
+
+- If a command expects the file to be in a particular extension, use it
+
+`unrar x {{compressed.rar}}`
+
+- If the example needs a generic extension, use `{{.ext}}`, but **only** if it helps to clarify the command.
+
+```
+Open all the files of a given extension in the current directory with the associated application:
+open {{*.ext}}
+```
+
+Here, its clear that `{{.ext}}` helps to clarify the command. But here, using `{{file}}` makes more sense
+
+`wc -l {{file}}` :white_check_mark:
+
+`wc -l {{file.txt}}` :x:
+
+- If multiple extensions are allowed, use this format
+
+`sass {{inputfile.scss|inputfile.sass}}`
+
+- Lastly, follow the `{{/path/to}}` convention when there is a path related command. Not when the file location is implicit.
+
+`cat {{file}}` :white_check_mark:
+
+`cat {{path/to/file}}` :x:
+
+But of course, use proper judgement, keeping simplicity and user friendliness as the top priority.
 
 In short, make it as intuitive as possible
 for the user to figure out how to use the command
 and fill it in with values.
-
-Stick to [`snake_case`](https://en.wikipedia.org/wiki/Snake_case) where possible.
-In some situations a command works with typical file extensions
-(like the `unrar` example above);
-you are encouraged to add these for demonstration.
 
 ## Submitting a pull request
 
@@ -87,20 +123,20 @@ Alternatively, you can do most of the process using the command line:
 
 - fork the repository on the github web interface
 
-- clone your fork locally:  
+- clone your fork locally:
   `git clone https://github.com/{{your_username}}/tldr.git && cd tldr`
 
-- create a feature branch, e.g. named after the command you plan to edit:  
+- create a feature branch, e.g. named after the command you plan to edit:
   `git checkout -b {{branch_name}}`
 
 - make your changes (edit existing files or create a new one)
 
-- commit the changes:  
+- commit the changes:
   `git commit --all -m "{{commit_message}}"`
-    
-- push to your fork:  
+
+- push to your fork:
   `git push`
-    
+
 - go to the github page for your fork and click the green pull request button.
 
 Please send only related changes in the same pull request.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,22 +58,22 @@ in order to allow clients to highlight them.
 Use [`snake_case`](https://en.wikipedia.org/wiki/Snake_case)
 for multi-word tokens.
 
-Keep the following guidelines in mind when writing the tokens:
+Keep the following guidelines in mind when choosing token names:
 
-- If the example is clearer with an actual value rather than a generic placeholder, use the actual value. For example - use `iostat {{2}}` rather than `iostat {{interval_in_secs}}`.
+- If the example is clearer with an actual value rather than a generic placeholder, use the actual value. For example, use `iostat {{2}}` rather than `iostat {{interval_in_secs}}`.
 
-- For any reference to a path or folder, use `{{path/to/<placeholder>}}`. For example - `ln -s {{path/to/file}} {{path/to/symlink}}`. In case of a possible reference both to a file or a folder, use `{{path/to/file_or_folder}}`
+- For any reference to paths to files or folders, use `{{path/to/<placeholder>}}`. For example - `ln -s {{path/to/file}} {{path/to/symlink}}`. In case of a possible reference both to a file or a folder, use `{{path/to/file_or_folder}}`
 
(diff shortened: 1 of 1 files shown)


ostera:
Looking good to me! 👍

waldyrious:
I actually think it's a little verbose. I can't give detailed commentary right now, but please hold off merging for a bit.

agnivade:
@waldyrious - ping :)

waldyrious:
I didn't forget -- just tackling smaller stuff in the pockets of time I find throughout the day :) I will likely tackle some more time-consuming discussions (here and elsewhere) at night during the workweek or in the weekend.

agnivade:
hehe .. I thought so. No worries :)

CONTRIBUTING.md
```diff
@@ -53,24 +53,60 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
 [`tldr tldrl`](https://github.com/tldr-pages/tldr/blob/master/pages/common/tldrl.md)
 
 ### Token syntax
-User-provided values should use the `{{token}}` syntax in order to allow clients
-to highlight them. 
+User-provided values should use the `{{token}}` syntax
+in order to allow clients to highlight them.
+Stick to [`snake_case`](https://en.wikipedia.org/wiki/Snake_case)
+when there are more than a word in the token.
```
rubenvereecken: You can remove this line, it's implied.

rubenvereecken:
This looks really daunting in size to new contributors

agnivade:
@rubenvereecken - I agree with you. But at the same time, there should be a place where these rules should be mentioned. That's what the original issue was about. I believe some rules can be omitted if they exist to serve only a edge case and can be enforced by team members in the PR.

I would ideally love if some of these rules can be automatically enforced. (Maybe they can !)

rubenvereecken:
@agnivade Ah thanks, I did miss out on the original discussion. Yeah, we can easily add enforcement in the linter. Even automatically format stuff.

Anyone familiar with Yacc and Flex? That's what it's written in - or at least, in a JS bastard version.

rubenvereecken:
Yeah I slipped there.

CONTRIBUTING.md
waldyrious: I agree it's more verbose than needed, but I like the explicitness. How about "use snake_case for multi-word tokens" instead?

CONTRIBUTING.md
```diff
@@ -87,20 +123,20 @@ Alternatively, you can do most of the process using the command line:
 
 - fork the repository on the github web interface
 
-- clone your fork locally:  
+- clone your fork locally:
```
waldyrious: Please revert the changes from this line onward. The double space at the end of a line is a markdown "feature" (albeit a poorly planned one, I admit) to insert a line break without starting a new paragraph. Keeping commands in their own lines improves readability and facilitates copy-pasting.

Also, it's not good practice to make unrelated changes in the same PR/commit (although I admit I'm often guilty of it!)

CONTRIBUTING.md
```diff
@@ -53,24 +53,60 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
...
+in order to allow clients to highlight them.
+Stick to [`snake_case`](https://en.wikipedia.org/wiki/Snake_case)
+when there are more than a word in the token.
 
-Some examples: 
-- `tar cf {{file}}`
-- `ln -s {{path/to/original/file}} {{path/to/link}}`
-- `mysql {{database_name}}`
-- `unrar x {{compressed.rar}}`
+Some conventions that are followed:
```
waldyrious: Sugestion: "Keep the following guidelines in mind when choosing tokens:"

waldyrious:
I've added a couple inline comments, but my main objection to this PR's current form is the overall length.

I would prefer at least keeping each guideline in a single bullet-point paragraph -- i.e., integrate the good/bad examples into the prose, as in "blah blah blah; for example `{{good_token}}` instead of `{{bad_token}}`".

This would make the page less lengthly and intimidating ("ugh, so many rules!..."), IMHO.

Also, I would ditch the cross/checkmark icons.

CONTRIBUTING.md
agnivade: Blame my editor for it. :stuck_out_tongue_closed_eyes: Its configured to remove trailing spaces at end of lines. 

Thanks for pointing out the feature.

agnivade:
I'll see what I can do. Too many points are indeed intimidating. I'll maybe have the most important ones in a prosaic format, and skip out the rare ones.

CONTRIBUTING.md
```diff
@@ -53,24 +53,32 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
...
+in order to allow clients to highlight them.
+Use [`snake_case`](https://en.wikipedia.org/wiki/Snake_case)
+for multi-word tokens.
 
-Some examples: 
-- `tar cf {{file}}`
-- `ln -s {{path/to/original/file}} {{path/to/link}}`
-- `mysql {{database_name}}`
-- `unrar x {{compressed.rar}}`
+Keep the following guidelines in mind when writing the tokens:
```
waldyrious: I think "when choosing tokens" or "when choosing token names" works better than "writing the tokens".

CONTRIBUTING.md
```diff
@@ -53,24 +53,32 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
 [`tldr tldrl`](https://github.com/tldr-pages/tldr/blob/master/pages/common/tldrl.md)
 
 ### Token syntax
-User-provided values should use the `{{token}}` syntax in order to allow clients
-to highlight them. 
+User-provided values should use the `{{token}}` syntax
+in order to allow clients to highlight them.
```
waldyrious: 👍 for semantic line-breaking :D

CONTRIBUTING.md
```diff
@@ -53,24 +53,32 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
...
+for multi-word tokens.
 
-Some examples: 
-- `tar cf {{file}}`
-- `ln -s {{path/to/original/file}} {{path/to/link}}`
-- `mysql {{database_name}}`
-- `unrar x {{compressed.rar}}`
+Keep the following guidelines in mind when writing the tokens:
+
+- If the example is clearer with an actual value rather than a generic placeholder, use the actual value. For example - use `iostat {{2}}` rather than `iostat {{interval_in_secs}}`.
```
waldyrious: Suggestion: `For example - use`--> `For example, use`

CONTRIBUTING.md
```diff
@@ -53,24 +53,32 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
...
+
+- For any reference to a path or folder, use `{{path/to/<placeholder>}}`. For example - `ln -s {{path/to/file}} {{path/to/symlink}}`. In case of a possible reference both to a file or a folder, use `{{path/to/file_or_folder}}`
```
waldyrious: How about `For any reference to a path or folder` --> `For references to paths to files or folders`?

CONTRIBUTING.md
```diff
@@ -53,24 +53,32 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
...
+
+- If a command expects the file to be in a particular extension, use it. For example - `unrar x {{compressed.rar}}`. In case needs a generic extension, use `{{.ext}}`, but **only** if it helps to clarify the command. For example, here -
```
waldyrious: `expects the file to be in a particular extension` --> `expects the file to have a particular extension`

CONTRIBUTING.md
```diff
@@ -53,24 +53,32 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
...
+
+```
```
waldyrious: I think you can make this block part of the list item it belongs to, if you indent it. Try editing in the browser and using the preview function :)

CONTRIBUTING.md
```diff
@@ -53,24 +53,32 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
...
+Open all the files of a given extension in the current directory with the associated application:
+open {{*.ext}}
+```
+
+Its clear that `{{.ext}}` helps to clarify the command. But here - `wc -l {{file}}`, using `{{file}}` makes more sense
```
waldyrious: `Its` --> `It's`

waldyrious: Sugestion: "But here - `wc -l {{file}}`, using `{{file}}` makes more sense" --> "But in a command like `wc -l {{file}}`, using `{{file}}` is sufficiently clear."

waldyrious: "clear (...) clarify" <-- rephrase to remove redundancy

CONTRIBUTING.md
```diff
@@ -53,24 +53,32 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
...
+
+- Lastly, follow the `{{/path/to}}` convention when there is a path related command. Not when the file location is implicit. But of course, use proper judgement, keeping simplicity and user friendliness as the top priority.
```
waldyrious: "path related" --> "path-related"

CONTRIBUTING.md
```diff
@@ -53,24 +53,32 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
...
-Some examples: 
-- `tar cf {{file}}`
-- `ln -s {{path/to/original/file}} {{path/to/link}}`
-- `mysql {{database_name}}`
-- `unrar x {{compressed.rar}}`
+Keep the following guidelines in mind when choosing token names:
+
+- If the example is clearer with an actual value rather than a generic placeholder, use the actual value. For example, use `iostat {{2}}` rather than `iostat {{interval_in_secs}}`.
+
+- For any reference to paths to files or folders, use `{{path/to/<placeholder>}}`. For example - `ln -s {{path/to/file}} {{path/to/symlink}}`. In case of a possible reference both to a file or a folder, use `{{path/to/file_or_folder}}`
```
waldyrious: Same suggestion as before: `For example - ...` --> `For example, ...`, or `For example: ...`. Although a dash indeed marks a pause in speech, the semantic meaning is different than the pause caused by a colon or comma.

CONTRIBUTING.md
```diff
@@ -53,24 +53,32 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
...
+
+- If a command expects the file to have a particular extension, use it. For example - `unrar x {{compressed.rar}}`. In case needs a generic extension, use `{{.ext}}`, but **only** if it helps to clarify the command. For example, here -
```
waldyrious: same as above.

CONTRIBUTING.md
```diff
@@ -53,24 +53,32 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
...
+
+  ```
+  Open all the files of a given extension in the current directory with the associated application:
+  open {{*.ext}}
+  ```
+
+  Here, using `{{.ext}}` helps to clarify the command. But in a command like `wc -l {{file}}`, using `{{file}}` is sufficiently clear.
```
waldyrious: "here" is duplicated. Maybe you could start the line with "...", since it's the continuation of the sentence before the block?

CONTRIBUTING.md
```diff
@@ -53,24 +53,32 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
...
+
+- Lastly, follow the `{{/path/to}}` convention when there is a path-related command. Not when the file location is implicit. But of course, use proper judgement, keeping simplicity and user friendliness as the top priority.
```
waldyrious: I'd suggest using `path/to/<placeholder>` here as well, for consistency with the one above. Also, the path convention usually starts without a `/`, since the most common case is that the path doesn't have to be absolute. (this also makes it consistent with what's used above)

waldyrious:
:+1: lgtm :)

PR #982, opened 2016-08-01
psql: Improve page

agnivade:
- Removed * , this was causing problems with rendering. The node client was
gobbling up the entire word !
- Changed 'query' to 'command' as it can be a postgres db command too.
- Used the -f flag for running multiple commands as its a better and scalable
option.

### DIFF
--- a/pages/common/psql.md
+++ b/pages/common/psql.md
@@ -2,7 +2,7 @@
 
 > PostgreSQL command-line client.
 
-- Connect to database. It connects to localhost using default port 5432 with default user:
+- Connect to database. It connects to localhost using default port 5432 with default user as currently logged in user:
 
 `psql {{database}}`
 
@@ -14,10 +14,10 @@
 
 `psql -h {{host}} -p {{port}} -U {{username}} -W {{database}}`
 
-- Execute a single command (sql or PostgreSQL-specific) against the given database. Note: useful in shell scripts:
+- Execute a single SQL query or PostgreSQL command on the given database. Note: useful in shell scripts:
 
 `psql -c '{{query}}' {{database}}`
 
-- Execute commands from a file against the given database:
+- Execute commands from a file on the given database:
 
 `psql {{database}} -f {{file.sql}}`
--- a/pages/common/psql.md
+++ b/pages/common/psql.md
@@ -2,22 +2,22 @@
 
 > PostgreSQL command-line client.
 
-- Connect to *database*. It connects to localhost using default port *5432* with default user:
+- Connect to database. It connects to localhost using default port 5432 with default user:
 
 `psql {{database}}`
 
-- Connect to *database* on given server *host* running on given *port* with *username* given, no password prompt:
+- Connect to database on given server host running on given port with username given, no password prompt:
 
 `psql -h {{host}} -p {{port}} -U {{username}} {{database}}`
 
-- Connect to *database*, user will be prompted for password:
+- Connect to database, user will be prompted for password:
 
 `psql -h {{host}} -p {{port}} -U {{username}} -W {{database}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/psql.md
```diff
@@ -2,22 +2,22 @@
 
 > PostgreSQL command-line client.
 
-- Connect to *database*. It connects to localhost using default port *5432* with default user:
+- Connect to database. It connects to localhost using default port 5432 with default user:
 
 `psql {{database}}`
 
-- Connect to *database* on given server *host* running on given *port* with *username* given, no password prompt:
+- Connect to database on given server host running on given port with username given, no password prompt:
```
waldyrious: "username given" --> "given username" (more natural, and matches the other arguments to this command).

Also, "no password prompt" --> "without a password prompt".

pages/common/psql.md
```diff
@@ -2,22 +2,22 @@
...
 
 `psql -h {{host}} -p {{port}} -U {{username}} {{database}}`
 
-- Connect to *database*, user will be prompted for password:
+- Connect to database, user will be prompted for password:
```
waldyrious: use a semicolon [instead of a comma](https://en.wikipedia.org/wiki/Run-on_sentence), since these are separate clauses.

pages/common/psql.md
```diff
@@ -2,22 +2,22 @@
...
 
 `psql -h {{host}} -p {{port}} -U {{username}} -W {{database}}`
 
-- Run single *query* against the given *database*. Note: useful in shell scripts:
+- Execute a single command(sql or internal) against the given database. Note: useful in shell scripts:
```
waldyrious: Missing space before the parenthesis. Also, what does "internal" mean here? perhaps "PostgreSQL-specific" would be clearer?

agnivade: agree.

pages/common/psql.md
```diff
@@ -2,22 +2,22 @@
 
 > PostgreSQL command-line client.
 
-- Connect to *database*. It connects to localhost using default port *5432* with default user:
+- Connect to database. It connects to localhost using default port 5432 with default user:
```
waldyrious: what is the default user?

pages/common/psql.md
```diff
@@ -2,22 +2,22 @@
...
 
 `psql -h {{host}} -p {{port}} -U {{username}} {{database}}`
 
-- Connect to *database*, user will be prompted for password:
+- Connect to database; user will be prompted for password:
 
 `psql -h {{host}} -p {{port}} -U {{username}} -W {{database}}`
 
-- Run single *query* against the given *database*. Note: useful in shell scripts:
+- Execute a single command (sql or PostgreSQL-specific) against the given database. Note: useful in shell scripts:
```
waldyrious: How about "Execute a single SQL query or PostgreSQL command (...)"?

Also, I would replace "against" with "on".

pages/common/psql.md
```diff
@@ -2,22 +2,22 @@
...
 
 `psql -c '{{query}}' {{database}}`
 
-- Run several queries against the given *database*. Note: useful in shell scripts:
+- Execute commands from a file against the given database:
```
waldyrious: same here. "against" may be idiomatic, but it's unnecessarily so. "on" works well and is less likely to confuse absolute beginners.

pages/common/psql.md
agnivade: The current user logged in to the machine.

waldyrious: Should we add this detail as a note in parenthesis?

waldyrious:
LGTM 👍 thanks for bearing with the nitpicking!

PR #983, opened 2016-08-01
repquota: add page

pearcem0:


pearcem0:
Actually, I think this should go in pages/common rather than pages/linux.

waldyrious:
Not sure. On OS X I get:

```
$ repquota -all
repquota: illegal option -- l
Usage:
	repquota [-v] [-g] [-u] -a
	repquota [-v] [-g] [-u] filesys ...
```

There are likely other inconsistencies as well. The version I have is:
`4.2 Berkeley Distribution       March 28, 2002       4.2 Berkeley Distribution`

Probably checking the BSD and GNU manpages for this and the other quota-related commands will reveal which can go on `common` and which need to be split into different platform versions.

pearcem0:
You're right, this will need to be classed as Linux after all. Thanks

pearcem0:
Hi, 
Is there anything else you need me to do for this pr?
Thanks,

pages/linux/repquota.md
```diff
@@ -0,0 +1,27 @@
+# repquota
+
+> Display a summary of existing file quotas for a filesystem.
+
+- Report on all existing quotas:
+
+`repquota -all`
+
+- Report  all quotas, even if there is no usage:
```
ostera: Double space between `Report` and `all`

pages/linux/repquota.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Report on quotas for users (default), only super-user may view quotas which are not their own:
+
+`repquota --user {{user}}`
+
+- Report on quotas for user groups:
+
+`repquota --group {{group}}`
+
+- Report on used space and limits in a human readable format:
```
ostera: I think elsewhere we've used `in human-readable format`. Saves a few bytes and maintains consistency. I'll search the repo and get back to you on this.

ostera: Yup: https://github.com/tldr-pages/tldr/search?utf8=%E2%9C%93&q=human-readable

Dashed.

ostera:
Hey, just left a few comments :) other than that it looks good. Fix them and it has my :+1:

pearcem0:
Thanks a lot. Made the changes.

pages/linux/repquota.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Report on all existing quotas:
+
+`repquota -all`
+
+- Report all quotas, even if there is no usage:
+
+`repquota -v`
+
+- Report on quotas for users (default), only super-user may view quotas which are not their own:
```
waldyrious: This should be a semicolon, not a comma (see https://en.wikipedia.org/wiki/Run-on_sentence). Also, it's unclear what "for users (default)" means... care to clarify?

pages/linux/repquota.md
```diff
@@ -0,0 +1,27 @@
+# repquota
+
+> Display a summary of existing file quotas for a filesystem.
+
+- Report on all existing quotas:
+
+`repquota -all`
+
+- Report all quotas, even if there is no usage:
```
waldyrious: Do you mean "even if there are no quotas defined", or "even if quotas are defined, but not used" (not sure the latter actually makes sense)?

pages/linux/repquota.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Report on quotas for users (default), only super-user may view quotas which are not their own:
+
+`repquota --user {{user}}`
+
+- Report on quotas for user groups:
+
+`repquota --group {{group}}`
+
+- Report on used space and limits in a human-readable format:
```
waldyrious: I'd be specific here and say "used quota space", or just "used quota"

pages/linux/repquota.md
pearcem0: How about - 
'Report on quotas for all users (unless specified); only super-user may view quotas which are not their own:' ?

pages/linux/repquota.md
pearcem0: okay, will do - Thanks!

pages/linux/repquota.md
waldyrious: So this command gives an error if not run with sudo and no users are specified? How about including sudo in its invocation?

agnivade:
Hi @pearcem0, any updates on this ?

pages/linux/repquota.md
pearcem0: When I say default, I mean running sudo repquota {{user}} will do the same thing as sudo repquota -u {{user}}. 
Okay, how about having two entries, one for reporting quota on your own user, and one for reporting quota on other users (as root) - 

- Report on quotas for current user:

`repquota --user {{current_user}}`

- Report on quotas for another user:

`sudo repquota --user {{another_user}}`

- Report on quotas for user groups:

`sudo repquota --group {{group}}`

I also added sudo for group. 

The command itself is generally an administrators tool, only the super-user may view quotas which are not their own. Running without sudo generates a permissions error. 
Should I add sudo to all of the commands?

pages/linux/repquota.md
pearcem0: Report quota, even for users those who do not consume resources. How about - 
Verbose mode - report all quotas, even for users those who do not consume resources:

pearcem0:
Hi @agnivade - So sorry for the delay. I've added some comments in reply to waldyrious so that we can continue.

pages/linux/repquota.md
waldyrious: So the quotas reported won't be the limits for each user, but actual usages that they make of their quota? Or both?

pages/linux/repquota.md
waldyrious: I think it makes sense to separate different usages, yes, and to add sudo to all commands that require it (except those related to viewing one's own quota)

pages/linux/repquota.md
pearcem0: Both.  Used (usage), soft limit, hard limit, time left on the quota timer etc.

waldyrious: Ok, so how about "Report quota stats for all users, even those who aren't using any of their quota"?

pages/linux/repquota.md
```diff
@@ -0,0 +1,27 @@
+# repquota
+
+> Display a summary of existing file quotas for a filesystem.
+
+- Report on all existing quotas:
```
waldyrious: In line with the suggested description below, this could be "Report stats for all quotas in use:"?

waldyrious:
Alright! I think we've given you enough headaches @pearcem0 😄 thanks so much for your patience! LGTM 👍

PR #984, opened 2016-08-01
quota: improve page

pearcem0:
Added information about how to enable quota for file system.

### DIFF
--- a/pages/common/quota.md
+++ b/pages/common/quota.md
@@ -1,6 +1,7 @@
 # quota
 
 > Display users' disk space usage and allocated limits.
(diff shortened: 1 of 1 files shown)


waldyrious:
Thanks! One question: is this location the same across different platforms?

pearcem0:
Good point, I see now that OSX also requires a .quota.ops.user file, which goes further out of the scope of tldr. I think it is best you decline the pull request and close?

waldyrious:
you can update the PR branch, and at the merge step we'll squash the changes so they'll look like a single clean commit. If you prefer opening another PR, that's fine too :)

PR #989, opened 2016-08-03
Add avrdude page

alx741:
Atmel AVR microcontrollers programming [libre] software


waldyrious:
Thanks! LGTM :+1:

agnivade:
Not to be a pedant, but it would be great if we can format the commit title to the standard -
`avrdude: Add page`

waldyrious:
@agnivade that can be changed in the merge step by using the "squash" workflow :) but thanks for pointing it out so that isn't overlooked.

Issue #190, opened 2014-04-23
New pages structure


rprieto:
_(copied form the conversation at #147)_

The current folder structure looks like:

```
pages
  |__ common
  |   |__ tar.md         # gnu command everyone should have
  |   |__ ssh.md         # gnu command everyone should have
  |   |__ npm.md         # just a tool that could be installed on any OS
  |__ linux
  |   |__ emerge.md      # clearly a linux-only tool
  |__ osx
      |__ ssh.md         #  OSX has a different version with different flags
```

This has worked so far, but
- the split can feel arbitrary
- you need to jump around to see if a command is already covered
- clients potentially need to make a lot of requests to get a given page

There seems to be a consensus among clients to move to a flatter folder structure, that would look like this:

```
pages
   |__ tar.md
   |__ ssh.md
   |__ ssh.osx.md
   |__ emerge.md
   |__ npm.md
```

By default, we can display `<command>.md`, but `<command>.<os>.md` should have precedence if available.

This means the clients would let OSX users query Linux commands, but after all why not, they might just be curious about it. Or they might be using `tldr` on a Mac while SSHing on a Linux box that doesn't have it.

To clear up platform constraints, the description for `emerge` for example could say "for Linux" or "Gentoo specific command".

This would be a breaking change though, so we need a plan of attack. One option is to address all open PRs to get to a stable state, then **copy** all pages to the new structure. The old structure can live on for a while until all clients update. The PR guidelines would say to push changes to the new structure only.

M3kH:
Can this done by the metadata syntax on MarkDown? maybe as additional to avoid duplicated files on categories/tags?
```
---
support:
 - osx
 - linux
 -- debian
...
---
```

igorshubovych:
@M3kH in the end of this optimization process we will build man :)

M3kH:
I would later prefer store informations in the markdown as `yml` instead of using the file names.
But this is my personal opinion :-)

waldyrious:
Quoting a comment by @agnivade in #1436, which replaced all references to OS X with the new name, macOS:

> Just a note that our platform folder is still named `osx`. And changing that might have consequences for various clients, so keeping it as is for now.

ibnesayeed:
[...]

PR #990, opened 2016-08-03
tree.md: add more useful examples

waldyrious:
and reorder entries by increasing complexity

### DIFF
--- a/pages/common/tree.md
+++ b/pages/common/tree.md
@@ -2,7 +2,7 @@
 
 > Show the contents of the current directory as a tree.
 
-- Show files and directories with a depth of 'num' (where 1 means the current directory):
+- Show files and directories up to 'num' levels of depth (where 1 means the current directory):
 
 `tree -L {{num}}`
 
--- a/pages/common/tree.md
+++ b/pages/common/tree.md
@@ -14,14 +14,14 @@
 
 `tree -a`
 
-- Print human readable size of files:
+- Print the tree without indentation lines, showing the full path instead:
 
-`tree -h`
+`tree -i -f`
 
-- Print the full path for each file:
+- Print the size of each node next to it, in human-readable format, with folders displaying their cumulative size (as in the `du` command):
 
-`tree -f`
+`tree -s -h --du`
 
-- Print the tree without lines and indentation. Useful when used with -f:
+- Filter the tree using a wildcard (glob) pattern, showing only the hierarchies leading to the matching entries:
 
-`tree -i`
+`tree -P {{*.txt}} --prune`
--- a/pages/linux/tree.md
+++ b/pages/linux/tree.md
@@ -18,10 +18,10 @@
 
 `tree -i -f`
 
-- Print the size of each node next to it, in human-readable format, with folders displaying their cumulative size (as in the `du` command):
+- Print the size of each node next to it, in human-readable format:
 
-`tree -s -h --du`
+`tree -s -h`
 
-- Filter the tree using a wildcard (glob) pattern, showing only the hierarchies leading to the matching entries:
+- Filter the tree using a wildcard (glob) pattern:
 
-`tree -P {{*.txt}} --prune`
+`tree -P {{*.txt}}`
--- a/pages/osx/tree.md
+++ b/pages/osx/tree.md
@@ -1,0 +1,31 @@
+# tree
+
+> Show the contents of the current directory as a tree.
+
+- Show files and directories up to 'num' levels of depth (where 1 means the current directory):
+
+`tree -L {{num}}`
+
+- Show directories only:
+
+`tree -d`
+
+- Show hidden files too:
+
+`tree -a`
+
+- Print the tree without indentation lines, showing the full path instead:
+
+`tree -i -f`
+
+- Print the size of each node next to it, in human-readable format, with folders displaying their cumulative size (as in the `du` command):
+
+`tree -s -h --du`
+
+- Find files within the tree hierarchy, using a wildcard (glob) pattern, and pruning out directories that don't contain matching files:
+
+`tree -P {{*.txt}} --prune`
+
+- Find directories within the tree hierarchy, pruning out directories that aren't ancestors of the wanted one:
+
+`tree -P {{directory_name}} --matchdirs --prune`


pages/common/tree.md
```diff
@@ -14,14 +14,14 @@
...
 `tree -a`
 
-- Print human readable size of files:
+- Print the tree without indentation lines, showing the full path instead:
 
-`tree -h`
+`tree -i -f`
 
-- Print the full path for each file:
+- Print the size of each node next to it, in human-readable format, with folders displaying their cumulative size (as in the `du` command):
```
agnivade: This feels like too many options in a single command. Wonder if we can cut down on one of them.

waldyrious: We could remove the --du option, but that makes the output less practical. I wonder if what makes it seem complex is the length of the description. Would something like this be better?

"Print the human-readable size of each node next to it (with folders displaying their cumulative size, as in the `du` command):"

waldyrious:
The -P option needs the -a to match dotfiles, and --matchdirs to match "dotdirs" like ".git".

I wonder if we could add `tree -a -P '.git' --prune --matchdirs` to fill this gap. THoughts?

pages/common/tree.md
```diff
@@ -14,14 +14,14 @@
...
 
-`tree -f`
+`tree -s -h --du`
 
-- Print the tree without lines and indentation. Useful when used with -f:
+- Filter the tree using a wildcard (glob) pattern, showing only the hierarchies leading to the matching entries:
 
-`tree -i`
+`tree -P {{*.txt}} --prune`
```
waldyrious: The -P option needs `-a` to match dotfiles, and --matchdirs to match "dotdirs" like ".git".

I wonder if we could add `tree -a -P '.git' --prune --matchdirs` to fill this gap, building on the simpler `-P` example proposed here. Thoughts?

pages/common/tree.md
agnivade: Its not the length of the description. Although a shorter description does have lesser cognitive load.

But yea .. I agree the 3 options really need to go together. In that case, your revised statement certainly looks better.

pages/common/tree.md
agnivade: This feels like straying too far from the 'most common usage' motto that we follow. This example seems sufficiently complex enough to be a last example. IMO, we don't need anything more complicated than this.

waldyrious: Fair enough, you've got a good point. I'll keep this particular invocation in my personal notes, though :)

pages/common/tree.md
waldyrious: Since you seem unsure still, I'll wait for someone else to chime in :)

agnivade: You got me there :sweat_smile:

ostera: I think we're good with `Print the full path for each file with human-readable size`. Anything other than that seems like over-explaining to me :'D

pages/common/tree.md
ostera: Ah, this is a good one. I typically end up `tree -a | grep -v`-ing out the things I don't want to see.

I'd stick with the simple and just use `-P {{pattern}}` and let people figure out the rest.

`--prune` seems excessive. Why filter out empty directories in this general an example? They won't hurt if `--matchdirs` isn't set as they contain nothing to match on.

`--matchdirs` is probably relevant to this pattern matching sample, but makes the command more daunting to deal with at first.

In my book, `-a` means `anarchic`. But most likely you're not trying to look for things inside `.git`. If you are, and you know of this folder, you likely know it's a _hidden_ folder and that it'll be ignored by default by most commands that follow this convention. 

I'd stick with the simple.

pages/common/tree.md
waldyrious: @ostera you probably missed that the actual command changed :) I reordered the commands to keep them sorted by complexity. Please check the full diff.

pages/common/tree.md
waldyrious: Good points, I actually meant to add `--matchdirs` here. But we could make this just about finding files within the tree, and make a separate `-P` example with `--matchdirs` and `--prune`, which I think is pretty useful too. That way we have both options, and the complexity doesn't suddenly jump up :D

pages/common/tree.md
ostera: @waldyrious oh my! You're right, completely missed the command change.

pages/common/tree.md
waldyrious: @agnivade @ostera actually, the `--prune` option is not implemented in linux's `tree`. Per the [manpage](http://linux.die.net/man/1/tree):

> ### Bugs

> Tree does not prune "empty" directories when the -P and -I options are used. Tree prints directories as it comes to them, so cannot accumulate information on files and directories beneath the directory it is printing.

That page also shows that the `--du` option does not exist for the linux version.

So I guess I'll change this entry to be simply `tree -P {{*.txt}}` and the one before it to be `tree -s -h`, then move the page to linux/tree.md, and create a separate osx/tree.md for the behavior including `--prune` and `--du`. Sounds good?

agnivade: Huh .. I wonder why this is so .. Anyway, SGTM.

waldyrious:
@ostera @agnivade PTAL

agnivade:
LGTM overall. Are we sure, this command isn't there for SunOS ?

ostera:
I have an idea for _extending_ the `common` page with platform-specific commands...I think we should talk about this before duplicating data.

waldyrious:
@ostera that's literally our oldest open issue (#190). I fully trust that we could come up with a nice solution in a few sessions of intensive brainstorming and implementation, but considering that issue's history and the need to update lots of clients over which we have no direct control, forgive me for not holding my breath :) I'll help with moving that forward, but would very much like to have this merged as I need to look up the syntax for some of these `tree` options all the time :P

waldyrious:
@agnivade, we (well, I) don't know. We can't really enforce that verification when most of our contributors don't have access to SunOS machines. Ideally we'd be able to set up different environments for CI, but from a cursory look Travis [only](https://docs.travis-ci.com/user/ci-environment/) offers [OS X and Linux (Ubuntu)](https://docs.travis-ci.com/user/multi-os/) images, and CircleCI [only has Linux (Ubuntu) ones](https://circleci.com/docs/build-images/).

That does give me an idea, though -- I wonder if we could set up automated tests for linux and osx, to verify that a command exists in a system (`type command`) and whether the command supports the options in the examples (something like `grep --foo > /dev/null 2>&1 || echo "fail"`?). But I'm not sure this could work in general. Just an idea :)

ostera:
@waldyrious haha I had forgotten about that! Then ignore me for now.

agnivade:
> We can't really enforce that verification when most of our contributors don't have access to SunOS machines.
I understand. No issues, lets move forward.

PR #991, opened 2016-08-03
osx/du.md: fix depth option, plus minor tweaks

waldyrious:
The version I have uses `-d` for the depth option, not `-depth`. Here's the version string as shown by `man du`:
```
BSD                              June 2, 2004                              BSD
```

In addition to this fix, this commit changes the tokens to clarify that they are paths, and changes "directory" to "folder", which is shorter and more accessible.

### DIFF
--- a/pages/osx/du.md
+++ b/pages/osx/du.md
@@ -4,16 +4,16 @@
 
 - Get a sum of the total size of a file/folder in human readable units:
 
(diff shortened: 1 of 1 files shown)


agnivade:
LGTM.

felixonmars:
:+1:

PR #993, opened 2016-08-03
osx/locate.md: add exact filename match

waldyrious:


agnivade:
Huh .. what an odd way to achieve this. They could have just added a `--exact` flag or sth and be done with it. Anyway, LGTM.

waldyrious:
yeah, this is definitely a workaround to an awkward feature of the command. I believe that behavior may be for historical compatibility, or maybe the creators assumed the *foo* usecase would be more common than "foo".

PR #994, opened 2016-08-05
mongo: Add page

agnivade:


pages/common/mongo.md
```diff
@@ -0,0 +1,19 @@
+# mongo
+
+> MongoDB shell client.
```
waldyrious: suggestion: shell --> command line

pages/common/mongo.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Connect to a database:
+
+`mongo {{database}}`
+
+- Connect to a database running on a given host on a given port:
+
+`mongo --host {{host}} --port {{port}} {{database}}`
+
+- Connect to a database with a given username, user will be prompted for password:
```
waldyrious: [semicolon](https://en.wikipedia.org/wiki/Run-on_sentence)

pages/common/mongo.md
```diff
@@ -0,0 +1,19 @@
...
+
+`mongo --username {{username}} {{database}} --password`
+
+- Run queries on the database:
```
waldyrious: "Run a query"?

pages/common/mongo.md
agnivade: Yea I had considered this. I think shell is a better choice here. Mainly because this actually creates a javascript shell environment wherein you can execute JS code apart from your usual DB queries. Even the man pages and official docs refer to it as a "shell" client.

pages/common/mongo.md
agnivade: Not really. You can actually evaluate complete javascript expressions which can run multiple queries. I agree `query` might be a bit misleading here. I thought of `javascript`, but that is too vague.

Open to other options here.

pages/common/mongo.md
waldyrious: Ah, that changes everything 😄 in that case, I would suggest "MongoDB interactive shell client", "Mongo REPL", or something like that. Just "shell client" makes it seem like it's a regular command-line client for the shell (whichever one you're using).

pages/common/mongo.md
waldyrious: {{javascript_expression(s)}}?

rubenvereecken: `query` is as close to the truth as you're going to get, it's a JS object with limited syntax

waldyrious: Then maybe we can clarify it in the command description instead...

agnivade: I think the concept of a "javascript expression" as a query is a bit hard to understand to a newbie. So, gave an example instead. And removed "query" from the description.

waldyrious:
LGTM 😄

PR #465, opened 2015-12-29
vim: add :w, :wq, i

shoeletz:
Vim commands for save, save then quit, and insert mode

PR #926, opened 2016-06-24
Updates vim and vimtutor pages

ostera:
Hi there fellow tldrers,

I took a shot at the `vim` and `vimtutor` pages that more or less made sense to me as a heavy vim user.

Thoughts? Comments? Rants? All welcome!

PR #996, opened 2016-08-06
vim: add :w, :wq, i (take II)

waldyrious:
Recovering a stale PR (#465) by @adamsulucz. Since the original repo and branch were deleted, I was unable to push the rebased and updated commits to the same PR, so I'm creating a new one and will close #465.

Also, since that PR had been thoroughly reviewed and was only missing a quick fix to get the OK for merging, I'll merge this right away, and we can then rebase #926 and continue the discussion from there.

/cc @sethwoodworth, @igorshubovych, @ostera and @rubenvereecken.

### DIFF
--- a/pages/common/vim.md
+++ b/pages/common/vim.md
@@ -20,7 +20,7 @@
 
 - Switch to insert mode:
 
-`<esc> i <enter>`
+`<esc> i`
 
 - Exit vim:
 
--- a/pages/common/vim.md
+++ b/pages/common/vim.md
@@ -14,18 +14,22 @@
 
 `view {{file}}`
 
-- Exit vim:
+- switch to normal mode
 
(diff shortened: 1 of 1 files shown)


waldyrious:
Hmm. In the future, an approach based on [this](http://stackoverflow.com/a/28822607/266309) may work for this use case. Noting here for future reference.

waldyrious:
No dice. Apparently it's [not possible](http://stackoverflow.com/a/32012637/266309) to update the `refs/pull/*/head` references. This is [described in the github documentation](https://help.github.com/articles/checking-out-pull-requests-locally/#modifying-an-inactive-pull-request-locally). There's a more prose-like and detailed discussion about this issue [here](https://github.com/isaacs/github/issues/591), which may help understanding the problem.

PR #998, opened 2016-08-06
kill.md: change signal explanation (closes #616)

waldyrious:
I've rebased, improved the descriptions, and fixed linting issues in #616. This had to be done in a separate PR for the reasons explained in #996, but the original commits by @lanzhiwang remain properly attributed.

/cc @igorshubovych, @markusthoemmes, @rubenvereecken, @ostera.

### DIFF
--- a/pages/common/kill.md
+++ b/pages/common/kill.md
@@ -1,32 +1,28 @@
 # kill
 
-> Sends a signal to a process.
-> Mostly used for stopping processes.
+> Sends a signal to a process, usually related to stopping the process.
+> All signals except for SIGKILL and SIGSTOP can be intercepted by the process to perform a clean exit.
 
-- Kill the process:
+- Terminate a program using the default SIGTERM (terminate) signal:
 
 `kill {{process_id}}`
 
-- List signal names:
+- List available signal names (to be used without the `SIG` prefix):
 
 `kill -l`
 
-- read configure file,reload
+- Terminate a program using the SIGHUP (hang up) signal. Many daemons will reload instead of terminating:
 
-`kill [-1][-SIGHUP] {{process_id}}`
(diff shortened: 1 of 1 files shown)


agnivade:
Love this page. I always forget the signal codes. LGTM !

jedahan:
Really great contribution

pages/common/kill.md
```diff
@@ -1,12 +1,28 @@
...
 `kill {{process_id}}`
 
-- List signal names:
+- List available signal names (to be used without the `SIG` prefix):
 
 `kill -l`
+
+- Terminate a program using the SIGHUP (hang up) signal. Many daemons will reload instead of terminating:
+
+`kill {{1|-HUP}} {{process_id}}`
```
ostera: Shouldn't this be `-1`? And if so, wouldn't it be more concise to have `-{{1|HUP}}` instead?

waldyrious: Good catch! I'll make the changes :)

ostera:
Good stuff guys, keep it up 👍

waldyrious:
@ostera I personally would have liked this one merged rather than squashed, since the commits were logically separated and properly described -- per the description above:

> I've rebased, improved the descriptions, and fixed linting issues in #616. 

 I'll assume it was just an oversight, but if you actually disagree, let me know :)

ostera:
I figured that since the other commits were unattributed it wouldn't be much of an issue – I agree that if we have multiple authors on the same PR, we shouldn't squash to preserve authorship on the individual changes.

waldyrious:
It's not as much about the authorship (although that counts too, to encourage participation). I was more concerned about encouraging (and preserving) the proper separation of commits as atomic changes, with informative descriptions.

PR #999, opened 2016-08-10
traceroute: Add example of specifying datagram size

zkendall:


pages/common/traceroute.md
```diff
@@ -17,3 +17,7 @@
 - Specify number of queries per hop:
 
 `traceroute -q 5 {{host}}`
+
+- Specify size in bytes of datagram:
+
+`traceroute {{host}} {{size<=32768}}`
```
agnivade: This is misleading. I almost thought that "size<=" was part of the syntax. Whereas its just the number of bytes.

Also, since the default is 60 for IPv4 and 80 for IPv6, its better to give a smaller no. 
`traceroute {{host}} {{45}}` seems good enough.

pages/common/traceroute.md
```diff
@@ -17,3 +17,7 @@
 - Specify number of queries per hop:
 
 `traceroute -q 5 {{host}}`
+
+- Specify size in bytes of datagram:
```
agnivade: "of datagram" -> "of probing packet" ?

zkendall:
@agnivade made suggested changes.

pages/common/traceroute.md
```diff
@@ -18,6 +18,6 @@
 
 `traceroute -q 5 {{host}}`
 
-- Specify size in bytes of datagram:
+- Specify size in bytes of probing packet:
 
-`traceroute {{host}} {{size<=32768}}`
+`traceroute {{host}} 42`
```
agnivade: This is still a token word. It should be `{{42}}`. https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#token-syntax

zkendall: I was copying the other values in this file, which are not in brackets. Are they wrong too, or am I missing something?

waldyrious: Good catch, @zkendall. Those should be changed too. You can either change it yourself in this PR, or we can merge as-is and change later, if you prefer.

ostera: My vote goes to fixing it on this PR, regardless of who does it.

waldyrious: > fixing it on this PR, regardless of who does it.

Actually, only @zkendall can edit this PR's branch, because it doesn't exist in the tldr-pages repo, but on his fork (where only he has commit rights).

If we are to fix it ourselves, it would be easier to merge the PR and then edit the page, than manually create a branch and open a new PR based on the changes of this one (which would mean this PR would have to be closed as the contents would be duplicated in the new branch)... it isn't worth the hassle IMO. So either we merge this and make the fix ourselves directly on master, or wait for @zkendall to say whether he's up for it.

zkendall: Did the thing.

waldyrious:
Thanks for the contribution :)

PR #1001, opened 2016-08-12
edquota: add page

pearcem0:


pages/linux/edquota.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Edit the user quota (default):
+
+`edquota --user {{user}}`
+
+- Edit the group quota:
+
+`edquota --group {{group}}`
+
+- Perform operations on given file system ONLY (default is to perform operations on all file systems with quota):
```
ostera: Something tells me it'd be better to add this default info on a second line right after the description of the command itself.

```markdown
> Edit quota for a user or group.
> Operates on all file systems by default.
```

But @waldyrious is our UX expert here, let's see what he has to say :)

pages/linux/edquota.md
```diff
@@ -0,0 +1,23 @@
+# edquota
+
+> Edit quota for a user or group.
+
+- Edit the user quota (default):
```
waldyrious: What does the "default" mean here?

pages/linux/edquota.md
waldyrious: I agree with @ostera. Also, it would be nice to clarify if this state is stored somewhere permanent, is only for the current session, only for the next command...

pages/linux/edquota.md
pearcem0: If equota is run without flags it is run for the current user.  Should I add a line for 'Edit quota of current user', then change line 5 to 'Edit quota for specified user'. ?

pages/linux/edquota.md
pearcem0: How about - 
> Edit quota for a user or group.
> By default it operates on all file systems with quotas. Quota information is stored permanently in the quota.user and quota.group files in the root of the file system.

pages/linux/edquota.md
waldyrious: Yes, that would be perfect :)

pages/linux/edquota.md
waldyrious: Sounds good -- only a minor nitpick: I'd split the two lines after the second period, rather than the first :) Like this:

> Edit quota for a user or group. By default it operates on all file systems with quotas.
> Quota information is stored permanently in the quota.user and quota.group files in the root of the file system.

pages/linux/edquota.md
```diff
@@ -0,0 +1,28 @@
+# edquota
+
+> Edit quota for a user or group. By default it operates on all file systems with quotas.
+> Quota information is stored permanently in the quota.user and quota.group files in the root of the file system.
+
+- Edit quota of current user (default):
+
+`edquota --user {{user}}`
```
waldyrious: @pearcem0 could the `--user` parameter be omitted altogether in this example?

pearcem0: It _could_, but I kept it in to be consistent with the repquota page.

waldyrious: I'd rather remove it in both (if repquota also allows it being omitted), since `edquota --user myusername` seems rather wasteful. That way, we could get rid of the "(default)" in the description. What do you think?

pearcem0: I agree, that does make more sense. I'll update both  now.

waldyrious: Sorry, I guess I wasn't clear :( I actually meant removing `--user {{username}}` altogether, assuming that the tool would run for the current user if no user is specified. Would the simple command without any parameters work both in `edquota` and `repquota`?

pearcem0: I see, sorry about that. The tool will not run without any parameters. I believe it is correct as it is now.

waldyrious:
LGTM 😄

agnivade:
:+1:

PR #1002, opened 2016-08-12
quotacheck: add page

pearcem0:


ostera:
Btw @pearcem0, you can just keep pushing commits to your branch and the pull request will be updated. At the end we squash all the commits so it will be tidy and neat :)

pages/linux/quotacheck.md
```diff
@@ -0,0 +1,28 @@
+# quotacheck
+
+> Scan a filesystem for disk usage, create, check and repair quota files.
```
waldyrious: The comma before "create" should be changed to a semicolon.

pages/linux/quotacheck.md
```diff
@@ -0,0 +1,28 @@
...
+> Should only be run by super-user.
```
waldyrious: "by *the* super-user."

pages/linux/quotacheck.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Check all mounted non-NFS filesystems:
```
waldyrious: "Check" --> "Check quotas on" (for additional clarity)

pages/linux/quotacheck.md
```diff
@@ -0,0 +1,28 @@
...
+
+`quotacheck --all`
+
+- Force check even if quotas are enabled:
```
waldyrious: I'm confused by this sentence. Did you perhaps mean "disabled"?

pages/linux/quotacheck.md
```diff
@@ -0,0 +1,28 @@
...
+
+`quotacheck --force {{filesystem}}`
+
+- Enable debugging mode:
+
+`quotacheck --debug {{filesystem}}`
```
waldyrious: Same comment as in the edquota PR: is this state stored somewhere permanent, or only for the current session, or only for the next command, or...?

waldyrious:
@pearcem0 please don't add commits so quickly while inline comments are still being added, as this may hide still valid comments as "comments on an outdated diff".

pages/linux/quotacheck.md
pearcem0: Sorry, no. It should be 'Force check if quotas are enabled:' - I'll change that

pages/linux/quotacheck.md
pearcem0: yes. Thanks, I'll add it

pages/linux/quotacheck.md
waldyrious: What does the "force" do here? Without it what would happen?

pearcem0: It forces checking (and writing) of new quota files on file systems. Which can be an issue as the created quota files may be out of sync.

Normally it would be recommended that quota is turned off while the check is run, so to prevent the above issues, quota check tries to remount the file system read-only before starting the scan. After the scan is done it remounts the file system read-write.

Do you think I should add an extra Description line - 
> Scan a file system for disk usage; create, check and repair quota files. 
> It is best to run quota check with quotas turned off to prevent damage or loss to quota files.

notice: I also removed the text about the command being run by super user - I think I should instead append sudo to the start of each example, do you agree?

pages/linux/quotacheck.md
```diff
@@ -0,0 +1,28 @@
+# quotacheck
+
+> Scan a filesystem for disk usage; create, check and repair quota files.
+> Should only be run by the super-user.
+
+- Check all mounted non-NFS filesystems:
```
pearcem0: I think that maybe it should be noted that actual usage may change while running quotacheck on an active file system. 
How would I best convey this, if you agree it needs to be added, @waldyrious ?

waldyrious: while running quotacheck? How long does the scan last, and what kind of changes are you talking about? An example would help me understand this better :)

waldyrious:
I think we've mulled over this one long enough. Thanks for the patient explanations, @pearcem0! LGTM 👍

pages/linux/quotacheck.md
```diff
@@ -0,0 +1,28 @@
...
+> Scan a filesystem for disk usage; create, check and repair quota files.
+> It is best to run quota check with quotas turned off to prevent damage or loss to quota files.
+
+- Check all mounted non-NFS filesystems:
+
+`sudo quotacheck --all`
+
+- Force check even if quotas are enabled:
+
+`sudo quotacheck --force {{filesystem}}`
```
agnivade: Its a bit unclear what `filesystem` refers to here. Does it mean a filesystem type (ext3,ext4,xfs) ? Or a filesystem partition (/dev/sda5,/dev/sda6) ? 

It most probably means the latter. It would be good if we can mention that with an example rather than just saying `filesystem`. Makes everything instantly clear.

waldyrious: ...or change the token to reflect the expected pattern.

pearcem0: Hmm, good point. I think most documentation I've used refers to it as file system, but I usually run the command with / or /home. So would it be more accurate to say something like {{mountpoint}} ?

agnivade: Yes, `{{mountpoint}}` would be good.

agnivade:
:+1:

PR #1005, opened 2016-08-16
Create fswatch.md

pellejacobs:


waldyrious:
Thanks for the contribution! A little more info would be useful, though. Specifically, what kinds of change are covered? if I `touch` the file, or `chmod` it, does that trigger fswatch? What if I overwrite it with `cp` or `mv`? Or delete it with `rm` or `find -delete`?

Also, can these different operations be filtered individually by fswatch (i.e. capture only deletions, or only file contents changes)?

pellejacobs:
All very good questions 😄  

- which kind of operations: literally everything: touching, removing, chmoding, etc. As long as it matches the path, it'll fire the event.

- on different operations, I assume so. However, I haven't looked into that. The main reason I wanted to add this command here, is that the man pages are quite extreme, while the main use case is "run a command when a file changes". This is functionality already implemented in eg. typescript with `tsc --watch` or `grunt-watch`, but it's nice to have if you're not using tools that currently implement watch functionality.

pages/common/fswatch.md
```diff
@@ -0,0 +1,7 @@
+# fswatch
+
+> A cross-platform file change monitor.
+
+- Run a single bash command on file addition, update or deletion. Use `{}` in your bash command as a placeholder for the absolute path to the file change.
```
waldyrious: Could the second sentence be used in the main description, above?

Also, instead of "file addition", I think "file creation" is more intuitive.

waldyrious:
> on different operations, I assume so. However, I haven't looked into that.

It'd be really great if you could clarify this point. The goal of the tldr-pages project is to help people *understand* the basics of the commands they use, rather than merely providing ready-made snippets for copy-pasting (that's what commandlinefu.com is for 😄)

pellejacobs:
Is this kind of what you would expect?

pages/common/fswatch.md
```diff
@@ -0,0 +1,15 @@
...
+
+`fswatch -0 {{path_to_watch}} | xargs -0 {{bash_command}}`
+
+- Use `{}` in your bash command as a placeholder for the absolute path to the file change:
+
+`fswatch -0 {{path_to_watch}} | xargs -0 -I {} {{bash_command}}`
+
+- Filter by certain event type, eg. Updated, Deleted or Created:
+
+fswatch -0 --event {{event_type}} {{path_to_watch}} | xargs -0 {{bash_command}}`
```
waldyrious: missing the ` at the start of this line

waldyrious:
Looks good @pellejacobs :) just address the comment I left inline and I'm ok with merging the PR.

pages/common/fswatch.md
agnivade: Please mention an actual event type(Updated/Deleted/Created) instead of `event_type`.   Enhances the example.

pages/common/fswatch.md
```diff
@@ -0,0 +1,15 @@
+# fswatch
+
+> A cross-platform file change monitor. 
+
+- Run a bash command on file creation, update or deletion:
+
+`fswatch -0 {{path_to_watch}} | xargs -0 {{bash_command}}`
```
agnivade: What sort of paths can be here ? If its only directories, I think we should go with `{{path/to/watch}}`

pages/common/fswatch.md
```diff
@@ -4,12 +4,12 @@
...
 
 - Use `{}` in your bash command as a placeholder for the absolute path to the file change:
 
-`fswatch -0 {{path_to_watch}} | xargs -0 -I {} {{bash_command}}`
+`fswatch -0 {{path/to/watch}} | xargs -0 -I {} {{bash_command}}`
 
 - Filter by certain event type, eg. Updated, Deleted or Created:
 
-fswatch -0 --event {{event_type}} {{path_to_watch}} | xargs -0 {{bash_command}}`
+`fswatch -0 --event Updated {{path/to/watch}} | xargs -0 {{bash_command}}`
```
agnivade: You don't need to remove the `{{}}`. `{{Updated}}` is fine. Refer to - https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#token-syntax

pellejacobs: Of course. Thanks for helping out 😄

waldyrious:
So should I take from your edits after @agnivade's [comment](https://github.com/tldr-pages/tldr/pull/1005#discussion_r75079284) that only directories are valid paths for fswatch?

pellejacobs:
Not really no. It says `{{path/to/watch}}` which means this can both be a directory or a file or even a combination of files with a glob. I made the change as it is consistent with the other tldr pages. 😄

waldyrious:
In that case I'd suggest a simple example using a single file, and an example using a glob pattern (otherwise identical, or perhaps changing the command that runs on changes, to give a concrete example of a use case). This would make these possibilities explicit.

pellejacobs:
👍 

An example of proper use would be, if you'd want to watch the `/tmp` directory and all local js files: 

```bash
fswatch -0 /tmp **/*.js | xargs -0 echo "That changed."
```

agnivade:
Hi @pellejacobs, 

Sorry for the multiple back and forth on this. I think we are nearly there. Just a couple of minor adjustments needed.

Here is what I feel should do the job -

- Run a bash command on file creation, update or deletion:

`fswatch -0 {{path/to/file}} | xargs -0 {{bash_command}}`

- Watch (multiple) directories or files

`fswatch -0 {{path/to/file}} {{path/to/directory}} {{path/to/another_directory/**/*.js}} | xargs -0 {{bash_command}}`

- Use `{}` in your bash command as a placeholder for the absolute path to the file change:

`fswatch -0 {{path/to/directory}} | xargs -0 -I {} {{bash_command}}`

- Filter by certain event type, eg. Updated, Deleted or Created:

`fswatch -0 --event {{Updated}} {{path/to/directory}} | xargs -0 {{bash_command}}`

I have started with a file for the first one. Because the description says "file creation ..etc.". Then moved on to watching specific file, directory, and multiple files. After that, its just further examples/variations of the command. And I have used {{path/to/directory}} since you have mentioned {} as a placeholder for the filename, which will be relevant only if we are watching a directory.

pages/common/fswatch.md
```diff
@@ -6,6 +6,10 @@
 
 `fswatch -0 {{path/to/watch}} | xargs -0 {{bash_command}}`
 
+- Watch (multiple) directories or files
```
waldyrious: 1) why is "multiple" in parenthesis? I don't think that adds anything to the description. 2) I think it's more natural to say "files or directories" than the reverse, but it's ok if you disagree. Also, you need a : at the end of the description :)

pages/common/fswatch.md
```diff
@@ -0,0 +1,19 @@
+# fswatch
+
+> A cross-platform file change monitor.
+
+- Run a bash command on file creation, update or deletion:
+
+`fswatch -0 {{path/to/watch}} | xargs -0 {{bash_command}}`
+
+- Watch (multiple) directories or files
```
waldyrious: 1) why is "multiple" in parenthesis? I don't think that adds anything to the description. 2) I think it's more natural to say "files or directories" than the reverse, but it's ok if you disagree.

Also, you need a : at the end of the description :)

pages/common/fswatch.md
```diff
@@ -0,0 +1,19 @@
...
+
+`fswatch -0 {{first/path/to/directory}} {{specificWatchedFile.js}} {{watchAllJsFiles/**/*.js}} | xargs -0 {{bash_command}}`
+
+- Use `{}` in your bash command as a placeholder for the absolute path to the file change:
```
waldyrious: "the file change" --> "the changed file"?

pages/common/fswatch.md
```diff
@@ -0,0 +1,19 @@
...
+
+`fswatch -0 {{path/to/watch}} | xargs -0 -I {} {{bash_command}}`
+
+- Filter by certain event type, eg. Updated, Deleted or Created:
```
waldyrious: I'd remove "certain" in this sentence.

pellejacobs:
You guys are amazing! Thanks for the feedback 😄

waldyrious:
@pellejacobs you are amazing for bearing with all our nitpicking :P 

After the last changes, this looks great! 👍 👍 I'm ok with merging. (Another maintainer will have to do it, per our guidelines)

PR #1006, opened 2016-08-17
Add vi command to save read only file

plessbd:
I always forget to sudo vim when I edit a file, this has saved me so much time.
http://stackoverflow.com/questions/2600783/how-does-the-vim-write-with-sudo-trick-work

### DIFF
--- a/pages/common/vim.md
+++ b/pages/common/vim.md
@@ -33,3 +33,7 @@
 - Save file then exit vim:
 
 `<esc> :wq <enter>`
+
+- Save file that is read only using sudo permissions:
+
+`<esc> :w !sudo tee % <enter>`


ostera:
Interesting hack @plessbd :) not exactly beginner friendly. It'd be easier to save elsewhere, and then `sudo mv` to the right directory, or even quit and reopen with `sudo vim`, than to understand that:

* `!`, is meant for external command execution
* `tee` will append to a file
* `%` is the name of the current file

And that invoking this command will be the equivalent to `cat file_b | sudo tee file_a`.

waldyrious:
btw, @ostera, we should resolve #926 ASAP, otherwise any PRs dealing with vim (like this one) will either make it harder to integrate, or will be made harder to integrate if #926 is merged first.

Regarding this PR's changed: I appreciate the usefulness, but vim is such a massive program that I don't think we would want to cover use cases like this one; instead, we should make sure the basic text editing operations are covered: insert/delete, copy/cut/paste, open/save, etc. That alone already gets us over the limit of examples per page established in our CONTRIBUTING.md guidelines. May I suggest you submitting this as a commandlinefu.com example, a self-answered SO post, or [cheat](https://github.com/chrisallenlane/cheat) entry?

ostera:
I think I can take care of integrating #926 afterwards, so we don't block other people frmo contributing. I'll make a point of doing it by the end of this week.

plessbd:
It is already on [commandlinefu](http://www.commandlinefu.com/commands/view/5147/save-a-file-you-edited-in-vim-without-the-needed-permissions-no-echo) and I think the [stackoverflow](http://stackoverflow.com/questions/2600783/how-does-the-vim-write-with-sudo-trick-work) I posted in the commit should suffice for there.

I just know I very frequently run into this or am helping someone that has run into it.

I find it to be quicker to explain this than to 
```
<esc> :w /var/tmp/dagnabit <enter> :q <enter>
sudo mv /tmp/dagnabit !$
```

However, I understand your decisions and reasons, thank you for sharing them instead of just saying no or closing. I will look into cheat I did not know of it before and will see about submitting to there.

waldyrious:
> I think I can take care of integrating #926 afterwards, so we don't block other people from contributing.

Absolutely agreed, it was just a heads-up :)

> I'll make a point of doing it by the end of this week.

Yay! Sorry again for messing with it in #996, btw -- but I hope you understand the reasoning :)

waldyrious:
> It is already on [commandlinefu](http://www.commandlinefu.com/commands/view/5147/save-a-file-you-edited-in-vim-without-the-needed-permissions-no-echo) and I think the [stackoverflow](http://stackoverflow.com/questions/2600783/how-does-the-vim-write-with-sudo-trick-work) I posted in the commit should suffice for there.

😅 ah, sorry, I should've looked closer.

> I understand your decisions and reasons, thank you for sharing them instead of just saying no or closing.

It is quite sad that you feel the need to thank us for this, isn't it? Unfortunately the FOSS community tends to foster communication patterns that would be considered plain rude in any other context. I hope our efforts here (in the tldr-pages project along with those in other parts of the FOSS ecosystem) will help that sometime in the future it'll feel more natural to admonish those who aren't nice rather than thanking those who are.

plessbd:
> It is quite sad that you feel the need to thank us for this, isn't it?

Yes it is

> Unfortunately the FOSS community tends to foster communication patterns that would be considered plain rude in any other context.

Which is why I like to express my gratitude when people are being an example of what it should be so they remember why they do it and dont become like the bad ones :).

ostera:
Closing this then. Thanks a million @plessbd, we can always use more of your attitude around here, so if you have more to contribute we'd gladly have a look :)

PR #1007, opened 2016-08-17
osx/pod.md: add page

waldyrious:
cherry-picked the relevant commits from #583 and made the changes discussed there so the pod.md page can be added to the repo. A new PR will need to be created for just the xcodebuild page, and then #583 can be closed.


pages/osx/pod.md
```diff
@@ -0,0 +1,31 @@
+# pod
+
+> Dependency manager for Swift and Objective-C Cocoa projects.
+
+- Create a Podfile for the current project with the default contents:
+
+`pod init`
+
+- Download all pod defined in Podfile and create an Xcode Pods library project in `./Pods`:
```
agnivade: "Download all pods" ?

pages/osx/pod.md
```diff
@@ -0,0 +1,31 @@
...
+
+`pod install`
+
+- Download a new pod for the current project and add it to the Podfile:
+
+`pod install {{pod_name}}`
+
+- Show the outdated pods in the current `Podfile.lock`:
```
agnivade: Will it be simpler to say "in the current project" ?

pages/osx/pod.md
```diff
@@ -0,0 +1,31 @@
...
+
+`pod init`
+
+- Download all pod defined in Podfile and create an Xcode Pods library project in `./Pods`:
+
+`pod install`
+
+- Download a new pod for the current project and add it to the Podfile:
+
+`pod install {{pod_name}}`
```
agnivade: Are you sure this option works ? I don't see one in the docs page. As far as I am seeing, the only way to add a new pod is to add it to the Podfile and run `pod install`.

agnivade:
Just to keep it similar to npm, can we add `pod list` too ? https://guides.cocoapods.org/terminal/commands.html#pod_list

pages/osx/pod.md
waldyrious: Good catch. I must have misread the docs. So it seems to install new pods one must add them to the Podfile and then run plain `pod install`. From [the docs](https://guides.cocoapods.org/using/pod-install-vs-update.html#pod-install):

> Every time the `pod install` command is run — and downloads and install new pods — it writes the version it has installed, for each pods, in the `Podfile.lock` file. (...)
> When you run `pod install`, it only resolves dependencies for pods that are not already listed in the Podfile.lock.

Do you agree with my reading? If so, I'll make the change accordingly.

waldyrious:
> Just to keep it similar to npm, can we add pod list too ?

Sure, as it seems we'll remove the second `pod install` command, this should fit :)

pages/osx/pod.md
agnivade: Correct. Further confirmation - http://stackoverflow.com/a/36355124/1027058

pages/osx/pod.md
waldyrious: Good point, but I will change it to "currently installed" instead, because pods can be defined in the Podfile but not actually installed (i.e. present in the `Podfile.lock`) until `pod install` is run.

waldyrious:
@agnivade how does it look now?

agnivade:
:+1:

PR #1008, opened 2016-08-17
xcodebuild.md: add page

waldyrious:
rebased, squashed and reworded commits from #583 in order to keep only xcodebuild.md (the pod.md changes were picked into #1007) and to provide more informative commit messages.

### DIFF
--- a/pages/osx/xcodebuild.md
+++ b/pages/osx/xcodebuild.md
@@ -1,15 +1,15 @@
 # xcodebuild
 
-> xcodebuild is build xcode project and generate ipa for command line tools
+> Build Xcode projects.
 
-- Build Workspace (ps: CocoaPods Generate Project)
+- Build workspace (CocoaPods Generate Project):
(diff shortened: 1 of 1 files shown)


agnivade:
Can we change the tokens to snake_case ? I am good with everything else.

waldyrious:
I can change that, but I'm still unsure about the "CocoaPods Generate Project". What do you read in that?

agnivade:
I don't think "CocoaPods Generate Project" is needed. It seems a bit off from the preceding line. Also, the man page doesn't talk about anything regarding generate project. Its very clear that if you pass `-workspace` and `-scheme`, it will build the workspace.

waldyrious:
I was wondering if the author meant "CocoaPods-generated project" or something like that. Alas, without his input we can't know. I'll remove the expression.

agnivade:
LGTM.

Issue #1000, opened 2016-08-10
Adding bash command completion


agnivade:
Hello,

I think adding command completion would be a nice feature addition to this project.

Adding it is very simple as we just have to place a file in `/etc/bash-completion.d/` and bash will source it everytime it loads (For zsh also, its something similar although I haven't researched much about it). The problem is the file needs to have the list of all commands that are present. 

Now, this is something I wanted to discuss about. I believe the responsibility of placing the file in the folder (`/etc/bash_completion.d/`) goes to the respective clients. So, I propose creating a completion file 'tldr', which will get updated with each push (through another function added in `scripts/build.sh`), and keeping it in the `scripts` folder. After this, every client will pull the file and place it in that folder in the install step.

Does this seem reasonable ?

ostera:
It'd be interesting to see how much of an overhead it becomes. 100 commands might not be a problem, but if there's 10,000 of them, then it might be problematic.

ostera:
Sorry, the overhead I meant was not when building this list, but when bash parsed it originally, and then whenever you `TAB` to autocomplete or see suggestions.

waldyrious:
if it's alphabetically sorted, I can't imagine it will be too expensive to traverse...

agnivade:
I just tested with around 8000 items. Seems to work pretty fine.

rubenvereecken:
Wooooh #1000

waldyrious:
@agnivade shall we close this, in light of #1010's closure?

agnivade:
Yes

Issue #824, opened 2016-03-15
improve `tldr tldr` with information how tldr should be updated


matkoniecz:
I wanted to update tldr so I checked `tldr tldr` ( http://www.ostera.io/tldr.jsx/#/tldr?_k=fk6knt )

>    Update the local cache of tldr pages:
> 
> tldr --update

I tried `tldr --update` what resulted in 

> usage: tldr [-h] [-o {linux,osx,sunos}] command [command ...]
> tldr: error: the following arguments are required: command

So I tried `tldr -o linux --update` what also resulted in

> usage: tldr [-h] [-o {linux,osx,sunos}] command [command ...]
> tldr: error: the following arguments are required: command

waldyrious:
@matkoniecz thanks for raising this issue. This is something we've considered in the past but haven't made a decision yet. The documentation for this option was added in #744 and in that PR we discussed making the contents of the tldr tldr output a formal requirement for clients to be featured in our README, but we didn't reach consensus. This issue is probably a good place to finalize that discussion and move forward.

May I ask what client you're using? python, node?...

matkoniecz:
python (I see tldr in `pip list` )

waldyrious:
Back-reference: #744

daylightbrightledlight:
So does this mean that this issue can be closed? I think we can close this issue now.

sbrl:
Hrm, I'm not sure @daylightbrightledlight. The thing is that although each client is has the same basic similarities, they also all have their own flags and options. I suspect that this is a case of that. I wonder if there's a formal spec for 'official' tldr clients? Perhaps @waldyrious or @agnivade would know.

agnivade:
We do have a minimal page here - https://github.com/tldr-pages/tldr/wiki/Minimum-specifications-for-TLDR-command-line-clients

Some more thoughts here - https://github.com/tldr-pages/tldr/issues/1065

But yes, I agree this is a client specific thing. I would say we should remove the `--update` flag as it is not there in python client.

sbrl:
Ah, ok @agnivade. I've opened #2272, which removes the --update flag.

PR #1010, opened 2016-08-19
Adding bash completion script

agnivade:
- Added as a build hook which will re-generate the list of
  commands and update the completion file
- The way to re-generate the file is not very elegant because of the sed-replace thingy in the template file. But it should do the job.

### DIFF
--- a/scripts/bash-complete.template
+++ b/scripts/bash-complete.template
@@ -1,0 +1,14 @@
+_complete_tldr() 
+{
+    local cur prev opts
+    COMPREPLY=()
+    cur="${COMP_WORDS[COMP_CWORD]}"
+    prev="${COMP_WORDS[COMP_CWORD-1]}"
+    opts="<placeholder>"
+
+    if [[ ${cur} == * ]] ; then
+        COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
+        return 0
+    fi  
+}
+complete -F _complete_tldr tldr
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -38,6 +38,12 @@ function upload_assets {
   git push -q
 }
 
+function update_bash_completion {
+  cd $TLDRHOME
+  commands=`find pages/ -type f -exec basename {} \; | sed 's/\.md//g' | tr '\n' ' '`
+  sed "s/<placeholder>/$commands/g" scripts/bash-complete.template > scripts/tldr
+}
+
 ###################################
 # MAIN
 ###################################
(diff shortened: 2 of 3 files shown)


agnivade:
xRef- #1000

rubenvereecken:
Oh man this was a long time coming. Sorry I never saw #1000, I should have responded there before. I actually have a zsh completion script, not sure where it is though, hang on.

rubenvereecken:
Alright this is really weird I can't find it anywhere on Github, although I know there was a discussion on this at some point. Maybe the link was removed at some point? Anywhere here goes https://gist.github.com/rubenvereecken/53d4c9173bf80245dff1155f902077a7. I guess we should add this in somewhere again.

What I wanted to share with you is this bit: `pages=$(tldr -a1)`. Now, I don't know about bash and I don't care to find out by myself right now, but could this approach work for bash as well? We updated the client to manage this exactly for this reason.

Here's the rest it can show

rubenvereecken:
Found it again, it's at the `tldr-node-client` repo: https://github.com/tldr-pages/tldr-node-client/blob/master/autocompletion/tldr.bash

agnivade:
@rubenvereecken - This is really great :) With your permission, I will include the zsh completion script as well.

My only concern is that the switch "-a1" might be a bit client specific, isn't it ? Are we sure other clients implement this switch ? 

Also, the "--*" flags seem to be client specific. IMO, the source repo should not be concerned about specific switches of every client. What do you think about it ?

waldyrious:
We've discussed in #824 and #744 about providing a specification for tldr clients in the tldr.md page, which all supported clients would have to conform to. This is clearly one of the use cases that would benefit from this :)

rubenvereecken:
The client specificity is exactly the reason why I think a completion script should not be bundled with the main repo. People will miss out on client-specific completions because they'll never install those.

We did want to make the listing behavior standard. After all, it's useful for all clients. Never got around to it I think.

agnivade:
Ah, interesting. I think we can enforce the list-all behavior on all clients. I don't care about the "-1". I can do the rest of it with bash.

But we are still undecided on the "--" flags. I agree with @rubenvereecken here because a proper completion script to complete "--" as well as the command argument. But that is not possible in the main repo. In that case, this task is better suited to be tackled by clients as done by the tldr-node-client.

agnivade:
@waldyrious - This PR is only open due to https://github.com/tldr-pages/tldr/issues/1065#issuecomment-247762407. Do you think this can be tracked in the parent issue itself ? Since this PR actually is never going to be merged (because autocompletion will be implemented by individual clients), I propose to either track it in the parent issue or open a new issue to track it.

waldyrious:
Agreed @agnivade -- I've added a comment there to make sure the intentions are clear.

PR #1011, opened 2016-08-21
Add `apt-get autoremove`

zdroid:
`apt-get autoremove` is surely one of the six most used `apt-get` commands, so it should be here among the others. For example, it is more common than `apt-get dist-upgrade`.


waldyrious:
LGTM 👍

PR #1013, opened 2016-08-22
apg: add page

michaelAndersonCampingWorld:


agnivade:
Hi @michaelAndersonCampingWorld, please fix the syntax errors pointed out by Travis. 

Mainly, there should be a one line gap between a description and an example. Also descriptions should end with a colon.

waldyrious:
The last example could be removed, since it's just a synthesis of two examples already present in the page (that would also put the page back within our guidelines of 5-6 examples per page).

agnivade:
This page needs some more work. I checked the man page, and there are a couple of useful flags left out. The current page mainly focuses only on the `-M` flag for selecting character set.

Please add other flags like `-a` (algo to select), `-x` (max passwd length), `-n` (# of passwords to generate). Also update the description of `-m` to minimum password length. (I know that in this case, it will generate a length of 16 chars because it has exceeded the max default length. But its slightly misleading for the users). We can condense the `-M` example.

Lastly, I would suggest to start off with simple `apg` and explain it, as that seems to work just fine. Then slowly start adding flags.

michaelAndersonCampingWorld:
@agnivade I think I disagree with you.  I'm not sure how other people use `apg`, I only discovered it yesterday, but I don't care about those other values. The reason I don't care about those other options is that is not how most people create passwords. If you use this for websites, then what you want is to quickly meet their requirements, i.e. 1 symbol, 1 uppercase, 1 lowercase and a number. You don't care about the algo, max password length or number of passwords. 

I'm new to the tldr community, but don't we want to focus on the common/useful parts of a command?

waldyrious:
@michaelAndersonCampingWorld I think what @agnivade is saying is that this page can still provide all the information it currently does, without spending so many examples in the -M option alone.

Just like you had combined the four individual options into `-M SNCL`, and even -M and -m into `-M SNCL -m 16`, we could for instance start with a simple, option-less `apg` call, then a `-M` example with a single argument, then `-M SNCL` to mention the other possible values of the -M option, and then a`-m` example.

This would cover everything that is currently in the page, leaving room for a couple more examples that demonstrate what else apg can do.

You're right that tldr-pages aims to demonstrate the most common invocations of a command, but note the plural in "invocations" -- if we restrict all examples to variations of one or two options, we're actually doing a disservice to those who don't know the command, by giving the impression that it might have only those options. In a way, it's about making good use of our self-imposed limited space: when so little room is available, using it all up in multiple instances of the same option is not very efficient.

Does this make sense to you?

agnivade:
Building up on @waldyrious said, the very reason I looked for the man page is because I wondered why would a command have only `-M` as its primary flag ? Something felt odd.

Therefore, from a newbie perspective who has just done a `tldr apg` for the first time, it will slightly mislead him to think that this is all the command is supposed to do. What we generally do here is use the space of 5/6 examples judiciously so that the most common flags are covered and attempt to explain them as simply as possible. We start with the simplest command invocation(`apg`), and then slowly build up from the simplest flags to the complex ones.

Also, keep in mind that `apg` might not just be used to generate passwords for websites. Its a generic password generator, so it can generate passwords for other apps like DBs or ssh keys. Therefore, I believe adding min/max password length would be helpful in case the user wants to remember a shorter or a longer password. And in case, you want to generate the password as a part of some script, you would want only a single password to be shown(`-n`). Choosing an algo is slightly bordering on being complex, we can keep it as a last example, or remove it depending on what other people say.

michaelAndersonCampingWorld:
@agnivade @waldyrious I think that you're right in general.

pages/common/apg.md
```diff
@@ -0,0 +1,23 @@
+# apg
+
+> Creates arbitrarily complex random passwords.
+
+- Create random passwords (default password length is 8):
+
+`apg`
+
+- Create a password with at least 1 symbol (S), 1 number(N), 1 uppercase(C), 1 lowercase(L):
```
waldyrious: Please add spaces before the parenthesis

pages/common/apg.md
```diff
@@ -0,0 +1,23 @@
...
+
+`apg -M SNCL`
+
+- Create a password with 16 chars:
```
waldyrious: "chars" --> "characters"

pages/common/apg.md
```diff
@@ -0,0 +1,23 @@
...
+
+`apg -m 16`
+
+- Create a password with max length of 16:
```
waldyrious: "max" --> "maximum"

pages/common/apg.md
```diff
@@ -0,0 +1,23 @@
...
+
+`apg -x 16`
+
+- Create a password that doesn't appear in a dictionary (you have to provide the dictionary):
+
+`apg -r dictfile`
```
waldyrious: "dictfile" --> "{{dictionary_file}}"

pages/common/apg.md
```diff
@@ -0,0 +1,23 @@
...
+
+`apg`
+
+- Create a password with at least 1 symbol (S), 1 number(N), 1 uppercase(C), 1 lowercase(L):
+
+`apg -M SNCL`
+
+- Create a password with 16 chars:
+
+`apg -m 16`
```
waldyrious: please add {{}} around the variable: {{16}}

pages/common/apg.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Create a password with max length of 16:
+
+`apg -x 16`
```
waldyrious: same here: add {{}}

waldyrious:
@michaelAndersonCampingWorld thanks for understanding and quick reaction :) I've left some comments inline, please take a look.

pages/common/apg.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Create a password with 16 characters:
+
+`apg -m {{16}}`
+
+- Create a password with maximum length of 16:
+
+`apg -x {{16}}`
+
+- Create a password that doesn't appear in a dictionary (you have to provide the dictionary):
```
agnivade: We generally try to avoid "you". How about "the dictionary file has to be provided" ?

waldyrious: side note: this should be added to CONTRIBUTING.md. [Wikipedia has some wording](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style#Second-person_pronouns) we might take inspiration from.

agnivade: noted.

waldyrious:
Thanks for the contribution, patience and quick reactions :) I think it's safe to assume LGTM from @agnivade, so I'll merge.

PR #1014, opened 2016-08-22
pwgen: make option names more intuitive

waldyrious:


agnivade:
LGTM. It took me a few moments to figure out the [y] thing. But yea, it makes sense. We should follow more of this convention when the flag does not match with the description.

ostera:
I like this convention too 👍

waldyrious:
If you guys like it we could add it to CONTRIBUTING.md :)

Issue #1018, opened 2016-08-27
Add convention to indicate origin of option names to Contributing.md


waldyrious:
Generally, if the short option letter corresponds to a given word, that word must be present in the command description (even if within parenthesis as a synonym to a more common word); otherwise, if it corresponds to a letter within a word, that letter should be highlighted (see for example chmod and pwgen).

If the letter was arbitrarily chosen and doesn't correspond to a word or a letter within a word, the description should strive to include a word that _could_ be applied to the option and starts with that letter.

ostera:
chmod page: https://tldr.ostera.io/chmod
pwgen: https://tldr.ostera.io/linux/pwgen

Note that the emphasis on both these pages is inconsistent (one is using `t(e)xt` and the other is using `t[e]xt`).

waldyrious:
@ericbn has changed chmod to use brackets in #1024 ([congrats btw!](http://www.xkcd.com/1000/)), so I suppose we can standardize in them rather than parenthesis.

He also pointed out that the `grep` page also uses this system.

marchersimon:
Can we close this?

PR #1019, opened 2016-08-27
curl.md: add more download examples

ericbn:
- Download using remote file name (instead of using `-o`)
- Download resuming previous file transfer
- Use `example.com` instead of `"URL"` or `localhost` as example host

### DIFF
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -5,7 +5,15 @@
 
 - Download a URL to a file:
 
-`curl "{{URL}}" -o {{filename}}`
+`curl {{http://localhost}} -o {{filename}}`
+
+- Download file using remote name:
+
+`curl -O {{http://localhost/filename}}`
+
+- Automatically continue/resume previous file transfer:
+
+`curl -O -C - {{http://localhost/filename}}`
 
 - Send form-encoded data:
 
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -3,17 +3,13 @@
 > Transfers data from or to a server.
 > Supports most protocols including HTTP, FTP, POP.
 
-- Download a URL to a file:
+- Download the contents of an URL to a file:
 
 `curl {{http://example.com}} -o {{filename}}`
 
-- Download file using remote name:
+- Download a file saving it under the same name as indicated in the URL, following redirects, and automatically resuming a previous file transfer:
 
-`curl -O {{http://example.com/filename}}`
-
-- Automatically continue/resume previous file transfer:
-
-`curl -O -C - {{http://example.com/filename}}`
+`curl -O -L -C - {{http://example.com/filename}}`
 
 - Send form-encoded data:
 
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -3,13 +3,13 @@
 > Transfers data from or to a server.
 > Supports most protocols including HTTP, FTP, POP.
 
-- Download the contents of an URL to a file:
+- Download the [o]utput of an URL to a file, following [L]ocation redirects:
 
-`curl {{http://example.com}} -o {{filename}}`
+`curl -o {{filename}} -L {{http://example.com}}`
 
-- Download a file saving it under the same name as indicated in the URL, following redirects, and automatically resuming a previous file transfer:
+- Download a file saving the [O]utput under the same filename as indicated in the URL, and automatically [C]ontinuing/resuming a previous file transfer:
 
-`curl -O -L -C - {{http://example.com/filename}}`
+`curl -O -C - {{http://example.com/filename}}`
 
 - Send form-encoded data:
 
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -5,40 +5,40 @@
 
 - Download a URL to a file:
 
-`curl {{http://localhost}} -o {{filename}}`
+`curl {{http://example.com}} -o {{filename}}`
 
 - Download file using remote name:
 
-`curl -O {{http://localhost/filename}}`
+`curl -O {{http://example.com/filename}}`
 
 - Automatically continue/resume previous file transfer:
 
-`curl -O -C - {{http://localhost/filename}}`
+`curl -O -C - {{http://example.com/filename}}`
 
 - Send form-encoded data:
 
-`curl --data {{name=bob}} {{http://localhost/form}}`
+`curl --data {{name=bob}} {{http://example.com/form}}`
 
 - Send JSON data:
 
-`curl -X POST -H "Content-Type: application/json" -d {{'{"name":"bob"}'}} {{http://localhost/login}}`
+`curl -X POST -H "Content-Type: application/json" -d {{'{"name":"bob"}'}} {{http://example.com/login}}`
 
 - Specify an HTTP method:
 
-`curl -X {{DELETE}} {{http://localhost/item/123}}`
+`curl -X {{DELETE}} {{http://example.com/item/123}}`
 
 - Head request:
 
-`curl --head {{http://localhost}}`
+`curl --head {{http://example.com}}`
 
 - Include an extra header:
 
-`curl -H "{{X-MyHeader: 123}}" {{http://localhost}}`
+`curl -H "{{X-MyHeader: 123}}" {{http://example.com}}`
 
 - Pass a user name and password for server authentication:
 
-`curl -u myusername:mypassword {{http://localhost}}`
+`curl -u myusername:mypassword {{http://example.com}}`
 
 - Pass client certificate and key for a secure resource:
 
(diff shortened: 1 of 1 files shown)


waldyrious:
I like the change from URL to an actual URL example, but how about using [example.com](http://example.com) instead of localhost?

ericbn:
@waldyrious, what about `http://host` instead of `http://localhost`?

waldyrious:
Do you think that's clearer than an actual (functioning) example url?

ericbn:
`example.com` looks cool. Updated the pull req.

pages/common/curl.md
```diff
@@ -5,32 +5,40 @@
 
 - Download a URL to a file:
```
agnivade: suggestion - "Download a URL" -> "Download the contents of an URL".

pages/common/curl.md
```diff
@@ -5,32 +5,40 @@
...
 
-`curl "{{URL}}" -o {{filename}}`
+`curl {{http://example.com}} -o {{filename}}`
+
+- Download file using remote name:
```
agnivade: This description can be improved a bit. How about - "Download the contents to a file where the filename is extracted from the last part of the url" ?

pages/common/curl.md
```diff
@@ -5,32 +5,40 @@
...
+
+`curl -O {{http://example.com/filename}}`
+
+- Automatically continue/resume previous file transfer:
```
agnivade: IMO, this option is a bit too advanced to show here. We can instead add the `-L` option here.

pages/common/curl.md
```diff
@@ -5,32 +5,40 @@
...
 
-`curl -X POST -H "Content-Type: application/json" -d {{'{"name":"bob"}'}} {{http://localhost/login}}`
+`curl -X POST -H "Content-Type: application/json" -d {{'{"name":"bob"}'}} {{http://example.com/login}}`
 
 - Specify an HTTP method:
 
-`curl -X {{DELETE}} {{http://localhost/item/123}}`
+`curl -X {{DELETE}} {{http://example.com/item/123}}`
 
 - Head request:
```
agnivade: While we are on this page, I feel we can do away with this option. Showing the `--head` option is redundant here immediately after showing the `-X METHOD` option. One can easily do a `-X HEAD` to make a head request. 

Given the limited space that we have, I think we should adhere to showing only one way to achieve something.

waldyrious: Can you look up the PR(s) that introduced these? IIRC the motivation was that there was a subtle difference between the two.

pages/common/curl.md
waldyrious: Agreed but I think this can be said more concisely.  Perhaps "Download a file saving it under the same name as indicated in the url"?

pages/common/curl.md
agnivade: will do.

pages/common/curl.md
ericbn: This is one of the main uses I have for curl. Also, the [`tldr wget`](https://github.com/tldr-pages/tldr/blob/master/pages/common/wget.md) page has the analogous option: "Continue an incomplete download".

pages/common/curl.md
ericbn: The [`tldr wget`](https://github.com/tldr-pages/tldr/blob/master/pages/common/wget.md) uses the exact same "Download a URL to a file" phrase.

agnivade: Well, if this makes it clearer to understand, we can update the wget page too. :)

pages/common/curl.md
agnivade: Well, this PR results in more examples than what wget page has. So, I was thinking that since we usually try to keep it to 5,6 examples, it might be better to use the `-L` option rather than `-C`. Lets take a decision based on what others comment.

waldyrious: What does -L do?

pages/common/curl.md
waldyrious: I agree @agnivade's proposal is clearer. I vote for updating wget.md accordingly.

pages/common/curl.md
agnivade: It automatically follows a redirect.

waldyrious: That seems like an auxiliary function which IMHO should be the default behavior. Perhaps we could include it in the first example (`-o`), mentioning what it does, and maybe even include it in all the other examples where it makes sense.

ericbn:
Updated the page with the suggested changes. As I believe both `-L` (follow redirects) and `-C -` (automatically resume previous file transfer), and well as `-O` (use remote filename) should be the default behavior, I merged them all in one example.

waldyrious:
@ericbn I appreciate the intention, but I think the sudden jump in complexity from the first example to the second one is too large. I'd rather introduce -L in the first example. -O and -C could stay together in the second, as long as the description indicates what behavior corresponds to what option name (e.g. by using the words [O]utput and [C]ontinue).

pages/common/curl.md
ericbn: `--head` example exists since the first commit (https://github.com/tldr-pages/tldr/commit/11264d9b19000734a2d35ecbdbdebc0b0b45aed9), `-H, --header` was introduced in https://github.com/tldr-pages/tldr/pull/631

ericbn:
Actually, the `-L -O -C -` combination is not an unusual one: https://twitter.com/nixcraft/status/476283005476945921, and about 1,710 results in [Google](https://www.google.com/#q=%22curl+-L+-O+-C%22) (if it's worth something)...

And as as the curl man page says:
> curl offers a busload of useful tricks like proxy support, user authentication, FTP
> upload,  HTTP  post,  SSL connections, cookies, file transfer resume, Metalink, and
> more. As you will see below, the number of features will make your head spin!

Maybe joining a few options that look good together like this in one example can keep this tldr page from growing even more...

waldyrious:
Yes, joining options is a good strategy for complex commands, but we usually attempt to introduce new options one or at most two at the time, to avoid making the step from one example to the next too overwhelming, especially for beginners. So yeah, I'd insist that adding -L to the first example and adding -O and -C to the second one (even if it still uses -L) would be the best approach in that light.

As for using the long options, that's certainly a possibility, but we have to balance verbosity and convenience versus understandability. In this case, since the long names are somewhat long, and we'd be using two of them in the same command, and there are words that can be used to indicate what the short options mean, I suppose we're better off using the short options.

ericbn:
Updated page.

waldyrious:
Thanks. The first example may be a bit daunting like this, but I don't think we have much alternative here unless we want to remove existing examples. LGTM :) 👍

ericbn:
"Send JSON data" example repeats the same parameters as "Specify an HTTP method" and "Include an extra header". Maybe something can be cleaned up here, and "Send JSON data" can more specifically be "Send data, specifying an HTTP method and including an extra header" (respective items would be enclosed in `{{}}`).

Also, isn't "Download the [o]utput of an URL to a file, following [L]ocation redirects" looking ugly (same for second example)? I haven't seen other examples using this approach. Maybe if we just remove the `[]`...?

pages/common/curl.md
agnivade: @waldyrious - They are indeed different. `--head` will print only the response headers. This is usually achieved with the HEAD http method. But you can also use the `--head` option with GET method, in which case it ignores the response body and only prints the headers. This is kind of unecessary IMO.

I would either do `curl -X HEAD http://example.com` or `curl http://example.com -v -o /dev/null`.

agnivade:
I agree that some examples can be removed. 

The order of the examples should also be changed. For eg - "Send JSON data" uses `-X` and `-H` which are again introduced later. Maybe, we should put "specify http method" and "include extra header" first, and then show this one.

I vote for removing the `--head` example as it can be achieved with the already given options. And given that this page is at `#10` of most searched commands in mixpanel, we should definitely work on making it as simple as possible.

waldyrious:
> Also, isn't "Download the [o]utput of an URL to a file, following [L]ocation redirects" looking ugly (same for second example)? I haven't seen other examples using this approach. Maybe if we just remove the `[]`...?

It may look a little weird, but is functional. See #1018 :)

ericbn:
Cool, #1018 could go to https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md.

Created a new pull req #1023, since this is getting big. I tried to organize the examples keeping the learning curve in mind. Count of examples didn't change. And in the end the new flags introduced were `-O`, `-L` and `-C -`, also `--data` was replaced by `-d`. Still didn't remove the `--head`.

ericbn:
Closing this is favor of #1023.

pages/common/curl.md
waldyrious: Just for future reference and convenience: `--head` was introduced in #631 as @ericbn said above, and changed to `-H` in #696.

PR #1020, opened 2016-08-28
Add grep flags for lines before, after the match

rahulkavale:


agnivade:
I am kind of neutral on this change. Will wait for others to comment.

waldyrious:
I'd rather say "use -A instead" and the rest could stay the same.

pages/common/grep.md
```diff
@@ -19,7 +19,7 @@
 
 `grep -E {{^regex$}} {{path/to/file}}`
 
-- Print 3 lines of context around each match:
+- Print 3 lines of context around each match (supports -A for lines after the match, -B before the match):
```
ostera: maybe something like this?

```markdown
- Print 3 lines of [C]ontext around, [A]fter, or [B]efore each match:

`grep -{{C|A|B}} 3 {{search_string}} {{path/to/file}}`
```

waldyrious: Nice! I like this :)

rahulkavale:
Makese sense, will update the PR

waldyrious:
LGTM. Given the other maintainers' comments, I believe it's safe to assume an OK for merging. Thanks!

PR #1021, opened 2016-08-28
ufw: Improve page

agnivade:
This closes #893

### DIFF
--- a/pages/linux/ufw.md
+++ b/pages/linux/ufw.md
@@ -21,7 +21,7 @@
 
 - Allow only TCP traffic from 192.168.0.4 to any address on this host, on port 22:
 
-`ufw allow from {{192.168.0.4}} to {{any}} port {{22}} proto {{tcp}}`
+`ufw allow proto {{tcp}} from {{192.168.0.4}} to {{any}} port {{22}}`
 
 - Deny traffic on port 80 on this host:
 
@@ -29,7 +29,7 @@
 
 - Deny all UDP traffic to port 22:
 
-`ufw deny from {{any}} to {{any}} port {{22}} proto {{udp}}`
+`ufw deny proto {{udp}} from {{any}} to {{any}} port {{22}}`
 
 - Remove a particular rule. The rule number can be retrieved from the `ufw status numbered` command:
 
--- a/pages/linux/ufw.md
+++ b/pages/linux/ufw.md
@@ -11,18 +11,26 @@
 
 `ufw disable`
 
-- Show ufw rules:
+- Show ufw rules, along with their numbers:
 
-`ufw status`
+`ufw status numbered`
 
-- Add ufw allow rule:
+- Allow incoming traffic on port 5432 on this host:
 
-`ufw allow {{port}} {{service_name}}`
+`ufw allow {{5432}}`
 
-- Example: Allow access to port 1337 from IP 1.2.3.4 using udp:
+- Allow only TCP traffic from 192.168.0.4 to any address on this host, on port 22:
 
-`ufw allow from 1.2.3.4 to any port 1337 proto udp`
+`ufw allow from {{192.168.0.4}} to {{any}} port {{22}} proto {{tcp}}`
 
-- Add ufw deny rule:
+- Deny traffic on port 80 on this host:
 
-`ufw deny {{port}} {{service_name}}`
+`ufw deny {{80}}`
+
+- Deny all UDP traffic to port 22:
+
+`ufw deny from {{any}} to {{any}} port {{22}} proto {{udp}}`
+
+- Remove a particular rule. The rule number can be retrieved from the `ufw status numbered` command:
+
+`ufw delete {{rule_number}}`
--- a/pages/linux/ufw.md
+++ b/pages/linux/ufw.md
@@ -31,6 +31,6 @@
 
 `ufw deny proto {{udp}} from {{any}} to {{any}} port {{22}}`
 
-- Remove a particular rule. The rule number can be retrieved from the `ufw status numbered` command:
+- Delete a particular rule. The rule number can be retrieved from the `ufw status numbered` command:
 
 `ufw delete {{rule_number}}`
--- a/pages/linux/ufw.md
+++ b/pages/linux/ufw.md
@@ -11,14 +11,18 @@
 
 `ufw disable`
 
+- Show ufw rules:
+
+`ufw status`
+
 - Add ufw allow rule:
 
 `ufw allow {{port}} {{service_name}}`
 
+- Example: Allow access to port 1337 from IP 1.2.3.4 using udp:
+
+`ufw allow from 1.2.3.4 to any port 1337 proto udp`
+
 - Add ufw deny rule:
 
 `ufw deny {{port}} {{service_name}}`
-
-- Show ufw rules:
-
-`ufw status`


pages/linux/ufw.md
```diff
@@ -15,14 +15,22 @@
...
+`ufw allow {{5432}}`
 
-- Example: Allow access to port 1337 from IP 1.2.3.4 using udp:
+- Allow traffic from 192.168.0.4 to port 22 using only tcp protocol:
 
-`ufw allow from 1.2.3.4 to any port 1337 proto udp`
+`ufw allow from {{192.168.0.4}} to {{any}} port {{22}} proto {{tcp}}`
 
-- Add ufw deny rule:
+- Deny incoming traffic on port 80 on this:
```
waldyrious: Perhaps "on this host", to match the first example (and to be more explicit)?

pages/linux/ufw.md
```diff
@@ -15,14 +15,22 @@
...
-`ufw deny {{port}} {{service_name}}`
+`ufw deny {{80}}`
+
+- Deny traffic to port 22 using only udp protocol:
+
+`ufw deny from {{any}} to {{any}} port {{22}} proto {{udp}}`
+
+- Delete the deny rule on port 80:
+
+`ufw delete deny {{80}}`
```
waldyrious: Reading this, I'd be unsure whether this command could potentially delete more rules than I wanted. How does one ensure it deletes the specific rule one is targeting?

pages/linux/ufw.md
```diff
@@ -15,14 +15,22 @@
...
+- Deny incoming traffic on port 80 on this:
 
-`ufw deny {{port}} {{service_name}}`
+`ufw deny {{80}}`
+
+- Deny traffic to port 22 using only udp protocol:
+
+`ufw deny from {{any}} to {{any}} port {{22}} proto {{udp}}`
+
+- Delete the deny rule on port 80:
```
waldyrious: Perhaps "remove" would work better than "delete" here.

waldyrious:
I like the direction of these changes, but the jump in complexity is too sudden and unexplained, e.g. from the generic allow rule to the specific one. I suppose "to any" is required if "from" is provided, but maybe if we integrated the word "any" in the example description the complexity would be more manageable?

agnivade:
Damn, mea culpa. I have been wondering how come I left the line as "on this" ? Now I see, I have uncommited changes ! :flushed: 

> maybe if we integrated the word "any" in the example description the complexity would be more manageable

I had exactly done that.

pages/linux/ufw.md
agnivade: A valid concern. A better way to delete a rule is by the rule #. To see the rule # we can do `ufw status numbered`. Then do `ufw delete [number]` I will make the changes accordingly.

pages/linux/ufw.md
```diff
@@ -11,14 +11,26 @@
 
 `ufw disable`
 
-- Add ufw allow rule:
+- Show ufw rules along with their numbers:
```
waldyrious: I'd add a comma after "rules", to better guide reading flow.

pages/linux/ufw.md
```diff
@@ -11,14 +11,26 @@
...
+`ufw status numbered`
 
-- Add ufw deny rule:
+- Allow incoming traffic on port 5432 on this host:
 
-`ufw deny {{port}} {{service_name}}`
+`ufw allow {{5432}}`
 
-- Show ufw rules:
+- Allow traffic from 192.168.0.4 to any address on this host on port 22 using only tcp protocol:
```
waldyrious: "using only tcp protocol" --> "using only the TCP protocol". Also, commas after "host" and "22".

pages/linux/ufw.md
```diff
@@ -11,14 +11,26 @@
...
 
-`ufw status`
+`ufw allow from {{192.168.0.4}} to {{any}} port {{22}} proto {{tcp}}`
+
+- Deny traffic on port 80 on this host:
+
+`ufw deny {{80}}`
+
+- Deny traffic to port 22 using only udp protocol:
```
waldyrious: Same as above: comma after "22", "udp" --> "the UDP"

pages/linux/ufw.md
```diff
@@ -11,14 +11,26 @@
...
+
+`ufw deny from {{any}} to {{any}} port {{22}} proto {{udp}}`
+
+- Remove a particular rule. The rule number can be retrieved from the "ufw status numbered" command:
```
waldyrious: Much better. I'd style the `status` command using ``, though

agnivade: Yea I tried it, then tested with `tldr --render`. Didn't make any difference though.

waldyrious: That's client-specific (although we *could* declare the node client as *the* spec for clients, we haven't done so, and I don't think we should). In any case, since we're using markdown, it's better to use the markdown syntax to indicate a code snippet.

pages/linux/ufw.md
```diff
@@ -11,26 +11,26 @@
...
+- Allow traffic from 192.168.0.4 to any address on this host, on port 22, using only the tcp protocol:
 
 `ufw allow from {{192.168.0.4}} to {{any}} port {{22}} proto {{tcp}}`
 
 - Deny traffic on port 80 on this host:
 
 `ufw deny {{80}}`
 
-- Deny traffic to port 22 using only udp protocol:
+- Deny traffic to port 22, using only the UDP protocol:
```
ostera: how about `Deny all UDP traffic to port 22` ?

agnivade: Had thought about that. But coming from the previous example which starts with "Deny traffic", then building on to a complex example, I thought adding the filters later would be more readable.

ostera: Hmm. I beg to disagree. My arguments being:

1. Less text is faster to scan/read/maintain (and also store and transfer)
2. The point of the example is that the UDP traffic will be blocked on port 22, emphasis on the kind of traffic and not the port

Perhaps we can do an example like:

```markdown

- Deny all [UDP,TCP] traffic to port 22:

`ufw deny from {{any}} to {{any}} port {{22}} proto {{udp|tcp}}`
```

Otherwise I stand by my suggestion, but I'd like to hear @waldyrious's opinion on it

waldyrious: I was originally inclined to agree with @agnivade but I can see how that wins on a more abstract, intellectual realm, whereas @ostera's arguments win on a more concrete and pragmatic perspective.

The combined example might actually even convey the grammar better than separate ones (where the reader would have to detect the pattern and build it on their heads), so I'd even approve of adopting that. But I'm ok with just rewording the description if we don't get consensus on the latter.

ps - we're discussing on a subset of the changes rather than on the full PR. Unfortunately github's UI seems to be making that distinction rather subtle lately.

agnivade: The combined example does seem to put on a bit of cognitive load to completely grasp the pattern. I am okay with changing the description. Agree with @ostera's thinking.

pages/linux/ufw.md
```diff
@@ -11,14 +11,26 @@
...
+- Allow incoming traffic on port 5432 on this host:
 
-`ufw deny {{port}} {{service_name}}`
+`ufw allow {{5432}}`
 
-- Show ufw rules:
+- Allow only TCP traffic from 192.168.0.4 to any address on this host, on port 22:
 
-`ufw status`
+`ufw allow from {{192.168.0.4}} to {{any}} port {{22}} proto {{tcp}}`
```
waldyrious: does the `proto` have to come at the end? It would be nice if we could read the actual command in roughly the same order as the description.

ostera: You mean as in `ufw allow proto {{tcp}} from ...` ? That'd be hella nice.

agnivade: Ask and ye shall receive :smile:

pages/linux/ufw.md
```diff
@@ -11,14 +11,26 @@
...
 
-- Show ufw rules:
+- Allow only TCP traffic from 192.168.0.4 to any address on this host, on port 22:
 
-`ufw status`
+`ufw allow proto {{tcp}} from {{192.168.0.4}} to {{any}} port {{22}}`
+
+- Deny traffic on port 80 on this host:
+
+`ufw deny {{80}}`
```
waldyrious: Maybe this description should say "incoming traffic", to match the `ufw allow` example.

pages/linux/ufw.md
```diff
@@ -11,14 +11,26 @@
...
+
+- Deny all UDP traffic to port 22:
+
+`ufw deny proto {{udp}} from {{any}} to {{any}} port {{22}}`
+
+- Remove a particular rule. The rule number can be retrieved from the `ufw status numbered` command:
```
waldyrious: Sorry for backtracking, but maybe it would be better after all to use the same word as the command actually uses, so "Delete" instead of "Remove". This is consistent with the other examples, e.g. "allow", "deny".

pages/linux/ufw.md
agnivade: I was going with the 3rd option. Make the first description verbose, then make the remaining ones implicitly follow the previous stuff. But I am okay with changing it if you guys feel so.

waldyrious: Nah, it's ok as it is. Let's not be too pedantic here 😅

waldyrious:
LGTM 😃

agnivade:
Yes, I want to keep the last 2 commits and squash the rest.

waldyrious:
Sure 👍. Let me know when you're ready then.

agnivade:
@waldyrious - This is failing because of the changes that you pushed in zip.md.

waldyrious:
Yes, I've noticed. I'll remove the [] for now until is https://github.com/tldr-pages/tldr-lint/pull/6 merged or #1018 is resolved otherwise.

PR #1023, opened 2016-08-30
curl: Add more download and rewrite data examples

ericbn:
- Based on discussion in #1019
- Download: first example introduces `-o`, second one introduces `-O`,
  third example introduces `-L` and `-C` using brackets as per #1018
- Data: first example introduces `-d` (default method is POST and
  default Content-Type is application/x-www-form-urlencoded), second
  example introduces `-X` and `-H`
- Count of examples is still 8 as before

### DIFF
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -3,15 +3,15 @@
 > Transfers data from or to a server.
 > Supports most protocols including HTTP, FTP, POP.
 
-- Download the output of an URL to a file:
+- Download the contents of an URL to a file:
 
 `curl {{http://example.com}} -o {{filename}}`
 
-- Download a file saving the output under the same filename as indicated in the URL:
+- Download a file saving it under the same name as indicated in the URL:
 
 `curl -O {{http://example.com/filename}}`
 
-- Download a file using the filename in the URL, following [L]ocation redirects, and automatically [C]ontinuing/resuming a previous file transfer:
+- Download a file, following [L]ocation redirects, and automatically [C]ontinuing/resuming a previous file transfer:
 
 `curl -O -L -C - {{http://example.com/filename}}`
 
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -3,34 +3,34 @@
 > Transfers data from or to a server.
 > Supports most protocols including HTTP, FTP, POP.
 
-- Download a URL to a file:
+- Download the output of an URL to a file:
 
-`curl "{{URL}}" -o {{filename}}`
+`curl {{http://example.com}} -o {{filename}}`
 
-- Send form-encoded data:
+- Download a file saving the output under the same filename as indicated in the URL:
 
-`curl --data {{name=bob}} {{http://localhost/form}}`
+`curl -O {{http://example.com/filename}}`
 
-- Send JSON data:
+- Following [L]ocation redirects, download a file using the filename in the URL, and automatically [C]ontinuing/resuming a previous file transfer:
 
-`curl -X POST -H "Content-Type: application/json" -d {{'{"name":"bob"}'}} {{http://localhost/login}}`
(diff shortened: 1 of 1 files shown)


ericbn:
@agnivade's comments from #1019 suggesting the removal of `--head`. This example exists since the first commit (11264d9).

> Showing the `--head` option is redundant here immediately after showing the `-X METHOD` option. One can easily do a `-X HEAD` to make a head request.
> They are indeed different. `--head` will print only the response headers. This is usually achieved with the HEAD http method. But you can also use the `--head` option with GET method, in which case it ignores the response body and only prints the headers. This is kind of unecessary IMO.
> I would either do `curl -X HEAD http://example.com` or `curl http://example.com -v -o /dev/null`.
> I vote for removing the `--head` example as it can be achieved with the already given options. And given that this page is at `#10` of most searched commands in mixpanel, we should definitely work on making it as simple as possible.

pages/common/curl.md
```diff
@@ -3,34 +3,34 @@
...
 > Supports most protocols including HTTP, FTP, POP.
 
-- Download a URL to a file:
+- Download the contents of an URL to a file:
 
-`curl "{{URL}}" -o {{filename}}`
+`curl {{http://example.com}} -o {{filename}}`
 
-- Send form-encoded data:
+- Download a file saving it under the same name as indicated in the URL:
```
waldyrious: Here the previous description ("Download a file saving the output under the same filename as indicated in the URL") would be more appropriate, IMO, to call attention to the `-O` option, which is different from `-o`.

Maybe we could shorten it a bit to "Download a file saving the output under the filename indicated by the URL", and we might consider writing [O]utput, to match the next examples.

Thoughts?

pages/common/curl.md
```diff
@@ -3,34 +3,34 @@
...
+`curl -O {{http://example.com/filename}}`
 
-- Send JSON data:
+- Download a file, following [L]ocation redirects, and automatically [C]ontinuing/resuming a previous file transfer:
 
-`curl -X POST -H "Content-Type: application/json" -d {{'{"name":"bob"}'}} {{http://localhost/login}}`
+`curl -O -L -C - {{http://example.com/filename}}`
 
-- Specify an HTTP method:
+- POST application/x-www-form-urlencoded data:
```
waldyrious: We should preserve "Send" in the description, for consistency and clarity.

pages/common/curl.md
```diff
@@ -3,34 +3,34 @@
...
+`curl -d {{'name=bob'}} {{http://example.com/form}}`
 
-- Head request:
+- Send data, specifying a custom HTTP method, and including an extra header:
 
-`curl --head {{http://localhost}}`
+`curl -d {{'{"name":"bob"}'}} -X {{PUT}} -H {{'Content-Type: application/json'}} {{http://example.com/users/1234}}`
 
-- Include an extra header:
+- Head request:
```
waldyrious: Let's just remove this one while we're at it, if we all agree it's redundant (I'll defer to @agnivade's comments on this).

pages/common/curl.md
ericbn: I was thinking that when there is just one parameter in the example, there should be no need for a `t[e]xt`.

pages/common/curl.md
ericbn: I agree, more verbosity here makes is clearer.

pages/common/curl.md
waldyrious: Yeah, I agree the [] could be omitted in such cases.

ericbn:
@waldyrious, removed "Head request"  and improved verbosity as discussed in the previous commits. Also changed "POP" to "POP3".

agnivade:
LGTM.

waldyrious:
LGTM as well. Great work and discussion, everyone, and thanks @ericbn for bearing with us :)

PR #1024, opened 2016-08-30
chmod: Use brackets instead of parenthesis

ericbn:
- To specify which single-letter parameter is being referenced
- As per #1018
- [grep](https://tldr.ostera.io/grep) also uses brackets


waldyrious:
Thanks! This is a straightforward improvement, so I'll merge right away. Also thanks for the reference to grep, I'll add that to #1018 as well.

PR #1026, opened 2016-08-30
wget: Use same example description as for curl

ericbn:
- For "Download a URL to a file" example
- Curl pull req is #1023
- Example titles discussed in #1019
- Also fix example formatting

### DIFF
--- a/pages/common/wget.md
+++ b/pages/common/wget.md
@@ -3,9 +3,9 @@
 > Download files from the Web.
 > Supports HTTP, HTTPS, and FTP.
 
-- Download a URL to a file:
+- Download the output of an URL to a file:
 
-`wget -O filename "{{url}}"`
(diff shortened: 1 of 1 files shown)


waldyrious:
I'm not really fond of the "output of an URL" expression. "Contents of an URL", as initially proposed in [#1019](https://github.com/tldr-pages/tldr/pull/1019#discussion_r76527794), sounds more natural to me. I see the -o here and in curl as referring to sending the output of the *command* into a file, rather than stdout.

Other than that, this change LGTM.

ericbn:
@waldyrious, updated this and `curl` accordingly.

waldyrious:
Thanks! This change seems pretty uncontroversial, and since we're 3 agreeing on the "contents" wording, I'll merge right away.

PR #1027, opened 2016-08-31
ping: add -a example

xxdavid:


agnivade:
We should probably add the following caveat - "It can give beep only from terminal number 1 through 7 and gnome-terminal ( It will not work in console )."

Relevant - http://askubuntu.com/questions/15963/why-cant-i-hear-audible-pings

waldyrious:
@agnivade I'm wondering if we could word the caveat in a more layman-friendly manner...

xxdavid:
`ping -a` uses the standard `\a` character for the bell. It's commonly used (though I can only think of Vim now). I think that if someone's terminal doesn't support the bell, he/she is probably aware of it and won't expect `ping -a` to ring the bell. What do you think?

agnivade:
@xxdavid - Not in my case. I wasn't aware that my terminal doesn't support the bell. If you look at the askubuntu question, it mentions that the "desktop effects" in compiz consumes the bell and therefore, no sound comes out. And also given the fact that it doesn't work in console, which is the major case for a lot of users, I feel this should be mentioned.

@waldyrious - Absolutely, I just copy-pasted the text here.

agnivade:
Hi @xxdavid , any updates on this ?

xxdavid:
What about `Ping host and ring the bell when a packet is received (if your terminal supports it)`?

agnivade:
Sure, please update the PR and we'll see what others have to say.

waldyrious:
You copied the "What about" 😄 please fix that. Other than that, LGTM! 👍

waldyrious:
oops, clicked the wrong button, sorry.

xxdavid:
Yay, sorry. 😄

waldyrious:
@agnivade what do you think? (asking explicitly since lgtm seems to have erased my previous approval)

agnivade:
:+1:

PR #1028, opened 2016-08-31
assimp: add page

waldyrious:


pages/common/assimp.md
```diff
@@ -0,0 +1,28 @@
...
+
+- List all supported import formats:
+
+`assimp listext`
+
+- List all supported export formats:
+
+`assimp listexport`
+
+- Convert a file to one of the supported output formats (see dox_cmd.h for a list of all common parameters):
```
agnivade: Just checking - the user seeing this page will know what does dox_cmd.h refer to right ?

agnivade:
The manpage for assimp on my ubuntu machine does not have the `listexport` option. Probably you can check that for platform compatibility.

waldyrious:
Apparently this was introduced in https://github.com/assimp/assimp/commit/eb4afe8, from 3 April 2011. I have the option in my machine, where `assimp version` returns "Version 3.2". Can you check what version you have installed?

pages/common/assimp.md
waldyrious: Not necessarily, but at least that's something googleable which should point them to the [dox_cmd.h](https://github.com/assimp/assimp/blob/master/doc/dox_cmd.h) file. Ideally, the rendered version (Doxygen output) should be available somewhere, but after a lot of searching, I could only find [this old archived version](https://laboratory.systems/p/apoc3d/source/file/1291/trunk/extra/assimp/doc/assimpcmd_html/common.html) (which doesn't even open in the browser, it triggers a download of the html page). Something closer to [this](http://htmlpreview.github.io/?https://github.com/ElviraBeck/GeKo/blob/master/dependencies/assimp/doc/assimpcmd_html/index.html) (with updated content) should be available either in http://assimp.sourceforge.net/lib_html/ or in https://assimp.github.io/assimp/ (or both!) but since it isn't, I figured at least pointing people to the source code file used to generate the documentation could be helpful, considering that it is easier to find than "the assimp_cmd docs" which the sub-commands' help mentions.

/cc @kimkulling

agnivade:
Ah ok. I have 3.0. The debian repo must be out of date. Did you do by `brew install`, or compiled from source ?

waldyrious:
IIRC I did `brew install`.

agnivade:
I think that's fine. We can leave it at that.

kimkulling:
Wenn habe an uploaded Version aß welk, Interesse?
Kim

Agniva De Sarker <notifications@github.com> schrieb am Mi., 7. Sep. 2016,
12:05:

waldyrious:
@kimkulling I tried to [google-translate your comment](https://translate.google.com/#auto/en/Wenn%20habe%20an%20uploaded%20Version%20a%C3%9F%20welk%2C%20Interesse%3F%0AKim), but I still couldn't figure it out ("If have to updated version ate withered, interest?"). Could you please clarify? :)

agnivade:
re-ping @kimkulling - Could you help us out here ?

kimkulling:
Auto correction: Do you need the latest Version?

Agniva De Sarker <notifications@github.com> schrieb am Sa., 24. Sep. 2016,
07:00:

waldyrious:
@kimkulling thanks for the clarification. I note that [Debian's stable version (jessie) still has the 3.0 version of assimp](https://packages.debian.org/search?keywords=assimp&searchon=names&suite=stable&section=all), but I'm not familiar with the update schedule of packages in Debian stable. Would it be possible to update the package there?

I'd appreciate if you could also comment [here](https://github.com/tldr-pages/tldr/pull/1028#r77702954) :)

waldyrious:
@agnivade any opinions on how to move forward with this?

agnivade:
I am okay with the `listexport` option. But am a little concerned about the `dox_cmd.h` reference. Perhaps, we can add another oneliner pointing people to the right stuff.

waldyrious:
@agnivade could you elaborate a bit on what the extra oneliner could look like?

agnivade:
You might say "look for the dox_cmd.h file in the source code github repo for assimp"

waldyrious:
@agnivade I took a stab at it, what do you think?

agnivade:
:+1:

waldyrious:
Ok I think the discussion has been open long enough for any additional concerns to be presented, so in the absence of these, I'll add a LGTM myself.

PR #1030, opened 2016-08-31
zip: make it more evident where the x comes from

waldyrious:
c.f. #782

### DIFF
--- a/pages/common/zip.md
+++ b/pages/common/zip.md
@@ -14,6 +14,6 @@
 
 `zip -d {{compressed.zip}} "{{foo/*.tmp}}"`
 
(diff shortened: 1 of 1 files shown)


agnivade:
Can we add the labels for the other flags too ? Like 'recursive' for `-r` and change "Remove" to "Delete" for `-d`.

waldyrious:
Yes, I thought about that. I'll work on this later today.

waldyrious:
@agnivade PTAL

agnivade:
perfect.

waldyrious:
The travis failure is because of the []. We need to edit the linter to allow this.

PR #1031, opened 2016-09-01
bash: add --version

waldyrious:


agnivade:
The added example looks good. I wanted to add

agnivade:
The added example looks good. I just wanted to put a comment on L12(which github wasn't allowing) that we should enclose the `{{command}}` in quotes `"{{command}}"`.

For eg - for echo hello to work, we need `bash -c "echo hello"`. Of course, within the command, we need to double escape if needed. But I guess that is not important here.

waldyrious:
Sure, I'll make that change.

agnivade:
LGTM. :+1:

PR #1034, opened 2016-09-02
du: rework page, add glob matching example (fixes #1032)

### LINKED ISSUE #1032: du: add pattern matching example
Get the total size of .jpg files at any depth within the current directory: `du -ch -- **/*.jpg | grep total$` ([source](http://unix.stackexchange.com/a/41635/30336)).

This [requires bash 4.0](http://unix.stackexchange.com/questions/34024/what-is-the-tool-that-allows-me-to-specify-arbitrary-directories-using#comment46158_34026), so I can't test it at the moment in OS X (`GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin14)`) -- if I try this the `**` does the same as `*`, i.e. it only matches a single depth level.

If someone can confirm the multi-level `**` behavior for this example in a shell that supports it, please say so.

waldyrious:


agnivade:
I don't have any issues with the example as such, but somehow I feel the grepping of "total" here is a bit hackish and deviating a bit from our motto of letting the user learn about the command slowly. This can break anytime depending on the output of the `du` command. We could do `tail -1` also   but even that doesn't feel very elegant. My vote is neutral on this.

waldyrious:
I hear you. Since the `-c` option specifically means "produce a grand total", I don't think tail -1 would be a bad idea. Otherwise, we could keep grep and use the expanded command name `--total`. Do you have any preference regarding these two options?

agnivade:
In that case, I prefer `tail -1`. 

Would love to see @ostera's take on this.

ostera:
Not trying to be a buzz kill, but I don't like this addition, considering its bash specific and requires special global options set.

Any chance it works elsewhere and on defaults?

waldyrious:
We could forgo the multi-level glob and just use `*.jpg`, sticking to files within the current directory. Or we could demonstrate a specific depth level, say `*/*.jpg`, which is more versatile and not too complex. I'd rather have the latter, but let me know what you think.

ostera:
After some thought and experimenting, I'd  still prefer `du -ch */*.jpg`. No pipe.

Note that the extra `--` is not necessary, and I feel that while using `tail` is far conciser and the `grep` command seemingly being better known (as per tld.jsx searches), the output can't be that bad since the total will still show at the bottom (and thus by the cursor) when it finishes running.

screen shot 2016-09-06 at 8 02 41 am

Ultimately, we could use `du -chs {{path}}`, which gives me something like this:

```zsh
~ λ du -chs ./Downloads
9.3G   	./Downloads
9.3G   	total
```

Or simply `du -hs {{path}}`:

```zsh
~ λ du -hs ./Downloads
9.3G   	./Downloads
```

But unfortunately this will print a summary _per argument_, meaning that using a glob to find all the jpeg's in a tree will _need_ the `-c` flag and either `tail` or `grep` for any humongous output.

Something tells me we should let people figure the tail/grep part by themselves if they start getting bothered by the long outputs.

waldyrious:
Sounds good to me, and thanks for the careful analysis :) I'll make the change.

waldyrious:
I went ahead and made a bunch of other fixes while I was at it; please take a look and let me know what you think.

pages/linux/du.md
```diff
@@ -1,23 +1,27 @@
...
+`du -k {{file_or_folder}}`
 
-- List file sizes of a directory and any subdirectories in MB:
+- List file sizes of a folder and any subfolders, in MB:
 
-`du -m {{file_or_directory}}`
+`du -m {{file_or_folder}}`
 
-- Get recursively, individual file/folder sizes in human readable form:
+- Get individual file/folder sizes, recursively, in human readable form:
```
agnivade: Can we include "all" in the description, for `-a` ?

waldyrious: Sure. I wasn't too happy with this wording anyway.

waldyrious:
I made a bunch of other edits again 😅 please check.

waldyrious:
I was forgetting to also edit the osx page. I now have synchronized both pages with the proposed edits. Here's how they differ: https://www.diffchecker.com/xKrezaBt

pages/linux/du.md
```diff
@@ -1,23 +1,27 @@
...
+> Disk usage: estimate and summarize file and folder space usage.
 
-- Get a sum of the total size of a file/folder in human readable units:
+- List the sizes of a folder and any subfolders, in the given unit (B/KB/MB):
 
-`du -sh {{file_or_directory}}`
+`du -{{b|k|m}} {{path/to/folder}}`
 
-- List file sizes of a directory and any subdirectories in KB:
+- List the sizes of a folder and any subfolders, in human-readable form (i.e. auto-selecting the appropriate unit for each size):
```
agnivade: I don't think the "i.e." part is needed. The `-h` flag is used at other pages too, and "human-readable form" looks good to me. :)

pages/linux/du.md
```diff
@@ -1,23 +1,27 @@
...
+`du -h {{path/to/folder}}`
 
-- List file sizes of a directory and any subdirectories in MB:
+- Show the size of a single folder, in human readable units:
 
-`du -m {{file_or_directory}}`
+`du -sh {{path/to/folder}}`
 
-- Get recursively, individual file/folder sizes in human readable form:
+- List the human-readable sizes of a folder and of all the files and folders within it (recursively):
```
agnivade: Why is "recursively" in brackets ?

waldyrious: Purely to make the sentence length and structure less daunting. It is semantically an annotation, rather than something that needs to be part of the sentence, so marking it as such gives a visual clue that it doesn't need to be loaded into working memory in order to process the preceding sentence. I don't know, I figure it makes it easier to read. Does that make any sense or am I blabbering? :P

pages/linux/du.md
waldyrious: I just like to seize opportunities to expose the mechanics behind terms that are otherwise somewhat subjective :) besides, it matches the previous description to better present the case that -h should always be used unless a specific size is needed. This is important because in OS X for example the default unit is the archaic 512 bytes, which may throw people off guard -- I know it did me.

agnivade: understood.

pages/linux/du.md
agnivade: I would say that "folder and of all the files and folders within it" already means "everything" inside it. Following your line of reasoning, since recursively doesn't "need" to be a part of the sentence, I would say it doesn't need to be there at all. Since the meaning is already captured. Just my 2 cents. Let me know what you think.

waldyrious: It is not entirely redundant since "folder and of all the files and folders within it" could be interpreted as the first level only, but yeah, I'm not opposed to removing it since it's not a big deal anyway. I'll remove since we have more to gain from the simplicity than from perfect explicitness in this case.

agnivade:
LGTM.

waldyrious:
I wonder if we can approve our own PRs, as project maintainers. Let's figure out: LGTM

waldyrious:
Yay, it works :)

PR #1037, opened 2016-09-04
mongorestore: add page

pdey:


waldyrious:
Solid entry, nice work! The only thing I'd change is "containing bson data dump" --> "containing a bson data dump".

pages/common/mongorestore.md
```diff
@@ -0,0 +1,19 @@
+# mongorestore
+
+> Utility to import a collection or database from a binary dump into a MongoDB instance.
+
+- Import database from folder containing a bson data dump:
```
agnivade: The language can be improved a bit. Its not very clear that the data is being imported "to" the database. I would love a "from" -> "to" sentence. How about "Import a bson data dump from a folder to a given database" ?

pages/common/mongorestore.md
```diff
@@ -0,0 +1,19 @@
...
+
+`mongorestore -h {{database_host}} -d {{database_name}} {{path/to/folder}}`
+
+- Import database with user authentication from a folder containing bson data dump:
```
agnivade: Maintaining parity with other sql-ish(see `mongo`, `myql`) pages, its best to prompt user for passwords rather than type it on the command prompt. The man page says to prompt for password, one should just drop the `--password` flag.

How does this sound -  "Import a bson data dump from a folder to a given database on a given host with a given username; user will be prompted for password" ? And we can just drop the `-p {{password}}` part.

agnivade:
A couple of points -

1. We try to use the spelled out version of the flags of a command. https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines. Lets keep it consistent with mongo and mongodump and use `--host`, `--db` etc.

2. The same options are repeated again for the last example. The way we write these pages are introduce different options in different examples and let the user slowly build up the full command. 
So, I would suggest for the first example to drop the `-h` as that's used in all the following examples. Then introduce --host,--port in the 2nd one. After that, just --username in the 3rd one. Now we can have an example for a different use case, combining all the previous options. I would recommend also adding the --collection flag since we have space for some more examples.

  This is how I would build up the flags

  ```
1.  --db
2. --host, --port, --db
3. --username, --db
4. --host, --username, --port, --db (import from file)
5. --collection, --db (import from folder, only to a particular collection)
```

**Comment by pdey** [comment]
Agree with most of the points. I feel --collection option for mongorestore need not be included here as it is rarely used ( such as collection name conflicts that will arise only if an exisiting database is merged with dump from another database).

File: pages/common/mongorestore.md [hunk 3]
```diff
@@ -0,0 +1,19 @@
+# mongorestore
+
+> Utility to import a collection or database from a binary dump into a MongoDB instance.
+
+- Import a bson data dump from a folder to a mongodb instance:
+
+`mongorestore --host {{database_host}} --port {{port}} --db {{database_name}} {{path/to/folder}}`
```
agnivade: I feel `--host` and `--port` can be dropped here since they are anyways used in the next examples. Lets lay it slowly on the user :smile:

pages/common/mongorestore.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Import a bson data dump from a folder to a mongodb instance with user authentication; user will be prompted for password:
+
+`mongorestore -h {{database_host:port}}  -d {{database_name}} --username {{username}} {{path/to/folder}} --password`
```
agnivade: Please expand the `-h` and `-d` flags.  And also mention in the description that the data is going to be imported to a given database host running at a given port. We usually try to match the words in the description with the flags.

pages/common/mongorestore.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Import a collection from a bson file to a mongodb instance:
+
+`mongorestore --host {{database_host}} --port {{port}} --db {{database_name}} {{path/to/file}}`
+
+- Import a collection from a bson file to a mongodb instance with user authentication; user will be prompted for password:
+
+`mongorestore -h {{database_host:port}} -d {{database_name}} -u {{username}} {{path/to/file}} -p`
```
agnivade: Same here, please expand the `-h`, `-d` and `-p` flags.

agnivade:
LGTM from my side. Over to @waldyrious for any final comments.

pages/common/mongorestore.md
```diff
@@ -0,0 +1,19 @@
+# mongorestore
+
+> Utility to import a collection or database from a binary dump into a MongoDB instance.
+
+- Import a bson data dump from a folder to a mongodb instance:
+
+`mongorestore --db {{database_name}} {{path/to/folder}}`
+
+- Import a bson data dump from a folder to a given mongodb database host running at a given port with user authentication; user will be prompted for password:
```
waldyrious: This is a bit of a mouthful. We should harness punctuation to better guide reading flow, e.g.

`- Import a bson data dump from a folder to a given mongodb database host, running at a given port, with user authentication (user will be prompted for password):`

pages/common/mongorestore.md
```diff
@@ -0,0 +1,19 @@
...
+
+`mongorestore --host {{database_host:port}}  --db {{database_name}} --username {{username}} {{path/to/folder}} --password`
```
waldyrious: There are two spaces before the --db option.

pages/common/mongorestore.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Import a collection from a bson file to a mongodb instance:
+
+`mongorestore --db {{database_name}} {{path/to/file}}`
+
+- Import a collection from a bson file to a given mongodb database host running at a given port with user authentication; user will be prompted for password:
```
waldyrious: Same comment as above.

waldyrious:
I've left some inline comments. Additionally, I would suggest the "MongoDB" capitalization to be used in all descriptions, for correctness and to match the form used in the command summary.

pages/common/mongorestore.md
pdey: Its true that multiple database can run on the same host, however one port can be bound to only one service (database or any other service accessible via network interface). Typically when someone tries to access a database hosted on a remote machine, and this is the typical usecase. The previous command omits the host and port options, and this works locally on the same machine where database is running. I hope these two combined together covers basic use cases.

waldyrious: If I'm reading you correctly, by specifying the host and the port, the --db option should be redundant, right? Could one say it works as a check to ensure one's connecting to the desired database and not another that happens to be hosted in the same host and port? I mean, what happens if you try to connect to a db giving the correct host address, port, and authentication, but a mismatching db name? Does the command fail?

And in the opposite case: if the port isn't provided in the command, and say the database isn't hosted in the default port that the command uses to connect; I suppose the --db parameter is not sufficient to make it work, right?

If both these assumptions are correct, then the --db option is redundant at best, and useless at worst, so I'd suggest omitting it since the command is quite long already. Otherwise, please correct any misconceptions on my part :)

pdey: Ah I see!.. i get your point. One mongodb server can run on a single port only but multiple (MongoDB) databases can be hosted on this server, hence we need to specify the database name ( --db option). Sorry for the confusion.

waldyrious: Oh, ok, got it, thanks :)
Do you think it would be clearer if the description said "mongodb server" instead of "mongodb database host"?

pdey: Makes total sense!.. will push and update.

waldyrious:
I'm happy with how the page turned out. Thanks again for a great contribution!

PR #1038, opened 2016-09-05
less, more: Update page up/down shortcut

ericbn:
- Previous `D`/`d` refers to scrolling half of a page, not a full page

### DIFF
--- a/pages/common/less.md
+++ b/pages/common/less.md
@@ -10,7 +10,7 @@
 
 - Page up / down:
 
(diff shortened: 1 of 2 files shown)


agnivade:
Ahh so this is it ! I always use PgUp/PgDn on my ubuntu box. And always got frustrated on a mac box where there's no PgUp/PgDn.

LGTM. :+1:

waldyrious:
Cool, thanks for the fix.

PR #421, opened 2015-12-28
Add new page about xargs command

jrziviani:
- xargs is a command that uses the standard input to create parameters
   for a command to execute.

Signed-off-by: Jose Ricardo Ziviani <jrziviani@gmail.com>

PR #1039, opened 2016-09-05
xargs: rework page to make it more beginner-friendly

waldyrious:
This page has been bugging me for a while and I have come across some bumps while using the command. I finally decided to dive in and learn about how it works, and this is the result. Hopefully the page is now more readable, intuitive and useful.

### DIFF
--- a/pages/common/xargs.md
+++ b/pages/common/xargs.md
@@ -1,19 +1,20 @@
 # xargs
 
-> Execute a command with piped arguments.
+> Execute a command with piped arguments coming from another command, a file, etc.
+> The input is treated as a single block of text and split into separate arguments on spaces, tabs, newlines and end-of-file.
 
-- Main use:
+- Main usage pattern:
 
-`{{arguments}} | xargs {{command}}`
+`{{arguments_source}} | xargs {{command}}`
 
-- Specific example: delete all files that start with 'M':
+- Delete all files with a `.backup` extension:
 
-`find . -name 'M*' | xargs rm`
+`{{find . -name '*.backup'}} | xargs {{rm -v}}`
 
-- Handle whitespace in arguments:
+- Use the NUL (`\0`) character as separator (useful to split the input on newlines but not on spaces):
 
-`{{arguments_null_terminated}} | xargs -0 {{command}}`
+`{{arguments_source}} | tr '\n' '\0' | xargs -0 {{command}}`
 
-- Insert arguments at chosen position, using '%' as the placeholder marker:
+- Execute the command once for each input line, replacing any occurrences of the placeholder (here marked as `_`) with the input line:
 
-`{{arguments}} | xargs -I '%' {{command}} % {{extra_arguments}}`
+`{{arguments_source}} | xargs -I _ {{command}} _ {{optional_extra_arguments}}`
--- a/pages/common/xargs.md
+++ b/pages/common/xargs.md
@@ -11,7 +11,7 @@
 
 `{{find . -name '*.backup'}} | xargs {{rm -v}}`
 
-- Use the NUL (`\0`) character as separator (useful to split the input on newlines but not on spaces):
+- Convert newlines in the input into NUL (`\0`) characters, and split on those only (useful if the input to xargs contains spaces):
 
 `{{arguments_source}} | tr '\n' '\0' | xargs -0 {{command}}`
 


pages/common/xargs.md
```diff
@@ -1,19 +1,20 @@
...
+`{{arguments_source}} | xargs {{command}}`
 
-- Specific example: delete all files that start with 'M':
+- Delete all files with a `.backup` extension:
 
-`find . -name 'M*' | xargs rm`
+`{{find . -name '*.backup'}} | xargs {{rm -v}}`
 
-- Handle whitespace in arguments:
+- Use the NUL (`\0`) character as separator (useful to split the input on newlines but not on spaces):
```
waldyrious: I was wondering if we should mention here that the simpler `xargs -I _ {{command}} _` also works for the same end, albeit with a different behavior (see http://stackoverflow.com/a/39335402/266309)

waldyrious: Also, both workarounds would be better served by the -d option, as pointed out by @jrziviani in #421, but that option does not exist in the version of xargs shipped with OSX, so we'd have to split the page, which sucks but may be the cleanest approach (see [osx/tree.md](../blob/master/pages/osx/tree.md) vs [linux/tree.md](../blob/master/pages/linux/tree.md), particularly in light of #990).

ostera: Perhaps better phrased as `Split the input on new lines (using the NUL character as separator)`, so we make _what_ it's doing very clearly. The _how_ is less important.

waldyrious: I'm inclined to disagree. The "what" is important, but if we don't emphasise the "how" (indeed, "why") it works, we risk defrauding our mission to allow people to *learn* the syntax of commands by example as opposed to memorizing obscure formulas. This was the problem in the previous description which made it hard for me to get a grasp on what was going on until I read the manpage -- which, by the way, describes the -0 option as "Change xargs to expect NUL (``\0'') characters as separators, instead of spaces and newlines." -- which is literally what it does, and is quite accessible already, IMO.

That said, it's not that the description you propose fails to explain the how, but I fear that the two phrases "split input in newlines" and "using the NUL character as separator" may make the understanding harder, since \n and \0 are different characters and xargs is explicitly using the latter -- the conversion is done by the user, but xargs is blind to that.

I'm not sure I'm being clear. Let me know if I'm rambling or if I need to clarify some of my points.

agnivade: Frankly, I am okay with both. @waldyrious' style is a bit more verbose, @ostera's version is a bit succinct. 

I think what would really help (IMO) is to explain in what scenario this option shines. And that is when the input contains spaces. Its a bit hard to grasp it without a concrete example. This would be my version -

```
Use the NUL (`\0`) character as separator (useful if a filename contains a space):
find . -type f | tr '\n' '\0' | xargs -0 {{command}}
```

I think this clearly explains "why" the `-0` flag is helpful. Just my 2 cents.

waldyrious: I'd support something very close to that, @agnivade. How does this sound?

    - Use the NUL (`\0`) character as separator (useful if the input to xargs contains spaces):
    
    `find . -type f | tr '\n' '\0' | xargs -0 {{command}}`

waldyrious: Re-reading @ostera's comment with a fresh mind, I understand the value on specifying that the input is split on newlines, which isn't indicated on neither mine nor @agnivade's proposals. I would like to come up with a good compromise. Any suggestions?

waldyrious: Also, find is not a good concrete example to demonstrate this, since -0 was precisely made to work with find's `-print0` 😄

agnivade: >  I understand the value on specifying that the input is split on newlines

We can say `Use the NUL (`\0`) character as separator to split input lines (useful if the input to xargs contains spaces):`
> since xargs' -0 was precisely made to work with find's -print0

That is true actually. Well, there is no harm in giving `find ... -print0` itself, we just have to be a bit more verbose and explain what `print0` does. And practically speaking, `xargs -0` is most used along with `find ... -print0`. If not, any other command is also fine by me.

waldyrious: I'd rather show a generic way this can work with any command, than rely on `find`'s custom behavior. I would keep the current example intact, i.e.

    `{{arguments_source}} | tr '\n' '\0' | xargs -0 {{command}}`

As for the wording... how about "Convert newlines in the input into NUL (\0) characters, and split on those only (useful if the input to xargs contains spaces)"? This, coupled with the main description, should provide both the "what" and the "how".

agnivade: Agree. Sounds good.

waldyrious: Yay! @ostera what about you?

agnivade: ping @ostera.

waldyrious:
Given the approval of both @ostera and @agnivade above, I'll merge even though those weren't declared in ways that the LGTM service recognizes.

PR #1040, opened 2016-09-07
launchctl: add page

waldyrious:
Built with info from several tutorials online. Unfortunately I seem to have discarded the links :(

### DIFF
--- a/pages/osx/launchctl.md
+++ b/pages/osx/launchctl.md
@@ -1,0 +1,32 @@
+# launchctl
+
+> A command-line interface to Apple's `launchd` manager for launch daemons (system-wide services) and launch agents (per-user programs).
+> `launchd` loads XML-based *.plist files placed in the appropriate locations, and runs the corresponding commands according to their defined schedule.
+
+- Activate a user-specific agent to be loaded into `launchd` whenever the user logs in:
+
+`launchctl load ~/Library/LaunchAgents/{{my_script}}.plist`
+
+- Activate an agent which requires root privileges to run and/or should be loaded whenever any user logs in (note the absence of `~` in the path):
+
+`sudo launchctl load /Library/LaunchAgents/{{root_script}}.plist`
+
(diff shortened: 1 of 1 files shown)


pages/osx/launchctl.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Activate a system-wide daemon to be loaded whenever the system boots up (even if no user logs in):
+
+`sudo launchctl load /Library/LaunchDaemons/{{system_daemon}}.plist`
+
+- Show all loaded agents/daemons, with the PID if the process they specify is currently running, and the exit code returned the last time they ran:
+
+`launchctl list`
+
+- Unload a currently loaded agent, e.g. to make changes (note: the plist is automatically loaded into `launchd` after a reboot and/or logging in):
```
agnivade: Doesn't "the plist file" sound better ? plist is just the extension right ?

And also, "e.g. to make changes" seems obvious to me. Why else would you unload an agent ? Either to stop it or to make changes.

waldyrious: It may be better to make it explicit that you must (I suppose?) unload an agent before changing the plist file, e.g. in case it is locked by the system, or would cause conflicts, or the changes wouldn't apply without unload/reload. Actually, I don't know which of these (if any) is the case, I just took up the wording from the tutorials I read. I would guess that it's the latter meaning, so we could reword this to make it clearer. Do you agree that the last option seems more plausible?

As for "the plist file", agreed.

agnivade: >  or the changes wouldn't apply without unload/reload. 

Yes, that might be the case. I feel its obvious. But sure, go ahead. Its always good to make it clearer.

waldyrious: I can't think of a way to be more explicit without being more verbose... I guess I'll leave it as it is for now, and if a better wording comes up, we can change it :)

agnivade:
LGTM.

waldyrious:
Great. I'll add my own LGTM as I'm a maintainer, so we can get this one behind us :P

PR #1041, opened 2016-09-07
cron: Add job definition example

agnivade:
Closes #822

### DIFF
--- a/pages/common/crontab.md
+++ b/pages/common/crontab.md
@@ -13,3 +13,16 @@
 - Remove all cron jobs for the current user:
 
 `crontab -r`
+
+- Configuration file:
+
+```
+# ┌───────────── min (0 - 59)
+# │ ┌────────────── hour (0 - 23)
+# │ │ ┌─────────────── day of month (1 - 31)
+# │ │ │ ┌──────────────── month (1 - 12)
+# │ │ │ │ ┌───────────────── day of week (0 - 6) (0 to 6 are Sunday to Saturday, or use names; 7 is Sunday, the same as 0)
+# │ │ │ │ │
+# │ │ │ │ │
+# * * * * *  command to execute
+```
--- a/pages/common/crontab.md
+++ b/pages/common/crontab.md
@@ -14,15 +14,18 @@
 
 `crontab -r`
 
-- Configuration file:
+- Cron job definition format:
 
-```
-# ┌───────────── min (0 - 59)
-# │ ┌────────────── hour (0 - 23)
-# │ │ ┌─────────────── day of month (1 - 31)
-# │ │ │ ┌──────────────── month (1 - 12)
-# │ │ │ │ ┌───────────────── day of week (0 - 6) (0 to 6 are Sunday to Saturday, or use names; 7 is Sunday, the same as 0)
-# │ │ │ │ │
-# │ │ │ │ │
-# * * * * *  command to execute
-```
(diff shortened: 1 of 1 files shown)


agnivade:
This is the best I could manage trying to incorporate everyone's suggestions within the framework that we have. :stuck_out_tongue_closed_eyes:

pages/common/crontab.md
```diff
@@ -13,3 +13,19 @@
 - Remove all cron jobs for the current user:
 
 `crontab -r`
+
+- Cron job definition format:
+
+`{{minute}} {{hour}} {{day_of_month}} {{month}} {{day_of_week}} {{command_to_execute}}`
```
waldyrious: How about adding this as a second line in the main description? (without the heavy token marks, maybe parenthesis or brackets)

waldyrious:
Very nice! I think we're in a good path here :) left one inlnine comment. Apart from that, I like this!

waldyrious:
btw, it was nice of you to preserve @kkk777kkk's commit, even though it's so different from what we're going for now :)

pages/common/crontab.md
```diff
@@ -13,3 +13,19 @@
...
+
+- Sample job which runs at 10:00 every day. * means any value:
+
+`0 10 * * * {{path/to/script.sh}}`
+
+- Sample job which runs every minute on the 3rd of April:
+
+`* * 3 4 * {{path/to/script.sh}}`
+
+- Sample job which runs at 02:30 every friday:
```
waldyrious: We definitely should mention the range of values (Monday=1 Saturday=6, Sunday=0/7).

We should also have an example using a weekday name (can we use names for other entries, e.g. months?).

agnivade: > can we use names for other entries, e.g. months?

What do you mean ? I already have an example for April.

waldyrious: I meant writing "april", "apr" or something like that, instead of "4" :)

agnivade: Got it. So if we are not doing nos., I feel there is no need to mention ranges.

waldyrious: I wasn't sure we should outright omit that numbers can be used, since that's a more convenient format, but I guess it's a good thing if we encourage crontabs out there to start becoming more legible xD I view that in the same light as our chmod page not mentioning octal markup, so there's precedent :) go ahead and remove the numbers and ranges where possible.

agnivade:
@waldyrious - PTAL.

pages/common/crontab.md
```diff
@@ -1,6 +1,7 @@
 # crontab
 
 > Schedule cron jobs to run on a time interval for the current user.
+> Job definition format - "(min) (hour) (day_of_month) (month) (day_of_week) command_to_execute".
```
waldyrious: I'd use `: ` instead of ` - `.

waldyrious:
looks good :) I left one minor comment inline.

ostera:
Besides @waldyrious minor correction, here's my LGTM

waldyrious:
LGTM 😄

waldyrious:
Re-triggering status check: LGTM

agnivade:
:+1:

waldyrious:
At last!

cooervo:
is there a tldr for cron? I didn't find it when:
`tldr cron`

agnivade:
There isn't. This is for `crontab`.

sbrl:
@cooervo You're welcome to contribute one :-)

PR #1043, opened 2016-09-08
date: add unix epoch timestamp

danielcompton:


agnivade:
Interesting, my version of `date` doesn't have a `-j` option at all. I am using 8.23 on an Ubuntu machine. What version are you using and in what platform ?

danielcompton:
I've removed the `-j` as I don't think it's needed here.

agnivade:
Can you check your commit ? I still see the `-j` :relaxed:

danielcompton:
Fixed it.

agnivade:
Thanks. 

I'm afraid this seems to be an OSX specific switch. On my machine -r refers to -
```
-r, --reference=FILE
              display the last modification time of FILE
```

Therefore, I get 
```
$date -r 1473305798
date: 1473305798: No such file or directory
```

On linux, the way to do this is - `date -d @1473305798`. So, you would have to either split this page into linux and osx folders with the different switches for each page, or leave out this switch.

danielcompton:
Hmm, yeah it does look like they have differing behaviours. I've split this out into two pages.

pages/osx/date.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Display the date in UTC and ISO 8601 format:
+
+`date -u +"%Y-%m-%dT%H:%M:%SZ"`
+
+- Display the number of seconds since the unix epoch:
+
+`date +%s`
+
+- Convert a unix expoch timestamp to a date:
```
agnivade: We usually try to include some words in the description that match the flags in the example. Can you update the description to contain a `j` and `r` ?

https://github.com/tldr-pages/tldr/issues/1018#issue-173587218

danielcompton: Neither `j` nor `r` are particularly illuminating flags in the man page:

screenshot of iterm2 8-09-16 9-55-54 pm

waldyrious: We can be creative and use words that have (ideally start with) those letters, even if that wasn't the origin of the abbreviation, as long as the meaning is made clearer for tldr users :)

danielcompton:
Updated, though it was a bit of a stretch :)

agnivade:
Its a stretch indeed. :smile:  @waldyrious - what do you think about this ?

waldyrious:
Hahah, points for the creativity! But yeah, I don't think forcing it that much makes things clearer to the reader. To be honest, I'm not sure I get what the -j does, even after checking the previous example description and the manpage option explanation. Do any of you think you can put it in other words just so we can be sure we are on the same page?

agnivade:
Nope .. :worried: Its totally out of whack from the explanation. I guess we have to let go of our convention for this one.

waldyrious:
It's not even about keeping the convention or not at this point -- as an explainer project, we certainly must know what we are explaining before we can simplify it!

agnivade:
Oops, I misunderstood you :expressionless:.

I think I vaguely understand it. You can use the `date` command to either set or display the date in a format. Usually you can use the `+` to format the date, but to use the `-f` flag to convert date from one format to other, you need to pass the `-j` flag. 

Pretty useless, if you ask me. I think its better not to show the `-j` option at all, as its causing more confusion than helping users at this point. And like @danielcompton said, `-j` is not even needed to show a unix epoch timestamp.

waldyrious:
Thanks for the explanation. I still don't get what we gain by using `-f` rather than the normal string format. In fact, the only example my (OS X) manpage gives of using the -f option is converting date's output to unix timestamp:

    `date -j -f "%a %b %d %T %Z %Y" "`date`" "+%s"`

...but that is already possible with the much simpler `date +%s`, so I really don't see the point.

I would perhaps tweak the last two examples to show how one can convert from date's output to unix epoch (`date +%s`), and vice-versa (`date -r 147330579`). i.e. they'd be worded as a pair of complementary commands. This would allow us to describe `-s` as "seconds" and `-r` as "reverse" :) What do you guys think?

agnivade:
Yep, makes sense to me.

agnivade:
@waldyrious - I took the opportunity to do the changes myself. How does it look to you ?

waldyrious:
I would have liked the last two example descriptions to be more symmetric, per my previous comment:

> convert from date's output to unix epoch (date +%s), and vice-versa (date -r 147330579). i.e. they'd be **worded as a pair of complementary commands**.

I can take a stab at it later if you prefer.

agnivade:
Sure, go ahead.

waldyrious:
@agnivade @danielcompton please take a look :) After playing a bit with these descriptions, I took another approach which I believe may work even better.

agnivade:
:+1:

waldyrious:
Awesome, let's wait for @danielcompton to comment.

danielcompton:
LGTM

waldyrious:
Great! Let's get this merged then :) :shipit:

PR #1045, opened 2016-09-09
hub: Add fork and pull-request examples

ericbn:
- Examples follow clone examples as a logical sequence

### DIFF
--- a/pages/common/crontab.md
+++ b/pages/common/crontab.md
@@ -1,6 +1,7 @@
 # crontab
 
 > Schedule cron jobs to run on a time interval for the current user.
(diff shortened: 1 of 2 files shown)


waldyrious:
@ericbn did my commit to master mess up your branch in ways it shouldn't have? Or was it the #1041 merge? Not sure what happened here.

ericbn:
Yeah, I was experimenting with some zsh aliases for git and I think I did mess up my branch... Guess I did `gco` (git checkout) instead of `gc` (git commit) once... Sorry.  :-)

waldyrious:
Ah ok. Just checking if the mistake was on my side so I could avoid it in the future 👍

PR #1046, opened 2016-09-09
hub: Add fork and pull-request examples

ericbn:
- Examples follow clone examples as a logical sequence
- Pull-request example stresses that one must first push to own fork
- This pull request created using the new example :-)

### DIFF
--- a/pages/common/hub.md
+++ b/pages/common/hub.md
@@ -11,11 +11,11 @@
 
 `hub clone {{github_username}}/{{repo_name}}`
 
-- Fork your own copy of a repository cloned from another user (creates a remote named after your GitHub username):
+- Fork your own copy of a repository cloned from another user (must be in the cloned repository folder, creates a remote named after your GitHub username):
 
 `hub fork`
 
-- Create a pull request from your fork:
+- Create a pull request from your fork (first pushing the changes to your fork remote, named after your GitHub username):
 
 `git push {{your_github_username}} {{current_branch}} && hub pull-request`
 
--- a/pages/common/hub.md
+++ b/pages/common/hub.md
@@ -7,17 +7,17 @@
 
 `hub clone {{repo_name}}`
 
-- Clone another user repository:
+- Clone another user's repository:
 
 `hub clone {{github_username}}/{{repo_name}}`
 
-- Fork repository cloned from another user (creates a remote named after your GitHub username):
+- Fork your own copy of a repository cloned from another user (creates a remote named after your GitHub username):
 
 `hub fork`
 
 - Create a pull request from your fork:
 
-`git push {{github_username}} && hub pull-request`
+`git push {{your_github_username}} {{current_branch}} && hub pull-request`
 
 - Open the current project's issues page:
 
--- a/pages/common/hub.md
+++ b/pages/common/hub.md
@@ -11,6 +11,14 @@
 
 `hub clone {{github_username}}/{{repo_name}}`
 
+- Fork repository cloned from another user (creates a remote named after your GitHub username):
+
+`hub fork`
+
+- Create a pull request from your fork:
+
+`git push {{github_username}} && hub pull-request`
+
 - Open the current project's issues page:
 
 `hub browse -- issues`


pages/common/hub.md
```diff
@@ -11,6 +11,14 @@
 
 `hub clone {{github_username}}/{{repo_name}}`
 
+- Fork repository cloned from another user (creates a remote named after your GitHub username):
```
waldyrious: This is a bit confusing, especially since fork and clone have closely related meaning. I'm wondering if we could rewords and/or mention a specific example of what would happen (in fact, I assume the command would interactively ask for information -- maybe we could mention what info the user would need to provide).

pages/common/hub.md
```diff
@@ -11,6 +11,14 @@
...
+
+`hub fork`
+
+- Create a pull request from your fork:
+
+`git push {{github_username}} && hub pull-request`
```
waldyrious: I'm not sure combining these commands is wise. The first one is usual git workflow, but it threw me back for a bit because I wondered if it had a special meaning here. In fact, does github_username here mean that the previous command created a branch named after your GH username? Otherwise I'm not sure I follow what's going on here.

pages/common/hub.md
```diff
@@ -11,6 +11,14 @@
```
waldyrious: while you're at it, you could fix "another user repository" --> "another user's repository" as well :)

waldyrious:
Awesome, I've been looking forward to contributions related to this. I've left some inline comments.

pages/common/hub.md
ericbn: `hub fork` does not ask for any extra information. You must be in a folder where the remote `origin` is the original GitHub repo (e.g. `https://github.com/tldr-pages/tldr.git`), then it will create a fork under your account and add it as a new remote named after your username (in my case, `ericbn 	git@github.com:ericbn/tldr.git`). Kind of complex to explain, but basically an automation of what you would get with the Fork button in GitHub...

pages/common/hub.md
ericbn: > does github_username here mean that the previous command created a branch named after your GH username? 

Not a branch, but a remote name. `git push origin` would push to the original repo, while `git push ericbn` would push to my fork, in my case. Maybe we can better say:

    git push {{github_username}} {{current_branch}}

`hub pull-request` asks for the pull request description, and you can manually state the base and head with `hub pull-request -b {{base}} -h {{head}}` but I think that's even more complex to explain. It has good defaults, so if I pushed the current branch to my fork before, it would assume the pull request is `ericbn:current_branch -> tldr:master`...

ericbn:
I'd say `hub fork` and `hub pull-request` are some of the great things `hub` can do, besides expanding repo names from shorter `tldr-pages/tldr` or `[my_user/]tldr` forms. Maybe we can use the `tldr-pages/tldr` repo as to give more concrete examples...

waldyrious:
Can you try adding the "another user's repository" to see if the clahub check goes away?

As for the rest of the comments, I'll be quite busy for the next two days (hackathon), but after that I'll try to read and respond

pages/common/hub.md
agnivade: How is the name of the remote decided ? Its slightly confusing when you say "username" when it actually refers to the remote. Maybe we can add some more explanation and say {{remote_name}} in the example.

pages/common/hub.md
```diff
@@ -7,10 +7,18 @@
 
 `hub clone {{repo_name}}`
 
-- Clone another user repository:
+- Clone another user's repository:
 
 `hub clone {{github_username}}/{{repo_name}}`
 
+- Fork your own copy of a repository cloned from another user (creates a remote named after your GitHub username):
```
agnivade: Might be good to mention that you need to be in the directory of the repo already cloned.

ericbn:
@agnivade, updated the examples with your suggestions from the previous commits.

waldyrious:
@ericbn apologies for having let this slip. The page looks understandable to me as it is now. I believe it can be further simplified, but I will propose changes later if I can come up with concise descriptions that don't sacrifice the current ones' clarity. LGTM 👍

agnivade:
Yea its a little bit verbose than usual. But it'll do the job IMO. LGTM.

PR #1047, opened 2016-09-09
add CLA to contributing.md

waldyrious:
(this also serves as a test for the clahub integration)


waldyrious:
Decided to go with cla-assistant for the following reasons (as mentioned in the project's gitter chat):

1. it not only updates the PR status but also comments when contributors have signed, giving extra visibility to the process
2. it maintains the text of the CLA in a [public gist](https://gist.github.com/waldyrious/e50feec13683e565769fbd58ce503d4e), so that could be a better way to handle eventual changes and improve transparency overall
3. it allows contributors to sign the CLA by simply logging in to their github account, while clahub requires the user to fill their info in a form

PR #1052, opened 2016-09-12
cpio: update descriptions

agnivade:
Closes #877

### DIFF
--- a/pages/common/cpio.md
+++ b/pages/common/cpio.md
@@ -1,16 +1,15 @@
 # cpio
 
-> Archiving utility.
+> Copies files to and from archives.
 
-- Create an archive from files
+- Take a list of file names from standard input and creates an archive from them:
 
-`echo "{{file1}} {{file2}} {{file3}}" | cpio -ov > {{archive.cpio}}`
+`echo "{{file1}} {{file2}} {{file3}}" | cpio -o > {{[o]utput_archive.cpio}}`
 
-- Create an archive from a directory
+- Create an archive containing all the files and folders in a directory in [v]erbose mode:
 
-`find {{directory}} | cpio -ov > {{archive.cpio}}`
+`find {{path/to/directory}} | cpio -ov > {{[o]utput_archive.cpio}}`
 
-- Extract an archive
+- Extract files from an archive, generating [d]irectories where needed, in [v]erbose mode:
 
-`cpio -idv < {{archive.cpio}}`
-
+`cpio -idv < {{[i]nput_archive.cpio}}`
--- a/pages/common/cpio.md
+++ b/pages/common/cpio.md
@@ -1,0 +1,16 @@
+# cpio
+
+> Archiving utility.
+
+- Create an archive from files
+
+`echo "{{file1}} {{file2}} {{file3}}" | cpio -ov > {{archive.cpio}}`
+
+- Create an archive from a directory
+
+`find {{directory}} | cpio -ov > {{archive.cpio}}`
+
+- Extract an archive
+
+`cpio -idv < {{archive.cpio}}`
+
--- a/pages/common/cpio.md
+++ b/pages/common/cpio.md
@@ -1,16 +1,16 @@
 # cpio
 
 > Copies files in and out of archives.
-> Supports following archive formats: binary, old ASCII, new ASCII, crc, HPUX binary, HPUX old ASCII, old tar, and POSIX.1 tar.
+> Supports the following archive formats: custom binary, old ASCII, new ASCII, crc, HPUX binary, HPUX old ASCII, old tar, and POSIX.1 tar.
 
-- Take a list of file names from standard input and copy them [o]ut to an archive in binary format:
+- Take a list of file names from standard input and add them [o]nto an archive in binary format:
 
 `echo "{{file1}} {{file2}} {{file3}}" | cpio -o > {{archive.cpio}}`
 
-- Copy [o]ut all files and folders in a directory to an archive, in [v]erbose mode:
+- Copy all files and folders in a directory and add them [o]nto an archive, in [v]erbose mode:
 
 `find {{path/to/directory}} | cpio -ov > {{archive.cpio}}`
 
-- Copy [i]n all files from an archive, generating [d]irectories where needed, in [v]erbose mode:
+- P[i]ck all files from an archive, generating [d]irectories where needed, in [v]erbose mode:
 
 `cpio -idv < {{archive.cpio}}`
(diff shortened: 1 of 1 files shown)


agnivade:
Tried hard to match the `i` and `o`. Used the filename from the example, otherwise I would have had to mention copy-out and copy-in in the description which would have been a bit weird.

pages/common/cpio.md
```diff
@@ -1,16 +1,15 @@
 # cpio
 
-> Archiving utility.
+> Copies files to and from archives.
 
-- Create an archive from files
+- Take a list of file names from standard input and creates an archive from them:
```
waldyrious: "creates" --> "create"

pages/common/cpio.md
```diff
@@ -1,16 +1,15 @@
 # cpio
 
-> Archiving utility.
+> Copies files to and from archives.
```
waldyrious: The description from https://www.gnu.org/software/cpio/ is quite readable and explains the acronym :)

> cpio copies files into or out of a cpio or tar archive

pages/common/cpio.md
```diff
@@ -1,16 +1,15 @@
...
 
-- Create an archive from files
+- Take a list of file names from standard input and creates an archive from them:
 
-`echo "{{file1}} {{file2}} {{file3}}" | cpio -ov > {{archive.cpio}}`
+`echo "{{file1}} {{file2}} {{file3}}" | cpio -o > {{[o]utput_archive.cpio}}`
```
waldyrious: I was slightly confused by the .cpio extension, as I had never heard of an archive format by that name. I understood it after reading the project page (emphasis mine):

> GNU **cpio supports the following archive formats: binary, old ASCII, new ASCII, crc, HPUX binary, HPUX old ASCII, old tar, and POSIX.1 tar.** The tar format is provided for compatability with the tar program. By default, cpio creates binary format archives, for compatibility with older cpio programs. **When extracting from archives, cpio automatically recognizes which kind of archive it is reading** and can read archives created on machines with a different byte-order.

I think we should incorporate that information into this description.

pages/common/cpio.md
```diff
@@ -1,16 +1,15 @@
...
 
-- Create an archive from a directory
+- Create an archive containing all the files and folders in a directory in [v]erbose mode:
```
waldyrious: I'd put a comma after "directory".

pages/common/cpio.md
waldyrious: Actually, I would venture saying that "create" is only accurate if the archive does not yet exist (in which case this tool is kind of redundant to other archiving tools). We should probably lead out with an example of copying files into an existing archive.

pages/common/cpio.md
```diff
@@ -1,16 +1,15 @@
...
 
-`find {{directory}} | cpio -ov > {{archive.cpio}}`
+`find {{path/to/directory}} | cpio -ov > {{[o]utput_archive.cpio}}`
 
-- Extract an archive
-
-`cpio -idv < {{archive.cpio}}`
+- Extract files from an archive, generating [d]irectories where needed, in [v]erbose mode:
```
waldyrious: I'd say "Extract all files (...)". Especially since we need a partial extraction example above this, for symmetry with the two previous examples, and to better demonstrate the core functionality of this tool (not full archiving/extraction, but partial).

pages/common/cpio.md
agnivade: Yes, like I mentioned before, I thought it was a bit weird to mention copy in and copy out. But since you are okay, I will change it.

waldyrious: I read that (and appreciated the effort), but didn't realize this wording fitted your perception of "weird". To me it sounds like perfectly reasonable (if a bit techy) English.

pages/common/cpio.md
```diff
@@ -0,0 +1,16 @@
+# cpio
+
+> Copies files in and out of archives.
+> Supports following archive formats: binary, old ASCII, new ASCII, crc, HPUX binary, HPUX old ASCII, old tar, and POSIX.1 tar.
+
+- Take a list of file names from standard input and copy them [o]ut to an archive in binary format:
```
waldyrious: "copy out to an archive" does sound awkward. "copy out of" would make sense but not "copy out to". Also, it seems strange that a redirection is needed even with the -o option, but hey, if that's how it works...

By the way, "binary format" is somewhat hand-wavy. We could perhaps mention in the main description that it is a *custom* binary format.

pages/common/cpio.md
```diff
@@ -0,0 +1,16 @@
+# cpio
+
+> Copies files in and out of archives.
+> Supports following archive formats: binary, old ASCII, new ASCII, crc, HPUX binary, HPUX old ASCII, old tar, and POSIX.1 tar.
```
waldyrious: "Supports following" --> "Supports the following".

pages/common/cpio.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Take a list of file names from standard input and copy them [o]ut to an archive in binary format:
+
+`echo "{{file1}} {{file2}} {{file3}}" | cpio -o > {{archive.cpio}}`
+
+- Copy [o]ut all files and folders in a directory to an archive, in [v]erbose mode:
```
waldyrious: Again, copy out to sounds awkward. I suppose we could just give up on describing the -o option, as much as it disappoints me.

pages/common/cpio.md
```diff
@@ -0,0 +1,16 @@
...
+
+`find {{path/to/directory}} | cpio -ov > {{archive.cpio}}`
+
+- Copy [i]n all files from an archive, generating [d]irectories where needed, in [v]erbose mode:
```
waldyrious: instead of "copy in files from an archive", the natural form is as it appears in the description: "copy files into an archive". Of course, these two sentences have the opposite meaning. Are you sure we're not reading the command options with their meanings reversed?

agnivade: Yep, its really like that. 

The man page says
```
cpio {-o|--create}

cpio {-i|--extract}
```

Do you want me to go back to the previous style of description again ?

waldyrious: What a perfectly missed opportunity they had there :disappointed: let's see how we can work around this. How about "add files onto an archive" for `-o` and "import files from an archive" for `-i`? Or maybe "p[i]ck files from an archive", since "import" tends to have a specific meaning that's not exactly matched here.

waldyrious:
Great :) we did it! LGTM

ostera:
LGTM!

PR #1053, opened 2016-09-12
column: add page

kidpixo:
Built following this man page : [column(1): columnate lists - Linux man page](http://linux.die.net/man/1/column)

Sadly the options don't have a long format, that's why is so hard for me to remember.

It is similar to [tldr/csvlook.md](https://github.com/tldr-pages/tldr/blob/aa60fa0f7ad3c9d5471b77329686a3e40e270106/pages/common/csvlook.md), but way quicker.

After some trouble, I installed  the linter/formatter locally via the [arch-npm/](https://hub.docker.com/r/nfnty/arch-npm/) docker image and aliased `tldrl` as
 `alias tldrl "docker exec npm tldrl"`, work like a charm :-)

Thanks for tldr!

### DIFF
--- a/pages/common/column.md
+++ b/pages/common/column.md
@@ -3,9 +3,9 @@
 > Format standard input or file into multiple columns.
 > Rows are filled before columns; default separator is whitespace.
 
-- Format output for a 10 columns wide display:
+- Format output for a 30 characters wide display:
 
-`printf "header1 header2\nbar foo\n" | column -c {{10}}`
+`printf "header1 header2\nbar foo\n" | column -c {{30}}`
 
 - Specify column delimiter character for the -t option (i.e. "," for csv); default is whitespace:
 
@@ -17,4 +17,4 @@
 
 - Fill columns before filling rows:
 
-`printf "header1\nbar\nfoobar\n" | column -c {{20}} -x`
+`printf "header1\nbar\nfoobar\n" | column -c {{30}} -x`
--- a/pages/common/column.md
+++ b/pages/common/column.md
@@ -1,0 +1,19 @@
+# column
+
+> Format standard or file input into multiple columns, rows are filled before columns.
+
+- Output is formatted for a display columns wide:
+
+`printf "header1 header2\nbar foo\n" | column -c 10`
+
+- Specify a set of characters to be used to delimit columns for the -t option, default is whitespace:
+
+`printf "header1,header2\nbar,foo\n" | column -s,`
+
+- Determine the number of columns the input contains and create a table:
+
+`printf "header1 header2\nbar foo\n" | column -t`
+
+- Fill columns before filling rows:
+
+`printf "header1\nbar\nfoobar\n" | column -c 20 -x`
--- a/pages/common/column.md
+++ b/pages/common/column.md
@@ -4,7 +4,7 @@
 
 - Output is formatted for a display columns wide:
 
-`printf "header1 header2\nbar foo\n" | column -c 10`
+`printf "header1 header2\nbar foo\n" | column -c {{10}}`
 
 - Specify a set of characters to be used to delimit columns for the -t option, default is whitespace:
 
@@ -16,4 +16,4 @@
 
 - Fill columns before filling rows:
 
-`printf "header1\nbar\nfoobar\n" | column -c 20 -x`
+`printf "header1\nbar\nfoobar\n" | column -c {{20}} -x`
--- a/pages/common/column.md
+++ b/pages/common/column.md
@@ -2,15 +2,15 @@
 
 > Format standard or file input into multiple columns, rows are filled before columns.
 
-- Output is formatted for a display columns wide:
+- Output is formatted for a 10 columns wide display:
 
 `printf "header1 header2\nbar foo\n" | column -c {{10}}`
 
-- Specify a set of characters to be used to delimit columns for the -t option, default is whitespace:
+- Specify characters to use to delimit columns for the -t option (i.e. "," for csv), default is whitespace:
 
-`printf "header1,header2\nbar,foo\n" | column -s,`
+`printf "header1,header2\nbar,foo\n" | column -s{{,}}`
 
-- Determine the number of columns the input contains and create a table:
+- Determine the number of columns and auto-align the output in a tabular format:
 
 `printf "header1 header2\nbar foo\n" | column -t`
 
--- a/pages/common/column.md
+++ b/pages/common/column.md
@@ -1,16 +1,17 @@
 # column
 
-> Format standard or file input into multiple columns, rows are filled before columns.
+> Format standard input or file into multiple columns.
+> Rows are filled before columns; default separator is whitespace.
 
-- Output is formatted for a 10 columns wide display:
+- Format output for a 10 columns wide display:
 
 `printf "header1 header2\nbar foo\n" | column -c {{10}}`
 
-- Specify characters to use to delimit columns for the -t option (i.e. "," for csv), default is whitespace:
+- Specify column delimiter character for the -t option (i.e. "," for csv); default is whitespace:
 
 `printf "header1,header2\nbar,foo\n" | column -s{{,}}`
 
-- Determine the number of columns and auto-align the output in a tabular format:
+- Split columns automatically and auto-align in a tabular format:
 
 `printf "header1 header2\nbar foo\n" | column -t`
 


ostera:
LGTM!

pages/common/column.md
```diff
@@ -0,0 +1,19 @@
+# column
+
+> Format standard or file input into multiple columns, rows are filled before columns.
+
+- Output is formatted for a display columns wide:
+
+`printf "header1 header2\nbar foo\n" | column -c 10`
```
agnivade: Please wrap 10 within {{ }} as its a token word.

kidpixo: done, thanks.

pages/common/column.md
```diff
@@ -0,0 +1,19 @@
+# column
+
+> Format standard or file input into multiple columns, rows are filled before columns.
+
+- Output is formatted for a display columns wide:
```
agnivade: If you are using a concrete value(10) in the example, the description should also use that. So its better to write - "Output is formatted for a 10 columns wide display"

pages/common/column.md
```diff
@@ -0,0 +1,19 @@
...
+
+`printf "header1 header2\nbar foo\n" | column -c {{10}}`
+
+- Specify a set of characters to be used to delimit columns for the -t option, default is whitespace:
+
+`printf "header1,header2\nbar,foo\n" | column -s,`
```
agnivade: `,` is also a token :)

pages/common/column.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Determine the number of columns the input contains and create a table:
```
agnivade: It seems this is copied from the man page verbatim. My suggestion - "Determine the  no. of columns and auto-align the output in a tabular format. "

pages/common/column.md
```diff
@@ -0,0 +1,19 @@
+# column
+
+> Format standard or file input into multiple columns, rows are filled before columns.
```
waldyrious: I'd use "Format standard input or file", or "Format file or standard input". Either is clearer and don't require the reader to parsed the whole expression before extracting its meaning.

waldyrious: Also, use a `;` instead of `,` to avoid a [run-on sentence](https://en.wikipedia.org/wiki/Run-on_sentence).

pages/common/column.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Output is formatted for a 10 columns wide display:
```
waldyrious: Instead of "Output is formatted", use "Format output" (we should probably mention the recommendation to use active rather than passive voide in the contributing guidelines)

pages/common/column.md
```diff
@@ -0,0 +1,19 @@
...
+
+`printf "header1 header2\nbar foo\n" | column -c {{10}}`
+
+- Specify characters to use to delimit columns for the -t option (i.e. "," for csv), default is whitespace:
```
waldyrious: Would it be clear enough to simply say "Specify column delimiter character for the -t option"? The parenthesized note can remain.

Also, replace the last comma with a semicolon, to avoid a run-on sentence.

pages/common/column.md
```diff
@@ -0,0 +1,19 @@
...
+
+`printf "header1,header2\nbar,foo\n" | column -s{{,}}`
+
+- Determine the number of columns and auto-align the output in a tabular format:
```
waldyrious: Would "Split columns automatically" be more accurate? I assume this either assumes tabs as separators, or looks up from an internal list of possible column delimiters. Is that the case?

kidpixo: "Split columns automatically" is fine, but the default separator is whitespace or the character passed with -s {{}}.

How does this sounds?

`- Split columns automatically; default separator is whitespace or passed by -s :`

waldyrious: But if you pass -s, does that still count as "automatically"? Also, was "and auto-align the output in a tabular format" redundant, considering that you removed it in your new proposal?

kidpixo: Well, the "automatically" refers to the "Determine the number of columns".
The other part was a proposal of @agnivade and I find if nice.
This is the original man page, how would you rewrite that?
```
-t
              Determine the number of columns the input contains and create
              a table.  Columns are delimited with whitespace, by default,
              or with the characters supplied using the --output-separator
              option.  Table output is useful for pretty-printing.
```

waldyrious: As I see it, letting the command detect columns without providing a separator is indeed automatic. By passing `-s`, the user explicitly tells the command where to split the lines, which dictates how many columns there are. I don't see how that can be described as automatic. So I'd refrain from mentioning `-s` here. I'd just say something like "Split columns automatically, using whitespace as column separators".

pages/common/column.md
waldyrious: On another note, I can't seem to get this option to produce any different behavior regardless of what number I use. Any hints?

pages/common/column.md
agnivade: I had done some testing with the command. We should also mentioned 'auto-align' because without `-t`, the output is just separated into columns. When used with `-t`, each column is left-aligned, like in a table. Hence my suggestion.

pages/common/column.md
agnivade: I know right ! I was wrestling with this option for quite some time. Couldn't get anywhere too :disappointed:

pages/common/column.md
waldyrious: Can you paste an example output with and without -t? I'm not sure I can see the column separation without the left alignment.

pages/common/column.md
kidpixo: try : 
```
>printf "header1 header2\nbar foo\n" | column
 header1 header2	bar foo
```
and  
```
>printf "header1 header2\nbar foo\n" | column -c 10
header1 header2
bar foo
```

in this case any number below 31 would not split the line (don't ask me why, it prints 24 characters in output...)

pages/common/column.md
kidpixo: I agree with @agnivade , just compare this two command: 
```
printf "header1 header2\nbar foo\n" | column -c 10
header1 header2
bar foo

>printf "header1 header2\nbar foo\n" | column -t
header1  header2
bar      foo
```
the second seems "automagically" aligned as table to me :-)

pages/common/column.md
waldyrious: Ok, so that's what I suspected. We should definitely not mix-up the two meanings of "column" here (the second being "character"). We're in luck since "characters" has the same starting letter, so we can describe the -c option with it easily :)

pages/common/column.md
waldyrious: @kidpixo ah that is very helpful indeed. So would it be fair to say that without the `-t`, the output is just wrapped when it hits the number of chars specified in `-c`? So if we add no options to the command, it just prints out whatever it receives, unaltered?

kidpixo: Sounds ok to me.

Should I add something somewhere specific?

waldyrious: This particular example seems fine to me given what we've discussed. You can address the other comments :)

agnivade:
LGTM.

waldyrious:
Please hold off merging :-1: there are still unaddressed comments.

pages/common/column.md
kidpixo: addressed in https://github.com/kidpixo/tldr/commit/207d2d38d7a8ccb55764918d5df4dd3f3eaac114

pages/common/column.md
kidpixo: this comment is ok then? check https://github.com/kidpixo/tldr/commit/207d2d38d7a8ccb55764918d5df4dd3f3eaac114 , please

waldyrious:
@kidpixo sorry if I wasn't clear. I meant that instead of "10 columns wide" we should say "10 characters wide" (probably using a more reasonable number, say 30 or more), to prevent confusion between the two meanings of "column" in the computing context.

kidpixo:
I agree on the `-c` argument numerical value.
For the other point, I understand what they mean in the original man page: normally the terminal width is referred as number of columns not characters 
I agree that the two meanings are misleading in the same context, I'll change it.

waldyrious:
Exactly :) LGTM now 👍

kidpixo:
Thanks, it was an interesting PR! 

Do you normally delete the branch at this point?

waldyrious:
Yes, you can delete the branch. It only exists as a real branch in your fork, actually (here it's just a raw git reference), so you're free to do do as you prefer :)

PR #1055, opened 2016-09-12
autojump: add new example

chuanconggao:


pages/common/autojump.md
```diff
@@ -2,6 +2,7 @@
 
 > Quickly jump among the directories you visit the most.
 > Aliases like j or jc are provided for even less typing.
+> TAB key can be used to list the top matches.
```
waldyrious: Our guidelines only admit up to two lines in the main description. I'd rather we don't open a precedent against that here. Could we maybe condense the second and third lines into a single one?

waldyrious:
Hmm. Not sure about this one. It's more of a shell example than an autojump example. I think I'd be fine if the additional stuff was serving an autojump example, but in this case it's the autojump example that's serving a separate command (ls, rm).

I'd like to hear the other maintainers' thoughts on this.

agnivade:
Agree. This is becoming more of a commandlinefu example. Ref to #1006.

agnivade:
@ostera - What are your thoughts on this ?

ostera:
I agree with you guys :)

waldyrious:
Ok then, I believe per the discussion above there's consensus that this change doesn't fit tldr-pages' ethos, so let's close. @chuanconggao, feel free to react anytime, we're open to discuss further if you disagree.

PR #1057, opened 2016-09-12
Add brew install command to README

danielcompton:
Relates to https://github.com/Homebrew/homebrew-core/pull/4637


waldyrious:
I'm fine with adding this and glad that this can help with Homebrew/homebrew-core#4637. I'd just ask you to mention that the brew command is OS X-specific, while the npm and web clients are platform-independent.

README.md
```diff
@@ -13,7 +13,7 @@
 
 A collection of simplified and community-driven man pages.
 
-Install it with `npm install -g tldr` or [try the web client](http://tldr.ostera.io).
+Install it with `npm install -g tldr`, `brew install tldr` (on OSX, using [Homebrew](http://brew.sh)), or [try the web client](http://tldr.ostera.io).
```
agnivade: There would be a slight disparity if we mention a platform specific method with a platform independent method. 

For eg. I would be fine if `brew install` is mentioned along with `apt-get install`, or if just `npm install` is mentioned. But mentioning `brew install` along with `npm install` slightly gives the indication that for OSX platforms, one should use `brew install`, which is not true at all.

README.md
```diff
@@ -47,7 +47,7 @@ You can access these pages on your computer using one of the following clients:
   - [tldroid](https://github.com/hidroh/tldroid), available on
     [Google Play](https://play.google.com/store/apps/details?id=io.github.hidroh.tldroid)
 - [Bash client](https://github.com/raylee/tldr)
-- [C++ client](https://github.com/tldr-pages/tldr-cpp-client): `brew install tldr-pages/tldr/tldr`
+- [C++ client](https://github.com/tldr-pages/tldr-cpp-client): `brew install tldr`
```
agnivade: Please also mention that for non-OSX clients, one has to compile from source.

README.md
danielcompton: How do you want this to read? I'm happy for you to modify this PR or make your own if that would be easier.

agnivade: I would refrain from mentioning this here, to keep the opening lines short and snappy. Mentioning it for the individual clients is fine.

waldyrious: Yeah, I could see this saying something like:

> Install it with `npm install -g tldr`, or [try the web client](http://tldr.ostera.io). On OS X, a client is also available using `brew install tldr`.

but as @agnivade said, it's a little awkward to mention one platform-specific installer and not mention others, like yum or apt-get. And if we do mention all those, we'd make this opening sentence too bulky, and downplay the other clients below, so yeah, I guess I'd also agree it's better just not to mention it here.

ericbn: My two cents: I would remove this line, as there are installation instructions in the Clients section already. If the Node client is the official one, this could be mentioned there too...

README.md
ericbn: Regardless of the change in line 16 above, this change could be merged already, as the `tldr` recipe is now officially being hosted in the homebrew repo...

waldyrious: @ericbn I've addressed this in e20cde3. Once this branch is rebased, we can focus on the change in the other discussion thread.

agnivade:
Hey @danielcompton - I am going ahead and closing this since according to the last discussion, we decided not to add a platform specific installation process for the top line.

PR #1063, opened 2016-09-15
rdesktop: add page

ramwin:
@waldyrious Thank you for your advice. I had reedited the rdesktop page. If there is any problem, don't hesitate to let me know. I'm pleased to fix it.


waldyrious:
I think this is a good start. We can tweak it later if needed. LGTM (Let's Get This Merged :P)

ostera:
LGTM

PR #1080, opened 2016-09-20
git-rebase: add --onto option

liuderchi:
`git rebase --onto <NEW_BASE> <OLD_BASE>` is a common usage if you want to keep history tidy.

### DIFF
--- a/pages/common/git-rebase.md
+++ b/pages/common/git-rebase.md
@@ -20,4 +20,4 @@
 
 - Rebase your local branch by specifying new base commit and old base commit:
 
-`git rebase --onto {{NEW-BASE-COMMIT}} {{OLD-BASE-COMMIT}}`
+`git rebase --onto {{new_base_commit}} {{old_base_commit}}`
--- a/pages/common/git-rebase.md
+++ b/pages/common/git-rebase.md
@@ -6,7 +6,7 @@
 
 `git rebase -i master`
 
-- Rebase your local branch  interactively with the latest changes from upstream:
+- Rebase your local branch interactively with the latest changes from upstream:
 
 `git fetch origin; git rebase -i origin/master`
 
@@ -17,3 +17,7 @@
 - Abort a rebase in-progress:
 
 `git rebase --abort`
+
+- Rebase your local branch by specifying new base commit and old base commit:
+
+`git rebase --onto {{NEW-BASE-COMMIT}} {{OLD-BASE-COMMIT}}`


pages/common/git-rebase.md
```diff
@@ -17,3 +17,7 @@
 - Abort a rebase in-progress:
 
 `git rebase --abort`
+
+- Rebase your local branch by specifying new base commit and old base commit:
+
+`git rebase --onto {{NEW-BASE-COMMIT}} {{OLD-BASE-COMMIT}}`
```
waldyrious: Please use lowercase and underscore in the tokens.

liuderchi: ok.
And found that `git-push.md` and `git-merge.md` not follow this style.

agnivade: Yes, we came up with a finalized style guide a bit later. You can fix those too in another PR if you would like to :smiley:

liuderchi: I'd love to :smiley: 
Maybe it's a whole-scan through `*.md` files.
And let me confirm that the token syntax guide is [here](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#token-syntax), right?

waldyrious: Exactly right @liuderchi. And a full-repo scan to fix tokens would be most welcome!

waldyrious:
One question, @liuderchi: would `git rebase --onto {{new_base_commit}}` (without specifying the `{{old_base_commit}}`) be the same as `git rebase {{new_base_commit}}`?

liuderchi:
> would git rebase --onto {{new_base_commit}} (without specifying the {{old_base_commit}}) be the same as git rebase {{new_base_commit}}?

@waldyrious  the ans is _no_ is most cases. (unless you config git with something)

TLDR：
- `--onto` option is used when you want _special base commit of current branch_, 
- otherwise use `git rebase [-i] master` is fine :+1: 


According `man git rebase` paragraph 2nd:
> If \<upstream\> is not specified, the upstream configured in
       branch.\<name\>.remote and branch.\<name\>.merge options will be used (see
       git-config(1) for details)

And in default it's not configured so this command fails and did nothing.

What's more, if you append _current branch_ (e.g. _myfeature_) as argument to the command
the result of the git history will be _only the current branch points to master_
and _commits in myfeature_ is lost.

waldyrious:
I may be reading you incorrectly, but "no in most cases" seems to contradict with the TLDR that you present afterwards. In any case, the change itself looks good to me. I'll revisit once I learn more about `rebase --onto`.

waldyrious:
LGTM

liuderchi:
@waldyrious sorry for confusing words I used.

* "no in most cases" means: 
    - without `{{old_base_commit}}` git would try to look up your git config, 
    - but in _default_ the target config entry is NOT set. In other words I just can't promise the ans is no in any cases.

* the previous TLDR section is for general scenarios, which is more straight forward.

agnivade:
:+1:

PR #1081, opened 2016-09-21
fixup: token string style

liuderchi:
- use underscore rather than minus
- use lower case rather than uppder case


waldyrious:
Great contribution! 👍 👍 👍 

LGTM :)

agnivade:
LGTM. Awesome work .. Keep 'em coming !

PR #1085, opened 2016-09-22
go: add page

agnivade:
Adding page for my current favorite language. :smile:

### DIFF
--- a/pages/common/go.md
+++ b/pages/common/go.md
@@ -6,7 +6,7 @@
 
 `go get {{package_path}}`
 
-- Compile and run a source file (it has to contain a "main" package):
+- Compile and run a source file (it has to contain a `main` package):
 
 `go run {{file}}.go`
 
@@ -14,7 +14,7 @@
 
 `go build`
 
-- Execute all test cases of the current package (files have to end with "_test.go"):
+- Execute all test cases of the current package (files have to end with `_test.go`):
 
 `go test`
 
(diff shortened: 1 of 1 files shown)


pages/common/go.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Compile and run a source file (it has to contain a "main" package):
+
+`go run {{file}}.go`
+
+- Compile the package present in the current directory:
+
+`go build`
+
+- Execute all test cases of the current package (files have to end with "_test.go"):
```
waldyrious: hm. github is trating the underscore as the start of italicized text, which then runs to the end of the document. How about using `` instead of ""?

waldyrious: I'm sorry, I meant
    `...`
instead of
    "..."

agnivade: I am thinking it might be a little inconsistent because I have main in "..." a few lines above, now this will be in `...`. Anyway, this is just an issue with github rendering, the client shows it fine. What do you think ?

waldyrious: I'd rather have "main" use `` too. In general, when presenting code samples, filenames, commands, etc., it's best to use ``. In this case it even has the nice side-effect of cancelling markdown syntax within it :)

waldyrious:
LGTM 👍

agnivade:
Yay .. lets go ! :shipit:

PR #1087, opened 2016-09-23
fold: add page

gamebusterz:
If anything else is not up to the mark, please let me know. Thank you for your time.

### DIFF
--- a/pages/common/fold.md
+++ b/pages/common/fold.md
@@ -1,0 +1,15 @@
+# fold
+
+> Wraps each line in an input file to fit a specified width.
+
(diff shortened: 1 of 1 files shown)


waldyrious:
LGTM.

pages/common/fold.md
```diff
@@ -0,0 +1,15 @@
+# fold
+
+> Wraps each line in an input file to fit a specified width.
```
agnivade: We should also mention that it outputs to the standard output. Initially I thought that it changes the file in-place.

agnivade:
LGTM

Issue #1004, opened 2016-08-15
add jq command


cultavix:

ostera:
Ah, this is a nice one. `jq`'s syntax can be fucking cryptic sometimes. I think we should proceed with care 😜

PR #1092, opened 2016-09-26
jq: add page

maximpertsov:
Basic usage for the jq json processing tool: https://stedolan.github.io/jq/

### DIFF
--- a/pages/common/jq.md
+++ b/pages/common/jq.md
@@ -6,7 +6,11 @@
 
 `cat {{file}} | jq`
 
-- Read JSON objects from file, into array, and output:
+- Output all elements from JSON array in file, or all key-value pairs from JSON objects in file:
+
+`cat {{file}} | jq .[]`
+
+- Read JSON objects from file, into array, and output (inverse of `jq .[]`):
 
 `cat {{file}} | jq --slurp`
 
--- a/pages/common/jq.md
+++ b/pages/common/jq.md
@@ -1,6 +1,6 @@
 # jq
 
-> A lightweight and flexible command-line JSON processor
+> A lightweight and flexible command-line JSON processor.
  
 - Output JSON file:
 
--- a/pages/common/jq.md
+++ b/pages/common/jq.md
@@ -1,11 +1,11 @@
 # jq
 
 > A lightweight and flexible command-line JSON processor.
- 
+
 - Output JSON file:
 
 `jq '' {{file}}`
- 
+
 - Read JSON stream into array and output:
 
 `jq -s '' {{file}}`
--- a/pages/common/jq.md
+++ b/pages/common/jq.md
@@ -4,20 +4,20 @@
 
 - Output JSON file:
 
-`jq '' {{file}}`
+`cat {{file}} | jq`
 
 - Read JSON objects from file, into array, and output:
 
-`jq '' -s {{file}}`
+`cat {{file}} | jq --slurp`
 
 - Output first element in JSON file:
 
-`jq '.[0]' {{file}}`
+`cat {{file}} | jq '.[0]'`
 
 - Output "key" of first element in JSON file:
 
-`jq '.[0].{{key}}' {{file}}`
+`cat {{file}} | jq '.[0].{{key}}'`
 
 - Output "key" of each element in JSON file:
 
-`jq 'map(.{{key}})' {{file}}`
+`cat {{file}} | jq 'map(.{{key}})'`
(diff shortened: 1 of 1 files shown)


agnivade:
closes #1004

agnivade:
ping @ostera for review since he was interested in this command. :)

pages/common/jq.md
```diff
@@ -0,0 +1,23 @@
...
+
+`jq '.[0]' {{file}}`
+
+- Output "key" of first element in JSON file:
+
+`jq '.[0].{{key}}' {{file}}`
+
+- Output "key" of each element in JSON file:
+
+`jq 'map(.{{key}})' {{file}}`
```
sbrl: This looks really useful. Perhaps adding an example that outputs multiple keys of a each object would be useful too? It's fine as is though.

pages/common/jq.md
```diff
@@ -0,0 +1,23 @@
+# jq
+
+> A lightweight and flexible command-line JSON processor.
+
+- Output JSON file:
+
+`jq '' {{file}}`
+
+- Read JSON stream into array and output:
```
waldyrious: Not sure I understand what a JSON stream is.

sbrl: Basically it's a text stream where every line has a json object on it.

pages/common/jq.md
```diff
@@ -0,0 +1,23 @@
+# jq
+
+> A lightweight and flexible command-line JSON processor.
+
+- Output JSON file:
+
+`jq '' {{file}}`
```
ostera: The `''` signifies an empty jq query. I'd say this one is rather uncommon. I find myself doing `cat {{ file }} | jq` _far more often_.

pages/common/jq.md
waldyrious: On a slightly tangential note, would it make a difference if the `''` came before the `-s`? If not, I think that'd be more intuitive.

maximpertsov: Nice @ostera, I like your description a lot more. Thanks for the historical context!

I also agree that `jq '' -s` is a bit more intuitive. Good call @waldyrious.

pages/common/jq.md
maximpertsov: I end up doing `cat <file> | jq` most of the time also. However, I wanted to call out that you need to add the empty query if you invoke jq directly, since this is a common error that users run into initially.

The empty string is also technically required if you pipe an empty jq query into another command (e.g. `cat <file> | jq | grep <text>` will not work, but `cat <file> | jq '' | grep <text>` will). Admittedly this usage is not so common either...

waldyrious: @maximpertsov it's laudable that you're aiming at preventing common rookie errors, but it would be nice to explain *why* the `''` is included, rather than just place it there and expect users to memorize the syntax as merely part of the tool's syntax.

Just mentioning that it represents an empty query would make this more understandable (it's just the what, not the why, but at least there aren't unexplained elements to the command).

Of course, if you can think of a better way to explain the `''` in concisely the example description, we're all ears :)

pages/common/jq.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Read JSON objects from file, into array, and output:
+
+`jq '' -s {{file}}`
```
waldyrious: @maximpertsov sorry to nitpick, but I would prefer if the expansion of `-s` ("stream", I suppose) was included in the command description, per the [guidelines](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines) (3rd bullet point). Ideally we should include the word and an explanation for it in parenthesis, or vice-versa.

pages/common/jq.md
```diff
@@ -0,0 +1,27 @@
+# jq
+
+> A lightweight and flexible command-line JSON processor.
+
+- Output JSON file:
+
+`cat {{file}} | jq`
```
agnivade: If the file already contains JSON text, what is this going to change ? Just pretty-print ? If yes. we should mention that.

waldyrious:
@maximpertsov I'll merge the PR, but please reply to @agnivade's comment above when you see this. Thanks for the contributon!

pages/common/jq.md
maximpertsov: That's right @agnivade, `jq` pretty-prints by default. You can add the flag `--compact-output` (or just `-c`) to print one json object per line instead. I'll add this note to the page.

Issue #978, opened 2016-07-25
add `fc-list` for  font list search


leoluyi:
`fc-list | grep {{font_name}} `

ostera:
Thanks for reporting! Feel free to [send a PR](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#submitting-a-pull-request) to propose this page — if in doubt, take a look at [CONTRIBUTING.md]
(https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) for some guidalines, and let us know if you need any help :)

mkoppmann:
I think we can close this issue.

waldyrious:
indeed @mkoppmann 👍

PR #811, opened 2016-02-29
fc-pattern: add page

sethwoodworth:
A useful tool in linux for looking up extended information on a font.

PR #1093, opened 2016-09-28
fc-list: add page

mkoppmann:
Fixes issue #978. I know that the format is not the right one, but I wanted to use the same style as in fc-match and fc-pattern.

### DIFF
--- a/pages/linux/fc-list.md
+++ b/pages/linux/fc-list.md
@@ -4,4 +4,4 @@
 
 - Return a list of installed fonts with given name:
 
-`fc-list | grep '{{Font Name}}'`
+`fc-list | grep '{{DejaVu Serif}}'`
--- a/pages/linux/fc-list.md
+++ b/pages/linux/fc-list.md
@@ -1,0 +1,7 @@
+# fc-list
+
+> List available fonts installed on the system.
+
+- Return a list of installed fonts with given name:
+
+`fc-list | grep '{{Font Name}}'`
--- a/pages/linux/fc-match.md
+++ b/pages/linux/fc-match.md
@@ -4,4 +4,4 @@
 
 - Return a sorted list of best matching fonts:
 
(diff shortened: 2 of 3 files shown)


waldyrious:
@mkoppmann thanks for the PR! the formatting is almost perfect, you just need to use {{snake_case}} for the token. Unless that is precisely what you meant by "the same style as in fc-match and fc-pattern"? Please confirm.

mkoppmann:
@waldyrious Yes, that’s what I meant. [Here](https://github.com/tldr-pages/tldr/pull/811) is the closed PR for fc-pattern and you said “Usually the tokens are provided in snake_case, but in this case I suppose it's better to illustrate what kind of strings one might pass to the command.”

waldyrious:
Thanks for the link, I didn't recall that :)

I wonder if we should provide a concrete example then, say "Times New Roman". That would make it clearer that the allowance was for the type of strings the command expects, rather than a possible overlook of the token syntax. What do you think

mkoppmann:
You mean instead of {{Font name}}? So `fc-list | grep 'DejaVu Serif'`. I chose DejaVu Serif because on Linux it’s more likely to be installed. Should I change the examples in the other fc-*.md also?

waldyrious:
Yeah, but keep the {{}} markers, since it's still a user-replaceable token -- it's just a specific rather than a generic one. And yes, if you could change the other fc-* pages, that'd be appreciated :)

waldyrious:
Could you can change the other pages in this PR too?

mkoppmann:
Hm? I changed fc-list, fc-match and fc-pattern. What other pages need to change? :)

waldyrious:
The browser cache must have played a trick on me. Sorry about the confusion 😄 LGTM.

agnivade:
:shipit:

PR #1095, opened 2016-09-28
Setting cache to false

agnivade:
- This was not letting new changes from tldr-lint repo to be fetched.
Therefore, preventing new checks like trailing spaces to appear.
- Also, its a good practice to get possible updates from npm packages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,7 @@ language: node_js
 node_js:
   - 'stable'
 
-cache:
-  directories:
-    - node_modules
+cache: false
 
 after_success:
 - bash scripts/build.sh


waldyrious:
👍

ostera:
I beg to disagree.

Naturally at this scale, and with only 4 dependencies, caching or not is _not that big a deal_, but letting `npm` update dependencies for us automatically could potentially break everything.

A better way would be to pin down exactly the version that we use of things, and doing [npm update](https://docs.npmjs.com/cli/update) periodically to update dependencies and see if things break or not with them. 

In this case, the bump from `v0.0.4` to `v0.0.5` in `tldr-lint` would've come with its corresponding PR to this repo to bump the `package.json`.

After all, new is not always better.

waldyrious:
IIUC, this only affects the testing environment, so even if updating stuff to the latest version breaks something, wouldn't we want to know early before that happens in a user's computer? Or am I completely misunderstanding the problem?

Otherwise, we can always use a service like http://greenkeeper.io (which I never used myself, but seems to work fine elsewhere)

ostera:
Well, if we pin down dependencies by exact version, we'll always run on a version that we know works, and anyone who `npm install`s will get exactly those version and not an accidentally upgraded one.

agnivade:
My thinking was that always bumping up versions whenever a change needed to be pushed would be very painstaking and not scalable. And since the no. of dependencies is so small, I would say that automatically updating dependencies should not be an issue.

Anyway, I had gone through the travis docs. Even if you bump up the package.json, it does not re-download the cache. Only after you flush the cache from its UI, does it trigger a re-download.

Let me know what do you guys think.

waldyrious:
Would Greenkeeper solve this problem?

ostera:
@agnivade, considering we handle only 1 dependency (tldr-lint), and that we rarely update it, it shouldn't be an issue.

My point is that I'd rather see something breaking and explicitly update a dependency to fix than have some dependency updated without me knowing what changed.

Greenkeeper could indeed work :)

agnivade:
Ok, so this is what I propose.

@ostera -  If I understand you correctly, your primary issue was with npm updating dependencies automatically. Not so much with setting cache to false (`tldr-lint` being a small repo). 

How about we set the dependencies for `glob`, `markdownlint-cli` and `husky` to its exact version, and set `tldr-lint` to a greater than or equal to mode ? Since I am planning to do some work on tldr-lint(already sent one [PR](https://github.com/tldr-pages/tldr-lint/pull/7) !), I will be fairly familiar with the codebase and have better control over its release cycles. 

So therefore, we will have strict controls over npm modules and at the same time, not worry about regularly purging cache or thinking whether to bump a version or not.

agnivade:
@ostera / @waldyrious - guys, what's the word on this ?

waldyrious:
Your proposal seems reasonable to me, @agnivade, but since my knowledge in nodejs development is basic at best, I'll defer to whatever you and @ostera decide.

ostera:
@agnivade, LGTM. Could you take the run-time of a few builds so we can compare the before/after?

agnivade:
Sure.

agnivade:
The runtimes seem to have increased by a margin of 10 secs on average.

After setting cache to false-
```
 PR #1043 date: add unix epoch timestamp 38 sec 
 PR #1105 vim: improve examples 47 sec 
 PR #1105 vim: improve examples 1 min 4 sec 
 PR #1001 edquota: add page 31 sec 
 PR #1103 Pin down the dependencies to their exact version 39 sec 
```

Before that - 
```
 PR #1088 git-reset: add page 35 sec 
 PR #1085 go: add page 32 sec 
 PR #1087 fold: add page 28 sec
 PR #1087 fold: add page 29 sec
```

ostera:
I can live with that :P

PR #1103, opened 2016-10-03
Pin down the dependencies to their exact version

agnivade:
- And keep tldr-lint on a greater than mode

As discussed in #1095

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,10 +3,10 @@
   "version": "1.0.0",
   "description": "Simplified, community-driven man pages",
   "dependencies": {
-    "glob": "=7.0.0",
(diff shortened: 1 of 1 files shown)


waldyrious:
LGTM 👍 per previous discussion in #1095.

zdroid:
Why write `=x.y.z` when it's [basically needless](https://docs.npmjs.com/files/package.json#dependencies)? `x.y.z` is enough to define you want an exact version. :wink:

Anyway, LGTM. :shipit:

agnivade:
That's true actually. I hadn't read up properly on semver. Will send an update.

agnivade:
LGTM

PR #1105, opened 2016-10-05
vim: improve examples

waldyrious:
I tried to condense the most useful basic commands into this page without surpassing our length limit.
I also expanded the main description -- I wonder if there's a way to phrase it which could allow us to get rid of all the <Esc>s...

I combined copy/cut/paste into a single command, since it was important to allow space for a slightly more advanced command, (the find/replace example), which introduces some important vim concepts, such as command application ranges.

PR #1106, opened 2016-10-10
vim: Normal mode is the normal mode

ericbn:
Once in normal mode, which is the mode user should stay most of the time in vim, one does not need to press `<Esc>` before each command. Insert mode is a temporary mode.
Make this more explicit in the examples (using the ...then... format also present in `less`, `more`, etc.) Also use ...then... for the search example.
Remove the yand and delete example, due to the complexity and the fact that there's `P` to paste before and `p` to paste after the cursor, and in favor of the quit without saving example.

### DIFF
--- a/pages/common/vim.md
+++ b/pages/common/vim.md
@@ -1,7 +1,7 @@
 # vim
 
 > Vi IMproved, a programmer's text editor, providing several modes for different kinds of text manipulation.
-> Pressing `i` enters edit mode; the normal mode (accessed via `<Esc>`) doesn't allow regular text editing.
+> Pressing `i` enters edit mode. `<Esc>` goes back to normal mode, which doesn't allow regular text insertion.
 
 - Open a file:
 
@@ -9,24 +9,28 @@
 
 - Enter text editing mode (insert mode):
 
-`<Esc> i`
+`<Esc>i`
 
 - Copy ("yank") or cut ("delete") the current line (paste it with `P`):
 
-`<Esc> {{yy|dd}}`
+`<Esc>{{yy|dd}}`
 
 - Undo the last operation:
 
-`<Esc> u`
+`<Esc>u`
 
-- Search for a pattern in the file (press `n` to go to the next result):
+- Search for a pattern in the file:
 
-`<Esc> /{{search_pattern}} <Enter>`
+`<Esc>/{{search_pattern}}<Enter>   then   n (next), N (previous)`
 
-- Perform a regex substitution in the whole file (from the start, `1`, to the end, `$`):
+- Perform a regex substitution in the whole file:
 
-`<Esc> :1,$s/{{pattern}}/{{replacement}}/g <Enter>`
+`<Esc>:%s/{{pattern}}/{{replacement}}/g<Enter>`
 
-- Save (write) the file, and quit vim:
+- Save (write) the file, and quit:
 
-`<Esc> :wq <Enter>`
+`<Esc>:wq<Enter>`
+
+- Quit without saving:
+
+`<Esc>:q!<Enter>`
--- a/pages/common/vim.md
+++ b/pages/common/vim.md
@@ -19,13 +19,13 @@
 
 `<Esc>u`
 
-- Search for a pattern in the file:
+- Search for a pattern in the file (press `n`/`N` to go to next/previous match):
 
-`<Esc>/{{search_pattern}}<Enter>   then   n (next), N (previous)`
+`<Esc>/{{search_pattern}}<Enter>`
 
-- Perform a regex substitution in the whole file:
+- Perform a regex substitution in the whole file (from the start, `1`, to the end, `$`):
 
-`<Esc>:%s/{{pattern}}/{{replacement}}/g<Enter>`
+`<Esc>:1,$s/{{pattern}}/{{replacement}}/g<Enter>`
 
 - Save (write) the file, and quit:
 


waldyrious:
`yy`/`dd`+`p` are probably the commands I use the most in Vim. I don't know about others, but I'd much rather have these instead of the quit without saving (especially since `u` is included to undo any changes).

ericbn:
I use `dd` (and `yy`) as well as variations like `dw`, `dW`, `v{motion}d`... Plus there's `p` and `P` to put after and before the cursor. I think these are too complex for a TL;DR. Maybe we could recommend users to check `vimtutor` instead...

waldyrious:
In my opinion simple replacements for Ctrl+C, Ctrl+X and Ctrl+V (`yy`, `dd` and `P`, respectively) are both absolutely essential for text editing, and can be conveyed in a simple enough manner, which is what I tried with the example included in the current version of this page.

Let's wait and see what others have to say :)

ostera:
Hej @ericbn, totally get what you mean. I'm a full-time vim user and yes, this looks wrong to me. **BUT** that's because I understand what a modal editor is, and in particular how vim works.

The average joe accidentally falling into vim has _no clue_, and needs a bit more guidance – and I can't emphasize this enough. That extra `<Esc>` is there so that they can just do as the `tldr` says, and get the result they expect.

**Re: the quitting example**, if they don't have permissions there's a good chance they won't know how to exit without saving to reopen with `sudo` (and we're not going to put the `r ! sudo tee` option here).

**Re: the copy/paste**, most people will get around just fine editing things slowly and manually and copying/pasting with their mouse.

Ultimately anyone who wants to learn vim will find far better resources than this. This is not a guide, this is not a getting started with, this is just a tldr.

agnivade:
Agree with @waldyrious here. I would call myself a basic level vim user, and `yy/dd/p` are the most common stuff that I use.

ericbn:
I get that we should not try to explain here how a modal editor works.

There are so many "basic" commands in vim, like also `.` (repeat last command) or `x` (delete char under cursor), and the list goes on..., and so many ways to yank and delete, one line or a word or from the cursor to the end of the line, or multiple lines, and the list goes on again..., that I still think that for the average Joe, he should better know how to quit without saving.

waldyrious:
There may be lots of ways to do copy/cut/paste operations, but we must present at least one of them, since they're fundamental operations in text editing. We don't have to stick with yy/dd/P, but unless we can provide an alternative (single) example that shows how to do these thee operations, I think it's a net negative to remove the current one.

ericbn:
I updated my commit with improvements we may all agree upon:
- Remove spaces between keypresses (as it could lead users to think they should type a space there).
- Simplify regexp substitution example.
- Add ...then... to search example.
- Add quit without saving example.

pages/common/vim.md
```diff
@@ -1,32 +1,36 @@
...
+- Search for a pattern in the file:
 
-`<Esc> /{{search_pattern}} <Enter>`
+`<Esc>/{{search_pattern}}<Enter>   then   n (next), N (previous)`
 
-- Perform a regex substitution in the whole file (from the start, `1`, to the end, `$`):
+- Perform a regex substitution in the whole file:
 
-`<Esc> :1,$s/{{pattern}}/{{replacement}}/g <Enter>`
+`<Esc>:%s/{{pattern}}/{{replacement}}/g<Enter>`
```
agnivade: Nice catch ! I always use `%s`, wonder how I missed it.

agnivade:
Nicely done ! This has my :+1:

pages/common/vim.md
waldyrious: Not sure this was a "catch" 😄 I deliberately used the non-shorthand notation, in order to introduce the concept of ranges preceding commands. I mentioned this decision explicitly in #1105:

> I combined copy/cut/paste into a single command, since it was important to allow space for a slightly more advanced command, (the find/replace example), which **introduces some important vim concepts, such as command application ranges.**

Although `%` is shorter, it is so by mere two characters, and obscures this meaning and instead introduces a single-purpose symbol that doesn't allow this example to be built upon, which is one of the [goals of TLDR-pages](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines):

> The goal is to allow people to *understand* the syntax of a command, not just *memorize* it.

So IMO we should both retain the `1,$` range, as well as the parenthetical note in the command's description.

pages/common/vim.md
```diff
@@ -1,32 +1,36 @@
...
 - Undo the last operation:
 
-`<Esc> u`
+`<Esc>u`
 
-- Search for a pattern in the file (press `n` to go to the next result):
+- Search for a pattern in the file:
 
-`<Esc> /{{search_pattern}} <Enter>`
+`<Esc>/{{search_pattern}}<Enter>   then   n (next), N (previous)`
```
waldyrious: I don't think mixing stuff to be typed and instructions together is a good idea, even if there's precedent (I'd rather we change it where it occurs). Using `N` is a nice addition, though -- how about we change the description to "(use `n` and `Shift+n` to navigate the results)"?

pages/common/vim.md
```diff
@@ -1,32 +1,36 @@
 # vim
 
 > Vi IMproved, a programmer's text editor, providing several modes for different kinds of text manipulation.
-> Pressing `i` enters edit mode; the normal mode (accessed via `<Esc>`) doesn't allow regular text editing.
+> Pressing `i` enters edit mode. `<Esc>` goes back to normal mode, which doesn't allow regular text insertion.
```
waldyrious: I like this change, it highlights that insert mode is meant to be temporary 👍

waldyrious:
I've added some inline comments. About the removal of spaces, I understand the motivation but I'm not sure it is worth the decreased readability (which is why they were included), especially for a command whose syntax is particularly terse. This one is a subjective matter, so although I'd prefer keeping the spaces, I will defer to the majority opinion in this thread.

pages/common/vim.md
ericbn: Okay, only `less` and `more` are also using this kind of example. We can fix them too.  I though it looked nice. 😄

pages/common/vim.md
agnivade: Right, this isn't technically a catch :) I meant from my point of view, I should have suggested this change. Your explanation already answered my question anyway.

I prefer the `%` because its shorter and you generally perform substitutions in the whole file. Atleast in my case. 

You are right, in your reasoning however, that `1,$` introduces the concept of ranges and allows this example to be built on. IMO, ranges are not required in a tldr page, and `%` does just fine. But then again, this is going into subjective territory. So my final vote is neutral on this.

agnivade:
> About the removal of spaces, I understand the motivation but I'm not sure it is worth the decreased readability 

I don't see any particular decrease in readability. On the contrary, I agree with @ericbn's thinking that first timers might think that a space is needed between the commands.

To conclude, I feel a lot of users feel strongly about `vim` since its so popular and people have developed so many ways to do one thing, therefore such a myriad of opinions. I have no strong objections to any of the changes which @waldyrious has suggested in the latest PR review. Whatever you guys come to agree on is fine with me.

pages/common/vim.md
ericbn: IMHO typed in ranges are too complex and rarely (if at all manually) used. The visual line mode, for example, is usually better for this kind of task... As for the purpose of explaining, we could describe it as "Perform a regex substitution of all occurrences ("g") in all lines ("%") of the file".

ericbn:
@agnivade, I completely agree that vim is so complex that it gives rise to many discussions for a TL;DR page. A TL;DR for vim, for many of its lovers, would be: "If you think it's too long and you don't want to read, then don't use it!" 😄 

It's the same for emacs. In fact, I'd like to know how can we do cut/copy/paste, undo, search, search/replace and save in emacs too... 😄

pages/common/vim.md
ericbn: There are already a lot of extra characters you have to type in to get this command. Two characters less is a good gain, and that's why people type `%` instead of `1,$` (even for `:%sort` or anything else one wants to apply to the whole file)...

I still think we're just showing a more complex and less common syntax there. And it's not as clear as to lack an explanation.

waldyrious: I understand your point, and agree to some extent; but we should keep in mind that tldr-pages is a didactic resource, not a cheatsheet of directly usable examples (that's more like [commandlinefu](http://www.commandlinefu.com/) territory). I don't think I can explain my motivations any better than I already have, so I'll just add that while I won't veto this change, I really, really would like us to keep `1,$` (and its corresponding note in the command description) for its didactic value.

ericbn:
Just added a new commit considering the observations on the last one.

waldyrious:
Thanks for the proposal and for bearing with us (me especially) 😄 LGTM! @ostera maybe you'll want to merge this one, so you'll have the chance for a final say?

PR #1107, opened 2016-10-11
less, more: Don't use ...then... in examples

ericbn:
Instead, explain additional keystrokes in the example title. See https://github.com/tldr-pages/tldr/pull/1106#discussion_r82836295

Also, change order of page up / down in `less` example to match title and to match the order of the other examples.

### DIFF
--- a/pages/common/less.md
+++ b/pages/common/less.md
@@ -8,21 +8,21 @@
 
 `less {{source_file}}`
 
-- Page up / down:
+- Page down / up:
 
-`<Space> (next), b (previous)`
+`<Space> (down), b (up)`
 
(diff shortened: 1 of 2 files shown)


waldyrious:
The change looks good overall, but I'm not sure the "up, down" order is the right one to present. Usually, we will have to page down before we even can page back up, so the natural sequence IMO would be "next" first, and then "previous".

ericbn:
Makes perfect sense. Just updated the commit.

waldyrious:
Thanks for addressing this as a follow-up to #1106 :) LGTM.

agnivade:
:+1:

PR #1113, opened 2016-10-13
edquota: minor tweaks to commands and descriptions

waldyrious:


pages/linux/edquota.md
```diff
@@ -1,21 +1,21 @@
...
-`edquota {{user}}`
+`edquota --user $(whoami)`
 
 - Edit quota of specific user:
 
-`sudo edquota {{user}}`
+`sudo edquota --user {{username}}`
 
-- Edit the group quota:
+- Edit quota a group:
```
pearcem0: should this be 'Edit quota for a group'?

waldyrious: Good catch! Will change.

waldyrious:
Thanks for the review @pearcem0 :) since you're the original author of the page, I think it's OK to merge without the two maintainers' thumbs-up rule.

PR #1116, opened 2016-10-14
add new guideline in CONTRIBUTING.md

yasharmaster:
Commands requiring 'sudo' should be mentioned directly in examples. It would be helpful to include it in the guidelines.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,13 +35,14 @@ Here are a few guidelines to get started:
    It's OK if the page doesn't cover everything; that's what `man` is for.
 2. When in doubt, keep new command-line users in mind.
    Err on the side of clarity rather than terseness.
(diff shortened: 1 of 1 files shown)


waldyrious:
That's a great point, @yasharmaster, but I don't think this specific rule fits well with the other examples which are more generic. How about appending it to guideline no. 2 as an example? i.e.

> 2. When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness. For example, commands that require `sudo` should include it directly in the examples.

(note also that I changed the '' to `` around `sudo`)

yasharmaster:
@waldyrious Changed it.

waldyrious:
Thanks @yasharmaster :) LGTM.

ostera:
LGTM!

PR #1119, opened 2016-10-16
quotacheck: minor tweaks

waldyrious:


waldyrious:
Thanks again @pearcem0 😄

PR #1123, opened 2016-10-20
msmtp: add page

josa42:


pages/common/msmtp.md
```diff
@@ -0,0 +1,16 @@
+# msmtp
+
+> An SMTP client.
+> In the default sendmail mode, msmtp reads a mail from standard input and sendscit to an SMTP server for delivery.
```
waldyrious: I feel this sentence should be shortened. Maybe cap it to 80 chars or so. Also, there's a typo: "sendscit" --> "sends it".

pages/common/msmtp.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Send a mail using the default account configured in `~/.msmtprc`:
```
waldyrious: "send a mail" sounds a little awkward, but maybe it's my ear. I'd have used "Send an email".

pages/common/msmtp.md
```diff
@@ -0,0 +1,16 @@
...
+
+`echo {{"Hello world"}} | msmtp {{to@example.org}}`
+
+- Send a mail using a specific account configured in `~/.msmtprc`:
+
+`echo {{"Hello world"}} | msmtp --account={{example}} {{to@example.org}}`
```
waldyrious: What kind of content goes in the account argument? An arbitrary string? An email address? Maybe the token should be formatted and/or renamed to clarify this.

pages/common/msmtp.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Send a mail without a configured account:
+
+`echo {{"Hello world"}} | msmtp --host={{localhost}} --port={{999}} --from={{from@example.org}} {{to@example.org}}`
```
waldyrious: How does this associate the email "from@example.org" with the example.org mail servers? I'm assuming the recipient would see something different in the "From" field when they use, say, gmail to read the message than if the actual mail servers at example.com had been used to send the message, no?

josa42: I am not entirely with this but as far as I know the "sent on behalf of" is shown when an optional head `Sender:` is set. So if i would send an email on someone else's behalf I would put my email address in `Sender:`.

But none of this has anything to do with this. The `--from` argument sets the content of the `From:` header and `--host` specifies the host the smtp server is running, nothing special about it.

waldyrious: Maybe I'm misunderstanding how this works. If you did `msmtp --host=smtp.gmail.com --from=josa42@gmail.com waldyrious@gmail.com`, would this command prompt you for the password of your gmail account? Otherwise, the email surely would arrive at the destination with different parameters (SMTP headers I suppose?).

agnivade: You are partially right. One can specify the authentication mode in the config file - http://msmtp.sourceforge.net/doc/msmtp.html#A-user-configuration-file. I think we should add some authentication options in the command line too, just to make it a complete example. What do you think ?

waldyrious: As long as it would make things clearer/more explicit, I fully agree.

waldyrious:
Note: there's an ongoing discussion thread which is hidden by github's interface because it started in a patch that has since been superseded. It can be seen here: https://github.com/tldr-pages/tldr/pull/1123#discussion_r84397854.

agnivade:
@josa42 - any updates on this ?

agnivade:
Closing this in favor of #1272

PR #1124, opened 2016-10-20
ssh-copy-id: move to common

josa42:


pages/common/ssh-copy-id.md
```diff
@@ -0,0 +1,11 @@
+# ssh-copy-id
+
+> Install your public key in a remote machine's authorized_keys.
+
+- Copy your key to a remot machine:
```
waldyrious: What's the default if there are several local keys available? Also, typo: "remot" --> "remote".

pages/common/ssh-copy-id.md
```diff
@@ -0,0 +1,11 @@
...
+
+`ssh-copy-id {{user@remote}}`
+
+- Copy a specific key to a remot machine:
+
+`ssh-copy-id -i {{~/.ssh/id_example.pub}} {{user@remote}}`
```
waldyrious: Do you think it'd be worth changing the token to "remote_host", for consistency with [ssh.md](https://github.com/tldr-pages/tldr/blob/master/pages/common/ssh.md), [scp.md](https://github.com/tldr-pages/tldr/blob/master/pages/common/scp.md), and probably others?

waldyrious:
Thanks for the PR! Please check the inline comments.

agnivade:
Nice PR! Very useful command. I keep telling people to stop copying their public key and manually adding to `authorized_keys`.

pages/common/ssh-copy-id.md
```diff
@@ -2,6 +2,14 @@
 
 > Install your public key in a remote machine's authorized_keys.
 
+- Copy your keys to the remote machine:
+
+`ssh-copy-id {{username@remote_host}}`
+
+- Copy your default key to the remote machine:
```
agnivade: This is not useful IMO. Copying your default key with ssh would require access to the machine in the first place. Which means you have to give password auth or some other key.

agnivade:
@josa42 - This was a long time pending. So just changed it myself. Let me know if you have any other thoughts.

PR #1125, opened 2016-10-21
http: Variable supplied parameters between {{}} missing in the last example.

ericbn:


waldyrious:
Thanks! This is simple and a straightforward fix, I'll merge right away.

PR #1126, opened 2016-10-23
pdftk: add page

frinkelpi:


pages/common/pdftk.md
```diff
@@ -0,0 +1,23 @@
+# pdftk
+
+> PDF toolkit.
+
+- Extract range:
```
waldyrious: This should be a little more descriptive, e.g. "Extract pages from a pdf file and save them as another file:". ("range" should probably be avoided since the syntax shows that the pages need not be a single continuous interval.)

pages/common/pdftk.md
```diff
@@ -0,0 +1,23 @@
...
+
+`pdftk {{input.pdf}} cat {{1-3 5 6-10}} output {{output.pdf}}`
+
+- Merge PDF:
```
waldyrious: Again, more verbosity would be preferrable; somethig like "Merge two pdf files and save the result as a third one:" would be good.

pages/common/pdftk.md
```diff
@@ -0,0 +1,23 @@
...
+
+`pdftk {{file1.pdf}} {{file2.pdf}} ... cat output {{output.pdf}}`
+
+- Separate PDF into individual pages:
```
waldyrious: Are these pages saved as separate pdf files, one per page? And what filename pattern do they use? It would be good for people to be able to predict the outcome of running a command without having to do experiments.

pages/common/pdftk.md
```diff
@@ -0,0 +1,23 @@
...
+
+`pdftk {{input.pdf}} burst`
+
+- Rotate third page by 90 degrees clockwise:
```
waldyrious: You should also mention that this is picking a set of pages. Or even better: make this an example of copying all pages from a source pdf, but rotating the third one.

waldyrious:
Nice one :) please take a look at the inline comments.

pages/common/pdftk.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Extract pages from a PDF file and save them as another one:
+
+`pdftk {{input.pdf}} cat {{1-3 5 6-10}} output {{output.pdf}}`
+
+- Merge (concatenate) a list of PDF files and save the result as another one:
+
+`pdftk {{file1.pdf}} {{file2.pdf}} ... cat output {{output.pdf}}`
+
+- Separate a PDF file into individual ones, with a given output pattern:
```
waldyrious: I liked that the previous version of this description mentioned that the splitting was done per page. How about something like "Split each page of a PDF file into a separate file, with a given output filename pattern"?

agnivade:
LGTM.

waldyrious:
@frinkelpi could you address [this comment](https://github.com/tldr-pages/tldr/pull/1126#discussion_r84594184), please?

pages/common/pdftk.md
frinkelpi: @waldyrious It would not work with a space. As far as I've understood, detailed syntax usage is not given in the tldr pages and the user should refer to the man page for more information, right ?

pages/common/pdftk.md
```diff
@@ -0,0 +1,23 @@
...
+
+> PDF toolkit.
+
+- Extract range:
+
+`pdftk {{input.pdf}} cat {{1-3 5 6-10}} output {{output.pdf}}`
+
+- Merge PDF:
+
+`pdftk {{file1.pdf}} {{file2.pdf}} ... cat output {{output.pdf}}`
```
frinkelpi: @ostera I have done it as in https://github.com/tldr-pages/tldr/blob/master/pages/common/convert.md

pages/common/pdftk.md
waldyrious: Sure, I was just suggesting that if a space would work, it would be better to use such a syntax, as that would IMO make the command more understandable.

Since that's not an option, I'm thinking maybe swapping this command and the last one would allow one to decipher the intended syntax more easily. This would also adhere more closely to the [project guidelines](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines):
> Introduce options gradually, starting with the simplest command invocations, and using more complex examples progressively.

frinkelpi: Good idea, done.

waldyrious:
Nice work, @frinkelpi :) LGTM.

PR #1129, opened 2016-10-25
git-diff: rework a bit for clarity

waldyrious:
- use {{commit}} instead of {{PATHSPEC}}
- start with the simplest invocation, git diff
- move command to show staged changes next to the first two, which it is related to
- clarify description of --summary example

### DIFF
--- a/pages/common/git-diff.md
+++ b/pages/common/git-diff.md
@@ -2,21 +2,25 @@
 
 > Show changes to tracked files.
 
-- Show changes to tracked files:
+- Show unstaged, uncommitted changes:
 
-`git diff {{PATHSPEC}}`
+`git diff`
 
-- Show only names of changed files:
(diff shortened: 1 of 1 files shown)


ostera:
LGTM

pages/common/git-diff.md
```diff
@@ -2,21 +2,25 @@
...
+- Show only names of changed files since a given commit:
 
-`git diff --staged`
+`git diff --name-only {{commit}}`
+
+- Output a summary of file creations, renames and mode changes since a given commit:
+
+`git diff --summary {{commit}}`
 
 - Create a patch file:
```
agnivade: Should we put the ".patch" out of the token ? IIRC, reviewboard/gerrit accepts only ".patch" files. Or I might be wrong.

agnivade:
:+1:

PR #1130, opened 2016-10-27
source: add page

Fykec:


pages/common/source.md
```diff
@@ -0,0 +1,11 @@
+# Source
+
+> The source command can be used to load any shell file into the current shell script or a command prompt.
```
agnivade: "load any shell file into the current shell script" ? Not sure if that makes sense to me.

ostera:
I have my reservations about this one, as it isn't exactly a command -- it's a _built-in_. And it varies from shell to shell rather than from os to os. Should we include shell specific built-ins in general?

All in all, `source` doesn't really load things, it reads and executes.

pages/common/source.md
waldyrious: Maybe "load any shell script into the current shell session"?

waldyrious:
@ostera I'm inclined to merge -- I'd rather have slightly incorrect information (i.e. not matching the current user's shell exactly) than no information at all. What do you say? @agnivade?

agnivade:
I have to agree with @ostera on this. My rule of thumb is if it has a man page, it can have a tldr page. `source` does not have a man page in Ubuntu. So, its a negative from me. I am open to counter points though.

waldyrious:
> My rule of thumb is if it has a man page, it can have a tldr page. 

@agnivade that's interesting, I'd actually use the rule of thumb that if it is an actual command line tool, it should have a tldr-pages entry, even if it doesn't have a man page (which is the case for many of the entries we do have currently). Please take a look at #1200 which poses a related question.

agnivade:
@waldyrious - so what's the take on this ?

waldyrious:
If my previous comment didn't sway neither you nor @ostera, I'll defer to the majority opinion.

agnivade:
Closing this as there was no further counter points against it.

PR #1132, opened 2016-10-27
yum: added "provides" as it is incredibly useful

andrea-82:
"yum provides" is incredibly useful when you come from a different distro, and you don't know exactly what package provides a particular command.


ostera:
LGTM

waldyrious:
LGTM 👍

PR #1134, opened 2016-10-28
vue: add page

gldraphael:


waldyrious:
Thanks for the PR @gldraphael, looks great! I've left only a minor comment inline, please take a look.

agnivade:
Nice, vue seems to be really picking up pace ! LGTM from me after fixing @waldyrious' comment.

gldraphael:
@waldyrious done!
@agnivade yes it is :D

waldyrious:
@gldraphael please remove the `/` as well, if you don't mind 😄

gldraphael:
@waldyrious so should it be `{{path/to/template}}` or `{{path/to/template_folder}}`?

ostera:
LGTM -- ha! the first `tldr.jsx` was written with vue :P

waldyrious:
If it's a folder rather than a file, then `{{path/to/template_folder}}` would be preferred :)

waldyrious:
LGTM!

PR #1135, opened 2016-10-28
alias: various tweaks

waldyrious:
I got caught off-guard about an alias I set not being permanently saved, so I added a caveat to the main description.

While I was at it, I also tweaked the descriptions of some examples.

### DIFF
--- a/pages/common/alias.md
+++ b/pages/common/alias.md
@@ -19,6 +19,6 @@
 
 `alias {{rm}}="{{rm -i}}"`
 
-- Create a `la` shortcut for `ls -a`:
+- Create `la` as a shortcut for `ls -a`:
 
 `alias {{la}}="{{ls -a}}"`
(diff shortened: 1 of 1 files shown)


pages/common/alias.md
```diff
@@ -10,14 +11,14 @@
...
 
 `alias -p`
 
-- Turn rm an interative command:
+- Turn rm into an interactive command:
 
 `alias {{rm}}="{{rm -i}}"`
 
-- Override la as ls -a:
+- Create a `la` shortcut for `ls -a`:
```
agnivade: Maybe we can reword this to "Create `la` as a shortcut for `ls -a`" ?

agnivade:
:+1:

ostera:
This one is also a built-in, such as `source`. It'd vary from shell to shell rather than from os to os.

Other than that, being pragmatic, LGTM.

waldyrious:
If there are actual significant differences, I agree we should probably remove it, or at least stick to behavior that is common. But that's a separate discussion :)

PR #1138, opened 2016-10-31
Add script to post Travid build errors on PRs

### LINKED ISSUE #1066: Post and update a Checklist with the errors to be fixed on each PR
As per https://github.com/tldr-pages/tldr/issues/980

While most people are happy with checking travis logs, it's a step less for us to just look for a message like the one below when a Travis build has failed, and make sure it's been all checked.

> Hi! We found some errors in your pages. Please attend to them soon so we can take a look at your page!
> - [ ] pages/linux/strace.md:21: TLDR005 Example descriptions should end in a colon
> - [ ] pages/linux/strace.md:21: TLDR006 Command name and description should be separated by an empty line

agnivade:
Fixes #1066

### DIFF
--- a/scripts/post_pr_comment.py
+++ b/scripts/post_pr_comment.py
@@ -1,0 +1,36 @@
+import json
+import os
+import sys
+import urllib2
+
+GITHUB_URL = 'https://api.github.com'
+
+def post_comment(pr_id, repo_slug, comment_body, user_token):
+	url = '{api_url}/repos/{slug}/issues/{number}/comments'.format(
+        api_url=GITHUB_URL, slug=repo_slug, number=pr_id)
+	req = urllib2.Request(url=url,
+                       data=json.dumps({'body': comment_body}))
+	req.add_header('Authorization', 'token ' + user_token)
+	f = urllib2.urlopen(req)
+	if f.getcode() != 201:
+		print f.read()
+
+
+if __name__ == '__main__':
+	PR_NUMBER = os.environ.get('TRAVIS_PULL_REQUEST')
+	REPO_SLUG = os.environ.get('TRAVIS_REPO_SLUG') # owner_name/repo_name
+	BOT_TOKEN = os.environ.get('TRAVIS_BOT_GITHUB_TOKEN')
+	BUILD_ID = os.environ.get('TRAVIS_BUILD_ID')
+
+	test_result = sys.stdin.read().strip()
+	comment = (
+        """
+The [build](https://travis-ci.org/tldr-pages/tldr/builds/{build_id}) for this PR has failed with the following message -
+```
+{comment_body}
+```
+Please rectify the error and make another push.
+        """).format(comment_body=test_result, build_id=BUILD_ID)
+
+	if PR_NUMBER != "false":
+		post_comment(PR_NUMBER, REPO_SLUG, comment, BOT_TOKEN)
--- a/scripts/post_pr_comment.py
+++ b/scripts/post_pr_comment.py
@@ -18,25 +18,25 @@ def post_comment(pr_id, repo_slug, comment_body, user_token):
 		print f.read()
 
 
-if __name__ == '__main__':
-	# Get the environment variables
-	PR_NUMBER = os.environ.get('TRAVIS_PULL_REQUEST')
-	REPO_SLUG = os.environ.get('TRAVIS_REPO_SLUG') # owner_name/repo_name
-	BOT_TOKEN = os.environ.get('TRAVIS_BOT_GITHUB_TOKEN')
-	BUILD_ID = os.environ.get('TRAVIS_BUILD_ID')
+# Get the environment variables
+PR_NUMBER = os.environ.get('TRAVIS_PULL_REQUEST')
+REPO_SLUG = os.environ.get('TRAVIS_REPO_SLUG') # owner_name/repo_name
+BOT_TOKEN = os.environ.get('TRAVIS_BOT_GITHUB_TOKEN')
+BUILD_ID = os.environ.get('TRAVIS_BUILD_ID')
 
-	# Read the test result output from stdin
-	test_result = sys.stdin.read().strip()
-	# Populate the template text
-	comment = (
-        """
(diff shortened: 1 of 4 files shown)


ostera:
wouldn't this run the tests twice?

waldyrious:
Adding some documentation to the script would be nice to allow others to chime in if edits are needed in the future :)

One thing I failed to figure out, for example, is where the `comment_body` contents comes from. Care to elucidate?

agnivade:
@waldyrious - Sure, will documentation. You mean the `comment_body` that is part of the string formatting ?

`comment_body` is just replaced by the `test_result` variable which gets its content from stdin. Let me know if its not clear, you can comment on the code diff so I can get the line no. exactly.

@ostera - That's right. But I couldn't get any other way to get the contents of the test failure. Travis has no such env var. There was one called `TRAVIS_FAILURE_RESULT` but that just has the exit code. I am open to any better ideas.

ostera:
@agnivade, [npm test 2>&1 > | tee test_results](https://tldr.ostera.io/tee)

agnivade:
Not bad ! And we should not be worried about cleaning the file later right ? Because its in travis env anyway.

ostera:
Wxactly, the file will just go away

waldyrious:
> `comment_body` is just replaced by the `test_result` variable which gets its content from stdin. 

Ok, that clarifies it, thanks. I had missed the `test_result = sys.stdin.read().strip()` line. Some comments (like the sentence I'm quoting above) would help following the data flow through the scripts.

agnivade:
@waldyrious - how does it look now ?

waldyrious:
@agnivade better, thanks! Are you still going to change the message to be posted, per the previous comments?

agnivade:
@waldyrious- PTAL

scripts/post_pr_comment.py
```diff
@@ -0,0 +1,42 @@
...
+	REPO_SLUG = os.environ.get('TRAVIS_REPO_SLUG') # owner_name/repo_name
+	BOT_TOKEN = os.environ.get('TRAVIS_BOT_GITHUB_TOKEN')
+	BUILD_ID = os.environ.get('TRAVIS_BUILD_ID')
+
+	# Read the test result output from stdin
+	test_result = sys.stdin.read().strip()
+	# Populate the template text
+	comment = (
+        """
+The [build](https://travis-ci.org/tldr-pages/tldr/builds/{build_id}) for this PR has failed with the following message:
```
waldyrious: Is there a reason the line wrapping couldn't be done?

waldyrious: line breaking :) to make the source code more readable.

agnivade: But that will split the output into separate lines. :( I just tried it.

waldyrious: Oh, ok. That's too bad, but no problem (that's why I asked whether there was something that made this undesirable).

waldyrious:
LGTM :D

scripts/post_pr_comment.py
```diff
@@ -0,0 +1,42 @@
...
+```
+{comment_body}
+```
+Please rectify the error and make another push.
```
ostera: `s/rectify/fix ` ? Plenty of our users won't have the broadest of vocabularies, we might as well make this more accessible.

ostera: Also `s/error/error(s)` -- it could potentially be many of them!

ostera: Lastly, `and push again` feels more natural than to `make another push`.

scripts/post_pr_comment.py
```diff
@@ -0,0 +1,42 @@
...
+	req = urllib2.Request(url=url,
+                       data=json.dumps({'body': comment_body}))
+	req.add_header('Authorization', 'token ' + user_token)
+	# Making the request
+	f = urllib2.urlopen(req)
+	if f.getcode() != 201:
+		print f.read()
+
+
+if __name__ == '__main__':
```
ostera: Considering we are invoking this through the interpreter, it shouldn't be necessary -- 

```python
import sys

print sys.stdin.read().strip()
```

And calling it as `python test.py < test.py` prints itself without the need of the wrapping block.

If we wanted to future-proof this, I'd put the block _within_ the if-name-main into a function that can be exported, and then within the block just call the function.

scripts/post_pr_comment.py
ostera: Here's an alternative way: 

```python
~ λ cat test.py

comment = (
"The [build]"
"(https://travis-ci.org/tldr-pages/tldr/builds/{build_id})"
" for this PR has failed with the following message:"
"\n```\n"
"{comment_body}"
"\n```\n"
).format(build_id=1234, comment_body="waat")

print comment
```

Which gives a result like this (the first backtick in the output is backslashed because Github's markdown renderer thought it was part of _it's_ markdown, and not the body of the block):
```markdown
~ λ py test.py
The [build](https://travis-ci.org/tldr-pages/tldr/builds/1234) for this PR has failed with the following message:
\```
waat
\```
```

And it looks like this:

----

The [build](https://travis-ci.org/tldr-pages/tldr/builds/1234) for this PR has failed with the following message:
```
waat
```

scripts/post_pr_comment.py
waldyrious: I think the build fails after the first error, so even if there are multiple problems, Travis will only report them one by one (not 100% sure though)

ostera: @waldyrious it would depend on how we run the tests, but the linter itself could fail with multiple linting errors.

scripts/post_pr_comment.py
agnivade: I get that since we are directly calling the file, we don't need the if-name-main check, but what is the significance of the `print sys.stdin.read().strip()` line ? Can you expand on that a bit ?

ostera: That was just part of my example --- we also happen to use that to get the stdin in your PR as well

agnivade: ah ok.

agnivade:
:+1:

PR #1139, opened 2016-10-31
fuck: add page

### LINKED ISSUE #1097: Add thefuck command
This should address #241 which at the time was just an alias and wasn't merged because of it. The tool is now a [full-fledged command line utility](https://github.com/nvbn/thefuck), and a very useful one at that, so we definitely should cover it.

assemusr:
fuck command has no options.
However, you should consider setting the alias first (for example in ~/.bashrc) as specified in 'thefuck' repository.
fixes #1097

### DIFF
--- a/pages/common/fuck.md
+++ b/pages/common/fuck.md
@@ -1,6 +1,7 @@
 # fuck
 
 > Corrects your previous console command.
+> Often repeated multiple times until it reaches a matched rule.
 
 - Try to match a rule for the previous command:
 
--- a/pages/common/fuck.md
+++ b/pages/common/fuck.md
@@ -1,0 +1,8 @@
+# fuck
+
+> Corrects your previous console command.
+
+- Try to match a rule for the previous command:
+
+`fuck`
+
--- a/pages/common/fuck.md
+++ b/pages/common/fuck.md
@@ -2,6 +2,10 @@
 
 > Corrects your previous console command.
 
+- Set the `fuck` alias to `thefuck` tool:
+
+`eval "$(thefuck --alias)"`
+
 - Try to match a rule for the previous command:
 
 `fuck`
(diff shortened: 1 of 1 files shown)


waldyrious:
I think it would be acceptable to add the alias-setting usage:

`eval "$(thefuck --alias)"`

assemusr:
The man page has a specific structure, where should I add the alias-setting?

waldyrious:
I'm not sure what you mean with "the man page has a specific structure". Do you mean the tldr page? I'd add this after the current example, as if a separate example. There's precedent for this kind of deviation from the standard format, see e.g. [cron](https://github.com/tldr-pages/tldr/blob/master/pages/common/crontab.md), whose last 3 "examples" are not really invocations of the command.

assemusr:
I see. But this is not supposed to be in the examples. All examples should have the 'command' argument within them, that's what any newcomer would think, and that's part of the Markdown format in CONTRIBUTING.md file.
However, the tool itself will let the user know if he didn't set-up the alias. Take a look at this:
screenshot from 2016-11-02 06-05-26

waldyrious:
@assemusr I understand, but CONTRIBUTING.md specifies guidelines, not rules set in stone. It's ok to bend the rules a little, when doing so provides more usefulness for the users. The cron page doesn't follow the rules strictly, but not providing the syntax there would be a disservice to users. Likewise, commands that start an interactive session, like vim, have examples that aren't actual invocations of the command, but that's the compromise we've found in order to provide the most useful information about the tool without deviating too much from the format.

So in this case, since this command has pretty much a single usage pattern, it's perfectly OK to use the remaining space to provide information that would spare users of the tool an extra wasted invocation (which would output an error message rather than perform the tool's actual job, as you showed).

waldyrious:
LGTM (and thanks for understanding).

agnivade:
:+1:

PR #1140, opened 2016-10-31
install: add page

### LINKED ISSUE #1015: add install command
http://linux.die.net/man/1/install

assemusr:
fixes #1015


pages/common/install.md
```diff
@@ -0,0 +1,24 @@
+# install
+
+> Copy files and set attributes.
+> Similar to "cp", but allows you to control the attributes of destination files.
```
waldyrious: We normally use backticks to wrap code samples, rather than quote marks.

pages/common/install.md
```diff
@@ -0,0 +1,24 @@
+# install
+
+> Copy files and set attributes.
```
waldyrious: It would be really useful to explain why the tool is called "install". See for example the description of the kill command.

waldyrious:
Nice one, but it seems you based your branch off `fuck_command`, rather than off `master`. Please rebase :)

assemusr:
Done :)

waldyrious:
@assemusr cool 👍 do take a look at the inline comments, please.

assemusr:
How about that? I couldn't find a specific reason of why it called 'install'

pages/common/install.md
```diff
@@ -0,0 +1,24 @@
...
+> Provide a standardized way of manipulating a file/directory's ownership and permissions while copying file or creating directory.
+
+- Copy files to destination (overwrite):
```
waldyrious: How about: "Copy files to destination, overwriting any existing files"?

pages/common/install.md
```diff
@@ -0,0 +1,24 @@
+# install
+
+> Copy files and set attributes.
+> Provide a standardized way of manipulating a file/directory's ownership and permissions while copying file or creating directory.
```
assemusr: There's no such explanation. However, from the linguistics point-of-view says: place (something) in a new position of authority. But that has nothing to do with a man page, for example the 'cron command' https://github.com/tldr-pages/tldr/blob/master/pages/common/crontab.md didn't say that it named after "chronos" the Greek word for time.

pages/common/install.md
assemusr: overwrite if destination was (a file), let's just delete that.

pages/common/install.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Copy files to destination (overwrite):
+
+`install {{path/to/source}} {{path/to/destination}}`
+
+- Copy and set ownership to destination:
```
assemusr: okay

pages/common/install.md
```diff
@@ -0,0 +1,24 @@
...
+
+`install -o {{user}} {{path/to/source}} {{path/to/destination}}`
+
+- Copy files to destination, setting their group ownership:
+
+`install -g {{user}} {{path/to/source}} {{path/to/destination}}`
+
+- Copy files to destination, setting their permissions:
+
+`install -m {{mode}} {{path/to/source}} {{path/to/destination}}`
```
agnivade: Does it accept both numbered mode(077) and alphabetical mode(+x) ?

agnivade:
LGTM

pages/common/install.md
waldyrious: Maybe a concrete example would make usage more evident. I'd favor an alphabetical mode example :) Note that if you do this, the word "mode" should be included in the description, since it wouldn't appear in the token itself.

pages/common/install.md
waldyrious: I was thinking more along the lines of the [kill command](https://github.com/tldr-pages/tldr/blob/master/pages/common/kill.md), which starts by describing what the command actually does ("Sends a signal to a process"), but follows that up by mentioning the use that gives the command its name ("usually related to stopping the process."). In this case, we could mention that a major use of this is to setup programs downloaded or built from source (e.g. copy files to a system location like /usr/local/bin, give them the appropriate permissions/ownership so they can be used by all users, and mark them as executable) hence "install". Does that sound reasonable to you?

waldyrious: Are you sure that's not the intent/main use of the command? Here's what I get from [this man page](https://linux.die.net/man/1/install):

> copies files (often just compiled) into destination locations you choose. If you want to download and install a ready-to-use package on a GNU/Linux system, you should instead be using a package manager like yum(1) or apt-get(1).

And from [this one](http://ss64.com/bash/install.html):

> It is typically used in Makefiles to copy programs into their destination directories.

We could either embrace the current meaning of "install" as in "install a program" and explain that this is the typical usage given to this command, as I've previously suggested, or we could revert to the strict etymological meaning of "install" as "place something in a given place", but in that case we'd still have to make the connection explicitly (mention "install" as a synonym for whatever expression we use). I think the first option is easier to accomplish in 1-2 lines of a main description.

assemusr: It really depends on the point-of-view, do you want the common usage or the overall technical capabilities. It still like `cp` with just some additional options, even if its often usage is in Makefiles.

Now the last two commits has all changes you suggested.

waldyrious:
Sorry for the delay, @assemusr. I'll think of something for the main description and propose it as a separate PR. Thanks!

PR #1144, opened 2016-10-31
bc: add a line about interactive usage with mathlib

kalebo:
Usually `bc` isn't very useful if it doesn't return floating point numbers where appropriate. e.g., 3/4 returns 0 when started with `-i` but for most users they would expect 0.75 to be returned so the tldr should mention that the argument `-l`does floating point arithmetic.

### DIFF
--- a/pages/common/bc.md
+++ b/pages/common/bc.md
@@ -2,11 +2,7 @@
 
 > Calculator.
 
-- Run calculator in interactive mode:
-
-`bc -i`
-
-- Run calculator in interactive mode with floating point support:
+- Run calculator in interactive mode using the standard math library:
 
 `bc -l`
 
--- a/pages/common/bc.md
+++ b/pages/common/bc.md
@@ -6,6 +6,10 @@
 
 `bc -i`
 
+- Run calculator in interactive mode with floating point support:
+
+`bc -l`
+
 - Calculate the result of an expression:
 
 `bc <<< "(1 + 2) * 2 ^ 2"`
--- a/pages/common/bc.md
+++ b/pages/common/bc.md
@@ -10,7 +10,7 @@
 
 `bc <<< "(1 + 2) * 2 ^ 2"`
 
-- Calculate with the given precision:
+- Calculate expression and force number of decimal places to 10:
 
 `bc <<< "scale=10; 5 / 3"`
 


waldyrious:
LGTM 👍

agnivade:
I am a bit confused on this. 

The man page for `bc` on my ubuntu box says `-l, --mathlib- Define the standard math library.` for this option. But the PR mentions `in interactive mode with floating point support`. First of all, will `-l` invoke interactive mode too ? And are floating point support and math library one and the same ? If yes, then we should use the later to denote [l].

kalebo:
@agnivade Yes, both `-i` and `-l` put `bc` into interactive mode. And yes, `-l` really means load the std math library. I agree the wording could be clearer. 

The main issue that  I see is that most people assume that a calculator should handle things like floating point division by default, but `bc` does not do that without specifying `-l`. I would almost say it's better to remove the line about `-i ` and replace it with a line that uses `-l` and perhaps state that it puts in into interactive mode with standard math libraries and floating point support. Or something to that effect.

agnivade:
No worries, just use `f[l]oating` and we should be good.

waldyrious:
IIUC the "l" in `-l` refers to "lib(rary)", right? How about "Run calculator using the standard math library, for floating point support (this enables interactive mode)"?

kalebo:
I reworded the entry as I realized that floating point support does not come from mathlib.You can run `scale=20` and get the same result in normal mode. However, enabling mathlib support also sets `scale=20` automatically. Ideally if a user just wants to run some quick calculations, including `-l` is easier than including `scale=x` which is more of a feature you'd want if you were writing a script. For this reason I replaced the line about `-i` with the one about `-l` since most users would expect their calculations not to be truncated at the decimal.

waldyrious:
What exactly is the scale? Number of decimal places for calculation results?

agnivade:
Hey @kalebo - Could you update the `scale` example to include the word 'decimal' too ? 'decimal precision' sounds much more clear than just 'precision'.

kalebo:
I switched up the wording in the scale example.  I'm a little torn between leaving in that example and ditching it. In short I don't want to make the tldr too in depth but bc's defaults are a bit of a stumbling block that I think would be nice to warn people about.

PR #1141, opened 2016-10-31
unar: add page

assemusr:
fixes #997

PR #1148, opened 2016-11-09
git-rebase: hopefully clarify some points

waldyrious:
I am not happy with the explanation of the `--onto` example, but I can't figure out a way to condense the information in https://github.com/tldr-pages/tldr/pull/1141#issuecomment-257587017 down to a single sentence + command... any help would be greatly appreciated.

### DIFF
--- a/pages/common/git-rebase.md
+++ b/pages/common/git-rebase.md
@@ -23,6 +23,6 @@
 
 `git rebase --onto {{new_base}} {{old_base}}`
 
-- Reapply the last 5 commits in-place, stopping to adjust them (reword, combine, reorder or remove):
+- Reapply the last 5 commits in-place, stopping to allow them to be reordered, omitted, combined or modified:
 
 `git rebase -i HEAD~5`
--- a/pages/common/git-rebase.md
+++ b/pages/common/git-rebase.md
@@ -5,7 +5,7 @@
 
 - Rebase the current branch on top of the master branch:
 
-`git rebase master`
+`git rebase {{master}}`
 
 - Start an interactive rebase, which allows the commits to be reordered, omitted, combined or modified:
 
@@ -25,4 +25,4 @@
 
 - Reapply the last 5 commits in-place, stopping to allow them to be reordered, omitted, combined or modified:
 
-`git rebase -i HEAD~5`
+`git rebase -i {{HEAD~5}}`
--- a/pages/common/git-rebase.md
+++ b/pages/common/git-rebase.md
@@ -1,23 +1,24 @@
 # git rebase
 
 > Apply local commits on top of another branch's history.
+> This effectively "moves" an entire branch to another base, by creating copies of the commits in the new location.
 
-- Rebase your local branch interactively with the latest changes in local master:
+- Rebase the current branch on top of the master branch:
 
-`git rebase -i master`
+`git rebase master`
 
-- Rebase your local branch interactively with the latest changes from upstream:
+- Start an interactive rebase, which allows the commits to be reordered, omitted, combined or modified:
 
-`git fetch origin; git rebase -i origin/master`
+`git rebase -i {{target_base_branch}}`
 
-- Handle an active rebase merge failure, after editing conflicting file(s):
+- Continue a rebase that was interrupted by a merge failure, after editing conflicting files:
 
 `git rebase --continue`
 
-- Abort a rebase in-progress:
+- Abort a rebase in progress (e.g. if it is interrupted by a merge conflict):
 
 `git rebase --abort`
 
-- Rebase your local branch by specifying new base commit and old base commit:
+- Rebase a branch starting from a specific base commit, rather than the common ancestor shared with the target branch:
 
 `git rebase --onto {{new_base_commit}} {{old_base_commit}}`
--- a/pages/common/git-rebase.md
+++ b/pages/common/git-rebase.md
@@ -1,7 +1,7 @@
 # git rebase
 
-> Apply local commits on top of another branch's history.
-> This effectively "moves" an entire branch to another base, by creating copies of the commits in the new location.
+> Reapply commits from one branch on top of another branch.
+> Commonly used to "move" an entire branch to another base, creating copies of the commits in the new location.
 
 - Rebase the current branch on top of the master branch:
 
@@ -19,6 +19,10 @@
 
 `git rebase --abort`
 
-- Rebase a branch starting from a specific base commit, rather than the common ancestor shared with the target branch:
+- Move part of the current branch onto a new base, providing the old base to start from:
 
-`git rebase --onto {{new_base_commit}} {{old_base_commit}}`
+`git rebase --onto {{new_base}} {{old_base}}`
+
+- Reapply the last 5 commits in-place, stopping to adjust them (reword, combine, reorder or remove):
+
+`git rebase -i HEAD~5`


pages/common/git-rebase.md
```diff
@@ -1,23 +1,24 @@
...
 
 `git rebase --continue`
 
-- Abort a rebase in-progress:
+- Abort a rebase in progress (e.g. if it is interrupted by a merge conflict):
 
 `git rebase --abort`
 
-- Rebase your local branch by specifying new base commit and old base commit:
+- Rebase a branch starting from a specific base commit, rather than the common ancestor shared with the target branch:
```
liuderchi: > `Rebase a branch ...`

- Would it be clearer if saying `Rebase current branch ...` ?

pages/common/git-rebase.md
```diff
@@ -1,23 +1,24 @@
...
 
 `git rebase --onto {{new_base_commit}} {{old_base_commit}}`
```
liuderchi: IMHO taking explanation in line 22 as context, it would be clearer if 

  * `{{new_base_commit}}` and `{{old_base_commit}}` are renamed using words appeared in line 22

      * e.g. `git rebase --onto {{target_commit}} {{starting_base_commit}}`

NOTE: `{{target_commit}}` is a wider usage and of course it can be specified with some target branch

pages/common/git-rebase.md
waldyrious: Indeed. I'll change.

pages/common/git-rebase.md
waldyrious: yes, I thought about using {{target_branch_name}} but that would lose the symmetry with the {{starting_base_commit}}. I wonder if we should make this example specifically about moving a branch B from being based on branch A (which is based on master) to being based on master directly (like the example in https://github.com/tldr-pages/tldr/pull/1141#issuecomment-257587017). That would allow both the source and target bases to be branches. But I'm struggling with how to word that description. What would you use?

liuderchi: Maybe we should step back and consider which usage (or the way we think of how `--onto` works) is most common.

---
Originally I think using `--onto` is like *moving bases*
so I describe it as `--onto {{new_base_commit}} {{old_base_commit}}`

---
But I found that more people consider it as:

Move *current branch* on top of *target branch* by saying some *starting commit* as base

follow that I guess we can forget symmetry say

`git rebase --onto {{target_branch}} {{starting_base_commit}}`

waldyrious: > I found that more people consider it as:

> Move *current branch* on top of *target branch* by saying some *starting commit* as base

What led you to that conclusion? Even the official documentation uses the "branch on top of branch" example, and I do think that's easier to grasp conceptually -- don't you?

liuderchi: I do agree with "branch on top of branch" example is more easier to understand.

but I guess `git rebase --onto master topicA topicB` is not often used 
since `topicA` can be growing so that base of starting base commit is not always `topicA` but more like maybe `topicA~3`

waldyrious: @liuderchi yes, that's a quite reasonable assumption, but it's a much harder situation to describe succinctly. Besides, tldr-pages' mission is not to provide extremely practical, copy-pastable one-liners (which is what commandlinefu provides), but rather to provide the set of examples that best allows users to *understand* what the command does, and then adapt the usage examples to match their needs.

So I would suggest we move forward with the `git rebase --onto master topicA topicB` example, which would best allow us to explain what that option does. What do you think?

waldyrious: Great that we're on the same page! Now, we still need a description for that example... how would you phrase that?

liuderchi: IMHO I would describe `git rebase --onto master topicA topicB` as:

rebase part of `topicB` onto branch `master` using `topicA` as base

how do you think?

waldyrious: ----
After thinking about this some more, I'd say that the main point of confusion is that `--onto` is implied in other invocations of rebase, but instead of introducing something like `--from` to disabiguate `git rebase <new-base> <old-base>` (note: this doesn't actually work with current git!) from, say, `git rebase <branch-to-be-moved> <new-base>`, they instead introduced the `--onto`. It would be much clearer to go from `git rebase new-base` to `git rebase new-base --from old-base`, but alas, they decided otherwise :(

waldyrious: Another option could be:

> Rebase the current branch branch starting from a manually specified base rather than the automatically detected one:  
> `git rebase --onto newbase oldbase`

Do you prefer this one, or the example I suggested above?

liuderchi: I prefer second one.

waldyrious: Hmm, I was actually wary that the second one could be a bit too verbose and jargon-y. On the other hand, it better describes what is going on. @ostera, @agnivade, what do you guys think?

pages/common/git-rebase.md
```diff
@@ -1,23 +1,24 @@
 # git rebase
 
 > Apply local commits on top of another branch's history.
```
waldyrious: Maybe "Reapply the commits from one branch to the tip of another branch" would be better?

liuderchi: I'm not contributor of this line but I think it's more descriptive.

pages/common/git-rebase.md
agnivade: Whatever you guys decide. IMO this concept is too confusing to be explained in a single line. I will always need a page with examples to grasp this :stuck_out_tongue_closed_eyes:

pages/common/git-rebase.md
waldyrious: Thanks @liuderchi, I'll change it accordingly.

pages/common/git-rebase.md
mhagger: If you ask me, the `--onto` usage is really confusing. Usually I come at it from a command like

    git rebase master

, which (in my head) means "rebase the current branch onto master". What it really means is "rebase the commits `master..HEAD` onto master", but that's not what I'm thinking. The "master" argument is used for **two things**: defining the start of the range of commits to rebase, and defining the attachment point for the rewritten commits.

Then I think to myself "no, I don't want to rebase all of the commits on my current branch, just the last three". What I'm changing relative to the earlier command is the starting point of the range of commits to rebase. It is totally unintuitive to me that I have to type

    git rebase --onto master HEAD^^^

because the `--onto` option makes it look like I'm changing the destination of the rebase. But that is almost never the case; when I resort to `--onto` I'm almost always adjusting the starting point of the range.

But all that whining doesn't change the status quo, which I would express as:

* `git rebase master` means roughly "transplant the commits `master..HEAD` onto branch `master`."
* `git rebase --onto master topicA` means roughly "transplant the commits `topicA..HEAD` onto commit `master`."
* `git rebase --onto master topicA topicB` means roughly "switch to branch `topicB`, then transplant the commits `topicA..topicB` onto commit `master`."

Remember that in all these cases, the notation `A..B` means "all of the commits that are in `B` but are not in `A`, as usual in Git.

ewels:
Suggestion for an extra example (as I have to google it myself more than I would like to admit):

- Interactively rebase the last 4 commits in the local branch:
```
git rebase -i HEAD~4
```

May be too similar to the first example though..?

pages/common/git-rebase.md
waldyrious: @mhagger thanks for your input. I left this discussion alone for some time to see if I could come back to it with a fresh head. My thoughts after reviewing the whole discussion, drawing deveral diagrams and testing multiple command invocations and the corresponding descriptions, are:

- we should use the simplest possible `--onto` invocation, which is already advanced enough for an introductory cheatsheet like this. That means using as many defaults as possible -- e.g. rebasing the current branch so the third argument can be omitted.
- I'd rather not introduce the `..` notation, which is its own beast to describe succinctly.
- Instead of an accurate description, we should strive for one that could be easily conceptually grasped, while simultaneously mapping to a simple enough command invocation (this includes choosing between using a third branch as a starting base versus walking back from the current branch using `~`).

That said, my current proposal would be:
```
- Transplant the current branch from one base to another:

`git rebase --onto {{new_base}} {{old_base}}`
```

(We could use "move" instead of "transplant", but the "base" intuition may be slightly less evident in that case)

What do you guys think? @liuderchi, @mhagger

waldyrious:
@ewels that's a nice one, but the description would need to be tweaked. I believe the intent there is to reapply the last 4 commits in-place, stopping to adjust them (reword, combine, reorder, remove, etc.) right? If so, would the description I just used be ok in your view?

ewels:
@waldyrious This is kind of a huge PR now so I'm not totally sure which description you mean 😁  But  yes, that's the intention. The main reason for my comment is the inclusion of the `HEAD~4` example bit though, basically an example of using within a single branch rather than across branches.

waldyrious:
@ewels I was talking about the description you gave, "Interactively rebase the last 4 commits in the local branch:". By saying "of using within a single branch rather than across branches" you confirm what I expected (rebase in-place). In that light, would this look good to you:

```
- Reapply the last 3 commits in-place, stopping to adjust them (reword, combine, reorder, remove, etc.)

`git rebase -i HEAD~4`
```
?

ewels:
Ah I see - yes that looks great 👍

agnivade:
@waldyrious - ping ?

waldyrious:
This is indeed quite tricky. Please take a look at the latest version -- @liuderchi, @mhagger, @ewels, @agnivade -- and let me know what you think :)

pages/common/git-rebase.md
```diff
@@ -1,23 +1,28 @@
...
 
-- Rebase your local branch by specifying new base commit and old base commit:
+- Move part of the current branch onto a new base, providing the old base to start from:
 
-`git rebase --onto {{new_base_commit}} {{old_base_commit}}`
+`git rebase --onto {{new_base}} {{old_base}}`
+
+- Reapply the last 5 commits in-place, stopping to adjust them (reword, combine, reorder or remove):
+
+`git rebase -i HEAD~5`
```
agnivade: wrap in tokens.

liuderchi: IMO interactive rebasing is also a common git technique. Good for me to add it.

Not only you can *reword, combine(fixup, squash), reorder, remove(drop)*
but also you can *split specific one commit into several ones (use edit command)*

waldyrious: I should have used the same description as the first -i example. @liuderchi would "reordered, omitted, combined or modified" be sufficiently inclusive?

liuderchi: IMO pretty inclusive with only one question.

let's checkout the original description:
```
# Commands:
# p, pick = use commit
# r, reword = use commit, but edit the commit message
# e, edit = use commit, but stop for amending
# s, squash = use commit, but meld into previous commit
# f, fixup = like "squash", but discard this commit's log message
# x, exec = run command (the rest of the line) using shell
# d, drop = remove commit
```
reword, edit --> *modified*
squash, fixup --> *combined*
drop --> Does *omitted* you mentioned means "drop or remove" ?

ewels:
Looks good! 👍

pages/common/git-rebase.md
waldyrious: > Does omitted you mentioned means "drop or remove" ?

Yes :) I would rather use "omitted" than "removed" because the original commit is not really deleted, and "drop" IMO is not as intuitive as "omitted" -- but I'd be willing to change to "drop" if you feel strongly about that.

ewels: Personally I'd be in favour of using the same terminology as git / the tool itself - anyone reading this help is going to be confronted with "drop" a few seconds after reading this message..

waldyrious: That could work, although as I said above, these words are meant to summarize the operations available on an interactive rebase, not match one-to-one. One could argue the reverse, for example: that by finding a direct connection with drop, the user could get confused that neither "modify" nor "combine" are available commands in the rebase.

Ultimately, it's a minor issue, so I'm ok picking whatever option reaches consensus :)

agnivade: I am okay with both approaches. Doesn't make much difference tbh.

waldyrious: @ewels, @liuderchi I'll wait for your final statements and then we can move ahead with whatever the consensus ends up being.

ewels: I'm fine with both approaches too. @waldyrious - you're the boss, if you prefer omitted then go for that :)

liuderchi: @waldyrious I'm ok with that. Thank you.

waldyrious:
@agnivade, please do a last check and approve if it looks good to you :)

waldyrious:
Took several months, but it's finally ready to merge! 🎉

PR #1153, opened 2016-11-15
ansible: add page

dikarel:
Add new page for `ansible`.

I've verified the commands listed in the page on my MacBook, with `localhost` as the test group

### DIFF
--- a/pages/common/ansible.md
+++ b/pages/common/ansible.md
@@ -1,0 +1,27 @@
+# ansible
+
+> Manage groups of computers remotely over SSH
+
+- List hosts belonging to a group:
+
+`ansible {{group}} --list-hosts`
+
+- Ping a group of hosts by invoking the ping module:
+
+`ansible {{group}} -m ping`
+
+- Display facts about a group of hosts by invoking the setup module:
+
+`ansible {{group}} -m setup`
+
+- Execute a command on a group of hosts by invoking command module with arguments:
+
+`ansible {{group}} -m command -a '{{my command}}'`
+
+- Execute a command with administrative privileges:
+
+`ansible {{group}} --become --ask-become-pass -m command -a '{{my command}}'`
+
+- Execute a command using a custom inventory file:
+
+`ansible {{group}} -i {{inventory_file}} -m command -a '{{my command}}'`
--- a/pages/common/ansible.md
+++ b/pages/common/ansible.md
@@ -1,6 +1,7 @@
 # ansible
 
-> Manage groups of computers remotely over SSH. Use the /etc/ansible/hosts file to add new groups/hosts
+> Manage groups of computers remotely over SSH.
+> Use the /etc/ansible/hosts file to add new groups/hosts.
 
 - List hosts belonging to a group:
 
--- a/pages/common/ansible.md
+++ b/pages/common/ansible.md
@@ -1,6 +1,6 @@
 # ansible
 
-> Manage groups of computers remotely over SSH
+> Manage groups of computers remotely over SSH. Use the /etc/ansible/hosts file to add new groups/hosts
 
 - List hosts belonging to a group:
 


pages/common/ansible.md
```diff
@@ -0,0 +1,27 @@
+# ansible
+
+> Manage groups of computers remotely over SSH
```
dikarel: This description is designed to communicate the following:
- This tool manages groups, and not just a single computer
- You'll need SSH on the remote machines as a pre-req
- It is a remote config/administration tool

pages/common/ansible.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Run command:
```
dikarel: This example is placed at the top, with the expectation that most of the time the user will be running commands

pages/common/ansible.md
```diff
@@ -0,0 +1,27 @@
...
+
+`ansible {{group}} -m command '{{command}}'`
+
+- Run command with sudo:
```
dikarel: A lot of admin-related commands will require sudo.

This one is a mutation of `Run command`, thus placed right after it for quick lookup.

pages/common/ansible.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Get system info:
+
+`ansible {{group}} -m setup`
+
+- List computers belonging to a group:
+
+`ansible {{group}} --list-hosts`
+
+- Using a custom inventory file:
```
dikarel: I added this command because my workflow uses this specific option a lot. I have multiple Ansible projects, each having their own inventory file.

I'm not sure if this assumption applies to everyone else, but I felt that the amount of documentation here was still manageable.

pages/common/ansible.md
waldyrious: What would you consider to be the getting-started / init command? Or the simplest one? These are better candidates for the first command than the most frequent one, accordign to [guideline n.4](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).

pages/common/ansible.md
```diff
@@ -0,0 +1,27 @@
+# ansible
+
+> Manage groups of computers remotely over SSH
+
+- Run command:
+
+`ansible {{group}} -m command -a '{{command}}'`
+
+- Run command with sudo:
```
waldyrious: "with sudo" --> "with administrative privileges"

pages/common/ansible.md
waldyrious: 6 examples is fine, but the command description should start with a verb.

pages/common/ansible.md
dikarel: AFAIK there isn't an init/setup command for this tool. Unless told otherwise, Ansible will simply read from a global config file ([`/etc/ansible/hosts`](http://docs.ansible.com/ansible/intro_inventory.html)).

In terms of simplicity, the following are good candidates:
- `command`: This is the bread and butter of this tool, making it possible to manage computers remotely
- `ping` or `setup`: These are simpler since it doesn't require any user input, but they're not the main reason someone would use Ansible

Note that `setup` does not actually perform any kind of initialization. It just [gathers facts about remote hosts](http://docs.ansible.com/ansible/setup_module.html)

pages/common/ansible.md
dikarel: Sounds good. I'll tweak the descriptions so that they use verbs and and unambiguous

pages/common/ansible.md
dikarel: Sounds good. Will do

pages/common/ansible.md
```diff
@@ -0,0 +1,27 @@
+# ansible
+
+> Manage groups of computers remotely over SSH
+
+- Execute a command on a group of hosts:
```
dikarel: **New changes**
- Use complete sentences that start with verbs
- Use phrasing similar to the [official docs](http://docs.ansible.com/ansible/modules_by_category.html)

RE: command simplicity, please see [my comments](https://github.com/tldr-pages/tldr/pull/1153#discussion_r88134311)

waldyrious: What does the -m stand for?

pages/common/ansible.md
```diff
@@ -0,0 +1,27 @@
...
+
+`ansible {{group}} -m command -a '{{command}}'`
+
+- Execute a command with administrative priviledges:
+
+`ansible {{group}} -bKm command -a '{{command}}'`
```
waldyrious: What do b and K stand for?

dikarel: - **b (become):** Run with [administrative priviledges](http://docs.ansible.com/ansible/become.html#become-privilege-escalation)
- **K (ask become pass):** Ask user to input password for sudo (otherwise Ansible will try to sudo using an empty password)

Excerpt from `ansible -h`:

screen shot 2016-11-16 at 7 09 19 am

pages/common/ansible.md
dikarel: Doing useful things in Ansible requires invoking modules ([full list](http://docs.ansible.com/ansible/list_of_all_modules.html)). Thus, we specify:

- **-m (module name):** The Ansible module to run
- **-a (module args):** Arguments for that module

The most straightforward module to use would be the [command module](docs.ansible.com/ansible/command_module.html), which lets users run ad-hoc commands on remote machines. It's usually shown in beginner tutorials. e.g. https://youtu.be/MfoAb50Br94?t=189

waldyrious: Ok, then perhaps something similar to "Run the command module on a group of hosts, passing the desired command as an argument" would be more helpful to describe this in a way that provides the right mental model for new users.

Btw, this might be a good reason to start with a module that doesn't require arguments, e.g. ping, since then we'd introduce the options gradually as recommended in the guidelines. The pages should be short enough that the most common commands not being first shouldn't pose a significant inconvenience to intermediate/experienced users.

pages/common/ansible.md
waldyrious: I'd separate -bK from -m, for conceptual clarity. Ideally we'd incorporate the words "become" and "as[k]" in the description, but it doesn't seem practical in this case.

pages/common/ansible.md
```diff
@@ -0,0 +1,27 @@
+# ansible
+
+> Manage groups of computers remotely over SSH
+
+- Display a list of hosts belonging to a group:
```
dikarel: I've re-ordered the commands so that the top 3 commands are as follows:

1. Display a list of hosts belonging to a group (simplest, not requiring a network connection and has the smallest output)
2. Ping a group of hosts
3. Display facts about a group of hosts (least simple, returning lots of data back and requiring a network connection)

pages/common/ansible.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Ping a group of hosts:
+
+`ansible {{group}} -m ping`
+
+- Display facts about a group of hosts:
+
+`ansible {{group}} -m setup`
+
+- Execute a command (using the command module) on a group of hosts:
```
dikarel: The following considerations were made for this description:

- The intent ("execute a command") should have higher priority over the mechanics ("command module"), as the user will be scanning the docs with a specific intention in mind. Thus, "Execute a command" is placed at the beginning of the sentence.

- The description should provide a hint at the correct mental model (as per comments by @waldyrious)

pages/common/ansible.md
```diff
@@ -0,0 +1,27 @@
...
+
+`ansible {{group}} -m command -a '{{my command}}'`
```
dikarel: I updated the placeholder command to `{{my command}}` to:

- Help users visually differentiate between the argument (`{{command}}`) and the module (`-m command`)
- Hint that it's okay to have spaces in the argument

We could go one step further and use a concrete example (e.g. `{{echo hello world}}`) to unambiguously demonstrate what's allowed, though I'll need a second opinion on that.

dikarel:
I'll squash my commits and fix the commit message once this passes review.

pages/common/ansible.md
waldyrious: This is a great choice for a first command 👍

pages/common/ansible.md
```diff
@@ -0,0 +1,27 @@
...
+
+`ansible {{group}} --list-hosts`
+
+- Ping a group of hosts:
+
+`ansible {{group}} --module-name ping`
+
+- Display facts about a group of hosts:
+
+`ansible {{group}} --module-name setup`
```
waldyrious: I think you could use the short option and mention the word "module" in the command description -- for example: "Invoke the setup module to display facts about a group of hosts". I feel that would hit a good balance between convenience and explicitness. But it's just a suggestion; if you prefer it this way, I'm fine with it.

pages/common/ansible.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Execute a command (using the command module) on a group of hosts:
+
+`ansible {{group}} --module-name command --args '{{my command}}'`
+
+- Execute a command with administrative priviledges:
```
waldyrious: typo: should be "privileges"

pages/common/ansible.md
```diff
@@ -0,0 +1,27 @@
...
+
+`ansible {{group}} --become --ask-become-pass --module-name command --args '{{my command}}'`
+
+- Execute a command using a custom inventory file:
+
+`ansible {{group}} --inventory-file {{path/to/inventory/file}} --module-name command --args '{{my command}}'`
```
waldyrious: Here the short options would be even more appropriate, since the command is quite long with the expanded option names. Perhaps we could introduce the options using the expanded versions, and from then on use the short ones?

pages/common/ansible.md
dikarel: By using single-letter options, it will conflict with contribution guideline point 3. I just want to confirm that this is okay in this case? I have no personal preference for either style, as long as we're consistent about it.

My thoughts on this:

- I agree that we should mention modules in the description to provide a pathway for beginners to discover advanced topics, as well as to hint at the correct mental model
- I feel it's easier for the beginner user to skim quickly for what they're looking for if we start the sentences with the desired outcome as opposed to technical specifics. My intuition tells me that the user will be thinking in terms of results, e.g. "I want to run a command" or "I want to ping hosts", as opposed to "I want to invoke module X".

My preferred solution would be something in the vein of "Display facts about a group of hosts, using the setup module", though I wouldn't mind being overridden as long as consistency is maintained.

waldyrious: > By using single-letter options, it will conflict with contribution guideline point 3.

You may be misreading the guideline. It says "Try to incorporate the spelled-out version of single-letter options **in the example's description**" (emphasis mine). If we can't add them to the description, expanding them in the command itself it an acceptable alternative, but it may even be that just the option's value can make it clear (e.g. in `-i {{path/to/inventory/file}}`, the token name allows one to infer that `-i` stands for "inventory". At the end of the day, this whole process is rather subjective, which is why there are guidelines, not hard-and-fast rules.

As for the phrasing, I agree that the second option is more helpful. As long as the relevant information is preserved, I see no problems putting the "what" first and the "how" later.

dikarel: My bad. I'll make adjustments to reflect this.

pages/common/ansible.md
dikarel: If by using descriptions we can hint at what each single-letter option stands for, then perhaps long-form options may be unnecessary and we can keep the commands succinct. Would the following pass muster?

- List hosts belonging to a group:

`ansible {{group}} --list-hosts`

- Ping a group of hosts by invoking the ping module:

`ansible {{group}} -m ping`

- Display facts about a group of hosts by invoking the setup module:

`ansible {{group}} -m setup`

- Execute a command on a group of hosts by invoking command module with arguments:

`ansible {{group}} -m command -a '{{my command}}'`

- Execute a command with administrative privileges:

`ansible {{group}} --become --ask-become-pass -m command -a '{{my command}}'`

- Execute a command using a custom inventory file:

`ansible {{group}} -i {{inventory_file}} -m command -a '{{my command}}'`

waldyrious: Yes, this seems ideal! 👌

agnivade: You can also use the `[]` convention like this - "setup [m]odule" for extra clarity.

waldyrious: I thought we had agreed that the [] syntax was superfluous when the short option letter is the first letter of the expanded word. Do you recall otherwise?

agnivade: > when the short option letter is the first letter of the expanded word

Oh was this condition there ? Anyways, makes sense. Was just a suggestion :stuck_out_tongue_closed_eyes:

waldyrious: Not sure what you mean by "there", but IIRC it was not documented in any permanent place. I made a suggestion to add that recommendation to contributing.md, but the ensuing discussion led to some downsides being pointed out, so we dropped the idea of adopting such a guideline. If I find the link to that discussion I'll add it here.

dikarel:
I've squashed the latest progress into a single commit. Should be ready to go

waldyrious:
LGTM, and thanks for the thorough commentary during the review, and patience with our nitpicking :)

agnivade:
2 things -

- Isn't there a command to show the list of groups ?
- Also, I feel we should explain how to add servers to a group. Maybe add a one-liner somewhere.

dikarel:
> Isn't there a command to show the list of groups ?

AFAIK there isn't any (see printout below). I maybe wrong though, in which case listing groups may also be a good first step for new users.

> Also, I feel we should explain how to add servers to a group. Maybe add a one-liner somewhere.

I agree it's a good idea to mention this briefly. FYI Ansible will point this out to the user the first time they try to run a command.

screen shot 2016-11-20 at 1 22 13 pm

Printout from running `ansible`:

```
Usage: ansible <host-pattern> [options]

Options:
  -a MODULE_ARGS, --args=MODULE_ARGS
                        module arguments
  --ask-vault-pass      ask for vault password
  -B SECONDS, --background=SECONDS
                        run asynchronously, failing after X seconds
                        (default=N/A)
  -C, --check           don't make any changes; instead, try to predict some
                        of the changes that may occur
  -D, --diff            when changing (small) files and templates, show the
                        differences in those files; works great with --check
  -e EXTRA_VARS, --extra-vars=EXTRA_VARS
                        set additional variables as key=value or YAML/JSON
  -f FORKS, --forks=FORKS
                        specify number of parallel processes to use
                        (default=5)
  -h, --help            show this help message and exit
  -i INVENTORY, --inventory-file=INVENTORY
                        specify inventory host path
                        (default=/etc/ansible/hosts) or comma separated host
                        list.
  -l SUBSET, --limit=SUBSET
                        further limit selected hosts to an additional pattern
  --list-hosts          outputs a list of matching hosts; does not execute
                        anything else
  -m MODULE_NAME, --module-name=MODULE_NAME
                        module name to execute (default=command)
  -M MODULE_PATH, --module-path=MODULE_PATH
                        specify path(s) to module library (default=None)
  --new-vault-password-file=NEW_VAULT_PASSWORD_FILE
                        new vault password file for rekey
  -o, --one-line        condense output
  --output=OUTPUT_FILE  output file name for encrypt or decrypt; use - for
                        stdout
  -P POLL_INTERVAL, --poll=POLL_INTERVAL
                        set the poll interval if using -B (default=15)
  --syntax-check        perform a syntax check on the playbook, but do not
                        execute it
  -t TREE, --tree=TREE  log output to this directory
  --vault-password-file=VAULT_PASSWORD_FILE
                        vault password file
  -v, --verbose         verbose mode (-vvv for more, -vvvv to enable
                        connection debugging)
  --version             show program's version number and exit

  Connection Options:
    control as whom and how to connect to hosts

    -k, --ask-pass      ask for connection password
    --private-key=PRIVATE_KEY_FILE, --key-file=PRIVATE_KEY_FILE
                        use this file to authenticate the connection
    -u REMOTE_USER, --user=REMOTE_USER
                        connect as this user (default=None)
    -c CONNECTION, --connection=CONNECTION
                        connection type to use (default=smart)
    -T TIMEOUT, --timeout=TIMEOUT
                        override the connection timeout in seconds
                        (default=10)
    --ssh-common-args=SSH_COMMON_ARGS
                        specify common arguments to pass to sftp/scp/ssh (e.g.
                        ProxyCommand)
    --sftp-extra-args=SFTP_EXTRA_ARGS
                        specify extra arguments to pass to sftp only (e.g. -f,
                        -l)
    --scp-extra-args=SCP_EXTRA_ARGS
                        specify extra arguments to pass to scp only (e.g. -l)
    --ssh-extra-args=SSH_EXTRA_ARGS
                        specify extra arguments to pass to ssh only (e.g. -R)

  Privilege Escalation Options:
    control how and which user you become as on target hosts

    -s, --sudo          run operations with sudo (nopasswd) (deprecated, use
                        become)
    -U SUDO_USER, --sudo-user=SUDO_USER
                        desired sudo user (default=root) (deprecated, use
                        become)
    -S, --su            run operations with su (deprecated, use become)
    -R SU_USER, --su-user=SU_USER
                        run operations with su as this user (default=root)
                        (deprecated, use become)
    -b, --become        run operations with become (does not imply password
                        prompting)
    --become-method=BECOME_METHOD
                        privilege escalation method to use (default=sudo),
                        valid choices: [ sudo | su | pbrun | pfexec | doas |
                        dzdo | ksu ]
    --become-user=BECOME_USER
                        run operations as this user (default=root)
    --ask-sudo-pass     ask for sudo password (deprecated, use become)
    --ask-su-pass       ask for su password (deprecated, use become)
    -K, --ask-become-pass
                        ask for privilege escalation password
```

agnivade:
>  I'm not quite sure where to put this one-liner while keeping descriptions short and snappy. 

Maybe you can add a second line in the main description mentioning to use the `/etc/ansible/hosts` file to add new groups/hosts.

pages/common/ansible.md
```diff
@@ -1,6 +1,6 @@
 # ansible
 
-> Manage groups of computers remotely over SSH
+> Manage groups of computers remotely over SSH. Use the /etc/ansible/hosts file to add new groups/hosts
```
agnivade: You can split this to a new line. See `crontab.md`

agnivade:
:+1:  Thanks for your patience !

dikarel:
👍  Thanks! Appreciate the patience from the maintainer side as well

PR #1154, opened 2016-11-16
logger: add page

agnivade:
Splitting page into 2 variants because mac does not have the expanded options. So it's difficult to remember -P (port) and -p (priority). Hence using expanded options in linux, and short options in mac.

### DIFF
--- a/pages/linux/logger.md
+++ b/pages/linux/logger.md
@@ -10,7 +10,7 @@
 
 `echo {{log_entry}} | logger`
 
-- Send the output to a remote syslog server running at a given port. Default port is `syslog`:
+- Send the output to a remote syslog server running at a given port. Default port is 514:
 
 `echo {{log_entry}} | logger --server {{hostname}} --port {{port}}`
 
--- a/pages/linux/logger.md
+++ b/pages/linux/logger.md
@@ -1,6 +1,6 @@
 # logger
 
-> Add messages to syslog.
+> Add messages to syslog (/var/log/syslog).
 
 - Log a message to syslog:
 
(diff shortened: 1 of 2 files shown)


pages/linux/logger.md
```diff
@@ -0,0 +1,23 @@
+# logger
+
+> Add messages to syslog.
```
waldyrious: maybe use a more descriptive name for "syslog"?

pages/linux/logger.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Log a message to syslog:
+
+`logger {{message}}`
+
+- Take input from stdin and log to syslog:
+
+`tail -f {{app.log}} | logger`
```
waldyrious: It's unclear where app.log comes from... care to explain a bit?

agnivade: Anywhere .. any file .. doesn't matter. I just need some data from stdin.

pages/linux/logger.md
agnivade: Actually `syslog` is the correct term because that's what it is. "system log" might mean that the logs are stored somewhere in the system. Should I add `/var/log/syslog` at the end ? Does that make it clearer ?

waldyrious: > Should I add `/var/log/syslog` at the end ? Does that make it clearer ?

I'd say it does, yeah.

pages/linux/logger.md
waldyrious: Hmm... maybe something like `echo {{"log entry"}}` would make it clearer, then. Or using cat instead of tail -f, and perhaps a path in the token (something like path/to/log_file). Otherwise it doesn't seem particularly generic. What do you think?

agnivade: yep, makes sense.

waldyrious:
LGTM

agnivade:
No negative comments so far. :+1: from me.

PR #1156, opened 2016-11-18
uname: split page to provide contextually-relevant info in the main desc.

waldyrious:
Instead of a "if you're on Linux, which is irrelevant to those who aren't, this change uses tldr-pages' platform-specific support to provide relevant information for Linux and OS X users.


agnivade:
:+1:

waldyrious:
Thanks for the review @agnivade. This is a small change, so let's :shipit:.

PR #1158, opened 2016-11-19
conda: add page

ewels:
Mostly taken from this cheat sheet: http://conda.pydata.org/docs/using/cheatsheet.html

I've tried to keep it short. If people think that there's room, options for extra examples are:

- Search for a package by name
`conda search {{package_name}}`

- List packages in current environment
`conda list`

Also, if the _tldr_ syntax allows it, then the activate / deactivate examples may be clearer together:

- Load / unload an environment
`source activate {{environment_name}}`
`source deactivate {{environment_name}}`


waldyrious:
We've recently discussed the option to expand the limit for the number of examples per page, but the maximum we've agreed on is 8, so only one extra exampled could be added in this case. The `source` example is not outright forbidden -- our main mission is to make the pages useful, not to make them comply with the rules to a T, so a little wiggle room is allowed, when sensible.

waldyrious:
Argh, sorry, clicked the wrong button

pages/common/conda.md
```diff
@@ -0,0 +1,31 @@
+# conda
+
+> Package, dependency and environment management for any language.
+
+- Create a new environment:
+
+`conda create --name {{environment_name}} {{packages}}`
```
waldyrious: Please mention the packages token in the example description.

pages/common/conda.md
```diff
@@ -0,0 +1,31 @@
...
+
+`source deactivate {{environment_name}}`
+
+- Delete an environment:
+
+`conda remove --name {{environment_name}} --all`
+
+- Install one or more conda packages:
+
+`conda install {{packages}}`
```
waldyrious: Does this happen in a specific environment, or globally in the current machine, or on a per-user directory, or...?

pages/common/conda.md
```diff
@@ -0,0 +1,31 @@
+# conda
+
+> Package, dependency and environment management for any language.
```
ewels: Absolutely, I just copied and pasted this from the website and chopped off the list of examples. "Programming language" is better.

pages/common/conda.md
```diff
@@ -0,0 +1,31 @@
...
+
+`conda info --envs`
+
+- Load an environment:
+
+`source activate {{environment_name}}`
+
+- Exit from an environment:
+
+`source deactivate {{environment_name}}`
```
ewels: Ah nice - yes that's better than having two examples.

pages/common/conda.md
ewels: Used like this it will install into the currently active environment. You can also specify `--name` to install a package into a specific named environment.

The description should specify this though, good call.

pages/common/conda.md
```diff
@@ -0,0 +1,35 @@
+# conda
+
+> Package, dependency and environment management for any programming language.
+
+- Create a new environment, installing named packages:
```
ewels: Not sure if it would be nice to give an example of common packages? _e.g.:_ `python=3.4`

pages/common/conda.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Delete an environment (remove all packages):
+
+`conda remove --name {{environment_name}} --all`
+
+- Search conda channels for a package by name:
+
+`conda search {{package_name}}`
+
+- Install one or more packages into the current environment:
```
ewels: NB: Haven't mentioned conda channels here at all. For example, many packages I used are part of [bioconda](https://bioconda.github.io/) and could be installed with `-c bioconda` (if the channel config isn't already saved).

ewels:
Thanks for your comments - I've made a few changes and pushed an update. I also added a couple more examples to get up to the limit of 8 examples.

pages/common/conda.md
waldyrious: Yes, specific examples are often helpful, e.g. for demonstrating the expected format. In this case if you think a concrete example would be more helpful, by all means go ahead.

pages/common/conda.md
waldyrious: There are almost always things that have to be left out when creating a tldr page. The best we can do is combine two or three options in a single example, but that can't make the example too overwhelming or unnecessarily complex. In this case I suspect leaving the notion of channels out is a reasonable compromise.

pages/common/conda.md
```diff
@@ -0,0 +1,35 @@
+# conda
+
+> Package, dependency and environment management for any programming language.
+
+- Create a new environment, installing named packages (eg. python=2.7):
```
waldyrious: I was thinking about making the change in the token itself. That'd be more directly understandable. Maybe an example with two packages would make it clear how to pass several packages to this tool.

ewels: Ah ok.. This is actually how I first wrote it, but then I changed it before submitting the pull request. I wasn't sure if it would be confusing as it would be difficult to distinguish between named parameters/flags and example strings? Are there any other _tldr_ pages that you know of which do this? How would you propose that we write this?

waldyrious: There certainly are, but I can't give you a concrete example from the top of my head, I'd have to look at a few pages looking for that. In any case, I don't see many ways this could be written: just use two common packages as you would use in a real command, and wrap them in the token markers. In the description, you could say something like "installing two named packages into it".

pages/common/conda.md
```diff
@@ -0,0 +1,35 @@
+# conda
+
+> Package, dependency and environment management for any programming language.
+
+- Create a new environment, installing named packages into it:
+
+`conda create --name {{environment_name}} {{python=2.7}} {{matplotlib}}`
```
waldyrious: I think it would be more semantically meaningful to write `{{python=2.7 matplotlib}}`, since they're part of the same token (packages). Do you agree?

ewels: Yup, true

waldyrious:
LGTM, and thanks for the patience :)

pages/common/conda.md
```diff
@@ -0,0 +1,35 @@
...
+
+`conda remove --name {{environment_name}} --all`
+
+- Search conda channels for a package by name:
+
+`conda search {{package_name}}`
+
+- Install one or more packages into the current environment:
+
+`conda install {{packages}}`
```
agnivade: ah ok then.

waldyrious: @ewels good point -- it would make sense to be consistent here and use the same format as the first example. Maybe replace the second package with something else to better illustrate that they can be switched around.

ewels: So to clarify - you mean swap `{{packages}}` here with `{{python=3.4 numpy}}` or something?

waldyrious: yes :)

agnivade:
LGTM

ewels:
Thanks both!

Issue #1084, opened 2016-09-22
add bioinformatic tools (bedtools, samtools, etc.)


mashehu:
Possible sources for [bedtools](http://bedtools.readthedocs.io/en/latest/content/example-usage.html) or for [samtools](http://davetang.org/wiki/tiki-index.php?page=SAMTools).

ewels:
There are loads of bioinformatics tools that could be nicely documented with _tldr_. @waldyrious - would it make sense to keep them separate somehow, so that they don't bloat the main package?

waldyrious:
This is something that has to be decided on a case-by-case basis, by those familiar with the tools. There certainly is precedent for it, see git and csvkit (#702).

waldyrious:
I might have misunderstood you. By main package do you mean tldr-pages repo? Separate page stores have been floated before by @ostera (tried to find an issue or PR reference, but I guess we chatted on Gitter rather than here). @ostera could you chime in?

ewels:
Yes sorry, I wasn't very clear. I meant the tldr-pages repo. Could be something like the existing `common`/`linux`/`osx` type thing? Or a separate repo. I was vaguely thinking along the lines of the previously mentioned channels with conda. A nice way to collect tools of a similar nature. If you look at the list of bioinformatics software in the [bioconda conda channel](https://bioconda.github.io/) that they have then that gives you an idea for the potential number of _tldr_ entries (in a perfect world!). Though in fairness not all of these are command line tools.

waldyrious:
Yes, we'd need to define a standard for where such pages could live (e.g. git repositories with a "pages" directory, like this one) and how existing clients could be made to point to them (adding a new registry to their internal list, rather than having the default tldr pages hardcoded.)

But until then, I don't think it's a bad idea to just add more tools to this repo, since as text files they don't really take that much space. The only issue would be tool name collisions, but hopefully those should be rare enough to be dealt with individually.

waldyrious:
Update: samtools have now been added with the merging of #1159, great job @ewels!

I'll create a separate issue to track bedtools, rename this one and close it. If we leave it open-ended, it will never be closed...

PR #1159, opened 2016-11-19
samtools: add page

ewels:
First stab at adding a new page for [samtools](http://www.htslib.org/doc/samtools.html). See #1084.

### DIFF
--- a/pages/common/samtools.md
+++ b/pages/common/samtools.md
@@ -3,15 +3,15 @@
 > Tools for handling high-throughput sequencing (genomics) data.
 > Used for reading/writing/editing/indexing/viewing of data in SAM/BAM/CRAM format.
 
-- Convert a SAM input file (-S) to BAM stream (-b) and save to file:
+- Convert a SAM input file to BAM stream and save to file:
 
-`samtools view -bS {{input.sam}} > {{output.bam}}`
+`samtools view -S -b {{input.sam}} > {{output.bam}}`
 
-- Take input from stdin (-) and output a region, including the file header (-h). Prints to stdout:
+- Take input from stdin (-) and print the SAM header and any reads overlapping a specific region to stdout:
 
 `{{other_command}} | samtools view -h - chromosome:start-end`
 
-- Sort a file and save using -o (format chosen by output file extension):
+- Sort file and save to BAM (the output format is automatically determined from the output file's extension):
 
 `samtools sort {{input}} -o {{output.bam}}`
 
@@ -29,7 +29,7 @@
 
 - Merge multiple files:
 
-`samtools merge {{output}} {{input_1}} [{{input_2}}..{{input_n}}]`
+`samtools merge {{output}} {{input_1}} [{{input_2}}...]`
 
 - Split input file according to read groups:
 
--- a/pages/common/samtools.md
+++ b/pages/common/samtools.md
@@ -17,7 +17,7 @@
 
 - Index a sorted BAM file (creates {{sorted_input.bai}}):
 
-`samtools index {{sorted_input}}`
+`samtools index {{sorted_input.bam}}`
 
 - Print alignment statistics about a file:
 
--- a/pages/common/samtools.md
+++ b/pages/common/samtools.md
@@ -15,7 +15,7 @@
 
 `samtools sort {{input}} -o {{output.bam}}`
 
-- Index a sorted BAM file (creates {{sorted_input.bai}}):
+- Index a sorted BAM file (creates {{sorted_input.bam.bai}}):
 
 `samtools index {{sorted_input.bam}}`
 
--- a/pages/common/samtools.md
+++ b/pages/common/samtools.md
@@ -1,0 +1,28 @@
+# samtools
+
+> Tools for handling high-throughput sequencing (genomics) data.
+> Reading/writing/editing/indexing/viewing of data in SAM/BAM/CRAM format.
+
+- Convert a SAM file to BAM (use - for stdin):
+
+`samtools view -bS {{input.sam}} > {{output.bam}}`
+
+- Print a region of a file, include the file header:
+
+`samtools view -h {{input}} chr:start-end > {{output.sam}}`
+
+- Sort a file:
+
+`samtools sort -m {{memory_per_cpu}} -@ {{cpus}} {{input}} -o {{output.bam}}`
+
+- Index a sorted BAM file:
+
+`samtools index {{sorted_input}}`
+
+- Print alignment statistics about a file:
+
+`samtools flagstat {{sorted_input}}`
+
+- Count alignments to each index (chromosome / contig):
+
+`samtools idxstats {{sorted_indexed_input}}`
--- a/pages/common/samtools.md
+++ b/pages/common/samtools.md
@@ -1,21 +1,21 @@
 # samtools
 
 > Tools for handling high-throughput sequencing (genomics) data.
-> Reading/writing/editing/indexing/viewing of data in SAM/BAM/CRAM format.
+> Used for reading/writing/editing/indexing/viewing of data in SAM/BAM/CRAM format.
 
-- Convert a SAM file to BAM (use - for stdin):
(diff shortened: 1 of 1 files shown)


ewels:
It's kind of a fun game trying to condense stuff down into such a short list of examples 😄  Note that there are a bunch of subcommands that I didn't add, plus some have so many options that they could probably have their own _tldr_ page (eg. `samtools view`).

I've done all of this based on the commands that I use most myself, others may have different opinions.

pages/common/samtools.md
```diff
@@ -0,0 +1,28 @@
+# samtools
+
+> Tools for handling high-throughput sequencing (genomics) data.
+> Reading/writing/editing/indexing/viewing of data in SAM/BAM/CRAM format.
```
waldyrious: Minor nitpick, but I'd prefer this sentence to have a subject.

pages/common/samtools.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Convert a SAM file to BAM (use - for stdin):
+
+`samtools view -bS {{input.sam}} > {{output.bam}}`
```
waldyrious: is the S an argument to the -b option, or a separate option? If so, what do those letters stand for?

pages/common/samtools.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Print a region of a file, include the file header:
+
+`samtools view -h {{input}} chr:start-end > {{output.sam}}`
```
waldyrious: What does "chr" stand for?

waldyrious:
The condensing challenge isn't easy, indeed. One of the things that helps is using the example descriptions to explain what the options mean, and thus create a mental model that would help infer other combinations of options. We're actually considering expanding the maximum number of examples to 8, but in this case, have you considered the option of splitting these commands into one page per subcommand, as is done with git?

pages/common/samtools.md
ewels: Yup, this is me being lazy and copying + pasting again.

pages/common/samtools.md
ewels: `-S` is a separate option. It specifies `SAM` input format. In the latest versions of samtools this flag is actually ignored, the format is autodetected from the input filename extension. However, a range of samtools versions are still common so I typically include it in scripts for backwards compatibility.

`-b` specifies that the output should be in `BAM` format instead of the default `SAM` for `stdout`. This isn't needed if using `-o {{output.bam}}` (autodetected by file extension again).

pages/common/samtools.md
ewels: Chromosome. I would hope that this is pretty common knowledge for anyone using the tool, but I could extend it to `chromosome:start-end`.

ewels:
I did think about expanding these into multiple pages, but I wanted to remain consistent. Not all commands have lots of options and probably don't deserve their own files. I think with a bit of tweaking based on your suggestions that this is probably a happy middle ground for most people - covering the most commonly used commands without going into too much depth.

ewels:
Thanks again for the suggestions - update pushed with better descriptions (explaining flags more explicitly). I also changed the example commands a little to demonstrate typical unix piping, as this is often how you use the tool. Finally I added a couple more command examples to push up 8.

pages/common/samtools.md
```diff
@@ -0,0 +1,36 @@
+# samtools
+
+> Tools for handling high-throughput sequencing (genomics) data.
```
waldyrious: would "sequencing (genomics) data" work better as "gene sequencing data" (or some variation thereof if that isn't technically correct)?

ewels: _"high-throughput sequencing"_ is a commonly used term for this (often abbreviated to HTS), so those words go together quite nicely. I added _(genomics)_ to help clarify. If I'm honest I can't think of any alternatives that sound better to my ear..

pages/common/samtools.md
```diff
@@ -0,0 +1,36 @@
...
+> Used for reading/writing/editing/indexing/viewing of data in SAM/BAM/CRAM format.
+
+- Convert a SAM input file (-S) to BAM stream (-b) and save to file:
+
+`samtools view -bS {{input.sam}} > {{output.bam}}`
+
+- Take input from stdin (-) and output a region, including the file header (-h). Prints to stdout:
```
waldyrious: To clarify, here it's ok to mention the link `stdin` --> `-`, because it's not obvious.

pages/common/samtools.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Index a sorted BAM file (creates {{sorted_input.bai}}):
+
+`samtools index {{sorted_input}}`
+
+- Print alignment statistics about a file:
+
+`samtools flagstat {{sorted_input}}`
+
+- Count alignments to each index (chromosome / contig):
```
waldyrious: Could "contig" be expanded into its full form?

pages/common/samtools.md
```diff
@@ -0,0 +1,36 @@
...
+
+`samtools idxstats {{sorted_indexed_input}}`
+
+- Merge multiple files:
+
+`samtools merge {{output}} {{input_1}} [{{input_2}}..{{input_n}}]`
```
waldyrious: I'm wondering if `[{{input_2}}...]` would be clearer. @agnivade, any thoughts?

pages/common/samtools.md
ewels: "Contig" is the full form :) https://en.wikipedia.org/wiki/Contig

pages/common/samtools.md
```diff
@@ -0,0 +1,36 @@
...
+
+`samtools sort {{input}} -o {{output.bam}}`
+
+- Index a sorted BAM file (creates {{sorted_input.bai}}):
+
+`samtools index {{sorted_input}}`
+
+- Print alignment statistics about a file:
+
+`samtools flagstat {{sorted_input}}`
```
ewels: It's a numerical value assigned to a DNA read when it's aligned to a reference genome. There are several of them and they can be combined. There's a nice tool here to decode them, which should give an idea of what they look like: https://broadinstitute.github.io/picard/explain-flags.html

ewels:
Thanks again! I agree with pretty much all of your suggestions, have just pushed a quick update to address your comments. Let me know what you think..

waldyrious:
LGTM! Great contribution, thanks!

pages/common/samtools.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Take input from stdin (-) and print the SAM header and any reads overlapping a specific region to stdout:
+
+`{{other_command}} | samtools view -h - chromosome:start-end`
+
+- Sort file and save to BAM (the output format is automatically determined from the output file's extension):
+
+`samtools sort {{input}} -o {{output.bam}}`
+
+- Index a sorted BAM file (creates {{sorted_input.bai}}):
```
agnivade: shouldn't this be `sorted_input.bam` ?

ewels: The input file can be `.bam` or `.cram` format, but yes it probably should be `sorted_input.bam` as I specify _BAM_ in the description (plus this is by far the most common). 👍

ewels: ..(updated)

ewels: Oh I see, nope! That's meant to be `bai` :) Short for for _bam index_. Though actually you are correct that there's a typo - it should be `sorted_input.bam.bai` now that we have the input file as `sorted_input.bam`

ewels: ps. If you're curious, here is the full description of the command output:

> If an output filename is given, the index file will be written to _out.index_. Otherwise, for a CRAM file _aln.cram_, index file _aln.cram.**crai**_ will be created; for a BAM file _aln.bam_, either _aln.bam.**bai**_ or _aln.bam.**csi**_ will be created, depending on the index format selected.

..confused yet? ;) 90% of people will be using this command with `.bam` files though, so I think it's clearest to make the example explicitly for that, creating `.bam.bai` files as we now have.

agnivade:
LGTM

PR #1048, opened 2016-09-09
add CLA to contributing.md

waldyrious:
All contributors to this project are invited to voice their opinion about this Contributor License Agreement (CLA) :) I crafted it after doing some research on CLAs, particularly https://github.com/clahub/clahub#whats-a-cla and https://opensource.com/law/11/7/trouble-harmony-part-1

The text of the CLA can be found at https://cla-assistant.io/tldr-pages/tldr. Copied below for convenience:

> ## TDLR Pages contributor agreement

> (1) I assert that any past, present and future contributions I make to the tldr-pages project are my own work, and that I have the right to license them.
>
> (2) I agree to license all my contributions to the tldr-pages project under the terms of the license included in the tldr-pages repository I’m contributing to.
>
> (3) I grant the tldr-pages organization a license to distribute my contributions under any open content license compliant with the Open Definition [a], or any open source software license approved by the Open Source Initiative [b], as applicable.
>
> a. http://opendefinition.org/licenses
>
> b. https://opensource.org/licenses/alphabetical

/cc all the maintainers, and the top 50 contributors (as of today): @igorshubovych, @oxguy3, @rubenvereecken, @rprieto, @notpeter, @onesuper, @ostera, @agnivade, @pando, @felixonmars, @Cvetomird91, @alx741, @dylanrees, @Dahie, @joelhy, @steelywing, @lord63, @sbrl, @akoenig, @naxoc, @sethwoodworth, @Like-all, @slash3b, @britter, @ericbn, @chirag64, @mumumu, @bobstrecansky, @Larry850806, @chuanconggao, @Duologic, @matthewgao, @denis-sokolov, @AloisMahdal, @Leandros, @lagelalegal, @martypenner, @fordhurley, @kuanyui, @pindexis, @egilkh, @d33tah, @bripkens, @Fallon9111, @eddieantonio, @RoryCrispin, @tjwudi, @jlems, @dbrgn, @kumon

(this PR also serves as a test for the cla-assistant integration)

PR #1164, opened 2016-11-25
apt: add page

mueller-ma:
apt is similar to apt-get:
https://askubuntu.com/questions/445384/what-is-the-difference-between-apt-and-apt-get#446484

### DIFF
--- a/pages/linux/apt.md
+++ b/pages/linux/apt.md
@@ -1,0 +1,27 @@
+# apt
+
+> Debian and Ubuntu package management utility.
+
+- Synchronize list of packages and versions available. This should be run first, before running subsequent apt commands:
+
+`apt update`
+
+- Install a new package:
+
+`apt install {{package}}`
+
+- Remove a package:
+
+`apt remove {{package}}`
+
+- Upgrade installed packages to newest available versions:
+
+`apt upgrade`
+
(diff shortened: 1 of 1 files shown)


pages/linux/apt.md
```diff
@@ -0,0 +1,27 @@
+# apt
+
+> Debian and Ubuntu package management utility.
+
+- Synchronize list of packages and versions available. This should be run first, before running subsequent apt commands:
```
agnivade: > This should be run first, before running subsequent apt commands

While useful, I don't think its compulsory to run it every time, which is what this line implies. It might be better to say "Useful to run before searching or installing new packages".

mueller-ma:
Do I really have to login on two sites that are getting access to my profile and private mail?

waldyrious:
@mueller-ma yes, our community has agreed to use that tool to facilitate signing into the CLA, which would otherwise be cumbersome. See #1048 for background.

mueller-ma:
@waldyrious I won't give someone my private mail address. I will close that PR and someone other can create the site for apt. Just simple 's/apt-get/apt/g'

waldyrious:
Sure, no problem. Thanks for the contribution anyway.

PR #1165, opened 2016-11-25
Adds dpkg note for finding which package owns a file

sethwoodworth:
This is the `apt` equivalent of `yum`'s `what-provides` command.

    dpkg -S `which dpkg`
    -> dpkg: /usr/bin/dpkg

### DIFF
--- a/pages/linux/dpkg.md
+++ b/pages/linux/dpkg.md
@@ -17,3 +17,7 @@
 - List package contents:
 
 `dpkg -L {{package_name}}`
+
+- Find out which package owns a file
+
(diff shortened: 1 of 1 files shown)


waldyrious:
LGTM! 👍

e0i:
I didn't know about this feature. Looks solid.

sethwoodworth:
@e0i It's not very well exposed as a feature IMO.  Over in fedora/redhat land has a `what-provides` sub-command.  After switching back and forth between distros, I missed the feature and had to find it.

PR #1166, opened 2016-11-25
Added adb push/pull commands

ente3000:


pages/common/adb.md
```diff
@@ -17,3 +17,15 @@
 - Push an Android application to an emulator/device:
 
 `adb install -r {{path/to/file.apk}}`
+
+- Copy a file/folder from the target device:
+
+`adb pull {{path/to/file.ext}} {{destination/path}}`
```
agnivade: If it can be a file or a folder, the style we follow is `{{path/to/file_or_folder}}`. Also, {{destination/path}} should be {{path/to/device}}, as its actually the source in this case. And we can use the same token for the next command also.

ente3000: `adb pull` is used as follows:
`adb pull (directory/file on a phone/emulator that is to be copied) (destination path on the device you're using the command with)`
Additionally, `{{path/to/device}}` is kinda confusing as users may think they can't copy their selection to a sub-folder.
However, I changed `{{path-to-file.ext}}` to `{{path/to/file_or_folder}}` and if you still want me to replace `{{destination/path}}`, I'll do that.

agnivade:
LGTM

waldyrious:
LGTM, thanks @ente3000!

lagelalegal:
Thanks @ente3000

PR #1170, opened 2016-11-28
nc: correct syntax for listen example

kuba-orlik:
On my machine the nc command requires the `-p` argument

### DIFF
--- a/pages/common/nc.md
+++ b/pages/common/nc.md
@@ -4,7 +4,7 @@
 
 - Listen on a specified port:
 
-`nc  -l {{port}}`
+`nc  -l -p {{port}}`
 
 - Connect to a certain port (you can then write to this port):
 


pages/common/nc.md
```diff
@@ -4,7 +4,7 @@
 
 - Listen on a specified port:
 
-`nc  -l {{port}}`
+`nc  -l -p {{port}}`
```
agnivade: Actually the man pages say something else. As far as I understand -

`-p` is used to specify the source port. And the port argument is the destination port. 

For use along side the `-l` flag, this is what the man page says -
```
-l      Used to specify that nc should listen for an incoming connection
             rather than initiate a connection to a remote host.  It is an
             error to use this option in conjunction with the -p, -s, or -z
             options.
```

Can you check again from your end why it does not work without the `-p` flag ?

kuba-orlik: Maybe it depends on the nc version? Mine is:

```
# nc --version
netcat (The GNU Netcat) 0.7.1
```

agnivade: It seems to be a platform specific difference. Suggest that you split the page into 2 versions. One in linux and one in osx.

kuba-orlik:
I wonder - should I just edit the text file to include both versions of syntax or is there a way to prepare separate entries for different operating systems?

waldyrious:
@kuba-orlik this kind of page split normally is done by moving the current version from `pages/common/` to `pages/linux/` or `pages/osx/` as appropriate, and creating a new page in the other platform's folder. However, in this case I'm not sure that's the issue. It seems to me that both you and @agnivade are using Linux, right? It might be just a versioning issue of the command itself. Could you guys verify that?

agnivade:
No, I guess @kuba-orlik is using osx. My netcat doesn't even have a `--version` option.

kuba-orlik:
I'm using Arch (manjaro).

agnivade:
Hey @kuba-orlik - so this seems to be an interesting problem. We usually have different variations for windows, linux, osx and sunos. But we don't handle variations inside a single platform itself. I am not sure how to go about this. Thoughts ?

agnivade:
Hey @kuba-orlik - it seems you are inactive. I am going to close this PR for now. Feel free to reopen this when you have some time again.

Thanks again for pointing out the issue. Much appreciated !

PR #1173, opened 2016-11-29
ftp: add page

ewels:


pages/common/ftp.md
```diff
@@ -0,0 +1,23 @@
+# ftp
+
+> Internet File Transfer Program
```
waldyrious: I'm pretty sure the P in FTP stands for Protocol.

pages/common/ftp.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Connect to a FTP server:
+
+`ftp {{url}}`
```
waldyrious: perhaps using an example url (ending in example.com) would be more illustrative.

pages/common/ftp.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Switch to binary transfer mode (graphics, compressed files etc):
+
+`binary`
+
+- Turn off prompts for each process:
```
waldyrious: What does "each process" refer to? Processes haven't been mentioned here at all until now.

pages/common/ftp.md
```diff
@@ -0,0 +1,23 @@
...
+
+`prompt off`
+
+- Download multiple files:
+
+`mget {{glob_expression}}`
```
waldyrious: Maybe move "glob expression" to the description, and use an actual example in the token.

pages/common/ftp.md
ewels: Yeah it does, but I guess the program subverts the acronym? I copied and pasted this from the FTP `man` command:
```
NAME
     ftp -- Internet file transfer program
```
But I agree that it's a bit confusing. I'll change this.

pages/common/ftp.md
ewels: For example, if you use `mput` to upload 100 files, it will ask you to confirm every file without this, one by one. _Transfer_ perhaps? Though it goes for stuff like deleting files too. _Interaction_ better maybe?

ewels: Ah, this also from the `man` page:
> `-i`  Turns off interactive prompting during multiple file transfers.

I'll copy that phrase instead.

pages/common/ftp.md
waldyrious: Hahah that's funny. But yeah I'm pretty sure it's better known as "protocol".

pages/common/ftp.md
```diff
@@ -0,0 +1,27 @@
+# ftp
+
+> Internet File Transfer Protocol
```
waldyrious: Ideally the description should explain what the command does, not just what it is. The expanded name in this case is pretty self-descriptive, but it would be nice to maintain the same pattern found in other commands.

pages/common/ftp.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Connect to a FTP server:
```
waldyrious: a --> an?

pages/common/ftp.md
```diff
@@ -0,0 +1,27 @@
...
+
+`ftp {{ftp.example.com}}`
+
+- Switch to binary transfer mode (graphics, compressed files etc):
```
waldyrious: please add a comma before, and a period after, the "etc".

pages/common/ftp.md
```diff
@@ -0,0 +1,27 @@
...
+
+`binary`
+
+- Turn off interactive prompting for multiple file transfers:
```
waldyrious: It would be clearer to explain what the prompt is for. One could make reasonable assumptions that this description doesn't clarify, e.g. it could be for confirmation or overwriting, and could activate only if glob expressions are used (not an explicit list of files), etc.

pages/common/ftp.md
ewels: I was about to disagree with you because F is not a vowel, but after some grammatical googling, I'll stand down.. 😉 👍

pages/common/ftp.md
```diff
@@ -0,0 +1,27 @@
...
+
+`mget {{*.png}}`
+
+- Upload multiple files (glob expression):
+
+`mput {{*.zip}}`
+
+- Delete multiple files on the remote machine (glob expression):
+
+`mdelete {{*.txt}}`
```
ewels: Yes, multiple. But it works on single files too, so I basically never use the non-m variants. I'm not really sure why they bother with separate commands to be honest. But fair enough, I'll switch this with it's singular version.

pages/common/ftp.md
ewels: Any suggestions for how to write this in a succinct manner? Here's the full `man` entry for `prompt`:

> `prompt`
>
> Toggle interactive prompting.  Interactive prompting occurs during multiple file transfers to allow the user to selectively retrieve or store files.  If prompting is turned off (default is on), any `mget` or `mput` will transfer all files, and any `mdelete` will delete all files.
> 
> When prompting is on, the following commands are available at a prompt:
> * `a`   Answer `yes` to the current file, and automatically answer `yes` to any remaining files for the current command.
> * `n`  Answer `no`, and do not transfer the file.
> * `p`  Answer `yes` to the current file, and turn off prompt mode (as is `prompt off` had been given).
> * `q`  Terminate the current operation.
> * `y`   Answer `yes`, and transfer the file.
> * `?`   Display a help message.
>
> Any other response will answer `yes` to the current file.

Perhaps something like _Transfer multiple files without interactive confirmations:_ or something?

ewels:
..the more I read the `man` page to answer your questions, the more commands I am reminded of which I think I should add! We now have `rename` as well as `delete` added as examples. There are a huge number of potential commands, but I think that this is probably a pretty complete list of the ones that I use routinely.

pages/common/ftp.md
agnivade: @waldyrious is our in-house grammar expert, if you haven't noticed. :wink: 

btw - Can someone explain the reason for the "an", for plebs like me ?

waldyrious: yes, that's it :)

pages/common/ftp.md
waldyrious: Yes, but confirmation of what? Just the transfer? That sounds weird, i.e. as if I'd say "copy files x y and z to foo", and the tool asked me for each of those "do you really want to copy x/y/z?".

I could understand for overwriting/deleting, and for commands using glob patterns, which could potentially catch more stuff than intended, but if it's indeed applied as a general rule, it strikes me as odd. I'm just trying to figure out which one is the case.

pages/common/ftp.md
waldyrious: We could not mention the single-file ones; the unintuitivess comes from the tool itself, so there isn't much we could do: either we'd recommend single-file commands that are not really useful, or we'd mention commands that contain an otherwise unexplained prefix for multiple. Let's just use the m versions, I suppose it's the lesser of the two evils.

pages/common/ftp.md
ewels: No, it is every transfer I think, as in your first example. That's why I always turn it off - it's really annoying!

pages/common/ftp.md
ewels: `ftp` contains quite a few evils ;) I'll revert this change then.. :+1:

pages/common/ftp.md
waldyrious: How about *Transfer multiple files without prompting for confirmation on every file:*? That way we're both more explicit, and we highlight the default behavior's silliness.

ewels: Sounds good 👍

waldyrious:
LGTM! :+1:

pages/common/ftp.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Upload multiple files (glob expression):
+
+`mput {{*.zip}}`
+
+- Delete multiple files on the remote server:
+
+`mdelete {{*.txt}}`
+
+- Rename a file on the remove server:
```
agnivade: typo -> "remove" :smile:

ewels: Oops, well spotted!

agnivade:
:shipit:

ewels:
Are you waiting for anything from me?

agnivade:
Nope, it good to go. But this lgtm system is screwing up again. Looks like @waldyrious has to again give a :+1: .

waldyrious:
It may be a problem with the commit having been added after my LGTM? Let's see: LGTM!

waldyrious:
That's weird, there seems to be a glitch in the LGTM recognition. No problem, we'll figure it out elsewhere. I'll override the check and merge. Thanks @ewels!

ewels:
Not sure how you guys are set up, but is this a new thing? https://lgtm.co/docs

> **LGTM Service Discontinued**
> _Please download the source code or Docker image to host the service on your own infrastructure._

Thanks for the merge!

waldyrious:
Oh, too bad :( it was a really nice idea. Oh well, time to wait until the next similar service to pop up...

PR #1191, opened 2016-12-05
sudo: use `shutdown -h` to halt rather than reboot

sourcefrog:
Makes the example consistent with the command


waldyrious:
Nice catch! This is a straightforward fix to an obvious error currently on the pages, so I'll merge right away without the two LGTMs from maintainers that we normally require.

PR #958, opened 2016-07-18
[RFC] new syntax for pages

waldyrious:
I'm opening this to serve as a discussion thread to decide whether we might want to simplify the pages' syntax.

The proposal is inspired on [Inkscape's man page](https://inkscape.org/en/doc/inkscape-man.html#examples) ([source code](https://github.com/inkscape/inkscape/blob/5e7291b/man/inkscape.pod.in#L426-L481)), and uses a format that I believe is more readable for humans and machines alike, for several reasons:
- the **main heading is changed from a `#` prefix to a row of `====` underneath**, which makes it stand out more in plain text mode;
- the current syntax involves bullet (unordered) lists of command descriptions with code samples --the actual commands-- _between_ them (rather than within each list item), which is both visually distracting in most markdown renderers, and semantically incorrect;
- **the bullet list syntax was changed to plain paragraphs**, which solves both the semantic issue, and a syntactic issue with some renderers not recognizing the code block if preceded by a list entry ([case in point](https://github.com/jedahan/tldr/blob/48e1286/pages/common/7za.md), from PR #758)
- **the example code is now indented**, so each description/code pair stands out better in the markdown source, and we can easily jump visually from example to example;
- the adding of indentation triggers markdown's code block formatting, so **now we don't have to wrap the examples in backticks** (which have been the source of many linting errors)
- by forgoing the backticks, the examples can be directly copied from the source markdown without the need to remove the backticks manually, which is more cumbersome than e.g. replacing tokens;
- since the markdown is more readable, clients don't need to be as sophisticated, opening the door to dumb "clients" like the one proposed in #527.

All in all, I believe this makes the pages be more inline with the markdown philosophy, which is to be human-readable, without the need for a renderer/client. Compare the plaintext source: [before](https://raw.githubusercontent.com/tldr-pages/tldr/master/pages/common/inkscape.md), [after](https://raw.githubusercontent.com/tldr-pages/tldr/waldyrious/alt-syntax/pages/common/inkscape.md); and the github-rendered output: [before](https://github.com/tldr-pages/tldr/blob/master/pages/common/inkscape.md), [after](https://github.com/tldr-pages/tldr/blob/waldyrious/alt-syntax/pages/common/inkscape.md).

[...]

PR #1197, opened 2016-12-15
Removing tee because it gobbles up the exit code

agnivade:


waldyrious:
I did wonder how changes like #958 stopped triggering Travis errors. For the curious, the regression was introduced [here](https://github.com/tldr-pages/tldr/commit/2a9060aef40d1912b19887a8d4b6951760b7a8c7#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) (not pointing fingers btw :) just adding context).

I don't have the technical expertise to assess the correctness of the fix, but I'll merge right away since the current pipeline is broken already; if this fix somehow introduces another issue, we'll deal with it accordingly.

agnivade:
Yea, it was a bit of oversight on my part. This last change of `tee` was suggested by @ostera because I was running `npm test` twice. I don't remember if I tested this last change or not. Anyways, I will raise another faulty PR just to test if everything works fine now or not.

waldyrious:
For those wondering about the verified badge in the commit: I messed up and added a typo when rewording the commit message, so I had to amend the commit and force-push. I had recently set up commit signing, so my signature was automatically applied to that amended commit.

PR #1196, opened 2016-12-14
git-cherry-pick: add page

alefir:
Added page for git cherry-pick

EDIT: Fixes #1155

PR #1198, opened 2016-12-15
git-cherry-pick: add examples to pick more commits

waldyrious:
See #1196 for context.


agnivade:
LGTM

waldyrious:
I'll wait for @alefir's input before merging :)

alefir:
I'd suggest changing 'commit_hash' to just 'commit' in the first example, since the '_hash' bit isn't used in any of the other examples. Besides that, looks fine to me.

waldyrious:
Good point, I'll change that. This is also consistent with the majority of our git pages, I believe.

PR #1201, opened 2016-12-15
git-tag: add --contains example

waldyrious:


pages/common/git-tag.md
```diff
@@ -22,3 +22,7 @@
 - Get updated tags from upstream:
 
 `git fetch --tags`
+
+- List tags that contain a given commit:
+
+`git tag --contains `
```
MarloweW: --contains should followed by {{commit_id}}.

waldyrious: Not sure how I missed this! THanks for the heads-up 👍

pages/common/git-tag.md
```diff
@@ -22,3 +22,7 @@
 - Get updated tags from upstream:
 
 `git fetch --tags`
+
+- List tags that contain a given commit:
```
waldyrious: IIUC this lists all tags whose ancestors include the given commit (also, I'm not sure that you can't add more than one tag to a commit). But I'll re-check.

waldyrious: Confirmed: http://stackoverflow.com/questions/36246615/how-does-git-tag-contains-work

As for the ability to add multiple tags to a commit, I found nothing suggesting such a restriction in neither https://git-scm.com/book/en/v2/Git-Basics-Tagging nor https://git-scm.com/docs/git-tag.

agnivade: >  this lists all tags whose ancestors include the given commit

We should include this then. Makes it clearer.

waldyrious: Sure :)

PR #1203, opened 2016-12-18
Change scp examples to match token syntax guidelines

alefir:


notpeter:
👍

waldyrious:
Thanks for the changes, @alefir :) I agree with pretty much all of them, but I'm unsure about swapping `{{~/.ssh/id_rsa}}` with `{{path/to/identity_file}}`. I understand the desire to generalize, but that may make the command harder to understand for those less familiar with these processes. I'd suggest either keeping `{{~/.ssh/id_rsa}}`, or at least starting the path with .ssh, e.g. `{{~/.ssh/identity_file}}` (or `{{~/.ssh/private_key}}`).

alefir:
I made that change because that is how it is referred to in the man page.
I've changed it because your suggestion makes more sense.

waldyrious:
Thanks. Looks good to me 👍. By the way, do you specifically prefer {{~/.ssh/identity_file}} to {{~/.ssh/private_key}}?

alefir:
Personally, I prefer identify_file because that's how it's referred to in the man page.

waldyrious:
Well, occasional user-unfriendliness of manpages is precisely why this project exists :) so we shouldn't bend over backwards to match their wording. I personally think private_key will be more intuitive for people, especially beginners, unless it's technically wrong. What do you think?

alefir:
That's a good point. I'll change it.

agnivade:
LGTM

Issue #1162, opened 2016-11-21
Add top CLI tools from "Who needs a GUI?" article


waldyrious:
@BryanLunduke's "[Who needs a GUI? How to live in a Linux terminal.](http://www.networkworld.com/article/3091139/linux/who-needs-a-gui-how-to-live-in-a-linux-terminal.html)" presents a list of the best CLI tools after attempting to go GUI-less for 30 days. These are probably good candidates for us to cover.

- [x] [mutt](https://en.wikipedia.org/wiki/Mutt_(email_client)) ([common](../blob/master/pages/common/mutt.md) / [linux](../blob/master/pages/linux/mutt.md))
- [x] [notmuch](https://notmuchmail.org/manpages/) ([common](../blob/master/pages/common/notmuch.md) / [linux](../blob/master/pages/linux/notmuch.md))
- [x] [alpine](https://www.washington.edu/alpine/mans.html) ([common](../blob/master/pages/common/alpine.md) / [linux](../blob/master/pages/linux/alpine.md))
- [x] [w3m](https://en.wikipedia.org/wiki/W3m) ([common](../blob/master/pages/common/w3m.md))
- [x] [nano](https://en.wikipedia.org/wiki/GNU_nano) ([common](../blob/master/pages/common/nano.md))
- [x] [wordgrinder](http://cowlark.com/wordgrinder/index.html) ([common](../blob/master/pages/common/wordgrinder.md) / [linux](../blob/master/pages/linux/wordgrinder.md))
- [x] [cmus](https://en.wikipedia.org/wiki/Cmus) ([common](../blob/master/pages/common/cmus.md) / [linux](../blob/master/pages/linux/cmus.md))
- [ ] [finch](https://developer.pidgin.im/wiki/Using%20Finch) ([common](../blob/master/pages/common/finch.md) / [linux](../blob/master/pages/linux/finch.md))
  - [Guide to Using Finch - Terminal-Based Chat Client](https://tuxarena.blogspot.com/2010/09/guide-to-using-finch-terminal-based.html)
  - [Finch - Howto use Pidgin via Terminal Console](https://coderstalk.blogspot.com/2008/09/finch-howto-use-pidgin-via-terminal.html)
- [x] [hangups](https://github.com/tdryer/hangups) ([common](../blob/master/pages/common/hangups.md) / [linux](../blob/master/pages/linux/hangups.md))
[...]

PR #1204, opened 2016-12-18
cmus: add page

alefir:
Addresses #1162 for cmus
Currently in pages/linux because I'm not sure if other operating systems have different commands.

### DIFF
--- a/pages/linux/cmus.md
+++ b/pages/linux/cmus.md
@@ -1,6 +1,6 @@
 # cmus
 
-> Commandline C* Music Player.
+> Commandline Music Player.
 > Use arrow keys to navigate, `<enter/return>` to select, and numbers 1-8 switch between different views.
 
 - Open cmus from specified directory:
--- a/pages/linux/cmus.md
+++ b/pages/linux/cmus.md
@@ -1,7 +1,7 @@
 # cmus
 
 > Commandline C* Music Player.
-> Arrow keys to navogate, `<enter/return>` to select, numbers 1-8 switch between different views.
+> Use arrow keys to navigate, `<enter/return>` to select, and numbers 1-8 switch between different views.
 
 - Open cmus from specified directory:
 
@@ -15,10 +15,10 @@
 
 `c`
 
-- Toggle shuffle:
+- Toggle shuffle mode on/off:
 
 `s`
 
-- Close cmus:
+- Quit cmus:
 
 `q`
--- a/pages/linux/cmus.md
+++ b/pages/linux/cmus.md
@@ -1,0 +1,24 @@
+# cmus
+
+> Commandline C* Music Player.
+> Arrow keys to navogate, `<enter/return>` to select, numbers 1-8 switch between different views.
+
+- Open cmus from specified directory:
+
+`cmus {{path/to/directory}}`
+
+- Add file/directory to library:
+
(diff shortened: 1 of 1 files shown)


pages/linux/cmus.md
```diff
@@ -0,0 +1,24 @@
+# cmus
+
+> Commandline C* Music Player.
+> Arrow keys to navogate, `<enter/return>` to select, numbers 1-8 switch between different views.
```
waldyrious: typo: "navogate". Also, please start the sentence with "Use...", and add "and" before "numbers".

pages/linux/cmus.md
```diff
@@ -0,0 +1,24 @@
+# cmus
+
+> Commandline C* Music Player.
```
waldyrious: C* is a somewhat obscure term IMO. Does it mean C/C++, or something else?

pages/linux/cmus.md
```diff
@@ -0,0 +1,24 @@
...
+
+`cmus {{path/to/directory}}`
+
+- Add file/directory to library:
+
+`:add {{/path/to/file_or_directory}}`
+
+- Pause/unpause current song:
+
+`c`
```
waldyrious: Does the "c" stand for anything in particular?

pages/linux/cmus.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Toggle shuffle:
+
+`s`
+
+- Close cmus:
```
waldyrious: Maybe "quit" instead of "close", so that the association with "q" can be more readily made.

pages/linux/cmus.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Add file/directory to library:
+
+`:add {{/path/to/file_or_directory}}`
+
+- Pause/unpause current song:
+
+`c`
+
+- Toggle shuffle:
```
waldyrious: This phrasing is slightly awkward. How about "Toggle shuffling of songs"?

pages/linux/cmus.md
alefir: I think it's there just because there are some similar commands in the same area of the keyboard (b is next song)

pages/linux/cmus.md
alefir: C* comes from the package description.
It was written in C and C++, and I think the * is being used as a regex-style wildcard.

waldyrious: I think we could just remove it. It's not really relevant to know what language a tool is written in, and here I'm afraid it just add noise.

pages/linux/cmus.md
waldyrious: Ok. It would be nice if we could come up with a nice mnemonic for it, even if it isn't the original meaning (if any). But that would be just a bonus :) it definitely won't hold up merging.

waldyrious: For example: "Cease or continue playing the current song" (I know, "cease" is kind of a stretch, but it's technically correct :P)

alefir: I cant think of anything that would make much sense.

waldyrious:
LGTM 👍

waldyrious:
Thanks for the contribution and the patience @alefir :)

PR #1206, opened 2016-12-20
less: open current file in an editor

willthemoor:


waldyrious:
Thanks for the contribution, @willthemoor :) I'm a bit concerned about the page size, though. The goal of the tldr-pages project is not to be exhaustive, but instead to cover the most common uses of a command, and give readers a good feel of how it works, so they can explore it further, e.g. by reading the manpage. so I'm not sure this additional option justifies the expansion to our maximum number of examples. What do you think?

willthemoor:
Totally understand (and appreciate!) that. I tend to use `v` a lot but get that others may not. My typical use case:

` $ less is-this-the-file-I-want-to-edit.html`

If yes, `v`. :)

In the spirit of keeping it brief, and get that this opinion is based on my version of the world, seems like more folks would get use out of knowing about `v` than `less -R` (ANSI color option).

fuerbringer:
@willthemoor According to the contribution guidelines the command descriptions should end in a colon `:`. Changing it to this: `- Open the current file in an editor:` would probably fix the Travis CI error :)

waldyrious:
Good point about `less -R`, albeit it could be argued that this is actually closer to the main task of `less`, which is to page files, than `-v`. I guess we ought to wait for more people to voice their opinion.

agnivade:
Yep, agree that `v` makes more sense to keep than `-R`.

agnivade:
Hey @willthemoor - I think according to the last discussion here, it was meant to remove the `-R` option and keep the `v`. Any updates on that ?

waldyrious:
One tear for the `-R` option 😢 you'll be missed! 😄

PR #1207, opened 2016-12-21
Reverting tldr bot reporting

agnivade:
Until we find a proper solution to this problem.


agnivade:
I will merge this myself as the current error reporting system is broken as of now.

PR #1208, opened 2016-12-21
curl: add --resolve

micahculpepper:


waldyrious:
Thanks for the contribution, @micahculpepper! Unfortunaltely, I must reiterate what I've [said in another recent PR](https://github.com/tldr-pages/tldr/pull/1206#issuecomment-268378416): tldr-pages entries, by definition, should aim to get the best balance between a page as short as possible and as informative as possible. Note "informative", not "comprehensive". In this vein, I am not really convinced that this additional example is common / illustrative enough to warrant bumping the number of examples all the way to the recommended maximum. Let me know what you think.

agnivade:
Agree with @waldyrious on this. 

I have had the need to do exactly this several times, I did that by either setting `/etc/hosts` entry, or doing it this way `curl http://<ip>/ -H "Host: www.example.com"`. Since the `-H` method is already shown, I am inclined to say this is not that important to be included.

micahculpepper:
@waldyrious I use the `--resolve` option quite frequently for work, and most of the time I check the curl man page, it's to look up this syntax. That's why I submitted it to tldr, in the hopes that it might help others. 

However, I see your point about not making the total entry too long. That's one of the core features of tldr, and part of the reason I installed it. If there's no desire from the community to add this, I can live with that.

Is it possible to add my own tldr modifications to, for example, a config file kept on my system? 

@agnivade `--resolve` is not the same as setting the host header. It is a relatively recent (Dec 2010) addition to curl, as it solves a particular problem. 

> when using HTTPS the TLS handshake happens before the server sees any HTTP headers
https://en.wikipedia.org/wiki/Server_Name_Indication#Background_of_the_problem

I use `--resolve` for testing HTTPS sites on servers or load balancers that host multiple domains, as it lets me manually set the hostname and IP of the request, ignoring DNS. It does effectively accomplish the same thing as setting an `/etc/hosts` entry, but it only does it for the scope of a single command, instead of affecting the whole system (which could impact the production environment).

agnivade:
I see. I think I used `-k` for those scenarios maybe, I don't remember exactly. Anyways, I still think this is going beyond the core set of curl functionalities.

And about adding adding local modifications, no, you can't do that yet. You can modify your local copy of curl.md. But then it will be overwritten in the next update. Another way is to create a fork and modify the `.tldrrc` to point to your fork. But then, you will have to frequently re-sync your fork.

waldyrious:
> Is it possible to add my own tldr modifications to, for example, a config file kept on my system?

I think @ostera has proposed that idea in the past, but I couldn't find a reference to it. The idea was to allow people to fetch tldr pages from alternative sources, but it hasn't been implemented so far. Maybe he can help pinpoint the past discussions, and then we can start a new issue to track the implementation of that functionality.

waldyrious:
> Another way is to create a fork and modify the `.tldrrc` to point to your fork.

Just to be clear: this refers to the node client, and the instructions can be found [here](https://github.com/tldr-pages/tldr-node-client#configuration).

waldyrious:
@micahculpepper thanks for taking the time to submit this PR. I'm sorry it didn't end up merged, but I trust that you understand why. Does the option to use a separate repository (in the node client) work for you? Let us know what can be done to improve that workflow.

PR #1211, opened 2016-12-25
gource: add page

fuerbringer:


pages/common/gource.md
```diff
@@ -0,0 +1,31 @@
+# gource
+
+> Visualize the history of Git, SVN, Mercurial and Bazaar repositories.
```
waldyrious: Perhaps mention that it produces a tree diagram of the project's directory structure, animated to show files and folders being created, modified or removed over time. Or some other description, hopefully more concise :)

pages/common/gource.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Run gource in a directory:
+
+`gource {{/path/to/repository/}}`
+
+- Run gource in fullscreen mode:
```
waldyrious: I'd mention that this assumes the command is run from the project's root directory.

pages/common/gource.md
```diff
@@ -0,0 +1,31 @@
+# gource
+
+> Visualize the history of Git, SVN, Mercurial and Bazaar repositories.
+
+- Run gource in a directory:
```
waldyrious: What happens if the directory isn't the root of the repository? Does the graph only show the portion of the project that's inside the given directory?

pages/common/gource.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Run gource in fullscreen mode:
+
+`gource -f`
+
+- Set the resolution:
+
+`gource -{{width}}x{{height}}`
+
+- Set the simulation time scale:
```
waldyrious: I'd use "animation", or "replay" rather than "simulation".

pages/common/gource.md
fuerbringer: I've tested this with various repositories including `tldr` and gource seems to jump back one level until it finds the `.git` files. So it does not display only a portion of the project when you enter a subdirectory.

Example: Running `gource` in `/tldr/` yields the same result as running it in `/tldr/pages/linux/`.

waldyrious: Ah, interesting. Could we reword the description to reflect this?

fuerbringer: Already made some changes earlier (because of other inline comments). The description got kinda long so I'm not sure how to reword it.

pages/common/gource.md
```diff
@@ -0,0 +1,32 @@
+# gource
+
+> Renders an animated tree diagram of Git, SVN, Mercurial and Bazaar repositories.
+> It shows files and folders being created, modified or removed over time.
+
+- Run gource in a directory:
+
+`gource {{/path/to/repository/}}`
```
agnivade: Please remove the ending slash

agnivade:
Very nice :) Just one minor comment.

pages/common/gource.md
```diff
@@ -0,0 +1,32 @@
...
+
+`gource -c {{time_scale_multiplier}}`
+
+- Set the amount of time to display each day:
+
+`gource -s {{seconds}}`
+
+- Set the background color:
+
+`gource -b {{hex_color_code}}`
```
waldyrious: Perhaps combine this with the -f option, to reduce the size of the page.

pages/common/gource.md
```diff
@@ -0,0 +1,32 @@
...
+> Renders an animated tree diagram of Git, SVN, Mercurial and Bazaar repositories.
+> It shows files and folders being created, modified or removed over time.
+
+- Run gource in a directory:
+
+`gource {{/path/to/repository/}}`
+
+- Run gource in fullscreen mode (assuming the command is run from the project's root directory):
+
+`gource -f`
```
waldyrious: Maybe split this into a plain `gource` invocation (placed above all other commands), with the remark in parenthesis, and a `gource -f` combined with the `gource -b` below.

pages/common/gource.md
```diff
@@ -0,0 +1,28 @@
+# gource
+
+> Renders an animated tree diagram of Git, SVN, Mercurial and Bazaar repositories.
+> It shows files and folders being created, modified or removed over time.
+
+- Run gource in a directory (assuming the command is run from the project's root directory):
+
+`gource {{path/to/repository}}`
+
+- Set the resolution:
```
waldyrious: This should mention that it assumes the current directory.

fuerbringer: So `(assuming the command is run from the project's root directory)` should be moved to this description?

waldyrious: I'd say it makes more sense in the first example; readers will likely assume that the same behavior carries over to the next ones.

But now that you mention it: the note in the command above should be reworded to match the command better. For example: "Run gource in a directory (if it isn't the repository's root directory, the root is seeked up from there)", or "Run gource in a directory (assuming it's the repository's root directory)", or something like that.

agnivade:
LGTM

PR #1212, opened 2016-12-27
cppcheck: add page

waldyrious:


waldyrious:
Note: this PR was created from the command line, using the `hub` command, thanks to the [hub.md](https://github.com/tldr-pages/tldr/blob/master/pages/common/hub.md) page :D

pages/common/cppcheck.md
```diff
@@ -0,0 +1,32 @@
...
+
+`cppcheck . 2> cppcheck.log`
+
+- Recursively check a given folder, and don't print progress messages:
+
+`cppcheck --quiet {{path/to/folder}}`
+
+- Check a given file, with all tests enabled (by default only errors are shown) except two:
+
+`cppcheck --enable=all --suppress={{check_id1}} {{path/to/file.cpp}}`
```
agnivade: Maybe its better to say "test_id1" because you have "tests" in the description ? 

Also, you say "except two", but you suppress just one check. Is this okay ?

pages/common/cppcheck.md
```diff
@@ -0,0 +1,32 @@
...
+
+`cppcheck --errorlist | grep "{{search pattern}}"`
+
+- Check a given file, ignoring specific tests:
+
+`cppcheck --suppress={{error_id1}} --suppress={{error_id2}} {{path/to/file.cpp}}`
+
+- Check the current folder, providing paths for include files located outside it (e.g. external libraries):
+
+`cppcheck -I {{include/folder_1}} -I {{include/folder_2"}} .`
```
agnivade: `"` intended ?

pages/common/cppcheck.md
waldyrious: Huh. I meant to add two, that's a mistake. And good point on the test_id1, I was using "checks" when I started working on this, as that's the term they use in their docuentation, and did a find/replace when I decided to switch to "test" to avoid ambiguity. Somehow I missed this one.

pages/common/cppcheck.md
waldyrious: Nope, I was clearly half-asleep when I wrote this xD

waldyrious:
@agnivade I removed the --suppress from the `--enable` example, in order to provide additional information (and since `--suppress` has a dedicated example, nothing is lost).

PR #1215, opened 2016-12-28
gource: tweak command descriptions

waldyrious:
Hopefully makes them clearer and more consistent.

### DIFF
--- a/pages/common/gource.md
+++ b/pages/common/gource.md
@@ -15,7 +15,7 @@
 
 `gource -c {{time_scale_multiplier}}`
 
-- Run gource specifying how long each day should be in the animation:
+- Run gource specifying how long each day should be in the animation (this combines with -c, if provided):
 
 `gource -s {{seconds}}`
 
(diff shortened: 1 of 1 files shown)


waldyrious:
I'm not really sure how `-c` and `-s` interact if used simultaneously. Any hints, @fuerbringer?

ps - I submitted these as a separate PR rather than request them in #1211 as that one had been nitpicked enough :)

waldyrious:
Thanks for the clarifications @fuerbringer :)

fuerbringer:
Looks great! I don't have anything to add. :+1:

agnivade:
I think the "Set" prefix was shorter and concise. Repeating "Run gource" sounds awkward. Maybe we can mention "Run gource" at the start, and then start the rest of the descriptions with "Set".

waldyrious:
I was aiming for consistency and maximum clarity (no parts of the description left implicit), but since you guys both seem to agree that starting with "Set" is sufficiently intuitive, I'll change.

PR #1216, opened 2016-12-29
pacman: add -Ql (List all files owned by a given package.) example

duesee:
Example output:

```
$ pacman -Ql hexedit
hexedit /usr/
hexedit /usr/bin/
hexedit /usr/bin/hexedit
hexedit /usr/share/
hexedit /usr/share/man/
hexedit /usr/share/man/man1/
hexedit /usr/share/man/man1/hexedit.1.gz
```

### DIFF
--- a/pages/linux/pacman.md
+++ b/pages/linux/pacman.md
@@ -30,6 +30,10 @@
 
 `pacman -Qo {{filename}}`
 
+- List all files owned by a given package:
+
+`pacman -Ql {{package_name}}`
+
 - Empty package cache to free up space:
 
 `pacman -Scc`


waldyrious:
Thanks for the PR, @duesee :) However, note that this change bumps the page over the limit (max 8 examples) as defined in CONTRIBUTING.md (after all, these are meant to be TL;DR pages, so being concise is the the primary characteristic we ought to preserve).

We can still replace one (or more) of the existing commands if this one seems more common or more useful to understand the command. Let me know what you'd suggest.

duesee:
Oh, didn't knew about the limit. I reviewed all other commands and must admit: they are better suited :-) So this can be closed then.

waldyrious:
Thanks for understanding and for the review :)

PR #1219, opened 2016-12-29
pacman: fix confusing example

duesee:


pages/linux/pacman.md
```diff
@@ -14,9 +14,9 @@
 
 `pacman -Rs {{package_name}}`
 
-- Search the package database for a keyword:
+- Search the package database for a regular expression or keyword:
 
-`pacman -Ss icon theme`
+`pacman -Ss {{regexp}}`
```
waldyrious: Maybe wrap the token in quote marks? or it isn't needed? Also, perhaps {{search_pattern}} would be better, since regular expressions are already mentioned in the description.

duesee: Quotation marks aren't needed (tested it right now), but wouldn't hurt either (being more future-proof?). Should we prioritize simplicity?

I kept the term "keyword" to handle simple examples, so "search_pattern" makes more sense, I agree :-)

waldyrious: I think quote marks are useful not only for future-proofness, but also because it signals explicitly to the reader that any string can be inserted here (including spaces).

PR #1223, opened 2017-01-02
dpkg: Fix travis errors

agnivade:


agnivade:
This happened because of merging #1165. At that time, the travis check was broken because of #1207.

PR #1224, opened 2017-01-03
alias: change main description; add view example

waldyrious:


pages/common/alias.md
```diff
@@ -1,12 +1,16 @@
 # alias
 
-> Creates an alias for a word when used as the first word of a command.
-> Note that these operations are not permanent, expiring when the current shell session ends.
+> Registers words that are replaced by a command string (which can comprise multiple commands, shell functions, etc).
```
agnivade: I would prefer to keep the word alias somewhere. Maybe something like "Creates an alias for a word that are replaced by a command string (which can comprise multiple commands, shell functions, etc)."

waldyrious: Well, the word *is* the alias, so it would have to be something like `Creates aliases -- words that are replaced by a command string (...)`. Maybe I'd even remove the (...) part, to make the sentence shorter. What do you think?

agnivade: sounds good

waldyrious: Ok to both suggestions (the rewording and the shortening)?

agnivade: Ok to rewording. Shortening is upto you, I am fine either way.

waldyrious: I think we should defer to the tldr spirit whenever possible, so I shortened the desc. PTAL :)

PR #1227, opened 2017-01-04
git-bisect: add page

waldyrious:


pages/common/git-bisect.md
```diff
@@ -0,0 +1,20 @@
...
+
+- For each commit that `git bisect` selects, mark it as "bad" or "good":
+
+`git bisect {{good|bad}}`
+
+- After `git bisect` pinpoints the faulty commit, end the bisect session and return to the previous branch:
+
+`git bisect reset`
+
+- Skip a commit during a bisect (e.g. one that causes a broken build due to another issue):
```
agnivade: Did not get this. Why would you need to skip a commit ? If its unrelated to the current bug, we can just mark it as good right ?

pages/common/git-bisect.md
```diff
@@ -0,0 +1,20 @@
+# git bisect
+
+> Use binary search to find the commit that introduced a bug.
+> Git automatically jumps back and forth in the commit graph to progressively narrow down the faulty commit.
+
+- Start a bisect section by passing a known good commit and a known bad commit:
```
agnivade: I was slightly stumped in understanding "known bad commit". How can I know the bad commit if I have to find it in the first place ? Then it dawned on me, that its a commit sufficiently back in time so that the actual commit which introduced the bug is in the range. 

Maybe we can rephrase it as - "Start a bisect section by passing a start and end commit which contains the commit we want to find".

pages/common/git-bisect.md
waldyrious: If a commit is broken for a different reason, you may not be able to tell whether the issue you're looking for is present in it or not -- the idea here is to tell git bisect to ignore this commit rather than pass it as a data point, as that could be a false positive or a false negative, both of which would pollute the binary search process.

pages/common/git-bisect.md
waldyrious: I think that makes sense. Will change.

ps - I just realized I meant to write "Start a bisect session", not "section"! Will fix that as well.

waldyrious:
@agnivade I went through a lot of back-and-forth attempting to produce concise yet clear descriptions. I'm not sure I managed it; PTAL.

PR #1229, opened 2017-01-04
brew.md: add Linuxbrew version

waldyrious:
See https://github.com/Linuxbrew/homebrew-core/pull/1472 for context.

### DIFF
--- a/pages/linux/brew.md
+++ b/pages/linux/brew.md
@@ -1,0 +1,27 @@
+# brew
+
+> The Homebrew package manager for Linux.
+
+- Search formula:
+
+`brew search {{text}}`
+
+- Install formula:
+
+`brew install {{formula}}`
+
+- List installed formulae [with matching name]:
+
+`brew list {{[text]}}`
+
+- Get latest version of installed formula (passing no formula updates all installed formulae):
+
+`brew upgrade {{[formula]}}`
+
+- Update brew:
+
+`brew update`
+
+- Switch version of formula:
+
+`brew switch {{formula}} {{version}}`


MarloweW:
Maybe we could add `brew info -f {{formula}}`, it's useful to find out files contained in a formula.

waldyrious:
@MarloweW this would be applicable to the osx version as well, right? If so, I'd rather do that in a separate PR.

MarloweW:
@waldyrious  I must have misremember it. 

`brew info {{formula}}` will just print description of formula and which directory current version of this formula is installed, `-f` has no effect on this command. To list all files provided by a formula, we should use `brew ls {{--verbose|-v}} {{formula}}`. And yes, they both are applicable to osx version and Linux version, checked source code.

A separate PR would be better.

pages/linux/brew.md
```diff
@@ -0,0 +1,27 @@
...
+
+`brew search {{text}}`
+
+- Install formula:
+
+`brew install {{formula}}`
+
+- List installed formulae [with matching name]:
+
+`brew list {{[text]}}`
```
agnivade: umm .. why the square brackets ?

waldyrious: This page is identical to osx/brew.md, except for the description. I'd rather make any changes in another PR. As for this particular notation, I suppose it means that the parameter is optional; without, the full list would be printed. But I agree it isn't super-intuitive. Let's change it in a separate PR.

agnivade: ok

PR #1232, opened 2017-01-05
git-worktree.md: add page

becomesaflame:


pages/common/git-worktree.md
```diff
@@ -0,0 +1,15 @@
+# git worktree
+
+> Manage multiple working trees attached to the same repository.
+
+- Create a new folder with the specified branch checked out into it:
+
+`git worktree add {{path}} {{branch}}`
+
+- List all the working directories attached to this repo:
```
waldyrious: I'd rather use "repository", as we don't gain much by abbreviating here.

waldyrious:
Nice one @becomesaflame :) Looks good to me, but please take a look at the inline comments.

becomesaflame:
Good suggestions.  I've updated it

pages/common/git-worktree.md
```diff
@@ -0,0 +1,15 @@
+# git worktree
+
+> Manage multiple working trees attached to the same repository.
+
+- Create a new folder with the specified branch checked out into it:
+
+`git worktree add {{path/to/folder}} {{branch}}`
```
agnivade: Please fix the inconsistency of options order in your 1st and 3rd examples. 

First one has `add {{folder}} {{branch}}` 
Third has `add -b {{branch}} {{folder}}`

Also, please reorder the examples to have the `add` commands first and then the `list` command.

becomesaflame: Is it better to have them grouped by type of command, or listed in order of most commonly used to least commonly used?

becomesaflame: Adding prune option is a good idea.  What are your thoughts on: 
```
- Remove a worktree: 

`rm -rf {path/to/folder}} && git worktree prune`
```

vs

```
- Remove a worktree (after deleting worktree folder): 

`git worktree prune`
```

agnivade: > Is it better to have them grouped by type of command, or listed in order of most commonly used to least commonly used?

The options should be ordered by the workflow. Usually we cite the most simple ones first, then slowly add complexity, building up on previous examples. In this case, we should put `add` first, then `list` and `prune` at last.

> What are your thoughts on:

I like the 2nd one.

waldyrious: > In this case, we should put `add` first, then `list` and `prune` at last.

I think he was asking whether `add -b` should be grouped with plain `add`, or moved down for being a less common example. I think this call is often subjective so there isn't a single answer, but in this case I'd say it makes more sense for both `add` examples to be together.

waldyrious:
Not sure what's the problem with the CLA status not being reported. Have you signed it, @becomesaflame?

becomesaflame:
Yes, and the check was showing up before as complete. I'm not sure what's going on with it.

agnivade:
Ok then, @waldyrious, I don't see a problem with force merging it.

PR #1236, opened 2017-01-08
git-push: revamp page

waldyrious:
Hopefully making it more useful and explicit.

Added a plain `git push` example, a `git push --all`, updated the remote branch deletion example, and reworded the descriptions.

### DIFF
--- a/pages/common/git-push.md
+++ b/pages/common/git-push.md
@@ -1,23 +1,31 @@
 # git push
 
-> Push commits to a remote repository.
+> Push commits to a remote (upstream) repository.
 
-- Publish local changes on a remote branch:
+- Send local changes in the current branch to its remote counterpart:
+
+`git push`
+
+- Send local changes in a given branch to its remote counterpart:
 
 `git push {{remote_name}} {{local_branch}}`
 
-- Publish local changes on a remote branch of different name:
+- Publish the current branch to a remote repository, setting the upstream branch name:
 
-`git push {{remote_name}} {{local_branch}}:{{remote_branch}}`
+`git push {{remote_name}} -u {{remote_branch}}`
 
-- Remove remote branch:
+- Send changes on all local branches to their remote counterparts:
 
-`git push {{remote_name}} :{{remote_branch}}`
+`git push --all`
 
-- Remove remote branches which don't exist locally:
+- Delete a branch in a remote repository:
+
+`git push {{remote_name}} --delete {{remote_branch}}`
+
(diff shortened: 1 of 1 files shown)


waldyrious:
added a plain `git push` example, a `git push --all`, updated the remote branch deletion example, and reworded the descriptions.

Dahie:
Signed, thank you!

pages/common/git-push.md
```diff
@@ -14,15 +14,15 @@
...
 
-`git push --all`
+`git push --all {{remote_name}}`
 
 - Delete a branch in a remote repository:
 
 `git push {{remote_name}} --delete {{remote_branch}}`
 
-- Remove local references to branches that have been deleted in a remote repository:
+- Remove remote branches that don't have a local counterpart:
```
waldyrious: Hi @liuderchi, I am not sure I understand what you're saying. What would be an example of "remote branches that don't have a counterpart in remote repository"?

liuderchi: I'm sorry I did not realize it's `push --prune` rather than `pull --prune`. 
I was misled by the previous description.

waldyrious: Yeah, that was the same mistake I made xD I think the problem was also that you looked at the commits individually rather than at the aggregate diff, which might have saved you the confusion ;)

pages/common/git-push.md
```diff
@@ -1,23 +1,31 @@
 # git push
 
-> Push commits to a remote repository.
+> Push commits to a remote (upstream) repository.
```
agnivade: Just wondering, is there is use of adding "upstream" here ? Doesn't remote always mean upstream ?

liuderchi: > Doesn't remote always mean upstream ?

No. `origin` is also a remote repository which is commonly used.

waldyrious: "upstream" wasn't used here as the concrete name of a remote (we could call them "origin", "upstream", "remote" or anything else if we wanted), but rather as a commonly used synonym -- so @agnivade's concern is valid.

My addition of the term here is because it is often used in that sense (including in several tldr pages). But we probably should avoid promoting that equivalence, since the terms "upstream" and "downstream" have specific meanings in the context of software dependency, which don't necessarily map with the meanings of "local" and "remote" in the context of computer networks in general (and distributed networks of git repositories in particular).

If you agree, I'll make the change.

agnivade: I agree.

PR #1238, opened 2017-01-09
git-checkout: also works for remote branches

waldyrious:
Not sure if this clarifies or obscures the example. Opinions welcome.

### DIFF
--- a/pages/common/git-checkout.md
+++ b/pages/common/git-checkout.md
@@ -6,7 +6,7 @@
 
 `git checkout -b {{branch_name}}`
 
-- Switch to an existing branch:
+- Switch to a branch that already exists locally or upstream:
 
 `git checkout {{branch_name}}`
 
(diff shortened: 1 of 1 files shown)


bobstrecansky:
Maybe change this to:
> Switch  to an existing or upstream branch

Thoughts?

agnivade:
I prefer @bobstrecansky's suggestion.

waldyrious:
Hmmm, not sure: an upstream branch already exists as well, just not locally. The distinction is not the existence, but the locality. How about this?

> Switch to an existing local or remote branch:

(Note that I'm using "remote" instead of "upstream" following the rationale described [here](https://github.com/tldr-pages/tldr/pull/1236#discussion_r96134742).

agnivade:
Yea makes sense. Go ahead.

waldyrious:
I'll just wait to see if @bobstrecansky has any thoughts about this :)

bobstrecansky:
Sounds good to me!

PR #1240, opened 2017-01-11
Update curl.md

jcherqui:
Add -I parameter to show headers informations.

### DIFF
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -31,6 +31,6 @@
 
 `curl -v -key {{key.pem}} -cacert {{ca.pem}} -cert {{client.pem}} -k {{https://example.com}}`
 
-- Show headers info only
+- Show headers info only:
 
 `curl -I {{http://example.com/}}`
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -30,3 +30,7 @@
 - Pass client certificate and key for a secure resource:
 
 `curl -v -key {{key.pem}} -cacert {{ca.pem}} -cert {{client.pem}} -k {{https://example.com}}`
+
+- Show headers info only
+
+`curl -I {{http://example.com/}}`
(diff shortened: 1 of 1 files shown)


waldyrious:
I'm inclined not to expand pages up to the maximum limit unless it's absolutely necessary. Is this such a common use case? And if so, could it replace one of the existing examples that may be less common/useful?

agnivade:
Its slightly common, but the `-v` already shows the request and response headers. So I believe we are covered. Actually we should improve the last example, and add the description that "it shows the request and response headers too".

waldyrious:
@agnivade that sounds good. What do you think, @jcherqui?

bobstrecansky:
From my cURL experience, i'd say they are both pertinent.  Many people use the -I flag to receive the headers, but don't actually know what it does.  the -v output can be time consuming to read if you only want the headers.  another option here is to use the -D or --dump-header flags, which will dump the headers to a file.  We could append the 
> Download a file saving the output under the filename indicated by the URL:
directive to include this if we want to save a little space.

jcherqui:
Ok, I prefer merge with `-u` example than the `-v`. It's more clearly. Good for you ?

agnivade:
Also, please add the verbose part to the last description.

waldyrious:
@agnivade I don't think we should require that change from @jcherqui, since it is not in the scope of this PR. I'd suggest that we merge this PR and make that change afterwards.

@bobstrecansky thanks for your input. I do agree with @jcherqui that merging with -u makes more sense, though.

agnivade:
Ok sure. It will be great if you can make that change before merging.

waldyrious:
I was about to make the change but there are a few issues: (1) the full-word options require double dashes; (2) the relationship between `--key`, `--cert` and `--cacert` is unclear to me (i.e. how the latter two differ, which depend on which, which are meant to be used together, etc.) and (3) the -v is not necessarily related to that example, so by including it (which requires mentioning it in the description, as you pointed out) makes it hard to make the description concise. I'll open a separate issue to track these.

PR #1241, opened 2017-01-11
dig.md: Add nssearch option

bobstrecansky:
Adding the nssearch command to show all authoritative name servers.


bobstrecansky:
My apologies on the CLAassistant only showing one of two users.  Was accidentially logged into my GHE account as well.

waldyrious:
@bobstrecansky no problem :) it seems to be all fixed now. As for this change, it looks good to me -- thanks for the submission!

PR #1242, opened 2017-01-11
sox.md: add volume option

jeeftor:
How to change volume of a wav file

### DIFF
--- a/pages/common/sox.md
+++ b/pages/common/sox.md
@@ -22,3 +22,7 @@
 - Print statistical data of an audio file:
 
 `sox {{input_audiofile}} -n stat`
+
+- Increase the volume of an audio file by 2x
+
+`sox -v 2.0 {{input_audiofile}} {{output_audiofile}}`
--- a/pages/common/sox.md
+++ b/pages/common/sox.md
@@ -23,6 +23,6 @@
 
 `sox {{input_audiofile}} -n stat`
 
-- Increase the volume of an audio file by 2x
+- Increase the volume of an audio file by 2x:
 
 `sox -v 2.0 {{input_audiofile}} {{output_audiofile}}`


jeeftor:
Signed

waldyrious:
Hi @jeeftor, thanks for the PR! I'm wondering, how does this differ from the `--norm` example, and why would one want to use this specifically, considering that it might clip the audio (IIUC)?

jeeftor:
@waldyrious it does clip the audio - however we had a use case where that was a desired effect - as we needed to keep some relative sounds together and not normalize the entire file.

waldyrious:
But is that such a common use case? Would it be justified to add it considering that we already have `--norm`? Remember that tldr-pages aim to provide a brief summary/overview, not cover the whole functionality set of the tool.

jeeftor:
Its something we do often where i work - its up to you guys whether you
want to pull it out.

waldyrious:
I would be inclined to leave it out, unless persuaded otherwise. Let's wait for others to have their say.

agnivade:
I am neutral on this as I have never used the command. I will defer to whatever the maintainers agree to.

Dahie:
I'd see it like this. tldr, not only provides the use-cases required most often, but also a quick overview of the functionality a tool might provide, by giving use case examples. It does not give an example for every feature a tool can do, but give an idea of some basics to know if its worth checking out the full manpages to learn about the full feature set. 
eg this new addition would tell me, there is a range of features for dealing with volume editing, what additional volume feature are out of scope here.

Based on this you could argue, nah, we have -norm and don't need another volume example or you look if there is another feature eg limiting, compression, etc that is worth including with a similar simple example.

waldyrious:
That's a good point, @Dahie. @ostera what do you think?

jeeftor:
Honestly sox is an audio processing tool so I would think   Volume manipulation is a pretty standard option which is why I included it. Any decision y'all make is cool however.

waldyrious:
@jeeftor that does make sense, but since the main focus of this project is conciseness, I'd rather not expand pages beyond the recommended size (5-6 examples) unless the benefits do justify it, e.g. when a new type of functionality is covered.

Having said that, the fact is this page has 5 examples at the moment, and this PR would add a sixth one -- well within our recommended limit. So I will agree to merging in this case, but just as a heads-up, we might consider replacing this example if someone proposes a new addition to this page in the future. You'll be pinged if that ever happens, though, so you'll get to have your say :)

PR #1243, opened 2017-01-11
brew.md: Improve explanation 

### LINKED ISSUE #1235: Improve brew.md
From [this comment](https://github.com/tldr-pages/tldr/pull/1229#issuecomment-270600770) by @MarloweW:

> Maybe we could add `brew info -f {{formula}}`, it's useful to find out files contained in a formula.

We should also revise the `list` example, as discussed [here](https://github.com/tldr-pages/tldr/pull/1229#discussion_r94899548), to get rid of the square brackets, which can be confusing. Ideally that could be done by rewording the description rather than splitting the example in two, since we're already at 6 examples and with the `info -f` we'd reach 7.

MarloweW:
Fixes #1235.

### DIFF
--- a/pages/linux/brew.md
+++ b/pages/linux/brew.md
@@ -10,9 +10,9 @@
 
 `brew install {{formula}}`
 
-- List installed formulae, argument `formula` is optional:
+- List all installed formulae:
 
-`brew list {{formula}}`
+`brew list`
 
 - Get latest version of installed formula (passing no formula updates all installed formulae):
 
--- a/pages/linux/brew.md
+++ b/pages/linux/brew.md
@@ -10,9 +10,9 @@
 
 `brew install {{formula}}`
 
-- List installed files for formulae or list all installed formulae if no formula specified, argument --verbose is to recursively list files of all subdirectories.:
+- List installed formulae, argument `formula` is optional:
 
-`brew list --verbose {{formula}}`
+`brew list {{formula}}`
 
 - Get latest version of installed formula (passing no formula updates all installed formulae):
 
--- a/pages/linux/brew.md
+++ b/pages/linux/brew.md
@@ -10,13 +10,13 @@
 
 `brew install {{formula}}`
 
-- List installed formulae [with matching name]:
+- List installed files for formulae, argument --verbose is to recursively list files of all subdirectories, it's optional:
 
-`brew list {{[text]}}`
+`brew list --verbose {{formula}}`
 
 - Get latest version of installed formula (passing no formula updates all installed formulae):
 
-`brew upgrade {{[formula]}}`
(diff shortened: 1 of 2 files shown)


MarloweW:
@waldyrious take a look at this please, let me know if something need reword.

waldyrious:
The transition "List installed formulae" --> "List installed files for formulae" is confusing me a little. Are these two descriptions meant to be equivalent? I interpret them to mean different things...

MarloweW:
I made some change to brew list description, please check it again.

waldyrious:
@MarloweW hmm, if I'm reading this correctly, the list command has two different behaviors if a formula name is provided or not. We normally try to avoid coupling two usage examples in one like this, so either we should choose one (I believe the one without parameters will be more commonly useful), or split them in two. My inclination is to do the former, but I'm happy to hear arguments otherwise.

MarloweW:
@waldyrious Yes, you're right about the different behaviors thing. I think it's a good idea avoiding multiple usage examples in one command description. `list` without parameters is indeed more commonly, I saw many questions about how to find installed files of a formula, think mention it in description may be useful too, but since it doesn't agree with guideline of this project, I prefer your former choice too.

pages/linux/brew.md
```diff
@@ -10,13 +10,13 @@
 
 `brew install {{formula}}`
 
-- List installed formulae [with matching name]:
+- List installed formulae, argument `formula` is optional:
 
-`brew list {{[text]}}`
+`brew list {{formula}}`
```
waldyrious: Saying that the argument is optional doesn't handle the fact that its inclusion makes this command do something different -- so these are still two commands in one.

Please remove the parameter from this example altogether. If you think it makes sense to include it (and that it justifies bumping the page to the maximum size), add a new example with `brew list {{formula}}` and the appropriate description.

agnivade: Or you can follow the structure of the below example and say "(passing no formula lists all installed formulae)" at the end. I think that's okay too.

waldyrious: I don't think the example below is equivalent, since both options do the same operation, but ok, your phrasing sounds reasonable and makes the page self-consistent.

MarloweW: I think I was trying to cover every detail in this command, which more or less went against guideline 1&3: 
> it's OK if the page doesn't cover everything
> The goal is to allow people to understand the syntax of the commands

Now, I prefer we just keep `brew list`, and nothing more about this command.

agnivade: Sure, I'm good with that too. :+1:

MarloweW: Updated.

PR #1244, opened 2017-01-12
tar.md: Add -C option

agnivade:
I use this option very frequently while installing a new version of node or golang, where the binary distribution is a tarball.

### DIFF
--- a/pages/common/tar.md
+++ b/pages/common/tar.md
@@ -30,3 +30,7 @@
 - List the contents of a tar file:
 
 `tar tvf {{source.tar}}`
+
+- Change to a different target directory, removing one leading component from file names. Useful in installing archived binary distributions:
+
+`tar -C {{/usr/local}} --strip-components {{1}} -xJf {{source}}.tar.xz`


waldyrious:
I think we should require a higher bar for inclusion of the 7th and especially 8th example in a page to justify stretching the page beyond the recommended limits. As in, a justification of (1) why the extra example is something that should not be left out even for a brief overview/summary of the command as tldr aims to provide, and (2) why none of the existing examples could be replaced or combined to avoid making the page too big.

If you agree, we could start here and I'll add this to CONTRIBUTING.md.

agnivade:
Hmm .. I understand your concern. And I am totally for keeping the no. examples to the absolute necessary. But I think adding more rules in CONTRIBUTING.md is going to complicate things further and might act as a repelling factor for new contributors.

I would like to keep it as ad-hoc. i.e. - If any maintainer thinks that the addition is unwarranted, he/she might ask the contributor for a reason as to why this is *absolutely* necessary. If not, then it should be fine. I would also love to hear @ostera's take on this.

As for this particular case, all the previous examples concern themselves with extracting, creating and listing archives. This example cites a practical usage of `tar`, which covers a common use case of installing binary distributions. Therefore, from the practical side of things, I think we should add it. Am open to other opinions too.

waldyrious:
Could you explain the parts of this command in more detail? I tried [explainshell](http://www.explainshell.com/explain?cmd=tar+-C+%7B%7B%2Fusr%2Flocal%7D%7D+--strip-components+%7B%7B1%7D%7D+-xJf+%7B%7Bsource%7D%7D.tar.xz) but that wasn't entirely elucidating.

agnivade:
Sure. 

I think `-C` is simple enough, it just changes the destination directory where the archive will be extracted. 
`-xJf` stands for extracting a `.xz` archive. 
Now comes the interesting `--strip-components` part. This strips of "n" leading parts from the filename while extracting the file at the destination location. To explain with an example, look here - http://serverfault.com/a/330131

If you think `--strip-components` is too much for a tldr example, we can remove it. But, usually `-C` and `--strip-components` are used together.

waldyrious:
I have a couple questions:
- is -xJf is equivalent to the xzf example earlier on?
- considering that we already have a -C and a xzf example, could it be more helpful to explain `--strip-components` separately, rather than attempting to provide a real-world usage which makes it harder to explain?
  Another option would be to introduce the `xzf`, `-C`, and `xzf/-C/--strip-components` examples in sequence, the latter building on the previous example's description to avoid re-describing that option.

agnivade:
Oh man, I think I was blind.  I didn't even see that `-C` was already there. You know what, I don't think this example is needed anymore. To just add `strip-components`, its not worth adding an 8th example. If you agree, just close the PR.

waldyrious:
Agreed :+1:

PR #1245, opened 2017-01-13
logstash: add page

abhijit-paul:


waldyrious:
LGTM, but please sign the CLA :)

pages/common/logstash.md
```diff
@@ -0,0 +1,12 @@
+# logstash
+
+> Migrate data from one source into another.
```
agnivade: "Migrate" might be a slightly misleading term for it. How about "Tool to consume data from various sources, transform it, and send it to various outputs." ?

pages/common/logstash.md
```diff
@@ -0,0 +1,12 @@
...
+> Usually used to ingest data from various other data sources into elasticsearch.
+
+- Check validity of a logstash configuration:
```
agnivade: I was going through logstash documentation. I think adding the `-e` example at first will help out beginners. Because the structure and syntax of the config file is not mentioned, and putting up something like - 
```
logstash -e 'input { stdin { } } output { stdout {} }'
```
will ease the users slowly into getting started with logstash.

pages/common/logstash.md
```diff
@@ -0,0 +1,12 @@
...
+
+`logstash --configtest --config logstash-config.conf`
+
+- Run logstash using configuration:
+
+`sudo logstash --config logstash-config.conf`
```
agnivade: No need to mention `sudo`.

pages/common/logstash.md
waldyrious: I happen to think "migrate" provides a nice concise description of that process. The only downside is implying a move rather than a copy. Another problem is that "another" implies "another source", but in fact it supposedly means "destination" or "target". A slight rephrase may be in order for these reasons, but I'd love to keep the original proposal's conciseness when doing so.

pages/common/logstash.md
waldyrious: Is sudo not needed in the most common use case? If it is, then we *should* include it, per [guideline #2](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines)

agnivade: good point.

pages/common/logstash.md
agnivade: "Migrate" omits a major part of logstash, which is intermediate data processing/filtering. Also, the implication of move rather than copy is pretty important I think.

pages/common/logstash.md
abhijit-paul: Sudo is needed. So, will go with the [guideline #2](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).

pages/common/logstash.md
abhijit-paul: Thre are two schools of thought regarding the "Migrate" word:
Those implying copy in tech:
* https://discussions.apple.com/thread/4475182?tstart=0
* https://www.techopedia.com/definition/1180/data-migration.

But, still the inherent meaning of migrate is to transfer:
* http://www.thesaurus.com/browse/migrate

Hence, we should avoid the confusion. 
I also agree with the ambiguity of "one source into another".
Rephrase: 
'**Replicate textual information between different kinds of databases**'

agnivade: I want to keep the part that we can transform the information too. Any way you can think of including that part ?

abhijit-paul: Essentially being an ETL tool, going with that avoids all these complexities we are having with the language. I should have thought about it before!

pages/common/logstash.md
```diff
@@ -0,0 +1,21 @@
+# logstash
+
+> An ETL tool, usually used to ingest data from various sources, like databases and flat log files into elasticsearch.
```
agnivade: Please expand the acronym within brackets. So that newbies don't have a hard time :)

pages/common/logstash.md
```diff
@@ -0,0 +1,21 @@
...
+
+- Boilerplate configuration for logstash to provide logic for extract, transform and load, respectively:
+
+`input {}`
+`filter {}`
+`output {}`
+
+- Check validity of a logstash configuration:
+
+`logstash --configtest --config {{logstash-config.conf}}`
```
agnivade: Change {{logstash-config}} -> {{logstash_config}}

pages/common/logstash.md
```diff
@@ -0,0 +1,21 @@
...
+
+- Run logstash using configuration:
+
+`sudo logstash --config {{logstash-config.conf}}`
```
agnivade: Change {{logstash-config}} -> {{logstash_config}}

pages/common/logstash.md
```diff
@@ -0,0 +1,21 @@
...
+
+- Run logstash with inline configuration string:
```
agnivade: Suggest changing it to "Run logstash with the most basic inline configuration string"

pages/common/logstash.md
```diff
@@ -0,0 +1,21 @@
+# logstash
+
+> An ETL tool, usually used to ingest data from various sources, like databases and flat log files into elasticsearch.
+
+- Boilerplate configuration for logstash to provide logic for extract, transform and load, respectively:
```
agnivade: We usually give command descriptions along with their examples here. This seems to be the configuration for logstash. Since, you are already mentioning the same thing in the last example, I think this is not needed anymore. 

Also, removing this would solve #1263

waldyrious:
Thanks for the contribution and patience, @abhijit-paul!

Issue #1082, opened 2016-09-21
Add more fields to the commands


waldyrious:
e.g. URL (mentioned in [#934](https://github.com/tldr-pages/tldr/pull/934#issuecomment-229016985)), see also (tracked in #481), maybe platform (related to #190)?

We will need to decide details in dedicated issues, but overall discussion about a potential new page format (including markdown syntax, see #784 and #958) should happen in a coordinated way, hence this issue.

waldyrious:
Another field proposal: original manpage (#1248)

waldyrious:
The first one (URL) is now properly tracked in #2649.

zdroid:
Isn't this all already implemented @waldyrious?

waldyrious:
@zdroid how so? IIUC only `URL` and `see also`. Even then, neither is implemented in a structured way.

Since we're already using markdown, one possibility to make these fields consistent and extensible would be to use the frontmatter metadata format, like this:

```md
---
foo: bar
baz: qux
---

(rest of markdown content)
```

That would make our pages compatible with other markdown processing tools and in particular with other cheatsheet repositories, as described by @chrisallenlane in https://github.com/chubin/cheat.sheets/issues/101#issuecomment-632032627.

chrisallenlane:
+1. Front-matter has worked really well for `cheat`. I think it would be the most painless approach to make the cheatsheet repositories portable, without imposing meaningful constraints on any of our individual projects.

Just shout if I can somehow be of service.

sbrl:
Ah, thanks for opening this issue @waldyrious!

Indeed, taking a more structured approach would be great.

Front-matter does sound like a good idea. It bothers me a bit that it's above the main title of the page though - but that's not a big deal.

I'm not sure that it's part of the CommonMark spec though, so it would require a breaking change update to the client specification.

We'd also have to perform an automated update to existing pages, and potentially rename the .zip archive too (which might be a good opportunity to deal with that other issue that was talking about restructuring the .zip file by splitting it into multiple languages, but I can't remember the issue number).

waldyrious:
Good point. @chrisallenlane do you know if these metadata blocks can be added to the bottom of the markdown file rather than the beginning? It would be great if we could keep our format `cat`/`less`-friendly.

Managor:
I'm not seeing any unresolved proposals in this issue. If I missed any, please reopen the issue.

PR #1248, opened 2017-01-18
Add links to full man pages

hginzel:
Add links to full man pages if available, please.
This commit was generated by `bash add_links.sh`.
http://man7.org/

### DIFF
--- a/pages/common/add_links.sh
+++ b/pages/common/add_links.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+#set -x	# debug
+
+verbose=1
+url="http://man7.org/linux/man-pages/man1/"
+ext=.1.html
+
+url="${url%/}/"	# add trailng slash if missing
+ext=".${ext#.}"	# add leading dot if missing
+
+check() {
+	test "`curl -s -o /dev/null -I -w "%{http_code}" "$url$1$ext"`" = "200"
+}
+
+# main code
+for fn in *
+do
+	c=${fn%.md}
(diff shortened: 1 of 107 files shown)


hginzel:
Signed.

agnivade:
Hi @hginzel, thanks for the PR. But I will have to deny this. Reasons being that we have a fixed markdown format which is consumed by all the clients. And adding a link in the main header breaks the formatting, as you can see from the failed travis test.

Also, I think if the users want to see the man pages, they can just do `man <command>` to get the page. It is unnecessary to open a link for that. And it also adds the requirement of having an internet connection.

And the files under the `pages` folder only contains markdown files following the format that we have. (You have added a bash script). Please take a look at [CONTRIBUTING.md](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) for that.

waldyrious:
@hginzel at the moment we can't change way the pages are structured, as @agnivade said, because many clients depend on the current format. See #958 for an example of how such a change can be carried out, including the client's maintainers in the discussion.

What you're proposing is reasonable, and such ideas have surfaced in the past; #1082 makes a summary of them. The fact that that issue is open means that we're indeed open to consider adding such additional fields, but they can't be done without community consensus. I'm sorry for the work you had preparing this PR, but at the moment we can't merge it.

My suggestion is to make a proposal similar to #958 (maybe after it is merged), including all the fields mentioned in #1082, and we can move forward from there.

PR #1251, opened 2017-01-22
brew: add doctor example

jonlabelle:
Add brew doctor example (checks system for potential problems).

### DIFF
--- a/pages/linux/brew.md
+++ b/pages/linux/brew.md
@@ -22,10 +22,6 @@
 
 `brew update`
 
-- Switch version of formula:
-
-`brew switch {{formula}} {{version}}`
-
 - Display information about formula, which contains formula version, installed path, dependencies, etc.:
 
 `brew info {{formula}}`
--- a/pages/linux/brew.md
+++ b/pages/linux/brew.md
@@ -29,3 +29,7 @@
 - Display information about formula, which contains formula version, installed path, dependencies, etc.:
 
 `brew info {{formula}}`
+
+- Check your system for potential problems:
+
+`brew doctor`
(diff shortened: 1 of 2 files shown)


MarloweW:
Does this command works in linux version? If it does, please add it to pages/linux/brew.md

agnivade:
Hi @jonlabelle - Since your example hits our maximum no. of allowed examples, we will have to be a bit cautious here and only add a new example if its absolutely necessary to do that. i.e. - If the `brew doctor` option is used commonly enough ? And if so, could it replace one of the existing examples that may be less common/useful?

jonlabelle:
Hi @agnivade. I've personally found the `doctor` option to be invaluable, on numerous occasions. Though I'm not sure it should replace any of the examples already listed.

waldyrious:
I'd suggest removing either `brew update` or `brew switch` (or maybe even both!). The first because brew itself tells you to update if it deems it necessary, and the second because it's probably less common (or at least useful to a smaller subset of people, IMO) than `brew doctor`.

jonlabelle:
I wouldn't recommend removing any of the core examples already listed. Though feel it's incomplete without it.

waldyrious:
@jonlabelle, could elaborate a bit? Specifically, I'd like to hear your thoughts about the two reasons I mentioned above to justify removing each of the removals I proposed.

jonlabelle:
I can only speak for myself... but I would replace the existing `switch` example, with the `doctor` example. I guess only because I've never needed to `switch` versions of a particular formula.

And even if brew can detect and report whether or not it needs updating, updating the package manager itself (and knowing how to manually) is too essential to leave out.

waldyrious:
Fair enough. Please go ahead and remove the `switch` example, then, to make room for the `doctor` one.

PR #1259, opened 2017-01-28
Bringing back tldr bot again

agnivade:
- Removed all the permissions of the bot, so that github won't remove
the token.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ cache: false
 after_success:
 - bash scripts/build.sh
 
+after_failure:
+- cat test_result | python scripts/post_pr_comment.py
+
 env:
   global:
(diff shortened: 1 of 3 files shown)


agnivade:
@waldyrious - can you approve this ?

ostera:
👍

ostera:
Even if the bot didn't have permissions, having the token exposed allows anyone to use it and have us rate-limited.

I like having the bot not have any permissions, but I'd rather not expose the token at all and keep using travis encrypted env variables.

agnivade:
You have a point. Its just that there's some weird github logic where travis encrypted variables are not available for PRs from foreign repos. Now, I am not sure after removing all permissions, this still applies or not. There's no way to test it out anyway without merging it first. Let me update it.

ostera:
Ah that would make sense. Otherwise anyone could fork a project, submit a
PR, and steal the secrets as they er unencrypted inside of the CI process.
sön 29 jan. 2017 kl. 15:51 skrev Agniva De Sarker <notifications@github.com

PR #1272, opened 2017-02-14
msmtp: add page

agnivade:
Taking off from #1123


agnivade:
@jeeftor - can you take a look at this ?

PR #1275, opened 2017-02-16
git-log: add graph view

jeeftor:
git-log-pic

added the pretty git-log graph function into the list

### DIFF
--- a/pages/common/git-log.md
+++ b/pages/common/git-log.md
@@ -14,9 +14,9 @@
 
 `git log --oneline`
 
-- Show all commits and branches in a graph format:
+- Show all commits and tags on all branches in a graph format:
 
-`git log --oneline --abbrev-commit --decorate --all --graph`
+`git log --oneline --decorate --all --graph`
 
 - Show only commits whose messages include a given string (case-insensitively):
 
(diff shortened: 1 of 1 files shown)


pages/common/git-log.md
```diff
@@ -14,6 +14,10 @@
 
 `git log --oneline`
 
+- Show all commits and branches in a graph format:
+
+`git log --oneline --abbrev-commit --decorate --all --graph`
```
agnivade: I think you can do without the `--abbrev-commit` option. Its good not to overload an example with too many options.

Also, please explain the `--decorate` and `--all` options. `--oneline` is already done above.

jeeftor: Ok so online is kind of useful to make the graph look nice.  It looks like `oneline` also runs the abrev-commit (show short commit hash instead of the 40 character one).

waldyrious:
I didn't have the opportunity to comment on this PR, but I just wanted to say I love it! ❤️ I have this alias (with a few extra tweaks) in my git config, and use it all the time!

ps - I'm happy with the final version that got merged, I wouldn't have changed anything :)

PR #1279, opened 2017-02-21
figlet: edit page

jeeftor:
Added additional info on how to get the list of available fonts.  

Also should this command be moved into common because you can run it on OS X??

### DIFF
--- a/pages/linux/figlet.md
+++ b/pages/linux/figlet.md
@@ -13,3 +13,7 @@
 - Pipe command output through figlet:
 
 `{{command}} | figlet`
+
+- Show available figlet fonts:
+
+`showfigfonts {{optional_string_to_display}}`


agnivade:
The command is `figlet`, but you have mentioned `showfigfonts`. Are both the same ?

jeeftor:
Figlet when you install comes with a separate command called showfogfonts. So the question is How do you tie the. Together. It runs figlet with a string and prints out the text for each font. So it's important to be tied into figlet if you want to do other fonts but not sure the best way to do so. A separate command seems like you wouldn't find it

agnivade:
Hmm, so linking 2 pages is possible. See the common/uname.md file. The node client will show a 'see also' section at the end of the command. You just have to add another line in the heading linking to `showfigfonts` command.

But if the user is probably never going to search for the `showfigfonts` as a standalone command, then for the sake of practicality, it makes sense to put them in a single page.

Thoughts ?

jeeftor:
Thoughts are they are never going to search for showfigfonts separately from figlet.

PR #1280, opened 2017-02-21
id3tag.md: add page

Dahie:
add id3tag for editing id3 tags of audio files

### DIFF
--- a/pages/common/id3tag.md
+++ b/pages/common/id3tag.md
@@ -1,10 +1,14 @@
 # id3tag
 
-> Tool for reading, writing, and manipulating ID3v1 and ID3v2 tags.
+> Tool for reading, writing, and manipulating ID3v1 and ID3v2 tags of MP3 files.
 
-- Set artist and title tag of an MP3:
+- Set artist and title tag of one MP3 file on the given path:
 
-`id3tag --artist={{artist}} --title={{title}} --album={{album}} {{path/to/file}}`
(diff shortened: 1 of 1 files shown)


agnivade:
Thanks for your PR ! Please fix the travis syntax errors first, then we'll get to bringing the PR to shape. :)

agnivade:
Also, your page title says `id3lib`, but the command is `id3tag`. If both are the same, lets use a consistent naming to avoid confusion.

agnivade:
@Dahie - It seems you have trailing whitespaces on your lines. I think the CI will pass once you fix those.

pages/common/id3tag.md
```diff
@@ -0,0 +1,11 @@
+# id3tag
+
+> Tool for reading, writing, and manipulating ID3v1 and ID3v2 tags.
```
agnivade: maybe mention "of mp3 files" at the end for clarity ?

pages/common/id3tag.md
```diff
@@ -0,0 +1,11 @@
...
+
+- Set artist and title tag of an MP3:
```
agnivade: of an MP3 -> of an MP3 file.

Also, you mention artist and title tag in the description, but the command also sets the album. If you want, we can add another example for album or just modify this description to add album too.

Dahie:
Hi, thanks for the feedback, I split the example into two, first one showing an edit of a single file and the second updates the album for all mp3 within the current folder.

PR #1284, opened 2017-02-28
qlmanage: Added new syntax

jeeftor:
qlmanage: Added reset example


agnivade:
@jeeftor - I made some changes. Can you check if they are alright ?

jeeftor:
@agnivade Looks good to me - i didnt even think to look over stuff :) - but you improved it i'd say.

PR #1285, opened 2017-03-01
lsof: Add description how to use by PID

jkubrynski:


agnivade:
@jeeftor - please take a look.

jkubrynski:
I see that few lines above you use `PID` so I think it'll be clear

PR #1289, opened 2017-03-02
find: ignore given directory

ramwin:
In some case, people want to find a file **excluding** the file in the library path.  
If you use python, you may want to find some file ignoring the file in `site-packages` directory.
If you use node, you may want to find some file ignoring the file in the `node_modules` directory.
I use `tldr find` and find that there is no such an example.  
I searched the google, and found the [Answer](http://stackoverflow.com/a/15736463/3601654). 
So, I add the example to the page of `find` and hope it will be helpful for other people.


ramwin:
Thank you for your attention. If there is any question, please let me konw. I will reply it as soon as possible.

agnivade:
Hi @ramwin  - thanks for your contribution. Unfortunately, this exceeds our max. no. of examples (8) that we allow per page. Our primary aim, as the name tldr suggests, is to mention the most commonly used examples of a command and keep it within limits.

But we can replace any of the existing examples with this one, provided sufficient reasons.

Let us know what you think.

ramwin:
Yes, I agree with you. It'r really hard for me to choose a command which is least frequently to be used.   
I discussed with my colleague. Maybe it's a good idea to replace the command `find root_path -empty`.  
In my daily jobs, I use this command least frequently.  
It's just my own option, there may be a better choice.

pages/common/find.md
```diff
@@ -33,3 +33,7 @@
 - Find files matching more than one search criteria:
 
 `find {{root_path}} -name '{{*.py}}' -or -name '{{*.r}}'`
+
+- Find files matching path pattern while excluding some certain path:
+
+`find {{root_path}} -name '{{*.py}}' -not -path '*/site-packages/*'`
```
jeeftor: You should fully tokenize the `-not -path` command

ramwin: Sorry, I can't understand the meaning of tokenize.  
I imitatethe structure of `find {{root_path}} -name '{{*.py}}' -or -name '{{*.r}}'` from `find.md`

ramwin:
To meet limition of the max. no. of examples (8) that we allow per page. I have deleted the command of finding empty files or directories. 
So far, I haven't read the source code of tldr-python-client.  It often occurs errors when I want to use the tldr for a second time. Like this:  
error_picture  
I'm willing to analysis the package and fix bug if it's within my reach.

pages/common/find.md
agnivade: It means  -not -path `'{{*/site_packages/*}}'`

ramwin: Thank you. I have changed it and updated the master branch.  
No offense, I think it would be more accurate if it was "You should fully tokenize the `*/site-packages/*` parameter".

agnivade:
@jeeftor - please merge(squash) if everything looks good to you.

PR #1291, opened 2017-03-03
Added tldr-bash-client

pepa65:
According to https://github.com/tldr-pages/tldr-pages.github.io/issues/19#issuecomment-283750531


waldyrious:
For reference, this PR was matched by an equivalent PR to the tldr-pages website: https://github.com/tldr-pages/tldr-pages.github.io/pull/20.

(We really ought to keep the list of clients on a single place...)

PR #1294, opened 2017-03-09
tomb: add page

mkaito:


mkaito:
There we go. I must have skimmed right over that part of the guidelines. I didn't bother installing the linter either, but I think it's okay now?

mkaito:
Those 4 lines of code *are* the basic use case. It's the response to the most basic question: "wait, how do I use this `tomb` thing again?". Sure, it's 4 lines, but they are all necessary for initial setup of a tomb.

I first wrote them all out in separate examples, but ended up with too many examples. And aside from opening a tomb, the other 3 commands are almost never used separately.

jeeftor:
I'm somewhat new to this project but I have yet to see any other TLDR pages
that have a 4 line block of of statements.  Thats why i'm waiting for
somebody else to weigh in.

mkaito:
Sure, just making my point assuming @agnivade (or whoever ends up reviewing) is not familiar with `tomb`.

pages/linux/tomb.md
```diff
@@ -0,0 +1,26 @@
+# tomb
+
+> Manage the creation and access of encrypted storage files.
```
agnivade: We should mention that `tomb` creates an encrypted "folder" to store files. From your description, it might appear that `tomb` creates encrypted files too.

mkaito:
I think this project might simply not be a good fit for a cheat sheet for `tomb` after all. I'm closing the PR.

waldyrious:
I'll give this one a shot. It's a good challenge to adapt various commands to a tldr-style format :)

waldyrious:
Please take a look @mkaito, @jeeftor and @agnivade.

pages/linux/tomb.md
```diff
@@ -0,0 +1,31 @@
+# tomb
+
+> Manage encrypted storage folders that can be safely transported and hidden in a filesystem.
+
+- Create a new tomb with an initial size of 100MB:
+
+`tomb dig -s 100 {{encrypted_folder.tomb}}`
```
agnivade: wrap 100 in tokens

pages/linux/tomb.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Create a new key file that can be used to lock a tomb, and set its password:
```
agnivade: I am guessing it asks for the password interactively ? If yes, we should add it in parentheses.

waldyrious: "and set its password via an interactive prompt:"?

pages/linux/tomb.md
waldyrious: 👍 Will do.

pages/linux/tomb.md
agnivade: I like the style we followed in `mongo` - "Connect to a database with a given username; user will be prompted for password".

waldyrious: Ok, sounds good. My only concern is not making clear that the prompt will request a password *for the key*, not the user's password... any ideas, or do you think it's not a big deal?

agnivade: Good point. How about - 

"Create a new key file that can be used to lock the tomb; user will be prompted to enter a password for it." ?

waldyrious: A bit long, but it works. I'll make the change.

waldyrious: I changed it slightly, to "Create a new key file that can be used to lock a tomb; user will be prompted for a password for the key:"

PR #1297, opened 2017-03-09
docker: added rm command

jeeftor:
Added copy command to docker

### DIFF
--- a/pages/common/docker.md
+++ b/pages/common/docker.md
@@ -25,3 +25,7 @@
 - Run a command inside of an already running container:
 
 `docker exec {{container}} {{command}}`
+
+- Copy a file from a docker container:
+
+`docker cp {{container}}:{{src_path}} {{dest_path}}`
--- a/pages/common/docker.md
+++ b/pages/common/docker.md
@@ -26,6 +26,6 @@
 
 `docker exec {{container}} {{command}}`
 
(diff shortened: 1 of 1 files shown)


agnivade:
Hey @jeeftor - since we are close to the max no. of examples, we have to be extra judicious in what we choose to add. In this case, I think adding the `rm` or `search` option is more likely to be used than `copy`. It also matches with the other commands already present like `start` and `stop`.

Let me know your thoughts.

jeeftor:
So the way i operate with TLDR pages is that if i do something and it isn't there then i add it.  If you want to add the other commands go for it 👍 - but i dont actually use docker that frequently ... so I leave it up to you in this case :)

agnivade:
> So the way i operate with TLDR pages is that if i do something and it isn't there then i add it.

That's absolutely perfect. But sometimes, what we use might not be the most appropriate example to add. The point here is that tldr being limited in scope, we always try to keep the most popular/practical examples from a newcomer's perspective. 

In that vein, I was saying that instead of adding `cp`, we should use `rm` or `search`, because those are the ones a newcomer will probably need before doing `cp`.

agnivade:
@jeeftor - I swapped `cp` with `rm`. Let me know if that's alright with you.

jeeftor:
All is good!

PR #1300, opened 2017-03-13
git: add a bunch of commands (from bropages)

sbibauw:
It think these are some of the more useful commands for Git. There are still many great commands and ideas on http://bropages.org/git that are worth importing here I think.
All the credit goes to the original bropages posters (anonymous).

### DIFF
--- a/pages/common/git.md
+++ b/pages/common/git.md
@@ -2,6 +2,36 @@
 
 > Distributed version control system.
 
+- Setup Git global configuration
+
+`git config --global user.name "John Doe"
+git config --global user.email johndoe@example.com
+git config --global core.editor vim`
+
+- Initiate version control in the current directory, creating a .git folder
+
+`git init`
+
+- Review unstaged changes (paginate), deciding whether or not to add them to the commit
+
+`git add -p`
+
+- Show diff of staged changes
+
+`git diff --cached`
+
+- Commit changes with a short description
+
+`git commit -m "Fixed something"`
+
+- Delete last commit
+
+`git reset --hard HEAD~1`
+
+- Create and checkout a new branch
+
+`git checkout -b mybranch`
+
 - Check the Git version:
 
 `git --version`
--- a/pages/common/git.md
+++ b/pages/common/git.md
@@ -4,9 +4,9 @@
 
 - Setup Git global configuration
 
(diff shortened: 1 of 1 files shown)


jeeftor:
Honestly I think `git status` is a pretty important command you don't have there.

sbibauw:
@jeeftor Thank you for your suggestion: you are absolutely right about `git status`. I added it.

agnivade:
Hey @sbibauw - All these commands already have tldr pages. The convention we have for commands split by space is to have the filename joined by '-'. Please do a `tldr <command>` before hand so as to avoid unecessary work. :smile:

sbibauw:
Thanks @agnivade. This explains why there were so few examples on `tldr git`. And yet I think it is totally necessary to have a (succinct) description of the main commands of a program like `git`. In the same way that you don't have to type `tldr tar xjf` but only `tldr tar`, because, if you're a beginner, you don't really know much about the commands of each program. So, while it's great to know that there is more under `tldr git {{command}}`, I do believe more examples in `git` are needed.

agnivade:
Hey @sbibauw - there is a difference between the `tar` example and the `git` example. The xjf are flags for the tar command itself. Whereas, status and commit for eg. are subcommands under git, which have flags separately. We have a lot of commands and sub-commands covered in this manner. Check out the `consul` pages for eg.

But its true that beginners might not know about sub-commands right away. In that case, its good to gently point them to that direction. We can say that -

```
Show help for a sub-command-
git {{sub-command}} --help
```

That way, the users will know there are sub-commands to use.

agnivade:
Hey @sbibauw - I am going to close this now. I believe I was able to clarify your concerns. Feel free to reply if you have something else to say.

Thanks for your efforts anyways ! Looking forward to more good stuff :)

sbibauw:
OK @agnivade. I still believe it would be more useful for people who need `tldr` (like me) to have an immediate summary of the main functions, but I understand the convention was set differently... :)

waldyrious:
@sbibauw I think #481 would solve the concern you have about the page for the main git command not giving hints about where to find the remaining information. If so, please follow that thread. I can't promise it will move ahead soon, but once it does we can address the git issue :)

PR #1302, opened 2017-03-15
yesod: add page

alx741:
The *yesod* CLI has a `--help` option but using it through the *stack* tool is a requirement and thus passing double dash arguments to it is cumbersome. This `TLDR` comes in handy to remember that clumsy commands to actually use the CLI tool.

### DIFF
--- a/pages/common/yesod.md
+++ b/pages/common/yesod.md
@@ -1,11 +1,11 @@
 # yesod
 
 > A framework to develop Haskell web applications.
-> All yesod commands are invoked through the stack project manager.
+> All yesod commands are invoked through the `stack` project manager.
 
-- Create a new scaffolded site:
+- Create a new scaffolded site with sqlite as backend in `my-project` directory:
 
-`stack new my-project yesod-sqlite`
+`stack new {{my-project}} {{yesod-sqlite}}`
 
 - Install the Yesod CLI tool within a Yesod scaffolded site:
 
--- a/pages/common/yesod.md
+++ b/pages/common/yesod.md
@@ -13,12 +13,12 @@
 
 - Start development server:
 
-`stack -- exec yesod devel`
+`stack exec -- yesod devel`
 
 - Touch files with altered Template Haskell dependencies:
 
-`stack -- exec yesod touch`
+`stack exec -- yesod touch`
 
 - Deploy application using Keter (Yesod's deployment manager):
 
-`stack -- exec yesod keter`
+`stack exec -- yesod keter`
--- a/pages/common/yesod.md
+++ b/pages/common/yesod.md
@@ -1,0 +1,19 @@
+# yesod
+
+> Yesod web framework CLI helper.
+
+- Install Yesod CLI tool in a Yesod scaffolding:
+
+`stack build yesod-bin cabal-install --install-ghc`
+
+- Start development server:
+
+`stack -- exec yesod devel`
+
+- Touch files with altered Template Haskell dependencies:
+
+`stack -- exec yesod devel`
+
+- Deploy application using Keter:
+
+`stack -- exec yesod keter`
--- a/pages/common/yesod.md
+++ b/pages/common/yesod.md
@@ -1,8 +1,13 @@
 # yesod
 
-> Yesod web framework CLI helper.
+> A framework to develop Haskell web applications.
+> All yesod commands are invoked through the stack project manager.
 
-- Install Yesod CLI tool in a Yesod scaffolding:
+- Create a new scaffolded site:
+
+`stack new my-project yesod-sqlite`
+
+- Install the Yesod CLI tool within a Yesod scaffolded site:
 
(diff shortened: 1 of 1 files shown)


pages/common/yesod.md
```diff
@@ -0,0 +1,19 @@
+# yesod
+
+> Yesod web framework CLI helper.
```
agnivade: You just defined a word with that word. :smile:  How about - 
```
A framework to develop Haskell web applications
```

pages/common/yesod.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Install Yesod CLI tool in a Yesod scaffolding:
```
agnivade: Our scope in tldr pages is limited. So we do not mention the installation step. You can remove this.

pages/common/yesod.md
```diff
@@ -0,0 +1,19 @@
...
+
+`stack build yesod-bin cabal-install --install-ghc`
+
+- Start development server:
+
+`stack -- exec yesod devel`
```
agnivade: I am a bit confused here. You mention the tool as `yesod`, but the command you use is `stack`. So should it be stack ?

pages/common/yesod.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Touch files with altered Template Haskell dependencies:
+
+`stack -- exec yesod devel`
+
+- Deploy application using Keter:
```
agnivade: Think of this description from a newcomer's perspective(like me). I have no idea what Keter is. Please describe that within parentheses.

pages/common/yesod.md
alx741: `yesod` here refers to the CLI tool, rather than the framework itself so I'm afraid that would be a bit misleading.

pages/common/yesod.md
alx741: Totally agree, however this doesn't refer to the installation of the tool, but rather the instantiation of the CLI in the scaffolding so it can be invoked (this is performed constantly, even on frequent updates of the underlying libraries being used in the project ), the term `installation` here allows that ambiguity in the context, but is the accepted term and well understood for the community.

Moreover the *installation* (instantiation) is both the most vital and the most forgettable part of the CLI usage, actually it constitutes the main reason for this *TLDR* page. Anyways if it is a major impediment I will remove it.

pages/common/yesod.md
alx741: Yes, the `yesod` tool needs to be invoked through the `stack` tool so it can used it context. It's the canonical way to use it, and unfortunately is exactly what makes it cumbersome to get inline help and remember how to use it.

The `stack` tool is the package manager for Haskell, and deserves another (separate) *TLDR* page, I will write one as well and propose it in a separate PR.

pages/common/yesod.md
alx741: Didn't thought of that, will make a change on that to see if it gets clearer

pages/common/yesod.md
agnivade: But yesod *is* the framework right ? The website literally says 'yesod web framework'. You can frame it in a different way if you want, just don't use the word `yesod`.

pages/common/yesod.md
agnivade: Got it. Please add a second line the in the main description that 'All commands are invoked through the `stack` package manager'. This will automatically refer users to the tldr stack page whenever its created.

pages/common/yesod.md
agnivade: Got it. So lets try to make the description a bit clearer. Try to explain what a "Yesod scaffolding" is. Think of it from a newcomer's point of view.

pages/common/yesod.md
alx741: That one is great!

pages/common/yesod.md
```diff
@@ -0,0 +1,24 @@
...
+
+`stack build yesod-bin cabal-install --install-ghc`
+
+- Start development server:
+
+`stack -- exec yesod devel`
+
+- Touch files with altered Template Haskell dependencies:
+
+`stack -- exec yesod devel`
```
agnivade: Both this and the previous example use the same command. Ideally we would use the same description if the command is the same. Can we do anything to combine these two ? Or maybe have different commands for each ?

alx741: oh! that is a big mistake, my bad. Fixed

agnivade:
@alx741 - I did some minor changes. Let me know if those look alright to you.

psibi:
@alx741 I think it would be better if the `--` option was placed after the exec subcommand:

```
stack exec -- yesod devel
```

IIRC, the official docs also recommend like that.

alx741:
@agnivade Oh right! the tokes, thanks!
@psibi definitely, thanks for that, fixed

PR #1307, opened 2017-03-24
zramctl: add page

sbrl:
zram is a linux kernel module, available since 3.14 and installed by default, that allows you to create virtual disks that are compressed on the fly and stored in ram.

While this PR adds a page about it, I don't expect it to be anywhere near perfect at first :P

http://karelzak.blogspot.co.uk/2014/08/zramctl.html

### DIFF
--- a/pages/linux/zramctl.md
+++ b/pages/linux/zramctl.md
@@ -1,7 +1,7 @@
 # zramctl
 
 > Setup and control zram devices.
-> Use `mke2fs` to format zram devices to partitions.
+> Use `mke2fs` or `mkswap` to format zram devices to partitions.
 
 - Check if zram is enabled:
 
--- a/pages/linux/zramctl.md
+++ b/pages/linux/zramctl.md
@@ -1,12 +1,13 @@
 # zramctl
 
 > Setup and control zram devices.
+> Use `mke2fs` to format zram devices to partitions.
 
 - Check if zram is enabled:
 
 `lsmod | grep -i zram`
 
-- Enable zram with 2 devices:
+- Enable zram with 2 devices. Use `zramctl` to configure the devices further:
 
 `sudo modprobe zram num_devices={{2}}`
 
--- a/pages/linux/zramctl.md
+++ b/pages/linux/zramctl.md
@@ -1,0 +1,23 @@
+# zramctl
+
+> Setup and control zram devices
+
+- Check if zram is enabled:
+
+`lsmod | grep -i zram`
+
+- Enable zram with 2 devices:
+
+`sudo modprobe zram num_devices={{2}}`
+
+- Initialise the next free zram device to a 2GB virtual drive:
+
+`sudo zramctl --find --streams {{compression_streams_count}} --size {{2GB}} --algorithm lz4`
+
+- List currently initialised devices:
+
+`zramctl`
+
+- Format a zram device to ext4:
+
+`sudo mkfs.ext4 {{/dev/zram0}}`
--- a/pages/linux/zramctl.md
+++ b/pages/linux/zramctl.md
@@ -12,7 +12,7 @@
 
 - Initialise the next free zram device to a 2GB virtual drive:
 
-`sudo zramctl --find --streams {{compression_streams_count}} --size {{2GB}} --algorithm lz4`
+`sudo zramctl --find --streams {{4}} --size {{2GB}} --algorithm lz4`
 
 - List currently initialised devices:
 
--- a/pages/linux/zramctl.md
+++ b/pages/linux/zramctl.md
@@ -1,6 +1,6 @@
 # zramctl
 
-> Setup and control zram devices
+> Setup and control zram devices.
 
 - Check if zram is enabled:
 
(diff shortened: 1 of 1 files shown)


agnivade:
@sbrl - Please fix the travis errors :)

sbrl:
@agnivade Fixed, I think :D

pages/linux/zramctl.md
```diff
@@ -0,0 +1,23 @@
...
+
+`sudo zramctl --find --streams {{compression_streams_count}} --size {{2GB}} --algorithm lz4`
+
+- List currently initialised devices:
+
+`zramctl`
+
+- Format a zram device to ext4:
+
+`sudo mkfs.ext4 {{/dev/zram0}}`
```
agnivade: This is not a zram command. Should this be included ?

pages/linux/zramctl.md
```diff
@@ -0,0 +1,23 @@
+# zramctl
+
+> Setup and control zram devices.
+
+- Check if zram is enabled:
+
+`lsmod | grep -i zram`
+
+- Enable zram with 2 devices:
```
agnivade: What does "device" mean here exactly ? Trying to understand zram a bit. IIUC, zram allows a sort of extended virtual memory to use RAM further and prevent paging to disk.

So how does "device" play in here. Is it a single zram "partition" / "disk" ? And if so, where is the size specified ?

pages/linux/zramctl.md
```diff
@@ -0,0 +1,23 @@
...
+
+`sudo modprobe zram num_devices={{2}}`
+
+- Initialise the next free zram device to a 2GB virtual drive:
+
+`sudo zramctl --find --streams {{compression_streams_count}} --size {{2GB}} --algorithm lz4`
```
agnivade: lz4 to be in tokens. And also, since you have already given a concrete no. for size, maybe you should also give a concrete no. for "compression_streams_count".

sbrl: Good point. I'll change that!

pages/linux/zramctl.md
sbrl: > What does "device" mean here exactly ?

By "device" I meant "linux device".

> IIUC, zram allows a sort of extended virtual memory to use RAM further and prevent paging to disk.

Sort of. You can create virtual disks that are stored compressed in RAM, which you can then format as a swap disk with `mkswap`. With that you can enable the new swap partition with a higher priority than any other swap partitions you may already have.

sbrl: > So how does "device" play in here. Is it a single zram "partition" / "disk" ?

Yep. When enabling zram, you can tell it how many virtual disks (called devices) to create.

sbrl: > And if so, where is the size specified ?

Simply creating the disk doesn't initialise it to have a size. You need to use `zramctl` to do that

I've written [a blog post](https://starbeamrainbowlabs.com/blog/article.php?article=posts%2F226-Linux-zram.html) on it, if it helps :-)

sbrl:
@agnivade Updated.

pages/linux/zramctl.md
agnivade: Great clarifications !

pages/linux/zramctl.md
```diff
@@ -0,0 +1,23 @@
...
+
+`lsmod | grep -i zram`
+
+- Enable zram with 2 devices:
+
+`sudo modprobe zram num_devices={{2}}`
+
+- Initialise the next free zram device to a 2GB virtual drive:
+
+`sudo zramctl --find --streams {{4}} --size {{2GB}} --algorithm lz4`
```
agnivade: What does "--streams" mean ? Can we add it to the description ?

sbrl: Yeah, that probably ought to be added to the description. It specifies the number of compression streams that should be attached to the device. I interpret that to be the number of concurrent reads/writes it can handle at once. Either way, I assume you'd want to set it to the number of cpus you have for best performance.

sbrl: I'm just not sure now how I'd work that into the description though.

sbrl: Suggestions?

agnivade: I will take a stab at it when I have some time. Now that I have a clear idea as to what the concepts are.

agnivade:
@sbrl - Made some changes. I removed the last example and added a pointer to that command in the main description. Also, removed the `--streams` option as I think this is too complicated to be mentioned in a tldr example.

Let me know what you think.

sbrl:
Yeah, that certainly simplifies it! I've tweaked the description just a bit.

jeeftor:
Please sign the license file and we can look into merging this in...

sbrl:
@jeeftor I hadn't signed it? Thanks for pointing that out! I thought for sure I'd signed it before. I've signed it again - hopefully that should work :-)

agnivade:
This is the old license check which somehow came. I am going to force-merge this.

sbrl:
Yay, thanks :D

PR #1308, opened 2017-03-26
id3tag: Trim trailing whitespace in first example

jonlabelle:
id3tag: Trim trailing whitespace in first example


agnivade:
Going to merge it right away since its a simple change.

PR #1309, opened 2017-03-26
docker: add some more commands

keithdadkins:
and trimmed a few earlier examples to keep the list down to 8 or less


pages/common/docker.md
```diff
@@ -2,30 +2,34 @@
...
-- Run a command inside of an already running container:
+- Run a command inside an already running container:
 
 `docker exec {{container}} {{command}}`
 
 - Remove a container:
 
 `docker rm {{container}}`
+
+- Remove all containers:
```
agnivade: For the `ps` example, you have combined that to a single one saying (add -a to see all). But here, you have mentioned 'remove all containers' in a separate example. I would suggest to either remove the 'remove all' example, or combine it with the previous one. 

Since, your remove all example is not a direct command, rather an aggregate of multiple commands, I would incline towards removing this example altogether. Unless, absolutely required, we strive not to include these kinds of examples.

pages/common/docker.md
```diff
@@ -2,30 +2,34 @@
...
+
+`docker rm $(docker ps -a -q)`
+
+- Remove an image:
+
+`docker rmi {{image}}`
+
+- Remove unused (dangling) images:
```
agnivade: This is beyond tldr's scope. Let us remove this one.

waldyrious:
Reopening as the "previous prompt" wasn't provided by @probot-stale. Please see #1350 and #1360 for details.

agnivade:
ping @keithdadkins

keithdadkins:
I think a better approach would be to break `docker` down into subcommands like `git` does. The `docker` command contains dozens of subcommands, some of which have subcommands of their own. I will spend some time working on this when i get a chance, so the basic _docker.md_ page as it is now should be ok.

waldyrious:
Sounds good, @keithdadkins 👍 looking forward to see what you propose!

PR #1310, opened 2017-03-27
timew: add page

abactel:
This will add the `timew` command. Timewarrior is advertised as:

> ... a command line time tracking application, which allows you to record time spent on activities.

[For more information see their website](https://taskwarrior.org/docs/timewarrior/)

### DIFF
--- a/pages/common/timew.md
+++ b/pages/common/timew.md
@@ -1,10 +1,10 @@
 # timew
 
-> Time tracker.
+> A time tracking tool used to measure durations.
 
-- Start a new stopwatch:
+- Start a new stopwatch, giving a tag name to the activity being tracked:
 
-`timew start {{item_to_time}}`
+`timew start {{activity_tag}}`
 
 - View running stopwatches:
 


pages/common/timew.md
```diff
@@ -0,0 +1,19 @@
+# timew
+
+> Time tracker.
+
+- Start a new stopwatch:
+
+`timew {{item_to_time}} start`
```
agnivade: What is "item_to_time" here ? The man [page](https://taskwarrior.org/docs/timewarrior/syntax.html) doesn't mention anything before the 'start' option.

abactel: The `{{item_to_time}}` is supposed to be the tag, an identifier for the thing the user is tracking. I didn't realize the man page showed the tag being placed after an initial `timew start`, as both work methods work equally. I've changed it in commit 2b48f56 to match the man page anyway.

agnivade:
@abactel - I included the word 'tag' itself as I think it fits better with the description. Let me know what you think of the changes.

abactel:
I agree with this, these changes make a lot of sense

agnivade:
@jeeftor - please check

sbrl:
What about stopping a single timer?

abactel:
I forgot about that. The command to stop a single stopwatch of a given tag is `timew stop activity_tag`. I'll create a commit to add this now.

PR #1313, opened 2017-03-30
find: add directories example

jereze:


agnivade:
Hi @jereze  - thanks for your contribution. Unfortunately, this exceeds our max. no. of examples (8) that we allow per page. Our primary aim, as the name tldr suggests, is to mention the most commonly used examples of a command and keep it within limits.

But we can replace any of the existing examples with this one, provided sufficient reasons.

Let us know what you think.

waldyrious:
Reopening as the "previous prompt" wasn't provided by @probot-stale. Please see #1350 and #1360 for details.

PR #1317, opened 2017-04-05
killall: add page

Raindeer44:


Raindeer44:
Killall is a way to send termination signals to processes by using the process name rather than PID. It also sends the signal to all instances of the process rather than a singular PID (if applicable).

jeeftor:
how does this differ from `pkill -f` ?

Raindeer44:
`killall` and `pkill` are basically the same, but `killall` requires an exact match on the process name by default, which keeps users from accidentally killing a process with a similar name. In my opinion, this is the safer command to use, especially for newer users. I could add in wording saying the process name has to be exact.

agnivade:
>  I could add in wording saying the process name has to be exact.

Yes, that would be helpful.

Raindeer44:
I added some clarifying wording about the name matching.

agnivade:
Hey @Raindeer44 - I think repeating "exact_process_name" multiple times is redundant. Mentioning it at the top should be enough. I also fixed one typo and simplified the last description. 

Let me know how you feel about the changes.

Raindeer44:
I like the changes

PR #1318, opened 2017-04-07
eyeD3.md: add page

sbrl:
I was going to create a page on composer, but I got distracted and did one for eyeD3 instead :P It's a tool for manipulating mp3 files which I use for my personal music collection, and I keep forgetting the syntax for setting the cover art :P

### DIFF
--- a/pages/linux/eyeD3.md
+++ b/pages/linux/eyeD3.md
@@ -1,6 +1,6 @@
 # eyeD3
 
-> Manipulate mp3 files.
+> Read and manipulate metadata of mp3 files.
 
 - View information about an mp3 file:
 
--- a/pages/linux/eyeD3.md
+++ b/pages/linux/eyeD3.md
@@ -1,19 +1,19 @@
 # eyeD3
 
-> Read and manipulate metadata of mp3 files.
+> Read and manipulate metadata of MP3 files.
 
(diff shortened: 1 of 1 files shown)


pages/linux/eyeD3.md
```diff
@@ -0,0 +1,19 @@
+# eyeD3
+
+> Manipulate mp3 files.
+
+- View information about an mp3 file:
+
+`eyeD3 {{Sounds.mp3}}`
```
agnivade: {{Sounds.mp3}} => {{filename.mp3}}

sbrl:
@agnivade Good suggestions - is that better?

And yeah, it does _only_ take mp3 files, but if you are that way inclined you can give all your mp3 files a different extension and it won't particularly care.

sbrl:
that new description reads much better :D

agnivade:
Yes, I was talking about extensions. If that does not matter, then its not needed to put it outside the tokens. 

I made some changes in the description to make the tool's function a bit clearer. Let me know what you think of those.

sbrl:
Ummm bump?

agnivade:
I am waiting for @jeeftor or @waldyrious to lend their thoughts.

sbrl:
Awesome, thanks @waldyrious :D

PR #1320, opened 2017-04-13
mailx: Improve the readability of the descriptions

sbrl:
I noticed that the wording was a bit awkward on the mailx command page, so I improved it a bit.

### DIFF
--- a/pages/common/mailx.md
+++ b/pages/common/mailx.md
@@ -2,15 +2,15 @@
 
 > Send and receive mail.
 
-- To send mail, the body is typed after the command and ended with Control-D:
+- Send mail (the content should be typed after the command, and ended with Control-D):
 
 `mailx -s "{{subject}}" {{to_addr}}`
 
-- Send mail with a short body:
+- Send mail with content passed from another command:
 
 `echo "{{content}}" | mailx -s "{{subject}}" {{to_addr}}`
 
-- Send mail with a body from a file:
+- Send mail with content read from a file:
 
 `mailx -s "{{subject}}" {{to_addr}} < {{content.txt}}`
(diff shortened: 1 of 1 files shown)


agnivade:
Hmm, I'm not so sure if it improves the wording. I understand that mail "body" is the right technical term for it. But, the word "contents" feels more approachable from a beginner's viewpoint. It might not be immediately apparent that the word "body" means the message inside the mail. Whereas, when I say "contents", it unambiguously refers to the message.

Let me know your thoughts.

sbrl:
@agnivade Thinking about it, I can see why "content" would be better than "body" for new users - I can change it if you like. A web page has a body too though, and if you're looking something up in tldr, it's possible you'd already know about that. I don't know :P

The major wording issue that I intend to fix in this commit is the `Send mail with content which written in a file` though. It just sounds _awkward_.

agnivade:
> The major wording issue that I intend to fix in this commit is the `Send mail with content which written in a file` though. It just sounds _awkward_.

Yea that's no issue. We can change it to "Send mail with content taken from a file".

Pinging @waldyrious / @jeeftor for their opinions on the body vs content thing.

waldyrious:
I agree that "content" is more natural and approachable.

waldyrious:
I've made some changes to the PR, please let me know if you're ok with them @sbrl :)

sbrl:
@waldyrious Yep, LGTM! :+1:

waldyrious:
Thanks :) @agnivade / @jeeftor I'll leave the review/merge to one of you since I made changes myself. Don't forget to use the rebase merging option!

PR #1321, opened 2017-04-14
Recursive scp arguments reversed

mterwill:
The command to "copy [...] on a remote host to a local
directory" in the current scp tldr page was backwards.

scp arguments take the form `scp from to`.


mterwill:
This issue was introduced in #826.

agnivade:
Yea, that's why I put the 'from-to' terminology for extra clarity.

PR #1266, opened 2017-02-03
Update cowsay.md

jeeftor:
Print for the different options

Issue #1265, opened 2017-02-03
cowsay


jeeftor:
So the `cowsay` command allows you to use lots of "different" options for the command.  Is it appropriate to update the TLDR page to print out how you could get a list of the different `cowsay` options?  

I've used the following command: 

`ls /usr/share/cowsay -1 | rev | cut -c 5- | rev | xargs -I _ cowsay -f _ cowsay -f _`

to generate a list but didnt know if it fit into the overall mission of TLDR.  For example the mac `say` command has a TLDR option to print out the available voices.  Some commands, however, like this one dont have an internal feature for doing this but rather you have to look in a directory.

Thoughts?

```
 _______________
< cowsay -f tux >
 ---------------
   \
    \
        .--.
       |o_o |
       |:_/ |
      //   \ \
     (|     | )
    /'\_   _/`\
    \___)=(___/

 _________________
< cowsay -f udder >
 -----------------
  \
   \    (__)
        o o\
       ('') \---------
          \           \
           |          |\
           ||---(  )_|| *
           ||    UU  ||
           ==        ==
 _________________
< cowsay -f vader >
 -----------------
        \    ,-^-.
         \   !oYo!
          \ /./=\.\______
               ##        )\/\
                ||-----w||
                ||      ||

               Cowth Vader
 _______________________
< cowsay -f vader-koala >
 -----------------------
   \
    \        .
     .---.  //
    Y|o o|Y//
   /_(i=i)K/
   ~()~*~()~
    (_)-(_)

     Darth
     Vader
     koala
 _______________
< cowsay -f www >
 ---------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||--WWW |
                ||     ||
```

waldyrious:
If there are 3 or 4 options, we can add them to the example description, or as multiple tokens separated by pipes. If there are many, then we could point out the usual location of the directory in the example description.

A similar case [happened with the assimp command](https://github.com/tldr-pages/tldr/pull/1028#issuecomment-257071872) and the discussion could be used as reference.

waldyrious:
For future reference: this was closed in #1266. I have opened #1323 as I was unable to comment in that discussion.

PR #1323, opened 2017-04-14
cowsay: remove overly complex example, reword desc

waldyrious:
I'm sorry @jeeftor, but I don't think the example introduced in #1266 adheres to the spirit of tldr-pages. It seems to be overly complex, too specific, and making use of way more commands than we normally accept as auxiliary to demonstrate the main command's functionality.

Looking forward to hear what you guys think.

### DIFF
--- a/pages/common/cowsay.md
+++ b/pages/common/cowsay.md
@@ -1,8 +1,7 @@
 # cowsay
 
 > Generate an ASCII character like a cow or sheep saying or thinking something.
-> Available characters are stored in the /usr/share/cowsay on Linux.
-> And /usr/local/share/cows/ on Mac.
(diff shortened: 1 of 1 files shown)


waldyrious:
Oh, I remember now, this was initiated in #1265. It would be helpful to mention that issue number in #1266, @jeeftor :)

And as I suggested in that issue, I believe this complex command is not the best way to point out the variety of characters that can be used to "say" things with this command. I'd prefer a simple solution such as listing the directory or simply mentioning it in the description of the `cowsay -f` example.

agnivade:
>  I'd prefer a simple solution such as listing the directory or simply mentioning it in the description of the cowsay -f example.

Agree with this.

PR #1264, opened 2017-02-03
x11vnc, xdotool: add pages

jeeftor:

PR #1324, opened 2017-04-14
x11vnc: wording tweaks

waldyrious:
Follow-up of #1264.

I would also like to clarify the last example: it's not clear to me what "monitor" means here (screen or display?) and why the description says 2 but the tokens say 0 and 1.


waldyrious:
I've made some more changes based on the information in http://stackoverflow.com/a/1210332/266309 and https://en.wikipedia.org/wiki/X_Window_System#Nomenclature. Hopefully this clarifies the usage of the terms "display" and "screen".

PR #1326, opened 2017-04-15
curl: improve last example per issue #1246

### LINKED ISSUE #1246: curl: improve last example
The last example in the [curl.md](https://github.com/tldr-pages/tldr/blob/master/pages/common/curl.md) page is:

    - Pass client certificate and key for a secure resource:
    
    `curl -v -key {{key.pem}} -cacert {{ca.pem}} -cert {{client.pem}} -k {{https://example.com}}`

There are a few issues with this:

- It [appears](http://manpages.ubuntu.com/manpages/precise/en/man1/curl.1.html) that the full-word options require double dashes. We need to confirm this and change accordingly
- The relationship between `--key`, `--cert` and `--cacert` is unclear to me (i.e. how the latter two differ, which depend on which, which are meant to be used together, etc.)
- The `-v` is not necessarily related to that example, so by including it (which requires mentioning it in the description, as pointed out in #1240) makes it hard to make the description concise; it would be better to remove it (maybe moving it a simpler example?).

I haven't used curl much, so I'll defer to @imjacobclark who originally contributed this example in #735, as well as @rubenvereecken, @igorshubovych and @ostera who participated in that discussion, and also @agnivade who raised the topic in #1240.

morrme:
Fixes #1246

### DIFF
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -29,4 +29,4 @@
 
 - Pass client certificate and key for a secure resource:
 
-`curl -v -key {{key.pem}} -cacert {{ca.pem}} -cert {{client.pem}} -k {{https://example.com}}`
+`curl --key {{key.pem}} --cacert {{ca.pem}} --cert {{client.pem}} -k {{https://example.com}}`


waldyrious:
Nice work, and fast as well! LGTM :+1:

Issue #1327, opened 2017-04-15
composer: add page


waldyrious:
See #975 and #1230 for previous discussion. @sbrl has expressed interest in doing this.

sbrl:
Opened #1329.

sbrl:
#1329 is now merged, so this issue can now be closed.

PR #1329, opened 2017-04-15
composer: add page

sbrl:
This commit adds a brand-new `composer` command page. It's not based on any previous work - I purposely didn't look at them to remind myself of them :wink: 

For #1327.

### DIFF
--- a/pages/common/composer.md
+++ b/pages/common/composer.md
@@ -6,7 +6,7 @@
 
 `composer require {{user/package-name}}`
 
-- Install all dependencies in `composer.json` for this project:
+- Install all the dependencies in this project's `composer.json`:
 
 `composer install`
 
--- a/pages/common/composer.md
+++ b/pages/common/composer.md
@@ -2,7 +2,7 @@
 
 > A package-based dependency manager for PHP projects.
 
-- Make a package a dependency for this project:
+- Add a package a dependency for this project:
 
 `composer require {{user/package-name}}`
 
--- a/pages/common/composer.md
+++ b/pages/common/composer.md
@@ -2,19 +2,19 @@
 
 > A package-based dependency manager for PHP projects.
 
-- Add a package a dependency for this project:
+- Add a package as a dependency for this project, adding it to `composer.json`:
 
 `composer require {{user/package-name}}`
 
-- Install all dependencies for this project:
+- Install all dependencies in `composer.json` for this project:
 
 `composer install`
 
-- Uninstall a package from this project, removing it as a dependency:
+- Uninstall a package from this project, removing it as a dependency from `composer.json`:
 
 `composer remove {{user/package-name}}`
 
-- Update all the dependencies of this project:
+- Update all the dependencies in `composer.json` for this project:
 
 `composer update`
 
--- a/pages/common/composer.md
+++ b/pages/common/composer.md
@@ -14,7 +14,7 @@
 
 `composer install`
 
-- Remove a package from this project:
+- Uninstall a package from this project, removing it as a dependency:
 
 `composer remove {{user/package-name}}`
 
--- a/pages/common/composer.md
+++ b/pages/common/composer.md
@@ -1,6 +1,6 @@
 # composer
 
-> A per-project package manager for PHP.
+> A package-based dependency manager for PHP projects.
 
 - Update composer to the latest version:
 
--- a/pages/common/composer.md
+++ b/pages/common/composer.md
@@ -2,10 +2,6 @@
 
 > A package-based dependency manager for PHP projects.
 
-- Update composer to the latest version:
-
-`composer self-update`
-
 - Make a package a dependency for this project:
 
 `composer require {{user/package-name}}`
@@ -21,3 +17,7 @@
 - Update all the dependencies of this project:
 
 `composer update`
+
+- Update composer to the latest version:
+
+`composer self-update`
--- a/pages/common/composer.md
+++ b/pages/common/composer.md
@@ -14,7 +14,7 @@
 
 `composer remove {{user/package-name}}`
 
-- Update all the dependencies in `composer.json` for this project:
+- Update all the dependencies in this project's `composer.json`:
 
 `composer update`
 
--- a/pages/common/composer.md
+++ b/pages/common/composer.md
@@ -1,0 +1,23 @@
+# composer
+
+> A per-project package manager for PHP.
+
+- Update composer to the latest version:
+
+`composer self-update`
+
+- Make a package a dependency for this project:
+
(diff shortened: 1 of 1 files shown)


pages/common/composer.md
```diff
@@ -0,0 +1,23 @@
+# composer
+
+> A per-project package manager for PHP.
+
+- Update composer to the latest version:
```
waldyrious: I don't think this should be the first example. In fact, it could probably be the last one.

pages/common/composer.md
```diff
@@ -0,0 +1,23 @@
...
+
+`composer self-update`
+
+- Make a package a dependency for this project:
```
waldyrious: What defines a project (e.g. "this" project)? Is there a file similar to package.json, or does the command simply assume we mean the current directory?

pages/common/composer.md
```diff
@@ -0,0 +1,23 @@
...
+
+`composer require {{user/package-name}}`
+
+- Install all dependencies for this project:
+
+`composer install`
+
+- Remove a package from this project:
```
waldyrious: Maybe "Remove a package dependency from this project"?

sbrl: That might work, but `composer remove` also deletes the package if it's installed too.

pages/common/composer.md
sbrl: Yeah, it's very similar to npm's `package.json`. It installs things local to the folder you're in (current directory) - except it installs to `vendor`, whereas npm installs to `node_modules`.

pages/common/composer.md
sbrl: It's important that composer is updated, as it might not work correctly if it isn't.

sbrl: I'll still move it down if you think it's a good idea though.

pages/common/composer.md
```diff
@@ -0,0 +1,23 @@
+# composer
+
+> A per-project package manager for PHP.
```
agnivade: Do we need to say "per-" ? I think "project manager for PHP applications" should be good enough ?

pages/common/composer.md
agnivade: Hey @sbrl - its about the natural flow of actions that a user will do. Usually the user will first use composer to install and remove package dependencies, and then update `composer` when needed. 

We try to mention the options in the order of the natural flow of usage.

pages/common/composer.md
sbrl: It's not a project manager though. It manages dependencies on a project-by-project level.

agnivade: How about "dependency manager for PHP projects" ?

sbrl: Yep, that sounds better

pages/common/composer.md
waldyrious: Do you see a way to make it clearer in the description what we mean by "this project", or do you think it's not needed?

pages/common/composer.md
waldyrious: I'm not sure I see the problem/difference.

pages/common/composer.md
sbrl: I suppose we could use "the current project" instead, but that's still a bit unclear. What does the npm page do?

pages/common/composer.md
sbrl: It does 2 things:

1. Uninstall the package
2. Removes the dependency from `composer.json` and `composer.lock`

pages/common/composer.md
waldyrious: It [directly mentions package.json](https://github.com/tldr-pages/tldr/blob/master/pages/common/npm.md), maybe we should do the same here as well.

sbrl: Hmmm yeah

pages/common/composer.md
waldyrious: Ok; I don't think "Remove a package from this project" makes it clear that (1) and (2) will happen, and I don't think "Remove a package dependency from this project" conveys (2) to the exclusion of (1), which is what I believe is your issue.

However, I can't think of a concise way to say both things, so feel free to keep the current description if you think it's sufficiently informative.

sbrl:
I've done some head scratching on the remove a package issue - is that better?

agnivade:
Looks good. Just the fixing of the order of examples remain.

waldyrious:
Better than anything I managed to come up with :) nice work.

sbrl:
@waldyrious Thanks :smile: 
@agnivade Fixing the order? Which one? I can't remember :P

waldyrious:
> @agnivade Fixing the order? Which one? I can't remember :P

@sbrl he is referring to [this thread](https://github.com/tldr-pages/tldr/pull/1329#discussion_r111668167).

pages/common/composer.md
sbrl: Ahhhh I see - yeah I'll change that :P

sbrl:
Thanks @waldyrious - I think I missed that comment. I've fixed that

pages/common/composer.md
```diff
@@ -0,0 +1,23 @@
+# composer
+
+> A package-based dependency manager for PHP projects.
+
+- Make a package a dependency for this project:
```
waldyrious: Can "Make" be changed to "Add"?

sbrl:
@waldyrious Good thinking

waldyrious:
@sbrl did you change your mind regarding mentioning composer.json? I.e. [this thread](https://github.com/tldr-pages/tldr/pull/1329#discussion_r111668202).

sbrl:
@waldyrious Thanks! Apparently I'm all forgetful now I'm on holiday :P

pages/common/composer.md
```diff
@@ -0,0 +1,23 @@
+# composer
+
+> A package-based dependency manager for PHP projects.
+
+- Add a package as a dependency for this project, adding it to `composer.json`:
+
+`composer require {{user/package-name}}`
+
+- Install all dependencies in `composer.json` for this project:
```
waldyrious: Just a nitpick, but I'd say "Install all dependencies in this project's `composer.json`" would be clearer (both here and in the second-to-last example in the page). What do you think?

waldyrious:
@sbrl my latest inline comment is the final one, I promise 😇

sbrl:
Heh heh heh - Good thinking though. This page has been far more work than I thought it would be - glad it's almost done.

waldyrious:
You forgot the "both here and in the second-to-last example in the page", but now that I know you agree, I'll take care of it myself ;)

sbrl:
Awesome~! :smile:

Issue #1239, opened 2017-01-09
add -pie example to perl page


waldyrious:
Something like this:

```
- Replace all occurrences of a string in a file, overwriting the file (i.e. in-place):

`perl -p -i -e 's/{{find}}/{{replace}}/g' {{filename}}`

- Same, saving the previous file to an extension:

`perl -p -i'.old' -e 's/{{find}}/{{replace}}/g' {{filename}}`
```

Inspired in http://www.itfromallangles.com/2012/05/bash-a-better-way-to-sed/

More info:
- [http://perlpie.com](https://web.archive.org/web/20161230202527id_/http://www.perlpie.com/)
- http://explainshell.com/explain?cmd=perl+-p
    > **perl -p**  causes Perl to assume the following loop around your program,
    >         which makes it iterate over filename arguments somewhat like sed:
    >```
    >LINE:
    >   while (<>) {
    >       ...             # your program goes here
    >   } continue {
    >       print or die "-p destination: $!\n";
    >   }
    >```

waldyrious:
Maybe add the `-0` option to the second example, for multi-line search-replace.

morrme:
Is this issue one that you'd just like your suggestions added and formatted onto the correct page?

agnivade:
Yes, its a suggestion to improve the page. Anyone is free to take this up and send a PR.

sbrl:
Since the attached PR has been merged, can we close this issue?

agnivade:
Yes !

(Ideally, the PR should have the word "Closes #<issue_no>" which auto-closes the issue. In this case, it must have missed out)

waldyrious:
Yeah, it was titled "per issue #1239", which understandably can't be automatically assumed to mean it will close the issue.

waldyrious:
I'm planning to add the `-0` example mentioned above. Do you guys have opinions about whether we could remove one of the existing examples on [the page](https://github.com/tldr-pages/tldr/blob/master/pages/common/perl.md)? I was thinking either the `-c` or the `-d` examples -- any thoughts, @morrme, @agnivade, @assemusr, @sbrl?

agnivade:
I vote for `-d`.

morrme:
`-c` and `-d` are really useful.  a new example would get the page to 8, right at the example limit, but i'd actually vote for `-M` if we *had* to remove one.

waldyrious:
[...]

PR #1330, opened 2017-04-15
perl: add pie example per issue #1239

morrme:


pages/common/perl.md
```diff
@@ -21,3 +21,11 @@
 - Run a Perl script in debug mode, using `perldebug`:
 
 `perl -d {{script.pl}}`
+
+- Replace all occurrences of a string in a file with another string, overwriting the file in-place:
+
+`perl -p -i -e 's/{{find}}/{{replace}}/g' {{filename}}`
+
+- Same, saving an unmodified copy of the previous file with a different extension:
```
waldyrious: I'd suggest changing "previous" --> "original".

waldyrious:
@morrme LGTM 👍 apart from the comment I left inline. Do you think it would make the examples too complex to add the `-0` option as described in #1239?

morrme:
@waldyrious i can give it a try. it may potentially need it's own example.

jeeftor:
@waldyrious -  in light of us making `cowsay` much simpler - is this not overly complex for a perl example?

waldyrious:
@jeeftor do you mean the `-pie` examples introduced in this PR are too complex? Or that the page as a whole is too complex? Or that adding a -0 to one of these examples would make them too complex? Please clarify :)

pages/common/perl.md
```diff
@@ -21,3 +21,11 @@
 - Run a Perl script in debug mode, using `perldebug`:
 
 `perl -d {{script.pl}}`
+
+- Replace all occurrences of a string in a file with another string, overwriting the file in-place:
+
+`perl -p -i -e 's/{{find}}/{{replace}}/g' {{filename}}`
```
agnivade: Its slightly difficult to grasp what each of these options stand for. I can make out `-e` is for in-place. But what about `-p` and `-i` ?

We usually aim to have atleast a word starting with the character of the option, or highlight the character mid-word li[k]e this. Can we do something like this here ?

pages/common/perl.md
```diff
@@ -21,3 +21,11 @@
...
+
+- Same, saving an unmodified copy of the original file with a different extension:
```
agnivade: I think its better to write the full description for clarity's sake. And also mention that the original file is saved with '.old' extension

pages/common/perl.md
waldyrious: `-e` refers to an expression (rather than a perl script), and is explained in the previous examples on this page (which refer to it as a "statement" -- perhaps we should tweak those to say "expression", to improve the memorability).

`-i` is the option that stands for in-place, which is easy enough to describe.

`-p`, as described in #1239, seems to stand for "program", but is not really something that can be explained succinctly -- I'd rather leave it unexplained, unless you guys can come up with something nice :)

pages/common/perl.md
waldyrious: I'm afraid it would be too long... do you have a suitable phrasing in mind?

morrme: could we reverse the order of the examples, so that the "longer" one is described first, then the second can be more succinct?

pages/common/perl.md
morrme: `-p` has more to do with the loop behavior than the program itself, if that makes sense.

pages/common/perl.md
agnivade: "Replace all occurrences of a string in a file with another string, saving the original file with a given extension" - is this too long ?

pages/common/perl.md
waldyrious: @morrme yes, I wondered if we could somehow squeeze the description of "p" into something like "pack" (I couldn't come up with another stand-in for "wrap" or "loop" including "p"). But I am also not really sure what this loop means, and what would happen without it. Probably understanding this mechanism better would help describing it :)

pages/common/perl.md
waldyrious: Not too long :) I'll change it. (I wasn't fond of "same" myself.)

pages/common/perl.md
morrme: in this case, our [e]xpression/program is 's/replacethis/withthat/g'
we are loo[p]ing or iterating through every line of a text file
and making the substitution [i]nplace 

we can't iterate through each line of the file without `-p`

waldyrious: "loo[p]# sounds like a great way to incorporate an explanation for `-p`! @agnivade what do you think of `Loo[p] over all lines of a file, editing them [i]n-place using a find/replace [e]xpression:`?

pages/common/perl.md
```diff
@@ -21,3 +21,11 @@
 - Run a Perl script in debug mode, using `perldebug`:
 
 `perl -d {{script.pl}}`
+
+- Replace all occurrences of a string in a file with another string, overwriting the file in-place:
+
+`perl -p -i -e 's/{{find}}/{{replace}}/g' {{filename}}`
+
+- Replace all occurrences of a string in a file with another string, saving the original file with a given extension:
```
waldyrious: Perhaps we could change `"saving the original file with a given extension"` to `"saving a back-up with a given extension"`?

agnivade: I like the first one better. :neutral_face:

waldyrious: Sure, fair enough.

pages/common/perl.md
waldyrious: @agnivade here too :)

agnivade: Umm .. loop seems like a stretch. I am okay with not explaining `-p`

waldyrious: Not sure I see what you mean by a stretch -- it's literally the word used in the [documentation](http://perldoc.perl.org/perlrun.html):

> perl -p causes Perl to assume the following loop around your program (...)

Perhaps by stretch you mean using the last letter of loop? I actually think it's a perfectly good anchor, just like the start of a syllable in the middle of a word would.

agnivade: ah ok then. go ahead.

pages/common/perl.md
```diff
@@ -21,3 +21,11 @@
...
 
 `perl -d {{script.pl}}`
+
+- Loo[p] over all lines of a file, editing them [i]n-place using a find/replace [e]xpression:
+
+`perl -p -i -e 's/{{find}}/{{replace}}/g' {{filename}}`
+
+- Replace all occurrences of a string in a file with another string, saving the original file with a given extension:
+
+`perl -p -i'.old' -e 's/{{find}}/{{replace}}/g' {{filename}}`
```
waldyrious: @agnivade just to be sure, are you ok with this example's description not following the same structure as the one before it? I think extending the previous one would make it excessively long (in fact, I would even propose swapping
`Replace all occurrences of a string in a file with another string`
with
`Edit a file in-place using a find/replace expression`
or the even shorter
`Run a find/replace expression on a file`
which, when joined with the rest of the sentence, wouldn't make the description more reasonably sized :)

agnivade: Sure, I like the second version.

waldyrious: Not sure you mean second of the two alternatives I presented, or second among the three descriptions I listed. I'll assume the former, but feel free to correct me afterwards :)

waldyrious:
Please don't merge -- I'll do an interactive rebase later today to clean up the commits before merging.

waldyrious:
Hmm, maybe we should avoid interactive rebases if possible, as that removes the history of commits in the PR itself -- we normally just want the clean history in master.

Squash-merging, for example, adds a single commit to master and marks the PR as merged, but doesn't remove the commits in the PR branch.

Ultimately, this is a failure of the github interface, in that we should be able to mark a PR as cleanly merged rather than closed, if the contents of the PR branch do match the contents of master, even if the commits of the PR branch don't match the commits added to the master branch.

But granted, that's tricky to do, so yeah, there's no simple answer to this -- it'll have to be done on a case-by-case basis, keeping the caveats above in mind.

waldyrious:
Following-up the comment above, there's some relevant discussion on [this reddit thread](https://www.reddit.com/r/programming/comments/67om7r/editing_your_git_history_with_rebase_for_cleaner/).

PR #1334, opened 2017-04-17
apt: improve page

agorf:


agnivade:
I think its a good tradeoff - removing autoremove with show. On the same vein, I propose that we remove 'full-upgrade' too. Its very rarely used and easily beyond tldr's scope.

agorf:
@agnivade Removed. How do you feel about adding `apt edit-sources`?

Also, almost all commands require `sudo` (or root privileges), but they don't have any. Maybe it's a good idea to add that as well.

agnivade:
> How do you feel about adding apt edit-sources?

I would avoid it.

> Also, almost all commands require sudo (or root privileges), but they don't have any. Maybe it's a good idea to add that as well.

Yes, please do. :smile:

agorf:
@agnivade Done.

pages/linux/apt.md
```diff
@@ -4,28 +4,24 @@
 
 - Update list of packages and versions available. This should be run before running further apt commands:
 
-`apt update`
+`sudo apt update`
```
jeeftor: Could you explain why its important to add `sudo` to this command.  Does it not function in user mode?  My gut tells me we shouldn't use sudo - and if you are using apt-get you should know to sudo it.  But maybe the other mods feel differently.

agorf: `apt update` requires super-user privileges in order to function because it writes system-level files.

Regarding the presence of `sudo`, see bullet point 2 in https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines

agnivade: Yep, @agorf is right. The contribution guidelines mention it - if a command requires `sudo`, it should be mentioned.

sbrl:
I'd say include `sudo` in the example if required. If the consensus is that it's ok with `sudo` in the examples, then I'll merge this PR.

agnivade:
Yep, go ahead !

sbrl:
Merge complete!

PR #1268, opened 2017-02-13
pvcreate: add page

rgcr:

PR #1336, opened 2017-04-17
pvcreate: tweaks to descriptions

waldyrious:
Hopefully make them slightly clearer. Follow-up of #1268.

### DIFF
--- a/pages/linux/pvcreate.md
+++ b/pages/linux/pvcreate.md
@@ -1,11 +1,11 @@
 # pvcreate
 
-> Initialize a disk or partition for use by LVM (Logical Volume Manager).
+> Initialize a physical volume (disk or partition) for use by the Logical Volume Manager (LVM).
 
-- Initialize `/dev/sda1` to use lvm:
+- Initialize the `/dev/sda1` device for use by LVM:
 
 `pvcreate {{/dev/sda1}}`
 
-- Force the creation without any confirmation:
+- Force the creation without any confirmation prompts:
 
 `pvcreate --force {{/dev/sda1}}`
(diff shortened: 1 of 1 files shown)


pages/linux/pvcreate.md
```diff
@@ -1,11 +1,11 @@
 # pvcreate
 
-> Initialize a disk or partition for use by LVM (Logical Volume Manager).
+> Initialize a physical volume (disk or partition) for use by the Logical Volume Manager (LVM).
 
-- Initialize `/dev/sda1` to use lvm:
+- Initialize the `/dev/sda1` device for use by LVM:
```
agnivade: In the main description, the terminology used is "disk or partition", but here you use "device". I think its good to be consistent. Maybe we can use `/dev/sda1` partition here ?

waldyrious: The "disk or partition" comes from the [pvcreate manpage](https://linux.die.net/man/8/pvcreate):

> pvcreate initializes *PhysicalVolume* for later use by the Logical Volume Manager (LVM). Each *PhysicalVolume* can be a disk partition, whole disk, meta device, or loopback file.

Reading the diagram in [this section](https://en.wikipedia.org/wiki/Logical_Volume_Manager_(Linux)#Features), it seems like "/dev/..." are only used for logical (as opposed to physical) volumes, so I'm not really sure what to use here.

Perhaps our best bet would be to play it safe and say "the `/dev/sda1` volume", which is also consistent with the manpage's synopsis ("pvcreate [options...] PhysicalVolume"). What do you think?

agnivade: Seems fine to me. Pinging @rgcr for a second opinion.

waldyrious:
Since it's been over a week with no feedback from @rgcr, I think we can move on with the safe bet described in the inline comment thread ("the `/dev/sda1` volume"). We can always improve the page if a better wording is proposed :)

PR #1340, opened 2017-04-18
curl: simplify last example (#1246)

waldyrious:
I wonder if `-k` is needed here. If so, should we use the spelled-out version (`--insecure`)?

### DIFF
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -27,6 +27,6 @@
 
 `curl -u myusername:mypassword -I {{http://example.com}}`
 
-- Pass client certificate and key for a secure resource:
+- Pass client certificate and key for a (possibly insecure) resource:
 
 `curl --cert {{client.pem}} --key {{key.pem}} --insecure {{https://example.com}}`
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -29,4 +29,4 @@
 
 - Pass client certificate and key for a secure resource:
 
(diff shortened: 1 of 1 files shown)


morrme:
the insecure option prevents the command from failing if the url is not secure. it basically gets curl to skip that check. 

-- insecure "spells" it out more since -k isnt intuitive.

agnivade:
> -- insecure "spells" it out more since -k isnt intuitive.

:+1:

waldyrious:
@morrme in that case, perhaps we should remove the "secure" from the description? The combination of "secure resource" and `--insecure` may sound confusing.

morrme:
@waldyrious i agree with that. i didn't want to change too much since i'm new to the pages but it doesn't make things plain.

waldyrious:
Ok, I changed the desc as well -- please take a look and let me know what yout think, @agnivade and @morrme

morrme:
i'd prefer something that says the option is just disabling or skipping  the certificate validation. sometimes the server could have a self signed certificate, or one that is not issued by a trusted CA...so it's "secure" just not veritably so, which would make the statement fail.   

...but if you stick with your term, even though the option is --insecure, the correct term for your description would be ` unsecure`

waldyrious:
@morrme how about "Pass client certificate and key for a resource, skipping certificate validation"?

morrme:
@waldyrious that works! :+1:

waldyrious:
Ok, given the approval of you both, I'll merge as soon as all checks report positive :)

PR #1353, opened 2017-04-27
telnet: improve descriptions

waldyrious:
also format "x" as a token in the last example.

Note: I'm not entirely sure of the last example's description. Confirmations of its correctness welcome :)

### DIFF
--- a/pages/common/telnet.md
+++ b/pages/common/telnet.md
@@ -1,19 +1,19 @@
 # telnet
 
-> Telnet is used to connect to a specified port of a host.
+> Connect to a specified port of a host using the telnet protocol.
 
 - Telnet to a certain port:
 
 `telnet  {{ip_address}} {{port}}`
 
-- To exit a telnet session:
+- Exit a telnet session:
 
 `quit`
 
-- Default escape character:
+- Emit the default escape character:
 
 `CTRL + ]`
 
-- Specify an escape character (x is the escape character):
+- Start telnet with "x" as the escape character:
 
-`telnet -e x {{ip_address}} {{port}}`
+`telnet -e {{x}} {{ip_address}} {{port}}`
--- a/pages/common/telnet.md
+++ b/pages/common/telnet.md
@@ -2,18 +2,22 @@
 
 > Connect to a specified port of a host using the telnet protocol.
 
-- Telnet to a certain port:
+- Telnet to the default port of a host:
 
-`telnet  {{ip_address}} {{port}}`
+`telnel {{host}}`
+
+- Telnet to a specific port of a host:
+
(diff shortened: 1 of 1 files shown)


pages/common/telnet.md
```diff
@@ -1,19 +1,19 @@
...
 
-- To exit a telnet session:
+- Exit a telnet session:
 
 `quit`
 
-- Default escape character:
+- Emit the default escape character:
 
 `CTRL + ]`
```
agnivade: Shouldn't this be 'CTRL+]' according to your new convention ?

pages/common/telnet.md
```diff
@@ -1,19 +1,19 @@
 # telnet
 
-> Telnet is used to connect to a specified port of a host.
+> Connect to a specified port of a host using the telnet protocol.
 
 - Telnet to a certain port:
```
agnivade: While we are on this page, maybe we should start with a simple 'telnet {{host}}' example and say that telnet connects to port 23 by default.

pages/common/telnet.md
waldyrious: It should, but that makes it hard to read. Shall we adopt a convention of adding spaces around the + sign? That sounds reasonable and in line with the project's ethos (prioritizing readability over compactness).

agnivade: sure :+1:

waldyrious: Done in #1356.

sbrl: As a user, I might not know what an 'escape character' does. Perhaps it should be made more clear that terminates the session?

waldyrious: "session escape character"? "session termination character"? In fact, I'm not sure it's intruitive to call this a character. Maybe "character combination", "keyboard shortcut", or something to that effect?

pages/common/telnet.md
waldyrious: Not sure how I missed this comment -- probably the notification linked directly to the second comment you made. Yes, agree, and will make this change as soon as we reach consensus in the second thread below.

pages/common/telnet.md
sbrl: Yeah, that would work I think

waldyrious: Which one? xD

jeeftor: Is this ready for review?

waldyrious: I'm waiting for @sbrl to reply :)

sbrl: Ahhh I see
My comment doesn't specify which one I'd prefer. I vote for character combination / keyboard shortcut

sbrl:
Merge complete!

PR #1357, opened 2017-04-28
ip: add page

haishanh:
iproute2(command name `ip`) is a great collection of utilities for working with networking.

It is considered the successor of `ifconfig`. It has many features designed for the complex modern network environment(especially in virtualization environment), like VXLAN tunneling, network namespace, ip rules, etc..

More and more Linux distro began to ship with iproute2.  On MacOS you can install using Homebrew:

```
brew install iproute2mac
``` 

See also introduction section [here](https://wiki.linuxfoundation.org/networking/iproute2)

### DIFF
--- a/pages/ip.md
+++ b/pages/ip.md
@@ -1,0 +1,27 @@
+# ip
+
+> Networking utilities collection.
+
+- Check address of all interfaces:
+
+`ip {{addr}}`
+
+- Add IPv4 address for eth1:
+
+`ip {{addr}} add 1.2.3.4/24 dev eth1`
+
+- Bring eth1 up:
+
+`ip {{link}} set eth1 up`
+
+- Add subnet route for eth1:
+
+`ip {{route}} add 1.2.3.0/24 dev eth1`
+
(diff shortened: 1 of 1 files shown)


agnivade:
Hey @haishanh - A page for `ip` already exists under 'linux/'. If you want to make some changes, please update that file :)

And FYI - we don't put a file directly under 'pages/', but we put it inside platform specific folders, or under 'common/' if its common for all platforms.

haishanh:
Ah, I'm on a Mac and I forgot to check the directory `linux/`, sorry about that.  I'm closing this one.

FYI, I was intended to put the file in `pages/common`, don't know how I end up with it in `pages/`

agnivade:
No worries. If you had intended to put it under common/, you can still do that in this PR itself. But we just need to verify that it works the same on both mac and linux.

waldyrious:
@haishanh could you please confirm whether the examples in [the ip page under pages/linux/](https://github.com/tldr-pages/tldr/blob/master/pages/linux/ip.md) work for you as described there?

PR #1358, opened 2017-04-28
Update tmux.md: detach should be lowercase d

mreinhardt:
Uppercase D chooses which client to detach, mostly used for detaching other users from your session. Lowercase d is for detaching the session being used.

### DIFF
--- a/pages/common/tmux.md
+++ b/pages/common/tmux.md
@@ -24,7 +24,7 @@
 
 - Detach from session:
 
-`Ctrl + B, D`
+`Ctrl + B, d`
 
 - Kill session:
 


agnivade:
@waldyrious - I must have messed up while resolving it :sweat: Help needed :sweat_smile:

waldyrious:
No problem, I'll take a look in a bit :)

waldyrious:
@agnivade this was due to a conflict with the changes made in #1356. Instead of merging the master branch (I suppose you used github's web ui for that, correct?), the solution was to rebase Michael's commit over the latest master. I just did that, so things should be ready for review.

pages/common/tmux.md
```diff
@@ -24,7 +24,7 @@
 
 - Detach from session:
 
-`Ctrl + B, D`
+`Ctrl + B, d`
```
sbrl: Since we use `Ctrl +` here, would it not be safe to assume that the `Ctrl + B` is `Ctrl + B` and not `Ctrl + Shift + B`? The letters on my keyboard, at least, are in uppercase. It's a shame markdown doesn't have a `<kbd>` tag equivalent.

In the same way, a single `D` should mean the `D` on your keyboard, and not `SHIFT + D`.

agnivade:
Good reasoning. Agree with you both.

waldyrious:
@mreinhardt given the consensus in the discussion above, I'll close this, but we're be happy to consider arguments to the contrary, so feel free to comment if you have differing thoughts. In any case, thank you for taking the time to improve the project :)

Issue #1350, opened 2017-04-26
Add ping functionaly to tldr-bot


waldyrious:
tldr-bot was first introduced in #1066, as one of the measures identified in #980 to streamline the contribution process. (The idea had been raised as early as #278.)

Besides reporting the linter errors inline for convenience, it would be nice to expand its functionality to automate as much of the maintenance process as possible, in order to avoid burn-out and prevent threads from going stale. The possibility of pinging people involved in a thread (which nicely complements the great work currently done by @mention-bot) was [mentioned in #1209](https://github.com/tldr-pages/tldr/issues/1209#issuecomment-270510219):

> In the meantime, @agnivade, do you think we could repurpose @tldr-bot to (also) perform the necessary pinging after the periods we define here expire? You've been doing that manually, but we should automate as much of this activity as possible, so as to free up contributor/maintainer time for the actual changes :)

> The bot could even close PRs as you did in #975 -- the wording you used there seems like a great template we could adopt: friendly, informative, and with clear action paths for alternative resolution.

waldyrious:
Note: we could try the [probot-stale](https://github.com/probot/stale) integration to do this. Of course, we should be careful about closing issues/PRs, both for the reasons described [here](https://github.com/github/choosealicense.com/issues/501#issuecomment-292249558):

> I don't mind stale issues. Honestly, I like them. When a legitimate decade old bug finally gets resolved in an open source project, I'm tickled. Here, all of the ones open I intend to resolve eventually.

but especially because this could become quite aggressive if we're not careful. See [here](https://github.com/benbalter/licensee/issues/114) for example -- note particularly the:

> probot-stale bot added the `wontfix` label [and commented]:
> "This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

Lots of wrong things here: the "wontfix" label (which, even though it isn't in all caps as Bugzilla does, is still pretty aggressive), the "It will be closed if no further activity occurs" -- which, together, make the well-meaning "Thank you for your contributions." sound actually quite sarcastic. And this is evidenced by the response it got: various thumbs-down, and a reply saying

> Seriously? That's a horrible way to manage issues. In that case: bump (because you're forcing me to)

[...]

PR #1360, opened 2017-04-30
create configuration file for probot-stale

waldyrious:
So this is my proposal for configurations to use in our [probot-stale](https://github.com/probot/stale) integration, as discussed in #1350.

Any feedback about the time periods, messages, or other details of the configuration? /cc @agnivade @ostera @jeeftor @sbrl @igorshubovych @rubenvereecken.

Note: merging this won't automatically enable probot-stale on this repo; we'll need to activate it in https://github.com/integration/probot-stale.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,30 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an issue or PR is considered stale
+daysUntilStale: 30
+
+# Number of days of inactivity before a stale issue or PR is closed
+daysUntilClose: 30
+
+# Issues or PRs with these labels will never be considered stale
+exemptLabels:
+  - up for grabs
+
+# Label to use when marking an issue or PR as stale
+staleLabel: waiting
+
+# Comment to post when marking an issue or PR as stale
+markComment: >
+  Hi all! This thread has not had any recent activity.
+  Can anyone comment on what's blocking progress?
+  Thanks!
+
+# Comment to post when closing a stale issue or PR
+closeComment: >
+  Hi everyone.
+  This thread is being closed as there was no response to the previous prompt.
+  However, please leave a comment whenever you're ready to resume,
+  so the thread can be reopened. Thanks again!
+
+# Limit to only `issues` or `pulls`
+only: pulls
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,7 +1,7 @@
 # Configuration for probot-stale - https://github.com/probot/stale
 
 # Number of days of inactivity before an issue or PR is considered stale
-daysUntilStale: 30
+daysUntilStale: 15
 
 # Number of days of inactivity before a stale issue or PR is closed
 daysUntilClose: 30
@@ -16,8 +16,7 @@ staleLabel: waiting
 # Comment to post when marking an issue or PR as stale
 markComment: >
   Hi all! This thread has not had any recent activity.
-  Can anyone comment on what's blocking progress?
-  Thanks!
+  Are there any updates? Thanks!
 
 # Comment to post when closing a stale issue or PR
 closeComment: >


.github/stale.yml
```diff
@@ -0,0 +1,30 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an issue or PR is considered stale
+daysUntilStale: 30
```
agnivade: I would tune it down to 15

.github/stale.yml
```diff
@@ -0,0 +1,30 @@
...
+exemptLabels:
+  - up for grabs
+
+# Label to use when marking an issue or PR as stale
+staleLabel: waiting
+
+# Comment to post when marking an issue or PR as stale
+markComment: >
+  Hi all! This thread has not had any recent activity.
+  Can anyone comment on what's blocking progress?
```
agnivade: This line feels a bit weird. I would rather change it to "Any updates on this ?'

.github/stale.yml
```diff
@@ -0,0 +1,30 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an issue or PR is considered stale
+daysUntilStale: 30
+
+# Number of days of inactivity before a stale issue or PR is closed
+daysUntilClose: 30
```
sbrl: Is this absolute, or days after `daysUntilStale`?

agnivade: The later. Its mentioned _before a stale issue_

agnivade:
I have a question though. Lets say a PR becomes stale, the bot marks it as such. Then someone comments on it. Will the bot remove the label "stale" ?

Logic suggests that it should otherwise, it doesn't make sense that when the PR becomes stale again for the 2nd time, it can't re-label something when its already labelled.

.github/stale.yml
waldyrious: Sure.

.github/stale.yml
waldyrious: How about "Are there any updates?"

What I like about the initial version is that it makes a specific call to action rather than placing a generic question. But it's not a big deal, and we can tweak the message later if we can find a way to combine the advantages of both sentences.

waldyrious:
@agnivade yes, it does remove the label -- the example I pointed out in #1350 shows this in action:

> -  probot-stale bot added the wontfix label 24 days ago
> - probot-stale bot commented 24 days ago: "This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions."
>- TheLastProject commented 24 days ago: "Seriously? That's a horrible way to manage issues. In that case: bump (because you're forcing me to)"
> - probot-stale bot removed the wontfix label 24 days ago

.github/stale.yml
agnivade: > How about "Are there any updates?"

looks good.

waldyrious: Any comments on the call to action part?

agnivade: I mean I don't see that the new version is _not_ a call to action in any way. Its not direct, that's true. And I would rather not be too direct. The new version seems to have the right tonality and still ask for some work to be done in a friendly way.

waldyrious: Ok, I can see that.

waldyrious:
I made the requested changes, @agnivade & @sbrl -- please review :)

waldyrious:
Alright! let's try this out :)

waldyrious:
Ok, this is now installed, as can be confirmed in https://github.com/organizations/tldr-pages/settings/installations. Let's see how it goes.

sbrl:
So it just works? We don't need to add the functionality to the upcoming bot for this?

waldyrious:
Not if the way this works is sufficient for our needs :)

PR #1361, opened 2017-04-30
jq: clarify descriptions

waldyrious:
Follow-up of #1092, to fix the question that was unanswered at the time of merging (i.e. what the simplest example does), and improve clarity of other example descriptions.

### DIFF
--- a/pages/common/jq.md
+++ b/pages/common/jq.md
@@ -2,7 +2,7 @@
 
 > A lightweight and flexible command-line JSON processor.
 
-- Output an entire JSON file, in pretty-print format:
+- Output a JSON file, in pretty-print format:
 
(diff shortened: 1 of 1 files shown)


waldyrious:
since @mention-bot is not reacting -- ping @maximpertsov

pages/common/jq.md
```diff
@@ -2,26 +2,26 @@
 
 > A lightweight and flexible command-line JSON processor.
 
-- Output JSON file:
+- Output an entire JSON file, in pretty-print format:
```
agnivade: IMHO "an entire" is unnecessary here. Its implicit.

waldyrious: Yes, agreed.

PR #1365, opened 2017-05-02
whatis: add glob example

sbrl:
Some might find the glob easier to use if they don't understand regular expressions.

### DIFF
--- a/pages/linux/whatis.md
+++ b/pages/linux/whatis.md
@@ -12,8 +12,8 @@
 
 - Display descriptions for all commands matching a glob:
 
-`whatis --wildcard {{command_glob}}`
+`whatis --wildcard {{net*}}`
 
 - Search man page descriptions with a regular expression:
 
-`whatis --regex {{regular_expression}}`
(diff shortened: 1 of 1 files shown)


pages/linux/whatis.md
```diff
@@ -10,6 +10,10 @@
 
 `whatis --long {{command}}`
 
+- Display descriptions for all commands matching a glob:
+
+`whatis --wildcard {{command_glob}}`
```
agnivade: Can we give concrete examples for both glob and regular_expression ? I think it will be easier for users to see something actual.

sbrl: Hrm yes. Suggestions? Perhaps replacing `{{{command_glob}}` with `{{net*}}` or something?

agnivade: Maybe we should differentiate a bit from glob and regex. net* or tcp* as a glob is good. For regex, can use "wish[0-9].[0-9]".

waldyrious: yes, I agree a concrete example is better in this case. I wonder if we should use more common command names for the search patterns instead, though.

PR #1369, opened 2017-05-07
New command: hostnamectl

sbrl:
I've been setting up a raspberry pi just now, and I noticed that `hostnamectl` isn't in tldr-pages! This PR rectifies the problem :P


sbrl:
:D

PR #1372, opened 2017-05-09
testing tldr-bot

agnivade:


agnivade:
Good bots !

waldyrious:
Now if we could only figure out why @mention-bot doesn't seem to be activating consistently... could it be that it only responds to PRs that are *not* initiated by repo owners/collaborators?

/cc @vjeux

waldyrious:
The [default configuration](https://github.com/facebook/mention-bot/blob/a7ed44e/server.js#L146) seems to be `"skipCollaboratorPR": false`, so it doesn't make sense that it would be the (intended) case.

PR #1373, opened 2017-05-09
vimdiff: add page

damianpeterson:


pages/common/vimdiff.md
```diff
@@ -0,0 +1,24 @@
+# vimdiff
+
+> Open up to four files in Vim and show their differences.
+> See Vim documentation for how to edit and navigate within a window.
+
+- Open files and show the differences:
```
waldyrious: Just for completeness, I'd write "and show the differences between them".

pages/common/vimdiff.md
```diff
@@ -0,0 +1,24 @@
...
+
+`vimdiff {{file1}} {{file2}}`
+
+- Use horizontal window split instead of the default vertical split:
```
waldyrious: This is just for the 2-file case, right? If so, maybe we should mention it (e.g. "Open two files with a horizontal...")

pages/common/vimdiff.md
```diff
@@ -0,0 +1,24 @@
...
+
+`vimdiff -o {{file1}} {{file2}}`
+
+- Close current window without saving changes:
```
agnivade: Do we have to mention `:q!` and `:w` ? I think these should remain in the vim page.

pages/common/vimdiff.md
```diff
@@ -0,0 +1,24 @@
+# vimdiff
+
+> Open up to four files in Vim and show their differences.
+> See Vim documentation for how to edit and navigate within a window.
```
agnivade: s/Vim/`vim`/

pages/common/vimdiff.md
damianpeterson: Yeah not sure about this. The moving between windows command is technically a Vim thing but I felt being able to do so as part of diffing files was key. Same if you're not too familiar with vimdiff and need to get out of there. Thoughts?

agnivade: Yea, the moving between windows is fine. For the quit and save part, that's why I mentioned to put the word "vim" in backticks in the heading section. The node-client will automatically read that and show suggestions for users to try a `tldr vim` too. 

I would suggest you change the description of "for how to edit and navigate within a window" and make it generic enough to include the functionality of `:q!` and `:w`. Therefore, users will do a `tldr vim` when needed.

damianpeterson: oh cool. I didn't know that and didn't notice the backticks. I've removed the `:q!` and `:w` and made the `Ctrl + w` consistent with other's documentation. Thanks :smile:

agnivade: > I would suggest you change the description of "for how to edit and navigate within a window" and make it generic enough to include the functionality of :q! and :w.

Would you like to do this ?

damianpeterson: I'm sorry I'm not clear on what you mean here. Are you saying to mention the ability to use those generic vim commands in that description or to alter the description to allow the reintroduction of those two commands as per previous commits?

agnivade: The first one. I was saying instead of saying "for how to edit and navigate within a window", make it something like 'for working with files and navigating within a window'

damianpeterson: Got you. Done. Thanks for your feedback!

pages/common/vimdiff.md
```diff
@@ -0,0 +1,16 @@
+# vimdiff
+
+> Open up to four files in vim and show their differences between them.
```
waldyrious: their --> the

pages/common/vimdiff.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Open two files and show the differences (up to four files can be compared):
+
+`vimdiff {{file1}} {{file2}}`
+
+- Open two files using a horizontal window split instead of the default vertical split:
+
+`vimdiff -o {{file1}} {{file2}}`
+
+- Move cursor to window left|right|up|down:
```
waldyrious: I'd write this as "Move the cursor to the window on the left, right, top or bottom:" (but I'm open to other phrasings, or arguments to keep the current one).

waldyrious:
Nice one @damianpeterson 👍 I think this is almost ready to merge: there's still a typo to fix, and I suggested a different phrasing to the last example -- let me know what you think.

pages/common/vimdiff.md
damianpeterson: good catch. done.

pages/common/vimdiff.md
damianpeterson: Cool. I've left the pipes in the description though as I feel it matches the pipes in the command example more closely. Happy to change if you feel strongly about it though.

PR #1375, opened 2017-05-13
perl: swap -M example with -0, for multiline regex

waldyrious:
For completeness: the -0 option is actually used to specify the input record separator
(as an octal or hexadecimal number).
Without it, the record separator is the newline character,
i.e. the files are processed line by line
(which doesn't allow find-replace expressions that include newlines to work).

According to the [documentation](http://perldoc.perl.org/perlrun.html#Command-Switches),
using plain `-0` is not guaranteed to have the effect of parsing the entire file,
because if the file does contain characters
with octal value equal to the parameter passed to (or implied by) the -0 option,
these characters will be treated as line breaks.
However, if the value exceeds 377<sub>8</sub> (i.e. 255),
it won't be matched to characters on the file.
777 is the preferred convention within that exceptional range,
as the highest value that keeps to 3 octal digits.

Here we're forgoing such details and using -0 anyway, since for most cases this will be enough.

See previous discussion in #1239.

### DIFF
--- a/pages/common/perl.md
+++ b/pages/common/perl.md
@@ -14,10 +14,6 @@
 
 `perl -e {{perl_statement}}`
 
-- Import module before execution of a perl statement:
-
-`perl -M{{module}} -e {{perl_statement}}`
-
 - Run a Perl script in debug mode, using `perldebug`:
 
 `perl -d {{script.pl}}`
@@ -25,6 +21,10 @@
 - Loo[p] over all lines of a file, editing them [i]n-place using a find/replace [e]xpression:
 
 `perl -p -i -e 's/{{find}}/{{replace}}/g' {{filename}}`
+
+- Run a multi-line find/replace expression on a file (i.e. including line breaks):
+
+`perl -0 -p -i -e 's/{{foo\nbar}}/{{foobar}}/g' {{filename}}`
 
 - Run a find/replace expression on a file, saving the original file with a given extension:
 
--- a/pages/common/perl.md
+++ b/pages/common/perl.md
@@ -22,10 +22,10 @@
 
 `perl -p -i -e 's/{{find}}/{{replace}}/g' {{filename}}`
 
-- Run a multi-line find/replace expression on a file (i.e. including line breaks):
-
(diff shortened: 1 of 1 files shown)


waldyrious:
I'll wait for @morrme to comment on the chosen example and description, before merging :)

morrme:
-p0e works but to fit the consistency of the rest of the file (wow I sound like yall now! 😄 ) ..... maybe the `-e` should be separate ???

waldyrious:
What would you prefer? -p0 -e? or -p -0 -e?

To clarify, I wrote it that way because with so many options it becomes cumbersome to type them all separately, and also because the option combination resembles a word ("poe"), which helps memorizing this invocation.

Ideally we'd do this consistently for the all examples in the page, but unfortunately -e needs to be right before the actual expression, and -i cannot be followed by another letter as that is assumed as the backup extension.

We could reorder the -p -i -e example to something like `-i -pe "..."` and the -0 example to `-0 -pe "..."`, but then we'd lose the "pie" mnemonic...

In the end, we have to compromise on some of these goals (consistency, memorability, typing convenience), keeping in mind the utility for the end user as the benchmark. This is why I chose the current formulation, but I'm happy to consider alternative interpretations of what a reasonable compromise should look like in this case.

agnivade:
ping @morrme

morrme:
@agnivade I thought this was already settled. I prefer the "separated" version but @waldyrious prefers the "combined" version. maintainer wins 😄

waldyrious:
@agnivade and/or @sbrl, perhaps you guys could break the tie then :)

morrme:
aww @waldyrious there's no tie, it's your project, so things  should go the way you want them to go.

agnivade:
Haha .. I like the snug fit of flags too. Too much typing otherwise :stuck_out_tongue:

waldyrious:
Not at all, @morrme -- we are very committed to keeping tldr-pages a community project where nobody's opinions are meant to count more than anyone else's -- the most reasonable arguments should prevail, and if unanimity can't be reached, consensus (i.e. voluntary compromise) among all participants in any given discussion should be used to make decisions 😉

PR #1376, opened 2017-05-14
brew cask: add page

waldyrious:


waldyrious:
`brew cask` is for binary packages, while `brew` normally builds packages from source (the Homebrew project prefers the latter whenever possible)

agnivade:
Oh, so its like `pip install` and `apt-get install`

waldyrious:
yes :)

agnivade:
@waldyrious - Anything blocking this from merging ?

waldyrious:
Nope, I was just waiting to see if another maintainer would do it (so that one of the two approvals wouldn't have to be the author's), but since there's been no objections for over a week, I guess we can just go ahead and merge.

PR #1379, opened 2017-05-16
cmake: add mkdir to the out-of-source build example

waldyrious:
This makes the command slightly more complex, but it better reflects the use case of when the project is first generated.

In any case, the `-p` in `mkdir` ensures the command does not error out if the build folder already exists, so now both cases are supported.

I'm considering a further PR replacing this command by a version using the [-H -B trick](https://www.reddit.com/r/programming/comments/6b9r19/how_to_build_a_cmakebased_project/dhlfh05/), which would change the command to be:

    cmake -H. -B{{build}} && make -C {{build}}

but that can be discussed separately.

### DIFF
--- a/pages/common/cmake.md
+++ b/pages/common/cmake.md
@@ -7,9 +7,9 @@
 
 `cmake && make`
 
-- Generate a Makefile and use it to compile a project in a separate folder (out-of-source build):
+- Generate a Makefile and use it to compile a project in a separate "build" folder (out-of-source build):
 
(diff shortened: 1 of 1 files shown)


waldyrious:
Yeah, me neither :) finding out about it was what prompted this PR, actually ;)

sbrl:
Yep, Looks ok to me! I think that the H, B trick might be worth adding actually.

PR #1382, opened 2017-05-17
new command: prosodyctl

sbrl:
Another somewhat obscure one for you :P

### DIFF
--- a/pages/common/prosodyctl.md
+++ b/pages/common/prosodyctl.md
@@ -1,0 +1,23 @@
+# prosodyctl
+
+> The control tool for the prosody XMPP server.
+
+- Show the status of the prosody server:
+
+`sudo prosodyctl status`
+
+- Ask the prosody server to reload it's configuration files:
+
+`sudo prosodyctl reload`
+
+- Add a user to the XMPP server:
+
+`sudo prosodyctl adduser user@example.com`
+
+- Set a user's password:
+
+`sudo prosodyctl passwd user@example.com`
+
+- Permanently delete a user:
+
+`sudo prosodyctl deluser user@example.com`
--- a/pages/common/prosodyctl.md
+++ b/pages/common/prosodyctl.md
@@ -12,12 +12,12 @@
 
 - Add a user to the Prosody XMPP server:
 
-`sudo prosodyctl adduser user@example.com`
+`sudo prosodyctl adduser {{user@example.com}}`
 
 - Set a user's password:
 
-`sudo prosodyctl passwd user@example.com`
+`sudo prosodyctl passwd {{user@example.com}}`
 
 - Permanently delete a user:
 
-`sudo prosodyctl deluser user@example.com`
+`sudo prosodyctl deluser {{user@example.com}}`
--- a/pages/common/prosodyctl.md
+++ b/pages/common/prosodyctl.md
@@ -1,16 +1,16 @@
 # prosodyctl
 
-> The control tool for the prosody XMPP server.
+> The control tool for the Prosody XMPP server.
 
-- Show the status of the prosody server:
+- Show the status of the Prosody server:
 
 `sudo prosodyctl status`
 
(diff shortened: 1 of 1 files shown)


pages/common/prosodyctl.md
```diff
@@ -0,0 +1,23 @@
+# prosodyctl
+
+> The control tool for the prosody XMPP server.
+
+- Show the status of the prosody server:
+
+`sudo prosodyctl status`
```
waldyrious: Out of curiosity, why does a read-only command require sudo?

pages/common/prosodyctl.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Ask the prosody server to reload it's configuration files:
```
waldyrious: it's --> its

pages/common/prosodyctl.md
```diff
@@ -0,0 +1,23 @@
...
+
+`sudo prosodyctl reload`
+
+- Add a user to the XMPP server:
```
waldyrious: You accidentally the verb :P

pages/common/prosodyctl.md
sbrl: Good question! I'm not sure either, but it appears to be because it needs access to it's configuration file - which is by default stored in a directory that's got the permissions 0750, and is owned by `prosody:prosody`.

pages/common/prosodyctl.md
```diff
@@ -0,0 +1,23 @@
+# prosodyctl
+
+> The control tool for the prosody XMPP server.
```
sbrl: Good thinking - fixed :D

pages/common/prosodyctl.md
sbrl: Hrm. Perhaps!

sbrl:
Thanks for the suggestions, @waldyrious! Is that better?

pages/common/prosodyctl.md
```diff
@@ -0,0 +1,23 @@
...
+
+`sudo prosodyctl status`
+
+- Reload the server's configuration files:
+
+`sudo prosodyctl reload`
+
+- Add a user to the Prosody XMPP server:
+
+`sudo prosodyctl adduser user@example.com`
```
agnivade: email address in tokens

pages/common/prosodyctl.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Set a user's password:
+
+`sudo prosodyctl passwd user@example.com`
+
+- Permanently delete a user:
```
agnivade: Is there anything like temporarily delete a user ? If no, then its simpler as "Delete a user"

waldyrious: Yeah, I had the same thought.

pages/common/prosodyctl.md
waldyrious: Damn, I always forget to check for these! 😅

pages/common/prosodyctl.md
sbrl: Nope, you can only permanently delete one. It deletes all of a user's associated data, too, so I thought it best to be explicit here - there's not way to restore it if you delete someone accidentally!

sbrl: Still, I can change these if needed.

agnivade: Well, I think its implicit that delete will permanently delete unless mentioned specifically. What do you think @waldyrious ?

waldyrious: Yes, it is the default behavior that deleting data is a permanent operation, especially in the command line. The "recycling" behavior is an exception and normally reserved for GUI environments.

And while I understand that it's better to be explicit, in this case it may actually raise doubts about whether non-permanent deletion is available.

That said, IMO the balance between these concerns is thin, so I'm ok with either decision.

PR #1384, opened 2017-05-19
mosh: add page

vivekkrish:
This PR adds example commands for the `mosh` mobile shell (https://mosh.org) application (a replacement to SSH).

Mosh is [primarily available](https://mosh.org/#getting) on MacOS and Linux (various distros) and also on Windows (via Chrome), which is why this document was added to `common`. If you think otherwise, please let me know, and I can duplicate it into `osx` and `linux`, and remove from `common`.

### DIFF
--- a/pages/common/mosh.md
+++ b/pages/common/mosh.md
@@ -1,16 +1,11 @@
 # mosh
 
-> Mobile Shell (`mosh`) is remote terminal application that allows roaming.
-> `mosh` is a more robust and responsive replacement for SSH.
-> Visit <https://mosh.org> for more information.
+> Mobile Shell (`mosh`) is a robust and responsive replacement for SSH.
+> `mosh` persists connections to remote servers while roaming between networks.
 
 - Typical usage:
 
 `mosh {{username}}@{{remote_host}}`
-
-- Usage when `mosh-server` binary is outside standard path:
-
-`mosh --server={{/path/to/bin/}}mosh-server {{remote_host}}`
 
 - Connect to a remote server with a specific identity (private key):
 
@@ -28,6 +23,6 @@
 
 `mosh -p {{124}} {{username}}@{{remote_host}}`
 
-- Disable instant echo:
+- Usage when `mosh-server` binary is outside standard path:
 
-`mosh --predict=never {{remote_host}}`
+`mosh --server={{/path/to/bin/}}mosh-server {{remote_host}}`
--- a/pages/common/mosh.md
+++ b/pages/common/mosh.md
@@ -1,0 +1,33 @@
+# mosh
+
+> Mobile Shell (`mosh`) is remote terminal application that allows roaming.
+> `mosh` is a more robust and responsive replacement for SSH.
+> Visit <https://mosh.org> for more information.
+
+- Typical usage:
+
+`mosh {{username}}@{{remote_host}}`
+
+- Usage when `mosh-server` binary is outside standard path:
+
+`mosh --server={{/path/to/bin/}}mosh-server {{remote_host}}`
+
+- Connect to a remote server with a specific identity (private key):
+
+`mosh --ssh="ssh -i {{/path/to/key_file}}" {{username}}@{{remote_host}}`
+
+- Connect to a remote server using a specific port:
+
+`mosh --ssh="ssh -p {{2222}}" {{username}}@{{remote_host}}`
+
+- Run a command on a remote server:
+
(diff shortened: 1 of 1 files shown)


pages/common/mosh.md
```diff
@@ -0,0 +1,33 @@
+# mosh
+
+> Mobile Shell (`mosh`) is remote terminal application that allows roaming.
+> `mosh` is a more robust and responsive replacement for SSH.
+> Visit <https://mosh.org> for more information.
```
agnivade: We always restrict the heading to 2 lines. Either delete this line or the one above this. Or merge them to a single line.

pages/common/mosh.md
```diff
@@ -0,0 +1,33 @@
...
+
+- Typical usage:
+
+`mosh {{username}}@{{remote_host}}`
+
+- Usage when `mosh-server` binary is outside standard path:
```
agnivade: I am wondering whether this is a common enough use-case for a beginner ?

pages/common/mosh.md
```diff
@@ -0,0 +1,33 @@
...
+
+- Run a command on a remote server:
+
+`mosh {{remote_host}} -- {{command -with -flags}}`
+
+- Select Mosh UDP port (useful when `{{remote_host}}` is behind a NAT):
+
+`mosh -p {{124}} {{username}}@{{remote_host}}`
+
+- Disable instant echo:
```
agnivade: What does "instant echo" mean ?

pages/common/mosh.md
waldyrious: I'd suggest removing the last line. We have plans for including links and other metadata in the future, where this would fit.

pages/common/mosh.md
```diff
@@ -0,0 +1,33 @@
+# mosh
+
+> Mobile Shell (`mosh`) is remote terminal application that allows roaming.
+> `mosh` is a more robust and responsive replacement for SSH.
+> Visit <https://mosh.org> for more information.
+
+- Typical usage:
```
waldyrious: We tend to prefer descriptions that actually describe what the command *does*. In this case, for example, "Connect to a remote server" would probably work.

pages/common/mosh.md
vivekkrish: Thanks for the review! I've pushed a new commit 0425d8b which reflects changes as suggested.

pages/common/mosh.md
vivekkrish: This is definitely a good use-case, since `mosh` (both the server and client components) can be installed by an unprivileged user.

pages/common/mosh.md
vivekkrish: `mosh` operates in a manner slightly different from SSH: SSH normally waits for the server to respond before it shows you what you've typed.

But `mosh` instantly echos back your typing, and does it in an adaptive manner, so as to compensate for any network issues (slowness, etc.). This makes the interface more responsive, even on high latency connections.

I've removed this example command, since the default setting (of echoing immediately) works best in most cases.

pages/common/mosh.md
vivekkrish: Thank you for the suggestion. Please see 0425d8b for the modified description.

agnivade: @vivekkrish - looks like you missed this one.

vivekkrish: Hi @agnivade, I thought that the sentence "`mosh` persists connections to remote servers while roaming between networks." addresses the point brought up by @waldyrious, while also promoting the capability of `mosh` to seamlessly persist connections to the remote server even when local connectivity is affected (e.g. loss of internet connectivity).

Would you suggest that we alter this statement any further?

Thank you!

waldyrious: I'm not sure I was clear. I meant that "Typical usage" does not really fit the format of descriptions we use in the project. We try to have every example stand alone, rather than depending on previous content of the page. Does that make sense to you?

vivekkrish: Ah, oops! I confused this comment with the one about removing the link: https://github.com/tldr-pages/tldr/pull/1384/files/ff9e94078b1ade9daa585b3a4722e4879aebc3aa#diff-32fbf11af0de15b9a5933f4ffe23e376

Sorry! I will make the change.

waldyrious: No problem :)

vivekkrish:
Thank you!

PR #1386, opened 2017-05-22
cmake: use -H -B trick for out-of-source build

waldyrious:
As initially discussed in #1379. This removes the need to use `mkdir` and `cd`.

Inspired by this post:
https://www.reddit.com/r/programming/comments/6b9r19/how_to_build_a_cmakebased_project/dhlfh05/


agnivade:
Merging this as no one has anything against it.

PR #1387, opened 2017-05-23
brew: expand example descriptions

waldyrious:


waldyrious:
Thanks for the reviews :)

PR #1388, opened 2017-05-23
quota: added example of using human readable output

kalebo:
Because `quota` has an odd syntax for printing sizes in human readable units, i.e, `-s` and not `-h` like in other commands, I think it should be included in the tldr.

### DIFF
--- a/pages/common/quota.md
+++ b/pages/common/quota.md
@@ -2,9 +2,9 @@
 
 > Display users' disk space usage and allocated limits.
 
-- Show disk quotas for the current user:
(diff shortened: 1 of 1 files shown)


waldyrious:
@kalebo LGTM, but what does -s stand for?

agnivade:
`-s` stands for the human-readable option. @kalebo has mentioned this in the top comment. :smile:

waldyrious:
Haha I got that :) I meant what word is "s" an abbreviation of.

sbrl:
@waldyrious I'm not sure! I'd love to know, though.

PR #1389, opened 2017-05-24
gist: add page

thalesmello:
It has five examples.

### DIFF
--- a/pages/common/gist.md
+++ b/pages/common/gist.md
@@ -2,17 +2,17 @@
 
 > Upload code to https://gist.github.com.
 
-- Upload any number of files:
+- Create a gist from any number of text files:
 
 `gist {{file.txt}} {{file2.txt}}`
 
-- Upload a private file with a description:
+- Create a private gist with a description:
 
-`gist -p {{file.txt}} -d {{"A meaningful description"}}`
+`gist -p -d {{"A meaningful description"}} {{file.txt}} `
 
-- Read contents from STDIN with a name:
+- Read contents from STDIN and create a gist from it:
 
-`{{echo "hello world"}} | gist -f {{name.rb}}`
+`{{echo "hello world"}} | gist`
 
 - List your public and private gists:
 
--- a/pages/common/gist.md
+++ b/pages/common/gist.md
@@ -18,6 +18,6 @@
 
 `gist -l`
 
-- Update existing gists with the `GIST_ID` in the URL:
+- Use the id from the gist URL to modify or include a file:
 
 `gist -u {{GIST_ID}} {{file.txt}}`
--- a/pages/common/gist.md
+++ b/pages/common/gist.md
@@ -1,0 +1,23 @@
+# gist
+
+> Upload code to https://gist.github.com
+
+- Upload any number of files:
+
+`gist {{file.txt}} {{file2.txt}}`
+
+- Upload a private file with a description:
+
+`gist -p {{file.txt}} -d {{"A meaningful description"}}`
+
+- Read contents from STDIN with a name:
+
+`{{echo hello world}} | gist -f {{name.rb}}`
+
+- List public and private gists for authed user
+
+`gist -l`
+
+- Update existing gists with the `GIST_ID` in the URL:
+
+`gist -u {{GIST_ID}} {{file.txt}}`
--- a/pages/common/gist.md
+++ b/pages/common/gist.md
@@ -1,6 +1,6 @@
 # gist
 
-> Upload code to https://gist.github.com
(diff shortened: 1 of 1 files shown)


pages/common/gist.md
```diff
@@ -0,0 +1,23 @@
...
+
+`gist {{file.txt}} {{file2.txt}}`
+
+- Upload a private file with a description:
+
+`gist -p {{file.txt}} -d {{"A meaningful description"}}`
+
+- Read contents from STDIN with a name:
+
+`{{echo hello world}} | gist -f {{name.rb}}`
```
agnivade: Might be better to wrap hello world in quotes.

pages/common/gist.md
```diff
@@ -0,0 +1,23 @@
...
+
+- List public and private gists for authed user:
```
agnivade: Shouldn't the user be authenticated for all the commands ? Is there any benefit of saying "authed user" ?

pages/common/gist.md
```diff
@@ -0,0 +1,23 @@
+# gist
+
+> Upload code to https://gist.github.com.
+
+- Upload any number of files:
```
waldyrious: These can only be text files, right? Would "Upload text files and create a gist out of them" be accurate, or does is a separate gist created for each file?

pages/common/gist.md
```diff
@@ -0,0 +1,23 @@
...
+
+`gist {{file.txt}} {{file2.txt}}`
+
+- Upload a private file with a description:
```
waldyrious: How about "Upload a file as a private gist with a description"?

pages/common/gist.md
```diff
@@ -0,0 +1,23 @@
...
+
+`gist -p {{file.txt}} -d {{"A meaningful description"}}`
+
+- Read contents from STDIN with a name:
```
waldyrious: I think "Read contents from STDIN and create a gist from it, with a name:" could be more accurate.

pages/common/gist.md
```diff
@@ -0,0 +1,23 @@
...
+
+`{{echo hello world}} | gist -f {{name.rb}}`
+
+- List public and private gists for authed user:
+
+`gist -l`
+
+- Update existing gists with the `GIST_ID` in the URL:
```
waldyrious: Does this replace all files in the gist, or only the file with the same name (if any)? And in the latter case, would the new file be appended in case it doesn't exist on the gist yet?

pages/common/gist.md
sbrl: I don't think you have to be authed to pull gists down. Also, "private" should be "secret" according to the wording on gist.github.com

pages/common/gist.md
sbrl: "private" -> "secret" - see above

pages/common/gist.md
sbrl: Oooh good questions. I think the answers should be made clear in the wording here.

Also, a random fact for you. Gists are full git repos, so you can clone & push them as you would any other repos - but you probably knew that already :P

pages/common/gist.md
thalesmello: I've changed the wording to make it clearer.

Also, by motivation in using "private" rather than "secret" is the `man gist` pages use the "private" term.

http://manpages.ubuntu.com/manpages/xenial/man1/gist-paste.1.html

Becuase the option is also "-p" for private, I supposed it would make it easier for the reader to relate them.

sbrl: Hum, yeah. I didn't bother to read the man page myself :P

sbrl: I don't see the commit though

pages/common/gist.md
thalesmello: I've updated the description to make it clearer.

pages/common/gist.md
thalesmello: Made the description clearer.

pages/common/gist.md
thalesmello: I decided to simply remove the naming part, as it's not mandatory.

pages/common/gist.md
thalesmello: @waldyrious I don't know. Let me test.

thalesmello: I've updated the description so it's clearer.

pages/common/gist.md
thalesmello: @sbrl Sorry. I commented before I pushed. It should be available now.

waldyrious:
Thanks @thalesmello! The final version of the page is quite nice :)

PR #1394, opened 2017-05-26
keepass2: add page

sbrl:
Another command! There's more that you could do with it, but this is a good starting point.

### DIFF
--- a/pages/common/keepass2.md
+++ b/pages/common/keepass2.md
@@ -8,8 +8,8 @@
 
 - Start KeePass 2, opening a specific password database:
 
-`keepass2 "{{/path/to/database.kbdx}}"`
+`keepass2 {{path/to/database.kbdx}}
 
 - Use a specific key file to open a password database:
 
-`keepass2 "{{/path/to/database.kbdx}}" -keyfile:{{/path/to/key/file.key}}`
+`keepass2 {{path/to/database.kbdx}} -keyfile:{{path/to/key/file.key}}`
--- a/pages/common/keepass2.md
+++ b/pages/common/keepass2.md
@@ -10,6 +10,6 @@
 
 `keepass2 "{{/path/to/database.kbdx}}"`
 
-- Use a specific key file to open a password database: 
+- Use a specific key file to open a password database:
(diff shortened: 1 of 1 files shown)


pages/common/keepass2.md
```diff
@@ -0,0 +1,15 @@
+# keepass2
+
+> A free, open source, light-weight and easy-to-use password manager.
```
waldyrious: FOSS is arguably irrelevant (most of the commands we cover are); "light-weight" is acceptable; but "easy-to-use" is too subjective and opinionated. I'd suggest going with a plain, objective description -- `A light-weight password manager` would work.

waldyrious:
LGTM apart from the description change I suggested inline.

sbrl:
Ah, right. I just lifted that description from the KeePass website

sbrl:
There - is that better?

pages/common/keepass2.md
```diff
@@ -0,0 +1,15 @@
...
+
+`keepass2`
+
+- Start KeePass 2, opening a specific password database:
+
+`keepass2 "{{/path/to/database.kbdx}}"`
+
+- Use a specific key file to open a password database:
+
+`keepass2 "{{/path/to/database.kbdx}}" -keyfile:{{/path/to/key/file.key}}`
```
agnivade: Shouldn't there be quotes around the key file path too ?

And how is the key file generated for a particular db ?

waldyrious: Or neither. We don't usually quote paths. Is it particularly necessary in this case?

By the way, do the paths need to be absolute? If not, maybe remove the leading slash.

sbrl: > Shouldn't there be quotes around the key file path too ?

Nope

> And how is the key file generated for a particular db ?

It's created when you first set up a database,  I think. I don't use it personally.

> Or neither. We don't usually quote paths. Is it particularly necessary in this case?

Nope! This is just my personal preference.

> By the way, do the paths need to be absolute? If not, maybe remove the leading slash.

No again!

PR #1395, opened 2017-06-05
fping: add page

jsonbruce:
A new command !

`fping` is a more powerful ping

### DIFF
--- a/pages/common/fping.md
+++ b/pages/common/fping.md
@@ -1,15 +1,15 @@
 # fping
 
-> A more powerful ping which ping multiple hosts
+> A more powerful ping which can ping multiple hosts.
 
-- List alive host within subnet via netmask
+- List alive hosts within a given subnet using a netmask:
 
 `fping -a -g 192.168.1.0/24`
 
-- List alive host within subnet via ip range
+- List alive hosts within a given subnet using a IP range:
 
 `fping -a -g 192.168.1.1 192.168.1.254`
 
-- List unreachable host within subnet
+- List unreachable hosts within a given subnet:
 
 `fping -u -g 192.168.1.0/24`
--- a/pages/common/fping.md
+++ b/pages/common/fping.md
@@ -1,0 +1,15 @@
+# fping
+
+> A more powerful ping which ping multiple hosts
+
+- List alive host within subnet via netmask
+
+`fping -a -g 192.168.1.0/24`
+
+- List alive host within subnet via ip range
+
+`fping -a -g 192.168.1.1 192.168.1.254`
+
+- List unreachable host within subnet
+
+`fping -u -g 192.168.1.0/24`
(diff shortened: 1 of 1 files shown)


pages/common/fping.md
```diff
@@ -0,0 +1,15 @@
+# fping
+
+> A more powerful ping which ping multiple hosts
```
waldyrious: "which ping" --> "which can ping". Also, you need a period at the end of the description.

pages/common/fping.md
```diff
@@ -0,0 +1,15 @@
...
+
+- List alive host within subnet via netmask
```
waldyrious: "host" --> "hosts"; I'd also suggest "via" --> "using a", which I think is clearer. By the way, does the `-g` stand for any word in particular?

pages/common/fping.md
```diff
@@ -0,0 +1,15 @@
...
+
+`fping -a -g 192.168.1.0/24`
+
+- List alive host within subnet via ip range
```
waldyrious: Pluralize "host" here as well, and make "ip" uppercase. "Via" could also be replaced by "using a".

pages/common/fping.md
```diff
@@ -0,0 +1,15 @@
...
+
+`fping -a -g 192.168.1.1 192.168.1.254`
+
+- List unreachable host within subnet
```
waldyrious: "hosts", as mentioned above.

jsonbruce: Thanks, I have pushed a new commit. By the way, `g` for Generate. @waldyrious

waldyrious:
Thanks for the quick reaction! LGTM 👍

pages/common/fping.md
```diff
@@ -0,0 +1,15 @@
+# fping
+
+> A more powerful ping which can ping multiple hosts.
+
+- List alive hosts within a given subnet using a netmask:
+
+`fping -a -g 192.168.1.0/24`
```
agnivade: What does `-g` mean ? Any way we can include it in the description ?

sbrl: "Generate list"?

jsonbruce: Yes, `g` for generate ! @agnivade 

Actually I have thought including generate in the description but I didn't get a comfortable way,
So, How do think the below

- List alive hosts within a **generated** subnet using a netmask

agnivade: How about `List alive hosts within a subnet generated using a netmask` ?

sbrl: Or even "Generate a list of alive hosts...."?

waldyrious: @sbrl the command does not generate the list of alive hosts: it generates a subnet, tests all hosts within it for aliveness, and returns the list of alive ones. I would rather go with something closer to @agnivade's proposal: `List alive hosts within a subnet generated from a netmask`.

PR #1396, opened 2017-06-05
Update curl.md

thomasfangyi:
Edit the description of send data to disambiguate.

Add extra text and mark to mark clearly what is data and what is header.

### DIFF
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -19,13 +19,13 @@
 
 `curl -d {{'name=bob'}} {{http://example.com/form}}`
 
-- Send a request with an extra [H]eader:
+- Send a request with an extra [H]eader using a custom HTTP method:
 
-`curl -H {{'X-My-Header: 123'}} {{http://example.com}}`
+`curl -H {{'X-My-Header: 123'}} -X {{PUT}} {{http://example.com}}`
 
-- Send [d]ata in JSON format, specifying a custom HTTP method, and including an extra [H]eader:
+- Send [d]ata in JSON format, specifying the appropriate content-type [H]eader:
 
-`curl -d {{'{"name":"bob"}'}} -X {{PUT}} -H {{'Content-Type: application/json'}} {{http://example.com/users/1234}}`
+`curl -d {{'{"name":"bob"}'}} -H {{'Content-Type: application/json'}} {{http://example.com/users/1234}}`
 
 - Pass a user name and password for server authentication and show headers info only:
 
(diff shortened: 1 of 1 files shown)


waldyrious:
IIUC, the data doesn't have to be in JSON format, and the header doesn't have to be the content-type one, right? If that's indeed the case, then our usual format is to just mention what parameters each example uses, not the values of those parameters (which should be evident if we manage to make it clear what option key corresponds to what parameter in the description).

thomasfangyi:
The intention of this edit is to make clear what is a header and what is the data. I personally find it useful to have an entry that tells me how to send JSON data as this is widely used. I'm afraid I didn't make my intention clear.

pages/common/curl.md
```diff
@@ -15,11 +15,15 @@
...
 `curl -d {{'name=bob'}} {{http://example.com/form}}`
 
-- Send data, specifying a custom HTTP method, and including an extra header:
+- Send a request with an extra [H]eader:
+
+`curl -H {{'X-My-Header: 123'}} {{http://example.com}}`
+
+- Send [d]ata in JSON format, specifying a custom HTTP method, and including an extra [H]eader:
 
 `curl -d {{'{"name":"bob"}'}} -X {{PUT}} -H {{'Content-Type: application/json'}} {{http://example.com/users/1234}}`
```
waldyrious: Perhaps it would be better to simplify this example to only use -d and -H, and change the description to something like:

`- Send [d]ata in JSON format, specifying the appropriate content-type [H]eader:`

The `-X` could be added to the new example you introduced, since it's more generic.

What do you think?

pages/common/curl.md
```diff
@@ -15,11 +15,15 @@
 
 `curl -O -L -C - {{http://example.com/filename}}`
 
-- Send form-encoded data (POST request of type application/x-www-form-urlencoded):
+- Send form-encoded [d]ata (POST request of type application/x-www-form-urlencoded):
```
waldyrious: It's not necessary to use brackets to identify options if they're sufficiently obvious. In this case, since there's only one option, I think it's reasonable to expect people to figure out what it means without adding visual noise to the description. But I'm open to consider other opinions.

pages/common/curl.md
thomasfangyi: Yeah, that makes it clearer, shall I separate the use of `-X` in a separate entry as well besides that?

waldyrious: No, because that would exceed our [maximum number of recommended examples](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines), which is 8.

waldyrious:
Thanks for the contribution and patience with the review process, @thomasfangyi :)

PR #1398, opened 2017-06-06
make: add multithreading example

sbrl:
Personally, I find myself always forgetting the letter to run jobs in parallel (I don't use it often enough! :P).


sbrl:
@agnivade I know, right! There should be a way to enable it by default! Perhaps I should alias `make` in my `.bash_aliases`.

PR #1399, opened 2017-06-06
badblocks: add page

sbrl:
I hope you never have to use this command. If you do, then hopefully this man page will make the experience a bit less stressful.

### DIFF
--- a/pages/common/badblocks.md
+++ b/pages/common/badblocks.md
@@ -6,10 +6,10 @@
 
 `sudo badblocks {{/dev/sda}}`
 
-- Search an unmounted disk for bad blocks with a non-descructive read-write test:
+- Search an unmounted disk for bad blocks with a non-destcructive read-write test:
 
 `sudo badblocks -n {{/dev/sda}}`
 
-- Search an unmounted disk fo r bad blocks with a DESTRUCTIVE write test:
+- Search an unmounted disk for bad blocks with a destructive write test:
 
 `sudo badblocks -w {{/dev/sda}}`
--- a/pages/common/badblocks.md
+++ b/pages/common/badblocks.md
@@ -1,7 +1,7 @@
 # badblocks
 
 > Search a device for bad blocks.
-> Some usages of badblocks can cause destructive actions, such as erasing all the data on a disk.
+> Some usages of badblocks can cause destructive actions, such as erasing all the data on a disk, including the partition table.
 
 - Search a disk for bad blocks by using a non-destructive read-only test:
 
--- a/pages/common/badblocks.md
+++ b/pages/common/badblocks.md
@@ -3,7 +3,7 @@
 > Search a device for bad blocks.
 > Some usages of `badblocks` can be destructive! This means it can erase all the data on a disk!
 
-- Search a disk for bad blocks by using a non-destructive read only test:
+- Search a disk for bad blocks by using a non-destructive read-only test:
 
 `sudo badblocks {{/dev/sda}}`
 
--- a/pages/common/badblocks.md
+++ b/pages/common/badblocks.md
@@ -1,6 +1,7 @@
 # badblocks
 
 > Search a device for bad blocks.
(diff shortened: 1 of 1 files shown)


pages/common/badblocks.md
```diff
@@ -0,0 +1,15 @@
+# badblocks
+
+> Search a device for bad blocks.
+
+- Search a disk for bad blocks by using a non-destructive read-only test:
+
+`sudo badblocks {{/dev/sda}}`
+
+- Search an unmounted disk for bad blocks with a non-descructive read-write test:
```
agnivade: descructive => typo again.

Are you deliberately trying to sneak in typos just to see whether we indeed review things or not ? :stuck_out_tongue_winking_eye:

pages/common/badblocks.md
```diff
@@ -0,0 +1,15 @@
...
+
+`sudo badblocks -n {{/dev/sda}}`
+
+- Search an unmounted disk fo r bad blocks with a DESTRUCTIVE write test:
```
agnivade: space between "for". And I think its okay to keep destructive in normal case.

pages/common/badblocks.md
sbrl: Nope! They are completely accidental :P

sbrl:
Fixed! Thanks, @agnivade

agnivade:
Umm, you replaced one typo with another - `non-destcructive` :laughing: 

And I think you missed my main comment in the review - 
> It would be good if you can describe what destructive means in the 2nd line of main heading.

sbrl:
@agnivade Ah, oops! I'll fix that. And sorry I missed that! Which bit are you referring to? The bit that starts with the `>`, or the line 1 down from there?

agnivade:
Yes, I was saying add a second line with > and explain what destructive means.

sbrl:
Umm what?

agnivade:
You are ending with a '!'. You have to end with a period.

sbrl:
@agnivade Awww, man! I really thought that it was appropriate in this case......

waldyrious:
I agree, @agnivade. As a project built upon collaborative authoring, it's really important that we retain a neutral language and writing style. For instance, Wikipedia [recommends](https://en.wikipedia.org/wiki/Wikipedia:Writing_better_articles#Information_style_and_tone) that the tone should "always remain formal, impersonal, and dispassionate."

I'd suggest rewording the second sentence in a purely descriptive manner, e.g. "Some usages of `badblocks` can cause destructive actions, such as erasing all the data on a disk."

sbrl:
On the contrary, I think there should be an exception here, since this particular command can destroy _all_ the data on an entire hard drive - _including the partition table_. When so much is at stake and could potentially go wrong, does it not warrant an extra exclamation mark? :P

I sense that I'm out-voted here, but I decided to make my point anyway. I'll change it if you _really_ think it should be, but for reasons outlined above I think this really ought to be an exception.

agnivade:
@sbrl - We can include the text "including the partition table", but I agree with @waldyrious that we should remain formal and dispassionate.

sbrl:
@agnivade Fiiiine - I've changed it :wink: :stuck

waldyrious:
@sbrl please keep standing up for the directions you feel are warranted for the project even if that means going against other contributors' views :).

In this case, I do understand your reasoning, but note that we do use neutral, descriptive language in all other cases (that I'm aware of), including in potentially dangerous commands, such as the infamous `rm -rf`, `git reset --hard`, etc. I searched the repo for instances of "dangerous", "careful", "attention", so I'm pretty confident we've been sticking to this implicit guideline so far.

That said, I did find a single instance of a "dangerous!", in [`umount -a`](https://github.com/tldr-pages/tldr/blob/master/pages/common/umount.md). I'm proposing to remove (or at least clarify) that note in #1408. We can certainly add the "including the partition table" note to the caveat in this command's description, but I don't feel it's right for us to make the decision about the level of seriousness of that effect on behalf of the user.

PR #1400, opened 2017-06-06
isoinfo: add page

sbrl:
Another strange one! You probably don't need this. Mounting isos instead is much more convenient! Still, it's useful for scripts, I think.

### DIFF
--- a/pages/linux/isoinfo.md
+++ b/pages/linux/isoinfo.md
@@ -1,0 +1,15 @@
+# isoinfo
+
+> Utility programs for dumping and verifying iso images.
+
+- List all the files packed into an iso:
+
+`isoinfo -f -i {{path/to/image.iso}}`
+
+- Extract a specific file to stdout:
+
+`isoinfo -i {{path/to/image.iso}} -x {{/PATH/TO/FILE/INSIDE/ISO.EXT}}`
+
+- Show header information for an iso:
+
+`isoinfo -d -i {{path/to/image.iso}}`
--- a/pages/linux/isoinfo.md
+++ b/pages/linux/isoinfo.md
@@ -1,15 +1,15 @@
 # isoinfo
 
-> Utility programs for dumping and verifying iso images.
+> Utility programs for dumping and verifying ISO disk images.
 
-- List all the files packed into an iso:
+- List all the files included in an ISO image:
 
 `isoinfo -f -i {{path/to/image.iso}}`
 
-- Extract a specific file to stdout:
+- E[x]tract a specific file from an ISO image and send it out stdout:
 
 `isoinfo -i {{path/to/image.iso}} -x {{/PATH/TO/FILE/INSIDE/ISO.EXT}}`
 
-- Show header information for an iso:
+- Show header information for an ISO disk image:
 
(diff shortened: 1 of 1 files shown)


pages/linux/isoinfo.md
```diff
@@ -0,0 +1,15 @@
+# isoinfo
+
+> Utility programs for dumping and verifying iso images.
```
waldyrious: maybe use "ISO disk images" here.

pages/linux/isoinfo.md
```diff
@@ -0,0 +1,15 @@
...
+
+- List all the files packed into an iso:
```
waldyrious: Suggestion: "into an ISO image"

pages/linux/isoinfo.md
```diff
@@ -0,0 +1,15 @@
...
+
+`isoinfo -f -i {{path/to/image.iso}}`
+
+- Extract a specific file to stdout:
```
waldyrious: How about "Extract a specific file from an ISO image and print it to stdout"?

pages/linux/isoinfo.md
```diff
@@ -0,0 +1,15 @@
...
+
+`isoinfo -i {{path/to/image.iso}} -x {{/PATH/TO/FILE/INSIDE/ISO.EXT}}`
+
+- Show header information for an iso:
```
waldyrious: "for an ISO disk image"

pages/linux/isoinfo.md
waldyrious: Or even clearer: "included in an ISO image"

pages/linux/isoinfo.md
agnivade: Good place to do E[x]tract ?

pages/linux/isoinfo.md
```diff
@@ -0,0 +1,15 @@
...
+
+> Utility programs for dumping and verifying iso images.
+
+- List all the files packed into an iso:
+
+`isoinfo -f -i {{path/to/image.iso}}`
+
+- Extract a specific file to stdout:
+
+`isoinfo -i {{path/to/image.iso}} -x {{/PATH/TO/FILE/INSIDE/ISO.EXT}}`
```
agnivade: What path to iso in caps ?

pages/linux/isoinfo.md
sbrl: Yeah, that might work. "print" implies text though, and this command supports binary files.

pages/linux/isoinfo.md
sbrl: It's part of the ISO disk image standard I think. All the ones I inspected also had `;1` appended to them too, which was essential when using this command I found.

pages/linux/isoinfo.md
sbrl: Good idea!

sbrl:
Thanks for the suggestions! Did I get everything?

pages/linux/isoinfo.md
agnivade: okk

PR #1401, opened 2017-06-07
nasm: add page

gttotev:
Add a page for nasm, the Netwide Assembler, a popular assembler for x86 machines.

### DIFF
--- a/pages/common/nasm.md
+++ b/pages/common/nasm.md
@@ -18,6 +18,6 @@
 
 `nasm -l {{list_file}} {{source.asm}}`
 
-- Add a directory to the include file search path before assembling:
+- Add a directory (written with trailing slash) to the include file search path before assembling:
 
-`nasm -i {{/path/to/include_dir}} {{source.asm}}`
+`nasm -i {{/path/to/include_dir/}} {{source.asm}}`
--- a/pages/common/nasm.md
+++ b/pages/common/nasm.md
@@ -1,0 +1,23 @@
+# nasm
+
+> The Netwide Assembler, a portable 80x86 assembler.
+
+- Assemble raw binary file `source` from `source.asm`:
+
+`nasm {{source.asm}}`
+
+- Assemble and generate an assembly listing file:
+
+`nasm -l {{list_file}} {{source.asm}}`
+
+- Add a directory to the include file search path before assembling:
+
+`nasm -i {{/path/to/include_dir}} {{source.asm}}`
+
+- Assemble binary file `output_file` of the specified format:
+
+`nasm -f {{format}} {{source.asm}} -o {{output_file}}`
+
+- View basic nasm help and valid output formats:
+
+`nasm -hf`
--- a/pages/common/nasm.md
+++ b/pages/common/nasm.md
@@ -6,6 +6,14 @@
 
 `nasm {{source.asm}}`
 
+- Assemble binary file `output_file` of the specified format:
+
+`nasm -f {{format}} {{source.asm}} -o {{output_file}}`
+
+- View valid output formats (along with basic nasm help):
+
+`nasm -hf`
+
 - Assemble and generate an assembly listing file:
 
 `nasm -l {{list_file}} {{source.asm}}`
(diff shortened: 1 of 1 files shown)


pages/common/nasm.md
```diff
@@ -0,0 +1,23 @@
...
+
+`nasm -l {{list_file}} {{source.asm}}`
+
+- Add a directory to the include file search path before assembling:
+
+`nasm -i {{/path/to/include_dir}} {{source.asm}}`
+
+- Assemble binary file `output_file` of the specified format:
+
+`nasm -f {{format}} {{source.asm}} -o {{output_file}}`
```
waldyrious: I would put this as the second example, since the "source" in the output filename of the first example threw me off for a second there. But I'm open to consider reasons why it makes more sense to place this example here.

waldyrious:
Very nice submission indeed, @gttotev, pretty much flawless :) I just left a minor inline comment.

pages/common/nasm.md
```diff
@@ -0,0 +1,23 @@
+# nasm
+
+> The Netwide Assembler, a portable 80x86 assembler.
+
+- Assemble raw binary file `source` from `source.asm`:
+
+`nasm {{source.asm}}`
+
+- Assemble and generate an assembly listing file:
```
sbrl: Assemble what? I assume a raw binary file

pages/common/nasm.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Add a directory to the include file search path before assembling:
+
+`nasm -i {{/path/to/include_dir}} {{source.asm}}`
+
+- Assemble binary file `output_file` of the specified format:
+
+`nasm -f {{format}} {{source.asm}} -o {{output_file}}`
+
+- View basic nasm help and valid output formats:
```
sbrl: this feels like it should be nearer the beginning.

pages/common/nasm.md
agnivade: Assemble an assembly file :) i.e. a program written in assembly language. And generate a listing file from it.

@gttotev - please correct me if I am wrong.

pages/common/nasm.md
gttotev: Yes, it makes more sense to place it second. I'll commit this change soon.

pages/common/nasm.md
gttotev: You're perfectly correct @agnivade, nasm assembles a binary file from the instructions in `source.asm` and additionally creates a listing file. Should I make this clearer in the example text?

pages/common/nasm.md
gttotev: Yes, I'll reorganize the examples a bit.

pages/common/nasm.md
```diff
@@ -0,0 +1,23 @@
+# nasm
+
+> The Netwide Assembler, a portable 80x86 assembler.
+
+- Assemble raw binary file `source` from `source.asm`:
+
+`nasm {{source.asm}}`
+
+- Assemble binary file `output_file` of the specified format:
```
waldyrious: It's confusing to use "raw binary file" in the first example and just "binary file" in the second, since this suggests that they might be different. I'd prefer if we used the same description in both cases.

gttotev: In the first example, the format of the binary file assembled is format raw. I'll make this explicitly clear.

waldyrious:
@gttotev the examples look good to merge already, but I'm not sure about the reordering. I would suggest the following as a natural sequence and wording for the first three commands:

>- Assemble `source.asm` into a binary file `source`, in the (default) raw binary format:
>
>`nasm {{source.asm}}`
>
>- List valid output formats (along with basic nasm help):
>
>`nasm -hf`
>
>- Assemble `source.asm` into a binary file `output_file`, in the specified format:
>
>`nasm -f {{format}} {{source.asm}} -o {{output_file}}`

The idea is to start with the simplest invocation, and build up from there. What do you think?

gttotev:
Yes, thank you, this is clear and instantly understandable wording I was aiming for! But, as for the ordering, I would put the example for listing the formats after the one for specifying a format. I think it would help to understand how to use formats before being able to list them out.

waldyrious:
Sure, that's reasonable as well.

waldyrious:
Thanks, let's :shipit: !

waldyrious:
(I'll do a squash merge to avoid this many commits for a single page, although I really appreciate the descriptive commit messages!)

gttotev:
Thanks! This is a really neat project and I'm glad I could help. I'll see what other commands I can contribute.

PR #1402, opened 2017-06-07
git-config: overhaul page

waldyrious:
- expand description of local and global options
- adjust example descriptions for clarity
- change the descriptions of the get/set/remove examples to be more generic 
- change the alias example to a more useful one

### DIFF
--- a/pages/common/git-config.md
+++ b/pages/common/git-config.md
@@ -3,15 +3,15 @@
 > Manage custom configuration options for git repositories.
 > These configurations can be local (for the current repository) or global (for the current user).
 
-- List only local configurations (stored in `.git/config` in the current repository):
+- List only local configuration entries (stored in `.git/config` in the current repository):
 
 `git config --list --local`
 
-- List only global configurations (stored in `~/.gitconfig`):
+- List only global configuration entries (stored in `~/.gitconfig`):
 
 `git config --list --global`
 
-- List all configurations that have been defined either locally or globally:
+- List all configuration entries that have been defined either locally or globally:
 
 `git config --list`
 
--- a/pages/common/git-config.md
+++ b/pages/common/git-config.md
@@ -1,27 +1,28 @@
 # git config
 
-> Get and set repository or global options.
+> Manage custom configuration options for git repositories.
+> These configurations can be local (for the current repository) or global (for the current user).
 
-- Print list of options for current repository:
+- List only local configurations (stored in `.git/config` in the current repository):
 
 `git config --list --local`
 
-- Print global list of options, set in ~/.gitconfig:
+- List only global configurations (stored in `~/.gitconfig`):
 
 `git config --list --global`
 
-- Get full list of options:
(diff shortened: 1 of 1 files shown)


pages/common/git-config.md
```diff
@@ -1,27 +1,28 @@
...
+`git config alias.unstage`
 
-- Set option alias.ls=status in file ~/.gitconfig:
+- Set the global value of a given configuration entry:
 
-`git config --global alias.ls "status"`
+`git config --global alias.unstage "reset HEAD --"`
 
-- Remove option alias.st from ~/.gitconfig:
+- Unset a global configuration entry (reverts to git's default behavior):
```
sbrl: Could this be more elegant? e.g. `Revert a global configuration entry to it's default`, perhaps? I'm not sure.

pages/common/git-config.md
```diff
@@ -1,27 +1,28 @@
...
 
 `git config --list --local`
 
-- Print global list of options, set in ~/.gitconfig:
+- List only global configurations (stored in `~/.gitconfig`):
 
 `git config --list --global`
 
-- Get full list of options:
+- List all configurations that have been defined either locally or globally:
```
sbrl: `configurations` -> `configuration options`, or `configuration directives`, I think. Perhaps `options`, because of the description.

pages/common/git-config.md
waldyrious: Yes, definitely more elegant :)

pages/common/git-config.md
waldyrious: I don't like calling them "options" because many of them are user-defined, and in particular aliases have both user-specified keys and user-specified values. "Options" feels more like picking from a list, which is definitely is the case for some of the entries, but not all of them.

"Directives" could work, but it's a bit uncommon/fancy -- how about "entries"? This would also be consistent with the other descriptions.

sbrl: Hrm. Not sure on "entries" I think it might fit better though.

sbrl: You are picking from a list of configuration variables though - it's just the values that are different

waldyrious: > You are picking from a list of configuration variables though - it's just the values that are different

Not necessarily -- as I mention above, the entries are in the form `<section>.<key> = <value>`, and for most of them only the value is user-provided (e.g. `color.ui = true` or `merge.tool = meld`), but for aliases, the actual key is user-provided as well (e.g. `alias.st = status`).

waldyrious:
I've updated all example descriptions to use "configuration entry/entries". It's actually more consistent that way, thanks @sbrl :)

sbrl:
@waldyrious Awesome! I think this is ready for merging naow! :smiley_cat:

waldyrious:
I did the squash-merge myself as the commit messages needed some cleanup. In any case, the PR had been approved by two other maintainers already.

sbrl:
:fireworks:

Issue #1117, opened 2016-10-14
Define progression path for contributor to maintainer


waldyrious:
Collecting notes I've put elsewhere:

From https://github.com/tldr-pages/tldr/pull/1048#issuecomment-247834196:

> [We should] ensure resilience of the project over the long term by making the barriers between user, contributor and maintainer as permeable as possible -- case in point: none of the the currently active maintainers were part of the original team. We want this fluid adaptation to continue into the future.

From https://github.com/tldr-pages/tldr/issues/980#issuecomment-235387355:

> One thing we clearly need is a way to handle the natural flow of maintainers: people's availability varies over time, so a hand-picked, small maintainer team won't be resilient in the long-run. I suggest something more permeable, inspired in the model used by Wikipedia: after a contributor has, say, 3 PRs merged, they should automatically get commit rights, and after X more commits they should be made part of the tldr-pages organization. If this could be implemented automatically, even better. That way the project is able to scale naturally, and avoid the stagnation that often plagues to such endeavors once the original team loses steam.

waldyrious:
There are interesting notes in http://docs.brew.sh/#maintainers which we could take some inspiration from. But let's try making the whole thing lighter (i.e. less documents to read, and more automated steps).

sbrl:
I'd want to field-test the whole thing for a while first in manual-mode - to make sure we get the right balance.

waldyrious:
From https://github.com/tldr-pages/tldr/pull/1407#issuecomment-308665847:

> when merging PRs by squashing, it's generally recommended to edit the commit message body and remove any superfluous details, or those that only make sense within the sequence of changes made during the PR (in effect, ensuring the whole thing is written as if it was a single commit from the start).
>
> For PRs that introduce new pages, it would suffice to use a single-line commit message (without a body) saying ": add page", unless the submitter includes details in the PR thread such as the URL for the manpage from where the tldr-page was derived, or other similar metadata.

waldyrious:
From the [Gitter chat room, on 20 April 2017](https://gitter.im/tldr-pages/tldr?at=58f8f91a8e4b63533de9ce0a):

> We generally prefer to have at least two maintainers review each PR, so for example if one reviews and approves it, the other can merge the PR once they agree as well (unless someone else raises objections).
> If other maintainers are too busy and a PR lingers around for too long without a second maintainer’s approval, they can be merged with a single maintainer’s approval, in the interests of providing a timely response to contributions

[...]

PR #1407, opened 2017-06-09
valgrind: add page

gttotev:
Add a page for Valgrind, a sophisticated suite of tools for profiling programs.

### DIFF
--- a/pages/common/valgrind.md
+++ b/pages/common/valgrind.md
@@ -1,0 +1,19 @@
+# valgrind
+
+> Wrapper to a set of expert tools for profiling, optimizing and debugging your programs.
+
+- Test `program` with a specified Valgrind tool:
+
+`valgrind --tool={{tool_name}} {{program}}`
+
+- Use the (default) Memcheck tool to report memory leaks of `program` in detail:
+
+`valgrind --leak-check=full --show-reachable=yes {{program}}`
+
+- Use the Cachegrind tool to profile and log CPU cache operations of `program`:
+
+`valgrind --tool=cachegrind {{program}}`
+
+- Use the Massif tool to profile and log heap memory and stack usage of `program`:
+
+`valgrind --tool=massif --stacks=yes {{program}}`
--- a/pages/common/valgrind.md
+++ b/pages/common/valgrind.md
@@ -2,7 +2,7 @@
 
 > Wrapper for a set of expert tools for profiling, optimizing and debugging programs.
 
-- Test `program` with a specified Valgrind tool:
+- Test `program` with the specified Valgrind tool. Commonly used tools include `memcheck`, `cachegrind`, `callgrind`, `massif`, `helgrind`, and `drd`:
 
 `valgrind --tool={{tool_name}} {{program}}`
 
--- a/pages/common/valgrind.md
+++ b/pages/common/valgrind.md
@@ -6,9 +6,9 @@
 
 `valgrind --tool={{tool_name}} {{program}}`
 
-- Use the (default) Memcheck tool to report memory leaks of `program` in detail:
+- Use the (default) Memcheck tool to report all possible memory leaks of `program` in detail:
 
-`valgrind --leak-check=full --show-reachable=yes {{program}}`
+`valgrind --leak-check=full --show-leak-kinds=all {{program}}`
 
 - Use the Cachegrind tool to profile and log CPU cache operations of `program`:
 
--- a/pages/common/valgrind.md
+++ b/pages/common/valgrind.md
@@ -1,12 +1,13 @@
 # valgrind
 
 > Wrapper for a set of expert tools for profiling, optimizing and debugging programs.
+> Commonly used tools include `memcheck`, `cachegrind`, `callgrind`, `massif`, `helgrind`, and `drd`.
 
-- Test `program` with the specified Valgrind tool. Commonly used tools include `memcheck`, `cachegrind`, `callgrind`, `massif`, `helgrind`, and `drd`:
+- Use the (default) Memcheck tool to show a diagnostic of memory usage by `program`.
 
-`valgrind --tool={{tool_name}} {{program}}`
+`valgrind {{program}}`
 
-- Use the (default) Memcheck tool to report all possible memory leaks of `program` in detail:
+- Use Memcheck to report all possible memory leaks of `program` in full detail:
 
 `valgrind --leak-check=full --show-leak-kinds=all {{program}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/valgrind.md
```diff
@@ -0,0 +1,19 @@
+# valgrind
+
+> Wrapper to a set of expert tools for profiling, optimizing and debugging your programs.
+
+- Test `program` with a specified Valgrind tool:
+
+`valgrind --tool={{tool_name}} {{program}}`
+
+- Use the (default) Memcheck tool to report memory leaks of `program` in detail:
```
agnivade: Can  you explain what `show-reachable` does in the description ?

pages/common/valgrind.md
```diff
@@ -0,0 +1,19 @@
+# valgrind
+
+> Wrapper to a set of expert tools for profiling, optimizing and debugging your programs.
```
waldyrious: I think "Wrapper for" would sound better than "Wrapper to". Also, please avoid addressing the reader directly ("your").

waldyrious:
Is there an option to list the available tools?

gttotev:
I looked around for an option, but it appears that the tools are only listed in `man valgrind` and online.

pages/common/valgrind.md
```diff
@@ -0,0 +1,19 @@
...
+
+> Wrapper for a set of expert tools for profiling, optimizing and debugging programs.
+
+- Test `program` with a specified Valgrind tool:
+
+`valgrind --tool={{tool_name}} {{program}}`
+
+- Use the (default) Memcheck tool to report all possible memory leaks of `program` in detail:
+
+`valgrind --leak-check=full --show-leak-kinds=all {{program}}`
```
waldyrious: Since the page only has 4 examples, I think it would be acceptable to split this one into separate examples, and perhaps use the simplest one as the first example in the page. How does that sound?

waldyrious:
That's too bad... how many are them? If they're 3/4, maybe we could list them explicitly in the main description.

gttotev:
There are about 10 tools total but only around half of them are commonly used.

waldyrious:
Ok, in that case let's leave a listing out. We might be able to include links in tldr pages later on (#481, #1082), in which case we might point to a page that presents a proper listing of the available tools.

pages/common/valgrind.md
gttotev: How about making a new first example, `valgrind {{program}}`, and leaving this one intact in its place to show more options for Memcheck?

waldyrious: That'd work, yeah. Maybe this one could be the second example then. And maybe mention "with full details" rather than just "in detail" just to make the option associations clearer?

gttotev: If I placed these two examples first, where would I put the general form of the command with the syntax for specifying a tool? At the bottom?

pages/common/valgrind.md
```diff
@@ -0,0 +1,19 @@
+# valgrind
+
+> Wrapper for a set of expert tools for profiling, optimizing and debugging programs.
+
+- Test `program` with the specified Valgrind tool. Commonly used tools include `memcheck`, `cachegrind`, `callgrind`, `massif`, `helgrind`, and `drd`:
```
waldyrious: Please move the "Commonly used tools..." part to the main description, as a second line.

pages/common/valgrind.md
waldyrious: I don't think a general form is needed if we're providing two specific examples (cachegrind and massif). So my suggestion would be:

    valgrind {{program}}
    valgrind --leak-check=full --show-leak-kinds=all {{program}}
    valgrind --tool=cachegrind {{program}}
    valgrind --tool=massif --stacks=yes {{program}}

pages/common/valgrind.md
```diff
@@ -0,0 +1,20 @@
+# valgrind
+
+> Wrapper for a set of expert tools for profiling, optimizing and debugging programs.
+> Commonly used tools include `memcheck`, `cachegrind`, `callgrind`, `massif`, `helgrind`, and `drd`.
+
+- Use the (default) Memcheck tool to show a diagnostic of memory usage by `program`:
```
sbrl: Is `Memcheck` supposed to be capitalised?

pages/common/valgrind.md
```diff
@@ -0,0 +1,20 @@
...
+
+`valgrind {{program}}`
+
+- Use Memcheck to report all possible memory leaks of `program` in full detail:
+
+`valgrind --leak-check=full --show-leak-kinds=all {{program}}`
+
+- Use the Cachegrind tool to profile and log CPU cache operations of `program`:
+
+`valgrind --tool=cachegrind {{program}}`
```
sbrl: Do each of the tools have aliases you can use instead?

This comment isn't a request to change - it's more of a question from someone who's curious :smiley_cat:

pages/common/valgrind.md
waldyrious: I think it makes sense to capitalize proper names in the descriptions, and use lowercase for the actual invocations in the commands.

gttotev: In the Valgrind manual, the tools are referred to by their capitalized proper names in the text while they are lowercase in the code.

pages/common/valgrind.md
gttotev: No, unfortunately. It gets old to have to keep typing them but that's why there's the `alias` bash command :)

sbrl: Awww, drat. Nevermind, then! Yah, `alias` could be handy here :P

waldyrious:
@sbrl ust a minor note: when merging PRs by squashing, it's generally recommended to edit the commit message body and remove any superfluous details, or those that only make sense within the sequence of changes made during the PR (in effect, ensuring the whole thing is written as if it was a single commit from the start).

For this one, as is normally the case for PRs that introduce new pages, it would suffice to use a single-line commit message (without a body) saying "<command>: add page", unless the submitter includes details in the PR thread such as the URL for the manpage from where the tldr-page was derived, or other similar metadata.

sbrl:
@waldyrious Ah, sure! I'll pay more attention there in the future. Thanks for the guidance :smiley_cat:

waldyrious:
No problem. It's my fault for not having codified these little details in the documentation as I promised I'd do ;)

agnivade:
Yea, actually you should put these down somewhere as 'notes for maintainers'.

waldyrious:
> you should put these down somewhere as 'notes for maintainers'.

I've been planning to do so for quite a while now (#1117) but other stuff keep popping up... :disappointed:

sbrl:
@waldyrious I think that it's very much a living standard, that we make up as we go along. I think that the most important thing is that we all follow the same guidelines :P

waldyrious:
Definitely, but it helps to have a place to document this living standard, so that we can avoid sync issues, e.g. new contributors or maintainers being unaware of previously agreed-upon guidelines whose discussion is spread through several unrelated issues/PRs.

PR #1408, opened 2017-06-11
umount: make descriptions more explicit

waldyrious:
Note: the "dangerous" remark was not clear about what the danger is -- we can mention possible caveats, but it's better to be explicit.

### DIFF
--- a/pages/common/umount.md
+++ b/pages/common/umount.md
@@ -1,6 +1,6 @@
 # umount
 
-> Revokes access to an entire filesystem mounted to a directory.
+> Unlink a filesystem from its mount point, making it no longer accessible.
 > A filesystem cannot be unmounted when it is busy.
 
 - Unmount a filesystem, by passing the path to the source it is mounted from:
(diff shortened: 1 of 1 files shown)


waldyrious:
Requesting review/merge from @sbrl and/or @rubenvereecken.

sbrl:
(I'm on my phone, and Octodroid doesn't support reviews yet) 

Those improvements look good to me! Though the description (especially the first line) is misleading. It's true that it revokes access, but it unmounts the device too.

waldyrious:
> Though the description (especially the first line) is misleading. It's true that it revokes access, but it unmounts the device too.

Sure we can change that line too. What would you suggest?

sbrl:
@waldyrious Perhaps "Unmounts an entire filesystem...."? I'm not sure - that uses "unmount", which is almost in the command name.

waldyrious:
Well, WIkipedia describes it [as follows](https://en.wikipedia.org/wiki/Mount_(Unix)):

> (...) `umount` instructs the operating system that the file system should be disassociated from its mount point, making it no longer accessible and may be removed from the computer.

Do you see anything we could use here for a compact description? Maybe "unlink a filesystem from its mount point"? Or perhaps something closer to "disassemble", since the filesystem ceases to exist as such when unmounted?

sbrl:
@waldyrious Yeah, something like that! How about "Unlink a filesystem from its mount point, making it no longer accessible."?

PR #1411, opened 2017-06-12
sshpass: add page

luk3yx:


waldyrious:
Hi @luk3yx, thanks for the contribution! I have a couple questions:

First, why do you use "non-interactive" in the main description? It sounds odd to me to start with a negative, unless the opposite option would be assumed by default. Is this the case? I.e. why would most users (and beginners, particularly) assume that an SSH password provider would be interactive by default?

And second, is there a simpler invocation of the command, with fewer options (i.e. using the command's defaults) that could be used as the first example? Otherwise it's quite a bit of information to process right away.

luk3yx:
@waldyrious I used non-interactive because that's how it's described in the manpage. Does _automatic_ sound better?
The command `sshpass -p password ssh -o StrictHostKeyChecking=no -- "command args"` doesn't strictly need the `--`.
The `-o StrictHostKeyChecking=no` part means if SSH hasn't seen the host key before, it'll approve of it.
sshpass does need `ssh`, as that's the SSH command it invokes.

waldyrious:
> Does automatic sound better?

I guess so, but I'd prefer to actually describe what it does more explicitly. Do you think you could take a stab at that?

As for simplifying the command, I suppose we could start with the simple invocation without `-o StrictHostKeyChecking=no` as a first example, and the second one including it (and then we'd also be able to describe it properly, which is a nice extra). Also, would the command be optional? Or is this non-interactive also in the sense that it doesn't create a session, but merely runs a command in the remote machine then returns?

luk3yx:
@waldyrious You can launch a shell using `sshpass`, if that's what you're asking.
`sshpass` creates a TTY, feeds the password into it, and then redirects stdin to the SSH session.

waldyrious:
Thanks for the additional explanation, @luk3yx, the main description is now much clearer, thanks :)

Unfortunately starting example descriptions with the verb form of "ssh" conflicts with the guidelines of starting descriptions with proper sentence case. How about "Start an ssh session ..."?

Also, perhaps the first example would be clearer with a specific file descriptor rather than just the generic token. What do you guys think?

agnivade:
Agree.

waldyrious:
Thanks @luk3yx!

PR #1417, opened 2017-06-25
curl: remove -I from auth example

### LINKED ISSUE #1406: Bug in curl.md
I have found a wrong command in [`curl.md`](https://github.com/tldr-pages/tldr/blob/87fb4e13bdcd58a465ab2dd7e859e778389d1cab/pages/common/curl.md), though I'm not sure if this goes for all platforms. The `-I` flag does not show just the headers, instead it performs a `HEAD` request. `-i` shows the headers, but not just the headers, also the response body.

Could others have a look at how their `curl` operates?

waldyrious:
(also, minor tweaks to some of the descriptions)

fixes #1406

### DIFF
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -1,13 +1,13 @@
 # curl
 
 > Transfers data from or to a server.
-> Supports most protocols including HTTP, FTP, POP3.
+> Supports most protocols, including HTTP, FTP, and POP3.
 
 - Download the contents of an URL to a file:
 
 `curl {{http://example.com}} -o {{filename}}`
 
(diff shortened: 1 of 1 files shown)


waldyrious:
@agnivade any particular reason you've removed the "(fixes #1406)" from the title of this PR? I think it won't automatically close the issue if it isn't present there -- or is the presence in the actual commit message sufficient?

agnivade:
I think your question is clarified now :smile: Putting "fixes #issue_num" in the commit message is the way to go if you want to auto-close an issue. Putting it in the title is also pretty ugly IMO.

PR #1418, opened 2017-06-27
Update echo.md to correct for possible misuse of english language

kun-zhou:
**I am not sure how to insert additional white spaces between words for display, so I instead inserted `[number of spaces]` to indicate the number of spaces.**

Quotes while using `echo` are not optional. For example, `echo hello[3 spaces here]world` and `echo "hello[3 spaces here]world"` are different. The first outputs "hello world" while the later "hello[3 spaces]world". 

This is because in the first command, `echo` takes two strings, "hello" and "world", while the later takes only one string, "hello[3 spaces]world".

### DIFF
--- a/pages/common/echo.md
+++ b/pages/common/echo.md
@@ -2,7 +2,7 @@
 
 > Print given arguments.
 
-- Print a text message. Note: quotes are optional:
+- Print a text message:
 
 `echo {{"Hello World"}}`
 


agnivade:
@kun-zhou - Great catch. In most of our commands, we put strings inside quotes because its the most common use-case for a beginner.

You are right that -
```
echo hello    world
```
and
```
echo "hello    world"
```

will have different outputs. However, if you see, for a beginner, usually the use-case will be the latter. And if we say that quotes are optional, naturally the question arises, why are they put in the first place if its optional. Therefore, to keep it simple and clean, we have put quotes.

Let me know what you think about it.

kun-zhou:
@agnivade I totally agree that putting quotes around the string is probably the most frequent use-case for a beginner. And I think that saying "quotes are optional" makes the sentence more confusing because it seems to suggest that quotes don't make a difference.

agnivade:
@sbrl - Just to be clear, I am reluctant to merge this PR. And @kun-zhou also agrees with me.

sbrl:
@agnivade Hrm. Maybe I'm reading this wrong then - I do have issues sometimes with things like that I think I'll stay out of this one then

agnivade:
Closing this one as nobody has spoken anything else.

@kun-zhou - Thank you for putting forward your thoughts. Please feel free to send PRs to improve any other pages you feel like. Grammar correctional PRs are always welcome !

PR #1431, opened 2017-07-20
dmesg: add page

7373Lacym:


pages/common/dmesg.md
```diff
@@ -0,0 +1,11 @@
+# dmesg
+
+> Write the kernel messages in Linux and other Unix-like operating systems to standard output.
```
agnivade: Since we would be splitting the page into per-os level, we can remove the "Linux and other Unix-like operating systems".

So there, I suggest - 
```
Display kernel messages to standard output.
```

pages/common/dmesg.md
```diff
@@ -0,0 +1,11 @@
...
+
+- Show how much physical memory is available on this system:
```
agnivade: I think we should first put the example of the simplest command usage, which is just `dmesg`.

7373Lacym:
Woo!  Thanks for the feedback!

agnivade:
@sbrl - ping for review

pages/linux/dmesg.md
```diff
@@ -0,0 +1,15 @@
...
+
+> Write the kernel messages to standard output.
+
+- Show kernel messages:
+
+`dmesg`
+
+- Show how much physical memory is available on this system:
+
+`dmesg | grep -i memory`
```
sbrl: That's a really cool trick that I hadn't thought of. Awesome :smiley_cat:

pages/linux/dmesg.md
```diff
@@ -0,0 +1,15 @@
+# dmesg
+
+> Write the kernel messages to standard output.
```
sbrl: This feels a teensy bit awkward, but it's nothing to worry about - I can't think of anything better :P

PR #1432, opened 2017-07-20
wall: clarify description

### LINKED ISSUE #1421: Question: `wall` super-user only wording
The current page on `wall` says:

> Only available to super-user.

But according to the wall man page:

> Only the superuser can write on the terminals of users who have chosen to deny messages or are using a program which automatically denies messages.

IMO, I think it makes sense to remove the current sentence about super-user entirely.  If folks agree I'll put up a PR.

sethwoodworth:
Closes #1421 

EDIT: Changed commit message to auto-close issue

### DIFF
--- a/pages/linux/wall.md
+++ b/pages/linux/wall.md
@@ -1,7 +1,6 @@
 # wall
 
 > Write a message on the terminals of users currently logged in.
-> Only available to super-user.
 
 - Send a message:
 


agnivade:
@sbrl - ping for review.

pages/linux/wall.md
```diff
@@ -1,7 +1,6 @@
 # wall
 
 > Write a message on the terminals of users currently logged in.
-> Only available to super-user.
```
sbrl: I think a "the" is needed here such that it reads "Only available to the super-user".

sethwoodworth: The PR removes the line entirely

agnivade: :rofl: @sbrl - The only thing the PR does is remove that line .. :rofl:

sbrl: Ahhhh! silly me. I think that line is useful, anyhow.

sethwoodworth: @sbrl #1421 contains a previous discussion that `wall` does _not_ require super-user privileges.  Users can send and receive `wall` messages by default.  **If** a user disables the messages a super-user can override that and send them a message anyway.  The line is incorrect as stands and the conclusion of the ticket was that we should drop it overall.

agnivade: Right. @sbrl - You dismissed your review. So, I'm not sure whether you are negative or positive on this. Have a look at #1421 for further context on this PR.

sbrl: Ahh I see. Yeah, it should be removed.

sbrl:
Looks ok to me! :smiley_cat:

PR #1436, opened 2017-07-26
Update macOS references

danielbayley:
Update references to 'OS X' to 'macOS'.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-# OS X filesystem custom folder attributes
+# macOS filesystem custom folder attributes
 .DS_Store
 
 # Bundler-specific files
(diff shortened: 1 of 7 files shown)


agnivade:
I'm not very sure, but aren't 'macos' and 'osx' the same ?

sbrl:
@agnivade I think they are effectively. Apple have apparently rebranded from osx to macos, but I haven't yet met anyone who uses the new name - probably because apple devices aren't very common around my area :P

danielbayley:
It was officially renamed a while ago now… this was just a precursor to my `open` update anyway.

danielbayley:
@sbrl `macOS` 😄 I think their rationale was to bring it in line with the naming scheme for their other products (`iOS`, `watchOS`, `tvOS`…)

agnivade:
Just a note that our platform folder is still named `osx`. And changing that might have consequences for various clients, so keeping it as is for now.

PR #1434, opened 2017-07-22
Create PULL_REQUEST_TEMPLATE.md

agnivade:
Closes #1067

PR #1437, opened 2017-07-26
open: improve page

danielbayley:


pages/osx/open.md
```diff
@@ -6,9 +6,13 @@
 
 `open {{file.ext}}`
 
-- Run a graphical MacOSX application:
+- Run a graphical macOS application:
 
-`open /Applications/{{Application.app}}`
+`open -a {{Application}}`
+
+- Run a graphical macOS app based on the bundle identifier:
```
agnivade: Just curious, do you think this example is commonly used enough for beginners to know ?

To me it seems like you have to first know the bundle identifier and then open the application with it.

agnivade:
Thank you for the PR ! Please take a look at the comment that I made.

pages/osx/open.md
danielbayley: @agnivade Using the bundle id is better since you aren’t making assumptions about where the app is actually located, so is much more reliable for use in say, scripts.

I also included the simpler `-a` example for common/quick command line use… again, hard-coding file paths should be avoided if possible.

agnivade: > Using the bundle id is better since you aren’t making assumptions about where the app is actually located, so is much more reliable for use in say, scripts.

Yes, that's true, but how are you going to get the bundle id in the first place ? Is there an easy way to get it.

> I also included the simpler -a example for common/quick command line use… again, hard-coding file paths should be avoided if possible.

nice.

danielbayley: @agnivade `osascript -e 'id of app "{{Application}}"'`

danielbayley: Do you think that should be in there?

agnivade: Right, so this brings a dependency on another command. That's why I was doubtful that this would be of use to beginners. And we usually avoid mentioning different commands on a single page if possible.

@sbrl / @waldyrious  - Thoughts ?

sbrl: Hrm. In this case I'd say that it's a useful command to have if the default macOS terminal doesn't autocomplete the application id, since you'd still need to go and look up what that other command was if the two go hand-in-hand.

I do understand though that we want to avoid implying that one command is part of the same package / program as another though.

agnivade: Can we say something like `Refer to osascript to get the application id` ? And create a separate page for osascript and put this option there ?

sbrl: @agnivade Yeah! That's a good idea.

agnivade: @danielbayley - Any updates  on this ?

danielbayley:
@agnivade Sorted now. Also added `osascript` in #1440

agnivade:
Looks good. And btw, you don't need to amend to the same commit and force push every time you make a change. You can create new commits and add them so that for a PR, we can see what changes have happened to keep context. And later, while merging, we can squash all the commits together is necessary.

waldyrious:
> you don't need to amend to the same commit and force push every time you make a change. You can create new commits and add them so that for a PR, we can see what changes have happened to keep context. And later, while merging, we can squash all the commits together is necessary.

Nicely written. This should go somewhere in the contribution guide -- perhaps we could split the ["using the command line" section](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#using-the-command-line) into a separate file, as I suggested in https://github.com/tldr-pages/tldr/pull/1434#issuecomment-323292726.

Issue #1177, opened 2016-11-30
page request: mysqld


dikarel:
This is the command that runs the server

https://linux.die.net/man/1/mysqld

morrme:
I would like to help with this one.  I do not see a PR, assignment or any other activity on it. Is there a process for "claiming" issues?

agnivade:
> I would like to help with this one.

Yes please !

>  Is there a process for "claiming" issues?

Just send a PR referring this issue in it. And please take a look at CONTRIBUTING.md.

russellg:
I've had a go at this. Let me know if I can improve it in any way.

bl-ue:
🎉

PR #1438, opened 2017-07-29
mysqld.md: add page

russellg:
Closes https://github.com/tldr-pages/tldr/issues/1177

### DIFF
--- a/pages/common/mysqld.md
+++ b/pages/common/mysqld.md
@@ -15,4 +15,5 @@
 `mysqld --log-error={{file}}`
 
 - Set the port number to use when listening for connections:
+
 `mysqld --port={{number}}`
--- a/pages/common/mysqld.md
+++ b/pages/common/mysqld.md
@@ -1,0 +1,18 @@
+# mysqld
+
+> The MySQL Server.
+
+- Display error log messages:
+
+`mysqld --console`
+
+- Log connections and queries to a file:
+
+`mysqld --log={{file}}`
+
+- Log errors and start messages a file:
+
+`mysqld --log-error={{file}}`
+
+- Set the port number to use when listening for connections:
+`mysqld --port={{number}}`


pages/common/mysqld.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Display error log messages:
+
+`mysqld --console`
+
+- Log connections and queries to a file:
+
+`mysqld --log={{file}}`
+
+- Log errors and start messages a file:
```
agnivade: I think one `--log` option should be sufficient.

pages/common/mysqld.md
```diff
@@ -0,0 +1,19 @@
+# mysqld
+
+> The MySQL Server.
+
+- Display error log messages:
+
+`mysqld --console`
+
+- Log connections and queries to a file:
```
agnivade: Can we make this statement without the "connections and queries" part ?

pages/common/mysqld.md
```diff
@@ -0,0 +1,19 @@
+# mysqld
+
+> The MySQL Server.
+
+- Display error log messages:
```
agnivade: I think you should also mention `Start the server`, since this is the first example and it should be clear that the mysqld command starts up a mysql server.

pages/common/mysqld.md
```diff
@@ -0,0 +1,19 @@
+# mysqld
+
+> The MySQL Server.
```
sbrl: Perhaps a more descriptive description would help here? Perhaps add a short sentence explaining what MySQL is.

russellg:
Thanks for the feedback!
I will review the change requests and update accordingly.
This is a great project. I look forward to contributing further in future.

russellg:
Apologies for not updating this. I will make further updates shortly

pages/common/mysqld.md
waldyrious: @agnivade why is that preferable? I generally think being explicit is nicer, provided it doesn't make the description too verbose or hard to read (which doesn't seem to be the case here).

pages/common/mysqld.md
waldyrious: But isn't one `--log` and the other `--log-error`, each doing different (i.e. mutually exclusive) things? Please correct me if I'm wrong, but otherwise, I don't think there's much to gain by using a single logging example, considering that the page only has 4 examples at the moment.

PR #1440, opened 2017-08-05
osascript: add page

danielbayley:
Also referred to from the open page.

### DIFF
--- a/pages/osx/osascript.md
+++ b/pages/osx/osascript.md
@@ -12,7 +12,7 @@
 
 - Run an AppleScript file:
 
-`osascript {{path/to/apple.scpt}}`
+`osascript {{path/to/apple}}.scpt`
 
 - Get the bundle identifier of an application (useful for `open -b`):
 
--- a/pages/osx/osascript.md
+++ b/pages/osx/osascript.md
@@ -1,0 +1,27 @@
+# osascript
+
+> Run AppleScript or JavaScript for Automation (JXA) from the command line.
+
+- Run an AppleScript command:
+
+`osascript -e '{{say "Hello world"}}'`
+
+- Run multiple AppleScript commands:
+
+`osascript -e '{{say "Hello"}}' -e '{{say "world"}}'`
+
+- Run an AppleScript file:
+
+`osascript {{path/to/apple.scpt}}`
+
+- Get the bundle identifier of an application (useful for `open -b`):
+
+`osascript -e 'id of app "{{Application}}"'`
+
+- Run a JavaScript command:
+
+`osascript -l JavaScript -e '{{console.log("Hello world");}}'`
+
+- Run a JavaScript file:
+
+`osascript -l JavaScript {{path/to/script.js}}`


pages/osx/osascript.md
```diff
@@ -0,0 +1,27 @@
...
+
+`osascript {{path/to/apple.scpt}}`
+
+- Get the bundle identifier of an application (useful for `open -b`):
+
+`osascript -e 'id of app "{{Application}}"'`
+
+- Run a JavaScript command:
+
+`osascript -l JavaScript -e '{{console.log("Hello world");}}'`
```
agnivade: Does `J` need to be in caps ?

danielbayley: Yes, for some bizarre reason it is case sensitive.

pages/osx/osascript.md
```diff
@@ -0,0 +1,27 @@
...
+
+`osascript -e '{{say "Hello world"}}'`
+
+- Run multiple AppleScript commands:
+
+`osascript -e '{{say "Hello"}}' -e '{{say "world"}}'`
+
+- Run an AppleScript file:
+
+`osascript {{path/to/apple.scpt}}`
```
agnivade: Does the extension matter ? If its not 'scpt', will it fail ? If yes, then it should probably be outside the token`{{}}`

pages/osx/osascript.md
```diff
@@ -0,0 +1,27 @@
+# osascript
+
+> Run AppleScript or JavaScript for Automation (JXA) from the command line.
```
sbrl: Perhaps `Run AppleScript or JavaScript files for.....` would read better?

danielbayley: "JavaScript for Automation"/"JXA" are the _official_ [silly] names given by Apple for their [shoddy] JavaScript implementation… again, not decisions made on my part! Whereas "AppleScript" is just called "AppleScript".

pages/osx/osascript.md
danielbayley: Well, it can either be `.scpt`, `.scptd` or `.applescript` (The former 2 are usually compiled, with `.scptd` being a 'script bundle' and the latter AppleScript written in plain text), but yeah it seems to choke unless one of those file extensions is used… So not sure what you want to do with that?

agnivade: Ok, then lets change it to `osascript {{path/to/file}}.scpt` to signify that one should always give the `.scpt` extension.

danielbayley: or one of the other 2…

pages/osx/osascript.md
sbrl: Ah, right! I thought it was a grammatical mistake :P

waldyrious:
FYI, I'm reviewing some PRs I didn't have the chance to look at before merging, and quite enjoyed reading through this one. Nice work with both the submission and review process, @danielbayley :+1:

PR #1442, opened 2017-08-09
z.md: add page

ranyitz:
Add a tldr page for the [z](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/z) plugin

### DIFF
--- a/pages/common/z.md
+++ b/pages/common/z.md
@@ -1,0 +1,27 @@
+# z
+
+> Tracks your most used directories, based on 'frecency'. After a short learning phase, z will take you to the most 'frecent' directory that matches ALL of the regexes given on the command line, in order.
+
+- cd to most frecent dir matching foo:
+
+`z foo`
+
+- cd to most frecent dir matching foo, then bar:
+
+`z foo bar`
+
+- cd to highest ranked dir matching foo
+
+`z -r foo`
+
+- cd to most recently accessed dir matching foo
+
+`z -t foo`
+
+- list all dirs matching foo (by frecency)
+
+`z -l foo`
+
+- remove the current directory from the datafile
+
+`z -x .`
--- a/pages/common/z.md
+++ b/pages/common/z.md
@@ -2,25 +2,25 @@
 
 > Tracks your most used directories and enable quick jumping using regexes from the command line.
 
-- Go to a directory that contains "foo":
+- Go to a directory that contains a regex:
 
-`z foo`
+`z {{regexp}}`
 
-- Go to a directory that contains "foo" and then "bar":
+- Go to a directory that contains regex1 and then regex2:
 
-`z foo bar`
+`z {{regex1}} {{regex2}}`
 
-- Go to the highest ranked directory matching foo:
+- Go to the highest ranked directory matching a regex:
 
-`z -r foo`
+`z -r {{regex}}`
 
-- Go to the most recently accessed directory matching foo:
+- Go to the most recently accessed directory matching a regex:
 
-`z -t foo`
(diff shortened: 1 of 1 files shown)


pages/common/z.md
```diff
@@ -0,0 +1,27 @@
+# z
+
+> Tracks your most used directories and enable quick jumping using regexes from the CLI.
```
agnivade: CLI => command line.

pages/common/z.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Go to the highest ranked directory matching foo:
+
+`z -r foo`
+
+- Go to the most recently accessed directory matching foo:
+
+`z -t foo`
+
+- List all directories matching foo (from z datafile):
```
agnivade: Is it necessary to mention "from z datafile" ? Will it be anything else ?

pages/common/z.md
```diff
@@ -0,0 +1,27 @@
...
+
+`z -l foo`
+
+- Remove the current directory from the datafile:
```
agnivade: Maybe "database" is better suited than "datafile" ?

pages/common/z.md
```diff
@@ -0,0 +1,27 @@
+# z
+
+> Tracks your most used directories and enable quick jumping using regexes from the CLI.
+
+- Go to a directory that contains "foo":
```
sbrl: "A directory"? Does this mean a subdirectory or a parent directory?

pages/common/z.md
sbrl: The z datafile is a file containing a list of all paths that z has seen so far I think, as opposed to the entire system.

agnivade: Yes, I was thinking that would be implicit to the user. But on further thought, its better to be explicity clear. I withdraw my concerns.

pages/common/z.md
ranyitz: A directory from the datafile, it can be a subdirectory or a parent.

pages/common/z.md
ranyitz: But it is datafile, the terminology is taken from the plugin's [README.md](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/z)

agnivade: I think it might be useful to use words that convey the "meaning" of the example rather than copy words from the README verbatim. 

From what I understand, the "datafile" is a file used by the z command to store the directory usages. In that context, it might be better to use the word "datastore" or "database". "datafile" is a bit of an uncommon word, and we want to avoid doubt as much as possible.

I suggest -
```
Remove the current directory from z database:
```

Same for the previous description too.

ranyitz:
@sbrl @agnivade, is that ok ?

agnivade:
Would be great if you can address this comment of mine - https://github.com/tldr-pages/tldr/pull/1442#discussion_r132384060

agnivade:
Ok, looks great. But I see that you changed foo to regex. Any particular reason for that ? We usually try to keep concrete examples like "foo" and "bar", rather than generic words like "regex". It helps to understand better and prevents the traditional 'man' page approach.

So, I would recommend using the "foo" , "bar" examples and using the tokens around them, rather than "regex".

ranyitz:
@agnivade - done

agnivade:
@sbrl - Please squash merge if everything looks good to you.

pages/common/z.md
sbrl: Ahhh ok 😺

sbrl:
😺

PR #1443, opened 2017-08-12
duplicity: add page

sbrl:
.....I seem to find all the difficult commands to document :P

On the plus side, it means less digging around in the duplicity man page for everyone :birthday: :smiley_cat:

### DIFF
--- a/pages/common/duplicity.md
+++ b/pages/common/duplicity.md
@@ -4,19 +4,19 @@
 
 - Backup a folder via ftps to a remote machine, encrypting with a password:
 
-`FTP_PASSWORD={{ftp_login_password}} PASSPHRASE={{encryption_password}} duplicity {{path/to/source/directory}} {{ftps://user@remotehost.tld/target/directory/path/}}`
+`FTP_PASSWORD={{ftp_login_password}} PASSPHRASE={{encryption_password}} duplicity {{path/to/source/directory}} {{ftps://user@hostname/target/directory/path/}}`
 
 - Backup a folder to Amazon S3, doing a full backup every month:
 
-`duplicity --full-if-older-than 1M --use-new-style s3://{{bucket_name[/prefix]}}`
+`duplicity --full-if-older-than {{1M}} --use-new-style s3://{{bucket_name[/prefix]}}`
 
 - Delete versions older than 1 year from a backup stored on a webdav share:
 
-`FTP_PASSWORD={{webdav_login_password}} duplicity remove-older-than 1Y --force {{webdav[s]://user@remotehost.tld[:port]/some_dir}}`
+`FTP_PASSWORD={{webdav_login_password}} duplicity remove-older-than {{1Y}} --force {{webdav[s]://user@hostname[:port]/some_dir}}`
 
 - List the files in a backup stored on a remote machine via ssh:
 
-`duplicity list-current-files --time {{timespec}} scp://{{user@domain.tld}}/path/to/backup/dir`
+`duplicity list-current-files --time {{timespec}} scp://{{user@hostname}}/path/to/backup/dir`
 
 - Restore a folder from a gpg-encrypted local backup to a folder:
 
--- a/pages/common/duplicity.md
+++ b/pages/common/duplicity.md
@@ -14,10 +14,14 @@
 
 `FTP_PASSWORD={{webdav_login_password}} duplicity remove-older-than {{1Y}} --force {{webdav[s]://user@hostname[:port]/some_dir}}`
 
+- List the backups available:
+
+`duplicity collection-status "file://{{absolute/path/to/backup/folder}}"
+
 - List the files in a backup stored on a remote machine via ssh:
 
 `duplicity list-current-files --time {{timespec}} scp://{{user@hostname}}/path/to/backup/dir`
 
 - Restore a folder from a gpg-encrypted local backup to a folder:
 
-`PASSPHRASE={{gpg_key_password}} duplicity restore --encrypt-key {{gpg_key_id}} --file-to-restore {{relative/path/restorefolder}} file://{{absolute/to/backup/folder}} {{path/to/directory/to/restore/to}}`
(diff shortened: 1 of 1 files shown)


pages/common/duplicity.md
```diff
@@ -0,0 +1,23 @@
+# duplicity
+
+> Creates incremental, compressed, encrypted, and versioned backups and optionally uploads them to a variety of backend services.
```
agnivade: No need for comma after "encrypted".

pages/common/duplicity.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Backup a folder via ftps to a remote machine, encrypting with a password:
+
+`FTP_PASSWORD={{ftp_login_password}} PASSPHRASE={{encryption_password}} duplicity {{path/to/source/directory}} {{ftps://user@remotehost.tld/target/directory/path/}}`
```
agnivade: Can we do without the environment variables ? They are cluttering the example and seem to be ftps specific rather than duplicity specific.

Also, remotehost.tld => hostname is slightly better.

pages/common/duplicity.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Backup a folder to Amazon S3, doing a full backup every month:
+
+`duplicity --full-if-older-than 1M --use-new-style s3://{{bucket_name[/prefix]}}`
```
agnivade: tokens around 1M

pages/common/duplicity.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Delete versions older than 1 year from a backup stored on a webdav share:
+
+`FTP_PASSWORD={{webdav_login_password}} duplicity remove-older-than 1Y --force {{webdav[s]://user@remotehost.tld[:port]/some_dir}}`
```
agnivade: Same on env vars.

Tokens around 1Y.  

Same on remotehost.

pages/common/duplicity.md
```diff
@@ -0,0 +1,23 @@
...
+
+- List the files in a backup stored on a remote machine via ssh:
+
+`duplicity list-current-files --time {{timespec}} scp://{{user@domain.tld}}/path/to/backup/dir`
```
agnivade: Can we give a concrete example on "timespec" ? I feel its better rather than a generic statement ?

Also, shouldn't this be `ssh` rather than `scp` ?

pages/common/duplicity.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Restore a folder from a gpg-encrypted local backup to a folder:
+
+`PASSPHRASE={{gpg_key_password}} duplicity restore --encrypt-key {{gpg_key_id}} --file-to-restore {{relative/path/restorefolder}} file://{{/path/to/backup/folder}} {{path/to/directory/to/restore/to}}`
```
agnivade: Same on env vars.

Remove the preceding slash from the paths.

pages/common/duplicity.md
sbrl: Ah, ok - I'll remove it. See [this](https://en.oxforddictionaries.com/explore/what-is-the-oxford-comma) though.

pages/common/duplicity.md
sbrl: Sorry, they're required! and `FTP_PASSWORD` is actually a duplicity thing to pass a password to a backend. While I do agree they clutter the example, unfortunately they are required 😕

Sure, I'll change that to `hostname`.

pages/common/duplicity.md
sbrl: Ah, oops! I must have missed that. I'll update that.

pages/common/duplicity.md
sbrl: I'll have to do some research.

Nope duplicity is weird - it requires `scp` and not `ssh` AFAIK.

pages/common/duplicity.md
sbrl: The preceding slask on `/path/to/backup/folder` is required. `file://` is the protocol / backend specifier, and `/path/to/backup/folder` is the absolute path to the folder to backup.

pages/common/duplicity.md
agnivade: > While I do agree they clutter the example, unfortunately they are required 

Hmm I see. Weird why they would make it an env var instead of a flag.

pages/common/duplicity.md
agnivade: I see. So how about this, lets change `/path/to/backup/folder` to `absolute/path/to/backup/folder`.  That way, the path naming convention remains the same and we also clarify that an absolute path is needed.

pages/common/duplicity.md
agnivade: Hmm that's strange. A `,` before an `and` just feels weird. I'm no grammar expert by any means. pinging @waldyrious if he has some time to give his expert comments.

pages/common/duplicity.md
sbrl: I think it's for security reasons. If it's a flag then it shows up in the cmd args associated with a process which you can extract with `ps`.

If there's a better way to present this, I'm all ears :slightly_smiling_face: :elephant:

pages/common/duplicity.md
sbrl: Hrm, can do! Personally I think that saying `absolute/path/` is misleading, but it's not a big deal :slightly_smiling_face:

pages/common/duplicity.md
sbrl: Found it hiding in the man page. Apparently the timespecs are weird and specific - good thing I checked! They have to be like this:

 - `2002-01-25T07:00:00+02:00`
 - `now`
 - time in seconds since 1970 (i.e. the unix epoch)
 - `1y53m2h6m` (or any combination thereof) - specifies the time in the past to which you want to travel
 - `yyyy-mm-dd`

I'm not sure how to represent that

agnivade: If they can be in any of those formats, I would recommend going with the last. It's the simplest and easiest to understand.

agnivade:
Looks good overall, just the timespec part needs to be taken care of and we should be good to do !

pages/common/duplicity.md
sbrl: Sure :D

sbrl:
Thanks for the review help, @agnivade :smiley_cat:

PR #1444, opened 2017-08-13
lldb.md: add page

kevinushey:
Add a page for the [LLDB](https://lldb.llvm.org/) debugger.

### DIFF
--- a/pages/linux/lldb.md
+++ b/pages/linux/lldb.md
@@ -1,6 +1,6 @@
 # lldb
 
-> The LLDB Debugger.
+> The LLVM Low-Level Debugger.
 
 - Debug an executable:
 
@@ -10,6 +10,6 @@
 
 `lldb -p {{pid}}`
 
-- Wait for process to launch and attach to it:
+- Wait for a new process to launch with the given name and attach to it:
 
 `lldb -w -n {{process_name}}`
--- a/pages/linux/lldb.md
+++ b/pages/linux/lldb.md
@@ -1,0 +1,15 @@
+# lldb
+
+> The LLDB Debugger.
+
+- Debug an executable:
+
+`lldb {{executable}}`
+
+- Attach lldb to running process with id {{pid}}:
+
+`lldb -p {{pid}}`
+
+- Wait for process to launch and attach to it:
+
+`lldb -w -n {{process_name}}`
--- a/pages/osx/lldb.md
+++ b/pages/osx/lldb.md
@@ -1,6 +1,6 @@
 # lldb
 
-> The LLDB Debugger.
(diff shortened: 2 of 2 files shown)


pages/common/lldb.md
```diff
@@ -0,0 +1,15 @@
...
+
+`lldb {{executable}}`
+
+- Attach lldb to running process with id {{PID}}:
+
+`lldb -p {{PID}}`
+
+- Wait for process to launch and attach to it:
+
+`lldb -w -n {{name}}`
```
agnivade: Lets change name to process_name to clarify things better. Also, a question - what if there are 2 processes with the same name. If I use the command without the `-w` flag, which process will it attach to ?

pages/common/lldb.md
```diff
@@ -0,0 +1,15 @@
+# lldb
+
+> The LLDB Debugger.
```
sbrl: What does LLDB stand for?

agnivade: low level debugger ? like llvm.

kevinushey: I followed in the steps of `tldr gdb` here, which also just has

```
gdb

The GNU Debugger.
```

Not sure if there's a better way to describe it? The same title is used at https://lldb.llvm.org/ FWIW.

pages/common/lldb.md
kevinushey: In this case, neither of those processes -- `lldb` will instead wait for a _new_ process to be launched with name `{{name}}`, and then attach to that. (This is incredibly handy when you need to attach to a child process launched by some other process)

kevinushey:
Addressed PR feedback (and force-pushed to keep PR history clean). Thanks for the review!

agnivade:
>  (and force-pushed to keep PR history clean).

FYI, you need not do that. Adding commits linearly helps keep track of the PR context. And then while merging, we have an option to squash all the commits together. So, that can be achieved while merging the PR. :smile:

pages/common/lldb.md
agnivade: Got it. In that case, lets change the description to -
`Wait for a new process to launch with the given name and attach to it:`

pages/common/lldb.md
agnivade: You can write the expanded version in brackets - LLDB(Low Level Debugger). That is a norm we follow in various pages.

kevinushey:
> FYI, you need not do that. Adding commits linearly helps keep track of the PR context. And then while merging, we have an option to squash all the commits together. So, that can be achieved while merging the PR. 😄

Ah, I didn't realize that!

Now updated to address PR comments -- thanks!

PR #1447, opened 2017-08-15
mogrify: add page

wefhy:


pages/common/mogrify.md
```diff
@@ -0,0 +1,23 @@
+# Mogrify
+
+> Tool that allows making operations on multiple image files ie. resizing, cropping, flipping, adding effects.
```
sbrl: Perhaps mention that it alters the original image (i.e. as compared with `convert`, which saves a new file)?

pages/common/mogrify.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Resize all jpg images in the folder to 50% of initial size:
```
sbrl: "of initial size" -> "of their initial size"

pages/common/mogrify.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Convert all png images in the folder to jpg:
+
+`mogrify -format jpg *.png`
+
+- Halve the saturation for all files:
+
+`mogrify -modulate 100,50 *`
+
+- Double the brightness:
```
sbrl: ....of all files?

pages/common/mogrify.md
```diff
@@ -0,0 +1,23 @@
+# Mogrify
```
agnivade: s/Mogrify/mogrify.

We always start a command with lower case.

pages/common/mogrify.md
```diff
@@ -0,0 +1,23 @@
...
+
+> Tool that allows making operations on multiple images ie. resizing, cropping, flipping, adding effects. Replaces the original file.
```
agnivade: Would `viz.` be a better replacement for `ie.` ?

pages/common/mogrify.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Resize all images starting with DSC to 800x600:
+
+`mogrify -resize 800x600 DSC*`
+
+- Convert all png images in the folder to jpg:
+
+`mogrify -format jpg *.png`
+
+- Halve the saturation for all files in the folder:
```
agnivade: When you say all files, it naturally raises the question - what files are supported ? 

I think you can add a second line in the heading section mentioning - "Supports image formats - jpg, png <fill others> ".

And in this description, instead of "all files", change it to "all supported files".

pages/common/mogrify.md
sbrl: What does `viz.`

pages/common/mogrify.md
sbrl: [imagemagick supports practically any image file you can think of](https://hastebin.com/irawibipit).

agnivade: Cool, then I think "for all image files" is appropriate.

pages/common/mogrify.md
agnivade: "namely". https://en.wikipedia.org/wiki/Viz.

pages/common/mogrify.md
wefhy: Hmm, I changed it to capital after bot complained, I must have misread the error message.
If we always start with lowercase, maybe add it also to bot's check?

pages/common/mogrify.md
wefhy: I would stay with `ie.`. I also didn't know what `viz.` means, it may be unclear for some people. I think, `ie.` fits there well.

agnivade: I recommended `viz.` because it is often used to cite examples or variants of something. And `ie.` is generally used to explain something in simpler terms.

But it is a minor thing, and if you guys think that `ie.` is well suited here, then I'm okay with it.

wefhy: OK, I'm not english native, if you think it fits better, I believe you, I'll also learn something ;)

agnivade: > OK, I'm not english native,

me neither :P

agnivade:
I'll wait for @sbrl to give a final opinion on the `viz` vs `ie` thing before merging it.

pages/common/mogrify.md
sbrl: I'd recommend either `i.e.` or `namely`, as an english native speaker :P

agnivade: Let's go with `ie` then. `namely` just sounds weird.

pages/common/mogrify.md
waldyrious: @agnivade, adding that check needs to be done in [tldr-lint](https://github.com/tldr-pages/tldr-lint), right?

pages/common/mogrify.md
waldyrious: Coming late to the discussion, but IMO we should avoid abbreviations if possible (and in any case, the spelling is `i.e.`, not `ie.`. I'll make a PR suggesting a change to `such as`, which IMO is the most beginner-friendly option :)

PR #1448, opened 2017-08-16
figlet: add page

guojikai:


guojikai:
@agnivade  I installed it via mac homebrew. I dont know whether it has a windows version or not but it has many versions covered almost every language. [https://github.com/search?p=2&q=figlet&ref=opensearch&type=Repositories](url)

agnivade:
I see. Ok, then please remove the page under `linux/` and merge this page with that. And then we'll see how it turns out.

pages/common/figlet.md
```diff
@@ -0,0 +1,15 @@
+# figlet
+
+> Display large characters made up of ordinary screen characters.
```
agnivade: I think the description of the previous page was pretty much to the point. IMO "ASCII banners" is a better and clearer description than saying "large characters".

 Lets keep the original description.

pages/common/figlet.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Use default settings:
```
agnivade: We generally use descriptions which are action oriented. Like describing what the command does.

In this case, how about - `Generate an ascii banner for "Hello World"` ?

pages/common/figlet.md
```diff
@@ -0,0 +1,15 @@
...
+
+`figlet {{Hello World}}`
+
+- To change the font, use the -f option:
+
+`figlet -f {{font}} {{Hello Wrld}}`
```
agnivade: Wrld => typo

pages/common/figlet.md
```diff
@@ -0,0 +1,15 @@
+# figlet
+
+> Display large characters made up of ordinary screen characters.
+
+- Use default settings:
+
+`figlet {{Hello World}}`
+
+- To change the font, use the -f option:
```
agnivade: The usage of the `-f` option is pretty clear from the command itself. So no need to mention it again in the description.

How about - `Generate output by specifying a given font`

waldyrious:
FYI, the removal of the "waiting" label was due to me making minor edits (typos, punctuation) to the comments above.

PR #1450, opened 2017-08-22
xmllint: various fixes

waldyrious:
- fix string example, and improve its description
- move xpath information to the main description
- add "tags" synonym for "nodes"
- add a html example
- simplify description of doctype example
- capitalize "xml" consistently

----

### DIFF
--- a/pages/common/xmllint.md
+++ b/pages/common/xmllint.md
@@ -6,7 +6,7 @@
 
 `xmllint --xpath "//{{foo}}" {{source_file.xml}}`
 
-- Return, as a string, the contents of the first node named "foo":
+- Return the contents of the first node named "foo" as a string:
 
 `xmllint --xpath "string(//{{foo}})" {{source_file.xml}}`
 
--- a/pages/common/xmllint.md
+++ b/pages/common/xmllint.md
@@ -1,24 +1,24 @@
 # xmllint
 
-> XML parser and linter.
+> XML parser and linter that supports XPath, a syntax for navigating XML trees.
 
-- Return all nodes named "foo":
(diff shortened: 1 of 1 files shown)


pages/common/xmllint.md
```diff
@@ -1,24 +1,24 @@
...
-> XML parser and linter.
+> XML parser and linter that supports XPath, a syntax for navigating XML trees.
 
-- Return all nodes named "foo":
+- Return all nodes (tags) named "foo":
 
 `xmllint --xpath "//{{foo}}" {{source_file.xml}}`
 
-- Return as string the contents of first node named "foo":
+- Return, as a string, the contents of the first node named "foo":
```
agnivade: Maybe this is better - "Return the contents of the first node named "foo" as a string"

waldyrious: I thought about that first, but wondered if it made sense to leave the "string" so late in the description since it's the first part of the xpath expression.

I don't think it's a big deal, though. Let me know if you don't think this concern outweighs the admittedly more awkward sentence construction I chose, in which case I'll make the change.

agnivade: > Let me know if you don't think this concern outweighs the admittedly more awkward sentence construction I chose, in which case I'll make the change.

I do. 😛

waldyrious: You *do* think the concern is more significant than the awkwardness? So no change, then?

agnivade: Dang it .. I knew that would be confusing ..

No, I want it to be changed. :P

waldyrious:
Thanks for the review, @agnivade.

PR #1452, opened 2017-08-22
duplicity: various tweaks

waldyrious:
I didn't have the chance to comment on #1443. These are some of the changes I'd propose to make the page more readable. I still have some questions -- hopefully @sbrl can answer them:

- In the WebDAV example, is the variable still FTP_PASSWORD, or was that a copy-paste error?
- In the GPG example, is that meant to be GPG or PGP? I always mix the two. I changed the abbreviation in the description to the less ambiguous GnuPG, but I might have made the wrong assumption.
- In any case, the phrasing of that example feels a little awkward:
  > Restore a **folder** from a gpg-encrypted local backup to a **folder**

  Could it be clarified somehow?

----

### DIFF
--- a/pages/common/duplicity.md
+++ b/pages/common/duplicity.md
@@ -23,6 +23,6 @@
 
 `duplicity list-current-files --time {{YYYY-MM-DD}} scp://{{user@hostname}}/path/to/backup/dir`
 
-- Restore a folder from a GnuPG-encrypted local backup to a folder:
+- Restore a subdirectory from a GnuPG-encrypted local backup to a given location:
 
 `PASSPHRASE={{gpg_key_password}} duplicity restore --encrypt-key {{gpg_key_id}} --file-to-restore {{relative/path/restorefolder}} file://{{absolute/path/to/backup/folder}} {{path/to/directory/to/restore/to}}`
--- a/pages/common/duplicity.md
+++ b/pages/common/duplicity.md
@@ -1,8 +1,9 @@
 # duplicity
 
-> Creates incremental, compressed, encrypted and versioned backups and optionally uploads them to a variety of backend services.
+> Creates incremental, compressed, encrypted and versioned backups.
+> Can also upload the backups to a variety of backend services.
 
-- Backup a folder via ftps to a remote machine, encrypting with a password:
+- Backup a folder via FTPS to a remote machine, encrypting it with a password:
 
 `FTP_PASSWORD={{ftp_login_password}} PASSPHRASE={{encryption_password}} duplicity {{path/to/source/directory}} {{ftps://user@hostname/target/directory/path/}}`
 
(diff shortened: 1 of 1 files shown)


sbrl:
Good catches! Thanks for those ;-)

 - Yep, it's still `FTP_PASSWORD` - no idea why :P
 - That's the right assumption to make, I think. PGP is for email as far as I understand, and GPG is the command in the terminal that encrypts / decrypts a file.
 - Perhaps it could be changed to `Restore a subdirectory from a gpg-encrypted local backup to a folder`?

agnivade:
On a side note, the PR template looks really good. 😄

waldyrious:
> That's the right assumption to make, I think. PGP is for email as far as I understand, and GPG is the command in the terminal that encrypts / decrypts a file.

Not so sure about that. From a brief foray into Wikipedia, GnuPG implements the OpenPGP standard, and other implementations exist (the original one being PGP, from which OpenPGP was defined):

> _**Pretty Good Privacy** (**PGP**) is an encryption program that provides cryptographic privacy and authentication for data communication. PGP is used for signing, encrypting, and decrypting texts, e-mails, files, directories, and whole disk partitions and to increase the security of e-mail communications. (...) PGP and similar software follow the OpenPGP standard (RFC 4880) for encrypting and decrypting data._

> _**GNU Privacy Guard** (**GnuPG** or **GPG**) is a free software replacement for Symantec's PGP cryptographic software suite. GnuPG is compliant with RFC 4880, which is the IETF standards track specification of OpenPGP. Modern versions of PGP are interoperable with GnuPG and other OpenPGP-compliant systems._

This seems to suggest that one does not need to use GPG, specifically, for this command, but please correct me if I'm wrong.

> Perhaps it could be changed to `Restore a subdirectory from a gpg-encrypted local backup to a folder`?

Maybe. Is "folder" here used to mean a specific location? If so, maybe we should use "location" or "path" instead.

sbrl:
> This seems to suggest that one does not need to use GPG, specifically, for this command, but please correct me if I'm wrong.

I believe it uses GPG under-the-hood.

> Maybe. Is "folder" here used to mean a specific location? If so, maybe we should use "location" or "path" instead.

Yeah, it's a specific location.

waldyrious:
> I believe it uses GPG under-the-hood.

Cool, then. Thanks for the clarification.

> Yeah, it's a specific location.

Ok, so how about `Restore a subdirectory from a gpg-encrypted local backup to a given location`?

sbrl:
Ping @waldyrious yep, that's a good idea :D

waldyrious:
OK, done!

waldyrious:
@sbrl, please merge using the squash option and clean up the commit message to only include the first commit's detail (the second one was deliberately written as a discardable message to be squashed away).

sbrl:
@waldyrious Done! That ok? :smiley_cat:

waldyrious:
Oh! I was under the impression that I had detailed each change as an extended message in the first commit. My previous comment sounds silly now, since that wasn't the case xD. In any case, yeah, that's perfect :)

PR #1453, opened 2017-08-24
gist: add two examples

jasonxia23:
----

- [x] The page (if new), does not already exist in the repo.

- [x] The page (if new), has been added to the correct platform folder:  
      `common/` if it's common to all platforms, `linux/` if it's Linux-specific, and so on.

- [x] The page has 8 or fewer examples.

- [x] The PR is appropriately titled:  
      "command: add page" for new pages, or "command: description of changes" for pages being edited

- [x] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines

PR #1456, opened 2017-08-25
improve PULL_REQUEST_TEMPLATE

waldyrious:
The current wording seems to be confusing (cf. PR #1453),
so hopefully with a more explicit placeholder notation, this won't trip other contributors. /cc @jasonxia23.

----

- [ ] ~~The page has 8 or fewer examples.~~

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,6 +3,8 @@
 <!-- Please perform the following checks and check all the boxes that apply. -->
 <!-- If a particular point is not applicable to your PR,
      strike-through the line by wrapping the text in ~~double tildes~~. -->
+<!-- If your PR does not create or edit a command page (e.g. README edits, etc.)
+     you can simply remove the entire checklist. -->
 
 - [ ] The page (if new), does not already exist in the repo.
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -12,6 +12,6 @@
 - [ ] The page has 8 or fewer examples.
 
 - [ ] The PR is appropriately titled:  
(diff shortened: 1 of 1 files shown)


waldyrious:
@agnivade I was wondering, perhaps we could add an additional sentence to the commented-out instructions, saying that if the PR doesn't pertain to a command page, the whole checklist can be removed, since pretty much all of the checks are page-specific, and the contribution guidelines are already shown at the top of the PR creation page anyway. Thoughts?

agnivade:
Sure, makes sense.

waldyrious:
Ok, I've added that change to this PR as well.

agnivade:
ping @sbrl

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -12,6 +14,6 @@
 - [ ] The page has 8 or fewer examples.
 
 - [ ] The PR is appropriately titled:  
-      "command: add page" for new pages, or "command: description of changes" for pages being edited
+      `<command name>: add page` for new pages, or `<command name>: <description of changes>` for pages being edited
```
sbrl: Perhaps we should change the `<` `>` out for `{{` and `}}`, as we have in the pages themselves?

waldyrious: I'd rather not. Angle brackets are quite common as a markup for placeholders, while the double curly brackets are quite specific to tldr-pages, and were chosen because of technical difficulties (to avoid conflicts with redirection operators), rather than because they were considered better than the angle brackets. But open to hear other arguments.

agnivade: I am neutral about this. Both are fine by me.

sbrl: Ah, ok then :smiley_cat:

PR #1457, opened 2017-08-28
peerflix: adds page

meodai:
----

### DIFF
--- a/pages/common/peerflix.md
+++ b/pages/common/peerflix.md
@@ -8,8 +8,8 @@
 
 - Stream a torrent to VLC:
 
-`peerflix "{{torrent url|magnet link}}" --{{vlc}}`
+`peerflix "{{torrent url|magnet link}}" --vlc`
 
 - Stream all files from a torrent to Airplay:
 
-`peerflix "{{torrent url|magnet link}}" --all --{{airplay}}`
+`peerflix "{{torrent url|magnet link}}" --all --airplay`
--- a/pages/common/peerflix.md
+++ b/pages/common/peerflix.md
@@ -2,14 +2,18 @@
 
 > Stream torrents to your favorite player
 
-- Stream a torrent
+- Stream a torrent: 
+
 `peerflix "{{torrent url|magnet link}}"`
 
-- Stream a torrent to a specific player
+- Stream a torrent to a specific player: 
+
 `peerflix "{{torrent url|magnet link}}" --{{player}}`
 
-- Stream a torrent to VLC
+- Stream a torrent to VLC: 
+
 `peerflix "{{torrent url|magnet link}}" --vlc`
 
-- Stream music from a torrent
+- Stream music from a torrent: 
+
 `peerflix "{{torrent url|magnet link}}" -a --{{player}}`
--- a/pages/common/peerflix.md
+++ b/pages/common/peerflix.md
@@ -1,19 +1,19 @@
 # peerflix
 
-> Stream torrents to your favorite player
+> Stream torrents to your favorite player.
 
-- Stream a torrent: 
+- Stream a torrent:
 
 `peerflix "{{torrent url|magnet link}}"`
 
-- Stream a torrent to a specific player: 
+- Stream a torrent to a specific player:
 
 `peerflix "{{torrent url|magnet link}}" --{{player}}`
 
-- Stream a torrent to VLC: 
+- Stream a torrent to VLC:
 
 `peerflix "{{torrent url|magnet link}}" --vlc`
 
-- Stream music from a torrent: 
+- Stream music from a torrent:
 
 `peerflix "{{torrent url|magnet link}}" -a --{{player}}`
--- a/pages/common/peerflix.md
+++ b/pages/common/peerflix.md
@@ -2,14 +2,14 @@
 
 > Stream torrents to your favorite player.
 
-- Stream a torrent:
+- Stream a all video-files in a torrent:
 
-`peerflix "magnet:?xt=urn:btih:ef330b39f4801d25b4245212e75a38634bfc856e"`
+`peerflix "magnet:?xt=urn:btih:00112233445566"`
 
 - Stream a torrent to VLC:
 
 `peerflix "{{torrent url|magnet link}}" --vlc`
 
-- Stream music from a torrent:
+- Stream all music music-files from a torrent:
 
-`peerflix "{{torrent url|magnet link}}" -all --vlc`
+`peerflix "{{torrent url|magnet link}}" --all --vlc`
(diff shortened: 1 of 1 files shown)


pages/common/peerflix.md
```diff
@@ -0,0 +1,19 @@
+# peerflix
+
+> Stream torrents to your favorite player.
+
+- Stream a torrent:
+
+`peerflix "{{torrent url|magnet link}}"`
```
sbrl: The quotes are required, right? Would a concrete example here be possible?

pages/common/peerflix.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Stream a torrent to a specific player:
+
+`peerflix "{{torrent url|magnet link}}" --{{player}}`
+
+- Stream a torrent to VLC:
```
sbrl: Is this needed considering the above example?

pages/common/peerflix.md
```diff
@@ -0,0 +1,19 @@
...
+
+`peerflix "{{torrent url|magnet link}}" --vlc`
+
+- Stream music from a torrent:
```
sbrl: How is this different to the above example? What does `-a` do (and is there a long form of it?)?

pages/common/peerflix.md
meodai: It works without them, but depending they are used in the official man, i guess it depends a lot on what type of characters are used in the magnet link.

I was not sure about the concrete example. I checked out examples like 'screenfetch' as inspiration.

pages/common/peerflix.md
meodai: probably not

pages/common/peerflix.md
meodai: -a stands for all (all files in the torrent), I did not even know the long form was implemented ;) I will change that right away

meodai:
@sbrl Thanks for your feedback. can you check again?

pages/common/peerflix.md
sbrl: Ah, ok.  Ping @agnivade /  @waldyrious - what do you think?

pages/common/peerflix.md
sbrl: Thanks :D

pages/common/peerflix.md
```diff
@@ -0,0 +1,15 @@
...
+
+`peerflix "magnet:?xt=urn:btih:ef330b39f4801d25b4245212e75a38634bfc856e"`
+
+- Stream a torrent to VLC:
+
+`peerflix "{{torrent url|magnet link}}" --vlc`
+
+- Stream music from a torrent:
+
+`peerflix "{{torrent url|magnet link}}" -all --vlc`
```
sbrl: Is this supposed to have just one dash?

pages/common/peerflix.md
```diff
@@ -0,0 +1,15 @@
+# peerflix
+
+> Stream torrents to your favorite player.
+
+- Stream a torrent:
+
+`peerflix "magnet:?xt=urn:btih:ef330b39f4801d25b4245212e75a38634bfc856e"`
```
waldyrious: IMO it would be nice to include a note in the description for this example saying what this magnet link refers to.

pages/common/peerflix.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Stream a torrent to VLC:
+
+`peerflix "{{torrent url|magnet link}}" --vlc`
```
waldyrious: Maybe the above example could mention (in the description) that it uses a magnet link, and this one could then be a (concrete) torrent URL, to demonstrate what format the command expects in this case. We can still include the `--vlc` here, since otherwise the syntax for this example is identical to the previous one.

pages/common/peerflix.md
waldyrious: Replied (kind of, since the source has changed) [here](https://github.com/tldr-pages/tldr/pull/1457#discussion_r135742461).

pages/common/peerflix.md
agnivade: Can we just dump a random string here ? Pointing to a real magnet url does not seem very elegant to me.

I want this to be similar to examples of urls like - `www.example.com`.

pages/common/peerflix.md
agnivade: Please update the description to mention that this will stream all files in the torrent.

pages/common/peerflix.md
waldyrious: That works too. Preferably something obviously example-like, such as `000011112222...eeeeffff`, or something cleverer inspired by https://en.wikipedia.org/wiki/Hexspeak#Notable_magic_numbers or http://bada55.io/

meodai:
@agnivade thanks for your feedback. I changed a few things. I did not add the `{{}}` around the magnet link, because in other tldr files it seams only to be used for strings that describe the passed param, not examples. But maybe I'm wrong let me know if I should add it.

agnivade:
@meodai - The `{{}}` syntax is used by tldr clients to highlight any sections of a command which can be replaced by user input. In this case, the magnet url you have mentioned is just a concrete example, but an example nonetheless. That means an user will replace that with their own magnet url. 

Hence, we need the `{{}}`.

pages/common/peerflix.md
```diff
@@ -0,0 +1,15 @@
+# peerflix
+
+> Stream video or audio based torrents to your favorite player.
+
+- Stream all video-files in a torrent:
```
agnivade: This command does not have the `--all` flag. But still you have mentioned `all video-files`. If it will indeed stream all the files, then perhaps we do not need the last example at all ?

And also, it's better to just mention "files" instead of "video-files", as you don't know what a torrent will contain.

pages/common/peerflix.md
```diff
@@ -0,0 +1,15 @@
...
+
+`peerflix "magnet:?xt=urn:btih:00112233445566"`
+
+- Stream a torrent to VLC:
+
+`peerflix "{{torrent url|magnet link}}" --vlc`
+
+- Stream all music music-files from a torrent:
```
agnivade: Same here. Just mention "files".

meodai:
@agnivade ok in this case i did put the `{{}}` around the player as well

agnivade:
@meodai - I removed the tokens around the `--vlc` and `--airplay` because they are constant according to the description of that command. If all looks good to you, we are ready to merge this !

meodai:
@agnivade thanks this is good to me. Sorry for the bumpy pull request, the next one will be smoother I hope ;)

agnivade:
Not at all. Thanks for being patient with our review process. I know it is painful sometimes.

PR #1458, opened 2017-08-28
ntfsfix: add page

jsonbruce:
----

### DIFF
--- a/pages/linux/ntfsfix.md
+++ b/pages/linux/ntfsfix.md
@@ -1,7 +1,7 @@
 # ntfsfix
 
-> Fix common errors and force Windows to check NTFS.
+> Fix some common NTFS problems in the situation that you have both Linux and Windows on your machine. You may run ntfsfix on an NTFS volume if you think  it  was  damaged  by Windows or some other way and it cannot be mounted.
 
-- Fix the Windows partition sdb2:
+- Fix the NTFS partition sdb2:
 
 `sudo ntfsfix /dev/sdb2`
--- a/pages/linux/ntfsfix.md
+++ b/pages/linux/ntfsfix.md
@@ -2,6 +2,6 @@
 
 > Fix common errors and force Windows to check NTFS.
 
-- Fix the Windows partition sdb2
+- Fix the Windows partition sdb2:
 
 `sudo ntfsfix /dev/sdb2`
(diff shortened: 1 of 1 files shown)


sbrl:
Ping @jsonbruce, I think you need a colon in there as @tldr-bot has mentioned.

jsonbruce:
The error has been fixed. @tldr-bot @sbrl

pages/linux/ntfsfix.md
```diff
@@ -0,0 +1,7 @@
+# ntfsfix
+
+> Fix common errors and force Windows to check NTFS.
```
sbrl: > common errors

Perhaps "common errors in NTFS partitions" would read better?

pages/linux/ntfsfix.md
```diff
@@ -0,0 +1,7 @@
...
+
+- Fix the Windows partition sdb2:
```
sbrl: "windows partition"? This feels inconsistent. Maybe "NTFS" would read better?

pages/linux/ntfsfix.md
```diff
@@ -0,0 +1,7 @@
...
+
+`sudo ntfsfix /dev/sdb2`
```
agnivade: So, I am a bit confused, you have added the page under `linux`, and the heading mentions `force Windows to check NTFS`.

Can you explain a bit what does that mean ? Does this tool check NTFS partitions from a linux OS ?

jsonbruce: Sorry for confusing description.

The situation is that **you have both Linux and Windows on your machine.**
You may run ntfsfix on an NTFS volume if you think  it  was  damaged  by Windows or some other way and it cannot be mounted.

I have updated the document.

agnivade: Understood. I simplified the description a bit. I only got confused because you mentioned "force Windows", so how could Windows do something if the command ran on linux. 

Now that it's clear that the command runs on linux and checks NTFS partitions, there is no need to mention the need to have "linux and windows on a machine". It might as well be a foreign partition mounted on an external drive. Not necessarily be an OS.

Also, I feel there is no need to mention the 2nd line because its pretty obvious that it repairs a damaged partition. So unless it was damaged in the first place by whatever means, you won't run this command.

I went ahead and made these changes as I hope they are pretty small. Let me know if you are okay with it.

sbrl:
Thanks, @jsonbruce! :smiley_cat:

PR #1459, opened 2017-08-29
ln: adds the syntax to create multiple symbolic links

jk:
----

### DIFF
--- a/pages/common/ln.md
+++ b/pages/common/ln.md
@@ -8,7 +8,7 @@
 
 - Create several symbolic links to one file (or folder):
 
-`ln -s {{path/to/file}} {{path/to/symlink1}}`
+`ln -s {{path/to/file}} {{path/to/symlink1}} {{path/to/symlink2}} {{path/to/symlink3}}`
 
 - Overwrite an existing symbolic to point to a different file:
 
--- a/pages/common/ln.md
+++ b/pages/common/ln.md
@@ -6,6 +6,10 @@
 
 `ln -s {{path/to/file}} {{path/to/symlink}}`
 
+- Create several symbolic links to one file (or folder):
+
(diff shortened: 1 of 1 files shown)


pages/common/ln.md
```diff
@@ -6,6 +6,10 @@
 
 `ln -s {{path/to/file}} {{path/to/symlink}}`
 
+- Create several symbolic links to one file (or folder):
+
+`ln -s {{path/to/file}} {{path/to/symlink1}}`
```
agnivade: Sorry, maybe I am misunderstanding something. Isn't this exactly the same as the previous example ? You seem to have just changed `symlink` to `symlink1`.

jk: @agnivade Sorry, you're right, my bad. I just forgot the additional symlinks. I pushed the fix to my branch. Should be fine now.

agnivade:
@jk - Thanks for the correction. Though I am not very sure if this is worth adding. We can easily run the first command multiple times to generate multiple symlinks. I will defer to others for the final decision.

/cc @sbrl , @waldyrious

jk:
@agnivade Alright, no problem, thanks for the feedback.

waldyrious:
@jk thanks for the contribution. I'm also hesitant about this -- could you please explain the motivation / use case for this invocation? We tend not to include too obscure usages of a command that most users won't be exposed to, to keep the signal-to-noise ratio up.

jk:
@waldyrious My use case is that I create several symbolic links to a newly deployed directory during a deployment. But @agnivade's objection is justified, it is equally possible to simply apply the command several times in succession. I'm closing the pull request.

waldyrious:
I appreciate your understanding :+1: Cheers!

PR #1460, opened 2017-08-30
docker-compose: add page

Paveltarno:
----

### DIFF
--- a/pages/common/docker-compose.md
+++ b/pages/common/docker-compose.md
@@ -10,9 +10,9 @@
 
 `docker-compose up --build`
 
-- Start all containers using a custom configuration file:
+- Start all containers using an alternate compose file:
 
-`docker-compose --file {{path/to/config}} up`
+`docker-compose --file {{path/to/file}} up`
 
 - Stop all running containers:
 
--- a/pages/common/docker-compose.md
+++ b/pages/common/docker-compose.md
@@ -1,0 +1,28 @@
+# docker-compose
+
+> Run and manage multi container docker applications  
+
+- Create and start containers in the background using a local `docker-compose.yml` file  
+
+`docker-compose up -d`
+
+- Start containers, rebuild if necessary
+
+`docker-compose up --build`
+
+- Start containers using a custom configuration file
+
+`docker-compose --file {{path/to/config}} up`
+
+- Stop running containers  
+
+`docker-compose stop` 
+
+- Stop and remove containers, networks, images, and volumes
+
+`docker-compose down`
+
+- Follow logs
+
+`docker-compose logs --follow`
+
--- a/pages/common/docker-compose.md
+++ b/pages/common/docker-compose.md
@@ -1,28 +1,27 @@
 # docker-compose
 
-> Run and manage multi container docker applications  
+> Run and manage multi container docker applications
 
-- Create and start containers in the background using a local `docker-compose.yml` file  
+- Create and start containers in the background using a local `docker-compose.yml` file
 
-`docker-compose up -d`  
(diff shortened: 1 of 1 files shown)


pages/common/docker-compose.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Start containers, rebuild if necessary:
+
+`docker-compose up --build`
+
+- Start containers using a custom configuration file:
+
+`docker-compose --file {{path/to/config}} up`
+
+- Stop running containers:
```
sbrl: Does this stop all containers, or just some of them?

pages/common/docker-compose.md
```diff
@@ -0,0 +1,27 @@
...
+
+`docker-compose stop`
+
+- Stop and remove containers, networks, images, and volumes:
```
sbrl: What's the difference between this command and the previous one? Does it delete containers / networks / images / volumes?

Paveltarno: @sbrl Yes, the `down` command stops and deletes the containers. The word `remove` is more common in docker (i.e `docker rm`) so IMO its better then `delete`.

pages/common/docker-compose.md
Paveltarno: Iv'e changed the description a bit to reflect these operations effect all containers

pages/common/docker-compose.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Create and start all containers in the background using a local `docker-compose.yml` file:
+
+`docker-compose up -d`
+
+- Start all containers, rebuild if necessary:
+
+`docker-compose up --build`
+
+- Start all containers using a custom configuration file:
```
agnivade: Just a quick question - how is this file different from the `docker-compose.yml` file ?

Paveltarno: When using the `--file` param, you can specify any compose file, without it `docker-compose` will look for a file named `docker-compose.yml` inside the working dir

agnivade: Ah, so its the same file. Please change it to "Start all containers using a given `docker-compose.yml` file." 

And change the token in the example to `{{path/to/file}}` to keep it consistent with other pages.

Paveltarno: sure, though I do want to emphasise that this file doesn't have to be called `docker-compose.yml`

agnivade: Ah I see. So if you don't mention anything. It looks for a file named "docker-compose.yml" by default in the current directory. And if you want to specify some other file, it can have any name.

I was going to comment to change the first description. But I like your new change. That works too.

pages/common/docker-compose.md
```diff
@@ -0,0 +1,27 @@
+# docker-compose
+
+> Run and manage multi container docker applications.
+
+- Create and start all containers in the background using a local `docker-compose.yml` file:
```
agnivade: Maybe better to use the "current directory" than "local". Makes it much more clear.

Paveltarno: agreed

pages/common/docker-compose.md
sbrl: Ah, ok - sounds good to me :smiley_cat:

pages/common/docker-compose.md
sbrl: Thanks :smiley_cat:

sbrl:
Thanks for the hard work, @Paveltarno :D

Paveltarno:
@sbrl NP, thanks for a great tool [=

PR #1461, opened 2017-08-31
timeout: add page

katekyan:
add a new page for timeout.
----

### DIFF
--- a/pages/linux/timeout.md
+++ b/pages/linux/timeout.md
@@ -6,6 +6,6 @@
 
 `timeout {{3s}} {{sleep 10}}`
 
-- Run `sleep 10` with a time limit of 5 seconds or kill it when interruptted:
+- Run `sleep 10` with a time limit of 5 seconds or kill it when interrupted:
 
 `timeout {{--signal INT}} {{5s}} {{sleep 10}}`
--- a/pages/linux/timeout.md
+++ b/pages/linux/timeout.md
@@ -2,10 +2,10 @@
 
 > Run a command with a time limit.
 
-- Run `sleep 10` with a time limit of 3 seconds:
+- Run `sleep 10` and kill it, if it's running after 3 seconds:
 
 `timeout {{3s}} {{sleep 10}}`
 
-- Run `sleep 10` with a time limit of 5 seconds or kill it when interrupted:
+- Specify the signal to be sent to the command after the time limit expires. (By default, TERM is sent):
 
-`timeout {{--signal INT}} {{5s}} {{sleep 10}}`
+`timeout --signal {{INT}} {{5s}} {{sleep 10}}`
--- a/pages/linux/timeout.md
+++ b/pages/linux/timeout.md
@@ -2,10 +2,10 @@
 
 > Run a command with a time limit.
 
-- Run sleep command (for 10 seconds) but kill it after 3 seconds:
+- Run "sleep 10" with a time limit of 3 seconds:
 
-`timeout 3s sleep 10`
+`timeout {{3s}} {{sleep 10}}`
 
-- Run sleep command (for 10 seconds) but kill it when interruptted or after 5 seconds:
+- Run "sleep 10" with a time limit of 5 seconds or kill it when interruptted:
 
-`timeout --signal INT 5s sleep 10`
+`timeout {{--signal INT}} {{5s}} {{sleep 10}}`
--- a/pages/linux/timeout.md
+++ b/pages/linux/timeout.md
@@ -1,0 +1,11 @@
+# timeout
+
+> Run a command with a time limit.
+
+- Run sleep command (for 10 seconds) but kill it after 3 seconds:
+
+`timeout 3s sleep 10`
+
+- Run sleep command (for 10 seconds) but kill it when interruptted or after 5 seconds:
+
+`timeout --signal INT 5s sleep 10`
--- a/pages/linux/timeout.md
+++ b/pages/linux/timeout.md
@@ -2,10 +2,10 @@
 
 > Run a command with a time limit.
 
-- Run "sleep 10" with a time limit of 3 seconds:
+- Run `sleep 10` with a time limit of 3 seconds:
 
 `timeout {{3s}} {{sleep 10}}`
 
-- Run "sleep 10" with a time limit of 5 seconds or kill it when interruptted:
+- Run `sleep 10` with a time limit of 5 seconds or kill it when interruptted:
 
 `timeout {{--signal INT}} {{5s}} {{sleep 10}}`


agnivade:
Congrats on your first github PR ! Would you mind signing the CLA ? Then we can get started with the review.

sbrl:
Ping @katekyan

katekyan:
Thanks for reminding me! I've signed the CLA but the status is still pending.

agnivade:
Hmm strange. It does behave like that very rarely. Maybe you should try re-signing from an incognito tab ? I don't know how to solve it, just an idea.

sbrl:
Do you have any script blockers / ad blockers enabled? Maybe they are causing an issue. Perhaps try disabling them too if @agnivade's suggestion doesn't work?

katekyan:
I tried different browser (IE and chrome). Both of them will show "You have signed the CLA for tldr-pages/tldr" if I click the CLA link.

sbrl:
That's really odd! In that case, I'll just continue with the review anyway. @waldyrious, any suggestions?

pages/linux/timeout.md
```diff
@@ -0,0 +1,11 @@
+# timeout
+
+> Run a command with a time limit.
+
+- Run sleep command (for 10 seconds) but kill it after 3 seconds:
+
+`timeout 3s sleep 10`
+
+- Run sleep command for 10 seconds but kill it after 3 seconds or when interruptted:
```
sbrl: Perhaps putting `for 10 seconds` in brackets like the first example would improve consistency?

And what's the difference between this command and the last?

katekyan: I agree. Adding the brackets will improve the consistency. The last command add the usage case for --signal INT on interrupt signal case. Either timeout or an interrupt signal will kill the process. In the first command, only timeout will kill the process. Both of them are common use case.

katekyan:
I noticed the CLA assitant comment now shows as: 
1 out of 2 committers have signed the CLA.
✅ katekyan
❌ Velik123
I signed the CLA. Am I not the only committer for this pull request?

agnivade:
Your first commit is sent from an email which is unrecognized. I guess that is the source of the issue. You can do a force push and update the email.

agnivade:
Something wrong with your github settings. It says
`kate committed with katekyan 29 days ago  ac42256` on the commit.

katekyan:
The issue is my work station has a different username than my github account name. When I do a git squash, git tried to merge my github account's change under the work station name. We can fix this by using the 'reset-author' option. This post is a life-saver[https://gist.github.com/carlosmcevilly/2221249](url).

agnivade:
Cool. The hyperlink in your comment is incorrect btw.

pages/linux/timeout.md
```diff
@@ -0,0 +1,11 @@
+# timeout
+
+> Run a command with a time limit.
+
+- Run sleep command (for 10 seconds) but kill it after 3 seconds:
+
+`timeout 3s sleep 10`
```
agnivade: Please wrap params in tokens.

pages/linux/timeout.md
```diff
@@ -0,0 +1,11 @@
...
+
+- Run sleep command (for 10 seconds) but kill it when interruptted or after 5 seconds:
```
agnivade: Typo in interruptted.

Also a question here - a running process should automatically quit when the SIGINT signal is sent. What is the need to specify it here with timeout ?

pages/linux/timeout.md
```diff
@@ -0,0 +1,11 @@
...
+
+`timeout --signal INT 5s sleep 10`
```
agnivade: Wrap with tokens.

pages/linux/timeout.md
katekyan: The timeout limit is required in case no signal is sent and some process could run indefinitely.

agnivade: My point was that a process will automatically quit when a SIGINT is received. So, is there a benefit of adding this example in the timeout page ?

pages/linux/timeout.md
```diff
@@ -0,0 +1,11 @@
+# timeout
+
+> Run a command with a time limit.
+
+- Run "sleep 10" with a time limit of 3 seconds:
```
waldyrious: When referring to other commands in the description texts, we normally use backticks so that they're rendered as code. Like so:

    - Run `sleep 10` with ...

pages/linux/timeout.md
```diff
@@ -0,0 +1,11 @@
...
+
+`timeout {{3s}} {{sleep 10}}`
+
+- Run "sleep 10" with a time limit of 5 seconds or kill it when interruptted:
```
waldyrious: I'm not sure I understand this description. Does it mean that the `timeout` command will kill the process after 5 seconds, using the interrupt signal? And if so, what's the default signal (or other mechanism) it uses to stop commands after the time limit?

agnivade: Ah, yea I think you are right. I got it wrong I believe. See here - https://github.com/tldr-pages/tldr/pull/1461#discussion_r141939945.

I thought that it will kill it when someone sends a SIGINT signal, which is supposed to happen anyways. But your analysis makes more sense, in that case, we need to update the description.

sbrl: Typo! Just spotted it. `interruptted` -> `interrupted`.

agnivade: Yes, I mentioned it here - https://github.com/tldr-pages/tldr/pull/1461#discussion_r141939945

sbrl: Ah oops! :P

katekyan: Actually, the default signal to stop the process is that the process will got killed after 5 seconds. If no one sends a SIGINT, the process will end after 5 seconds. Otherwise, the process ends as soon as it gets a SIGINT.

agnivade: @katekyan  I think we are going in circles again. I looked up the man page. @waldyrious is right. Let me explain if I may - 

> is that the process will got killed after 5 seconds.

Yes, but with which signal ? The answer is TERM.

> If no one sends a SIGINT, the process will end after 5 seconds. 

I think you are incorrect in understanding this. No one has to send a signal at all. This will all be done by the `timeout` command. The command will terminate the process using the given signal, `INT` in this case instead of the default `TERM`. 

> Otherwise, the process ends as soon as it gets a SIGINT.

Umm yea, which a process should do anyway :) That has nothing to do with the timeout command.

Let me update the page with some changes which will make things clear.

agnivade:
@katekyan - I have updated the page. Let me know what do you think about it.

pages/linux/timeout.md
katekyan: I see what you mean now. Thanks for clarifying it and updating the description!

katekyan:
Yeah, let's do this! Do you need anything from me?

sbrl:
Nope! All merged thanks to @agnivade :smiley_cat:

katekyan:
That's great! Thank you all!

PR #1466, opened 2017-09-03
pssh: improve page

waldyrious:
I'm not 100% sure that my edits are sensible or even correct. Requesting review from @deekim :)

----

### DIFF
--- a/pages/common/pssh.md
+++ b/pages/common/pssh.md
@@ -4,24 +4,24 @@
 
 - Run a command on two hosts, and print its output on each server inline:
 
-`pssh -i -H "{{host1}} {{host2}}" {{echo "foo"}}`
+`pssh -i -H "{{host1}} {{host2}}" {{ls -l}}`
 
 - Run a command and save the output to separate files:
 
-`pssh -H {{host1}} -H {{host2}} -o {{path/to/output_dir}} {{echo "foo"}}`
+`pssh -H {{host1}} -H {{host2}} -o {{path/to/output_dir}} {{ls -l}}`
 
 - Run a command on multiple hosts, specified in a new-line separated file:
 
-`pssh -i -h {{path/to/hosts_file}} {{echo "foo"}}`
+`pssh -i -h {{path/to/hosts_file}} {{ls -l}}`
 
 - Run a command as root (this asks for the root password):
 
-`pssh -i -h {{path/to/hosts_file}} -A -l {{root_username}} {{echo "foo"}}`
+`pssh -i -h {{path/to/hosts_file}} -A -l {{root_username}} {{ls -l}}`
 
 - Run a command with extra SSH arguments:
 
-`pssh -i -h {{path/to/hosts_file}} -x "{{-O VisualHostKey=yes}}" {{echo "foo"}}`
+`pssh -i -h {{path/to/hosts_file}} -x "{{-O VisualHostKey=yes}}" {{ls -l}}`
 
-- Run a command using the maximum number of concurrent connections:
+- Run a command limiting the number of parallel connections to 10:
 
 `pssh -i -h {{path/to/hosts_file}} -p {{10}} '{{cd dir; ./script.sh; exit}}'`
--- a/pages/common/pssh.md
+++ b/pages/common/pssh.md
@@ -1,27 +1,27 @@
 # pssh
 
-> Parallel ssh program.
+> Parallel SSH program.
 
-- Run a command on host1 and host2 to print the date on each server inline:
+- Run a command on two hosts to print the current date on each server inline:
 
-`pssh -i -H "{{host1}} {{host2}}" {{date}}`
+`pssh -i -H "{{host1}} {{host2}}" {{echo "foo"}}`
 
 - Run a command and save the output to separate files:
 
-`pssh -H {{host1}} -H {{host2}} -o {{path/to/output_dir}} {{date}}`
+`pssh -H {{host1}} -H {{host2}} -o {{path/to/output_dir}} {{echo "foo"}}`
 
-- Run a command on each host specified in a new-line separated host file:
+- Run a command on multiple hosts, specified in a new-line separated file:
 
-`pssh -i -h {{path/to/hosts_file}} {{date}}`
+`pssh -i -h {{path/to/hosts_file}} {{echo "foo"}}`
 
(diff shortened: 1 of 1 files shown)


deekim:
lgtm. :)

pages/common/pssh.md
```diff
@@ -1,27 +1,27 @@
...
+- Run a command on multiple hosts, specified in a new-line separated file:
 
-`pssh -i -h {{path/to/hosts_file}} {{date}}`
+`pssh -i -h {{path/to/hosts_file}} {{echo "foo"}}`
 
-- Run a command as root; asks for root password:
+- Run a command as root (this asks for the root password):
 
-`pssh -i -h {{path/to/hosts_file}} -A -l {{root}} {{date}}`
+`pssh -i -h {{path/to/hosts_file}} -A -l {{root_username}} {{echo "foo"}}`
```
sbrl: Why `root_username` annd not `username` or just `root`?

pages/common/pssh.md
```diff
@@ -1,27 +1,27 @@
...
 
 - Run a command with extra SSH arguments:
 
-`pssh -i -h {{path/to/hosts_file}} -x "{{-O VisualHostKey=yes}}" {{date}}`
+`pssh -i -h {{path/to/hosts_file}} -x "{{-O VisualHostKey=yes}}" {{echo "foo"}}`
 
-- Run a command using a maximum number of concurrent connections:
+- Run a command using the maximum number of concurrent connections:
```
sbrl: Perhaps "specifying" here?

pages/common/pssh.md
```diff
@@ -1,27 +1,27 @@
 # pssh
 
-> Parallel ssh program.
+> Parallel SSH program.
 
-- Run a command on host1 and host2 to print the date on each server inline:
+- Run a command on two hosts, and print its output on each server inline:
 
-`pssh -i -H "{{host1}} {{host2}}" {{date}}`
+`pssh -i -H "{{host1}} {{host2}}" {{echo "foo"}}`
```
sbrl: The quotes won't be transferred through, right?

waldyrious: Not sure what you mean, please expand a bit.

pages/common/pssh.md
waldyrious: Just `root` (what it was previously) is ambiguous IMO (is it a username, the word "root", something else?). Just `username` wouldn't make it sufficiently clear that the user must be root. I think the expanded token strikes an acceptable balance between explicitness and conciseness.

pages/common/pssh.md
waldyrious: Ah, that does make it clearer. IIUC, then, the limit is set by the user, otherwise the default is to use as many as possible -- is that correct?

pages/common/pssh.md
sbrl: Ah, oops! What I meant by this was that the quotes around `foo` won't be read by `pssh` without escaping them first, right?

pages/common/pssh.md
sbrl: Ah, cool :D

pages/common/pssh.md
sbrl: I think so, yeah

pages/common/pssh.md
waldyrious: Oh, so if an echo command with a multi-word (or variable-containing) parameter is sent in this manner, the hosts won't actually see the quotes, you mean? I guess we can use something else then, for simplicity's sake. How about `ls -l`?

pages/common/pssh.md
waldyrious: @deekim please confirm, just to be extra sure :)

deekim: http://www.theether.org/pssh/docs/0.2.3/pssh-HOWTO.html
> By default, pssh uses at most 32 ssh processes in parallel to ssh to the various nodes.

32 i believe.

pages/common/pssh.md
sbrl: Yeah, that's it - `ls -l` would work better

waldyrious:
@sbrl @deekim please re-review. I chose a slightly different wording, and used "parallel" instead of "concurrent" to make the connection with the `-p` clearer.

deekim:
seems okay to me.

pages/common/pssh.md
agnivade: What was the problem with `date` ? It's a simple command. Generates a single line. Very good for examples like these.

agnivade:
>  I chose a slightly different wording, and used "parallel" instead of "concurrent" to make the connection with the -p more evident.

Technically, "concurrent" is actually the right word here. "parallel" means how many tasks are running in multiple cpus at the same time. "concurrent" means multiple tasks are running, but some might be waiting for some result, and not blocking other tasks from running. It's a subtle, but important difference. https://blog.golang.org/concurrency-is-not-parallelism

However, for tldr's scope and mission - which is aimed at newbies, I think this is a fair compromise to make.

I just left one comment here - https://github.com/tldr-pages/tldr/pull/1466#discussion_r137362766. Other than that, LGTM.

pages/common/pssh.md
waldyrious: It actually threw me off for a bit, as I didn't immediately realize it that parameter was meant to be an arbitrary command. My reasoning was that something like `echo`, `grep` or `ls` would make this point clearer.

agnivade: Interesting, it was pretty clear to me that this was an arbitrary command. Maybe my familiarity with the command biased me. 

Anyway, I still think `date` is the right command in this scenario, because if you have hundreds of servers, `ls -l` of the home directory will typically give you the same result. However, `date` will give you the current local time of that server. It's just a one line output and you also get to know what timezones your servers belong to ;)

waldyrious: How about something like `hostname -i`, then? I included the `-i` option to give the visual hint that this isn't just a token word but an actual command.

waldyrious:
@agnivade I think you migt find the discussion at https://github.com/JuliaLang/julia/pull/20486 pretty interesting. Let me know if you gain any insights from it :)

pages/common/pssh.md
agnivade: Sure.

PR #1467, opened 2017-09-03
latexmk: minor tweaks to descriptions

waldyrious:
Manually pinging @Nagiza for review, since @mention-bot seems to be failing again :/

----


pages/linux/latexmk.md
```diff
@@ -2,18 +2,18 @@
 
 > Automated LaTeX document generation.
 
-- Compile filename.tex to pdf:
+- Compile a LaTeX file to PDF:
```
sbrl: What about `Render a LaTeX file as a PDF`?

pages/linux/latexmk.md
```diff
@@ -2,18 +2,18 @@
...
 
 `latexmk -pdf {{filename.tex}}`
 
-- Force latexmk to generate document despite errors:
+- Force `latexmk` to generate a document even if there are errors:
 
 `latexmk -f {{filename.tex}}`
 
-- Clean up current directory from files generated by latexmk except; dvi, postscript and pdf:
+- Clean up the current directory from all files generated by `latexmk` except DVI, PostScript and PDF:
```
sbrl: `from all files generated` -> `by deleting all files generated`?

waldyrious: I agree the current wording is a little awkward, but I don't think avoiding it justifies making the sentence that long...

pages/linux/latexmk.md
waldyrious: That could work, but I don't see what's wrong with the current wording. Is it the "compile" that you want to get rid of?

pages/linux/latexmk.md
Nagiza: Other sources has phrased it " Clean up all temporary files in current directory"
But i am not sure it's better, because you only know which temporary files it's referring to if you are familiar with compiling .tex files.

pages/linux/latexmk.md
sbrl: It's mainly "to" which feels awkward actually, though compiling feels like compiling a program into it's executable :P

pages/linux/latexmk.md
sbrl: I'm mainly confused about the action referred to here. Are the files deleted? Or does something else happen to them?

pages/linux/latexmk.md
waldyrious: Well, it certainly is converting a plaintext file into a binary format :) would you be OK with "into a PDF file", or do you still prefer the "render" variant? I'm fine with either.

pages/linux/latexmk.md
waldyrious: Yes, they are deleted.

pages/linux/latexmk.md
sbrl: Yeah, "into a PDF file" would work just fine!

pages/linux/latexmk.md
sbrl: Ah, ok

waldyrious: So what do you propose here?

sbrl: Hrm, I'm not sure. Maybe "Cleanup and delete"? Maybe it's ok as it is.

waldyrious: How about "Delete all `latexmk`-generated temporary files from the current directory"? Maybe the detail on which ones are kept goes beyond the scope of tldr.

sbrl: yeah, that sounds better! Thanks :smile_cat:

Nagiza: Yeah, thats good and clear 👍

PR #1468, opened 2017-09-03
wodim: shorten main desc; clarify "burn" vs record

waldyrious:
----

### DIFF
--- a/pages/linux/wodim.md
+++ b/pages/linux/wodim.md
@@ -7,7 +7,7 @@
 
 `wodim --devices`
 
-- Record ("burn") pure CD-DA audio to a disc:
+- Record ("burn") an audio-only disc:
 
 `wodim dev=/dev/{{optical_drive}} -audio {{track*.cdaudio}}`
 
--- a/pages/linux/wodim.md
+++ b/pages/linux/wodim.md
@@ -1,14 +1,13 @@
 # wodim
 
-> Command for record/burning data using a optical drive to CD or DVD.
-> This command is also called cdrecord on some systems, but it is wodim underneath.
-> Some usage of wodim can cause destructive actions, such as erasing all the data on a disc.
+> Command (aliased as `cdrecord` on some systems) for recording data to CDs or DVDs.
+> Some invocations of wodim can cause destructive actions, such as erasing all the data on a disc.
 
-- Display optical drives available to wodim:
+- Display optical drives available to `wodim`:
 
 `wodim --devices`
 
-- Record pure CD-DA audio to a disc:
+- Record ("burn") pure CD-DA audio to a disc:
 
 `wodim dev=/dev/{{optical_drive}} -audio {{track*.cdaudio}}`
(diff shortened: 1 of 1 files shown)


pages/linux/wodim.md
```diff
@@ -1,14 +1,13 @@
 # wodim
 
-> Command for record/burning data using a optical drive to CD or DVD.
-> This command is also called cdrecord on some systems, but it is wodim underneath.
-> Some usage of wodim can cause destructive actions, such as erasing all the data on a disc.
+> Command (aliased as `cdrecord` on some systems) for recording data to CDs or DVDs.
```
sbrl: `aliased as` -> `aliased to`?

pages/linux/wodim.md
```diff
@@ -1,14 +1,13 @@
...
+> Some invocations of wodim can cause destructive actions, such as erasing all the data on a disc.
 
-- Display optical drives available to wodim:
+- Display optical drives available to `wodim`:
 
 `wodim --devices`
 
-- Record pure CD-DA audio to a disc:
+- Record ("burn") pure CD-DA audio to a disc:
```
sbrl: `pure CD-DA audio` -> `a pure CD-DA audio file`?

pages/linux/wodim.md
waldyrious: That suggests the wrong direction of the relationship, but it may be due to English nuances. For example, I feel it's appropriate to say "`foo` is a link to `bar`" or "`bar` is linked from `foo`" -- always making it clear that "foo" is the secondary object, and "bar" the primary one.

"Aliased to" kind of sounds like "linked to" to my ears, but again, I'm not a native English speaker, so I'll defer to whatever the community decides on this one.

pages/linux/wodim.md
waldyrious: Yeah, that makes it clearer, thanks. I'm not sure what "pure" means in this context, though. Any hints, @Nagiza, @agnivade?

Nagiza: As i understand it "pure" means that the burned CD only contains audio tracks and no data track, such as a album cover.
I found that the "CD-DA" is a unnecessary detail. 
`Record ("burn") pure CD-DA audio to a disc:` -> `Record ("burn") a audio only disc:`

pages/linux/wodim.md
sbrl: It's a really small point, so either will be just fine :-)

waldyrious: I guess my arguments didn't sway you, then :P let's agree to disagree?

pages/linux/wodim.md
waldyrious: Nice, that's even more readable. I'll change when I'm back on a desktop.

pages/linux/wodim.md
sbrl: Yep :P

PR #1470, opened 2017-09-04
moro: minor tweaks

waldyrious:
These PR lowercases the title, as we normally do, and makes the descriptions more in line with the langyuage other tldr-page entries use ("To do X" --> "Do X")

I'm still unsure about one the first example, "Run moro once right after starting the working day, and once more before leaving work": is that an instruction for the user, who should run that command manually every day, or is it a way to start moro which from then on will automatically do the start/stop dance by itself (using the default day start / day ent times)? /cc @omidfi.

----

- [_] The page (if new), does not already exist in the repo.

- [_] The page (if new), has been added to the correct platform folder:  

### DIFF
--- a/pages/common/moro.md
+++ b/pages/common/moro.md
@@ -22,9 +22,9 @@
 
 `moro note {{3 hours on project Foo}}`
 
-- Invoke `moro` without parameters a third time, to show a report of the current working day:
+- Show a report of time logs and notes for the current working day:
 
-`moro`
+`moro report`
 
 - Show a report of time logs and notes for all working days on record:
 
--- a/pages/common/moro.md
+++ b/pages/common/moro.md
@@ -22,7 +22,7 @@
 
 `moro note {{3 hours on project Foo}}`
 
-- Invoke `moro` without parameters a third time, to show a report of the current working day
+- Invoke `moro` without parameters a third time, to show a report of the current working day:
 
 `moro`
 
--- a/pages/common/moro.md
+++ b/pages/common/moro.md
@@ -2,22 +2,30 @@
 
 > Track work time.
 
-- Run moro once right after starting the working day, and once more before leaving work:
+- Invoke `moro` without parameters, to set the current time as the start of the working day:
 
 `moro`
 
-- Adjust the start of the working day:
+- Specify a custom time for the start of the working day:
 
 `moro hi {{09:30}}`
 
-- Adjust the end of the working day:
+- Invoke `moro` without parameters a second time, to set the current time at the end of the working day:
+
+`moro`
+
+- Specify a custom time for the end of the working day:
 
 `moro bye {{17:30}}`
 
-- Add a note on the working day:
+- Add a note on the current working day:
 
 `moro note {{3 hours on project Foo}}`
 
-- Show the full report of all records and notes:
+- Invoke `moro` without parameters a third time, to show a report of the current working day
+
+`moro`
+
+- Show a report of time logs and notes for all working days on record:
 
 `moro report --all`
--- a/pages/common/moro.md
+++ b/pages/common/moro.md
@@ -1,4 +1,4 @@
-# Moro
+# moro
 
 > Track work time.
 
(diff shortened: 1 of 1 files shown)


omidfi:
@waldyrious thanks for the edit. 
Yes, moro needs to be run manually every day.

pages/common/moro.md
```diff
@@ -1,23 +1,23 @@
...
 
 `moro hi {{09:30}}`
 
-- To adjust the end of the working day:
+- Adjust the end of the working day:
 
 `moro bye {{17:30}}`
 
-- To add a note on the working day:
+- Add a note on the working day:
```
sbrl: "the current", perhaps?

pages/common/moro.md
```diff
@@ -1,23 +1,23 @@
...
 
 `moro note {{3 hours on project Foo}}`
 
-- To see the full report of all records and notes:
+- Show the full report of all records and notes:
```
sbrl: `Show the full report` -> `Show a full report`

pages/common/moro.md
waldyrious: Sounds good to me, will change.

pages/common/moro.md
waldyrious: Agreed.

waldyrious:
@omidfi so, does one ever run `moro` with no parameters as shown in the first example, or does it need to be always `moro hi`, `moro bye`, etc.?

omidfi:
@waldyrious moro without argument works, that's the easier way to use it.
The first time moro is run it's exactly like: 

  moro hi {current time}

Second time it's equivalent to 

  moro bye {current time}

And the next time after that it will be the same as:

  moro report

waldyrious:
@omidfi thanks for the clarification. I'm not sure how to convey that information succinctly in this page... any suggestions appreciated!

waldyrious:
That's way too long :sweat_smile: the rule of thumb we use for tldr pages is that each example description should take up about a single line in an average-width console window.

omidfi:
Huh huh :D 

It's because it does 2 things based on the order of execution. What if we put it in 2 seperate examples? 

To register clock in time 
moro

To register clock out time: (clock in should be registered already) 
moro

waldyrious:
Hmm, that might work. Let me give it a try.

waldyrious:
@sbrl and @omidfi, please take a look at the new text and let me know what you think :)

omidfi:
Looks great!

pages/common/moro.md
```diff
@@ -1,23 +1,31 @@
-# Moro
+# moro
 
 > Track work time.
 
-- Run moro once right after starting the working day, and once more before leaving work:
+- Invoke `moro` without parameters, to set the current time as the start of the working day:
```
sbrl: Don't think that comma is needed.

waldyrious: It isn't needed in the sense that the sentence would be incorrect without it, but I think it eases the reading flow, by marking semantic pauses in long sentences like this one.

Otherwise the reader has to parse the entire sentence to understand its structure and break it down into its major components (clauses).

sbrl: True, let's leave it in then.

pages/common/moro.md
```diff
@@ -1,23 +1,31 @@
...
 
 `moro bye {{17:30}}`
 
-- To add a note on the working day:
+- Add a note on the current working day:
 
 `moro note {{3 hours on project Foo}}`
 
-- To see the full report of all records and notes:
+- Invoke `moro` without parameters a third time, to show a report of the current working day:
```
agnivade: Can we update this to just `moro report` ? And say - `Show a report of the current working day`.

This is much clearer and avoids the confusion of trying to recall when does `moro` record the time, and when will it show a report.

omidfi:
Maybe we can also turn them all to their explicit alternatives: 

clock in as the current time

moro hi

clock out at the current time

moro bye

Show a report of the current working day.

moro report

waldyrious:
Explicit is probably the safest approach for a first introduction to the tool, but perhaps a middle ground removing the third bare call as @agnivade suggests would be a good compromise. I'm personally leaning more to that., but happy to abide by whatever the consensus is.

agnivade:
Looks like @omidfi is okay with my approach.

omidfi:
Sure go ahead!

waldyrious:
Done :)

PR #1471, opened 2017-09-04
z: reword descriptions for clarity

waldyrious:
I'm still unsure about the "foo" then "bar" example. Can you clarify how that works, @ranyitz?

Also, I think we should include at least one (specific) regex example. And what about wildcards such as `*`? Do those work, or is `*` always interpreted in the regex sense?

----

- [_] The page (if new), does not already exist in the repo.

- [_] The page (if new), has been added to the correct platform folder:  

### DIFF
--- a/pages/common/z.md
+++ b/pages/common/z.md
@@ -1,27 +1,27 @@
 # z
 
-> Tracks your most used directories and enable quick jumping using regexes from the command line.
+> Tracks the most used directories and enables quickly navigating to them using string or regex patterns.
 
-- Go to a directory that contains foo:
+- Go to a directory that contains "foo" in the name:
 
 `z {{foo}}`
 
-- Go to a directory that contains foo and then bar:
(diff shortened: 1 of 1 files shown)


agnivade:
> I'm still unsure about the "foo" then "bar" example. Can you clarify how that works, @ranyitz?

I think it will go to a path which contains "foo" before "bar". 

> Also, I think we should include at least one (specific) regex example.

Yes, agree.

waldyrious:
>> I'm still unsure about the "foo" then "bar" example. Can you clarify how that works, @ranyitz?

> I think it will go to a path which contains "foo" before "bar".

So for example `path/to/foo/something-bar` would work, or would it have to be `path/to/foo/something/bar`, or something else?

sbrl:
The order they are generally used is foo -> bar -> baz as far as I know

sbrl:
Ping @agnivade. Are we ok to merge this, or are further changes required?

PR #1475, opened 2017-09-07
vault: add page

### LINKED ISSUE #1186: Add vault
Distributed secret management tool

https://www.vaultproject.io

fluxw42:
----

This should resolve #1186

### DIFF
--- a/pages/common/vault.md
+++ b/pages/common/vault.md
@@ -6,22 +6,22 @@
 
 `vault init`
 
-- Unseal or 'unlock' the vault by providing one of the key shares needed to access the encrypted data store:
+- Unseal (unlock) the vault, by providing one of the key shares needed to access the encrypted data store:
 
 `vault unseal {{key-share-x}}`
 
-- Authenticate the CLI client against the Vault server using an authentication token:
+- Authenticate the CLI client against the Vault server, using an authentication token:
 
 `vault auth {{authentication-token}}`
 
-- Store a new secret in the vault using the generic back-end, called 'secret':
+- Store a new secret in the vault, using the generic back-end called "secret":
 
 `vault write secret/{{hello}} value={{world}}`
 
-- Read a value from the vault using the generic back-end, called 'secret':
+- Read a value from the vault, using the generic back-end called "secret":
 
 `vault read secret/{{hello}}`
 
-- Seal or 'lock' the Vault server by removing the encryption key of the data store from memory:
+- Seal (lock) the Vault server, by removing the encryption key of the data store from memory:
 
 `vault seal`
--- a/pages/common/vault.md
+++ b/pages/common/vault.md
@@ -6,7 +6,7 @@
 
 `vault init -key-shares={{5}} -key-threshold={{2}}`
 
-- Unseal the vault:
+- Unseal the vault by providing one of the keyshares. Repeat with necessary key-shares until unsealed:
 
 `vault unseal {{key-share-x}}`
 
--- a/pages/common/vault.md
+++ b/pages/common/vault.md
@@ -1,0 +1,27 @@
+# vault
+
+> HashiCorp Vault, a tool for managing secrets.
+
+- Create a new vault, requiring at least 2 out of 5 keyshares to unseal:
+
+`vault init -key-shares={{5}} -key-threshold={{2}}`
+
+- Unseal a vault, by providing one of the keyshares. Repeat with necessary key-shares until unsealed:
+
+`vault unseal {{key-share-x}}`
+
+- Authenticate client against vault, using an authentication token:
+
+`vault auth {{authentication-token}}`
+
+- Store a new secret in the vault:
+
+`vault write {{secret/hello}} value={{world}}`
+
+- Read a secret from the vault:
+
+`vault read {{secret/hello}}`
+
+- Seal the vault:
+
+`vault seal`
--- a/pages/common/vault.md
+++ b/pages/common/vault.md
@@ -2,26 +2,26 @@
 
 > A CLI to interact with HashiCorp Vault.
 
-- Create a new vault, requiring at least 2 out of 5 keyshares to unseal:
+- Connect to a Vault server and initialize a new encrypted data store:
 
-`vault init -key-shares={{5}} -key-threshold={{2}}`
+`vault init`
 
-- Unseal the vault by providing one of the keyshares. Repeat with necessary key-shares until unsealed:
+- Unseal or 'unlock' the vault by providing one of the key shares needed to access the encrypted data store:
 
 `vault unseal {{key-share-x}}`
 
-- Authenticate client against vault, using an authentication token:
+- Authenticate the CLI client against the Vault server using an authentication token:
 
 `vault auth {{authentication-token}}`
 
-- Store a new secret in the vault using the generic back-end called secret:
+- Store a new secret in the vault using the generic back-end, called 'secret':
 
 `vault write secret/{{hello}} value={{world}}`
 
-- Read a value from the vault using the generic back-end called secret:
+- Read a value from the vault using the generic back-end, called 'secret':
 
 `vault read secret/{{hello}}`
 
-- Seal the vault again:
+- Seal or 'lock' the Vault server by removing the encryption key of the data store from memory:
 
 `vault seal`
--- a/pages/common/vault.md
+++ b/pages/common/vault.md
@@ -14,13 +14,13 @@
 
 `vault auth {{authentication-token}}`
 
-- Store a new secret in the vault:
+- Store a new secret in the vault using the generic back-end called secret:
 
-`vault write {{secret/hello}} value={{world}}`
+`vault write secret/{{hello}} value={{world}}`
 
-- Read a secret from the vault:
+- Read a value from the vault using the generic back-end called secret:
 
-`vault read {{secret/hello}}`
+`vault read secret/{{hello}}`
 
 - Seal the vault again:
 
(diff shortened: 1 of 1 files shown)


pages/common/vault.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Create a new vault, requiring at least 2 out of 5 keyshares to unseal:
+
+`vault init -key-shares={{5}} -key-threshold={{2}}`
+
+- Unseal a vault, by providing one of the keyshares. Repeat with necessary key-shares until unsealed:
+
+`vault unseal {{key-share-x}}`
+
+- Authenticate client against vault, using an authentication token:
```
sbrl: What precisely is meant by "authenticate client against vault"?

pages/common/vault.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Store a new secret in the vault:
+
+`vault write {{secret/hello}} value={{world}}`
+
+- Read a secret from the vault:
+
+`vault read {{secret/hello}}`
+
+- Seal the vault:
```
sbrl: Perhaps "seal the vault again"? Just to make it clear :P

pages/common/vault.md
```diff
@@ -0,0 +1,27 @@
+# vault
+
+> HashiCorp Vault, a tool for managing secrets.
+
+- Create a new vault, requiring at least 2 out of 5 keyshares to unseal:
+
+`vault init -key-shares={{5}} -key-threshold={{2}}`
+
+- Unseal a vault, by providing one of the keyshares. Repeat with necessary key-shares until unsealed:
```
sbrl: Don't think that the comment is needed here.

pages/common/vault.md
fluxw42: HashiCorps Vault is a server or daemon process. You have to authenticate the client process 'vault' before you can do any key-related actions.

pages/common/vault.md
fluxw42: Ok, I'll make this shorter.

pages/common/vault.md
```diff
@@ -0,0 +1,27 @@
...
+
+`vault unseal {{key-share-x}}`
+
+- Authenticate client against vault, using an authentication token:
+
+`vault auth {{authentication-token}}`
+
+- Store a new secret in the vault:
+
+`vault write {{secret/hello}} value={{world}}`
```
agnivade: Is the "/" necessary ? Can it be just "hello" ?

pages/common/vault.md
```diff
@@ -0,0 +1,27 @@
+# vault
+
+> HashiCorp Vault, a tool for managing secrets.
```
agnivade: Can we remove "HashiCorp Vault" ? It does not add any value to the explanation.

pages/common/vault.md
fluxw42: Vault uses different back-ends (secret, auth, audit, ..), which are modeled as a kind of file system. The 'secret/*' prefix in the example I provided points to their generic back-end to read and write arbitrary secrets.

You can mount other types of back-ends, eg. AWS, PKI, SSH, TOTP, ... but that would be too much to demonstrate in a tldr.

pages/common/vault.md
fluxw42: Would it be better if I changed it like this?

    > A CLI to interact with HashiCorp Vault

The 'vault' tool does nothing on it's own, you only use it to interact with a Vault server (locally or remotely).

pages/common/vault.md
fluxw42: Done.

pages/common/vault.md
agnivade: I see. So I recommend changing the description to `Store a new secret in the vault using the generic back-end called secret`

And move out the "secret" part out of the tokens- `vault write secret{{/hello}} value={{world}}`

fluxw42: Makes sense: Fixed!

agnivade:
Looks okay to me. However, I feel that the language can still be improved a bit. Waiting for @waldyrious' approval.

pages/common/vault.md
sbrl: Oops  I meant comma, @Florent2! The comment is needed lol

pages/common/vault.md
sbrl: Ahhh ok

pages/common/vault.md
sbrl: Awesome :smiley_cat:

sbrl:
Oops! Looks like  mistyped in my comment above. See the inline comment.

pages/common/vault.md
fluxw42: Haha :-) Well, the *comment* is back but the *comma* is gone.

fluxw42: Haha :-) Well, the *comment* is back but the *comma* is gone.

pages/common/vault.md
```diff
@@ -0,0 +1,27 @@
+# vault
+
+> A CLI to interact with HashiCorp Vault.
+
+- Create a new vault, requiring at least 2 out of 5 keyshares to unseal:
```
waldyrious: I'm a little wary of starting out with this much jargon from the get-go ("keyshares"? "unseal"?).

Would it be possible to start with a simpler example (say, just `vault init`, relying on the defaults for the parameters)?

pages/common/vault.md
```diff
@@ -0,0 +1,27 @@
...
+
+`vault init -key-shares={{5}} -key-threshold={{2}}`
+
+- Unseal the vault by providing one of the keyshares. Repeat with necessary key-shares until unsealed:
```
waldyrious: What exactly does "unseal" mean in this context? Would it be possible to add a synonym or equivalent expression in parenthesis?

Also, we probably should use consistent spelling ("keyshares" or "key-shares").

pages/common/vault.md
```diff
@@ -0,0 +1,27 @@
...
+
+`vault unseal {{key-share-x}}`
+
+- Authenticate client against vault, using an authentication token:
```
waldyrious: I'm not really comfortable with the wording "Authenticate client against vault". Any change we could word it some other way?

pages/common/vault.md
```diff
@@ -0,0 +1,27 @@
...
+
+`vault auth {{authentication-token}}`
+
+- Store a new secret in the vault using the generic back-end called secret:
```
waldyrious: Please correct me if I'm reading this wrong, but is "secret" the name of the generic/default back-end? If so, maybe add quote marks around it to make that clearer, and perhaps a comma before "called".

pages/common/vault.md
```diff
@@ -0,0 +1,27 @@
...
+
+`vault write secret/{{hello}} value={{world}}`
+
+- Read a value from the vault using the generic back-end called secret:
+
+`vault read secret/{{hello}}`
+
+- Seal the vault again:
```
waldyrious: Here we could take the opportunity that the description is short, to explain what seal/unseal does exactly. E.g., does this command unmount a folder, encrypt a file, mark stuff as read-only, or something else?

waldyrious:
Overall looks good (nice work, @fluxw42, and welcome back -- hadn't seen you in a while!), but the descriptions can be improved IMO. I've left a few inline comments.

pages/common/vault.md
waldyrious: c.f. https://github.com/tldr-pages/tldr/pull/1475#discussion_r137553819

pages/common/vault.md
fluxw42: Using `vault init` would work, I'll simplify the example. This uses a default of 5 key shares requiring 3 of them to unseal.

pages/common/vault.md
fluxw42: When a vault is 'sealed' it means that the master key is missing to access the stored key- and config data. The master key is split up in a number of key shares (by default 5) and can only be reconstructed when a configured number of shares are combined again (by default 3 out of 5).

Unsealing is the process of combining these key shares to construct the original master key again. So when a vault is *sealed*, nobody can access it - not even authenticated clients -  since the master key is missing. When a vault is *unsealed*, it can be used by authenticated clients.

Would `unlocking` as a synonym for `unsealing` be better?

pages/common/vault.md
fluxw42: It deletes the master encryption key from memory, rendering all data inaccessible until someone *unseals* (or *unlocks*) the vault again.

fluxw42:
I think I've addressed all remarks that came up during the review. It's harder than expected to explain these basic commands in one concise sentence, there's a lot going on under the hood ...

Please take a look at it and let me know if I could still change something :)

waldyrious:
Looks good @fluxw42. We understand your pain, we have to deal with that struggle all the time when reviewing PRs. I think the page is quite acceptable as it is, but I'd only replace the single quote marks by double quotes, and use parenthesis for the synonyms, e.g.

> Seal (lock) the Vault server by removing the encryption key of the data store from memory:

Finally, you might want to throw in a few commas, to make the logical splits of clauses in the longer sentences more evident.

fluxw42:
Should be fixed in 0ddfb4f: 
 * Single quotes `'` in description are now double quotes `"`.
 * Used parenthesis for *lock* and *unlock* synonyms.
 * Added commas where necessary, also removed two superfluous ones.

waldyrious:
@sbrl, the ball's on your court :smile:

sbrl:
Looks ok here :slightly_smiling_face:

PR #1477, opened 2017-09-08
ffmpeg: add extract frame example and other tweaks

agnivade:
- Applying the proper file path convention to all examples.
- Re-arranged the examples to have -i param to be the first one.
- Brought the file extension out from the token where the
file type is explicitly mentioned in the description.

----

- ~~[ ] The page (if new), does not already exist in the repo.~~

- ~~[ ] The page (if new), has been added to the correct platform folder:  

### DIFF
--- a/pages/common/ffmpeg.md
+++ b/pages/common/ffmpeg.md
@@ -8,16 +8,16 @@
 
 - Convert frames from a video or GIF into individual numbered images:
 
-`ffmpeg -i {{path/to/video_or_gif}} {{image%d.png}}`
+`ffmpeg -i {{path/to/video|gif}} {{image%d.png}}`
 
 - Combine numbered images (image1.jpg, image2.jpg, etc) into a video or GIF:
 
-`ffmpeg -i {{path/to/image%d.jpg}} -f image2 {{video.mpg_or_video.gif}}`
+`ffmpeg -i {{path/to/image%d.jpg}} -f image2 {{video.mpg|video.gif}}`
 
 - Extract a single frame from a video at time mm:ss and save it as a 128x128 resolution image:
 
-`ffmpeg -i {{path/to/video}} -s {{mm:ss}} -frames 1 -s {{128x128}} -f image2 {{image.png}}`
+`ffmpeg -i {{path/to/video}} -ss {{mm:ss}} -frames 1 -s {{128x128}} -f image2 {{image.png}}`
 
 - Convert AVI video to MP4. AAC Audio @ 128kbit, Video @ 1250Kbit:
 
-`ffmpeg -i {{path/to/video}}.avi -acodec libfaac -ab 128k -vcodec mpeg4 -b 1250K {{out}}.mp4`
+`ffmpeg -i {{path/to/input_video}}.avi -acodec libfaac -ab 128k -vcodec mpeg4 -b 1250K {{output_video}}.mp4`
--- a/pages/common/ffmpeg.md
+++ b/pages/common/ffmpeg.md
@@ -4,20 +4,20 @@
 
 - Extract the sound from a video and save it as MP3:
 
-`ffmpeg -i {{path/to/video}} -vn -ar 44100 -ac 2 -ab 192 -f mp3 {{sound}}.mp3`
+`ffmpeg -i {{video}} -vn -ar 44100 -ac 2 -ab 192 -f mp3 {{sound}}.mp3`
 
 - Convert frames from a video or GIF into individual numbered images:
 
-`ffmpeg -i {{path/to/video|gif}} {{image%d.png}}`
+`ffmpeg -i {{video|gif}} {{frame_%d.png}}`
 
-- Combine numbered images (image1.jpg, image2.jpg, etc) into a video or GIF:
+- Combine numbered images (frame_1.jpg, frame_2.jpg, etc) into a video or GIF:
 
-`ffmpeg -i {{path/to/image%d.jpg}} -f image2 {{video.mpg|video.gif}}`
+`ffmpeg -i {{frame_%d.jpg}} -f image2 {{video.mpg|video.gif}}`
 
 - Extract a single frame from a video at time mm:ss and save it as a 128x128 resolution image:
 
-`ffmpeg -i {{path/to/video}} -ss {{mm:ss}} -frames 1 -s {{128x128}} -f image2 {{image.png}}`
+`ffmpeg -i {{video}} -ss {{mm:ss}} -frames 1 -s {{128x128}} -f image2 {{image.png}}`
 
 - Convert AVI video to MP4. AAC Audio @ 128kbit, Video @ 1250Kbit:
 
-`ffmpeg -i {{path/to/input_video}}.avi -acodec libfaac -ab 128k -vcodec mpeg4 -b 1250K {{output_video}}.mp4`
+`ffmpeg -i {{input_video}}.avi -acodec libfaac -ab 128k -vcodec mpeg4 -b 1250K {{output_video}}.mp4`
--- a/pages/common/ffmpeg.md
+++ b/pages/common/ffmpeg.md
@@ -4,16 +4,20 @@
 
 - Extract the sound from a video and save it as MP3:
 
-`ffmpeg -i {{video_filename}} -vn -ar 44100 -ac 2 -ab 192 -f mp3 {{sound.mp3}}`
+`ffmpeg -i {{path/to/video}} -vn -ar 44100 -ac 2 -ab 192 -f mp3 {{sound}}.mp3`
 
 - Convert frames from a video or GIF into individual numbered images:
 
-`ffmpeg -i {{video_or_gif_filename}} {{image%d.png}}`
+`ffmpeg -i {{path/to/video_or_gif}} {{image%d.png}}`
 
 - Combine numbered images (image1.jpg, image2.jpg, etc) into a video or GIF:
 
-`ffmpeg -f image2 -i {{image%d.jpg}} {{video.mpg_or_video.gif}}`
+`ffmpeg -i {{path/to/image%d.jpg}} -f image2 {{video.mpg_or_video.gif}}`
+
+- Extract a single frame from a video at time mm:ss and save it as a 128x128 resolution image:
+
+`ffmpeg -i {{path/to/video}} -s {{mm:ss}} -frames 1 -s {{128x128}} -f image2 {{image.png}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/ffmpeg.md
```diff
@@ -4,16 +4,20 @@
 
 - Extract the sound from a video and save it as MP3:
 
-`ffmpeg -i {{video_filename}} -vn -ar 44100 -ac 2 -ab 192 -f mp3 {{sound.mp3}}`
+`ffmpeg -i {{path/to/video}} -vn -ar 44100 -ac 2 -ab 192 -f mp3 {{sound}}.mp3`
```
sbrl: Is the filename required to me `.mp3`?

pages/common/ffmpeg.md
```diff
@@ -4,16 +4,20 @@
...
 
-`ffmpeg -i {{video_or_gif_filename}} {{image%d.png}}`
+`ffmpeg -i {{path/to/video_or_gif}} {{image%d.png}}`
 
 - Combine numbered images (image1.jpg, image2.jpg, etc) into a video or GIF:
 
-`ffmpeg -f image2 -i {{image%d.jpg}} {{video.mpg_or_video.gif}}`
+`ffmpeg -i {{path/to/image%d.jpg}} -f image2 {{video.mpg_or_video.gif}}`
+
+- Extract a single frame from a video at time mm:ss and save it as a 128x128 resolution image:
```
sbrl: Very cool example! :smiley_cat:

agnivade: I know ! I just used it at work. The first thing I did was check `tldr ffmpeg` and saw this wasn't there. Hence this PR. 😄

pages/common/ffmpeg.md
agnivade: The example mentions "save it as MP3". This is to make that clearer. And since its in `common/`, so sometimes Windows barfs if it doesn't see `.mp3` extension.

sbrl: Ahhh ok. Cool :smiley_cat:

agnivade:
ping @waldyrious

pages/common/ffmpeg.md
```diff
@@ -4,16 +4,20 @@
...
 
 - Convert frames from a video or GIF into individual numbered images:
 
-`ffmpeg -i {{video_or_gif_filename}} {{image%d.png}}`
+`ffmpeg -i {{path/to/video_or_gif}} {{image%d.png}}`
 
 - Combine numbered images (image1.jpg, image2.jpg, etc) into a video or GIF:
 
-`ffmpeg -f image2 -i {{image%d.jpg}} {{video.mpg_or_video.gif}}`
+`ffmpeg -i {{path/to/image%d.jpg}} -f image2 {{video.mpg_or_video.gif}}`
```
waldyrious: Nice change, the order is indeed more intuitive this way. :+1:

waldyrious: Actually, on second thought, I'm unsure what `-f image2` means here. Additionally, we should probably use the pipe syntax `{{video.mpg|video.gif}}`.

agnivade: `-f image2` is ffmpeg's internal syntax. Even I don't know what its supposed to mean. I am guessing down the line, their image decoder changed, but they couldn't remove the old codecs due to compatibility, hence this weird flag. I could be totally wrong though.

Good suggestion on the pipe syntax.

pages/common/ffmpeg.md
```diff
@@ -4,16 +4,20 @@
...
+
+- Extract a single frame from a video at time mm:ss and save it as a 128x128 resolution image:
+
+`ffmpeg -i {{path/to/video}} -s {{mm:ss}} -frames 1 -s {{128x128}} -f image2 {{image.png}}`
```
waldyrious: Does this command really use the `-s` option twice, with different meanings? If so, we definitely should clarify that -- perhaps by using the long option name, if the description would be too long.

pages/common/ffmpeg.md
```diff
@@ -4,16 +4,20 @@
...
 
 - Convert AVI video to MP4. AAC Audio @ 128kbit, Video @ 1250Kbit:
 
-`ffmpeg -i {{in.avi}} -acodec libfaac -ab 128k -vcodec mpeg4 -b 1250K {{out.mp4}}`
+`ffmpeg -i {{path/to/video}}.avi -acodec libfaac -ab 128k -vcodec mpeg4 -b 1250K {{out}}.mp4`
```
waldyrious: I think the `in.avi` / `out.mp4` symmetry should be preserved. maybe `path/to/input.avi` and `path/to/output.mp4`?

pages/common/ffmpeg.md
agnivade: So sorry ! The first `-s` is supposed to be `-ss` :rofl:

agnivade:
@waldyrious - I have kept the word "video" because the other examples using "sound", "image" etc. And kept the "path/to" convention only in the source param to keep it consistent.

pages/common/ffmpeg.md
```diff
@@ -4,16 +4,20 @@
 
 - Extract the sound from a video and save it as MP3:
 
-`ffmpeg -i {{video_filename}} -vn -ar 44100 -ac 2 -ab 192 -f mp3 {{sound.mp3}}`
+`ffmpeg -i {{path/to/video}} -vn -ar 44100 -ac 2 -ab 192 -f mp3 {{sound}}.mp3`
 
 - Convert frames from a video or GIF into individual numbered images:
 
-`ffmpeg -i {{video_or_gif_filename}} {{image%d.png}}`
+`ffmpeg -i {{path/to/video|gif}} {{image%d.png}}`
```
waldyrious: Having `path/to/` in the first part of the token but not on the second looks confusing, I think we can remove it altogether in this case, for simplicity.

pages/common/ffmpeg.md
```diff
@@ -4,16 +4,20 @@
...
 
 - Combine numbered images (image1.jpg, image2.jpg, etc) into a video or GIF:
 
-`ffmpeg -f image2 -i {{image%d.jpg}} {{video.mpg_or_video.gif}}`
+`ffmpeg -i {{path/to/image%d.jpg}} -f image2 {{video.mpg|video.gif}}`
```
waldyrious: The "numbered images (image1.jpg, **image2**.jpg, etc)" part is confusing regarding `-f image2`. Perhaps we could use a different prefix instead? Say, `frame_%d.png` (with an underscore to make the `%d` more visible, and yeah, I also would suggest using png rather than jpg just to encourage lossless formats).

pages/common/ffmpeg.md
```diff
@@ -4,16 +4,20 @@
...
+
+- Extract a single frame from a video at time mm:ss and save it as a 128x128 resolution image:
+
+`ffmpeg -i {{path/to/video}} -ss {{mm:ss}} -frames 1 -s {{128x128}} -f image2 {{image.png}}`
 
 - Convert AVI video to MP4. AAC Audio @ 128kbit, Video @ 1250Kbit:
 
-`ffmpeg -i {{in.avi}} -acodec libfaac -ab 128k -vcodec mpeg4 -b 1250K {{out.mp4}}`
+`ffmpeg -i {{path/to/input_video}}.avi -acodec libfaac -ab 128k -vcodec mpeg4 -b 1250K {{output_video}}.mp4`
```
waldyrious: Same here: I think the lack of `path/to` in both cases may raise more questions than its presence settles. I'd propose either using it on both, or removing it on both.

pages/common/ffmpeg.md
agnivade: Good point.

Regarding png/jpg, it actually depends on your use case. Using jpgs for real life photographs makes sense, otherwise you will blow up storage if you save them in pngs. pngs are more suitable for line drawings, clips and gifs. So, I would suggest keeping jpg as is.

waldyrious: Yeah but are jpgs actually used that commonly for frames that can be assembled into meaningful videos? In my experience such frames tend to be more like (temporary) generated images rather than actual photographs, and the final video is then compressed in order to take better advantage of the temporal coherence between frames.

I mean, I know that nowadays there are many cameras with the continuous shooting / sport mode feature, but I wonder how likely it is that people take such sequences only to merge them later as a video (as opposed to actually shooting video in the first place).

But of course, this is a minor point. Let me know what you think.

agnivade: > Yeah but are jpgs actually used that commonly for frames that can be assembled into meaningful videos? 

Yes, I believe so. I have used the exact same thing to create a gif from a series of images. And I had to use jpg images.

waldyrious:
As an additional side note, what you think about using the `(input|output)_(video|image|sound).ext` format for all tokens? I think it may make it easier for people to reason about the different examples and how they relate to each other. It's also more explicit, which is always good when that doesn't make things too verbose :)

agnivade:
Hmm .. I dont know. I think having input_ and output_ makes sense when both of them are of the same type. i.e. if both input and outputs are a video or audio or image. 

If you are converting a video to an video, isn't it clear enough if we just mention {{video}} and {{audio}} ?

For eg, in this one - 
```
- Extract the sound from a video and save it as MP3:

`ffmpeg -i {{video}} -vn -ar 44100 -ac 2 -ab 192 -f mp3 {{sound}}.mp3`
```
it just sounds odd if we change it to - 
```
`ffmpeg -i {{input_video}} -vn -ar 44100 -ac 2 -ab 192 -f mp3 {{output_sound}}.mp3`
```

I don't have a strong logic to back my viewpoint. Just what sounds right to me.

pages/common/ffmpeg.md
waldyrious: Ok :+1:

waldyrious:
> I think having input_ and output_ makes sense when both of them are of the same type. i.e. if both input and outputs are a video or audio or image.

Makes sense. I was mostly aiming at consistency within each example. So in the one you mention, I'd rather have it as

    ffmpeg -i {{video.mp4}} -vn -ar 44100 -ac 2 -ab 192 -f mp3 {{sound}}.mp3

(with the extension for the video inside the token marks, since it's variable, unlike the mp3 extension).

Same for the other example:

    ffmpeg -i {{video|gif}} {{frame_%d.png}}

would become

    ffmpeg -i {{video.mpg|video.gif}} {{frame_%d.png}}

and as a bonus, that would then be consistent with the next example :)

agnivade:
Because the description mentioned "convert avi video to mp4". Therefore, for clarity's sake.

PR #1478, opened 2017-09-08
groff: add page

waldyrious:
Sources: https://embswit.wordpress.com/2013/02/08/render-man-page-from-source/, and the EXAMPLES section of `man groff` :)

----

### DIFF
--- a/pages/common/groff.md
+++ b/pages/common/groff.md
@@ -1,0 +1,24 @@
+# groff
+
+> Typesetting program that reads plain text mixed with formatting commands and produces formatted output.
+> It is the GNU replacement for the troff and nroff Unix commands for text formatting.
+
+- Render a man page as plain text, and display the result:
+
+`groff -man -T utf8 {{manpage.1}}`
+
+- Render a man page using the ASCII output device, and display it using a pager:
+
+`groff -man -T ascii {{manpage.1}} | less`
+
+- Render a man page into an HTML file:
+
+`groff -man -T html {{manpage.1}} > page.html`
+
+- Process a roff file using the `tbl` and `pic` preprocessors, and the `me` macro set:
+
+`groff -t -p -me -T utf8 {{foo.me}}`
+
+- Run a `groff` command with preprocessor and macro options guessed by the `grog` utility:
+
+`eval "$(grog -T utf8 {{foo.me}})"`
--- a/pages/common/groff.md
+++ b/pages/common/groff.md
@@ -1,7 +1,7 @@
 # groff
 
 > Typesetting program that reads plain text mixed with formatting commands and produces formatted output.
-> It is the GNU replacement for the troff and nroff Unix commands for text formatting.
+> It is the GNU replacement for the `troff` and `nroff` Unix commands for text formatting.
 
 - Render a man page as plain text, and display the result:
 
@@ -13,7 +13,7 @@
 
 - Render a man page into an HTML file:
 
-`groff -man -T html {{manpage.1}} > page.html`
+`groff -man -T html {{manpage.1}} > {{page.html}}`
 
 - Process a roff file using the `tbl` and `pic` preprocessors, and the `me` macro set:
 


pages/common/groff.md
```diff
@@ -0,0 +1,24 @@
...
+
+`groff -man -T utf8 {{manpage.1}}`
+
+- Render a man page using the ASCII output device, and display it using a pager:
+
+`groff -man -T ascii {{manpage.1}} | less`
+
+- Render a man page into an HTML file:
+
+`groff -man -T html {{manpage.1}} > page.html`
```
agnivade: tokens around page.html

pages/common/groff.md
```diff
@@ -0,0 +1,24 @@
+# groff
+
+> Typesetting program that reads plain text mixed with formatting commands and produces formatted output.
+> It is the GNU replacement for the troff and nroff Unix commands for text formatting.
```
agnivade: Should we wrap troff and nroff with backticks ?

pages/common/groff.md
```diff
@@ -0,0 +1,24 @@
...
+
+`groff -man -T html {{manpage.1}} > page.html`
+
+- Process a roff file using the `tbl` and `pic` preprocessors, and the `me` macro set:
+
+`groff -t -p -me -T utf8 {{foo.me}}`
+
+- Run a `groff` command with preprocessor and macro options guessed by the `grog` utility:
+
+`eval "$(grog -T utf8 {{foo.me}})"`
```
agnivade: Hmm .. this seems a bit arcane. Does the `grog` command itself output the `groff` command ?

waldyrious: Yeah. This example is actually [mentioned in the groff manpage](http://man.gnu.org.ua/manpage/?1+groff#EXAMPLES).

pages/common/groff.md
waldyrious: Yes, good catch.

waldyrious:
By the way, I noticed there's additional example-based documentation in the manual [current version](http://www.hep.by/gnu/groff/Invocation-Examples.html#fnd-1), [old version](http://web.cecs.pdx.edu/~trent/gnu/groff/groff.html#SEC13).

Do you see anything there that could be added here, or even replace / improve some of the examples in this page?

pages/common/groff.md
sbrl: That's weird! Looks ok then to me. Either that or something like `grog command | bash` I guess. I can see how `eval` is more explicit though - I vote for eval then.

pages/common/groff.md
```diff
@@ -0,0 +1,24 @@
+# groff
+
+> Typesetting program that reads plain text mixed with formatting commands and produces formatted output.
+> It is the GNU replacement for the `troff` and `nroff` Unix commands for text formatting.
+
+- Render a man page as plain text, and display the result:
+
+`groff -man -T utf8 {{manpage.1}}`
+
+- Render a man page using the ASCII output device, and display it using a pager:
```
sbrl: `device`? That's a bit strange. Could we not use `plugin` or `module` here? I for one associate `device` with a physical device or chip on my motherboard.

waldyrious: "output device" is the term they use in the [documentation](http://man.gnu.org.ua/manpage/?1+groff#OPTIONS):

```
−T dev: Set output device to dev.

  Real devices in groff are:

  - dvi     TeX DVI format.
  - html
  - xhtml   HTML and XHTML output.
  - lbp     Canon CAPSL printers (LBP-4 and LBP-8 series laser printers).
  - lj4     HP LaserJet4 compatible (or other PCL5 compatible) printers.
  - ps      PostScript output.

  For the following TTY output devices, −T selects the output encoding:
	
  - ascii   7bit ASCII.
  - cp1047  Latin-1 character set for EBCDIC hosts.
  - latin1  ISO 8859-1.
  - utf8    Unicode character set in UTF-8 encoding.

  (...)

  The default device is ps.
```

sbrl: Ah, ok! That's really weird.

waldyrious: Yeah... I guess when some of these older stuff were created in a time when naming conventions weren't as established as they are today (e.g. maybe "device" wasn't as strongly correlated to physical devices as it is today)

PR #1479, opened 2017-09-09
curl: display header information

andrewlouis93:
* Knowing how to only display request & response header information is pretty key imho. 
* Saw some related discussion over here: https://github.com/tldr-pages/tldr/issues/1406, and I think this addresses most concerns raised there.

r/ @waldyrious


andrewlouis93:
closing in favour of https://github.com/tldr-pages/tldr/pull/1480

PR #1480, opened 2017-09-09
curl: custom hostname resolution & only display header information

andrewlouis93:
This command combines two curl functionalities that I use very often. 
* The first is to configure hostname resolution to some custom IP - functionally a command line `/etc/hosts` alternative.
* The second is to only display request & response header information. Saw some related discussion over here: #1406, and I think this addresses most concerns raised there.

r/ @waldyrious

### DIFF
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -27,6 +27,10 @@
 
 `curl -d {{'{"name":"bob"}'}} -H {{'Content-Type: application/json'}} {{http://example.com/users/1234}}`
 
+- Pass a custom IP address for a specific host and port pair:
+
+`curl --resolve example.com:443:192.168.0.1 -vo /dev/null https://example.com/filename`
+
 - Pass a user name and password for server authentication:
 
 `curl -u myusername:mypassword {{http://example.com}}`
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -27,7 +27,7 @@
 
 `curl -d {{'{"name":"bob"}'}} -H {{'Content-Type: application/json'}} {{http://example.com/users/1234}}`
 
-- Pass a custom IP address for a specific host and port pair:
+- Pass a custom IP address for a specific host and port pair, and only display header information:
 
 `curl --resolve example.com:443:192.168.0.1 -vo /dev/null https://example.com/filename`
 


agnivade:
Thank you for the PR ! Unfortunately this exceeds the maximum no. of examples that we allow for a command, which is 8. This was on the PR template, which you might have noticed while creating it.

To add your example, we would have to remove something else. Over the years, we have had lots of updates to the curl page. I think it comprises 8 of the most used options a beginner is going to use. Let me know what you think.

sbrl:
Oooh that's a nice example! As @agnivade said though, we've got a maximum number of examples per page of 8. Is there no way we can split curl up into multiple pages? There's just so much it can do, and 8 examples is rather limiting I agree. At the same the same time though, I do understand why the limit exists.

agnivade:
It's true `curl` can do so much more. And so can lot of other commands too. But we have to draw a line somewhere. And that is the fun of tldr, trying to distill it down to the most common and useful examples. ;)

andrewlouis93:
@sbrl @agnivade I totally missed the limit of 8 examples per page 🙈 

imho since there's a limit of 8 examples per command, the goal should be to maximize (common) use-case coverage in each example - I think [example 3](https://github.com/tldr-pages/tldr/pull/1480/files#diff-1120b1c2e3e5874195d879c597bcb855R16) does this really well.

That being said, I think we could potentially merge [example 7](https://github.com/tldr-pages/tldr/pull/1480/files#diff-1120b1c2e3e5874195d879c597bcb855R34) into another example. For instance:

> 
>  - Download a file from a basic auth protected endpoint, and save the output under the filename indicated by the URL:
>  		 
>  `curl -u myusername:mypassword -O {{http://example.com/filename}}`
> 

I can only speak from lived experience - but, being able to easily produce header information is a major use-case, and should be on this list even if the ability to specify custom hostname resolution isn't.

waldyrious:
@andrewlouis93, we normally aim to combine options when they are somehow related, not as a workaround to push in more content than the limit of examples would normally allow.

In other words, the limit is a specific amount and we could choose to interpret it literally (as in "the letter of the law"), but the *purpose* of that limit is to ensure the complexity and readability of the pages remains within the scope of the tldr-pages project (i.e. "the spirit of the law"). It seems to me that the merging you're proposing falls closer to the former than to the latter. Do you agree?

andrewlouis93:
@waldyrious Fair, and I agree that I'm arguing to shoehorn that functionality in there - but I really think the ability to only show request/response headers via `curl` should be in here somewhere :/ There's a reason you guys have had a lot of discussion about it - and if the only path forwards is to push the limit of the examples a little bit - I'd be okay with it, but then again I'm pretty new to this community :)

waldyrious:
@andrewlouis93 I don't think anyone's arguing that `-I` isn't a useful curl option. But is it more relevant for an introductory page than any of the existing examples? Again, we're not aiming for completeness here (it would be impossible anyway, so we have to draw the line somewhere).

I actually think 8 examples is already a lot -- I would actually be more inclined to support lowering the limit than raising it. So within the current limits, would you say that any of the current examples in the page should be removed in favor of the `-I` one?

agnivade:
ping @andrewlouis93

agnivade:
@andrewlouis93 - I am going ahead and closing this. Please feel free to reopen if you have differing opinions than what @waldyrious has mentioned.

PR #1490, opened 2017-09-13
zip: minor change in -x example

ericbn:
Use quotes instead of escaping the `*` in the `-x` example. Also making examples more meaningful, and with token highlighting.

### DIFF
--- a/pages/common/zip.md
+++ b/pages/common/zip.md
@@ -8,7 +8,7 @@
 
 - E[x]clude unwanted files from being added to the compressed archive:
 
-`zip -r {{compressed.zip}} {{path/to/dir}} -x \*.git\* \*node_modules\* ...`
+`zip -r {{compressed.zip}} {{path/to/dir}} -x {{path/to/exclude}}`
 
 - Archive a directory and its contents with the highest level [9] of compression:
 


ericbn:
(I personally prefer single quotes, but that is a rabbit-hole discussion) 😄

pages/common/zip.md
```diff
@@ -8,7 +8,7 @@
 
 - E[x]clude unwanted files from being added to the compressed archive:
 
-`zip -r {{compressed.zip}} {{path/to/dir}} -x \*.git\* \*node_modules\* ...`
+`zip -r {{compressed.zip}} {{path/to/dir}} -x "{{.*}}" "{{node_modules}}" ...`
```
agnivade: I think we should keep the ".git". It's common enough to warrant a mention here.

ericbn:
So rewriting the original example, which is:

    zip -r {{compressed.zip}} {{path/to/dir}} -x \*.git\* \*node_modules\* ...

using quoted strings instead of quoting each `*` character, we would have:

    zip -r compressed.zip path/to/dir -x "*.git*" "*node_modules*" ...

which is where I think it is a confusing example. We don't need to glob `*.git*` or `*node_modules*`. They are directory names that don't have varying prefixes or suffixes, so just saying `.git` and `node_modules` is enough. That's why I wanted an example that makes more sense, while still having a glob:

    zip -r compressed.zip path/to/dir -x ".*" "node_modules" ...

So any file/directory that starts with `.` and the `node_modules` one. The double quotes for `node_modules` are of course not needed, and just there for consistency. Maybe both should use globs in a better example...

agnivade:
Understood. Then I think we should keep the node_modules and .git and add another option for globbing. Something like - 
```
zip -r {{compressed.zip}} {{path/to/dir}} -x "{{.git}}" "{{node_modules}}" "{{tmp_*}}"
```

pages/common/zip.md
waldyrious: I would normally disagree, since hidden files of any kind are, by definition, more common than the `.git` directory. However, in this case I'm afraid the wildcard expression can be confusing, since that same combination (`.*`) has a specific meaning as a regex pattern.

waldyrious:
I agree with @agnivade: a clearer glob pattern (i.e. other than `.*`) would make this easier to understand.

ericbn:
I think we're not quite following the 5th item of the [Guidelines](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines) by getting on the details and variations that can be given to the `-x` parameters.

Maybe just

    zip -r {{compressed.zip}} {{path/to/dir}} -x {{path/to/exclude}}

ericbn:
Updated commit with simpler example

agnivade:
ping @sbrl

sbrl:
Looks ok to me! 😺

PR #1427, opened 2017-07-11
brew: expand description for `brew update` and add `outdated` example

vivekkrish:

PR #1495, opened 2017-09-16
brew: minor reword, don't use token syntax in desc

waldyrious:
This is a follow-up to #1427, since I wasn't able to join the comment there at the time.

----

### DIFF
--- a/pages/linux/brew.md
+++ b/pages/linux/brew.md
@@ -14,7 +14,7 @@
 
 `brew list`
 
-- Update an installed formula (if no formula name is given, updates all installed formulae):
+- Update an installed formula (if no formula name is given, all installed formulae are updated):
 
 `brew upgrade {{formula}}`
 
--- a/pages/linux/brew.md
+++ b/pages/linux/brew.md
@@ -14,7 +14,7 @@
 
 `brew list`
 
-- Update an installed formula (passing no `{{formula}}` updates all installed formulae):
+- Update an installed formula (if no formula name is given, updates all installed formulae):
 
 `brew upgrade {{formula}}`
 
@@ -22,7 +22,7 @@
 
 `brew update`
 
-- Show formulae that have an updated version available:
+- Show formulae that have a more recent version available:
 
 `brew outdated`
 
--- a/pages/osx/brew.md
+++ b/pages/osx/brew.md
@@ -14,7 +14,7 @@
 
 `brew list`
 
-- Update an installed formula (if no formula name is given, updates all installed formulae):
+- Update an installed formula (if no formula name is given, all installed formulae are updated):
 
 `brew upgrade {{formula}}`
 
--- a/pages/osx/brew.md
+++ b/pages/osx/brew.md
@@ -14,7 +14,7 @@
 
 `brew list`
 
-- Update an installed formula (passing no `{{formula}}` updates all installed formulae):
+- Update an installed formula (if no formula name is given, updates all installed formulae):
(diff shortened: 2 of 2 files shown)


vivekkrish:
Hi @waldyrious, looks good! Apologies for using tokens in the description.

Would definitely help if [CONTRIBUTING.md](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) was updated with this rule.

waldyrious:
@vivekkrish you're right, of course. I'm a little bogged down with other projects, to be honest -- would you like to submit that change to CONTRIBUTING.md?

agnivade:
I would suggest not adding any extra rules in CONTRIBUTING.md unless required. People already skip reading the document citing that it's too long.

This seems to be a corner case to me. IMO this is better left to be checked during review time.

waldyrious:
Indeed, that is a concern I have as well. But I don't like having implicit rules all over the place and making contributors feel they've wasted their time reading the documentation (when they do). Perhaps splitting that document into a main, concise overview, and subpages with additional details for corner cases would be a better solution.

agnivade:
I wouldn't go as far as to say that they have wasted their time. It's totally okay for the review process to bring a PR into shape (with corrections that are not mentioned in a CONTRIBUTING page).

And splitting into multiple pages doesn't really help because now a contributor has to read more rules, which is bad. And if he just reads the main section and sends a PR and we correct that by saying this rule was in the supplementary section, its just as bad. IMO, I don't think we would have gained anything.

Case in point - the infinitive tense rule. Its not mentioned, but still enforced.

Just my thoughts. Open to hear other opinions.

waldyrious:
I was thinking more along the lines of the main CONTRIBUTING.md file being the "official" list of guidelines we'd expect contributors to read, and the others more like appendices. In other words, the main page would be a manual-type document, while the others would be reference-type documents where one could look up specific information about how the project chooses to deal with a given edge case.

That sort of resource also saves time in the long run, for both contributors and maintainers, because all those small decisions that we make here and there (and decide aren't big enough a deal to justify making the guidelines longer) won't have to be re-taken, and the discussions won't have to be repeated among people who didn't participate in the previous ones, and so on.

So I definitely see the value in these subpages, as long as they are developed as auxiliary documents, not "required reading" :)

agnivade:
Hmm, you do have a point though. I would be okay with an "auxiliary" document like that. For an average contributor, only reading the CONTRIBUTING.md should be sufficient. 

The wording has to be carefully done to give that impression. I will leave you to it.

pages/linux/brew.md
```diff
@@ -14,15 +14,15 @@
 
 `brew list`
 
-- Update an installed formula (passing no `{{formula}}` updates all installed formulae):
+- Update an installed formula (if no formula name is given, updates all installed formulae):
```
sbrl: Perhaps it would read better if it said `all installed formulae are updated` instead of `updates all installed formulae`?

pages/osx/brew.md
```diff
@@ -14,15 +14,15 @@
 
 `brew list`
 
-- Update an installed formula (passing no `{{formula}}` updates all installed formulae):
+- Update an installed formula (if no formula name is given, updates all installed formulae):
```
sbrl: Same here.

pages/linux/brew.md
waldyrious: Agreed.

PR #1497, opened 2017-09-17
fdisk: add page

skymfc:
----


skymfc:
First comment

skymfc:
The first comment isn't easy cause the specify requriments

agnivade:
Hi @skymfc - We have a specific format for the pages because all the clients need to display it correctly.  The format is [here](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#markdown-format). 

And you can run the tests locally also by just doing an `npm test`. No need to push to github every time to check the format.

pages/linux/fdisk.md
```diff
@@ -0,0 +1,19 @@
+# fdisk
+
+> Manipulate disk partition table.
+
+- List  the  partition  tables for the sda disk and then exit:
```
sbrl: You've got some double spaces in here.

Also "and then exit"/ Does that mean that the other forms of this command don't exit?

pages/linux/fdisk.md
```diff
@@ -0,0 +1,19 @@
+# fdisk
+
+> Manipulate disk partition table.
```
sbrl: `table` -> `tables`?

pages/linux/fdisk.md
```diff
@@ -0,0 +1,19 @@
...
+
+- List  the  partition  tables for the sda disk and then exit:
+
+`fdisk -l /dev/sda`
+
+- Print the size (in blocks) of /dev/sda1:
+
+`fdisk -s /dev/sda1`
+
+- Create/remove/change partitions in the device:
```
sbrl: Ah, yeah - I see. it's this command that remains as a sort of interactive shell. In that case, perhaps we could alter this description to reflect that rather than saying "and then exit" above?

pages/linux/fdisk.md
```diff
@@ -0,0 +1,19 @@
...
+
+`fdisk /dev/sda`
+
+- Print help and then exit:
```
sbrl: Not sure that this is needed.

@agnivade / @waldyrious thoughts?

agnivade: Its fine. Some commands do have their -h/--help variants. So a user can see extra options with this.

PR #1499, opened 2017-09-19
port: add page

personalnadir:
Proposed page for mac ports port cli.

----

### DIFF
--- a/pages/osx/port.md
+++ b/pages/osx/port.md
@@ -1,0 +1,27 @@
+# port
+
+> macOS package manager
+
+- Search for package:
+
+`port search {{search_term}}`
+
+- Install package:
+
+`sudo port install {{package_name}}`
+
+- List installed packages:
+
+`port installed`
+
+- Update ports tree:
+
+`sudo port selfupdate`
+
+- Upgrade outdated packages:
+
+`sudo port upgrade outdated`
+
+- Remove inactive port versions:
+
+`sudo port uninstall inactive`
--- a/pages/osx/port.md
+++ b/pages/osx/port.md
@@ -22,6 +22,6 @@
 
 `sudo port upgrade outdated`
 
-- Remove inactive package versions:
+- Remove old versions of installed packages:
 
 `sudo port uninstall inactive`
--- a/pages/osx/port.md
+++ b/pages/osx/port.md
@@ -1,6 +1,6 @@
 # port
 
-> macOS package manager
+> Package manager for macOS.
 
 - Search for package:
 
--- a/pages/osx/port.md
+++ b/pages/osx/port.md
@@ -2,11 +2,11 @@
 
 > Package manager for macOS.
 
-- Search for package:
+- Search for a package:
 
 `port search {{search_term}}`
 
-- Install package:
+- Install a package:
 
 `sudo port install {{package_name}}`
 
@@ -14,7 +14,7 @@
 
 `port installed`
 
-- Update ports tree:
+- Update port and fetch latest list of available packages:
 
 `sudo port selfupdate`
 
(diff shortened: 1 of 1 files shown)


pages/osx/port.md
```diff
@@ -0,0 +1,27 @@
+# port
+
+> Package manager for macOS.
+
+- Search for package:
```
agnivade: "a" package

pages/osx/port.md
```diff
@@ -0,0 +1,27 @@
...
+
+`port search {{search_term}}`
+
+- Install package:
```
agnivade: "a" package

pages/osx/port.md
```diff
@@ -0,0 +1,27 @@
...
+
+`sudo port install {{package_name}}`
+
+- List installed packages:
+
+`port installed`
+
+- Update ports tree:
```
agnivade: Might be better to remove "tree". How about -
```
Update `port` to the latest version
```

personalnadir: That's a good point. I think it still needs to refer to the update to the tree in some form though as that's its most common use-case. How about something like:

`Update `port` and fetch latest list of available packages`

agnivade: Maybe I don't understand clearly, what exactly do you mean by "tree" here ?

personalnadir: The port tree is what Mac Ports refers to the list of packages as. From their documentation for this command: "The local ports tree is a collection of files that contain information on which packages are available through MacPorts and how they can be installed. You should regularly update your ports tree to get access to updated versions of software and bug fixes. To do that, use selfupdate"

`port selfupdate` also upgrades the port program as well if a new version is available. The output looks something like:

```` --->  Updating MacPorts base sources using rsync
MacPorts base version 2.4.1 installed,
MacPorts base version 2.4.1 downloaded.
--->  Updating the ports tree
--->  MacPorts base is already the latest version

The ports tree has been updated. To upgrade your installed ports, you should run
  port upgrade outdated

File: pages/osx/port.md [hunk 4]
```diff
@@ -0,0 +1,27 @@
+# port
+
+> Package manager for macOS.
```
sbrl: So `port` is macOS's primary package manager?

pages/osx/port.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Update port and fetch latest list of available packages:
+
+`sudo port selfupdate`
+
+- Upgrade outdated packages:
+
+`sudo port upgrade outdated`
+
+- Remove inactive package versions:
```
sbrl: What does `inactive` mean in this context?

pages/osx/port.md
personalnadir: No, there's also Homebrew which has its brew command already in the TLDR pages. I took that line directly from the brew page. I think port is the more venerable of the two.

pages/osx/port.md
personalnadir: port does not remove older versions of installed packages after an upgrade, so in case an update goes wrong you can revert to the already installed versions. These older versions are referred to as inactive by the command. 

Perhaps obsolete or superseded might be clearer? Or something like:

`- Remove old versions of installed packages `

pages/osx/port.md
sbrl: Ah, ok :smiley_cat:

pages/osx/port.md
sbrl: Ah, I see. Yeah, that's a much better way of putting it :smiley_cat:

pages/osx/port.md
personalnadir: Sorry, forgot I hadn't made that change, just added it now

sbrl:
Thanks, @personalnadir 😺

PR #1500, opened 2017-09-21
apt-get: example for upgrading a single package

waldyrious:
Can anyone confirm whether this is also true for `apt` (or whether there's a more appropriate syntax in that case)? We should update [apt.md](https://github.com/tldr-pages/tldr/blob/master/pages/linux/apt.md) in the same way, for consistency.

Due credit to [these excellent answers at AskUbuntu](https://askubuntu.com/questions/44122/how-to-upgrade-a-single-package-using-apt-get) :)

----

### DIFF
--- a/pages/linux/apt-get.md
+++ b/pages/linux/apt-get.md
@@ -10,13 +10,13 @@
 
 `apt-get install {{package}}`
 
-- Upgrade all installed packages to their newest available versions:
-
-`apt-get upgrade`
-
 - Remove a package:
 
 `apt-get remove {{package}}`
+
+- Upgrade all installed packages to their newest available versions:
+
+`apt-get upgrade`
 
 - Remove all packages that are no longer needed:
 
--- a/pages/linux/apt-get.md
+++ b/pages/linux/apt-get.md
@@ -14,9 +14,13 @@
 
 `apt-get remove {{package}}`
 
-- Upgrade installed packages to newest available versions:
+- Upgrade all installed packages to newest available versions:
 
 `apt-get upgrade`
+
+- Upgrade a single package to the newest available version:
+
+`apt-get install --only-upgrade {{package}}`
 
 - Remove no longer needed packages:
 
--- a/pages/linux/apt-get.md
+++ b/pages/linux/apt-get.md
@@ -6,7 +6,7 @@
 
 `apt-get update`
 
-- Install a new package, or update an existing package to its latest available version:
+- Install a package, or update it to the latest available version:
 
 `apt-get install {{package}}`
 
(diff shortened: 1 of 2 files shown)


waldyrious:
PR number 1500, woo!

pages/linux/apt-get.md
```diff
@@ -14,10 +14,14 @@
...
 `apt-get remove {{package}}`
 
-- Upgrade installed packages to newest available versions:
+- Upgrade all installed packages to newest available versions:
 
 `apt-get upgrade`
 
+- Upgrade a single package to the newest available version:
+
+`apt-get install --only-upgrade {{package}}`
```
agnivade: Umm .. you don't need to pass the `--only-upgrade` flag. By default, if the package is installed and you do `apt-get install`, it will upgrade it.

waldyrious: I believe that `--only-upgrade` is precisely to ensure that the package isn't installed if it isn't already present in the machine. Can you confirm? But I can agree that it may be a niche case. What do you guys think, @agorf, @fuerbringer, @lord63, @zdroid?

agnivade: > Can you confirm?

Yes, that's right. But almost always, you would want to actually _install_ the package if you don't have it. So, its very much a corner case, where you want to upgrade a package, but you don't want to accidentally install it if it doesn't exist.

agorf: True, trying to install a missing package with `--only-upgrade` skips the package installation in my system. But this sounds like an edge case to me: "do not install this package but only upgrade it if it's already installed". You'll probably already know if a package is installed or not. Whereas `apt-get install` always works like how you'd expect ("install this package if missing or upgrade it if present - either way I want it at the latest version").

waldyrious: Fair enough. Should we then, instead of introducing a new example, change the description of the existing `apt-get install {{package}}` example to "Install a new package or update it to the latest available version"?

I'm a bit wary of mixing concepts / intentions here, which is why I initially added this next to the "upgrade all" option, as it semantically makes more sense there.

agorf: +1 for a single example with the detailed description you mention.

agnivade: > Should we then, instead of introducing a new example, change the description of the existing apt-get install {{package}} example to "Install a new package or update it to the latest available version"?

By all means, yes. :)

waldyrious: Ok then. Can you confirm that this is true for both `apt` and `apt-get`?

agnivade: Yes.

waldyrious:
I made the agreed change, and also a bunch of other edits to improve both pages. The commits should be squashed on merge.

PR #1501, opened 2017-09-22
wget: reorder examples and add additional details

waldyrious:
- start with the simplest invocation.
- use example tokens rather than a single meta name, to make the specificities of each example clearer
- remove -O and -P, since the target files/folders can be easily renamed using standard tools;
  this leaves space for more tool-specific options.
- add a single-page + resources example, introducing --page-requisites before the full-blown crawl example.
- expand all option names in the full crawl example, for consistency.
- group all network-related control options at the end of the page.

----

### DIFF
--- a/pages/common/wget.md
+++ b/pages/common/wget.md
@@ -19,7 +19,7 @@
 
 `wget --ftp-user={{username}} --ftp-password={{password}} {{ftp://example.com}}`
 
-- Limit download speed:
+- Limit download speed to 200 kB/s:
 
 `wget --limit-rate={{200k}} {{https://example.com}}`
 
--- a/pages/common/wget.md
+++ b/pages/common/wget.md
@@ -3,11 +3,11 @@
 > Download files from the Web.
 > Supports HTTP, HTTPS, and FTP.
 
-- Download the contents of an URL to a file:
+- Download the contents of an URL to a file (named "foo" in this case):
 
-`wget {{https://example.com}}`
+`wget {{http://example.com/foo}}`
 
-- Download a single web page and all its associated resources:
+- Download a single web page and all its resources (scripts, stylesheets, images, etc.):
 
 `wget --page-requisites --convert-links {{https://example.com/somepage.html}}`
 
--- a/pages/common/wget.md
+++ b/pages/common/wget.md
@@ -5,7 +5,7 @@
 
 - Download the contents of an URL to a file (named "foo" in this case):
 
-`wget {{http://example.com/foo}}`
+`wget {{https://example.com/foo}}`
 
 - Download a single web page and all its resources (scripts, stylesheets, images, etc.):
 
--- a/pages/common/wget.md
+++ b/pages/common/wget.md
@@ -5,24 +5,28 @@
 
 - Download the contents of an URL to a file:
 
-`wget -O {{filename}} {{url}}`
+`wget {{https://example.com}}`
+
+- Download a single web page and all its associated resources:
+
+`wget --page-requisites --convert-links {{https://example.com/somepage.html}}`
+
+- Download a full website, with 3-second intervals between requests:
+
+`wget --mirror --page-requisites --convert-links --wait=3 {{https://example.com}}`
+
+- Download the contents of an URL via authenticated FTP:
+
+`wget --ftp-user={{username}} --ftp-password={{password}} {{ftp://example.com}}`
 
 - Limit download speed:
 
-`wget --limit-rate={{200k}} {{url}}`
+`wget --limit-rate={{200k}} {{https://example.com}}`
 
 - Continue an incomplete download:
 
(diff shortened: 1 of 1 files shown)


pages/common/wget.md
```diff
@@ -5,24 +5,28 @@
...
+`wget {{https://example.com}}`
 
-- Limit download speed:
+- Download a single web page and all its associated resources:
 
-`wget --limit-rate={{200k}} {{url}}`
+`wget --page-requisites --convert-links {{https://example.com/somepage.html}}`
 
-- Continue an incomplete download:
+- Download a full website, with 3-second intervals between requests:
```
sbrl: `with a 3 second delay`?

waldyrious: Just that? It may not be sufficiently clear that the delay is between each request rather than, say, before starting the entire process. That's an unlikely use case, sure but I'd prefer avoiding the ambiguity if we can.

sbrl: Ah, true. Let's leave it as is then :smiley_cat: 

I meant to send this as a general comment, not a change request, oops :P

pages/common/wget.md
```diff
@@ -5,24 +5,28 @@
...
 - Download the contents of an URL to a file:
 
-`wget -O {{filename}} {{url}}`
+`wget {{https://example.com}}`
 
-- Limit download speed:
+- Download a single web page and all its associated resources:
 
-`wget --limit-rate={{200k}} {{url}}`
+`wget --page-requisites --convert-links {{https://example.com/somepage.html}}`
```
agnivade: What does `--convert-links` do ?

pages/common/wget.md
agnivade: The original example had a "target_folder" too. Are you sure we don't need it anymore ?

pages/common/wget.md
```diff
@@ -5,24 +5,28 @@
...
+`wget --mirror --page-requisites --convert-links --wait=3 {{https://example.com}}`
 
-`wget -c {{url}}`
+- Download the contents of an URL via authenticated FTP:
 
-- Download a full website:
+`wget --ftp-user={{username}} --ftp-password={{password}} {{ftp://example.com}}`
 
-`wget --mirror -p --convert-links -P {{target_folder}} {{url}}`
+- Limit download speed:
```
agnivade: Maybe better to mention the speed too ?
`Limit download speed to 200kBps`.

Because just 200k might give the impression of size, not speed.

pages/common/wget.md
waldyrious: It rewrites links present in the content of the page (in this case links to the scripts, styles, images, etc.) to point to the local versions.

pages/common/wget.md
waldyrious: > The original example had a "target_folder" too. Are you sure we don't need it anymore ?

(replying in the correct location, for clarity)

We don't *strictly* need it, as by default the url is used as the folder name. Of course, if we decide that a custom name is an important part of this example, then the `-P` is needed in that sense. However, as I mentioned in the commit message, I deliberately chose to "remove -O and -P, since the target files/folders can be easily renamed using standard tools; this leaves space for more tool-specific options."

pages/common/wget.md
waldyrious: I've replied below in the thread corresponding to the removed line.

pages/common/wget.md
agnivade: Ok

pages/common/wget.md
agnivade: Alright, so then lets mention that the site will be downloaded to a folder with the same name as url, for extra clarity.

waldyrious: Wouldn't that make it too long, though? The description currently says:

> Download a full website, with 3-second intervals between requests:

agnivade: Right, I see that you have removed the `-O` option from the first example. How about we mention this info in the first example, and keep this as is ? So this does not increase in length and the user can deduce that the same file output rules apply.

waldyrious: Good point. I'm not sure how to say it compactly, though, because the default filename will either be the last part of the URL (after the last slash), or default to `index.html` if pointing directly to the root domain of a website (that is, assuming I'm not missing any edge cases).

agnivade: We can add a url path to the root domain like this - 'http://example.com/foo' and say that the downloaded file name will be foo. The users can deduce that for the root domain it will be `index.html`.

waldyrious: Ok 👍

waldyrious:
@agnivade how does it look now?

agnivade:
Looks good. Any reason you changed https to http ?

waldyrious:
> Any reason you changed https to http?

Copy-paste error :P I'll change back.

PR #1502, opened 2017-09-23
ffmpeg: simplify convert to audio example

slicer69:
The ffmpeg command detects the output file type and automatically selects the proper output format. We do not need to specify format or sample rate. This greatly simplifies the extraction of audio from video files.


agnivade:
Thank you. Please sign the CLA so that we can get started on the review process.

sbrl:
That is a huge simplification. Thanks 😺

PR #1513, opened 2017-10-04
awk: add examples to filter lines

lindenb:
two examples added for awk.

----

### DIFF
--- a/pages/common/awk.md
+++ b/pages/common/awk.md
@@ -22,10 +22,10 @@
 
 `awk '{s+=$1; print $1} END {print "--------"; print s}' {{filename}}`
 
-- Keep the lines where the second column is "Earth"
+- Keep the lines where the second column is "Earth":
 
 `awk '($2 == "Earth")' {{filename}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/awk.md
```diff
@@ -21,3 +21,11 @@
...
 
 `awk '{s+=$1; print $1} END {print "--------"; print s}' {{filename}}`
+
+- Keep the lines where the second column is "Earth":
+
+`awk '($2 == "Earth")' {{filename}}`
+
+- Keep the lines where the second column is greater than 10  or lower than 0 :
+
+`awk '($2 > 10 || $2 < 0)' {{filename}}`
```
agnivade: This is a counter-intuitive example. I recommend changing it to `$2>0 && $2<10` so that the second column is within the range of 0-10.

I think that is a more common use-case than the complement of that.

ericbn: The parenthesis are unnecessary in both new examples. And not sure if "Keep the lines..." is as clear as "Print lines..." (also because "print" is used in other examples)

PR #1516, opened 2017-10-04
bedops: add page

alexpreynolds:
----

### DIFF
--- a/pages/common/bedops.md
+++ b/pages/common/bedops.md
@@ -5,8 +5,6 @@
 
 - Take the multiset union of two or more BED or Starch files and write the sorted result to "output_file":
 
-`bedops --everything {{path/to/file_1}} {{path/to/file_2}} > {{path/to/output_file}}`
-
 `bedops --everything {{path/to/file_1}} {{path/to/file_2}} [{{path/to/file_3}} ...] > {{path/to/output_file}}`
 
 - Find elements from reference set "reference_file" that overlap one or more query sets by one or more bases:
--- a/pages/common/bedops.md
+++ b/pages/common/bedops.md
@@ -3,22 +3,24 @@
 > BEDOPS: A fast, highly scalable, and easily-parallelizable genome analysis toolkit.
 > Apply efficient set operations on any number of sorted BED or Starch files such as element-of, intersection, merge, difference, partition, etc. and their logical inverses.
 
-- Take the multiset union of two or more BED or Starch files and write the sorted result to {{path/to/output_file}}:
+- Take the multiset union of two or more BED or Starch files and write the sorted result to "output_file":
+
+`bedops --everything {{path/to/file_1}} {{path/to/file_2}} > {{path/to/output_file}}`
 
 `bedops --everything {{path/to/file_1}} {{path/to/file_2}} [{{path/to/file_3}} ...] > {{path/to/output_file}}`
 
-- Find elements from reference set {{path/to/reference_file}} that overlap one or more query sets by one or more bases:
+- Find elements from reference set "reference_file" that overlap one or more query sets by one or more bases:
 
 `bedops --element-of 1 {{path/to/reference_file}} {{path/to/query_file_1}} [{{path/to/query_file_2}} ...] > {{path/to/output_file}}`
 
-- Merge the overlapping intervals in one or more files and write the sorted result to {{path/to/output_file}}:
+- Merge the overlapping intervals in one or more files and write the sorted result to "output_file":
 
 `bedops --merge {{path/to/file_1}} [{{path/to/file_2}} {{path/to/file_3}} ...] > {{path/to/output_file}}`
 
-- Partition overlapping intervals in one or more files into disjoint intervals and write the sorted result to {{path/to/output_file}}:
+- Partition overlapping intervals in one or more files into disjoint intervals and write the sorted result to "output_file":
 
 `bedops --partition {{path/to/file_1}} [{{path/to/file_2}} {{path/to/file_3}} ...] > {{path/to/output_file}}`
 
-- For all operations, efficiently apply set operations on specified {{chromosome}} and ignore other chromosomes:
+- For all operations, efficiently apply set operations on specified chromosome and ignore other chromosomes:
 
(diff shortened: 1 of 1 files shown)


pages/common/bedops.md
```diff
@@ -0,0 +1,24 @@
+# bedops
+
+> BEDOPS: A fast, highly scalable, and easily-parallelizable genome analysis toolkit.
```
agnivade: The user already knows the name of the command. So I think we can remove "BEDOPS:".

pages/common/bedops.md
```diff
@@ -0,0 +1,24 @@
...
+> Apply efficient set operations on any number of sorted BED or Starch files such as element-of, intersection, merge, difference, partition, etc. and their logical inverses.
+
+- Take the multiset union of two or more BED or Starch files and write the sorted result to {{path/to/output_file}}:
```
agnivade: We can just mention "output_file" like this. The full path is more suited in the actual example.

pages/common/bedops.md
```diff
@@ -0,0 +1,24 @@
...
+
+`bedops --merge {{path/to/file_1}} [{{path/to/file_2}} {{path/to/file_3}} ...] > {{path/to/output_file}}`
+
+- Partition overlapping intervals in one or more files into disjoint intervals and write the sorted result to {{path/to/output_file}}:
+
+`bedops --partition {{path/to/file_1}} [{{path/to/file_2}} {{path/to/file_3}} ...] > {{path/to/output_file}}`
+
+- For all operations, efficiently apply set operations on specified {{chromosome}} and ignore other chromosomes:
+
+`bedops --chrom {{chromosome}} --operation {{files...}} > {{path/to/output_file}}`
```
agnivade: Is `--chrom` correct ? Seems like a typo.

pages/common/bedops.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Merge the overlapping intervals in one or more files and write the sorted result to {{path/to/output_file}}:
+
+`bedops --merge {{path/to/file_1}} [{{path/to/file_2}} {{path/to/file_3}} ...] > {{path/to/output_file}}`
+
+- Partition overlapping intervals in one or more files into disjoint intervals and write the sorted result to {{path/to/output_file}}:
+
+`bedops --partition {{path/to/file_1}} [{{path/to/file_2}} {{path/to/file_3}} ...] > {{path/to/output_file}}`
+
+- For all operations, efficiently apply set operations on specified {{chromosome}} and ignore other chromosomes:
```
agnivade: I think the {{}} can be removed from the descriptions. They are generally used in the examples. I suggest ".. efficiently apply set operations on a specified chromosome and ignore .. "

pages/common/bedops.md
```diff
@@ -0,0 +1,24 @@
+# bedops
+
+> BEDOPS: A fast, highly scalable, and easily-parallelizable genome analysis toolkit.
+> Apply efficient set operations on any number of sorted BED or Starch files such as element-of, intersection, merge, difference, partition, etc. and their logical inverses.
+
+- Take the multiset union of two or more BED or Starch files and write the sorted result to {{path/to/output_file}}:
+
+`bedops --everything {{path/to/file_1}} {{path/to/file_2}} [{{path/to/file_3}} ...] > {{path/to/output_file}}`
```
agnivade: I get what you are trying to do with the [ ...] notation. In this project, we lean more towards giving concrete examples and letting the user build up complicated examples from them. From what I see, all commands take multiple input files. So, what we can do is - keep this example to have just 2 files, and then have the next example to have 2 query files.

In this way, the user can figure out that they can add more files on command line. And we also avoid the [ ...] notation, which takes a second to figure out.

What do you think ?

pages/common/bedops.md
alexpreynolds: BEDOPS is the name of the toolkit. The `bedops` binary is one tool in this kit. There are other binaries that are part of BEDOPS.

pages/common/bedops.md
alexpreynolds: That's fine.

pages/common/bedops.md
alexpreynolds: I think having some way to denote that three or more files are also possible is useful. Perhaps we could have two examples, one with two files, and another with four or five files, to make it clear that multiple files are allowed.

pages/common/bedops.md
alexpreynolds: I think I just reviewed how other binaries were explained in their own Markdown files, and I saw this convention used there. I'm happy to do things in whichever way you would prefer.

pages/common/bedops.md
alexpreynolds: No, `--chrom` is correct. This is how the user specifies that she is working with a specific chromosome in the input files, and no other chromosomes.

pages/common/bedops.md
agnivade: If this convention is used in other files, that is incorrect and needs to be removed. Sometimes, these things skip reviews and get committed to the repo.

pages/common/bedops.md
agnivade: Sure, I would think three files would be sufficient. We wouldn't want to make the example too long. But up to you.

pages/common/bedops.md
agnivade: Ah I see. Ok we can keep it in that case.

alexpreynolds:
I pushed a commit to this branch that reflects the changes you requested in your comments. Please let me know if there are any remaining issues.

agnivade:
We need to remove the `[ .. ]` convention used in the examples. And just use {{/path/to/file1}} {{/path/to/file2}}

sbrl:
Ping @alexpreynolds

alexpreynolds:
Is there a summary of what changes are needed? I'm trying to use existing tldr documents as a template and I'm not sure what's left to change. Thanks!

agnivade:
Please look at my [comment](https://github.com/tldr-pages/tldr/pull/1516#issuecomment-340139092) right above :point_up:

PR #1519, opened 2017-10-05
ufw: Added an example of rule comments

kalebo:
One of the more obscure features of UFW is being able to add comments to your rules and was only added last year. This feature helps significantly when managing a complicated rule set. 

I tried to shim it in to an existing example without introducing too much complexity as the entry already has eight examples.

----

### DIFF
--- a/pages/linux/ufw.md
+++ b/pages/linux/ufw.md
@@ -15,7 +15,7 @@
 
 `ufw status numbered`
 
-- Allow incoming traffic on port 5432 on this host and an identifing comment:
+- Allow incoming traffic on port 5432 on this host with a comment identifying the service:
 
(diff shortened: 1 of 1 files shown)


pages/linux/ufw.md
```diff
@@ -15,9 +15,9 @@
 
 `ufw status numbered`
 
-- Allow incoming traffic on port 5432 on this host:
+- Allow incoming traffic on port 5432 on this host and an identifing comment:
```
agnivade: typo => identifing

And I recommend using "with an" instead of "and an" to signify that this comment is paired with that rule.

sbrl:
Looks good to me :smiley_cat:

PR #1520, opened 2017-10-05
find: improve -mtime example

ericbn:
Make the description clearer and more specific to the given `-1` parameter. Remove unnecessary `-name '{{}}'` from example.


ericbn:
I know this file has lots of examples. Just trying to improve an existing one.

PR #1521, opened 2017-10-06
apt-get: add purge

Darexon:
----

Added an `apt-get purge` description to the apt-get page.

### DIFF
--- a/pages/linux/apt-get.md
+++ b/pages/linux/apt-get.md
@@ -14,6 +14,10 @@
 
 `apt-get remove {{package}}`
 
+- Remove a package (like `remove`), but also remove any associated configuration files:
+
+`apt-get purge {{package}}`
+
 - Upgrade all installed packages to their newest available versions:
 
 `apt-get upgrade`


agnivade:
Thanks for the PR.

I am however reluctant to merge this. The `--purge` is an advanced option which is not needed by a beginner. Please keep in mind that tldr pages are not a replacement for man pages. We intentionally have a limit of 8 examples max so that we can keep the page short and contain the most useful examples. It's tldr for a reason. :wink: 

In this case, we are 7 which is cutting it pretty close without any strong reason to do so. Hence, my reluctance.

Darexon:
Purge is included in the Aptitude page which is why I feel it should be included in the apt-get page as well.

agnivade:
I would want it removed from aptitude page also then. We try hard to keep all pages consistent, but with so many commands and its variants across platforms, its not possible to do that always.

Overall, the central theme should be to keep it limited and tight. And mention only practical examples.

agnivade:
Appreciate your understanding on this @Darexon. I hope this does not deter you from sending more PRs in the future. Feel free to send a PR to update the aptitude page.

PR #1524, opened 2017-10-06
r: add page

YangVincent:
----

### DIFF
--- a/pages/common/r.md
+++ b/pages/common/r.md
@@ -1,0 +1,15 @@
+# R
+
+> R language interpreter
+
+- Call a R interactive shell (REPL):
+
+`R`
+
+- Check R version:
+
+`R --version`
+
+- Run file as input
+
+`R -f file.R`
--- a/pages/common/r.md
+++ b/pages/common/r.md
@@ -1,6 +1,6 @@
 # R
 
-> R language interpreter
+> R language interpreter.
 
 - Call a R interactive shell (REPL):
 
@@ -10,6 +10,6 @@
 
 `R --version`
 
-- Run file as input
+- Run file as input:
 
 `R -f file.R`
--- a/pages/common/r.md
+++ b/pages/common/r.md
@@ -2,7 +2,7 @@
 
 > R language interpreter.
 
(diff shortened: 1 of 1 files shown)


pages/common/r.md
```diff
@@ -0,0 +1,15 @@
+# R
```
sbrl: It really is a single letter command that's uppercase? That's rather unusual.

pages/common/r.md
```diff
@@ -0,0 +1,15 @@
...
+
+> R language interpreter.
+
+- Call a R interactive shell (REPL):
```
sbrl: I'd prefer `an` instead of `a` in this instance, as `R` is a single character.

pages/common/r.md
YangVincent: Yup! Though I've seen many versions where it calls for the lower case, though that may just be aliased.

pages/common/r.md
YangVincent: Fixed!

pages/common/r.md
agnivade: I would really hate to break convention here. All our pages have the heading as lowercase. Unless there is a different command with a lower case `r`, I would recommend changing this to lowercase.

pages/common/r.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Call an R interactive shell (REPL):
+
+`R`
+
+- Check R version:
+
+`R --version`
+
+- Run file as input:
```
agnivade: Change this to  - "Execute a file" to keep it consistent with pages of other programming languages.

pages/common/r.md
```diff
@@ -0,0 +1,15 @@
...
+
+`R -f file.R`
```
agnivade: Wrap file.R in tokens

pages/common/r.md
```diff
@@ -0,0 +1,15 @@
+# R
+
+> R language interpreter.
+
+- Call an R interactive shell (REPL):
```
agnivade: Call => Start

PR #1533, opened 2017-10-08
pup: add page

deekim:
https://github.com/EricChiang/pup

----


deekim:
Per the contributing guidelines,
> If a command expects the file to have a particular extension, use it. For example, unrar x {{compressed.rar}}. In case a generic extension is needed, use {{.ext}}, but only if an extension is required. For instance, in find.md's example "Find files by extension" (find {{root_path}} -name '{{*.ext}}') using {{*.ext}} explains the command without being unnecessarily specific; But in a command like wc -l {{file}}, using {{file}} (without extension) is sufficient.

I was not sure it was okay to use `{{index.html}}` as part of the example, as the file extension is not required.

pages/common/pup.md
```diff
@@ -0,0 +1,27 @@
+# pup
+
+> HTML parsing tool.
+
+- Transform raw HTML file into a cleaned, indented, and colored format:
```
sbrl: `Transform raw HTML` -> `Transform a raw HTML`?

pages/common/pup.md
```diff
@@ -0,0 +1,27 @@
...
+
+`cat {{index.html}} | pup --color`
+
+- Filter HTML by element tag name:
```
sbrl: With the way this and further examples on this page are worded, it feels like the HTML it takes in isn't 'raw'? Perhaps the first example needs rewording - I'm not sure.

agnivade: Why not ? By raw HTML, I think its meant that whitespace is removed and the tags are not indented properly. The wording looks okay to me.

pages/common/pup.md
```diff
@@ -0,0 +1,27 @@
...
+
+`cat {{index.html}} | pup '{{tag}}'`
+
+- Filter HTML by id:
+
+`cat {{index.html}} | pup '{{tag#id}}'`
```
agnivade: Change the tag to `div` here to give a practical example.

pages/common/pup.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Filter HTML by attribute value:
+
+`cat {{index.html}} | pup '{{tag[attribute="value"}}'`
```
agnivade: Change the filter value to `input[type="text"]` to give another concrete example.

pages/common/pup.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Print all text from selected nodes and children:
```
agnivade: How about "Print all text from the filtered HTML elements and their children"

pages/common/pup.md
deekim: maybe `unformatted` would be clearer?

pages/common/pup.md
```diff
@@ -0,0 +1,27 @@
...
+
+> Command line HTML parsing tool.
+
+- Transform a raw HTML file into a cleaned, indented, and colored format:
+
+`cat {{index.html}} | pup --color`
+
+- Filter HTML by element tag name:
+
+`cat {{index.html}} | pup '{{div}}'`
```
agnivade: We can keep this as {{tag}} to match with "element tag name". Since we are giving a lot of concrete examples later anyway.

pages/common/pup.md
sbrl: Ah, ok. Let's leave it then :-)

PR #1517, opened 2017-10-05
popd: add page

juan88:
----

- [x] The page (if new), does not already exist in the repo.

- [x] The page (if new), has been added to the correct platform folder:  
      `common/` if it's common to all platforms, `linux/` if it's Linux-specific, and so on.

- [x] The page has 8 or fewer examples.

- [x] The PR is appropriately titled:  
      `<command name>: add page` for new pages, or `<command name>: <description of changes>` for pages being edited

- [x] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines

PR #1535, opened 2017-10-09
pushd: added platform specific examples in windows, osx and linux. Removed from common.

juan88:
----

As mentioned in #1517 the windows' cmd specific options of this command do not go along the examples provided in the common/pushd.md version. So, the page is created for each platform with the according examples.


pages/windows/pushd.md
```diff
@@ -0,0 +1,7 @@
+# pushd
+
+> Place a directory on a stack so it can be accessed later.
```
agnivade: Add another line as subheading like - 
"Also look at `popd` to switch back to original directory"

sbrl: Or even `See also` instead of `Also look`

sbrl:
Perhaps add the extra `see also` line to the osx version of this page too?

sbrl:
Thanks, @juan88 😺

PR #1536, opened 2017-10-09
ldconfig: add page

agnivade:
----

### DIFF
--- a/pages/linux/ldconfig.md
+++ b/pages/linux/ldconfig.md
@@ -2,11 +2,11 @@
 
 > Configure symlinks and cache for shared library dependencies.
 
-- Update symlinks and rebuild the cache. (Usually run when a new library is installed):
+- Update symlinks and rebuild the cache (usually run when a new library is installed):
 
 `sudo ldconfig`
 
-- Update symlinks for a given directory:
+- Update the symlinks for a given directory:
 
 `sudo ldconfig -n {{path/to/directory}}`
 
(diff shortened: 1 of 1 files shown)


pages/linux/ldconfig.md
```diff
@@ -0,0 +1,15 @@
+# ldconfig
+
+> Configure symlinks and cache for shared library dependencies.
+
+- Update symlinks and rebuild the cache. (Usually run when a new library is installed):
```
sbrl: Don't think the full stop is needed here. Also, I'm not sure that `Usually` needs a capital letter at the beginning.

pages/linux/ldconfig.md
```diff
@@ -0,0 +1,15 @@
...
+
+`sudo ldconfig`
+
+- Update symlinks for a given directory:
```
sbrl: Perhaps `Update symlinks` -> `Update the symlinks` reads better?

pages/linux/ldconfig.md
```diff
@@ -0,0 +1,15 @@
...
+
+`sudo ldconfig -n {{path/to/directory}}`
+
+- Print the libraries in the cache and check whether a given library is present:
```
sbrl: Perhaps just including the `ldconfig -p` bit would be ok? Not sure.

agnivade: You would almost always never use `ldconfig -p` without wanting to check for a particular library. This is how the -p flag is commonly used and in the interest of putting useful and practical examples, I have added the grep line.

pages/linux/ldconfig.md
agnivade: done

pages/linux/ldconfig.md
sbrl: Ahh ok. Looks good to me then!

PR #1539, opened 2017-10-10
diffstat: add page

harrymt:
----

### DIFF
--- a/pages/common/diffstat.md
+++ b/pages/common/diffstat.md
@@ -1,0 +1,11 @@
+# diffstat
+
+> Make histogram of diff-output
+
+- Display changes in a histogram:
+
+`diff {{file1}} {{file2}} | diffstat`
+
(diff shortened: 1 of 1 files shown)


pages/common/diffstat.md
```diff
@@ -0,0 +1,11 @@
+# diffstat
+
+> Make histogram of diff-output.
```
sbrl: Perhaps `Create a histogram from the output of the diff command` (with backticks around "diff") would read better?

pages/common/diffstat.md
```diff
@@ -0,0 +1,11 @@
...
+
+- Display changes in a histogram:
+
+`diff {{file1}} {{file2}} | diffstat`
+
+- Display inserted, deleted, modified changes in table format:
```
sbrl: I think you need an `and` after the last comma in that list.

Perhaps `as a table` instead of `in table format` would read better?

PR #1540, opened 2017-10-10
last: add example that avoids hostname truncation

kalebo:
Last by default puts its hostname information in a fixed width column. When you are trying to track down a long host name it is really inconvenient to have it truncated. The `-a` flag used in the  example resolves this issue. I have also included an example of using the `-i` flag to just get the ip address associated with the login

The last two examples were worded in such a way as to imply that the most recent entry would be returned, however the complete list of logins would be returned. I have changed the wording to reflect this reality.

----

### DIFF
--- a/pages/common/last.md
+++ b/pages/common/last.md
@@ -18,10 +18,10 @@
 
 `last {{user_name}} -i`
 
-- View all recorded reboots (last logins of the pseudo user reboot):
+- View all recorded reboots (i.e., the last logins of the pseudo user "reboot"):
 
 `last reboot`
 
-- View all recorded shutdowns (last logins of the pseudo user shutdown):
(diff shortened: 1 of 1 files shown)


pages/common/last.md
```diff
@@ -10,18 +10,18 @@
...
+`last -F -a`
 
-- View the last login by a specific user:
+- View all logins by a specific user and show the ip address instead of the hostname:
 
-`last {{user_name}}`
+`last {{user_name}} -i`
 
-- View the last reboot (last login of the pseudo user reboot):
+- View all recorded reboots (last logins of the pseudo user reboot):
```
agnivade: Maybe we should put the user "reboot" in quotes. Just to clarify that reboot is an user and not the term "user reboot"

PR #1544, opened 2017-10-13
update-alternatives: add page

jsonbruce:
----

### DIFF
--- a/pages/linux/update-alternatives.md
+++ b/pages/linux/update-alternatives.md
@@ -4,20 +4,16 @@
 
 - Add a symbolic link:
 
-`sudo update-alternatives --install {{link}} {{name}} {{path}} {{priority}}`
+`sudo update-alternatives --install {{path/to/symlink}} {{command_name}} {{path/to/command_binary}} {{priority}}`
 
-- Add a symbolic link for "java" with a priority of "300":
+- Configure a symbolic link for "java":
 
-`sudo update-alternatives --install /usr/bin/java java /opt/java/jdk1.8.0_102/bin/java 300`
-
-- Configure a symbolic link:
-
-`sudo update-alternatives --config {{name}}`
+`sudo update-alternatives --config {{java}}`
 
 - Remove a symbolic link:
 
-`sudo update-alternatives --remove {{name}} {{path}}`
+`sudo update-alternatives --remove {{java}} {{/opt/java/jdk1.8.0_102/bin/java}}`
 
 - Display information about a specified command:
 
-`update-alternatives --display {{name}}`
+`update-alternatives --display {{java}}`
--- a/pages/linux/update-alternatives.md
+++ b/pages/linux/update-alternatives.md
@@ -1,0 +1,19 @@
+# update-alternatives
+
+> A convenient tool for maintaining symbolic links determining default commands.
+
+- Add a symbolic link:
+
+`sudo update-alternatives --install /usr/bin/java java /opt/java/jdk1.8.0_102/bin/java 300`
+
+- Config a symbolic link:
+
+`sudo update-alternatives --config java`
+
+- Remove a symbolic link:
+
+`sudo update-alternatives --remove java /usr/bin/java`
+
+- Display information about the link group:
+
+`update-alternatives --display java`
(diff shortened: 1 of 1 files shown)


pages/linux/update-alternatives.md
```diff
@@ -0,0 +1,19 @@
+# update-alternatives
+
+> A convenient tool for maintaining symbolic links determining default commands.
```
sbrl: Perhaps `determining` should be `that determine`?

pages/linux/update-alternatives.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Add a symbolic link:
+
+`sudo update-alternatives --install /usr/bin/java java /opt/java/jdk1.8.0_102/bin/java 300`
+
+- Config a symbolic link:
```
sbrl: Perhaps expanding `Config` to `Configure` would read better?

pages/linux/update-alternatives.md
```diff
@@ -0,0 +1,19 @@
...
+
+`sudo update-alternatives --config java`
+
+- Remove a symbolic link:
+
+`sudo update-alternatives --remove java /usr/bin/java`
+
+- Display information about the link group:
```
sbrl: Perhaps `about the link group` would read better as `about the specified link group`?

pages/linux/update-alternatives.md
```diff
@@ -0,0 +1,19 @@
+# update-alternatives
+
+> A convenient tool for maintaining symbolic links that determining default commands.
```
agnivade: Umm the grammar got messed up a bit. I think @sbrl was trying to say "that determines". I actually prefer "to determine".

What do you say @sbrl ?

pages/linux/update-alternatives.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Add a symbolic link:
```
agnivade: Since all your examples are java based, we should mention this at the start. Then it will be implied. So I suggest - `Add a symbolic link for "java"`

pages/linux/update-alternatives.md
```diff
@@ -0,0 +1,19 @@
...
+
+`sudo update-alternatives --install /usr/bin/java java /opt/java/jdk1.8.0_102/bin/java 300`
```
agnivade: Please wrap the parameters in tokens. Also, what does 300 stand for ?

pages/linux/update-alternatives.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Configure a symbolic link:
+
+`sudo update-alternatives --config java`
+
+- Remove a symbolic link:
+
+`sudo update-alternatives --remove java /usr/bin/java`
+
+- Display information about the specified link group:
```
agnivade: What is meant by a link group here ?

pages/linux/update-alternatives.md
sbrl: Hrm, it has a bit! Yeah, it's a rather awkward situation here actually. I'd be happy with either "that determine" or "to determine".

pages/linux/update-alternatives.md
```diff
@@ -1,19 +1,19 @@
 # update-alternatives
 
-> A convenient tool for maintaining symbolic links that determining default commands.
+> A convenient tool for maintaining symbolic links to determin default commands.
```
agnivade: => determine

pages/linux/update-alternatives.md
agnivade: I didnt want you to remove all concrete examples with their generic counterparts :smile:. Those are fine. Just change the description to `Add a symbolic link for "java" with a priority of "300"`

jsonbruce: Yes, but I think a more generic expression maybe better. Or can we add a generic expression and a concrete examples

agnivade: We try to put concrete examples instead of generic ones so that beginners can have a better idea as to how the command is used. If you want, we can start with a generic example and then have concrete ones after that.

agnivade:
@jsonbruce I made some slight modifications of my own which I believe clarifies some things better. Please let me know your thoughts on it.

sbrl:
Thanks for sticking with the review process, @jsonbruce! :smiley_cat:

PR #1549, opened 2017-10-15
rainbowstream: add page

jitakirin:
Ref #1162

----

### DIFF
--- a/pages/common/rainbowstream.md
+++ b/pages/common/rainbowstream.md
@@ -3,26 +3,26 @@
 > Terminal-based Twitter Client. Realtime tweetstream, compose, search , favorite … and much more.
 > Online help with `h`, up and down arrow for history, tab auto-complete, 2-tab suggestion.
 
-- show your timeline
+- Show your timeline:
 
 `home [{{num_of_last_tweets}}]`
 
-- show profile of given user
+- Show profile of given user:
 
 `whois @{{user}}`
 
-- tweet the message as-is
+- Tweet the message as-is:
 
 `t {{message}}`
 
-- retweet the tweet with given id (id beside the time)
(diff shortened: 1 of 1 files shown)


agnivade:
Thanks for the contribution @jitakirin. The language and explanations need to be improved a bit. 

I haven't looked in-depth. But here are a couple of points which you might want to look at - 
1. The main heading seems to have been copied verbatim from the documentation page. Perhaps we can summarize that in a more clear fashion. We can also get rid of the ellipsis (...). It's an informal language use which we tend to avoid.
2. Please start the page with the normal command invocation `rainbowstream` and explain what it does.
3. Then move to explaining how to get into the interactive mode and the different commands in that mode.

Once these are taken care of, we can move to a more in-depth review.

jitakirin:
Hi @agnivade, I based this on the page for [rtv](https://github.com/tldr-pages/tldr/blob/master/pages/common/rtv.md), you wouldn't have a better example (especially for an interactive app) so I can get a better idea of preferred language and style?

I'll get the points addressed anyway, thanks for the review 😃

agnivade:
Yes, rtv seems to be as close to this I can find.

sbrl:
The `alpine` and `ansible` pages might be worth a little look if you want more of an idea of the kind of style we go for :-)

agnivade:
Yes, `alpine` works.

jitakirin:
I tried to address all the points, please let me know if there are further improvements I can make.

agnivade:
Sorry for the delay on this @jitakirin. I made some small changes. Its ready now.

sbrl:
Thanks, @jitakirin! :smiley_cat: 

Just 2 more commands left in #1162 to go!

PR #1561, opened 2017-10-20
perf: add page

mfrw:
----

### DIFF
--- a/pages/linux/perf.md
+++ b/pages/linux/perf.md
@@ -2,9 +2,9 @@
 
 > Framework for linux performance counter measurements.
 
-- Display basic performance counter stats for a command `gcc -c -O3 hello.c`:
+- Display basic performance counter stats for a command:
 
-`perf stat {{gcc -c -O3 hello.c}}`
+`perf stat {{gcc hello.c}}`
 
 - Display system-wide real time performance counter profile:
 
--- a/pages/linux/perf.md
+++ b/pages/linux/perf.md
@@ -2,18 +2,18 @@
 
 > Framework for linux performance counter measurements.
 
-- Gather basic performance counter stats for a command:
+- Display basic performance counter stats for a command `gcc -c -O3 hello.c`:
 
-`perf stat {{command}}`
+`perf stat {{gcc -c -O3 hello.c}}`
 
 - Display system-wide real time performance counter profile:
 
 `sudo perf top`
 
-- Run a command and record its profile into perf.data:
+- Run a command and record its profile into "perf.data":
 
 `sudo perf record {{command}}`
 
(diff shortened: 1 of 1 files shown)


pages/linux/perf.md
```diff
@@ -0,0 +1,19 @@
+# perf
+
+> Framework for linux performance counter measurements.
+
+- Gather basic performance counter stats for a command:
```
agnivade: It also displays the stats in the console right ? So maybe Display is a better word than Gather. Since Gather does not specify whether it will be showed or not.

pages/linux/perf.md
```diff
@@ -0,0 +1,19 @@
...
+
+`perf stat {{command}}`
```
agnivade: I think it might be clearer to mention an actual command so that users can get a better idea. Also add a `-` after stat so that extra options can be applied to a command.

pages/linux/perf.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Display system-wide real time performance counter profile:
+
+`sudo perf top`
+
+- Run a command and record its profile into perf.data:
```
agnivade: perf.data within quotes("")

pages/linux/perf.md
```diff
@@ -0,0 +1,19 @@
...
+
+`sudo perf record {{command}}`
+
+- Read perf.data (created by perf record) and display the profile:
```
agnivade: perf.data within quotes("") and perf record within backticks.

pages/linux/perf.md
mfrw: Sure..

pages/linux/perf.md
```diff
@@ -0,0 +1,19 @@
+# perf
+
+> Framework for linux performance counter measurements.
+
+- Display basic performance counter stats for a command `gcc -c -O3 hello.c`:
```
agnivade: No need to mention the gcc command in the description.

pages/linux/perf.md
```diff
@@ -0,0 +1,19 @@
...
+
+`perf stat {{gcc -c -O3 hello.c}}`
```
agnivade: Please add a - after stat. Also, would you like to remove the -c and -O3 flags to keep it simple ?

mfrw: sure, I would.
I can't add `-` after stat, as it works without it, whatever you write after `perf stat` is taken as an executable.
Also adding a `-` throws an error as the program is not written to handle it.

agnivade: Ok, I must have been looking at older documentation.

sbrl:
Thanks, @mfrw! :smiley_cat:

PR #1562, opened 2017-10-20
pacman: changed package install instruction to pacman -Su

CorePoint:
pacman -S $PACKAGE calls for problems due to potential partial updates.

### DIFF
--- a/pages/linux/pacman.md
+++ b/pages/linux/pacman.md
@@ -8,7 +8,7 @@
 
 - Install a new package:
 
-`pacman -S {{package_name}}`
+`pacman -Su {{package_name}}`
 
 - Remove a package and its dependencies:
 


agnivade:
Can you explain what the `-u` option does ?

CorePoint:
It upgrades packages that have newer versions according to your synced database.
It is explained [here](https://wiki.archlinux.org/index.php/System_maintenance#Partial_upgrades_are_unsupported) why it it's not a good idea.
Granted, if you didn't sync your database in between nothing will happen.
But it's simply a bad habbit to install packages with `pacman -S`.

Could you please commit the change your self if you add my change?
I don't want to grant a third party access to my data over a one character commit.

agnivade:
If you do not agree to the CLA, the repo cannot accept a PR from you. I can do the changes in that case.

I have a question though - the command is supposed to install a new package. So your response - 
> It upgrades packages that have newer versions according to your synced database.

is not exactly clear to me. It is supposed to install one package. And since its new, it will obviously install the latest version. So where is the question of upgrade coming from ?

CorePoint:
The command will install the package you specified (and the dependencies if needed) and simultaneously upgrade all packages that are not up to date according to your local database.

The explanation why it's needed is the following:

Lets say you updated your system a few days ago, then you synced your local database with one of your choosen mirrors in one of the following days.
If you now install a package with `pacman -S` it will only install the most recent version of the package and it's dependencies, lets say one of those dependencies is an important lib.
You will now have the most recent lib according to your local database installed on your system, but all installed packages that depend on this important lib are on the version that was recent at the last time you updated (or installed them in the first place for that matter), and they are of course build against a different version of the lib than the one that is installed on your system.

sbrl:
I can see how installing with `-Su` would be good practice - though coming from Ubuntu I've always run `sudo pacman -Syu` to do a full system upgrade _before_ I even think about installing anything new. However, I think that the description for this command may need updating to reflect this.

CorePoint:
Yeah, you can also install a package with `pacman -Syu $PACKAGE` but that would be not desirable of bandwith constricted connections.

Yeah i see how the description should be updated.
I'm not sure how to do it wihout being confusing.
If you would change the description to something like 

> upgrades your package to the recent version according to your database and installs defined package

it would be pretty long and would not explain why there isn't any command listed to just install a package.

Giving `pacman -Su` its own entry and adding "you should never do this" to the description of `pacman -S` would add a useless command but it would probably be the better way.

Maybe you have another ideas?

sbrl:
From what I'm reading, it both installs the package and updates all it's dependencies, correct? SO perhaps something like

```
Install a package and the latest versions of its dependencies?
```

CorePoint:
No, it updates all packages installed on your system.

sbrl:
Hrm, ok. I don't get the difference between `sudo pacman -Su package_name` and `sudo pacman -Syu` then - other than one installs a new package and the other doesn't.

CorePoint:
The `y` syncs your local DB to the mirror you've choosen. So you can `pacman -Syu` and then `pacman -Su ` for how often you want, the `pacman -Su` won't do anything because you already got the updates with `pacman -Syu` and as long don't sync your DB again you wont get any new updates.

agnivade:
How about 

```
Install a given package and update all other packages to their latest versions
```

CorePoint:
Yeah, sounds good.

agnivade:
Thanks. I will make the changes.

Foxboron:
This shouldn't have been merged. `-S` can never lead to partial updates unless `-Sy` is used, where the local databases have been updated but a system update isn't done. `-Su` is not considered best practise and is at best confusing.

agnivade:
ping @CorePoint

CorePoint:
He's a member of the Arch team so he surely knows what he is talking about.
But let me give you my perspective about not using -Su to install for someone unexpirienced.

Somebody who uses tldr for such a basic thing as his package manager, probably doesn't even know if when he synced his DB the last time and/or if he made a full update since then.
Maybe he didn't even do hit knowing what's happening in the background, it could be something like `pacman -Syu` and then closing your terminal, or your pc battery running out while doing that.

So for the scope of something like tldr, where you want to give people short instructions that are easy to grasp, you have a few choices:

1: Don't have an entry for pacman at all, and force the user to look it up, hopes that he does it in the wiki and understands what he's doing.

2: Don't have an entry for installing packages and assume the user just either is so unexpirienced that he has to look it up elsewhere like in number 1, or that he doesn't need this basic information anyway and hopes to see some tips for other (less basic, less ciritical) operations.

3: Having a `pacman -S` entry but with a warning and a maybe a link to the Arch wiki page for pacman.

4: Have a warning at the top of the tldr entry for pacman.

I just can tell you that partial updates are not a rare sight in the support channels for Arch, sometimes even users who aren't exactly new to the Distro do that and ask for help because they don't know what they did wrong. Having just a simple entry like `pacman -S to install a package` would definitly reinforce that problem.
That's why people  often recommend to just install with `pacman -Su`, because you should either know what you're doing when using `pacman -S` (and somebody who knows what he's doing probably wouldn't ask for the most basic operation) or you need to read the manual in whole.

Foxboron:
This is bikeshedding over a non existing problem. The tldr pages shouldn't care for deep explanations as to why, for the same reason the Arch rosetta[1] on pacman shouldn't. It doesn't matter and isn't in the scope of the project. The pages tells you that upgrading should be done with `-Syu`, thats it. If your computers stops in the middle of the upgrade tldr won't help you. It's not in the scope.


> Having just a simple entry like pacman -S to install a package would definitly reinforce that problem.

It won't.

> That's why people often recommend to just install with pacman -Su

This is the first time I have ever seen anyone recommend `-Su` as a way to install packages in pacman. 

[1] https://wiki.archlinux.org/index.php/Pacman/Rosetta

agnivade:
I have never used Arch. So I don't have much idea. If you guys can come to a decision on this, I can make the necessary changes.

@Foxboron - Are you saying we should revert to `-S` again ?

CorePoint:
If you do maybe add something like:

`pacman -Sy`
 
 
 - Syncs your package database with your mirror(s) [YOU SHOULD NEVER DO THIS]

agnivade:
That would over complicate things. We just want to keep it simple and to the point. Our first aim is to keep the no. of examples to a minimum. So adding an extra example only to say "YOU SHOULD NEVER DO THIS" does not make sense from tldr's perspective.

CorePoint:
Ok, i see. I can't really add anything more to this, besides telling you that i saw dozens of people borking their system because they installed packages with `pacman -S`, because they didn't update their packages after syncing their database.

Foxboron:
@CorePoint if someone borks their system with `-S` it is because they ran `-Sy`. This isn't recommended. It's not listed as something you should do. And shouldn't be mentioned. Telling users to run `-Su` because they could maybe be running `-Sy` is just wrong.

@agnivade Yes, please revert the change.

sbrl:
Fascinating discussion here. I started using an arch derivative on my travel ling laptop a few months ago, so it's helpful to read this :D

PR #1569, opened 2017-10-23
ffmpeg: update examples for modern syntax, add commentary on CRF vs CBR options, add remuxing example

hvindin:
----


pages/common/ffmpeg.md
```diff
@@ -4,7 +4,7 @@
 
 - Extract the sound from a video and save it as MP3:
 
-`ffmpeg -i {{video.mp4}} {{sound}}.mp3`
+`ffmpeg -i {{video.mp4}} -vn -codec:audio {{sound}}.mp3`
```
agnivade: We always try to give the simplest possible example as the first one. Is this absolutely necessary ?

pages/common/ffmpeg.md
```diff
@@ -14,10 +14,18 @@
 
 `ffmpeg -i {{frame_%d.jpg}} -f image2 {{video.mpg|video.gif}}`
 
-- Extract a single frame from a video at time mm:ss and save it as a 128x128 resolution image:
+- Quickly extract a single frame from a video at time mm:ss and save it as a 128x128 resolution image:
 
-`ffmpeg -i {{video.mp4}} -ss {{mm:ss}} -frames 1 -s {{128x128}} -f image2 {{image.png}}`
+`ffmpeg -ss {{mm:ss}} -i {{video.mp4}} -frames 1 -s {{128x128}} -f image2 {{image.png}}`
```
agnivade: You seem to have just changed the order. How does this speed up the execution time of the command ?

pages/common/ffmpeg.md
```diff
@@ -14,10 +14,18 @@
...
 
-- Convert AVI video to MP4. AAC Audio @ 128kbit, Video @ 1250Kbit:
+- Convert AVI video to MP4. AAC Audio @ 128kbit, h264 Video @ 1250kbit (Gives a predictable file size at lower quality):
```
agnivade: Gives => gives.

pages/common/ffmpeg.md
```diff
@@ -14,10 +14,18 @@
...
 
-`ffmpeg -i {{input_video}}.avi -acodec libfaac -ab 128k -vcodec mpeg4 -b 1250K {{output_video}}.mp4`
+`ffmpeg -i {{input_video}}.avi -codec:audio aac -b:audio 128k -codec:video libx264 -b:video 1250K {{output_video}}.mp4`
+
+- Convert AVI video to MP4. AAC Audio @ 128kbit, h264 Video @ CRF 23, Preset of very slow (Good trade off with file size/speed/quality), also allow faststart by moving the MOOV atom to the start of the file after processing:
```
agnivade: The following 2 examples seems to be an advanced usage of `ffmpeg` to me. Please note that a tldr page is mostly aimed at a beginner to get familiar with the basics of the command. For an exhaustive list, the man page is always there. Also, this takes the no. of examples to 7 which we don't usually go to unless absolutely necessary.

Do you think these examples are required for someone just starting to know the ffmpeg command ? Let me know what you think,

pages/common/ffmpeg.md
hvindin: I'd argue that the -vn is necessary as it's not uncommon to wind up with an MP3 file where ffmpeg has taken a guess and embedded something it can convert to an image out of the video stream into to tags.

Although you are right that specifying audio codec is probably superfluous.

pages/common/ffmpeg.md
hvindin: If the -ss is after the input file then ffmpeg re-encodes the whole file up to that point.

If -ss is specified before the input file then ffmpeg just starts from that point onwards.

See here for more details: https://trac.ffmpeg.org/wiki/Seeking

pages/common/ffmpeg.md
hvindin: Seems like a reasonable ask of a quick summary guide. Although I would suggest that speciting a bitrate for the video us almost never the correct option and suspect that the better "default" for a new user would be to specify a crf and let ffmpeg figure out the but rates.

The remuxing example is actually probaby the most common thing that is asked about on stack exchange/mailing lists that everyone defaults to the wrong action. If a user uses `ffmpeg -I {{video}}.mkv {{video}}.mp4` then ffmpeg will re-encode the whole thing using what it guesses are the best codecs for the format. If you specify to copy then the remux takes seconds and achieves what 99% of users are actually trying to do.

I'll update this pull with a smaller, simplified list of examples of just the bits I think are the most commonly asked questions just from what I've seen, and their most straightforward answers.

pages/common/ffmpeg.md
agnivade: Wow, I guess you live and learn. Thanks for the link. Also filled me with fond memories of Princess Mononoke :relaxed:

pages/common/ffmpeg.md
```diff
@@ -14,10 +14,14 @@
...
 
-- Convert AVI video to MP4. AAC Audio @ 128kbit, Video @ 1250Kbit:
+- Convert AVI video to MP4. AAC Audio @ 128kbit, h264 Video @ CRF 23:
 
-`ffmpeg -i {{input_video}}.avi -acodec libfaac -ab 128k -vcodec mpeg4 -b 1250K {{output_video}}.mp4`
+`ffmpeg -i {{input_video}}.avi -codec:audio aac -b:audio 128k -codec:video libx264 -crf 23 {{output_video}}.mp4`
+
+- Remux MKV video to MP4. No re-encoding of video or audio streams:
+
+`ffmpeg -i {{inpupt_video}}.mkv -codec copy  {{output_video}}.mp4`
```
agnivade: inpupt_video => input_video.

And also double space after "copy"

pages/common/ffmpeg.md
```diff
@@ -14,10 +14,14 @@
...
-`ffmpeg -i {{video.mp4}} -ss {{mm:ss}} -frames 1 -s {{128x128}} -f image2 {{image.png}}`
+`ffmpeg -ss {{mm:ss}} -i {{video.mp4}} -frames 1 -s {{128x128}} -f image2 {{image.png}}`
 
-- Convert AVI video to MP4. AAC Audio @ 128kbit, Video @ 1250Kbit:
+- Convert AVI video to MP4. AAC Audio @ 128kbit, h264 Video @ CRF 23:
 
-`ffmpeg -i {{input_video}}.avi -acodec libfaac -ab 128k -vcodec mpeg4 -b 1250K {{output_video}}.mp4`
+`ffmpeg -i {{input_video}}.avi -codec:audio aac -b:audio 128k -codec:video libx264 -crf 23 {{output_video}}.mp4`
+
+- Remux MKV video to MP4. No re-encoding of video or audio streams:
```
agnivade: How about "Remux a MKV video to MP4 without re-encoding the video or audio streams". ?

agnivade:
Looks like your latest commit got messed up.

hvindin:
Yikes. that's what I get for trying to do stuff on my phone!

Accidentally rebased to this branch rather than master. I'll close this PR and raise another one with only the relevant changes :)

PR #1570, opened 2017-10-23
nmap: added examples for scanning a host with nse scripts and scanning enabled SSL/TLS ciphers/protocols on a host on port 443

hvindin:

### DIFF
--- a/pages/common/nmap.md
+++ b/pages/common/nmap.md
@@ -26,3 +26,11 @@
 - Perform TCP and UDP scanning (use -sU for UDP only, -sZ for SCTP, -sO for IP):
 
 `nmap -sSU {{address_or_addresses}}`
+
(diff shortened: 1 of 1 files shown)


pages/common/nmap.md
```diff
@@ -26,3 +26,11 @@
 - Perform TCP and UDP scanning (use -sU for UDP only, -sZ for SCTP, -sO for IP):
 
 `nmap -sSU {{address_or_addresses}}`
+
+- Perform scan using an nmap nse script:
```
agnivade: I think this example is a bit corner case. I am okay with adding the next example.

hvindin: After reading the review of my ffmpeg pull I was actually just thinking that the generic "any nmap script" example probably wasn't enirely fitting. Will update this pull accordingly.

PR #1571, opened 2017-10-24
fatlabel: add page

sbrl:
Probably doesn't actually need a page, but I created one anyway because it's actually really useful (it's the only way I know to set a mixed-case label on a fat32 filesystem).

----

### DIFF
--- a/pages/linux/fatlabel.md
+++ b/pages/linux/fatlabel.md
@@ -2,10 +2,10 @@
 
 > Sets or gets the label of a FAT32 partition.
 
-- Get the label of a FAT32 partition: 
+- Get the label of a FAT32 partition:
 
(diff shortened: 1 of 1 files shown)


pages/linux/fatlabel.md
```diff
@@ -0,0 +1,11 @@
+# fatlabel
+
+> Sets or gets the label of a FAT32 partition.
+
+- Get the label of a FAT32 partition: 
+
+`fatlabel {{/dev/sdXY}}`
```
agnivade: Maybe better to just say sda1 or sda5 instead of sdXY

sbrl:
Thanks for the tip, @agnivade!

PR #1573, opened 2017-10-25
softwareupdate: add page

IanOxborrow:


pages/osx/softwareupdate.md
```diff
@@ -0,0 +1,19 @@
...
+
+`softwareupdate -ia`
+
+- Download and install all recommended updates:
+
+`softwareupdate -ir`
+
+- Download and install a specific app:
+
+`softwareupdate -i {{update_name}}`
```
agnivade: Should it be {{app_name}} instead of {{update_name}} ?

IanOxborrow: I thought it would be better to use update_name over app_name because the OS can also be updated through this method.

agnivade: I see. Alright.

PR #1574, opened 2017-10-25
leave: add page

m4grio:
----

### DIFF
--- a/pages/common/leave.md
+++ b/pages/common/leave.md
@@ -1,0 +1,20 @@
+# leave
+
+> Remind you when you have to leave.
+> To remove reminders use `kill $(pidof leave)`.
+
+- Set a reminder:
+
+`leave {{time_to_leave}}`
+
+- Remind you to leave at noon:
+
+`leave 1200`
+
+- Set a reminder in a specific amount of time:
+
+`leave +{{amount_of_time}}`
+
+- Remind you in 4 hours and 4 minutes:
+
+`leave +0404`
--- a/pages/common/leave.md
+++ b/pages/common/leave.md
@@ -1,13 +1,13 @@
 # leave
 
-> Remind you when you have to leave.
+> Remind when it's time to leave.
 > To remove reminders use `kill $(pidof leave)`.
 
 - Set a reminder:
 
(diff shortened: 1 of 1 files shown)


m4grio:
Hi, @sbrl. It reminds you just by feeding `'Time to leave'` to the terminal 5 minutes before, at the time and every minute thereafter until you logout or manually kill the process/es – its intention is to remind you when to leave.

I normally use it to set quick reminders like for example when I'm playing a game but I need to wait certain amount of time to do something further.

The synopsis is `leave [[+]hhmm]`, where `hh` is the time in hours and `mm` the time in minutes – one nedds to set the exact time of the day to be reminded, unless the argument is preceded by `+`, which means the reminder will be triggered in `hh` hours and `mm` minutes from now.

I actually struggled to decide whether should I clarify that in the arguments with the options currently available in the `tldr` lexis and came up with the current proposal, leaving further clarification to the `man` page – which is, by the way, pretty clear and easy to understand.

Happy to discuss any further change. Couple of options:
- More examples – clarify through examples how the argument is assembled.
- Detailed explanation on the `hh`, `+`, and `mm` usage in the first example.
- Different wording.

sbrl:
Ah, cool! Looking at it, perhaps we could put the last example 2nd? I think that should make it more obvious.

m4grio:
I think it would be rather confusing that way.

This is the rational behind why I decided to put the examples in the order of my proposal:
1. The synopsis of remind-me-at-x command.
2. An example of remind-me-at-x command.
3. The synopsis of remind-me-in-x command.
4. An example of remind-me-in-x command.

Maybe the issue is not the *order* but the *clarity* of the descriptions – shall we try to reword it? A bit of help would help me a lot.

m4grio:
Haha no worries.

agnivade:
Hi @m4grio - we usually keep the language indirect and from a 3rd person perspective. So we try to avoid mentioning 'you'.

Can you update the PR to change the tone so that 'you' is not mentioned ?

agnivade:
@m4grio made some slight changes of my own. Let me know if they look okay to you.

PR #1581, opened 2017-10-25
dotnet: add page

gldraphael:
The CLI is quite extensive, I've tried to cover the most frequently used commands.
A similar page could be added for `mono`.

----

### DIFF
--- a/pages/common/dotnet.md
+++ b/pages/common/dotnet.md
@@ -4,7 +4,7 @@
 
 - Initialize a new .NET project:
 
-`dotnet new project-type`
+`dotnet new template-short-name`
 
 - Restore nuget packages:
 
--- a/pages/common/dotnet.md
+++ b/pages/common/dotnet.md
@@ -1,6 +1,6 @@
 # dotnet
 
-> .NET Command Line Tools
+> Cross Platform .NET Command Line Tools for .NET Core
 
 - Initialize a new .NET project:
 
--- a/pages/common/dotnet.md
+++ b/pages/common/dotnet.md
@@ -1,6 +1,6 @@
 # dotnet
 
-> Cross Platform .NET Command Line Tools for .NET Core
+> Cross Platform .NET Command Line Tools for .NET Core.
 
 - Initialize a new .NET project:
 
--- a/pages/common/dotnet.md
+++ b/pages/common/dotnet.md
@@ -14,6 +14,6 @@
 
 `dotnet run`
 
-- Run a packaged dotnet application (only needs the runtime, the rest require the .NET Core SDK installed):
+- Run a packaged dotnet application (only needs the runtime, the rest of the commands require the .NET Core SDK installed):
 
 `dotnet {{path/to/application.dll}}`
--- a/pages/common/dotnet.md
+++ b/pages/common/dotnet.md
@@ -4,7 +4,7 @@
 
 - Initialize a new .NET project:
 
-`dotnet new template-short-name`
+`dotnet new {{template_short_name}}`
 
 - Restore nuget packages:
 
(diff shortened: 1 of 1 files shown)


pages/common/dotnet.md
```diff
@@ -0,0 +1,19 @@
+# dotnet
+
+> .NET Command Line Tools
+
+- Initialize a new .NET project
```
sbrl: Don't know what's up with Travis CI here! Anyway, we prefer a colon at the end of each command description.

pages/common/dotnet.md
```diff
@@ -0,0 +1,19 @@
+# dotnet
+
+> .NET Command Line Tools
```
sbrl: Perhaps this description could be expanded into a full sentence?

pages/common/dotnet.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Restore nuget packages
+
+`dotnet restore`
+
+- Build and execute the .NET project in the current directory
+
+`dotnet run`
+
+- Run a packaged dotnet application (only needs the runtime, the rest require the .NET Core SDK installed)
```
sbrl: By "the rest" you mean a compiled CIL executable? Also, `require` should be `requires`.

gldraphael: By rest I meant the rest of the commands.

There are two options for installing .NET Core:

1. The complete SDK (for development and build environments)
1. Only the runtime (for production environments)

The last command is the only command that the runtime supports. I used require because it refers to the _rest of the commands_ which is plural. Do you think we should get rid of the text in the parenthesis? Because what I've put there is not clear enough.

sbrl: Ahhhh cool! I see. With mono there's the `mono` command (the runtime), and `mcs` (the compiler).

And ahhhhh I see why that's not pluralised - I completely misinterpreted that lol! Perhaps `the rest of the commands require` would work better then.

pages/common/dotnet.md
```diff
@@ -0,0 +1,19 @@
+# dotnet
+
+> Cross Platform .NET Command Line Tools for .NET Core
```
agnivade: Period at the end.

pages/common/dotnet.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Initialize a new .NET project:
+
+`dotnet new project-type`
```
agnivade: shouldn't project-type be the project name ? What is project-type supposed to mean ?

pages/common/dotnet.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Restore nuget packages:
```
agnivade: Maybe good to explain what does restore do actually.

sbrl: Restore is a nuget-specific term that basically means "make all the installed packages equal those defined in the nuget configuration file" - which is basically like `npm install`.

gldraphael: Yes, so I'm guessing dotnet users will be familiar with nuget packages.

pages/common/dotnet.md
gldraphael: I'll change `project-type` to `template-short-name`. Sounds more correct.

Here's the possible values it can take:
dotnet-new

I would have added them to the example command if they were just a few.

agnivade: I meant they need to be wrapped in tokens. I have added them now.

agnivade:
@gldraphael - how does it look now ?

pages/common/dotnet.md
```diff
@@ -1,6 +1,6 @@
 # dotnet
 
-> Cross Platform .NET Command Line Tools for .NET Core.
```
gldraphael: It's called title-case 😅

gldraphael:
Looks good to me 🙃

agnivade:
> Is there no way to comment on a commit message!

In gerrit you can. Not here in github.

sbrl:
Merging anyway since @travis-ci seems to be confused again.

PR #1584, opened 2017-10-26
tee: Add usecase for process substitution

pollev:
----

### DIFF
--- a/pages/common/tee.md
+++ b/pages/common/tee.md
@@ -12,4 +12,4 @@
 
 - Create a folder called "example", print "example" and write "example" to the specified file:
 
-`echo "example" | tee >(xargs mkdir) >(cat) >path/to/file`
+`echo "example" | tee >(xargs mkdir) >(cat) >{{path/to/file}}`
--- a/pages/common/tee.md
+++ b/pages/common/tee.md
@@ -1,6 +1,6 @@
 # tee
 
-> Read from standard input and write to standard output and files.
+> Read from standard input and write to standard output and files (or commands).
 
 - Copy standard input to each FILE, and also to standard output:
 
@@ -9,3 +9,7 @@
 - Append to the given FILEs, do not overwrite:
 
 `echo "example" | tee -a {{FILE}}`
+
+- Create a folder called "example", print "example" and write "example" to FILE:
+
+`echo "example" | tee >(xargs mkdir) >(cat) >{{FILE}}`
--- a/pages/common/tee.md
+++ b/pages/common/tee.md
@@ -10,6 +10,6 @@
 
 `echo "example" | tee -a {{FILE}}`
 
-- Create a folder called "example", print "example" and write "example" to the specified file:
+- Create a folder called "example", count the number of characters in "example and write "example" to the terminal:
 
-`echo "example" | tee >(xargs mkdir) >(cat) >{{path/to/file}}`
+`echo "example" | tee >(xargs mkdir) >(wc -c)
--- a/pages/common/tee.md
+++ b/pages/common/tee.md
@@ -10,6 +10,6 @@
 
 `echo "example" | tee -a {{FILE}}`
 
-- Create a folder called "example", count the number of characters in "example and write "example" to the terminal:
+- Create a folder called "example", count the number of characters in "example" and write "example" to the terminal:
 
 `echo "example" | tee >(xargs mkdir) >(wc -c)`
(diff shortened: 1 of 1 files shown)


pages/common/tee.md
```diff
@@ -9,3 +9,7 @@
 - Append to the given FILEs, do not overwrite:
 
 `echo "example" | tee -a {{FILE}}`
+
+- Create a folder called "example", print "example" and write "example" to FILE:
+
+`echo "example" | tee >(xargs mkdir) >(cat) >{{FILE}}`
```
sbrl: Perhaps we could change this to `path/to/file`? That's the convention we've used in other pages.

pages/common/tee.md
```diff
@@ -9,3 +9,7 @@
 - Append to the given FILEs, do not overwrite:
 
 `echo "example" | tee -a {{FILE}}`
+
+- Create a folder called "example", print "example" and write "example" to FILE:
```
sbrl: Perhaps `to the specified file` would read better?

pollev:
I always forget as well, hence why I decided to add it here :D 

Besides it usefulness as a logging tool, I believe tee is the only standard linux tool for duplicating the stdout of a command into multiple other commands.

Anyway, your remarks are great, I will edit the page

pages/common/tee.md
```diff
@@ -10,6 +10,6 @@
 
 `echo "example" | tee -a {{FILE}}`
 
-- Create a folder called "example", print "example" and write "example" to FILE:
+- Create a folder called "example", print "example" and write "example" to the specified file:
 
-`echo "example" | tee >(xargs mkdir) >(cat) >{{FILE}}`
+`echo "example" | tee >(xargs mkdir) >(cat) >path/to/file`
```
sbrl: Oops! Don't forget the `{{` and `}}` either side of `path/to/file` :P

pollev: Ah I figured it was one or the other. will update

pages/common/tee.md
```diff
@@ -9,3 +9,7 @@
 - Append to the given FILEs, do not overwrite:
 
 `echo "example" | tee -a {{FILE}}`
+
+- Create a folder called "example", print "example" and write "example" to the specified file:
```
agnivade: This seems a bit complicated. The idea is useful though. I think we can remove the "xargs mkdir" one, and just keep the cat and file output.

It still shows the idea that something can be read from standard input and can be redirected to commands and files.

pollev: I'm sorry but I have to disagree, the entire point is that tee makes it possible to duplicate output streams to the input streams of multiple commands. 
The goal is not to redirect output. You could do that by just using regular pipes.

Tee is, as far as I know, the only basic unix utility that is capable of doing this.

However, this is just my take on what makes a clear example. I used mkdir and cat as these will probably be commands most people are familiar with. Actual use cases are much more likely to be stream editors like sed or even tools like grep

agnivade: Ok fair enough. How about we split this concept into 2 examples, so that we go easy on a newcomer. First example can be the cat and file output. And second can be the mkdir and cat one. What do you think ?

pollev: I think we're getting there. Honestly it's pretty hard to think of an easy example.
I think it is actually cat that is the bad command in this example.

Let me explain why I think so and maybe you have an idea to improve this example?

echo "example" | tee >(command1) >(command2)
The above example will:
- feed 'example' into command 1.
- feed 'example' into command 2.
- print 'example' on the terminal

this happens because tee duplicates the output not two but three times. Once for each file or command listed. And one additional time as regular output.

echo "example" | tee >(command1) >(command2) >/dev/null
will:
- feed 'example' into command 1.
- feed 'example' into command 2.

because now we just send the 'normal' stdout to /dev/null.

echo "example" | tee >(cat)
will:
- print 'example on the terminal.
- print 'example on the terminal a second time

Once due to normal stdout and once due to the cat command.

So what I am trying to say here is that cat is perhaps the worst command to use as an example. because if you do not redirect the normal output, it will be sent to the terminal anyway.

A real example would be:
echo "this string contains numbers 12345 and symbols #{$" | tee >(grep -o "..3..") >(grep -o ".{.")
output:
- this string contains numbers 12345 and symbols #{$
- 12345
- #{$

But maybe grep is too confusing of a command itself to use in such an example?

agnivade: Ah, I didn't realise that it will print to the terminal anyway. How about a `wc -c` instead of `cat` ? The `mkdir` seems fine.

pollev:
Proposed new changes. I think this might be better?

agnivade:
Great. I just added a missing quote.

sbrl:
Thanks, @pollev :smiley_cat:

PR #1585, opened 2017-10-26
modinfo: add page

aydwi:
----

### DIFF
--- a/pages/linux/modinfo.md
+++ b/pages/linux/modinfo.md
@@ -1,7 +1,11 @@
 # modinfo
 
-> Extract information about a Linux Kernel module.
+> Extract information about a Linux kernel module.
 
-- Print value(s) of the specified field(s) instead of all the fields:
+- List all attributes of a kernel module:
 
-`modinfo -F {{author/description/license/parm/filename}} {{kernel_module_name}}`
+`modinfo {{kernel_module}}`
+
+- List value(s) of the specified field(s) instead of all the fields:
+
+`modinfo -F {{author|description|license|parm|filename}} {{kernel_module}}`
--- a/pages/linux/modinfo.md
+++ b/pages/linux/modinfo.md
@@ -1,0 +1,7 @@
+# modinfo
+
+> Extract information about a Linux Kernel module.
+
+- Print value(s) of the specified field(s) instead of all the fields:
+
+`modinfo -F {{author/description/license/parm/filename}} {{kernel_module_name}}`
--- a/pages/linux/modinfo.md
+++ b/pages/linux/modinfo.md
@@ -6,6 +6,6 @@
 
 `modinfo {{kernel_module}}`
 
-- List value(s) of the specified field(s) instead of all the fields:
+- List the specified attribute(s) only:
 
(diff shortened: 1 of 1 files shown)


pages/linux/modinfo.md
```diff
@@ -0,0 +1,7 @@
+# modinfo
+
+> Extract information about a Linux Kernel module.
```
agnivade: linux kernel. Lower case.

pages/linux/modinfo.md
```diff
@@ -0,0 +1,7 @@
...
+
+- Print value(s) of the specified field(s) instead of all the fields:
```
agnivade: I think we should start with `modinfo {{kernel_module}}` and explain what it does.

and then go to `-F`.

Also, instead of /, use |.

pages/linux/modinfo.md
aydwi: Made kernel lowercase. I think Linux should stay as it is.

pages/linux/modinfo.md
aydwi: Updated the description

pages/linux/modinfo.md
```diff
@@ -0,0 +1,11 @@
+# modinfo
+
+> Extract information about a Linux kernel module.
+
+- List all attributes of a kernel module:
+
+`modinfo {{kernel_module}}`
+
+- List value(s) of the specified field(s) instead of all the fields:
```
agnivade: You mention "attributes" above and "fields" in this example. Please use any one to be consistent. 
 
Also, no need to say "instead of all the fields". The first example clarifies it that not setting any flags shows all attributes.

aydwi: Updated it further. Please check if it is appropriate now.

pages/linux/modinfo.md
```diff
@@ -0,0 +1,11 @@
...
+
+> Extract information about a Linux kernel module.
+
+- List all attributes of a kernel module:
+
+`modinfo {{kernel_module}}`
+
+- List the specified attribute(s) only:
+
+`modinfo -F {{author|description|license|parm|filename}} {{kernel_module}}`
```
sbrl: It's a bit ambiguous here how multiple attributes are delimited. Is it with the vertical bar?

agnivade: Yes the vertical bar signifies multiple values which can be used. It's what we have been using.

agnivade: Ah no, I don't think so. ping @aydwi

aydwi: Actually, multiple attributes can be specified, separated by a space.

agnivade: Ok, then in that case change it to - `List the specified attribute(s) only. (You can specify multiple attributes separated by space)`

aydwi: After some testing, it looks like it does not accept multiple attributes in a single command. I've fixed the mistake and pushed again.

agnivade:
Thanks for verifying.

PR #1590, opened 2017-10-27
trans: add page

AnimiVulpis:
----

### DIFF
--- a/pages/common/trans.md
+++ b/pages/common/trans.md
@@ -16,8 +16,8 @@
 
 - Translate a word from German to English:
 
-`trans {{de}}:{{en}} {{handy}}`
+`trans {{de}}:{{en}} {{Schmetterling}}`
 
 - Behave like a dictionary to get the meaning of a word:
 
-`trans -d {{handy}}`
+`trans -d {{word}}`
--- a/pages/common/trans.md
+++ b/pages/common/trans.md
@@ -1,0 +1,23 @@
+# trans
+
+> **Translate Shell** is a command-line translator.
+
+- Translate a word (language is detected automatically):
+
+`trans "{{word_or_sentence_to_translate}}"`
+
+- Get a brief translation:
+
+`trans --brief "{{word_or_sentence_to_translate}}"`
+
+- Translate a word into french:
+
+`trans :fr {{word}}`
+
+- Translate from a specific language:
+
+`trans de:en {{handy}}`
+
+- Use `trans` as a dictionary:
+
+`trans -d handy`
--- a/pages/common/trans.md
+++ b/pages/common/trans.md
@@ -1,6 +1,6 @@
 # trans
 
-> **Translate Shell** is a command-line translator.
+> Translate Shell is a command-line translator.
 
 - Translate a word (language is detected automatically):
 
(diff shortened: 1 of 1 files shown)


pages/common/trans.md
```diff
@@ -0,0 +1,23 @@
+# trans
+
+> **Translate Shell** is a command-line translator.
```
agnivade: Why the ** ? The clients do not parse and display according to exact markdown rules. So the ** will be shown as ** itself, not bold text.

pages/common/trans.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Get a brief translation:
+
+`trans --brief "{{word_or_sentence_to_translate}}"`
+
+- Translate a word into french:
+
+`trans :fr {{word}}`
+
+- Translate from a specific language:
```
agnivade: Maybe "Translate a word from German to English" sounds more clear and specific.

And why "handy" ?

pages/common/trans.md
```diff
@@ -0,0 +1,23 @@
...
+
+`trans de:en {{handy}}`
+
+- Use `trans` as a dictionary:
```
agnivade: Maybe better to say "Behave like a dictionary to get the meaning of a word".

pages/common/trans.md
```diff
@@ -0,0 +1,23 @@
...
+
+`trans -d handy`
```
agnivade: Wrap in tokens.

pages/common/trans.md
AnimiVulpis: I just wanted to keep the description like on this page: https://www.soimort.org/translate-shell/
But didn't notice that this will not work. Will remove it

pages/common/trans.md
AnimiVulpis: True, will change.
**handy** because:
1. I expect most users of `tldr` to get the english auto-detection when using `trans`, which would result in a dictionary style explanation and not a translation. Thus making the use of `de:en` necessary (in this case)
2. The word might seem english (even to a lot of germans), but in fact is not and so people might learn something 😄

agnivade: Can we use a German-sounding word, so that its clear that we are translating from German to English ?

AnimiVulpis: Of course. Do you have a suggestion? As someone who speaks german day-in day-out I might not be the best person to judge the german-ity of german words.
Some suggestions:
- Bier
- Schmetterling
- Kreuzung
- Bahnhof
- Krankenwagen

agnivade: I could guess the first one :wink: The 2nd one looks very german-sounding to me. Another reason to vote for 2nd is because it means butterfly, which is a nice thing.

AnimiVulpis: I will take this then. Only _downside_ to this is that **Schmetterling** does not really need the `source` language defined, because google would auto-determine the source language correctly.
But I think that **this** does not matter as much in this case of `tldr`.
Easier to understand is better in this cace.

AnimiVulpis:
I am not sure if/where I should add a note that not all languages support dictionary mode.
The list [here](https://github.com/soimort/translate-shell/wiki/Languages) shows which languages are supported.

It is not even mentioned in the man page. Only [here](https://www.soimort.org/translate-shell/#dictionary-mode) is a **note** about it.

agnivade:
You can add a second line in the heading section.

sbrl:
Looks ok to me! Thanks, @AnimiVulpis :smiley_cat:

PR #1591, opened 2017-10-27
roll: add page

AnimiVulpis:
----

### DIFF
--- a/pages/common/roll.md
+++ b/pages/common/roll.md
@@ -1,0 +1,27 @@
+# roll
+
+> Rolls a user-defined dice sequence.
+
+- Roll 3 6-sided dices and sums the results:
+
+`roll 3d`
+
+- Roll 1 8-sided dice, add 3 and sum the results:
+
+`roll d8 + 3`
+
+- Roll 4 6-sided dices, keep the 3 highest results and sum the results:
+
+`roll 4d6h3`
+
+- Roll 2 12-sided dices 2 times and show every roll:
+
+`roll --verbose 2{2d12}`
+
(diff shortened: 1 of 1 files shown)


pages/common/roll.md
```diff
@@ -0,0 +1,27 @@
+# roll
+
+> Rolls a user-defined dice sequence.
+
+- Roll 3 6-sided dices and sums the results:
```
sbrl: Isn't the singular of dice "die"?

AnimiVulpis: Checked back on this and you are right:
One **die**
Multiple **dice**
**dices** is a verb if I understood correctly (again something learned ^^)

Will change accordingly

AnimiVulpis:
Did I tokenize correctly?

agnivade:
You have to put the {{ }} only once. So something like {{3d}}, {{4d6h3}}.

pages/common/roll.md
```diff
@@ -0,0 +1,27 @@
...
+
+`roll {{4d6h3}}`
+
+- Roll 2 12-sided dice 2 times and show every roll:
+
+`roll --verbose {{2{2d12}}}`
+
+- Roll 2 20-sided dice until the result is bigger than 10:
+
+`roll "{{2d20>10}}"`
```
agnivade: Just curious why is this example in quotes. Others don't have quotes.

AnimiVulpis: It's necessary because otherwise the output of `roll 2d20` would be piped into a file called `10`

sbrl:
Thanks, @AnimiVulpis! That's another useful command added :smiley_cat:

PR #1595, opened 2017-10-28
ffmpeg: update examples for modern syntax, add commentary on CRF vs CBR options, add remuxing example

hvindin:
fixed copy of #1569 @agnivade 

----

### DIFF
--- a/pages/common/ffmpeg.md
+++ b/pages/common/ffmpeg.md
@@ -29,4 +29,3 @@
 - Remux MKV video to MP4. No re-encoding of video or audio streams, convert subtitles stream to MP4 compatible codec:
 
 `ffmpeg -i {{inpupt_video}}.mkv -codec:audio copy -codec:video copy -codec:s mov_text --movflags +faststart {{output_video}}.mp4`
-
--- a/pages/common/ffmpeg.md
+++ b/pages/common/ffmpeg.md
@@ -4,7 +4,7 @@
 
 - Extract the sound from a video and save it as MP3:
 
-`ffmpeg -i {{video.mp4}} {{sound}}.mp3`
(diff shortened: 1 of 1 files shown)


pages/common/ffmpeg.md
```diff
@@ -14,10 +14,14 @@
...
 `ffmpeg -i {{frame_%d.jpg}} -f image2 {{video.mpg|video.gif}}`
 
-- Extract a single frame from a video at time mm:ss and save it as a 128x128 resolution image:
+- Quickly extract a single frame from a video at time mm:ss and save it as a 128x128 resolution image:
 
-`ffmpeg -i {{video.mp4}} -ss {{mm:ss}} -frames 1 -s {{128x128}} -f image2 {{image.png}}`
+`ffmpeg -ss {{mm:ss}} -i {{video.mp4}} -frames 1 -s {{128x128}} -f image2 {{image.png}}`
 
-- Convert AVI video to MP4. AAC Audio @ 128kbit, Video @ 1250Kbit:
+- Convert AVI video to MP4. AAC Audio @ 128kbit, h264 Video @ CRF 23:
```
sbrl: What's CRF?

hvindin: Constant Rate Factor, it's the recommended way to simply encode a video file. Unless you have some pretty specific file size requirements or are encoding for a very specific purpose your best bet is usually just to use the CRF options, which will let ffmpeg figure out the bitrate for you.

Check here: https://trac.ffmpeg.org/wiki/Encode/H.264 for the details, but if the tldr pages are looking to keep things as simple as reasonably possible, using crf instead of cbr is the way to do it. Things start to either return unpredictable results or get very complicated trying to do cbr encoding.

PR #1609, opened 2017-10-28
pwgen: Duplicated command from Linux to OSX

PurpleBooth:
pwgen is available on OSX as well as Linux. This will move the
documentation to common so it works on both environments.

----

### DIFF
--- a/pages/osx/pwgen.md
+++ b/pages/osx/pwgen.md
@@ -1,0 +1,15 @@
+# pwgen
+
+> Generate pronounceable passwords.
+
+- Generate random password with s[y]mbols:
+
+`pwgen -y {{length}}`
+
+- Generate secure, hard-to-memorize passwords:
+
+`pwgen -s {{length}}`
+
+- Generate password with at least one capital letter in them:
+
+`pwgen -c {{length}}`


agnivade:
If it is not on windows, and just on linux and osx, then it needs to be duplicated across both folders.

PurpleBooth:
Duplicated as advised.

pages/osx/pwgen.md
```diff
@@ -0,0 +1,15 @@
+# pwgen
+
+> Generate pronounceable passwords.
+
+- Generate random password with s[y]mbols:
```
sbrl: What's `y` doing in square brackets here?

PurpleBooth: Unsure, it came from the original from linux, is it best to make a new PR to remove it from there, or do it from both here?

agnivade: Haha, its a convention we came up with. If the parameter in the example is a single letter, then at least a word in the description has to start with it.

Here, the parameter is `-y` and it stands for symbol. Which is weird. Hence [y] to denote the `-y`.

sbrl: Ahh cool! Haven't seen that elsewhere.

sbrl:
Thanks, @PurpleBooth!

PR #1611, opened 2017-10-29
git-diff: Adds comparison of file on current branch to other branch

deiga:
----

### DIFF
--- a/pages/common/git-diff.md
+++ b/pages/common/git-diff.md
@@ -29,3 +29,7 @@
 - Compare a single file between two branches or commits:
 
 `git diff {{branch_1}}..{{branch_2}} [--] {{path/to/file}}`
+
+- Compare a single file from current branch to other branch
+
+`git diff {{branch}}[:{{path/to/file}}] {{path/to/file}}`
--- a/pages/common/git-diff.md
+++ b/pages/common/git-diff.md
@@ -30,6 +30,6 @@
 
 `git diff {{branch_1}}..{{branch_2}} [--] {{path/to/file}}`
 
-- Compare a single file from current branch to other branch:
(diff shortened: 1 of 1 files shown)


deiga:
@agnivade It does, but it's a more succinct syntax in comparing files from the current branch

deiga:
And it allows you to compare files which are named differently

agnivade:
Can we change the example to compare two different files and thereby avoid the [] ? The use of [] is a bit confusing.

deiga:
In that case I'd rather make 2 examples, since my purpose of adding was the easy comparison of current branch file to other branch

agnivade:
Unfortunately, we cannot make it 2 examples because that would exceed the max quota that we have of 8 examples. 

You can choose the one you think is the most important and practical one and we can go with that.

deiga:
Why do you think that `[]` is confusing? It's a common practice in manpages to denote optional params etc.

agnivade:
It is because of the description. You mention compare a single file from current branch to other branch, so the natural thought process is that the required input is the file, and the other branch. But then there is another file within []. So, that makes the user think what is that about. It takes some time to realize that the same command can be used to compare a different file too.

Here in tldr, we frequently do not use the man page syntax and focus on more practical and real-world examples. While using optional parameters are succinct, they only increase the cognitive load for a user. Therefore, my suggestion to use just one method.

Please note that we do not aim to cover all possible syntaxes. That is what the man page is for. We want to focus on 8 (and if possible less than that) most practical and real-world examples.

Let me know your thoughts.

deiga:
@agnivade I went with the "compare 2 different files"

deiga:
@sbrl jup :)

PR #1612, opened 2017-10-29
rspamc: add page

sbrl:
I keep forgetting how to train my spam filter, and I realised the reason was that I hadn't created a tldr page yet :P

Apparently you can even adjust the weighting of individual rules with this command, but I haven't yet had to do that, so I'm not too sure on how it's done - which probably means it's a more advanced usage of the command.

----


sbrl:
@agnivade Lol! Yeah, it's an email that you want to keep that isn't spam :P

PR #1636, opened 2017-11-19
cal: Add use case for spanning date

nzachow:
For clarity I changed "-m" to "--monday".
Also, I've removed the "cal -y" for simplicity.
You could just user "cal {year}" to get the same information.

----

~~- [ ] The page (if new), has been added to the correct platform folder:  ~~
~~      `common/` if it's common to all platforms, `linux/` if it's Linux-specific, and so on.~~

### DIFF
--- a/pages/linux/cal.md
+++ b/pages/linux/cal.md
@@ -6,7 +6,7 @@
 
 `cal`
 
-- Display previous, current and next month: 
+- Display previous, current and next month:
 
 `cal --three`
 
--- a/pages/linux/cal.md
+++ b/pages/linux/cal.md
@@ -6,13 +6,13 @@
 
 `cal`
 
+- Display previous, current and next month: 
+
+`cal --three`
+
 - Use monday as the first day of the week:
 
-`cal -m`
-
-- Display a calendar for the current year:
-
-`cal -y`
+`cal --monday`
 
 - Display a calendar for a specific year (4 digits):
 


nzachow:
On Arch linux cal from util-linux 2.31 works as intended.
[Online documentation](http://man7.org/linux/man-pages/man1/cal.1.html) also suggests that these options should work.

What you think is the best approach here?

agnivade:
Seems like ubuntu is taking cal from the freebsd variant https://www.unix.com/man-page/freebsd/1/cal/. Since we already have the proper page for cal under the osx folder. This can be more linux like.

pages/linux/cal.md
```diff
@@ -6,13 +6,13 @@
...
+- Display previous, current and next month:
 
-`cal -m`
+`cal --three`
 
-- Display a calendar for the current year:
+- Use monday as the first day of the week:
 
-`cal -y`
+`cal --monday`
```
agnivade: Please wrap monday in tokens ({{}})

nzachow: Sorry, but I'm not sure I understood what you meant.
I got the impression that you want this: `cal {{--monday}}`
Is this right ?

agnivade: cal --{{monday}}. https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#token-syntax

nzachow: This wouldn't cause confusion as someone could think that `{{monday}}` could be swapped for any other day ?

agnivade: Oh I see that only monday or sunday is applicable. I initially thought any day can be used.

agnivade:
ping @sbrl

sbrl:
Thanks for the PR, @nzachow 😺

PR #1639, opened 2017-11-22
Sudo is not needed for npm install

eddiejaoude:
----

- [ ] ~~The page has 8 or fewer examples.~~

- [ ] ~~The PR is appropriately titled:  
      `<command name>: add page` for new pages, or `<command name>: <description of changes>` for pages being edited~~

- [ ] ~~The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@
 
 A collection of simplified and community-driven man pages.
 
-Install it with `sudo npm install -g tldr`
+Install it with `npm install -g tldr`
 or [try the web client](http://tldr.ostera.io).
 
 ## What is tldr?


sbrl:
Not true. `sudo` is required on my system if I want to install a package globally. Indeed, it is not required to install locally, but it is (on my system at least) required to install globally.

eddiejaoude:
> Not true. sudo is required on my system if I want to install a package globally. Indeed, it is not required to install locally, but it is (on my system at least) required to install globally.

Your system is not setup correctly! From the [npm docs](https://docs.npmjs.com/getting-started/installing-npm-packages-globally)

> To download packages globally, you simply use the command anpm install -g <package>a, e.g.:
> `npm install -g jshint`
> If you get an EACCES error, you should fix your permissions. You could also try using sudo, but this should be avoided:

> **You could also try using sudo, but this should be avoided:**

sbrl:
Ok then! Thanks for the correction, @eddieantonio 😺

eddiejaoude:
You are welcome @sbrl, I hope that helps 👍

agnivade:
@sbrl - FYI - everywhere else the `npm install` is without sudo. In the node client repo, there is a section which mentions why root is not needed. Hence this was done to make everything uniform.

PR #1640, opened 2017-11-22
gtop: add page

natesilva:
Added page for `gtop` (https://github.com/aksakalli/gtop).

----

- [ x ] The page (if new), does not already exist in the repo.

- [ x ] The page (if new), has been added to the correct platform folder:  

- [ x ] The page has 8 or fewer examples.

- [ x ] The PR is appropriately titled:  

- [ x ] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines

### DIFF
--- a/pages/common/gtop.md
+++ b/pages/common/gtop.md
@@ -1,0 +1,19 @@
+# gtop
+
+> System monitoring dashboard for the terminal.
+
+- Show the system stats dashboard:
+
+`gtop`
+
+- Fix question marks or other incorrect characters:
+
+`LANG=en_US.utf8 TERM=xterm-256color gtop`
+
+- Sort by CPU usage:
+
+`c`
+
+- Sort by memory usage:
+
+`m`
--- a/pages/common/gtop.md
+++ b/pages/common/gtop.md
@@ -6,10 +6,6 @@
 
 `gtop`
 
-- Fix question marks or other incorrect characters:
-
-`LANG=en_US.utf8 TERM=xterm-256color gtop`
-
 - Sort by CPU usage:
 
 `c`


pages/common/gtop.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Show the system stats dashboard:
+
+`gtop`
+
+- Fix question marks or other incorrect characters:
+
+`LANG=en_US.utf8 TERM=xterm-256color gtop`
+
+- Sort by CPU usage:
```
sbrl: I'm not sure it's clear whether this is a command or a keyboard key that you should press. @agnivade, thoughts?

pages/common/gtop.md
```diff
@@ -0,0 +1,19 @@
+# gtop
+
+> System monitoring dashboard for the terminal.
+
+- Show the system stats dashboard:
+
+`gtop`
+
+- Fix question marks or other incorrect characters:
```
agnivade: I want to improve this description a bit. Can you explain the context behind this ?

I would ideally want to split this into 2 examples clearly mentioning the motive behind tuning each variable. For eg. in my system, the value of TERM is already `xterm-256color`.

pages/common/gtop.md
agnivade: Its a keyboard key I would assume. Since the command is similar to `top`

pages/common/gtop.md
natesilva: On certain terminals where the `LANG` or `TERM` are not set correctly, `gtop` displays question marks or other junk instead of its normal output. Using this command format fixes it. (This is documented on the Github page for `gtop`.)

I have to admit, one reason I wanted to create this `tldr` is because I wanted to have this command handy instead of looking it up on the Github page for the project.

natesilva: The Github page says, for this form of the command:

> If you see question marks or other different characters, try to run it with these environment variables:
>
> `$ LANG=en_US.utf8 TERM=xterm-256color gtop`

That full description is too long for `tldr`, if you have a better suggestion let me know.

agnivade: This seems more like a way to fix some issues rather than some command specific thing of `gtop` itself. @sbrl - Thoughts ?

pages/common/gtop.md
sbrl: Ok then, let's leave it then

pages/common/gtop.md
sbrl: Yeah, I'd agree. This is a way to fix a setup issue rather than do something with the command. If you've got `gtop` installed, it's highly likely that you've read the README - since it isn't bundled by default with any distribution I know of.

It's useful sure, but I'm not sure it fits here.

agnivade: Yea I think so too. For the sake of practicality, we allow a command to be used sometimes with other commands. But this case seems to be a bit specific to the user environment.

@natesilva - Let's remove this option.

sbrl:
Thanks, @natesilva! 😺

PR #1641, opened 2017-11-22
pasuspender: add page

sbrl:
----


sbrl:
Good point, @agnivade! I don't think so. I've removed them to simplify the page.

agnivade:
Thanks. I have another doubt though. Should hw:0 be in tokens ? Also, it would be good to explain what it means.

sbrl:
True. I think that the whole command should be in tokens though - as this command basically allows you to pause `pulseaudio` while another command is running.

PR #1642, opened 2017-11-23
 npm: additional module listing use case

raylas:
Added a use case for listing all globally installed modules with a tree depth of 0.

### DIFF
--- a/pages/common/npm.md
+++ b/pages/common/npm.md
@@ -27,6 +27,10 @@
 
 `npm list`
 
+- List a tree of globally installed modules:
+
+`npm list -g --depth=0`
+
 - Interactively create a package.json file:
 
 `npm init`
--- a/pages/common/npm.md
+++ b/pages/common/npm.md
@@ -27,7 +27,7 @@
 
 `npm list`
 
-- List a tree of globally installed modules:
+- List globally installed modules:
 
 `npm list -g --depth=0`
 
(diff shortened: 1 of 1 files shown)


pages/common/npm.md
```diff
@@ -27,6 +27,10 @@
```
agnivade: Please change this description to `List of tree of installed modules referenced in package.json`

pages/common/npm.md
```diff
@@ -27,6 +27,10 @@
 
 `npm list`
 
+- List globally installed modules:
+
+`npm list -g --depth=0`
```
agnivade: --depth={{0}}

pages/common/npm.md
```diff
@@ -27,6 +27,10 @@
 
 `npm list`
 
+- List globally installed modules:
```
agnivade: `List top-level globally installed modules`

pages/common/npm.md
sbrl: Shouldn't `package.json` be in backticks?

agnivade: Backticks are only for commands, not filenames like these. I also checked the yarn.md page. It also refers to package.json simply without any backticks.

sbrl: Ah, ok 😺

sbrl:
Thanks, @raylas 😺

PR #1644, opened 2017-11-23
ps: added examples for current user processes

### LINKED ISSUE #1610: ps: add more CLI options
Need to add more command line options for ps command.

SubOptimal:
Fixes #1610

----

### DIFF
--- a/pages/common/ps.md
+++ b/pages/common/ps.md
@@ -16,8 +16,8 @@
 
 - List all processes of the current user in extra full format:
 
-`ps --user $(id --user) -F`
+`ps --user $(id -u) -F`
 
 - List all processes of the current user as a tree:
 
-`ps --user $(id --user) f`
+`ps --user $(id -u) f`
--- a/pages/common/ps.md
+++ b/pages/common/ps.md
@@ -18,6 +18,6 @@
 
 `ps --user $(id --user) -F`
 
-- List all processes of the current user as ASCII art process hierarchy:
+- List all processes of the current user as a tree:
 
 `ps --user $(id --user) f`
--- a/pages/common/ps.md
+++ b/pages/common/ps.md
@@ -14,10 +14,10 @@
 
 `ps aux | grep {{string}}`
 
-- List all processes of the current user in extra full format
+- List all processes of the current user in extra full format:
 
 `ps --user $(id --user) -F`
 
-- List all processes of the current user as ASCII art process hierarchy
+- List all processes of the current user as ASCII art process hierarchy:
 
 `ps --user $(id --user) f`
--- a/pages/common/ps.md
+++ b/pages/common/ps.md
@@ -13,3 +13,11 @@
 - Search for a process that matches a string:
 
 `ps aux | grep {{string}}`
+
+- List all processes of the current user in extra full format
+
+`ps --user $(id --user) -F`
+
+- List all processes of the current user as ASCII art process hierarchy
+
+`ps --user $(id --user) f`


pages/common/ps.md
```diff
@@ -13,3 +13,11 @@
 - Search for a process that matches a string:
 
 `ps aux | grep {{string}}`
+
+- List all processes of the current user in extra full format:
```
sbrl: What do you mean by "extra full format" here?

pages/common/ps.md
```diff
@@ -13,3 +13,11 @@
...
+
+`ps --user $(id --user) -F`
+
+- List all processes of the current user as ASCII art process hierarchy:
```
sbrl: Perhaps `user as an ASCII art` would read better?

pages/common/ps.md
agnivade: The extra full format is already covered by `auxww`. Let's just keep the next example.

pages/common/ps.md
agnivade: How about changing it to - `List all processes of the current user as a tree:` ? And also, you can change `id --user` to `id -u` for brevity. As that is how its mentioned in the id.md page.

pages/common/ps.md
SubOptimal: The explanation of `-F - extra full format` I took from the `man` page.

It's the same as `--format user,pid,ppid,c,sz,rss,psr,stime,tty,time,cmd`

| format specifier| description |
| --- | --- |
| user | effective user name |
| pid | process ID |
| ppid | parent process ID |
| c | processor utilization |
| sz | size in physical pages of the core image of the process |
| rss | size of the non-swapped physical memory |
| psr | processor that process is currently assigned to |
| stime | time the command started |
| tty | the terminal |
| time | cumulative CPU time |
| cmd | command with all its arguments |

SubOptimal: It's not really the same

using `auxww`

    USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
    suboptim    5503  0.0  0.1  22056  5888 pts/11   Ss   12:28   0:00 bash

using `-F`

    UID        PID  PPID  C    SZ   RSS PSR STIME TTY          TIME CMD
    suboptim  5503  3129  0  5514  5888   0 12:28 pts/11   00:00:00 bash

pages/common/ps.md
SubOptimal: I will update it to `as a tree`.

But will keep `id --user` as this is suggested in https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#user-content-guidelines 
> Try to incorporate the **spelled-out version** of single-letter options in the example's description.

agnivade: No you did not misread. There are several implicit rules so its not practical to mention every one of them. The addendum to this rule is, if the single-letter version is easily understandable, then use that. In this case, `-u` can be easily remembered for `user`. But if it was something like `-m|--user`, then `--user` would be preferred. 

Hope that makes sense.

pages/common/ps.md
agnivade: I see. It is indeed not a perfect match. What I meant was that it's more of an extended format like `auxww`. I wanted to have just one option with an extended format. But if you feel this is a helpful option from a beginner's perspective, we can keep it.

sbrl: Ah, ok. Sure, this works then 😺

pages/common/ps.md
SubOptimal: Ok. Now it's more clear to me.

pages/common/ps.md
SubOptimal: As the usage of `ps` is really divers I would suggest to add one more advanced example for beginner's, which shows how to find the parent process `PPID` easily.
I can't really judge if it is helpfull for most of the beginners or not. Why not leaving it in till someone propose an example maybe more useful for beginner's?

agnivade: Since we are now just at 5 examples, it should not be a big issue.

sbrl:
Thanks for sticking with the review process, @SubOptimal :D

PR #1646, opened 2017-11-23
mediainfo: add page

agnivade:
----

### DIFF
--- a/pages/common/mediainfo.md
+++ b/pages/common/mediainfo.md
@@ -6,7 +6,7 @@
 
 `mediainfo {{file}}`
 
-- Store the output to a given file apart from displaying in the console:
+- Store the output to a given file along with displaying in the console:
 
 `mediainfo --Logfile={{out.txt}} {{file}}`
 
--- a/pages/common/mediainfo.md
+++ b/pages/common/mediainfo.md
@@ -1,0 +1,15 @@
+# mediainfo
+
+> Display metadata from video and audio files.
+
+- Display metadata for a given file in the console:
+
+`mediainfo {{file}}`
+
+- Store the output to a given file apart from displaying in the console:
+
+`mediainfo --Logfile={{out.txt}}  {{file}}`
+
+- Display the list of metadata attributes that can be extracted:
+
+`mediainfo --Info-Parameters`
--- a/pages/common/mediainfo.md
+++ b/pages/common/mediainfo.md
@@ -8,7 +8,7 @@
 
 - Store the output to a given file apart from displaying in the console:
 
-`mediainfo --Logfile={{out.txt}}  {{file}}`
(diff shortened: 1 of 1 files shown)


pages/common/mediainfo.md
```diff
@@ -0,0 +1,15 @@
+# mediainfo
+
+> Display metadata from video and audio files.
+
+- Display metadata for a given file in the console:
+
+`mediainfo {{file}}`
+
+- Store the output to a given file apart from displaying in the console:
```
sbrl: "apart from displaying in the console" feels a bit confusing. Does it display the data into the console too? Or not?

Also, the `--Logfile` bit has a capital letter. Was this intentional, or does it not work without it?

pages/common/mediainfo.md
```diff
@@ -0,0 +1,15 @@
+# mediainfo
+
+> Display metadata from video and audio files.
+
+- Display metadata for a given file in the console:
+
+`mediainfo {{file}}`
```
sbrl: Perhaps `path/to/file.ext` might be more consistent with the style we've used on other pages?

pages/common/mediainfo.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Store the output to a given file apart from displaying in the console:
+
+`mediainfo --Logfile={{out.txt}} {{file}}`
+
+- Display the list of metadata attributes that can be extracted:
```
sbrl: This is useful - it seems to give a list of parameters and then a description for each. Different file types may have different metadata attributes that you can extract I assume though. Is there a way to list them for a specific file type too?

pages/common/mediainfo.md
agnivade: Ah yes, I should change it to "along with displaying in the console".

> Also, the --Logfile bit has a capital letter.

Hmm, it does work with `--logfile`. But `--Logfile` is what it says in its help section. So I would rather be safe.

pages/common/mediainfo.md
agnivade: Ah no. I didn't find anything like that.

sbrl: Shame! Just a thought, anyway.

pages/common/mediainfo.md
sbrl: Ah, ok. Let's stick with the safe option then.

pages/common/mediainfo.md
agnivade: We usually use the `file.ext` syntax when adding the extension makes the command more clear. https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#token-syntax see pt. 5

agnivade:
ping @sbrl

sbrl:
Looks ok to me!

PR #1647, opened 2017-11-24
update tail to indicate -F

plytro:
If I open the file with tail like this:

$ tail -f /var/log/messages
... and if the log rotation facility on my machine decides to rotate that log file while I'm watching messages being written to it ("rotate" means delete or move to another location etc.), the output that I see will just stop.

If I open the file with tail like this:

$ tail -F /var/log/messages
... and again, the file is rotated, the output would continue to flow in my console because tail would reopen the file as soon as it became available again, i.e. when the program(s) writing to the log started writing to the new /var/log/messages.

----

### DIFF
--- a/pages/common/tail.md
+++ b/pages/common/tail.md
@@ -17,3 +17,8 @@
 - Keep reading file until `Ctrl + C`:
 
 `tail -f {{file}}`
+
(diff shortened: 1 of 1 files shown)


sbrl:
Thanks, @plytro! 😺

PR #1651, opened 2017-11-24
chmod: Add recursive 755/644 chmod example

mathomp4:
Might not be approved, but this is the classic 755 on all directories, 644 on all files in a tree a la: https://stackoverflow.com/questions/18817744/change-all-files-and-folders-permissions-of-a-directory-to-644-755

As I just discovered `tldr`, I had to try my hand at updating a page with a command I have to Google every single time.

----

### DIFF
--- a/pages/common/chmod.md
+++ b/pages/common/chmod.md
@@ -22,6 +22,6 @@
 
 `chmod o=g {{file}}`
 
-- Change permissions [-R]ecursively giving [g]roup and [o]thers the abililty to [w]rite:
+- Change permissions recursively giving [g]roup and [o]thers the abililty to [w]rite:
 
 `chmod -R g+w,o+w {{directory}}`
--- a/pages/common/chmod.md
+++ b/pages/common/chmod.md
@@ -22,6 +22,6 @@
 
 `chmod o=g {{file}}`
 
-- Change permissions [-R]ecursively giving the [u]ser [r]ead, [w]rite, and e[X]ecute rights, and give [g]roup and [o]thers [r]ead and e[X]ecute rights but not [-w]rite rights:
+- Change permissions [-R]ecursively giving [g]roup and [o]thers the abililty to [w]rite:
 
-`chmod -R u+rwX,go+rX,go-w {{directory}}`
+`chmod -R g+w,o+w {{directory}}`
--- a/pages/common/chmod.md
+++ b/pages/common/chmod.md
@@ -21,3 +21,7 @@
 - Give [o]thers (not in the file owner's group) the same rights as the group:
 
 `chmod o=g {{file}}`
+
+- Change permissions [-R]ecursively giving the [u]ser [r]ead, [w]rite, and e[X]ecute rights, and give [g]roup and [o]thers [r]ead and e[X]ecute rights but not [-w]rite rights:
+
+`chmod -R u+rwX,go+rX,go-w {{directory}}`


pages/common/chmod.md
```diff
@@ -21,3 +21,7 @@
 - Give [o]thers (not in the file owner's group) the same rights as the group:
 
 `chmod o=g {{file}}`
+
+- Change permissions [-R]ecursively giving the [u]ser [r]ead, [w]rite, and e[X]ecute rights, and give [g]roup and [o]thers [r]ead and e[X]ecute rights but not [-w]rite rights:
```
sbrl: I like what you've you've done here to aid in the understanding of the command! Cool 😃

pages/common/chmod.md
```diff
@@ -21,3 +21,7 @@
...
+
+`chmod -R u+rwX,go+rX,go-w {{directory}}`
```
agnivade: Can we just keep this example to "user" and remove the "group" and "others" ? I fear the description is getting a bit too long. And the "group" and "others" are already explained in earlier examples. Since, this example just introduces the `-R` option, I think the users can build up the `go` options themselves.

And also, I think we can use `x` instead of `X`. The man page does say that X is for `execute/search only if the file is a directory or already has execute permission for some user (X)`, but I did not find any difference with `x` in my tests.

mathomp4: Actually, capital `X` is usually preferred because `x` can often do things in a way you wouldn't suspect:

* https://www.franzoni.eu/chmod-and-the-capital-x/
* https://metashell.net/index.php/2016/05/10/chmod-capital-x-gunna-give-it-to-you/
* https://prasadlinuxblog.wordpress.com/2012/05/09/the-uppercase-x-in-chmod/

Indeed, looking at that last link, this:

`chmod -R u=rwX,g=rX,o=rX {{directory}}`

might be a better command example. With this you are expressly saying `u` gets `rwX` and `g` and `o` get `rX`.

agnivade: I see. My only concern is earlier we use `x` but in this example, we use `X` without really explaining the difference. And if we attempt to explain the difference, the description will become huge.

Therefore, in the spirit of simplicity, I wanted to go with `x`.

I am open to any other suggestions you have.

mathomp4: Hmm. Well, I can just close the request in the sake of non-confusion.

Do you know if there is a way to provide a "custom page path" to tldr? Such that I could add my `pages/common/chmod.md` so my tldr sees it, but there isn't a need to get it in the official tree?

agnivade: We don't need to close it. I would still want the `-R` option to be added.

> Do you know if there is a way to provide a "custom page path" to tldr? 

If you use the node client, yes there is. https://github.com/tldr-pages/tldr-node-client#configuration. See last section. Though its not exactly what you want.

mathomp4: Well, if you want to not use `X` we should avoid it completely. For a recursive use perhaps we could add `chmod -R g+w,o+w`:

```
$ tree -p Dir1/
Dir1/
├── [-rw-r--r--]  file1
└── [-rw-r--r--]  file2

0 directories, 2 files
$ chmod -R g+w,o+w Dir1/
$ tree -p Dir1/
Dir1/
├── [-rw-rw-rw-]  file1
└── [-rw-rw-rw-]  file2

0 directories, 2 files
```
That adds write to both group and other.

agnivade:
@sbrl - Can you take a look at this PR again ? I had some concerns due to which the PR has changed a bit. Please look at my comments here - https://github.com/tldr-pages/tldr/pull/1651#discussion_r153714809. Let me know if you are good with these.

sbrl:
Yeah sure, @agnivade.

It still looks ok - much simpler and easy to understand :D

sbrl:
Thanks, @mathomp4!

PR #1653, opened 2017-11-24
Added eg to similar projects

orschiro:


sbrl:
I don't see why not. Are you the owner of that project?

orschiro:
@sbrl I am not the owner of this project. Just found it while browsing [1] and thought I am going to mention it. :-)

[1] https://www.reddit.com/r/linux/comments/7ez5il/tldr_tar_simplified_and_communitydriven_man_pages/

agnivade:
ping @srsudar.

srsudar:
Sounds good--what do you need from me?

I talk about how I think `eg` relates to these other tools [on this issue](https://github.com/srsudar/eg/issues/4). Basically I think they are all useful. Most importantly, when I first created `eg` I didn't realize there were other tools, but in retrospect of course there were.

I would describe `eg` a little differently. Individual commands appear to have more examples in `eg`, I think, but there are **way** more supported commands in `tldr`, and you guys correctly handle different platforms. `eg` doesn't, but instead mostly combines specific behavior in different sections.

And in general of course `tldr` is much more mature, with much broader support and tool integration, etc.

Aside from supported commands, I think the biggest difference is that `eg` allows you to add custom examples and custom commands. When I poked around last time, I didn't think it was supported in `tldr` but I could have missed it.

I would describe `eg` more like:

[eg](https://github.com/srsudar/eg) provides detailed examples with explanations on the command line. Examples come from the repository, but `eg` allows you to display custom examples and commands alongside the defaults.

Let me know if you need something else!

agnivade:
Just wanted to check with you on the wording in this PR. @orschiro - Can you update the PR with the language that @srsudar has mentioned ?

orschiro:
@agnivade updated the language thanks to @srsudar!

sbrl:
Thanks, @orschiro and @srsudar 😺

PR #1656, opened 2017-11-24
ember: add page

alexlafroscia:


agnivade:
I actually prefer the current description. It is to the point and describes accurately what the tool does. In general, we tend to avoid using subjective words like "ambitious". And mentioning "command line interface for web applications" does not provide much useful information.

pages/common/ember.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Run the development server:
+
+`ember serve`
+
+- Run the test suite:
+
+`ember test`
+
+- Generate files:
```
agnivade: Can you expand this a bit further ? It's not exactly clear what you mean by generate files.

alexlafroscia:
@jimfilippou yeah, that might be a bit more clear. I grabbed the tagline from [the Github repo](https://github.com/ember-cli/ember-cli) and used that.

pages/common/ember.md
alexlafroscia: For sure. I actually just grabbed the headings and examples from the Ember CLI readme, but I can definitely expand it more.

sbrl:
Looks ok to me! Thanks, @alexlafroscia 😺

PR #1657, opened 2017-11-24
ssh: add command for using tar via pipes

jwilkins:
----

### DIFF
--- a/pages/common/ssh.md
+++ b/pages/common/ssh.md
@@ -30,3 +30,7 @@
 - Enable the option to forward the authentication information to the remote machine (see `man ssh_config` for available options):
 
 `ssh -o "ForwardAgent=yes" {{username}}@{{remote_host}}`
+
+- Copy files (preserving links, permissions, etc) between hosts using pipes and tar
+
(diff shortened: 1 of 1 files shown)


agnivade:
Hi @jwilkins, thanks for the contribution.

However, I think this is an advanced usage of ssh. Our scope here is very limited, and we try to keep the examples from a beginner's perspective. If we mention all possible examples, this will become the man page itself. Keeping that in mind, I think this should not be added to the page.

Let me know your thoughts.

sbrl:
I'm inclined to agree with @agnivade here, unfortunately - though I do believe that this command is rather cool - and while it's a shame to not include it, it does feel like an advanced usage.

However, we haven't yet got an example that executes a command on a remote host. To that end, perhaps this could be simplified to just `cat` a file from a remote host? Advanced users should be able to figure out that you can use `tar` etc. intead to send whole directories I think.

agnivade:
> However, we haven't yet got an example that executes a command on a remote host. 

We do. It's the 4th example. :smile:

agnivade:
Hey @jwilkins - Let me know if you agree with my suggestion to re-purpose this PR to update the remote command example to a concrete example.

PR #1658, opened 2017-11-25
smartctl: add page

sommd:
----

### DIFF
--- a/pages/common/smartctl.md
+++ b/pages/common/smartctl.md
@@ -1,23 +1,24 @@
 # smartctl
 
-> View a disk's [SMART](https://en.wikipedia.org/wiki/S.M.A.R.T.) data and other information.
+> View a disk's SMART data and other information.
+> See https://en.wikipedia.org/wiki/S.M.A.R.T for more information.
 
 - View SMART health summary:
 
-`sudo smartctl --health {{/dev/<device>}}`
+`sudo smartctl --health {{/dev/sda}}`
 
 - View device information:
 
-`sudo smartctl --info {{/dev/<device>}}`
+`sudo smartctl --info {{/dev/sda}}`
 
-- Begin a (short) self-test:
+- Begin a short self-test:
 
-`sudo smartctl --test {{short}} {{/dev/<device>}}`
+`sudo smartctl --test short {{/dev/sda}}`
 
 - View current/last self-test status and other SMART capabilities:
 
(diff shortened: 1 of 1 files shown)


pages/common/smartctl.md
```diff
@@ -0,0 +1,23 @@
...
+
+- View SMART health summary:
+
+`sudo smartctl --health {{/dev/<device>}}`
+
+- View device information:
+
+`sudo smartctl --info {{/dev/<device>}}`
+
+- Begin a (short) self-test:
```
agnivade: You can take short out of brackets and remove the tokens around "short" in the example. That way, the example directly matches the description.

pages/common/smartctl.md
```diff
@@ -0,0 +1,23 @@
+# smartctl
+
+> View a disk's [SMART](https://en.wikipedia.org/wiki/S.M.A.R.T.) data and other information.
```
agnivade: This is a console tool. So `[xyz](link)` will be displayed as is. I suggest you keep the wikipedia link in a second line.

pages/common/smartctl.md
```diff
@@ -0,0 +1,23 @@
...
+
+- View SMART health summary:
+
+`sudo smartctl --health {{/dev/<device>}}`
```
agnivade: For filesystems, we usually give a concrete example- like `/dev/sda1`. It makes it more clear. Please do that for all examples.

agnivade:
Thanks for the PR ! This is a very important command. I gave some minor comments that needs some work.

pages/common/smartctl.md
```diff
@@ -0,0 +1,24 @@
...
+
+- View SMART health summary:
+
+`sudo smartctl --health {{/dev/sda}}`
+
+- View device information:
+
+`sudo smartctl --info {{/dev/sda}}`
+
+- Begin a short self-test:
```
sbrl: So this doesn't actually output the result of the short self-test?

sommd: Self tests run in the background, this command exits after starting the test.

There are tests that run in the foreground, but that pauses all other disk operations, which isn't a good idea if testing a mounted drive.

Check the results is shown in the next commands.

sbrl: Ah, ok. That's ok then - just wanted to check :-)

sbrl:
Thanks, @sommd 😺

PR #1675, opened 2017-11-26
git: Added examples for standard workflow

Noah-Huppert:
----

- ~~[ ] The page (if new), does not already exist in the repo.~~

- ~~[ ] The page (if new), has been added to the correct platform folder:  

I modified the git tldr page to include some more commonly used examples. I think it now covers most of my day to day git usage.

### DIFF
--- a/pages/common/git.md
+++ b/pages/common/git.md
@@ -2,18 +2,22 @@
 
 > Distributed version control system.
 
-- Check the Git version:
-
-`git --version`
-
-- Call general help:
+- Show help
 
 `git --help`
 
-- Call help on a command:
+- Show file changes
 
-`git help {{command}}`
+`git status`
 
-- Execute Git command:
+- Track file changes
 
-`git {{command}}`
+`git add {{path}}`
+
+- Commit file changes
+
+`git commit -m "{{Message}}"`
+
+- Upload changes to remote server
+
+`git push {{remote name}} {{branch}}`
--- a/pages/common/git.md
+++ b/pages/common/git.md
@@ -2,22 +2,22 @@
 
 > Distributed version control system.
 
-- Show help
+- Show help:
 
 `git --help`
 
-- Show file changes
+- Show file changes:
 
 `git status`
 
-- Track file changes
+- Track file changes:
 
 `git add {{path}}`
 
-- Commit file changes
+- Commit file changes:
 
 `git commit -m "{{Message}}"`
 
-- Upload changes to remote server
+- Upload changes to remote server:
 
 `git push {{remote name}} {{branch}}`
--- a/pages/common/git.md
+++ b/pages/common/git.md
@@ -12,7 +12,7 @@
 
 - Commit file changes:
 
-`git commit -m "{{Message}}"`
+`git commit -m "{{message}}"`
 
 - Upload changes to remote server:
 
--- a/pages/common/git.md
+++ b/pages/common/git.md
@@ -2,10 +2,6 @@
 
 > Distributed version control system.
 
-- Show help:
-
-`git --help`
-
 - Show file changes:
 
 `git status`
@@ -20,4 +16,12 @@
 
 - Upload changes to remote server:
 
-`git push {{remote name}} {{branch}}`
+`git push {{server name}} {{branch}}`
+
+- Switch to a branch:
+
+`git checkout {{branch}}`
+
+- Create a new branch:
+
+`git checkout -b {{branch}}`
--- a/pages/common/git.md
+++ b/pages/common/git.md
@@ -16,7 +16,7 @@
 
 - Upload changes to remote server:
 
-`git push {{server name}} {{branch}}`
+`git push {{remote repository}} {{branch}}`
 
 - Switch to a branch:
 
--- a/pages/common/git.md
+++ b/pages/common/git.md
@@ -14,7 +14,7 @@
 
 `git commit -m "{{message}}"`
 
-- Upload changes to remote server:
+- Upload changes to remote repository:
 
 `git push {{remote repository}} {{branch}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/git.md
```diff
@@ -2,18 +2,26 @@
...
+- Commit file changes:
 
-`git help {{command}}`
+`git commit -m "{{message}}"`
 
-- Execute Git command:
+- Upload changes to remote server:
 
-`git {{command}}`
+`git push {{server name}} {{branch}}`
```
JacobCallahan: I'd recommend changing "server name" to "repository name" for more clarity. "remote repository" could work as well.

JacobCallahan:
These changes are much more useful than the version they replace!

pages/common/git.md
Noah-Huppert: I'll change it to "remote repository" to introduce a dash of the official Git terminalolgy.

agnivade:
Hi @Noah-Huppert - We have separate dedicated pages for git subcommands. You can do `tldr git status`, `tldr git commit` for example. This page is only for the `git` main command, not subcommands

JacobCallahan:
If someone new doesn't know what subcommands exist within git (or other utilities) they wouldn't necessarily know what subcommands even exist to tldr. Perhaps this could be a wider area for improvement (e.g. adding a "see also: git pull, git push, git rebase, etc..") at the bottom of the tldr output. This could be spread the all high level commands that have subcommand pages.

agnivade:
Yes, we have given some thought on that. There is an open issue which tracks exactly that - https://github.com/tldr-pages/tldr/issues/481. It's up to the clients to implement that and would also need a change in the markdown format. We just haven't been able to get to it.

dospunk:
Perhaps some text at the bottom of the git tldr page explaining how info on how subcommands can be found would be helpful?

agnivade:
#481 deals exactly with that. Like I mentioned above, we need to decide on the markdown format and then notify all the client maintainers so that they can implement the change. It's not trivial. Hence been pending for some time.

bangbambang:
can't we just add `git help tutorial | everyday | workflows` or `git help -g`

Those would help to explains the available subcommands and still under the 8 examples limit.

agnivade:
We can do a `git help everyday`. 

@waldyrious - thoughts on this ? It seems a lot of people do just `tldr git`, expecting to list all the subcommands. And its not obvious that `tldr git <subcommand>` is also available.

waldyrious:
The right solution for the subcommands is clearly the "see also" section, aka #481. We'll have to make do with alternative solutions until that's available. I see two ways to mitigate this problem (and we can do both):

1. Expanding the command description with a second line mentioning the existence of subcommand pages (similar to how the [uname page](https://github.com/tldr-pages/tldr/blob/master/pages/linux/uname.md) mentions the `lsb_release` one)
2. Including `git help --guides` in this page and an example of how to invoke one of them (`git help everyday` sounds like the most appropriate one). That was a great call, @bangbambang!

So my concrete suggestion for this PR is:

- Keep the original page contents
- Add a second line to the main description mentioning the existence of subcommand tldr pages
- Swap the position of the first and last examples (`git --version` and `git {{command}}`)
- Add `git help --guides` and `git help everyday` after the `git help {{command}}` example.

Does that sound reasonable?

agnivade:
> Add a second line to the main description mentioning the existence of subcommand tldr pages

Yea I have already tried this. The node client does not detect multi word commands currently. :cry:  Need to fix that.

> Add git help --guides and git help everyday after the git help {{command}} example.

I think `git help everyday` should be sufficient no ? Any reason you want `--guides` too ?

waldyrious:
> The node client does not detect multi word commands currently.

That doesn't sound related to this. Do you mean that we can't currently do `tldr git commit` using the node client?

> Any reason you want `--guides` too ?

Because it contextualizes `git help everyday` as one of several guides available, and lists the other guides (`glossary`, `tutorial` and `workflows`, in particular, would probably be useful for the tldr audience)

agnivade:
> That doesn't sound related to this. 

Sorry, I misunderstood maybe. For the `uname` page, if you do `tldr uname`, you will get a `See also: lsb_release` at the bottom of the command output. I thought you wanted something like that here too. I think you just want to mention that commands like `git status` and `git commit` exist, right ?

> Because it contextualizes git help everyday as one of several guides available,

Hmm .. okay. I think you will be needed to touch up the language a bit once this PR is updated :stuck_out_tongue_closed_eyes: 

@Noah-Huppert - Can you re-purpose this PR with the suggestions made by @waldyrious above ?

waldyrious:
> For the `uname` page, if you do `tldr uname`, you will get a `See also: lsb_release` at the bottom of the command output.

Yeah, that sounds like a client-specific feature. I was talking about the contents of the page itself. Literally adding a second line to the main description (which currently only says "Distributed version control system"), with a note just like the one in the uname page. I think doing that might go quite a long way towards solving the main problem this PR attempts to tackle :)

Noah-Huppert:
Any final say on this? My opinion is that we should just merge this in as in. Because it makes the git help just a bit easier to use.  

Maybe some of the issues mentioned above could be continued in other threads?

agnivade:
Yes, please look at my last comment. Specifically, we want this PR to have the changes that @waldyrious suggested here - https://github.com/tldr-pages/tldr/pull/1675#issuecomment-347468097.

Noah-Huppert:
Finals rn, but I have the list now:

- [ ] Add note about tldr also supporting git subcommands
- [ ] Add a git subcommand example, and make it first
- [ ] Add a `git help` example
    - mention the `--guides` flag 
- [ ] Add a git guide example
    - `git help everyday`

PR #1683, opened 2017-11-27
sudo: Add example to run an interactive shell

ChaoticLaw:
This adds the example `sudo -i`, which allows the user to enter an interactive shell with sudo.

Running sudo version 1.8.21p2 on Arch Linux, there is also a spelled flag `--login`, but this doesn't work on my macOS 10.11.6 system (sudo 1.7.10p9), hence using the syntax above.

----

- [ ] ~~The page (if new), has been added to the correct platform folder:~~
    ~~`common/` if it's common to all platforms, `linux/` if it's Linux-specific, and so on.~~

### DIFF
--- a/pages/common/sudo.md
+++ b/pages/common/sudo.md
@@ -17,7 +17,3 @@
 - Repeat the last command as sudo:
 
 `sudo !!`
-
-- Run an interactive shell with superuser privileges
-
-`sudo -i`
--- a/pages/common/sudo.md
+++ b/pages/common/sudo.md
@@ -18,6 +18,6 @@
 
 `sudo !!`
 
-- Run an interactive shell as sudo:
+- Launch a shell with root privileges:
 
 `sudo -i`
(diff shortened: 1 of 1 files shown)


agnivade:
Can you explain what does "interactive shell" mean ? Does it mean all sudo commands run in interactive mode ?

ChaoticLaw:
For this example, interactive shell means the launching of a shell process as the root user, meaning that any subsequent commands will be run as root instead of the current user, until they exit the shell.

Straying a bit from this specific example, the flag `-i` in conjunction with `-u {{user}}` will run the specified user's set shell instead—so if user foo has a shell of `/bin/bash`, and user bar has a shell of `/bin/fish`, when user foo executes `sudo -u bar -i`, they will run `/bin/fish` as user bar, and any commands issued will be run as user bar until they exit.

Sorry about not being concise; I'm a bit new to this.

agnivade:
Ah I see. Then it may be better to change the description to - `Launch a shell with root privileges:`

sbrl:
What about `Launch the default shell with root privileges`?

sbrl:
Thanks, @ChaoticLaw!

PR #1686, opened 2017-11-27
apachectl: add page

basiclines:


agnivade:
A page should be in the platform specific folder. In this case, it should be inside the osx folder. This instruction was explicitly mentioned in the PR template.

Please move this page under the osx folder.

basiclines:
@agnivade Just moved the page to the osx folder 😅

pages/osx/apachectl.md
```diff
@@ -0,0 +1,15 @@
+# apachectl
+
+> Apache HTTP Server Control Interface for OSX.
```
sbrl: I'm not sure that `Control Interface` needs to be in title case. Perhaps lowercase would read better here?

basiclines: Fixed!

pages/osx/apachectl.md
```diff
@@ -0,0 +1,15 @@
+# apachectl
+
+> Apache HTTP Server control interface for macOS.
+
+- Load the org.apache.httpd launchd job:
```
agnivade: I am a bit curious about why the words "Load" and "Unload" are used and not "start" and "stop".

pages/osx/apachectl.md
```diff
@@ -0,0 +1,15 @@
...
+
+`apachectl start`
+
+- Unload the org.apache.httpd launchd job:
+
+`apachectl stop`
+
+- Unload, then loads the org.apache.httpd launchd job:
```
agnivade: You don't have to mention "org.apache.httpd" after the 2nd time. It is implicit.

pages/osx/apachectl.md
basiclines: That's from the official `man` page but it's true that it doesn't fit the needs of this repo.

agnivade: Yes, we try to avoid copying text verbatim from the man page. And try to change it to something simple and clear.

PR #1691, opened 2017-11-27
dd: state that /dev/urandom is not secure

Peyphour:
Using /dev/urandom is not recommended to securely generate random bytes.

----

### DIFF
--- a/pages/linux/dd.md
+++ b/pages/linux/dd.md
@@ -10,7 +10,7 @@
 
 `dd if=/dev/{{source_drive}} of=/dev/{{dest_drive}} bs=4M conv=noerror status=progress`
 
-- Generate a file of 100 random bytes by using kernel random driver:
+- Generate a file of 100 (non-secure) random bytes by using kernel random driver:
 
 `dd if=/dev/urandom of={{random_file}} bs=100 count=1`
 


agnivade:
For usages in commands like `dd`, `/dev/urandom` is the preferred way I think. Using `/dev/random` might block if the entropy pool is running low, which might be a hassle. Only in the case of a freshly booted machine with low entropy, it might pose a security risk. But that is a corner case.

For all practical purposes though, `/dev/urandom` will generate good quality random numbers. It is indeed possible that the attacker might exhaust the entropy pool, after which it will fall back to a pseudorandom number generator as mentioned in the man page - 
```
If there is not  sufficient  entropy,  a  pseudorandom
       number  generator is used to create the requested bytes.  As a result, in this case the returned values are theoretically vulnerable
       to a cryptographic attack on the algorithms used by the driver.  Knowledge of how to do this is not available in the current unclas‐
       sified  literature,  but  it is theoretically possible that such an attack may exist.
```
There is a very good discussion here which points to some flaws of this argument - https://security.stackexchange.com/a/3939

Did you have something else in mind to use other than `/dev/random` or `/dev/urandom` ?

Would love to hear your thoughts.

sbrl:
Interesting discussions here. I'm not actually sure how one uses `/dev/random`, actually - it appears to block for lengthy periods of time on all the Linux machines I've tested it on any time I try to get more than a few bytes out of it.

PR #1693, opened 2017-11-27
wget: Add subdirectory example

a-raccoon:
Added `wget -m -np`

----

### DIFF
--- a/pages/common/wget.md
+++ b/pages/common/wget.md
@@ -7,10 +7,6 @@
 
 `wget {{https://example.com/foo}}`
 
-- Mirror all listed files within a directory and its sub-directories only (no parent):
-
-`wget -m -np {{https://example.com/somepath/}}`
-
 - Download a single web page and all its resources (scripts, stylesheets, images, etc.):
 
 `wget --page-requisites --convert-links {{https://example.com/somepage.html}}`
@@ -18,6 +14,10 @@
 - Download a full website, with 3-second intervals between requests:
 
 `wget --mirror --page-requisites --convert-links --wait=3 {{https://example.com}}`
+
+- Download all listed files within a directory and its sub-directories (does not download embedded page elements):
+
+`wget --mirror --no-parent {{https://example.com/somepath/}}`
 
 - Download the contents of an URL via authenticated FTP:
 
--- a/pages/common/wget.md
+++ b/pages/common/wget.md
@@ -7,6 +7,10 @@
 
 `wget {{https://example.com/foo}}`
 
+- Mirror all listed files within a directory and its sub-directories only (no parent):
+
+`wget -m -np {{https://example.com/somepath/}}`
+
 - Download a single web page and all its resources (scripts, stylesheets, images, etc.):
 
 `wget --page-requisites --convert-links {{https://example.com/somepage.html}}`


pages/common/wget.md
```diff
@@ -7,6 +7,10 @@
 
 `wget {{https://example.com/foo}}`
 
+- Mirror all listed files within a directory and its sub-directories only (no parent):
```
sbrl: Perhaps `sub-directories only, without mirroring any parent directories` would read better?

agnivade: Please move this below the "Download a full website" example.

And also, the word "mirror" can be a bit confusing to newbies. Let's use the same wording as the "Download a full website" example. And update the description to use the directory name "foo". Something like - `Download only foo and its sub-directories without its parent directories.`

pages/common/wget.md
```diff
@@ -7,6 +7,10 @@
...
+
+`wget -m -np {{https://example.com/somepath/}}`
```
agnivade: Let's use the expanded version of the options to keep it consistent with the rest of the pages. And change the subdirectory to foo to keep it consistent.

agnivade:
Sorry, didn't mean to approve it. Github doesn't allow editing a review. :cry:

pages/common/wget.md
a-raccoon: agnivade: Let's use the contracted version of the options here, and on the rest of the pages, because it's TL;DR.  I'm not digging the wrapped lines of 20 character length switches.

pages/common/wget.md
a-raccoon: I used the word "no parent" to emphasize the meaning of -np without actually having to expand to the TL;DR `--no-parent` switch name.

sbrl:
@agnivade Strange. I've always been able to edit reviews before ❓

pages/common/wget.md
agnivade: Our first aim is to be clear in what we write. We can use contracted versions but not at the cost of clarity. Upon reading `-np` it's not immediately clear what it means. Hence, my suggestion. 

Using expanded versions when required is also one of our CONTRIBUTING guidelines - https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines (pt 3). Mentioning it here again for reference - 

`
Try to incorporate the spelled-out version of single-letter options in the example's description. The goal is to allow people to understand the syntax of the commands, not just memorize it.
`

pages/common/wget.md
agnivade: @a-raccoon - Our aim here in tldr is to let people understand the command, rather than memorize the syntax. Please see my comment [here](https://github.com/tldr-pages/tldr/pull/1693#discussion_r153699659)

pages/common/wget.md
a-raccoon: It's why I used the word "no parent" in my example.  My main objection is that wget is one of those few projects that supports long switch names, and it seems strange to me that TooLong;Didn'tRead makes it its mission to use the too-long switch names, at the expense of brevity and consistency with other programs like `tar`.  Maybe I should rip all the long switch names out of wget just so they cannot be used here.

But I digress.  Its your project.  I think usage of long switch names is silly, nay, garbage.

sbrl:
Is this PR waiting an an action from me?

pages/common/wget.md
agnivade: Our first and foremost mission is to make our pages beginner friendly. Yes, the description has "no parent". But IMO, its not immediately clear what it means. More so, because its at the end in brackets without any context. At least to me, my mind asks the question - no parent for what ? It takes me some time to realize parent and child directories are being talked about.

This is what I would recommend -

```
Download only foo and its sub-directories without its parent directories.

wget --mirror --no-parent {{https://example.com/somepath/}}
```

Let me know your thoughts. Also, please move this below the "Download a full website" example.

> But I digress. Its your project. 

It's not. :smile:  It's an open-source project where everybody's opinion counts. We just try to follow the spirit of the project in the pages. Yes, in case of a dispute, unfortunately we go with the popular opinion. But not before we've have a proper discussion about it.

a-raccoon: Mirror all listed files within a directory and its sub-directories only (no parent)
  Download only foo and its sub-directories without its parent directories.

I chose my wording carefully with these considerations in mind: `-m` doesn't download embedded page elements (like images) only href links.  It is used mostly for mirroring directory file listings; pages of linked files -- I avoided the word `download` because directory structure is preserved.  I used the words `sub-directories only (no parent)` to further emphasize the style of mirrored download, and also educate the meaning of `-np`.  Foo in this context could be a little confusing, because it doesn't work on anything else but pages of href links.  Specify anything else, and only one file will be downloaded; that file.

agnivade: I see. Your description makes sense to me. Only issue is that we have an example with `--mirror` and that does not mention "mirror", but it says "download". That has `--convert-links` though. I just want descriptions to be consistent. Is there a way we can get that ?

sbrl:
Ping @agnivade?

agnivade:
No, it's waiting on @a-raccoon

a-raccoon:
I am without further opinion or interest.

agnivade:
@a-raccoon - Sorry about that. Our review process can be a bit tedious at times. I will update the PR myself.

agnivade:
@sbrl , @a-raccoon - can you check if this looks okay to you guys ? I tried to incorporate the original description as much as possible while keeping it consistent with the rest of the page.

agnivade:
Merging this now. @a-raccoon - Feel free to send another PR if you would like something to be modified.

PR #1701, opened 2017-11-28
eval: add page

Neph0:
----

### DIFF
--- a/pages/linux/eval.md
+++ b/pages/linux/eval.md
@@ -2,7 +2,7 @@
 
 > Execute arguments as a single command in the current shell and return its result.
 
-- Call `echo` with the `foo` argument:
+- Call `echo` with the "foo" argument:
 
 `eval "{{echo}}" "{{foo}}"`
 
--- a/pages/linux/eval.md
+++ b/pages/linux/eval.md
@@ -1,0 +1,11 @@
+# eval
+
+> Execute arguments as a single command in the current shell and return its result.
+
+- Call `echo` with the `foo` argument:
+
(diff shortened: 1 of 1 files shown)


Neph0:
@sbrl : This is to highlight a difference between `eval` and `exec`. To be honest, this built-in doesn't do much, but I suppose it should be in here for the completeness of `tldr` 😄

pages/linux/eval.md
```diff
@@ -0,0 +1,11 @@
+# eval
+
+> Execute arguments as a single command in the current shell and return its result.
+
+- Call `echo` with the `foo` argument:
```
agnivade: Put foo inside double quotes and remove backticks. It is mainly used to denote commands.

pages/linux/eval.md
```diff
@@ -0,0 +1,11 @@
...
+
+`eval "{{echo}}" "{{foo}}"`
```
agnivade: I think `eval {{"echo foo"}}` might be simpler.

Neph0: The whole point of `eval` is to create commands from separate arguments.

agnivade: Practically I don't think it matters much. The point is to be able to construct the string. If the value "foo" comes from a variable, then you can either do `eval "echo" $foo` or `eval "echo $foo"`. Both will give the same result. Conceptually, it might be simpler to imagine the whole command within quotes and that `eval` is evaluating the entire string. Hence my suggestion.

Let me know your thoughts.

PR #1705, opened 2017-11-28
unzip: include example of extracting to console.

WrathZA:
Unzipping a file to a console allows for easy grepping of the contents.

----

- [ ] ~~The page (if new), has been added to the correct platform folder:~~
~~`common/` if it's common to all platforms, `linux/` if it's Linux-specific, and so on.~~

- [x] The page has 8 or fewer examples

### DIFF
--- a/pages/common/unzip.md
+++ b/pages/common/unzip.md
@@ -14,6 +14,6 @@
 
 `unzip -l {{file.zip}}`
 
-- Display (pipe) the contents of the file(s) inside a zip file to the console:
+- Extract the contents of the file(s) to stdout alongside the extracted file names
 
 `unzip -c {{file.zip}}`
--- a/pages/common/unzip.md
+++ b/pages/common/unzip.md
@@ -12,8 +12,8 @@
 
 - List the contents of a zip file without extracting:
 
-`unzip -l {{file}}`
+`unzip -l {{file.zip}}`
 
-- Extract the contents of a zip file to the console:
+- Display (pipe) the contents of the file(s) inside a zip file to the console:
 
-`unzip -c {{file}}`
+`unzip -c {{file.zip}}`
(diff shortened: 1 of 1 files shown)


pages/common/unzip.md
```diff
@@ -13,3 +13,7 @@
 - List the contents of a zip file without extracting:
 
 `unzip -l {{file}}`
+
+- Extract the contents of a zip file to the console:
```
agnivade: The wording can be improved a bit. Contents of a zip file is the unzipped file. We need to clarify that it will display the contents of the unzipped file, not the contents of the zip file, in which case it becomes the same as `-l`.

How about - `Display the contents of the file(s) inside a zip file to the console`

pages/common/unzip.md
```diff
@@ -13,3 +13,7 @@
...
+
+`unzip -c {{file}}`
```
agnivade: While we are at it, please rename {{file}} to {{file.zip}} everywhere for clarity.

agnivade:
ping @ericbn for taking a look.

ericbn:
I agree with @agnivade's comments. I suggest something like - `Extract file to stdout alongside the extracted file names`.

I'm not sure if I would use `-c` or `-p` for the example (`-p` does not print the extracted file names).

agnivade:
I think `-c` is better. I just tested with some binary files. Its practically impossible to tell when one file has ended and one has started without the file name.

pages/common/unzip.md
agnivade: I had thought about that. This page does use `compressed.zip`, but only in the extract case where you need to mention 2 files - compressed_file and extracted_file. If 2 files are not needed, then I think for simplicity's sake, just mentioning `file.zip` is good enough.

sbrl:
Looks ok to me!

PR #1709, opened 2017-11-29
nm: add example of demangling C++ symbols

ry:


pages/osx/nm.md
```diff
@@ -13,3 +13,7 @@
 - List all symbols, even debugging symbols:
 
 `nm -a {{file.o}}`
+
+- Demangle C++ symbols:
```
agnivade: Can we do without the `-p` and `-i` ? If they are required / recommended, please explain what they do in the description.

ry: They can be removed and I've done that in a new commit.

sbrl:
Thanks, @ry!

waldyrious:
See #538, where this page was introduced and the demangling behavior was discussed.

PR #1713, opened 2017-11-29
gradle: adds additional gradle examples

bavaga:
----

### DIFF
--- a/pages/common/gradle.md
+++ b/pages/common/gradle.md
@@ -1,10 +1,18 @@
 # gradle
 
-> Gradle is the official build system for Android Studio.
+> Gradle is an advanced open source build automation system.
 
 - Compile a package:
 
 `gradle build`
+
+- Exclude test task:
+
+`gradle build -x test`
+
+- Run in offline mode:
+
+`gradle build --offline`
 
 - Clear the build folder:
 
@@ -13,3 +21,7 @@
 - Compile and Release package:
 
 `gradle assembleRelease`
+
+- Create Gradle wrapper:
+
+`gradle wrapper --gradle-version {{gradle_version}}`
--- a/pages/common/gradle.md
+++ b/pages/common/gradle.md
@@ -8,9 +8,9 @@
 
 - Exclude test task:
 
-`gradle build -x test`
+`gradle build -x {{test}}`
 
-- Run in offline mode:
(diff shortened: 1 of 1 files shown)


pages/common/gradle.md
```diff
@@ -1,15 +1,27 @@
...
 
 - Compile a package:
 
 `gradle build`
 
+- Exclude test task:
+
+`gradle build -x test`
+
+- Run in offline mode:
```
agnivade: Can you expand the description a bit to explain what this offline mode is ? It is not immediately clear to me what it does.

pages/common/gradle.md
```diff
@@ -1,15 +1,27 @@
...
-> Gradle is the official build system for Android Studio.
+> Gradle is an advanced open source build automation system.
 
 - Compile a package:
 
 `gradle build`
 
+- Exclude test task:
+
+`gradle build -x test`
```
agnivade: Shouldn't `test` be inside tokens (`{{}}`) ?

pages/common/gradle.md
```diff
@@ -1,15 +1,27 @@
...
+
 - Clear the build folder:
 
 `gradle clean`
 
 - Compile and Release package:
 
 `gradle assembleRelease`
+
+- Create Gradle wrapper:
```
agnivade: We have to also explain the options used in the command. Either change the description to `Create gradle wrapper with a given gradle version`, or change the example to exclude specifying the gradle-version. 

Remember to always focus on the most common and practical examples. If specifying the gradle-version is not a common thing, then we can drop it.

agnivade:
I see that you had closed a previous PR and opened a new one. That is not necessary. You can update the same PR again by pushing new commits.

pages/common/gradle.md
bavaga: Thanks. Done.

pages/common/gradle.md
bavaga: 👍

pages/common/gradle.md
bavaga: I've removed that example. It seems not to be a very common use case.

bavaga:
Thank you for the feedback. I integerated it in the new commit.

sbrl:
Looks ok to me! Thanks, @bavaga 😺

PR #1714, opened 2017-11-29
mix: add page

### LINKED ISSUE #1667: page request: elixir
mix commands

alxers:
----

Fix #1667

### DIFF
--- a/pages/common/mix.md
+++ b/pages/common/mix.md
@@ -2,11 +2,7 @@
 
 > Mix is a build tool that provides tasks for creating, compiling, and testing Elixir projects, managing its dependencies, and more.
 
-- List all mix commands:
-
-`mix help`
-
-- Execute particular file:
+- Execute a particular file:
 
 `mix run {{my_script.exs}}`
 
@@ -18,10 +14,10 @@
 
 `mix compile`
 
-- Start an iex session inside the project:
-
-`iex -S mix`
-
 - Run project tests:
 
 `mix test`
+
+- List all mix commands:
+
+`mix help`
--- a/pages/common/mix.md
+++ b/pages/common/mix.md
@@ -24,4 +24,4 @@
 
 - Run project tests:
 
- `mix test`
+`mix test`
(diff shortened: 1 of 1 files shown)


sbrl:
Hey - thanks for the new page, @alxers! Looks like you're having some issues with the linter. We've got the style we prefer documented over [here](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#markdown-format), and a few good examples are the pages for `tar`, `eyeD3`, and `ln`. If you need any assistance, just ask :smiley_cat:

pages/common/mix.md
```diff
@@ -0,0 +1,27 @@
+# mix
+
+> Mix is a build tool that provides tasks for creating, compiling, and testing Elixir projects, managing its dependencies, and more.
+
+- List all mix commands:
+
+`mix help`
+
+- Execute particular file:
```
sbrl: Perhaps `Execute a particular file` would read better here?

pages/common/mix.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Create a new project:
+
+`mix new {{project_name}}`
+
+- Compile project:
+
+`mix compile`
+
+- Start an iex session inside the project:
```
sbrl: This looks like a different command. Perhaps it should have its own page? We can always add a second line to the description saying `See also`.

pages/common/mix.md
```diff
@@ -0,0 +1,27 @@
+# mix
+
+> Mix is a build tool that provides tasks for creating, compiling, and testing Elixir projects, managing its dependencies, and more.
+
+- List all mix commands:
```
agnivade: This should be the last example. Think of the natural order of commands that a user is expected to use. First, a user will want to run a file and see if it works or not. Then move on to creating projects and compiling them. And lastly, they would want to see all commands.

pages/common/mix.md
agnivade: On second thoughts, if this is a very integral part of the workflow. We can keep it and put iex within backticks. This would be something like the locate.md page, where the `updatedb` command is in the same page because it is synonymous with `locate`. But only if it is used very regularly along with `mix`, otherwise this should move to a separate page.

sbrl:
Thanks, @alxers! 😺

alxers:
😉

PR #1715, opened 2017-11-29
Improvements for docker 

taygius:
Added examples for "docker images", "docker stats" and "docker inspect". 
Updated "Run a command inside of an already running container": added parameter -i(--interactive) for making it more useful


agnivade:
This exceeds our maximum allowed examples for a page, which is 8. It is explicitly mentioned in the PR template and in CONTRIBUTING guidelines.

PR #1723, opened 2017-11-30
csslint: add page

### LINKED ISSUE #1598: page request: csslint
webpage here - http://csslint.net/
cli documentation here - https://github.com/CSSLint/csslint/wiki/Command-line-interface

Please go through the [CONTRIBUTING](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) page to understand what is the format we use for our pages.

andrewda:
----

Closes #1598

### DIFF
--- a/pages/common/csslint.md
+++ b/pages/common/csslint.md
@@ -16,12 +16,12 @@
 
 - Specify certain rules as errors (which result in a non-zero exit code):
 
-`csslint --errors={{rule1}},{{rule2}},{{rule3}} {{file.css}}`
+`csslint --errors={{errors,universal-selector,imports,etc.}} {{file.css}}`
 
 - Specify certain rules as warnings:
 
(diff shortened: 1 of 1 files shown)


pages/common/csslint.md
```diff
@@ -0,0 +1,27 @@
...
+
+`csslint {{file1.css}} {{file2.css}} {{file3.css}}`
+
+- List all possible style rules:
+
+`csslint --list-rules`
+
+- Specify certain rules as errors (which result in a non-zero exit code):
+
+`csslint --errors={{rule1}},{{rule2}},{{rule3}} {{file.css}}`
```
agnivade: I would prefer to mention some concrete rules here. Something like `csslint --errors={{floats,important}} {{file.css}}`. And same for the next 2 examples.

agnivade:
@andrewda  - I have removed the "etc." from the tokens. Keep in mind that the tokens are supposed to be user-specified values which should _work_ if used as is in the command.

PR #1724, opened 2017-12-01
vim: add new examples for opening vim and refer to vimtutor

frazjp65:
Add a way to start vim cleanly and a way to start in read-only mode. Remove information on how to use vim and add a link to vimtutor.

----

### DIFF
--- a/pages/common/vim.md
+++ b/pages/common/vim.md
@@ -1,36 +1,16 @@
 # vim
 
 > Vi IMproved, a programmer's text editor, providing several modes for different kinds of text manipulation.
-> Pressing `i` enters edit mode. `<Esc>` goes back to normal mode, which doesn't allow regular text insertion.
+> For more information on how to use vim, see [vimtutor](vimtutor.md).
 
 - Open a file:
 
 `vim {{file}}`
 
-- Enter text editing mode (insert mode):
+- Open a file in read-only mode:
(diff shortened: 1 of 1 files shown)


frazjp65:
I removed the link to vimtutor and instead referred to it. Apparently they do not display properly from the command line.

agnivade:
Hi @frazjp65, thanks for contributing.

I would rather keep the examples on how to use vim in this page. The prime reason being usefulness and practicality. Almost, everyone who use vim will need to use these basic search and edit commands. Otherwise, simply mentioning the `vim` command and its options is not much useful. You can see a similar thing being done for the less.md page.

However, this does not apply everywhere though. For eg. we do not mention console specific commands  for tools like `psql` or `mysql`. Those are much more tool specific and the main command has much more useful options that needs to be considered within the short limit of 8 examples that we have.

In short, its a balancing act. We try to keep a page as useful and practical possible within our limited scope.

frazjp65:
I re-added the commands and added `vim -R`. Does it look better now?

agnivade:
Unfortunately, we allow only 8 examples in a page. It is mentioned explicitly in the PR template.

frazjp65:
To be fair, only 4 examples are how to use the command. However, I understand the need for the other examples.

PR #1728, opened 2017-12-01
hg: add page

pxgamer:
----

### DIFF
--- a/pages/common/hg.md
+++ b/pages/common/hg.md
@@ -1,6 +1,6 @@
 # hg
 
-> Provides an interface for Mercurial, a distributed source control management system.
+> Provides a command line interface for Mercurial, a distributed source control management system.
 
 - Check the Mercurial version:
 
--- a/pages/common/hg.md
+++ b/pages/common/hg.md
@@ -1,6 +1,6 @@
 # hg
 
-> The `hg` command provides an interface for Mercurial, a distributed source control management system.
+> Provides an interface for Mercurial, a distributed source control management system.
 
 - Check the Mercurial version:
 
--- a/pages/common/hg.md
+++ b/pages/common/hg.md
@@ -1,6 +1,6 @@
 # hg
 
-> Mercurial distributed source control management.
+> The `hg` command provides an interface for Mercurial, a distributed source control management system.
 
 - Check the Mercurial version:
 
--- a/pages/common/hg.md
+++ b/pages/common/hg.md
@@ -2,9 +2,9 @@
 
 > Provides a command line interface for Mercurial, a distributed source control management system.
 
-- Check the Mercurial version:
+- Execute Mercurial command:
 
-`hg --version`
+`hg {{command}}`
 
 - Call general help:
 
@@ -14,6 +14,6 @@
 
 `hg help {{command}}`
 
-- Execute Mercurial command:
+- Check the Mercurial version:
 
-`hg {{command}}`
+`hg --version`
--- a/pages/common/hg.md
+++ b/pages/common/hg.md
@@ -1,6 +1,6 @@
 # hg
 
-> Provides a command line interface for Mercurial, a distributed source control management system.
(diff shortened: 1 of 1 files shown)


pages/common/hg.md
```diff
@@ -0,0 +1,19 @@
+# hg
+
+> Mercurial distributed source control management.
```
sbrl: This feels like it could be a be more descriptive, maybe?

pages/common/hg.md
```diff
@@ -0,0 +1,19 @@
...
+
+`hg --version`
+
+- Call general help:
+
+`hg help`
+
+- Call help on a command:
+
+`hg help {{command}}`
```
sbrl: I assume that there are/will be separate pages for each command like we have with `git`? Maybe updating the description with s see also might help there.

pxgamer: Yes, I was going to do separate pages because there are a lot of commands.

A 'see also' would be good, but as I haven't yet added any additional commands would it be required yet?

pages/common/hg.md
pxgamer: Maybe something such as?
> The `hg` command provides an interface for Mercurial, a distributed source control management system.

pages/common/hg.md
pxgamer: Also, should additional commands (i.e. `hg-add`, `hg-clone`, etc.) be added as separate PRs?

sbrl: Ah, I see your point! 😛 

Yeah, let's leave it until you've actually created those pages lol

pages/common/hg.md
```diff
@@ -0,0 +1,19 @@
+# hg
+
+> The `hg` command provides an interface for Mercurial, a distributed source control management system.
```
sbrl: Sorry for nitpicking, but I don't think we need to mention the command's name `hg` in the description, since it's already in the title.

Other than that it looks ok to me!

pages/common/hg.md
```diff
@@ -0,0 +1,19 @@
+# hg
+
+> Provides an interface for Mercurial, a distributed source control management system.
```
agnivade: How about `Command line interface for Mercurial ..` ?

pages/common/hg.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Check the Mercurial version:
```
agnivade: This should be the last option

pages/common/hg.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Call general help:
+
+`hg help`
+
+- Call help on a command:
+
+`hg help {{command}}`
+
+- Execute Mercurial command:
```
agnivade: This should be the first option.

pages/common/hg.md
agnivade: Just keep in mind to update the page whenever you submit a PR for a hg subcommand.

agnivade:
I think you made this page mostly similar to git.md page. The order is incorrect in that page too. We already have an open PR to change that page. I will take care to change the order there.

pages/common/hg.md
pxgamer: I didn't know if `Command line` was necessary as it is used on the command line. :thinking: But ok.

pxgamer:
Interesting. I'd thought that the `git` tldr page would have been one of the correct pages. Considering how widely used Git is. Guess I'll change the order.

Do you mind linking to the open PR @agnivade? Cheers.

agnivade:
https://github.com/tldr-pages/tldr/pull/1675

pxgamer:
Cheers. Looking into it.

pages/common/hg.md
agnivade: Simply saying "interface" might leave some doubt as to what kind of interface. Hence, to be extra clear, we can say 'command line interface'

PR #1738, opened 2017-12-03
Added vim and ssh examples

neilhwatson:
----

~~- [ ] The page has 8 or fewer examples.~~

~~- [] The PR is appropriately titled:  
      `<command name>: add page` for new pages, or `<command name>: <description of changes>` for pages being edited~~

I've offered a few additions that do go beyond your 8 examples, but not much. Another option might be an a page of advanced example.

### DIFF
--- a/pages/common/ssh.md
+++ b/pages/common/ssh.md
@@ -31,6 +31,6 @@
 
 `ssh -o "ForwardAgent=yes" {{username}}@{{remote_host}}`
 
-- Copy files en masse via ssh
+- Copy files en masse via ssh:
 
 `tar czf - {{path/to/dir}} | ssh {{username}}@{{host}} 'tar -C "/" -xzf -'`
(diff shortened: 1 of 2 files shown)


agnivade:
Thanks for the  PR !

As you yourself have mentioned, this PR increases beyond our limit of 8 examples. Unfortunately, in that case, we cannot merge it. This is a limit that we apply to ourselves because our scope is limited and we try to distill the most common and practical usages within these 8 examples.

> Another option might be an a page of advanced example.

Unfortunately, that is not an option too. We deliberately do not want to add any advanced options. The pages are mainly focused to keep the most commonly used examples that can be easily understood by a beginner and not to merely provide copy-pastable snippets (like, say, commandlinefu.com) but rather to allow readers to understand the language of the commands.

Hope you can understand. Thanks again for your contribution.

Issue #1732, opened 2017-12-02
command-request: xman


vitalkanev:
Original man page: [here](https://gist.github.com/vitalkanev/d7e6e829524da089aa784a443d423bea)

sbrl:
Thanks for requesting this page, @vitalkanev! Do you want to have a go yourself?

Pages are best created by people who understand the command they are about.

If you do, take a quick read of [CONTRIBUTING.md](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md), which guides you through the process of creating a new page here. Then, just simply send a pull request to this repository - and if you need any assistance, just ask :-)

PR #1740, opened 2017-12-03
xman: add page

vitalkanev:
Related: #1732

----

### DIFF
--- a/pages/linux/xman.md
+++ b/pages/linux/xman.md
@@ -1,0 +1,19 @@
+# xman
+
+> Manual page viewer for X Window System
+
+- Start xman in three-button window:
+
+`xman`
+
+- Open helpfile other than default:
+
+`xman -helpfile {{filename}}`
+
+- Show both manual page and directory:
+
+`xman -bothshown`
+
+- Specify the geometry (size and location) of the top menu:
+
+`xman -geometry {{width}}+{{height}}+{{x}}+{{y}}`


pages/linux/xman.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Open helpfile other than default:
+
+`xman -helpfile {{filename}}`
+
+- Show both manual page and directory:
+
+`xman -bothshown`
+
+- Specify the geometry (size and location) of the top menu:
```
agnivade: Is this a commonly used option ? I would guess that the default geometry should be good enough for most cases.

pages/linux/xman.md
```diff
@@ -0,0 +1,19 @@
+# xman
+
+> Manual page viewer for X Window System.
+
+- Start xman in three-button window:
+
+`xman`
+
+- Open helpfile other than default:
```
agnivade: Can you expand on this a bit ? What does a non-default helpfile mean ?

vitalkanev: Non-default helpfile stands for the man page (Used `man {{command}} > {{command}}.txt` to make these as a text file), which can be used, when pushing the `Manual page` button on the default window. (As I remember)

pages/linux/xman.md
vitalkanev: I guess not. Will remove

pages/linux/xman.md
agnivade: So this is just the output of the man page stored to a file ? If so, can we say `Open manual page output stored in a given file` ?

agnivade:
@vitalkanev - Can you take a look at the comment I gave here - https://github.com/tldr-pages/tldr/pull/1740#discussion_r154750467 ?

pages/linux/xman.md
vitalkanev: I guess yes

agnivade:
ping @sbrl

PR #1749, opened 2017-12-04
snap: add page

eridanired:
Create a new page for the **snap** package management command.

----

### DIFF
--- a/pages/linux/snap.md
+++ b/pages/linux/snap.md
@@ -1,17 +1,17 @@
 # snap
 
 > Tool for managing the "snap" self-contained software packages.
-> Similar to what "apt" is for ".deb".
+> Similar to what `apt` is for ".deb".
 
-- Search for an app:
+- Search for a package:
 
-`snap find {{part_of_a_package_name}}`
+`snap find {{package_name}}`
 
 - Install a package:
 
 `snap install {{package_name}}`
 
-- Display basic info. about installed snap software:
+- Display basic information about installed snap software:
 
 `snap list`
 
--- a/pages/linux/snap.md
+++ b/pages/linux/snap.md
@@ -1,24 +1,24 @@
 # snap
 
-> Tool for managing the *snap* self-contained softwawre packages.
-> Similar to what *apt* is for *.deb*.
+> Tool for managing the "snap" self-contained software packages.
+> Similar to what "apt" is for ".deb".
 
-- Look for an app:
+- Search for an app:
 
-`snap find {{package}}`
+`snap find {{part_of_a_package_name}}`
 
-- Install software:
+- Install a package:
 
-`snap install {{package}}`
+`snap install {{package_name}}`
 
-- Find installed *snap* software:
+- Display basic info. about installed snap software:
 
 `snap list`
 
-- Remove programs:
+- Uninstall a package:
 
-`snap remove {{package}}`
+`snap remove {{package_name}}`
 
-- Check for recent *snap* changes:
+- Check for recent snap changes in the system:
 
 `snap changes`
--- a/pages/linux/snap.md
+++ b/pages/linux/snap.md
@@ -3,22 +3,22 @@
 > Tool for managing the *snap* self-contained softwawre packages.
 > Similar to what *apt* is for *.deb*.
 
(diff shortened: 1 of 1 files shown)


pages/linux/snap.md
```diff
@@ -0,0 +1,24 @@
+# snap
+
+> Tool for managing the *snap* self-contained softwawre packages.
```
sbrl: > softwawre

Typo?

pages/linux/snap.md
```diff
@@ -0,0 +1,24 @@
...
+> Similar to what *apt* is for *.deb*.
+
+- Look for an app:
+
+`snap find {{package}}`
+
+- Install software:
```
sbrl: `software` seems a bit vague. Perhaps `a package` would be better?

pages/linux/snap.md
```diff
@@ -0,0 +1,24 @@
...
+
+`snap install {{package}}`
```
sbrl: Perhaps `package_name` would work better here?

pages/linux/snap.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Find installed *snap* software:
+
+`snap list`
+
+- Remove programs:
```
sbrl: Perhaps `packages` would increase clarity here?

pages/linux/snap.md
agnivade: Please remove the markdown formatting `**` from snap. Use backticks around "apt" and double quotes around ".deb".

pages/linux/snap.md
elopio: or app, to be consistent with the previous entry.

pages/linux/snap.md
```diff
@@ -0,0 +1,24 @@
+# snap
+
+> Tool for managing the *snap* self-contained softwawre packages.
+> Similar to what *apt* is for *.deb*.
+
+- Look for an app:
```
elopio: Why not `Find an app`? I see no need to find a synonym for the action.

pages/linux/snap.md
```diff
@@ -0,0 +1,24 @@
...
+
+`snap find {{package}}`
+
+- Install software:
+
+`snap install {{package}}`
+
+- Find installed *snap* software:
```
elopio: As my previous suggestion about synonyms for actions, I would use `List` instead of `Find` here.

pages/linux/snap.md
elopio: apps or packages sound good to me.

elopio:
This is great @eridanired, thank you!
Could I convince you to also make a page about snapcraft? :D

pages/linux/snap.md
eridanired: I like "software" because snap doesn't necessarily carries "applications" in the mainstream sense of the word. Snap can also be only a library or some other kind of pile of code.

I didn't used package, because newbies not always know what "package" actually means. But yeah "package" could do since it can be used in an equally broad sense as "software".

pages/linux/snap.md
eridanired: Synonym is just to reinforce in the imaginary of the user the meaning and sense of the action that the command performs. We can leave it at "Find" but that would be just being redundant while wasting a chance for providing a bit more of additional help.

pages/linux/snap.md
eridanired: Submitted it, please check it again; made other changes too taking into consideration the other comments.

Thank you.

pages/linux/snap.md
eridanired: Took into consideration the received comments and submitted it again, please give it look. Thanks!

agnivade:
@eridanired - I made some minor corrections which make the page more consistent with the other examples and also in line with the apt.md page. Let me know if it looks good to you.

pages/linux/snap.md
sbrl: It's a word that's used by both `apt` and `npm`, so I thought it might be best to maintain consistency.

eridanired:
Looks great to me, thank you @agnivade and everyone for helping out.

PR #1750, opened 2017-12-05
grep: remove 2 examples

### LINKED ISSUE #1695: Grep page has 10 examples
This might be overly pedantic, but the contribution guidelines specify no more than 8 examples. Grep has 10.

agnivade:
Make the page conform to our maximum limit of 8 examples

Fixes #1695

----

### DIFF
--- a/pages/common/grep.md
+++ b/pages/common/grep.md
@@ -23,17 +23,9 @@
 
 `grep -{{C|B|A}} 3 {{search_string}} {{path/to/file}}`
 
-- Print the count of matches instead of the matching text:
-
-`grep -c {{search_string}} {{path/to/file}}`
-
 - Print line number for each match:
 
 `grep -n {{search_string}} {{path/to/file}}`
-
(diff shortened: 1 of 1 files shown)


agnivade:
I have removed the 2 examples which I feel are least used amongst all of the others. Feel free to comment if you think some other option needs to be removed.

sbrl:
I agree about the count one, but perhaps we could merge the line numbers and filename ones to get us under the limit? I find the filenames one to be rather useful when grepping multiple files myself (i.e. with a glob).

agnivade:
Noted. @waldyrious - Do you have any thoughts ?

gnebehay:
If we keep the filename example, could we change it from `-l` to `-H`? The problem with `-l` is that it suppresses all other output, while `-H` just adds the filename in the beginning of the line, which is what I imagine is more useful.

agnivade:
Done

PR #1752, opened 2017-12-05
tmux inside tmux

venoel:
----

### DIFF
--- a/pages/common/tmux.md
+++ b/pages/common/tmux.md
@@ -29,3 +29,7 @@
 - Kill session:
 
 `tmux kill-session -t {{name}}`
+
+- Use tmux inside tmux:
+
+`Ctrl + B, Ctrl + B`


agnivade:
I have the same question as @sbrl. @venoel - Is this something that you use very commonly ? Or you just added this option because it is possible to do it ?

venoel:
I have the same answer as for @sbrl. :)

For example. I start tmux locally, log in to the remote server and start tmux in remote console.

At the same time I do not insist. It is just proposal (first time in github). I like tmux and I excited of your project.

agnivade:
Why not just type `tmux` in the remote console ? Does that not work ?

venoel:
It works. But then you need control remote tmux, isn't it? You can achieve it by `Ctrl + B, Ctrl + B, c` for create another windows in remote tmux, for examle.

venoel:
It works. But then you need control remote tmux, isn't it? You can achieve it by `Ctrl + B, Ctrl + B, c` for create another windows in remote tmux, for examle.

agnivade:
I think this falls into the category of advanced usage. It would have been okay if the no. of examples would have been 5, but since this stretches it all the way to our max 8, I am reluctant to add this.

PR #1755, opened 2017-12-05
scala: add page

hohonuuli:
----

### DIFF
--- a/pages/common/scala.md
+++ b/pages/common/scala.md
@@ -1,6 +1,6 @@
 # scala
 
-> Scala application launcher and interactive interpreter
+> Scala application launcher and interactive interpreter.
 
 - Call a Scala interactive shell (REPL):
 
@@ -17,8 +17,3 @@
 - Execute a single Scala command in the command line:
 
 `scala -e {{command}}`
-
-
-
-
-
--- a/pages/common/scala.md
+++ b/pages/common/scala.md
@@ -2,7 +2,7 @@
 
 > Scala application launcher and interactive interpreter.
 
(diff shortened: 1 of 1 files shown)


pages/common/scala.md
```diff
@@ -0,0 +1,19 @@
+# scala
+
+> Scala application launcher and interactive interpreter.
+
+- Call a Scala interactive shell (REPL):
```
agnivade: Perhaps "Launch" is better suited than "Call"

pages/common/scala.md
```diff
@@ -0,0 +1,19 @@
...
+
+`scala`
+
+- Execute a script in a given Scala file:
```
agnivade: How about "Execute a given Scala script" ?

pages/common/scala.md
hohonuuli: "Launch" sounds better to me too. There's no consensus on terms though. [python](https://github.com/hohonuuli/tldr/blob/master/pages/common/python.md) uses "Call", [ruby](https://github.com/hohonuuli/tldr/blob/master/pages/common/ruby.md) uses "Open". The `man scala` page uses "Start". Perhaps we should follow the trend of using a different word for every page ... maybe I should use "Instantiate"? ;-) I'll change it to "Start" to keep it in line with the scala man.

pages/common/scala.md
hohonuuli: How about "Execute a Scala script:"

hohonuuli:
👍

sbrl:
Looks ok to me! Thanks, @hohonuuli :D

PR #1762, opened 2017-12-06
wget: added -O example

Neph0:
----


Neph0:
I was surprised when I noticed that this example wasn't part of the 7 already in the page. This is always the option that I look for in the man. Since it's not standard (would usually be `-o` or `-f`), I think it has its place in here.

agnivade:
Wow, I was surprised too.

PR #1768, opened 2017-12-07
strace: add example of tracing file operations

ad-m:
I think this is quite a common case, especially since everything is a file in Linux.


ad-m:
@agnivade , fixed.

PR #1773, opened 2017-12-08
ember: Add `ember build -prod` command

Turbo87:
----

/cc @alexlafroscia

### DIFF
--- a/pages/common/ember.md
+++ b/pages/common/ember.md
@@ -15,6 +15,10 @@
 
 `ember build`
 
+- Build the project in production mode:
+
+`ember build -prod`
+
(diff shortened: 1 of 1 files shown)


alexlafroscia:
Fine by me! As long as there aren't too many examples, I know the target is like 5-7 and we're pushing the upper limit on that.

agnivade:
The max limit is 8. With this example, it touches that. We don't usually reach the limit unless for a good reason. I think building for prod is a good enough reason.

sbrl:
Thanks, @Turbo87 and @alexlafroscia 😺

PR #1788, opened 2017-12-12
p5:add page

hill:
----


hill:
All good @sbrl - thanks for that!

pages/common/p5.md
```diff
@@ -0,0 +1,19 @@
+# p5-manager
```
agnivade: Shouldn't this be `p5` ? Since the command name is p5.

hill: Yeah wasn't sure what was correct to use here - the actual package is called p5-manager (as p5 is a seperate javascript library). Should this be changed to the command name or package name?

sbrl: I think the idea is that it's the command name - which should be the same as the filename IIRC.

hill: Ok, most recent commit has name changed 😁

PR #1791, opened 2017-12-12
pipenv: add page

heimann:
----

### DIFF
--- a/pages/common/pipenv.md
+++ b/pages/common/pipenv.md
@@ -1,6 +1,6 @@
 # pipenv
 
-> Python development workflow for humans.
+> Simple and unified Python development workflow.
 > Manages packages and the virtual environment for a project.
 
 - Create a new project:
--- a/pages/common/pipenv.md
+++ b/pages/common/pipenv.md
@@ -1,7 +1,7 @@
 # pipenv
 
-> Python Development Workflow for Humans.
-> Manages packages and virtual environment for a project.
+> Python development workflow for humans.
+> Manages packages and the virtual environment for a project.
 
 - Create a new project:
 
@@ -15,7 +15,7 @@
 
 `pipenv install {{package_name}}`
 
-- Install all dependencies for a project (including dev):
+- Install all the dependencies for a project (including dev):
 
 `pipenv install --dev`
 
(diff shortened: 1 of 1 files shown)


pages/common/pipenv.md
```diff
@@ -0,0 +1,28 @@
+# pipenv
+
+> Python Development Workflow for Humans.
```
sbrl: Perhaps this would read better in sentence case rather than title case?

pages/common/pipenv.md
```diff
@@ -0,0 +1,28 @@
...
+> Manages packages and virtual environment for a project.
```
sbrl: Perhaps `the virtual environment` would read better here?

pages/common/pipenv.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Create a new project using Python 3:
+
+`pipenv --three`
+
+- Install a package:
+
+`pipenv install {{package_name}}`
+
+- Install all dependencies for a project (including dev):
```
sbrl: perhaps `all the dependencies` would read better here?

pages/common/pipenv.md
```diff
@@ -0,0 +1,28 @@
...
+
+`pipenv install --dev`
+
+- Uninstall a package:
+
+`pipenv uninstall {{package_name}}`
+
+- Activate a shell within the created virtual environment:
```
sbrl: Perhaps `Start` would aid clarity here?

heimann: `Start a shell within...` etc?

pages/common/pipenv.md
heimann: I agree

pages/common/pipenv.md
```diff
@@ -0,0 +1,28 @@
+# pipenv
+
+> Python development workflow for humans.
```
agnivade: The "for humans" part makes it a bit informal. Can we make it "Simple and unified Python packaging tool" ?

heimann: How about "Python packaging and virtual environments made simple."? Key part of the program is that it handles not just package management but the whole environment around it.

Another option could be: "Simple and unified Python development workflow"?

agnivade: I like the second one.

heimann: Ok done. Thanks for the feedback!

PR #1796, opened 2017-12-13
dep: add page

### LINKED ISSUE #1659: page request: dep
Webpage here - https://github.com/golang/dep#usage

Please go through the [CONTRIBUTING](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) page to understand what is the format we use for our pages.

ravernkoh:
dep is a tool for dependency management in the Go ecosystem.

Fixes #1659.

----

### DIFF
--- a/pages/common/dep.md
+++ b/pages/common/dep.md
@@ -1,12 +1,12 @@
 # dep
 
-> Tool for dependency management in the Go ecosystem.
+> Tool for dependency management in Go projects.
 
-- Initialize a directory to be a project:
+- Initialize the current directory as the root of a Go project:
 
 `dep init`
 
-- Install the project dependencies:
+- Install any missing dependencies (Scans Gopkg.toml and your .go files):
 
 `dep ensure`
 
@@ -14,7 +14,7 @@
 
 `dep status`
 
-- Add a new depedency:
+- Add a dependency to the project:
 
 `dep ensure -add {{project_url}}`
 
--- a/pages/common/dep.md
+++ b/pages/common/dep.md
@@ -1,0 +1,23 @@
+# dep
+
+> Tool for dependency management in the Go ecosystem.
+
+- Initialize a directory to be a project:
+
+`dep init`
+
+- Install the project dependencies:
+
+`dep ensure`
+
+- Report the status of the project's depedencies:
+
+`dep status`
+
+- Add a new depedency:
+
+`dep ensure -add {{project_url}}`
+
+- Update the locked versions of all dependencies:
+
+`dep ensure -update`
--- a/pages/common/dep.md
+++ b/pages/common/dep.md
@@ -10,13 +10,13 @@
 
 `dep ensure`
 
-- Report the status of the project's depedencies:
+- Report the status of the project's dependencies:
 
 `dep status`
 
 - Add a dependency to the project:
 
-`dep ensure -add {{project_url}}`
+`dep ensure -add {{package_url}}`
 
 - Update the locked versions of all dependencies:
 


sbrl:
Thanks for the pull request, @ravernkoh! Once we get a second review we can get to merging :D

Is this PR related to Google Code-In?

pages/common/dep.md
```diff
@@ -0,0 +1,23 @@
+# dep
+
+> Tool for dependency management in the Go ecosystem.
```
waldyrious: How about "Tool for dependency management in Go projects"?

pages/common/dep.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Initialize a directory to be a project:
```
waldyrious: I'd suggest something like "Initialize a directory as the root of a Go project" instead. Thoughts?

pages/common/dep.md
```diff
@@ -0,0 +1,23 @@
...
+
+`dep init`
+
+- Install the project dependencies:
+
+`dep ensure`
```
waldyrious: The description is fine, but just so it matches the actual command word, maybe "Install any missing dependencies of the project". Also, I assume this only checks for Go dependencies, right? In which case, I'd specify that as well ("any missing Go dependencies", or another wording that you find more sensible).

pages/common/dep.md
```diff
@@ -0,0 +1,23 @@
...
+
+`dep status`
+
+- Add a new depedency:
+
+`dep ensure -add {{project_url}}`
+
+- Update the locked versions of all dependencies:
+
+`dep ensure -update`
```
waldyrious: Is there a way to manually lock/unlock dependencies? If so, I think it should be mentioned before this example, to introduce the concept.

waldyrious:
It certainly looks to be connected to [task #5102697852174336](https://codein.withgoogle.com/dashboard/task-instances/5102697852174336/), yeah.

I've left some comments -- mostly editorial, i.e. the page could still be merged as-is if time presses, but I'd like to hear your thoughts.

pages/common/dep.md
ravernkoh: I agree with this!

pages/common/dep.md
ravernkoh: Yes. I will change the wording.

pages/common/dep.md
ravernkoh: Nope. Unfortunately I don't think there is a way to lock & unlock dependencies. To change the locked version I guess you can set the version `Gopkg.toml` and then run this update command.

How should I communicate the fact that the user is supposed to edit the Gopkg.toml? From what I've seen, most of the descriptions are of a single line, describing what the command should do.

ravernkoh:
Your suggestions have been very helpful.

I have updated the files to include your suggestions. However, from my current knowledge of looking through the usage docs and the help pages, there's no real way of locking and unlocking dependencies :(

pages/common/dep.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Initialize the current directory as the root of a Go project:
+
+`dep init`
+
+- Install any missing dependencies (Scans Gopkg.toml and your .go files):
+
+`dep ensure`
+
+- Report the status of the project's depedencies:
```
agnivade: typo => depedencies

pages/common/dep.md
agnivade: This is fine @ravernkoh. @waldyrious - the users are expected to go through the `dep` command. So essentially, until you do an `update`, everybody gets the same version of the dependencies.

pages/common/dep.md
```diff
@@ -0,0 +1,23 @@
...
+
+`dep status`
+
+- Add a dependency to the project:
+
+`dep ensure -add {{project_url}}`
```
agnivade: Please change project_url => package_url

ravernkoh:
Updated.

waldyrious:
Nice work @ravernkoh!

PR #1800, opened 2017-12-13
split file command as it takes different args on macOS

gingerbeardman:
`-i` on linux
`-I` or `--mime` on macOS


agnivade:
Thanks.

PR #1801, opened 2017-12-13
git-commit: auto stage and commit

flexdinesh:
Added entry for auto staging all changes during commit

### DIFF
--- a/pages/common/git-commit.md
+++ b/pages/common/git-commit.md
@@ -1,6 +1,6 @@
 # git commit
 
-> Commit staged files to the repository.
+> Commit files to the repository.
 
 - Commit staged files to the repository with comment:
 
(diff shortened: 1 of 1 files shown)


pages/common/git-commit.md
```diff
@@ -1,6 +1,6 @@
 # git commit
 
-> Commit staged files to the repository.
```
sbrl: Why remove `staged` here?

flexdinesh: Since the new content describes how to commit files even before staging (auto stage and commit) them, I thought removing 'staged' in the original desc made more sense.

sbrl: Ah, ok. Sure!

PR #1802, opened 2017-12-14
sshuttle: add page

sbrl:
----

### DIFF
--- a/pages/linux/sshuttle.md
+++ b/pages/linux/sshuttle.md
@@ -13,4 +13,4 @@
 
 - Forward all IPv4 and IPv6 traffic via a remote SSH server:
 
-`sudo sshuttle --remote={{username}}@{{sshserver}} --method=tproxy {{0.0.0.0/0}} {{::/0}} --exclude={{your_local_ip_address}} --exclude={{ssh_server_ip_address}}
+`sudo sshuttle --remote={{username}}@{{sshserver}} --method=tproxy {{0.0.0.0/0}} {{::/0}} --exclude={{your_local_ip_address}} --exclude={{ssh_server_ip_address}}`
--- a/pages/linux/sshuttle.md
+++ b/pages/linux/sshuttle.md
@@ -11,6 +11,6 @@
 
 `sshuttle --dns --remote={{username}}@{{sshserver}} {{0.0.0.0/0}}`
 
-- Forward all IPv4 and IPv6 traffic:
+- Use the tproxy method to forward all IPv4 and IPv6 traffic:
 
-`sudo sshuttle --remote={{username}}@{{sshserver}} --method=tproxy {{0.0.0.0/0}} {{::/0}} --exclude={{your_local_ip_address}} --exclude={{ssh_server_ip_address}}`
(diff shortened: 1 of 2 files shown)


pages/linux/sshuttle.md
```diff
@@ -0,0 +1,16 @@
...
+
+`sshuttle --remote={{username}}@{{sshserver}} {{0.0.0.0/0}}`
+
+- Forward all IPv4 TCP and DNS traffic via a remote SSH server:
+
+`sshuttle --dns --remote={{username}}@{{sshserver}} {{0.0.0.0/0}}`
+
+- Forward all IPv4 and IPv6 traffic via a remote SSH server:
+
+`sudo sshuttle --remote={{username}}@{{sshserver}} --method=tproxy {{0.0.0.0/0}} {{::/0}} --exclude={{your_local_ip_address}} --exclude={{ssh_server_ip_address}}`
```
agnivade: I see you have added `--exclude`. Should we introduce this option in a next example ? Or do you think we can remove it ?

pages/linux/sshuttle.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Forward all IPv4 TCP traffic via a remote SSH server:
+
+`sshuttle --remote={{username}}@{{sshserver}} {{0.0.0.0/0}}`
+
+- Forward all IPv4 TCP and DNS traffic via a remote SSH server:
+
+`sshuttle --dns --remote={{username}}@{{sshserver}} {{0.0.0.0/0}}`
+
+- Forward all IPv4 and IPv6 traffic via a remote SSH server:
```
agnivade: No need to mention "via a remote SSH server" after the 2nd time.

agnivade: No need to mention "via a remote SSH server" after the 2nd time.

pages/linux/sshuttle.md
sbrl: Hrm. Apparently it's required here because otherwise in `tproxy` mode it bounces the incoming traffic from the SSH server back again, causing infinite loops. It might also bounce the outgoing traffic in weird ways too without both excludes here.

pages/linux/sshuttle.md
sbrl: Ok, sure.

pages/linux/sshuttle.md
agnivade: My point was that all your previous examples could be matched to their descriptions.

Like "remote ssh server" for `--remote`
DNS traffic for `--dns`

And then suddenly the description says "IPV4 and IPV6 traffic"  along with which, there is a slew of new flags like `--method` and `--exclude` which are not mentioned in the description. It would be good if we can explain them in the description. Does that sound okay ?

sbrl: Ahhhh. Right. I see what you're getting at! Good thinking. Let me see if I can rearrange this a bit.

sbrl: Yeah, it's certainly an awkward one. How's that? The issue is that the previous methods only handle IPv4 traffic, and this more direct method of tunnelling traffic, though convenient once set up, comes with several caveats that must be taken care of with additional flags, unfortunately.

PR #1806, opened 2017-12-15
hg-clone: add page

pxgamer:
----

### DIFF
--- a/pages/common/hg-clone.md
+++ b/pages/common/hg-clone.md
@@ -1,0 +1,23 @@
+# hg clone
+
+> Create a copy of an existing repository in a new directory.
+
+- Clone a repository to a specified directory:
+
+`hg clone {{remote_repository_source}} {{destination_path}}`
+
+- Clone a repository on a specific branch:
+
+`hg clone --branch {{branch}} {{remote_repository_source}}`
+
+- Clone a repository with a clean working directory:
+
+`hg clone --noupdate {{remote_repository_source}}`
+
+- Clone a repository to a specific revision, tag or branch:
+
+`hg clone --updaterev {{revision}} {{remote_repository_source}}`
+
+- Clone a repository up to a specific revision without any newer history:
+
+`hg clone --rev {{revision}} {{remote_repository_source}}`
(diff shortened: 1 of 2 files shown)


pages/common/hg.md
```diff
@@ -1,7 +1,7 @@
 # hg
 
 > A command line interface for Mercurial, a distributed source control management system.
-> See the `hg-add` and `hg-commit` pages for additional information.
+> See the `hg-clone`, `hg-add` and `hg-commit` pages for additional information.
```
agnivade: Maybe you should add `and other pages` because it does not make sense to keep on adding commands here as you add more pages. People should get the general idea that other pages exist and they can just tldr it.

pages/common/hg-clone.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Clone a repository to a specified directory:
+
+`hg clone {{remote_repository_source}} {{destination_path}}`
+
+- Clone a repository on a specific branch:
+
+`hg clone --branch {{branch}} {{remote_repository_source}}`
+
+- Clone a repository with a clean working directory:
```
agnivade: What does "clean working directory" mean ? Does it mean you have to be inside a directory which already has the code ? Or it should be an empty directory ?

pages/common/hg-clone.md
```diff
@@ -0,0 +1,23 @@
...
+
+`hg clone --noupdate {{remote_repository_source}}`
+
+- Clone a repository to a specific revision, tag or branch:
```
agnivade: So the `--updaterev` flag takes revision and branch both ? Why should someone use the `--branch` flag then ? And not use `--updaterev` always.

pages/common/hg.md
pxgamer: Makes sense. :+1:

pages/common/hg-clone.md
pxgamer: `--updaterev` will checkout the specified revision, tag or branch.

I'll mention that `--branch` only clones the specified branch. And ignores the others unless they are pulled.

pages/common/hg-clone.md
pxgamer: Yeh, that makes sense it's not very clear.

It just means that it only creates the `.hg` directory and doesn't check out any files.

sbrl:
Thanks again, @pxgamer!

PR #1807, opened 2017-12-15
dc: add page

mathomp4:
----

I figured `bc` had a `tldr`, so `dc` should as well.

### DIFF
--- a/pages/common/dc.md
+++ b/pages/common/dc.md
@@ -21,4 +21,3 @@
 - To 100 decimal places [100 k], calculate the golden ratio, phi: sqrt(5) [5 v] plus 1 [1 +], divided by 2 [2 /], and [p]rint result:
 
 `dc -e "100 k 5 v 1 + 2 / p"`
-
--- a/pages/common/dc.md
+++ b/pages/common/dc.md
@@ -1,0 +1,19 @@
+# dc
+
+> An arbitrary precision calculator. Uses reverse polish notation (RPN).
+
+- Run calculator in interactive mode:
+
+`dc`
+
+- Calculate 4 times 5 [4 5 *] subtract 17 [17 -] and [p]rint the output :
+
+`echo "4 5 * 17 - p"| dc`
+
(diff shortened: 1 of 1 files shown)


kritixilithos:
Maybe you could also mention the `-e` flag in the page as an alternative to piping in the input (for example `dc -e "10 8^1-p"`).

mathomp4:
Huh. I actually did not know about that! I must be old. I'm used to using `echo` or `<<<`. I refrained from the `<<<` version since it's `bash` only. Let me add one.

pages/common/dc.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Execute dc script in file:
+
+`dc -f {{file}}`
+
+- Calculate 4 times 5 [4 5 *], subtract 17 [17 -], and [p]rint the output (using echo):
+
+`echo "4 5 * 17 - p"| dc`
+
+- Set number of decimal places to 10 [k 7], calculate 5 divided by -3 [5 _3 /] and [p]rint (using dc -e):
```
agnivade: Shouldn't this be [10 k] ? Considering you have [100 k] in the next example ?

mathomp4: Sigh. Yes. Errant vim keystrokes on home laptop. I'll use 7 places since the code actually does too.

pages/common/dc.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Calculate 4 times 5 [4 5 *], subtract 17 [17 -], and [p]rint the output (using echo):
+
+`echo "4 5 * 17 - p"| dc`
+
+- Set number of decimal places to 7 [7 k], calculate 5 divided by -3 [5 _3 /] and [p]rint (using dc -e):
+
+`dc -e "7 k 5 _3 / p"`
+
+- To 100 decimal places [100 k], calculate the golden ratio, phi: sqrt(5) [5 v] plus 1 [1 +], divided by 2 [2 /], and [p]rint result:
```
agnivade: Let's use a consistent description here too - `Set number of decimal places`

PR #1813, opened 2017-12-17
s: add page

mstruebing:
s is a websearch CLI.
https://github.com/zquestz/s

----

### DIFF
--- a/pages/common/s.md
+++ b/pages/common/s.md
@@ -16,4 +16,4 @@
 
 - Use a specified binary to perform the search query:
 
-`s --binary {{binary}} {{query}}`
+`s --binary "{{binary}} {{arguments}}" {{query}}`
--- a/pages/common/s.md
+++ b/pages/common/s.md
@@ -2,18 +2,18 @@
 
 > Web search from the terminal.
 
-- Search for query on Google(default provider):
+- Search for a query on Google(default provider):
 
 `s {{query}}`
-
-- Search for a query on provider:
-
-`s --provider {{provider}} {{query}}`
 
 - List all providers:
 
 `s --list-providers`
 
-- Search for query in binary:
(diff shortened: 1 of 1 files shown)


pages/common/s.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Search for query on Google(default provider):
+
+`s {{query}}`
+
+- Search for a query on provider:
+
+`s --provider {{provider}} {{query}}`
+
+- List all providers:
```
agnivade: I think this should be above "search with a given provider". Because first, a person needs to list the providers to get to know what all providers are there to search with something other than google.

pages/common/s.md
```diff
@@ -0,0 +1,19 @@
+# s
+
+> Web search from the terminal.
+
+- Search for query on Google(default provider):
+
+`s {{query}}`
+
+- Search for a query on provider:
```
agnivade: on provider => with a given provider. Just to make things a bit clearer.

pages/common/s.md
```diff
@@ -0,0 +1,19 @@
...
+
+`s --provider {{provider}} {{query}}`
+
+- List all providers:
+
+`s --list-providers`
+
+- Search for query in binary:
```
agnivade: Perhaps - `Use a specified binary to perform the search query` fits better. Also are the quotes necessary ? I believe it just needs the path to the binary file.

mstruebing: The quotes aren't really necesseray, but mostly you would do something like: `chromium --incognito` and to get the arguments passed to `the binary` it is necessary, otherwise they would be interpreted by `s`. I can remove than or change them to `"{{binary}} {{arguments}}"`?

agnivade: Yes the latter is better.

PR #1815, opened 2017-12-17
lpass: add page

hdhauk:
----


pages/common/lpass.md
```diff
@@ -0,0 +1,23 @@
+# lpass
+
+> Command line interface for LastPass Password Manager.
+
+- Login to your LastPass account, by entering your Master Password when promted:
```
agnivade: typo at "promted"

pages/common/lpass.md
```diff
@@ -0,0 +1,23 @@
...
+
+- List all sites grouped by category:
+
+`lpass ls`
+
+- Show password for a specified entry (usually the website url):
+
+`lpass show {{identifier}} --password`
+
+- Generate a new password and add to LastPass:
```
agnivade: This should go above the previous example. And let's use concrete examples for url and identifier. And use the same identifier for the `show` command.

Also typo at password_lenght

pages/common/lpass.md
```diff
@@ -0,0 +1,23 @@
+# lpass
+
+> Command line interface for LastPass Password Manager.
```
agnivade: Should P and M be in small case ?

pages/common/lpass.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Login to your LastPass account, by entering your Master Password when prompted:
```
agnivade: Same here, maybe M and P should be lower case ?

pages/common/lpass.md
```diff
@@ -0,0 +1,23 @@
...
+
+`lpass status`
+
+- List all sites grouped by category:
+
+`lpass ls`
+
+- Generate a new password and add to LastPass:
+
+`lpass generate --username {{username}} --url {{gmail.com}} {{gmail.com}} {{password_length}}`
```
agnivade: Probably making the identifier different from the site name would make it clearer. Also we should update the description saying that `Generate a new password for <gmail.com> with identifier <myinbox> and add to LastPass`. That makes it clear what is what.

pages/common/lpass.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Show password for a specified entry (usually the website url):
```
agnivade: Since we are making the entry different from the website, let us remove the "usually the website url" part. The user can figure out that he/she can always make the identifier same as the website.

hdhauk: I put that in there as I assume most of the passwords in a users vault are created by a browser plugin which uses the domain name as identifier. But I can change if you would like?

agnivade: Hmm ok. But won't it be confusing that the example will show an identifier that is _not_ the website url, but the description mentions that usually it's the website url. That is my only concern.

hdhauk: I'm not sure if I understand you fully. 

To clarify, the `lpass generate`-example will create an entry similar to the one a browser-plugin would, with the identifier in this case being `gmail.com`. In the next example where we show the password, we again use the same identifier `gmail.com` to extract the password that was generated.

hdhauk: Never mind, I didn't see your other comments before. Have amended the two last ones according to your input.

hdhauk:
Looks good!

PR #1817, opened 2017-12-18
lastlog.md: add page

jsonbruce:
----

### DIFF
--- a/pages/linux/lastlog.md
+++ b/pages/linux/lastlog.md
@@ -1,0 +1,19 @@
+# lastlog
+
+> Show the most recent login of all users or of a given user.
+
+- Display the most recent login of all users:
+
+`lastlog`
+
+- Display lastlog record of the specified user:
+
+`lastlog -u {{username}}`
+
+- Display only lastlog records before than 7 days:
+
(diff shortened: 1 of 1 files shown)


pages/linux/lastlog.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Display the most recent login of all users:
+
+`lastlog`
+
+- Display lastlog record of the specified user:
+
+`lastlog -u {{username}}`
+
+- Display only lastlog records before than 7 days:
```
agnivade: Perhaps `Display only records older than 7 days` reads better. No need to mention "lastlog" again.

pages/linux/lastlog.md
```diff
@@ -0,0 +1,19 @@
...
+
+`lastlog -b 7`
+
+- Display the lastlog records more recent than 3 days:
```
agnivade: Same here. Remove "the lastlog"

pages/linux/lastlog.md
jsonbruce: Yes, but I think `before` can be hint of the arg `-b`

PR #1818, opened 2017-12-18
getfacl.md: add page

jsonbruce:
----

### DIFF
--- a/pages/linux/getfacl.md
+++ b/pages/linux/getfacl.md
@@ -1,0 +1,15 @@
+# getfacl
+
+> Get file access control lists.
+
+- Display the file access control list:
+
+`getfacl -a {{path/to/file_or_folder}}`
+
+- Display the file access control list with numeric user and group IDs:
+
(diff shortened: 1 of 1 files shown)


pages/linux/getfacl.md
```diff
@@ -0,0 +1,15 @@
+# getfacl
+
+> Get file access control lists.
+
+- Display the file access control list:
+
+`getfacl -a {{path/to/file_or_folder}}`
```
agnivade: This seems to give the same result as without `-a`. Let's remove it in that case.

jsonbruce: Yes, you're right. The default operation of `getfacl` is `getfacl -a`

jsonbruce: But I don't think `getfacl -PR` will be useful for most situations. The output format of `getfacl -PR` is not as good as `ls -al`.
So I think `getfacl` is more suit for single file.

agnivade: Understood. Let's do the first part then.

PR #1823, opened 2017-12-18
serverless: add page

heimann:
----


pages/common/serverless.md
```diff
@@ -0,0 +1,27 @@
+# serverless
+
+> Toolkit for deploying and operating serverless architectures on AWS, Google Cloud, Azure and IBM OpenWhisk.
+
+- Create a serverless project:
+
+`sls create`
+
+- Create a serverless project from template:
```
sbrl: Perhaps ...`from a template` would read better?

pages/common/serverless.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Deploy to a cloud provider:
+
+`sls deploy`
+
+- Display information about a serverless project:
+
+`sls info`
+
+- Follow logs for a project:
```
sbrl: Perhaps `Follow the logs`... would read better?

pages/common/serverless.md
```diff
@@ -0,0 +1,27 @@
+# serverless
```
sbrl: You've called the page `serverless`, but the command name is `sls`? We generally prefer pages to be filed under their command names.

heimann: So this one is a little tricky I guess. The package is called [serverless](https://serverless.com), and it comes with two command aliases, `serverless` and `sls`. `sls` is shorter to type so I tend to go with that and I figured a good "tldr" of the command would show that usage as well. 

I see three options for resolving this:
1) Renaming all commands to `serverless` i.e. `serverless create`
2) Adding a description at the top indicating that it can be run as either `serverless` or `sls`.
3) Renaming the page `sls` (this one is my least favorite, I think it'll be confusing).
4) Having two pages, one called sls.md and one serverless.md with the same commands but using the different aliases. 

Is there one of these (or an option I couldn't think of) that you feel is more in line with convention?

agnivade: Please do 1 and 2. You can add a second line saying that the command can be run either as `serverless` or `sls`.

heimann: Ok changed, let me know what you think. Thanks for the feedback y'all

pages/common/serverless.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Deploy to a cloud provider:
+
+`serverless deploy`
+
+- Display information about a serverless project:
+
+`serverless info`
+
+- Follow the logs for a project:
```
agnivade: This should be the last example.

pages/common/serverless.md
```diff
@@ -0,0 +1,28 @@
...
+
+`serverless logs -t`
+
+- Display metrics for a function:
```
agnivade: Should we replace this by the invoke function command ? Seems to me that invoking a function is more important that displaying metrics. Thoughts ?

pages/common/serverless.md
sbrl: Yeah, #1 and #2 sound like the best plan to me.

pages/common/serverless.md
heimann: So remove display metrics?

heimann: Or just move it to the bottom?

pages/common/serverless.md
heimann: Hmm, that's an interesting thought. I certainly find myself using invoke more during development but metrics once something is live. You're probably right though in that in general invoking is more important, I'll switch it out.

heimann:
@sbrl Are you still waiting for me to do more changes? It says you requested changes but as far as I can see I addressed all of them. (Sorry if I'm missing something obvious here, I'm new to this process)

sbrl:
Oops, nope! All looks good to me :D 
Looks like this PR might have slipped through the net. Thanks for pinging me here!

PR #1824, opened 2017-12-19
export: add page

Crease29:


pages/common/export.md
```diff
@@ -0,0 +1,15 @@
...
+
+> Command to set environment variables.
+
+- Set a new environment variable:
+
+`export {{VARIABLE}}={{value}}`
+
+- Echo the value of an environment variable (don't forget the Dollar sign):
+
+`echo {{$VARIABLE}}`
```
sbrl: This example doesn't appear to be for the `export` command?

pages/common/export.md
```diff
@@ -0,0 +1,15 @@
+# export
+
+> Command to set environment variables.
```
sbrl: Can't the export command be used to 'export' a variable's value to sub-processes too?

pages/common/export.md
Crease29: You're right. Shall I remove that example?

pages/common/export.md
Crease29: It's open to interpretation, but yes. `export` defines variables in the **current** environment and therefore sub-processes can access them as well.
Shall I point out in the description that variables that are set with `export` are only present in the current session and will not be persisted?

pages/common/export.md
agnivade: Yes.

agnivade:
`export` is a shell built-in actually and not a command. And also, does this work in Windows ? If not, please create duplicate copies inside the linux and osx folder.

Crease29:
Thanks for your feedback @agnivade. Please review the changes I've made.

sbrl:
Looks ok to me! Thanks, @Crease29 😺

PR #1826, opened 2017-12-20
nmon: add page

jsonbruce:
----

### DIFF
--- a/pages/linux/nmon.md
+++ b/pages/linux/nmon.md
@@ -1,6 +1,6 @@
 # nmon
 
-> A system administrator, tuner, benchmark tool.
+> A system administrator, tuner, and benchmark tool.
 
 - Start nmon:
 
--- a/pages/linux/nmon.md
+++ b/pages/linux/nmon.md
@@ -1,0 +1,15 @@
+# nmon
+
+> A system administrator, tuner, benchmark tool.
+
+- Start nmon:
+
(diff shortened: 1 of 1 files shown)


pages/linux/nmon.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Start nmon:
+
+`nmon`
+
+- Save records to file (default args "-s300 -c288" for 300 seconds between refreshing the screen and 288 refreshes):
+
+`nmon -f`
+
+- Save records to file with 30 seconds refreshing gap and 240 refreshes:
```
agnivade: What does "240 refreshes" mean ? Does it mean that within a single interval, it will take 240 measurements ?

jsonbruce: No. It means the system will take 240 measurements and the gap between each measurement is 30 seconds. So the whole time monitored is 2 hours.
I've drew a illustration to explain this command:

selection_200

agnivade: I see. In that case, I would slightly change the description to something like - 
`Save records to file with a total of 240 measurements by taking 30 seconds between each measurement.`

jsonbruce: Great, I love this sentence!

jsonbruce:
LGTM. Thanks! @sbrl

PR #1827, opened 2017-12-20
hardinfo.md: add page

jsonbruce:
----

### DIFF
--- a/pages/linux/hardinfo.md
+++ b/pages/linux/hardinfo.md
@@ -5,3 +5,11 @@
 - Start hardinfo:
 
 `hardinfo`
+
+- Print report to standard output:
+
+`hardinfo -r`
+
+- Save report to HTML file:
+
+`hardinfo -r -f html > hardinfo.html`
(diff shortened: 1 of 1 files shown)


pages/linux/hardinfo.md
```diff
@@ -0,0 +1,7 @@
+# hardinfo
+
+> Show hardware information in GUI window.
+
+- Start hardinfo:
```
sbrl: Perhaps we could include another example or two? `hardinfo --help` outputs this for me: https://starbeamrainbowlabs.com/bin/ms05

jsonbruce: Yes, You're right.
Actually most of the time we use `hardinfo` as GUI tool, but it also can output hardware information to stdout.
I should add the other situation even if we rarely use it.

sbrl: Ah, ok. I guess the generation of a report might be useful in an automation scenario - e.g. where one has 100 machines and needs information on all their hardware for an audit or something.

jsonbruce: Yes, it's a good application situation.

PR #1831, opened 2017-12-20
i7z.md: add page

jsonbruce:
----

### DIFF
--- a/pages/linux/i7z.md
+++ b/pages/linux/i7z.md
@@ -1,6 +1,6 @@
 # i7z
 
-> A CPU (only i3, i5 and i7) realtime reporting tool.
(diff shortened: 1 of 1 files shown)


pages/linux/i7z.md
```diff
@@ -0,0 +1,7 @@
+# i7z
+
+> A CPU (only i3, i5 and i7) realtime reporting tool.
```
agnivade: Perhaps better to add "Intel" for extra clarity ?

jsonbruce: Yeah, I've considered that but there is no other CPU models called i3/5/7. So I'm not sure if I should use description 
`An Intel CPU (only i3, i5 and i7) realtime reporting tool.`

PR #1835, opened 2017-12-21
pg_ctl: add page

heimann:
----

### DIFF
--- a/pages/common/pg_ctl.md
+++ b/pages/common/pg_ctl.md
@@ -1,0 +1,19 @@
+# pg_ctl
+
+> Utility for controlling a PostgreSQL server and database cluster.
+
+- Start a PostgreSQL server:
+
+`pg_ctl -D {{data_directory}} -l {{log_file_name}}`
+
+- Initialize a PostgreSQL database cluster:
+
+`pg_ctl -D {{data_directory}} init`
+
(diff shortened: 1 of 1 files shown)


sbrl:
Thanks, @heimann! 😺

pages/common/pg_ctl.md
```diff
@@ -0,0 +1,19 @@
...
+
+`pg_ctl -D {{data_directory}} -l {{log_file_name}}`
+
+- Initialize a PostgreSQL database cluster:
+
+`pg_ctl -D {{data_directory}} init`
+
+- Stop a PostgreSQL server:
+
+`pg_ctl -D {{data_directory}} stop`
```
agnivade: Don't we need to be root for the start/stop operations ?

heimann: No, I think it's best practice to not run those as root. If you initialize the db and root owns the data directory then yes you need to be root to stop it, but the recommended way is to actually create a separate 'postgres' user and run it as that user:
```su postgres -c 'pg_ctl start -D /usr/local/pgsql/data -l serverlog'```
as per: https://www.postgresql.org/docs/9.1/static/server-start.html

agnivade: Right, thanks.

PR #1842, opened 2017-12-28
tree.md: add example for exclude directory

n4nagappan:
----

### DIFF
--- a/pages/linux/tree.md
+++ b/pages/linux/tree.md
@@ -26,6 +26,6 @@
 
 `tree -P {{*.txt}}`
 
-- Exclude a directory:
+- Ignore files that match a wildcard (glob) pattern:
 
-`tree -I node_modules`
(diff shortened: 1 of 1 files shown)


pages/linux/tree.md
```diff
@@ -25,3 +25,7 @@
 - Filter the tree using a wildcard (glob) pattern:
 
 `tree -P {{*.txt}}`
+
+- Exclude a directory:
```
agnivade: Perhaps "Ignore" is better than "Exclude" because the option is `-I` and we always strive to use the word used in the flag.

Also, the documentation says that `-I` can be a pattern like `-P`. Perhaps we should clarify that rather than saying just a "directory" ?

n4nagappan: @agnivade Fixed with 68ccb63f989bfdbac1ed6dab93edd7338aedacd5

PR #1850, opened 2017-12-30
java: add custom class command

J053Fabi0:
Custom input and output when compiling a class with the java command


pages/common/java.md
```diff
@@ -13,3 +13,7 @@
 - Display JDK, JRE and HotSpot versions:
 
 `java -version`
+
+- Custom input and output files. Preferably .txt files:
```
agnivade: Move this to the second example. And we should have a clearer description explaining what the command is doing. I am guessing this allows to execute a .class file where the filename is different than the class name. We should mention that to make it clearer. I am doubtful on what the output file is for though. I would guess normally people would want to execute a .class file ?

J053Fabi0:
I didn't knew that! I thought it was only for the java command. 😆

agnivade:
No problem :)

PR #1504, opened 2017-09-25
Create ISSUE_TEMPLATE

agnivade:
----

~- [ ] The page (if new), does not already exist in the repo.~

~- [ ] The page (if new), has been added to the correct platform folder:  
      `common/` if it's common to all platforms, `linux/` if it's Linux-specific, and so on.~

~- [ ] The page has 8 or fewer examples.~

~- [x] The PR is appropriately titled:  
      `<command name>: add page` for new pages, or `<command name>: <description of changes>` for pages being edited~

~- [ ] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines~

Closes #1069

PR #1851, opened 2017-12-30
ISSUE_TEMPLATE: add missing placeholder

waldyrious:
The placeholder was inadvertently made invisible by Github's HTML sanitization while rendering the text [here](https://github.com/tldr-pages/tldr/pull/1504#issuecomment-354375185). I've updated that comment to escape the `<` and `>`, and this PR fixes the actual template file.


agnivade:
Oops, thanks.

PR #1852, opened 2017-12-31
yarn: add yarn list and yarn run example

zkcode:
----

### DIFF
--- a/pages/common/yarn.md
+++ b/pages/common/yarn.md
@@ -26,7 +26,7 @@
 
 `yarn why {{module_name}}`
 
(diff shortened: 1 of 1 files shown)


sbrl:
Hey, thanks for the addition, @zkcode! Unfortunately, this brings us above the maximum number of examples per page, which is 8. Is there any way we could reduce the number of examples on this page to bring us under that limit at all?

agnivade:
I would say we can let go of the last 2 examples. They are not very commonly used.

zkcode:
thanks for suggestion, anyway we should adhere to the predefined rules.

PR #1857, opened 2018-01-02
npx: add page

baixiaoji:
----

### DIFF
--- a/pages/common/npx.md
+++ b/pages/common/npx.md
@@ -1,12 +1,12 @@
 # npx
 
-> Execute  npm packages binaries.
+> Execute binaries from `npm` packages(most of the cli tools) .
 
-- Execute the npm packages binaries:
+- Execute the npm packages binaries(Package will be installed if it does not exist in node_modules or $PATH):
 
 `npx {{module_name}}`
 
-- Define the package to be installed, then execute the binaries:
+- Define the package name to be installed, then execute the binaries:
 
 `npx -p {{module_name}}`
 
--- a/pages/common/npx.md
+++ b/pages/common/npx.md
@@ -6,7 +6,7 @@
 
 `npx {{module_name}}`
 
-- Define the package name to be installed, then execute the binaries:
+- Want to use multiple binaries, define the package name to be installed, then execute the binaries:
 
 `npx -p {{module_name}}`
 
--- a/pages/common/npx.md
+++ b/pages/common/npx.md
@@ -8,7 +8,7 @@
 
 - Want to use multiple binaries, define the package name to be installed, then execute the binaries:
 
-`npx -p {{module_name}}`
+`npx -p {{package_name}} {{module_name}}`
 
 - View help contents:
 
--- a/pages/common/npx.md
+++ b/pages/common/npx.md
@@ -1,0 +1,15 @@
+# npx
+
+> Execute binaries from npm packages.
+
+- Execute the command:
+
+`npx <command>[@version] [command-arg]`
+
+- Define the package to be installed, then execute the binaries:
+
+`npx [options] [-p|--package <pkg>]... <command> [command-arg]...`
+
+- Call help see other options meaning with local languange:
+
+`npx --help`
--- a/pages/common/npx.md
+++ b/pages/common/npx.md
@@ -1,12 +1,12 @@
 # npx
 
-> Execute binaries from `npm` packages(most of the cli tools) .
+> Execute binaries from `npm` packages.
 
-- Execute the npm packages binaries(Package will be installed if it does not exist in node_modules or $PATH):
+- Execute the binary from a given npm module:
 
 `npx {{module_name}}`
 
-- Want to use multiple binaries, define the package name to be installed, then execute the binaries:
+- In case a package has multiple binaries, specify the package name along with the binary:
 
 `npx -p {{package_name}} {{module_name}}`
 
--- a/pages/common/npx.md
+++ b/pages/common/npx.md
@@ -1,15 +1,15 @@
 # npx
 
-> Execute binaries from npm packages.
+> Execute  npm packages binaries.
 
-- Execute the command:
+- Execute the npm packages binaries:
 
-`npx <command>[@version] [command-arg]`
+`npx {{module_name}}`
 
 - Define the package to be installed, then execute the binaries:
 
-`npx [options] [-p|--package <pkg>]... <command> [command-arg]...`
+`npx -p {{module_name}}`
 
-- Call help see other options meaning with local languange:
+- View help contents:
 
 `npx --help`


pages/common/npx.md
```diff
@@ -0,0 +1,15 @@
+# npx
+
+> Execute binaries from npm packages.
+
+- Execute the command:
+
+`npx <command>[@version] [command-arg]`
```
agnivade: This is copied verbatim from the official description. We do not use syntax like <>. Please have a look at the CONTRIBUTING guidelines for the token syntax. 

And start with a simple example like `npm {{command}}` and then move to complex examples.

pages/common/npx.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Define the package to be installed, then execute the binaries:
+
+`npx [options] [-p|--package <pkg>]... <command> [command-arg]...`
+
+- Call help see other options meaning with local languange:
```
agnivade: This seems to be grammatically incorrect. And there is no need to add "local language". Perhaps "View help contents" is good enough.

pages/common/npx.md
```diff
@@ -0,0 +1,15 @@
...
+
+> Execute binaries from npm packages.
+
+- Execute the command:
+
+`npx <command>[@version] [command-arg]`
+
+- Define the package to be installed, then execute the binaries:
+
+`npx [options] [-p|--package <pkg>]... <command> [command-arg]...`
```
agnivade: This is not the syntax we use. Please refer to my comment above.

pages/common/npx.md
baixiaoji: sorry, I will try again

pages/common/npx.md
```diff
@@ -0,0 +1,15 @@
+# npx
+
+> Execute  npm packages binaries.
```
agnivade: How about 
```
Execute binaries from `npm` packages
```

pages/common/npx.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Execute the npm packages binaries:
```
agnivade:
```
Execute the installed npm module. (Package will be installed if it does not exist)
```

pages/common/npx.md
```diff
@@ -0,0 +1,15 @@
...
+
+`npx {{module_name}}`
+
+- Define the package to be installed, then execute the binaries:
+
+`npx -p {{module_name}}`
```
agnivade: I think you missed specifying the package parameter for `-p`.

pages/common/npx.md
```diff
@@ -0,0 +1,15 @@
+# npx
+
+> Execute  npm packages binaries.
+
+- Execute the npm packages binaries:
+
+`npx {{module_name}}`
+
+- Define the package to be installed, then execute the binaries:
```
agnivade:
```
Specify a given package name for an executable, in case the binary name does not match the package name
```

baixiaoji: sorry, you means  I can give a example to replace the module_name?

pages/common/npx.md
baixiaoji: Thx for you advice.

pages/common/npx.md
agnivade: The `-p` flag is necessary when a package has multiple binaries. Otherwise, the package name has to be the same as the binary name. Therefore, my description clarifies in which case you need a package_name and a module_name. Does that make sense ?

pages/common/npx.md
agnivade: @baixiaoji - Maybe you missed this comment ?

pages/common/npx.md
baixiaoji: sorry, I can't fully understand what's your saying, but I will try my best to description, wait monent.

pages/common/npx.md
baixiaoji: you can see the files changed, I especially add some explaination  for your comment.

agnivade: Yes, you changed the description. But you need to update the example to to specify the package name. Shouldn't it be like `npx -p {{package_name}} {{module_name}}` ? That is what I meant.

baixiaoji: you can see again

pages/common/npx.md
```diff
@@ -0,0 +1,15 @@
+# npx
+
+> Execute binaries from `npm` packages(most of the cli tools) .
```
agnivade: What is the use of "(most of the cli tools)" here ?

pages/common/npx.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Execute the npm packages binaries(Package will be installed if it does not exist in node_modules or $PATH):
```
agnivade: I thought `Execute the installed npm module.` sounded better. But if you want to keep your style, we can do `Execute the binary from a given npm module`. I have used module rather than package to match with the example below since you have used "module_name".

Also, give a space before "(". 

And I intentionally did not mention the node_modules or $PATH locations for simplicity here. Essentially, we like to keep things at a high level and not bother to go into details unless they are required to understand the command. I feel that just informing the user that the package will be installed if it is not, is good enough. There is not need to explain how it detects whether a package is installed or not. However, if you think that it is important, you can keep it.

pages/common/npx.md
```diff
@@ -0,0 +1,15 @@
...
+
+`npx {{module_name}}`
+
+- Want to use multiple binaries, define the package name to be installed, then execute the binaries:
```
agnivade: This sentence is not grammatically correct. How about `In case a package has multiple binaries, specify the package name along with the binary` ?

pages/common/npx.md
baixiaoji: to explain the packages, usually the packages is cli tools to excute like webpack

pages/common/npx.md
baixiaoji: ha, I make a mistake, want explain specific thing however it will more  details.

pages/common/npx.md
baixiaoji: Thx

pages/common/npx.md
```diff
@@ -0,0 +1,15 @@
+# npx
+
+> Execute binaries from `npm` packages.
+
+- Execute the binary from a given npm module:
+
+`npx {{module_name}}`
+
+- In case a package has multiple binaries, specify the package name along with the binary:
```
sbrl: What is meant by a package name here?

agnivade: The name of the npm package. Usually the package name is the same as the binary name. Sometimes, a package has multiple binaries, or the binary name is different from the package name. In that case, we need to differentiate between the two.

baixiaoji: So,it's ok ?

agnivade: Yes, I believe so. Unless @sbrl wants to clarify something further in the description.

baixiaoji: ok

sbrl: Ah, ok. Maybe it was the `module_name` that threw me because that's actually the binary name.

PR #1861, opened 2018-01-04
latexmk: merge linux/latexmk with common/latexmk

### LINKED ISSUE #1735: Merge linux/latexmk.md into common/latexmk.md
The common page was created earlier (#655), but seems to have been missed in the PR that introduced the linux page (#1428). /cc @85pando, @Nagiza.

agnivade:
Fixes #1735

----


pages/common/latexmk.md
```diff
@@ -15,9 +15,9 @@
 
 `latexmk -pdf {{source.tex}}`
 
-- Clean up all temporary tex files in the folder:
+- Force the generation of a document inspite of errors:
```
waldyrious: The correct spelling here would be "in spite of errors", but I think it would be clearer as "even if there are errors" or "disregarding any errors".

pages/linux/latexmk.md
```diff
@@ -1,19 +0,0 @@
...
-
-- Compile a LaTeX file into a PDF file:
-
-`latexmk -pdf {{filename.tex}}`
-
-- Force `latexmk` to generate a document even if there are errors:
-
-`latexmk -f {{filename.tex}}`
-
-- Clean up the current directory from all files generated by `latexmk` except DVI, PostScript and PDF:
```
waldyrious: This description is a bit verbose, but maybe a compromise with the shorter "Clean up temporary tex and output files" could be found. How about "Clean up all temporary files in the current directory"?

agnivade: Umm .. this page is removed actually. Look at the common page. Since it had `-c {{source.tex}}` and `-C`, simply keeping `-c` was kinda pointless.

waldyrious: Oh, I didn't notice the capitalization difference. That explains the "and output files" in the page being kept -- that confused me. So, with that cleared up, I'd still say `-c` is more useful than `-C`. It's more common to want to clean up the folder (tex processing produces lots of intermediate files) than to remove everything altogether, including the final outputs.

(Btw, I commented here because github doesn't allow commenting on unchanged context lines.)

PR #1864, opened 2018-01-06
cpufreq-set: add page

jsonbruce:
----

### DIFF
--- a/pages/linux/cpufreq-set.md
+++ b/pages/linux/cpufreq-set.md
@@ -1,20 +1,20 @@
 # cpufreq-set
 
 > A tool to modify CPU frequency settings.
-> Requires root privileges.
+> New frequency value range between the output of command `cpufreq-info -l`.
 
 - Set the CPU frequency policy of CPU 1 to "userspace":
 
 `sudo cpufreq-set -c {{1}} -g {{userspace}}`
 
-- Set the current minimum CPU frequency of CPU 1, value range between the output of command `cpufreq-info -l`:
+- Set the current minimum CPU frequency of CPU 1:
 
 `sudo cpufreq-set -c {{1}} --min {{min_frequency}}`
 
-- Set the current maximum CPU frequency of CPU 1, value range between the output of command `cpufreq-info -l`:
+- Set the current maximum CPU frequency of CPU 1:
 
 `sudo cpufreq-set -c {{1}} --max {{max_frequency}}`
 
-- Set the current work frequency of CPU 1 , value range between the output of command `cpufreq-info -l` (requires "userspace" policy to be loaded):
+- Set the current work frequency of CPU 1:
 
 `sudo cpufreq-set -c {{1}} -f {{work_frequency}}`
--- a/pages/linux/cpufreq-set.md
+++ b/pages/linux/cpufreq-set.md
@@ -7,14 +7,14 @@
 
 `sudo cpufreq-set -c {{1}} -g {{userspace}}`
 
-- Set the current minimum CPU frequency of CPU 1, value range between the output of command "cpufreq-info -l":
+- Set the current minimum CPU frequency of CPU 1, value range between the output of command `cpufreq-info -l`:
 
 `sudo cpufreq-set -c {{1}} --min {{min_frequency}}`
 
-- Set the current maximum CPU frequency of CPU 1, value range between the output of command "cpufreq-info -l":
+- Set the current maximum CPU frequency of CPU 1, value range between the output of command `cpufreq-info -l`:
 
 `sudo cpufreq-set -c {{1}} --max {{max_frequency}}`
 
-- Set the current work frequency of CPU 1 , value range between the output of command "cpufreq-info -l" (requires "userspace" policy to be loaded):
+- Set the current work frequency of CPU 1 , value range between the output of command `cpufreq-info -l` (requires "userspace" policy to be loaded):
 
 `sudo cpufreq-set -c {{1}} -f {{work_frequency}}`
--- a/pages/linux/cpufreq-set.md
+++ b/pages/linux/cpufreq-set.md
@@ -1,7 +1,7 @@
 # cpufreq-set
 
 > A tool to modify CPU frequency settings.
-> New frequency value range between the output of command `cpufreq-info -l`.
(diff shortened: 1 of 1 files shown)


jsonbruce:
I think the concrete values maybe not suit for other people's computer, as we often run the examples showed by `tldr` directly.
So I used the placeholder.

agnivade:
Users are not expected to blindly copy paste the examples. The reason I ask for concrete values is because if I have to run the command, I will have to search for the value to be set. But if I am shown a value, then at least I have some idea around what can be set as a value.

If you think that the downside of providing a concrete value is more than the usefulness it provides, then we can go with a placeholder.

jsonbruce:
Yes, I can understand this kind of use-case.
But for commands like `cpufreq-set`, concrete values may bring more confusion, as it has value range.

I've added value range part to the command description.

=======

I tested these two commands again: `cpufreq-info` and `cpufreq-set`.

### Step 1
The command `cpufreq-info -c 1 -l` shows the allowed minimum and maximum CPU frequency of core 1.
```
➜  cpufreq-info -c 1 -l
1200000 2501000
```
We can only change the **current minimum frequency**, **current maximum frequency** and **current work frequency** between this range.

### Step 2
The command `cpufreq-info -o` show the current minimum and maximum CPU frequency of all cores.
```
➜  cpufreq-info -o
          minimum CPU frequency  -  maximum CPU frequency  -  governor
CPU  0      1200000 kHz ( 47 %)  -    2501000 kHz (100 %)  -  userspace
CPU  1      1200000 kHz ( 47 %)  -    2501000 kHz (100 %)  -  userspace
CPU  2      1200000 kHz ( 47 %)  -    2501000 kHz (100 %)  -  performance
CPU  3      1200000 kHz ( 47 %)  -    2501000 kHz (100 %)  -  performance
CPU  4      1200000 kHz ( 47 %)  -    2501000 kHz (100 %)  -  ondemand
CPU  5      1200000 kHz ( 47 %)  -    2501000 kHz (100 %)  -  ondemand
CPU  6      1200000 kHz ( 47 %)  -    2501000 kHz (100 %)  -  conservative
CPU  7      1200000 kHz ( 47 %)  -    2501000 kHz (100 %)  -  conservative
CPU  8      1200000 kHz ( 47 %)  -    2501000 kHz (100 %)  -  powersave
CPU  9      1200000 kHz ( 47 %)  -    2501000 kHz (100 %)  -  powersave
```

### Step 2
The command `cpufreq-info -c 1 -f` shows the current work frequency of core 1.
```
➜  cpufreq-info -c 1 -f
2501000
```

### Step 4
The command `cpufreq-set -c 1 --min 1200001 --max 2500000` changes the current min-frequency and max-frequency of core 1 to given value.

Then let's see the result:
```
➜  cpufreq-info -o
          minimum CPU frequency  -  maximum CPU frequency  -  governor
CPU  0      1200000 kHz ( 47 %)  -    2501000 kHz (100 %)  -  userspace
CPU  1      1200001 kHz ( 47 %)  -    2500000 kHz ( 99 %)  -  userspace
CPU  2      1200000 kHz ( 47 %)  -    2501000 kHz (100 %)  -  performance
```

**Notice**: We can do single digit modification. Like `1200000` to `1200001`.

### Step 5
The command `cpufreq-set -c 1 -f 2400001` changes the current work frequency of core 1 to given value.

Then let's see the result:
```
➜  ~ cpufreq-info -c 1 -f
2500000
```

**Notice**: The single digit modification will be rounded up `GHz` unit. (default unit is `kHz`, 1 GHz=10^6 kHz).
Like `2400001` will be rounded up to `2500000`

So we can use command `cpufreq-set -c 1 -f 2.4GHz` to set work frequency.


======

So I'll update the documentation `cpufreq-info.md` too.

pages/linux/cpufreq-set.md
```diff
@@ -0,0 +1,20 @@
+# cpufreq-set
+
+> A tool to modify CPU frequency settings.
+> Requires root privileges.
+
+- Set the CPU frequency policy of CPU 1 to "userspace":
+
+`sudo cpufreq-set -c {{1}} -g {{userspace}}`
+
+- Set the current minimum CPU frequency of CPU 1, value range between the output of command `cpufreq-info -l`:
```
agnivade: My only concern is that the "value range between the output of command" is repeated 3 times. Is there any way we can just mention it once ? Maybe put this as the subheading and remove the "Requires root privileges" part since `sudo` is already there in the example ?

jsonbruce:
My review stuck on `pending` @agnivade

image

agnivade:
If we try to add each and every guideline in the style guide, it will become huge and deter newcomers from reading it. Some things are fine as an unspoken rule.

For some commands, "Requires root privileges" are fine by me. 

Yes, the text in the heading and sub-heading applies to all examples in the page. But there is nothing wrong in mentioning that one has to run `cpufreq-info -l` to get the CPU freq range. It is still a valid statement and the user can figure out that for the commands that need a CPU freq value, one has to run this command.

jsonbruce:
> Make Problem More Simple and Clear.

OK. Let's do this.

jsonbruce:
LGTM

agnivade:
@jsonbruce - We usually either rebase or squash when merging a PR. https://github.com/tldr-pages/tldr/blob/master/contributing-guides/maintainers-guide.md#maintainers-guide.

jsonbruce:
> When merging PRs, use the strategy that produces a clean git history: Use squash if there's a single commit in the PR, or if the multiple commits are not independent changes. If the PR author took the time to craft individual, informative commit messages for each commit, use rebase to honor that work and preserve the history of the changes.
> 
> A simple heuristic to follow is that if there are more "dirty" commits than "clean" commits, then prefer squash, else do a rebase.
> 

For learning:
According to this rule, I should use `rebase` in this case, right? @agnivade

agnivade:
In my thinking, 6 commits doing very minor changes like wrapping with backticks and adding value ranges are "dirty" commits and I usually squash them. But the choice is yours. Go with what you feel is the best. But avoid merge.

jsonbruce:
Copy that. @agnivade 

I've been thinking `dirty commits` for a while about how to define it.  :smile:

PR #1869, opened 2018-01-08
grep: add usage fore grep and find for content searching

ycmjason:
----

- [ ] ~The page (if new), does not already exist in the repo.~

- [ ] ~The page (if new), has been added to the correct platform folder:  


agnivade:
Hey @ycmjason, thanks for the contribution.

Unfortunately, we have a hard limit of 8 for the no. of examples that we allow in a page. This was written explicitly in the PR template. The reason is to challenge ourselves to provide the most useful set of examples in a short space. For a complete documentation, there is always the man page.
 
The `grep` and `find` pages have gone through a number of iterations over the years. I would rather let them remain as is.

PR #1879, opened 2018-01-10
hg-log: add page

pxgamer:
----

### DIFF
--- a/pages/common/hg-log.md
+++ b/pages/common/hg-log.md
@@ -10,11 +10,11 @@
 
 `hg log --graph`
 
-- Display the revision history with names matching a specified pattern:
+- Display the revision history with file names matching a specified pattern:
 
 `hg log --include {{pattern}}`
 
-- Display the revision history, excluding names that match a specified pattern:
+- Display the revision history, excluding file names that match a specified pattern:
 
 `hg log --exclude {{pattern}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/hg-log.md
```diff
@@ -0,0 +1,35 @@
+# hg log
+
+> Display the revision history of the repository.
+
+- Display the entire history of the repository:
```
sbrl: Perhaps `revision history` instead of `history` to keep it consistent would be better?

pxgamer: Sounds good 👍

pages/common/hg-log.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Display the entire revision history of the repository:
+
+`hg log`
+
+- Display the revision history with an ASCII graph:
+
+`hg log --graph`
+
+- Display the revision history with names matching a specified pattern:
```
agnivade: Should we say "file names" instead of names to make it more clear ? Or do you think its good enough ?

pages/common/hg-log.md
```diff
@@ -0,0 +1,35 @@
...
+
+`hg log --include {{pattern}}`
+
+- Display the revision history, excluding names that match a specified pattern:
```
agnivade: Same here.

pages/common/hg-log.md
pxgamer: Sounds good on both. :+1:

PR #1887, opened 2018-01-12
maintainers-guide.md: Suggest to use rebase rather than merge

agnivade:
/cc @waldyrious

### DIFF
--- a/contributing-guides/maintainers-guide.md
+++ b/contributing-guides/maintainers-guide.md
@@ -25,7 +25,7 @@ for the behavior expected of tldr-pages maintainers:
   or if the multiple commits are not independent changes.
   If the PR author took the time to craft individual,
   informative commit messages for each commit,
-  use regular merge to honor that work and preserve the history of the changes.
+  use rebase to honor that work and preserve the history of the changes.
 
 - **Know when and how to say no**.
   Sometimes requests or contributions need to be declined,
--- a/contributing-guides/maintainers-guide.md
+++ b/contributing-guides/maintainers-guide.md
@@ -27,6 +27,9 @@ for the behavior expected of tldr-pages maintainers:
   informative commit messages for each commit,
   use rebase to honor that work and preserve the history of the changes.
 
+  A simple heuristic to follow is that if there are more "dirty" commits
+  than "clean" commits, then prefer squash, else do a rebase.
+
 - **Know when and how to say no**.
   Sometimes requests or contributions need to be declined,
   at least in their current form.


waldyrious:
Indeed I used to prefer rebases, which make the git history tidier, but now I believe a merge is more useful, for various reasons:

- If coupled with a rebase, it keeps the git history reasonably clean but not to the point of losing information about which commits were part of the same changeset (pull request). I learned recently that this strategy is called "semi-linear history". As mentioned, it requires a rebase followed by a merge (to avoid interleaving branches), but since that's not yet possible via the GitHub web interface (I requested just that at https://github.com/isaacs/github/issues/1017), I chose to skip the requirement of a rebase, to prevent making the PR merging workflow too cumbersome for maintainers. (Of course, maintainers are still free to do the rebase first via the command line, if they don't mind the extra work)
- It creates a commit for each PR merged, that automatically includes a reference to the PR number, which allows frictionless cross-referencing.
- Since the merge commit is authored by the PR merger (rather than the PR author), that provides a way to check how many merges a maintainer has performed (which is one of the criteria we decided to use to measure maintenance activity).

But I might be overlooking some aspect that makes rebases better than merges. Open to hear other opinions :)

agnivade:
Hmm .. I hadn't thought about the information it retains about who merges the commit. But in that case, I believe we have to be consistent, because if you squash commits then that info is lost. So judging maintainer activity from commits wont be very accurate if we squash sometimes and merge sometimes.

Anyway, I still feel that the mess of commit history that it creates, overpowers the info retained in using merge rather than rebase. My vote is still with rebase.

But since this is not a code repo, and `git bisect`s are not a common occurrence here, if you want to keep merge behavior, I am okay.

waldyrious:
> But in that case, I believe we have to be consistent, because if you squash commits then that info is lost.

Agreed. We could suggest merging even if there is a single commit to always preserve the branch structure. The motivation behind the current wording was to allow preserving the information about commit grouping without increasing the burden to maintainers. GitHub only allows commits to be merged, squashed or rebased, so any other strategy means having to go to the command line, which introduces friction, and we don't want to ask that regularly of maintainers.

To help reason about this issue, I decided to make a table to cover all of the possible scenarios and where GitHub's PR closing modes fit, for the option of (1) discarding branch structure info and having a purely linear history:

| | PR branch base matches `master` at time of merge | PR branch base behind `master` at time of merge |
| --- | --- | --- |
| (A) one commit in the PR | GH: squash | GH: rebase |
| (B) multiple commits, all of which should be combined | GH: squash | GH: squash |
| (C) multiple commits, some of which should be combined | CLI: `rebase -i && merge` | CLI: `rebase -i && merge` |
| (D) multiple commits, none of which should be combined | GH: rebase | GH: rebase |

... and (2) preserving branch structure with a semi-linear history:

| | PR branch base matches `master` at time of merge | PR branch base behind `master` at time of merge |
| --- | --- | --- |
| (A) one commit in the PR | GH: merge | CLI: `rebase && merge --no-ff` |
| (B) multiple commits, all of which should be combined | CLI: `rebase -i && merge --no-ff` | CLI: `rebase -i && merge --no-ff` |
| (C) multiple commits, some of which should be combined | CLI: `rebase -i && merge --no-ff` | CLI: `rebase -i && merge --no-ff` |
| (D) multiple commits, none of which should be combined | GH: merge | CLI: `rebase -i && merge --no-ff` |

Scenario (C) is obviously the one that complicates everything the most (it also includes the case where commits need to be reworded, reordered, etc.). I'd say we could define, as a baseline, a heuristic of defaulting to the (B) resolutions if there are more "dirty" commits than "clean" commits, and to the (D) actions otherwise. And if the maintainer feels like it, they can go to the command line and clean the branch neatly.

So with that, we can see that option (1) can always be done from the GH's web interface, while (2) currently requires using the command line pretty much all the time (at least until isaacs/github#1017 is fixed).

For that reason, although I still prefer the semi-linear graph structure, I recognize that it isn't practical for our GitHub-based project yet, and thus agree with your proposal to change the guidelines to use rebase instead of merge. :+1:

waldyrious:
Please include the heuristic I mentioned above for scenario (C), for completeness.

sbrl:
That table complicates things. Is it going to be part of the maintainers guide too? I think it might complicate things a bit too much.

I've heard multiple people talk about 'keeping the commit history clean' in multiple different repos, but I'm not sure I understand. Isn't the commit history of a branch always linear?

agnivade:
> I think it might complicate things a bit too much.

We will just add one line - " a heuristic of defaulting to the (B) resolutions if there are more "dirty" commits than "clean" commits, and to the (D) actions otherwise", to decide when to squash and when to rebase. I think it's an important point to have in the guide. 

>  Isn't the commit history of a branch always linear?

Not really. If you have `gitk` installed, you will see the real picture. Its a mess of lines cris-crossing each other. Just go to the tldr repo and do a `gitk`.

waldyrious:
For future reference: I've just disabled merge commits in the repository settings, to reflect the decision we made here (and now included in `maintainers-guide.md`).

agnivade:
Ah great, I didn't bother to check the settings. This is perfect !

waldyrious:
Yeah, I also forgot it could be enforced there :)

PR #1891, opened 2018-01-13
feh: add page

fuerbringer:
A tldr-page for the [feh](https://feh.finalrewind.org/) image utility  ([GitHub](https://github.com/derf/feh)). I've added it to `common` since it supports GNU/Linux, FreeBSD and OpenBSD.

----


agnivade:
Hmm .. we seriously need to create a `unix-common` platform. `common` includes windows too. @fuerbringer - please create copies of the page under `linux` and `macOS`.

agnivade:
I apologize for not being clear. The page should be removed from common and only be in linux and macOS. Since, it is not supported in windows.

agnivade:
The problems are that 

1. We have to move all the pages that separately lie in linux and osx folders to unix-common. And also there are several pages that mistakenly lie in common, which should be in unix-common. Those need to be moved too.

2. All the clients need to be updated. Since it's a breaking change and if the clients don't look for the new directory, they won't return any result for pages which actually exist.

PR #1892, opened 2018-01-13
hg-serve: add page

pxgamer:
----

### DIFF
--- a/pages/common/hg-serve.md
+++ b/pages/common/hg-serve.md
@@ -18,7 +18,7 @@
 
 `hg serve --name {{name}}`
 
-- Start a web server instance using the specified theme:
+- Start a web server instance using the specified theme (see the templates directory):
 
 `hg serve --style {{style}}`
 
--- a/pages/common/hg-serve.md
+++ b/pages/common/hg-serve.md
@@ -22,6 +22,6 @@
 
 `hg serve --style {{style}}`
 
-- Start a web server instance using the specified SSL certificate:
+- Start a web server instance using the specified SSL certificate bundle:
 
 `hg serve --certificate {{path/to/certificate}}`
--- a/pages/common/hg-serve.md
+++ b/pages/common/hg-serve.md
@@ -1,0 +1,27 @@
+# hg serve
+
+> Start a standalone Mercurial web server for browsing repositories.
+
+- Start a web server instance:
+
+`hg serve`
+
+- Start a web server instance on the specified port:
+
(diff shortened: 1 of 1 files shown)


pages/common/hg-serve.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Start a web server instance with a specific name:
+
+`hg serve --name {{name}}`
+
+- Start a web server instance using the specified theme:
+
+`hg serve --style {{style}}`
+
+- Start a web server instance using the specified SSL certificate:
```
agnivade: Where would one specify the ssl key ?

pages/common/hg-serve.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Start a web server instance on the specified listening address:
+
+`hg serve --address {{address}}`
+
+- Start a web server instance with a specific name:
+
+`hg serve --name {{name}}`
+
+- Start a web server instance using the specified theme:
```
agnivade: Is there a predefined list of themes one can use ? Where do I get the list of themes ?

pages/common/hg-serve.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Start a web server instance on the specified port:
+
+`hg serve --port {{port}}`
+
+- Start a web server instance on the specified listening address:
+
+`hg serve --address {{address}}`
+
+- Start a web server instance with a specific name:
```
agnivade: What does it mean for a web server to have a name ?

pxgamer: The name is what appears in the title of the pages. And also it is shown in tabs.

I wasn't quite sure of a good way to shorten it.

pages/common/hg-serve.md
pxgamer: It is predefined, but you can also specify a custom theme if it's added to the template directory.

The downside is that there is no way (as far as I know) to list themes, other than to do an `ls` in the themes directory.

agnivade: Ok, then perhaps we can mention that in brackets.

pages/common/hg-serve.md
agnivade: How about "identifier" or "label" ?

pages/common/hg-serve.md
pxgamer: It looks like it needs to be a bundled `.PEM` file containing both the key and the certificate.

PR #1896, opened 2018-01-15
ps: get the parent pid

zlbabe:


jsonbruce:
I love this feature. Thanks! @zlbabe

Issue #1878, opened 2018-01-10
Role transitions to match new governance guidelines


waldyrious:
Now that the project finally has ~~[explicit criteria for role transitions](https://github.com/tldr-pages/tldr/blob/master/GOVERNANCE.md#ii-role-transitions)~~ ([updated link](https://github.com/tldr-pages/tldr/blob/master/COMMUNITY-ROLES.md#when-to-change-roles) after [#1897](https://github.com/tldr-pages/tldr/pull/1897#issuecomment-358395036)), the first thing we need to do is update the current contributors' roles to match them.

Here are the changes that should be done, according to the document:

- [x] a) **Contributors to be added as collaborators in this repo**: actually, I'd rather not collate this particular list myself, since that would entail opening PRs which are still on my unread notifications list, and would therefore disappear from it without the review I'd wish to give them. So I'll just describe the process and ask for someone else to step in. What I did was go to https://github.com/tldr-pages/tldr/graphs/contributors and click on the "x commits" links where x >= 5, starting at the bottom. Then I checked whether the commits actually corresponded to 5 or more distinct, non-trivial PRs, or whether some of them were multiple commits from the same PR. One example of the latter is (at the time of writing) `@Larry850806`. See also `@chuanconggao` as an example of someone with 5 merged PRs but one of them being a trivial one (#642). The first user who matches the criteria (again, counting from the bottom) is @zdroid (PRs #1101, #1110, #1011, #1090, #1091), so he'd be one of those to be added as a collaborator. I'd appreciate if someone else could repeat this process for the other contributors and complete this list; otherwise, that's going to have to wait until I'm done clearing my notification backlog.

- [x] b) **Collaborators to be added as members of the organization**: Nothing to do here, unless I'm missing something. @jeeftor (added after the exchanges [here](https://github.com/tldr-pages/tldr/issues/1209#issuecomment-280909917) and [here](https://gitter.im/tldr-pages/tldr?at=58ab24e7f045df0a2270cec9)) is the only collaborator in the repo who's not a member of the org, and I believe he hasn't showed interest in performing maintenance tasks. Please correct me if I'm wrong.

[...]

PR #1899, opened 2018-01-17
Using new travis token

agnivade:
Used my token here.

Travis seems to be down. Will test it once it's up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,8 @@
 sudo: false
-
 language: node_js
 
 node_js:
-  - 'stable'
-
+- stable
 cache: false
 
 after_success:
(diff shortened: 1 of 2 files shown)


agnivade:
xref #1878

waldyrious:
Are you planning to remove the other token?

Also, don't forget the documentation :) from https://github.com/tldr-pages/tldr/issues/1878#issuecomment-356573527:

> we definitely should add a comment to that file noting the token on .travis.yml, and the comment on `.travis.yml` should mention the `upload_assets()` function on this file.

agnivade:
First I wanted to verify that this thing works. :) Will update the documentation once its working.

Yes, there can only be one token at a time.

waldyrious:
Cool. Btw, looks like Travis is working well again, so I've restarted the build. Let's wait and see!

agnivade:
You would only be able to test it once you merge with master. Yes, now it works !

waldyrious:
Right, forgot about that detail :P

Issue #1579, opened 2017-10-25
Let's get the new windows section started!


sbrl:
As a sort of celebration of the new `windows` platform folder, here's a list of 10 useful commands that are specific to windows machines.

 - [x] sfc
 - [x] sigverif
 - [x] driverquery
 - [x] nslookup
 - [x] ping
 - [x] pathping
 - [x] ipconfig
 - [x] repair-bde
 - [x] tasklist
 - [x] taskkill

_From [this tech republic article](http://www.techrepublic.com/blog/10-things/10-windows-7-commands-every-administrator-should-know/)_

itsjzt:
some more are 

- [ ] cls
- [ ] color
- [ ] set
- [ ]  systeminfo
- [ ] tree
- [ ] wmic

Leandros:
You forgot the most import ones!

- [ ] dir
- [ ] xcopy
- [ ] rmdir
- [ ] del

and the most important one:

- [ ] How to install gnu coreutils

sbrl:
Good thinking, @Leandros! Not sure about `How to install gnu coreutils` though - unless that's a command.

Leandros:
More or less a joke. The gnu coreutils are all available on Windows, and work lovely. Often better than their Windows "native" counterparts.

sbrl:
Ah, ok! Yeah, they are really good, aren't they?

_(Sorry, I sometimes miss jokes :P)_

pxgamer:
@sbrl, the `sigverif` command is just a GUI and has no options. Should that still be added?

Leandros:
Added a couple more tools, including the super useful `doskey`.

sbrl:
@pxgamer Hrm, good question. If it has no options at all, then I'm not sure it would be particularly useful as a page - except to ensure people that its behaviour can be altered through the use of flags. Someone can go ahead if they want I guess!

@agnivade, thoughts?

agnivade:
We do have a lot of pages for GUI. Check `rainbowstream` and `rtv`. Originating from this issue - https://github.com/tldr-pages/tldr/issues/1162. 

We can go ahead with `sigverify`.

pxgamer:
@agnivade, fine with me. 👍

agnivade:
We just have one page [wmic](https://docs.microsoft.com/en-us/windows/desktop/WmiSdk/wmic) left to close this issue. Anybody wants to take a stab ?

pxgamer:
[...]

PR #1901, opened 2018-01-17
color: add page

pxgamer:
Part of #1579

----

### DIFF
--- a/pages/windows/color.md
+++ b/pages/windows/color.md
@@ -1,0 +1,15 @@
+# color
+
+> Set the console foreground and background colors.
+
+- Set the console colors to the default values:
+
+`color`
+
+- Set the console foreground and background to a specific color:
+
+`color {{color_hex}}`
+
+- List available color values and detailed information:
+
+`color /?`
--- a/pages/windows/color.md
+++ b/pages/windows/color.md
@@ -8,7 +8,7 @@
 
 - Set the console foreground and background to a specific color:
 
-`color {{color_hex}}`
+`color {{foreground_code}}{{background_code}}`
 
 - List available color values and detailed information:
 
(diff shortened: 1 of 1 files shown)


pages/windows/color.md
```diff
@@ -0,0 +1,15 @@
+# color
+
+> Set the console foreground and background colors.
+
+- Set the console colors to the default values:
+
+`color`
+
+- Set the console foreground and background to a specific color:
```
agnivade: So this sets both the foreground and background to the same color ? That will be weird. Since you won't be able to differentiate.

pages/windows/color.md
```diff
@@ -0,0 +1,15 @@
...
+
+`color {{color_hex}}`
```
agnivade: Is the "#" needed ? I suggest `color {{hex_code}}`

pages/windows/color.md
```diff
@@ -0,0 +1,15 @@
...
+
+- List available color values and detailed information:
```
agnivade: This should be above `color {{hex_code}}` as the users will need to know what values are present to be able to set a color.

pages/windows/color.md
pxgamer: You provide it with 2 digits, such as `70` which will have a foreground colour of white on a black background.

pages/windows/color.md
pxgamer: I guess it's slightly misleading, it's not a standard hex colour. It's 2 hex digits.

0 - F are different colours and a combination of 2 of them (such as `70` or `F3`) is what you provide.

pages/windows/color.md
pxgamer: That's a good point. I'll move it above.

pages/windows/color.md
pxgamer: I guess that would be fine actually. :+1:

Even though it's a bit longer, it shows how the command actually works.

PR #1902, opened 2018-01-18
Added documentation about travis token

agnivade:
/cc @waldyrious

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -28,6 +28,7 @@ function build_archive {
 }
 
 function upload_assets {
+  # ${GH_TOKEN} is referred from the secure variable inside .travis.yml
   git clone --quiet --depth 1 https://${GH_TOKEN}@${SITE_URL} $SITE_HOME
   mv -f $TLDR_ARCHIVE $SITE_HOME/assets/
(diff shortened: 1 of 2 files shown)


waldyrious:
Nice :+1:

Please also mention that the upload is done via https://github.com/tldr-pages/tldr/blob/master/scripts/build.sh#L31, and add a comment to that file as well with similar documentation.

waldyrious:
For future reference, this PR is a follow-up to #1899.

agnivade:
@waldyrious - PTAL

scripts/build.sh
```diff
@@ -28,6 +28,7 @@ function build_archive {
 }
 
 function upload_assets {
+  # ${GH_TOKEN} is referred from the secure variable inside .travis.yml
```
waldyrious: How about "referred from the secure variable inside .travis.yml" --> "defined as a secure variable in .travis.yml"?

waldyrious: Also, I'd add a comment at the top of this file, saying something like:

    # This script is executed by Travis CI when a PR is merged (i.e. in the `after_success` step).

agnivade: Done

PR #1904, opened 2018-01-18
taskkill: add page

pxgamer:
Part of #1579

----

### DIFF
--- a/pages/windows/taskkill.md
+++ b/pages/windows/taskkill.md
@@ -12,15 +12,15 @@
 
 - Forcefully terminate a specified process (must be combined with a process id or name):
 
-`taskkill /f`
+`taskkill /pid {{process_id}} /f`
 
 - Terminate a process and its child processes (must be combined with a process id or name):
 
-`taskkill /t`
+`taskkill /im {{process_name}} /t`
 
 - Terminate a process on a remote machine:
 
-`taskkill /s {{remote_name}}`
+`taskkill /pid {{process_id}} /s {{remote_name}}`
 
 - Display information about the usage of the command:
 
--- a/pages/windows/taskkill.md
+++ b/pages/windows/taskkill.md
@@ -1,16 +1,16 @@
 # taskkill
 
-> Terminate a process by it's process id or name.
+> Terminate a process by its process id or name.
 
 - Display information about the usage of the command:
 
 `taskkill /?`
 
-- Terminate a process by it's id:
+- Terminate a process by its id:
 
 `taskkill /pid {{process_id}}`
 
-- Terminate a process by it's name:
+- Terminate a process by its name:
 
 `taskkill /im {{process_name}}`
 
@@ -18,7 +18,7 @@
 
 `taskkill /f`
 
-- Terminate a process and it's child processes:
+- Terminate a process and its child processes:
 
 `taskkill /t`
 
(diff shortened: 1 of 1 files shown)


pages/windows/taskkill.md
```diff
@@ -0,0 +1,27 @@
...
+
+`taskkill /pid {{process_id}}`
+
+- Terminate a process by its name:
+
+`taskkill /im {{process_name}}`
+
+- Forcefully terminate a specified process:
+
+`taskkill /f`
```
sbrl: How do you specify the process to terminate? Perhaps we could merge this with one of the other examples?

pages/windows/taskkill.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Terminate a process and its child processes:
+
+`taskkill /t`
```
sbrl: Perhaps we could add `/pid {{process_id}}` to this or something for the same reason?

pages/windows/taskkill.md
pxgamer: I did think about that, but I wasn't sure as it covers the `/pid` and `/im` earlier. I didn't really want to duplicate stuff.

I guess could it mention that it needs to be combined with a process id or name?

sbrl: Yeah, probably. Unless we combine the two examples, I guess?

pxgamer: Could I add `(must be combined with a process id or name)` to the end of both descriptions?

agnivade: I think it's ok. It is natural to combine the options from previous examples to have more complex examples and is something the users are expected to do. If you say `/f` and `/pid` and `/im` are mentioned earlier, then that means `/f` can be combined with either of them.

pages/windows/taskkill.md
```diff
@@ -0,0 +1,27 @@
+# taskkill
+
+> Terminate a process by its process id or name.
+
+- Display information about the usage of the command:
```
agnivade: Move this to the end.

TIP: Arrange the examples in the order you expect the users to use them. Usually we keep the `-h` and its variants at the end, as those are expected to let the user know about further usages of the command.

pages/windows/taskkill.md
agnivade: Sorry, I think I was not clear. I meant that it was okay as it was before, i.e. without the `(must be combined with a process id or name)`. But this is OK too.

pages/windows/taskkill.md
```diff
@@ -0,0 +1,27 @@
...
+
+`taskkill /f`
+
+- Terminate a process and its child processes (must be combined with a process id or name):
+
+`taskkill /t`
+
+- Terminate a process on a remote machine:
+
+`taskkill /s {{remote_name}}`
```
agnivade: Shouldn't we need to give the process id or name here too ?

pxgamer: Yes, it can be combined with any of the above commands, so I wasn't sure if it needed to mention it.

agnivade: Dunno, looks confusing to me. Mentioning it thrice is also redundant. Best way is to go with @sbrl's original recommendation. Use `/pid` for `/f`, `/im` for `/t`, and again `/pid` for `/s`. That wil clearly indicate that pid or process_name can be mixed. And remove the text in `()`.

pxgamer: Sounds good to me. :+1: I'll make that change in a bit.

PR #1907, opened 2018-01-19
git-stash: Update deprecated commands

jaseemabid:
"git stash save" has been deprecated in favour of "git stash push" since
Git 2.15

Ref: https://github.com/git/git/blob/master/Documentation/RelNotes/2.16.0.txt#L34

----

### DIFF
--- a/pages/common/git-stash.md
+++ b/pages/common/git-stash.md
@@ -4,7 +4,7 @@
 
 - Stash current changes, except new (untracked) files:
 
-`git stash [save {{optional_stash_message}}]`
+`git stash [push {{optional_stash_message}}]`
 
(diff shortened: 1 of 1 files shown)


agnivade:
Ubuntu 16.04 upstream is still quite behind at 2.7.4 (https://packages.ubuntu.com/xenial/git-core). Brew is latest (http://formulae.brew.sh/formula/git) because it compiles from the git repo itself. 

I would rather not change it unless ubuntu repos have caught up. Thoughts ?

sbrl:
Perhaps we could include a note in the description saying to use `save` for older versions of git?

agnivade:
We can do that. Also mention the version no. 2.16.

PR #1908, opened 2018-01-19
reg: add page

pxgamer:
Part of #1579

This command will be split across pages for the following commands:

- reg query (#1909)
- reg add (#1910)
- reg delete (#1911)
- reg copy (#1912)
- reg save (#1913)
- reg restore (#1914)
- reg load (#1915)
- reg unload (#1916)
- reg compare (#1923)
- reg export
- reg import
- reg flags

----

- [x] The page (if new), does not already exist in the repo.

- [x] The page (if new), has been added to the correct platform folder:  
      `common/` if it's common to all platforms, `linux/` if it's Linux-specific, and so on.

- [x] The page has 8 or fewer examples.

- [x] The PR is appropriately titled:  
      `<command name>: add page` for new pages, or `<command name>: <description of changes>` for pages being edited

- [x] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines

PR #1911, opened 2018-01-19
reg-delete: add page

pxgamer:
Part of #1908

----

### DIFF
--- a/pages/windows/reg-delete.md
+++ b/pages/windows/reg-delete.md
@@ -1,0 +1,19 @@
+# reg delete
+
+> Delete keys or their values from the registry.
+
+- Delete a specific registry key:
+
+`reg delete {{key_name}}`
+
+- Delete a value under a specific key:
+
+`reg delete {{key_name}} /v {{value}}`
+
+- Delete all values recursively under the specified key:
+
+`reg delete {{key_name}} /va`
+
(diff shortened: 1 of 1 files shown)


pages/windows/reg-delete.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Delete a value under a specific key:
+
+`reg delete {{key_name}} /v {{value}}`
+
+- Delete all values recursively under the specified key:
+
+`reg delete {{key_name}} /va`
+
+- Forcefully deletes the key or value without a prompt:
```
agnivade: Hmm .. you mention "key or value". I thought values were data under a key. So will this be the same as `/va` ? But if you want to delete a particular value, don't you need to specify the parent key also ?

pxgamer: It's mainly to be combined with any of the other commands above.

So to delete a value, you'd combine it with `/v {{value}}`

agnivade: Understood. But the description does not match with the example here, you mention "key or value" but the example just has key_name. Maybe it's better if you do a `/f /va` and mention that `Forcefully delete all values recursively under a key`. That would indicate that `/f` can be combined with `/v {{value}}` too.

PR #1919, opened 2018-01-20
dmesg: show kernel error messages

zlbabe:
Show only kernel error messages


pages/linux/dmesg.md
```diff
@@ -17,3 +17,7 @@
 - Show kernel messages 1 page at a time:
 
 `dmesg | less`
+
+- Show kernel error messages:
```
agnivade: Please move this to the 2nd example.

PR #1926, opened 2018-01-23
top: adding memory options

alexandresardinha:
----

### DIFF
--- a/pages/linux/top.md
+++ b/pages/linux/top.md
@@ -18,6 +18,10 @@
 
 `top -p $(pgrep -d ',' {{process_name}})`
 
+- Sort by memory usage with memory units (k/M/G):
+
+`top -aM`
+
 - Get help about interactive commands:
 
 `?`


pages/linux/top.md
```diff
@@ -18,6 +18,10 @@
 
 `top -p $(pgrep -d ',' {{process_name}})`
 
+- Sort by memory usage with memory units (k/M/G):
+
+`top -aM`
```
pxgamer: Just tested this and I seem to get an error of `top: unknown option 'a'`. Is this a standard command?

alexandresardinha: Sorry! As you described, it's not standard.

It works on:

$ cat /etc/redhat-release
CentOS release 6.9 (Final)

So, as it's not standard, I think it's not a good idea.
Apologies for any inconvenience.

pxgamer: No problem. 👍

I'll leave this here to see if it should be included anyway.
@sbrl @agnivade

agnivade: Probably not a good idea.

sbrl: Hrm. What version of `top` are you using? If I do `top -v`, I get this:

```
  procps-ng 3.3.12
Usage:
  top -hv | -bcHiOSs -d secs -n max -u|U user -p pid(s) -o field -w [cols]
```

....I'm just wondering if it's a similar case to `netcat` - i.e. there's both a BSD `netcat` and a GNU one.

agnivade: I see the same here for centOS - https://www.unix.com/man-page/centos/1/top/. But this does not contain the `-aM` option.

pxgamer: I was using 3.3.12 on Ubuntu 17.10.

I've also tested on my NAS and CentOS server. 👍

alexandresardinha:
$ top -v
        top: procps version 3.2.8
usage:  top -hv | -abcHimMsS -d delay -n iterations [-u user | -U user] -p pid [,pid ...]

agnivade:
CentOS 6.9 https://github.com/tldr-pages/tldr/pull/1926#discussion_r163518253

sbrl:
According to [wikipedia](https://en.wikipedia.org/wiki/CentOS#CentOS_releases#End-of-support_schedule), CentOS 7 is the latest version. I assume that you tested with CentOS 7, @pxgamer and @agnivade?

pxgamer:
@sbrl, I was using CentOS 7.

sbrl:
@pxgamer As I thought. The option, which is present in CentOS 6.9, appears to have been removed in CentOS 7.

PR #1927, opened 2018-01-23
reg-export: add page

pxgamer:
Part of #1908

----


pages/windows/reg-export.md
```diff
@@ -0,0 +1,11 @@
+# reg export
+
+> Export the specified sub keys, and values into a file.
```
sbrl: Not sure that a comma is needed here.

pxgamer: You're right. I think I added that comma due to it previously being a list.

PR #1930, opened 2018-01-23
case: add page

colinmorris:
Technically this is a built-in shell keyword rather than a command, but I assume that's okay, since there's a page for `for`.

I did the examples as one-liners to follow the example of the `for` page (and because the linter complained when I tried spreading them over multiple lines). This made the second example rather long, but the linter didn't complain.

----


agnivade:
Please remove the file from common. And put it separately in linux and osx.

sbrl:
To clarify, it's because pages under `common` must be available natively on all platforms - including windows :-)

colinmorris:
When you say 'windows', could you clarify what shell environment that means? My first guess would have been cmd.exe, but I think it has essentially none of the commands in `common/` (it looks like it does have e.g. a `find` command, but the usage is totally different from the posix `find` documented in `common/`).

(When deciding where to put this, I wasn't really sure how to check whether it had a Windows implementation according to the criteria of this repo. I just chose `common/` because that was where the lion's share of commands are listed, and I assumed that if whatever Windows environment under consideration had `for`, it probably had `case` too.)

agnivade:
The current list of commands under `common` is incorrect. For historical reasons. We do not want to add on to the mistakes. 

`for` should not be under common. (Somebody needs to manually go through each command under common and test whether it works in windows or not). Hence, my request.

Hope that clarifies your concerns.

colinmorris:
Okay, I've made the requested change, but I don't think this organizational scheme is ideal. I opened an issue with a proposed alternative here: #1960.

colinmorris:
Based on the discussion in #1960, it sounds like this is okay to go in `common` after all, so I've reverted back to the original proposal.

agnivade:
So you can use bash commands in Windows now ? Wow did not know that.

colinmorris:
Not sure if you're being facetious, but yes, you can run bash commands using [Git for Windows](https://gitforwindows.org/) (as mentioned by @pxgamer in #1960). It would seem you can also get a bash shell using [Windows Subsystem for Linux](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux) though I'm not too familiar with it. I was going off the (seemingly pretty expansive) definition you gave here:

> Anything that you can "install" or "compile and run" in Windows is fair game. Can you install bash in windows ? (I am not talking about cygwin shells, that would be different) If yes, then sure case can be added. If not, then no.

Maybe Git for Windows and WSL don't count for the same reason cygwin doesn't count? Though I don't know what that reason is.

I think #1965 would help avoid misunderstandings like this in the future. (And let us bring this misunderstanding to a close!)

sbrl:
This is an awkward issue. In other OSes, we do have pages for commands that aren't installed by default - but I guess it's slightly different for 'core' commands? But then there's the problem of defining what a core command is.....

agnivade:
I was being serious. I have not used Windows for a _very_ long time. I should have googled myself though before asking. Now I feel dumb. 😰 

Yes, let us define the boundaries in #1965 completely.

PR #1932, opened 2018-01-25
nft: add page

rowanfreeman:
----

(note: this is my first-ever contribution to open source software)


agnivade:
Hi @rowanfreeman - Thanks for the page !

We would need this page to be under 8 examples. It's a hard limit that we have set ourselves. I would recommend leaving out examples in the order of least-used from a beginner's perspective.

pages/linux/nft.md
```diff
@@ -0,0 +1,36 @@
...
+
+`sudo nft --handle --numeric list chain {{family}} {{table}} {{chain}}`
+
+- Delete a rule:
+
+`sudo nft delete rule {{family}} {{table}} {{chain}} handle {{handle}}`
+
+- Save current configuration:
+
+`sudo nft list ruleset > /etc/nftables.conf`
```
agnivade: Put tokens (`{{}}`) around `/etc/nftables.conf`

pages/linux/nft.md
```diff
@@ -0,0 +1,36 @@
...
+
+`sudo nft add chain {{family}} {{table}} {{chain}} \{ type {{type}} hook {{hook}} priority {{priority}} \; \}`
+
+- Add a new rule:
+
+`sudo nft {{[add | insert]}} rule {{family}} {{table}} {{chain}} {{[position]}} {{statement}}`
+
+- Example rule:
+
+`sudo nft add rule inet filter input tcp dport \{ telnet, ssh, http, https \} accept`
```
agnivade: This can be merged with the "Add a new rule" example. It is how we usually do it. You can expand the description a bit more mentioning the family, table and other terms. But we need to be careful to be make it too large. And use this concrete example there.

You can see the `ufw.md` page as a reference.

pages/linux/nft.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Show rule handles:
+
+`sudo nft --handle --numeric list chain {{family}} {{table}} {{chain}}`
+
+- Delete a rule:
+
+`sudo nft delete rule {{family}} {{table}} {{chain}} handle {{handle}}`
```
agnivade: This can also use a concrete example if you are doing it in the add command, just to be consistent.

pages/linux/nft.md
```diff
@@ -0,0 +1,32 @@
+# nftables
```
sbrl: There appears to be a mismatch between the name of the page and the command name. Perhaps we could change this to `nft`, and then include it's full name in the description somehow? @agnivade, what do you think?

rowanfreeman:
Sorry about many needless commits/reverts. Just making mistakes in Git.

pages/linux/nft.md
agnivade: Right. If the command name is `nft`, that should be the title.

sbrl:
No problem, @rowanfreeman. Thanks for the PR! 😺

PR #1939, opened 2018-01-25
grep: show filenames only

zlbabe:
Show filenames only

### DIFF
--- a/pages/common/grep.md
+++ b/pages/common/grep.md
@@ -34,3 +34,7 @@
 - Invert match for excluding specific strings:
 
 `grep -v {{search_string}}`
+
+- Show filenames only:
+
+`grep -l {{search_string}}`


agnivade:
Hey @zlbabe , thanks for the contribution.

Unfortunately, we have a hard limit of 8 for the no. of examples that we allow in a page. This was written explicitly in the PR template. The reason is to challenge ourselves to provide the most useful set of examples in a short space. For a complete documentation, there is always the man page.

The grep page has gone through a lot of iterations. I don't see anything that warrants to be removed.

zlbabe:
>I don't see anything that warrants to be removed.

Well, if you look at example 
```
- Use the standard input instead of a file:

cat path/to/file | grep search_string
```
I don't think that it's very useful, it's about using pipe, it does not even concern grep command. Also, why use `cat+grep` when grep alone can do the job. Maybe we can remove it and replace it by the one I suggest.

agnivade:
Its true that purists will say it is a [useless use of cat](http://porkmail.org/era/unix/award.html), and you can do `grep search_string < file`. But the primary motive behind this example is to show that grep can take input from standard input. Your example does not show that. And since doing a `cat + grep` is so ubiquitous in your day-to-day life, this warrants a place here.

zlbabe:
I don't see the value added by the example of ` some command | grep ` as everyone knows it. I mean every beginner with command line knows it. So it isn't wise to put some other stuff instead which brings some new useful information.
It's okay now if you want to close it.

agnivade:
I wouldn't want to assume what everyone knows or does not know. But I have seen people struggling to understand the concept of piping. And since this project is _aimed_ at beginners, it only makes sense to include simple commands that we use everyday, rather than esoteric ones.

Thanks for your understanding @zlbabe.

PR #1943, opened 2018-01-26
restic: add page

hdhauk:
----

### DIFF
--- a/pages/common/restic.md
+++ b/pages/common/restic.md
@@ -18,6 +18,6 @@
 
 `restic -r {{path/to/repository}} restore {{snapshot_id}} {{path/to/target}}`
 
-- Clean up repository and keep only the most recent snapshot of each unique backup:
+- Clean up the repository and keep only the most recent snapshot of each unique backup:
 
 `restic forget --keep-last 1 --prune`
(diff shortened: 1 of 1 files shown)


agnivade:
@hdhauk - Can you do a force push and remove the older commits ? There seems to be a merge commit at the end. If you just get the latest from master and apply just the new commit and then force push, I think that should do it.

hdhauk:
@agnivade Looks better now?

pages/common/restic.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Show backup snapshots currently stored in the repository:
+
+`restic -r {{/path/to/repository/}} snapshots`
+
+- Restore a specific backup snapshot to a target directory:
+
+`restic -r {{/path/to/repository/}} restore {{snapshot_id}} {{/path/to/target}}`
+
+- Clean up repository and keep only the most recent snapshot of each unique backup (same host and directory):
```
agnivade: Do we need to mention "same host and directory" ? I thought that was implied.

hdhauk: There are no default interval when snapshots are taken, they are taken whenever you run the backup command. For regular automatic backups you'll need to use cron or something.

You're right, its probably redundant – removed!

pxgamer:
I've merged this as the grammatical change is pretty minor. 👍

PR #1944, opened 2018-01-27
babel: add page

pxgamer:
----

### DIFF
--- a/pages/common/babel.md
+++ b/pages/common/babel.md
@@ -1,0 +1,35 @@
+# babel
+
+> A compiler for writing next generation JavaScript.
+
+- Compile a specified input file, and output to stdout:
+
+`babel {{path/to/file}}`
+
+- Compile a specified input file, and output to a specific file:
+
+`babel {{path/to/input_file}} --out-file {{path/to/output_file}}`
+
+- Compile the input file every time it is changed:
+
+`babel {{path/to/input_file}} --watch`
+
+- Compile a whole directory of files:
+
+`babel {{path/to/input_directory}}`
+
+- Ignore specified comma-separated files in a directory:
+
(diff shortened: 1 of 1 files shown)


pages/common/babel.md
```diff
@@ -0,0 +1,35 @@
+# babel
+
+> A compiler for writing next generation JavaScript.
```
agnivade: Shouldn't it be transpile ?

pxgamer: Yes, I think.

sbrl: Yes. From [here](https://www.stevefenton.co.uk/2012/11/compiling-vs-transpiling/):

> Compiling
>  - is the general term for taking source code written in one language and transforming into another
> 
> Transpiling
>  - is a specific term for taking source code written in one language and transforming into another language that has a similar level of abstraction

pages/common/babel.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Transpile the input file every time it is changed:
+
+`babel {{path/to/input_file}} --watch`
+
+- Transpile a whole directory of files:
+
+`babel {{path/to/input_directory}}`
+
+- Ignore specified comma-separated files in a directory:
```
sbrl: Can this contain globs?

pages/common/babel.md
```diff
@@ -0,0 +1,35 @@
+# babel
+
+> A transpiler for writing next generation JavaScript.
```
agnivade: I think the language can be improved as - `A transpiler which converts code from JavaScript ES6/ES7 syntax to ES5 syntax.`. Although it supports other targets too. But as the most general case, this is what it is used for.

pages/common/babel.md
pxgamer: Yes, it can contain globs.

PR #1948, opened 2018-01-28
maintainers-guide: cleanup msg when squashing

waldyrious:
Just a small addition to the maintainers' guide about how to handle squash-type merges, plus small wording changes in the PR merging section.

### DIFF
--- a/contributing-guides/maintainers-guide.md
+++ b/contributing-guides/maintainers-guide.md
@@ -84,9 +84,10 @@ as a guideline for current and future maintainers.
   If there's a single commit in the PR,
   or if the multiple commits are not semantically independent changes,
   use the `Squash and merge` method.
-  If the PR author took the time to craft
+  (Don't forget to clean up the body of the squashed commit message.)
+  If instead the PR author took the time to craft
   individual, informative messages for each commit,
-  use the `Rebase and merge` method,
+  then use the `Rebase and merge` method,
   to honor that work and preserve the history of the changes.
   For less clear-cut cases, a simple heuristic you can follow
   is that if there are more "dirty" commits than "clean" commits,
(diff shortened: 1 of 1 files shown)


waldyrious:
Hey @jsonbruce, thanks for the review. Next time please use squash for this type of PR (notice how the second commit message was generic and not really indicative of the changes therein). Does that make sense to you?

jsonbruce:
Oops..., I misread the second commit message. :sob: 
I'm really sorry for this failed practice.

And I'm thinking how can I revert this wrong merge action?

agnivade:
No worries :smile: 

And you cannot revert changes to master. master is a protected branch :) So no `git push -f` :stuck_out_tongue_closed_eyes:

waldyrious:
> No worries 😄

Yeah, that's not a big deal. Just look at [our past graph](https://gitlab.com/waldyrious/tldr-mirror/network/master) and you'll see how we haven't sweat it too much when less clean merges have happened (also due to an explicit policy only recently being adopted, via this document).

But the graph should also illustrate the desirability of a cleaner graph, especially for a project with so many contributors. So just keep an eye out for that in future merges ;)

waldyrious:
by the way, the fact that we're using a linear history probably warrants a note in the documentation recommending that every commit mentions the page it refers to (e.g. `page_foo: description of changes`). Otherwise some of the commit messages (in PRs with multiple commits) may be hard to understand without the context of surrounding commits. That's the case for some of @pxgamer's recent commits, for instance

Do you guys think it's worth adding this note?

pxgamer:
@waldyrious, I'm used to following the standard format of Tim Pope, I'm happy to start using that format if it's needed. 👍

I think in the case of this project, it does make more sense to use the page name for all commits.

PR #1955, opened 2018-01-30
trizen: new page

rowanfreeman:
----

### DIFF
--- a/pages/linux/trizen.md
+++ b/pages/linux/trizen.md
@@ -2,7 +2,7 @@
 
 > Arch Linux utility for building packages from the Arch User Repository (AUR).
 
-- Synchronize and update all packages (including AUR):
+- Synchronize and update all AUR packages (excludes non-AUR packages):
 
 `trizen -Syua`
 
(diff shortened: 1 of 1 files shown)


pages/linux/trizen.md
```diff
@@ -0,0 +1,27 @@
+# trizen
+
+> Arch Linux utility for building packages from the Arch User Repository (AUR).
+
+- Synchronize and update all AUR packages (excludes non-AUR packages):
```
agnivade: maybe we can update the description to `Synchronize and update all non-AUR packages` ? Then we don't need to mention "includes AUR packages" for every description after this.

pages/linux/trizen.md
```diff
@@ -0,0 +1,27 @@
...
+
+`trizen -S {{package}}`
+
+- Remove a package and its dependencies (includes AUR packages):
+
+`trizen -Rs {{package}}`
+
+- Search the package database for a keyword (including AUR):
+
+`trizen -Ss {{package}}`
```
agnivade: Should it be "keyword" instead of "package" ? Because you are entering the keyword to search for the package.

sbrl:
Looks ok to me! Thanks, @rowanfreeman :D

PR #1963, opened 2018-01-31
del: add page

pxgamer:
Part of #1579

----

### DIFF
--- a/pages/windows/del.md
+++ b/pages/windows/del.md
@@ -1,0 +1,31 @@
+# del
+
+> Delete one or more files.
+
+- Delete one or more space-separated files or patterns:
+
+`del {{file_pattern}}`
+
+- Prompt for confirmation before deleting each file:
+
+`del {{file_pattern}} /p`
+
+- Force the deletion of read-only files:
+
+`del {{file_pattern}} /f`
+
+- Recursively delete files from all subdirectories:
+
+`del {{file_pattern}} /s`
+
+- Do not prompt when deleting files based on a global wildcard:
+
+`del {{file_pattern}} /q`
+
+- Display the help and list available attributes:
+
+`del /?`
+
+- Delete files based on specified attributes:
+
+`del {{file_pattern}} /a {{attribute}}`
--- a/pages/windows/del.md
+++ b/pages/windows/del.md
@@ -14,7 +14,7 @@
 
 `del {{file_pattern}} /f`
 
-- Recursively delete files from all subdirectories:
+- Recursively delete file(s) from all subdirectories:
 
 `del {{file_pattern}} /s`
 


pages/windows/del.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Delete one or more space-separated files or patterns:
+
+`del {{file_pattern}}`
+
+- Prompt for confirmation before deleting each file:
+
+`del {{file_pattern}} /p`
+
+- Force the deletion of read-only files:
```
agnivade: Why read-only ? Does it mean it won't delete writable files ?

pages/windows/del.md
```diff
@@ -0,0 +1,31 @@
...
+
+`del {{file_pattern}} /f`
+
+- Recursively delete files from all subdirectories:
+
+`del {{file_pattern}} /s`
+
+- Do not prompt when deleting files based on a global wildcard:
```
agnivade: What does "based on a global wildcard" mean ?

pages/windows/del.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Prompt for confirmation before deleting each file:
+
+`del {{file_pattern}} /p`
+
+- Force the deletion of read-only files:
+
+`del {{file_pattern}} /f`
+
+- Recursively delete files from all subdirectories:
```
agnivade: Perhaps change this to `file(s)` ?

pages/windows/del.md
pxgamer: By default, `del` only removes files that aren't set to `Read-only` on the file system (there's a checkbox in the file explorer).

Using `/f` will remove any files that have the `Read-only` status set, as well as other files.

pages/windows/del.md
pxgamer: Sounds good 👍

pages/windows/del.md
pxgamer: An example would be `del *` or `del *.*` (both do the same thing).

pages/windows/del.md
agnivade: I see. Got it.

pages/windows/del.md
agnivade: But that has nothing to do with the `/q` switch right ? You can do `del file.txt /q` and `del *.txt /q` and both will do the same thing. The description makes it look like the the opposite is also true - i.e. when files are deleted without a wildcard, it prompts even if you specify `/q`.

pxgamer: The `/q` switch is only used to allow deletion of everything in a directory using either `*` or `*.*`. Specifying a wildcard with an extension (e.g. `*.txt`) won't prompt. You're right that `/q` won't do anything for `*.txt` or `file.txt` as it only applies when a wildcard is used that will delete all content.

agnivade: Ah I see now. Perhaps you should just specify * or *.* in brackets. Because it applies only on those cases.

pxgamer: Sounds good. :+1: Would improve the readability and reduce confusion.

pages/windows/del.md
pxgamer: Ping @agnivade. Is this ok? 🤔

sbrl:
Merged due to inactivity and the PR being made by a maintainer.

PR #1964, opened 2018-01-31
Render PDF copies of tldr pages

aydwi:
Added a Python script and required resources to generate a PDF file of tldr pages


agnivade:
A couple of overall points before I take a deeper look.
1. Please move the directory inside "scripts".
2. Ideally, we would want just a single pdf file with sections inside for linux, windows, osx and common. Think of the pdf file as the combined data of all the pages. Having separate pdf files is not very useful because if a user wants to look up a command for a different platform, he/she has to open another file.
3. There is no need to check-in the generated pdf files. I think of them as supplementary data. A user can generate the pdf if needed. Plus, they will get outdated with every commit if we keep them (which would need auto-generating the file for every commit).

Also, the code needs a lot of work. There is no `if __name__ == "__main__":`. No shebang. Unnecessary use of global variables.

@felixonmars - Can you do a review ?

aydwi:
@agnivade 1. Will do

2. That's a good idea. I'll adjust it to generate file in this format.

3. Yeah, I understand they will get outdated fast, but I committed them just to let others see how they look. Thinking again, not very useful. I'll remove them.

Further, I'll add the shebang. Also, I did not use main because I don't think this standalone script is going to be imported as a module. Anyways, I'll add it if required.

pxgamer:
I agree with @agnivade's points. Other than that, not much to say. I don't have much experience with Python. :+1:

agnivade:
Thanks, looks better now. A couple of comments before I look in detail - 

1. Use the argparse library to specify the arguments. That will automatically generate the `-h` contents and will allow you to specify the theme values. Also, it takes care of checking required parameters.
2. Parse the arguments and then pass them to the `main()` function.

sbrl:
Looks good! I'd investigate the use of `virtualenv` to isolate dependencies though.

aydwi:
I've been somewhat busy. I'll make these changes soon.

aydwi:
@agnivade @sbrl I've made and pushed the changes. The arguments are parsed using `argparse` now, and then passed to `main()`. Also made some other minor changes. Sorry for the delay!

sbrl:
Nice and portable - cool :sunglasses: 

I like that section on requirements - it's too often overlooked.

scripts/pdf/render.py
```diff
@@ -0,0 +1,132 @@
...
+        loc = loc + "/"
+
+    #Writing names of all directories inside 'pages' to a list
+    for os_dir in os.listdir(loc):
+        oslist.append(os_dir)
+
+    oslist.sort()
+
+    #Required strings to create intermediate HTML files
+    header = "<html><head><link rel=\"stylesheet\" type=\"text/css\" href=\"" + colorscheme + ".css\"></head><body>\n"
```
agnivade: Use triple-quotes to avoid unnecessary escaping of the double-quote. Or you can avoid the quotes altogether. The spec does not require quotes. This - `<style type=text/css>` is completely valid. :smile:

scripts/pdf/render.py
```diff
@@ -0,0 +1,132 @@
...
+    allmd = []
+    group = []
+    ap = []
+
+    #Checking correctness of path
+    if not os.path.isdir(loc):
+        print("Invalid directory. Please try again!", file = sys.stderr)
+        sys.exit(1)
+
+    if not loc.endswith('/'):
```
agnivade: I don't think this is required. It works fine without the trailing slash. See my comment below.

scripts/pdf/render.py
```diff
@@ -0,0 +1,132 @@
...
+        dir_title = "<h2 class=\"titledir\">" + operating_sys.capitalize() + "</h2></body></html>"
+
+        #Creating directory title page for current directory
+        with open("dir_title.html", 'w') as os_html:
+            os_html.write(header + dir_title)
+        
+        group.append(HTML('dir_title.html').render())
+
+        #Creating a list of all md files in the current directory
+        for temp in glob.glob(os.path.join(loc + operating_sys, '*.md')):
```
agnivade: Change this line to - `os.path.join(loc, operating_sys, '*.md')` to avoid adding `/` at the end of loc earlier.

scripts/pdf/render.py
```diff
@@ -0,0 +1,132 @@
...
+        f.write(header + title_content)
+
+    group.append(HTML('title.html').render())
+
+    for operating_sys in oslist:
+
+        i = 1
+
+        #Required string to create directory title pages
+        dir_title = "<h2 class=\"titledir\">" + operating_sys.capitalize() + "</h2></body></html>"
```
agnivade: Follow the same convention as before - either use triple quotes, or remove the double-quotes.

scripts/pdf/render.py
```diff
@@ -0,0 +1,132 @@
...
+        for temp in glob.glob(os.path.join(loc + operating_sys, '*.md')):
+            allmd.append(temp)
+
+        #Sorting all filenames in the directory, to maintain the order of the PDF
+        allmd.sort()
+
+        #Conversion of md to HTML
+        for md in allmd:
+
+            output_file = open("htmlout.html", "a")
```
agnivade: You are not doing anything with this. Mistake ?

scripts/pdf/render.py
```diff
@@ -0,0 +1,132 @@
...
+
+            with open(md, "r") as inp:
+                text = inp.readlines()
+
+            purge = open("htmlout.html", "w")
+            purge.close()
+
+            with open("htmlout.html", "a") as out:
```
agnivade: If you open the file with `w` and start writing, it will automatically overwrite the file. No need to do a purge and then open with `a`.

scripts/pdf/render.py
```diff
@@ -0,0 +1,132 @@
...
+                    if re.match(r'^>', line):
+                        line = line[:0] + '####' + line[1:]
+
+                    html = markdown.markdown(line)
+                    out.write(html)
+
+                out.write(footer)
+
+            group.append(HTML('htmlout.html').render())
+            print("Rendered page " + str(i) + " of the directory " + operating_sys)
```
agnivade: Use string formatting - `"Rendered page %d of the directory %s" % (i, operating_sys)`

scripts/pdf/render.py
```diff
@@ -0,0 +1,132 @@
...
+            purge = open("htmlout.html", "w")
+            purge.close()
+
+            with open("htmlout.html", "a") as out:
+                out.write(header)
+
+                for line in text:
+                    if re.match(r'^>', line):
+                        line = line[:0] + '####' + line[1:]
+
```
agnivade: Remove these empty lines on L85 and L88. It is readable enough.

scripts/pdf/render.py
aydwi: Made the change - not using quotes now

scripts/pdf/render.py
aydwi: Thanks for pointing this out. Made the change

scripts/pdf/render.py
aydwi: Removed the quotes

scripts/pdf/render.py
aydwi: I missed it somehow. Removed this line

scripts/pdf/render.py
aydwi: When I wrote this, I was experimenting with something and created the `purge` block. Many thanks for pointing this out. I've improved it.

scripts/pdf/render.py
aydwi: Now using string formatting

`print("Rendered page {} of the directory {}".format(str(i), operating_sys))`

scripts/pdf/render.py
aydwi: Removed

aydwi:
@agnivade Thanks for pointing these out. I've made the changes.

PR #1967, opened 2018-02-01
sed: adding remove trailing and leading whitespace commands

codydiehl:
Added removing leading and trailing whitespace [tabs, spaces] from end and beginning of lines.
useful for code cleanup.

----

### DIFF
--- a/pages/common/sed.md
+++ b/pages/common/sed.md
@@ -25,3 +25,11 @@
 - Replace separator / by any other character not used in the find or replace patterns, e.g., #:
 
 `sed 's#{{find}}#{{replace}}#' {{filename}}`
+
+- Remove all leading whitespace (spaces, tabs) from front of each line, overwriting the file:
+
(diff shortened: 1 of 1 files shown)


pxgamer:
Hi @codydiehl, thanks for the edits.

Personally I'd say these are already slightly covered by the other examples (find and replace in-place).

But I'll leave it for the others to decide. 👍

agnivade:
Hey @codydiehl - thanks for the PR. This is just a specific case of the `-i` example already added. If we add this, we will be setting precedents for a lot of commands to have a generic example and then a specific example. Not sure if we want to do that.

codydiehl:
Understood and reasonable. We can close this. 
I had mainly made the request to expand on some available regex type options available with sed. However that may not be fit for the sed tldr, and rather have better place in a regex specific entry.

agnivade:
Thanks !

PR #1970, opened 2018-02-02
phpcs: add page

pxgamer:
----

## Basic Check

## References

Official Page: https://github.com/squizlabs/PHP_CodeSniffer

### DIFF
--- a/pages/common/phpcs.md
+++ b/pages/common/phpcs.md
@@ -1,0 +1,35 @@
+# phpcs
+
+> Tokenize PHP, JavaScript and CSS files to detect violations of a defined set of coding standards.
+
+- Sniff the specified directory for issues:
+
+`phpcs {{path/to/directory}}`
+
+- Display a list of installed coding standards:
+
+`phpcs -i`
+
+- Specify a coding standard to validate against:
+
+`phpcs {{path/to/directory}} --standard {{standard}}`
+
+- Specify file extension(s) to include when sniffing:
+
+`phpcs {{path/to/directory}} --extensions {{file_extension(s)}}`
+
+- Specify the format of the output report (e.g. `full`, `xml`, `json`, `summary`):
+
+`phpcs {{path/to/directory}} --report {{format}}`
+
+- Set config variables to be used during the process:
+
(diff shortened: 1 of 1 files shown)


pages/common/phpcs.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Display a list of installed coding standards:
+
+`phpcs -i`
+
+- Specify a coding standard to validate against:
+
+`phpcs {{path/to/directory}} --standard {{standard}}`
+
+- Specify file extension(s) to include when sniffing:
```
sbrl: Is there a way to exclude files too?

pxgamer: There's an `--ignore` option which can be used.

pxgamer:
@jsonbruce, I thought I'd try it out. At least so we can see how it looks. :+1:

pages/common/phpcs.md
```diff
@@ -0,0 +1,35 @@
+# phpcs
+
+> Tokenize PHP, JavaScript and CSS files to detect violations of a defined set of coding standards.
+
+- Sniff the specified directory for issues:
```
agnivade: What is the default coding standard it checks against ?

pxgamer: It defaults to the PEAR standard.

agnivade: Ok, perhaps we can specify that in brackets to make it extra clear. It was a natural question that came to my mind.

pxgamer: Makes sense. :smiley:

pages/common/phpcs.md
sbrl: Perhaps we could one of the lesser used options out for that, or include it in this example? It sounds useful to me. Thoughts?

pxgamer: I guess maybe `--ignore` instead of `-l` (no recursion)? 🤔

sbrl: Yeah, I guess. I guess you can always use shell globs with it (i.e. `badfolder/**` will grab _everything_ in `badfolder/`)

PR #1974, opened 2018-02-06
rabin2: add page

alufers:
Added page for the `rabin2` program.
https://radare.gitbooks.io/radare2book/content/rabin2/intro.html

----

### DIFF
--- a/pages/common/rabin2.md
+++ b/pages/common/rabin2.md
@@ -1,7 +1,7 @@
 # rabin2
 
 > Get information about binary files - symbols, sections, linked libraries, etc.
-> Comes bundled with radare2
+> Comes bundled with radare2.
 
 - Display general information about a binary (architecture, type, endianness):
 
--- a/pages/common/rabin2.md
+++ b/pages/common/rabin2.md
@@ -1,7 +1,7 @@
 # rabin2
 
-> Get information about binary files - symbols, sections, linked libraries, etc.
-> Comes bundled with radare2.
+> Get information about binary files (ELF, PE, Java CLASS, Mach-O) - symbols, sections, linked libraries, etc.
+> Comes bundled with `radare2`.
 
 - Display general information about a binary (architecture, type, endianness):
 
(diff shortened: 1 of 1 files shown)


pages/common/rabin2.md
```diff
@@ -0,0 +1,24 @@
+# rabin2
+
+> Get information about binary files - symbols, sections, linked libraries, etc.
+> Comes bundled with radare2.
```
agnivade: Wrap radare2 within backticks (` `` `)

pages/common/rabin2.md
```diff
@@ -0,0 +1,24 @@
+# rabin2
+
+> Get information about binary files - symbols, sections, linked libraries, etc.
```
agnivade: What sort of binaries does it support ? ELF ? Mach-O ?

pages/common/rabin2.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Display symbols imported from libraries:
+
+`rabin2 -i {{path/to/binary}}`
+
+- Display strings contained in the binary:
+
+`rabin2 -z {{path/to/binary}}`
+
+- Output in JSON (works for every subcommand):
```
agnivade: Maybe it's better to say `Display the output in JSON:` and remove the -I option. You can combine any of the flags with other commands also (Please correct me if I am wrong). So I guess it is a bit redundant to say that.

alufers: You are right, but the `-j` option does not work on its own (it requires another flag), and I don't want to include examples that are broken.

agnivade: Oh I see. Alright then, just change the description. No need to remove `-I`

alufers: I have changed the description in the latest commit.

agnivade: I also wanted to remove the `(works for every subcommand)`. Because you can intermix the previous flags also.

alufers: Done.

agnivade: Thanks !

PR #1978, opened 2018-02-06
apk: add page

PlrEx:
----

### DIFF
--- a/pages/linux/apk.md
+++ b/pages/linux/apk.md
@@ -22,6 +22,6 @@
 
 `apk search {{keyword}}`
 
-- List the packages installed in the System:
+- Get info about a specific package:
 
 `apk info {{package}}`
--- a/pages/linux/apk.md
+++ b/pages/linux/apk.md
@@ -25,4 +25,3 @@
 - List the packages installed in the System:
 
 `apk info {{package}}`
-
--- a/pages/linux/apk.md
+++ b/pages/linux/apk.md
@@ -1,0 +1,28 @@
+# apk
+
+> Alpine Linux package management tool.
+
+- Update repository indexes from all remote repositories:
+
+`apk update`
+
+- Install a new package:
+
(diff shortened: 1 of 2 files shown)


pages/linux/apk.md
```diff
@@ -0,0 +1,27 @@
+# apk
+
+> Alpine Linux package management tool.
+
+- Update repository indexes from all remote repositories:
```
agnivade: Shouldn't this be after the `add` option ? What is the natural flow of actions here ? Does the user first do an update or add a package ?

pages/linux/apk.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Repair package or upgrade it without modifying main dependencies:
+
+`apk fix {{package}}`
+
+- Search package via keyword:
+
+`apk search {{keyword}}`
+
+- List the packages installed in the System:
```
agnivade: "System" should be lowercase. Also, the description says "List the packages installed", but the example has "{{package}}". So does it mean it will show all packages with the given string ? Or is there a mistake ?

pages/linux/apk.md
sbrl: I believe one has to first update the indexes, as with `apt`.

pxgamer: Yes, looking at their docs that is the correct flow. It updates the package indexes. :+1:

pages/linux/apk.md
pxgamer: Looks like `apk info` will list all packages (like the description says) and `apk info {{package}}` only displays for a specific package so I think this may be a mistake.

PlrEx: I have made the changes to **apk info** command. Thank you for pointing out.

pages/linux/apk.md
agnivade: great.

PR #1980, opened 2018-02-08
zip: added multi-part zip compression

hendrikelsner:

### DIFF
--- a/pages/common/zip.md
+++ b/pages/common/zip.md
@@ -30,6 +30,6 @@
 
 `zip -d {{compressed.zip}} "{{foo/*.tmp}}"`
 
-- Archive a directory and its contents to a multi-part [s]plit zip file (i.e. 3GB parts):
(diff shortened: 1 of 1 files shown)


pages/common/zip.md
```diff
@@ -29,3 +29,7 @@
 - Delete files from an existing zip file:
 
 `zip -d {{compressed.zip}} "{{foo/*.tmp}}"`
+
+- Archive a directory and its contents to a multi-part [s]plit zip file (i.e. 3GB parts):
```
pxgamer: I assume other zip sizes can be specified?

It may be better to use `e.g.` instead of `i.e.` 👍 `i.e.` implies that it may have to be 3GB, whereas `e.g.` would be saying that 3GB is an example and can be replaced by other sizes.

hendrikelsner: yes, you can specify other sizes. thank you for the correction. 👍

PR #1982, opened 2018-02-09
ping6: add page

mstruebing:
https://linux.die.net/man/8/ping6
https://www.tutorialspoint.com/unix_commands/ping6.htm

----

### DIFF
--- a/pages/common/ping6.md
+++ b/pages/common/ping6.md
@@ -2,7 +2,7 @@
 
 > Send ICMP ECHO_REQUEST packets to network hosts via IPv6 address.
 
-- Ping host:
+- Ping a host:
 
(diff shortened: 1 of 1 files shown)


pages/common/ping6.md
```diff
@@ -0,0 +1,23 @@
+# ping6
+
+> Send ICMP ECHO_REQUEST packets to network hosts via IPv6 address.
+
+- Ping host:
```
pxgamer: Could you change this to `Ping a host:` for consistency. Same with the other ones. 👍

agnivade: btw, this issue is there in the original ping.md page too. We should fix that.

PR #1986, opened 2018-02-11
fish: fix typo, omf: add page

TempPlaceHolder:
omf: add page

### DIFF
--- a/pages/common/omf.md
+++ b/pages/common/omf.md
@@ -1,0 +1,28 @@
+# omf
+
+> Oh My Fish, the Fishshell Framework.
+> Install packages to extend and modify the fish shell.
+
+- Install one or more packages:
+
+`omf install {{name}}`
+
+- List installed packages:
+
+`omf list`
+
+- List available themes:
+
+`omf theme`
+
+- Apply a theme:
+
+`omf theme {{name}}`
+
+- Remove a theme or package:
+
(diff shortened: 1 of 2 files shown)


pages/common/omf.md
```diff
@@ -0,0 +1,28 @@
+# omf
+
+> Oh My Fish, the Fishshell Framework.
+> Install packages to extend and modify the fish shell.
+
+- Install one or more packages:
```
agnivade: How are multiple packages specified ? Separated by space ? We should mention that.

pages/common/omf.md
```diff
@@ -0,0 +1,28 @@
...
+
+- List available themes:
+
+`omf theme`
+
+- Apply a theme:
+
+`omf theme {{name}}`
+
+- Remove a theme or package:
```
agnivade: What if there is a theme and package with the same name ? Both will be removed ?

pages/common/omf.md
pxgamer: Yeh, they are space separated according to the docs. :+1:

pages/common/omf.md
TempPlaceHolder: Hm themes are also packages, so it doesn't matter.

agnivade: I see. Ok.

PR #1989, opened 2018-02-12
forever: add page

sakalys:
- [v] The page (if new), does not already exist in the repo.

- [v] The page (if new), has been added to the correct platform folder:  

- [v] The page has 8 or fewer examples.

- [v] The PR is appropriately titled:  

- [v] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines.

### DIFF
--- a/pages/common/forever.md
+++ b/pages/common/forever.md
@@ -1,15 +1,15 @@
-# forever - nodejs tool
+# forever
 
-> Server-side JavaScript application that makes sure node applications run indefinitely (restarts after crash).
+> Server-side JavaScript application that makes sure Node.js applications run indefinitely (restarts after exit).
 
 - Start running a file forever (as a daemon):
 
-`forever {{file}}.js`
+`forever {{script}}`
 
-- List forever running processes:
+- List running "forever" processes (along with IDs and other details of "forever" processes):
 
 `forever list`
 
-- Stop a forever running process:
+- Stop a running "forever" process:
 
-`forever stop {{ID from "forever list"}}`
+`forever stop {{ID|pid|script}}`
--- a/pages/common/forever.md
+++ b/pages/common/forever.md
@@ -1,6 +1,6 @@
-# forever (nodejs/npm tool)
+# forever - nodejs tool
 
-> Server-side JavaScript application that makes sure node applications run indefinitely (restarts after crash)
+> Server-side JavaScript application that makes sure node applications run indefinitely (restarts after crash).
 
 - Start running a file forever (as a daemon):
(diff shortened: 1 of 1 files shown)


pages/common/forever.md
```diff
@@ -0,0 +1,15 @@
+# forever - nodejs tool
+
+> Server-side JavaScript application that makes sure node applications run indefinitely (restarts after crash).
+
+- Start running a file forever (as a daemon):
+
+`forever {{file}}.js`
+
+- List forever running processes:
```
Hoi15A: That sentence sounds a bit odd to me. Maybe `List running forever processes:` is a better fit?

pages/common/forever.md
```diff
@@ -0,0 +1,15 @@
...
+
+`forever list`
+
+- Stop a forever running process:
```
Hoi15A: Same as above. I feel like `Stop a running forever process:` sounds clearer.

pages/common/forever.md
```diff
@@ -0,0 +1,15 @@
+# forever - nodejs tool
+
+> Server-side JavaScript application that makes sure node applications run indefinitely (restarts after crash).
```
Hoi15A: I havent really used `forever` before. Does it only restart after crashes or also just normal exiting (i.e. exit code 0)? If so maybe its better to say `(restarts after exit).`

pages/common/forever.md
```diff
@@ -0,0 +1,15 @@
+# forever - nodejs tool
```
sbrl: The generally prefer the title of a page to be the same as the name of command.

pages/common/forever.md
agnivade: Mention that IDs are also listed in the output, so that you don't have to mention - ID from "forever list" in the next example.

sakalys: Well, if you say so, then the following would be best IMO:
> List running "forever" processes

pages/common/forever.md
sakalys: It also restarts after a clean exit (0). I guess you're right, it could say "exit".

pages/common/forever.md
sakalys: I agree

pages/common/forever.md
sakalys: Let me know if you're asking me to change it to "forever"

sakalys:
I addressed all the suggested changes. Let me know if there's anything else.

PR #1994, opened 2018-02-13
terminator: add page

### LINKED ISSUE #1987: page request: terminator
[Terminator](https://linux.die.net/man/1/terminator) - Multiple GNOME terminals in one window

TheYahya:
Fixes #1987

### DIFF
--- a/pages/linux/terminator.md
+++ b/pages/linux/terminator.md
@@ -1,19 +1,23 @@
 # terminator
 
-> Multiple GNOME terminals in one window.
+> Arrange multiple GNOME terminals in one window.
+
+- Start terminator window:
+
+`terminator`
 
 - Start with a fullscreen window:
 
 `terminator -f`
 
-- Split terminals Horizontally:
+- Split terminals horizontally:
 
-`Ctrl+Shift+O`
+`Ctrl + Shift + O`
 
-- Split terminals Vertically:
+- Split terminals vertically:
 
-`Ctrl+Shift+E`
+`Ctrl + Shift + E`
 
 - Open new tab:
 
-`Ctrl+Shift+T`
+`Ctrl + Shift + T`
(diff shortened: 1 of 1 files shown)


pages/linux/terminator.md
```diff
@@ -0,0 +1,19 @@
+# terminator
+
+> Multiple GNOME terminals in one window.
```
agnivade: How about - `Arrange multiple GNOME terminals in one window` ?

pages/linux/terminator.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Start with a fullscreen window:
+
+`terminator -f`
+
+- Split terminals Horizontally:
```
agnivade: Is there a reason to capitalize H and V ?

pages/linux/terminator.md
```diff
@@ -0,0 +1,19 @@
+# terminator
+
+> Multiple GNOME terminals in one window.
+
+- Start with a fullscreen window:
```
agnivade: We usually start with the simplest command invocation - i.e. `terminator` and explain what it does.

pages/linux/terminator.md
TheYahya: It's because I just copy/paste it from `man` page :see_no_evil:, I will fix it.

pages/linux/terminator.md
TheYahya: 👍

pages/linux/terminator.md
TheYahya: I thought it would be better if we use the original description of it in `man` page. don't you think?

agnivade: Not always. `man` page descriptions are terse and meant to be comprehensive. Our motto is to make things simple and easy to understand for beginners. In that context, it is always good to have descriptions which are absolutely crystal clear. Hence, my suggestion.

If we always copy from the man pages, the user might as well do `man terminator` instead of `tldr terminator` :laughing:

pages/linux/terminator.md
TheYahya: You're right, ok :ok_hand:

PR #1996, opened 2018-02-14
swapon: add page

shashank2806:
----

### DIFF
--- a/pages/linux/swapon.md
+++ b/pages/linux/swapon.md
@@ -1,0 +1,23 @@
+# swapon
+
+> Enables device or file for paging or swapping.
+
+- Get swap information:
+
+`swapon -s`
+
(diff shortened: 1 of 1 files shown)


pages/linux/swapon.md
```diff
@@ -0,0 +1,23 @@
+# swapon
+
+> Enables device or file for paging or swapping.
```
agnivade: This line seems to be picked directly from the man page. I have one small suggestion - "paging" is a general concept. "swapping" is an implementation of paging to disk. We should probably just use "swapping" here for simplicity.

shashank2806: Thanks for suggestion !

shashank2806:
Indeed, You guessed it right !

PR #2002, opened 2018-02-17
Calculating command coverage

### LINKED ISSUE #1070: Create lists of commands to test coverage parity against

agnivade:
- Removed the files badge because it was taking too much space. And added this badge in its place.
- For now, just going with the linux platform. But will add other platforms too later.

DO NOT MERGE THIS YET. Some lines need to be un-commented. 

But review is most welcome. @waldyrious @sbrl @pxgamer @jsonbruce 

Fixes #1070


agnivade:
@waldyrious - would you be able to take a look ?

waldyrious:
Just a quick note: I'll review this later today.

waldyrious:
Ok, I finally was able to set some time aside to do a proper review of this. Overall this seems to be progressing nicely.

I'm wondering how much work you expect to still be necessary for to the wiki page parsing (e.g. for getting the lists of commands for other platforms). I am afraid it can be a little fragile because the content of those pages can easily change in incompatible ways regarding the structure the code is expecting.

I suspect it would be much easier to get the data directly in a structured form, namely as queries to [Wikidata](https://www.wikidata.org). Here's an [example query](https://query.wikidata.org/#SELECT%20%3Fitem%20%3FitemLabel%20WHERE%20%7B%0A%20%20%3Fitem%20wdt%3AP31%20wd%3AQ18343316.%0A%20%20SERVICE%20wikibase%3Alabel%20%7B%20bd%3AserviceParam%20wikibase%3Alanguage%20%22en%22%20%7D%0A%7D%0ALIMIT%20100) to get all entries that are instances of "standard Unix utility" (we can combine additional properties to fetch only Linux commands, or widen the search to cover non-Unix commands). These queries can be fetched via a dedicated API; there's some information [here](https://www.wikidata.org/wiki/Wikidata:SPARQL_query_service), which even mentions a Python wrapper.

If you decide to proceed via Wikidata, I'll be able to help out in filling gaps on the data side, to ensure that at least the information currently in the Wikipedia pages is obtainable via Wikidata. Let me know.

As for the SVG badge itself -- it looks great, nice work there :) but I wonder if you're aware of shields.io's [custom badge feature](http://shields.io/#your-badge) that allows building personalized badges by simply using a particular image URL format: `https://img.shields.io/badge/<SUBJECT>-<STATUS>-<COLOR>.svg`. For example, `https://img.shields.io/badge/command%20coverage-30%25-brightgreen.svg` produces this:
> command coverage

(More complex and procedural generation is also possible, e.g. of hex color codes based on the coverage percentage, etc.). What do you think about leaving the image generation to them?

Finally, just a minor comment: I'd suggest keeping the unrelated changes (the capitalization of the license badge, and removal of the Tokei.rs badge) in a separate commit, so that the git history, once merged, can be more readable. But again, this is a minor issue, so it shouldn't hold up merging.

agnivade:
> These queries can be fetched via a dedicated API; there's some information here, which even mentions a Python wrapper.

This sounds great. I did not know about this. I will look into incorporating it.

> If you decide to proceed via Wikidata, I'll be able to help out in filling gaps on the data side, to ensure that at least the information currently in the Wikipedia pages is obtainable via Wikidata. Let me know.

That would be mighty fine :)

> As for the SVG badge itself -- it looks great, nice work there :) but I wonder if you're aware of shields.io's custom badge feature that allows building personalized badges by simply using a particular image URL format: https://img.shields.io/badge/<SUBJECT>-<STATUS>-<COLOR>.svg.

I didn't ! Very cool. Thanks for the tip

> Finally, just a minor comment: I'd suggest keeping the unrelated changes (the capitalization of the license badge, and removal of the Tokei.rs badge) in a separate commit, so that the git history, once merged, can be more readable. But again, this is a minor issue, so it shouldn't hold up merging.

Sure.

waldyrious:
Awesome. Once you get to the Wikidata queries then, let me know if you need any help with either tweaking the queries or fixing up the data. Ideally we should be able to filter directly at the source (e.g. by adding additional constraints to the query), rather than have to clean up the result.

agnivade:
Currently, I am very involved with the Go community and most of my time is being focused there. As of now, this PR needs non-trivial effort to be spent to bring it in shape.

I am going to close this for the time being and re-open when I find some more time in the future. My apologies for not being able to see this to the end.

waldyrious:
No problem, @agnivade, that's totally understandable (I'm also quite unavailable to provide continued attention to this project for the time being, due to various reasons as you know). When you eventually find the time and energy to work on this, do let me know if I can assist in any way, as I mentioned above.

zdroid:
@agnivade Should this branch be deleted or kept for your future work?

sbrl:
I think we ought to keep it, @zdroid. There's no harm in this, anyhow :P

agnivade:
I don't think I will be able to address this in the near future. We can keep this if somebody wants to reuse my work.

PR #2005, opened 2018-02-19
hexdump: add page

tzhenghao:


tzhenghao:
Hey @pxgamer, good question. I believe it's installable on Windows, but I'm not sure where this should live. Should I still slot it under `linux` directory? btw, this command is readily available on MacOS.

tzhenghao:
Another option is to put it under both `/linux` and `/osx` directories. I'm looking at another command like `htop` being slotted under `/linux`, but it works the same on MacOS. However, this markdown file isn't in the `/osx` directory.

agnivade:
Yes, let's go with both directories. Please ensure that the command usage is correct by testing on macOS :)

pxgamer:
Sounds good to me. :+1:

tzhenghao:
Hi @agnivade @pxgamer, I've made the appropriate changes and tested that it works on both Linux an MacOS. Please have a look when you get the chance. Thanks!

PR #2014, opened 2018-03-03
fdisk: add page

avalanchy:
I think the most common usage, listing available devices.

### DIFF
--- a/pages/linux/fdisk.md
+++ b/pages/linux/fdisk.md
@@ -1,0 +1,7 @@
+# fdisk
+
+> Manipulate disk partition table
+
+- View all available partitions:
+
+`fdisk -l`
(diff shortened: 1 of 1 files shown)


pages/linux/fdisk.md
```diff
@@ -0,0 +1,7 @@
+# fdisk
+
+> Manipulate disk partition table.
+
+- View all available partitions:
```
agnivade: Add the option to change partition table - `fdisk {{/dev/sda}}` first.

avalanchy: Ok fine, I added that but as a second option. Example of editing before listing does not make any sense.

agnivade: Agree.

avalanchy:
:tada:

PR #2015, opened 2018-03-06
gzip: Remove gz extension from output file

DonoA:
From my testing in Ubuntu 16.04, the compressed files do not have the gz extension. Removed this to alleviate a little confusion.

----

### DIFF
--- a/pages/common/gzip.md
+++ b/pages/common/gzip.md
@@ -12,7 +12,7 @@
 
 - Compress a file specifying the output filename:
 
-`gzip -c {{file.ext}} > {{compressed_file.ext}}.gz`
+`gzip -c {{file.ext}} > {{compressed_file.ext}}`
 
 - Uncompress a gzipped file specifying the output filename:
 
@@ -20,4 +20,4 @@
 
 - Specify the compression level. 1=Fastest (Worst), 9=Slowest (Best), Default level is 6:
 
-`gzip -9 -c {{file.ext}} > {{compressed_file.ext}}.gz`
+`gzip -9 -c {{file.ext}} > {{compressed_file.ext}}`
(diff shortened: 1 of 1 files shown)


agnivade:
Thanks for the contribution @DonoA.

Technically you are right, but it will look more confusing if both the original file and the compressed file have the same extension. Especially for a binary file like .gz to have an extension like .txt. Also, putting the right extension shows the proper icon in UI and a different color in the console. Not having the extension will deliberately break that convention.

DonoA:
I would completely agree, however when decompressing a file the gz is added implicitly which would imply it is added implicitly when compressing.

This lead me to accidentally overwrite the very document I was trying to compress!

agnivade:
I think you meant while decompressing, the gz extension is _removed_ implicitly. While compressing, it is _indeed_ added implicitly, but you have updated the wrong example. The first example compresses a file, and creates a new file with the gz extension added. But it also deletes the original file. If you want to keep both files, then you need to redirect output and specify a filename.

I guess you want to use the 2nd method ?

Removing the .gz extension from the file name still has nothing to do with your usecase IIUC. A gzipped file should always have a .gz extension to avoid confusion.

DonoA:
I think you are correct that the tldr page should support the fact that the file should always have a gz extension when compressing, however I changed it to be more like the tar page which includes the gz extension inside the name to show it is recommended.

agnivade:
Should we update the `-d` cases too to move the extension inside the token ?

DonoA:
Its completely up to you but my original point earlier was that the gz extension is actually added automatically by gzip (at least on ubuntu 16.04, gzip 1.6) so it makes sense to have it the way it is.

agnivade:
Agree

sbrl:
Thanks, @DonoA!

PR #2021, opened 2018-03-08
git-sizer: add page

kevin91liu:
----

For additional details, see here:
https://github.com/github/git-sizer

### DIFF
--- a/pages/common/git-sizer.md
+++ b/pages/common/git-sizer.md
@@ -1,9 +1,8 @@
 # git sizer
 
 > Computes various Git repository size metrics and alerts you to any that might cause problems or inconvenience.
-> Change to the directory containing your repository before running this command.
 
-- Report only statistics that have >0 level of concern:
+- Report only statistics that have a level of concern greater than 0:
 
 `git sizer`
 
--- a/pages/common/git-sizer.md
+++ b/pages/common/git-sizer.md
@@ -11,6 +11,6 @@
 
 `git sizer -v`
 
-- See command line options:
+- See additional options:
 
 `git sizer -h`
--- a/pages/common/git-sizer.md
+++ b/pages/common/git-sizer.md
@@ -1,0 +1,16 @@
+# git sizer
+
+> Computes various Git repository size metrics and alerts you to any that might cause problems or inconvenience.
+> Change to the directory containing your repository before running this command.
+
+- Report only statistics that have >0 level of concern:
+
+`git sizer`
+
+- Report all statistics:
+
+`git sizer -v`
+
+- See command line options:
+
+`git sizer -h`


pages/common/git-sizer.md
```diff
@@ -0,0 +1,16 @@
+# git sizer
```
sbrl: Are you sure this is spelt correctly? I get a `sizer` is not a git command.

pages/common/git-sizer.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Report only statistics that have >0 level of concern:
+
+`git sizer`
+
+- Report all statistics:
+
+`git sizer -v`
+
+- See command line options:
```
sbrl: Perhaps `See additional help` might work better here? `See command line options` would imply to me that we aren't covering _any_ command like options in this tldr page, which is obviously not the case.

pages/common/git-sizer.md
pxgamer: If it's in the path, it should work with a space I think. Otherwise `git-sizer` should be used.

Looks like it is spelt correctly though.

kevin91liu: You can invoke it by running either `git-sizer` or `git sizer`. I've done it with both.

pages/common/git-sizer.md
kevin91liu: Sure, I'll change it to `See additional options`

pages/common/git-sizer.md
agnivade: Yep, this is fine.

pages/common/git-sizer.md
```diff
@@ -0,0 +1,16 @@
+# git sizer
+
+> Computes various Git repository size metrics and alerts you to any that might cause problems or inconvenience.
+> Change to the directory containing your repository before running this command.
```
agnivade: Is there a reason to write this line ? You always need to run any git commands inside a git repo. That is required for _all_ git commands. I am wondering whether we should mention this explicitly since its pretty commonly known.

pages/common/git-sizer.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Report only statistics that have >0 level of concern:
```
agnivade: Let's change this to `Report only statistics that have a level of concern greater than 0`

pages/common/git-sizer.md
pxgamer: I agree that it probably should be removed. It's implied, and `git sizer` returns an error anyway to notify the user when it's not run in a Git directory. 👍

pages/common/git-sizer.md
sbrl: Huh. It must be an extension or bundled as part of a newer version of git then.

PR #2026, opened 2018-03-12
wget.md: add quiet mode

kishaningithub:
----

### DIFF
--- a/pages/common/wget.md
+++ b/pages/common/wget.md
@@ -27,17 +27,9 @@
 
 `wget --ftp-user={{username}} --ftp-password={{password}} {{ftp://example.com}}`
 
-- Limit download speed to 200 kB/s:
-
-`wget --limit-rate={{200k}} {{https://example.com}}`
-
 - Continue an incomplete download:
 
 `wget -c {{https://example.com}}`
-
-- Retry a given number of times if the download doesn't succeed at first:
-
-`wget -t {{number_of_retries}} {{https://example.com}}`
 
 - Make the progress bar disappear:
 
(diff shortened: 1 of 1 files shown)


pxgamer:
Hi @kishaningithub, thanks for the page edit.

Unfortunately, we have a hard limit of 8 for the number of examples that we allow in a page. I'm not actually sure why there are already 9 examples in the wget page. 🤔

agnivade:
Because the linter does not catch those errors yet.

agnivade:
@kishaningithub - Do you see any option that is worth removing so that your option can be added ? I think limit download-rate is a pretty obscure one and can be replaced with this. `-q` is more likely to be helpful that limiting the download rate.

Also, maybe we can remove the `-t` option to bring it to 8.

ping @pxgamer , @sbrl - thoughts ?

pxgamer:
@agnivade, sounds fine to me. 👍

pages/common/wget.md
```diff
@@ -27,14 +27,10 @@
...
-- Limit download speed to 200 kB/s:
-
-`wget --limit-rate={{200k}} {{https://example.com}}`
-
 - Continue an incomplete download:
 
 `wget -c {{https://example.com}}`
 
-- Retry a given number of times if the download doesn't succeed at first:
+- Make the progress bar disappear:
```
agnivade: We usually prefer a formal tone, and it would be good to start a word with the flag. Perhaps - `Enable quiet mode to suppress output` is better. This also clarifies that not only the progress bar is hidden but all error messages and information.

PR #2030, opened 2018-03-13
certbot: add page

sbrl:
----


pages/linux/letsencrypt.md
```diff
@@ -0,0 +1,15 @@
+# letsencrypt
+
+> The Let's Encrypt Agent for automatically obtaining and renewing TLS certificates.
+
+- Obtain a new certificate via webroot authorization:
+
+`sudo letsencrypt --webroot --webroot-path {{path/to/webroot}} --domain {{subdomain.example.com}}`
+
+- Oobtain an untrusted test certificate instead:
```
pxgamer: Minor typo here with `Obtain`.

agnivade:
Hmm, I am a bit confused. I thought people used certbot ? There is a CLI for letsencrypt also ? Or are both the same ?

pxgamer:
I was actually thinking of Certbot. 🤔 Looking through some repos, this may be the predecessor to Certbot.

sbrl:
Huh :thinking: I thought that `letsencrypt` was the new one, @agnivade? Can't remember. It works for me :P

@pxgamer: Fixed - thanks for spotting that :-)

agnivade:
Ah, found it here https://github.com/certbot/certbot

> Until May 2016, Certbot was named simply letsencrypt or letsencrypt-auto, depending on install method. Instructions on the Internet, and some pieces of the software, may still refer to this older name.

I think it might be better to use `certbot` rather than `letsencrypt`. What do you think ? 

Also, this is a very important tool, so I would like to expand a bit more to cover more usecases. Specifically I would like to cover all these points - 

```
(default) run   Obtain & install a certificate in your current webserver
    certonly        Obtain or renew a certificate, but do not install it
    renew           Renew all previously obtained certificates that are near expiry
   -d DOMAINS       Comma-separated list of domains to obtain a certificate for

  --apache          Use the Apache plugin for authentication & installation
  --standalone      Run a standalone webserver for authentication
  --nginx           Use the Nginx plugin for authentication & installation
  --webroot         Place files in a server's webroot folder for authentication
  --manual          Obtain certificates interactively, or using shell script hooks

   -n               Run non-interactively
  --test-cert       Obtain a test certificate from a staging server
  --dry-run         Test "renew" or "certonly" without saving any certificates to disk
```

Let's start begin with `certonly --webroot` and then add `--apache` and `--nginx`. Then `renew` and then finally move to `dry-run` and `test-cert` at the end.

sbrl:
Are the switches the same as those I've detailed? I don't actually have `certbot` installed on my server, so I can't test it.

Furthermore, I can't test the `--nginx` and `--apache` options, since I've got it set up for `--webroot` - and I don't want it messing with my configuration files :-(

agnivade:
I think it's okay if we do not practically test every command. We can read the documentation and explain each options clearly. That should be good enough. I can help put up a rough draft if you want.

Yes, the switches are the same I believe.

sbrl:
Ah, ok :-)

How's that, @agnivade?

pages/linux/certbot.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Obtain a new certificate via nginx authorization, installing the new certificate automatically:
+
+`sudo certbot --nginx --domain {{subdomain.example.com}}`
+
+- Obtain a new certificate via apache authorization, installing the new certificate automatically:
+
+`sudo certbot --apache --domain {{subdomain.example.com}}`
+
+- Renew all Let's Encrypt certificates that expire soon (don't forget to restart any servers that use them afterwards):
```
agnivade: What is the actual expiry threshold ?

pages/linux/certbot.md
```diff
@@ -0,0 +1,28 @@
+# certbot
+
+> The Let's Encrypt Agent for automatically obtaining and renewing TLS certificates.
+> Successor to `letsencrypt`.
+
+- Obtain a new certificate via webroot authorization:
```
agnivade: We should clarify that it just obtains the cert  but does not install it.

pages/linux/certbot.md
pxgamer: As far as I know it's 30 days. :thinking:

agnivade: Ok, we should just mention the exact days then. Much more helpful IMO.

sbrl:
Great points, @agnivade and @pxgamer! How's that?

agnivade:
Cool. Do you think we should add the `--manual` option too ?

sbrl:
Hrm. I think that might be an advanced usage, since the man page specifically details that it's laborious - it's much better to automate the process I think. If you don't want the certificate installed automatically, I think that the `--webroot` option should cover almost all other use cases.

agnivade:
Ok.

pxgamer:
All looks good to me. :+1:

PR #2033, opened 2018-03-17
sips: add page

martin-juul:
----


pages/osx/sips.md
```diff
@@ -0,0 +1,28 @@
+# sips
+
+> Apple Scriptable Image Processing System.
+> Raster/Query images and ColorSync ICC Profiles.
+
+- Specify an output directory so originals does not get modified:
```
agnivade: so _that_ originals _do_ not get modified.

pages/osx/sips.md
```diff
@@ -0,0 +1,28 @@
...
+
+`sips --out /path/to/out ...`
```
agnivade: path/to/out should be in tokens `{{}}`. Also, what are the `...` for ?

pages/osx/sips.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Resample image at specified size, Image aspect ratio may be altered:
```
agnivade: Should it be resample or resize ?

martin-juul: this [page](http://www.richmediacs.com/user_manuals/RMCS_PS_Training/ResizeVsResample.html) explains the difference between resizing and resampling well.

The tldr, is when resizing you're redistributing the amount of pixels. Compared to when you're resampling you're adding or subtracting the amount of pixels in the image.

pages/osx/sips.md
martin-juul: I will fix that, the dots were meant to imply the rest of the users command (appending to another command). Should i change it to {{rest_of_command}} or just remove it?

pages/osx/sips.md
martin-juul: Sorry, i'll fix that.

PR #2035, opened 2018-03-19
opkg: add page

monsieurlo:
----

### DIFF
--- a/pages/linux/opkg.md
+++ b/pages/linux/opkg.md
@@ -29,11 +29,3 @@
 - List all the available packages:
 
 `opkg list`
-
-- List all the installed packages:
-
-`opkg list-installed`
-
-- List all the upgradable packages:
-
-`opkg list-upgradable`
--- a/pages/linux/opkg.md
+++ b/pages/linux/opkg.md
@@ -1,6 +1,6 @@
 # opkg
 
-> The opkg utility is a lightweight package manager used to install OpenWrt packages.
+> A lightweight package manager used to install OpenWrt packages.
 
 - Install a package:
 
--- a/pages/linux/opkg.md
+++ b/pages/linux/opkg.md
@@ -14,9 +14,13 @@
 
 `opkg update`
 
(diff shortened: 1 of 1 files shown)


pages/linux/opkg.md
```diff
@@ -0,0 +1,35 @@
+# opkg
+
+> The opkg utility is a lightweight package manager used to install OpenWrt packages.
```
agnivade: Remove "The opkg utility is"

pages/linux/opkg.md
```diff
@@ -0,0 +1,35 @@
...
+
+`opkg remove {{wget}}`
+
+- Update the list of available packages:
+
+`opkg update`
+
+- Upgrade installed packages:
+
+`opkg upgrade`
```
agnivade: Isn't there a way to upgrade a single package ?

pxgamer: Yes there is. `opkg upgrade {{package(s)}}` (where packages are a space-separated list).

agnivade:
Thanks.

But now the no. of examples have exceeded 8. I suggest that we remove `opkg list-installed` and `opkg list-upgradable` since there is already a `list` example.

pages/linux/opkg.md
```diff
@@ -0,0 +1,31 @@
+# opkg
+
+> A lightweight package manager used to install OpenWrt packages.
+
+- Install a package:
+
+`opkg install {{wget}}`
```
pxgamer: Just a minor query, would it be better to use `{{package}}` instead of `{{wget}}` for each example. I understand that `wget` is an example, but it would just add consistency with the `{{package(s)}}` usage.

pxgamer: Just a minor query, would it be better to use `{{package}}` instead of `{{wget}}` for each example. I understand that `wget` is an example, but it would just add consistency with the `{{package(s)}}` usage.

PR #2038, opened 2018-03-19
units: add page

ryangreenberg:

### DIFF
--- a/pages/linux/units.md
+++ b/pages/linux/units.md
@@ -1,0 +1,24 @@
+# units
+
+> Provide the conversion between two units of measure.
+> Typing 'search {{text}}' in the prompt will display a list of all of the units containing {{text}}.
+
+- Run in interactive mode:
+
+`units`
+
+- Show the conversion between two simple units:
+
+`units {{quarts}} {{tablespoons}}`
+
+- Convert between units with quantities:
+
+`units {{15 pounds}} {{kilograms}}`
+
+- Show the conversion between two compound units:
+
+`units {{"meters / second"}} {{"inches / hour"}}`
+
+- Show the conversion between units with different dimensions:
+
+`units {{"acres"}} {{"ft^2"}}`
--- a/pages/osx/units.md
+++ b/pages/osx/units.md
@@ -1,0 +1,23 @@
+# units
+
+> Provide the conversion between two units of measure.
+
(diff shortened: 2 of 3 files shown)


pages/common/units.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Show the conversion between two simple units:
+
+`units quarts tablespoons`
+
+- Show the conversion between two compound units:
+
+`units "meters / second" "inches / hour"`
+
+- Run `units` in interactive mode:
```
agnivade: This should be the first example.

pages/common/units.md
```diff
@@ -0,0 +1,15 @@
+# units
+
+> Provide the conversion between two units of measure.
+
+- Show the conversion between two simple units:
+
+`units quarts tablespoons`
```
agnivade: Values should be inside tokens (`{{}}`).

pages/common/units.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Show the conversion between two compound units:
+
+`units "meters / second" "inches / hour"`
```
agnivade: Same as above.

pages/common/units.md
```diff
@@ -0,0 +1,15 @@
+# units
+
+> Provide the conversion between two units of measure.
+
```
agnivade: Add this line in a sub-heading-  `Typing 'search {{text}}' in the prompt  will display a list of all of the units containing {{text}}.`

ryangreenberg: Looks like that is the behavior of GNU units but not the units that is installed by default on, e.g. OS X. How do you want to handle the difference?

ryangreenberg:
@agnivade Thank you for the feedback! I added a new example that included quantities (to help people understand they can use `units` with or without them).

pages/common/units.md
agnivade: Ah, in that case, you need to create separate copies of the page inside linux, windows and osx folders.

agnivade:
We need to take care of this https://github.com/tldr-pages/tldr/pull/2038#discussion_r175869313

ryangreenberg:
@agnivade I added your suggested search message to the linux page. I don't know what (if anything) the windows version supports so I left it the same as the osx page.

PR #2043, opened 2018-03-24
ncat: add -k option

zlbabe:
This option let ncat accept multiple connections and keep them open after they are closed.

### DIFF
--- a/pages/linux/ncat.md
+++ b/pages/linux/ncat.md
@@ -6,7 +6,7 @@
 
 `ncat -l {{port}} > {{/path/to/file}}`
 
-- Accept multiple connections and keep them open after they have been closed:
+- Accept multiple connections and keep ncat open after they have been closed:
 
 `ncat -lk {{port}}`
 
--- a/pages/linux/ncat.md
+++ b/pages/linux/ncat.md
@@ -4,7 +4,7 @@
 
 - Listen for input on the specified port and write it to the specified file:
 
-`ncat -l {{port}} > {{/path/to/file}}`
+`ncat -l {{port}} > {{path/to/file}}`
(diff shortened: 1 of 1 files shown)


pages/linux/ncat.md
```diff
@@ -9,3 +9,7 @@
 - Write output of specified file to the specified host on the specified port:
 
 `ncat {{address}} {{port}} < {{/path/to/file}}`
+
+- Accept multiple connections and keep them open after they have been closed:
```
sbrl: What is meant by "keep them open after they have been closed"? Why would I want it to do that?

Just curious.

agnivade: I think this should be the second example ?

agnivade: The wording is incorrect here. It is not the _connection_ that remains open, rather _ncat_ itself which is open.

This section from the man page will clarify - 

```
-k, --keep-open (Accept multiple connections) .
           Normally a listening server accepts only one connection and then quits when the connection is closed. This option makes it
           accept multiple simultaneous connections and wait for more connections after they have all been closed. It must be combined with
           --listen. In this mode there is no way for Ncat to know when its network input is finished, so it will keep running until
           interrupted. This also means that it will never close its output stream, so any program reading from Ncat and looking for
           end-of-file will also hang.
```

agnivade:
@zlbabe - Would you like to update the wording as discussed here https://github.com/tldr-pages/tldr/pull/2043#discussion_r176913151 ?

pages/linux/ncat.md
sbrl: Ah, I see.

PR #2069, opened 2018-04-13
jq: accepts a file as last parameter

ericbn:
Examples look simpler with file as last parameter than with using `cat`,
also they explicitly show that you must use `.` as the default
all-matching rule.

Keep the last two examples with piping, and change their descriptions
accordingly.

### DIFF
--- a/pages/common/jq.md
+++ b/pages/common/jq.md
@@ -4,24 +4,24 @@
 
 - Output a JSON file, in pretty-print format:
 
-`cat {{file}} | jq`
+`jq . {{file.json}}`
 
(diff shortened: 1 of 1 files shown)


agnivade:
How about we leave the last 2 examples to read from stdin ? Several times, I try to filter a json stream from other sources. We should keep some of those.

ericbn:
@agnivade, totally agree. Should we change the description of these examples to show some differentiation?

agnivade:
Yep.

ericbn:
Updated commit.

PR #2070, opened 2018-04-15
cu: add page

tulik:

### DIFF
--- a/pages/common/cu.md
+++ b/pages/common/cu.md
@@ -5,7 +5,7 @@
 
 - To connect to a remote system venus, enter:
 
-`cu venus`
+`cu {{venus}}`
 
 - To log in to a system connected by a hardwired line asynchronous line, enter:
 
--- a/pages/common/cu.md
+++ b/pages/common/cu.md
@@ -3,7 +3,7 @@
 > Call up another system.
 > It can also do simple file transfers with no error checking.
 
-- To connect to a remote system venus, enter: 
+- To connect to a remote system venus, enter:
 
 `cu venus`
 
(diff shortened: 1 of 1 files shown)


pages/common/cu.md
```diff
@@ -0,0 +1,12 @@
+# cu
+
+> Call up another system.
+> It can also do simple file transfers with no error checking.
```
sbrl: You mention that it can do simple file transfers, but you don't include an example for this?

pxgamer: According to the docs, it looks like these are done with the following:

Send a file to the remote system, assuming that the remote expects it:
`cu ~> {{path/to/file}}`
Receive a file from the remote system, accepting data until the contents of the eofread variable are seen:
`cu ~<`
Send a file to a remote Unix system:
`cu ~p {{path/to/local_file}}  {{path/to/remote_file}}`
Retrieve a file from a remote Unix system:
`cu ~t {{path/to/remote_file}} {{path/to/local_file}}`

sbrl: Ah. Maybe we should look at including some of those in the page, then? And are those tilde `~` characters there? In bash at least those would probably need to be in single quotes to avoid getting auto-expanded to `$HOME`.

pxgamer: Oh, actually I'm just being ignorant.

The current examples do seem to be the only ones really. It looks like the transfer things I mentioned are once you've run the command. I assume it opens kind of a subshell.

agnivade: I think it should be fine to add those to make the page complete. Since, the secondary header mentions file transfer, the page will be incomplete without adding a few examples showing how to do that.

sbrl: What's the status on this?

agnivade: pending on OP.

agnivade:
ping @tulik - any update on this ?

PR #2072, opened 2018-04-16
dbus-daemon: add page

### LINKED ISSUE #937: page request: dbus-daemon
dbus-daemon man page request

pxgamer:
Closes #937

----

### DIFF
--- a/pages/linux/dbus-daemon.md
+++ b/pages/linux/dbus-daemon.md
@@ -1,0 +1,31 @@
+# dbus-daemon
+
+> The D-Bus message daemon, allowing multiple programs to exchange messages.
+
+- Run the daemon with a configuration file:
+
+`dbus-daemon --config-file {{path/to/file}}`
+
+- Run the daemon with the standard per-login-session message bus configuration:
+
+`dbus-daemon --session`
+
+- Run the daemon with the standard systemwide message bus configuration:
+
+`dbus-daemon --system`
+
+- Set the address to listen on and ignore the configuration for this:
+
+`dbus-daemon --address {{address}}`
+
+- Output the process id to stdout:
+
+`dbus-daemon --print-pid`
+
+- Force the message bus to write to the system log for messages:
+
+`dbus-daemon --syslog`
+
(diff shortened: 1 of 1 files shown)


pages/linux/dbus-daemon.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Run the daemon with the standard per-login-session message bus configuration:
+
+`dbus-daemon --session`
+
+- Run the daemon with the standard systemwide message bus configuration:
+
+`dbus-daemon --system`
+
+- Set the address to listen on and ignore the configuration for this:
```
agnivade: What does "this" mean here ? Do you mean the configuration file ?

pages/linux/dbus-daemon.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Output the process id to stdout:
+
+`dbus-daemon --print-pid`
+
+- Force the message bus to write to the system log for messages:
+
+`dbus-daemon --syslog`
+
+- Disable the creation of a pid file and ignore the configuration for this:
```
agnivade: Is this a common enough use-case ? I believe usually you would want to use a pid file ? (Also, same question as above)

pxgamer: I guess probably not really. 🤔

pages/linux/dbus-daemon.md
pxgamer: "this" is referring to the configuration setting inside the config file. There should be a value in the file such as `<listen>unix:path=/tmp/foo</listen>`. I guess maybe `and override any value in the configuration file`? Or something like that.

agnivade: Ah ok, I think it maybe be better written as - `Set the address to listen on and override the configuration value for it.`

pages/linux/dbus-daemon.md
pxgamer: Sure, I'll remove it when I get to work. :+1:

PR #2088, opened 2018-05-03
gnuplot: add page

sbrl:
I've been using this for my University work and I saw it didn't have a page, so I've created one :smiley_cat: 

----

### DIFF
--- a/pages/common/gnuplot.md
+++ b/pages/common/gnuplot.md
@@ -1,6 +1,6 @@
 # gnuplot
 
-> Optionally interactive graph plotter that outputs in several formats.
+> A graph plotter that outputs in several formats.
 
 - Start the interactive graph plotting shell:
 
@@ -10,9 +10,9 @@
 
 `gnuplot {{path/to/definition.plt}}`
 
-- Set the output format by executing a command before loading the definition file (in this case output a PNG to stdout):
+- Set the output format by executing a command before loading the definition file:
 
-`gnuplot -e "{{set terminal png size 1024,768}}" {{path/to/definition.plt}} >{{path/to/output_graph.png}}`
+`gnuplot -e "{{set output "path/to/filename.png" size 1024,768}}" {{path/to/definition.plt}}`
 
 - Persist the graph plot preview window after gnuplot exits:
 
--- a/pages/common/gnuplot.md
+++ b/pages/common/gnuplot.md
@@ -12,7 +12,7 @@
 
 - Set the output format by executing a command before loading the definition file (in this case output a PNG to stdout):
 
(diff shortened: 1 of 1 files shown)


pages/common/gnuplot.md
```diff
@@ -0,0 +1,19 @@
+# gnuplot
+
+> Optionally interactive graph plotter that outputs in several formats.
```
agnivade: What does "Optionally" mean here ? It's not very clear.

pages/common/gnuplot.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Start the interactive graph plotting shell:
+
+`gnuplot`
+
+- Plot the graph for the specified graph definition file:
+
+`gnuplot {{path/to/definition.plt}}`
+
+- Set the output format by executing a command before loading the definition file (in this case output a PNG to stdout):
```
agnivade: "output a PNG to stdout" means dumping the binary file contents to stdout. That is probably
 not what is happening. Maybe better to say "redirect output to a png file".

pages/common/gnuplot.md
```diff
@@ -0,0 +1,19 @@
...
+
+`gnuplot -e "{{set terminal png size 1024,768}}" {{path/to/definition.plt}} >{{path/to/output/graph.png}}`
```
agnivade: path/to/output_graph.png

pages/common/gnuplot.md
sbrl: Basically you can either interact with it via a sort of shell, or put the same commands in a file and tell it to parse it (which is what's done most of the time).

pages/common/gnuplot.md
sbrl: Hrm, yeah. I had trouble with this, because whether it outputs to stdout or not depends on the graph definition file. For example, I could say this:

```
set terminal png size 1024,768
```

...which would output the png to the standard output, or this:

```
set output "filename.png"
```

....which would output to a file directly..

agnivade: Maybe just go with the most common usage and mention `set output "filename.png"` ? No need to cover all scenarios and make the description long in that process

pages/common/gnuplot.md
agnivade: Got it. I think it's an implementation detail which need not be mentioned in the heading. "A graph plotter that ..." sounds just fine to me.

pages/common/gnuplot.md
sbrl: True. Interestingly enough, it's actually `set terminal png` that you see all over stackoverflow - even though it's not as useful as `set output`. Still,  agree - I'll change it.

PR #2098, opened 2018-05-06
rtcwake: add page

FraYoshi:
Added rtcwake Linux page and examples
----


pages/linux/rtcwake.md
```diff
@@ -0,0 +1,36 @@
+# rtcwake
+
+> Enter a system sleep state until specified wakeup time relative to your bios clock.
+> Commands need to be run as superuser since shutdowns are involved.
+
+- Check your hardware clock info and exit:
+
+`rtcwake -m show -v`
+
+- Show if an allarm is set:
```
sbrl: Typo here? `allarm` -> `alarm`

pages/linux/rtcwake.md
```diff
@@ -0,0 +1,36 @@
...
+
+`rtcwake -m show`
+
+- Disable a previous set alarm:
+
+`rtc -m disable`
+
+- Suspend to ram and wakeup after 10 seconds:
+
+`# rtcwake -m mem -s {{10}}`
```
sbrl: Is the hash needed here? If it's denoting a superuser command, perhaps you could use `sudo ` instead?

pages/linux/rtcwake.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Suspend to disk _(higher power saving)_ and wakeup 10 minutes later:
```
sbrl: Not sure on italics and whether it renders correctly in the terminal. @agnivade, could you confirm?

pages/linux/rtcwake.md
```diff
@@ -0,0 +1,36 @@
...
+
+`# rtcwake -m disk --date +{{15}}m`
+
+- Suspend to disk and wakeup 1 hour later so play some music with mpv:
```
sbrl: Maybe ....`so play some music with mpv` would read better as ....`and play some music with mpv`?

pages/linux/rtcwake.md
pxgamer: I don't think it usually does, at least not with a few of the clients. :thinking:

pages/linux/rtcwake.md
FraYoshi: Yes, typo.. gonna correct thanks =)

pages/linux/rtcwake.md
FraYoshi: It's needed to be superuser, sincerely, I was unsure about using the hash or the sudo.. Gonna switch to the last one =)

pages/linux/rtcwake.md
FraYoshi: I was also unsure about the use of `mpv`.. It's a cool combination but maybe "and run a command" would be better because not every system has `mpv` installed.. i.e. they could have `vlc`

pages/linux/rtcwake.md
agnivade: It works fine on the node client. But usually we avoid markdown in the description, and this seems to be an incorrect usage of italicizing a text anyways. We should remove the underscores.

pages/linux/rtcwake.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Disable a previous set alarm:
+
+`rtc -m disable`
+
+- Suspend to ram and wakeup after 10 seconds:
+
+`sudo rtcwake -m mem -s {{10}}`
+
+- Suspend to disk (higher power saving) and wakeup 10 minutes later:
```
agnivade: The description mentions 10 minutes, but example has 15 minutes. Please choose one.

Also, we should be consistent in the language. The description before has "after 10 seconds", but this one has "10 minutes later". We should choose one and be consistent.

pages/linux/rtcwake.md
```diff
@@ -0,0 +1,36 @@
...
+
+`sudo rtcwake -m disk --date +{{15}}m`
+
+- Suspend to disk and wakeup 1 hour later and play some music with mpv:
```
agnivade: Why is this example necessary ? I don't see it adding anything new except using hour as a unit.

pages/linux/rtcwake.md
```diff
@@ -0,0 +1,36 @@
...
+
+`sudo rtcwake -m disk --date +{{1}}h && mpv {{music_folder/*}}`
+
+- Freeze the system (more efficent than suspend-to-ram but linux > 3.9 required) and wakeup at a given date and time:
+
+`sudo rtcwake -m freeze --date {{YYYYMMDDhhmm}}`
+
+- Run a test (CTRL-C to abort) in which the computer is waked at a given time:
```
agnivade: What does the "test" do ? Is this sort of a dry run ?

pages/linux/rtcwake.md
FraYoshi: Yes, I thought that adding more examples was better for a higher understanding of how the thing works.. so s for seconds, m for minutes and so on.. Or maybe Is better a longer description?

pages/linux/rtcwake.md
```diff
@@ -0,0 +1,36 @@
+# rtcwake
+
+> Enter a system sleep state until specified wakeup time relative to your bios clock.
+> Commands need to be run as superuser since shutdowns are involved.
+
+- Check your hardware clock info and exit:
```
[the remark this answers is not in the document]
FraYoshi: So removing It but keeping the `-v` flag since avoiding implies a CTRL-C command after.

pages/linux/rtcwake.md
FraYoshi: Yes, a sort of dry run. So maybe better adding It at the top?

pages/linux/rtcwake.md
FraYoshi: It's a useful usage example imo... for example, I use the command mainly as a wake-up alarm.

pages/linux/rtcwake.md
FraYoshi: Or maybe a standard time like 1h and the thing written in 3 different ways in 3 different examples like: 3600s, 60m and 1h!

agnivade: We already have 15m and 1h, so I think it is clear enough that different units can be used.

pages/linux/rtcwake.md
agnivade: This is showing composition of one command with another command. You are combining the wakeup action with playing music after it to make it behave like an alarm. Strictly speaking, this behavior is outside the scope of a tldr page for rtcwake, which should just be focused on explaining the command usage. A user can easily construct an example like this if he/she understands the command well enough (which should be our primary aim).

pages/linux/rtcwake.md
agnivade: It's fine. But perhaps we can modify the description to - `Perform a dry run to wakup the computer at a given time. (Press Ctrl + C to abort):`

FraYoshi:
Now It should be more clear.
I take a moment to thank you for enduring my mistakes ^^"

now there's an example involving `-s` (seconds) and one involving `--date` (time_units)
do you think It's fine keeping both?

agnivade:
Yes it's fine. Thanks.

agnivade:
@sbrl / @pxgamer - PTAL. Merge if you think it's fine.

PR #2100, opened 2018-05-06
ipfs: add page

FraYoshi:
Added ipfs to common.
It actually has 10 examples.. Any suggestion on how to reduce them? Actually, I find important all of them as a not daily user reference .-.
----

### DIFF
--- a/pages/common/ipfs.md
+++ b/pages/common/ipfs.md
@@ -3,15 +3,15 @@
 > Inter Planetary File System.
 > A peer-to-peer hypermedia protocol to make the web faster, safer, and more open.
 
-- Add a file from local to the file system and pin It:
+- Add a file from local to the file system and pin it:
 
 `ipfs add {{filename}}`
 
-- Add a folder and its files recursively from local to the File System:
+- Add a folder and its files recursively from local to the file system:
 
 `ipfs add -r {{folder}}`
 
-- Save a remote file and give It a name but not pin It:
+- Save a remote file and give it a name but not pin it:
 
 `ipfs get {{hash}} -o {{filename}}`
 
--- a/pages/common/ipfs.md
+++ b/pages/common/ipfs.md
@@ -7,11 +7,11 @@
 
 `ipfs get {{hash}} -o {{filename}}`
 
-- Add a file from local to the File System with the filename _(-w)_, pin It and print the hash:
+- Add a file from local to the file system with the filename (-w) and pin It:
 
 `ipfs add {{filename}} -w`
 
-- Add a folder and files in It from local to the File System, pin It and print the hash:
+- Add a folder and files in It from local to the File System and pin It:
 
 `ipfs add -r {{folder}}`
 
@@ -27,11 +27,7 @@
 
 `ipfs pin rm {{hash}}`
 
-- Create a new ipfs ID with a given name:
-
-`ifps key gen --type=rsa --size=2048 {{name}}`
-
-- Add an hash to the IPNS with the given ID _(omit --key for default)_. If is the hash of a folder, It will add the whole folder with files:
+- Add an hash to the IPNS with the given ID (omit --key for default). If is the hash of a folder, It will add the whole folder with files:
 
 `ipfs name publish --key={{ID}} {{hash}}`
 
--- a/pages/common/ipfs.md
+++ b/pages/common/ipfs.md
@@ -3,11 +3,11 @@
 > Inter Planetary File System.
 > A peer-to-peer hypermedia protocol to make the web faster, safer, and more open.
 
-- Add a file from local to the file system and pin it:
+- Add a file from local to the file system, pin it and print the relative hash:
 
 `ipfs add {{filename}}`
 
-- Add a folder and its files recursively from local to the file system:
+- Add a folder and its files recursively from local to the file system and print the relative hash:
 
 `ipfs add -r {{folder}}`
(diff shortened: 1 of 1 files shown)


FraYoshi:
not sure how to solve this error.......

agnivade:
The error is clear. On line 11, it is expecting a newline, but you have an empty space. Always configure your editor to trim trailing white spaces.

pages/common/ipfs.md
```diff
@@ -0,0 +1,44 @@
+# ipfs
+
+> Inter Planetary File System.
+> A peer-to-peer hypermedia protocol to make the web faster, safer, and more open.
+
+- Save a remote file and give It a name but not pin It:
+
+`ipfs get {{hash}} -o {{filename}}`
```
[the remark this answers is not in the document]
FraYoshi: It is actually the hash of the file... the IPFS is all about these... A file is retrieved inserting the hash and not the address to the file.

pages/common/ipfs.md
```diff
@@ -0,0 +1,44 @@
...
+
+- Add a file from local to the File System with the filename _(-w)_, pin It and print the hash:
```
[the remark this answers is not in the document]
FraYoshi: correcting the lower case thing! maybe the whole description could be changed like so:
"Add a file and Its name to the file system and pin It. Omit the -w flag to add just the file"

I see that on the related command I wrote {{filename}}... maybe could be better {{file_to_add}} or just {{file}}

pages/common/ipfs.md
```diff
@@ -0,0 +1,44 @@
...
+
+- Check pinned files:
+
+`ipfs pin ls`
+
+- Unpin a file from the local storage:
+
+`ipfs pin rm {{hash}}`
+
+- Create a new ipfs ID with a given name:
```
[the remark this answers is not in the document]
FraYoshi: ock, removed.

pages/common/ipfs.md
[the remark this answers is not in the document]
agnivade: So we should mention the add examples first. Otherwise, the user does not know from where to get the file hash. We always arrange the commands in the order of usage, if a command needs the output of another command we need to mention that first.

pages/common/ipfs.md
```diff
@@ -0,0 +1,32 @@
+# ipfs
+
+> Inter Planetary File System.
+> A peer-to-peer hypermedia protocol to make the web faster, safer, and more open.
+
+- Save a remote file and give It a name but not pin It:
+
+`ipfs get {{hash}} -o {{filename}}`
+
+- Add a file from local to the file system with the filename (-w) and pin It:
```
agnivade: Is there a need to mention "-w" ? If you think so, we also need to explain the flag, which is "wrap with directory".

We should also mention that this command prints the file hash which is necessary in the first command.

pages/common/ipfs.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Add a folder and files in It from local to the File System and pin It:
+
+`ipfs add -r {{folder}}`
+
+- Pin a remote file, locally:
+
+`ipfs pin add {{hash}}`
+
+- Check pinned files:
```
agnivade: Maybe "Display" is a better word than "Check". It just displays the pinned files. We are not checking for anything here.

pages/common/ipfs.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Add a file from local to the file system with the filename (-w) and pin It:
+
+`ipfs add {{filename}} -w`
+
+- Add a folder and files in It from local to the File System and pin It:
+
+`ipfs add -r {{folder}}`
+
+- Pin a remote file, locally:
```
agnivade: No need for the comma.

pages/common/ipfs.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Save a remote file and give It a name but not pin It:
+
+`ipfs get {{hash}} -o {{filename}}`
+
+- Add a file from local to the file system with the filename (-w) and pin It:
+
+`ipfs add {{filename}} -w`
+
+- Add a folder and files in It from local to the File System and pin It:
```
agnivade: We should explain the "-r" flag. Perhaps - "Add a folder and its files recursively from ..."

pages/common/ipfs.md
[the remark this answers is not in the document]
FraYoshi: Perfect, going to modify with this style =)

pages/common/ipfs.md
FraYoshi: To simplify is maybe better avoiding the -w flag.. actually It's not necessarily needed.. just a fashion :)

pages/common/ipfs.md
```diff
@@ -0,0 +1,32 @@
+# ipfs
+
+> Inter Planetary File System.
+> A peer-to-peer hypermedia protocol to make the web faster, safer, and more open.
+
+- Add a file from local to the file system and pin it:
```
agnivade: Mention that it also prints out the file hash.

pages/common/ipfs.md
```diff
@@ -0,0 +1,32 @@
...
+
+`ipfs add {{filename}}`
+
+- Add a folder and its files recursively from local to the file system:
```
agnivade: Same here

PR #2102, opened 2018-05-07
man: add example for choosing page section

fetlang:
I feel like this is a useful enough option to include in the tldr page

### DIFF
--- a/pages/common/man.md
+++ b/pages/common/man.md
@@ -6,6 +6,10 @@
 
 `man {{command}}`
 
+- Display man page for a command from section 7:
+
+`man {{command}}.7`
+
 - Display path searched for manpages:
 
 `man --path`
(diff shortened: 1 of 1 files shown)


pages/common/man.md
```diff
@@ -6,6 +6,10 @@
 
 `man {{command}}`
 
+- Display man page for a command from section 7:
+
+`man {{command}}.7`
```
sbrl: Perhaps we could put `7` in `{{` and `}}`?

fetlang: oh yeh that makes sense

agnivade: In several examples, I see `man command <section>` being used. Is that incorrect ?

fetlang: On Linux, `man command <section>` will pull up two man pages: one for command, and one for section(which will undoubtedly not exist).

agnivade: Ah thanks !

agnivade:
Moved . out of tokens. Thanks.

PR #2104, opened 2018-05-09
jrnl: add page

OrangeDugongo:
----


pages/common/jrnl.md
```diff
@@ -0,0 +1,23 @@
+# jrnl
+
+> A simple journal application for your command line.
+
+- Insert new entry with your editor:
```
sbrl: Maybe `Insert a new entry`... would read better?

pages/common/jrnl.md
```diff
@@ -0,0 +1,23 @@
...
+
+`jrnl`
+
+- Quickly insert new entry:
```
sbrl: Maybe ...`insert a new entry:` would read better?

pages/common/jrnl.md
```diff
@@ -0,0 +1,23 @@
...
+
+`jrnl {{timestamp}}: {{title}}. {{content}}`
+
+- View last ten entries:
```
sbrl: Maybe `View the last ten entries:` would read better?/

pages/common/jrnl.md
```diff
@@ -0,0 +1,23 @@
...
+
+`jrnl -n {{10}}`
+
+- View everything that happened from the start of last year to the start of last march:
+
+`jrnl -from {{"last year"}} -until {{march}}`
+
+- Edit all entries tagged with @texas and @history:
+
+`jrnl {{@texas}} -and {{@history}} --edit`
```
agnivade: I guess the tags are "texas", "history" and not "@texas", "@history". In that case, please move the @ out of the token. And also, we can remove the @ from the descriptions.

pages/common/jrnl.md
```diff
@@ -0,0 +1,23 @@
...
+
+> A simple journal application for your command line.
+
+- Insert a new entry with your editor:
+
+`jrnl`
+
+- Quickly insert a new entry:
+
+`jrnl {{timestamp}}: {{title}}. {{content}}`
```
agnivade: "timestamp" is not immediately apparent as to the format of the timestamp to be used. I think it's better we use a concrete example for this.

pages/common/jrnl.md
OrangeDugongo: Yes, you're right but `-and` accepts other filters and not only tags. I guess to remove the @ from the description but i left @ inside the token

agnivade: Cool, let's do that.

PR #2105, opened 2018-05-09
trash-cli: add page

cfsanderson:


agnivade:
These seem to have different sub-commands. Ideally, we should have separate pages for each of the commands. And just link them from the main page.

cfsanderson:
🤔 not sure what you mean @agnivade. `trash-cli` is the name of the software but isn't a command per se.
https://github.com/andreafrancia/trash-cli

agnivade:
I thought this was something like git. `git` is the main command, but we have separate pages for `git-push`, `git-commit` and so on. Do each of the commands like `trash-put` and `trash-empty` have flags and other details which need to be mentioned ?

pages/common/trash-cli.md
```diff
@@ -0,0 +1,23 @@
+# trash-cli
+
+> Trash-cli is a less permanent way of removing files via the command line.
```
agnivade: "less permanent" seems a bit unclear. Perhaps - "A command line interface to the trashcan APIs" sounds better.

pages/common/trash-cli.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Trash files and directories:
+
+`trash-put {{file_name}}`
+
+- Empty the trashcan(s):
```
agnivade: So does this mean there can be multiple trashcans ? How do we control which trash can to dump a file to ?

sbrl:
Maybe we should have a page for each command then? I'm not sure.

agnivade:
Ideally we should, but then each page will have just one example. So maybe the ease of use outweighs the need for consistency ?

Also, another question is will the user search for something like `tldr trash-put`, `tldr trash-rm` ? Then certainly we need separate pages because if they are all under `trash-cli`, then it won't return any result. But if they just search for `trash-cli`, then I think it should be fine to have it in a single page.

ping @waldyrious - thoughts ?

pages/common/trash-cli.md
cfsanderson: Sounds good.

pages/common/trash-cli.md
cfsanderson: No, I'll get rid of that

PR #2110, opened 2018-05-12
tldr as container image

nutellinoit:
Added the ability to use `tldr` by running a container. This is my first pull request i hope i'm not going against contributions guide!

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,9 @@ You can access these pages on your computer using one of the following clients:
 - Web clients:
     - [tldr.jsx](https://github.com/ostera/tldr.jsx): http://tldr.ostera.io/
     - [DistroWatch](https://distrowatch.com/dwres.php?resource=man-pages)
+- Docker container:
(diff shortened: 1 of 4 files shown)


sbrl:
So this runs a tldr client in a docker container?

agnivade:
Yes.

pxgamer:
Is this ok to close as it hasn't had any activity for 25 days. @agnivade, @sbrl does this seem alright?

@nutellinoit, I can see you've made a repo for this but it's empty at the moment. I'd be happy to do some testing once it's set up. :smile: :confetti:

agnivade:
Yep.

nutellinoit:
After a lot of time I managed to create the project. The link is https://github.com/nutellinoit/tldr-docker

I set jenkins so that every hour goes to check the repository and download the submodules, then will trigger the new docker build and push on docker hub

😊

agnivade:
Sure, I don't see a reason not to. We should create a separate section "Docker images".

PR #2111, opened 2018-05-14
xargs: Show delimiter flag usage

okaymaged:

### DIFF
--- a/pages/common/xargs.md
+++ b/pages/common/xargs.md
@@ -7,9 +7,9 @@
 
 `{{arguments_source}} | xargs {{command}}`
 
-- Delete all files with a `.backup` extension:
+- Delete all files with a `.backup` extension. The `-d` flag changes the delimiter to allow filenames with whitespace:
 
-`find . -name {{'*.backup'}} | xargs rm -v`
+`find . -name {{'*.backup'}} | xargs -d '\n' rm -v`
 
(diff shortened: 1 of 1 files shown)


pages/common/xargs.md
```diff
@@ -9,7 +9,7 @@
 
 - Delete all files with a `.backup` extension:
```
agnivade: What difference is the `--delimiter` option making here ? The output from `find` is already delimited by newlines.

okaymaged: It's changing the delimeter on `xargs` not `find`. By default `xargs` uses any whitespace as its delimiter.

okaymaged:
> I meant to add the part about newline in the command description itself, not another sub-heading. We only have 2 lines for the heading.

Done. 

> And also, I see that you have removed the example with -0. Any reason for that ?

I missed it initially. It doesn't make sense to transform '\n' to null characters then use null characters as the delimiters rather than just using '\n' as the delimiter (which the `find` example already shows).

Alternatively, we can use: 
```find . -name {{'*.backup'}} -print0 | xargs -0 rm -v```
This is safest in the rare case that new line characters are part of the filename (and shows usage of `-0`)

agnivade:
> This is safest in the rare case that new line characters are part of the filename (and shows usage of -0)

Great, let's do that.

okaymaged:
Done

agnivade:
@sbrl - Are you okay with the changes ?

PR #2113, opened 2018-05-16
mysqldump: make correct dump of UTF-8 databases

SpartakusMd:
Make sure to use the `-r <filename>` option, instead of redirecting command-line output to a file `> <filename>`. Redirection does not capture UTF characters properly, while the -r option does.

As well the import part was changed to fix the same issue.

----

- [ x] The page has 8 or fewer examples. (this is an edit)

- [ x] The PR is appropriately titled:  

- [ x] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines.

### DIFF
--- a/pages/common/mysql.md
+++ b/pages/common/mysql.md
@@ -18,6 +18,6 @@
 
 `mysql --socket {{path/to/socket.sock}}`
 
-- Execute SQL statements in a script file (batch file):
+- Execute SQL statements in a script file (batch file), user will be prompted for a password:
 
-`mysql -h {{database_host}} -u {{user}} --password {{database_name}} -e "source {{/path/to/backup-file.sql]}" {{database_name}}`
+`mysql -u {{user}} --password -e "source {{filename.sql}}" {{database_name}}`
--- a/pages/common/mysql.md
+++ b/pages/common/mysql.md
@@ -20,4 +20,4 @@
 
 - Execute SQL statements in a script file (batch file), user will be prompted for a password:
 
-`mysql -u {{user}} --password -e "source {{filename.sql}}" {{database_name}}`
+`mysql -e "source {{filename.sql}}" {{database_name}}`
--- a/pages/common/mysql.md
+++ b/pages/common/mysql.md
@@ -18,6 +18,6 @@
 
 `mysql --socket {{path/to/socket.sock}}`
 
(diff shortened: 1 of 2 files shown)


pages/common/mysql.md
```diff
@@ -20,4 +20,4 @@
 
 - Execute SQL statements in a script file (batch file):
 
-`mysql {{database_name}} < {{script.sql}}`
+`mysql -h {{database_host}} -u {{user}} --password {{database_name}} -e "source {{/path/to/backup-file.sql]}" {{database_name}}`
```
agnivade: Perhaps you meant to remove the {{database_name}} after password ? Since I see you have added it to the end.

Also remove the ] after `.sql`. And the / before path.

pages/common/mysqldump.md
```diff
@@ -4,8 +4,8 @@
 
 - Create a backup, user will be prompted for a password:
 
-`mysqldump -u {{user}} --password {{database_name}} > {{filename.sql}}`
+`mysqldump -u {{user}} --password {{database_name}} -r {{filename.sql}}`
 
 - Restore a backup, user will be prompted for a password:
 
-`mysql -u {{user}} --password {{database_name}} < {{filename.sql}}`
+`mysql -u {{user}} --password {{database_name}} -e "source {{/path/to/backup-file.sql]}" {{database_name}}`
```
agnivade: Same as before.

pages/common/mysql.md
```diff
@@ -18,6 +18,6 @@
 
 `mysql --socket {{path/to/socket.sock}}`
 
-- Execute SQL statements in a script file (batch file):
+- Execute SQL statements in a script file (batch file), user will be prompted for a password:
 
-`mysql {{database_name}} < {{script.sql}}`
+`mysql -u {{user}} --password -e "source {{filename.sql}}" {{database_name}}`
```
agnivade: Sorry, please let's not add the user and password options back again. Those weren't there for a reason.

The tldr project is more of an explainer than a cheatsheet -- in other words, the underlying mission is not to merely provide copy-pastable snippets (like, say, commandlinefu.com) but rather to allow readers to understand the language of the commands, so that they can then derive new commands from the provided examples, or compose several simple examples into more complex ones that are useful to them.

PR #2114, opened 2018-05-17
git-push: Add entry for pushing to a different branch name

hdra:


agnivade:
The example `git push {{remote_name}} -u {{remote_branch}}` does the same. And I think that is a more common example than mentioning the full local branch and remote branch.

PR #2115, opened 2018-05-17
msbuild: add page

pxgamer:
----

### DIFF
--- a/pages/common/msbuild.md
+++ b/pages/common/msbuild.md
@@ -1,6 +1,6 @@
 # msbuild
 
-> The Microsoft build tool for .NET projects.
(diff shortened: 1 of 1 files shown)


pages/common/msbuild.md
```diff
@@ -0,0 +1,27 @@
+# msbuild
+
+> The Microsoft build tool for .NET projects.
+
+- Build a specific project file:
```
sbrl: By default, if presented with no arguments, `msbuild ` will automagically find a project or solution file in the current directory and build that. Perhaps we could include that here too?

Also, `msbuild` will build Visual Studio solutions. Perhaps we might want to update the wording in the description at least to reflect this?

PR #2125, opened 2018-05-24
lz4: add page

janwh:
Hi! As lz4 is IMHO a very nice alternative to gzip or xz compression, it should have a tldr page as well. I hope my suggested examples are good enough even for novice users.  Cheers! 🤙

----

### DIFF
--- a/pages/common/lz4.md
+++ b/pages/common/lz4.md
@@ -1,0 +1,27 @@
+# lz4
+
+> Compress or decompress .lz4 files.
+
+- Compress a file:
+
+`lz4 {{file}}`
+
+- Decompress a file:
+
+`lz4 -d {{file.lz4}}`
+
+- Decompress a file and write to stdout:
+
+`lz4 -dc {{file.lz4}}`
+
+- Package and compress a directory and its contents:
+
+`tar cvf - {{path/to/dir}} | lz4 - {{dir.tar.lz4}}`
+
+- Decompress and unpack a directory and its contents:
+
+`lz4 -d {{dir.tar.lz4}} | tar -xv`
+
+- Compress a file using the best compression:
+
+`lz4 -9 {{file}}`


pages/common/lz4.md
```diff
@@ -0,0 +1,27 @@
...
+
+`lz4 -d {{file.lz4}}`
+
+- Decompress a file and write to stdout:
+
+`lz4 -dc {{file.lz4}}`
+
+- Package and compress a directory and its contents:
+
+`tar cvf - {{/path/to/dir}} | lz4 - {{dir.tar.lz4}}`
```
pxgamer: Could you change the absolute path to `{{path/to/dir}}`?

janwh: Done!

pages/common/lz4.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Package and compress a directory and its contents:
+
+`tar cvf - {{path/to/dir}} | lz4 - {{dir.tar.lz4}}`
+
+- Decompress and unpack a directory and its contents:
+
+`lz4 -d {{dir.tar.lz4}} | tar -xv`
+
+- Compress a file using the best compression:
```
agnivade: This should have been the second example.

sbrl: Ah, oops! That might have been a good idea.

Most of the time, the best compression mode doesn't actually compress it much more than the standard compression does, whilst taking much longer. So to that end it's not actually that useful in day-to-day circumstances.

janwh: I thought so, too at first. I decided against it as for me LZ4's major benefit is its default mode (`-1`): the space savings on compressable files are nicely balanced with the speed of processing on incompressable files. The improved compression of using `-9` are diminishing the benefits of processing speed, and as tldr is making an effort to appeal to novice users, I thought I'd intentionally discourage the use of `-9` by putting it at the end.

agnivade: Ok, fair enough.

PR #2133, opened 2018-06-03
rbenv: add page

thezeroalpha:
----

### DIFF
--- a/pages/common/rbenv.md
+++ b/pages/common/rbenv.md
@@ -1,0 +1,31 @@
+# rbenv
+
+> A tool to easily install Ruby versions and manage application environments.
+
+- Install one or more space-separated versions of Ruby:
+
+`rbenv install {{version(s)}}`
+
+- Display a list of installed versions:
+
+`rbenv versions`
+
+- Use a specific version of Ruby across the whole system:
+
+`rbenv global {{version}}`
+
+- Use a specific version of Ruby for an application/project folder:
+
+`rbenv local {{version}}`
+
+- Show the currently selected Ruby version:
+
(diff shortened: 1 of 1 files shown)


pages/common/rbenv.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Show the currently selected Ruby version:
+
+`rbenv version`
+
+- Uninstall a version of Ruby:
+
+`rbenv uninstall {{version}}`
+
+- List Ruby versions that contain an executable program:
```
agnivade: What does "contain an executable program" mean ? I guess you mean to say what is the version of a ruby binary ? Won't that be shown easily by `ruby -v` ? In that case, I don't think this example is much beneficial to be in a tldr page.

pxgamer: It displays all ruby versions that contain the specified executable. Say you have a Gem on multiple versions of Ruby, then it will display these. For example, `rbenv whence travis` would display which versions of Ruby one has Travis installed for. Afaik.

thezeroalpha: Yes, @pxgamer is right, that's exactly what it does. Should I reword it to make that clearer?

thezeroalpha: Well, @pxgamer's description also includes the word "contain", but I agree that it's better. I'll change it.

agnivade: I meant this - 
> For example, rbenv whence travis would display which versions of Ruby one has Travis installed for. Afaik.

agnivade:
Going to merge this one, and send a new PR with something that I feel might be clearer.

PR #2135, opened 2018-06-06
conky: add page

f4ww4z:
----

### DIFF
--- a/pages/linux/conky.md
+++ b/pages/linux/conky.md
@@ -10,18 +10,18 @@
 
 `conky -C > ~/.conkyrc`
 
-- Load a config file:
+- Launch conky with a given config file:
 
-`conky -c {{/path/to/config}}`
+`conky -c {{path/to/config}}`
 
 - Start in the background (daemonize):
 
 `conky -d`
 
-- Align specified config file on screen:
+- Align conky on the desktop:
 
-`conky -c {{path/to/config}} -a {{{top,bottom,middle}_{left,right,middle}}}`
+`conky -a {{{top,bottom,middle}_{left,right,middle}}}`
 
-- Wait 5 seconds and launch the conky config in a specified path:
+- Pause for 5 seconds at startup before launching:
 
-`conky -p 5 -c {{path/to/config}}`
+`conky -p {{5}}`
--- a/pages/linux/conky.md
+++ b/pages/linux/conky.md
@@ -1,0 +1,27 @@
+# conky
+
+> Light-weight system monitor for X.
+
+- Launch with default, built-in config:
+
+`conky`
+
+- Create a new, default config:
+
+`conky -C > ~/.conkyrc`
+
+- Load a config file:
+
+`conky -c {{/path/to/config}}`
+
+- Start in the background (daemonize):
+
+`conky -d`
+
+- Align specified config file on screen:
+
+`conky -c {{path/to/config}} -a {{{top,bottom,middle}_{left,right,middle}}}`
+
+- Wait 5 seconds and launch the conky config in a specified path:
+
+`conky -p 5 -c {{path/to/config}}`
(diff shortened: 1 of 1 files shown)


pages/linux/conky.md
```diff
@@ -0,0 +1,27 @@
...
+
+`conky`
+
+- Create a new, default config:
+
+`conky -C > ~/.conkyrc`
+
+- Load a config file:
+
+`conky -c {{/path/to/config}}`
```
pxgamer: Could you remove the prefixing `/` here?

pages/linux/conky.md
```diff
@@ -0,0 +1,27 @@
...
+
+`conky -d`
+
+- Align specified config file on screen:
+
+`conky -c {{path/to/config}} -a {{{top,bottom,middle}_{left,right,middle}}}`
+
+- Wait 5 seconds and launch the conky config in a specified path:
+
+`conky -p 5 -c {{path/to/config}}`
```
pxgamer: Put the 5 in tokens. :+1:

pages/linux/conky.md
```diff
@@ -0,0 +1,27 @@
+# conky
+
+> Light-weight system monitor for X.
+
+- Launch with default, built-in config:
+
+`conky`
+
+- Create a new, default config:
```
pxgamer: I don't think the comma is necessary here. :thinking:

pages/linux/conky.md
```diff
@@ -0,0 +1,27 @@
...
+
+`conky -c {{/path/to/config}}`
+
+- Start in the background (daemonize):
+
+`conky -d`
+
+- Align specified config file on screen:
+
+`conky -c {{path/to/config}} -a {{{top,bottom,middle}_{left,right,middle}}}`
```
pxgamer: Is the final format like `{1,2,3_4,5,6}`? If Not, I think it's fine to only have tokens around the 2 parts maybe. Rather than around the whole thing.

pages/linux/conky.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Launch with default, built-in config:
+
+`conky`
+
+- Create a new, default config:
+
+`conky -C > ~/.conkyrc`
+
+- Load a config file:
```
agnivade: "Launch conky with a given config file" matches with the first description.

pages/linux/conky.md
agnivade: What does `-p` stand for ? Can we use a word in the description that matches with that ?. And no need to use the `-c` option again. It is already specified above. And users are expected to combine options.

pages/linux/conky.md
```diff
@@ -0,0 +1,27 @@
...
+
+`conky -c {{/path/to/config}}`
+
+- Start in the background (daemonize):
+
+`conky -d`
+
+- Align specified config file on screen:
```
agnivade: This is a bit unclear to me. What does it mean to align a config file on screen ?

pages/linux/conky.md
f4ww4z: Yes, the final format is `{top,bottom,middle}_{left,right,middle}` with an underline (a token).

pages/linux/conky.md
f4ww4z: the `-p` option is to pause for {SECS} seconds at startup before launching. I will change it to that I think.

pages/linux/conky.md
f4ww4z: It's to align the display on the desktop. So for example, `conky -a top_right` will place the system monitor on the top-right corner of the desktop.

pages/linux/conky.md
pxgamer: Oh, then this is fine. 👍

pages/linux/conky.md
f4ww4z: Yeah I agree, it's a [cumulative adjective](https://www.quickanddirtytips.com/education/grammar/commas-with-adjectives) :smile:

PR #2136, opened 2018-06-07
complete: add page

emmabukacek:
----

### DIFF
--- a/pages/common/complete.md
+++ b/pages/common/complete.md
@@ -1,0 +1,15 @@
+# complete
+
+> Provides argument autocompletion to shell commands.
+
+- Apply an autocomplete function to a command:
+
+`complete -F {{function}} {{command}}`
+
+- Apply an autocomplete command to a command:
+
+`complete -C {{autocomplete_command}} {{command}}`
+
+- Apply autocompletion without appending a space to the completed word:
+
+`complete -o nospace -F {{function}} {{command}}`
--- a/pages/common/complete.md
+++ b/pages/common/complete.md
@@ -2,11 +2,11 @@
 
 > Provides argument autocompletion to shell commands.
 
-- Apply an autocomplete function to a command:
+- Apply a function, that performs autocompletion, to a command:
 
 `complete -F {{function}} {{command}}`
 
-- Apply an autocomplete command to a command:
+- Apply a command, that performs autocompletion, to another command:
 
 `complete -C {{autocomplete_command}} {{command}}`
 
--- a/pages/common/complete.md
+++ b/pages/common/complete.md
@@ -2,11 +2,11 @@
 
 > Provides argument autocompletion to shell commands.
 
-- Apply a function, that performs autocompletion, to a command:
+- Apply a function that performs autocompletion to a command:
 
 `complete -F {{function}} {{command}}`
 
-- Apply a command, that performs autocompletion, to another command:
+- Apply a command that performs autocompletion to another command:
 
 `complete -C {{autocomplete_command}} {{command}}`
 


pages/common/complete.md
```diff
@@ -0,0 +1,15 @@
+# complete
+
+> Provides argument autocompletion to shell commands.
+
+- Apply an autocomplete function to a command:
+
+`complete -F {{function}} {{command}}`
+
+- Apply an autocomplete command to a command:
```
emmabukacek: Admittedly this phrasing isn't great, so I would love to hear alternatives.

pages/common/complete.md
```diff
@@ -0,0 +1,15 @@
...
+
+`complete -C {{autocomplete_command}} {{command}}`
+
+- Apply autocompletion without appending a space to the completed word:
+
+`complete -o nospace -F {{function}} {{command}}`
```
emmabukacek: Wasn't sure how many examples I should include. I've only used `complete` at the basic level, so I'd love to hear any thoughts on any additional examples we should or shouldn't include here.

pages/common/complete.md
agnivade: "autocomplete sub-command" ?

pages/common/complete.md
agnivade: This is fine.

pages/common/complete.md
agnivade: @sbrl , @pxgamer - comments ?

pxgamer: I was just trying to actually work out what this does as I had no idea. Done a bit of testing and I understand now, but I have no idea for how it should be worded.

emmabukacek: I've been thinking about this a bit.. Some thoughts:

The core issue here is the fact that two identical nouns located together may create confusion. That being said, either command can hypothetically be an absolute path, on the `$PATH` variable, or any other way we access commands in the shell. Since they are indeed the same thing, albeit one with autocompletion functionality, it may not be erroneous to use the same word. So instead, perhaps we can add an adjective to the latter command or update the first adjective. Perhaps something like this:

```markdown
- Apply a command that performs autocompletion to another command:
```

```markdown
- Apply an autocomplete command to a different command
```

Thoughts?

sbrl: What about `Apply an autocomplete provider (function/command) to a specified command name`? (replace in brackets as appropriate)

pages/common/complete.md
sbrl: Up to 8 is fine!

pages/common/complete.md
emmabukacek: This is a step in the right direction, but I'm still worried `provider` obfuscates the meaning. In addition, having `function/command` might be confusing considering the two options, `-F` and `-C`, specifically delineate between function and command. :grimacing: I promise I'm not trying to be difficult! Hehe. Just hoping we can find the clearest explanation here.

agnivade: > - Apply a command that performs autocompletion to another command:

I like this. Although slightly longer, it is perfectly clear what's happening.

emmabukacek:
I went ahead and changed both examples to elaborate on the autocomplete function / command for the sake of parity, but happy to revert the first example. Thanks for the quick response, y'all! :smile:

pages/common/complete.md
```diff
@@ -0,0 +1,15 @@
+# complete
+
+> Provides argument autocompletion to shell commands.
+
+- Apply a function, that performs autocompletion, to a command:
```
agnivade: Don't think the commas after `function` and `command` are needed.

emmabukacek: Not a problem. I'll drop those.

agnivade:
Weird, CLA causing issues again. @emmabukacek - Have you signed the CLA ?

emmabukacek:
@agnivade I did! Not sure why it's not registering though.. I went ahead and signed it again. Perhaps that will fix it?

agnivade:
The CLA assistant has caused issues in the past and seems to be acting up again. I will force merge this.

PR #2153, opened 2018-06-24
windows/mkdir: add page

sbrl:
For #1579

----

### DIFF
--- a/pages/windows/mkdir.md
+++ b/pages/windows/mkdir.md
@@ -6,6 +6,6 @@
 
 `mkdir {{directory_name}}`
 
-- Recursively create a nested direcotry tree:
+- Recursively create a nested directory tree:
 
 `mkdir {{path/to/sub_directory_name}}`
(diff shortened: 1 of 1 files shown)


pages/windows/mkdir.md
```diff
@@ -0,0 +1,11 @@
+# mkdir
+
+> Creates a directory.
+
+- Create a directory:
+
+`mkdir {{directory_name}}`
+
+- Recursively create a nested direcotry tree:
```
pxgamer: Typo in directory. :+1:

pages/windows/mkdir.md
```diff
@@ -0,0 +1,11 @@
...
+
+`mkdir {{path/to/sub_directory_name}}`
```
agnivade: There seems to be no difference between the first and second command. Is this correct ?

pages/windows/mkdir.md
sbrl: Good spot, thanks!

pages/windows/mkdir.md
sbrl: Yep. I wanted to show explicitly that the windows variant of `mkdir` recursively creates directories by default, whereas the Linux one doesn't

PR #2155, opened 2018-06-26
finger: add page

pxgamer:
----

### DIFF
--- a/pages/windows/finger.md
+++ b/pages/windows/finger.md
@@ -1,0 +1,20 @@
+# finger
+
+> Return information about one or more users on a specified system.
+> The remote system must be running the Finger service.
+
+- Display information about a specific user:
+
+`finger {{user}}@{{host}}`
+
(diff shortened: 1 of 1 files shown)


pages/windows/finger.md
```diff
@@ -0,0 +1,20 @@
...
+
+`finger {{user}}@{{host}}`
+
+- Display information about all users on the specified host:
+
+`finger @{{host}}`
+
+- Display information in a longer format:
+
+`finger {{user}}@{{host}} -l`
```
sbrl: You use a `-l` here, but a `/?` further down. Typo?

pxgamer: Not a typo, Microsoft is just inconsistent. You can use `--help` (instead of `/?`) but I wasn't sure if that's available in older versions of the command.

sbrl: Ah, I see! Good ol' microsoft :P

PR #2159, opened 2018-06-28
xar: add page

johan456789:
----

### DIFF
--- a/pages/linux/xar.md
+++ b/pages/linux/xar.md
@@ -4,12 +4,12 @@
 
 - Create a xar archive of all files in {{directory}}:
 
-`xar -cf sample.xar {{directory}}`
+`xar -cf {{archive.xar}} {{path/to/directory}}`
 
 - List the contents of the xar archive sample.xar:
 
-`xar -tf sample.xar`
+`xar -tf {{archive.xar}}`
 
 - Extract the contents of sample.xar to the current working directory:
 
-`xar -xf sample.xar`
+`xar -xf {{archive.xar}}`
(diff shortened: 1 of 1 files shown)


pages/linux/xar.md
```diff
@@ -0,0 +1,15 @@
+# xar
+
+> Manage .xar archives.
+
+- Create a xar archive of all files in {{directory}}:
+
+`xar -cf sample.xar {{directory}}`
```
sbrl: Perhaps this would be better as `{{path/to/directory}}` with a description something like ....`in the specified directory:`? Then it would be more consistent with the existing pages we've got already.

pages/linux/xar.md
```diff
@@ -0,0 +1,15 @@
...
+
+- List the contents of the xar archive sample.xar:
+
+`xar -tf sample.xar`
+
```
sbrl: Could `sample.xar` be `{{path/to/filename.xar}}` please?

pages/linux/xar.md
```diff
@@ -0,0 +1,15 @@
...
+- Extract the contents of sample.xar to the current working directory:
+
+`xar -xf sample.xar`
```
sbrl: Same here.

Wesalius:
if the problem is only to tokenize, than I would do it, but I dont have the permissions

agnivade:
That is fine. We usually ask the contributors first if they have any other input before going ahead to make the change (unless it is a minor thing and super obvious). In this case, since @johan456789 has not replied, I will do the changes myself.

PR #2161, opened 2018-07-03
ls: add `--quoting-style=shell-escape`

HaleTom:
Most of the quoting styles put `?` instead of non-printable characters. This one works.

----

### DIFF
--- a/pages/common/ls.md
+++ b/pages/common/ls.md
@@ -25,3 +25,7 @@
 - Long format list of all files, sorted by modification date (oldest first):
 
 `ls -ltr`
+
+- Quote strange filenames (delete using `rm -- NAME`):
+
+`ls --quoting-style=shell-escape`


pages/common/ls.md
```diff
@@ -25,3 +25,7 @@
 - Long format list of all files, sorted by modification date (oldest first):
 
 `ls -ltr`
+
+- Quote strange filenames (delete using `rm -- NAME`):
```
agnivade: Can we define "strange" ? Perhaps non-printable characters is better. Also, is adding the delete statement necessary ?

pxgamer: I think that `non-printable` would be better, if this is what they are.

bumper314:
Is this really "common"?  `ls` on Mac doesn't take the --quoting-style option and "shell-escape" isn't a valid option on Ubuntu 14.04 LTS:

```
ls: invalid argument ‘shell-escape’ for ‘--quoting-style’
Valid arguments are:
  - ‘literal’
  - ‘shell’
  - ‘shell-always’
  - ‘c’
  - ‘c-maybe’
  - ‘escape’
  - ‘locale’
  - ‘clocale’
```

HaleTom:
@bumper314 what does Mac support in terms of `--quoting-style=`?

bumper314:
@HaleTom Sorry for the delay.  Mac does not support `--quoting-style` at all.

```
$ ls --quoting-style=shell-escape
ls: illegal option -- -
usage: ls [-ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1] [file ...]
```

Nothing related to "quote" in the man page on Mac either, so there doesn't appear to be a short flag equivalent.

agnivade:
Since there is no mac equivalent for this option, I think it is a bit out of place in a tldr page. I think it is simpler to have a common page for ls, rather than split the page into linux and macos and have this option there.

Closing this. Please comment if anyone disagrees.

PR #2163, opened 2018-07-04
mycli: move from linux to common; add uri example;

gonejack:
Signed-off-by: gonejack <igonejack@gmail.com>

----

About moving: mycli is python based which is also available for macOS.

### DIFF
--- a/pages/common/mycli.md
+++ b/pages/common/mycli.md
@@ -13,3 +13,11 @@
 - Connect to a database on the specified host with the specified user:
 
 `mycli -u {{user}} -h {{host}} {{database_name}}`
+
+- Connect to a database with URI:
+
+`mycli mysql://{{user}}@{{host}}`
+
+- Connect to a database with URI more directly:
+
+`mycli mysql://{{user}}:{{password}}@{{host}}:{{port}}/{{database_name}}`


pages/common/mycli.md
```diff
@@ -13,3 +13,11 @@
 - Connect to a database on the specified host with the specified user:
 
 `mycli -u {{user}} -h {{host}} {{database_name}}`
+
+- Connect to a database with URI:
+
+`mycli mysql://{{user}}@{{host}}`
```
agnivade: We should stick with only one type of usage. The above examples show `-u` and `-h`. This seems to be just a short hand representation of it.

pages/common/mycli.md
```diff
@@ -13,3 +13,11 @@
...
+
+- Connect to a database with URI more directly:
+
+`mycli mysql://{{user}}:{{password}}@{{host}}:{{port}}/{{database_name}}`
```
agnivade: Same as before. Also, passwords should not be passed in the command line. The command will ask for password and the user should enter. Please see our pages for `psql` and `mysql`. We should try to keep this page consistent with others.

pages/common/mycli.md
gonejack: It is hard to say which one is better, But people like me are often given the latter one because URI is easy to share and keep.

pages/common/mycli.md
gonejack: As before, If the URI is like saved in project development config file and password is required, Then the password don't have to sit in separate line.

agnivade: Ideally, in settings like that you can pass the password in an environment variable or through something like `.pgpass`. You should almost never set password plainly like this.

pages/common/mycli.md
agnivade: Sure, I am not saying this is better or worse. I am just concerned about consistency. And since other pages like mysql and psql follow the existing format, I raised a concern regarding the benefits of changing such a convention.

agnivade:
However, the move from linux to common was totally fine. Would you like to reopen this PR and update your commit so that only the file move is there ?

gonejack:
It's too trivial to have a PR, just have it in your next update.

sbrl:
We've had PRs for less! Generally, we end up doing everything through PRs anyway :P

Issue #1966, opened 2018-02-01
Big reorganisation needed


gingerbeardman:
For those that aren't aware [the tldr project was originally created for just macOS X commands](https://github.com/tldr-pages/tldr/tree/93444b4d839df511da00a33657d5ad3e57dc7179). since then other platforms have been added: linux, sunOS and most recently Windows. as well as the "common" folder. 

IMHO the introduction of `common` was a mistake and now we are all suffering because of it. whilst command names may be common even between BSD and Linux the arguments or parameters accepted by identically named commands are different. 

ideals:
- common would only contain commands that are truly common and interoperable across platforms (a big ask!). there may not be many with if that criteria is upheld to the letter, especially now that Windows has been added.

- clients should look into the folder of the OS they are running on first, and then refer to common only if the OS folder does not contain the command. for me this is just the obvious way to do it, but it is not currently in the client spec

- during the migration away from common, it could contain stubs that inform users of the change, information of the pending deprecation

those are my initial thoughts on the matter.

NargiT:
+1

waldyrious:
What are your thoughts on #190?

gingerbeardman:
@waldyrious my thoughts are that a flat structure is a better approach. crazy that that issue is 4 years old! 

Another thought I had would be for "us" to implement updates to each client to ensure all clients are onboard with any new structure from day one.

waldyrious:
> @waldyrious my thoughts are that a flat structure is a better approach.

I was asking more in the lines of whether you see any overlap in that issue and this one.

(As for the clients, yes, that's always been our policy: we tend to make sure client authors are at least aware of upcoming changes, and give them ample time to adapt if they need so.)

gingerbeardman:
Yes 190 is just what I think is needed. 

With regards to clients, what I meant is that we make the code changes ans submit PRs in those projects in the clients we use as part of of this endeavor. So the client authors simply have to merge and release. 

For example, I use the MacOS shell client, and the Dash docset client. I would be happy to make the changes needed for those clients to meet any new tldr organisation. I think we need to roll up our sleeves here and get stuck in.

notpeter:
Hi folks, 

[...]

Issue #1065, opened 2016-09-16
Define spec for officially sanctioned clients


waldyrious:
Background:
- https://github.com/tldr-pages/tldr/pull/744
- https://github.com/tldr-pages/tldr/wiki/Minimum-specifications-for-TLDR-command-line-clients

agnivade:
We should close #1010 and #1000 and make it part of this issue.

Mentioning a comment made by @rubenvereecken for reference -
> The client specificity is exactly the reason why I think a completion script should not be bundled with the main repo. People will miss out on client-specific completions because they'll never install those.

> We did want to make the listing behavior standard. After all, it's useful for all clients. Never got around to it I think.

The main idea was to enforce a `-a|--list-all` flag on all clients, by which clients can further implement bash|zsh command completion.

waldyrious:
Are you saying that every sanctioned client **must** implement auto-completion? If not (which is my position), I'd say we should connect those issues to this one, but not close them. That way we can track implementation of that feature without blocking this one. A `--list-all` feature can certainly be part of the minimal spec, though (IMO).

agnivade:
> Are you saying that every sanctioned client must implement auto-completion?

Not at all. But I was saying **if** they are to be implemented at all, it should be implemented by clients(because a proper auto-completion should also complete the helper flags and not just commands, and the parent repo has no way to know all the helper flags of all the clients). That's what we concluded at the end of the PR. Therefore, I want a `--list-all` option to be part of the minimum spec, which will allow a standard way for all clients to implement auto-completion.

waldyrious:
Ok, in that case those issues should remain open until (1) we define what a sanctioned client is --i.e. this issue--, (2) we identify which clients are (or we can easily help become) sanctioned, and finally (3) we ensure that the sanctioned clients all support auto-completion.

And this issue will make sure that `--list-all` is part of the official spec. Sounds good?

agnivade:
Sure.

agnivade:
Writing down some notes that I collected.

**Architecture**
- Clients must have a local cache to query from.
- Local cache should be stored under .tldr folder inside user's home directory.
- Fresh data should be downloaded from here - [page cache](http://tldr-pages.github.io/assets/tldr.zip), [index]( http://tldr-pages.github.io/assets/index.json).

I think nothing else should be specified regarding how the client wishes to store the pages internally. If it wishes to store the metadata in a RocksDB cache, that should be allowed.

**Features**

[...]

Issue #787, opened 2016-02-22
Support for aliases of commands with multiple names


oxguy3:
It'd be really useful if there were a way to note that one command is an alias of another. Many tools are distributed under multiple names, and the TLDR client ought to figure this out instead of leaving it to the user to decipher. Examples:
- A lot of GNU utilities get prefixed with a "g" when installed on systems like OS X.
- `python2` and `python3` are typically aliases of `python`, and so both those pages ought to redirect to the python.md page. Same applies to the executables of many other languages.
- In many cases, two different commands will actually just be slight variations of the same underlying tool, and don't really need separate pages. `cowthink` is just a variant of `cowsay`, `7z`/`7za`/`7zr` are all minor variants on the exact same utility, etc. In general, if two commands share the same man page, they probably ought to share the same tldr page as well.

Here's an example of how this might be implemented. The OS X tool "gdate" is really just the GNU tool "date". So _pages/osx/gdate.md_ could perhaps look like this:

```
# gdate

> REDIRECT: common/date
```

TLDR clients that recognize this alias syntax could automatically load the common/date page (obviously there'd have to be delay between announcing this and implementing it so clients could update to support it).

Implementation note: It'd be ideal if the build server checked that all redirects are valid, and that there are no redirect chains (i.e. you shouldn't be able redirect to another redirect. That would just create unnecessary complexity, and it'd also cause lots of extraneous HTTP requests for any client that queries the live repo instead of using a cached local copy.)

jedahan:
why not just use symlinks?

oxguy3:
Huh, guess I forgot git supported those. Will definitely need to make sure that it doesn't break anything (I wonder how GitHub handles it when, say, a JS-based client tries to download a symlinked file), but yeah I guess that might work better.

waldyrious:
Do they work seamlessly on all platforms?

jedahan:
As far as I can tell, symlinks work fine on osx, linux, free/openbsd. They do not seem to work on windows (barring ugly post-download hacks), but almost all of these commands require some sort of *nix environment, and are meant for quick lookup. So I think its OK.

The browser-as-client question is way more interesting, and we should find a way to make that work well. But I think that might be a peculiarity of that client, that can be worked around - it would indeed be nice if only the JS client needed changing to support symlinks, rather than us coming up with our own format and requiring all clients to implement it.

ostera:
tldr: already supported by the web client. You have my :+1: @oxguy3 

[...]

PR #2165, opened 2018-07-04
uptime: Fill out some more options

sbrl:
....I suspect that this belongs in `linux/` too. Thoughts?

----

### DIFF
--- a/pages/common/uptime.md
+++ b/pages/common/uptime.md
@@ -5,3 +5,15 @@
 - Print current time, uptime, number of logged-in users and other information:
 
 `uptime`
+
+- Show only the amount of time the system has been booted for:
+
+`uptime --pretty`
+
+- Print the date and time the system booted up at:
+
+`uptime --since`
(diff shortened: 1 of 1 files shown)


agnivade:
Yea, the platform thing is tricky. People try these commands from different systems. We need a comprehensive solution. The simplest way would be for the client to look into other folders and mention that the same page exists for a different platform.

sbrl:
Indeed. Do we have a spec for 'official' clients or something? If we did, then we could include it in there.

Alternatively there's the symbolic link solution - which [has issues on Windows](https://github.com/git-for-windows/git/wiki/Symbolic-Links). It's that age old question in #1966 and #787 again....

agnivade:
We have https://github.com/tldr-pages/tldr/issues/1065 which is not yet closed. It is more up @waldyrious' alley.

sbrl:
Indeed, @agnivade

PR #2169, opened 2018-07-06
snapper: add page

lasersharks:
----

### DIFF
--- a/pages/linux/snapper.md
+++ b/pages/linux/snapper.md
@@ -1,6 +1,6 @@
 # snapper
 
-> Filesystem snapshot management tool
+> Filesystem snapshot management tool.
 
 - List snapshot configs:
 
--- a/pages/linux/snapper.md
+++ b/pages/linux/snapper.md
@@ -1,0 +1,27 @@
+# snapper
+
+> Filesystem snapshot management tool
+
+- List snapshot configs:
+
+`snapper list-configs`
+
+- Create snapper config:
+
+`snapper -c {{config}} create-config {{path/to/subvolume}}`
+
+- List snapshots for a config:
+
+`snapper -c {{config}} list`
+
+- Create a new snapshot:
+
+`snapper -c {{config}} snapshot`
+
+- Delete a snapshot:
+
+`snapper -c {{config}} delete {{snapshot_number}}`
+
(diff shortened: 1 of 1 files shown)


pages/linux/snapper.md
```diff
@@ -0,0 +1,27 @@
...
+
+> Filesystem snapshot management tool.
+
+- List snapshot configs:
+
+`snapper list-configs`
+
+- Create snapper config:
+
+`snapper -c {{config}} create-config {{path/to/subvolume}}`
```
agnivade: Let's mention a concrete example of path/to/subvolume

pages/linux/snapper.md
```diff
@@ -0,0 +1,27 @@
...
+
+`snapper -c {{config}} snapshot`
+
+- Delete a snapshot:
+
+`snapper -c {{config}} delete {{snapshot_number}}`
+
+- Delete multiple snapshots:
+
+`snapper -c {{config}} delete {{snapshot_1}}-{{snapshot_2}}`
```
agnivade: For more than 2 snapshots, do you put `-` in between all of them ?

lasersharks: I think this may be a better example of this instead,
```
-- Delete multiple snapshots:
+- Delete a range of snapshots:
-`snapper -c {{config}} delete {{snapshot_1}}-{{snapshot_2}}`
+`snapper -c {{config}} delete {{snapshot_X}}-{{snapshot_Y}}`
```
A real world example being `snapper -c home delete 101-120`

pages/linux/snapper.md
lasersharks: Thanks for the comments - I'm not a real world example would work to well, as the subvolume could be any directory in the filesystem.
As example being `snapper -c home create-config /home`

pxgamer: Is it always a directory?

lasersharks: Yes - a subvolume would always be a directory

agnivade: Then how about just `path/to/directory` ?

lasersharks: I have added and pushed this :)

PR #2174, opened 2018-07-09
msg: add page

pxgamer:
----

### DIFF
--- a/pages/windows/msg.md
+++ b/pages/windows/msg.md
@@ -14,6 +14,10 @@
 
 `msg /server:{{server_name}} {{username|session_name|session_id}}`
 
+- Send a message to all users:
+
+`msg *`
+
 - Set a delay in seconds for a message:
 
 `msg /time:{{seconds}}`
--- a/pages/windows/msg.md
+++ b/pages/windows/msg.md
@@ -8,7 +8,7 @@
 
 - Send a message from stdin:
 
-`echo "{{message}}" | msg`
+`echo "{{message}}" | msg {{username|session_name|session_id}}`
 
 - Send a message to a specific server:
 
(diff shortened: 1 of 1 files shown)


pages/windows/msg.md
```diff
@@ -0,0 +1,19 @@
...
+
+> Send a message to a specific user or session.
+
+- Send a message to a specified user or session:
+
+`msg {{username|session_name|session_id}} {{message}}`
+
+- Send a message from stdin:
+
+`{{message}} | msg`
```
agnivade: Shouldn't it be echo {{message}} or whatever equivalent of that is in windows ? Simply typing "message" works ?

pxgamer:
@sbrl, to send a message to all users on a server, you need to do:
`msg /server:{{servername}} *`

The `/server` option needs to be combined with a user/`*` so I should probably include that 🤔

agnivade:
Yes, let's do that.

pages/windows/msg.md
```diff
@@ -0,0 +1,23 @@
+# msg
+
+> Send a message to a specific user or session.
+
+- Send a message to a specified user or session:
+
+`msg {{username|session_name|session_id}} {{message}}`
+
+- Send a message from stdin:
```
agnivade: In this case, which user|session does the message go to ?

pages/windows/msg.md
```diff
@@ -0,0 +1,23 @@
...
+
+`echo "{{message}}" | msg`
+
+- Send a message to a specific server:
+
+`msg /server:{{server_name}} {{username|session_name|session_id}}`
+
+- Send a message to all users:
```
agnivade: Maybe clarify that this sends to all users of the current machine ? Since this example is right after an example to send to a specific server.

pages/windows/msg.md
pxgamer: Good point, it needs to be specified. 🤦‍♂️

PR #2188, opened 2018-07-13
dnf: removed bad example

julienc91:
The `update` command of `dnf` doesn't work like `apt-get`'s : `dnf` automatically handles synchronizing list of packages.
In fact, it is a *deprecated* alias for the `upgrade` command.

Source : https://dnf.readthedocs.io/en/latest/command_ref.html#update-command

I also added `sudo` to all the listed commands, as all of them require root privileges (second point of the `CONTRIBUTING.md` file).

### DIFF
--- a/pages/linux/dnf.md
+++ b/pages/linux/dnf.md
@@ -4,16 +4,16 @@
 
 - Install a new package:
 
-`dnf install {{package}}`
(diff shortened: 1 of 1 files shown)


agnivade:
Thanks for the separate commits. :)

PR #1895, opened 2018-01-14
truncate: add page

felixonmars:
- [x] The page (if new), does not already exist in the repo.

- [x] The page (if new), has been added to the correct platform folder:  
      `common/` if it's common to all platforms, `linux/` if it's Linux-specific, and so on.

- [x] The page has 8 or fewer examples.

- [x] The PR is appropriately titled:  
      `<command name>: add page` for new pages, or `<command name>: <description of changes>` for pages being edited.

- [x] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines.

PR #2196, opened 2018-07-15
truncate: add page

felixonmars:
----

### DIFF
--- a/pages/linux/truncate.md
+++ b/pages/linux/truncate.md
@@ -1,0 +1,15 @@
+# truncate
+
+> Shrink or extend the size of a file to the specified size.
+
+- Set a size of 10 GB to an exsting file, or create a new file with the specified size:
+
+`truncate -s 10G {{filename}}`
+
+- Extend the file size by 50M, fill with holes (which reads as zero bytes):
+
+`truncate -s +50M {{filename}}`
+
+- Shrink the file by 2GiB, by removing data from the end of file:
+
+`truncate -s -2G {{filename}`


felixonmars:
Following https://github.com/tldr-pages/tldr/pull/1895

felixonmars:
Oops, missed the tokens. Thanks!

amejiarosario:
How do I update? I still get
```bash
$ tldr truncate
This page doesn't exist yet!
Submit new pages here: https://github.com/tldr-pages/tldr
```

sbrl:
I think you want `tldr --update`, @amejiarosario

amejiarosario:
thanks @sbrl, but still doesn't have it

```bash
$ tldr --update

$ tldr truncate
This page doesn't exist yet!
Submit new pages here: https://github.com/tldr-pages/tldr
```

felixonmars:
It works fine here. Are you running tldr on Mac or Windows? The PR only submitted the page into linux/ subfolder.

sbrl:
Ah, good point @mfrw. If you're on windows, you may need to do `tldr linux/truncate`.

amejiarosario:
Ah, yes, using Mac. It should be in common because I can have truncate in
Mac installing core-utils

felixonmars:
I am not sure about the policy here... Mac doesn't come with GNU coreutils, but it's installable with something like homebrew.

sbrl:
Hrm. If it's installable, then I think it's ok. This command is available in the _Windows subsystem for Linux_, right?

felixonmars:
Yes, since it's in the coreutils. I have opened a PR.

amejiarosario:
Actually giving some more thought, I think All documentation should be
available regardless of the host OS. That would avoid “not found” when
there’s documentation for it.

sbrl:
@amejiarosario - it depends on your client there. Some clients will search other platforms automatically, and others won't. I'd recommend opening an issue against the repository of the client you're using.

PR #2207, opened 2018-07-19
nim: add page

Calinou:

This adds a page for the [Nim](https://nim-lang.org/) compiler.

### DIFF
--- a/pages/common/nim.md
+++ b/pages/common/nim.md
@@ -5,20 +5,20 @@
 
 - Compile a source file:
 
-`nim compile {{file}}`
+`nim compile {{file.nim}}`
 
 - Compile and run a source file:
 
-`nim compile -r {{file}}`
+`nim compile -r {{file.nim}}`
 
 - Compile a source file with release optimizations enabled:
 
-`nim compile -d:release {{file}}`
+`nim compile -d:release {{file.nim}}`
 
 - Build a release binary optimized for low file size:
 
-`nim compile -d:release --opt:size {{file}}`
+`nim compile -d:release --opt:size {{file.nim}}`
 
-- Generate HTML documentation for a module:
+- Generate HTML documentation for a module (output will be placed in the current directory):
 
-`nim doc {{file}}`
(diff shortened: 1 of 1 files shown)


pages/common/nim.md
```diff
@@ -0,0 +1,24 @@
+# nim
+
+> The Nim compiler.
+> Processes, compiles and links Nim language source files.
+
+- Compile a source file:
+
+`nim compile {{file}}`
```
sbrl: Maybe `{{path/to/file.ext}}`, where `.ext` is the default extension for a nim source file? Also on the other examples below.

pages/common/nim.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Compile a source file with release optimizations enabled:
+
+`nim compile -d:release {{file}}`
+
+- Build a release binary optimized for low file size:
+
+`nim compile -d:release --opt:size {{file}}`
+
+- Generate HTML documentation for a module:
```
sbrl: Where does it put the documentation?

pages/common/nim.md
Calinou: The `.nim` extension can be ommitted, e.g. a file called `main.nim` can be compiled using `nim compile main`.

pages/common/nim.md
Calinou: The documentation is placed in the same directory as the module source file. Should I note that in the example?

pages/common/nim.md
agnivade: But can a file called main.txt be compiled using `nim compile main` ? If not, then I think we should add the extension.

Calinou: @agnivade No, it only detects the `.nim` extension; I think it makes sense to add the extension in the example.

agnivade:
We actually prefer spelled out options rather than short ones. For eg. we use `npm install`, rather than `npm i`. Once the user has grasped the command, he/she is free to use shorter options.

PR #2217, opened 2018-07-23
chkconfig: add page

tunght13488:
----

### DIFF
--- a/pages/linux/chkconfig.md
+++ b/pages/linux/chkconfig.md
@@ -8,20 +8,20 @@
 
 - Show a service's runlevel:
 
-`chkconfig --list ntpd`
+`chkconfig --list {{ntpd}}`
 
 - Enable service at boot:
 
-`chkconfig sshd on`
+`chkconfig {{sshd}} on`
 
 - Enable service at boot for runlevels 2, 3, 4, and 5:
 
-`chkconfig --level 2345 sshd on`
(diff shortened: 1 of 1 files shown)


pages/linux/chkconfig.md
```diff
@@ -0,0 +1,27 @@
+# chkconfig
+
+> Manage service runlevel.
```
sbrl: I assume that this command is part of a service manager of some description? I've got ubuntu with systemd - and it's not installed.

Perhaps we could expand this description to include the service manager that this works with? For example `Manage insert_service_manager_name_here service runlevels.` would work?

tunght13488: @sbrl this is the service manager for Centos 6. I'm not sure about Red Hat/Fedora though as I'm not using any of them.

For latest Centos with systemd we can do something like `systemctl {enable|disable} <service_name>`. The equivalent on Centos 6 is `chkconfig <service_name> on`

sbrl: Ah, I see. Perhaps we could reword this to `Manage the runlevel of services on CentOS 6`?

tunght13488: @sbrl sounds good to me. Updated

sbrl:
Thanks, @tunght13488 :smiley_cat:

PR #2222, opened 2018-07-24
makensis: add page

idleberg:
----

### DIFF
--- a/pages/common/makensis.md
+++ b/pages/common/makensis.md
@@ -1,0 +1,16 @@
+# makensis
+
+> Cross-platform compiler for NSIS installers.
+> It compiles a NSIS script into an Windows installer executable.
+
+- Compile a NSIS script:
+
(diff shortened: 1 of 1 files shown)


pages/common/makensis.md
```diff
@@ -0,0 +1,16 @@
+# makensis
+
+> Cross-platform compiler for NSIS installers.
+> It compiles a NSIS script into an Windows installer executable.
```
sbrl: Does this mean that it only works on Windows? Or does it work on all platforms (even though it only spits out installers for Windows)?

idleberg: The compiler itself is available for all POSIX platform, it's only the resulting installer that's for Windows. So adding it to `commons` was a deliberate choice.

sbrl: Ah, excellent! Just wanted to confirm :-)

PR #2231, opened 2018-07-27
apt-get: add pointer to apt-cache when searching for pacakges

sschlesier:
----

I realize this is a bit of a stretch, but I have personally had to repeatedly google how to search for a package w/ apt-get only to re-learn it was apt-cache many times.  I suspect I'm not the only person who has had this expectation of apt-get.

- [x ] The page has 8 or fewer examples.

- [x ] The PR is appropriately titled:  

- [ x] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines.

### DIFF
--- a/pages/linux/apt-get.md
+++ b/pages/linux/apt-get.md
@@ -1,6 +1,7 @@
 # apt-get
 
 > Debian and Ubuntu package management utility.
+> Search for packages using `apt-cache`.
 
 - Update the list of available packages and versions (it's recommended to run this before other `apt-get` commands):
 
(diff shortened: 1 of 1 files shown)


sbrl:
Hey, thanks for the PR, @sschlesier! Indeed, a reference to `apt-cache` would be splendid here. Perhaps we can create a second line of the description instead of adding an extra example here? I feel that this would be a neater solution.

sschlesier:
I'm glad that a reference to apt-cache is welcome.  

I'm not sure I fully understood the suggestions, but I've made an update that I think matches what was suggested.

If it's not correct, I'd be happy for a pointer to a page using the suggested style.

agnivade:
That is correct. To see what I meant try - `tldr uname` using the node client.

PR #2232, opened 2018-08-01
simplified video trimming

FraYoshi:
I think setting a start and end time simpler than a given time from the start time, let me know your thoughts about : )


agnivade:
@hvindin - thoughts ?

PR #2234, opened 2018-08-04
godoc: add options to use index

agnivade:
----

### DIFF
--- a/pages/common/godoc.md
+++ b/pages/common/godoc.md
@@ -18,6 +18,6 @@
 
 `godoc -write_index -index_files={{path/to/file}}`
 
-- Use the index file to enable search through the documentation:
+- Use the given index file to search the docs:
 
 `godoc -http=:{{6060}} -index -index_files={{path/to/file}}`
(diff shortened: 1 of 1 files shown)


pages/common/godoc.md
```diff
@@ -13,3 +13,11 @@
 - Serve documentation as a web server on port "6060":
 
 `godoc -http=:{{6060}}`
+
+- Create an index file:
+
+`godoc -write_index -index_files={{path/to/file}}`
+
+- Use the index file to enable search through the documentation:
```
agnivade: I don't like this sentence structure. What I am trying to say is using the index file will allow you to enable the search functionality, which can be used to search through the docs.

How do I put this in a short sentence ? 

@sbrl , @pxgamer

agnivade:
ping @pxgamer @sbrl

pages/common/godoc.md
pxgamer: I'm not really sure how to word it. :thinking:

agnivade: That is not exactly what it does. Using the `-index` option enables the search. Without it, you wouldn't be able to search at all.

sbrl: Ah, I see. What about `Use the specified index to search the docs`? That implies that you need to specify an index to use when searching, I think.

agnivade: Hmm .. ok. Not the best, but still better than mine. I will take it.

PR #2235, opened 2018-08-07
convert: add example for creating a gif

ShaunKarran:


pages/common/convert.md
```diff
@@ -17,3 +17,7 @@
 - Horizontally append images:
 
 `convert {{image1.png}} {{image2.png}} {{image3.png}} +append {{image123.png}}`
+
+- Create gif from series of images:
```
agnivade: Create a gif from a series of images with a given delay in (unit?) between them.

pages/common/convert.md
```diff
@@ -17,3 +17,7 @@
...
+
+`convert {{image1.png}} {{image2.png}} {{image3.png}} -delay 100 {{animation.gif}}`
```
agnivade: 100 in tokens.

pages/common/convert.md
ShaunKarran: Thanks for the comments! :)

pages/common/convert.md
ShaunKarran: The other examples set a specific value for the purpose of the example. I was following that format.

agnivade: Yes, we prefer specific values in examples. I just want the value to be in tokens.

PR #2238, opened 2018-08-09
chsh: added list chells command

telmotrooper:
----

- [x] The page has 8 or fewer examples.

- [x] The PR is appropriately titled:  
      `<command name>: add page` for new pages, or `<command name>: <description of changes>` for pages being edited.

- [x] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines.

PR #2242, opened 2018-08-13
Revert "chsh: added list chells command"

agnivade:
Reverts tldr-pages/tldr#2238

The flag is not present on all platforms


agnivade:
Going ahead and merging myself because this was an incorrect push and should be in platform specific folders.

PR #2243, opened 2018-08-14
sed: introduce print command in osx

mira01:
----

### DIFF
--- a/pages/osx/sed.md
+++ b/pages/osx/sed.md
@@ -1,6 +1,6 @@
 # sed
 
-> Run replacements based on regular expressions.
+> Edit text in a scriptable manner.
 
 - Replace the first occurrence of a string in a file, and print the result:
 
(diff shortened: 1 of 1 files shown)


mira01:
@agnivade I use this often when when I need to view only a part of a large configuration file. (eg. it is useful in INI files from header of a section to the end of section that is empty line)

The reason I introduced command in the form '{{line_number}},/^$/' instead of '/regex/, /regex/' is that I wanted to show usage of absolute addressing in sed. (This can be also used in replace command).

I found it quite sad that in so much powerful tool as *sed* there is presented only one its feature, the replace. Six variants of replace command.  I wanted to improve it, as I use it and I find it useful, but from time to time I forget the right syntax.

I hope i explained it understandably.

mira01:
Thank you @agnivade & @sbrl

PR #2254, opened 2018-08-23
brew-cask: Changed ‘search‘ as old was deprecated, added ‘zap‘

melgu:
----


agnivade:
If it is deprecated, we should mention the new way to search. You have removed it altogether. Does that mean users are not allowed to search at all ?

PR #2255, opened 2018-08-24
scripts: add python3 compatibility for the bot

### LINKED ISSUE #2185: print() is a function in Python 3
Python 3 syntax error:
* https://github.com/cclauss/tldr/blob/master/scripts/send_to_bot.py#L16

pxgamer:
Reapply the changes by cclauss from #2248.

Closes #2185

### DIFF
--- a/scripts/send_to_bot.py
+++ b/scripts/send_to_bot.py
@@ -1,7 +1,10 @@
 import json
 import os
 import sys
-import urllib2
+try:
+  import urllib.request as urllib2
+except ImportError:
+  import urllib2
 
 URL = 'https://tldr-bot.starbeamrainbowlabs.com/'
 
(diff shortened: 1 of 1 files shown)


MasterOdin:
You also need to modify the urllib2 line to make it python3 compatible.

If I were you folks, I'd also:
Change [line 4](https://github.com/tldr-pages/tldr/blob/55533b6f35dd727b9226e0982468b1256857558f/scripts/send_to_bot.py#L4) to be:
```
try:
    import urllib.request as urllib2
except ImportError:
    import urllib2
```

pxgamer:
@MasterOdin, cheers. I don't use Python personally, so I'm probably not the right person to do this change, I just thought it was worth having for future usage.

I've updated it with your suggested change. Does that make sense? 🤔

sbrl:
Does merging this mean that I'll need to we'll also need to update tldr-pages/tldr-bot, or change anything on my server (I host the tldr-bot)?

agnivade:
Nope.

agnivade:
Yep.

PR #2259, opened 2018-08-26
COMMUNITY-ROLES.md: format messages as quotes

waldyrious:
This allows the template messages to retain the semantic line-breaking
that improves the readability of their source code version control diffs,
while at the same time they can be copy-pasted to issues
without the line breaks (which is slightly problematic
because github includes the line breaks
even when rendering content as rich-text prose).

(Case in point: this very PR, which reuses the extended commit message.)

Also, slightly reword a sentence in the "Past owners" section,
to make it more general and cover more cases.

### DIFF
--- a/COMMUNITY-ROLES.md
+++ b/COMMUNITY-ROLES.md
@@ -203,7 +203,7 @@ and are responsible for performing role changes in the community.
 ### Past owners
 
 The following people used to be owners in the tldr-pages organization,
-and have reduced their participation or otherise moved on to other projects.
+and have reduced their participation or otherwise moved on to other projects.
 Their contributions and dedication have ensured the success of the tldr project,
 and for that they'll always be appreciated.
 
--- a/COMMUNITY-ROLES.md
+++ b/COMMUNITY-ROLES.md
@@ -87,25 +87,23 @@ using one of the template messages below as a base.
 
 Open an issue with the following message:
 
-```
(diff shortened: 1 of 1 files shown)


COMMUNITY-ROLES.md
```diff
@@ -211,7 +203,7 @@ and are responsible for performing role changes in the community.
 ### Past owners
 
 The following people used to be owners in the tldr-pages organization,
-and have since moved on to other projects.
```
agnivade: typo at "otherwise" :smile:

waldyrious: Haha thanks, let me fix it up real quick :)

waldyrious:
IMO it's OK not to have these documents in mind at all times, but it's good to be generally aware that the sort of information they contain is described somewhere in the repo (and if not, it should), and be able to find it when the situation arises.

agnivade:
@waldyrious - CLA issues again ?

waldyrious:
@agnivade I haven't (sadly) been merging TLDR-pages PRs recently to notice whether CLA-assistant has been failing consistently, but I did notice elsewhere that a lot of CI-like checks seem to be failing to report lately (not failing the check, but simply not reporting any status at all).

I believe GitHub may have changed something recently about how such integrations work, but that's mostly speculation. Have you guys noticed anything strange recently with the checks in this repo or others you're active in?

Issue #2213, opened 2018-07-22
List of GNU Core Utilities commands missing in tldr


Wesalius:
I compared [GNU Core Utilities commands list](https://en.wikipedia.org/wiki/List_of_GNU_Core_Utilities_commands) with contents of tldr/pages and found out that the following are missing from tldr:

* [x] arch
* [x] b2sum
* [x] chcon
* [x] csplit
* [x] dircolors
* [x] dirname
* [x] factor
* [x] false
* [x] fmt
* [x] groups
* [x] hostid
* [x] join
* [x] link
* [x] logname
* [x] mknod
* [x] nproc
* [x] numfmt
* [x] pathchk
* [x] pinky
* [x] pr
* [x] printenv
* [x] ptx
* [x] realpath
* [x] runcon
* [x] stdbuf
* [x] stty
* [x] true
* [x] tsort
* [x] unlink
* [x] users
* [x] vdir

We are missing 31 from the 101 commands listed in the list, so I am posting them here just in case anyone is looking for ideas for what to add to tldr :-)

sbrl:
Hey, thanks for the list! I've edited your comment to add checkboxes. Do you want to have a go yourself?

Pages are best created by people who understand the command they are about.

If you do, take a quick read of [CONTRIBUTING.md](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md), which guides you through the process of creating a new page here. Then, just simply send a pull request to this repository - and if you need any assistance, just ask :-)

ronniegane:
`csplit` added in https://github.com/tldr-pages/tldr/pull/2571

andrik:
chcon added in #2895

schneiderl:
Congrats for those involved in closing this 😄

sbrl:
Hey, wow! Looks like we're got em all :D

Closing this tracking meta-issue.

(Does anyone have any other cool lists we could work on next?)

waldyrious:
> (Does anyone have any other cool lists we could work on next?)

There's #1162 (which is only missing `notmuch` and `finch`), and #2214 (which still has a bunch to be added). 😉

sbrl:
Great idea @waldyrious! I'll do that now.

PR #2270, opened 2018-08-28
true: add page

pxgamer:
Part of #2213

----


pages/linux/true.md
```diff
@@ -0,0 +1,7 @@
+# true
+
+> Does nothing, but return a successful exit status code.
```
sbrl: Perhaps ...`but returns a successful exit status code of 0`... would read better?

pages/linux/true.md
```diff
@@ -0,0 +1,7 @@
+# true
+
+> Does nothing, but returns a successful exit status code of 0.
```
agnivade: Let us remove "Does nothing, but".

Also, can we add a sub-heading mentioning - "Use this with the || operator to make a command always exit with 0" ?

sbrl: Can do!

melgu:
The `true` and `false` commands also exist on macOS. Wouldn't it be better to place those in `common/` instead of `linux/`?

agnivade:
Do they exist on WSL ?

pxgamer:
@agnivade, pretty sure they do exist on WSL. So maybe I'll move them to `common`. I'll check in about 25 minutes when I'm at work. :+1:

pxgamer:
I can confirm that it is available on WSL. 👍

PR #2271, opened 2018-08-28
mutool: add page

fer22f:
Hello! This is the first draft at documenting `mutool`, it's very useful and fast but the CLI syntax is quite weird. It can basically replace some Xpdf tools to convert into text, images, merge PDFs and extract and query resources.

One of the difficulties is that there are two commands that do the same thing: `draw` and `convert`, but have different, better defaults. For example, to convert to images, I used `convert` because it has a better resolution default, but to convert to text I used `draw` because it outputs to standard output instead of creating a `out.txt` file.

Another one is that `mutool` doesn't support long parameters, just short ones, the syntax should be familiar for the most part, but it can be obtuse. Also, there is no documentation on the page selection syntax, but it [can be found in the source code](http://git.ghostscript.com/?p=mupdf.git;a=blob;f=source/fitz/string.c;hb=b7df29cce7f4815a2109f7b4339674c1e662e883#l488) and it doesn't support expressing "last two pages", for example.

In general, this tool does too much, it could've been split up into more pages, but these examples should suffice and inspire some exploration. Since converting to images and text is more common, they were placed at the top, and then merging and extracting resources.

----

### DIFF
--- a/pages/common/mutool.md
+++ b/pages/common/mutool.md
@@ -1,0 +1,23 @@
+# mutool
+
+> Convert PDF files, query information and extract data.
+
+- Convert pages 1-10 into 10 PNG images:
+
+`mutool convert -o {{image%d.png}} {{file.pdf}} {{1-10}}`
+
+- Convert pages 2, 3 and 5 of a PDF into text in the standard output:
+
+`mutool draw -F {{txt}} {{file.pdf}} {{2,3,5}}`
+
+- Concatenate two PDFs:
+
+`mutool merge -o {{output.pdf}} {{input1.pdf}} {{input2.pdf}}`
+
+- Query information about all content embedded in a PDF:
+
+`mutool info {{input.pdf}}`
+
+- Extract all images, fonts and resources embedded in a PDF out into the current directory:
+
+`mutool extract {{input.pdf}}`


pages/common/mutool.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Convert pages 1-10 into 10 PNG images:
+
+`mutool convert -o {{image%d.png}} {{file.pdf}} {{1-10}}`
+
+- Convert pages 2, 3 and 5 of a PDF into text in the standard output:
+
+`mutool draw -F {{txt}} {{file.pdf}} {{2,3,5}}`
+
+- Concatenate all pages skipping the first from two PDFs into a new one:
```
sbrl: The wording here seems a bit confusing to me. Perhaps something like this would read better?:

```markdown
- Merge 2 PDFs by concatenation, skipping the first page of each:
```

What do you think?

pages/common/mutool.md
```diff
@@ -0,0 +1,23 @@
...
+
+`mutool merge -o {{output.pdf}} {{input1.pdf}} 2-N {{input2.pdf}} 2-N`
+
+- Query information about all content embedded in a PDF:
+
+`mutool info {{input.pdf}}`
+
+- Extract all images, fonts and resources embedded in a PDF in the current directory:
```
sbrl: Perhaps ...`in a PDF out into the current`..... would read better?

fer22f: Yep, it would!

pages/common/mutool.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Convert pages 1-10 into 10 PNG images:
+
+`mutool convert -o {{image%d.png}} {{file.pdf}} {{1-10}}`
+
+- Convert pages 2, 3 and 5 of a PDF into text in the standard output:
+
+`mutool draw -F {{txt}} {{file.pdf}} {{2,3,5}}`
+
+- Concatenate two PDFs, skipping the first page of each:
```
agnivade: Can we change this example to just concatenate two pdfs ? I think that is the base version that we should mention. Rather than a complicated one.

fer22f: Sure! There's two options however:

* Use `1-N` syntax.
* Use no syntax at all, after all, it's optional.

agnivade: If it makes things simpler to explain, let us not use any syntax.

fer22f:
Ah, interesting, now I see that `git` has been split up into `git-commit`, `git-clone`, etc. In the future I could improve `mutool` and split it up like so.

agnivade:
Yes, sure. If the sub-commands have multiple usages, then you can split it into separate pages.

PR #2273, opened 2018-08-28
false: add page

sbrl:
Since @pxgamer  opened a PR for `true` (#2270), I decided to do one for `false`.

Part of #2213.

----


pages/linux/false.md
```diff
@@ -0,0 +1,8 @@
+# false
+
+> Does nothing, unsucessfully.
```
agnivade: I wonder if this line is helpful. Removing this doesn't change much.

pxgamer: I think keeping `true` (#2270) and `false` in the same format would be good.

sbrl: @pxgamer ...so remove this line/

agnivade: Yes. I think it makes sense to keep both pages in sync.

PR #2274, opened 2018-08-28
link: add page

pxgamer:
Part of #2213

----

### DIFF
--- a/pages/linux/link.md
+++ b/pages/linux/link.md
@@ -1,7 +1,8 @@
 # link
 
-> Create a link to an existing file.
+> Create a hard link to an existing file.
+> For more options, see the `ln` command.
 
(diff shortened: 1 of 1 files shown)


pages/linux/link.md
```diff
@@ -0,0 +1,7 @@
+# link
+
+> Create a link to an existing file.
```
sbrl: Does it create a symbolic link? If so, then maybe we could reword this to `Create a symbolic link`....?

Also, what about a see also `ln` if more advanced options are required?

pxgamer: It doesn't create a symlink. Only hard links. `ln` seems to be much better as it has symlinks and hard links, as well as error checking.

PR #2278, opened 2018-08-30
tsort: add page

MasterOdin:
A command from the list in #2213

### DIFF
--- a/pages/common/tsort.md
+++ b/pages/common/tsort.md
@@ -2,6 +2,6 @@
 
 > Perform a topological sort.
 
-- Performs a topological sort consistent with a partial sort per line of input broken up into elements separated by blanks:
+- Perform a topological sort consistent with a partial sort per line of input broken up into elements separated by blanks:
 
 `tsort {{file}}`
--- a/pages/common/tsort.md
+++ b/pages/common/tsort.md
@@ -1,0 +1,7 @@
+# tsort
+
+> Perform a topological sort.
+
+- Performs a topological sort consistent with a partial sort per line of input broken up into elements separated by blanks:
+
+`tsort {{file}}`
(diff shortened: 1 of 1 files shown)


pages/common/tsort.md
```diff
@@ -0,0 +1,7 @@
+# tsort
+
+> Perform a topological sort.
+
+- Perform a topological sort consistent with a partial sort per line of input broken up into elements separated by blanks:
```
sbrl: Erm I'm confused! What does this do? That's a lot of words jumbled up in that sentence.

Perhaps we could add some additional clarity in a second line of the command description too - maybe something like this?:

```
> In other words ......
```

MasterOdin: Yeah, that was my worry when I wrote this.

Essentially, it takes a file where each line is some list of "dependencies":
```
a b c
d
e f
b c d e
```
which can be read as:
- a is before b which is before c
- e is before f
- b is before c which is before d which is before e

and produces a topological sort of it:
```
a
b
c
d
e
f
```

So the gist of that sentence is:
- Produces a topological sort of a file
- Each line of the file gives a partial ordering of elements
- Each element is separated by a space

I'm not sure how wordy of an explanation you guys allow to expand on that singular sentence.

MasterOdin: The use case would be to see the ordering of dependencies within a directed acyclic graph (a common example is function calls within a program).

agnivade: We can condense it to - `Perform a topological sort consistent with a partial sort per line of input separated by blanks:`

sbrl: Ah, ok! Perhaps we could add a 2nd description line saying `A common use is on function calls in a program.`?

PR #2279, opened 2018-08-30
COMMUNITY-ROLES.md: clarify steps to remove owners

### LINKED ISSUE #2266: Clarify steps to remove organization owners
We have detailed descriptions of project governance processes in the [COMMUNITY-ROLES.md](https://github.com/tldr-pages/tldr/blob/master/COMMUNITY-ROLES.md) document, but some of them are clearer than others. In particular, the process for retiring inactive organization owners (of which I was a test subject in #2257) is a bit vague.

To fix this ambiguity, describing the process step-by-step would be ideal. As an initial proposal, I'd suggest changing the following passage in the section "For removing inactive organization members":

> Once the role change is completed, make sure to update the lists below accordingly.

... to read something like this:

> Once they acknowledge the message (or after a week without any reaction), go to github.com/orgs/tldr-pages/people and [...].
>
> Afterwards, move their name to the list of former owners, below in this document.
>
> Finally, once the membership changes are complete (both removal from the organization and adding as collaborator), and the lists are updated, close the issue opened to track this process.
 
The `[...]` part is what I'm unsure about. I suppose it would be something like "remove them from the org and then invite them as collaborators to the repos where they had participated", but then does that process offer the ability to convert them to external collaborators? And if so, is this the same thing as repo-specific collaborators? I suspect not, and if my suspicion is correct, then the latter (repo-specific collaborators) would be preferred, I think.

In addition, there should probably be an extra step to verify that the collaborator permissions are "write" rather than "admin" (this ensures the sets of available actions, and therefore responsibilities, are clearly separated between the different stages of participation, so that there's no confusion about what is expected from each person). It also reduces the risk of damaging actions in case the accounts are abandoned for longer periods and happen to be compromised.

I have suggestions for making all transition steps more defined, but this is what IMO would bring the "retirement" process in par with the other ones, so we should probably start there.

Thoughts?

waldyrious:
Fixes #2266.

### DIFF
--- a/COMMUNITY-ROLES.md
+++ b/COMMUNITY-ROLES.md
@@ -180,7 +180,7 @@ as is also described in the guidelines.
 In any case, we wish you the best of luck in your new endeavors!
 
 Once they acknowledge the message (or after a week without any reaction),
-go to github.com/orgs/tldr-pages/people, click the gear icon in their row,
+go to https://github.com/orgs/tldr-pages/people, click the gear icon in their row,
 and select the "Convert to outside collaborator" menu entry.
 
 Afterwards, edit this file to move their name to the "Past owners" section below.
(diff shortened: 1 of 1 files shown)


sbrl:
Merging with 3 approvals.

PR #2284, opened 2018-08-31
pinky: add page

pxgamer:
Part of #2213

----


pages/linux/pinky.md
```diff
@@ -0,0 +1,27 @@
+# pinky
+
+> Print user information using the `finger` protocol.
+
+- Display details for the current user:
```
sbrl: Perhaps `about` would be better than `for` here?

PR #2289, opened 2018-09-01
iw: add page

Conaclos:

### DIFF
--- a/pages/linux/iw.md
+++ b/pages/linux/iw.md
@@ -1,0 +1,19 @@
+# iw
+
+> Show and manipulate wireless devices.
+
+- Scan for available wireless networks:
+
+`iw dev {{wlp}} scan`
+
+- Join an open wireless network:
+
+`iw dev {{wlp}} connect {{SSID}}`
+
+- Close the current connection:
+
+`iw dev {{wlp}} disconnect`
+
+- Show information about the current connection:
+
+`iw dev {{wlp}} link`


pages/linux/iw.md
```diff
@@ -0,0 +1,19 @@
+# iw
+
+> Show / manipulate wireless devices.
```
sbrl: Perhaps `Show and manipulate`... would read better?

pages/linux/iw.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Scan available wireless networks:
```
sbrl: Perhaps `Scan for available`... would read better?

pages/linux/iw.md
Conaclos: I agree.
However, I choose this sentence form because it is already used for [`ip` tldr page](https://github.com/tldr-pages/tldr/blob/master/pages/linux/ip.md).

Conaclos:
@sbrl 
Is not important to stay consistent with existing tldr pages like [`ip` tldr page](https://github.com/tldr-pages/tldr/blob/master/pages/linux/ip.md)?

Is there tldr conventions about wireless interface name? Is `wlp` ok?

Maybe `wireless-interface` is better?

agnivade:
`wlp` is fine. `wireless-interface` is not better because it is still a generic value.

Keeping consistency is important. But subject to the page of course. In this case, it is a one letter change, which I am neutral to.

sbrl:
`wlp` isn't an actual interface though? Personally I'd prefer either `wlan0` or `path/to/wireless-interface`, as they're more descriptive.

Conaclos:
@sbrl 
I think that wlan0 is outdated.

sbrl:
True, @Conaclos, but the point is that everyone should recognise it :P

This should be fine though, I guess.

Issue #2214, opened 2018-07-22
List of util-linux commands missing in tldr


Wesalius:
I compared [List of util-linux commands](https://en.wikipedia.org/wiki/Util-linux#Included) with contents of tldr/pages and found out that the following are missing from tldr:

* [x] addpart
* [ ] agetty
* [x] blkdiscard
* [x] blkid
* [ ] blockdev
* [x] cfdisk
* [x] chcpu
* [x] chfn
* [x] chrt → #5398
* [ ] col
* [ ] colcrt
* [x] colrm
* [x] ctrlaltdel
* [x] delpart → #11135
* [x] eject
* [ ] fdformat
* [x] findfs
* [x] flock
* [ ] fsck.cramfs
* [ ] fsck.minix
* [ ] fsfreeze
* [x] fstrim
* [x] getopt
* [ ] Hex
* [x] ipcmk
* [x] ipcrm
* [x] isosize
* [ ] ldattach
* [x] line
* [x] login
* [x] lslocks
* [x] lslogins
* [x] mcookie
* [x] mesg
* [x] mkfs
* [ ] mkfs.bfs
* [x] mountpoint
* [x] namei
* [x] newgrp
* [x] nologin
* [x] nsenter
* [x] partx
* [ ] pg
* [ ] pivot_root
* [x] prlimit→ #7113
* [x] raw → #6557 
* [ ] readprofile
* [ ] resizepart
* [x] runuser
* [x] script
* [x] scriptreplay
* [ ] setarch
* [ ] setpriv
* [x] setsid → #7841
* [ ] setterm
* [ ] sfdisk
* [ ] sulogin
* [x] swaplabel → #10462
* [ ] switch_root
* [ ] tailf
* [x] taskset
* [ ] tunelp
* [x] ul
* [x] unshare → #7920
* [x] utmpdump
* [x] uuidd
* [x] uuidgen
* [x] vipw
* [x] wdctl→ #7112
* [x] wipefs

We are missing 35 from the 107 commands listed in the list, so I am posting them here just in case anyone is looking for ideas for what to add to tldr :-)

sbrl:
Thanks for this list! I've updated your comment to add checkboxes. Do you want to have a go yourself?

Pages are best created by people who understand the command they are about.

[...]

PR #2291, opened 2018-09-02
eject: add page

pxgamer:
Part of #2214

----

### DIFF
--- a/pages/linux/eject.md
+++ b/pages/linux/eject.md
@@ -16,16 +16,16 @@
 
 - Toggle whether a device's tray is open or closed:
 
-`eject -T {{name_of_device}}`
+`eject -T {{/dev/cdrom}}`
 
 - Eject a cd drive:
 
-`eject -r {{name_of_device}}`
+`eject -r {{/dev/cdrom}}`
 
 - Eject a floppy drive:
 
-`eject -f {{name_of_device}}`
+`eject -f {{/dev/cdrom}}`
 
 - Eject a tape drive:
 
-`eject -q {{name_of_device}}`
+`eject -q {{/dev/cdrom}}`
--- a/pages/linux/eject.md
+++ b/pages/linux/eject.md
@@ -12,7 +12,7 @@
 
 - Eject a specific device (the default order is cd-rom, scsi, floppy and tape):
 
-`eject {{name_of_device}}`
+`eject {{/dev/cdrom}}`
 
 - Toggle whether a device's tray is open or closed:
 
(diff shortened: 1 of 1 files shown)


pages/linux/eject.md
```diff
@@ -0,0 +1,31 @@
...
+
+> Eject cds, floppy disks and tape drives.
+
+- Display the default device:
+
+`eject -d`
+
+- Eject the default device:
+
+`eject`
```
mfrw: I would suggest swapping this item with the first one.

pages/linux/eject.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Eject a specific device:
+
+`eject {{name_of_device}}`
+
+- Toggle whether a device's tray is open or closed:
+
+`eject {{name_of_device}} -T`
```
mfrw: Shouldn't the command line options come before the positional parameters?

pages/linux/eject.md
pxgamer: When I was creating this, I actually wasn't sure which way round it should be.

I was going to put it first as it's the call without any arguments.

But, I then also thought, as people may not know what their default drive is, should the check be first in a logical order?

pages/linux/eject.md
pxgamer: It doesn't seem to make any difference, but I'll maybe move them to the start as that's how man pages order them.

pages/linux/eject.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Display the default device:
+
+`eject -d`
+
+- Eject the default device:
+
+`eject`
+
+- Eject a specific device:
```
agnivade: In this case, what does it eject ? CD, floppy or tape ? We should mention that.

pxgamer: It's ordered as:
`cd-rom, scsi, floppy and tape`

agnivade: Let's mention that in parentheses.

agnivade:
Perhaps also mention a concrete value for {{device}} ?

pxgamer:
@agnivade, sure. Sorry, somehow I missed your comment before.

agnivade:
Umm, sorry if I wasn't clear. I actually meant for all the examples. :P

pxgamer:
Ah, whoops. Should I do different ones for each? Like, `/mnt/cdrom`, `/dev/cdrom`, `/mnt/floppy`

Or just do all of them as `/dev/cdrom`?

agnivade:
/dev/cdrom is fine.

PR #2292, opened 2018-09-03
dpkg: add -c option (list local deb file contents)

muendelezaji:
Add example for listing package contents of a downloaded `.deb` file, because `dpkg -L` only works with the `apt` name of the package (and only when installed).

----

### DIFF
--- a/pages/linux/dpkg.md
+++ b/pages/linux/dpkg.md
@@ -4,7 +4,7 @@
 
 - Install a package:
 
-`dpkg -i {{/path/to/file}}`
+`dpkg -i {{path/to/file.deb}}`
 
 - Remove a package:
 
@@ -17,6 +17,10 @@
 - List package contents:
 
 `dpkg -L {{package_name}}`
+
+- List contents of a local package file:
+
+`dpkg -c {{path/to/file.deb}}`
 
 - Find out which package owns a file:
 


pages/linux/dpkg.md
```diff
@@ -21,3 +21,7 @@
 - Find out which package owns a file:
 
 `dpkg -S {{file_name}}`
+
+- List contents of local package file:
+
+`dpkg -c {{/path/to/file}}`
```
pxgamer: Remove the prefixing slash from here. 👍

pages/linux/dpkg.md
```diff
@@ -21,3 +21,7 @@
 - Find out which package owns a file:
 
 `dpkg -S {{file_name}}`
+
+- List contents of local package file:
```
pxgamer: ...`of a local`...

pages/linux/dpkg.md
muendelezaji: For the sake of consistency, should it be removed from the first entry as well? i.e. `dpkg -i` example (looks like it was missed during #947)

pages/linux/dpkg.md
```diff
@@ -21,3 +21,7 @@
 - Find out which package owns a file:
 
 `dpkg -S {{file_name}}`
+
+- List contents of a local package file:
```
agnivade: Please move this one example up.

muendelezaji: Done

muendelezaji:
> For further clarity, I would suggest changing path/to/file to path/to/file.deb everywhere.

Updated. Note that some of the options actually expect the apt name rather than the deb file name. So those are correct as they are.

PR #2294, opened 2018-09-04
set: add page (#2268)

### LINKED ISSUE #2268: page request: set
Requesting a page for the built-in command [set](http://linuxcommand.org/lc3_man_pages/seth.html)

mfrw:
Closes #2268 

----

### DIFF
--- a/pages/common/set.md
+++ b/pages/common/set.md
@@ -1,0 +1,7 @@
+# set
+
+> Display, Set or Unset value of shell attributes and positional parameters.
+
+- Display the names and values of shell variables
+
+`set`
--- a/pages/common/set.md
+++ b/pages/common/set.md
@@ -16,4 +16,4 @@
 
 - Set various options, e.g. enable `vi` style line editing:
 
-`set -o vi`
+`set -o {{vi}}`
--- a/pages/common/set.md
+++ b/pages/common/set.md
@@ -2,6 +2,6 @@
 
 > Display, Set or Unset value of shell attributes and positional parameters.
 
-- Display the names and values of shell variables
+- Display the names and values of shell variables:
 
 `set`
--- a/pages/common/set.md
+++ b/pages/common/set.md
@@ -1,6 +1,6 @@
 # set
 
-> Display, Set or Unset value of shell attributes and positional parameters.
+> Display, set or unset value of shell attributes and positional parameters.
 
 - Display the names and values of shell variables:
 
--- a/pages/common/set.md
+++ b/pages/common/set.md
@@ -5,3 +5,15 @@
 - Display the names and values of shell variables:
 
 `set`
+
+- Mark variables that are modified or created for export:
+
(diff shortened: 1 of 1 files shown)


agnivade:
Is this the only example common across all shells ?

mfrw:
@agnivade no, there are more but I did not check across all available shells.
Can you suggest which shells should I consider?
I have strong feeling options should be the same for all, but I am not sure.

agnivade:
Since it is in common, windows support is must. So I would say powershell, bash, zsh are essential.

mfrw:
@agnivade if we have to include `powershell`  then I would suggest not adding anything else.

agnivade:
You can also split the page into 2. Put one inside windows/ and the other with more options in common/. Windows clients will automatically take the more specific page.

mfrw:
@agnivade I think this is a very good idea. Let me split them up into two pages :)
Thanks for the suggestion.

mfrw:
@agnivade It turns out that `set` already had a windows page.

pages/common/set.md
```diff
@@ -0,0 +1,19 @@
+# set
+
+> Display, Set or Unset value of shell attributes and positional parameters.
```
sbrl: Not sure that `Set` and `Unset` need to be capitalised here.

mfrw: yes sure...

pages/common/set.md
```diff
@@ -0,0 +1,19 @@
...
+
+`set -a`
+
+- Notify of job termination immediately:
+
+`set -b`
+
+- Set various options, e.g. enable `vi` style line editing:
+
+`set -o vi`
```
sbrl: Perhaps `vi` should be surrounded by `{{` and `}}`?

pages/common/set.md
```diff
@@ -0,0 +1,19 @@
+# set
+
+> Display, set or unset value of shell attributes and positional parameters.
```
sbrl: Perhaps `value` would read better as `values` here?

pages/common/set.md
mfrw: Ahhh... yes :(

pages/common/set.md
sbrl: Yeah, that would sound better than `... value of shell attributes ...`, as the `shell attributes` is plural, so the `value` should also be plural.

mfrw: sure

mfrw: @sbrl If you could help me in squashing all the commits as a single commit.
Just like gerrit does.

sbrl: Yeah, when we merge a PR, we have a special button that lets us squash all the commits automagically - so you don't need to do anything there :slightly_smiling_face:

mfrw: ahhh ! thanks

PR #2298, opened 2018-09-05
multitail: add page

mattsches:
----

### DIFF
--- a/pages/common/multitail.md
+++ b/pages/common/multitail.md
@@ -1,0 +1,19 @@
+# multitail
+
+> Monitor last part of multiple files in multiple windows in a terminal.
+
+- Monitor two files in two separate windows:
+
+`multitail {{file1}} {{file2}}`
+
+- Merge two files in one window:
+
+`multitail {{file1}} -I {{file2}}`
+
+- Show two files in two windows side by side:
+
+`multitail -s {{file1}} {{file2}}`
+
+- Show output of two commands:
+
+`multitail -l "ping {{host1}}" -l "ping {{host1}}"`


pages/common/multitail.md
```diff
@@ -0,0 +1,19 @@
...
+
+`multitail {{file1}} -I {{file2}}`
+
+- Show two files in two windows side by side:
+
+`multitail -s {{file1}} {{file2}}`
+
+- Show output of two commands:
+
+`multitail -l "ping {{host1}}" -l "ping {{host1}}"`
```
agnivade: Actually, the entire command should be in `{{}}`. So maybe write something like - `multitail -l "{{echo hello}}" -l "{{echo world}}"`

Also - what is the difference between this command and simply doing `echo hello && echo world` ?

pages/common/multitail.md
```diff
@@ -0,0 +1,19 @@
+# multitail
+
+> Monitor last part of multiple files in multiple windows in a terminal.
```
agnivade: What does "Monitor" mean here ? Does it continuously read like `tail -f` ? If so, we should clarify that behavior.

pages/common/multitail.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Monitor two files in two separate windows:
+
+`multitail {{file1}} {{file2}}`
+
+- Merge two files in one window:
```
agnivade: Merge _the output of_ two files ..

pages/common/multitail.md
mattsches: `ping` continuously adds data, so `multitail -l "{{ping host1}}" -l "{{ping host2}}"` will keep adding lines for both commands in parallel. So what about:

```
- Keep showing appended output of two commands:
```

pages/common/multitail.md
mattsches: Rephrasing … what do you think about the following?

```
Keep showing continuously added data of multiple files in multiple windows.
```

pages/common/multitail.md
mattsches: Maybe

```
- Follow the merged output two files in one window
```

mattsches:
I wasn't really sure what `common` is about, so it should be moved to `linux`, I guess.

agnivade:
> I wasn't really sure what common is about, so it should be moved to linux, I guess.

It says right in the PR template.

"The page (if new), has been added to the correct platform folder:
common/ if it's common to all platforms, linux/ if it's Linux-specific, and so on."

Let us know if it wasn't clear so that we can clarify it further.

pages/common/multitail.md
agnivade: I like - 'Follow the interleaved output from two commands:'. I think `follow` is a better term which explains what is going on, rather than `show appended output` which stutters a bit.

pages/common/multitail.md
agnivade: Keeping in line with my previous comment - `Follow output of multiple files in separate windows.`

pages/common/multitail.md
agnivade: .. merged output _of_ two files ..

mattsches:
Sorry, can't find the time right now to pursue this any further. Closing ...

PR #2301, opened 2018-09-05
jpegoptim: added max size param example

nutellinoit:
----

Hi again, I find it useful to have an example of recursive optimization of `jpegoptim`
I use it only this way , also with --size param

`find {{.}} -name "{{*.jpg}}" -exec jpegoptim --strip-all --all-progressive --size={{250k}}  {} \;`

### DIFF
--- a/pages/common/jpegoptim.md
+++ b/pages/common/jpegoptim.md
@@ -14,6 +14,10 @@
 
 `jpegoptim --all-progressive {{image1.jpeg}} {{image2.jpeg}} {{imageN.jpeg}}`
 
+- Force the output images to be fixed max size:
+
+`jpegoptim --size={{250k}} {{image1.jpeg}} {{image2.jpeg}} {{imageN.jpeg}}`
+
 - Recursive folder optimization:
 
-`find {{.}} -name "{{*.jpg}}" -exec jpegoptim --strip-all --all-progressive {} \;`
+`find {{.}} -name "{{*.jpg}}" -exec jpegoptim --strip-all --all-progressive --size={{250k}} {} \;`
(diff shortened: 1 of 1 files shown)


pages/common/jpegoptim.md
```diff
@@ -13,3 +13,11 @@
 - Force the output images to be progressive:
 
 `jpegoptim --all-progressive {{image1.jpeg}} {{image2.jpeg}} {{imageN.jpeg}}`
+
+- Force the output images to be fixed max size:
+
+`jpegoptim --size={{250k}} {{image1.jpeg}} {{image2.jpeg}} {{imageN.jpeg}}`
+
+- Recursive folder optimization:
```
agnivade: Sorry, but we do not usually mention examples like these. It is out of scope for tldr. This falls into the category of cheat sheets.

Rarely we might consider one or two exceptions, but I think this is fairly straightforward to construct, given that `tldr find` already shows the -exec usage.

sbrl: I'm inclined to agree here.

If you're curious for an alternative way of doing that that maxes out all CPU cores though, I've got this one-liner stashed away:

```
find . -regextype egrep -iregex '^.*\.jpe?g$' -print0 | xargs -0 -n1 -P4 jpegoptim --preserve --all-progressive
```

nutellinoit: @sbrl thank you! 

i changed the page only with --size line

pages/common/jpegoptim.md
```diff
@@ -13,3 +13,7 @@
 - Force the output images to be progressive:
 
 `jpegoptim --all-progressive {{image1.jpeg}} {{image2.jpeg}} {{imageN.jpeg}}`
+
+- Force the output images to be fixed max size:
```
agnivade: I think "fixed upto a max size" is grammatically correct.

sbrl: Perhaps `to have a fixed maximum size` would read better?

PR #2314, opened 2018-09-08
mpv: add page

mortenfyhn:
----

Thought this could be useful for some. Is there a specific style for keyboard shortcuts that can be used while the command/program runs? mpv accepts a bunch of shortcuts while playing video, and some of those would be nice to add to the tldr page, but I'm not sure if that's a) something you want in the tldr pages, and b) what syntax to use for that.

### DIFF
--- a/pages/common/mpv.md
+++ b/pages/common/mpv.md
@@ -9,3 +9,7 @@
 - Play a video at specified speed (0.01 to 100, default 1):
 
 `mpv --speed {{speed}} {{video_file}}`
+
+- Play a video using a profile defined in your `mpv.conf` file:
+
+`mpv --profile {{profile_name}} {{video_file}}`
--- a/pages/common/mpv.md
+++ b/pages/common/mpv.md
@@ -1,0 +1,11 @@
+# mpv
+
+> A media player.
+
+- Play a video:
+
+`mpv {{video_file}}`
+
+- Play a video at specified speed (0.01 to 100, default 1):
+
+`mpv --speed {{speed}} {{video_file}}`
--- a/pages/common/mpv.md
+++ b/pages/common/mpv.md
@@ -22,6 +22,6 @@
 
 `mpv --speed {{speed}} {{file}}`
 
-- Play a file using a profile defined in your `mpv.conf` file:
+- Play a file using a profile defined in the `mpv.conf` file:
 
 `mpv --profile {{profile_name}} {{file}}`
--- a/pages/common/mpv.md
+++ b/pages/common/mpv.md
@@ -1,15 +1,27 @@
 # mpv
 
-> A media player.
+> A audio/video player based on MPlayer.
 
-- Play a video:
+- Play a video or audio file:
 
-`mpv {{video_file}}`
+`mpv {{file}}`
 
-- Play a video at specified speed (0.01 to 100, default 1):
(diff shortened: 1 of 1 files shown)


roblav96:
@mortenfyhn Dude this is so awesome! =]

mortenfyhn:
Yeah, I know only mentioning `--speed` is a bit a limited. I could add some more, but I'm honestly not sure which of the many other options are most used / useful. I added one more though.

Any input on my keyboard shortcut question?

pages/common/mpv.md
```diff
@@ -0,0 +1,11 @@
+# mpv
+
+> A media player.
```
mfrw: Perhaps `A video player based on MPlayer` would sound better...

pages/common/mpv.md
```diff
@@ -9,3 +9,7 @@
 - Play a video at specified speed (0.01 to 100, default 1):
 
 `mpv --speed {{speed}} {{video_file}}`
+
+- Play a video using a profile defined in your `mpv.conf` file:
```
mfrw: How about we **drop** the `your`.
'Play a video using a profile defined in `mpv.conf` file:'

mfrw:
@mortenfyhn Thank you for the PR. If you want to add shortcuts which you think are useful, you could have a look at `emacs` or `vim` page for pointers.
I have also left some review comments.

pages/common/mpv.md
mortenfyhn: It's not just a video player though, it plays audio too (and maybe more, I'm not sure). I could add a line for "play an audio file" just to make that clear.

pages/common/mpv.md
mortenfyhn: I don't really agree, because the user has to define his own `mpv.conf` file. I could write 'in _the_ `mpv.conf`' file, but that would make it sound like it already exists.

sbrl: I think this is fine as-is here, as the `mpv.conf` file is likely to be in the user's _own_ home directory.

pages/common/mpv.md
sbrl: Perhaps something like this would read better?:

```
> A audio and video player.
> Based on MPlayer.
```

pages/common/mpv.md
```diff
@@ -0,0 +1,15 @@
+# mpv
+
+> A media player.
+
+- Play a video:
+
+`mpv {{video_file}}`
+
+- Play a video at specified speed (0.01 to 100, default 1):
```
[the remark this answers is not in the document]
mortenfyhn: I agree.

pages/common/mpv.md
mortenfyhn: I've updated this.

pages/common/mpv.md
mfrw: I agree

agnivade: FYI - We try to avoid personal pronouns and aim to keep the language impersonal. If you want to keep `your`, I would suggest `one's` or something similar which makes it sound impersonal.

mortenfyhn: I've pushed a new one using `the`, I think it reads better than `one's`.

mortenfyhn:
Thanks!

PR #2316, opened 2018-09-08
flatpak: add page

YtvwlD:
----

### DIFF
--- a/pages/linux/flatpak.md
+++ b/pages/linux/flatpak.md
@@ -1,0 +1,27 @@
+# flatpak
+
+> Build, install and run flatpak applications and runtimes.
+
+- Run an installed application:
+
+`flatpak run {{name}}`
+
+- Install an application from a remote source:
+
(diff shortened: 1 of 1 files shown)


pages/linux/flatpak.md
```diff
@@ -0,0 +1,27 @@
+# flatpak
+
+> Build, install and run flatpak applications and runtimes.
+
+- Run an installed application:
+
+`flatpak run {{name}}`
+
+- Install an application from a remote source:
```
pxgamer: Perhaps move this first if you have to install them before running them? :thinking:

YtvwlD: Well, yeah. But you would also have to add a source, first.
I don't know. I put running an application first, because that's pretty straightforward.

pxgamer: We usually try to make tldr pages in the order that users will use them. I'm going to say this is fine for now. :+1:

PR #2317, opened 2018-09-09
asar: add page

idleberg:
----

### DIFF
--- a/pages/common/asar.md
+++ b/pages/common/asar.md
@@ -1,19 +1,19 @@
 # asar
 
-> A file archiver for the Electron platform
+> A file archiver for the Electron platform.
 
-- Archive a file or folder
+- Archive a file or folder:
 
 `asar pack {{path/to/file}} {{archived.asar}}`
 
-- Extract an archive
+- Extract an archive:
 
 `asar extract {{archived.asar}}`
 
-- Extract a specific file from an archive
+- Extract a specific file from an archive:
 
 `asar extract-file {{archived.asar}} {{file}}`
 
-- List files of an archive
+- List files of an archive:
 
 `asar list {{archived.asar}}`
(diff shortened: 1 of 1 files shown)


mfrw:
Merged, thanks! 
👍

pages/common/asar.md
```diff
@@ -0,0 +1,19 @@
+# asar
+
+> A file archiver for the Electron platform.
```
sbrl: I know this got merged already, but what's the `Electron platform` in this instance, and how does that relate to file archiving? The only _Electron_ I'm aware of is the shell that the _Atom_ editor uses.

I'm confused.

pages/common/asar.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Extract an archive:
+
+`asar extract {{archived.asar}}`
+
+- Extract a specific file from an archive:
+
+`asar extract-file {{archived.asar}} {{file}}`
+
+- List files of an archive:
```
sbrl: Perhaps ...`inside an archive:` would read better?

pages/common/asar.md
sbrl: I know this got merged already, but what's the `Electron platform` in this instance, and how does that relate to file archiving? The only _Electron_ I'm aware of is the shell that the _Atom_ editor uses.

I'm confused.

idleberg: It's the same thing and its standard archive format.

https://github.com/electron/asar

I don't know if it's worth mentioning that's made for Electron. To me it seems more useful than the official description „Simple extensive tar-like archive format with indexing”, but that's up to you guys :)

pages/common/asar.md
idleberg: I agree. Or perhaps streamline it with other archivers, e.g. `7z` which uses „List the contents of an archive file”

pages/common/asar.md
sbrl: Opened #2325.

This description does seem more useful, I'm just curious as to whether we can clarify it a bit more :-)

PR #2324, opened 2018-09-11
unlink: add page

pxgamer:
Part of #2213

----

### DIFF
--- a/pages/common/unlink.md
+++ b/pages/common/unlink.md
@@ -1,0 +1,7 @@
+# unlink
+
+> Remove a file from a filesystem.
+
+- Remove the specified file:
+
+`unlink {{path/to/file}}`


pages/common/unlink.md
```diff
@@ -0,0 +1,7 @@
+# unlink
+
+> Remove a file from a filesystem.
+
+- Remove the specified file if it is the last link:
```
agnivade: This makes the user wonder what happens if it is _not_ the last link. I find the man page description clear enough. How about - "Delete a name from the file system. (If the name was the last link, the file is removed too)".

pages/common/unlink.md
```diff
@@ -0,0 +1,7 @@
+# unlink
+
+> Remove a file from a filesystem.
```
agnivade: Let us change this to "Delete a name from the file system" since this is more precise. It does not always delete a file.

mfrw: How about:
`Decrement the link count of a file, deleting the file if the link count reduces to '0'`
This is in view of the change suggested by @agnivade 
I wanted to use the word `remove` instead of `deleting` but am unable to decide...

sbrl:
See my comment in reply to @mfrw's original comment on 9071bd9.

PR #2327, opened 2018-09-11
nmap: add example for scanning live hosts on a network.

jottr:
----

### DIFF
--- a/pages/common/nmap.md
+++ b/pages/common/nmap.md
@@ -30,3 +30,7 @@
 - Perform TLS cipher scan against a host to determine supported ciphers and SSL/TLS protocols:
 
 `nmap --script ssl-enum-ciphers {{address_or_addresses}} -p 443`
+
+- List live hosts with hostnames on a specific network:
+
+`nmap -sn 10.0.0.1/24`


pages/common/nmap.md
```diff
@@ -30,3 +30,7 @@
 - Perform TLS cipher scan against a host to determine supported ciphers and SSL/TLS protocols:
 
 `nmap --script ssl-enum-ciphers {{address_or_addresses}} -p 443`
+
+- List live hosts with hostnames on a specific network:
+
+`nmap -sn 10.0.0.1/24`
```
pxgamer: Is this different to the first example?

agnivade:
We already have a -sn example.

jottr:
True.
I overlooked the other example. 
Imho, I find my example to be clearer than: `nmap -sn ip_or_hostname optional_another_address` . 
Following that example I would do: `nmap -sn 192.168.0.1` instead of  `nmap -sn 192.168.0.1/24`

agnivade:
Sure, in that case I would ask you to change all the other examples to be ip addresses too. We always prefer concrete values rather than generic text like ip_address or hostname.

pages/common/nmap.md
mebeim: It's actually the same as the first example. This PR should be closed.

mebeim:
Closing this since the added example is equal to the first one.

PR #2328, opened 2018-09-12
users: add page

pxgamer:
Part of #2213

----

### DIFF
--- a/pages/linux/users.md
+++ b/pages/linux/users.md
@@ -1,0 +1,11 @@
+# users
+
+> Display a list of logged in users.
+
+- Display a list of logged in users:
+
+`users`
+
+- Display a list of logged in users according to a specific file:
+
+`users {{path/to/file}}`
--- a/pages/linux/users.md
+++ b/pages/linux/users.md
@@ -8,4 +8,4 @@
 
 - Display a list of logged in users according to a specific file:
 
-`users {{path/to/file}}`
+`users {{/var/log/wmtp}}`


pages/linux/users.md
```diff
@@ -0,0 +1,11 @@
+# users
+
+> Display a list of logged in users.
```
mfrw: How about ..
`Lists the login names of the users currently on the system, in sorted order`

agnivade: I think "logged in users" is fine.

pages/linux/users.md
```diff
@@ -0,0 +1,11 @@
...
+
+- Display a list of logged in users:
+
+`users`
+
+- Display a list of logged in users according to a specific file:
```
agnivade: What does "according to a file" mean ? What does the file specify ?

pxgamer: You can specify a custom file. By default it checks `/var/run/utmp`. But there's also `/var/log/wmtp` which is a historical dump of `utmp`. And `btmp` which displays failed login attempts.

pages/linux/users.md
sbrl: I think that's a bit too verbose, @mfrw. If possible, we prefer more concise descriptions as opposed to longer ones as they're easier to understand quickly :-)

pages/linux/users.md
agnivade: Ah I see. I thought it will always show the current logged in users. Perhaps we can mention "/var/log/wmtp" instead of path/to/file ? That may make things more clear.

PR #2334, opened 2018-09-13
crystal: add page

aa-tan:
----

### DIFF
--- a/pages/common/crystal.md
+++ b/pages/common/crystal.md
@@ -14,7 +14,7 @@
 
 `crystal play`
 
-- Create a project directory for a crystal application: 
+- Create a project directory for a crystal application:
 
 `crystal init app {{name of application}}`
 
--- a/pages/common/crystal.md
+++ b/pages/common/crystal.md
@@ -1,0 +1,23 @@
+# crystal
+
+> Crystal language compiler
+
+- Run a crystal file:
+
+`crystal {{path/to/file.cr}}`
+
+- Compile a file and all dependencies to a single executable:
+
+`crystal build {{path/to/file.cr}}`
+
+- Start a local interactive server for testing the language:
+
+`crystal play`
+
+- Create a project directory for a crystal application: 
+
+`crystal init app {{name of application}}`
+
+- Display all help options
+
+`crystal help`
--- a/pages/common/crystal.md
+++ b/pages/common/crystal.md
@@ -1,6 +1,6 @@
 # crystal
 
-> Crystal language compiler
+> Crystal language tool for compiling.
 
 - Run a crystal file:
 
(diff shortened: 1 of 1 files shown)


aa-tan:
Any idea why commit 8a49735 still did not pass the style guidelines? I ended up recommitting after deleting and rewriting line 17. Possibly some invisible character?

aa-tan:
Disregard that, there was a space at the end of the line. Also, apologies that the two commits that followed are not in line with the contribution guidelines, let me know if this PR should be deleted, I can submit one again after making the proper changes if necessary.

mfrw:
@aa-tan Thank you for your contribution. No, you don't have to send in a new PR. We will try to work with this :)
Don't worry about the mistakes, it happens to all of us :)
Welcome aboard!

pages/common/crystal.md
```diff
@@ -14,7 +14,7 @@
 
 `crystal play`
 
-- Create a project directory for a crystal application: 
+- Create a project directory for a crystal application:
 
 `crystal init app {{name of application}}`
```
mfrw: Can we have a concrete example here instead of a generic one?

pages/common/crystal.md
```diff
@@ -14,7 +14,7 @@
 
 `crystal play`
```
mfrw: Are there any other common usages that we could add to this page?

pages/common/crystal.md
```diff
@@ -0,0 +1,23 @@
...
+
+`crystal build {{path/to/file.cr}}`
+
+- Start a local interactive server for testing the language:
+
+`crystal play`
+
+- Create a project directory for a crystal application:
+
+`crystal init app {{name of application}}`
```
mfrw: Can we have a concrete example here instead of a generic one?

pxgamer: Just to note, this should be changed to token syntax like `name_of_application` (or `application_name`). Unless it gets changed to a concrete example. :thinking:

pages/common/crystal.md
```diff
@@ -0,0 +1,23 @@
+# crystal
+
+> Crystal language tool for compiling.
```
agnivade: We should make this consistent with other pages like `go`. Like - `Tool for managing crystal source code`

pages/common/crystal.md
aa-tan: Would `crystal init app my_blog` serve as a more concrete example?

pages/common/crystal.md
aa-tan: Fixed with b5de468

agnivade:
I updated application name myself. Made it consistent with other pages like `npm`, `dpkg`, `pacman` where we have package_name, module_name etc.

PR #2336, opened 2018-09-16
findstr: add page

pxgamer:
Part of #1579

----

### DIFF
--- a/pages/windows/findstr.md
+++ b/pages/windows/findstr.md
@@ -1,0 +1,35 @@
+# findstr
+
+> Find specified text within one or more files.
+
+- Find space-separated string(s) in all files:
+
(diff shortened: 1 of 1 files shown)


pages/windows/findstr.md
```diff
@@ -0,0 +1,35 @@
+# findstr
+
+> Find specified text within one or more files.
+
+- Find space-separated string(s) in all files:
+
+`findstr "{{query}}" *`
+
+- Find space-separated string(s) in all files recursively:
```
agnivade: Let us use the `[]` idiom here. Since the option is `/s` for recursively, whereas `/r` means regular expressions, I think we should help users remember the options better.

So let's use `recur[s]ively`, `e[x]actly`

PR #2337, opened 2018-09-17
rdfind: add page

derphilipp:
The tool is a available from https://rdfind.pauldreik.se/

----

### DIFF
--- a/pages/common/rdfind.md
+++ b/pages/common/rdfind.md
@@ -1,16 +1,16 @@
 # rdfind
 
-> finds duplicate files and gets rid of them
+> Finds duplicate files and gets rid of them.
 
 - Identify all duplicates in a given directory and output summary:
 
 `rdfind -n true {{path/to/directory}}`
 
-- Replace all duplicates with hardlinks
+- Replace all duplicates with hardlinks:
 
 `rdfind -makehardlinks true {{path/to/directory}}`
 
-- Replace all duplicates with symlinks/soft links
+- Replace all duplicates with symlinks/soft links:
 
 `rdfind -makesymlinks true {{path/to/directory}}`
 
--- a/pages/common/rdfind.md
+++ b/pages/common/rdfind.md
@@ -1,0 +1,19 @@
+# rdfind
+
+> finds duplicate files and gets rid of them
+
+- Identify all duplicates in a given directory and output summary:
+
+`rdfind -n true {{path/to/directory}}`
+
+- Replace all duplicates with hardlinks
+
+`rdfind -makehardlinks true {{path/to/directory}}`
+
+- Replace all duplicates with symlinks/soft links
+
+`rdfind -makesymlinks true {{path/to/directory}}`
+
+- Delete all duplicates, do not ignore empty files:
+
+`rdfind -deleteduplicates true -ignoreempty false {{/mnt/backup}}`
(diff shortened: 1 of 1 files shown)


pages/common/rdfind.md
```diff
@@ -0,0 +1,19 @@
+# rdfind
+
+> Finds duplicate files and gets rid of them.
+
+- Identify all duplicates in a given directory and output summary:
```
sbrl: Perhaps ...`output a summary` would read better?

pages/common/rdfind.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Replace all duplicates with hardlinks:
+
+`rdfind -makehardlinks true {{path/to/directory}}`
+
+- Replace all duplicates with symlinks/soft links:
+
+`rdfind -makesymlinks true {{path/to/directory}}`
+
+- Delete all duplicates, do not ignore empty files:
```
sbrl: Perhaps ...`duplicates, and do not ignore`.... would read better?

pages/common/rdfind.md
derphilipp: Good point! 👍

pages/common/rdfind.md
derphilipp: Valid point, i like that one better; Thank you

pages/common/rdfind.md
```diff
@@ -14,6 +14,6 @@
 
 `rdfind -makesymlinks true {{path/to/directory}}`
 
-- Delete all duplicates, do not ignore empty files:
+- Delete all duplicates, and do not ignore empty files:
```
mfrw: Drop the `,`

derphilipp: Done 😄

pages/common/rdfind.md
mfrw: I think the `,` is superfluous. We should probably drop it.

derphilipp: Agreed, removed it;

pages/common/rdfind.md
```diff
@@ -0,0 +1,19 @@
...
+
+`rdfind -makehardlinks true {{path/to/directory}}`
+
+- Replace all duplicates with symlinks/soft links:
+
+`rdfind -makesymlinks true {{path/to/directory}}`
+
+- Delete all duplicates and do not ignore empty files:
+
+`rdfind -deleteduplicates true -ignoreempty false {{/mnt/backup}}`
```
agnivade: Why is this suddenly `/mnt/backup` ?

pages/common/rdfind.md
```diff
@@ -0,0 +1,19 @@
+# rdfind
+
+> Finds duplicate files and gets rid of them.
```
agnivade: I think its clearer to mention - "Finds files with duplicate names". Duplicate files might mean files with the same content too.

derphilipp: But it does find files of the same content, not the same name; I try to clarify that

agnivade: Ah, I understood the opposite. More the reason we should fix this. I suggest - `Find files with duplicate content ..`

derphilipp: Added a pull request to resolve the issue:  https://github.com/tldr-pages/tldr/pull/2341

pages/common/rdfind.md
sbrl: Ooh good spot. I guess `path/to/backup` would have been better?

derphilipp: or just set it to a generic pathname like the rest;

PR #2343, opened 2018-09-20
tee: add `tee /dev/tty` mid-pipeline example

ticky:

### DIFF
--- a/pages/common/tee.md
+++ b/pages/common/tee.md
@@ -12,7 +12,7 @@
 
 - Print standard input to the terminal, and also pipe it into another program for further processing:
 
-`echo "example" | tee /dev/tty | xargs printf "\033[1;34m%s\033[m\n"`
+`echo "example" | tee {{/dev/tty}} | {{xargs printf "[%s]"}}`
 
 - Create a folder called "example", count the number of characters in "example" and write "example" to the terminal:
 
--- a/pages/common/tee.md
+++ b/pages/common/tee.md
@@ -10,6 +10,10 @@
 
 `echo "example" | tee -a {{FILE}}`
 
+- Print standard input to the terminal, and also pipe it into another program for further processing:
+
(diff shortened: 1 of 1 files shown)


pages/common/tee.md
```diff
@@ -10,6 +10,10 @@
 
 `echo "example" | tee -a {{FILE}}`
 
+- Print standard input to the terminal, and also pipe it into another program for further processing:
+
+`echo "example" | tee /dev/tty | xargs printf "\033[1;34m%s\033[m\n"`
```
agnivade: /dev/tty in tokens `{{}}` please. Same for printf. Also, no need to colorize the output and make the example complicated. We want the pages to be simple and readable. printf "[%s]" is fine.

ticky: Updated in c6bb56d651b0e78b966f1daf18c10010334f473f. I wrapped the entire last command in the tokens; I hope that’s correct, but I couldn’t quite figure out from the docs what the intended approach for tokens representing entire commands!

pages/common/tee.md
```diff
@@ -12,7 +12,7 @@
 
 - Print standard input to the terminal, and also pipe it into another program for further processing:
 
-`echo "example" | tee /dev/tty | xargs printf "\033[1;34m%s\033[m\n"`
+`echo "example" | tee {{/dev/tty}} | {{xargs printf "[%s]"}}`
```
mfrw: I think `echo "example" | tee | xargs printf "[%s]"` will also do the same.
What benefit do we have from including `/dev/tty`

ticky: That’s not correct:

image

As the description says, this both prints to your terminal *and* passes it to the next command. Tee with no args does nothing.

mfrw: I don't know what was I thinking :(
You are correct, I apologise.

PR #2345, opened 2018-09-21
cmd: add page

pxgamer:
Part of #1579

----

### DIFF
--- a/pages/windows/cmd.md
+++ b/pages/windows/cmd.md
@@ -6,7 +6,7 @@
 
 `cmd`
 
-- Run the specified command and terminate:
+- Run the specified command and then exit:
 
 `cmd /c "{{command}}"`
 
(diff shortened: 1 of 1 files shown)


pages/windows/cmd.md
```diff
@@ -0,0 +1,35 @@
+# cmd
+
+> The Windows command interpreter.
+
+- Start a new instance of the command interpreter:
+
+`cmd`
+
+- Run the specified command and terminate:
```
sbrl: Perhaps ...`and then exit` would read better?

My thinking is that the example below could be altered to says something like `and then enter an interactive shell` or something.

pxgamer: Sorry, been away this weekend. :+1: I think that definitely is more clear. :slight_smile:

pxgamer:
@sbrl @agnivade is this ok to be merged?

sbrl:
Yep! I thought that @agnivade was going to merge it, but I've just done so anyway :-)

PR #2346, opened 2018-09-21
php artisan: add page

ImLiam:
This adds a page for the command line interface for the popular PHP framework [Laravel](https://laravel.com/docs/master/artisan)

-----

### DIFF
--- a/pages/common/php-artisan.md
+++ b/pages/common/php-artisan.md
@@ -1,0 +1,19 @@
+# php artisan
+
+> Laravel's Artisan command line interface.
+
+- Display a list of all available commands:
+
+`php artisan help`
+
+- Start PHP's built-in web server for the current Laravel application:
+
+`php artisan serve`
+
+- Start an interactive PHP command line interface:
+
+`php artisan tinker`
+
+- Generate a new Eloquent model class with a migration, factory and resource controller:
+
+`php artisan make:model {{ModelName}} --all`
--- a/pages/common/php-artisan.md
+++ b/pages/common/php-artisan.md
@@ -2,10 +2,6 @@
 
 > Laravel's Artisan command line interface.
 
-- Display a list of all available commands:
-
-`php artisan help`
-
 - Start PHP's built-in web server for the current Laravel application:
 
 `php artisan serve`
@@ -17,3 +13,7 @@
 - Generate a new Eloquent model class with a migration, factory and resource controller:
 
 `php artisan make:model {{ModelName}} --all`
+
+- Display a list of all available commands:
+
+`php artisan help`


pages/common/php-artisan.md
```diff
@@ -0,0 +1,19 @@
+# php artisan
+
+> Laravel's Artisan command line interface.
+
+- Display a list of all available commands:
+
+`php artisan help`
```
sbrl: Wouldn't this get PHP to execute a file called `artisan` in the current directory?

I assume that a dash `-` was missed out here or something?

I'm not familiar with _Artisan_, so I could be wrong.

pages/common/php-artisan.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Start PHP's built-in web server for the current Laravel application:
+
+`php artisan serve`
+
+- Start an interactive PHP command line interface:
+
+`php artisan tinker`
+
+- Generate a new Eloquent model class with a migration, factory and resource controller:
```
sbrl: Is `Eloquent` meant to be capitalised?

pages/common/php-artisan.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Display a list of all available commands:
+
+`php artisan help`
+
+- Start PHP's built-in web server for the current Laravel application:
+
+`php artisan serve`
+
+- Start an interactive PHP command line interface:
```
sbrl: What's the different between this and `php -a`?

pages/common/php-artisan.md
ImLiam: Indeed, `artisan` is a file included with Laravel to be executed with PHP. I'm not entirely sure if the tldr project intends to cover things like this since it's not a global command, just one _very_ frequently used inside project directories by Laravel users.

I noticed there's some other command pages with spaces in (such as `git add`) but obviously, these are global. Is there any recommended way to handle this?

pages/common/php-artisan.md
ImLiam: Yes, [Eloquent](https://laravel.com/docs/master/eloquent) is a noun - it is the name of a module in the Laravel framework.

pages/common/php-artisan.md
ImLiam: Tinker is a REPL Laravel offers that gives an interactive interface with specific integrations (such as bootstrapping the framework), it is different to the `php -a` tool offered by PHP's native CLI.

pages/common/php-artisan.md
sbrl: Cool, thanks :-)

pages/common/php-artisan.md
sbrl: Ah, ok - thanks for clearing that up :-)

pages/common/php-artisan.md
sbrl: Ah, I see. It's definitely something we should cover - though I'm unsure as to whether the page name should be `php artisan` or just simply `artisan`.

I think `php artisan` should either be `php ./artisan` or `./artisan` personally. @agnivade, what do you think?

ImLiam: Perhaps one of these would be better, but some comments:

- `./artisan` would not be appropriate, as Windows wouldn't know what to execute the file as since it doesn't support the shebang.
- All the documentation, guides, tutorials, etc. around the framework refers to it as `php artisan` consistently, I'm fairly certain that anyone looking up the tl;dr of this command will understand that.
- If it is named `php ./artisan` would people still be able to access this page as `tldr php artisan`?

sbrl: If the official documentation calls it `php artisan`, then let's stick with that :-)

pages/common/php-artisan.md
```diff
@@ -0,0 +1,19 @@
+# php artisan
+
+> Laravel's Artisan command line interface.
+
+- Display a list of all available commands:
```
agnivade: I think this should be the last example.

sbrl:
Is this ready for merging?

pxgamer:
Think so @sbrl :+1: I wasn't sure if we'd wait for confirmation from the PR creator.

ImLiam:
Ah sorry - I never even saw this request for change! The change makes sense to me 👍

pxgamer:
Thanks @ImLiam, this seems good to merge then. :+1:

PR #2350, opened 2018-09-23
GOVERNANCE: number principles to allow x-refs

waldyrious:
This is a simple change that doesn't affect the actual content in any way; it only changes the presentation, by using a numbered list (rather than a bulleted one) to make it possible to refer directly to specific governance rules.


sbrl:
Thanks, @waldyrious :smiley_cat:

waldyrious:
My pleasure, as always ^^

PR #2361, opened 2018-10-01
apt: added to description.

thenarcissist:
Added that apt should be preferred over apt-get in Ubuntu systems which are 16.04 and later.

----

- [✓] The page has 8 or fewer examples.

- [✓] The PR is appropriately titled:  

- [✓] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines.

### DIFF
--- a/pages/linux/apt.md
+++ b/pages/linux/apt.md
@@ -1,6 +1,7 @@
 # apt
 
 > Package management utility for Debian based distributions.
+> Recommended replacement for apt-get in Ubuntu versions 16.04 and later.
 
 - Update the list of available packages and versions (it's recommended to run this before other `apt` commands):
 


pages/linux/apt.md
```diff
@@ -1,6 +1,7 @@
 # apt
 
 > Package management utility for Debian based distributions.
+> Recommended replacement for apt-get in Ubuntu versions 16.04 and later.
```
agnivade: I think it's more useful in the `apt-get` page. Because people searching for `apt` are already likely to use the `apt` command. So its more helpful to redirect them from `apt-get` to `apt`. But `apt-get` already has 2 lines. So I doubt whether we will be able to fit this in somewhere.

thenarcissist: The idea behind putting that up was that I had people asking me about the difference.

agnivade: Sure, in that case, they can just do `tldr apt-get` and know the difference.

MasterOdin:
I feel it should be slightly noted that apt is only recommended for interactive use and not for scripts, where apt-get is preferred.

thenarcissist:
I was of the thought that the people who would use scripts would know which
one to prefer.

tonycavella:
Unsubscribe

agnivade:
> I feel it should be slightly noted that apt is only recommended for interactive use and not for scripts, where apt-get is preferred.

Ah, did not know about that. But I don't think we can fit all that information in a tldr page.

pages/linux/apt.md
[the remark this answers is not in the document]
agnivade: Sounds good

sbrl:
Hrm, it looks like it's not letting me alter this PR. I'm going to have to close it and open a new one. I'll preserve the commit made on this branch though for authorship information.

sbrl:
Hrm, it looks like it's not letting me alter this PR. I'm going to have to close it and open a new one. I'll preserve the commit made on this branch though for authorship information.

I've opened #3264.

PR #2364, opened 2018-10-01
erl: add page

mira01:
----

### DIFF
--- a/pages/common/erl.md
+++ b/pages/common/erl.md
@@ -1,0 +1,15 @@
+# erl
+
+> Run and manage programs in the Erlang programming language.
+
+- Compile and run sequential Erlang program as a common script and then exit:
+
+`erlc {{files}} && erl -noshell '{{mymodule:myfunction(arguments)}}, init:stop().'`
+
+- Connect to a running Erlang node:
+
+`erl -remsh {{nodename}}@{{hostname}} -sname {{custom_shortname}} -hidden -setcookie {{cookie_of_remote_node}}`
+
+- Tell the Erlang shell to load modules from a directory:
+
+`erl -pa {{directory_with_beam_files}}`


pages/common/erl.md
```diff
@@ -0,0 +1,15 @@
+# erl
+
+> Run and manage programs written in Erlang programming language.
```
sbrl: Perhaps ...`in the Erland programming`... would read better?

pages/common/erl.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Run sequential erlang program as a common script and then exit:
```
sbrl: Should `erlang` be capitalised here, or should it be lowercase in the description?

pages/common/erl.md
```diff
@@ -0,0 +1,15 @@
...
+
+`erlc *.erl && erl -noshell '{{mymodule:myfunction(arguments)}}, init:stop().'`
+
+- Connect to a running erlang node:
+
+`erl -remsh {{nodename}}@{{hostname}} -sname {{custom_shortname}} -hidden -setcookie {{cookie_of_remote_node}}`
+
+- Tell the erlang shell to load another modules or other versions:
```
sbrl: Perhaps ....`another module or version:` would read better?

pages/common/erl.md
mira01: Ok, thanks

pages/common/erl.md
mira01: I think capitalising here is better, going to fix it

pages/common/erl.md
mira01: Going to reword to "another modules or version". Is it ok?

pages/common/erl.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Run sequential Erlang program as a common script and then exit:
+
+`erlc *.erl && erl -noshell '{{mymodule:myfunction(arguments)}}, init:stop().'`
+
+- Connect to a running Erlang node:
+
+`erl -remsh {{nodename}}@{{hostname}} -sname {{custom_shortname}} -hidden -setcookie {{cookie_of_remote_node}}`
+
+- Tell the Erlang shell to load another modules or version:
```
sbrl: Not sure that `modules` needs to be plural here.

mira01: I believe that yes as there is given a path to the directory where multiple files could be found

sbrl: I see. In that case, we'll need to rework it a bit. How about ....`to load modules from a directory:`?

mira01: Thanks, that sounds good.

pages/common/erl.md
```diff
@@ -0,0 +1,15 @@
...
+
+`erl -pa {{directory_with_beam_files}}`
```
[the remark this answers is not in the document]
mira01: Unfortunately it is not, as in erlang there are three directories the user typically interacts with: top level directory, directory with source code and directory with compiled modules (beam files). These directories could, but do not neccessarily be the same.

pages/common/erl.md
```diff
@@ -0,0 +1,15 @@
+# erl
+
+> Run and manage programs in Erlang programming language.
+
+- Run sequential Erlang program as a common script and then exit:
+
+`erlc *.erl && erl -noshell '{{mymodule:myfunction(arguments)}}, init:stop().'`
```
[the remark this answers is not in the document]
mira01: Thanks, changing to {{files}}

agnivade:
Unless you have any comments, then yes. :)

sbrl:
Merging due to inactivity.

PR #2365, opened 2018-10-01
elm: add page

mira01:
----

### DIFF
--- a/pages/common/elm.md
+++ b/pages/common/elm.md
@@ -2,7 +2,7 @@
 
 > Run and manage programs in Elm programming language.
 
-- Start an interactive Elm shell (repl):
+- Start an interactive Elm shell (REPL):
 
 `elm --repl`
 
--- a/pages/common/elm.md
+++ b/pages/common/elm.md
@@ -1,0 +1,15 @@
+# elm
+
+> Run and manage programs in Elm programming language.
+
+- Start an interactive Elm shell (repl):
+
+`elm --repl`
+
+- Initialize an empty Elm project:
+
+`elm --init`
+
+- Compile the current project and serve it via the local web server:
+
+`elm --reactor`


pages/common/elm.md
```diff
@@ -0,0 +1,15 @@
+# elm
+
+> Run and manage programs in Elm programming language.
+
+- Run interactive elm shell (repl):
```
sbrl: Should `elm` be capitalised here, or should it be lowercase in the description?

pages/common/elm.md
```diff
@@ -0,0 +1,15 @@
...
+
+`elm --repl`
+
+- Initialize empty elm project:
```
sbrl: Perhaps `Initialize an empty`... would read better?

pages/common/elm.md
```diff
@@ -0,0 +1,15 @@
...
+
+`elm --init`
+
+- Compile your project and serve it via local web server for easy development:
```
sbrl: Perhaps `Compile the current project and serve it via the local web server for easy development:` would read better?

mira01: Thanks, going to fix it

pages/common/elm.md
mira01: Thanks, going to fix it

pages/common/elm.md
mira01: Thanks, I think capitalising here is better

pages/common/elm.md
```diff
@@ -0,0 +1,15 @@
+# elm
+
+> Run and manage programs in Elm programming language.
+
+- Run interactive Elm shell (repl):
```
agnivade: "Start an" instead of "Run" sounds better. Although, looking at the `node`, `python` and `ruby` pages, all have slightly different styles.

pages/common/elm.md
```diff
@@ -0,0 +1,15 @@
...
+
+`elm --repl`
+
+- Initialize an empty Elm project:
+
+`elm --init`
+
+- Compile the current project and serve it via the local web server for easy development:
```
agnivade: We can drop the "for easy development".

pages/common/elm.md
mira01: Ok, I will change it

pages/common/elm.md
mira01: Ok, thanks

Issue #2333, opened 2018-09-13
page request: xsv


mfrw:
Add documentation for [xsv](https://github.com/BurntSushi/xsv).
A tool for manipulating csv files.

sbrl:
Hey, @szepnapot! @jedahan already has a pull request open for this command. However, you're welcome to tackle any other command you like :smile:

PR #2366, opened 2018-10-01
xsv: add page

jedahan:

---

A first crack at #2333

- Should this use `{{path/to/worldcities.csv}}` instead of `{{worldcities.csv}}`?
- Should I describe that worldcities.csv can be downloaded with `curl -LO http://burntsushi.net/stuff/worldcitiespop.csv`?

### DIFF
--- a/pages/common/xsv.md
+++ b/pages/common/xsv.md
@@ -1,0 +1,27 @@
+# xsv
+
+> A fast CSV command line toolkit written in Rust.
+
+- Inspect the headers of a file:
+
+`xsv headers {{worldcities.csv}}`
+
+- Count the number of entries:
+
+`xsv count {{worldcities.csv}}`
+
+- Get an overview of the shape of entries:
+
+`xsv stats {{worldcities.csv}} | xsv table`
+
+- Select a few columns:
+
+`xsv select {{Country,AccentCity}} {{worldcities.csv}}`
+
+- Show 10 random sample of entries:
+
+`xsv sample {{10}} {{worldcities.csv}}`
+
+- Join a column from one file to another:
+
+`xsv join --no-case {{Country}} {{worldcities.csv}} {{Abbrev}} {{countrynames.csv}} | xsv table`
--- a/pages/common/xsv.md
+++ b/pages/common/xsv.md
@@ -4,24 +4,24 @@
 
 - Inspect the headers of a file:
 
-`xsv headers {{worldcities.csv}}`
+`xsv headers {{path/to/worldcities.csv}}`
 
 - Count the number of entries:
 
-`xsv count {{worldcities.csv}}`
+`xsv count {{path/to/worldcities.csv}}`
 
 - Get an overview of the shape of entries:
 
-`xsv stats {{worldcities.csv}} | xsv table`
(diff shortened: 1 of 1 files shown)


mfrw:
> * Should this use `{{path/to/worldcities.csv}}` instead of `{{worldcities.csv}}`?

`{{wordcities.csv}}` should suffice.
> * Should I describe that worldcities.csv can be downloaded with `curl -LO http://burntsushi.net/stuff/worldcitiespop.csv`?

I don't think we generally provide this information but in this case, we could, for coherency.
\cc @agnivade @sbrl @pxgamer

sbrl:
> * Should this use `{{path/to/worldcities.csv}}` instead of `{{worldcities.csv}}`?

We do prefer `{{path/to/worldcities.csv}}`, if I recall correctly, yeah.

pages/common/xsv.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Get an overview of the shape of entries:
+
+`xsv stats {{worldcities.csv}} | xsv table`
+
+- Select a few columns:
+
+`xsv select {{Country,AccentCity}} {{worldcities.csv}}`
+
+- Show 10 random sample of entries:
```
sbrl: Perhaps `Show a 10 random entries:` would read better:

pages/common/xsv.md
```diff
@@ -0,0 +1,27 @@
+# xsv
+
+> A fast CSV command line toolkit written in Rust.
+
+- Inspect the headers of a file:
+
+`xsv headers {{path/to/worldcities.csv}}`
```
agnivade: worldcities sounds odd. We generally use path/to/file.csv. Can we use that everywhere ?

pages/common/xsv.md
```diff
@@ -0,0 +1,27 @@
...
+
+`xsv count {{path/to/worldcities.csv}}`
+
+- Get an overview of the shape of entries:
+
+`xsv stats {{path/to/worldcities.csv}} | xsv table`
+
+- Select a few columns:
+
+`xsv select {{Country,AccentCity}} {{path/to/worldcities.csv}}`
```
agnivade: You can change it to column_1, column_2

pages/common/xsv.md
```diff
@@ -0,0 +1,27 @@
...
+
+`xsv select {{column_a,column_b}} {{path/to/file.csv}}`
+
+- Show 10 random entries:
+
+`xsv sample {{10}} {{path/to/file.csv}}`
+
+- Join a column from one file to another:
+
+`xsv join --no-case {{key_a}} {{path/to/file/a.csv}} {{column_b}} {{path/to/file/b.csv}} | xsv table`
```
agnivade: Question: Why is the first one called key_ and the second one column_ ? Shouldn't both be "column_x"

jedahan: I was trying to hint at what join does but maybe its more confusing than not? will switch

agnivade: If it is like a sql-join where values of one column should be equal to values of another column, then yes I think column_a and column_b sounds better. I was curious to know why you mentioned "key".

sbrl:
Thanks for sticking with the review process, @jedahan :smiley_cat:

PR #2370, opened 2018-10-01
update license year

frazjp65:

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 ## The MIT License (MIT)
 
-Copyright (c) 2014 the [TLDR team](https://github.com/orgs/tldr-pages/people) and [contributors](https://github.com/tldr-pages/tldr/graphs/contributors)
+Copyright (c) 2018 the [TLDR team](https://github.com/orgs/tldr-pages/people) and [contributors](https://github.com/tldr-pages/tldr/graphs/contributors)
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


agnivade:
I don't think this is meant to be updated. The year denotes when the license file was created.

@waldyrious

waldyrious:
There are [various approaches](https://ben.balter.com/2015/06/03/copyright-notices-for-websites-and-open-source-projects/#the-date), the two most common being to use the creation year, and to use a range from the creation year to the current year. Although the argumentation for a range format (2014-2018) makes a lot of sense, I personally prefer using only the creation year, simply because it would take manual editing every year to keep the range from growing out-of-sync with the actual repo contents.

That said, I'd welcome an integration that automatically kept this up to date (similar to, say, what [greenkeeper](https://greenkeeper.io/) does for dependencies). Do you know of any such tools, @frazjp65?

agnivade:
In all of the open source projects, I have always seen just the creation date. A tool is fine, but there is additional overhead to maintain the tool. I don't see a good ROI on using a tool just for this.

Just the creation date seems fine to me.

waldyrious:
If it's a third-party integration that we'd just have to turn on, I don't see that as much of a burden. Of course, if it's something we'd have to manage more closely (running our own instance, managing login credentials, etc) then I agree the cost-benefit wouldn't look as good.

In any case, doing that or not would be orthogonal to this PR, so I guess we can close it. @frazjp65 let us know if you disagree -- we are open to reconsider, as always :)

PR #2378, opened 2018-10-02
slapt-get: add page

alinepegas:
----

### DIFF
--- a/pages/linux/slapt-get.md
+++ b/pages/linux/slapt-get.md
@@ -1,28 +1,28 @@
 # slapt-get
 
 > An apt like system for Slackware package management.
-> Package sources need to be configured on the slapt-getrc file.
+> Package sources need to be configured in the slapt-getrc file.
 
-- Update the list of available packages and versions (it's recommended to run this before other `slapt-get` commands):
+- Update the list of available packages and versions:
 
 `slapt-get --update`
 
 - Install a package, or update it to the latest available version:
 
-`slapt-get --install {{package}}`
+`slapt-get --install {{package_name}}`
 
 - Remove a package:
 
-`slapt-get --remove {{package}}`
+`slapt-get --remove {{package_name}}`
 
-- Upgrade all installed packages to their newest available versions:
+- Upgrade all installed packages to their latest available versions:
 
-`slapt-get --upgrade {{package}}`
+`slapt-get --upgrade {{package_name}}`
 
 - Locate packages of interest by the package name, disk set, or version:
 
-`slapt-get --search {{package}}`
+`slapt-get --search {{package_name}}`
 
-- Show information about packages:
+- Show information about a package:
 
-`slapt-get --show {{package}}`
+`slapt-get --show {{package_name}}`
--- a/pages/linux/slapt-get.md
+++ b/pages/linux/slapt-get.md
@@ -1,0 +1,28 @@
+# slapt-get
+
+> An apt like system for Slackware package management.
+> Package sources need to be configured on the slapt-getrc file.
+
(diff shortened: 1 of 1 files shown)


pages/linux/slapt-get.md
```diff
@@ -0,0 +1,28 @@
+# slapt-get
+
+> An apt like system for Slackware package management.
+> Package sources need to be configured on the slapt-getrc file.
```
sbrl: `on` -> `in`?

pages/linux/slapt-get.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Update the list of available packages and versions (it's recommended to run this before other `slapt-get` commands):
```
sbrl: Not sure that the `it's recommended` bit in brackets is needed. Generally, we order things in a tldr-page in the order that one would execute them. Additionally, it's rather similar to the `apt update` paradigm that's used in multiple other package managers.

What do you think?

pages/linux/slapt-get.md
```diff
@@ -0,0 +1,28 @@
...
+
+`slapt-get --update`
+
+- Install a package, or update it to the latest available version:
+
+`slapt-get --install {{package}}`
```
sbrl: Perhaps `package_name` would be more descriptive?

pages/linux/slapt-get.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Remove a package:
+
+`slapt-get --remove {{package}}`
+
+- Upgrade all installed packages to their newest available versions:
```
sbrl: Perhaps ...`to their latest versions` would be more consistent?

pages/linux/slapt-get.md
```diff
@@ -0,0 +1,28 @@
...
+
+`slapt-get --upgrade {{package}}`
+
+- Locate packages of interest by the package name, disk set, or version:
```
sbrl: What's a _disk set_? Not a _Slackware_ user, so I'm unsure :P

pages/linux/slapt-get.md
```diff
@@ -0,0 +1,28 @@
...
+
+`slapt-get --search {{package}}`
+
+- Show information about packages:
```
sbrl: Perhaps ...`about a package` would read better?

pages/linux/slapt-get.md
alinepegas: Sure

pages/linux/slapt-get.md
alinepegas: Well, I agree that it's unnecessary. I put it there to mimic what was done on the `apt-get.md` page. Should I keep it or remove it?

pages/linux/slapt-get.md
alinepegas: Yes, I'll make that change.

pages/linux/slapt-get.md
alinepegas: I agree, I will make the change.

pages/linux/slapt-get.md
alinepegas: Slackware used to be distribuited as a set of iso images (multiple CDs) where each package category (packages for gnome, packages for kde, packages for developers)  were in a different disk. Now there is only one ISO, but the package categories remain and are still called disk sets.

pages/linux/slapt-get.md
alinepegas: Yes, I agree. I will make the change.

pages/linux/slapt-get.md
sbrl: Ah, I see :-)

PR #2384, opened 2018-10-03
groups: add page

supermamie:
Updates #2213.

----

### DIFF
--- a/pages/linux/groups.md
+++ b/pages/linux/groups.md
@@ -1,0 +1,11 @@
+# groups
+
+> Print the groups a user is in
+
+- Display for current user:
+
+`groups`
+
+- Display for another user:
+
+`groups {{username}}`
--- a/pages/linux/groups.md
+++ b/pages/linux/groups.md
@@ -1,11 +1,15 @@
 # groups
 
-> Print the groups a user is in.
+> Print group memberships for a user.
 
-- Display for current user:
+- Print group memberships for the current user:
 
 `groups`
 
-- Display for another user:
+- Print group memberships for a specific user:
 
 `groups {{username}}`
+
+- Print group memberships for a list of users:
+
+`groups {{username1}} {{username2}} {{username3}}`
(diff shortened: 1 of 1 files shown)


pages/linux/groups.md
```diff
@@ -0,0 +1,11 @@
+# groups
+
+> Print the groups a user is in.
```
sbrl: Perhaps ....`groups that a user`.... would read better?

pages/linux/groups.md
```diff
@@ -0,0 +1,11 @@
...
+
+- Display for current user:
```
sbrl: Perhaps `Display the groups of the current user` would read better?

pages/linux/groups.md
```diff
@@ -0,0 +1,11 @@
...
+
+`groups`
+
+- Display for another user:
```
sbrl: Perhaps `Display the groups for`..... would read better?

pages/linux/groups.md
supermamie: Do you think "Print group memberships for a user" would be better ?

pages/linux/groups.md
supermamie: Do you think "Print group memberships for the current user" would be better ?

pages/linux/groups.md
supermamie: Do you think "Print group memberships for users" would be better ?
(I'll also modify the doc and add an example with a list of users)

pages/linux/groups.md
sbrl: Go for it!

pages/linux/groups.md
sbrl: Erm not sure here. That implies multiple users - we've got a single user here.

sbrl: Perhaps `Print group memberships for a specific user` might work instead?

PR #2385, opened 2018-10-03
hyperfine: update basic example to include the minimum number of runs

russsaidwords:
----


agnivade:
We should not add inconsistent examples within a page. The rest of the page uses `make` examples, but suddenly this one uses cat and bat. We should have used different make targets to do a comparison.

I will send a PR.

PR #2386, opened 2018-10-03
numfmt: add page

russsaidwords:
Part of #2213

----

### DIFF
--- a/pages/linux/numfmt.md
+++ b/pages/linux/numfmt.md
@@ -1,0 +1,15 @@
+# numfmt
+
+> Convert numbers from/to human-readable strings.
+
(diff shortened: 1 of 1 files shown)


pages/linux/numfmt.md
```diff
@@ -0,0 +1,15 @@
+# numfmt
+
+> Convert numbers from/to human-readable strings.
```
mfrw: Perhaps `to or from` reads better.

sbrl: Or `to and from`?

russsaidwords: That change is all set, I went with "to and from" because that has the right "sound" to my ears. Thanks all.

PR #2391, opened 2018-10-04
luac: add page

matusf:
----

### DIFF
--- a/pages/common/luac.md
+++ b/pages/common/luac.md
@@ -1,11 +1,11 @@
 # luac
 
-> Lua byte code compiler.
+> Lua bytecode compiler.
 
-- Compile a lua source code to a lua byte code:
+- Compile Lua source file to Lua bytecode:
 
 `luac -o {{byte_code.luac}} {{source.lua}}`
 
-- Don't include a debug information in the byte code:
+- Don't include a debug information in the output:
 
 `luac -s -o {{byte_code.luac}} {{source.lua}}`
--- a/pages/common/luac.md
+++ b/pages/common/luac.md
@@ -2,7 +2,7 @@
 
 > Lua bytecode compiler.
 
-- Compile Lua source file to Lua bytecode:
+- Compile a Lua source file to Lua bytecode:
 
 `luac -o {{byte_code.luac}} {{source.lua}}`
 
--- a/pages/common/luac.md
+++ b/pages/common/luac.md
@@ -6,6 +6,6 @@
 
 `luac -o {{byte_code.luac}} {{source.lua}}`
 
-- Don't include a debug information in the output:
+- Don't include debug information in the output:
 
 `luac -s -o {{byte_code.luac}} {{source.lua}}`
--- a/pages/common/luac.md
+++ b/pages/common/luac.md
@@ -2,10 +2,10 @@
 
 > Lua byte code compiler.
 
-- Compile lua source code to lua byte code:
+- Compile a lua source code to a lua byte code:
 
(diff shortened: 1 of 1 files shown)


pages/common/luac.md
```diff
@@ -0,0 +1,11 @@
+# luac
+
+> Lua byte code compiler.
+
+- Compile lua source code to lua byte code:
+
+`luac -o {{byte_code.luac}} {{source.lua}}`
+
+- Stop including debugging information in byte code:
```
stellarhopper: It probably reads better if you say "Strip debug information from the output" or "Don't include debug information in the byte code".  "Stop including.." implies some sort of a lasting effect..

matusf: I'd go for "Don't include debug information in the byte code"

pages/common/luac.md
```diff
@@ -0,0 +1,11 @@
+# luac
+
+> Lua byte code compiler.
```
sbrl: Not sure that there should be a space in `bytecode`.

pages/common/luac.md
```diff
@@ -0,0 +1,11 @@
...
+
+- Compile a lua source code to a lua byte code:
```
sbrl: Perhaps ...`Lua source file to Lua bytecode would read better?

pages/common/luac.md
```diff
@@ -0,0 +1,11 @@
...
+
+`luac -o {{byte_code.luac}} {{source.lua}}`
+
+- Don't include a debug information in the byte code:
```
sbrl: Perhaps `Don't include debug information in the output` would read better?

pages/common/luac.md
matusf: My spellchecker underlined it but on wiki it says bytecode. I'll change it

pages/common/luac.md
matusf: I agree. I'll change it.

pages/common/luac.md
matusf: Ok, but the article should probably stay there. So that it's grammatically correct. What do you think?

stellarhopper: 'information' is not a countable noun, so "a debug information" is incorrect.

pages/common/luac.md
```diff
@@ -1,11 +1,11 @@
...
-> Lua byte code compiler.
+> Lua bytecode compiler.
 
-- Compile a lua source code to a lua byte code:
+- Compile Lua source file to Lua bytecode:
 
 `luac -o {{byte_code.luac}} {{source.lua}}`
 
-- Don't include a debug information in the byte code:
+- Don't include a debug information in the output:
```
stellarhopper: 'information' is not a countable noun, so "a debug information" is incorrect.

pages/common/luac.md
```diff
@@ -0,0 +1,11 @@
+# luac
+
+> Lua bytecode compiler.
+
+- Compile Lua source file to Lua bytecode:
```
sbrl: Oops! Think I missed something here. Perhaps `Compile a Lua`.... would read better?

pages/common/luac.md
```diff
@@ -0,0 +1,11 @@
+# luac
+
+> Lua bytecode compiler.
+
+- Compile a Lua source file to Lua bytecode:
+
+`luac -o {{byte_code.luac}} {{source.lua}}`
+
+- Don't include debug information in the output:
```
agnivade: How about "Do not include debug symbols in the output:" ? So that we match the flag `-s` with a word in the description.

PR #2399, opened 2018-10-05
logname: add page

ckauf:

### DIFF
--- a/pages/linux/logname.md
+++ b/pages/linux/logname.md
@@ -1,0 +1,7 @@
+# logname 
+
+> Shows the users login name.
+
+- Display login user name:
+
+`logname`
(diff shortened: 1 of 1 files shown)


ckauf:
This is for issue #2213

pages/linux/logname.md
```diff
@@ -0,0 +1,7 @@
+# logname 
+
+> Shows the users login name.
```
sbrl: Perhaps `user's` instead of `users` would read better?

pages/linux/logname.md
```diff
@@ -0,0 +1,7 @@
...
+
+- Display login user name:
```
sbrl: Perhaps `Display the currently logged in user's name` would read better?

pages/linux/logname.md
ckauf: Yup. Done that.

pages/linux/logname.md
ckauf: Yup. Done that.

Issue #2393, opened 2018-10-04
page request: vncviewer


LachlanMarnham:
docs: https://tigervnc.org/doc/vncviewer.html
repo (with an 'about' in the README): https://github.com/TigerVNC/tigervnc

I'll take the PR if that's ok?

mfrw:
@LachlanMarnham please go right ahead :)

PR #2401, opened 2018-10-05
vncviewer: add page

LachlanMarnham:
----

issue: #2393

### DIFF
--- a/pages/linux/vncviewer.md
+++ b/pages/linux/vncviewer.md
@@ -10,7 +10,7 @@
 
 `vncviewer -FullScreen {{host}}:{{display_number}}`
 
-- Launch a VNC client with specific screen geometry:
+- Launch a VNC client with a specific screen geometry:
 
 `vncviewer --geometry {{width}}x{{height}} {{host}}:{{display_number}}`
 
--- a/pages/linux/vncviewer.md
+++ b/pages/linux/vncviewer.md
@@ -1,0 +1,19 @@
+# vncviewer
+
+> Launches a VNC (Virtual Network Computing) client.
+
+- Launch a VNC client which connects to a host on a given display:
+
+`vncviewer {{host}}:{{display_number}}`
+
+- Launch in full-screen mode:
+
+`vncviewer -FullScreen {{host}}:{{display_number}}`
+
+- Launch a VNC client with specific screen geometry:
+
+`vncviewer --geometry {{width}}x{{height}} {{host}}:{{display_number}}`
+
+- Launch a VNC client which connects to a host on a given port:
+
+`vncviewer {{host}}::{{port}}`


pages/linux/vncviewer.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Launch a VNC client which connects to a host on a given display:
+
+`vncviewer {{host}}:{{display_number}}`
+
+- Launch in full-screen mode:
+
+`vncviewer -FullScreen {{host}}:{{display_number}}`
+
+- Launch a VNC client with specific screen geometry:
```
sbrl: Perhaps .....`with a specific`..... would read better?

pages/linux/vncviewer.md
```diff
@@ -0,0 +1,19 @@
+# vncviewer
+
+> Launches a VNC (Virtual Network Computing) client.
+
+- Launch a VNC client which connects to a host on a given display:
+
+`vncviewer {{host}}:{{display_number}}`
```
agnivade: From where does one get the display_number ?

pages/linux/vncviewer.md
```diff
@@ -0,0 +1,19 @@
...
+
+`vncviewer -FullScreen {{host}}:{{display_number}}`
+
+- Launch a VNC client with specific screen geometry:
+
+`vncviewer --geometry {{width}}x{{height}} {{host}}:{{display_number}}`
+
+- Launch a VNC client which connects to a host on a given port:
+
+`vncviewer {{host}}::{{port}}`
```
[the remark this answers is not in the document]
LachlanMarnham: @agnivade Intentional. If one types host:number, then number is parsed as a display number. If one types host::number, number is parsed as a port.

pages/linux/vncviewer.md
LachlanMarnham: @sbrl yes I think so, I will make the change.

pages/linux/vncviewer.md
LachlanMarnham: @agnivade It is the display that the VNC server is running on. I erred away from explaining this concept due to the line in the contributing guidelines "...and avoid explaining general UNIX concepts...". I also decided to keep this consistent with vncserver.md, which had been submitted by someone else already, and which refers to the display_number without definition. ie:

- Kill an instance of VNC Server running on a specific display:
`vncserver --kill :{{display_number}}`

Perhaps I should reword to "Launch a VNC client which connects to a host which is serving on a given display", to make it clear? Any ideas?

agnivade: Yes, I saw the vncserver.md page. Now that I think a bit more, I think this should be fine. I initially thought this to be some vnc specific concept. But if it is a general UNIX concept, we can let it be.

Thanks for the clarification.

PR #2409, opened 2018-10-06
lsusb: add page

f4ww4z:
----

### DIFF
--- a/pages/linux/lsusb.md
+++ b/pages/linux/lsusb.md
@@ -1,0 +1,23 @@
+# lsusb
+
+> Display info about connected USB devices.
+
+- List all the USB ports available:
+
+`lsusb`
+
+- Hierarchy view, with each device's supported speed:
+
+`lsusb -t`
+
+- Get detailed info of a USB device:
+
+`lsusb -D {{device}}`
+
+- Show only devices with the specified vendor and product ID:
+
+`lsusb -d {{vendor}}:{{product}}`
+
+- List all mass storages:
+
+`lsusb -v | grep -Ei '(idVendor|Mass Storage)'`
--- a/pages/linux/lsusb.md
+++ b/pages/linux/lsusb.md
@@ -10,6 +10,10 @@
 
 `lsusb -t`
 
+- List verbose information about USB devices:
+
+`lsusb --verbose`
+
 - List detailed information about a USB device:
 
 `lsusb -D {{device}}`
@@ -17,7 +21,3 @@
 - List devices with a specified vendor and product id only:
 
 `lsusb -d {{vendor}}:{{product}}`
-
-- List verbose information about connected mass storage devices:
-
-`lsusb -v | grep -Ei '(idVendor|Mass Storage)'`
--- a/pages/linux/lsusb.md
+++ b/pages/linux/lsusb.md
@@ -1,23 +1,23 @@
 # lsusb
 
-> Display info about connected USB devices.
+> Display information about USB buses and devices connected to them.
 
-- List all the USB ports available:
+- List all the USB devices available:
 
 `lsusb`
 
-- Hierarchy view, with each device's supported speed:
+- List the USB hierarchy as a tree:
 
 `lsusb -t`
 
-- Get detailed info of a USB device:
+- List detailed information about a USB device:
 
 `lsusb -D {{device}}`
 
-- Show only devices with the specified vendor and product ID:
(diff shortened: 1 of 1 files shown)


pages/linux/lsusb.md
```diff
@@ -0,0 +1,23 @@
+# lsusb
+
+> Display info about connected USB devices.
+
+- List all the USB ports available:
```
mfrw: I think renaming `ports` to `devices` would be more correct.

pages/linux/lsusb.md
```diff
@@ -0,0 +1,23 @@
...
+
+`lsusb`
+
+- Hierarchy view, with each device's supported speed:
```
mfrw: Perhaps rephrasing to  `List the USB hierarchy as a tree` sounds a little better.

pages/linux/lsusb.md
```diff
@@ -0,0 +1,23 @@
...
+
+`lsusb -t`
+
+- Get detailed info of a USB device:
```
mfrw: How about `List detailed information about a USB device`

pages/linux/lsusb.md
```diff
@@ -0,0 +1,23 @@
...
+
+`lsusb -D {{device}}`
+
+- Show only devices with the specified vendor and product ID:
```
mfrw: Again just to maintain the coherence of the page,
`List devices with a specified vendor and product id only`

pages/linux/lsusb.md
```diff
@@ -0,0 +1,23 @@
...
+
+`lsusb -d {{vendor}}:{{product}}`
+
+- List all mass storages:
```
mfrw: Perhaps renaming this example to `List verbose information about USB devices`
also, change the example to only `lsusb -v`.

mfrw:
@f4ww4z Thank you for the PR.
I have some review comments for you to address.

pages/linux/lsusb.md
f4ww4z: Thanks for the review @mfrw .

How about `List verbose information about connected mass storage devices`? Since I want to make the examples gradually more complex, and also I think 'verbose' is a common option that users may know already.

sbrl: Sounds good to me, @f4ww4z! Let's do that.

f4ww4z: Thanks @sbrl , I have updated the description in my new commit.

mfrw: @f4ww4z instead of mass storage, use USB devices, as this tool is not limited to storage.

pages/linux/lsusb.md
```diff
@@ -1,23 +1,23 @@
...
 
 `lsusb -D {{device}}`
 
-- Show only devices with the specified vendor and product ID:
+- List devices with a specified vendor and product id only:
 
 `lsusb -d {{vendor}}:{{product}}`
 
-- List all mass storages:
+- List verbose information about connected mass storage devices:
```
mfrw: Perhaps `List verbose information about USB devices`

We should not  mention mass storage devices as the tool gives info about the usb root-hub and any other usb device.

pages/linux/lsusb.md
sbrl: @mfrw Indeed it isn't. I think what @f4ww4z is getting at is that they'd like to show how you _could_ limit it to mass storage devices. The other examples above show that you can list all other types of devices too :-)

pages/linux/lsusb.md
f4ww4z: Thanks, @mfrw. I've updated the desc like you said, and moved it near the top so that the examples are more complex going down to the bottom.

pages/linux/lsusb.md
```diff
@@ -0,0 +1,23 @@
...
+
+- List verbose information about USB devices:
+
+`lsusb --verbose`
+
+- List detailed information about a USB device:
+
+`lsusb -D {{device}}`
+
+- List devices with a specified vendor and product id only:
```
agnivade: I don't think you can have multiple devices with the same vendor and product id. In that case, this should be "List a device with a .."

mfrw: I think It will be the same, but I can understand what @agnivade is referring to.

agnivade: Ah that's right. Yes, then I think it is possible to have multiple devices with same vendor and product id.

PR #2419, opened 2018-10-09
whoami: clarify description and add sudo example

### LINKED ISSUE #2418: question: whoami
Hello,
I noticed that logname was added to tldr and was immediately intrigued by the difference between logname and whoami (from --help for each on Linux):

- logname: Print the name of the current user.
- whoami: Print the user name associated with the current effective user ID.

Next I looked for an example to showcase the difference (with sudo rights):
- `sudo logname` -> your user name
- `sudo whoami` -> root

There are two whoami entries, `windows/whoami` (no problem here) and `commons/whoami`; the latter falsely states "# Show the username of the current user." and "Display currently logged user name:".

**Questions:**
- Should commons/whoami be corrected to "Print the user name associated with the current effective user ID." (and a sudo example added)
- Or is this too specific for commons and should be added to linux, sunos and osx where it also applies?

Thanks, Ben

mfrw:
Fixes #2418
----

### DIFF
--- a/pages/common/whoami.md
+++ b/pages/common/whoami.md
@@ -1,7 +1,11 @@
 # whoami
 
-> Show the username of the current user.
+> Print the username associated with the current effective user ID.
 
-- Display currently logged user name:
+- Display currently logged username:
 
 `whoami`
+
+- Display the username after a change in the user ID:
+
+`sudo whoami`


pages/common/whoami.md
```diff
@@ -1,7 +1,11 @@
...
 
-> Show the username of the current user.
+> Print the username associated with the current effective user ID.
 
-- Display currently logged user name:
+- Display currently logged username:
 
 `whoami`
+
+- Display the username after a change in the user ID:
```
agnivade: Does saying "Display the root username help" ?

mfrw: I think uid 0 is always printed as root.

agnivade: Ah ok.

PR #2426, opened 2018-10-11
hr: add page

arielallon:
tool is here: https://github.com/LuRsT/hr

----

### DIFF
--- a/pages/common/hr.md
+++ b/pages/common/hr.md
@@ -12,4 +12,4 @@
 
 - Print a multiline horizontal rule:
 
-`hr {{string}} {{string}} {{string...}}`
+`hr {{string_a}} {{string_b}} {{string_c}}`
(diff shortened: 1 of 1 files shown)


pages/common/hr.md
```diff
@@ -0,0 +1,15 @@
...
+
+`hr`
+
+- Print a horizontal rule with a custom string:
+
+`hr {{string}}`
+
+- Print a multiline horizontal rule:
+
+`hr {{string}} {{string}} {{string...}}`
```
agnivade: Remove the ... from the last string.

sbrl: You could always do `string_a`, `string_b`, and `string_c` instead.

arielallon: I'll go with the latter to indicate they are different.
I wanted to indicate that there is no limit to the number of strings, but since tldr is aimed at examples more than a full man page, I'm ok with not explicitly saying that.

PR #2433, opened 2018-10-12
rdpsign: add page

pxgamer:
----

### DIFF
--- a/pages/windows/rdpsign.md
+++ b/pages/windows/rdpsign.md
@@ -18,6 +18,6 @@
 
 `rdpsign {{path/to/file.rdp}} /v`
 
-- Test the signing and output the result without updating the input:
+- Test the signing by displaying the output to console without updating the file:
 
 `rdpsign {{path/to/file.rdp}} /l`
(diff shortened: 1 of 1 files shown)


pages/windows/rdpsign.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Enable quiet output:
+
+`rdpsign {{path/to/file.rdp}} /q`
+
+- Display verbose warnings, messages and statuses:
+
+`rdpsign {{path/to/file.rdp}} /v`
+
+- Test the signing and output the result without updating the input:
```
agnivade: Maybe just say - "Test the signing by displaying the output to console without updating the file" ?

pxgamer: Just interested, is it better to say `output to console` or `output to stdout`?

sbrl: I'd probably go with `stdout` myself, as it makes sense in both interactive and non-interactive contexts.

agnivade: I'm fine with both. I think in other pages, we have a mixture of both console and stdout.

pxgamer: Went with `stdout` :+1:

PR #2440, opened 2018-10-13
banner: add page

glenacota:
----

### DIFF
--- a/pages/common/banner.md
+++ b/pages/common/banner.md
@@ -2,7 +2,7 @@
 
 > Print the given argument as a large ASCII art.
 
-- Print the text message as a large banner. Note: quotes are optional:
+- Print the text message as a large banner (quotes are optional):
 
 `banner {{"Hello World"}}`
 
--- a/pages/common/banner.md
+++ b/pages/common/banner.md
@@ -1,0 +1,15 @@
+# banner
+
+> Print large ASCII art version of the given argument.
+
+- Print the text message as a large banner. Note: quotes are optional:
+
+`banner {{"Hello World"}}`
+
+- Print the text message as a banner with a given width:
+
+`banner -w 50 {{"Hello World"}}`
+
+- Read text from the standard input and print it as a large banner:
+
+`banner`
--- a/pages/common/banner.md
+++ b/pages/common/banner.md
@@ -10,6 +10,6 @@
 
 `banner -w {{50}} {{"Hello World"}}`
 
-- Read text from stdin and print it as a large banner:
+- Read text from stdin:
 
 `banner`
--- a/pages/common/banner.md
+++ b/pages/common/banner.md
@@ -8,7 +8,7 @@
 
 - Print the text message as a banner with a width of 50 characters:
 
-`banner -w 50 {{"Hello World"}}`
+`banner -w {{50}} {{"Hello World"}}`
 
 - Read text from stdin and print it as a large banner:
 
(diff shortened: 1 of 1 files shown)


pages/common/banner.md
```diff
@@ -0,0 +1,15 @@
...
+
+> Print large ASCII art version of the given argument.
+
+- Print the text message as a large banner. Note: quotes are optional:
+
+`banner {{"Hello World"}}`
+
+- Print the text message as a banner with a given width:
+
+`banner -w 50 {{"Hello World"}}`
```
pxgamer: Put `50` in token syntax :+1:

pages/common/banner.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Read text from the standard input and print it as a large banner:
```
pxgamer: Might be easier to use `from stdin` instead of `from the standard input`?

glenacota: Agree :)

pages/common/banner.md
glenacota: 👍

pages/common/banner.md
```diff
@@ -0,0 +1,15 @@
+# banner
+
+> Print large ASCII art version of the given argument.
```
sbrl: Perhaps `Prints text as large ASCII art.` would read better?

glenacota: Thanks for your suggestion. My proposed description takes inspiration from the one of the _echo_ command. I would keep the root form of the verb, as in the majority of the other commands. But I like "Print the given argument as a large ASCII art."
What do you think?

sbrl: Yeah, go for `Print the given argument as a large ASCII art.` then :slightly_smiling_face:

pages/common/banner.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Print the text message as a large banner. Note: quotes are optional:
+
+`banner {{"Hello World"}}`
+
+- Print the text message as a banner with a width of 50 characters:
+
+`banner -w 50 {{"Hello World"}}`
+
+- Read text from stdin and print it as a large banner:
```
agnivade: Just "Read text from stdin" is fine

pages/common/banner.md
```diff
@@ -0,0 +1,15 @@
+# banner
+
+> Print the given argument as a large ASCII art.
+
+- Print the text message as a large banner. Note: quotes are optional:
```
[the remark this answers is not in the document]
glenacota: Ok, but I wanted to be consistent with the format used for the _echo_ command. Do you think we should change the sentence also there?

pages/common/banner.md
glenacota: Thanks for the suggestion, but - just for my understanding - you mean that "and print it as a large banner" here is redundant because there are other examples before?

Also, I wonder if there is a general rule for it - see, for example, the pages of commands _gist_ and _bash_. The review process would be faster for you if there is a shared, consistent way to describe examples and descriptions. What do you think?

agnivade: Your understanding is correct.

In general, if we have explained something 2 times, there is no need to repeat that as it becomes implicit. However, it is a rule of thumb and depends on the context and the page. Therefore, we don't have it as a written rule.

I apologize if you have looked through other pages and seen inconsistencies. Sometimes, other pages have been committed years ago and reviewers also miss something. We always try to keep everything consistent, but sometimes things slip through cracks.

Thanks for your understanding. And I appreciate you taking the time to look through other pages !

PR #2442, opened 2018-10-14
coreutils commands: move pages to common/ folder

felixonmars:
As suggested in #2196. The commands in coreutils are installable under
Mac and Windows, so let's move them to common/ folder.

----


agnivade:
Why is the sort file deleted ?

felixonmars:
It includes one different example, but the rest are the same as the one in common/. If the example is important we should add it into the common/ one instead.

agnivade:
Sure, let's do that. I wouldn't want to lose information. If we want to remove any example, we should do it in a separate PR. But since this PR is just about moving pages, I would like all examples to remain.

felixonmars:
OK, updated.

sbrl:
Thanks, @felixonmars :smiley_cat:

PR #2449, opened 2018-10-15
shutdown: add page

pxgamer:
----

### DIFF
--- a/pages/windows/shutdown.md
+++ b/pages/windows/shutdown.md
@@ -26,7 +26,7 @@
 
 `shutdown /s /c "{{reason}}"`
 
-- Abort a shutdown sequence:
+- Abort a shutdown sequence whose timeout is yet to expire:
 
 `shutdown /a`
 
--- a/pages/windows/shutdown.md
+++ b/pages/windows/shutdown.md
@@ -1,0 +1,35 @@
+# shutdown
+
+> A tool for shutting down, restarting or logging off a machine.
+
+- Shutdown the current machine:
+
+`shutdown /s`
+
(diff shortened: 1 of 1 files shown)


pages/windows/shutdown.md
```diff
@@ -0,0 +1,35 @@
...
+
+`shutdown /s /c "{{reason}}"`
+
+- Abort a shutdown sequence:
+
+`shutdown /a`
+
+- Shutdown a remote machine:
+
+`shutdown /m {{\\computer}}`
```
mfrw: This seems unclear. Should we mention specify if its the computer name or hostname/IP?

pxgamer: I guess I can add `using it's hostname or computer name:` or something along those lines.

sbrl: Maybe we could update the example instead to `\\hostname`?

agnivade: Yes, `\\hostname` is definitely clearer

pages/windows/shutdown.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Specify a timeout in seconds to wait before shutting down:
+
+`shutdown /s /t {{seconds}}`
+
+- Specify a comment for the shutdown reason:
+
+`shutdown /s /c "{{reason}}"`
+
+- Abort a shutdown sequence:
```
agnivade: I am guessing this aborts an active shutdown when the timeout is yet to expire ? In that case, I think we can clarify it like "Abort a shutdown sequence whose timeout is yet to expire"

PR #2452, opened 2018-10-15
mdutil: add page

glenacota:
----

### DIFF
--- a/pages/osx/mdutil
+++ b/pages/osx/mdutil
@@ -1,0 +1,19 @@
+# mdutil
+
+> Manage the metadata stores used by Spotlight for indexing.
+
+- Show the indexing status of the startup volume:
+
+`mdutil -s {{/}}`
+
+- Turn off the Spotlight indexing for the volume mounted on /Volumes/Foo:
+
+`mdutil -i off {{/Volumes/Foo}}`
+
+- Turn on the Spotlight indexing:
+
+`mdutil -i on {{/Volumes/Foo}}`
+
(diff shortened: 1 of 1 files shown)


sbrl:
Hey, thanks for the new page, @glenacota! I think you forgot the `.md` file extension, though.

glenacota:
> Hey, thanks for the new page, @glenacota! I think you forgot the `.md` file extension, though.

Oh, my :sweat_smile: sorry :)

pages/osx/mdutil.md
```diff
@@ -0,0 +1,19 @@
+# mdutil
+
+> Manage the metadata stores used by Spotlight for indexing.
+
+- Show the indexing status of the startup volume:
+
+`mdutil -s {{/}}`
+
+- Turn off the Spotlight indexing for the volume mounted on /Volumes/Foo:
```
pxgamer: Possibly merge this with the one below?
```
mdutil -i {{on|off}} {{path/to/volume}}
```

Also, is `/Volumes` an example storage directory? :slightly_smiling_face:

glenacota: Merging the two examples is a very good suggestion, thanks! 
As for the second question, do you mean that /Volumes/Foo was not a good example?

agnivade: We usually use `path/to/volume` unless a concrete directory makes an example clearer. If /Volumes/Foo is just a normal mounted path, I think path/to/volume is a better choice.

PR #2472, opened 2018-10-20
gnomon: add page

Aracki:
[Gnomon](https://github.com/paypal/gnomon) is the utility tool for finding slow processes made by Paypal.

### DIFF
--- a/pages/common/gnomon.md
+++ b/pages/common/gnomon.md
@@ -16,8 +16,8 @@
 
 - High threshold. If the elapsed time for a line is equal to or higher than this value in seconds, then the timestamp will be colored bright red:
 
-`{{npm test}} | gnomon --high 0.5`
+`{{npm test}} | gnomon --high {{0.5}}`
 
 - Medium threshold. If the elapsed time for a line is equal to or higher than this value in seconds, then the timestamp will be colored bright yellow:
 
-`{{npm test}} | gnomon --medium 0.5`
+`{{npm test}} | gnomon --medium {{0.5}}`
--- a/pages/common/gnomon.md
+++ b/pages/common/gnomon.md
@@ -14,10 +14,6 @@
 
 `{{npm test}} | gnomon --type=absolute`
 
-- Format the absolute timestamp, using PHP date format strings:
-
-`{{npm test}} | gnomon --type=absolute --format "H:i:s.u 0"`
-
 - High threshold. If the elapsed time for a line is equal to or higher than this value in seconds, then the timestamp will be colored bright red:
 
 `{{npm test}} | gnomon --high 0.5`
--- a/pages/common/gnomon.md
+++ b/pages/common/gnomon.md
@@ -10,7 +10,7 @@
 
 `{{npm test}} | gnomon --type=elapsed-total`
 
-- Show an absolute timestamp in UTS:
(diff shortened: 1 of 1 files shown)


pages/common/gnomon.md
```diff
@@ -0,0 +1,11 @@
+# gnomon 
+
+> Utility to annotate console logging statements with timestamps and find slow processes.
+
+- Use UNIX (or DOS) pipes to pipe the stdout of any command through gnomon:
+
+`npm test | gnomon`
+
+- Use command-line options to adjust gnomon's behavior for any command:
```
sbrl: This example seems rather general. The point of a tldr-page is to demonstrate the most useful arguments and combinations so that you don't have to look them up :-)

Aracki: Maybe it's enough `npm test | gnomon` command?

sbrl: Looking at the README, there are a bunch of potentially useful flags there. We could cover `--type=elapsed-total`, `--high` & `--medium`, and even `--real-time=false` (not all in the same example, of course :P

sbrl: Looking back, I don't think I explained myself very well. It's the `Use command-line options` in this example that I'm unsure about, as it implies to me that I need to read the man page - and that's what tldr-pages should help you have to do less :P

pages/common/gnomon.md
```diff
@@ -4,8 +4,24 @@
...
-`npm test | gnomon`
+`{{npm test}} | gnomon`
 
-- Use command-line options to adjust gnomon's behavior for any command:
+- Show number of seconds since the start of the process:
 
-`{{any command}} | gnomon --type=elapsed-total --high=8.0`
+`{{npm test}} | gnomon --type=elapsed-total`
+
+- Show an absolute timestamp in UTS:
```
sbrl: What's `UTS`?

Aracki: It should be UTC. I need to stop making typos.

pages/common/gnomon.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Show an absolute timestamp in UTC:
+
+`{{npm test}} | gnomon --type=absolute`
+
+- Format the absolute timestamp, using PHP date format strings:
+
+`{{npm test}} | gnomon --type=absolute --format "H:i:s.u 0"`
+
+- High threshold. If the elapsed time for a line is equal to or higher than this value in seconds, then the timestamp will be colored bright red:
```
agnivade: This is too long :). How about  - "Set a high threshold of 0.5 seconds for the elapsed time; exceeding which the timestamp will be colored bright red". ?

pages/common/gnomon.md
```diff
@@ -0,0 +1,27 @@
...
+
+`{{npm test}} | gnomon --high 0.5`
+
+- Medium threshold. If the elapsed time for a line is equal to or higher than this value in seconds, then the timestamp will be colored bright yellow:
```
agnivade: "Set a medium threshold of 0.2 second. (Timestamp will be colored bright yellow)". We do not need to repeat the entire explanation again.

Also, set the threshold to 0.2 second, to differentiate from high

sbrl:
Thanks, @Aracki :smiley_cat:

PR #2473, opened 2018-10-20
m: add page

Aracki:
[m-cli](https://github.com/rgcr/m-cli) alias `m` is Swiss Army Knife for macOS.  
I have added 5 simple examples what `m` is able to do.

### DIFF
--- a/pages/osx/m.md
+++ b/pages/osx/m.md
@@ -1,0 +1,23 @@
+# m
+
+> Swiss Army Knife for macOS
+
+- Get the battery status:
+
+`m battery status`
+
+- Turn off the bluetooth:
+
+`m bluetooth off`
+
+- List available filesystems for formatting:
+
+`m disk filesystems`
+
+- Enable Dock's auto hide feature:
+
+`m docker autohide YES`
+
+- Disable firewall:
+
+`m firewall disable`
--- a/pages/osx/m.md
+++ b/pages/osx/m.md
@@ -6,7 +6,7 @@
 
 `m battery status`
 
-- Turn off the bluetooth:
+- Turn off bluetooth:
 
 `m bluetooth off`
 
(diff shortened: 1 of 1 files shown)


pages/osx/m.md
```diff
@@ -0,0 +1,23 @@
+# m
+
+> Swiss Army Knife for macOS.
+
+- Get the battery status:
+
+`m battery status`
+
+- Turn off the bluetooth:
```
sbrl: Not sure that `the` is needed here.

pages/osx/m.md
```diff
@@ -0,0 +1,23 @@
...
+
+`m bluetooth off`
+
+- List available filesystems for formatting:
```
sbrl:
```suggestion
- List available the filesystems for formatting:
```

Perhaps this reads better?

pages/osx/m.md
```diff
@@ -0,0 +1,23 @@
...
+
+`m disk filesystems`
+
+- Enable Dock's auto hide feature:
+
+`m docker autohide YES`
+
+- Disable firewall:
```
sbrl:
```suggestion
- Disable the firewall:
```

Perhaps this reads better?

pages/osx/m.md
Aracki: Applied!

pages/osx/m.md
Aracki: No. I'm sure there is no need for "the" here.

pxgamer: I agree, `the` should either be before the `available` (i.e. `List the available...`) or left as is (which I feel makes more sense). :+1:

Aracki: I would definitely leave as it is.

PR #2474, opened 2018-10-20
source: add page

Aracki:
Add `source` page.

### DIFF
--- a/pages/common/source.md
+++ b/pages/common/source.md
@@ -2,6 +2,6 @@
 
 > Execute commands from a file in the current shell.
 
-- Evaluates contents of file:
+- Evaluate contents of a given file:
 
 `source {{path/to/file}}`
--- a/pages/common/source.md
+++ b/pages/common/source.md
@@ -2,14 +2,6 @@
 
 > Execute commands from a file in the current shell.
 
-- Cause bash to re-read its initialization file:
+- Evaluates contents of file:
 
-`source ~/.bashrc`
-
-- Cause bash to re-read its aliases:
-
-`source ~/.bash_aliases`
-
-- Cause fish to re-read its initialization file:
-
-`source ~/.config/fish/config.fish`
+`source {{path/to/file}}`
--- a/pages/common/source.md
+++ b/pages/common/source.md
@@ -1,0 +1,7 @@
+# source
+
+> Evaluate contents of file.
+
+- Causes fish to re-read its initialization file:
+
+`source ~/.config/fish/config.fish`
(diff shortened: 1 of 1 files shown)


pages/common/source.md
```diff
@@ -0,0 +1,7 @@
+# source
+
+> Evaluate contents of file.
```
mfrw: How about `A shell builtin command to executes the content of the file passed as argument`

pages/common/source.md
```diff
@@ -0,0 +1,7 @@
...
+
+- Causes fish to re-read its initialization file:
```
mfrw: This would only work if we are in `fish` shell. I think we should mention that.

pages/common/source.md
```diff
@@ -0,0 +1,7 @@
+# source
+
+> Evaluate contents of file.
+
+- Cause fish to re-read its initialization file:
```
agnivade: We need an example which shows a basic usage. Something like source {{file.sh}} would do.

pages/common/source.md
Aracki: There is another explanation from bash shell: `Execute commands from a file in the current shell.` What do you think about that?

pages/common/source.md
Aracki: How you would change the example of mine: `source ~/.config/fish/config.fish`?

pages/common/source.md
sbrl: I like the latter there :-)

pages/common/source.md
sbrl: This isn't a `fish`-specific command - `bash` has it too. Perhaps we could adapt the description here a bit?

Aracki: @sbrl I have added two more commands. And changed description too.

pages/common/source.md
Aracki: I've added bash examples alongside fish. Do you think it's enough?

pages/common/source.md
sbrl: I still see `Cause fish to re-read its initialization file` here. Was it a different description you changed?

sbrl: Sounds good to me! A generic `{{path/to/file}}` example is all that's needed I think.

Aracki: Maybe `Evaluates contents of file` or `Execute commands from a file` (same as main description)?

agnivade: `Evalute contents of a given file` sounds good.

sbrl:
Merging as @mfrw's comments have been resolved.

PR #2476, opened 2018-10-22
ag: fixed misleading information

NightMachinary:
Fixed a misleading description that wasted quite a bit of my time.

----

### DIFF
--- a/pages/common/ag.md
+++ b/pages/common/ag.md
@@ -2,26 +2,26 @@
 
 > The Silver Searcher. Like ack, but faster.
 
-- Find files containing "foo", and print the line matches in context:
+- Find files containing "foo", and print the line matches in context (`-F` is an alias for `--literal`):
 
-`ag {{foo}}`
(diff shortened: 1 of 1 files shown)


pages/common/ag.md
```diff
@@ -2,26 +2,26 @@
 
 > The Silver Searcher. Like ack, but faster.
 
-- Find files containing "foo", and print the line matches in context:
+- Find files containing "foo", and print the line matches in context (`-F` is an alias for `--literal`):
```
sbrl: What does `--literal` do?

mfrw: @sbrl As per the documentation `Don't parse PATTERN as a regular expression`

agnivade: I would like to keep the first example without any flags. We should change the description instead.

agnivade: Also, maybe bring up the penultimate example to the top. I think that might clarify that foo may be a normal text as well as a regular expression.

PR #2478, opened 2018-10-22
command: add page

Aracki:
I have added Unix basic command `command` which is useful when trying to override aliases for example.

There are `--all` and `--search` flags but I didn't put them on purpose cause I think there are no use cases for them.


pages/common/command.md
```diff
@@ -0,0 +1,7 @@
+# command
+
+> Command forces the shell to execute the program and ignore any functions, builtins and aliases with the same name.
+
+- Causes shell to execute the ls program, even if an ls alias exists:
```
sbrl: Looks like you've got a linter error. What about this?

```suggestion
- Causes shell to execute the ls program literally, even if an ls alias exists:
```

agnivade:
It's in the common folder. So does it exist in windows too ?

Aracki:
> It's in the common folder. So does it exist in windows too ?

Good point @agnivade. I will copy the file in `osx` and `linux` folders. I have no knowledge of `sunos` so far. 😕

PR #2479, opened 2018-10-22
sudo: general revision of entry

kalebo:
My changes are as follows:

1. Changed the more complex examples *involving* `sudo` to simple examples showing the *usage* of `sudo`.
2. Separate the examples of editing a file and doing something as another user
3. Standardized the wording to use superuser instead of root
4. Clarified the usage of `sudo !!` which is not a feature of sudo, but the shell you use. I'm half tempted to say we should just take it out, but it is a very useful shortcut that more people should know about.
5. Added the `-e` flag that allows you to edit files using your default editor as specified by the environment variables `VISUAL` or `EDITOR`. This has several other benefits over editing the file directly with a command like `sudo vim /etc/fstab` as well.
6. Fixed the odd syntax for the `shutdown` example which was visually confusing as to which flags were being passed to `shutdown` and which to `sudo`

I'm also don't like using examples like `sudo shutdown now` or anything else that could be harmful if blindly copy and pasted, so I'd be curious if there is a more innocuous command that we could use but still demonstrate the effects of `sudo`.

----

### DIFF
--- a/pages/common/sudo.md
+++ b/pages/common/sudo.md
@@ -1,23 +1,23 @@
 # sudo
 
-> Executes a single command as another user.
+> Executes a single command as the superuser or another user.
 
-- List the contents of an unreadable directory:
+- Run a command as the superuser:
 
-`sudo {{ls}} {{/usr/local/scrt}}`
+`sudo {{shutdown now}}`
 
-- Edit a file as the user www:
+- Edit a file as the superuser with your default editor:
 
-`sudo -u {{www}} {{vi}} {{/var/www/index.html}}`
+`sudo -e {{/etc/fstab}}`
 
-- Shut down the machine:
+- Run a command as another user and/or group:
 
-`sudo {{shutdown}} -h +10 {{"Cya soon!"}}`
+`sudo -u {{user}} -g {{group}} {{id}}`
 
-- Repeat the last command as sudo:
+- Repeat the last command prefixed with "sudo" (only in bash, zsh, etc.):
 
 `sudo !!`
 
-- Launch the default shell with root privileges:
+- Launch the default shell with superuser privileges:
 
 `sudo -i`
--- a/pages/common/sudo.md
+++ b/pages/common/sudo.md
@@ -4,7 +4,7 @@
 
 - Run a command as the superuser:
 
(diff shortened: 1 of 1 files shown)


pages/common/sudo.md
```diff
@@ -1,23 +1,23 @@
...
+- Edit a file as the superuser with your default editor:
 
-`sudo -u {{www}} {{vi}} {{/var/www/index.html}}`
+`sudo -e {{/etc/fstab}}`
 
-- Shut down the machine:
+- Run a command as another user and/or group:
 
-`sudo {{shutdown}} -h +10 {{"Cya soon!"}}`
+`sudo -u {{user}} -g {{group}} {{id}}`
```
agnivade: Maybe we can change `id` to `whoami` to make it more clear ?

kalebo: I actually used `whoami` when I first rewrote the example; however, the trouble with `whoami` is that the output does not change when including `-g` flag in sudo whereas `id`does. Although I do worry that `id` may be mistaken for another variable rather than a command. Is another command besides these two we could use?

sbrl: I didn't even know that executing as another group was a thing! Could we use an `echo` command to output the contents of an environment variable?

agnivade: Yea, even I thought id was another flag or something. If there is no better option, then I am fine with it.

kalebo: I've thought about this some more and I can't come up with another command that is as succinct as `id` and shows the effects of `sudo` as well as `id` does. One possibility to make it clearer that it is a command rather than another flag to sudo is to use `id -a` instead. The `-a` flag is ignored in most versions, but it's there for compatibility reasons. What do you guys think?

agnivade: Great, let's do that.

sbrl:
Merging due to inactivity, since the described issues appear to have been fixed.

agnivade:
According to this comment of mine https://github.com/tldr-pages/tldr/pull/2479/files#r232543198, we were still waiting for @kalebo to add a `-a` option. However, we can still do it in a new PR.

PR #2488, opened 2018-10-24
m4: add page

arunisaac:

### DIFF
--- a/pages/common/m4.md
+++ b/pages/common/m4.md
@@ -1,0 +1,11 @@
+# m4
+
+> Macro processor.
+
+- Process macros in file:
+
+`m4 {{file}}`
+
+- Define a macro before processing files:
+
+`m4 -D{{macro_name}}={{macro_value}} {{file}}`


pages/common/m4.md
```diff
@@ -0,0 +1,11 @@
+# m4
+
+> Macro processor.
+
+- Process macros in file:
```
sbrl: Perhaps this will read better?

```suggestion
- Process macros in a file:
```

Also, what is mean  by `process`? Does it execute them, or simply parse them?

Perhaps we can change `{{file}}` to `{{path/to/file}}`?

pages/common/m4.md
```diff
@@ -0,0 +1,11 @@
...
+
+`m4 {{file}}`
+
+- Define a macro before processing files:
+
+`m4 -D{{macro_name}}={{macro_value}} {{file}}`
```
sbrl: Is there a concrete example we can use here?

arunisaac:
> Perhaps this will read better?
>
> ```suggestion
> - Process macros in a file:
> ```

Agreed.

> Also, what is mean  by `process`? Does it execute them, or simply parse them?

m4 expands macros in the specified file and prints out the results. m4
is a Turing complete macro language. So, the macro expansion can involve
any arbitrary computation. IMO, `process` is the word that best
describes this. Also, that is the word the m4 man page uses. But, we
could instead say `expand` if you prefer that.

> Perhaps we can change `{{file}}` to `{{path/to/file}}`?

Sure.

> +- Define a macro before processing files:
> +
> +`m4 -D{{macro_name}}={{macro_value}} {{file}}`
>
> Is there a concrete example we can use here?

I'm not entirely sure what kind of a concrete example we can provide. Do
you have something in mind?

In m4, defining a macro before processing a file, is roughly equivalent
to setting a variable before executing a script in some scripting
language. As such, any concrete example we provide would require us to
also show the contents of the file and how that macro is used in that
file.

agnivade:
> As such, any concrete example we provide would require us to
also show the contents of the file and how that macro is used in that
file.

Let's leave it then.

sbrl:
Process seems fine, @arunisaac :-)

PR #2489, opened 2018-10-24
guix-package: add page

arunisaac:

### DIFF
--- a/pages/linux/guix-package.md
+++ b/pages/linux/guix-package.md
@@ -1,0 +1,27 @@
+# guix package
+
+> Install, upgrade and remove Guix packages, or rollback to previous configurations.
+
+- Install a new package:
+
+`guix package -i {{package_name}}`
+
+- Remove a package:
+
+`guix package -r {{package_name}}`
+
+- Search the package database for a regular expression:
+
+`guix package -s "{{search_pattern}}"`
+
+- List installed packages:
+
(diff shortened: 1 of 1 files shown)


pages/linux/guix-package.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Search the package database for a regular expression:
+
+`guix package -s "{{search_pattern}}"`
+
+- List installed packages:
+
+`guix package -I`
+
+- List generations:
```
sbrl: What's a generation?

arunisaac:
> Do you have plans to write a `guix` page too?

I have plans to write a guix-environment page and a guix-system page. Do
we need to have a separate guix page?

Can the guix-environment page and the guix-system page be separate pull
requests?

> +- List generations:
>
> What's a generation?

In Guix, a generation is a system state, that is, the set of all
packages installed. Guix remembers previous system states (or
generations), and it is possible to roll back to an earlier state.

In analogy with git, guix generations are like git commits. Just like it
is possible to revert git commits and go back to an earlier repo state,
in Guix, it is possible to revert package installs and go back to an
earlier system state.

sbrl:
Thanks for clearing that up, @arunisaac!

Without a `guix` page, typing `tldr guix` will yield a misleading 404.

PR #2494, opened 2018-10-25
git checkout: add example for creating new branch based on remote branch with different name

mikerogne:
Adds an example for creating and checking out a new branch that's based on an existing remote/source branch with a different name.

It's similar to https://github.com/tldr-pages/tldr/pull/1359 - but shows the example of allowing the new branch being created having a different branch name than the source.

It's something I don't use super often, but end up having to Google each time because it's not in `tldr git checkout` :)

----

### DIFF
--- a/pages/common/git-checkout.md
+++ b/pages/common/git-checkout.md
@@ -6,6 +6,10 @@
 
 `git checkout -b {{branch_name}}`
 
+- Create and switch to a new branch based on an existing remote branch:
+
+`git checkout -b {{branch_name}} {{remote/source_branch}}`
+
 - Switch to an existing local branch:
 
 `git checkout {{branch_name}}`
--- a/pages/common/git-checkout.md
+++ b/pages/common/git-checkout.md
@@ -6,9 +6,9 @@
 
 `git checkout -b {{branch_name}}`
 
-- Create and switch to a new branch based on an existing remote branch:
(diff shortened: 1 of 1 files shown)


pages/common/git-checkout.md
```diff
@@ -6,6 +6,10 @@
 
 `git checkout -b {{branch_name}}`
 
+- Create and switch to a new branch based on an existing remote branch:
+
+`git checkout -b {{branch_name}} {{remote/source_branch}}`
```
mfrw: Usually when we checkout a remote branch, we also want to track it. Should we add that option too or will it be too complicated? @agnivade @sbrl @pxgamer 
```suggestion
`git checkout -b {{branch_name}} {{remote/source_branch}} --track`
```

mikerogne: Heyas! Thanks for checking it out so fast. No pun intended 👈👈😎.

My use-case for it was to demonstrate creating a new branch based on a source branch that has a different name, even from a remote. There was some --track/--tracking option mentioned in the other PR I referenced in the description. Not sure. Up to you guys!

mikerogne:
I think https://github.com/tldr-pages/tldr/pull/2494/commits/c2c7434f645a3702289c97da55ad34c150b1388c is the clearest description I can come up with. Valid examples are included. Tried my best to keep it very succinct.

FatBoyXPC:
Pretty slick change, I used to have to google this quite often!

PR #2500, opened 2018-10-26
opt: add page

guilhermeleobas:
Hi,

I've added a page for the LLVM [`opt`](https://linux.die.net/man/1/opt) tool. 

----

### DIFF
--- a/pages/common/opt.md
+++ b/pages/common/opt.md
@@ -4,12 +4,12 @@
 
 - Run an optimization or analysis on a bitcode file:
 
-`opt {{-passname}} {{file.bc}} -S -o {{file_opt.bc}}`
+`opt {{-passname}} {{path/to/file.bc}} -S -o {{file_opt.bc}}`
 
-- Print the Control Flow Graph of a function to a "dot" file:
+- Output the Control Flow Graph of a function to a "dot" file:
 
-`opt {{-dot-cfg}} -S {{file.bc}} -disable-output`
+`opt {{-dot-cfg}} -S {{path/to/file.bc}} -disable-output`
 
-- Optimize the program with "-O2" and write its output to another file:
+- Optimize the program at level 2 and output the result to another file:
 
-`opt -O2 {{file.bc}} -S -o {{file_opt.bc}}`
+`opt -O2 {{path/to/file.bc}} -S -o {{file_opt.bc}}`


pages/common/opt.md
```diff
@@ -0,0 +1,15 @@
+# opt
+
+> LLVM .bc -> .bc modular optimizer and analysis printer. `opt` takes LLVM source files as input, runs the specified optimizations or analyses on it, and then outputs the optimized file or the analysis results.
```
sbrl: This seems rather wordy. Can we make it more concise at all?

Also, what is meant by `.bc`?

pages/common/opt.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Run an optimization or an analysis in a bitcode file:
```
sbrl: Perhaps

```suggestion
- Run an optimization or analysis on a bitcode file:
```

....will read better?

Also, can `file.bc` be `path/to/file.bc` please?

pages/common/opt.md
```diff
@@ -0,0 +1,15 @@
...
+
+`opt {{-passname}} {{file.bc}} -S -o {{file_opt.bc}}`
+
+- Print the Control Flow Graph of a function to a "dot" file:
```
sbrl: Perhaps `Output` would be a better term than `Print` here if it's going to a file and not a terminal?

Also, can `file.bc` be `path/to/file.bc` please?

pages/common/opt.md
```diff
@@ -0,0 +1,15 @@
...
+
+`opt {{-dot-cfg}} -S {{file.bc}} -disable-output`
+
+- Optimize the program with "-O2" and write its output to another file:
```
sbrl: Perhaps we could specify what `-O2` actually does instead? How about this?

```suggestion
- Optimize the program at level 2 and output the result to another file:
```

Also, perhaps `file_opt.bc` can be `path/to/file_opt.bc` instead?

pages/common/opt.md
guilhermeleobas: > This seems rather wordy. Can we make it more concise at all?

Ok. I will think on something else.

> Also, what is meant by .bc?

`.bc` is a bitcode file while `.ll` is a human-readable version of it.

guilhermeleobas:
Hi @sbrl, thanks for the review. I've applied your suggestions.

pages/common/opt.md
```diff
@@ -0,0 +1,15 @@
+# opt
+
+> LLVM .bc -> .bc modular optimizer and analysis printer. `opt` takes LLVM source files as input, runs the specified optimizations or analyses on it, and then outputs the optimized file or the analysis results.
+
+- Run an optimization or analysis on a bitcode file:
+
+`opt {{-passname}} {{path/to/file.bc}} -S -o {{file_opt.bc}}`
```
agnivade: `-` of passname should be outside the tokens.

pages/common/opt.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Output the Control Flow Graph of a function to a "dot" file:
+
+`opt {{-dot-cfg}} -S {{path/to/file.bc}} -disable-output`
+
+- Optimize the program at level 2 and output the result to another file:
+
+`opt -O2 {{path/to/file.bc}} -S -o {{file_opt.bc}}`
```
agnivade: path/to/output_file.bc

pages/common/opt.md
agnivade: Yea, this is a bit too long for a tldr page. How about - "A tool that takes LLVM source files and runs specified optimizations or analyses on them" ?

guilhermeleobas:
Hi @agnivade and @sbrl, I've applied your suggestions but my changes aren't being reflected in this PR. Maybe it's because I naively deleted the fork of my other PR (#2499). 

I've tried to resync another fork with this PR [(ref)](https://github.com/isaacs/github/issues/168#issuecomment-374201226) but that didn't work. Maybe the best is if I open a new PR. What do you guys think?

Sorry for the trouble.

sbrl:
Hrm. Each PR should end up on a new branch, @guilhermeleobas.

Personally, I use the Web UI for PRs & editing them for tldr-pages :P

guilhermeleobas:
I am aware of that but look the header of this PR. It says the name of the base branch is **unknown branch** instead of **opt**:
image

sbrl:
Ah, I see!

Yeah, I suspect that this PR and your previous one were from the same branch.

guilhermeleobas:
I think I was using different branches. Because otherwise, my commits from the previous PR would appear in this one. Anyway, do you think the best approach is to open a new PR with the changes asked by @agnivade?

agnivade:
Sure, let's do that.

Issue #2509, opened 2018-10-27
Inconsistencies in token definitions


glenacota:
Hello,
I'd like to start uniforming the definition of tokens among TLDR pages, while also enforcing the [syntax guidelines](https://github.com/tldr-pages/tldr/blob/master/contributing-guides/style-guide.md). I believe that more consistency will not only help TLDR users, but also new contributors.

For convenience, I was thinking about creating a number of narrow-scoped pull-requests instead of a single massive one. Each pull-request will refer to this issue. 

If you also believe this is a good idea, please feel free to join me in the process. If instead you don't think this is the right approach, let's discuss how to improve it! 😄 

Cheers,
Guido

waldyrious:
Hey @glenacota! Your suggestions are most welcome, and individual narrow-scoped pull requests indeed seem like a good approach for mass changes.

I'm not really sure I understand what kinds of changes you have in mind, though. Before starting making PRs, it might be a good idea to lay out your plans here, so they can be discussed and refined if needed.

glenacota:
Hi @waldyrious, the kind of changes I had in mind were something like:
- enforce the syntax guidelines, e.g., use only snake case notations, no capital letters
- use a common definition for the same token, e.g., ```filename``` instead of ```file_name```. As a rule of thumb to select the common definition, I'd choose the most recurrent correct definition among all pages

agnivade:
Thanks for taking this initiative. I had done that long time ago, after which we actually created the syntax guideline. Before that it was the wild west ! But sometimes things fall through the cracks and it's always good to clean things up.

Your first point seems fine. Regarding the second point, I am willing to let it be. Reason being, it will cause unnecessary churn at no good cost. We also allow "file" and "path/to/file" where it seems appropriate. Bottom line is that a file token is so common, keeping it 100% consistent is a near impossible task. So unless there is a glaring issue, I would rather let it be.

waldyrious:
I do think changes such as those in #2510 have value (so I think its closure was too hasty). Granted, some of the changes there might be arguable, as the variations in capitalization or choice of word separators might represent conventions in the ecosystem of the corresponding tools; but overall the changes seemed sensible. Even whether we should reflect context-specific naming conventions in tldr pages might be worth discussing, IMO.

[...]

PR #2510, opened 2018-10-27
token definitions: improve consistency of tokens containing "name"

glenacota:
This PR aims to solve inconsistencies in tokens that contain the substring "name". 
In particular, I renamed ```host_name``` -> ```hostname```, ``` user_name``` -> ```username```, and ```file_name``` -> ```filename```. This is because the version without the underscore was not only correct, but also the most used in TLDR overall.

This PR is related to issue #2509.

### DIFF
--- a/pages/common/bosh.md
+++ b/pages/common/bosh.md
@@ -4,7 +4,7 @@
 
 - Create local alias for director:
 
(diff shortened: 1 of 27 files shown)


sbrl:
Hey, @glenacota! Looks like you're experiencing some issues. Is there anything I can help with?

Issue #2339, opened 2018-09-17
Support for multiple (human) languages


arunisaac:
It would be nice if tldr pages existed in other (human) languages too, and not just English. Is there any plan for tldr pages to support multiple languages?

agnivade:
I am sorry, but the entire page parsing logic is based around English punctuation marks. Like `:`, `>`.  We cannot _fully_ shift to other languages. English punctuation marks are compulsory.

Just simply converting the pages to other languages is possible. Although it will take architectural changes, which need to be implemented by every client out there. So, even if we make pages in other languages, until the client you are using implements it, you won't be able to use tldr.

But more importantly, someone has to _write_ the pages. This is the first request to support multiple languages. If there is sufficient interest, then we can decide on how to lay out the pages, and then each client can implement it gradually.

arunisaac:
For the most part, technical documentation is non-existent in most
Indian languages. And, owing to their size, translating existing
technical documentation requires a lot of labour to achieve even a small
degree of completeness. So, my thinking was that due to its small size,
the tldr pages could be translated with less effort and quickly reach
sufficient completeness to attract more translators and bootstrap the
process of translating bigger manuals. There are large sections of
society that would come forward to learn more about computers if that
knowledge was presented to them in their mother tongue.

pxgamer:
I wonder if something such as [GitLocalize](https://github.com/marketplace/gitlocalize) could help with this?

We could have it in a structure such as:
<details>

```
pages/
  en/
    common/
    linux/
    osx/
    windows/
  etc/
```

</details>

-----

But yes, as @sbrl mentioned, this would require more reviewers for other languages. 🤔

agnivade:
Breaking down into sub-folders will break all clients immediately. I would go with the usual github idiom for storing READMEs in multiple languages.

Like - <filename>_<2-letter-lang-code>.md. And leave the default as english. So that we can add new files without breaking anything. And clients can implement new language support gradually at ease.

pxgamer:
Hmm. That would definitely make more sense. 🤔 I wasn't really thinking.

arunisaac:
> Like - <filename>_<2-letter-lang-code>.md. And leave the default as
> english.

This sounds great. May I send pull requests for translated tldr pages
now?

> We'll need someone who can speak the target languages to review the
> PRs as we do for English. Quite often we suggest wording and
> grammatical changes etc. to improve the clarity of the page.

[...]

PR #2512, opened 2018-10-28
Tamil pages proof of concept

arunisaac:
This is a proof of concept for Tamil pages as discussed in #2339


arunisaac:
> Great work, @arunisaac!

Thanks! :-)

> We should probably define a common 'specification' for how it's going
> to be implemented in clients - I think we had an issue for that
> somewhere.
>
> Is there an environment variable that we could use to auto-detect?

How about using gettext's conventions?

https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html

gettext also has means to specify a priority list of languages. See
https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html

agnivade:
This is the [issue](https://github.com/tldr-pages/tldr/issues/1065). Please comment on it so that we have everything at one place.

agnivade:
I this it is the other way around. Clients cannot support translated pages, if this PR is blocked.

arunisaac:
If everything is fine, can this PR be merged? It would be easier to work
on client support if this PR is merged.

sbrl:
Merged! I agree that client support is much easier to deal with if we merge this PR - we can always revert it if need be :P

Also, I've guessed a standard PR / commit message format:

```
cp,ls,mkdir,mv,rm,rmdir: add translations [Tamil] (#2512)
```

waldyrious:
> I've guessed a standard PR / commit message format

Actually @arunisaac specifically [asked](https://github.com/tldr-pages/tldr/issues/2339#issuecomment-433880949) if the pattern he had used for individual commits (`<command>: add Tamil page`) was a good idea, and the agreement was that it was. I do agree your guess is a good counterproposal. Perhaps we could go with something like `cp: add Tamil translation`? What do you think @arunisaac @agnivade @sbrl?

waldyrious:
On a separate note, I'm sad that @arunisaac's care with separating the commits was not reflected in the final merge into the master branch. The [maintainers' guide](https://github.com/tldr-pages/tldr/blob/master/contributing-guides/maintainers-guide.md#ii-handling-prs) suggests that:

> If [...] the PR author took the time to craft individual, informative messages for each commit, then use the Rebase and merge method, to honor that work and preserve the history of the changes.

But I guess all these guides and documentation are becoming too verbose and more of a burden than a helpful cheatsheets for day-to-day operation of the project. Or maybe it was my fault to propose them and not stick around for long enough to ensure they get assimilated into the maintainer culture (as other principles did). In any case, I hope @arunisaac does not get discouraged from making atomically separated commits in the future :)

sbrl:
Oops! I think I missed both those points. On second thought, it would have been a better choice to rebase here rather than squash - I haven't rebased in ages on here! I'll try and consider this more  in the future. I can revert & re-merge here if you like?

This conversation has gotten rather long here and in #2339, so I think I missed the bit on the commit message standardisation too.

sbrl:
In terms of guides & documentation - especially in terms of the client spec, I think a `CLIENTS.md` document is needed that details the spec there. In terms of the multi-language stuff, adding it tot he maintainers guide (which was an interesting re-read, actually) sounds like a good idea. Maybe in a new sub-heading?

agnivade:
>  I haven't rebased in ages on here!

Yes it's rare. But I rebase from time to time, when I see good commits.

Regarding, commit message, I think we should update the PR template too. Since that mentions how a commit message should be.

arunisaac:
> In any case, I hope @arunisaac does not get discouraged from making
> atomically separated commits in the future :)

No, since you mentioned it, I won't be discouraged. :-)

Regarding the commit message format, I have no strong opinions. Perhaps,
I slightly prefer `cp: add Tamil page` to `cp: add Tamil translation`
since the former is slightly shorter.

waldyrious:
> I can revert & re-merge here if you like?

That would make the history even messier IMO. The only way to fix it would be to force-push to the master branch, but that's a pretty huge hammer that I don't think is justified in this case.

> Perhaps, I slightly prefer `cp: add Tamil page` to `cp: add Tamil translation` since the former is slightly shorter.

That's a fair point, but IMO the 5 characters that we save are worth less than the extra clarity (not everyone knows the name of all languages in the world). So my preference still remains `cp: add Tamil translation`. Thoughts, @agnivade and @sbrl?

> Regarding the multi-language stuff, adding it to the maintainers guide (which was an interesting re-read, actually) sounds like a good idea. Maybe in a new sub-heading?

I'm not sure the maintainers guide is the best place for that. Or do you mean the commit message format? Though even then, I think the general CONTRIBUTING.md guide would be a better place, don't you agree?

> Regarding, commit message, I think we should update the PR template too. Since that mentions how a commit message should be.

Good point. Once we decide on the format, let's update that too.

agnivade:
>  So my preference still remains cp: add Tamil translation. Thoughts, @agnivade and @sbrl?

Sure.

sbrl:
Yep, `cp: add Tamil translation` looks good to me. On a related note, could we add a link to the maintainer's guide in `CONTRIBUTING.md` please?

waldyrious:
I've made the discussed changes to CONTRIBUTING.md and the PR template at #2575.

I've added a link to the maintainer's guide in more places at #2576.

gowthamgts:
@arunisaac - thanks for this. It's really great to see tamil translations to oss projects.

arunisaac:
Thank you! :-) It would be nice to have many more translated pages,
though. I haven't worked on this in a long time. Do contribute
translations if you find time.

PR #2516, opened 2018-10-28
cat: add two new examples

mendesmiguel:
----

### DIFF
--- a/pages/common/cat.md
+++ b/pages/common/cat.md
@@ -17,3 +17,11 @@
 - Number all output lines:
 
 `cat -n {{file}}`
+
+- Display "$" at end of each line.
+
+`cat -E {{file}}`
+
+- Display TAB characters as ^I.
+
+`cat -T {{file}}`
--- a/pages/common/cat.md
+++ b/pages/common/cat.md
@@ -18,10 +18,10 @@
 
 `cat -n {{file}}`
 
-- Display "$" at end of each line.
+- Display "$" at end of each line:
 
 `cat -E {{file}}`
 
-- Display TAB characters as ^I.
+- Display TAB characters as ^I:
 
 `cat -T {{file}}`


sbrl:
Hey - thanks for the edit, @mendesmiguel! Looks like you're having some issues with the linter. We've got the style we prefer documented over [here](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#markdown-format), and a few good examples are the pages for `tar`, `eyeD3`, and `ln`. If you need any assistance, just ask :smiley_cat:

pages/common/cat.md
```diff
@@ -17,3 +17,11 @@
 - Number all output lines:
 
 `cat -n {{file}}`
+
+- Display "$" at end of each line:
+
+`cat -E {{file}}`
+
+- Display TAB characters as ^I:
```
agnivade: What is the usefulness of this ? I can still understand why `-E` may be useful. But `-T` just seems like a corner case to me.

pages/common/cat.md
```diff
@@ -18,10 +18,10 @@
 
 `cat -n {{file}}`
 
-- Display "$" at end of each line.
+- Display "$" at end of each line:
 
 `cat -E {{file}}`
 
-- Display TAB characters as ^I.
```
rhoio: On MacOS X I get:
cat: illegal option -- E
cat: illegal option -- T

Those two options seem to work only for the linux cat command.

sbrl: Maybe we should have a macOS-specific `cat` page?

rhoio: Has this issue not come up before? There are other commands where options slightly differ depending on the OS (sed, ps and grep come to my mind). Best way IMO would be to restrict examples to those which works on all systems.

agnivade: We have platform specific folders. Any pages which are different in other platforms should have pages along with those options in those folders.

sbrl: Yeah, it has come up a time or 2 before. We usually solve it by copying said page to the platform-specific folder and adjusting that copy instead, since a client will always prefer a platform-specific version of a page over a `common/` one or one for another platform.

PR #2528, opened 2018-10-30
jest: add page

huntie:
Create entry for the [Jest](https://jestjs.io/) JavaScript testing platform.

----

### DIFF
--- a/pages/common/jest.md
+++ b/pages/common/jest.md
@@ -1,0 +1,23 @@
+# jest
+
+> A zero-configuration JavaScript testing platform.
+
+- Run all tests:
+
+`jest`
+
+- Run only tests with a name that matches the regex pattern:
+
+`jest -t {{spec_name}}`
+
+- Run only tests relating to uncommitted files:
+
+`jest --onlyChanged`
+
+- Watch files for changes and rerun related tests:
+
+`jest --watch`
+
+- Show help:
+
+`jest --help`


sbrl:
Hey - thanks for the new page, @huntie! Looks like you're having some issues with the linter. We've got the style we prefer documented over [here](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#markdown-format), and a few good examples are the pages for `tar`, `eyeD3`, and `ln`. If you need any assistance, just ask :smiley_cat:

pages/common/jest.md
```diff
@@ -0,0 +1,23 @@
+# jest
+
+> Delightful JavaScript Testing.
```
agnivade: Please change this to something more objective, like - "A framework for testing javascript code" or something similar.

pages/common/jest.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Run all tests:
+
+    jest
+
+- Run tests filtered by spec name:
+
+    jest -t name-of-spec
```
agnivade: spec_name

pages/common/jest.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Run tests relating to uncommitted files:
+
+    jest -o
+
+- Run using interactive watch mode to rerun tests for changed files:
```
agnivade: How about - "Automatically re-run tests on changes to files."

pages/common/jest.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Run all tests:
+
+    jest
+
+- Run tests filtered by spec name:
+
+    jest -t name-of-spec
+
+- Run tests relating to uncommitted files:
```
agnivade: Is there an expanded version of the flag ? `-o` does not match with uncommitted.

huntie: Yeah, it's `--onlyChanged`. Do we have a convention for specifying expanded flags alongside the short version?

pages/common/jest.md
huntie: That would be shorter, but I felt it was more descriptive to indicate that this command runs with a REPL. From `jest --help`:

> Watch files for changes and rerun tests related to changed files.

pages/common/jest.md
huntie: There's no precedence for this formatting in the [Jest CLI Options](https://jestjs.io/docs/en/cli) docs. The `jest --help` version however is probably more precise:

> Run only tests with a name that matches the regex pattern.

pages/common/jest.md
huntie: This is, as far as I can tell, the project tagline, but I appreciate it is highly subjective. Will look into authoring a concise description. 👍

huntie:
Thanks for the feedback, will address in the next couple of days.

Any reason that this project doesn't use Markdown code blocks? Advantages:

- Semantically a multiline block of code separate to a paragraph, as opposed to an inline symbol or snippet.
- The indented syntax can be used as I have done without use of backticks, relevant where this project does not highlight code of a particular language.
- Non-ambiguous paragraph whitespace handling (albeit web only).

pages/common/jest.md
agnivade: We always try to explain a command in our own way. Otherwise, it would be just another man page. In all our pages, we try to keep the descriptions as objective as possible.

pages/common/jest.md
agnivade: Sorry, I don't understand. You wrote "name-of-spec". Since our pages follow the snake case convention, I suggested spec_name, which also matches tokens like package_name in other pages.

If you change the description to regex, you can change the token to a concrete value like "booking*"

pages/common/jest.md
agnivade: I suggested this to match with other pages we have (webpack). Wherever possible, we strive to remain consistent throughout the pages. If we can tweak an example to be more lucid in its meaning, then we should use that by all means. If we just copy over the documentation, then it just becomes another man page copy :)

But if you want to keep that wording, I suggest - "Watch files for changes and re-run the tests in those files". Just avoided using "change" twice.

pages/common/jest.md
agnivade: Yes, we do.  https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines pt. 4.

The heuristic is, keep the single word flag if it is clear enough or there is a word in the description starting with it. Otherwise, if there is a spelled out version, use that.

agnivade:
> Any reason that this project doesn't use Markdown code blocks?

No specific reason. We already have an open issue to switch to a new format. But the challenge is _all_ pages need to be converted to the new format, and _all_ clients need to be updated. Also, we need a period where both versions of pages are supported, to prevent breaking older clients for other folks.

pages/common/jest.md
huntie: Ah right, I misunderstood. Will change for the command example 👍

huntie:
I've amended my commit including all changes just now. Thanks for the detailed responses – let me know if I should address anything else.

pages/common/jest.md
```diff
@@ -0,0 +1,23 @@
...
+
+`jest`
+
+- Run only tests with a name that matches the regex pattern:
+
+`jest -t {{spec_name}}`
+
+- Run only tests relating to uncommitted files:
+
+`jest --onlyChanged`
```
huntie: What I really would like to do here is show both the full and shorthand versions of this option, say with `jest -o` as the second line, but couldn't find examples like this to follow.

sbrl: Yeah. We generally prefer the long options only in tldr-pages. I think the idea is that you can then look up the shortcuts as a more advanced user if you find yourself typing an option a lot.

Also, please could you avoid squashing your commits? It make it harder to see what you've changed, and we've got a special button that does it for us at merge-time :-)

huntie: I amended the original commit since it was a direct adjustment to the feature I was adding, that has not yet reached your stable branch. It is also in my clone repository 😛.

When merged I will delete this clone anyway – so given you will squash changes on merge this history is lost anyway except in the history of this PR on GitHub.

agnivade: The point is it makes it difficult for us to review your PR.

huntie:
Withstanding the Git workflow comments above, is this branch good to merge?

PR #2533, opened 2018-10-30
chromium: add page

BinaryBrain:
----

### DIFF
--- a/pages/common/chromium.md
+++ b/pages/common/chromium.md
@@ -1,0 +1,27 @@
+# chromium
+
+> Open-source web browser from Google.
+
+- Open a file:
+
+`chromium {{file.html}}`
+
+- Open an URL:
+
+`chromium {{example.com}}`
+
+- Open in incognito mode:
+
+`chromium --incognito {{example.com}}`
+
+- Open in a new window:
+
+`chromium --new-window {{example.com}}`
+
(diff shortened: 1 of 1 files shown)


pages/common/chromium.md
```diff
@@ -0,0 +1,27 @@
+# chromium
+
+> Open-source web browser from Google.
+
+- Open a file:
+
+`chromium {{file.html}}`
```
sbrl: Could `file.html` be `path/to/file.html` please?

pages/common/chromium.md
```diff
@@ -0,0 +1,27 @@
...
+
+`chromium --new-window {{example.com}}`
+
+- Open in app mode (without toolbars, URL bar, buttons, etc.):
+
+`chromium --app='{{https://example.com}}'`
+
+- Use a proxy server:
+
+`chromium --proxy-server="{{socks5://foobar:66}}" {{example.com}}`
```
agnivade: please change foobar to hostname.

BinaryBrain:
@agnivade Because the protocol is optional everywhere except with the `--app` option apparently.
So if we add the protocol everywhere as you suggest, the use may not know it is optional.

agnivade:
Thanks for clarifying !

PR #2539, opened 2018-10-31
ansiweather: add page

AnimiVulpis:
----

### DIFF
--- a/pages/common/ansiweather.md
+++ b/pages/common/ansiweather.md
@@ -1,0 +1,15 @@
+# ansiweather
+
+> AnsiWeather is a Shell script for displaying the current weather conditions in your terminal.
+
+- Display forecast using metric units for the next five days for Rzeszow, Poland:
+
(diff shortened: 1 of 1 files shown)


pages/common/ansiweather.md
```diff
@@ -0,0 +1,15 @@
+# ansiweather
+
+> AnsiWeather is a Shell script for displaying the current weather conditions in your terminal.
```
sbrl: Not sure we need `AnsiWeather is` here. Also, I don't think `Shell` should be capitalised.

pages/common/ansiweather.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Display forecast using metric units for the next five days for Rzeszow, Poland:
```
sbrl:
```suggestion
- Display a forecast using metric units for the next five days for Rzeszow, Poland:
```

Perhaps this will read better? Same goes for the other examples.

PR #2540, opened 2018-10-31
brightness: add page

AnimiVulpis:
----

### DIFF
--- a/pages/osx/brightness.md
+++ b/pages/osx/brightness.md
@@ -1,12 +1,12 @@
 # brightness
 
-> Set and obtain the brightness level of all internal and certain external displays.
+> Get and set the brightness level of all internal and certain external displays.
 
-- Set 100% brightness:
+- Set the brightness to 100%::
 
 `brightness {{1}}`
 
-- Set 50% brightness:
+- Set the brightness to 50%::
 
 `brightness {{0.5}}`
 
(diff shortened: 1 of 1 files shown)


pages/osx/brightness.md
```diff
@@ -0,0 +1,15 @@
+# brightness
+
+> Set and obtain the brightness level of all internal and certain external displays.
+
+- Set 100% brightness:
```
sbrl: Perhaps this will read better?

```suggestion
- Set the brightness to 100%:
```

The same goes for the next example down.

pages/osx/brightness.md
```diff
@@ -0,0 +1,15 @@
...
+
+`brightness {{1}}`
+
+- Set 50% brightness:
+
+`brightness {{0.5}}`
+
+- Show current brightness:
```
agnivade: This should be the first example

pages/osx/brightness.md
```diff
@@ -0,0 +1,15 @@
+# brightness
+
+> Set and obtain the brightness level of all internal and certain external displays.
```
agnivade: Maybe "Get" instead of obtain ?

So then it should be - "Get and set .."

pages/osx/brightness.md
AnimiVulpis: You're right. That sounds better.

pages/osx/brightness.md
AnimiVulpis: Sounds easier to understand which is desirable ^^

PR #2543, opened 2018-10-31
lorem: add page

AnimiVulpis:
----

### DIFF
--- a/pages/common/lorem.md
+++ b/pages/common/lorem.md
@@ -1,0 +1,19 @@
+# lorem
+
+> Create more or less random lorem ipsum text.
+
+- Print the number of words you want:
+
+`lorem -n {{20}}`
+
+- Print 10 lines of Goethe's Faust:
+
+`lorem -l {{10}} --faust`
+
+- Print 5 sentences of Poe's Raven:
+
+`lorem -s {{5}} --raven`
+
+- Print random words from Boccaccio's Decameron for 40 characters:
+
+`lorem --decamerone --randomize -c {{40}}`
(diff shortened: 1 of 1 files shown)


pages/common/lorem.md
```diff
@@ -0,0 +1,19 @@
+# lorem
+
+> Create more or less random lorem ipsum text.
+
+- Print the number of words you want:
```
sbrl: Generally, we try to avoid using `you` in tldr-pages. Perhaps we can rephrase this?
```

Comment (pxgamer) [reply pages/common/lorem.md]: Perhaps:
```suggestion
- Print the specified number of words:
```

pages/common/lorem.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Print 10 lines of Goethe's Faust:
+
+`lorem -l {{10}} --faust`
+
+- Print 5 sentences of Poe's Raven:
+
+`lorem -s {{5}} --raven`
+
+- Print random words from Boccaccio's Decameron for 40 characters:
```
agnivade: I would keep it consistent with the previous descriptions - 

"Print 40 random characters from Boccacio's Decameron".

pages/common/lorem.md
```diff
@@ -0,0 +1,19 @@
...
+
+`lorem --decamerone --randomize -c {{40}}`
```
agnivade: Let us move `decamerone` to the end.

pages/common/lorem.md
AnimiVulpis: My idea was to show that the order options does not matter.
But the inconsistency might be more confusing than helping.

PR #2544, opened 2018-10-31
noti: add page

AnimiVulpis:
----

### DIFF
--- a/pages/common/noti.md
+++ b/pages/common/noti.md
@@ -1,0 +1,15 @@
+# noti
+
+> Monitor a process and trigger a notification.
+
+- Display a notification when tar finishes compressing files:
+
+`noti {{tar -cjf example.tar.bz2 example/}}`
+
+- Display a notification even when you put it after the command to watch:
+
+`{{command_to_watch}}; noti`
+
+- Monitor a process by PID and trigger a notification when the pid disappears:
+
+`noti -w {{process_id}}`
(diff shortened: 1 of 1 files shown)


pages/common/noti.md
```diff
@@ -0,0 +1,15 @@
+# noti
+
+> Monitor a process and trigger a notification.
```
sbrl: - Is it a graphical notification, or a textual one?
 - Does it show only when the command finishes executing?

pages/common/noti.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Display a notification when tar finishes compressing files:
+
+`noti {{tar -cjf example.tar.bz2 example/}}`
+
+- Display a notification even when you put it after the command to watch:
+
+`{{command_to_watch}}; noti`
+
+- Monitor a process by PID and trigger a notification when the pid disappears:
```
agnivade: The latter PID should be in caps.

pages/common/noti.md
AnimiVulpis: Yes, it's graphical, using the default OS notification banner/popup

As far as I know it's supposed indicate a finished process, but I am pretty sure you could somehow plug it in the middle of a chain of commands (with some shell magic like `tee` perhaps?).

Should that, or parts of that, be mentioned in the description?

agnivade: Maybe not, since this is a tldr page. The basic usage should be fine.

AnimiVulpis: Then we should leave it like that

PR #2547, opened 2018-11-01
psgrep: add page

alex0112:
----

Note:
New to the whole forking system of creating PRs.  Sorry about the extra commits.

### DIFF
--- a/pages/common/psgrep.md
+++ b/pages/common/psgrep.md
@@ -6,7 +6,7 @@
 
 `psgrep {{'foo'}}`
 
-- Find process lines containing a specific string, do not include headers:
+- Find process lines containing a specific string excluding headers:
 
 `psgrep -n {{'foo'}}`
 
--- a/pages/common/psgrep.md
+++ b/pages/common/psgrep.md
@@ -1,0 +1,15 @@
+# psgrep
+
+> Search through processes with grep.
+
+- Find process lines containing a specific string:
+
+`psgrep {{'foo'}}`
+
+- Find process lines containing a specific string, do not include headers:
+
+`psgrep -n {{'foo'}}`
+
+- Search using simplified format (PID, user, command):
+
+`psgrep -s {{'foo'}}`
--- a/pages/common/scheme.md
+++ b/pages/common/scheme.md
@@ -1,6 +1,6 @@
 # scheme
 
-> MIT Scheme language interperter and REPL (interactive shell).
+> MIT Scheme language interpreter and REPL (interactive shell).
 
 - Open an interactive shell (REPL):
 
--- a/pages/common/scheme.md
+++ b/pages/common/scheme.md
@@ -1,0 +1,23 @@
+# scheme
+
+> MIT Scheme language interperter and REPL (interactive shell)
+
+- Open an interactive shell (REPL):
+
+`scheme`
+
+- Run a scheme program (with no REPL output):
+
+`scheme --quiet < script.scm`
+
+- Load a scheme program into the REPL:
+
+`scheme --load {{script.scm}}`
+
+- Load scheme expressions into the REPL:
+
+`scheme --eval "(define foo 'x)"`
+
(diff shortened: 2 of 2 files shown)


pages/common/psgrep.md
```diff
@@ -0,0 +1,15 @@
+# psgrep
+
+> Search through processes with grep.
+
+- Find process lines containing a specific string:
+
+`psgrep {{'foo'}}`
+
+- Find process lines containing a specific string, do not include headers:
```
pxgamer: Perhaps this would read better:
```suggestion
- Find process lines containing a specific string excluding headers:
```

pages/common/psgrep.md
```diff
@@ -0,0 +1,15 @@
+# psgrep
```
sbrl: Did you mean `pgrep`? `psgrep` isn't a thing on my Linux laptop here, but that could just be my distribution.

agnivade:
Can you show me the man page for psgrep ? I cannot find any

alex0112:
@agnivade   
```
PSGREP(1)                        psgrep Manual                       PSGREP(1)

NAME
       psgrep - search the process list using the power of grep

SYNOPSIS
       psgrep [options] PATTERN

DESCRIPTION
       psgrep  is a small Bash shell script that searches the process list (as
       obtained by ps(1) ) using the awesome utility grep(1)  for  its  power.
       It  is  more  useful  than  pgrep(1) because not only can it search the
       process list and return a PID, it can give more useful information such
       as  its  UID,  GID,  memory/CPU usage, niceness, and anything else sup-
       ported by ps.  It is also more useful than using something like
              ps aux | grep PATTERN
       because it also prevents the matching of the ps or grep processes  used
       for  searching the list.  In fact, the main reasoning behind the origi-
       nal writing of this script was entirely due to this annoyance with  the
       given method.

OPTIONS...
```

What distro are you on?

pages/common/psgrep.md
alex0112: `pgrep` is a separate command from `psgrep`.  To quote the manpage: 

```
psgrep  is  a small Bash shell script that searches the process list (as obtained by ps(1) ) using the awesome utility grep(1) for its power.  It is more
       useful than pgrep(1) because not only can it search the process list and return a PID, it can give more useful information such as  its  UID,  GID,  mem-
       ory/CPU usage, niceness, and anything else supported by ps.
```

alex0112: I'm running MacOS High Sierra.  The manpage doesn't specify that this is for OSX but, your comment and @agnivade makes me question.

pages/common/psgrep.md
alex0112: I like this suggestion.

agnivade:
Then this should be in the macos folder, not common. I am on ubuntu.

agnivade:
Oh I guess this is installable everywhere. Just that this wasn't available by default in Ubuntu.

pages/common/psgrep.md
```diff
@@ -0,0 +1,15 @@
...
+
+> Search through processes with grep.
```
agnivade: grep in backticks

pages/common/psgrep.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Find process lines containing a specific string:
+
+`psgrep {{'foo'}}`
```
agnivade: Are the single quotes necessary ? And I would rather have process_name than foo to match with the pgrep page.

alex0112:
Apologies for the lack of activity.  I mistakenly branched off of another feature branch.  I'm going to delete this PR and resubmit with a clean history.

pages/common/psgrep.md
alex0112: It doesn't appear that the quotes are necessary.  I can remove them.
`process_name` is fine with me.

alex0112:
See #2607 for the cleaned up version

alex0112:
See #2609 for the clean version

PR #2548, opened 2018-11-01
xfce4-terminal: add page

sbrl:
----

### DIFF
--- a/pages/linux/xfce4-terminal.md
+++ b/pages/linux/xfce4-terminal.md
@@ -6,6 +6,10 @@
 
 `xfce4-terminal`
 
+- Set the initial title:
+
+`xfce4-terminal --initial-title "{{initial_title}}"
+
 - Open a new tab in the current terminal window:
 
 `xfce4-terminal --tab`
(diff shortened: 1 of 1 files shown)


pages/linux/xfce4-terminal.md
```diff
@@ -0,0 +1,27 @@
...
+
+`xfce4-terminal --tab`
+
+- Execute a command in a new terminal window:
+
+`xfce4-terminal --command {{command_with_args}}`
+
+- Keep the terminal around after the executed command finishes executing:
+
+`xfce4-terminal --command {{command}} --hold`
```
pxgamer: In the example above, the token name is `command_with_args`, does that mean this example doesn't support arguments? 🙂

sbrl: Ah, good point! I was actually fiddling around with the difference between `--execute` and `--command` here. Turns out that with `--command` it can be used multiple times, but you need to surround everything in quotes. With `--execute` it says "ignore any other arguments and treat the rest of the command line as a command I should execute".

So both do indeed support arguments, but I haven't represented that here. Thanks :D

PR #2551, opened 2018-11-02
tslint: add page

mstruebing:
----
see: https://github.com/palantir/tslint

### DIFF
--- a/pages/common/tslint.md
+++ b/pages/common/tslint.md
@@ -6,7 +6,7 @@
 
 `tslint --init`
 
-- Lint on files:
+- Lint on a given set of files:
 
 `tslint {{filename}}.js {{filename1}}.js`
 
--- a/pages/common/tslint.md
+++ b/pages/common/tslint.md
@@ -1,0 +1,19 @@
+# tslint
+
+> A pluggable linting utility for TypeScript.
+
+- Create tslint config:
+
+`tslint --init`
+
+- Lint on files:
+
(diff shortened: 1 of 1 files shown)


pages/common/tslint.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Lint on files:
+
+`tslint {{filename}}.js {{filename1}}.js`
+
+- Fix lint issues:
+
+`tslint --fix`
+
+- Lint with config in project root:
```
sbrl: Is this a config file in a specific directory, or a specific config file?

mstruebing: The config `tslint.json` needs to be in the project root path. Similiar to others `--config {{path/to/config}}` but there you need to specify the root path where the config is located.

pages/common/tslint.md
```diff
@@ -0,0 +1,19 @@
+# tslint
+
+> A pluggable linting utility for TypeScript.
+
+- Create tslint config:
+
+`tslint --init`
+
+- Lint on files:
```
agnivade: Let's make the same change as `xo` and this should be good to go.

Lint on a given set of files.

pages/common/tslint.md
sbrl: Ah, I see. Perhaps this will read better then?

```suggestion
- Lint with the config file in the project root:
```

PR #2552, opened 2018-11-02
xo: add page

mstruebing:
----
see: https://github.com/xojs/xo

### DIFF
--- a/pages/common/xo.md
+++ b/pages/common/xo.md
@@ -1,0 +1,23 @@
+# xo
+
+> A pluggable zero configuration linting utility for JavaScript.
+
+- Lint files in `src`:
+
+`xo`
+
+- Lint on files:
+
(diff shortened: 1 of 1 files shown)


pages/common/xo.md
```diff
@@ -0,0 +1,23 @@
+# xo
+
+> A pluggable zero configuration linting utility for JavaScript.
+
+- Lint files in `src`:
```
agnivade: src should be in double quotes. Backticks should be only for commands.

pages/common/xo.md
```diff
@@ -0,0 +1,23 @@
...
+
+`xo`
+
+- Lint on files:
```
agnivade: on _a given set of_ files

mstruebing: I've used the `eslint` page as a template but okay, will change that.

PR #2557, opened 2018-11-05
csc: add page

sbrl:
I bring you a page for the esteemed the C&sharp; Compiler! Believe it or not, this is now multi-platform with recent versions of _Mono_.

----

### DIFF
--- a/pages/common/csc.md
+++ b/pages/common/csc.md
@@ -20,7 +20,7 @@
 
 - Embed a resource:
 
-`csc /resource:{{path/to/input_file.cs}},{{namespace.string.name}}`
+`csc /resource:{{path/to/target_file.cs}},{{namespace.string.name}} {{path/to/input_file.cs}}`
 
 - Automatically generate XML documentation:
 
--- a/pages/common/csc.md
+++ b/pages/common/csc.md
@@ -2,7 +2,7 @@
 
 > The Microsoft C# Compiler.
 
-- Compile one or more C# files to a CIL executable
+- Compile one or more C# files to a CIL executable:
 
 `csc {{path/to/input_file_a.cs}} {{path/to/input_file_b.cs}}` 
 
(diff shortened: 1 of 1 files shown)


pages/common/csc.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Automatically generate XML documentation:
+
+`csc /doc:{{path/to/output.xml}} {{path/to/input_file.cs}}`
+
+- Specify an icon:
+
+`csc /win32icon:{{path/to/icon.ico}} {{path/to/input_file.cs}}`
+
+- Strongly-name the resulting assembly with a keyfile:
```
pxgamer: What does `Strongly-name` mean here? 🙂

pages/common/csc.md
```diff
@@ -0,0 +1,35 @@
...
+
+`csc /target:library {{path/to/input_file.cs}}`
+
+- Reference another assembly:
+
+`csc /reference:{{path/to/library.dll}} {{path/to/input_file.cs}}`
+
+- Embed a resource:
+
+`csc /resource:{{path/to/input_file.cs}},{{namespace.string.name}}`
```
agnivade: Is this missing path/to/input_file.cs ?

pages/common/csc.md
sbrl: Strong-naming is a concept specific to .NET assemblies in which the IL in an assembly is hashed, and then the hash encrypted with a private key. The public key is then bundled along with the assembly. In this manner, it can be assured that the code inside a .NET assembly has not been altered since it was built.

Combined with an install-time verification system such as `signtool.exe` (Code-Signing certificates; Windows) or GPG in a package manager (Linux), a full chain of trust can be established to ensure that the executing IL has both not been altered since build time, and was built by a trusted party.

pages/common/csc.md
sbrl: Yep! Thanks :P

pages/common/csc.md
pxgamer: Cheers for the explanation. :smile:

sbrl: np!

PR #2558, opened 2018-11-05
monodis: add page

sbrl:
As with #2557, this was very difficult to keep it down to just 8 examples. For both, I've tried to pick the most common options and present them in a form that's logic to one's workflow. Particularly with `csc` though it's unlikely that you'll be interfacing with it directly (since `msbuild` does that for you!) unless you've got a really odd setup.

----

### DIFF
--- a/pages/common/monodis.md
+++ b/pages/common/monodis.md
@@ -1,0 +1,31 @@
+# monodis
+
+> The Mono Common Intermediate Language (CIL) disassembler.
+
+- Disassemble an assembly to textual CIL:
+
+`monodis {{path/to/assembly.exe}}`
+
+- Save the output to a file:
+
+`monodis --output={{path/to/output.il}}`
+
+- Show information about an assembly:
+
+`monodis --assembly {{path/to/assembly.dll}}`
+
+- List the references of an assembly:
+
+`monodis --assemblyref {{path/to/assembly.exe}}`
+
+- List all the methods in an assembly:
+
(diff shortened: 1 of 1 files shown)


pages/common/monodis.md
```diff
@@ -0,0 +1,31 @@
...
+
+> The Mono Common Intermediate Language (CIL) disassembler.
+
+- Disassemble an assembly to textual CIL:
+
+`monodis {{path/to/assembly.exe}}`
+
+- Save the output to a file:
+
+`monodis --output={{path/to/output.il}}`
```
agnivade: path/to/assembly.exe missing ?

pages/common/monodis.md
```diff
@@ -0,0 +1,31 @@
+# monodis
+
+> The Mono Common Intermediate Language (CIL) disassembler.
+
+- Disassemble an assembly to textual CIL:
```
agnivade: Should this be "Disassembly a binary" ? And same below where `exe`s are used.

pages/common/monodis.md
sbrl: Good catch!

pages/common/monodis.md
sbrl: Indeed they are binaries, but `assembly` is the technical .NET term for a _Common Intermediate Language_ binary.

sbrl:
I think you're broken, @stale bot.

PR #2560, opened 2018-11-07
cradle: add page

pxgamer:
The command line for the [Cradle Framework](https://cradlephp.github.io).

----

### DIFF
--- a/pages/common/cradle.md
+++ b/pages/common/cradle.md
@@ -5,7 +5,7 @@
 
 - Connect to a server:
 
-`cradle connect {{connection_name}}`
+`cradle connect {{server_name}}`
 
 - Call general help:
 
--- a/pages/common/cradle.md
+++ b/pages/common/cradle.md
@@ -1,0 +1,20 @@
+# cradle
+
+> The Cradle PHP framework.
+> See `cradle-install`, `cradle-deploy` and other pages for additional information.
+
+- Connect to a server:
+
+`cradle connect {{connection_name}}`
+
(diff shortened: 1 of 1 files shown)


pages/common/cradle.md
```diff
@@ -0,0 +1,20 @@
+# cradle
+
+> The Cradle PHP framework.
+> See `cradle-install`, `cradle-deploy` and other pages for additional information.
+
+- Connect to a server:
+
+`cradle connect {{connection_name}}`
```
agnivade: I am guessing this is something like "user:password@hostname:port" ? Can we write something like that to make it clearer ?

pxgamer: It's just something like `app` or whatever you want it to be called. The details for it are specified in the `config/deploy.php` file.

https://github.com/CradlePHP/Cradle/blob/master/config/deploy.php

pages/common/cradle.md
```diff
@@ -0,0 +1,20 @@
+# cradle
+
+> The Cradle PHP framework.
+> See `cradle-install`, `cradle-deploy` and other pages for additional information.
+
+- Connect to a server:
+
+`cradle connect {{server_name}}`
+
+- Call general help:
```
sbrl: `Call` sounds rather odd here. What about `Show` instead? Same for the next example down.

pxgamer: Yeah, I don't normally add the `help` commands to tldr pages. But I was looking at `git` for the format.

PR #2564, opened 2018-11-07
asciinema: add page

waldyrious:
This page is heavily based on the excellent output of `asciinema -h`.

### DIFF
--- a/pages/common/asciinema.md
+++ b/pages/common/asciinema.md
@@ -1,0 +1,35 @@
+# asciinema
+
+> Record and replay terminal sessions, and optionally share them on asciinema.org.
+
+- Associate the local install of asciinema with an asciinema.org account:
+
+`asciinema auth`
+
+- Make a new recording (to be uploaded or saved locally):
+
+`asciinema rec`
+
+- Make a new recording and save it to a local file:
+
+`asciinema rec {{path/to/file}}.cast`
+
+- Replay a terminal recording from a local file:
+
+`asciinema play {{path/to/file}}.cast`
+
+- Replay a terminal recording hosted on asciinema.org:
+
+`asciinema play https://asciinema.org/a/{{cast_id}}`
+
+- Make a new recording, limiting any idle time to at most 2.5 seconds:
+
+`asciinema rec -i 2.5`
+
+- Print the full output of a locally saved recording:
+
+`asciinema cat {{path/to/file}}.cast`
+
+- Upload a locally saved terminal session to asciinema.org:
+
+`asciinema upload {{path/to/file}}.cast`
--- a/pages/common/asciinema.md
+++ b/pages/common/asciinema.md
@@ -2,7 +2,7 @@
 
 > Record and replay terminal sessions, and optionally share them on asciinema.org.
 
-- Associate the local install of asciinema with an asciinema.org account:
+- Associate the local install of `asciinema` with an asciinema.org account:
 
 `asciinema auth`
 
--- a/pages/common/asciinema.md
+++ b/pages/common/asciinema.md
@@ -6,7 +6,7 @@
 
 `asciinema auth`
 
-- Make a new recording (to be uploaded or saved locally):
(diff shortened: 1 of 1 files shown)


pages/common/asciinema.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Replay a terminal recording hosted on asciinema.org:
+
+`asciinema play https://asciinema.org/a/{{cast_id}}`
+
+- Make a new recording, limiting any idle time to at most 2.5 seconds:
+
+`asciinema rec -i 2.5`
+
+- Print the full output of a locally saved recording:
```
sbrl: So is this like printing a transcript?

pages/common/asciinema.md
```diff
@@ -0,0 +1,35 @@
...
+
+`asciinema play {{path/to/file}}.cast`
+
+- Replay a terminal recording hosted on asciinema.org:
+
+`asciinema play https://asciinema.org/a/{{cast_id}}`
+
+- Make a new recording, limiting any idle time to at most 2.5 seconds:
+
+`asciinema rec -i 2.5`
```
agnivade: 2.5 in tokens

pages/common/asciinema.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Associate the local install of `asciinema` with an asciinema.org account:
+
+`asciinema auth`
+
+- Make a new recording (to be uploaded or saved locally):
+
+`asciinema rec`
+
+- Make a new recording and save it to a local file:
```
agnivade: So where does it save when no file params are passed ?

pages/common/asciinema.md
waldyrious: I'm getting a little rusty it seems :)

pages/common/asciinema.md
waldyrious: It asks when the recording is ended. Instructions are provided interactively to start, stop and save/upload the cast in the relevant moments of the capture process.

Do you think there's a way to make this clearer (without making the line too long, that is)?

pages/common/asciinema.md
waldyrious: I suppose so. I haven't tried this one TBH :sweat_smile:. I imagine it will either print both the typed commands and their output, or just the command outputs. The former seems more useful, so I'd venture saying it's more likely (I can't test it ATM).

waldyrious:
> Good to see some PRs from you @waldyrious !

It's great to submit them! There's plenty more where these came from, time permitting. :)

pages/common/asciinema.md
waldyrious: I gave it a try, check it out :)

pages/common/asciinema.md
waldyrious: @sbrl just confirmed: this indeed prints both the commands and the output, with formatting and everything. So it's basically an instantaneous version of the `play` subcommand.

Do you think the command description could be changed to make this clearer?

waldyrious:
It seems like most of the subcommands accept either a local file or an URL. Perhaps we could tweak the tokens to make this move evident. Thoughts?

pages/common/asciinema.md
sbrl: Hrm, now that you mention it, I think this is clear enough :-)

sbrl:
It might also be worth including an example where you execute a custom command. Use-case: I want to record a demo of a program I've written and include it in a [blog post](https://starbeamrainbowlabs.com/blog/article.php?article=posts%2F325-AI-Microsoft-Text-Recognizers.html).

waldyrious:
@sbrl I'm not sure I understand exactly what you mean by custom command. Can you expand a bit?

sbrl:
Sure. For example, I've written an awesome _nyan cat_ program that I want to show off. I don't want to reveal my username / hostname, so I use `asciinema rec -c '{{command}}'` to get asciinema to start the program directly instead of my shell.

PR #2565, opened 2018-11-07
svn: add `svn log` example, remove unnecessary `svn help` example

Calinou:

The `svn help` example isn't of much use, so I replaced it with a more useful example which describes basic usage of `svn log`.

### DIFF
--- a/pages/common/svn.md
+++ b/pages/common/svn.md
@@ -20,4 +20,8 @@
 
 - Display changes from the last 10 revisions, showing modified files for each revision:
 
-`svn log -vl 10`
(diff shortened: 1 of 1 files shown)


pages/common/svn.md
```diff
@@ -18,6 +18,6 @@
 
 `svn ci -m {{commit log message}} {{[PATH...]}}`
 
-- Show detailed help:
```
agnivade: We can keep both. And 10 should be in tokens (`{{}}`)

Calinou: Should be fixed now.

PR #2567, opened 2018-11-08
TravisCI: include pages of all languages in archive

arunisaac:
Now that we are supporting pages in multiple natural languages (#2339), we need to include all those pages in the tldr archive.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -25,7 +25,7 @@ function build_archive {
   rm -f $TLDR_ARCHIVE
 
   cd $TLDRHOME/
-  zip -r $TLDR_ARCHIVE pages/ pages.*/ LICENSE.md
+  zip -r $TLDR_ARCHIVE pages*/ LICENSE.md
 }
 
 function upload_assets {
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -25,7 +25,7 @@ function build_archive {
   rm -f $TLDR_ARCHIVE
 
   cd $TLDRHOME/
-  zip -r $TLDR_ARCHIVE pages/ LICENSE.md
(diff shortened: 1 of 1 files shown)


scripts/build.sh
```diff
@@ -25,7 +25,7 @@ function build_archive {
   rm -f $TLDR_ARCHIVE
 
   cd $TLDRHOME/
-  zip -r $TLDR_ARCHIVE pages/ LICENSE.md
+  zip -r $TLDR_ARCHIVE pages/ pages.*/ LICENSE.md
```
agnivade: Why not just `pages*/` ?

waldyrious: Does it even work with the `.` there, considering that the English pages directory does not have a suffix?

waldyrious:
@arunisaac just for reference, github's squash option does not automatically interpret the `fixup!` directive in the commit message. I had to manually remove the second commit's message when merging. It's certainly something that would be nice for them to support.

scripts/build.sh
pxgamer: @waldyrious, I assume it would in the initial one as it has `pages/ pages.*/` so that would catch both.

arunisaac:
Waldir Pimenta <notifications@github.com> writes:

> @arunisaac just for reference, github's squash option does not
> automatically interpret the `fixup!` directive in the commit message.

Yes, I suspected as much. In the future, should I use the "fixup!"
prefix? Or, is there some other preferred way for github to
automatically recognize commit messages to be squashed?

waldyrious:
It's up to you. Maintainers should ask in case they need git actions from the contributor, anyway. In this case, what needed to happen was pretty straightforward (and could be done in the merge process) so I just went ahead and did it.

PR #2571, opened 2018-11-09
csplit: add page

ronniegane:

### DIFF
--- a/pages/linux/csplit.md
+++ b/pages/linux/csplit.md
@@ -1,19 +1,19 @@
 # csplit
 
-> Split a file into pieces
+> Split a file into pieces.
 
-- Split a file at lines 5 and 23
+- Split a file at lines 5 and 23:
 
 `csplit {{file}} 5 23`
 
-- Split a file every 5 lines (-k flag necessary if not divisible by 5)
+- Split a file every 5 lines (-k flag necessary if not divisible by 5):
 
 `csplit -k {{file}} 5 {*}`
 
-- Split a file and use a given prefix for the output files
+- Split a file and use a given prefix for the output files:
 
 `csplit {{file}} 5 -f {{prefix}}`
 
-- Split a file at a line matching a regular expression
+- Split a file at a line matching a regular expression:
 
 `csplit {{file}} /{{regex}}/`
--- a/pages/linux/csplit.md
+++ b/pages/linux/csplit.md
@@ -7,7 +7,11 @@
 
 `csplit {{file}} {{5}} {{23}}`
 
-- Split a file every {{5}} lines (-k flag necessary if not divisible by {{5}}):
+- Split a file every {{5}} lines (this will fail if the total number of lines is not divisible by {{5}}):
+
+`csplit {{file}} {{5}} {*}`
+
+- Split a file every {{5}} lines, ignoring exact-division error:
 
 `csplit -k {{file}} {{5}} {*}`
 
--- a/pages/linux/csplit.md
+++ b/pages/linux/csplit.md
@@ -1,6 +1,7 @@
 # csplit
 
-> Split a file into pieces.
+> Split a file into pieces.  
+> This generates files named "xx00", "xx01", and so on.
 
 - Split a file at lines 5 and 23:
 
--- a/pages/linux/csplit.md
+++ b/pages/linux/csplit.md
@@ -3,17 +3,17 @@
 > Split a file into pieces.  
 > This generates files named "xx00", "xx01", and so on.
 
-- Split a file at lines 5 and 23:
+- Split a file at lines {{5}} and {{23}}:
 
-`csplit {{file}} 5 23`
+`csplit {{file}} {{5}} {{23}}`
 
-- Split a file every 5 lines (-k flag necessary if not divisible by 5):
+- Split a file every {{5}} lines (-k flag necessary if not divisible by {{5}}):
 
-`csplit -k {{file}} 5 {*}`
+`csplit -k {{file}} {{5}} {*}`
 
-- Split a file and use a given prefix for the output files:
+- Split a file at line {{5}} and use a given prefix for the output files:
 
-`csplit {{file}} 5 -f {{prefix}}`
+`csplit {{file}} {{5}} -f {{prefix}}`
 
 - Split a file at a line matching a regular expression:
 
--- a/pages/linux/csplit.md
+++ b/pages/linux/csplit.md
@@ -3,19 +3,19 @@
 > Split a file into pieces.
 > This generates files named "xx00", "xx01", and so on.
 
-- Split a file at lines {{5}} and {{23}}:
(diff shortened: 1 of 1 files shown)


pages/linux/csplit.md
```diff
@@ -0,0 +1,19 @@
+# csplit
+
+> Split a file into pieces.
```
agnivade: We should add a sub-heading - "This generates files named "xx00", "xx01" and so on."

pages/linux/csplit.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Split a file at lines 5 and 23:
+
+`csplit {{file}} 5 23`
```
agnivade: 5 and 23 in tokens (`{{}}`) here and everywhere else.

pages/linux/csplit.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Split a file every 5 lines (-k flag necessary if not divisible by 5):
```
agnivade: We can have a separate example showing the importance of `-k` and not mix it with this example.

pages/linux/csplit.md
```diff
@@ -0,0 +1,19 @@
...
+
+`csplit -k {{file}} 5 {*}`
+
+- Split a file and use a given prefix for the output files:
```
agnivade: Split a file _at line 5_

pages/linux/csplit.md
ronniegane: Added in 4c464e0c3

pages/linux/csplit.md
ronniegane: I added an example in cd8321e15e8 but I'm not sure of the wording, it feels a bit clunky.

pages/linux/csplit.md
ronniegane: Clarified in 4c464e0c3c878

pages/linux/csplit.md
ronniegane: Added in 4aa995b31953

pages/linux/csplit.md
```diff
@@ -0,0 +1,24 @@
+# csplit
+
+> Split a file into pieces.
+> This generates files named "xx00", "xx01", and so on.
+
+- Split a file at lines {{5}} and {{23}}:
```
agnivade: 5 and 23 need not be inside `{{}}`. We use that only for examples.

pages/linux/csplit.md
```diff
@@ -0,0 +1,24 @@
...
+
+`csplit {{file}} {{5}} {{23}}`
+
+- Split a file every {{5}} lines (this will fail if the total number of lines is not divisible by {{5}}):
+
+`csplit {{file}} {{5}} {*}`
```
agnivade: Does * need to be inside {} ?

pages/linux/csplit.md
ronniegane: Removed from this line and the other example descriptions in fe8cd972

agnivade:
@waldyrious - Could you take a look at this ? I believe it can be improved slightly. But I don't have anything concrete to suggest.

waldyrious:
I think the page is well-written, but the command itself is not very intuitive, so only by reading these descriptions it's unclear what will actually happen. Several things puzzle me about it:

- Are the splitted files literally named xx00 and so on, or is the xx a placeholder for the original filename?
- If the prefix is indeed always xx, what happens if I split a second file in the same directory? Does the counter keep incrementing globally? How do I know what pieces belong to each original file?
- How is the division performed? Do the given lines become part of the piece above them, or the one below?
- As I commented inline: for the regex option, what happens if more than one line matches the regex? And if none does?

I hope I can help with making specific proposals to clarify the page if these issues are clarified.

pages/linux/csplit.md
```diff
@@ -0,0 +1,24 @@
...
+
+`csplit {{file}} {{5}} {{23}}`
+
+- Split a file every 5 lines (this will fail if the total number of lines is not divisible by 5):
+
+`csplit {{file}} {{5}} {*}`
+
+- Split a file every 5 lines, ignoring exact-division error:
+
+`csplit -k {{file}} {{5}} {*}`
```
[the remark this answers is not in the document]
ronniegane: yes, it is `--keep-files`. Should I write that in full in the code line, the description line, or both?

ronniegane:
@waldyrious I'll just test these examples
```
➜  example ls
example.txt
➜  example cat example.txt 
one
two
three
four
five
six
seven
eight
nine
ten
eleven
twelve
➜  example csplit example.txt 5
19
44
➜  example ls
example.txt  xx00  xx01
➜  example cat xx00
one
two
three
four
➜  example cat xx01
five
six
seven
eight
nine
ten
eleven
twelve
```
Splitted files are literally named `xx00`, unless the custom prefix flag is used.  
Matching lines become part of the piece _below_ them (line 5 is in xx01)
```
➜  example ls
example2.txt  example.txt  xx00  xx01
➜  example csplit example2.txt 8
37
16
➜  example cat xx00
uno
dos
tres
cuatro
cinco
seis
siete
```
Splitting again overwrites the existing `xx00` etc files. No warning or error is given.
```
➜  example csplit example.txt /s/
24
39
➜  example cat xx00
one
two
three
four
five
➜  example cat xx01
six
seven
eight
nine
ten
eleven
twelve
```
Using a regex stops at the very first match (first line with "s" in it in this case), unless the `{INTEGER}` or `{*}` keywords are used after the regex pattern
```
➜  example csplit example.txt /s/ {\*}
24
4
35
➜  example ls
example2.txt  example.txt  xx00  xx01  xx02
```
(I have to escape the `*` in my shell (zsh) to prevent globbing, but I'm not sure if that's common to other shells)
```
➜  example csplit example.txt /z/     
csplit: ‘/z/’: match not found
63
```
If there is no match for the regex, an error is thrown. No files will be created _unless_ the `--keep-files` flag is used, in which case it will just make one file xx00.

pages/linux/csplit.md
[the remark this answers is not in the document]
waldyrious: Hm, that's... unexpected. Let me see if I understand the semantics of that option name: without the `-k` the partial files are created, but then deleted once the final one generates an error? Otherwise, how do you explain the choice of name? It's not clear to me.

waldyrious:
@ronniegane many thanks for the detailed examples. Later today I'll try to propose reworded command descriptions that might make these behaviors more predictable.

waldyrious:
@ronniegane @agnivade unfortunately I became unexpectedly busy for the past few days. Perhaps it's better to merge the PR as-is, and I'll submit a change proposal afterwards.

agnivade:
SGTM

waldyrious:
Ok, I'll merge then. Sorry for the inconvenience, @ronniegane, and thanks again for the contribution and patience :)

PR #2572, opened 2018-11-09
Add OCaml client

RosalesJ:
----

Just add an OCaml client to the list of tldr clients.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ You can access these pages on your computer using one of the following clients:
   `stack install tldr`
 - [Node.js client](https://github.com/tldr-pages/tldr-node-client):
   `npm install -g tldr`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -96,6 +96,7 @@ You can access these pages on your computer using one of the following clients:
   `stack install tldr`
 - [Node.js client](https://github.com/tldr-pages/tldr-node-client):
   `npm install -g tldr`
+- [OCaml client](https://github.com/RosalesJ/tldr.git): `opam install tldr`
```
pxgamer: Is this mean to be pointing to the following?
```suggestion
- [OCaml client](https://github.com/RosalesJ/tldr-ocaml): `opam install tldr`
```

PR #2574, opened 2018-11-10
Add C# client

principis:
I wrote a C# client because... Why not.
It doesn't have many functions at the moment, only base functionality.
I saw some mentions of a spec for clients but nothing 'official' so I followed the wiki.

Let me know if I need to make adjustments!


principis:
Yes that seems like a good idea. It's a bit confusing what is expected of a client atm.

PR #2575, opened 2018-11-10
Update Contribution guidelines and PR template

waldyrious:
This commit updates the commit message guidelines in CONTRIBUTING.md:
- Add example commit messages for translations and for changes that affect multiple pages;
- Prefix each example with a description of the type of change it refers to.

It also updates the PR template, to make it more compact and reduce duplication of information included in CONTRIBUTING.md:
- Remove horizontal rule at the top
  (the comments are sufficient as visual separation);
- Shorten the comments to make them more compact and more likely to be read;
- Make each checklist entry take a single line;
- Remove blank lines between checklist entries.

This is a follow-up to the discussion at #2512.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,19 +1,9 @@
-----
 <!-- Thank you for sending a PR! -->
-<!-- Please perform the following checks and check all the boxes that apply. -->
-<!-- If your PR does not create a command page,
-     you can remove the first two checklist items. -->
-<!-- If your PR neither creates nor edits a command page (e.g. README edits, etc.)
-     you can simply remove the entire checklist. -->
+<!-- Please perform the following checks and mark all the boxes accordingly. -->
+<!-- You can remove the checklist items that don't apply to your PR. -->
 
 - [ ] The page (if new), does not already exist in the repo.
-
-- [ ] The page (if new), has been added to the correct platform folder:  
-      `common/` if it's common to all platforms, `linux/` if it's Linux-specific, and so on.
-
+- [ ] The page is in the correct platform folder (`common/`, `linux/`, etc.)
 - [ ] The page has 8 or fewer examples.
-
-- [ ] The PR is appropriately titled:  
-      `<command name>: add page` for new pages, or `<command name>: <description of changes>` for pages being edited.
-
-- [ ] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines.
+- [ ] The PR title conforms to the recommended [templates](../blob/master/CONTRIBUTING.md#commit-message).
+- [ ] The page follows the [content guidelines](../blob/master/CONTRIBUTING.md#guidelines).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,9 +87,10 @@ For the commit message, use the following format:
     <command>: type of change
 
 Examples:
-  - `ls: add page`
-  - `cat: fix typo`
-  - `git-push: add --force example`
+  - For a new page addition: `ls: add page`
+  - For a page edit: `cat: fix typo`, `git-push: add --force example`
+  - For a new translation of an existing page: `cp: add Tamil translation`
+  - For related changes to several pages: `grep, find, locate: synchronize format of wildcards`
 
 ## Licensing
 


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,19 +1,9 @@
...
-      `common/` if it's common to all platforms, `linux/` if it's Linux-specific, and so on.
-
+- [ ] The page is in the correct platform folder (`common/`, `linux/`, etc.)
 - [ ] The page has 8 or fewer examples.
-
-- [ ] The PR is appropriately titled:  
-      `<command name>: add page` for new pages, or `<command name>: <description of changes>` for pages being edited.
-
-- [ ] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines.
+- [ ] The PR title conforms to the recommended [templates](CONTRIBUTING.md#commit-message).
```
agnivade: template ?

waldyrious: My thinking was that there are several formats, depending on the type of change. I considered each of them a template for the commit message and PR title for that kind of change.

sbrl: This will display as markdown, right? Just curious if the links will show up. Even if not, I think it should be easy enough to work out which file & section it's referring to.

pxgamer: It will display as markdown.

Just wondering though if using `CONTRIBUTING.md` is right. I would have thought that it would point to `https://github.com/tldr-pages/tldr/pull/CONTRIBUTING.md`?

sbrl: Indeed it probably will, unless GH does some mangling. This is a test: [test](CONTRIBUTING.md)

sbrl: Nope, no mangling here.

pxgamer: Maybe it's better to do this?
`https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message`

waldyrious: It will display as markdown, but only in the preview mode when composing the PR message, or after saving the PR.

The reasoning for the short link is to make the template as compact as possible, in order not to overwhelm contributors when they're creating a PR.

In any case, relative links are a feature of markdown, so even though it would indeed be more explicit to include the full URL, it's not really wrong to use the short link, if that makes the template cleaner and more likely to be read in full.

That said, I just noticed, thanks to your conversation, that the relative link from a PR page does not produce the intended URL. Writing `[foo](CONTRIBUTING.md)` produces a link pointing to

>  https://github.com/tldr-pages/tldr/pull/CONTRIBUTING.md

rather than 

>  https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md

So we'll have to tweak the relative link, like so: `[foo](../blob/master/CONTRIBUTING.md)`. I'll make the change now.

pxgamer: Yeah, I was originally going to do that, but realised that when you go to make a PR it's from a different url. i.e. `...tldr-pages/tldr/compare/master...{user}:{branch}`

If using the relative one, then when they are writing their PR and filling in the details, it will actually point to: `...tldr-pages/tldr/compare/blob/master/CONTRIBUTING.md`

waldyrious:
@sbrl and @pxgamer, I've updated the PR to fix the relative URL to CONTRIBUTING.md so that it works from the PR page, as discussed in the inline comments above.

.github/PULL_REQUEST_TEMPLATE.md
sbrl: You can prefix it with a `/` to make a url relative to the current domain, if that makes it any more concise.

waldyrious: @pxgamer I just tested your hypothesis to be sure. The `../` in the relative link URL works just like in unix paths, i.e. it moves one level up, and that applies to both the URL where PRs are created from (`...tldr-pages/tldr/compare/...`) and to the one where the PR lives after it's created (`...tldr-pages/tldr/pull/...`).

So in this case, `[foo](../blob/master/CONTRIBUTING.md)` does indeed work as intended, producing an URL to

    https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md

So this PR can be merged as it is, after all :)

waldyrious:
I've confirmed the current URL format works as intended (see the inline conversation above for details). Feel free to merge this.

.github/PULL_REQUEST_TEMPLATE.md
pxgamer: I realised why it was failing for me, because I use GitFlow formatting for my branches. So they contain a `/` (e.g. `feature/command`)

So as long as people don't use slashes in branch names (which most newcomers probably wont) it should be fine. :+1:

waldyrious: Hmmm, that's an interesting edge case. It will fail then, indeed. Perhaps to keep it safe we should use the trick @sbrl mentioned then? i.e.

    [foo](/tldr-pages/tldr/blob/master/CONTRIBUTING.md)

which at least is shorter than the full-fledged

    [foo](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md)

WDYT?

pxgamer: D'oh. Merged to quickly. :yum:

I think that sounds good. :+1:

waldyrious: Great! I'll submit a new PR.

waldyrious:
Oh no, @pxgamer, you squash-merged my carefully crafted separate commits :cry:

Not a big deal of course :) but in the future, please keep in mind the [second-to-last bullet point here](https://github.com/tldr-pages/tldr/blob/master/contributing-guides/maintainers-guide.md#ii-handling-prs) when merging PRs.

pxgamer:
Ah, damn it. I apologise, I completely forgot about commit history.

waldyrious:
No problem! :smiley: this sort of thing is better gradually acquired as part of the development culture/process, rather than by memorizing rules exhaustively.

PR #2583, opened 2018-11-12
cut: fix a command example & minor README.md improvements

PetrusKiendys:
----

- Fixup command example for **cut**
- Minor changes in README.md


scripts/pdf/README.md
```diff
@@ -17,13 +17,13 @@ This directory contains the script and related resources to generate a PDF copy
 
 ## Requirements
 
-The PDF is generated by first converting the markdown files to HTML, and then rendering those HTML files as PDF. It depends on the following libraries-
+The PDF is generated by first converting the markdown files to HTML, and then rendering those HTML files as PDF. It depends on the following libraries:
 
-#### Python-Markdown
+**Python-Markdown**
```
agnivade: Let us revert these. `####` is a correct markdown syntax and is consistent with `##` above.

PetrusKiendys: Alright sure, I have allowed edits to be done from maintainers.
Will you edit the PR or should I do it?

pages/common/cut.md
```diff
@@ -12,7 +12,7 @@
 
 - Cut out everything from the 3rd character to the end of each line:
 
-`cut -c{{3-}}`
+`cut -c {{3-}}`
```
sbrl: This was valid syntax before. The space here is completely optional actually. Personally, I'd like to leave this as-is - as some commands (e.g. 7-zip) require the lack of space, and that can be confusing if it's a one-off thing.

agnivade: I think it is more to do with consistency. If you see the above 2 examples, they also use `-c` but have a space. I do not mind either way, but if we are not using space, then I would change the above examples too.

sbrl: Ah. Let's change this then. I've just seen a number of PRs making this kind of change recently.

PetrusKiendys: Yes, exactly as @agnivade puts it, this change is for the sake of consistency. Either we omit the space or insert it, either one of those but preferably the same in all examples.

Sorry for replying late to this, luckily @agnivade motivated the change for me, thanks!

PR #2589, opened 2018-11-13
find: add searching multiple patterns simultaneously 

ab-pivot:
Add OR search condition

----

### DIFF
--- a/pages/common/find.md
+++ b/pages/common/find.md
@@ -6,9 +6,9 @@
 
 `find {{root_path}} -name '{{*.ext}}'`
 
-- Find files by multiple search patterns, pattern_1 OR ending with case insensitive pattern_2:
+- Find files by matching multiple patterns:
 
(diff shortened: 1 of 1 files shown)


pages/common/find.md
```diff
@@ -6,6 +6,10 @@
 
 `find {{root_path}} -name '{{*.ext}}'`
 
+- Find files by multiple search patterns, pattern_1 OR ending with case insensitive pattern_2:
```
agnivade: No need to introduce iname and complicate this example. We already have `-iname` below. Please just introduce `-or` by saying "Find files by matching multiple patterns:" You are also missing a * at the end of `pattern_2`.

ab-pivot: I omitted trailing `*` intentionally to illustrate
>OR **ending** with case insensitive pattern_2:

My intention was to show as many as possible configuration details in one example. It sounds like you would prefer more focused example instead. Works for me.

PR #2590, opened 2018-11-14
ctest: add page

lbonn:
- [X] The PR title conforms to the recommended [templates](../blob/master/CONTRIBUTING.md#commit-message).
- [X] The page follows the [content guidelines](../blob/master/CONTRIBUTING.md#guidelines).

### DIFF
--- a/pages/common/ctest.md
+++ b/pages/common/ctest.md
@@ -1,0 +1,15 @@
+# ctest
+
+> CMake test driver program.
+
+- Run all tests defined in the CMake project, executing 4 jobs at a time in parallel:
+
+`ctest -j{{4}} --output-on-failure`
+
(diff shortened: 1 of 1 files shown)


lbonn:
Thanks for the proofread @sbrl!

I've tweaked it a bit to show the most common invocation first. My initial motivation was that the cli was really not intuitive for the other two common operations but showing the most basic use case is also important.

lbonn:
I changed it again sorry. A test suite is usually run with some level of parallelism.

lbonn:
Sorry, I was not aware of that. I'm usually quite wary of github rewriting the history for me but squashing makes a lot of sense in this case.

I'll keep that in mind :).

PR #2594, opened 2018-11-15
TravisCI: Include languages in index file.

arunisaac:
This PR includes a `language` field in the `index.json` file as discussed in #2339. I have kept the field name singular (`language` instead of `languages`) because `platform` is also in the singular.

Since this `index.json` file is now common to all languages, should it not be outside the `pages` folder in the top level directory of the archive? I do realize this will break existing clients...

### DIFF
--- a/scripts/build-index.js
+++ b/scripts/build-index.js
@@ -24,7 +24,9 @@ function buildPagesIndex(files) {
       if (!index[page].platform.includes(os)) {
         index[page].platform.push(os);
       }
-      index[page].language.push(language);
+      if (!index[page].language.includes(language)) {
+        index[page].language.push(language);
+      }
     } else {
       index[page] = {name: page, platform: [os], language: [language]};
     }
--- a/scripts/build-index.js
+++ b/scripts/build-index.js
@@ -11,8 +11,8 @@ function parsePagename(pagefile) {
 }
 
 function parseLanguage(pagefile) {
-  var language = pagefile.split(/\//)[0].replace(/^pages/, '');
-  return language == '' ? 'en' : 'ta';
+  var pagesFolder = pagefile.split(/\//)[0];
+  return pagesFolder == 'pages' ? 'en' : pagesFolder.replace(/^pages./, '');
 }
 
 function buildPagesIndex(files) {
--- a/scripts/build-index.js
+++ b/scripts/build-index.js
@@ -10,30 +10,27 @@ function parsePagename(pagefile) {
   return pagefile.split(/\//)[2].replace(/\.md$/, '');
 }
 
-function buildShortPagesIndex(files) {
+function parseLanguage(pagefile) {
+  var language = pagefile.split(/\//)[0].replace(/^pages/, '');
+  return language == '' ? 'en' : 'ta';
+}
+
+function buildPagesIndex(files) {
   var reducer = function (index, file) {
     var os = parsePlatform(file);
     var page = parsePagename(file);
+    var language = parseLanguage(file);
     if (index[page]) {
-      index[page].push(os);
+      if (!index[page].platform.includes(os))
+        index[page].platform.push(os);
+      index[page].language.push(language);
     } else {
-      index[page] = [os];
+      index[page] = {name: page, platform: [os], language: [language]};
     }
     return index;
   };
 
-  return files.reduce(reducer, {});
-}
-
-function buildPagesIndex(shortIndex) {
-  return Object.keys(shortIndex)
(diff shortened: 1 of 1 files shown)


scripts/build-index.js
```diff
@@ -10,30 +10,27 @@ function parsePagename(pagefile) {
...
-      index[page] = [os];
+      index[page] = {name: page, platform: [os], language: [language]};
     }
     return index;
   };
 
-  return files.reduce(reducer, {});
-}
-
-function buildPagesIndex(shortIndex) {
```
agnivade: This seems wrong. Any reason you removed this function ?

scripts/build-index.js
```diff
@@ -10,30 +10,27 @@ function parsePagename(pagefile) {
   return pagefile.split(/\//)[2].replace(/\.md$/, '');
 }
 
-function buildShortPagesIndex(files) {
+function parseLanguage(pagefile) {
+  var language = pagefile.split(/\//)[0].replace(/^pages/, '');
+  return language == '' ? 'en' : 'ta';
```
agnivade: This will not work for future languages. Please just return 'en' if the language variable is empty. Else return the value.

scripts/build-index.js
```diff
@@ -10,30 +10,27 @@ function parsePagename(pagefile) {
...
+}
+
+function buildPagesIndex(files) {
   var reducer = function (index, file) {
     var os = parsePlatform(file);
     var page = parsePagename(file);
+    var language = parseLanguage(file);
     if (index[page]) {
-      index[page].push(os);
+      if (!index[page].platform.includes(os))
```
agnivade: Please add braces.

scripts/build-index.js
```diff
@@ -10,30 +10,27 @@ function parsePagename(pagefile) {
...
+        index[page].platform.push(os);
+      index[page].language.push(language);
     } else {
-      index[page] = [os];
+      index[page] = {name: page, platform: [os], language: [language]};
```
agnivade: You have changed the format here. This will break all clients.

agnivade:
> Since this index.json file is now common to all languages, should it not be outside the pages folder in the top level directory of the archive? I do realize this will break existing clients...

It is not in the pages folder at all. At a later step, it is copied over to the assets folder in the tldr-pages.github.io repo.

arunisaac:
I was away from my computer for a few days. Hence, the delayed reply.

> -function buildShortPagesIndex(files) {
> +function parseLanguage(pagefile) {
> +  var language = pagefile.split(/\//)[0].replace(/^pages/, '');
> +  return language == '' ? 'en' : 'ta';
>
> This will not work for future languages. Please just return 'en' if
> the language variable is empty. Else return the value.

Indeed, what an embarassing mistake! I'll fix it.

>>      if (index[page]) {
> -      index[page].push(os);
> +      if (!index[page].platform.includes(os))
>
> Please add braces.

Ok, will do.

> -  return files.reduce(reducer, {});
> -}
> -
> -function buildPagesIndex(shortIndex) {
>
> This seems wrong. Any reason you removed this function ?
>
>> @@ -10,30 +10,27 @@ function parsePagename(pagefile) {
>    return pagefile.split(/\//)[2].replace(/\.md$/, '');
>  }
>  
>>      } else {
> -      index[page] = [os];
> +      index[page] = {name: page, platform: [os], language: [language]};
>
> You have changed the format here. This will break all clients.

The build-index.js script only needs to create the "pages index". It
does not need to create the "short pages index". The earlier code first
created the short pages index and then expanded it to the pages
index. My modification creates the pages index directly.

I don't think I have changed the format. And, I have not removed the
buildPagesIndex function. I only removed the buildShortPagesIndex
function. The git diff is misleading.

agnivade:
>  And, I have not removed the buildPagesIndex function. I only removed the buildShortPagesIndex
function.

Yes, I meant that. Any reason you did that ? This will break the node client.

arunisaac:
I'm sorry I don't understand. How exactly does removing buildShortPagesIndex break the node client? If I understand correctly, the short pages index is not saved to any file. And, the final zip archive at http://tldr.sh/assets/tldr.zip does not have any short pages index. The node client has its own code to generate the short pages index from the full pages index.

agnivade:
Ah you are right. Sorry about that. I need to check further since this is an important change.

agnivade:
Sorry I am not getting time to look at this.

agnivade:
I have taken the opportunity to fix this in #2696. Closing this in favor of that.

Issue #2592, opened 2018-11-14
Restructure docker page


MasterOdin:
In #1505, the docker page was restructured such that it has `docker container` in front of every command with the vague line that the old CLI interface might be gotten rid of. It never did (or was going to be) and in most resources I've seen, you'll see people using and recommending the shorter core commands over the longer form (like `docker ps` over `docker container ls`).

Would it restructure the docker page such that it has the short commands then have a `docker-container`, `docker-image`, etc. pages that has the longer form commands?

mfrw:
@MasterOdin Should we close this now?

waldyrious:
@mfrw we still need the `docker-container` and `docker-image` pages, for completeness. The former, in particular, would be required for us not to lose information that was on the `docker` page prior to #2597.

Aracki:
@waldyrious @mfrw This should be closed after #3061 & #3062 gets merged.

agnivade:
Thanks for taking care of this.

Aracki:
#3061 & #3062 merged. Closing this.

PR #1505, opened 2017-09-26
docker: update docker commands to conform to Docker 1.13

djung460:
----

~~- [ ] The page (if new), does not already exist in the repo.~~

~~- [ ] The page (if new), has been added to the correct platform folder: 
      `common/` if it's common to all platforms, `linux/` if it's Linux-specific, and so on.~~

- [x] The page has 8 or fewer examples.

- [x] The PR is appropriately titled:  
      `<command name>: add page` for new pages, or `<command name>: <description of changes>` for pages being edited

- [x] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines

PR #2597, opened 2018-11-15
docker: simplify examples

MasterOdin:
- [x] The PR title conforms to the recommended [templates](../blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](../blob/master/CONTRIBUTING.md#guidelines).

This is the first of PRs that is attached to the raised issue #2592, reverting the changes of #1505 to the more common real-world suggested usage, with the intent to create further docker command specific pages. 

I'm not sure how you'd want to handle the fact that it has 9 examples though. You could probably merge the two run examples, but not sure how to specify an optional `{{command}}` parameter.

### DIFF
--- a/pages/common/docker.md
+++ b/pages/common/docker.md
@@ -4,19 +4,19 @@
 
 - List currently running docker containers:
 
-`docker container ls`
+`docker ps`
 
 - List all docker containers (running and stopped):
 
-`docker container ls -a`
+`docker ps -a`
 
 - Start a container from an image, with a custom name:
 
-`docker container run --name={{container_name}} {{image}}`
+`docker run --name {{container_name}} {{image}}`
 
 - Start or stop an existing container:
 
-`docker container {{start|stop}} {{container_name}}`
+`docker {{start|stop}} {{container_name}}`
 
 - Pull an image from a docker registry:
 
@@ -24,16 +24,16 @@
 
 - Start a container from an image and get a shell inside of it:
 
-`docker container run -it {{image}} bash`
+`docker run -it {{image}} sh`
 
 - Run a command inside of an already running container:
 
-`docker container exec {{container_name}} {{command}}`
+`docker exec {{container_name}} {{command}}`
 
 - Remove a stopped container:
 
-`docker container rm {{container_name}}`
+`docker rm {{container_name}}`
 
 - Fetch and follow the logs of a container:
 
-`docker container logs -f {{container_name}}`
+`docker logs -f {{container_name}}`
--- a/pages/common/docker.md
+++ b/pages/common/docker.md
@@ -24,7 +24,7 @@
 
 - Open a shell inside of an already running container:
 
-`docker exec -it {{container_name}} sh`
+`docker exec -it {{container_name}} {{sh}}`
 
 - Remove a stopped container:
 
--- a/pages/common/docker.md
+++ b/pages/common/docker.md
@@ -12,7 +12,7 @@
 
 - Start a container from an image, with a custom name:
 
-`docker run --name {{container_name}} {{image}}`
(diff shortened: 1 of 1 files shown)


pages/common/docker.md
```diff
@@ -4,36 +4,36 @@
 
 - List currently running docker containers:
 
-`docker container ls`
+`docker ps`
```
sbrl: Was this meant to be `ls` instead of `ps`?

MasterOdin: Nope, that's the actual command. I assume it's one of the many reasons why they did the CLI restructure, adding in the newer sub-commands like `container`, `image`, etc.

agnivade:
@MasterOdin - Could you remove one example from the page so that we are under our budget of 8 ?

pages/common/docker.md
sbrl: I see. If the `container` commands are newer, why are we changing the page to the older commands? I'm confused.

pxgamer: @sbrl, see #2592. Looks like the short commands were going to be removed, but then most people tend to use the short commands. It looks like @MasterOdin was thinking of using additional pages for the long commands. So `docker-container`, etc.

sbrl: Ah, I see. This change is ok then :-)

Maybe once those pages are done we can revisit and add a `See also`.

MasterOdin:
Removed a 9th example from the file, but I'm a bit conflicted. Would it be appropriate to add a `{{command}}` to the end of the `docker run` example and reword it to be "Start a container from an image with a custom name and run the command (leave off to use default Dockerfile command)" or something like that or just leave it as is?

pages/common/docker.md
MasterOdin: Yeah, it's kind of a weird spot Docker is in. They added a bunch of sub-commands (container, image, etc.) to allow adding a lot more commands for those specific actions, but left the top level largely alone. As such, while all of the top level commands do have an equivalent action within a sub-command, most docs and things really just use these handful of key top level commands still.

MasterOdin:
It's not mandatory and if not included, a default command will be executed by the developer of the Dockerfile. The most common usecase would be to not include the command though.

sbrl:
Perhaps we should leave it off then. If it isn't commonly used, then I think it's better left to a more in-depth guide or tutorial :-)

pages/common/docker.md
```diff
@@ -4,36 +4,32 @@
...
 `docker pull {{image}}`
 
-- Start a container from an image and get a shell inside of it:
-
-`docker container run -it {{image}} bash`
-
 - Run a command inside of an already running container:
 
-`docker container exec {{container_name}} {{command}}`
+`docker exec -it {{container_name}} {{command}}`
```
agnivade: I think it will be better if we change command to bash. That might preserve the information we already had in the page.

MasterOdin: Changed it to `sh` as that's more guaranteed to exist on a container than bash ([Alpine Linux](https://alpinelinux.org/) is a popular container image that doesn't have bash for example).

agnivade: Sure

pages/common/docker.md
```diff
@@ -4,36 +4,32 @@
...
 
-- Start a container from an image and get a shell inside of it:
+- Open a shell inside of an already running container:
 
-`docker container run -it {{image}} bash`
-
-- Run a command inside of an already running container:
-
-`docker container exec {{container_name}} {{command}}`
+`docker exec -it {{container_name}} sh`
```
sbrl: Shouldn't `sh` be enclosed in `{{` `}}`?

MasterOdin: Done

PR #2600, opened 2018-11-16
drill: add page

zoresvit:
Add TL;DR page for a `drill` command from `ldns` package.

### DIFF
--- a/pages/common/drill.md
+++ b/pages/common/drill.md
@@ -1,0 +1,31 @@
+# drill
+
+> Get information out of DNS.
+
+- Lookup the IP(s) associated with a hostname (A records):
+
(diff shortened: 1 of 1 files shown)


pages/common/drill.md
```diff
@@ -0,0 +1,31 @@
+# drill
+
+> Get information out of DNS.
```
sbrl: Perhaps `Perform various DNS queries.` would be a better description here?

agnivade: FWIW - we deliberately try to avoid picking up lines verbatim from the man page. Otherwise a user can just look up the man page itself instead of checking the tldr page.

PR #2601, opened 2018-11-19
comp: add page

pxgamer:

### DIFF
--- a/pages/windows/comp.md
+++ b/pages/windows/comp.md
@@ -1,0 +1,32 @@
+# comp
+
+> Compare the contents of two files or sets of files.
+> Use wildcards (*) to compare sets of files.
+
+- Compare files interactively:
+
+`comp`
+
+- Compare two specified files:
+
+`comp {{path/to/file_1}} {{path/to/file_2}}`
+
+- Display differences in decimal format:
+
+`comp /d {{path/to/file_1}} {{path/to/file_2}}`
+
+- Display differences in ASCII format:
+
+`comp /a {{path/to/file_1}} {{path/to/file_2}}`
+
+- Display line numbers for differences:
+
+`comp /l {{path/to/file_1}} {{path/to/file_2}}`
+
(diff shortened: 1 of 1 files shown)


pages/windows/comp.md
```diff
@@ -0,0 +1,32 @@
+# comp
+
+> Compare the contents of two files or sets of files.
+> Use wildcards (*) to compare sets of files.
```
sbrl: Cool! Looks ok to me. Will this be an issue for clients? GitHub's diff renderer doesn't appear to like it. Should we wrap it in `*`?

pxgamer: I think it's a feature of comp. I was going to add an example, but it seemed a bit weird doing;

```powershell
comp {{path/to/*}} {{path/to/*}}
```

Or something like that.

jedahan: I don't think its that weird as long as you write the explanation above:

```markdown
- compare sets of files:

`comp {{path/to/folder/a/*}} {{path/to/folder/b/*}}`
```

pages/windows/comp.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Display differences in ASCII format:
+
+`comp /a {{path/to/file_1}} {{path/to/file_2}}`
+
+- Display line numbers for differences:
+
+`comp /l {{path/to/file_1}} {{path/to/file_2}}`
+
+- Use a case-insensitive comparison:
```
agnivade: "Compare files case-insensitively" might sound more consistent with the previous descriptions.

pages/windows/comp.md
pxgamer: @sbrl, using a backslash seems to work well. But I'm only testing on the Node client. 🤔

With `***` it just renders all 3 of them for me.

sbrl: Ah. Let's leave it then :-)

PR #2605, opened 2018-11-21
restic: add how to restore a specific file

bbigras:

I think it must be frequent that people only want to restore 1 file (or a specific path).

### DIFF
--- a/pages/common/restic.md
+++ b/pages/common/restic.md
@@ -18,6 +18,10 @@
 
 `restic -r {{path/to/repository}} restore {{snapshot_id}} {{path/to/target}}`
 
+- Restore a specific path from a specific backup to a target directory:
+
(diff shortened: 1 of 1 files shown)


pages/common/restic.md
```diff
@@ -18,6 +18,10 @@
 
 `restic -r {{path/to/repository}} restore {{snapshot_id}} {{path/to/target}}`
 
+- Restore a specific file from a specific backup to a target directory:
```
bbigras: I think I should change 'specific file' to 'specific path'

agnivade: Can it restore folders too ?

bbigras: Yes. If you specify something like `/home/user` it will restore all the files matching this path. So the folder recursively.

bbigras: I just pushed a new commit with 'path' instead of 'file'.

bbigras:
The new commit also include 'restore' that I forgot last time. :man_facepalming:

PR #2610, opened 2018-11-23
rclone: add page

Linguista:
Commands for Rclone <http://www.rclone.org>

### DIFF
--- a/pages/common/rclone.md
+++ b/pages/common/rclone.md
@@ -1,0 +1,52 @@
+# rclone
+
+> CLI program to copy/sync/move files and directories to and from many cloud services.
+
+
+- Interactively configure an rclone remote at the cloud service of your choice:
+
+`rclone config`
+
+
+- List contents of a folder on an rclone remote:
+
+`rclone lsf {{remote_name}}:{{path/to/folder}}`
+
+
+- Copy file or folder from local source to remote destination:
+
+`rclone copy {{/path/to/source_file_or_folder}} {{remote_name}}:{{path/to/destination_folder}}`
+
+
+- Copy file or folder from remote source to local destination:
+
+`rclone copy {{remote_name}}:{{path/to/source_file_or_folder}} {{/path/to/destination_folder}}`
+
+
+- Sync local source to remote destination, changing the destination only:
+
+`rclone sync {{/path/to/file_or_folder}} {{remote_name}}:{{path/to/folder}}`
+
+
+- Move file or folder from local source to remote destination:
+
+`rclone move {{/path/to/file_or_folder}} {{remote_name}}:{{path/to/folder}}`
+
+
+- Delete remote file or folder (use `--dry-run` to test, remove it to actually delete):
+
+`rclone --dry-run delete {{remote_name}}:{{path/to/file_or_folder}}`
+
+
+- Remove empty remote folder:
+
+`rclone rmdir {{remote_name}}:{{path/to/file_or_folder}}`
+
+
+- Mount rclone remote (experimental):
+
+`rclone mount {{remote_name}}:{{path/to/folder}} {{/path/to/mount_point}}`
+
+- Unmount rclone remote if CTRL-C fails (experimental)
+
+`fusermount -u {{/path/to/mount_point}}`


pages/common/rclone.md
```diff
@@ -0,0 +1,35 @@
...
+
+`rclone lsf {{remote_name}}:{{path/to/folder}}`
+
+- Copy file or folder from local source to remote destination:
+
+`rclone copy {{/path/to/source_file_or_folder}} {{remote_name}}:{{path/to/destination_folder}}`
+
+- Copy file or folder from remote source to local destination:
+
+`rclone copy {{remote_name}}:{{path/to/source_file_or_folder}} {{/path/to/destination_folder}}`
```
agnivade: Remove preceding `/` from `/path` here and elsewhere.

pages/common/rclone.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Move file or folder from local source to remote destination:
+
+`rclone move {{/path/to/file_or_folder}} {{remote_name}}:{{path/to/folder}}`
+
+- Delete remote file or folder (use `--dry-run` to test, remove it to actually delete):
+
+`rclone --dry-run delete {{remote_name}}:{{path/to/file_or_folder}}`
+
+- Mount rclone remote (experimental):
```
agnivade: If these are experimental, are they necessary for a tldr page ? Remember, a tldr page is mainly for beginners trying to get introduced to a command. For further details, there is the man page anyways.

pages/common/rclone.md
Linguista: In some cases the preceding `/` is required, and in some it shouldn't be used. All commands here were tested as shown, and are confirmed to work, so I wouldn't recommend removing any slashes.

pages/common/rclone.md
Linguista: It's a pretty important function, and it's been listed as "experimental" for ages -- at this point, it's really just the author being overly cautious, I think.

agnivade: Ok

pages/common/rclone.md
agnivade: Then in cases where `/` is required, we should move the `/` out of the tokens (`{{}}`).

pages/common/rclone.md
```diff
@@ -0,0 +1,35 @@
...
+
+`rclone mount {{remote_name}}:{{path/to/folder}} {{/path/to/mount_point}}`
+
+- Unmount rclone remote if CTRL-C fails (experimental):
+
+`fusermount -u {{/path/to/mount_point}}`
```
agnivade: This should be in its own page I think, because this is not an rclone command. Unless this command is always used with rclone.

Linguista: Hmm... it is indeed a different command than `rclone`, but at the same time it's the only way to do this with `rclone` mounts -- there's no native equivalent. What do you recomment?

pages/common/rclone.md
Linguista: I just did that and was about to commit the change, but when I saw how it actually looks, I stopped. `/` is an integral part of the path, and placing it outside of the double braces makes it appear to be part of the command's syntax, which is not the case. I think that's why we don't see things like `rm` **/**`var/logs/my.log`. 

So are you really sure about this request?

Linguista: Changes made.

pages/common/rclone.md
Linguista: I understand your rationale, but I think such a change would make it impossible to provide this useful information to users, since a page for `fusermount` couldn't possibly include usage examples for all the programs that make use of it. Plus, the change would only save a handful of characters. In other words, something lost and essentially nothing gained. I'd really recommend not implementing it.

pages/common/rclone.md
agnivade: Umm.. I don't see the `/` outside the token. Am I missing something ?

Linguista: Do you not see [this](https://github.com/tldr-pages/tldr/pull/2610/commits/ed968e43634f405837a8699f56648197417e40ff)? Not a big github user, so I may have neglected to do something besides make the edits.

Linguista: By the way, in light of your comment on not using absolute paths, I decided to eliminate _all_ leading slashes, though I don't think this is a good solution.

`rclone` does _not_ take a leading slash before the first directory in a remote share (so it's `remote:home/user/bin`, even though the same path locally is `/home/user/bin`). The absence of something can't be shown, of course, so the only way to make this clear in an example is to put the leading slashes in all _non_-remote directories, so that their absence in remotes is obviously intentional, and communicates the idea that they shouldn't be used.

This clashes with the `tldr` style guide. 

So the choice is one of stylistic consistency versus accuracy. I personally think the latter is the best choice, but I suspect that insisting on this point is a losing battle, so I won't insist.

It might be worthwhile to consider just how iron-clad the style rules should be, in general.

agnivade: > By the way, in light of your comment on not using absolute paths, I decided to eliminate all leading slashes

According to this [comment](https://github.com/tldr-pages/tldr/pull/2610#discussion_r236039871) of mine, I thought we decide on pulling the `/` out of the token. But it seems you have removed it. I thought you meant those were required in some cases ?

> So the choice is one of stylistic consistency versus accuracy. I personally think the latter is the best choice, 

Sure, you can just add a line in the sub-heading mentioning this. So that it is extra clear.

sbrl:
Hey, @agnivade! What's remaining before we can get this merged?

PR #2611, opened 2018-11-24
sort: add --ignore-case example

cslee:

### DIFF
--- a/pages/common/sort.md
+++ b/pages/common/sort.md
@@ -10,6 +10,10 @@
 
 `sort -r {{filename}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/sort.md
```diff
@@ -10,6 +10,10 @@
 
 `sort -r {{filename}}`
 
+- Sort a file in case-insensitive way:
+
+`sort -f {{filename}}`
```
agnivade: I would recommend to use the longer version - `--ignore-case` because it is easier to understand. `-f` stands for fold which means we would have to use that word in the description.

PR #2613, opened 2018-11-24
entr: add page

Jegp:

Also linted with tldr-lint

### DIFF
--- a/pages/linux/entr.md
+++ b/pages/linux/entr.md
@@ -1,0 +1,23 @@
+# entr
+
+> Run arbitrary commands when files change.
+
+- Rebuild with `make` if any files in any subfolders change:
+
+`ag -l | entr make`
+
+- Rebuild and test with `make` if `.c` sources in current directory changes:
+
+`ls *.c | entr 'make && make test'`
+
+- Reload the MuPDF viewer whenever the source `.pdf` changes:
+
+`ls *.pdf | entr pkill -HUP mupdf`
+
+- Reload the spawned childprocess with `SIGTERM` before restarting:
+
+`ls *.rb | entr -r ruby main.rb`
+
+- Run a command with the changed file (`/_`) as input:
+
+`ls *.sql | entr psql -f /_`
--- a/pages/linux/entr.md
+++ b/pages/linux/entr.md
@@ -18,6 +18,6 @@
 
 `ls *.rb | entr -r ruby main.rb`
 
-- Run a command with the changed file (`/_`) as input:
+- Run a command with the changed file (`/_`) as an argument:
 
 `ls *.sql | entr psql -f /_`
--- a/pages/linux/entr.md
+++ b/pages/linux/entr.md
@@ -6,7 +6,7 @@
 
 `ag -l | entr make`
 
-- Rebuild and test with `make` if `.c` sources in current directory changes:
+- Rebuild and test with `make` if any `.c` source files in the current directory change:
 
 `ls *.c | entr 'make && make test'`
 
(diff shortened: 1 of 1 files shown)


pages/linux/entr.md
```diff
@@ -0,0 +1,23 @@
+# entr
+
+> Run arbitrary commands when files change.
+
+- Rebuild with `make` if any files in any subfolders change:
+
+`ag -l | entr make`
```
sbrl: Perhaps we could enclose `ag -l` and `make` in quotes? This denotes that they can be changed by the reader.

The same goes for the examples further down.

pages/linux/entr.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Rebuild and test with `make` if `.c` sources in current directory changes:
```
sbrl: Perhaps this will read better?

```suggestion
- Rebuild and test with `make` if any `.c` source files in the current directory change:
```

pages/linux/entr.md
```diff
@@ -0,0 +1,23 @@
...
+
+`ls *.c | entr 'make && make test'`
+
+- Reload the MuPDF viewer whenever the source `.pdf` changes:
```
sbrl: What's MuPDF? I'm unsure as to how this is different to the previous example.

pages/linux/entr.md
```diff
@@ -0,0 +1,23 @@
...
+
+`ls *.pdf | entr pkill -HUP mupdf`
+
+- Reload the spawned childprocess with `SIGTERM` before restarting:
```
sbrl: Perhaps `subprocess` or `child process` would read better here?

Also, what exactly do you mean by "reload"?

pages/linux/entr.md
```diff
@@ -0,0 +1,23 @@
...
+
+`ls *.rb | entr -r ruby main.rb`
+
+- Run a command with the changed file (`/_`) as input:
```
sbrl: Perhaps this will read better?

```suggestion
- Run a command with the changed file (`/_`) as an argument:
```

pages/linux/entr.md
Jegp: It's actually not different, just another example on how to send signals using the command. I've added an updated description below. Alternatively it could be left out. What are your thoughts?
```suggestion
- Reload a PDF viewer (`mupdf`) by sending a `SIGHUP` whenever any `.pdf` sources change in the current folder:
```

pages/linux/entr.md
Jegp: Is this more clear?
```suggestion
- The reload (`-r`) flag sends a `SIGTERM` to any previously spawned ruby subprocesses before executing `ruby main.rb`:
```

pages/linux/entr.md
agnivade: Yes, I believe it can be left out.

pages/linux/entr.md
sbrl: Oops! I meant the token syntax :P

pages/linux/entr.md
sbrl: I think so! Perhaps we could trim it back to this?

```
Send a `SIGTERM` to any previously spawned ruby subprocesses before executing `ruby main.rb`
```

Jegp: Definitely! I've updated the commit.

pages/linux/entr.md
Jegp: Aha, that makes more sense. I've added a commit (5d0376) that puts all non-flag commands in `{{` and `}}` brackets. Is that what you meant?

PR #2617, opened 2018-11-26
kubens: add page

Aracki:
I have added `kubens`. An utility to switch between `kubectl` namespaces.


pages/common/kubens.md
```diff
@@ -0,0 +1,15 @@
+# kubens
+
+> Utility which is able to switch between Kubernetes namespaces.
```
sbrl: Perhaps we can simplify this?

```suggestion
> Utility to switch between Kubernetes namespaces.
```

PR #2578, opened 2018-11-11
stdbuf: add page

kseistrup:
----

- [x] The page (if new), does not already exist in the repo.

- [x] The page (if new), has been added to the correct platform folder:  
      `common/` if it's common to all platforms, `linux/` if it's Linux-specific, and so on.

- [x] The page has 8 or fewer examples.

- [x] The PR is appropriately titled:  
      `<command name>: add page` for new pages, or `<command name>: <description of changes>` for pages being edited.

- [x] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines.

PR #2618, opened 2018-11-26
stdbuf: add page

sbrl:
A follow-on from #2578.

### DIFF
--- a/pages/common/stdbuf.md
+++ b/pages/common/stdbuf.md
@@ -2,7 +2,7 @@
 
 > Run a command with modified buffering operations for its standard streams.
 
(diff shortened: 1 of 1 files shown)


mfrw:
Should we pen-down somewhere that our review process can be a little involved. It seems people get pissed off after some time?

mfrw:
@pxgamer Yes, I completely agree with what you pointed out :)

sbrl:
Erm I thought I did, @pxgamer? It may not have gone through - hrm.

_(Sorry about the inactivity. Got 3 deadlines for my masters, and I needed to get on top of that first)_

PR #2621, opened 2018-11-27
yarn: make the install example more explicit

waldyrious:
Also format code in other examples with backticks.

### DIFF
--- a/pages/common/yarn.md
+++ b/pages/common/yarn.md
@@ -6,19 +6,19 @@
 
 `yarn global add {{module_name}}`
 
-- Install all dependencies referenced in the package.json file:
+- Install all dependencies referenced in the `package.json` file (the `install` is optional):
 
-`yarn`
+`yarn install`
 
-- Install a module and save it as a dependency to the package.json file (add --dev to save as a dev dependency):
+- Install a module and save it as a dependency to the `package.json` file (add `--dev` to save as a dev dependency):
 
 `yarn add {{module_name}}@{{version}}`
 
-- Uninstall a module and remove it from the package.json file:
+- Uninstall a module and remove it from the `package.json` file:
 
 `yarn remove {{module_name}}`
 
-- Interactively create a package.json file:
+- Interactively create a `package.json` file:
 
 `yarn init`
 


waldyrious:
Oh, I see. Too bad that doesn't allow us to format other elements as code. I'll change in this PR, but out of curiosity, would it be possible/efficient for the clients to check the captured potential "see more" commands against the index of actually existent pages? Otherwise even if they are legitimate commands, we'd be telling people to look up pages don't exist.

waldyrious:
Alternatively, we could come up with a markup that would be explicit about listing other commands, then clients could extract them confidently, plus they'd be backward-compatible (including for plain-text rendering of the pages!). How about using `[[...]]`? That's already used for wikicode internal links, and is similar to markdown links too.

waldyrious:
> I just wanted to be clear in the page itself. Using backticks would not help 

Oh, great. In that case, I'd rather keep the code formatting I had originally, if that's OK with you.

My reasoning is that the page itself should have its content formatted in a meaningful manner, so that e.g. a simple markdown renderer would produce appropriate output, and in fact even reading the plain text version would provide adequate hints about the type of content each markup is wrapping.

I also don't think it's legitimate to take over markdown's code formatting using backticks and restrict it to a specific use case. If clients can already handle backtick-wrapped strings that aren't pages in the database, and avoid adding them to a see also list, then there's no downside in using that formatting where it makes sense.

WDYT?

agnivade:
Hmm .. that is true. However, I am afraid that I enforced this policy on a lot of reviews. So folks might blame me for inconsistencies later :stuck_out_tongue_closed_eyes:  Your logic is sound though. Somehow, mentally I just wanted to separate the two.

waldyrious:
Nah, it's perfectly OK to change one's mind :) if people complain about the inconsistency, an honest explanation should be sufficient to appease the confusion. So I'll revert my commit changing the backticks to quotes.

As for separating the two, I don't disagree with that :) IMO a specialized markup for links to other pages would be quite welcome, and I believe there's plenty of precedent for using `[[...]]`, since that's what's used by mediawiki (one of the most widely used lightweight markup languages, along with markdown). But we can discuss that elsewhere if you guys prefer making that decision more carefully.

PR #2636, opened 2018-12-04
dvtm: add page

Jipok:

### DIFF
--- a/pages/common/dvtm.md
+++ b/pages/common/dvtm.md
@@ -10,9 +10,9 @@
 
 `Ctrl+g c`
 
-- Focus the **n**-th window:
+- Focus the n-th window:
 
-`Ctrl+g **n**`
+`Ctrl+g n`
 
 - Close focused window:
 
--- a/pages/common/dvtm.md
+++ b/pages/common/dvtm.md
@@ -23,7 +23,7 @@
 
 `Ctrl+g Space`
 
-- Toggle dvtm mouse grabbing:
+- Toggle mouse grabbing:
 
 `Ctrl+g M`
 
--- a/pages/common/dvtm.md
+++ b/pages/common/dvtm.md
@@ -1,0 +1,32 @@
+# dvtm
+
+> Dynamic virtual terminal manager.
+
+- Show manual page:
+
+`Ctrl+g ?`
+
+- Create a new shell window:
+
+`Ctrl+g c`
+
+- Focus the **n**-th window:
+
+`Ctrl+g **n**`
+
+- Close focused window:
+
+`Ctrl+g x x`
+`Ctrl+d (for shell only)`
+
+- Toggle between defined layouts:
+
+`Ctrl+g Space`
+
+- Toggle dvtm mouse grabbing:
+
+`Ctrl+g M`
+
+- Quit dvtm:
+
+`Ctrl+g q q`


Jipok:
http://www.brain-dump.org/projects/dvtm/
https://github.com/martanne/dvtm

pages/common/dvtm.md
```diff
@@ -0,0 +1,32 @@
+# dvtm
+
+> Dynamic virtual terminal manager.
+
+- Show manual page:
```
agnivade: Please show a basic invocation of dvtm. And then move to these commands.

pages/common/dvtm.md
```diff
@@ -0,0 +1,32 @@
...
+
+`Ctrl+g c`
+
+- Focus the n-th window:
+
+`Ctrl+g n`
+
+- Close focused window:
+
+`Ctrl+g x x`
```
agnivade: We do not use multiple consecutive examples. Please separate them.

pages/common/dvtm.md
```diff
@@ -0,0 +1,32 @@
...
+`Ctrl+d (for shell only)`
+
+- Toggle between defined layouts:
+
+`Ctrl+g Space`
+
+- Toggle dvtm mouse grabbing:
```
agnivade: No need to mention dvtm here and the last example. It is understood.

pages/common/dvtm.md
Jipok: In most cases it exec through abduco. You don't need to invocate it manually. And since multiple consecutive examples are forbidden, what should I choose from this:
`dvtm`
`abduco -A {{name}}`
`abduco -A {{name}} dvtm`

pages/common/dvtm.md
Jipok: I noticed that most people don't know that ctrl+d sends \0 which the shell processes as logout. And in most cases, such a window closing is much better than killing it with ctrl+g x x. 
How to write this best?

pages/common/dvtm.md
Jipok: Ok. I just took it from the manual.

agnivade: We always try to avoid the language in the manual. Otherwise it will simply be a copy-paste job and the user can anyways lookup the manual. There will be no added value for this page. :)

Jipok:
Need help with invocation example and window closing example.

pages/common/dvtm.md
agnivade: Then please add a sub-heading saying - Refer to the abduco command to see how to invoke this.

pages/common/dvtm.md
agnivade: Go with the most common scenario. I think that should be Ctrl+d.

Jipok: Is it ok that Ctrl-d is not a dvtm hotkey, but little known(i think) standard hotkey for a shell?

agnivade: As I said before, choose the command you would recommend normally. This is a tldr page, so we cannot add everything. If ctrl-d is a better choice than ctrl-g x x, then go for it. Even if it's not a dvtm hotkey, if it is what the users would normally use, we should use that.

We always prefer practical and useful commands rather than sticking with a man-page like approach.

PR #2639, opened 2018-12-08
tar: Add Example About Extracting Files From Nested Directories

ElderMael:


agnivade:
Hi @ElderMael - I had already sent a PR for this long ago https://github.com/tldr-pages/tldr/pull/1244. But we decided not to merge that because it will unnecessarily add more than what is necessary in a tldr page.

We need a strong reason to exclude other options and include this.

mebeim:
Closing this since there hasn't been activity for >5 months, and as noted this change was already refused in the past. @ElderMael feel free to comment if you'd like to explain a reason why this example should be preferred over any other.

PR #2640, opened 2018-12-08
apt-file: add page

sbrl:

### DIFF
--- a/pages/linux/apt-file.md
+++ b/pages/linux/apt-file.md
@@ -6,7 +6,7 @@
 
 `sudo apt update`
 
-- Search for packages that contain the specified file:
+- Search for packages that contain the specified file or path:
 
 `apt-file search {{part/of/filename/or/path}}`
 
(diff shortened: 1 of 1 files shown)


pages/linux/apt-file.md
```diff
@@ -0,0 +1,15 @@
...
+
+> Search for files in apt packages, including ones not yet installed.
+
+- Update the metadata database:
+
+`sudo apt update`
+
+- Search for packages that contain the specified file:
+
+`apt-file search {{part/of/filename/or/path}}`
```
agnivade: In the description, you mention "file", but in the example, you add that it can be a part of the filename too. I am wondering if this info can be moved to the description, and the example be kept simple.

sbrl: I see. Doing some research reveals that it uses a substring search. How about this?

```
Search for packages that contain the specified file or path
```

agnivade: SGTM

agnivade: Did you want to change the token too, or keep it the same ?

sbrl: Oops! Forgot about it :P

PR #2645, opened 2018-12-12
lastb: add page

mstruebing:

### DIFF
--- a/pages/linux/lastb.md
+++ b/pages/linux/lastb.md
@@ -1,0 +1,23 @@
+# lastb
+
+> Show a listing of last logged in users.
+
+- Show a list of all last logged in users:
+
+`lastb`
+
+- Show a list of all last logged in users since:
+
+`lastb --since {{time}}`
+
+- Show a list of all last logged in users until:
+
+`lastb --until {{time}}`
+
+- Show a list of all logged in users at a specific time:
+
+`lastb --present {{time}}`
+
+- Show a list of all last logged in users and translate the IP into a hostname:
+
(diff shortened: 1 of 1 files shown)


pages/linux/lastb.md
```diff
@@ -0,0 +1,23 @@
+# lastb
+
+> Show a listing of last logged in users.
+
+- Show a list of all last logged in users:
```
agnivade: This commands needs sudo to run. Let us add that.

pages/linux/lastb.md
```diff
@@ -0,0 +1,23 @@
...
+
+`lastb`
+
+- Show a list of all last logged in users since:
+
+`lastb --since {{time}}`
```
agnivade: It becomes much clear if you mention the time format. I struggled with it and had to look up the man page. You can mention - "YYYY-MM-DD" or "hh:mm" also if you prefer.

pages/linux/lastb.md
mstruebing: You are right. But not everyone has `sudo` installed, for someone it might would be `su -c 'lastb'`.
What about beginning the line with a `#` to indicate that this command must be run as root?

Which of these suggestions do you prefer? Still `sudo`?

mstruebing:
I've fixed the comments from @agnivade but I'm not entirly sure about the time-format description.

agnivade:
@mstruebing - I inserted the format examples in the token itself. And used YYYY-MM-DD and hh:mm in separate examples so that its clear that both formats can be used.

PR #2646, opened 2018-12-12
zstd: add page

dcaspi:

### DIFF
--- a/pages/common/zstd.md
+++ b/pages/common/zstd.md
@@ -1,0 +1,23 @@
+# zstd
+
+> Compress or decompress files with Zstandard compression
+
+- Compress a file into a new file with the .zst suffix:
+
+`zstd {{file}}`
+
+- Decompress a file:
+
+`zstd -d {{file}}.zstd`
+
+- Decompress to stdout:
+
+`zstd -dc {{file}}.zstd`
+
+- Compress a file, while specifing the compression level. 0 = Worst, 19 = Best (Default level is 3):
+
+`zstd -{{level}} {{file}}`
+
+- Compress a file with high compression levels, using more memory during compression and decompression. Max level is 22:
+
+`zstd --ultra -{{level}} {{file}}`
--- a/pages/common/zstd.md
+++ b/pages/common/zstd.md
@@ -18,6 +18,6 @@
 
 `zstd -{{level}} {{file}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/zstd.md
```diff
@@ -0,0 +1,23 @@
...
+
+> Compress or decompress files with Zstandard compression.
+
+- Compress a file into a new file with the .zst suffix:
+
+`zstd {{file}}`
+
+- Decompress a file:
+
+`zstd -d {{file}}.zstd`
```
agnivade: This has ".zstd", but the previous example has ".zst". Are both correct ?

pages/common/zstd.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Decompress to stdout:
+
+`zstd -dc {{file}}.zstd`
+
+- Compress a file, while specifing the compression level. 0 = Worst, 19 = Best (Default level is 3):
+
+`zstd -{{level}} {{file}}`
+
+- Compress a file with high compression levels, using more memory during compression and decompression. Max level is 22:
```
agnivade: Remove "during compression and decompression". Convert last sentence to inside parenthesis, as the description above.

pages/common/zstd.md
dcaspi: Nice catch
Thanks!

pages/common/zstd.md
dcaspi: I've done as you said, but I think it adds value to mention that it uses more memory during decompression as well, as it's not obvious.
Would appreciate your feedback on that.

agnivade: Hmm .. that's right. You mean it will take more memory while decompressing the file as well right ? I agree that it is not obvious, but at the same time it makes the sentence a bit longer than I am comfortable with. @sbrl @pxgamer - thoughts ?

agnivade: I think it still doesn't clarify the fact that it will take more memory during decompression too.

sbrl: Ah, I see. Hrm. What about this then:

```
Use more memory (both when compressing and decompressing) to achieve a higher compression ratio:
```

agnivade: SGTM.

sbrl:
Done, @agnivade

PR #2647, opened 2018-12-12
aws: add page

Aracki:
We already have [`aws-s3`](https://github.com/tldr-pages/tldr/blob/master/pages/common/aws-s3.md) page, but I think there is need for a more general one like `aws`.

### DIFF
--- a/pages/common/aws.md
+++ b/pages/common/aws.md
@@ -1,0 +1,23 @@
+# aws
+
+> The official CLI tool for Amazon Web Services.
+
+- List all IAM users:
+
+`aws iam list-users`
+
+- List all EC2 instances from a specific region:
+
+`aws ec2 describe-instances --region {{us-east-1}}`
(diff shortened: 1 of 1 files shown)


pages/common/aws.md
```diff
@@ -0,0 +1,23 @@
...
+
+`aws sqs receive-message --queue-url {{https://queue.amazonaws.com/546123/Test}}`
+
+- Publish message to the specific SNS topic:
+
+`aws sns publish --topic-arn {{arn:aws:sns:us-east-1:54633:testTopic}} --message "Message"`
+
+- To see help text for AWS RDS commands:
+
+`aws rds help`
```
agnivade: Let us change this to a generic help just like the `git` page. - aws {{command}} help

Aracki: seems legit!

PR #2648, opened 2018-12-12
terminalizer: add page

Aracki:
[Terminalizer](https://github.com/faressoft/terminalizer) is a utility tool for recording your terminal and making a gif out of it.

### DIFF
--- a/pages/common/terminalizer.md
+++ b/pages/common/terminalizer.md
@@ -2,7 +2,7 @@
 
 > Utility program which records the terminal and generate animated gifs or share a video.
 
-- Create a global config directory:
+- Create the global config directory:
 
 `terminalizer init`
 
--- a/pages/common/terminalizer.md
+++ b/pages/common/terminalizer.md
@@ -1,16 +1,16 @@
 # terminalizer
 
-> Utility program which records your terminal and generate animated gifs or share a web player.
+> Utility program which records the terminal and generate animated gifs or share a video.
 
 - Create a global config directory:
 
 `terminalizer init`
 
-- Record your terminal and create a recording file:
+- Record the terminal and create a recording file:
 
 `terminalizer record {{file_name}}`
 
-- Play a recorded file on your terminal:
+- Play a recorded file on the terminal:
 
 `terminalizer play {{file_name}}`
 
@@ -18,6 +18,6 @@
 
 `terminalizer render {{file_name}}`
 
-- Upload a recording file and get a link for an online player:
+- Upload a video to terminalizer.com:
 
 `terminalizer share {{file_name}}`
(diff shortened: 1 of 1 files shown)


pages/common/terminalizer.md
```diff
@@ -0,0 +1,23 @@
+# terminalizer
+
+> Utility program which records your terminal and generate animated gifs or share a web player.
```
agnivade: Let us avoid using "your". Also, what do you mean by "share a web player" ?

pages/common/terminalizer.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Create a global config directory:
+
+`terminalizer init`
+
+- Record your terminal and create a recording file:
```
agnivade: Record _the_ terminal

pages/common/terminalizer.md
```diff
@@ -0,0 +1,23 @@
...
+
+`terminalizer record {{file_name}}`
+
+- Play a recorded file on your terminal:
```
agnivade: on _the_ terminal

pages/common/terminalizer.md
```diff
@@ -0,0 +1,23 @@
...
+
+`terminalizer play {{file_name}}`
+
+- Render a recording file as an animated gif image:
+
+`terminalizer render {{file_name}}`
+
+- Upload a recording file and get a link for an online player:
```
agnivade: Where does it upload ?

pages/common/terminalizer.md
agnivade: I think just calling them "videos" keeps things simple and clear.

pages/common/terminalizer.md
agnivade: Just mentioning the upload destination should be fine IMO.

pages/common/terminalizer.md
Aracki: LGTM

pages/common/terminalizer.md
Aracki: I simplified it...

pages/common/terminalizer.md
```diff
@@ -0,0 +1,23 @@
+# terminalizer
+
+> Utility program which records the terminal and generate animated gifs or share a video.
+
+- Create a global config directory:
```
sbrl: Does this create said directory in a pre-defined location, or in the current directory?

Aracki: Default location is `$HOME/.terminalizer`. I think it's known that all "global" configs goes into `$HOME`.

Aracki: Or you should think that need to be mentioned?

sbrl: Ah, I see. I don't think we need specify the directory it puts it in. What about tweaking it to this?

```suggestion
- Create the global config directory:
```

Using the definite article here helps imply that we're talking about 1 specific instance, instead 1 out of potentially many instances.

PR #2652, opened 2018-12-15
Add another Go client

elecprog:
I wrote a(nother) client in Go, that uses a database for performance and Go's cross compilation to support most platforms (providing only common pages on 'non-tldr platforms').

As far as I can see I support all required features (ie. page lookup and downloading the pages from <https://tldr.sh/assets/tldr.zip>).

If you think there is anything missing, please let me know. :slightly_smiling_face:


waldyrious:
Much appreciated, @elecprog! Nice work :)

PR #2656, opened 2018-12-18
tlp-stat: add page

mikesprague:
Signed-off-by: Michael Sprague <mikesprague@gmail.com>

### DIFF
--- a/pages/linux/tlp-stat.md
+++ b/pages/linux/tlp-stat.md
@@ -1,6 +1,6 @@
 # tlp-stat
 
-> Show configuration, system information, active power saving settings and battery data. See also `tlp`.
+> A tool to generate TLP status reports. See also `tlp`.
 
 - Generate status report with configuration and all active settings:
 
--- a/pages/linux/tlp-stat.md
+++ b/pages/linux/tlp-stat.md
@@ -2,7 +2,7 @@
 
 > TLP status reports. See also `tlp`.
 
-- Status report with configuration and all active settings:
+- Generate status report with configuration and all active settings:
 
 `sudo tlp-stat`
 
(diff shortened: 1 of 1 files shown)


pages/linux/tlp-stat.md
```diff
@@ -0,0 +1,15 @@
+# tlp-stat
+
+> TLP status reports. See also `tlp`.
+
+- Status report with configuration and all active settings:
```
sbrl: Perhaps this will read better?

```suggestion
- Generate status report with configuration and all active settings:
```

mikesprague: Sure thing, just committed the update. Thanks!

pages/linux/tlp-stat.md
```diff
@@ -0,0 +1,15 @@
+# tlp-stat
+
+> TLP status reports. See also `tlp`.
```
agnivade: How about - "A tool to generate TLP status reports." ?

mikesprague: I'm going to recommend we go with what's on the [man page](https://www.mankier.com/8/tlp-stat) if you want more descriptive text. How about the following:

> Show configuration, system information, active power saving settings and battery data.

I'll update my PR with the above. If you'd still prefer your suggested verbiage, let me know. Thanks!

agnivade: Your original description was fine. There is no need to describe everything in the heading. I just wanted to add "A tool to generate" to clarify what it does. Otherwise if you just say "TLP status reports", then it is incomplete, and questions like "Where, What" arise.

mikesprague: Makes sense. I will revert and update with your suggestion.

FWIW - as `tlp-stat` installs with `tlp`, and isn't used on it's own, I had made the incorrect assumption that level of specificity wasn't needed.

Thanks for clarification. Will update now.

Issue #2649, opened 2018-12-13
Add links to projects / tools


Jab2870:
Firstly, great tool. Thanks for creating it.

Some tools are pretty obsucure. In my case it was `x_x`. I came accross it by acident but I'm glad I did. The description caught my interest but searching for the repo wasn't as easy as a websearch.

It would be very helpful in this case, and I am sure others, if the pages included a link to a repo or a project page.

agnivade:
Thanks, I think this was brought before sometime in the past. I am not against this. But this requires a change in the page format, which needs to be discussed and approved and implemented by all clients. A non-trivial amount of work.

waldyrious:
This had originally been proposed in #934. Thanks for opening an issue -- now we can track this properly, as a sub-item of #1082.

sbrl:
Do you have some suggested syntax, @Jab2870? Ideally, it should be something that doesn't break existing clients, feels like markdown, and looks good even if not supported by a particular client.

Jab2870:
Well, a link in MarkDown is this:

```markdown
[Link Text](URL)
```
So that would probably be a startign point.

Perhaps we could have a section called relavent links?

```markdown
## Relevant Links

* [Github](https://github.com/tldr-pages/tldr/)
* [Project Page](https://tldr.sh/)
```

Potentially a more future proof solution would be addign YAML headers to the files. This would mean that clients would need to be updated but would have the advantage of alowing us to add a lot more "meta" information to the pages. Potentially things like date created, date modified, a list of authors, original manpage etc although that is probably out of scope for this issue. Probably more relevant for #1082

waldyrious:
I like the idea of a metadata section, but perhaps it would make more sense at the bottom of the pages, rather than at the beginning.

Jab2870:
The only reason I say top is that there are other Markdown Syntaxes that use Yaml at the top [Pandoc](https://pandoc.org/MANUAL.html#extension-yaml_metadata_block) is the one I am most familier with although I think R Markdown does this as well.

jedahan:
I'd advocate for links to just be part of the natural description up top, where we provide context. A single link to a homepage feels in the spirit of `tl;dr`

waldyrious:
Yes, I agree the main link of the project feels central enough to justify placing it at the top, as part of the main description (rather than inside a metadata block).

From the options you presented, I like the autolink one the best.

sbrl:
[...]

PR #2660, opened 2018-12-20
multiple pages: add homepages

sbrl:
Finally! Got there in the end. Thanks for the assistance, @waldyrious :D

I ended up cherry-picking the commit from master, remembering that a cherry pick will change the commit hash.

For #2649.

### DIFF
--- a/pages/common/ansible-galaxy.md
+++ b/pages/common/ansible-galaxy.md
@@ -1,6 +1,7 @@
 # ansible-galaxy
 
 > Create and manage Ansible roles.
+> Homepage: <https://www.ansible.com/>.
 
 - Install a role:
 
--- a/pages/common/ansible-playbook.md
+++ b/pages/common/ansible-playbook.md
@@ -1,6 +1,7 @@
 # ansible-playbook
 
 > Execute tasks defined in playbook on remote machines over SSH.
+> Homepage: <https://www.ansible.com/>.
 
 - Run tasks in playbook:
 
--- a/pages/common/npm.md
+++ b/pages/common/npm.md
@@ -2,6 +2,7 @@
 
 > JavaScript and Node.js package manager.
 > Manage Node.js projects and their module dependencies.
+> Homepage: <http://npm.org/>.
 
 - Download and install a module globally:
 
--- a/pages/common/virtualenv.md
+++ b/pages/common/virtualenv.md
@@ -1,6 +1,7 @@
 # virtualenv
 
 > Create virtual isolated Python environments.
+> Homepage: <https://virtualenv.pypa.io/en/latest/>.
 
 - Create a new environment:
 
--- a/pages/common/7z.md
+++ b/pages/common/7z.md
@@ -1,6 +1,7 @@
 # 7z
 
 > A file archiver with high compression ratio.
+> Homepage: <https://www.7-zip.org/>.
 
 - Archive a file or folder:
 
--- a/pages/common/7za.md
+++ b/pages/common/7za.md
@@ -2,6 +2,7 @@
 
 > A file archiver with high compression ratio.
 > A standalone version of `7z` with support for fewer archive types.
+> Homepage: <https://www.7-zip.org/>.
 
 - Archive a file or folder:
 
--- a/pages/common/7zr.md
+++ b/pages/common/7zr.md
@@ -2,6 +2,7 @@
 
 > A file archiver with high compression ratio.
 > A standalone version of `7z` that only supports .7z files.
+> Homepage: <https://www.7-zip.org/>.
 
 - Archive a file or folder:
 
--- a/pages/common/ansible.md
+++ b/pages/common/ansible.md
@@ -2,6 +2,7 @@
 
 > Manage groups of computers remotely over SSH.
 > Use the /etc/ansible/hosts file to add new groups/hosts.
+> Homepage: <https://www.ansible.com/>.
 
 - List hosts belonging to a group:
 
--- a/pages/common/git.md
+++ b/pages/common/git.md
@@ -1,6 +1,7 @@
 # git
 
 > Distributed version control system.
+> Homepage: <https://git-scm.com/>.
 
 - Check the Git version:
 
--- a/pages/common/hyperfine.md
+++ b/pages/common/hyperfine.md
@@ -1,6 +1,7 @@
 # hyperfine
 
 > A command-line benchmarking tool.
+> Homepage: <https://github.com/sharkdp/hyperfine>.
 
 - Run a basic benchmark, performing at least 10 runs:
 
--- a/pages/common/mysql.md
+++ b/pages/common/mysql.md
@@ -1,6 +1,7 @@
 # mysql
 
 > The MySQL command-line tool.
+> Homepage: <https://www.mysql.com/>.
 
 - Connect to a database:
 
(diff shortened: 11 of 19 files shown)


pages/common/ansible-playbook.md
```diff
@@ -1,6 +1,7 @@
 # ansible-playbook
 
 > Execute tasks defined in playbook on remote machines over SSH.
+> Homepage: <https://www.ansible.com/>.
```
waldyrious: Wouldn't these subcommand pages benefit from a more specific link, similar to what you did with `mysqldump`? IMO the root URL should only be used for the main command.

pages/common/npm.md
```diff
@@ -2,6 +2,7 @@
 
 > JavaScript and Node.js package manager.
 > Manage Node.js projects and their module dependencies.
+> Homepage: <http://npm.org/>.
```
MasterOdin: Should this be https://www.npmjs.com/ or https://www.npmjs.com/package/npm?

pxgamer: Yeah, definitely npmjs.org 👍

agnivade:
> When I render this on Mac, the link doesn't seem to render:

Yeah .. I think the node client thinks of it as markdown and removes anything in `<>`

waldyrious:
> Nice! Didn't know we supported this.

There's been extensive discussion in #2649, check it out for more context :)

> the node client thinks of it as markdown and removes anything in `<>`

That's valid markdown, though, isn't it? Or does the node client not follow CommonMark?

agnivade:
It is actually the `:` which is causing this. Not sure why. Anybody is welcome to investigate.

waldyrious:
> It is actually the `:` which is causing this. Not sure why. Anybody is welcome to investigate.

Huh, that's weird. I would have assumed that it's some sort of html santitization pass that could be removing non-whiteslisted "tags". How did you conclude that the issue is the `:`?

agnivade:
By removing things from `Homepage: <http://example.org>` until the text rendered. :stuck_out_tongue:

pages/common/ansible-playbook.md
sbrl: Oh, good point! I haven't actually used _ansible_, so I must have skipped it. I'll do that now.

pages/common/npm.md
sbrl: As far as I'm aware, the domain is actually npmjs.org. As least, that's what I've always gone to.

sbrl:
What's blocking this from being merged? Did I forget something

pxgamer:
@sbrl, it's broken on the Node client for me still.

pages/common/npm.md
waldyrious: > As far as I'm aware, the domain is actually npmjs.org.

That's exactly what @MasterOdin is saying :)
Note that you wrote https://npm.org/ instead (without the "js"), which is the website for the National Association of Pastoral Musicians (of the US).

pages/common/ansible-galaxy.md
```diff
@@ -1,6 +1,7 @@
 # ansible-galaxy
 
 > Create and manage Ansible roles.
+> Homepage: <https://docs.ansible.com/ansible/latest/reference_appendices/galaxy.html#the-command-line-tool>.
```
waldyrious: Why not https://docs.ansible.com/ansible/latest/cli/ansible-galaxy.html, to match what you used for the `ansible-playbook` page?

I understand that the reference page is more informative, but I'm not sure it justifies the much harder to read URL. But in any case, we should use a consistent URL format for similar pages.

sbrl: Didn't notice! The Ansible documentation is rather confusing :P

I'll change it now.

sbrl:
I see, @pxgamer. In that case I think we're waiting on @agnivade then

agnivade:
It seems like @jedahan has already done that. It is just pending some finishing touches.

waldyrious:
@sbrl make sure to also address the NPM URL as discussed in the inline comments above :)

pages/common/virtualenv.md
```diff
@@ -1,6 +1,7 @@
 # virtualenv
 
 > Create virtual isolated Python environments.
+> Homepage: <https://virtualenv.pypa.io/en/latest/>.
```
waldyrious: I think the `/en/latest/` is appended automatically when visiting https://virtualenv.pypa.io, so IMO we could simplify the link here.

sbrl:
Done, @waldyrious :smiley_cat:

waldyrious:
I might suggest removing the slashes for the root-type URLs (`<https://foo.com/>` --> `<https://foo.com>`), just so they're a tad more readable.

pages/common/npm.md
```diff
@@ -2,6 +2,7 @@
 
 > JavaScript and Node.js package manager.
 > Manage Node.js projects and their module dependencies.
+> Homepage: <http://npm.com/>.
```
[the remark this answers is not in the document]
sbrl: Oops!

sbrl:
Interesting idea, @waldyrious!

I think I prefer the trailing slash though, as the url doesn't look complete without it.

waldyrious:
> I think I prefer the trailing slash though, as the url doesn't look complete without it.

I understand; I just think it looks inconsistent, since the github repo and npm package URLs don't contain slashes at the end, even though they don't point to a specific file like the URLs ending in `.html` do. I think we should either add it consistently, or remove it consistently. WDYT?

sbrl:
Didn't notice that! I suggest we add them consistently.

principis:
I just read #2649 
This is the first thing I've seen about this, why isn't this change mentioned on the wiki or somewhere else? It would break practically every client.

agnivade:
Any text is allowed in the command description. So we are not breaking any rules.

What _might_ break is the rendering of the text. We are waiting for a fix in node client before we merge this.

principis:
> Any text is allowed in the command description. So we are not breaking any rules.
> 
> What _might_ break is the rendering of the text. We are waiting for a fix in node client before we merge this.

Yes sure, I just think it should've been mentioned somewhere, it's annoying for clients to suddenly notice that something is broken. :) But I know it's difficult to go and notify everyone. Maybe we should add a clear specification on the wiki? Or maybe a specific page of upcoming changes to the spec. (I know the contributing guide exists).
Anyway I've fixed it. (Not that anyone but myself uses my client)

@elecprog Update your client. ;) ^

agnivade:
> Maybe we should add a clear specification on the wiki?

What do you want to mention ? A text like this might as well be on any page. So then, by that logic, all clients are already broken.

There no change in the spec here. We are just adding another line in the sub-heading. If that breaks any client, it was already broken in the first place.

principis:
@agnivade Nope you're right sorry. I was confused about <url> which I thought isn't valid markdown but it is... It's just that the spec for pages is a bit loose and up to interpretation. Maybe I sounded a bit rude which wasn't my intention.

Different question. If we add a "homepage" url do we always link to the homepage or sometimes just to the documentation?
For example: https://docs.ansible.com/ansible/latest/cli/ansible-galaxy.html
Or another example git-add https://git-scm.com/docs/git-add

sbrl:
@principis We're drafting a client spec over in #1065 if you're interested!

agnivade:
@sbrl - Feel free to merge this if there are no more pending tasks

sbrl:
Thanks, @agnivade! I don't think there are.

PR #2661, opened 2018-12-20
uprecords: add page

sbrl:
With my fancy new bash command, I identified that we didn't have a page for this already. Although uprecords, by default, shows you the top 10 longest uptimes, it's actually really useful as it tracks _every_ reboot you've done, what your kernel version was, and how long you were down for. All it needs to make it perfect is a reason column......

Here's said bash command:

```bash
while read nextcmd; do tldr "${nextcmd}" >/dev/null 2>&1; if [[ $? -ne 0 ]]; then echo $nextcmd; fi done < <(history | awk '{print $2}' | sort | uniq | grep -iP '^[a-zA-Z0-9]+$')
```

It'll search your entire shell history and display a list of all the commands you've used that don't yet have a tldr page.

### DIFF
--- a/pages/linux/uprecords.md
+++ b/pages/linux/uprecords.md
@@ -1,0 +1,23 @@
+# uprecords
+
+> Displays a summary of historical uptime records
+
+- Display a summary of the top 10 historical uptime records:
+
+`uprecords`
+
+- Display the top 25 records:
+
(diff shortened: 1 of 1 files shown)


pxgamer:
Merged as @rebeccabj's review didn't seem to include any suggestions.

PR #2663, opened 2018-12-21
add-apt-respository/adduser/alpine:add chinese translation

HugueLiu:


agnivade:
If we know someone, then sure. Otherwise I don't think we can do much. I have run a couple of them under google translate just as a smoke test.

pages.cn/linux/apk.md
```diff
@@ -0,0 +1,27 @@
+# apk
+
+> Alpine linux的包管理工具.
+
+- 从所有的远程仓库中更新仓库索引
```
sbrl: I believe a colon is needed at the end of this line, at least in English. Is this not the case in Chinese?

HugueLiu: yes, there should be a colon, it's my fault. I have corrected the error

waldyrious:
@HugueLiu shouldn't the tokens (`{{...}}`) be translated as well?

HugueLiu:
> @HugueLiu shouldn't the tokens (`{{...}}`) be translated as well?

I have though about that, but I'm not sure whether the contents in {{}} should be translated. If should, I will translate them.

agnivade:
Yes, please go ahead and translate them.

PR #2670, opened 2018-12-26
nodetool: add page

mspear:

### DIFF
--- a/pages/common/nodetool.md
+++ b/pages/common/nodetool.md
@@ -1,0 +1,11 @@
+# nodetool
+
+> Command line interface for managing a cassandra cluster.
+
+- Check the health of your cluster:
+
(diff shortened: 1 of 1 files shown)


pages/common/nodetool.md
```diff
@@ -0,0 +1,11 @@
+# nodetool
+
+> Command line interface for managing a cassandra cluster.
+
+- Check the health of your cluster:
+
+`nodetool --host {{host}} --username {{username}} --password {{password}} status`
```
sbrl: Specifying a password on the command-line isn't very secure. Is there a way of doing this without doing that?

agnivade: There seems to be a `--password-file` which takes the path to a password file. But I think the tool should prompt for the password if it is not passed explicitly. That is the ideal behavior.

PR #2671, opened 2018-12-26
pivpn: add page

sbrl:
Found this recently - it works like a charm! Just don't forget to allow it through your firewall like I did.....

### DIFF
--- a/pages/linux/pivpn.md
+++ b/pages/linux/pivpn.md
@@ -1,0 +1,29 @@
+# pivpn
+
+> Easy security-hardened OpenVPN setup and manager.
+> Originally designed for the Raspberry Pi, but works on other Linux devices too.
+> Homepage: <http://www.pivpn.io/>.
+
+- Install PiVPN:
+
+`curl -L https://install.pivpn.io | bash`
+
+- Add a new client device:
+
+`sudo pivpn add`
+
+- List all client devices:
+
+`sudo pivpn list`
+
+- List currently connected devices and their statistics:
+
+`sudo pivpn clients`
+
+- Revoke a previously authed device:
+
+`sudo pivpn revoke`
+
+- Uninstall PiVPN:
+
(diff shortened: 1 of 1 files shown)


pages/linux/pivpn.md
```diff
@@ -0,0 +1,29 @@
+# pivpn
+
+> Easy security-hardened OpenVPN setup and manager.
+> Originally designed for the Raspberry Pi, but works on other Linux devices too.
+> Homepage: <http://www.pivpn.io/>.
+
+- Install PiVPN:
```
agnivade: We usually do not mention install steps. Let us remove this.

pages/linux/pivpn.md
```diff
@@ -0,0 +1,29 @@
...
+
+- List all client devices:
+
+`sudo pivpn list`
+
+- List currently connected devices and their statistics:
+
+`sudo pivpn clients`
+
+- Revoke a previously authed device:
```
agnivade: authenticated.

pages/linux/pivpn.md
```diff
@@ -0,0 +1,29 @@
+# pivpn
+
+> Easy security-hardened OpenVPN setup and manager.
+> Originally designed for the Raspberry Pi, but works on other Linux devices too.
```
agnivade: I wonder if this line is necessary ?

pages/linux/pivpn.md
sbrl: Can do!

pages/linux/pivpn.md
sbrl: I think so. `Pi` is in the name, which is implies that it only work on Raspberry Pis.

pages/linux/pivpn.md
sbrl: Oops! Totally didn't see this comment here. It must have slipped through the net!

Changed.

PR #2676, opened 2018-12-30
inotify-wait: add page

sbrl:


pages/linux/inotify-wait.md
```diff
@@ -0,0 +1,27 @@
...
+
+`while inotifywait --recursive {{path/to/directory}} --exlude '{{regex}}'; do {{command}}; done`
+
+- Wait at most 30 seconds:
+
+`while inotifywait --timeout {{30}} {{path/to/file}}; do {{command}}; done`
+
+- Only watch for file modification events:
+
+`while inotifywait --event modify {{path/to/file}}; do {{command}}; done`
```
agnivade: What other event types are there ? If the list is small, then perhaps, we can show them in the command itself - {{modify|create|delete}}.

sbrl: The list is rather considerable, unfortunately. Here it is in full:

access, modify, attrib, close_write, close_nowrite, close, open, moved_to, moved_from, move, move_self, create, delete, delete_self, unmount

PR #2682, opened 2019-01-06
chflags: add page

yuriybash:

### DIFF
--- a/pages/osx/chflags.md
+++ b/pages/osx/chflags.md
@@ -1,6 +1,6 @@
 # chflags
 
-> Change file flags
+> Change file flags.
 
 - Set the `hidden` flag for a file:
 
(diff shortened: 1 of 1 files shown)


pages/osx/chflags.md
```diff
@@ -0,0 +1,19 @@
+# chflags
+
+> Change file or directory flags.
+
+- Set the `hidden` flag for a file:
+
+`chflags hidden {{/path/to/file}}`
```
agnivade: Please remove the preceeding `/`. And also put hidden within tokens (`{{}}`).

Same applies for all other examples too.

Also, is there a way to know all the flags supported by this command ?

yuriybash: Changes made. You can find a list of available flags [here](https://ss64.com/osx/chflags.html). My understanding is that `tldr` pages are meant to be fairly short, so it didn't seem appropriate to include a list of available flags in the page itself.

agnivade: Understood. I just wanted to check if there was a command to list all the flags.

PR #2684, opened 2019-01-07
Multiple pages: Add homepages

principis:
Added some more homepages, see: #2660 
It seemed handy for all the git pages to link directly to the docs although it's not really a homepage. But it is the same website so... Let me know if I need to change anything.

### DIFF
--- a/pages/common/gitk.md
+++ b/pages/common/gitk.md
@@ -1,6 +1,8 @@
 # gitk
 
 > A graphical git repository browser.
+> Homepage: <https://git-scm.com/docs/gitk>.
+> Homepage: <https://git-scm.com/docs/gitk>.
 
 - Show the repository browser for the current git repository:
 
--- a/pages/common/gitk.md
+++ b/pages/common/gitk.md
@@ -2,7 +2,6 @@
 
 > A graphical git repository browser.
 > Homepage: <https://git-scm.com/docs/gitk>.
-> Homepage: <https://git-scm.com/docs/gitk>.
 
 - Show the repository browser for the current git repository:
 
--- a/pages/common/youtube-dl.md
+++ b/pages/common/youtube-dl.md
@@ -1,6 +1,7 @@
 # youtube-dl
 
 > Download videos from YouTube and other websites.
+> Homepage: <http://rg3.github.io/youtube-dl/>.
 
 - Download a video or playlist:
 
--- a/pages/common/code.md
+++ b/pages/common/code.md
@@ -1,6 +1,7 @@
 # code
 
 > Visual Studio Code.
+> Homepage: <https://github.com/microsoft/vscode>.
 
 - Open VS Code:
 
--- a/pages/common/git-add.md
+++ b/pages/common/git-add.md
@@ -1,6 +1,7 @@
 # git add
 
 > Adds changed files to the index.
+> Homepage: <https://git-scm.com/docs/git-add>.
 
 - Add a file to the index:
 
--- a/pages/common/git-am.md
+++ b/pages/common/git-am.md
@@ -2,6 +2,7 @@
 
 > Apply patch files. Useful when receiving commits via email.
 > See also `git format-patch`, which can generate patch files.
+> Homepage: <https://git-scm.com/docs/git-am>.
 
 - Apply a patch file:
 
--- a/pages/common/git-bisect.md
+++ b/pages/common/git-bisect.md
@@ -2,6 +2,7 @@
 
 > Use binary search to find the commit that introduced a bug.
 > Git automatically jumps back and forth in the commit graph to progressively narrow down the faulty commit.
+> Homepage: <https://git-scm.com/docs/git-bisect>.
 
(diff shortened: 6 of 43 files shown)


pages/common/gitk.md
```diff
@@ -1,6 +1,8 @@
 # gitk
 
 > A graphical git repository browser.
+> Homepage: <https://git-scm.com/docs/gitk>.
+> Homepage: <https://git-scm.com/docs/gitk>.
```
blunket: You probably didn't mean to dupe this one :slightly_smiling_face:

blunket:
(Sorry for my previous comment -- I screwed up, lol)

principis:
> (Sorry for my previous comment -- I screwed up, lol)

NP :) my script messed it up.

waldyrious:
@sbrl what happened to your insistence on using the slashes at the end of URLs? ;)

In all seriousness, though, I think it might be better if we don't include them because people are more likely to write links without a trailing slash than with them. If we insist in keeping them, we'd have to constantly remind people of them -- case in point -- which is a bit of a drag for everyone.

Btw, @principis, I agree that using specific pages for each subcommand is more useful than linking all of them to the homepage of the main command. Perhaps we should use a different term than "homepage" for the main link, to avoid giving out the wrong idea. WDYT?

principis:
@waldyrious I added slashes consistently with the websites as was said in #2660 I find them ugly so I would rather remove them but it's not that big of a deal.

> Perhaps we should use a different term than "homepage" for the main link, to avoid giving out the wrong idea. WDYT?

Ow I forgot to mention that sorry :laughing: It was my intention to mention it in my PR.
I couldn't think of a good term but I think it's a good idea.

jedahan:
maybe we don't specify and let people put whatever they think is most appropriate? Most likely it will be
`homepage: http://mycoolproject.org`
but sometimes
`documentation: http://project.readthedocs.org`
and maybe even
`community: http://project.gitter.im`

sbrl:
@waldyrious: Oops! Missed that :P I'm unsure that it's worth ensuring that _all_ urls are formatted in _exactly_ the same way though.

@jedahan: Sounds like a great idea!

sbrl:
Sounds good to me, @waldyrious

pages/common/youtube-dl.md
```diff
@@ -1,6 +1,7 @@
 # youtube-dl
 
 > Download videos from YouTube and other websites.
+> Homepage: <http://rg3.github.io/youtube-dl/>.
```
vladimyr: Prefer `https` if it is available :bowtie:

vladimyr: @sbrl Is this linter (https://github.com/tldr-pages/tldr-lint) still used for validating tldr pages? If so this should be added as rule there...

agnivade: Did you mean to comment on this PR ? What has tldr-lint got to do with adding a homepage link ?

vladimyr: In ideal world tldr-lint or any other linter used for checking tldr pages should report if http (home)page is linked and https version is available. It should report dead links also?

agnivade: Oh I see. I think that would put too much logic into the linter. I would want the heading text to be free-form without any strict format. The `Homepage:` format is just something we agreed to. I think it is not necessary to have linter checks for that.

vladimyr: Maybe I wasn't clear enough. I wasn't talking about homepage specifically but all urls page may contain. If url is found and uses less secure scheme or points to dead end that should be reported...

agnivade: I see. I am neutral to it.

sbrl: Some commands' homepages may be only available over http and not https though.

vladimyr: Quoting myself:
> if http (home)page is linked and https version is available.

That is relatively easy to check.

vladimyr: This is probably not a place to discuss it but parsing tldr pages with bison is like killing fly with orbital ion cannon 🙄
After all tldr page is markdown document and there are plenty of markdown parsers around that can be adapted/wrapped to check page format. Such parser could provide hooks for specific page segment like urls which can in turn be used for http/https and/or dead link checking.

sbrl:
What's remaining for getting this and #2660 merged, @agnivade?

agnivade:
We were waiting for the node client to have home pages support. Currently, the node client breaks trying to display the home page link. 

@vladimyr - You had mentioned about subclassing the markdown parser to add the support. Are you planning to do that ?

vladimyr:
Not atm :disappointed: Right now I don't have enough free time for doing it properly including extensive test suite but maybe some time in the future...

vladimyr:
Sure, totally agree 👍

agnivade:
Done. Merging this.

pages/common/youtube-dl.md
agnivade: Couldn't agree more. Sadly, that is the state of affairs now. If someone wants to rewrite the linter completely, they are more than welcome.

PR #2685, opened 2019-01-07
sed: fix wrong example description.

mebeim:

---

Just fixed the first example on the `sed` page. Sed by default replaces **all** occurrences of a string in a file, not only the first one.


mebeim:
@sbrl oh, you're totally right, I got mislead by my own edit. I updated the PR, now is much clearer.

PR #2679, opened 2019-01-04
replace cn with zh; apt/aptitude/arch: add chinese translations

HugueLiu:
- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform folder (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).

Closes #2677

PR #2687, opened 2019-01-08
move Chinese pages from pages.cn to pages.zh

waldyrious:
Also add new Chinese translations for apt, aptitude and arch.

Note: this is a re-submission of #2679, which I inadvertently closed when trying to clean up its commits. @HugueLiu's authorship of the changes is preserved.


HugueLiu:
Thanks very much for your work.

agnivade:
Yep

PR #2694, opened 2019-01-09
Add support for building snaps

igorljubuncic:
**TL;DR:**

Please add support for snaps.

**TL:**

This PR adds support for building a snap package of tldr.

Snaps are cross-distro Linux software packages. One snap can be installed on all supported Ubuntu LTS and non LTS releases from 14.04 onward. Additionally, they can installed on Debian, Manjaro, Fedora, OpenSUSE and many others distributions. If you make a snap of tldr, it will be available in the Snap Store (snapcraft.io/store), which is used an audience of millions, so this could also give you extra exposure to your project. Snaps also come with automatic updates, so any update you make to the man pages, will be automatically reflected.

I built tldr locally on a stock Ubuntu 18.04 system, with the following steps:

`snap install snapcraft --classic --beta`
`git clone https://github.com/igorljubuncic/tldr.git`
`cd tldr`
`git checkout add-snapcraft`
`snapcraft`

The generates a <tldr-version>.snap file, something like tldr_1.0_amd64.snap.

`snap install tldr_1.0_amd64.snap --dangerous`

This install and tests tldr LOCALLY (before going to the store). The --dangerous flag is necessary because we’re installing an app that hasn’t gone through the snap store review process and isn't signed.

Once installed, the (namespaced) tldr command can be executed, e.g.:

`snap run tldr tar`

If landed, you will need to:

- Register a developer account in the snap store https://snapcraft.io/account.
- Register the tldr name in the store. 

`snapcraft login`
`snapcraft register`

Upload a built snap to the store:

`snapcraft push tldr_1.0_amd64.snap --release edge`

Test install from the store:

`snap install tldr --edge`

The store supports multiple risk levels as “channels” - edge, beta, candidate and stable. Typically, you will upload to edge and after testing, promote to other channels. Once you push tldr to stable, you can then update the store page with the logo and description, and we can help you with the promotion. We'd be glad to feature your application in our store.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,19 @@
+name: tldr
+version: '1.0'
+summary: Simplified and community-driven man pages
+description: |
+  Simplified and community-driven man pages
+grade: stable
+confinement: strict
+base: core18
+parts:
+  tldr:
+    plugin: python
+    python-version: python2
+    source: https://github.com/tldr-pages/tldr-python-client.git
+apps:
+  tldr:
+    command: bin/tldr
+    plugs:
+      - home
+      - network


agnivade:
We already have a snap for tldr. See the node client.

Moreover, this should be in a client repo. This is the repo for pages.

PR #2699, opened 2019-01-10
Add an explanation of subcommands

baruchiro:

I searched for a record for a `git checkout` and nearly opened an issue on it.
I think if there is a user who is less engaged in open source, he would not know how to get records for subcommands.

### DIFF
--- a/pages/common/git.md
+++ b/pages/common/git.md
@@ -14,6 +14,10 @@
 
 `git help {{command}}`
 
+- Call tldr on a command:
+
+`tldr {{git-command}}`
+
 - Execute Git command:
 
 `git {{command}}`
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -5,3 +5,7 @@
 - Get typical usages of a command (hint: this is how you got here!):
 
 `tldr {{command}}`
+
+- For typical usages of a subcommand:
+
+`tldr {{command-subcommand}}`


pages/common/tldr.md
```diff
@@ -5,3 +5,7 @@
 - Get typical usages of a command (hint: this is how you got here!):
 
 `tldr {{command}}`
+
+- For typical usages of a subcommand:
+
+`tldr {{command-subcommand}}`
```
agnivade: This is not true. You don't need the `-`. You can do "tldr git merge" and it works fine.

pages/common/git.md
```diff
@@ -14,6 +14,10 @@
 
 `git help {{command}}`
 
+- Call tldr on a command:
+
+`tldr {{git-command}}`
```
agnivade: There should not be a tldr command in a git page.

pages/common/tldr.md
baruchiro:
```bash
$ tldr git merge
Invalid argument `merge'

Usage: tldr [-v|--version] [--update] COMMAND
  tldr Client program

```

Ubuntu 18.10

pages/common/git.md
baruchiro: I understand, but when a user search for help with `git checkout`, how he can find the result?
If you say that he gets the result from `tldr tldr`, I will accept.

Accurate0: $ tldr git-checkout

git checkout

Checkout a branch or paths to the working tree.
...

Although most clients automatically add a "-" between words.

baruchiro: @Accurate0 , I know this now, but

> I searched for a record for a `git checkout` and nearly opened an issue on it.
> I think if there is a user who is less engaged in open source, he would not know how to get records for subcommands.

Just after traveling in this repo, I found the solution. So if there is a user that not familiar with Github or open-source, how he would find the solution?

pages/common/tldr.md
agnivade: That is a client specific thing. The node client handles it fine. Since this is not in any spec, I don't think we should mention it.

pages/common/git.md
agnivade: The most basic tendency is to write the exact command; which in this case is "git checkout". So if you do "tldr git checkout", it should give the correct result. And the node client indeed does that. So I don't see any confusion here.

Now, if some client does not implement this feature, and the require you to pass a `-` or whatever, I would expect the person to file an issue with the appropriate client repo asking what the solution is.

Please note that this repo is just for pages, and we do not control the client authors. There are lots of clients in lots of programming languages. If you have any confusion regarding how to find a command, you should ask that client's owner.

Does that clarify things ?

pages/common/tldr.md
sbrl: So does the bash client. Perhaps you could open an issue against your client's repo?

_(For my own reference, this is worth including in the official client spec that's being drafted in #1065)_

pages/common/git.md
baruchiro: Yes, Thank you

PR #2702, opened 2019-01-11
tr: improve page

waldyrious:
- Add source data to all examples
- Add an example for piping from another command
- Improve punctuation in main description
- Add token markup {{}} to one example that was missing it
- Remove redundant "and print result" from the last two examples
  (it's true for all the ones shown)

### DIFF
--- a/pages/common/tr.md
+++ b/pages/common/tr.md
@@ -1,27 +1,31 @@
 # tr
 
-> Translate characters - run replacements based on single characters and character sets.
+> Translate characters: run replacements based on single characters and character sets.
 
 - Replace all occurrences of a character in a file, and print the result:
 
-`tr {{find_characters}} {{replace_characters}} < {{filename}}`
+`tr {{find_character}} {{replace_character}} < {{filename}}`
+
+- Replace all occurrences of a character from another command's output:
+
+`echo {{text}} | tr {{find_character}} {{replace_character}}`
 
 - Map each character of the first set to the corresponding character of the second set:
 
-`tr 'abcd' 'jkmn' < {{filename}}`
+`tr '{{abcd}}' '{{jkmn}}' < {{filename}}`
 
 - Delete all occurrences of the specified set of characters from the input:
 
-`tr -d '{{input_characters}}'`
+`tr -d '{{input_characters}}' < {{filename}}`
 
 - Compress a series of identical characters to a single character:
 
-`tr -s '{{input_characters}}'`
+`tr -s '{{input_characters}}' < {{filename}}`
 
-- Translate the contents of the file to upper-case and print result:
+- Translate the contents of a file to upper-case:
 
 `tr "[:lower:]" "[:upper:]" < {{filename}}`
 
-- Strip out non-printable characters from the file and print result:
+- Strip out non-printable characters from a file:
 
 `tr -cd "[:print:]" < {{filename}}`


mebeim:
IMHO for the last two examples saying "and print the result" is clearer since the file is not modified in place (see e.g. [`sed`](https://github.com/tldr-pages/tldr/blob/master/pages/common/sed.md), [`node`](https://github.com/tldr-pages/tldr/blob/master/pages/common/node.md), etc).

pages/common/tr.md
```diff
@@ -1,27 +1,31 @@
...
 
 - Map each character of the first set to the corresponding character of the second set:
 
-`tr 'abcd' 'jkmn' < {{filename}}`
+`tr '{{abcd}}' '{{jkmn}}' < {{filename}}`
 
 - Delete all occurrences of the specified set of characters from the input:
 
-`tr -d '{{input_characters}}'`
+`tr -d '{{input_characters}}' < {{filename}}`
```
sbrl: The extra space after the `<` can cause unexpected effects on some systems.

pages/common/tr.md
```diff
@@ -1,27 +1,31 @@
...
 
 - Compress a series of identical characters to a single character:
 
-`tr -s '{{input_characters}}'`
+`tr -s '{{input_characters}}' < {{filename}}`
```
sbrl: Same here.

pages/common/tr.md
waldyrious: Oh, really? I didn't know about that. Can you point me to some resources about this?

waldyrious:
> IMHO saying "and print the result" is clearer since the file is not modified in place.

Yes, agreed.

> You think it would be too pedantic to add it to each example?

Pedantic is not the word I'd use -- more like redundant (see below).

> I think mentioning `and print the result` once on the first example is fine - it's consistent with the style of other pages.

It's not merely a matter of following existing conventions (in fact, we should welcome people questioning them, as @mebeim did). In this case, it actually can be directly justified from the [content guidelines](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines), particularly item 4:

> Introduce options gradually, starting with the simplest command invocations, and using more complex examples progressively.

I.e., commands are _expected_ to build from the previous ones, so that more advanced notions can be presented without making each description increasingly longer. This also forces us to make the first few examples simple and well described enough to be confident that people will understand the pattern and not need the same concepts to be repeated in the later examples' descriptions.

pages/common/tr.md
sbrl: I can't, actually! I got so confused I've opened [a question on the Unix StackExchange](https://unix.stackexchange.com/questions/494506/spaces-after-io-redirection-operator) about it.

waldyrious: I see. I'm no expert but IIUC 2>&1 is a special redirect construct. [This answer](https://stackoverflow.com/a/818265/266309) and [this one](https://stackoverflow.com/a/41319194/266309) may help.

agnivade: From the answer [here](https://unix.stackexchange.com/questions/494506/spaces-after-io-redirection-operator), looks like spaces are fine.

agnivade: btw, this example was plain wrong. I wonder how the submitter missed this. From the git log, I see this was submitted in 2015 - https://github.com/tldr-pages/tldr/commit/70017a62004c2c2440bda4f209e43e6ac23d6628#diff-c850ca507cb0f59baf2c23812405abaa.

agnivade: yes.

PR #2703, opened 2019-01-11
Add contribution guidelines for translations.

mebeim:
Added guidelines on how to contribute with new translations and clear specification of top level language specific directory names, [as previously agreed](https://gitter.im/tldr-pages/tldr/archives/2019/01/11).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,29 +73,27 @@ refer to the [style guide](contributing-guides/style-guide.md).
 
 ## Translations
 
-Translation of pages can be done by simply creating the corresponding page in the appropriate top level language specific directory under the appropriate platform subdirectory.
-When translating a page into a language that is not present yet, a new language specific directory must be created first.
+Translation of pages can be done by simply creating the corresponding page within the appropriate language-specific directory, creating that as well if it does not already exist.
 
-Language specific directories must follow the pattern `pages.<language_tag>`, where `<language_tag>` is a [BCP 47](https://tools.ietf.org/html/bcp47) conforming tag in the form `<language>-<region>`, where:
+Language specific directories must follow the pattern `pages.<language_tag>`, where `<language_tag>` is a [BCP 47](https://tools.ietf.org/html/bcp47) conforming tag in the form of `<language>[-<region>]`, where:
 
  - `<language>` is the shortest [ISO 639](https://en.wikipedia.org/wiki/ISO_639) language code for the chosen language (see [here](https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes) for a complete list).
  - `<region>` is the two-letter [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) region code for the chosen region (see [here](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements) for a complete list).
 
-Some examples:
+The `<region>` code is optional and should only be added when it is needed to distinguish between an already existing `<language>` and one of its regional dialects. As an example, both `fr-FR` and `fr-BE` should fall under the same `pages.fr` directory, since there virtually is no difference in writing between standard French and Belgian French.
 
- - English (United Kingdom): `pages.en-GB`.
- - Chinese (Taiwan): `pages.zh-TW`.
- - Portuguese (Brazil): `pages.pt-BR`.
+Some examples of valid BCP 47 tags:
+
+ - French: `fr`.
+ - Chinese: `zh`.
+ - Chinese (Singapore): `zh-SG`.
+ - Portuguese (Brazil): `pt-BR`.
 
 You can check the validity of BCP 47 tags [here](http://schneegans.de/lv/).
 
 ### Default language for newly added pages
 
-The default language used for pages is English (US). Pages written in English are stored in the default `pages` directory (notice the absence of a specific language tag). Translations should be added only for pages that already exist in the default language. If you'd like to add a new page in a different language, please consider creating the English page first.
-
-### Different regions for the same language
-
-For the same language, if there virtually is no difference in writing between the country of origin and another country in which the language is spoken, then the former should always be preferred. In such cases, please choose the region code of the country of origin, to avoid fragmentation of pages for the same language. As an example, both `fr-FR` and `fr-BE` should fall under the same `pages.fr-FR` directory, since there virtually is no difference in written French between France (country of origin) and Belgium. The same goes for `en-US`, `en-GB`, `en-CA` (and so on), which should all fall under the default `pages` directory.
+The default language used for pages is English (US). Pages written in English are stored in the default `pages` directory (notice the absence of a specific language tag). Although not strictly required, if you'd like to add a new page in a different language, please consider creating the English page too.
 
 ## Submitting a pull request
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,6 +71,24 @@ As a quick reference, the format of each page should match the following templat
 For more detailed page formatting guidelines,
 refer to the [style guide](contributing-guides/style-guide.md).
 
+## Translations
+
+Translation of pages can be done by simply creating the corresponding page in the appropriate top level language specific directory under the appropriate platform subdirectory.
+When translating a page into a language that is not present yet, a new language specific directory must be created first.
+
+Language specific directories must follow the pattern `pages.<language_tag>`, where `<language_tag>` is a [BCP 47](https://tools.ietf.org/html/bcp47) conforming tag in the form `<language>-<region>`, where:
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -71,6 +71,24 @@ As a quick reference, the format of each page should match the following templat
...
+When translating a page into a language that is not present yet, a new language specific directory must be created first.
+
+Language specific directories must follow the pattern `pages.<language_tag>`, where `<language_tag>` is a [BCP 47](https://tools.ietf.org/html/bcp47) conforming tag in the form `<language>-<region>`, where:
+
+ - `<language>` is the shortest [ISO 639](https://en.wikipedia.org/wiki/ISO_639) language code for the chosen language (see [here](https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes) for a complete list).
+ - `<region>` is the two-letter [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) region code for the chosen region (see [here](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements) for a complete list).
+
+Some examples:
+
+ - English (United Kingdom): `pages.en-GB`.
```
sbrl: I'm unsure whether we should include English as an example here, as the pages are, by default, already in English.

mebeim: Changed!

mebeim:
Added some clarifications on the default language and different regions using the same language. Not sure if it could be worded better, comments are welcome 🙂

CONTRIBUTING.md
```diff
@@ -71,6 +71,32 @@ As a quick reference, the format of each page should match the following templat
 For more detailed page formatting guidelines,
 refer to the [style guide](contributing-guides/style-guide.md).
 
+## Translations
+
+Translation of pages can be done by simply creating the corresponding page in the appropriate top level language specific directory under the appropriate platform subdirectory.
```
agnivade: top-level

CONTRIBUTING.md
```diff
@@ -71,6 +71,32 @@ As a quick reference, the format of each page should match the following templat
...
+When translating a page into a language that is not present yet, a new language specific directory must be created first.
+
+Language specific directories must follow the pattern `pages.<language_tag>`, where `<language_tag>` is a [BCP 47](https://tools.ietf.org/html/bcp47) conforming tag in the form `<language>-<region>`, where:
```
agnivade: in the form _of_ ..

mebeim:
Applied requested changes 😁

CONTRIBUTING.md
```diff
@@ -71,6 +71,32 @@ As a quick reference, the format of each page should match the following templat
 For more detailed page formatting guidelines,
 refer to the [style guide](contributing-guides/style-guide.md).
 
+## Translations
+
+Translation of pages can be done by simply creating the corresponding page in the appropriate top-level language specific directory under the appropriate platform subdirectory.
```
waldyrious: We can make the sentence a bit easier to chew by forgoing some accuracy, since that's explained in more detail below. Here's what I'd suggest:

```suggestion
Translation of pages can be done by simply creating the corresponding page in the appropriate language-specific directory.
```

CONTRIBUTING.md
```diff
@@ -71,6 +71,32 @@ As a quick reference, the format of each page should match the following templat
...
+When translating a page into a language that is not present yet, a new language specific directory must be created first.
```
waldyrious: I'd suggest simplifying here too, merging this with the next paragraph:

> If the language-specific directory for the desired language does not yet exist, it must be created first, following the pattern ...

CONTRIBUTING.md
```diff
@@ -71,6 +71,32 @@ As a quick reference, the format of each page should match the following templat
...
+
+ - Chinese (Taiwan): `pages.zh-TW`.
+ - Portuguese (Brazil): `pages.pt-BR`.
+ - German (Germany): `pages.de-DE`.
+
+You can check the validity of BCP 47 tags [here](http://schneegans.de/lv/).
+
+### Default language for newly added pages
+
+The default language used for pages is English (US). Pages written in English are stored in the default `pages` directory (notice the absence of a specific language tag). Translations should be added only for pages that already exist in the default language. If you'd like to add a new page in a different language, please consider creating the English page first.
```
waldyrious: What's the motivation for requiring the English page first?

waldyrious:
I made some inline comments, but the main change I'd propose is that the region code should be optional (which is also what BCP 47 specifies, IIRC), and only added when it's necessary for disambiguation. So there would be `pages.fr` and `pages.de`, rather than `pages.fr-FR` and `pages.de-DE`.

This heuristic also allows us to drop the last section, which is quite important, since the longer documentation is, the less likely people are to read it in full.

agnivade:
> I made some inline comments, but the main change I'd propose is that the region code should be optional (which is also what BCP 47 specifies, IIRC), and only added when it's necessary for disambiguation.

@waldyrious - Hmm, I see. I thought they would be compulsory. So in case of languages that can be represented by more than one code, what will be the default region ? According to you, `pt` and `pt-BR` can coexist, so we don't need the region for `pt` ?
See https://gitter.im/tldr-pages/tldr?at=5c389d6aa57b501bcfea16ee

CONTRIBUTING.md
mebeim: What about:

> Translation of pages can be done by simply creating the corresponding page within the appropriate language-specific directory, creating that as well if it does not already exist.

mebeim:
Thanks, @waldyrious, I reworded simplifying the first paragraph following your suggestion. Looks way cleaner now. 

About the BCP47 codes, as @agnivade pointed out the choice was a result of discussing this on Gitter. I thought always having the region too would simplify the work for clients, but it's rather simple to handle either way. So I also think that using the language code alone would be cleaner, I changed that too.

As for the English being the default language, I just assumed it myself since it applies to the current situation. I made it only a suggestion thought, not a strict rule as it previously was.

---

Let me guys know what do you think about the currend draft.

waldyrious:
> According to you, pt and pt-BR can coexist, so we don't need the region for pt ?

Yes, that's what I typically see with localization of software projects.

> I also thought always having the region too would simplify the work for clients

That's a good point, but it depends on how the client decides what language to show. There are several possible approaches: only when the user explicitly asks for it, by passing an argument like `--lang=xx` every time tldr is invoked; or with a configuration file where the user stores their preferred language (or even a fallback chain); or the client could try detecting the OS interface language or keyboard layout; web clients may want to filter by IP, etc.

So there's no telling whether a consistent xx-XX format will help with the implementation or not in advance. Not to mention that clients may want to, say, show "pt-BR" if "pt" or "pt-PT" is requested but nonexistent -- so there could be a set of nontrivial heuristic mapping between language codes that would not get that much help from the code being constant length.

> I also thought that English would probably be a standard fallback for clients if a page is not present in the chosen language

Again, that depends a lot on implementation choices and user preferences. I, for one, would prefer English at the top of my fallback chain, rather than the opposite. Others may want it in the middle.

The point is, if I identify N languages I can read tldr pages in, regardless of the order, the client should show the first one it finds, and show an error if the page does not exist in any of the languages I specify. There's no need to handle English differently for the scheme to work well.

> Let me know what you guys think about the current draft.

I'll re-review this later today, time permitting. Thanks for the submission btw :)

waldyrious:
@mebeim would you like to keep multiple commits in the final merge? Looking at the PR, I appreciate that you separated the commits and described them individually, but I'm not sure they all represent stages we'd want to preserve in the repository history (for example, I don't think f12a2d0 and 39005e2 stand alone as atomic commits). Would you like to clean up the branch manually, or should we squash the commits into a single one when merging?

mebeim:
@waldyrious cleaned up, thanks for the suggestion.

waldyrious:
@mebeim sorry, I didn't mean to present those two commits as an exhaustive listing -- as I said, they were just examples. I still don't think it makes sense to have 7f651f2 and 5674e90 as separate commits, since one makes a very minor change, and the other reworks other text introduced earlier in this same PR. That would mean including back-and-forth changes in the same PR, which is fine for the review process, but not that useful as part of the repo's definitive history. Does that make sense?

That said, I don't mean to make this cumbersome to you. If you feel this cleanup is too much work, let us know how you'd prefer proceeding.

mebeim:
Oh yes, you're totally right, sorry for the misunderstanding. I tidied up the branch and it now looks fine to me. I don't think there's the need, but feel free to apply further edits yourself if you want.

waldyrious:
It's cool, 3 commits with hand-crafted messages is more than we usually get :) Thanks for your patience! Let's get this merged.

agnivade:
Thank you for your patience @mebeim. Much appreciated.

mebeim:
np guys, love to support such a beautiful project ;)

PR #2706, opened 2019-01-12
CLIENT_SPECIFICATION.md: Write initial client specification

### LINKED ISSUE #1065: Define spec for officially sanctioned clients

sbrl:
It's finally happened! It's taken me all afternoon, but I've taken my previous notes & all the feedback, and folded it into a full specification, which I'm officially proposing with this PR.

Closes #1065 (finally!)


CLIENT_SPECIFICATION.md
```diff
@@ -0,0 +1,194 @@
...
+Finally, the [`LANGUAGE` environment variable](https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html) specifies a priority list of languages that a user wishes to read in. If present, a client MUST use the defined priority list to decide on the language that to display in.
+
+If a page is not available in the user's preferred language, then a client MUST respect the user's priority list defined in the `LANGUAGE` variable (if specified), and MAY choose to notify the user that a page in their chosen language couldn't be found (perhaps along with a link to the [translations section of the contributing guide](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#translations)).
+
+## Other Considerations
+This section contains a number of other items that don't neatly fit into any of the sections defined above.
+
+If appropriate, it is RECOMMENDED that clients implement a cache of pages. If implemented, clients MUST download the _entire_ archive from **UNKNOWN_LOCATION**.
+
+Additionally, clients MAY automatically the cache on a regular basis.
```
mebeim: May automatically _update_ the cache

pepa65: Some hopefully helpful comments:
* Why is it called `tldr-pages`? Is the name `tldr` reserved for something else??
* `It is _not_ a specification of the format of the pages themselves` -- I am guessing you mean the output format, but this could also be taken to refer to the storage format for the pages, so a clarifying word would be helpful.
* Why MUST `-a`/`--list-all` list only one page per line? I think this is usually not what people want for ease of use. I would prefer an additional separate commandline option `-n`/`--no-formatting` that lists everything plain: no colours, only 1 line per page in a listing.
* If this is **only a specification of how a user should be able to interface with an official client** then why specify the `Directory Structure`? That should be irrelevant, and in no need of specification. Same for `Translation`. But if it is important to include these sections, better adjust the language in the introduction, so that it is not only about the interface.

CLIENT_SPECIFICATION.md
```diff
@@ -0,0 +1,194 @@
...
+linux/sh
+osx/brew
+```
+
+Example invocations are as follows:
+
+```bash
+tldr windows/type
+# --foo is a custom argument, as described above
+tldr --foo bar common/git merge
```
jedahan: I have never seen a commandline application that uses `/` as anything but a path separator. I teach an introduction to the commandline class and it will be very confusing for students to see this formatting. It is also telling that this kind of formatting is rarely supported well by cli argument parsing libraries.

I am not sure if this is the best place to argue for a commandline option to disambiguate via `--os`. I feel like it will be similar to `--language`.

CLIENT_SPECIFICATION.md
```diff
@@ -0,0 +1,194 @@
...
+## Directory Structure
+This section documents the directory structure that contains the pages themselves.
+
+The master version of every page is stored inside (but not directly) the `pages` directory. Inside this directory, there is a folder for each platform - for example `windows`, `linux`, and the special `common` platform:
+
+ - `pages/`
+	 + `common/`
+	 + `linux/`
+	 + `windows/`
+	 + `osx/`
```
jedahan: should this be macos instead of osx?

CLIENT_SPECIFICATION.md
jedahan: (and I've seen clients that allow you to just give a path to a markdown file to render it)

CLIENT_SPECIFICATION.md
jedahan: Sorry, proposal to rename osx to macos should be a separate issue. This document describes the state of the pages now, and is therefore the right thing to do.

CLIENT_SPECIFICATION.md
mebeim: It could be a good idea to allow something like `tldr platform command` like `man` does with `man section command` :)

vladimyr:
👍for your effort and produced result. Looks really nice.

However I spotted one _wild_ inconsistency. In the same time supporting `--update` (cache) cli flag is mandated but later on whole cache functionality is declared as optional? 🙄
Am I missing something? 🤔

CLIENT_SPECIFICATION.md
jedahan: so the only tricky thing there is the spec mentions `tldr platform command` should be disambiguated as `tldr platform-command`. That said, matching `linux|macos|winndows` wouldnt be so difficult.

vladimyr: And then you have following issue. What's expected result of this hypothetical case: `tldr osx cleaner`. Is it `cleaner` app tldr page on `osx` platform or simply an app named `osx cleaner` with `osx-cleaner` tldr page? 🤔

pepa65: Your proposal of `tldr PLATFORM PAGE NAME` syntax could work, because so far there are no pages that start with any of the platforms that are supported. It would be even easier to type, and take care of @jedahan's objections. I initially chose `/` because in the current archive implementation, it does boil down to being a path separator (and it is shorter to type than some --prefix and easier to remember). In the current tldr-bash-client I could easily start supporting both forms. (This won't really help the client specification though..!)

pepa65: (By the way, the syntax to specify the path to a markdown file for rendering does require a commandline flag, and doesn't clash with the use of `/` in a normal page lookup.)

sbrl: I see. The pages do actually have 1 folder per platform though. Also, I see the `/` as a 'category' separator, sort-of meaning "inside".

So my typing `tldr linux/tar`, I'm requesting the `tar` page in the `linux` platform.

CLIENT_SPECIFICATION.md
sbrl: Probably, but it's probably worth opening an issue for that and handing that as a separate issue. The issue with that rename is that it'll break clients.

CLIENT_SPECIFICATION.md
agnivade: What is the issue with having a flag for platform, like `--platform=osx` ? Seems unambiguous to me.

sbrl: The thing is that `tldr platform/page` requires much less typing that `--platform=os_name`

vladimyr: By going with forward slashes it becomes path-like so it is hard to guess what's really happening here (for newcomers) as @jedahan already pointed out. Also it will look superweird on windows because of forward slashes. If you really want to avoid flags that much use something less ambigous/extravagant like `tldr platform:page` :bulb:

CLIENT_SPECIFICATION.md
vladimyr: FWIW it should probably be darwin rather than macos but I suspect that to be user _unfriendly_ 🤔 TBH this platform division is cut at wrong layer completely. Interesting bit isn't an os or kernel but userland tools family - whether it is gnu or bsd.

CLIENT_SPECIFICATION.md
agnivade: Hmm .. IMO I think the amount of time saved by not typing a few characters is not worth the confusion caused due to it.

jedahan: `--platform, -p` or `--os, -o` if its kinda short.
I set this anyway because its usually what I want: `h() { tldr -q $* || tldr -q -o linux $* || tldr -q -o windows $* }`

CLIENT_SPECIFICATION.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Command-Line Interface
+This section describes the standardised command-line interface (CLI) for clients implementing one. Clients that do not provide a CLI can ignore this section.
+
+### Arguments
+A number of arguments MUST be supported (unless otherwise specified if a CLI is implemented:
+
+Argument		        | Meaning
+------------------------|--------------------
+`--update`				| Updates the offline cache of pages. MUST NOT be implemented if cache is not supported.
```
jedahan: how about
`Updates the offline cache of pages. MUST be implemented if cache is supported`

CLIENT_SPECIFICATION.md
```diff
@@ -0,0 +1,194 @@
...
+`--version`, `-v`		| Shows the current version of the client, and the version of this specification that it implements.
+`--list-all`, `-a`		| Lists all the pages in the current platform to the standard output. One page name per line. Additional decoration MAY be printed to the standard error.
```
jedahan: how about `--list`, `-l`? This way it is similar to `ls -l` (one per line), and shorter to type.

CLIENT_SPECIFICATION.md
sbrl: In that case, we need to decide between `--platform` or `--os`. I do like `platform:page`  though.

CLIENT_SPECIFICATION.md
sbrl: Hrm, you raise a good point. What do you think, @agnivade / @pxgamer / @waldyrious?

pxgamer: I think this was because the node client uses `--list-all` :thinking_face:

sbrl: I think `--list-all` is rather confusing, actually. It implies that all pages are listed, but in reality it's only the pages in the current platform & common that are listed.

Perhaps we can disambiguate this somehow?

CLIENT_SPECIFICATION.md
vladimyr: Lets make sure that we all are on same page. @agnivade pointed out that saving few keystrokes wont save a day comparing to possible confusion it could create and I fully agree with that one. In the same time @sbrl pointed out that `platform/page` is less tedious to write than `--platform=<os_name>` to which I can totally relate. @jedahan introduced short options into the story as some kind of compromise?

Folks, few things to have in mind:
1. We created discussion around long options while in reality it goes other way around.
Original UNIX had only short options and those long ones came with GNU tools as a way of expanding possible options pool. Rule of thumb is that every option should have short flag while optionally long one can be added for user ergonomics. That's not some dogmatic rule that's [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html).
OFC that experienced users (familiar with particular tool) would favor `-p <platform>` over `--platform=<platform>` while latter provides more information to newcomers. That's not my prediction, that's [GNU's message to developers](https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html).
At the end of the day who doesn't want to be both POSIX and GNU compliant and that is why most of well known tools these days support both short & long options.

2. @agnivade in his replies always used `--platform=<platform>` form.
Personally, that's worst (or at least most tedious) form of GNU style one could use; let me explain way. Aside from `--` and _long option_ itself you also use `=` which is located on _edge of typing space_ on standard keyboard.
I hope this isn't news for anyone but just to make sure: [relevant section of TAO of Unix Programming](http://www.catb.org/esr/writings/taoup/html/ch10s05.html) quoted here:

    > An option argument (if any) can be separated by either whitespace or a single “=” (equal sign) character.

    Which means you could use `--platform <platform>`.

    As meme says __I don't usually do GNU style options but when I do I use whitespace instead.__ You can probably guess my rationale, yeah it is easier to hit on keyboard. That's just my personal preference, you may or may not find it true in your case; I just wanted to point that out.

    **GNU gives you choice**:
    - `--platform=<platform>` :white_check_mark:
    - `--platform <platform>` :white_check_mark:

    Pick your poison :wink:

3. Regarding `--os` vs `--platform` and `-o` vs `-p`.

    First pair is debating between concepts. Is it really an OS or a platform? I would say neither but in my previous reply https://github.com/tldr-pages/tldr/pull/2706#discussion_r I explained why my vote goes to `--platform`.

    Second pair is debating between expectations: [standard options](http://www.catb.org/esr/writings/taoup/html/ch10s05.html). My rationale here is that `-p` as _port_ makes sense only in context of networking tools while `-o` as _output_ applies to more wider variety of usecases. So my vote here goes to `-p`.

    `--platform` + `-p` = `-p, --platform`

## Proposal

Clients **MUST** implement `-p, --platform`. That covers both, ergonomic needs of experienced users and learning needs of newcomers and it is POSIX and GNU compliant. Clients **MAY** additionally support `platform:page` shorthand in lieu of regular options to provide both experienced and new users with expressive and intuitive query syntax.

Thoughts? :speaking_head:

vladimyr: To further support my case lets examine what @mebeim brought to discussion: https://github.com/tldr-pages/tldr/pull/2706#discussion_r247359500:
>It could be a good idea to allow something like `tldr platform command` like man does with `man section command` :) Sounds pretty straightforward to me.

According to [`man man`](https://www.freebsd.org/cgi/man.cgi?query=man&apropos=0&sektion=0&manpath=FreeBSD+12.0-RELEASE+and+Ports&arch=default&format=html) itself:
```
SYNOPSIS
     man [-adho] [-t | -w] [-M manpath] [-P pager] [-S mansect]
     [-m arch[:machine]] [-p [eprtv]] [mansect] page ...
     man -f keyword ...
     man -k keyword ...
```

Optional `[mansect]` argument is just convenience shorthand for `[-S mansect]`. Take for example:
```
$ man 1 printf
$ man 3 printf
```
those are just shortcuts for:
```
$ man printf -S1
$ man printf -S3
```
_(FWIW `-S` is actually a colon delimited range so `-S1` and `-S1:2` would produce same result.)_

or if you hate hitting that <kbd>Shift</kbd> key you can leverage System V compatibility through:
```
$ man printf -s1
$ man printf -s3
```
Strangely enough this is not even listed as possible but simply implemented behind curtains.

Man sections are not really analogous to platforms but there is another interesting option here:
```
[-m arch[:machine]
```
It enables overriding `MACHINE` and `MACHINE_ARCH` environment variables _(e.g. `MACHINE=macppc MACHINE_ARCH=powerpc` for those old PowerPC powered Macs)_ in order to load **hardware platform** specific manpages. This was probably less exotic back at the time while currently it does not serve much of a purpose.
Take a closer look on syntax tho: `arch:machine` _(e.g. macppc:powerpc for my previous example)_ I knew that someone implanted those archaic ideas in my hand :angry:

>**This gives us a history lesson about following standards and extending them with and unique expressive bits and pieces.**

pepa65: In the bash-tldr client using the `tldr platform/page` syntax is standard. I am happy to support `tldr platform:page` as well, and would be willing to also implement `tldr platform page`. I might go all out and support whatever commandline options are going to be mandated -- as long as they don't conflict of course.

CLIENT_SPECIFICATION.md
vladimyr: Personally I don't see any value in listing every possible page, crossplatform (which I would assume `--list-all` does as @jedahan pointed out) because that means we should print out paths to disambiguate pages covering same tool on different platforms. That essentially puts us in business of re-implementing `ls -R`. If user really wants that they should be able to get cache location somehow and do that on they own.
So my vote goes to `-l, --list` to get list of pages for assumed/specified platform.

OTOH discussion about list formatting is (sorry but I have to say it) aiming at completely wrong target. It simply **does NOT matter** at all how it will look and there is good background reason for that. Take our beloved `ls` for example. Default output of `ls` is columned or is it?! Pipe that through `cat/less` and you'll get _one-per-line_ list. I'm too lazy to dig out proper POSIX/TAOUP/whatever spec/recommendation so rule of thumb instead:

Output of _first class citizen_ *NIX CLI tool:
1. **MAY** be aesthetically pleasing with all kind of wonders you decide to throw in
**if `STDOUT` is NOT `TTY`**
2. **MUST** be formatted as _one-per-line_ list (to enable easy manipulation using *NIX tools)
**if `STDOUT` is a `TTY`**

Can we just replace _blah blah CLI tool_ with `tldr` client and put it inside spec? :wink:

agnivade: I think listing all pages are useful to know what all commands are there (all platforms or current platforms comes after). This is a tool aimed for beginners and we should not expect them to go to the cache directory and manually check the folder.

Now regarding listing pages from all platforms or a single platform is up for debate. I think it's reasonable to want to see _all_ pages from all platforms, otherwise people will need to type all the platforms along with `--list` to get all the pages. Disambiguation will be needed, yes.

In any case, `--list-all` should mean display all pages from all platforms. `--list` should be for current platform and can be combined with `--platform` flag.

vladimyr: If you envision future where `tldr --list --platform macos` is a thing, can we just do `tldr --list --platform all`?

agnivade: I would be okay with `all`, but do you think there are other places where we can use `all` or just for this ? If it's just for this, then why not change `--list` to mean `--list-all` ? Better not to attach special meaning to any word, if we can keep things simple.

vladimyr: Well, that's just my opinion but I give more importance to listing current platform pages than to dumping everything and options should then follow that assumption. Meaning `-l,--list` does that unless you add _pro_ type of option that will turn it into dump everything. Most useful option should be easiest to type.

agnivade: Got it. I am kinda opposite to that notion. I think list (or whatever) should dump all platforms by default until I want just a specific platform.

So I propose - `-l,--list` shows all platforms by default, unless you add a `--platform` flag, upon which it just shows pages for that platform.

That is just my opinion. We will decide whatever the community wants.

agnivade: Ah that makes sense. Thanks.

jedahan: if `tldr --list` with no platform defaults to all, that is the inverse behaviour of `tldr <command>` with no platform defaulting to the detected platform.

CLIENT_SPECIFICATION.md
sbrl: Thanks for the excellent analysis, @vladimyr! Here are my thoughts:

 - I think `tldr platform page name` is slightly less clear on which bit is the platform and which the page.
 - I agree on the subject of os vs platform. We might support something like `android` as a platform in the future for example - which is technically linux as an OS, but a completely different platform with a completely different set of commands.
 - I like the `-p page`, `--platform page` proposal a lot.
 - I like `platform:page` a lot too.

I agree with @vladimyr's proposal for MUST `-p`, `--platform` and MAY `platform:page`.

CLIENT_SPECIFICATION.md
sbrl: Great points.

I think @jedahan has a point about consistency. What about `tldr --list` defaults to the current platform, `tldr --list platform` lists the pages on a specific platform, and `tldr --list all` lists all pages inn all platforms? While I like the idea of using a `*`, that could be interpreted by the shell to mean "all files in the current directory" and get substituted in, so I'd like to avoid that if possible.

I'm unsure how the `common` platform fits into all of this though.

Great point @vladimyr on if the output is a TTY. I think I'll add that.

vladimyr: @sbrl Depends on language/tooling you use but this list covers pretty much all popular options: https://rosettacode.org/wiki/Check_output_device_is_a_terminal In, a nutshell there is either `isTTY(fd)` utility function provided or a property on `stdout` itself.

pepa65: I still think if people want to do something that technical, they could either just turn to the cache directory and get anything they need there, or, they should use the `-t`/`--tty` flag or `-n`/`--noformat` or some such. Because if people pipe their output into `less`, they probably don't want to automatically get no columns or colors. If you use a flag, the user can decide, if you don't, some users are going to request it as a feature.

vladimyr: @pepa65

> If you use a flag, the user can decide, if you don't, some users are going to request it as a feature.

And that's like only thing I agree with. :disappointed:

> I still think if people want to do something that technical,

Calling list manipulation with standard cli tools technical is bit far fetched if you ask me; after all isn't one of selling points of `tldr` pages is to get people to that level more easily? :upside_down_face:
What so technical in grepping large lists? :thinking:

> or, they should use the `-t`/`--tty`

I'm very curious to see example of tool implementing given option. I never seen something like that and I highly doubt I ever will; having in mind it is not considered standard:
- http://www.catb.org/esr/writings/taoup/html/ch10s05.html
- https://www.gnu.org/prep/standards/html_node/Option-Table.html#Option-Table
    _(`tty` used by `gdb` is not switch but an option with an argument: https://manpages.ubuntu.com/manpages/precise/en/man1/gdb.1.html)_

>Because if people pipe their output into less, they probably don't want to automatically get no columns or colors.

Sorry but that's just your opinion while common practice teaches us exact opposite. I may or may not agree with you but you are actually arguing with **facts**:
- https://superuser.com/questions/36022/getting-colored-results-when-using-a-pipe-from-grep-to-less
- https://unix.stackexchange.com/questions/19317/can-less-retain-colored-output

So, yeah you are right, people like ansi colors but _like_ and _expect_ by default are two different things. By default colors are off. Which brings me to the next thing you said:

> or -n/--noformat or some such.

Again, you are going against common practice established by *NIX cli tools. Formatting is off by default if output is tty. If you really want to retain it you need to explicitly request tool generating output to do so. And it makes perfect sense why you need extra step for colors/formatting to be outputted/preserved: **not every tool you may use to process output understands and/or is able to strip out ansi escapes!** If user is positive that his tool of choice (e.g. `less`) is capable of doing so, ok, you have choice to do `--color=always`, `--color`, `--pretty`, you name it...

In conclusion - to actually address your concern we should probably support `--pretty` option that would force colors and general formatting being preserved when output is not `tty`. I'm ok with that and FWIW you have my :+1: there but in form of recommendation, such thing shouldn't be a **MUST** but **MAY** instead:
- `stdout is NOT TTY` - Generate formatted output _(mandatory)_
- `stdout is TTY and --pretty` - Preserve formatting _(optional)_
- `stdout is TTY` - Generate `\n` separated list _(mandatory)_

That being said, please try to align your proposals with established CLI tools design practices and user expectations. TBH I understand your motivation and would sometimes give everything for a chance to _rewrite_ history of our discipline but that's simply not possible. `tldr` clients are not first tools of such class and less exotic they are more successful and useful they'll become!

vladimyr: @pepa65
>Don't you feel you are being unnecessarily redundant with `tldr --list --platform macos`? I just do `tldr --list macos` and it works. There is no need there for a `--platform` marker.

Yes and no :thinking:

Sure it is verbose, I personally don't like it too but...
OTOH lets compare this with what `man` does (I like to consider `tldr` client as _`man` for humans_ same way as [`httpie`](https://httpie.org)  is _`curl`for humans_; correct me if my understanding is wrong :bowtie:).

In order to get all `man` pages you need to run:
```
$ man -k .
```
where `.` is regex meaning everything. If you want to narrow that down using sections:
```
$ man -k . -s1
```
lists only manpages inside 1st section.

If you now swap `-k .` with `-l` and `-s1` with `-p <platform>` you'll get original form I proposed:
```
$ tldr -l -p macos
$ # or if you like it more
$ tldr --list --platform macos
```

You could OFC argue that manpage listing I just described has mandatory search param while tldr listing does not thus putting us in better position by allowing what you suggested. Yeah, that's true but might be also an indication that we are doing something wrong :warning:
Maybe right way to do the listing is to copy good ol' man? :upside_down_face:

@jedahan
Could not support you more in that consistency effort :100:

@sbrl
I wanted to do some wildcard like thingie and ended up doing lousy job :man_facepalming:

I know it isn't constructive at all but this whole thing with what should _list (almost) everything_ print is painfully convoluted and more we evolve it less value I see in it. Can someone ELI5 in what particular case user wants to print out every or almost every possible command out apart from pure curiosity and explorer's spirit?
Aren't the _real life_ usecases:
(step zero: you know what you want to do)
1. You **know** which tool can help you but you don't know how to use it so you do:
```
$ tldr <tool>
```
2. You **do NOT know** which tool you should use so you do:
```
$ tldr --search <query>
```
**Which isn't even mentioned in the spec itself!** :warning:

In both cases you might be searching answers for a platform you are currently not working on so you'll add `--platform <target_platform>` too.

I mean this is exactly what `man` supports also:
```
$ man <tool>
$ man -k <query> # analogous to `--search`
```
while listing all pages is available by means of _empty_ query but is certainly isn't something that is endorsed by having dedicated option like `--list-(almost/maybe/platform/common)-all` :upside_down_face:

agnivade: Alright then, let's go with default being current platform and having a reserved platform `all`.

CLIENT_SPECIFICATION.md
```diff
@@ -0,0 +1,194 @@
...
+## Command-Line Interface
+This section describes the standardised command-line interface (CLI) for clients implementing one. Clients that do not provide a CLI can ignore this section.
+
+### Arguments
+A number of arguments MUST be supported (unless otherwise specified if a CLI is implemented:
+
+Argument		        | Meaning
+------------------------|--------------------
+`--update`				| Updates the offline cache of pages. MUST be implemented if cache is supported.
+`--version`, `-v`		| Shows the current version of the client, and the version of this specification that it implements.
```
vladimyr: I believe we need `-h, --help` here too?
BTW options are typically written in reverse order: `-v, --version` :bowtie:

jedahan:
Also note as a general philosophy I think we should strive to only cover the most important features as minimally as possible to allow clients to experiment with ux/ui while giving us a decent checklist if we should include a client in the readme/docs/wiki/whatever.

I.e. if there is any contention in how the interface acts, feel free to let different clients behave differently. For example, if people have opinions about the formatting of list that differ, allow the clients to differ in how they display lists. We can always add restrictions or suggestions to specification, but its much harder to take them away.

As a user, the only things I care about are:
* I can find a page with `tldr <command>`
* I can get the help for a given tldr client with `tldr` (or ideally `tldr tldr`)

Every other feature/idea is *extra* - certainly useful, and will make me choose one client over another. But if the specification defines the implementation for all clients, then I think its actually worse for the user in choosing a client. If they all work the same which should I use?

It would be a good smoke test of our specification - can the specification of tldr clients be sufficiently covered by `tldr tldr`?

CLIENT_SPECIFICATION.md
sbrl: Great discussion! I'll update the spec to reflect this.

Met me know if I miss anything :P

sbrl:
Currently, we have just this in `tldr tldr`:

```markdown
#tldr

- Simplified man pages.

> Get typical usages of a command (hint: this is how you got here!):
 tldr command
```

...because every client is different enough that `tldr command` is the only commonality between them all :P

CLIENT_SPECIFICATION.md
```diff
@@ -0,0 +1,206 @@
...
+```
+
+#### Specifying the Platform
+As pages are grouped by platform, a user may want to access a platform-specific version of a page. This MUST be supported by prefixing the page name as follows:
+
+```
+windows/type
+linux/sh
+osx/brew
+```
```
vladimyr: Isn't this supposed to get removed from the spec? 😕

CLIENT_SPECIFICATION.md
```diff
@@ -0,0 +1,206 @@
...
+# In the below, "--foo" is a custom argument that takes a parameter.
+tldr --foo bar bash
+```
+
+Furthermore, clients may OPTIONALLY support a colon to specify the platform name in addition to the `--platform`, `-p` syntax described above Here are some examples:
+
+```bash
+tldr platform:page
+tldr windows:type
+tldr common:git checkout
```
vladimyr: This is **really really** confusing. People used to cli conventions are likely read it like two arguments instead. If _prefix_ syntax is used kebab case should be mandated:
```
$ tldr common:git-checkout
```

mebeim:
I would recommend only making the short command line options required, with *optional* support to the long options. That is, for example: require `-l`, and *optionally* allow `--list`. I also only see `--update`, but not `-u`, which IMHO would make more sense to have by default

The reason I say this is because parsing long options is something that is not well standardized in low-level (mostly compiled) languages, and would force whoever wants to write a client to go through the seemingly unneeded hassle of creating a cross-platform long option parser or detect and use an existing one.

---

Stated the above, I would suggest the following changes.

First of all, rephrase the "Arguments" section that currently says:

> A number of arguments MUST be supported [...]
> *(table of options)*

to the following:

> A set of options needs to be supported according to the following table:
> *(table of options)*

Secondly, rearrange the table of options like this:

Option | Meaning | Notes
-|-|-
`-u` | Update the offline cache of pages. | MUST be supported if the program implements caching.
`-v` | Show the current version of the client, and the version of this specification that is implemented.
`-l [platform>` | Lists all the pages in the specified platform to standard output. If the special platform `all` is specified a list of all pages in all platforms MUST be displayed.
`--pretty` | OPTIONAL. Forces the output to contain all additional decorations, even if the standard output is not a TTY.
`--platform`, `-p` | Specifies the platform that should be used for resolving page names. If specified, the named platform MUST be checked first instead of the host platform  as described below.

mebeim:
Also, posting this as a separate comment since the previous is pretty long: I think that having two options that take a `platform` as a parameter is overkill. I think only having the `-p` option to take a platform as parameter and then use it as a modifier would be way cooler. It would become:

Command | Behavior
-|-
`tldr -l` | List all pages for the default (current) platform.
`tldr -l -p linux` | List all pages for the `linux` platform.
`tldr -lp all` | List all pages for all the platforms.
`tldr -p linux command` | Search a command for the `linux` platform.

vladimyr:
@mebeim It plays nicely with ability of joining short options as you showcased 😎 and I fully support you regarding this one. However this gets us back to https://github.com/tldr-pages/tldr/pull/2706#discussion_r248397662 which already got blessed despite me being using similar arguments against it https://github.com/tldr-pages/tldr/pull/2706#discussion_r248530657?

mebeim:
@vladimyr I guess is a matter of personal taste at this point. Saying that `tldr --list --platform xxx` is "unnecessarily redundant" does not make much sense to me: of course it is, you're using two long options when you could instead do `tldr -lp xxx`. This is actually also why I am in favour of making the `-l,--list` option *optional*. It does not make much sense to me to require it by default, because as we saw in this PR's comments it's basically a purely subjective matter what it should or should not do. 

Also, seeing arguments about making `-l` list *all* the platforms by default: you also need to consider the fact that duplicated pages exist (e.g. `file` is both under `linux` and `osx`). Should then a command be listed more times or only one? The latter would require much more work. This has to be taken into consideration (and supports my point of making `-l` optional imho).

vladimyr:
> I guess is a matter of personal taste at this point. Saying that `tldr --list --platform xxx` is "unnecessarily redundant" does not make much sense to me: of course it is, you're using two long options when you could instead do `tldr -lp xxx`.

Congratulations! You nailed it! :clap: 
Personal taste is great criteria if you are making an order in a restaurant. Writing spec is not picking stuff from some giant menu. Established (historical) practices and standards aren't optional sauces you can pick or discard because they don't match your personal taste. Users, both client writers and actual end users, have their needs and expectations and they need to be expressed, perceived, validated and appropriately addressed. They base them on their experience gained using other cli tools so this shouldn't be test bed for any sort of extravagance.
What that means is in order to be constructive everyone needs to hop into user's shoes. If it is _unnecessarily redundant_ to me, or if _I don't like GNU style options_ means literally nothing because that's not voice of diverse user base you should try to represent!

> Also, seeing arguments about making `-l` list _all_ the platforms by default: you also need to consider the fact that duplicated pages exist (e.g. `file` is both under `linux` and `osx`). Should then a command be listed more times or only one? The latter would require much more work. This has to be taken into consideration (and supports my point of making `-l` optional imho).

This is just names listing so squashing duplicates doesn't really require much more work but some sort of supported platforms indication should exist in that case:
1. You can either print it once following by list of supported platforms (e.g. `<tool> (<platform_1>, <platform_2> ...)`); or
2. Print it multiple times in path-like style `<platform>/<name>`

which I already noted here https://github.com/tldr-pages/tldr/pull/2706#discussion_r248132144:
> Personally I don't see any value in listing every possible page, crossplatform (which I would assume `--list-all` does as @jedahan pointed out) because **that means we should print out paths to disambiguate pages covering same tool on different platforms. That essentially puts us in business of re-implementing `ls -R`. If user really wants that they should be able to get cache location somehow and do that on they own.**
So my vote goes to `-l, --list` to get list of pages for assumed/specified platform.

I questioned it once more as part of https://github.com/tldr-pages/tldr/pull/2706#discussion_r248530657:
> I know it isn't constructive at all but this whole thing with what should list (almost) everything print is painfully convoluted and more we evolve it less value I see in it. **Can someone ELI5 in what particular case user wants to print every or almost every possible command out apart from pure curiosity and explorer's spirit?**

I never got that explanation unless I take that _personal taste_ criteria as possible one.

> I think creating a client spec that is simple and reasonable leaving freedom to the developers is way more important than giving a pedantic and difficult to follow spec just to be more fancy and precise. 

Well said!

> If it was for me, the spec would have just been:
> 
> > You need to implement `-u` to update, `-p` to choose a different platform. Other functionality should be well described by the help function of the client and the readme of the repo (if present).

Agree with your point but phrasing is bit off, avoid extremes. In reality that is not enough to put spec label on it but _less is more_ is in fact a good advice to follow.
Also `-u` for update mandates offline cache which is something that IMHO shouldn't be requested these days. Times are changing, everybody is using SO these days and I somehow doubt that they have local copy of whole SO content somewhere :trollface: 

> What's the point of having dozens of different clients if they are all written the same way? The fact that they are different is one of the most important things, because one can pick whichever they prefer the most.

Again, fully agree. Diversity is good because it fosters innovation!

vladimyr:
> The reason I say this is because parsing long options is something that is not well standardized in low-level (mostly compiled) languages, and would force whoever wants to write a client to go through the seemingly unneeded hassle of creating a cross-platform long option parser or detect and use an existing one.

This is an interesting statement :open_mouth: I don't regularly use system programming languages so I can't verify it but I find it hard to believe that there isn't cross platform c library for parsing both short and long style options? (which by means of FFI eliminates this issue elsewhere too) I know that POSIX defines only `getopt` for BSD style options but still... :thinking: 

If that is really true than I fully support your proposal to make GNU style options optional.

CLIENT_SPECIFICATION.md
sbrl: Oops! Thanks - I must have missed that :P

CLIENT_SPECIFICATION.md
sbrl: I see. Good idea! I'll change that now.

sbrl:
Great comments! Could you propose a PR against this PR with those changes please, @vladimyr? I'm getting somewhat confused keeping track of everything here - as evidenced by the bunch of stuff I missed from the last round of commenting :P

mebeim:
> This is just names listing so squashing duplicates doesn't really require much more work but some sort of supported platforms indication should exist in that case

Oh, right, agreed! Totally missed that.

> Times are changing, everybody is using SO these days and I somehow doubt that they have local copy of whole SO content somewhere :trollface:

You still have to consider that, at least the last time I checked, GitHub does not advise constantly requesting stuff from their site as if it was a static content server. See [section 5](https://help.github.com/articles/github-terms-of-service/#5-scraping) and [section 7](https://help.github.com/articles/github-terms-of-service/#7-excessive-bandwidth-use) of the Terms of Service. I also wouldn't really trust a client that does not do caching and requests pages every single time, that's just lazy stuff, at least cache single pages when I request them... but this is obviously only a personal opinion.

> I find it hard to believe that there isn't cross platform c library for parsing both short and long style options.

There *might* be one, but consider the following points taking the C language as an example:

- You find that library: now what do you do?

    - Link it statically and include it in your project? Well, you have to take a look at the license of the library and be sure to follow its guidelines and maybe also need to change your own license if that license does not allow it. You want to distribute on Linux and that library is provided as a package? Well then you should dynamically link it, statically linking stuff is considered bad!

    - So no static, how do you detect if the target system has that package? Really boring stuff.

- T

Take C for example: using `getopt` already makes your source not `c89` or `c99` compliant anymore, but `gnu99`, which is why people often write their own `getopt` if they want to port their application somewhere else where the GNU standard is not taken for granted.

vladimyr:
> You still have to consider that, at least the last time I checked, GitHub does not advise constantly requesting stuff from their site as if it was a static content server. See [section C5](https://help.github.com/articles/github-terms-of-service/#5-scraping) and [section C7](https://help.github.com/articles/github-terms-of-service/#7-excessive-bandwidth-use) of the Terms of Service.

You kinda mixed apples and oranges here. You can consume tldr pages directly from github either by using api call or through raw links (e.g. https://github.com/tldr-pages/tldr/raw/master/pages/common/env.md). Neither is considered scraping so section C5 has nothing to do with it.
OTOH section C7 handles excessive bandwidth usage but who will produce it? End users of tldr clients? That is highly unlikely and if they really want to get banned that's their problem 😉 

> I also wouldn't really trust a client that does not do caching and requests pages every single time, that's just lazy stuff, at least cache single pages when I request them... but this is obviously only a personal opinion.

I agree with you that caching single pages is a _must_ and second your laziness statement. That gives us two options:
1. clients who eagerly cache whole tldr _db_ (providing offline support)
2. clients who cache page requests on-fly (require internet connection)

Personally I don't see any of these two approaches as wrong/right one. For me it is also a matter of preference and most important I firmly believe official spec shouldn't exclude and/or disallow  neither one!

vladimyr:
> Great comments! Could you propose a PR against this PR with those changes please, @vladimyr? I'm getting somewhat confused keeping track of everything here - as evidenced by the bunch of stuff I missed from the last round of commenting :P

@sbrl It is not you. This review thread is awful medium for discussing these kind of things. Ideally this would be some sort of forum thread but that's so yesterday 😛

vladimyr:
@mebeim So you are telling me that there isn't header only option parsing library?
👉 https://attractivechaos.github.io/klib/#Ketopt%3A%20parsing%20command-line%20arguments

I'm just trolling 😁 I don't know how good this thing is and is it good at all...
Jokes aside that was educative breakdown of struggles C developer goes through... 👍

mebeim:
@vladimyr Oh, right, I did not know raw links were treated differently. So that point I made is actually not a concern at all, awesome. Sorry for mixing stuff up 😅

> Personally I don't see any of these two approaches as wrong/right one.

Totally agree, caching or not people should be free to do what they want :slightly_smiling_face:

sbrl:
@mebeim: Good idea! I'll rename it now.

@vladimyr: There's always #1065 if you'd like even _moar_ discussion about it.....
There has to be a better way of drafting specifications.

sbrl:
Hey, go away @stale! This PR took way too much effort for you to just close. 

What are the outstanding issues on this?

@vladimyr @mebeim

mebeim:
@sbrl IMHO the biggest issue is the table of required options, which should be updated to better clarify what is required and what is not, like I said in my comment above. Right now it is just confusing. Secondly it should be decided whether the `-p, --platform` option should be used as a modifier (like I proposed in my second comment) or if both `-p` and `-l` should get the same platform parameter.

sbrl:
Right.

How about an extra column to the argument table with a heading like "required?"?

Hrm. I do like the idea of `--platform` being a modifier that affects the page listing too. We'd have to ensure we clearly define the default behaviour though. What happens if it's not specified?

mebeim:
@sbrl for the table, yeah, that's exactly what I suggested in my comment. For the `-p` option, well, the behavior if not specified is the usual: search for the current OS and fallback to common.

sbrl:
Right. I've added that column. How would you suggest we re-word the `--platform ` argument?

mebeim:
@sbrl hmm, what about:

> Specifies the platform to be used to perform the action (either listing or searching). If this option is specified, the selected platform MUST be checked first instead of the current platform as described below.

Right now I can't think of a nicer way, but I'm sure there is. Maybe someone else could improve it.

CLIENT-SPECIFICATION.md
```diff
@@ -28,12 +28,12 @@ This section describes the standardised command-line interface (CLI) for clients
...
-`-p`, `--platform`   | Specifies the platform that should be used for resolving page names. If specified, the named platform MUST be checked first instead of the host platform as described below.
-`-u`, `--update`     | Updates the offline cache of pages. MUST be implemented if cache is supported.
+Argument			| Required?	| Meaning
+--------------------|--------------------
+`-v`, `--version`	| Yes		| Shows the current version of the client, and the version of this specification that it implements.
+`-l`, `--list`		| No		| Lists all the pages in the current platform to the standard output. If the special platform `all` is specified a list of all pages in all platforms MUST be displayed.
+`-p`, `--platform`	| Yes		| Specifies the platform that should be used for resolving page names. If specified, the named platform MUST be checked first instead of the host platform as described below.
+`-u`, `--update`	| Only if cache is implemented	| Updates the offline cache of pages. MUST be implemented if cache is supported.
```
jedahan: `Conditional` for Required? is a bit shorter

CLIENT-SPECIFICATION.md
```diff
@@ -0,0 +1,174 @@
...
+
+
+## Command-Line Interface
+This section describes the standardised command-line interface (CLI) for clients implementing one. Clients that do not provide a CLI can ignore this section.
+
+### Arguments
+A number of arguments MUST be supported (unless otherwise specified) if a CLI is implemented:
+
+Argument			| Required?	| Meaning
+--------------------|--------------------
```
mebeim: Table is broken:

```suggestion
--------------------|--------------------|--------------------
```

Also, the columns are not aligned right now.

sbrl:
Thanks for the suggestions! I've applied them.

mebeim:
I think this looks good enough to be introduced as first ever version of the client specification. Can't really keep track of _all_ that has been said above, but looks fine, am I right? Major points and issues seem to have been addresses by @sbrl, so I'd say let's just remove the draft notes in the document and change `pages/common/tldr.md` to only show the options that MUST be supported and then it's ready to merge. 

What do you guys think?

agnivade:
Sounds good. This has been sitting for too long. Let's give it one more day to see if anyone has any more comments, then it can be merged.

CLIENT-SPECIFICATION.md
```diff
@@ -0,0 +1,174 @@
...
+
+ - Chinese (Taiwan): `pages.zh-TW`.
+ - Portuguese (Brazil): `pages.pt-BR`.
+ - German (Germany): `pages.de-DE`.
+
+You can check the validity of BCP 47 tags [here](http://schneegans.de/lv/).
+
+The structure inside these translation folders is identical to that of the main `pages` folder.
+
+**Note to self: A portion of this section is from [PR #2703](https://github.com/tldr-pages/tldr/pull/2703).**
```
mebeim:
```suggestion
```

pages/common/tldr.md
```diff
@@ -5,3 +5,15 @@
 - Get typical usages of a command (hint: this is how you got here!):
 
 `tldr {{command}}`
+
+- List available pages:
+
+`tldr --list`
```
mebeim:
```suggestion
```
`-l|--list` is optional.

pages/common/tldr.md
```diff
@@ -5,3 +5,15 @@
...
+
+- Show the tar tldr page for linux:
+
+`tldr --platform {{linux}} {{tar}}`
```
mebeim:
```suggestion
`tldr -p {{linux}} {{tar}}`
```
`--platform` is optional, `-p` is not.

sbrl:
Thanks everyone! I would have taken a look at this sooner, but I've been busy with work from University.

Merging now, since I've handled @mebeim's suggestions :smiley_cat:

sbrl:
Oh yeah, we should add a link to it to the README too, and possibly look at which clients adhere tot he spec.

PR #2708, opened 2019-01-14
youtube-dl: fix custom filename example

vastpeng:
image


waldyrious:
@vastpeng can you explain why the previous version failed? Looking at the diff it's not entirely clear to me. By the way, thanks for the contribution :)

agnivade:
`youtube-dl` is a python tool. So it needs a format character in the string formatting. Here - `%(title) by`, there is none, so it incorrectly detects `b` as the format character and throws an error.

PR #2709, opened 2019-01-14
timeout: clarify the default signal sent to the process

Calinou:
The second example mentioned SIGTERM as the default signal, but the first example was worded in a way that may have implied SIGKILL is sent by default.

### DIFF
--- a/pages/common/timeout.md
+++ b/pages/common/timeout.md
@@ -2,7 +2,7 @@
 
 > Run a command with a time limit.
 
-- Run `sleep 10` and kill it, if it's running after 3 seconds:
+- Run `sleep 10` and terminate it, if it's running for more than 3 seconds:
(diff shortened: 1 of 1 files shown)


pages/common/timeout.md
```diff
@@ -2,7 +2,7 @@
 
 > Run a command with a time limit.
 
-- Run `sleep 10` and kill it, if it's running after 3 seconds:
+- Run `sleep 10` and terminate it, if it's running for more than 3 seconds:
```
sbrl: Perhaps this will read better?

```suggestion
- Run `sleep 10` and terminate it, if it runs for more than 3 seconds:
```

Calinou: Done :slightly_smiling_face:

PR #2710, opened 2019-01-14
split: remove split into n files, add split by pattern

jedahan:
osx does not have a `-n` option, but does have a `-p` option

### DIFF
--- a/pages/osx/split.md
+++ b/pages/osx/split.md
@@ -6,9 +6,9 @@
 
 `split -l {{10}} {{filename}}`
 
-- Split a file into 5 files. File is split such that each split has same size (except the last split):
+- Split a file by a pattern (the pattern is the head of each split):
 
-`split -n {{5}} {{filename}}`
+`split -p {{pattern}} {{filename}}`
 
 - Split a file with 512 bytes in each split (except the last split; use 512k for kilobytes and 512m for megabytes):
 
--- a/pages/osx/split.md
+++ b/pages/osx/split.md
@@ -6,9 +6,9 @@
 
 `split -l {{10}} {{filename}}`
 
-- Split a file by a pattern (the pattern is the head of each split):
+- Split a file by a regular expression. The matching line will be the first line of the next output file:
 
-`split -p {{pattern}} {{filename}}`
+`split -p {{regular expression}} {{filename}}`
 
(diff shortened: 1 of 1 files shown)


pages/osx/split.md
```diff
@@ -6,9 +6,9 @@
 
 `split -l {{10}} {{filename}}`
 
-- Split a file into 5 files. File is split such that each split has same size (except the last split):
+- Split a file by a pattern (the pattern is the head of each split):
```
sbrl: Is this pattern a regular expression?

Also, any reason for removing this example and adding a new one? We haven't reached the limit for this page yet.

jedahan: at least on my macos, `-n` is not an option for split.

pattern is interpreted as a regular expression, though the manual does not specify which kind

sbrl: Ah, I see! This is the macos-specific version of the page. Got it!

pages/osx/split.md
```diff
@@ -6,9 +6,9 @@
 
 `split -l {{10}} {{filename}}`
 
-- Split a file into 5 files. File is split such that each split has same size (except the last split):
+- Split a file by a regular expression. The matching line will be the first line of the next output file:
 
-`split -n {{5}} {{filename}}`
+`split -p {{regular expression}} {{filename}}`
```
agnivade: A concrete example of a regex would be helpful here.

jedahan: should `{{filename}}` be like `{{animals.txt}}`?

pages/osx/split.md
```diff
@@ -8,7 +8,7 @@
 
 - Split a file by a regular expression. The matching line will be the first line of the next output file:
 
-`split -p {{regular expression}} {{filename}}`
+`split -p {{cat|[d|h]og}} {{filename}}`
```
jedahan: @agnivade should `{{filename}}` be like `{{animals.txt}}`?

sbrl: Should `[d|h]` be `(d|h)`?

jedahan: good catch!

jedahan: so my test file was:
```
bat
cat
gnat
dog
hog
bog
dhog
```
and this worked for splitting only on cat, dog, and hog (and not splitting on dhog)

PR #2711, opened 2019-01-15
now: add page

musewick:

### DIFF
--- a/pages/common/now.md
+++ b/pages/common/now.md
@@ -2,7 +2,7 @@
 
 > Cloud platform for serverless deployment.
 
-- Deploy current directory:
+- Deploy the current directory:
 
 `now`
 
--- a/pages/common/now.md
+++ b/pages/common/now.md
@@ -1,6 +1,6 @@
 # now
 
-> Cloud platform for serverless deployment
+> Cloud platform for serverless deployment.
 
 - Deploy current directory:
 
--- a/pages/common/now.md
+++ b/pages/common/now.md
@@ -6,7 +6,7 @@
 
 `now`
 
-- List of deployments:
+- Display a list of deployments:
 
 `now list`
 
@@ -18,10 +18,10 @@
 
 `now remove {{deployment_id}}`
 
-- Log in into account or create a new one:
+- Log in into an account or create a new one:
 
 `now login`
 
-- Initialize an example project:
+- Initialize an example project (a new directory will be created):
 
 `now init`
--- a/pages/common/now.md
+++ b/pages/common/now.md
@@ -1,0 +1,27 @@
+# now
+
+> Cloud platform for serverless deployment
+
+- Deploy current directory:
+
+`now`
+
+- List of deployments:
+
+`now list`
+
+- Display information related to a deployment:
+
+`now inspect {{deployment_url}}`
+
+- Remove a deployment:
+
+`now remove {{deployment_id}}`
+
+- Log in into account or create a new one:
+
+`now login`
+
+- Initialize an example project:
+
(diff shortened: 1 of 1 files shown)


pages/common/now.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Display information related to a deployment:
+
+`now inspect {{deployment_url}}`
+
+- Remove a deployment:
+
+`now remove {{deployment_id}}`
+
+- Log in into account or create a new one:
```
sbrl: Perhaps this will read better?

```suggestion
- Log in into an account or create a new one:
```

Also, I assume this is the first thing a user will probably want to do. Could we move this to the top of the page please? tldr-pages usually present commands in the order that they are likely to be used.

pages/common/now.md
```diff
@@ -0,0 +1,27 @@
...
+
+`now login`
+
+- Initialize an example project:
```
sbrl: Same thing here. Could this be the 2nd example please? Also, perhaps this would be clearer?

```suggestion
- Initialize an example project in the current directory:
```

pages/common/now.md
```diff
@@ -0,0 +1,27 @@
+# now
+
+> Cloud platform for serverless deployment.
+
+- Deploy current directory:
+
+`now`
+
+- List of deployments:
```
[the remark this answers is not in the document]
musewick: Last deployments that done to date, per app. Different folder = different app.

pages/common/now.md
musewick: I thought write about the login command as a hint to way to change the account, to a new one or to existing.

If not authorized user tries to deploy by command now it will start the login process first.
image

pages/common/now.md
```diff
@@ -0,0 +1,27 @@
+# now
+
+> Cloud platform for serverless deployment.
+
+- Deploy current directory:
```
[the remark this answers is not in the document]
musewick: You are right

pages/common/now.md
musewick: Actually this not like `npm init` command, I don't know why it is named init (`now example` would be better), but it is not necessary command. But interesting command for reveal now's opportunities, that's why I propose it at the end.

pages/common/now.md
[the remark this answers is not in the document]
pxgamer: Perhaps:
```suggestion
List recent deployments for the current project:
```

pages/common/now.md
pxgamer: If you push the new commit, the review comments should still apply (or will show outdated if you change content marked by a review).

pages/common/now.md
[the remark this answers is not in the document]
musewick: image
Each line is the app, they are located in different directories. In column `url` listed the last url of deployment of the app. I checked --help and there formulation is:
image
I would like to write a "list of apps", but to maintain consistency I did not.

pages/common/now.md
sbrl: I see. If it prompts for a login automatically, then is this command commonly used?

pages/common/now.md
sbrl: Sounds good to me!

pages/common/now.md
musewick: You mean is `now` commonly used command? Yes, after enter `now` occur fully automated deployment with auto-copy generated url to the clipboard and it is the main service thing.

sbrl: Ah, I mean the `now login` command.

PR #2726, opened 2019-01-23
Client spec

jedahan:
The main differences between this and #2706 are:

* ~~addition of `-t, --translation` for language-tag specification~~
* removal of `{{platform}}:{{command}}` syntax
* removal of `--pretty`
* allow clients to not implement the long form of commands
* filling out tldr.md
* switching from tabs to spaces for tables


CLIENT-SPECIFICATION.md
```diff
@@ -31,15 +33,17 @@ Pages are grouped by platform. A platform is an operating system. For example, `
...
-`--pretty`				| OPTIONAL. Forces the output to contain all additional decorations, even if the standard output is not a TTY.
-`--platform`, `-p`		| Specifies the platform that should be used for resolving page names. If specified, the named platform MUST be checked first instead of the host platform  as described below.
+Argument		         | Meaning
+---------------------|--------------------
+`-v`, `--version`    | Shows the current version of the client, and the version of this specification that it implements.
+`-l`, `--list`       | Lists all the pages in the current platform to the standard output. If the special platform `all` is specified a list of all pages in all platforms MUST be displayed.
+`-p`, `--platform`   | Specifies the platform that should be used for resolving page names. If specified, the named platform MUST be checked first instead of the host platform as described below.
+`-t, `--translation` | Specifies the translation that should be used for resolving page names. If specified, the named language-tag MUST be checked first instead of the host language as described below.
```
pxgamer:
```suggestion
`-t`, `--translation` | Specifies the translation that should be used for resolving page names. If specified, the named language-tag MUST be checked first instead of the host language as described below.
```

vladimyr:
> It would make sense to have the user choose the language only once through some kind of global configuration or detect it automatically from the environment and then cache only that specific language along with the default `/pages` as a fallback.

Inside current version of spec there is whole section dedicated to i18n environment variables and detecting language from environment. Having in mind your previous concerns about option parsing portability I would love to hear your opinions about making this work cross platform without killing yourself by writing l10n framework... To make things more challenging lets throw WSL into the mix... 😉

vladimyr:
Personally I would avoid even mentioning `LC_*` stuff because that is nice to have fallback (configuration should be goto way of detecting language) and inherently kills portability unless you decide to go great lengths in ordert to support Windows too.

mebeim:
@vladimyr looking for `LC_*` does no harm IMHO, whether it is Windows or not. On Win they would just not be defined so no problem with portability etc.

If I did not made it clear enough, what I was talking about was something like this:

 1. Check global configuration: is language set? If so, use that.
 2. If not set, then check for environment vars: is there any of them? If so, update global configuration setting that as language.
 3. No environment vars? Update global config setting default to `/pages/`.
 4. Regardless of which path is taken between the above three, use the language which is now configured and fallback to `/pages/` if the page is not found.

mebeim:
By the way, I agree with @jedahan that having first version of the spec 1.0 would be better than a random X.Y. I don't get why you guys are updating the numbers in the first place if everything is still to be made public for the first time.

agnivade:
Is this based off https://github.com/tldr-pages/tldr/pull/2706 ? Which one are we supposed to review ?

jedahan:
I misread a comment from #2706 inviting people to create proposed changes, which is why I made this.

sbrl:
Yeah, this is a PR that gets merged into the main client spec PR. I asked @jedahan for this because I was finding it difficult to keep track of some of the finer changes that needed making.

> Why are you guys even updating the version number?

That's an artifact left over from before this was a PR :P

pages/common/tldr.md
```diff
@@ -5,3 +5,19 @@
...
+
+- List available pages:
+
+`tldr --list`
+
+- Show the tar tldr page for linux:
+
+`tldr --platform {{linux}} {{tar}}`
+
+- Get help for a command in Chinese:
```
sbrl: Could we remove this example please? `-t` isn't in the spec.

mebeim: You probably meant the following example, not this one.

sbrl:
Back to #2706!

PR #2728, opened 2019-01-23
dmesg updates

perfecto25:
- [ ] The page is in the correct platform folder (`common/`, `linux/`, etc.)

### DIFF
--- a/pages/linux/dmesg.md
+++ b/pages/linux/dmesg.md
@@ -33,4 +33,3 @@
 - Colorize output (available in kernels 3.5.0 and newer):
 
 `dmesg -L`
-
--- a/pages/linux/dmesg.md
+++ b/pages/linux/dmesg.md
@@ -21,3 +21,16 @@
 - Show kernel messages 1 page at a time:
 
 `dmesg | less`
+
+- Show kernel messages with a timestamp (available in kernels 3.5.0 and newer):
+
+`dmesg -T`
+
+- Show kernel messages in human-readable form (available in kernels 3.5.0 and newer):
+
+`dmesg -H`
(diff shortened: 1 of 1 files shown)


schneiderl:
Should that tab exist on the start of the example lines?

pages/linux/dmesg.md
```diff
@@ -4,20 +4,32 @@
...
 
 - Show how much physical memory is available on this system:
 
-`dmesg | grep -i memory`
+    `dmesg | grep -i memory`
 
 - Show kernel messages 1 page at a time:
 
-`dmesg | less`
+    `dmesg | less`
```
sbrl: Could you remove the spaces at the beginning of every line please?

pages/linux/dmesg.md
```diff
@@ -4,20 +4,32 @@
...
+
+- Show kernel messages with a timestamp (available in kernels 3.5.0 and newer):
```
sbrl: I'm unsure that we need to include `(available in kernels 3.5.0 and newer)` here. Are there any current Linux distributions that use a kernel older than this in their latest release?

perfecto25: kernels like  2.6.32-642.4.2.el6.x86_64  dont have these options, centos 6 for example which is still fairly common

sbrl: ....I see. Checking the Centos website reveals that there's still a year or two of official support left there, so this should be ok then :-)

It just surprised me - I haven't run into a 2.6.x kernel for _ages_! (The last one was the OVH VPS kernel - ewwwwwww........)

PR #2729, opened 2019-01-24
at, batch: consistency between examples.

mebeim:

---

Improve consistency between the examples of `at` and `batch`, which both have a similar function and identical usage.

Not quite sure how to word the commit message, let me know if you guys think it may be worded better.


mebeim:
@agnivade you're right, that sounds simpler, thanks. Edited.

PR #2734, opened 2019-01-28
xfce4-screenshooter: add page.

mebeim:

---

I decided to add 7 examples because I personally use each one of them regularly to take screenshots (mostly combined with `--delay`) and I feel like they are quite essential to know.

I use these commands as keyboard shortcuts, and I added the first one (option-less) just because it's the simplest, but I actually think it is very rarely used from command line. Would it be better to remove it?

Options not covered are:

- `--mouse`: to show the mouse pointer. You usually don't want your mouse in a screenshot.
- `--display`: to change display. You usually take a screenshot of the current display.
- `--upload`: to launch a GUI for uploading to ZimageZ (requires user/passw).
- `--imgur`: same for Imgur but no user/passw required.

### DIFF
--- a/pages/linux/xfce4-screenshooter.md
+++ b/pages/linux/xfce4-screenshooter.md
@@ -1,0 +1,31 @@
+# xfce4-screenshooter
+
+> The XFCE4 screenshot tool.
+
+- Launch the screenshooter GUI:
+
+`xfce4-screenshooter`
+
+- Take a screenshot of the entire screen and launch the GUI to ask how to proceed:
+
+`xfce4-screenshooter --fullscreen`
+
+- Take a screenshot of the entire screen and save it in the specified directory:
+
+`xfce4-screenshooter --fullscreen --save {{path/to/directory}}`
+
+- Wait some time before taking the screenshot:
+
+`xfce4-screenshooter --delay {{seconds}}`
+
+- Take a screenshot of a region of the screen (select using the mouse):
+
+`xfce4-screenshooter --region`
+
+- Take a screenshot of the active window, and copy it to the clipboard:
+
+`xfce4-screenshooter --window --clipboard`
+
+- Take a screenshot of the active window, and open it with a choosen program:
+
+`xfce4-screenshooter --window --open {{gimp}}`


pages/linux/xfce4-screenshooter.md
```diff
@@ -0,0 +1,31 @@
+# xfce4-screenshooter
+
+> The XFCE4 screenshot tool.
+
+- Launch the screenshooter GUI:
+
+`xfce4-screenshooter`
+
+- Take a screenshot of the entire screen and launch the GUI to ask how to proceed:
```
mfrw: How about?
```suggestion
- Take a screenshot of the entire screen and save it interactively:
```

pages/linux/xfce4-screenshooter.md
```diff
@@ -0,0 +1,31 @@
...
+
+`xfce4-screenshooter --fullscreen`
+
+- Take a screenshot of the entire screen and save it in the specified directory:
+
+`xfce4-screenshooter --fullscreen --save {{path/to/directory}}`
+
+- Wait some time before taking the screenshot:
+
+`xfce4-screenshooter --delay {{seconds}}`
```
mfrw:
```suggestion
`xfce4-screenshooter --delay {{s}}`
```

pages/linux/xfce4-screenshooter.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Take a screenshot of the entire screen and launch the GUI to ask how to proceed:
+
+`xfce4-screenshooter --fullscreen`
+
+- Take a screenshot of the entire screen and save it in the specified directory:
+
+`xfce4-screenshooter --fullscreen --save {{path/to/directory}}`
+
+- Wait some time before taking the screenshot:
```
mfrw:
```suggestion
- Wait for "s" seconds before taking the screenshot:
```

pages/linux/xfce4-screenshooter.md
mebeim: Hm, I wouldn't say "save it". The GUI offers various options apart from saving: copy to clipboard, discard it, upload to the web, etc.

pages/linux/xfce4-screenshooter.md
mebeim: I've never seen `{{s}}`. I always saw `{{seconds}}`. Which I think is also what is suggested in the guidelines

> Try to incorporate the spelled-out version of single-letter options in the example's description. The goal is to allow people to understand the syntax of the commands, not just memorize it.

I think it applies to option values too.

pages/linux/xfce4-screenshooter.md
mebeim: Again, never saw this in any page and I feel like it is clear enough how I wrote it.

pages/linux/xfce4-screenshooter.md
sbrl: I guess if you wanted to use that turn of phrase, you'd be after something like `process it` instead of save it. Personally, I think it's probably ok as-is (though perhaps `a GUI` would read slightly better?)

pages/linux/xfce4-screenshooter.md
sbrl: Yeah, @mebeim - I think we always prefer verbosity here, @mfrw. To this end, I would value `seconds` over `s`.

pages/linux/xfce4-screenshooter.md
sbrl: I'm unsure that this needs changing, @mfrw. If you wanted to change it, you could do something like this, for example:

```suggestion
- Wait a specified amount of time before taking the screenshot:
```

PR #2740, opened 2019-01-30
multiple pages: add even more homepages

sbrl:
Took the time to add homepages to another batch of pages. I've started to streamline the process with a few simple bash commands, which I'll put here for future reference:

```bash
git checkout -b NEW_BRANCH_NAME # VERY important! Don't want to push to master.........
read -p "Page name: " pagename && gedit $(find -name "${pagename}.md") && git commit -am "${pagename}: add homepage"
git push
```

### DIFF
--- a/pages/common/nmap.md
+++ b/pages/common/nmap.md
@@ -2,6 +2,7 @@
 
 > Network exploration tool and security / port scanner.
 > Some features only activate when Nmap is run with privileges.
+> Homepage: <https://nmap.org>
 
 - Try to determine whether the specified hosts are up and what are their names:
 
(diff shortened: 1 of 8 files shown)


mebeim:
Is there a rule in the linter to ensure these are correctly formatted?

    > Homepage: <xxx>

And also only max one per description and at the last line of description?

If not, I think this would be important to add to ensure consistency and client support.

agnivade:
There is no rule like that. The linter is currently written lex/yacc which makes it difficult to modify. You are welcome to submit a PR.

There is no need for extra client support, the clients are just mean to display the heading. If they want to do additional parsing, that is fine. But that is not a requirement.

mebeim:
Oh I see, I guess enforcing such a rule wouldn't really be needed then, it's just part of the description anyway.

PR #2748, opened 2019-02-03
aria2: add page

mfrw:

### DIFF
--- a/pages/common/aria2.md
+++ b/pages/common/aria2.md
@@ -12,7 +12,7 @@
 
 `aria2c {{http://mirror1.org/myLinux.iso}} {{http://mirror2.org/myLinux.iso}}`
 
-- Download using "2" connections per host:
+- Download using 2 connections per host:
 
 `aria2c -x{{2}} {{http://example.org/myLinux.iso}}`
 
--- a/pages/common/aria2.md
+++ b/pages/common/aria2.md
@@ -1,0 +1,33 @@
+# aria2
+
+> A lightweight multi-protocol & multi-source command-line download utility.
+> Supports HTTP, HTTPS, FTP, SFTP, BitTorrent and Metalink.
+> Homepage: <https://aria2.github.io/>.
+
+- Download a web resource:
+
+`aria2c {{http://example.org/myLinux.iso}}`
+
+- Download a resource from multiple sources:
+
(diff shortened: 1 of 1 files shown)


mebeim:
Oh wait a second, the build seems to be failing, I'll open an issue for that.

pages/common/aria2.md
```diff
@@ -0,0 +1,33 @@
...
+
+- Download a web resource:
+
+`aria2c {{http://example.org/myLinux.iso}}`
+
+- Download a resource from multiple sources:
+
+`aria2c {{http://mirror1.org/myLinux.iso}} {{http://mirror2.org/myLinux.iso}}`
+
+- Download using "2" connections per host:
```
mebeim:
```suggestion
- Download using {{2}} connections per host:
```

mebeim: @pxgamer oh, then I must have wrongly assumed it was preferred since I saw it in some pages:

	marco:~/Documents/git/tldr$ grep -r '^-.*{{' pages/
	pages/linux/cryptsetup.md:- Open a LUKS volume and create a decrypted mapping at /dev/mapper/{{target}}:
	pages/common/samtools.md:- Index a sorted BAM file (creates {{sorted_input.bam.bai}}):
	pages/common/su.md:- Switch to user {{username}} (password required):
	pages/common/su.md:- Switch to user {{username}} and simulate a full login shell:
	pages/common/bedtools.md:- Intersect two files with respect to the sequences' strand and save the result to {{path/to/output_file}}:
	pages/common/bedtools.md:- Intersect two files with a left outer join, i.e. report each feature from {{file_1}} and NULL if no overlap with {{file_2}}:
	pages/common/bedtools.md:- Group file {{path/to/file}} based on the first three and the fifth column and summarize the sixth column by summing it up:
	pages/common/bedtools.md:- Find for all features in {{file_1}}.bed the closest one in {{file_2}}.bed and write their distance in an extra column (input files must be sorted):
	pages/common/mosh.md:- Select Mosh UDP port (useful when `{{remote_host}}` is behind a NAT):

I guess those are different scenarios though. So just ignore my comment! Sorry.

sbrl: I don't think the `2` here needs to be surrounded by anything.
```suggestion
- Download using 2 connections per host:
```

PR #2750, opened 2019-02-03
Cleaner deployment process (fixes #2749)

### LINKED ISSUE #2749: Travis build failure on PRs due to impossibility to decrypt the deploy key
[Travis fails to build pull requests](https://travis-ci.org/tldr-pages/tldr/builds/488184528) because on PRs the variables `$encrypted_973441be79af_key` and `$encrypted_973441be79af_iv` are not defined (for obvious security reasons). Since those are not defined, the `ssh-add` command waits for a passphrase on standard input, making the build time out and fail.

My eye didn't unfortunately spot this rather simple mistake when reviewing #2747, whoops :(

The solution is simple: only deploy when on the master branch. I'm working on a PR to solve this and will submit it as soon as possible.

mebeim:
Cleaned up the Travis configuration, now utilizing the `deploy` section. This section is only evaluated after a successful build, and also only on the master branch (not in PRs). The steps needed to decrypt the deploy key were [moved in the script](https://github.com/mebeim/tldr/blob/better-deployment/scripts/deploy.sh#L21). Checks for branch or PR where removed from the script since the script now only gets called when needed and therefore doesn't need to perform additional checks.

Also, since that's more of a *deploy* rather than a *build* script, I renamed it accordingly. Not really needed, but makes things clearer IMHO.


mfrw:
@mebeim @sbrl @agnivade merging this, as this my PR is blocked on this fix :)

PR #2760, opened 2019-02-07
rsync: add -r example with trailing backslash

coreysharris:
rsync -r src/bar /dest
is different from 
rsync -r src/bar/ /dest
and it's hard to remember which is which.

### DIFF
--- a/pages/common/rsync.md
+++ b/pages/common/rsync.md
@@ -19,6 +19,10 @@
 
 `rsync -r {{remote_host_name}}:{{remote_folder_location}} {{local_folder_location}}`
 
+- Transfer directory contents (but not the directory itself) from a remote into local:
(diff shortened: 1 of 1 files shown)


pages/common/rsync.md
```diff
@@ -19,6 +19,10 @@
 
 `rsync -r {{remote_host_name}}:{{remote_folder_location}} {{local_folder_location}}`
 
+- Transfer directory contents (but not the directory itself) from a remote into local:
```
sbrl: What do you mean by `into` local here? Is this "into a local directory?

coreysharris: This was a mistake... I've now replaced "into" with "to", so it matches the language of the example above it.

PR #2763, opened 2019-02-08
nc: fix mistake in command examples

lucianoq:

### DIFF
--- a/pages/common/nc.md
+++ b/pages/common/nc.md
@@ -16,11 +16,11 @@
 
 - Serve a file:
 
-`nc -l {{port}} < {{file}}`
+`nc {{ip_address}} {{port}} < {{file}}`
 
 - Receive a file:
 
-`nc {{ip_address}} {{port}} > {{file}}`
+`nc -l {{port}} > {{file}}`
 
 - Server stay up after client detach:
 


pages/common/nc.md
```diff
@@ -16,11 +16,11 @@
 
 - Serve a file:
 
-`nc -l {{port}} < {{file}}`
+`nc {{ip_address}} {{port}} < {{file}}`
```
sbrl: These appear to work for me. What seems to be the issue you're experiencing?

`-l` causes `nc` to **l**isten on the specified port number and serve the standard input. Without it, `nc` does the opposite: It connects to a listener and downloads content as a client, which it outputs to the standard output (you can also send content to the listening server via the standard input, IIRC).

mebeim: I think there's been a misunderstanding since to "receive" a file one could either listen for it or get it from a server which is serving it. Either of the following options works:

1. Serve with `nc -l port < file`, receive with `nc server_addr port > file`.
2. Wait for the file with `nc -l port > file`, send with `nc client_addr port < file`.

The second one is what I usually do too. I usually *wait for a file* with the `-l` option and have the other party *send it to me* with the classic `nc addr port < file`. That's also more secure IMHO since you cannot mistakenly send the file to someone else, while using the first option anyone could connect to you and get the file.

So yeah, that's basically what @lucianoq was thinking when changing this page. ¯\_(ツ)_/¯

mebeim:
I think there's been a misunderstanding since to "receive" a file one could either listen for it or get it from a server which is serving it. Either of the following options works:

1. Serve with `nc -l port < file`, receive with `nc server_addr port > file`.
2. Wait for the file with `nc -l port > file`, send with `nc client_addr port < file`.

The second one is what I usually do too. I usually *wait for a file* with the `-l` option and have the other party *send it to me* with the classic `nc addr port < file`. That's also more secure IMHO since you cannot mistakenly send the file to someone else, while using the first option anyone could connect to you and get the file.

So yeah, that's basically what @lucianoq was thinking when changing this page. ¯\_(ツ)_/¯

agnivade:
Ah I see now. I think the first option is more common to setup nc as a server. I feel the current examples are fine as is.

mebeim:
@agnivade agree :slightly_smiling_face:

lucianoq:
@mebeim got exactly I meant. `nc` to send and `nc -l` to receive. I didn't think about the other way.
But it's fine.

PR #2779, opened 2019-02-13
sam: add page

tmoon8730:
Added a page for the AWS SAM CLI tool

### DIFF
--- a/pages/linux/sam.md
+++ b/pages/linux/sam.md
@@ -1,6 +1,7 @@
 # sam
 
 > AWS Serverless Application Model (SAM) CLI.
+> Homepage: <https://github.com/awslabs/aws-sam-cli>
 
 - Initialize a serverless application:
 
@@ -10,7 +11,7 @@
 
 `sam init --runtime {{python3.7}}`
 
-- Package a SAM application (Alias of *aws cloudformation package*):
+- Package a SAM application:
 
 `sam package`
 
@@ -22,6 +23,6 @@
 
 `sam local start-api`
 
-- Deploy an AWS SAM application (Alias of *aws cloudformation deploy*):
+- Deploy an AWS SAM application:
 
 `sam deploy`
--- a/pages/linux/sam.md
+++ b/pages/linux/sam.md
@@ -1,7 +1,7 @@
 # sam
 
 > AWS Serverless Application Model (SAM) CLI.
-> Homepage: <https://github.com/awslabs/aws-sam-cli>
+> Homepage: <https://github.com/awslabs/aws-sam-cli>.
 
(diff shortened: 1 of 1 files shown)


pages/linux/sam.md
```diff
@@ -0,0 +1,27 @@
...
+
+> AWS Serverless Application Model (SAM) CLI.
+
+- Initialize a serverless application:
+
+`sam init`
+
+- Initialize a serverless application with a specific runtime:
+
+`sam init --runtime python3.7`
```
mebeim:
```suggestion
`sam init --runtime {{python3.7}}`
```

Parameters entered by the user should be enclosed in double curly brackets.

tmoon8730: Updated :+1:

pages/linux/sam.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Initialize a serverless application:
+
+`sam init`
+
+- Initialize a serverless application with a specific runtime:
+
+`sam init --runtime {{python3.7}}`
+
+- Package a SAM application (Alias of *aws cloudformation package*):
```
sbrl: I'm not sure we need to mention an alias here. Perhaps we could write a page for `aws cloudformation` and put

```
> See also `aws cloudformation`.
```

as a 2nd description line?

pages/linux/sam.md
```diff
@@ -0,0 +1,27 @@
+# sam
+
+> AWS Serverless Application Model (SAM) CLI.
+
```
sbrl: Perhaps we could add a new line here:

```
> Homepage: <https://github.com/awslabs/aws-sam-cli>
```

This is the new homepage syntax we defined in #2649.

pages/linux/sam.md
```diff
@@ -0,0 +1,27 @@
...
+- Build your Lambda function code:
+
+`sam build`
+
+- Run your serverless application locally:
+
+`sam local start-api`
+
+- Deploy an AWS SAM application (Alias of *aws cloudformation deploy*):
```
sbrl: Same here.

pages/linux/sam.md
tmoon8730: Added. Thanks for the suggestion!

pages/linux/sam.md
tmoon8730: Great suggestion! For not I have removed the references all together as adding pages for the AWS CLI will take some time as it is pretty large. If there is interest in adding those pages I would be happy to do so and update this page to reference it.

pages/linux/sam.md
tmoon8730: See comment above

PR #2781, opened 2019-02-13
vdir: add page

mebeim:

---

Another one from the GNU Coreutils (#2213).

### DIFF
--- a/pages/common/vdir.md
+++ b/pages/common/vdir.md
@@ -7,7 +7,7 @@
 
 `vdir`
 
-- List files with size displayed using human readable units (KB, MB, GB):
+- List files with sizes displayed in human readable units (KB, MB, GB):
 
 `vdir -h`
 
--- a/pages/common/vdir.md
+++ b/pages/common/vdir.md
@@ -3,30 +3,30 @@
 > List directory contents.
 > Drop-in replacement for `ls -l`.
 
-- List files in the current directory, one per line, with details:
+- List files and directories in the current directory, one per line, with details:
 
 `vdir`
 
-- List files with sizes displayed in human readable units (KB, MB, GB):
+- List with sizes displayed in human readable units (KB, MB, GB):
 
 `vdir -h`
 
-- List all files, including hidden files (starting with a dot):
+- List including hidden files (starting with a dot):
 
 `vdir -a`
 
-- List files sorting them by size (largest first):
+- List files and directories sorting entries by size (largest first):
 
 `vdir -S`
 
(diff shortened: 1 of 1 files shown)


pages/common/vdir.md
```diff
@@ -0,0 +1,32 @@
+# vdir
+
+> List directory contents.
+> Drop-in replacement for `ls -l`.
+
+- List files in the current directory, one per line, with details:
+
+`vdir`
+
+- List files with size displayed using human readable units (KB, MB, GB):
```
sbrl: Perhaps this will read better?

```suggestion
- List files with sizes displayed in human readable units (KB, MB, GB):
```

pages/common/vdir.md
```diff
@@ -0,0 +1,32 @@
+# vdir
+
+> List directory contents.
+> Drop-in replacement for `ls -l`.
+
+- List files in the current directory, one per line, with details:
```
sbrl: So this doesn't list subdirectories?

pages/common/vdir.md
mebeim: I just used the same wording [as the `ls` command](https://github.com/tldr-pages/tldr/blob/master/pages/common/ls.md), but yeah I guess this could be updated if you think is clearer. It is clear either way to me so I can go with whichever you prefer.

pages/common/vdir.md
mebeim: Nope, it doesn't. It is a drop-in replacement for `ls -l` as I said. See: 

image

sbrl: Ah, by "subdirectories" I meant whether it listed directories that are subdirectories fot he directory that you're listing the contents of.  By that screenshot it looks like it does.

pages/common/vdir.md
sbrl: I think my suggestion is a little clearer.

pages/common/vdir.md
mebeim: @sbrl Oh, yeah, it does! Do you think examples would be better rewritten saying "files and directories" instead of just "files" everywhere?

pages/common/vdir.md
mebeim: Applied :+1:

pages/common/vdir.md
sbrl: Ah, I see. I think it would probably be clearer if we did use `files and directories` in this instance.

mebeim: You're right, pushed a reword.

PR #2784, opened 2019-02-14
npm-why: add page

sbrl:
Another day, another page on my continuing quest to have a page for every command in my shell history :P

PR #2785, opened 2019-02-15
yarn-why: add page

pxgamer:
Looking at #2784 brought me to a Yarn equivalent of `npm-why`. Thanks @sbrl 

### DIFF
--- a/pages/common/yarn-why.md
+++ b/pages/common/yarn-why.md
@@ -1,0 +1,8 @@
+# yarn-why
+
+> Identifies why a Yarn package has been installed.
+> Homepage: <https://www.npmjs.com/package/yarn-why>.
+
+- Show why a Yarn package is installed:
+
+`yarn-why {{package_name}}`


pxgamer:
Yarn lock files actually lock the dependencies (like Composer). Yarn also used to be much faster at installing, etc. before `npm v6`.

I've recently moved back from Yarn to npm to be honest though as the speed differences are barely noticeable I've found.

agnivade:
After switching to yarn, life has brought me peace. I don't `rm -rf node_modules/` anymore !

sbrl:
I see! `npm` does support locking dependencies via `package-lock.json` now though.

pxgamer:
@sbrl, but unless you use `npm ci` the default doesn't really use the lock.

In Composer and Yarn, if you run an install, it will use the lock. In `npm install` it will update the lock.

agnivade:
Yes, I think it was using package-lock.json from v 3 onwards. I have faced problems even with it.

sbrl:
@pxgamer: TIL! Didn't know that.

PR #2788, opened 2019-02-15
paccache: add page

DerPhysikeR:

Thanks for tldr! I find it really useful and found paccache to be missing, so I created a page for it.

### DIFF
--- a/pages/linux/paccache.md
+++ b/pages/linux/paccache.md
@@ -1,0 +1,19 @@
+# paccache
+
+> A flexible pacman cache cleaning utility.
+
+- Remove all but the 3 most recent package versions from the pacman cache:
+
+`paccache -r`
+
+- Set the number of package versions to keep:
+
+`paccache -rk {{num_versions}}`
+
+- Perform dryrun and show the number of candidate packages:
+
+`paccache -d`
+
+- Move candidate packages to a directory instead of deleting them:
+
+`paccache -m {{directory}}`
--- a/pages/linux/paccache.md
+++ b/pages/linux/paccache.md
@@ -10,10 +10,10 @@
 
 `paccache -rk {{num_versions}}`
 
-- Perform dryrun and show the number of candidate packages:
+- Perform a dry-run and show the number of candidate packages for deletion:
 
 `paccache -d`
 
 - Move candidate packages to a directory instead of deleting them:
 
-`paccache -m {{directory}}`
+`paccache -m {{path/to/directory}}`


pages/linux/paccache.md
```diff
@@ -0,0 +1,19 @@
+# paccache
+
+> A flexible pacman cache cleaning utility.
+
+- Remove all but the 3 most recent package versions from the pacman cache:
+
+`paccache -r`
```
sbrl: Is there a long form of this option? We generally prefer the longer form of options here at tldr-pages, as they are more descriptive for new users.

The same goes for the arguments in subsequent examples.

pages/linux/paccache.md
```diff
@@ -0,0 +1,19 @@
...
+
+`paccache -rk {{num_versions}}`
+
+- Perform dryrun and show the number of candidate packages:
+
+`paccache -d`
+
+- Move candidate packages to a directory instead of deleting them:
+
+`paccache -m {{directory}}`
```
sbrl: Could `directory` be `path/to/directory` please? It'll make it more consistent with other pages.

pages/linux/paccache.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Remove all but the 3 most recent package versions from the pacman cache:
+
+`paccache -r`
+
+- Set the number of package versions to keep:
+
+`paccache -rk {{num_versions}}`
+
+- Perform dryrun and show the number of candidate packages:
```
sbrl: Perhaps this will read better?

```suggestion
- Perform a dry-run and show the number of candidate packages for deletion:
```

DerPhysikeR: Sounds good, I will change it.

pages/linux/paccache.md
DerPhysikeR: Sounds good too, I will change it.

pages/linux/paccache.md
DerPhysikeR: I thought  that according to point 3 of the contribution guidelines it is enough to mention the long version of the option in the description of the command, since neither the pages for pacman nor pacaur use the long version of their options.
But if you prefer the long version, that's fine by me. Should I also change the other options?
-r -> --remove, -k -> --keep, -m -> --move, -d -> --dryrun

sbrl: Might be a good idea. Though _pacman_-based commands appear to be somewhat of an exception.

What do you suggest, @agnivade / @pxgamer?

agnivade: If the short-form option is unambiguously clear, which in this case, I think it is. Then it is fine to keep it. If `-r` meant something like update packages, then sure. Here since `-r` means `remove`, it is fine to use the short-form.

pxgamer: I normally use long forms for most examples, but I think it seems fine using `-r` for `remove` here. 👍

DerPhysikeR:
I made the discussed changes and kept the short-form option -r.
For -k, -m and -d I also kept the short-form, I hope that's ok.

PR #2795, opened 2019-02-21
nkf: add page

pr0xy-t:

### DIFF
--- a/pages/common/nkf.md
+++ b/pages/common/nkf.md
@@ -1,8 +1,9 @@
 # nkf
 
+> Network kanji filter
 > Convert character code and new line code.
 
-- Stdout utf-8 code:
+- Convert to UTF-8 encoding:
 
 `nkf -w {{path/to/file.txt}}`
 
@@ -10,7 +11,7 @@
 
 `nkf -s {{path/to/file.txt}}`
 
-- Overwrite as utf-8 code:
+- Convert to UTF-8 encoding and overwrite the file:
 
 `nkf -w --overwrite {{path/to/file.txt}}`
 
--- a/pages/common/nkf.md
+++ b/pages/common/nkf.md
@@ -1,7 +1,7 @@
 # nkf
 
 > Network kanji filter
-> Convert character code and new line code.
+> Converts kanji code from one encoding to another.
 
 - Convert to UTF-8 encoding:
 
--- a/pages/common/nkf.md
+++ b/pages/common/nkf.md
@@ -7,7 +7,7 @@
 
 `nkf -w {{path/to/file.txt}}`
 
-- Stdout shift_jis code:
+- Convert to SHIFT_JIS encoding:
 
 `nkf -s {{path/to/file.txt}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/nkf.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Stdout shift_jis code:
+
+`nkf -s {{path/to/file.txt}}`
+
+- Overwrite as utf-8 code:
+
+`nkf -w --overwrite {{path/to/file.txt}}`
+
+- Set new line code to LF and overwrite(unix type):
```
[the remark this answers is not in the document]
imaiproxy: It becomes easy to see😃

pages/common/nkf.md
```diff
@@ -0,0 +1,27 @@
+# nkf
+
+> Convert charcter code and new line code.
```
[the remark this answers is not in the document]
imaiproxy: I will be careful😔

pages/common/nkf.md
```diff
@@ -0,0 +1,27 @@
+# nkf
+
+> Convert character code and new line code.
```
agnivade: Please add "Network kanji filter" as the main heading.

From the man page here https://linux.die.net/man/1/nkf, it seems that it converts input kanji code from one encoding to another. I think that is more clear. The current description does not mention what does it convert to.

pages/common/nkf.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Stdout utf-8 code:
```
agnivade: Stdout is not a verb. Please change it to "Convert to UTF-8 encoding"

pages/common/nkf.md
```diff
@@ -0,0 +1,27 @@
...
+
+`nkf -w {{path/to/file.txt}}`
+
+- Stdout shift_jis code:
```
agnivade: same as above.

pages/common/nkf.md
```diff
@@ -0,0 +1,27 @@
...
+
+`nkf -s {{path/to/file.txt}}`
+
+- Overwrite as utf-8 code:
```
agnivade: Convert to UTF-8 encoding and overwrite the file.

Also, I see --overwrite has a suffix value in the man page. Is that not true ?

pages/common/nkf.md
agnivade: Seems to be another encoding from https://linux.die.net/man/1/nkf. I would suggest the same format as above.

pages/common/nkf.md
agnivade: Possible. But it would be good to confirm this.

pages/common/nkf.md
sbrl: Done.

zdroid:
I checked the manpage and it seems all examples are written correctly. `--overwrite` requires a suffix which is the file name in this case.

agnivade:
If I am not mistaken, the overwrite suffix should mean what gets appended to the end of the filename that gets overwritten. The file name is obviously part of the command, but suffix is something that gets added at the end, if specified. But it's optional anyways, so I think it's fine.

PR #2802, opened 2019-02-24
multiple pages: change Unicode characters to ASCII

urbalazs:
Change Unicode characters to ASCII:

- non-breaking spaces (\xc2\xa0) to normal spaces (\x20)
- apostrophes (there are only 2 Unicode apostrophes, all others are already ASCII)
- hyphen (−) to minus (-)
- fix wording (STD‐OUT to "standard output")

### DIFF
--- a/pages/common/flutter.md
+++ b/pages/common/flutter.md
@@ -1,6 +1,6 @@
 # flutter
 
-> Google’s free, open source, and cross-platform mobile app SDK.
+> Google's free, open source, and cross-platform mobile app SDK.
 
 - Check the Flutter version:
 
--- a/pages/osx/launchctl.md
+++ b/pages/osx/launchctl.md
@@ -23,10 +23,10 @@
 
 `launchctl unload ~/Library/LaunchAgents/{{my_script}}.plist`
 
-- Manually run a known (loaded) agent/daemon, even if it isn’t the right time (note: this command uses the agent's label, rather than the filename):
+- Manually run a known (loaded) agent/daemon, even if it is not the right time (note: this command uses the agent's label, rather than the filename):
 
 `launchctl start {{my_script}}`
 
-- Manually kill the process associated with a known agent/daemon, if it's running:
+- Manually kill the process associated with a known agent/daemon, if it is running:
 
 `launchctl stop {{my_script}}`
--- a/pages/common/mpv.md
+++ b/pages/common/mpv.md
@@ -24,4 +24,4 @@
 
 - Play a file using a profile defined in the `mpv.conf` file:
 
-`mpv --profile {{profile_name}} {{file}}`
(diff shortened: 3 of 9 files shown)


pages/osx/launchctl.md
```diff
@@ -23,10 +23,10 @@
 
 `launchctl unload ~/Library/LaunchAgents/{{my_script}}.plist`
 
-- Manually run a known (loaded) agent/daemon, even if it isn’t the right time (note: this command uses the agent's label, rather than the filename):
+- Manually run a known (loaded) agent/daemon, even if it is not the right time (note: this command uses the agent's label, rather than the filename):
 
 `launchctl start {{my_script}}`
 
-- Manually kill the process associated with a known agent/daemon, if it's running:
```
agnivade: I don't think this was necessary. We have a lot of places with "it's" in our repo.

pages/common/flutter.md
```diff
@@ -1,6 +1,6 @@
 # flutter
 
-> Google’s free, open source, and cross-platform mobile app SDK.
+> Google's free, open source, and cross-platform mobile app SDK.
```
sbrl: Most terminals can render unicode now. Why change it to ascii?

urbalazs: To be consistent with other texts. All apostrophes are ASCII, only this one was Unicode.

waldyrious: Not only it was inconsistent, but subtle differences like these can also introduce hard-to-diagnose issues when e.g. searching the content.

pages/osx/launchctl.md
waldyrious: I noticed that, and I agree it's not necessary, but I let it go in because it is equivalent, arguably more "explicit", and the other changes were worthwhile. Sorry if I jumped the gun :sweat_smile: I have a sweet spot for such cleanup changes, as a [wikignome](https://en.wikipedia.org/wiki/Wikipedia:WikiGnome) myself.

agnivade: Haha, it's just that this change stood out to me :laughing: . I think we should maintain a balance between usefulness and verbosity. It is not a huge deal, but in future PRs, let us go with whatever the author chooses to use.

pages/common/flutter.md
pxgamer: Windows is still pretty useless with most unicode characters. 😉

sbrl: Ah, I see! Thanks for the clarification :-)

Regarding searching though, generally a decent search engine will [transliterate to ascii automagically before indexing](https://starbeamrainbowlabs.com/blog/article.php?article=posts/305-Search-Engine-Optimisation-The-Curious-Question.html).

PR #2803, opened 2019-02-25
Add more Chinese translations

starccy:


agnivade:
Thanks for the PR. Please also translate the text within the tokens.

/cc @HugueLiu

starccy:
Do you mean I should translate the text between `{{}}` ? If yes, I am sorry for missing it. I will fix that.     
What's more, I note that someone would like translate by using automation scripts. Whether it means there is no need to translate manually any more?

agnivade:
> Do you mean I should translate the text between `{{}}` ?

Correct.

> Whether it means there is no need to translate manually any more?

For now, that workflow is not applied yet. So you do not need to worry about that.

sbrl:
I believe the automation scripts you refer to is a system by why translators can get notified about changes to the original versions of pages they have translated, but since that's still a work-in-progress, let's continue with the existing structure for now :-)

starccy:
Ok, I learned that. Thanks.

sbrl:
Are we ok to merge, @agnivade / @HugueLiu?

agnivade:
I still see some tokens untranslated.

starccy:
@agnivade I checked the files, Those untranslated tokens stand for filenames, I think it's unnecessary to translate them.

agnivade:
I see tokens like foo and bash. I meant that these should be transliterated to Chinese characters, not translated to their Chinese equivalents.

starccy:
I apologize for not replying you earlier, because I'm preparing for the exam. I would improve these stuff immediately.

PR #2806, opened 2019-02-26
grep: add file or directory exclusion

liumingc:
update grep with `--exclude-dir|--exclude` options


agnivade:
This exceeds our max limit of 8. If you have an example in mind that can be replaced with this, let us know. Although, the page has gone through several revisions. It mostly contains the top 8 usages of grep.

liumingc:
It's hard to choose which example to drop, so I think it's ok to remain the origin examples.

PR #2807, opened 2019-02-27
nix-env: add example of showing installed packages

aaronduino:

### DIFF
--- a/pages/common/nix-env.md
+++ b/pages/common/nix-env.md
@@ -2,6 +2,10 @@
 
 > Manipulate or query Nix user environments.
 
+- Show installed packages:
+
+`nix-env -q`
+
 - Show available package with name or without name:
 
 `nix-env -qa {{pkg_name}}`
--- a/pages/common/nix-env.md
+++ b/pages/common/nix-env.md
@@ -2,7 +2,7 @@
 
 > Manipulate or query Nix user environments.
 
-- List installed packages:
+- List all installed packages:
 
 `nix-env -q`
 
--- a/pages/common/nix-env.md
+++ b/pages/common/nix-env.md
@@ -6,13 +6,13 @@
 
 `nix-env -q`
 
-- Show available package with name or without name:
+- Query available packages:
 
 `nix-env -qa {{pkg_name}}`
 
-- Show the status of available packages:
+- Query the status of available packages:
 
-`nix-env -qas`
+`nix-env -qsa`
 
 - Install package:
 
--- a/pages/common/nix-env.md
+++ b/pages/common/nix-env.md
@@ -12,7 +12,7 @@
 
 - Query available packages:
 
-`nix-env -qa {{pkg_name}}`
+`nix-env -qa {{search_term}}`
 
 - Install package:
 
--- a/pages/common/nix-env.md
+++ b/pages/common/nix-env.md
@@ -2,17 +2,17 @@
 
 > Manipulate or query Nix user environments.
 
-- Show installed packages:
+- List installed packages:
 
 `nix-env -q`
+
+- Query installed packages:
+
+`nix-env -q {{search_term}}`
 
 - Query available packages:
 
 `nix-env -qa {{pkg_name}}`
-
-- Query the status of available packages:
-
-`nix-env -qsa`
 
 - Install package:
 


pages/common/nix-env.md
```diff
@@ -2,6 +2,10 @@
 
 > Manipulate or query Nix user environments.
 
+- Show installed packages:
+
+`nix-env -q`
+
 - Show available package with name or without name:
```
agnivade: I don't understand this one. How can it show a package without a name ?

pxgamer: I don't think that is part of this PR, but I think it should be `package(s)`. You can either list all packages, or ones matching the specified name query.

waldyrious: Yeah, the description could be improved, but that's out of scope of this PR. That said, if @aaronduino wants to do it, it would be appreciated :)

agnivade: Yeah, it's not part of the PR. But now the first description becomes part of this, since that also shows installed packages. I think we need to change the description to remove the "without name" part. Also, I would like a clarification on the difference between `-q` and `-a`. If `-qa` is available and `-q` is installed, what is `-a` ?

pxgamer: `-q` is shorthand for `--query`, `-a` is shorthand for `--available`.

waldyrious: We should consider incorporating those terms in the descriptions (as per [guideline n.3](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines)); that would help with understanding and memorizing the command's syntax.

agnivade: @aaronduino - Could you update the PR with the changes ?

aaronduino: How should I go about incorporating the word "query," considering that -q by itself does not let a user query the installed packages, but rather provides a full list of everything installed?

However, I'll definitely use the word "query" for the second and third examples.

aaronduino: Should I submit changes to the second and third examples as a separate PR or just as separate commits?

waldyrious: > How should I go about incorporating the word "query," considering that -q by itself does not let a user query the installed packages, but rather provides a full list of everything installed?

We can always use the [full version](https://nixos.org/nix/manual/#synopsis-6) `--query --instaled` (which is more verbose but not out of place in tldr-pages, since our mission is not to provide the shortest commands but understandable ones). Ideally we should be able to do `-qi` and use the spelled words in the description, but there isn't a short form of the `--installed` option :(

> Should I submit changes to the second and third examples as a separate PR or just as separate commits?

Additional commits in this PR are fine, thanks! 👍

pages/common/nix-env.md
```diff
@@ -6,13 +6,13 @@
...
-- Show available package with name or without name:
+- Query available packages:
 
 `nix-env -qa {{pkg_name}}`
 
-- Show the status of available packages:
+- Query the status of available packages:
 
-`nix-env -qas`
+`nix-env -qsa`
```
aaronduino: I changed the order of the flags so that `qsa` aligns with _query_ the _status_ of _available_ packages

aaronduino:
Turns out I was wrong. Installed packages can in fact be queried with a search term.

However, I think it is important to show that running `nix-env -q` without a search term lists all installed packages.

Focusing on using a description that matches the used flags, the following probably makes the most sense:

```markdown
Query installed packages:

`nix-env -q {{search_term}}`
```

However, focusing on what users most likely want to do, I think:

```markdown
List installed packages:

`nix-env -q`
```
would be more helpful.

Maybe we do both, replacing the `--status` example?

The [`--status` flag](https://nixos.org/nix/manual/#queries) doesn't seem to be very helpful to a novice user, especially since understanding its output of `--status` would require reading a more elaborate explanation.

The main use of the `--status` flag to a novice user would be to check if a package is installed. Maybe they do so by instead seeing if anything is returned by `nix-env -q {{pkg_name}}` (a query of installed packages)?

waldyrious:
@aaronduino what you write sounds sensible. Do you think adding `nix-env -q {{search_term}}` next to the `nix-env -q` example would lead to a more logical sequence?

aaronduino:
> Do you think adding `nix-env -q {{search_term}}` next to the `nix-env -q` example would lead to a more logical sequence?

Yes, I think so.

I'll commit the changes I made.

pages/common/nix-env.md
```diff
@@ -2,13 +2,17 @@
...
+
+- Query installed packages:
 
-- Show the status of available packages:
+`nix-env -q {{search_term}}`
 
-`nix-env -qas`
+- Query available packages:
+
+`nix-env -qa {{pkg_name}}`
```
aaronduino: Should I change this to use `{{search_term}}` like the previous example, instead of `{{pkg_name}}`?

If you already know the exact name of a package (as suggested by `{{pkg_name}}` but not `{{search_term}}`), it wouldn't be very helpful to search for it.

waldyrious: Agreed. It is more consistent, and gives more meaning to the term "query".

pages/common/nix-env.md
```diff
@@ -2,13 +2,17 @@
 
 > Manipulate or query Nix user environments.
 
-- Show available package with name or without name:
+- List installed packages:
```
waldyrious: Perhaps phrasing this as "List all installed packages" will tie better with the next example, so that this one can more easily be mentally shifted towards the correct meaning (as in "oh, so that was also a query, but with no query terms, so it returns all results"). I feel that the "all" can help in gently nudging the brain towards this interpretation. Just a thought — WDYT?

aaronduino: I agree, good catch. I'll make that change right now

pages/common/nix-env.md
waldyrious: With the recent changes, I don't think using `--query --installed` is needed anymore, so I'll mark this conversation as resolved.

waldyrious:
Thanks for the contribution and thoughtful discussion, @aaronduino!

aaronduino:
I don't mind at all. Thanks for all the helpful feedback.

It's exciting to give back to the community!

waldyrious:
We're glad to have you! Hope you'll stick around. Be careful, it gets a little addictive ;)

PR #2745, opened 2019-02-01
Fix ellipsis (...) usage and consistency in pages

mebeim:
From a comment on #2742 by @agnivade:
> Yeah, there does seem to be a few -
> 
>     -~/play/tldr/pages$git grep "\.\.\." | grep "\`"
>     common/7z.md:`7z a -t {{zip|gzip|bzip2|tar|...}} {{archived.7z}} {{path/to/file}}`
>     common/7za.md:`7za a -t{{zip|gzip|bzip2|tar|...}} {{archived}} {{path/to/file}}`
>     common/cowsay.md:`cowthink -s "I'm just a cow, not a great thinker ..."`
>     common/nmap.md:`nmap -p {{port1,port2,...,portN}} {{address_or_addresses}}`
>     common/pdftk.md:`pdftk {{file1.pdf}} {{file2.pdf}} ... cat output {{output.pdf}}`
>     common/samtools.md:`samtools merge {{output}} {{input_1}} [{{input_2}}...]`
>     common/svn.md:`svn add PATH...`
>     common/svn.md:`svn ci -m {{commit log message}} {{[PATH...]}}`
>     common/tig.md:`tig {{path1}} {{path2}} ...`
>     linux/ss.md:`ss {{state/exclude}} {{bucket/big/connected/synchronized/...}}`
>
>We can remove them from 7z, 7za, pdftk, samtools, svn and tig.

---

I removed the three dots where not needed (`7z`, `7za`, `svn`), and I converted ellipsis from the Unicode character `…` (U+206) to three dots in `calibredb` (both English and Italian).

I feel like on some commands, namely `calibredb`, `pdftk`, `samtools` and `tig`, the use of `...` makes the command more clear than its absence. So I edited those to make the use of `...` consistent and clear.

Let me know what you guys think about this.

PR #2808, opened 2019-02-27
nix: add page

aaronduino:

While users can already search for packages using `nix-env`, `nix search` searches both package names and descriptions instead of just package names.

For example, `nix search browser` lists firefox, chromium, etc while `nix-env -qa browser` returns no results.


waldyrious:
I'd rather the page had at least 2 or 3 examples, if that's not too much of a bother :)

aaronduino:
No problem at all. The file now has 5 examples.

I learned a few helpful sub-commands in the process.

pages/common/nix.md
```diff
@@ -0,0 +1,23 @@
+# nix
+
+> Utilities for the Nix language and store.
+
+- Search for a package via its name or description:
+
+`nix search {{search_term}}`
+
+- Start a nix shell with the specified packages available:
```
agnivade: N should be capital here to be consistent with the rest of the page.

aaronduino: Nice catch!

mebeim:
Looks ok to me, but shouldn't this be moved under `linux`?

pages/common/nix.md
waldyrious: @aaronduino this doesn't seem to have been changed yet — marking as unresolved so this doesn't get lost in the conversation.

pages/common/nix.md
```diff
@@ -5,3 +5,19 @@
 - Search for a package via its name or description:
 
 `nix search {{search_term}}`
+
+- Start a nix shell with the specified packages available:
+
+`nix run {{nixpkgs.pkg_names}}`
```
waldyrious: It's unclear from this syntax how one would provide multiple packages. Could we rework this example to make that clearer?

aaronduino: Perhaps `nix run {{nixpkgs.pkg1}} {{nixpkgs.pkg2}} {{nixpkgs.pkg3...}}`?

The `nixpkgs` string indicates that the the packages must be prefixed by the "channel" they come from (almost always `nixpkgs` for non-advanced users).

aaronduino:
@mobeim nix.md is now located in the linux directory.

aaronduino:
Actually, `nix` works on both osx and Linux. I think that means it should be in `common/`

agnivade:
That means copies of it should be in osx and linux directories separately.

pages/common/nix.md
agnivade: We can do {{nixpkgs.pkg1 nixpkgs.pkg2 nixpkgs.pkg3…}}

Note that … is an unicode ellipsis.

aaronduino: Weird, the comment box must have done that. Would it be okay to commit it I replace the unicoded ellipsis with 3 periods?

waldyrious: I agree it makes more sense to wrap the entire list in the token markers, rather than having multiple separate tokens with an ellipsis inside the last one.

As for the Unicode ellipsis, it seems to me that @agnivade was suggesting to use the Unicode one rather than three ASCII periods. However, that kinda goes against #2802 (/cc @urbalazs), so I would vote for using a regular ASCII ellipsis. That said, I don't have a strong opinion on the matter.

waldyrious:
> That means copies of it should be in osx and linux directories separately.

@agnivade did we ever define a concrete policy for when to place pages in `common/` vs. the platform-specific directories? I was under the impression that separate copies were only appropriate when the actual contents differ. After all, we don't make the claim that a page being in `common/` means it is available in all systems.

aaronduino:
Oops, my mistake misreading that.

I agree that unicode could be problematic, especially since not all terminal emulators support it.

I'll combine the markers into a list the way you did—it looks much nicer.

-------- Original Message --------

pages/common/nix.md
agnivade: @mebeim recently made a commit which changed all 3 ASCII periods to unicode ellipsis. I don't have much expertise on this matter. Just want things to be consistent.

aaronduino: In that case I assume it would make most sense to comply with the rest of the project? If unicode is a problem then it should be handled for the entire repository, not just this file.

pages/common/nix.md
aaronduino: Looking at the most recent file view (clicking on files changed, not on this page), it seems to be fixed, no? https://github.com/tldr-pages/tldr/pull/2808/files

agnivade:
> After all, we don't make the claim that a page being in `common/` means it is available in all systems.

Hmm .. this policy has always confused me. I thought common means it is available in all systems. I know that there are currently pages in common which are not available in windows, but I am not sure whether that is intentional or something to be fixed.

I like your logic better though. Going forward, we should document this somewhere so that I don't forget it again :stuck_out_tongue:

pages/common/nix.md
waldyrious: Well, #2802 did not edit those (which got in earlier, in #2745), so my guess is that the intention was not to remove all non-ASCII characters altogether, but focus on particularly confusable / easily missed ones.

On the other hand, I'm still wary for the same reason I argued against requiring trailing slashes in URLs [here](https://github.com/tldr-pages/tldr/pull/2684#issuecomment-452132239): it seems to me that people are more likely to use regular ellipsis than Unicode ones when authoring pages, so we'd have to regularly go through all pages to find inconsistencies (or create automated checks).

Just food for thought -- if you guys still prefer Unicode ellipsis, I won't insist further (in this thread 😜)

aaronduino:
Okay, that makes sense. I moved the file back to common.

pages/common/nix.md
pxgamer: @waldyrious, is this ok to merge?

pages/common/nix.md
agnivade: Yeah, I don't mind either way. @sbrl and @mebeim were both in favor of adding unicode ellipsis and I was neutral.

Guys, what are your thoughts on @waldyrious' reasoning ?

agnivade:
@mebeim - Did you see this comment https://github.com/tldr-pages/tldr/pull/2808#discussion_r261947741 of mine ? It's easy to miss, so just wanted to confirm.

pages/common/nix.md
waldyrious: Yep, this is indeed fixed.

sbrl:
I was under the impression that things could only go in `common/` if they were available in all systems.

mebeim:
@agnivade oh sorry about that, missed the comment. I actually don't know where to draw the line about those ellipsis. I think Unicode looks better but some terminals don't support it or display it wrongly, so probably three standard dots are better... not a big deal though as far as it's understandable it's the same IMHO.

As per where pages should go, that's right, currently there's no clear specification about what should or should not go under `common`. Not an easy choice to make :\ and of course it brings the problem of keeping track of different versions of pages if there are different versions.

agnivade:
Alright, should we revert to 3 dots then ?

mebeim:
@agnivade probably, seems more compatible 😅

PR #2809, opened 2019-02-28
nixos-rebuild: add page

aaronduino:

This command is central to the NixOS operating system; [official documentation](https://nixos.org/nixos/manual/#sec-changing-config)


aaronduino:
Turns out I've learned tons about `nixos-rebuild` by writing this tldr page. I think my modifications are complete (apart from acting on maintainer feedback). I didn't intend to add anything after the initial submission.

pages/common/nixos-rebuild.md
```diff
@@ -0,0 +1,31 @@
+# nixos-rebuild
+
+> Reconfigure a NixOS machine.
+
+- Build and switch to the new configuration, making it the boot default:
```
mebeim: I see this example has basically the same description as the 5th one. What's the difference? I think one of them (if not both) should be reworded to be clearer.

pages/common/nixos-rebuild.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Build and switch to the new configuration, making it the boot default and installing updates:
+
+`sudo nixos-rebuild switch --upgrade`
+
+- Rollback changes to the configuration, switching to the previous generation:
+
+`sudo nixos-rebuild switch --rollback`
+
+- Build the new configuration and make it the boot default:
```
mebeim: See above.

aaronduino:
@mebeim the way it works is that nixos-rebuild does a combination of switching the OS to a new configuration while it is running and adding a boot entry,

Name | Switch | Boot
---|---|---
Switch| x| x
Boot||x
Test|x|

aaronduino:
I'm moving this to the Linux folder since it is only a available on the NixOS distro of Linux.

pages/common/nixos-rebuild.md
aaronduino: Does the change I made fix this issue? https://github.com/tldr-pages/tldr/pull/2809/commits/926f8a8c4889d5859cf8c90b5bdfcc4143a167f0

pages/linux/nixos-rebuild.md
```diff
@@ -0,0 +1,31 @@
+# nixos-rebuild
+
+> Reconfigure a NixOS machine.
```
sbrl: Perhaps we could add an additional line below this one that links to the documentation with the new homepage syntax we drafted up in #2649?

```
> Documentation: https://nixos.org/nixos/manual/#sec-changing-config
```

aaronduino: Okay, I just committed the above change

aaronduino:
Is that true? Couldn't adding a period to a URL be problematic?

pages/linux/nixos-rebuild.md
sbrl: Cool! I've just realised that I got the syntax incorrect in my above comment. I've made quick commit to fix it, if you don't mind?

sbrl:
Oops, @aaronduino! See my comment on the original thread.

pages/linux/nixos-rebuild.md
aaronduino: No problem at all. Thanks!

PR #2810, opened 2019-02-28
sd: add page

jesse-c:

### DIFF
--- a/pages/common/sd.md
+++ b/pages/common/sd.md
@@ -16,4 +16,4 @@
 
 - Find and replace across a project:
 
-`sd -i 'from "react"' 'from "preact"' $(fd -t f)`
+`sd -i {{'from "react"'}} {{'from "preact"'}} $(fd -t f)`
--- a/pages/common/sd.md
+++ b/pages/common/sd.md
@@ -16,4 +16,4 @@
 
 - Find and replace across a project:
 
-`sd -i {{'from "react"'}} {{'from "preact"'}} $(fd -t f)`
+`sd -i {{'from "react"'}} {{'from "preact"'}} $(find . -type f)`
--- a/pages/common/sd.md
+++ b/pages/common/sd.md
@@ -4,7 +4,7 @@
 
 - Trim some whitespace using regex:
 
-`echo 'lorem ipsum 23   ' | sd '\s+$' ''`
+`{{echo 'lorem ipsum 23   '}} | sd '\s+$' ''`
 
 - Replace words using capture groups:
 
--- a/pages/common/sd.md
+++ b/pages/common/sd.md
@@ -1,0 +1,19 @@
+# sd
+
+> `sd` is an intuitive find & replace CLI.
+
+- Trim some whitespace using regex:
+
+`echo 'lorem ipsum 23   ' | sd '\s+$' ''`
+
+- Replace words using capture groups:
+
+`echo 'cargo +nightly watch' | sd '(\w+)\s+\+(\w+)\s+(\w+)' 'cmd: $1, channel: $2, subcmd: $3'`
+
+- Find and replace in a file:
+
+`sd -i 'window.fetch' 'fetch' http.js`
+
+- Find and replace across a project:
+
+`sd -i 'from "react"' 'from "preact"' $(fd -t f)`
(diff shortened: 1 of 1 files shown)


pages/common/sd.md
```diff
@@ -0,0 +1,19 @@
+# sd
+
+> `sd` is an intuitive find & replace CLI.
+
+- Trim some whitespace using regex:
+
+`echo 'lorem ipsum 23   ' | sd '\s+$' ''`
```
pxgamer: We usually add tokens around custom input examples.
```suggestion
`{{echo 'lorem ipsum 23   '}} | sd '\s+$' ''`
```

pages/common/sd.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Replace words using capture groups:
+
+`echo 'cargo +nightly watch' | sd '(\w+)\s+\+(\w+)\s+(\w+)' 'cmd: $1, channel: $2, subcmd: $3'`
```
pxgamer: Same here
```suggestion
`{{echo 'cargo +nightly watch'}} | sd '(\w+)\s+\+(\w+)\s+(\w+)' 'cmd: $1, channel: $2, subcmd: $3'`
```

pages/common/sd.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Find and replace in a file:
+
+`sd -i 'window.fetch' 'fetch' http.js`
```
pxgamer:
```suggestion
`sd -i '{{window.fetch}}' '{{fetch}}' {{http.js}}`
```

pages/common/sd.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Find and replace across a project:
+
+`sd -i 'from "react"' 'from "preact"' $(fd -t f)`
```
pxgamer:
```suggestion
`sd -i '{{from "react"}}' '{{from "preact"}}' $(fd -t f)`
```

pages/common/sd.md
jesse-c: The changes look good, but just for my own sake, I'm unfamiliar with the double-curly brace. I've used singles before in this context. I tried the updated diff in both zsh and bash, and neither work for me. Am I missing something obvious? 😅

pxgamer: It's actually the token syntax for the `tldr` project. 👍 It allows the `tldr` clients to highlight customisable entries in the pages. The [style guide](https://github.com/tldr-pages/tldr/blob/master/contributing-guides/style-guide.md#token-syntax) explains how this works.

jesse-c: 🤦🏼‍♂️

Thank you.

jesse-c:
Would it be best to squash the commits locally and force push onto the branch?

pxgamer:
I think it should be fine, we have a button to squash commits when we merge the PR. 👍

jesse-c:
Sounds good. 🙏🏻

pages/common/sd.md
```diff
@@ -0,0 +1,19 @@
...
+
+`{{echo 'cargo +nightly watch'}} | sd '(\w+)\s+\+(\w+)\s+(\w+)' 'cmd: $1, channel: $2, subcmd: $3'`
+
+- Find and replace in a file:
+
+`sd -i {{'window.fetch'}} {{'fetch'}} {{http.js}}`
+
+- Find and replace across a project:
+
+`sd -i {{'from "react"'}} {{'from "preact"'}} $(fd -t f)`
```
sbrl: What's `fd`?

jesse-c: A Rust alternative to `find`. [1]

[1] https://github.com/sharkdp/fd

sbrl: I see. Generally, a page should document the functionality of the command itself, instead of introducing combinations of other commands.

In this instance, if we were to include the `fd` here, I'd be inclined to use a regular-old `find`, as most users will have that installed by default.

jesse-c: That's a good point. I've updated it to use `find`. Thanks for the feedback!

jesse-c:
Thank you!

sbrl:
np :-)

PR #2811, opened 2019-02-28
ack: clarify file types syntax meaning

### LINKED ISSUE #2805: ack: meaningless value
In [ack](https://github.com/tldr-pages/tldr/blob/master/pages/common/ack.md): `ack --ruby {{each_with_object}}`

What does it mean "each_with_object"? I tried to find out from [ack documentation](https://beyondgrep.com/documentation/), but without success.
Please use a more descriptive value name!

negamorgan:
fix #2805

### DIFF
--- a/pages/common/ack.md
+++ b/pages/common/ack.md
@@ -6,7 +6,7 @@
 
 `ack {{foo}}`
 
-- Find files scoped to a specific programming language:
+- Find files of a specific type:
 
 `ack --ruby {{foo}}`
 
@@ -17,3 +17,7 @@
 - Show the file names containing "foo" and number of matches in each file:
 
 `ack -cl {{foo}}`
+
+- List all valid types:
+
+`ack --help=types`
--- a/pages/common/ack.md
+++ b/pages/common/ack.md
@@ -6,9 +6,9 @@
 
 `ack {{foo}}`
 
-- Find files in a specific language:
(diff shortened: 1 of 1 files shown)


pages/common/ack.md
```diff
@@ -6,9 +6,9 @@
 
 `ack {{foo}}`
 
-- Find files in a specific language:
+- Find files scoped to a specific programming language:
```
agnivade: Pardon my ignorance, but what does "scoped" exactly mean here ? Will it only check files with ".rb" extension ? Or does it mean something else ?

negamorgan: Pretty much. From the docs: 

> The following is the list of filetypes supported by ack.  You can specify a file type with the --type=TYPE format, or the --TYPE format.  For example, both --type=perl and --perl work.

> Note that some extensions may appear in multiple types.  For example, .pod files are both Perl and Parrot.

> --[no]ruby         .rb .rhtml .rjs .rxml .erb .rake .spec; Rakefile; first line matches /^#!.*\bruby/

agnivade: Then I think the previous description was clearer. By "scope", I was thinking of "variable scoping" and was wondering how does this apply to files.

agnivade: Made the changes myself.

PR #2824, opened 2019-03-07
npm: remove unnecessary --save flag and simplify descriptions

KyleW:
Updates and simplifies the npm page.
Specifically:
* Removes unnecessary `--save` flag (added by default in npm > 5).
* Adds an example that installs a package wihtout a version since thats a simpler case.
* Moves the global install example down the page since it's a more complicated case
* Switch to the longversion of global option (`--global`) instead of `-g` to comply with guidelines
* Simplified description of adding a dependency
* Switched from the terms "dependency" and "module" to "package" where possible since package is more idiomatic in this context.
* Removed `npm list --global` example to stay under 8 example limit

### DIFF
--- a/pages/common/npm.md
+++ b/pages/common/npm.md
@@ -4,34 +4,34 @@
 > Manage Node.js projects and their module dependencies.
 > Homepage: <https://www.npmjs.com/>.
 
-- Download and install a module globally:
+- Interactively create a package.json file:
 
-`npm install -g {{module_name}}`
+`npm init`
 
-- Download all dependencies referenced in package.json:
+- Download all the packages listed as dependencies in package.json:
 
 `npm install`
 
-- Download a given dependency required for the application to run, and add it to the package.json:
+- Download a specific version of a package and add it to the list of dependencies in package.json:
 
-`npm install {{module_name}}@{{version}} --save`
+`npm install {{module_name}}@{{version}}`
 
-- Download a given dependency for development purposes, and add it to the package.json:
+- Download a package and add it to the list of dev dependencies in package.json:
 
-`npm install {{module_name}}@{{version}} --save-dev`
+`npm install {{module_name}} --save-dev`
 
-- Uninstall a module:
+- Download a package and install it globally:
+
+`npm install -g {{module_name}}`
+
+- Uninstall a package and remove it from the list of dependencies in package.json:
 
 `npm uninstall {{module_name}}`
 
-- List a tree of installed modules referenced in package.json:
+- Print a tree of installed, project-level packages:
 
 `npm list`
 
 - List top-level globally installed modules:
 
 `npm list -g --depth={{0}}`
-
-- Interactively create a package.json file:
-
-`npm init`
--- a/pages/common/npm.md
+++ b/pages/common/npm.md
@@ -28,7 +28,7 @@
 
 `npm uninstall {{module_name}}`
 
-- Print a tree of installed, project-level packages:
+- Print a tree of locally-installed dependencies:
 
 `npm list`
 


agnivade:
Thanks for the PR. Couple of general comments

- The `npm list --global` was a really helpful command. I would like it to remain. I think we can remove the `npm install {{module_name}}` example because we already have a `--save-dev` example with just the module_name. So the user can figure out that a simple install is also possible with just the module name.
- Let us bring back `-g`. The guidelines are there only when the short-form is not apparent. In this case, it is very clear that `g` stands for `global`.

pages/common/npm.md
```diff
@@ -4,34 +4,34 @@
...
-- Download all dependencies referenced in package.json:
+- Download all the packages listed as dependencies in package.json:
 
 `npm install`
 
-- Download a given dependency required for the application to run, and add it to the package.json:
+- Download a package and add it to the list of dependencies in package.json:
 
-`npm install {{module_name}}@{{version}} --save`
+`npm install {{module_name}}`
```
sbrl: Don't you need to add `--save` for this to actually add to `package.json`? I'm not sure that removing this is a good idea.

Did this behaviour change at some point - could you point me at some documentation for this?

pages/common/npm.md
```diff
@@ -4,34 +4,34 @@
...
+`npm install --global {{module_name}}`
 
-- List top-level globally installed modules:
+- Uninstall a package and remove it from the list of dependencies in package.json:
 
-`npm list -g --depth={{0}}`
+`npm uninstall {{module_name}}`
 
-- Interactively create a package.json file:
+- Print the dependency tree defined in package.json:
```
sbrl: This isn't technically correct, I don't think. The `list` subcommand lists the currently-installed dependency tree, and doesn't depend on the contents of `package.json`. By this I mean that you can have a package installed but not defined in package.json, and `npm list` will still show it to you.

pages/common/npm.md
KyleW: No. Saving is the default behavior starting in npm 5. Relevant section of official docs: https://docs.npmjs.com/cli/install.html

pages/common/npm.md
KyleW: Good catch. I'll rephrase.

KyleW:
> Thanks for the PR. Couple of general comments
> 
> * The `npm list --global` was a really helpful command. I would like it to remain. I think we can remove the `npm install {{module_name}}` example because we already have a `--save-dev` example with just the module_name. So the user can figure out that a simple install is also possible with just the module name.
> * Let us bring back `-g`. The guidelines are there only when the short-form is not apparent. In this case, it is very clear that `g` stands for `global`.

Ok, I'll switch `--global` back to`-g`.

And I can see the value of addling the `npm list -g` example back in. But I also would like to keep`npm install {{packageName}}` example. It really is the simplest thing you can do with npm. Maybe there's another example we can drop instead.

Do we need the `--save-dev` example? While folks can use it for a lot of things, my sense is that it primarily matters for package developers, not new npm users.

pages/common/npm.md
KyleW: Hmm. on second thought, I could use a suggestion here. That part of the description was in the prior version of the doc as well. I think the intent is to indicate that `npm list` only tells you about your project dependencies and not your global dependencies. 

Does anyone have a good idea on how to phrase that?

KyleW:
Revised based on comments. 

Note: there's still only a single commit in the pr because I squashed the changes before pushing. I think that's the right thing to do to keep the history clean, but let me know if there's a better convention.

agnivade:
> And I can see the value of addling the npm list -g example back in. But I also would like to keep npm install {{packageName}} example. It really is the simplest thing you can do with npm. Maybe there's another example we can drop instead.

Sure, if you want to keep it. I don't think we can remove anything else. `npm list -g` takes the least priority amongst others.

pages/common/npm.md
sbrl: I see. Thanks! That'll be a hard habit to break lol :P

pages/common/npm.md
sbrl: What about this?

```suggestion
- Print a tree of locally-installed dependencies:
```

sbrl:
No need to squash before pushing, as we've got a button to do that automagically when we merge the commit. In fact not doing so makes it easier to see what's changed as we go back and forth :D

pages/common/npm.md
KyleW: I like it! I'll make the change and repush.

KyleW:
I think I've incorporated all the feedback. Please take another look.

agnivade:
Did you not want to add npm install {{package}} and remove npm list -g ?

KyleW:
> Did you not want to add npm install {{package}} and remove npm list -g ?

I think this version which includes npm list -g and leaves out the install without a version is just fine. If the maintainers are comfortable, I say merge it!

agnivade:
Sounds good.

PR #2826, opened 2019-03-11
pages.zh/windows: add Chinese translation

starccy:


starccy:
@agnivade To be honest, I may want to ask the same question...

sbrl:
lol!

I'm sure that with the number of pages we have now in other languages, client authors can implement support. @pepa65, is this something that the bash client could support - especially with #2706 being recently closed?

pepa65:
I was planning to implement that once sufficient interest (read: available translated pages) is there.
But note that my client is no longer compliant under the new specs, because it uses the `tldr platform/page` syntax, and that's the way I like it. I don't want to be typing `tldr -p platform page`, a bit stubborn perhaps, but I am sure users can adapt..?

pepa65:
Are these translated pages going to be added to the zip archive??

sbrl:
They should be in the zip archive already I think, because of #2567, @pepa65 - though if you're not seeing them there might be a bug

Personally I like the `tldr platform/page` syntax best too. I hope your client keeps supporting it :D

agnivade:
@starccy - So you are spending non-trivial amount of time translating these pages without any direct result ? That is some serious dedication. Thanks !

pepa65:
I must have the wrong zip... :-1: What is the right URL? (I'm using http://tldr.sh/assets/tldr.zip which has an `index.json` that includes more languages, but no actual pages...)

sbrl:
The new [Client Spec](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md) should contain the right URL. If that doesn't work, could you open an issue please, @pepa65

pepa65:
The link in the spec just opens up de general repo. I think the build instructions for the zipfile on http://tldr.sh/assets should be made to include all pages* directories.

pxgamer:
@pepa65, it does when I download it. 🤔

mttbernardini:
@pepa65 I can reproduce what happens to @pxgamer

pxgamer:
Oh. Are you talking about adding an `index.json` to each directory? That makes sense.

pepa65:
My mistake, my client was only unpacking the pages directory, not the whole archive..!
I don't use the index.json anymore, but if someone does, it would be important for every pages* directory to have one (that, or one general one in the top level, but that will break old clients that use the index.json).

pxgamer:
Glad you worked it out 👍

I'd see it as being better for an `index.json` per pages directory. 🤔

pepa65:
I opened a new Issue for getting a index.json built for every language.

PR #2833, opened 2019-03-15
git grep: add page

EpocSquadron:

### DIFF
--- a/pages/common/git-grep.md
+++ b/pages/common/git-grep.md
@@ -16,6 +16,6 @@
 
 `git grep --recurse-submodules {{search_string}}`
 
-- Search for a string in another point in history:
+- Search for a string in at a specific point in history:
 
 `git grep {{search_string}} HEAD~2`
--- a/pages/common/git-grep.md
+++ b/pages/common/git-grep.md
@@ -17,7 +17,3 @@
 - Search for a string in another point in history:
 
 `git grep {{search_string}} HEAD~2`
-
-`git grep {{search_string}} some-branch-name`
-
-`git grep {{search_string}} {{git_ref}}`
--- a/pages/common/git-grep.md
+++ b/pages/common/git-grep.md
@@ -1,6 +1,6 @@
 # git-grep
 
-> Find strings inside files as they existed anywhere in repo history. Accepts a lot of the same flags as regular grep. Learn more here: https://git-scm.com/docs/git-grep
+> Find strings inside files as they existed anywhere in repo history. Accepts a lot of the same flags as regular grep. Learn more here: https://git-scm.com/docs/git-grep.
 
 - Search for a string in working directory:
 
--- a/pages/common/git-grep.md
+++ b/pages/common/git-grep.md
@@ -16,6 +16,6 @@
 
 `git grep --recurse-submodules {{search_string}}`
 
-- Search for a string in at a specific point in history:
+- Search for a string at a specific point in history:
 
 `git grep {{search_string}} {{HEAD~2}}`
--- a/pages/common/git-grep.md
+++ b/pages/common/git-grep.md
@@ -4,18 +4,18 @@
 > Accepts a lot of the same flags as regular `grep`.
 > Documentation: <https://git-scm.com/docs/git-grep>.
 
-- Search for a string in working directory:
+- Search for a string in tracked files:
 
 `git grep {{search_string}}`
 
-- Search for string in files matching a pattern in working directory:
+- Search for a string in files matching a pattern in tracked files:
 
 `git grep {{search_string}} -- {{file_glob_pattern}}`
 
-- Search for a string in working directory, including submodules:
+- Search for a string in tracked files, including submodules:
 
 `git grep --recurse-submodules {{search_string}}`
 
 - Search for a string in at a specific point in history:
 
-`git grep {{search_string}} HEAD~2`
+`git grep {{search_string}} {{HEAD~2}}`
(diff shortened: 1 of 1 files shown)


pages/common/git-grep.md
```diff
@@ -0,0 +1,19 @@
+# git-grep
+
+> Find strings inside files as they existed anywhere in repo history. Accepts a lot of the same flags as regular grep. Learn more here: https://git-scm.com/docs/git-grep.
```
sbrl: Perhaps this will read better?

```suggestion
> Find strings inside files anywhere in a repository's history.
> Accepts a lot of the same flags as regular `grep`.
> Documentation: https://git-scm.com/docs/git-grep.
```

pages/common/git-grep.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Search for a string in working directory:
```
sbrl: Perhaps this will read better?
```suggestion
- Search for a string in the current working directory:
```

pages/common/git-grep.md
```diff
@@ -0,0 +1,19 @@
...
+
+`git grep {{search_string}}`
+
+- Search for string in files matching a pattern in working directory:
```
sbrl: Perhaps this will read better?

```suggestion
- Search for string in files matching a pattern in the current working directory:
```

pages/common/git-grep.md
```diff
@@ -0,0 +1,19 @@
...
+
+`git grep {{search_string}} -- {{file_glob_pattern}}`
+
+- Search for a string in working directory, including submodules:
```
sbrl: Perhaps this will read better?

```suggestion
- Search for a string in the current working directory, including submodules:
```

pages/common/git-grep.md
```diff
@@ -0,0 +1,19 @@
...
+
+`git grep --recurse-submodules {{search_string}}`
+
+- Search for a string in another point in history:
```
sbrl: Perhaps this will read better?

```suggestion
- Search for a string in at a specific point in history:
```

pages/common/git-grep.md
pxgamer: @sbrl, does this need the angle brackets around the URL?

pages/common/git-grep.md
```diff
@@ -0,0 +1,19 @@
...
+
+`git grep {{search_string}} HEAD~2`
```
pxgamer:
```suggestion
`git grep {{search_string}} {{HEAD~2}}`
```

pages/common/git-grep.md
sbrl: Yes. I forgot :P

pages/common/git-grep.md
EpocSquadron: There is a slight distinction here that I didn't capture well, though I'm not sure its appropriate to describe in full in tldr. The command works on files that aren't ignored if its run against HEAD, otherwise it only works on files that are actually in that commit. The man page calls it the "work tree" in the former case.

How about if we changed it to "Search for a string in non-ignored files in the current directory:"?

pages/common/git-grep.md
EpocSquadron: Again, "current working directory" is misleading. Maybe just say "Search for a string in files matching a pattern:"

pages/common/git-grep.md
EpocSquadron: One more. "Search for a string in non-ignored files in the current directory and all submodules:"

pages/common/git-grep.md
EpocSquadron: @pxgamer Originally I wrote this as `{{ref_or_commit}}` but revised it to a specific example after reviewing some of the other git- subcommands in this repo, and noting the contributing guides' recommendation not to explain larger concepts. In particular, [here in the git-reset page](https://github.com/tldr-pages/tldr/blob/master/pages/common/git-reset.md) uses `HEAD~2` explicitely rather than explaining the concept of a ref.

I'm happy to accept the change, but wonder what your opinion on the problem of being explicit without constraining the user's idea of what can go there.

pages/common/git-grep.md
agnivade: We also need to mention subdirectories. Let us use the word "tracked" rather than "non-ignored". Better to use a positive than double-negative. How about - "Search for a string in all tracked files in the working directory and its subdirectories" ?

pages/common/git-grep.md
EpocSquadron: I haven't heard back so I'm going to accept it for now. It can always be changed later.

pxgamer: Sorry, not sure why I missed replying 🤔 That sounds good. 👍

pages/common/git-grep.md
```diff
@@ -0,0 +1,21 @@
...
+
+- Search for a string in files matching a pattern in tracked files:
+
+`git grep {{search_string}} -- {{file_glob_pattern}}`
+
+- Search for a string in tracked files, including submodules:
+
+`git grep --recurse-submodules {{search_string}}`
+
+- Search for a string in at a specific point in history:
```
sbrl: Perhaps this will read better?
```suggestion
- Search for a string at a specific point in history:
```

pxgamer: @sbrl, should this just be committed as it's such a minor change?

sbrl:
Thanks, @EpocSquadron :smiley_cat:

PR #2834, opened 2019-03-15
xar: add page

johan456789:

### DIFF
--- a/pages/osx/xar.md
+++ b/pages/osx/xar.md
@@ -1,0 +1,15 @@
+# xar
+
+> Manage .xar archives.
+
+- Create a xar archive of all files in a given directory:
+
+`xar -cf {{archive.xar}} {{path/to/directory}}`
+
+- List the contents of a given xar archive:
+
+`xar -tf {{archive.xar}}`
+
+- Extract the contents of a given xar archive to the current directory:
+
+`xar -xf {{archive.xar}}`
--- a/pages/osx/xar.md
+++ b/pages/osx/xar.md
@@ -1,4 +1,5 @@
 # xar
+> Homepage: <https://mackyle.github.io/xar/>.
 
 > Manage .xar archives.
 
--- a/pages/osx/xar.md
+++ b/pages/osx/xar.md
@@ -1,4 +1,5 @@
 # xar
+
 > Manage .xar archives.
 > Homepage: <https://mackyle.github.io/xar/>.
 
--- a/pages/osx/xar.md
+++ b/pages/osx/xar.md
@@ -1,13 +1,13 @@
 # xar
 
 > Manage .xar archives.
-> Homepage: <https://mackyle.github.io/xar/>.
+> Homepage: <https://linux.die.net/man/1/xar>.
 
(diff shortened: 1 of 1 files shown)


pages/osx/xar.md
```diff
@@ -0,0 +1,15 @@
+# xar
+
+> Manage .xar archives.
+
```
sbrl: Perhaps we could insert a line here that says this?

```markdown
> Homepage: <https://mackyle.github.io/xar/>.

File: pages/osx/xar.md [hunk 2]
```diff
@@ -0,0 +1,15 @@
...
+- Create a xar archive of all files in a given directory:
+
+`xar -cf {{archive.xar}} {{path/to/directory}}`
+
+- List the contents of a given xar archive:
+
+`xar -tf {{archive.xar}}`
```
agnivade: Let's use `--list` instead of `-t`

johan456789: I tried but `--list` doesn't work. There's also no such usage [here](https://linux.die.net/man/1/xar).

agnivade: This page https://mackyle.github.io/xar/xar.html mentions that `--list` is a synonym for `-t`. Is it a version related issue ? Or the doc is incorrect ?

pxgamer: It looks like it was originally in subversion (the old version) and then forked to GitHub and updated.

agnivade: Even the github page has the `--list` option - https://github.com/mackyle/xar/blob/master/xar/src/xar.1#L29

pxgamer: Yeah, the new versions do. The man page on Linux.die is the original older version.

johan456789:
```
$ which xar
/usr/bin/xar
```
Does this mean it's preinstalled on mac?

agnivade: Not necessarily. Any package is free to install a command at `/usr/bin`. Is there a brew version of xar that you can get the latest version of ?

johan456789:
```
$ brew search xar
==> Formulae
xar-mackyle                              xaric
$ brew info xar-mackyle
xar-mackyle: stable 1.6.1 (bottled)
eXtensible ARchiver
https://mackyle.github.io/xar/
Not installed
```
I just checked my friend's mac. It also has `xar`.  I'm guessing `xar` comes preinstalled on mac and the github version of `xar` is different from the one on mac.
Also, mac version:
```
$ xar --version
xar 1.8dev
```

agnivade: Ok, so the mac version is more recent than the brew version ? And that still doesn't have the --list option ?

johan456789: It seems so. Should I remove the GitHub link as homepage?

agnivade: Sure. Then put this https://linux.die.net/man/1/xar as the home page. And change `list` to `lis[t]`

johan456789: Sure, changed it

PR #2838, opened 2019-03-18
vpnc: minor grammatical fix

sbrl:
Proposing in a new PR because the original author was a first-time contributor, and I didn't want to bog them down in details.


sbrl:
Merging 'cause it's a trivial change.

PR #2844, opened 2019-03-21
Remove broken Go Client

frioux:
Always returns 404, even when other clients work.


frioux:
I don't want to sign a CLA to delete a line of a readme; feel free to do it another way :)

pxgamer:
Ah, I was thinking that would be the reason. :+1: Completely fair enough.

PR #2856, opened 2019-03-29
pg_dump: added -a and -s examples

andrik:

### DIFF
--- a/pages/common/pg_dump.md
+++ b/pages/common/pg_dump.md
@@ -20,7 +20,7 @@
 
 - Dump only database data into a SQL-script file:
 
-`pg_dump -a {{db_name}} > {{output_file.sql}}`
+`pg_dump -a {{db_name}} > {{path/to/output_file.sql}}`
 
 - Dump only schema (data definitions) into a SQL-script file:
 
--- a/pages/common/pg_dump.md
+++ b/pages/common/pg_dump.md
@@ -18,7 +18,7 @@
 
 `pg_dump -Fc {{db_name}} > {{output_file.dump}}`
 
-- Dump only database data into a SQL-script file:
+- Dump only database data into an SQL-script file:
 
 `pg_dump -a {{db_name}} > {{path/to/output_file.sql}}`
 
--- a/pages/common/pg_dump.md
+++ b/pages/common/pg_dump.md
@@ -17,3 +17,11 @@
 - Dump a database into a custom-format archive file:
 
 `pg_dump -Fc {{db_name}} > {{output_file.dump}}`
+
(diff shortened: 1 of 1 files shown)


pages/common/pg_dump.md
```diff
@@ -17,3 +17,11 @@
 - Dump a database into a custom-format archive file:
 
 `pg_dump -Fc {{db_name}} > {{output_file.dump}}`
+
+- Dump only database data into a SQL-script file:
```
sbrl: Perhaps this will read better?

```suggestion
- Dump only database data into an SQL-script file:
```

pages/common/pg_dump.md
```diff
@@ -17,3 +17,11 @@
...
+
+`pg_dump -a {{db_name}} > {{output_file.sql}}`
```
sbrl: Perhaps this could be change to this please?

```suggestion
`pg_dump -a {{db_name}} > {{path/to/output_file.sql}}`
```

pages/common/pg_dump.md
```diff
@@ -17,3 +17,11 @@
...
+
+- Dump only schema (data definitions) into a SQL-script file:
```
sbrl: Perhaps this will read better?
```suggestion
- Dump only schema (data definitions) into an SQL-script file:
```

pages/common/pg_dump.md
```diff
@@ -17,3 +17,11 @@
...
+
+`pg_dump -s {{db_name}} > {{output_file.sql}}`
```
sbrl: Perhaps we could change it to this?

```suggestion
`pg_dump -s {{db_name}} > {{path/to/output_file.sql}}`
```

pages/common/pg_dump.md
andrik: Hi @sbrl 

I think it's **a** because the next word starts with consonant.

pages/common/pg_dump.md
andrik: Hi @sbrl 

The same as before.

pages/common/pg_dump.md
sbrl: True, but there are exceptions to the rule. In this case, it sounds better if we use `an` instead of `a`. It's because SQL is an abbreviation, I believe.

andrik: English is not my main language :relaxed: 
So I know just the basic rules, I'll take the suggestions

pxgamer: To me, `an` doesn't sound right. :p

But that's cause I pronounce it Sequel, rather than S-Q-L. `An` is correct though. :+1:

PR #2859, opened 2019-03-30
client spec: fix page platform section

sbrl:
This updates the page platform section to match what we actually do, following a discussion with @agnivade via Gitter.

@agnivade, how to you mean that the `Page Resolution` section needs clarification?

### DIFF
--- a/CLIENT-SPECIFICATION.md
+++ b/CLIENT-SPECIFICATION.md
@@ -1,6 +1,6 @@
 # tldr-pages client specification
 
- - **Current Specification Version:** 1.0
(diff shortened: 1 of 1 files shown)


agnivade:
We can add a line saying that - It is possible that due to this page resolution logic, the client may show a page which does not belong to the host platform, because a page can reside in `common`, and not in the host platform. So the user must not assume that a shown command is always runnable in the host platform.

There is also this issue that now we do not know _exactly_ which platforms are supported with a command. That would need a page structure change to add metadata to specify this.

prfcttone2:
Great idea

PR #2862, opened 2019-04-02
Update building of index.json to be at the top level of the generated ZIP

### LINKED ISSUE #2829: Pages archive doesn't have index.json for other languages
In the pages archive on `http://tldr.sh/assets/tldr.zip` only the English language pages/ directory has an index.json, the other languages don't. This will make it harder for clients to implement displaying pages in other languages when they rely on or use index.json.

mebeim:
The `index.json` file contains information about *all* the pages and *all* the languages, and it should be located at the top level of the ZIP file which is generated when deploying to tldr-pages.github.io, not inside the `/pages/` folder (which is for English pages only).

I updated the build and deploy scripts to produce a copy of said file and add it at the top of the generated `tldr.zip`. I did not however disable the generation of `/pages/index.json`, since some clients rely on this file. 

Since this is kind of an architectural change, if this is merged I will notify each client by creating an issue on its repository, and also create a new issue in this repo to track which clients are up to date with the change. Once they all are or we think enough time has passed, we can then close said issue creating another PR to completely remove the duplicated `/pages/index.json` file from the ZIP.

PS: @sbrl the fact that the archive as well as the index can be retrieved from tldr-pages.github.io should probably be added into the client spec. When that happens we could also edit/remove the wiki page.

Fixes #2829

### DIFF
--- a/scripts/build-index.js
+++ b/scripts/build-index.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var glob = require('glob');
+const glob = require('glob');
 
 function parsePlatform(pagefile) {
   return pagefile.split(/\//)[1];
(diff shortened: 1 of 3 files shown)


scripts/build-index.js
```diff
@@ -11,48 +11,59 @@ function parsePagename(pagefile) {
...
+  return pagesFolder == 'pages' ? 'en' : pagesFolder.replace(/^pages\./, '');
 }
 
 function buildPagesIndex(files) {
-  var reducer = function (index, file) {
-    var os = parsePlatform(file);
-    var page = parsePagename(file);
-    var language = parseLanguage(file);
+  let reducer = function (index, file) {
+    let os       = parsePlatform(file),
```
agnivade: Can we just have single declarations per line like before ? It just seems odd to introduce this spacing convention when it is not used anywhere else in the code.

scripts/build-index.js
```diff
@@ -11,48 +11,59 @@ function parsePagename(pagefile) {
...
 
-  var obj = files.reduce(reducer, {});
+  let obj = files.reduce(reducer, {});
+
   return Object.keys(obj)
       .sort()
       .map(function(page) {
         return {
-          name: page,
+          name    : page,
```
agnivade: let's revert this too.

mebeim:
@agnivade updated, sorry if it's a force push I did not want to add a commit for such a trivial change.

agnivade:
Yes

PR #2870, opened 2019-04-05
pax: add page

codesoap:

`pax` is a core util and think this tldr page would be useful for some. I included only one example, where pipes / I/O redirection is used, because this feature of `pax` is not needed in the basic use cases.

### DIFF
--- a/pages/common/pax.md
+++ b/pages/common/pax.md
@@ -12,16 +12,16 @@
 
 - Create an archive from files:
 
-`pax -wf {{target.tar}} {{file1 file2 file3}}`
+`pax -wf {{target.tar}} {{path/to/file1}} {{path/to/file2}} {{path/to/file3}}`
 
 - Create an archive from files, using output redirection:
 
-`pax -w {{file1 file2 file3}} > {{target.tar}}`
+`pax -w {{path/to/file1}} {{path/to/file2}} {{path/to/file3}} > {{target.tar}}`
 
 - Extract an archive into the current directory:
 
 `pax -rf {{source.tar}}`
 
-- Copy to a directory, while keeping the original metadata; target/ must exist:
+- Copy to a directory, while keeping the original metadata; `target/` must exist:
 
-`pax -rw {{file1 directory1 directory2}} {{target/}}`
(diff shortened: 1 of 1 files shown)


pages/common/pax.md
```diff
@@ -0,0 +1,27 @@
+# pax
+
+> Archiving and copying utility.
```
sbrl: Perhaps we could add a line after this one for the homepage of this program, like this?

```
> Homepage: <INSERT_URL_HERE>.
```

pages/common/pax.md
```diff
@@ -0,0 +1,27 @@
...
+
+`pax -f {{archive.tar}}`
+
+- List the contents of a gzipped archive:
+
+`pax -zf {{archive.tar.gz}}`
+
+- Create an archive from files:
+
+`pax -wf {{target.tar}} {{file1 file2 file3}}`
```
sbrl: Perhaps each file could be enclosed in its own set of `{{` and `}}`, and be prepended with `path/to/` please? The same goes for other examples here.

pages/common/pax.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Create an archive from files, using output redirection:
+
+`pax -w {{file1 file2 file3}} > {{target.tar}}`
+
+- Extract an archive into the current directory:
+
+`pax -rf {{source.tar}}`
+
+- Copy to a directory, while keeping the original metadata; target/ must exist:
```
sbrl: Maybe `target/` could be enclosed in backticks here?

codesoap: Sounds reasonable, I've committed this change.

pages/common/pax.md
codesoap: I had oriented myself on `tar.md`, but sure I can do that. The changes are pushed now.

pages/common/pax.md
codesoap: I'm afraid I couldn't really find a homepage. `pax` is a standard core utility, that is defined by the POSIX.1-2001 standard and there are different implementations out there. Should I link to the [entry in the POSIX standard](http://pubs.opengroup.org/onlinepubs/009695399/utilities/pax.html)?

agnivade: No need if it's a core utility.

PR #2872, opened 2019-04-08
tshark: add capture filter

abitrolly:

### DIFF
--- a/pages/linux/tshark.md
+++ b/pages/linux/tshark.md
@@ -6,11 +6,11 @@
 
 `tshark`
 
-- Catch only packets matching a specific `capture filter`:
+- Only capture packets matching a specific capture filter:
 
 `tshark -f '{{udp port 53}}'`
 
-- Show only packets matching a specific `output filter`:
+- Only show packets matching a specific output filter:
 
 `tshark -Y '{{http.request.method == "GET"}}'`
 
--- a/pages/linux/tshark.md
+++ b/pages/linux/tshark.md
@@ -6,11 +6,11 @@
 
 `tshark`
 
-- Capture only specific packets:
+- Catch only packets matching a specific `capture filter`:
 
 `tshark -f '{{udp port 53}}'`
 
-- Only output captured packets matching a specific wireshark filter:
+- Show only packets matching a specific `output filter`:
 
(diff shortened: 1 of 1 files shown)


abitrolly:
It is also possible to remove `tshark` as it is kind of obvious.

pages/linux/tshark.md
```diff
@@ -6,6 +6,10 @@
 
 `tshark`
 
+- Capture only specific packets:
+
+`tshark -f 'udp port 53'`
```
pxgamer: Could you update this to token syntax?
```suggestion
`tshark -f '{{udp port 53}}'`
```

pxgamer:
Regarding https://github.com/tldr-pages/tldr/pull/2872#issuecomment-480689593, I think it's fine to keep this. :+1: We usually do for most pages.

pxgamer:
Ping @mebeim, @sbrl, @schneiderl 🙂

pages/linux/tshark.md
```diff
@@ -6,6 +6,10 @@
 
 `tshark`
 
+- Capture only specific packets:
```
schneiderl: By taking a look at the example I assume that with "specific" you mean packages going through port 53. 

IMHO it makes sense to have this clearly stated. 
The word "Specific" here does not explain much about what is the filter.

mebeim: Perhaps something like:

> Capture only packet matching a specific wireshark filter:

would read better?

agnivade: That is nearly the same as the description after. I would prefer to clearly distinguish between -f and -Y.

mebeim: @agnivade welp, that's a good point, I agree. I got confused the first time I read this PR because I thought the two examples were doing the exact same thing. The result is the same, but what the program does is very different.

abitrolly: I've added more explicit mention of `capture filter` and `output filter`. Hopefully that wlll clear the first time confusion.

schneiderl:
Sorry for the time it took for me to approve @abitrolly. 

Thank you for the PR.

abitrolly:
No problem. Now I will remember it. :D

PR #2876, opened 2019-04-08
COMMUNITY-ROLES.md: tweak phrasing

waldyrious:
Just small wording changes to make the progression steps' summaries more explicit and link better with each other (and with the contents of MAINTAINERS.md).

In particular, organization members are no longer conflated with the term "maintainer", which is more of an umbrella term covering all of the roles described here.

### DIFF
--- a/COMMUNITY-ROLES.md
+++ b/COMMUNITY-ROLES.md
@@ -32,7 +32,7 @@ exceptions can always be considered, via open community discussion.)
   This means they will be able to push commits to that repository,
   as well as merge PRs, label and close issues, among other things.
 
-- **Repo collaborators who show interest in maintenance work should be added as org members.**
+- **Repository collaborators who regularly perform maintenance tasks should be added as organization members.**
   (Maintenance work means facilitating contributions by other people,
   which in this project typically consists in reviewing and/or merging PRs.)
   Specifically: once a repository collaborator has _merged at least 10 PRs_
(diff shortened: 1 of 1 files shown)


COMMUNITY-ROLES.md
```diff
@@ -32,9 +32,9 @@ exceptions can always be considered, via open community discussion.)
   This means they will be able to push commits to that repository,
   as well as merge PRs, label and close issues, among other things.
 
-- **Collaborators who perform maintenance tasks should be made org members.**
-  (Maintenance work is understood as facilitating contributions by other people,
-  which in this project consists primarily of reviewing and/or merging PRs.)
+- **Repo collaborators who show interest in maintenance work should be added as org members.**
```
agnivade: Isn't "showing interest" more vague ? I would think performing maintenance tasks is clear and succinct.

waldyrious: I see. My thinking was that "perform" sounds more like a one-off thing, while "show interest" implies a sustained activity. Perhaps we could instead change it to "who _regularly_ perform maintenance tasks"?

waldyrious:
> I'd suggest using "Repository" instead of "Repo" and "Organization" instead of "Org".

I thought about that, but didn't want to make those lines too long. I suppose using jargon can be worse, though. Since you guys seem to agree, per the comment and reactions above, I'll make the change.

COMMUNITY-ROLES.md
agnivade: Yeah, we can do that.

agnivade:
I think org and repo are fine. But in general, we always tend to use expanded form of a word for extra clarity. So in that sense, it is good to be explicit.

waldyrious:
Made the changes discussed above as a new commit, to be squashed upon merge.

COMMUNITY-ROLES.md
mebeim: I agree, that sounds way better.

PR #2877, opened 2019-04-09
Add pacman line for installing tldr-python-client on arch

smt923:
Adds a line for installing the python client onto arch, mirroring my [PR to the github.io site](https://github.com/tldr-pages/tldr-pages.github.io/pull/33)


waldyrious:
Should we add "(on Arch Linux)"? The presence of `pip` is a reasonable assumption (especially for someone looking to install a Python-based tool), but `pacman` isn't as generally available.

PR #2880, opened 2019-04-09
refactor: fix newlines removing carriage returns

mebeim:
Fixed pages where Windows style (CRLF) newlines were used instead of Unix style (LF).


mebeim:
@pxgamer Hmmm it should already be handled by [`.editorconfig`](https://github.com/tldr-pages/tldr/blob/master/.editorconfig), the person who created these pages probably used an editor which did not honor `.editorconfig`, so probably adding a `.gitattributes` file with `* text=auto` is a good idea. Well, unless for some strange reason (like chinese characters) `git` decides to use CRLF, in that case `*.md text eol=lf` would do. I could add it to this PR.

pxgamer:
Yeah, most editors don't support EditorConfig by default afaik. Or at least, the editors I've used.

Regarding `*.md text eol=lf`, shouldn't `* text=auto` standardise line endings to `LF` anyway?

mebeim:
@pxgamer yes it should, on files that detects as text. If a file is not detected as text then the eol is not changed, so I was wondering if lots of unicode chars are correctly recognized... the answer is probably yes.

Issue #2874, opened 2019-04-08
Add member: @mebiem


sbrl:
Hi, @mebeim! After joining as a collaborator in the repository, you have been regularly performing maintenance tasks. Thank you for that! According to [COMMUNITY-ROLES.md](https://github.com/tldr-pages/tldr/blob/master/COMMUNITY-ROLES.md), you've now met the thresholds to be effectively considered an active maintainer of the project ([LINKS TO THE RELEVANT PRs]). To publicly acknowledge that fact, we'll add you to the tldr-pages organization.

If you accept the invitation, we ask you to make your membership public, and (in case you don't already) start hanging out in our Gitter chat room. You'll now be one of the public faces of the tldr-pages project. Welcome aboard!

mebeim:
I'd sure love to be a member, thank you :smile: Looking forward to keep contributing and expanding this awesome project! :tada:

agnivade:
Thanks for all your work @mebeim. @sbrl feel free to send out the invitation if you haven't already.

waldyrious:
I know this is a chore, but for future reference it would be nice to add the links to some PRs in the opening comment above, instead of the placeholder :innocent:

sbrl:
Invite sent! And yeah, I probably should @waldyrious. Is there a way to filter on who merged a PR? I couldn't find one

mebeim:
@sbrl there is not, unfortunately :disappointed:. The only way is using GitHub APIs filtering manually, like [this SO answer](https://stackoverflow.com/a/29768423/3889449) suggests.

mebeim:
Accepted the invite :tada:

waldyrious:
Reopening in light of #2881 -- we still need to update MAINTAINERS.md with the new information. @mebeim, wanna do the honors? :)

mebeim:
@waldyrious sure :+1:

pxgamer:
Congrats!

PR #2881, opened 2019-04-09
COMMUNITY-ROLES: streamline the How-to section

waldyrious:
Looks like every time we make role changes new wrinkles are found in the process description :)

In this case, #2874 was closed with no mention of MAINTAINERS.md needing an update ~~(it was actually done in #2746, but without a backlink to the issue)~~. Actually, my bad -- the update was not done after all, that was the older change from when he became a collaborator. More evidence that the process was missing steps :)

While on it, I also made some formatting changes and assorted adjustments to streamline the process. Here is the full list of changes:
- Use preformatted blocks for the message templates
  (instead of quote blocks, which required escaping of markdown formatting,
  yet still didn't prevent "smart" links from being rendered by github).
- Use a numbered list to make the role change processes easier to follow.
- Emphasize that the messages are templates, which should be edited if needed.
- Change 1 week to 2 weeks of when removing inactive owners.
  (to allow for periods of unavailability, illness, personal reasons, etc.)
- Correct references to "below" to point to MAINTAINERS.md instead.
- Add the step to update MAINTAINERS.md to all role change checklists.

### DIFF
--- a/COMMUNITY-ROLES.md
+++ b/COMMUNITY-ROLES.md
@@ -115,10 +115,10 @@ using one of the template messages below as a base.
    go to https://github.com/tldr-pages/tldr/settings/collaboration
    and add them to the repository as collaborator with write permissions.
 
-3. Afterwards, add their name to the "Current repository collaborators"
+3. Open a PR adding their name to the "Current repository collaborators" section
    in [MAINTAINERS.md](MAINTAINERS.md).
-
-4. Once the steps above are completed, close the issue.
+   Make sure to include "Closes #<issue number>" in the PR description.
+   The issue will then be automatically closed once the PR is merged.
 
 ### For adding new organization members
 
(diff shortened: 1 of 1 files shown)


COMMUNITY-ROLES.md
```diff
@@ -89,108 +89,129 @@ using one of the template messages below as a base.
 
 ### For adding new collaborators
```
mebeim: While we're at it, I'll suggest changing these headings removing the leading "For". "Adding new collaborators" sounds way better as a heading IMHO.

```suggestion
 ### Adding new collaborators
```

COMMUNITY-ROLES.md
```diff
@@ -89,108 +89,129 @@ using one of the template messages below as a base.
...
+   — you're welcome to contribute to the project in whatever ways suit your liking.
+
+   If you do decide to start performing maintenance tasks, though,
+   we only ask you to get familiar with the
+   [maintainer's guide](https://github.com/tldr-pages/tldr/blob/master/contributing-guides/maintainers-guide.md).
+
+   Thanks for all your work so far!
+   ```
+
+2. Once they acknowledge the message,
```
mebeim:
```suggestion
2. Once they acknowledge the message, if they accept,
```

I'd specify that one should actually accept the proposed role change (same goes for the other sections, except the last one probably).

pxgamer: Seems logical 😋

COMMUNITY-ROLES.md
waldyrious: Sounds good, will change.

COMMUNITY-ROLES.md
waldyrious: Yeah, makes sense :)

waldyrious:
Thanks for the suggestions @mebeim! I've applied them, please take another look when you can.

waldyrious:
@mebeim I just realized that the template messages leave no room for the recipient to accept or decline the proposal :smile:

Oh well, another update is in order...

PR #2893, opened 2019-04-11
trawl: add page

Aracki:
I have added [trawl](https://github.com/robphoenix/trawl) - The package that does pretty printing network interface information to the console, much like `ifconfig/ipconfig/ip/ifdata`.

### DIFF
--- a/pages/common/trawl.md
+++ b/pages/common/trawl.md
@@ -1,0 +1,19 @@
+# trawl
+
+> Prints out network interface information to the console, much like ifconfig/ipconfig/ip/ifdata.
+
+- Show column names:
+
+`trawl -n`
+
+- Filter interface names using a case insensitive regular expression:
+
+`trawl -f wi`
+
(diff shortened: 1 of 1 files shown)


pages/common/trawl.md
```diff
@@ -0,0 +1,19 @@
+# trawl
+
+> Prints out network interface information to the console, much like ifconfig/ipconfig/ip/ifdata.
+
```
sbrl: Could we add a line here for the homepage?

```markdown
> Homepage: <https://github.com/robphoenix/trawl>.

Aracki: I didn't see that in other pages; but LGTM.

pxgamer: It's relatively new, so most pages don't have them yet. :+1:

Aracki: I agree with injecting links into the markdown!

PR #2896, opened 2019-04-11
multiple pages: add homepage

schneiderl:


mebeim:
Hey @schneiderl thanks for the PR! 

I see you are using the readthedocs.io subdomain for [`borg`](https://github.com/borgbackup/borg), [`bedtools`](https://github.com/arq5x/bedtools2) and [`cabal`](https://github.com/haskell/cabal). I was wondering if it would make more sense to use the official GitHub repositories instead (which I just linked). What do you think? 

cc @sbrl

schneiderl:
Thank you for the comment @mebeim 
It does make sense. Specially because we are sure the official GitHub will always be up to date.
I will just wait for @sbrl comments and commit the changes if he agrees with it.

sbrl:
Personally, I'd recommend linking to the specific CLI documentation if it's available - whether that be on GH or RTD.

mebeim:
Well, either way the GH repo can be reached from RTD and vice-versa, so at this point both versions are ok to me. I was just wondering if it was already decided which one to prefer.

PR #2897, opened 2019-04-11
Use Chinese punctuation in Chinese pages

telnetning:


sbrl:
Hey, @telnetning! I'm unsure what this PR does. Could you explain please? The PR title doesn't explain it very well.

schneiderl:
This PR seems to be adjusting the punctuation marks to the Chinese set of punctuation marks as it makes no sense to use European punctuation on texts written in Chinese. More information about that can be found [here](https://en.wikipedia.org/wiki/Chinese_punctuation).

Am I right @telnetning ?
In my opinion this PR changes do make a lot of sense.

pages.zh/linux/apt-key.md
```diff
@@ -1,23 +1,23 @@
...
-- 从可信密钥库中移除一个密钥:
+- 从可信密钥库中移除一个密钥：
 
 `apt-key del {{密钥id}}`
 
-- 向可信密钥库中添加一个远程密钥:
+- 向可信密钥库中添加一个远程密钥：
 
-`wget -qO - {{https://host.tld/filename.key}} | apt-key add -`
+`wget -qO - {{https：//host.tld/filename.key}} | apt-key add -`
```
mebeim:
```suggestion
`wget -qO - {{https://host.tld/filename.key}} | apt-key add -`
```

pages.zh/linux/arp-scan.md
```diff
@@ -1,19 +1,19 @@
...
-- 扫描自定义范围内的IP网络:
+- 扫描自定义范围内的IP网络：
 
 `arp-scan {{127.0.0.0}}-{{127.0.0.31}}`
 
-- 扫描带有自定义子网掩码的IP网络:
+- 扫描带有自定义子网掩码的IP网络：
 
-`arp-scan {{10.0.0.0}}:{{255.255.255.0}}`
+`arp-scan {{10.0.0.0}}：{{255.255.255.0}}`
```
mebeim:
```suggestion
`arp-scan {{10.0.0.0}}:{{255.255.255.0}}`
```

telnetning:
@sbrl @schneiderl  yes, schneiderl is right.
And thanks @mebeim , the punctuation inside commands really shouldn't be changed.

I make this change according to  _Guides For Free Software Localization Working_(自由软件中文化工作指南, sorry for it's only available in chinese).  

This document suggests that the punctuations(except `...`/`--`/`()`) should be translated into Chinese on texts.

mebeim:
On second thought I would like to hear some other opinions before actually deciding whether this should be merged or not. Excuse my rushed approval, but since such changes wouls obviously mess with the current page formatting rules, and possibly the clients, it should be discussed.

This would basically break the current page format rules. I don't necessarily see that as a problem, but we need to consider whether it is actually appropriate to do this for the sake of correctness or not. After all we're talking about almost identical punctuation marks, but obviously very different in terms of the actual characters used.

Clients still basically have no translation support, so it's not really a problem right now on that side, but this would surely impact clients' ability of correctly parsing pages since they'd have to parse these pages differently.

Furthermore, allowing slightly different formats for different languages is not simple to manage.

Looking to hear from you guys @agnivade @pxgamer @sbrl @waldyrious.

sbrl:
I'd say that we should always ensure that the page format is followed, regardless of language. I think in a recent PR we chose to use plain-ASCII over smart quotes.

telnetning:
@sbrl just as you said, translation criteria is important. Indeed, in my opinion, the first thing needed to do is setting a page format.    

Page format may be a translation guide like _Guides For Free Software Localization Working_ or others. But anyway, a guide doc is needed.    

I adjusted some punctuation marks , such as`.` to `。`, as all translation guides I have seen do like this. And I didn't adjusted `...`/`--`/`()` as the guide above suggested that.  

In addition, I noticed `pages/osx` also contains many pages. So my patch may should not be merged otherwise there will be two formats.   

However, I think the format issue should be discussed again and a guide document is needed for all pages in Chinese.

agnivade:
This was my first concern in the original translation issue - https://github.com/tldr-pages/tldr/issues/2339#issuecomment-422293597. It is risky to change punctuation characters as some clients may depend on it. I am fine with changing stuff within sentences or tokens. But any trailing characters like `:` or `.` should be left alone.

Exactly what @mebeim said

> A compromise could be to allow commas and other punctuation inside phrases, but not for critical parts of the page like ending periods or ending colons.

sbrl:
The other thing is that the linter also depends on these characters.

mebeim:
So, @telnetning, could you fix this by keeping the original english punctuation characters for critical parts of the pages (like final description dot and final example description colon)? 

I get your point, and also anything in the description is free to be changed (brackets, accents, whatever, not only commas), but those characters are more than needed both for linting purposes and for client compatibility. We could work on allowing those in the future, but it is not really feasible right now.

telnetning:
OK, I have understood and agreed it's not the time to bring Chinese punctuation characters into translated text. After all, Usability is more important than aesthetics feeling.   
So, I will close this issue, and we can re-discuss this in the future. 
Thanks you all again.

PR #2899, opened 2019-04-11
MAINTAINERS.MD: Add @schneiderl as contributor

schneiderl:
Just added myself as collaborator on Maintainers.md

Closes #2899

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -26,6 +26,7 @@ If you are an owner of the organization, you should be able to see an automated 
    [7 September 2018](https://github.com/tldr-pages/tldr/issues/2306) — present
  - Lucas Schneider ([@schneiderl](https://github.com/schneiderl)):
    [11 April 2019](https://github.com/tldr-pages/tldr/issues/2898) — present
+   
 
 ## Current organization members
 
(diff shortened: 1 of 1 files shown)


mebeim:
I'd rather say "Add @schneiderl as contributor" since it's an addition rather than an a role change. Also you lost a newline there, let's keep the formatting consistent, shall we? Oh, and by the way: welcome aboard! :smile: :tada:

schneiderl:
@mebeim made the necessary changes :)

waldyrious:
@schneiderl please keep the double lines before level-2 headings :blush:

mebeim:
@waldyrious they are kept. GitHub displays the diff in a weird way, but there still are two lines before the next heading.

By the way @schneiderl with my previous comment I was talking more about the commit message than the title of the PR :sweat_smile:, that's what is going to be in the history. I'm approving anyway since it can be changed when merging so that's not really a big problem.

waldyrious:
> @waldyrious they are kept. GitHub displays the diff in a weird way, but there still are two lines before the next heading.

Right, my bad. It's what I get for reviewing PRs late at night! 😄

PR #2905, opened 2019-04-12
smartctl: Fix link to Wikipedia

mueller-ma:
There is a redirection on Wikipedia from `S.M.A.R.T` to `S.M.A.R.T.`, but it's nicer to get to `S.M.A.R.T.` directly.

### DIFF
--- a/pages/common/smartctl.md
+++ b/pages/common/smartctl.md
@@ -1,7 +1,7 @@
 # smartctl
 
 > View a disk's SMART data and other information.
-> See https://en.wikipedia.org/wiki/S.M.A.R.T for more information.
+> See [en.wikipedia.org/wiki/S.M.A.R.T.](https://en.wikipedia.org/wiki/S.M.A.R.T.) for more information.
 
 - View SMART health summary:
 
--- a/pages/common/smartctl.md
+++ b/pages/common/smartctl.md
@@ -1,7 +1,7 @@
 # smartctl
 
 > View a disk's SMART data and other information.
-> See [en.wikipedia.org/wiki/S.M.A.R.T.](https://en.wikipedia.org/wiki/S.M.A.R.T.) for more information.
+> More information: <https://en.wikipedia.org/wiki/S.M.A.R.T.>.
 
 - View SMART health summary:
 


pages/common/smartctl.md
```diff
@@ -1,7 +1,7 @@
 # smartctl
 
 > View a disk's SMART data and other information.
-> See https://en.wikipedia.org/wiki/S.M.A.R.T for more information.
+> See [en.wikipedia.org/wiki/S.M.A.R.T.](https://en.wikipedia.org/wiki/S.M.A.R.T.) for more information.
```
sbrl: Hey! Could we perhaps tweak this to look like this please?

```suggestion
> More information: <https://en.wikipedia.org/wiki/S.M.A.R.T.>.
```

schneiderl: Hi @mueller-ma, 

Thank you for the contribution 👍 
I believe we should not be using markdown because the text from this command description might be rendered in clients that do not support that. 
I think the direct link (without the square brackets and parenthesis) should be used in this case.

schneiderl:
```suggestion
> See https://en.wikipedia.org/wiki/S.M.A.R.T. for more information.
```

mueller-ma: Thanks for the suggestion. I committed it.

sbrl: Actually, most markdown is ok - the spec requires clients implement full CommonMark support.

It's just that the regular link syntax isn't well supported (despite the fact that there is now [a standard](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) for displaying it in a terminal).

mueller-ma: I personally like having "https://" hidden in weblinks. I suggest you add a sentence about links to your Style Guide.

Off topic: You can also add a sentence about mail addresses and urls in example must be `@example.com`, because this is exactly why example.com exists.

sbrl: Great ideas!

mueller-ma: I can do so, if you tell me how you want to have hyperlinks.

sbrl: I think we'd prefer the prefix to be shown in hyperlinks, and use the `<....>` syntax if possible.

PR #2911, opened 2019-04-14
COMMUNITY-ROLES: fix usage of angle brackets

waldyrious:
Without wrapping in code spans using backticks, angle brackets are interpreted as HTML tags, and stripped out from the rendered version.

Hopefully this will be the last update in the recent series of changes to this page. (Maybe I should stop re-reading the page once the PRs are merged :P)


mebeim:
O wow, who would've noticed it, pretty funny. Thanks again @waldyrious :)

PR #2912, opened 2019-04-15
7z: add space after option -p and -o

andrik:

### DIFF
--- a/pages/common/7z.md
+++ b/pages/common/7z.md
@@ -17,7 +17,7 @@
 
 - Extract an archive with user-defined output path:
 
-`7z x {{archived.7z}} -o {{path/to/output}}`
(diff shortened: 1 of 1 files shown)


andrik:
It's not mandatory using space, but as suggested by @schneiderl on the pt_br version I think it looks better using it.

waldyrious:
@andrik is there a long-form of the `-o` option that we could use? Perhaps that would accept a space or an equal sign, making the command more readable. (Though that's not a big deal, it can be confusing to have a space after `-p` and no space after `-o`).

sbrl:
I don't think there is.

Also, the official 7z manual page does not include a space, I've realised. I think I would vote for reverting this change.

andrik:
Yeah, it's better to follow the manual. I'll change it.

Issue #2056, opened 2018-04-05
page request: ldapsearch


grafuls:
ldapsearch command can be pretty complex, would be nice to see this here.
https://linux.die.net/man/1/ldapsearch

sbrl:
Thanks for requesting this page, @grafuls! Do you want to have a go yourself?

Pages are best created by people who understand the command they are about.

If you do, take a quick read of [CONTRIBUTING.md](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md), which guides you through the process of creating a new page here. Then, just simply send a pull request to this repository - and if you need any assistance, just ask :-)

lamar-frankie:
I'd like to claim this issue. If @grafuls hasn't started working on it.

grafuls:
@lamar-frankie I haven't got the time to get to it, all yours.

lbonanomi:
Is this issue up-for-grabs?

sbrl:
It looks like it, @lbonanomi. Go ahead :-)

lbonanomi:
LDAP is mature enough to have spawned several "standard" clients across different platforms, so I opened a draft PR (https://github.com/tldr-pages/tldr/pull/2915) for the generic features that appear across both Linux and SunOS clients. 

The major difference seems to be support for timelimit, sizelimit and extensions; Linux has these, SunOS' stock ldap client doesn't. Would it be appropriate to maintain Linux and SunOS pages independently to show Linux users all their options without teasing SunOS users?

schneiderl:
Hi @lbonanomi 
I think that the way to go: having two pages.
1 under common/linux and 1 under common/sunos.

Would you send a PR with the proposed changes?

lbonanomi:
https://github.com/tldr-pages/tldr/pull/2915 has been upgraded to a regular PR.

I'd appreciate feedback about any examples maintainers think could be dropped, I'd like to add a line for result-paging on the Linux client but don't want to violate the 8 examples limit without a formal go-ahead.

sbrl:
Thanks, @lbonanomi! I'm not sure which example we could drop.

lbonanomi:
@sbrl I hate to nag but is there anything left to be done on this issue before it gets closed?

agnivade:
Nope. In future, you can use the "Fixes #issue_no" string in a PR, which will automatically close an issue when the PR is merged.

lbonanomi:
Thank you @agnivade!

PR #2915, opened 2019-04-15
ldapsearch: add page

lbonanomi:
Query LDAP server for all values of matching items
Query LDAP server for certain values of matching items 
Query LDAP server with AND filter
Query LDAP server with FOR filter
Query LDAP server with AND NOT filter


pages/linux/ldapsearch.md
```diff
@@ -0,0 +1,35 @@
+# ldapsearch
+
+> CLI utility for querying an LDAP directory.
+
+- Query LDAP for all items that match filter "{{memberOf}}={{group1}}" and return {{displayName}}:
```
sbrl: Perhaps this would read better?

```suggestion
- Query an LDAP server for all items that are a member of the given group and return the display name:
```

pages/linux/ldapsearch.md
```diff
@@ -0,0 +1,35 @@
...
+
+`ldapsearch -D '{{admin_DN}}' -w '{{password}}' -h {{ldap_host}} -b {{base_ou}} '{{memberOf}}={{group1}}' {{displayName}}`
+
+- Query LDAP with a no-newline password file and return {{displayName}}:
```
sbrl: Perhaps this will read better?
```suggestion
- Query an LDAP server with a no-newline password file and return the display name:
```

pages/linux/ldapsearch.md
```diff
@@ -0,0 +1,35 @@
...
+
+`ldapsearch -D '{{admin_DN}}' -y '{{password_file}}' -h {{ldap_host}} -b {{base_ou}} '{{memberOf}}={{group1}}' {{displayName}}`
+
+- Query LDAP for up-to 5 items that match filter "{{memberOf}}={{group1}}" and return {{displayName}}:
```
sbrl: Perhaps this will read better?
```suggestion
- Return the first 5 items that match the given filter:
```

pages/linux/ldapsearch.md
```diff
@@ -0,0 +1,35 @@
...
+
+`ldapsearch -D '{{admin_DN}}' -w '{{password}}' -h {{ldap_host}} -b {{base_ou}} '{{memberOf}}={{group1}}' -z 5 {{displayName}}`
+
+- Query LDAP for up-to 7 seconds for all items that match filter "{{memberOf}}={{group1}}" and return {{displayName}}:
```
sbrl: Perhaps this will read better?
```suggestion
- Wait up to 7 seconds for a response:
```

pages/linux/ldapsearch.md
```diff
@@ -0,0 +1,35 @@
...
+
+`ldapsearch -D '{{admin_DN}}' -w '{{password}}' -h {{ldap_host}} -b {{base_ou}} '{{memberOf}}={{group1}}' -l 7 {{displayName}}`
+
+- Query LDAP for all items that DO NOT match filter "{{memberOf}}={{group1}}" and return {{displayName}}:
```
sbrl: Perhaps this will read better?
```suggestion
- Invert the filter:
```

pages/linux/ldapsearch.md
```diff
@@ -0,0 +1,35 @@
...
+
+`ldapsearch -D '{{admin_DN}}' -w '{{password}}' -h {{ldap_host}} -b {{base_ou}} '(!({{memberOf}}={{group1}}))' {{displayName}}`
+
+- Query LDAP for all items that match filter "memberOf=group1 AND memberOf=group2 AND memberOf=group3" and return {{displayName}} value for each item:
```
sbrl: Perhaps this will read better?
```suggestion
- Return all items that are part of multiple groups, returning the display name for each item:
```

pages/linux/ldapsearch.md
```diff
@@ -0,0 +1,35 @@
...
+
+`ldapsearch -D '{{admin_DN}}' -w '{{password}}' -h {{ldap_host}} '(&(memberOf=group1)(memberOf=group2)(memberOf=group3))' "{{displayName}}"`
+
+- Query LDAP for all items that match filter "memberOf=group1 OR memberOf=group2 OR memberOf=group3" and return {{displayName}} value for each item:
```
sbrl: Perhaps this will read better?
```suggestion
- Return all items that are members of at least 1 of the specified groups:
```

pages/linux/ldapsearch.md
```diff
@@ -0,0 +1,35 @@
...
+
+`ldapsearch -D '{{admin_DN}}' -w '{{password}}' -h {{ldap_host}} '(|(memberOf=group1)(memberOf=group1)(memberOf=group3))' {{displayName}}`
+
+- Query LDAP for all items that match filter "memberOf=group1 AND memberOf=group2 AND NOT memberOf=group3" and return {{displayName}} value for each item:
```
sbrl:
```suggestion
- Combine multiple boolean logic filters:
```

pages/linux/ldapsearch.md
lbonanomi: Would "return the object's `display name` value", fit? I feel like the user should be nudged to not think of object values as universal.

pages/linux/ldapsearch.md
lbonanomi: Placeholder.

Changes here will take a cue from line 5's changes.

lbonanomi:
WRT the page duplication: Does the plan to split Linux and SunOS versions of ldapsearch https://github.com/tldr-pages/tldr/issues/2056#issuecomment-483267812 still stand, or should the OS-specific switches get removed and this page be relocated to common?

schneiderl:
@lbonanomi if the plan is to have different examples SunOS and Linux we should have multiple pages. If it got decided to drop those SunOS specific examples there is no need to have that duplicated and it should be relocated.

lbonanomi:
@schneiderl understood; I'm standing-by for a maintainer's decision on splitting between OSes.

pages/linux/ldapsearch.md
lbonanomi: @sbrl my read of LDAP docs suggests that results aren't sorted by default, I committed your change but I'm worried that this could be misleading.

pages/linux/ldapsearch.md
sbrl: Sounds good to me!

pages/linux/ldapsearch.md
sbrl: I see. What about this then?
```suggestion
- Return 5 items that match the given filter:
```

sbrl:
If the pages are identical, then `common` is the correct place I think now according to the new conventions.

If the pages are different, then they should stay where they are.

I've been leaving comments on the `linux` copy of the page since they looks similar, so in theory we can update the `sunos` page to match with little difficulty once we're done with the `linux` copy

pages/linux/ldapsearch.md
lbonanomi: LGTM, thank you!

pages/linux/ldapsearch.md
lbonanomi: Line now reads "Query an LDAP server with a no-newline password file for all items that are a member of the given group and return the object's {{displayName}} value:"

I'd be happy for a suggestion on phrasing, but tripped over the newline issue myself during testing.

pages/linux/ldapsearch.md
lbonanomi: After much incremental editing, I think I have this sorted.

lbonanomi:
I've dropped the SunOS branch. Everything in the Linux branch shows common functionality at this point due to size-constraints, and it appears that the SunOS machines I use are all back-rev enough to make lousy general examples.

sbrl:
Ah, I see. Looks ok to me :-)

pages/common/ldapsearch.md
```diff
@@ -0,0 +1,35 @@
+# ldapsearch
+
+> CLI utility for querying an LDAP directory.
+
+- Query an LDAP server for all items that are a member of the given group and return the object's {{displayName}} value:
+
+`ldapsearch -D '{{admin_DN}}' -w '{{password}}' -h {{ldap_host}} -b {{base_ou}} '{{memberOf}}={{group1}}' {{displayName}}`
```
agnivade: Change displayName in the description to `displayName`

Change displayName to display_name everywhere.

Also, it looks like `memberOf` is compulsory to get the member of a group. In that case, it should be outside the token.

lbonanomi: @agnivade: I see what you're saying, but I'm shy about changing the fields provided in the examples. `ldapsearch` works with an LDAP server's locally defined schema, so any fields provided as arguments must be treated as literal values. 

The {{memberOf}} and {{displayName}} fields are part of the default schema for Active Directory, which seems common enough to warrant being made an example.

What do you think?

agnivade: Then please remove them from the tokens.

sbrl: I think what he's trying to say is that they are concrete examples. What we might want to do is `{{memberOf=group1}}` instead of `{{memberOf}}={{group1}}`

pages/common/ldapsearch.md
```diff
@@ -0,0 +1,35 @@
+# ldapsearch
+
+> CLI utility for querying an LDAP directory.
+
+- Query an LDAP server for all items that are a member of the given group and return the object's displayName value:
+
+`ldapsearch -D '{{admin_DN}}' -w '{{password}}' -h {{ldap_host}} -b {{base_ou}} 'memberOf=group1' displayName`
```
agnivade: group1 should be inside tokens.

pages/common/ldapsearch.md
```diff
@@ -0,0 +1,35 @@
...
+
+`ldapsearch -D '{{admin_DN}}' -w '{{password}}' -h {{ldap_host}} -b {{base_ou}} 'memberOf={{group1}}' -l 7 displayName`
+
+- Invert the filter:
+
+`ldapsearch -D '{{admin_DN}}' -w '{{password}}' -h {{ldap_host}} -b {{base_ou}} '(!(memberOf={{group1}}))' displayName`
+
+- Return all items that are part of multiple groups, returning the display name for each item:
+
+`ldapsearch -D '{{admin_DN}}' -w '{{password}}' -h {{ldap_host}} '(&(memberOf=group1)(memberOf=group2)(memberOf=group3))' "displayName"`
```
agnivade: groups should be inside tokens. Same for the later examples.

pages/common/ldapsearch.md
lbonanomi: I drafted a change for the above. 

Honestly I seem to be having a devil of a time with formatting, so maintainers _please_ feel free to nudge any changes you think are correct.

sbrl: No problem, @lbonanomi! Yeah, our format is somewhat odd. It's based on [CommonMark](https://commonmark.org/), but it's got some customisations to it.

And in this case, I posted a comment 'cause I wanted @agnivade's feedback

lbonanomi:
We did it! Thank you everyone!

PR #2916, opened 2019-04-15
jq: add 'output to string' example

bbigras:
I needed to figure this out to export restic's json data to prometheus/node_exporter's textfile collector.

For some reason, only the opening parenthesis needs to be escaped it seems.

You can test with:
`echo '{"k": "v"}' | jq --raw-output '"test: \(.k)"'`


pages/common/jq.md
```diff
@@ -25,3 +25,7 @@
 - Output the value of a given key of each element in a JSON text from stdin:
 
 `cat {{file.json}} | jq 'map(.{{key_name}})'`
+
+- Output the value of a given key to a string:
```
agnivade: We should also explain the raw-output part.

bbigras: > With this option, if the filter’s result is a string then it will be written directly to standard output rather than being formatted as a JSON string with quotes. This can be useful for making jq filters talk to non-JSON-based systems.

https://stedolan.github.io/jq/manual/

Would something like a " (and disable JSON output)" suffix be enough?

agnivade: sure

bbigras:
I did the change and resolved the conflict.

PR #2920, opened 2019-04-16
7z: remove spaces after options and fix characters

andrik:

### DIFF
--- a/pages.pt-BR/common/7z.md
+++ b/pages.pt-BR/common/7z.md
@@ -5,11 +5,11 @@
 
 - Compactar um arquivo ou diretório:
 
-`7z a {{arquivo_compactado.7z}} {{caminho/arquivo_ou_diretorio}}`
+`7z a {{arquivo_compactado.7z}} {{caminho/arquivo_ou_diretório}}`
 
 - Criptografar um arquivo existente (incluindo cabeçalhos):
 
-`7z a {{arquivo_criptografado.7z}} -p {{senha}} -mhe {{arquivo.7z}}`
+`7z a {{arquivo_criptografado.7z}} -p{{senha}} -mhe {{arquivo.7z}}`
 
 - Descompactar um arquivo mantendo a estrutura de diretórios original:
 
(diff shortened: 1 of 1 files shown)


mebeim:
Thanks for the PR @andrik, we usually use spaces after options though, so I'd suggest adding a space on the example which is missing it rather than removing it from the other examples.

andrik:
Hi @mebeim, I forgot to mention that this PR refers to #2912

mebeim:
Thanks for the link @andrik, I missed those comments, my bad. I just noticed that the man page doesn't use spaces, makes sense to be consistent then.

pages.pt-BR/common/7z.md
```diff
@@ -5,23 +5,23 @@
 
 - Compactar um arquivo ou diretório:
 
-`7z a {{arquivo_compactado.7z}} {{caminho/arquivo_ou_diretorio}}`
+`7z a {{arquivo_compactado.7z}} {{caminho/arquivo_ou_diretório}}`
 
 - Criptografar um arquivo existente (incluindo cabeçalhos):
 
-`7z a {{arquivo_criptografado.7z}} -p {{senha}} -mhe {{arquivo.7z}}`
+`7z a {{arquivo_criptografado.7z}} -p{{senha}} -mhe {{arquivo.7z}}`
```
waldyrious: Maybe here it makes sense to split -mh from -e, and remove the space between -e and the filename? Or is it unrelated?

sbrl: The man page isn't clear, but I think that `-mhe` is all one option.

waldyrious: Wow, really? These pages seem to have quite cryptic interface :/

mebeim: `-mhe` is the `-m` option with parameter `he`, as documented in the man page: 

```-m{Parameters}
-mhe=on|off
7z format only : enables or disables archive header encryption (Default : off)```

andrik: Hey guys, I found this:

> he=[off | on]
> Enables or disables archive header encryption. The default mode is he=off.

I looked the file (/usr/share/doc/p7zip/DOC/MANUAL/cmdline/switches/method.htm) suggested on manual.

waldyrious: I'm tempted to suggest `-m"he"` or `-m'he'`, just to be sure this is properly understood. I suppose that'd work, right?

mebeim: It's the same for me really, I don't have a preference, both look clear enough.

waldyrious: @andrik what do you think?

andrik: I think it's better using -mhe=on

andrik: May I change?

waldyrious: > May I change?

No objections to the change were raised, so please go ahead :)

andrik: Done

PR #2934, opened 2019-04-18
mkcert: add page

Aracki:
I have added [mkcert](https://github.com/FiloSottile/mkcert).

PR #2939, opened 2019-04-21
mkcert: minor grammar fix

waldyrious:

This PR is a follow-up to #2934. It applies a change that was [suggested there](https://github.com/tldr-pages/tldr/pull/2934#pullrequestreview-228655114) but was overlooked before merging.

### DIFF
--- a/pages/common/mkcert.md
+++ b/pages/common/mkcert.md
@@ -11,7 +11,7 @@
 
 `mkcert {{example.org}}`
 
-- Generate certificate and private key for the multiple domains:
+- Generate certificate and private key for multiple domains:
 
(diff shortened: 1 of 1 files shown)


mebeim:
Dammit. Sorry for that. I don't get why people mark conversations as solved when they actually are not. @Aracki please only mark something as solved when it actually is... :slightly_frowning_face:

Aracki:
@mebeim I forced push 4 commits and got an email that PR got merged. I think I haven't marked it as solved. However, I will be more careful in the future. 👍

mebeim:
@Aracki well it says "marked as solved by Aracki". By the way no problem it can happen, and also I should've looked better :') thanks again for the page :slightly_smiling_face:

PR #2940, opened 2019-04-22
csc: clarify example

mebeim:
Clarify that a resource does not have to be another `.cs` source file, but can be any kind of file. Also, the identifier following the resource name is not needed since it defaults to the embedded resource filename.

### DIFF
--- a/pages/common/csc.md
+++ b/pages/common/csc.md
@@ -20,7 +20,7 @@
 
 - Embed a resource:
 
-`csc /resource:{{path/to/target_file.cs}},{{namespace.string.name}} {{path/to/input_file.cs}}`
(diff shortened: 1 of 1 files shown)


sbrl:
What is the default embedded resource filename though? I assumed that it was non-obvious (I at least don't know what it is), so I included the explicit resource name here.

I guess this is fine though :-)

mebeim:
Well if you include `path/to/some_file` it would be `some_file`. I could re-add that part if you think it makes things clearer.

sbrl:
Ah, I see! I must be getting confused because then you include an embedded resource as part of a project, it takes on the default namespace as a prefix.

sbrl:
Ah, I see! I must be getting confused because then you include an embedded resource as part of a project, it takes on the default namespace as a prefix.

PR #2941, opened 2019-04-22
ab: add Brazilian Portuguese translation

andrik:

### DIFF
--- a/pages.pt-BR/common/ab.md
+++ b/pages.pt-BR/common/ab.md
@@ -1,13 +1,12 @@
 # ab
-
-> Ferramenta de benchmarking do servidor Apache. A ferramenta mais simples para realizar testes de carga.
+> Ferramenta do servidor Apache para realizar benchmarking e testes de carga em servidores web.
 > Página oficial: <https://httpd.apache.org/docs/2.4/programs/ab.html>.
 
 - Executar 100 requisições HTTP do tipo GET para uma determinada URL:
 
 `ab -n {{100}} {{url}}`
 
-- Executar 100 requisições HTTP do tipo GET para uma determinada URL. Sendo que as requisições concorrentes acontecerão em grupos de 10:
+- Executar 100 requisições HTTP do tipo GET para uma determinada URL, executando 10 requisições simultâneas de cada vez:
 
 `ab -n {{100}} -c {{10}} {{url}}`
 
--- a/pages.pt-BR/common/ab.md
+++ b/pages.pt-BR/common/ab.md
@@ -1,0 +1,20 @@
+# ab
+
+> Ferramenta de _benchmarking_ do servidor Apache. A ferramenta mais simples para realizar testes de carga.
+> Página oficial: <https://httpd.apache.org/docs/2.4/programs/ab.html>.
+
+- Executar 100 requisições HTTP do tipo GET para uma determinada URL:
+
+`ab -n {{100}} {{url}}`
+
+- Executar 100 requisições HTTP do tipo GET para uma determinada URL. Sendo que as requisições concorrentes acontecerão em grupos de 10:
+
+`ab -n {{100}} -c {{10}} {{url}}`
+
+- Utilizar a funcionalidade HTTP _Keep Alive_, permitindo que várias requisições sejam feitas em uma sessão HTTP:
+
+`ab -k {{url}}`
+
+- Definir o tempo total do _benchmarking_, em segundos:
+
+`ab -t {{60}} {{url}}`
--- a/pages.pt-BR/common/ab.md
+++ b/pages.pt-BR/common/ab.md
@@ -1,6 +1,6 @@
 # ab
 
-> Ferramenta de _benchmarking_ do servidor Apache. A ferramenta mais simples para realizar testes de carga.
+> Ferramenta de benchmarking do servidor Apache. A ferramenta mais simples para realizar testes de carga.
 > Página oficial: <https://httpd.apache.org/docs/2.4/programs/ab.html>.
 
 - Executar 100 requisições HTTP do tipo GET para uma determinada URL:
@@ -11,10 +11,10 @@
 
 `ab -n {{100}} -c {{10}} {{url}}`
 
-- Utilizar a funcionalidade HTTP _Keep Alive_, permitindo que várias requisições sejam feitas em uma sessão HTTP:
+- Utilizar a funcionalidade HTTP Keep Alive, permitindo que várias requisições sejam feitas em uma sessão HTTP:
 
 `ab -k {{url}}`
 
-- Definir o tempo total do _benchmarking_, em segundos:
+- Definir o tempo total do benchmarking, em segundos:
 
 `ab -t {{60}} {{url}}`
--- a/pages.pt-BR/common/ab.md
+++ b/pages.pt-BR/common/ab.md
@@ -1,5 +1,6 @@
 # ab
-> Ferramenta do servidor Apache para realizar benchmarking e testes de carga em servidores web.
+
+> Ferramenta da Apache para realizar benchmarking e testes de carga em servidores web.
 > Página oficial: <https://httpd.apache.org/docs/2.4/programs/ab.html>.
 
 - Executar 100 requisições HTTP do tipo GET para uma determinada URL:


pages.pt-BR/common/ab.md
```diff
@@ -0,0 +1,20 @@
+# ab
+
+> Ferramenta de _benchmarking_ do servidor Apache. A ferramenta mais simples para realizar testes de carga.
```
schneiderl: @mebeim @sbrl @agnivade How are we dealing with foreign words on translated pages? Is this the correct way to approach it?

schneiderl: I havent seen anything about this on CONTRIBUTING.MD. If we decide this is the way to do it I believe this should be clearly stated there.

mebeim: I don't really see this as a problem. Some English words are ok to use in other languages, and some are better translated. There isn't really a rule to follow here, depends on the language and the specific situation. For example, "benchmarking" in Italian is a commonly used term in informatics, and I used it without problem in [the translation I made](https://github.com/tldr-pages/tldr/blob/master/pages.it/common/ab.md). Probably this is the case for Portuguese too, isn't it?

By the way, I would remove the underscores (`_`) around the word, it doesn't need any particular emphasis.

schneiderl: Oh maybe I did not made clear what I was trying to say. I also don't see any problems on using english words in other languages. 

I was talking about the underscores there. Should we keep or remove them?

mebeim: @schneiderl oooh I see, I totally misunderstood haha. Regarding the underscores, I think we should remove them, there really is no need to highlight anything in that description IMHO.

sbrl: What do you mean "foreign" words? As far as I understand it, I'd recommend following any existing guidelines you know of to work out which words should be translated. How would you naturally say it in the target language?

The only things we can't really translate are the command names, anything that would break the actual command invocation, and the syntax that the linter uses to validate the page.

mebeim: Well, "benchmarking" is "benchmarking" in Italian, same for "file", "directory", and many other informatic words. What @schneiderl was talking about were the surrounding underscores (`_`), not the translation, I mixed things up with my reply :')

mebeim:
```suggestion
> Ferramenta de benchmarking do servidor Apache. A ferramenta mais simples para realizar testes de carga.
```
I'd suggest removing those underscores surrounding "benchmark", there's no need to emphasize the word.

sbrl: Ah, the underscores mean _italics_ in markdown :P

mebeim: @sbrl yes I know, that's why I'm saying there's no need for it.

andrik: Hi guys, I decided to put underscores(italic) because in Brazil it's mandatory on academic papers, so I'm used to doing that. 

I agree that we don't need that on TLDR

andrik: :+1:

pages.pt-BR/common/ab.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Executar 100 requisições HTTP do tipo GET para uma determinada URL:
+
+`ab -n {{100}} {{url}}`
+
+- Executar 100 requisições HTTP do tipo GET para uma determinada URL. Sendo que as requisições concorrentes acontecerão em grupos de 10:
+
+`ab -n {{100}} -c {{10}} {{url}}`
+
+- Utilizar a funcionalidade HTTP _Keep Alive_, permitindo que várias requisições sejam feitas em uma sessão HTTP:
```
mebeim:
```suggestion
- Utilizar a funcionalidade HTTP Keep Alive, permitindo que várias requisições sejam feitas em uma sessão HTTP:
```

pages.pt-BR/common/ab.md
```diff
@@ -0,0 +1,20 @@
...
+
+`ab -k {{url}}`
+
+- Definir o tempo total do _benchmarking_, em segundos:
```
mebeim:
```suggestion
- Definir o tempo total do benchmarking, em segundos:
```

pages.pt-BR/common/ab.md
andrik: @mebeim Sorry, I didn't remember that was more than one.

pages.pt-BR/common/ab.md
```diff
@@ -0,0 +1,20 @@
+# ab
+
+> Ferramenta de benchmarking do servidor Apache. A ferramenta mais simples para realizar testes de carga.
```
waldyrious: > A ferramenta mais simples para realizar testes de carga.

This is subjective, and borders on ad-like. I'd suggest removing it.

pages.pt-BR/common/ab.md
```diff
@@ -0,0 +1,20 @@
...
+> Página oficial: <https://httpd.apache.org/docs/2.4/programs/ab.html>.
+
+- Executar 100 requisições HTTP do tipo GET para uma determinada URL:
```
waldyrious: I'd suggest wrapping HTTP and GET in backticks.

pages.pt-BR/common/ab.md
```diff
@@ -0,0 +1,20 @@
...
+
+`ab -n {{100}} {{url}}`
+
+- Executar 100 requisições HTTP do tipo GET para uma determinada URL. Sendo que as requisições concorrentes acontecerão em grupos de 10:
```
waldyrious: Eu simplificava

> Sendo que as requisições concorrentes acontecerão em grupos de 10

para apenas

> em grupos de 10

ou possivelmente

> em grupos concorrentes de 10 requisições cada

pages.pt-BR/common/ab.md
mebeim: Why would that be necessary? Those are not other command names or special tokens. I've never seen backticks used in such cases in other pages.

See the following pages for example:

https://github.com/tldr-pages/tldr/blob/9c59940f45fd5371b190c2f170425d789a742cbb/pages/linux/mke2fs.md#L5

https://github.com/tldr-pages/tldr/blob/9c59940f45fd5371b190c2f170425d789a742cbb/pages/common/nasm.md#L5

https://github.com/tldr-pages/tldr/blob/9c59940f45fd5371b190c2f170425d789a742cbb/pages/common/less.md#L17

pages.pt-BR/common/ab.md
mebeim: I see this was translated literally from the English page, you're right it's subjective, it should be removed from there too. Perhaps something like "a simple tool for load testing", instead of "the simplest" would be ok, what do you think?

schneiderl: @mebeim totally agree with that. Writing it that way almost look like an ad.

waldyrious: I don't see why we need two sentences saying "a tool to do X". If we want to keep "testes de carga", then we could combine the sentences, e.g. something like this:

> Ferramenta do servidor Apache para realizar benchmarking e testes de carga.

What do you guys think?

pages.pt-BR/common/ab.md
waldyrious: @mebeim in my understanding, backticks serve to identify specific technical terms amidst regular English prose, such as commands, filenames, keyboard shortcuts, etc.

Granted, in this sentence HTTP could perhaps be considered a regular acronym, like URL, and be left unmarked; but GET is a specific HTTP verb that carries technical meaning in this context, so I still think it should be marked as code.

pages.pt-BR/common/ab.md
andrik: You're are right guys. 

What do you think about adding a section of CONTRIBUTING.md showing some good and bad examples about descriptions, translations, etc?

pages.pt-BR/common/ab.md
mebeim: I see. I would personally only use backticks for tokens and command names, but I get your point.

pages.pt-BR/common/ab.md
mebeim: @waldyrious you're right, that's repetitive. We should go with something like:

> Apache Benchmarking tool to perform load testing.

@andrik that could be a good idea, I wouldn't know where precisely to add it though. I'll take a closer look when I've got time.

pages.pt-BR/common/ab.md
andrik: I agree

pages.pt-BR/common/ab.md
andrik: Que tal?

> Executar 100 requisições HTTP do tipo GET para uma determinada URL. Em grupos de 10 concorrentemente.

pages.pt-BR/common/ab.md
waldyrious: I think such a section would make more sense in the [style guide](https://github.com/tldr-pages/tldr/blob/master/contributing-guides/style-guide.md), but overall it's a good idea, yes.

waldyrious: > Apache Benchmarking tool

@mebeim that got me thinking -- is this a tool of the Apache HTTP server, or a tool under the Apache organization's umbrella? If it's the latter, the Portuguese description needs to be changed, because it refers to the Apache server.

pages.pt-BR/common/ab.md
waldyrious: Hm, acho que assim não fica claro se o que é concorrente são os grupos ou as chamadas dentro de cada grupo. (Já agora, podes confirmar qual dessas é a interpretação correta?)

pages.pt-BR/common/ab.md
mebeim: @waldyrious it's developed by Apache, but it's general purpose, for *any* web server.

andrik: Even though you can use ab on other web servers, it's distributed [within](https://httpd.apache.org/docs/2.4/programs/) Apache Web Server 

As we can't install it as a standalone program I think it would be better described as Apache Web Server Tool. What do you think?

pages.pt-BR/common/ab.md
andrik: > Number of multiple requests to perform at a time. Default is one request at a time.

Descrição retirada do site oficial, então é o segundo caso. Acho que o termo grupo não está muito bom para essa descrição. Que tal:

> Executar 100 requisições HTTP do tipo GET para uma determinada URL, definindo 10 requisições simultâneas de cada vez.

pages.pt-BR/common/ab.md
mebeim: @andrik it is not really part of the Apache Web Server, it is part of `apache2-utils`, which includes [various utilities](https://packages.debian.org/stretch/apache2-utils). Anyway, the appropriate description IMHO would still be:

> Apache Benchmarking tool to perform load testing.

Or, maybe, more detailed:

> Apache Benchmarking tool to perform web server load testing.

andrik: Hi @mebeim, you are right

I was fooled by the first sentence of [docs](https://httpd.apache.org/docs/2.4/programs/) of Apache :man_facepalming: 

> This page documents all the executable programs included with the Apache HTTP Server.

pages.pt-BR/common/ab.md
sbrl: I'm not sure that HTTP warrants backticks as it's an abbreviation, right?

mebeim: @sbrl Acronym, but yeah, that was my point. Same for GET/POST etc, never saw those wrapped in backticks.

pages.pt-BR/common/ab.md
andrik: > @andrik it is not really part of the Apache Web Server, it is part of `apache2-utils`, which includes [various utilities](https://packages.debian.org/stretch/apache2-utils), and the package description says `Provides some add-on programs useful for any web server.`.
> 
> Anyway, the appropriate description IMHO would still be:
> 
> > Apache Benchmarking tool to perform load testing.
> 
> Or, maybe, more detailed:
> 
> > Apache Benchmarking tool to perform web server load testing.

I think the latter is good, does everybody agree?

pages.pt-BR/common/ab.md
andrik: So, do we have a consensus about that? For me, the protocol name and verbs should be without backticks?

pages.pt-BR/common/ab.md
andrik: @waldyrious ou @schneiderl, o que vcs acham?

pages.pt-BR/common/ab.md
mebeim: Well dunno, looks like a Mexican standoff :thinking: :rofl:

pages.pt-BR/common/ab.md
schneiderl: @andrik acho que fica mais claro. Concordo com a alteração.

waldyrious: Também concordo. Talvez usaria "executando" ou "realizando" em vez de "definindo", mas fora isso, parece-me bem :)

pages.pt-BR/common/ab.md
waldyrious: My suggestion above, was:

> HTTP could perhaps be considered a regular acronym, like URL, and be left unmarked

so I'd only insist in "backticking" `GET` and `POST` (which are not acronyms, btw), again for the reason I suggested above:

> `GET` is a specific HTTP verb that carries technical meaning in this context, so I still think it should be marked as code.

Is that still a deal-breaker to you, @mebeim? From my reading of the discussion above it looks like this would be fine with @sbrl (but please confirm as well).

pages.pt-BR/common/ab.md
waldyrious: > I think the latter is good, does everybody agree?

I agree 👍

pages.pt-BR/common/ab.md
mebeim: @waldyrious I mean, for what I see that's not coherent with the rest of the pages. If we want to change GET into `GET` and POST into `POST` then we'd also have to change it elsewhere, not only in this page. Few examples:

https://github.com/tldr-pages/tldr/blob/dbf4e997968223ac4772c0634aa044cac0648685/pages/common/curl.md#L18

https://github.com/tldr-pages/tldr/blob/dbf4e997968223ac4772c0634aa044cac0648685/pages/common/mitmdump.md#L10

https://github.com/tldr-pages/tldr/blob/dbf4e997968223ac4772c0634aa044cac0648685/pages/linux/httpie.md#L5

https://github.com/tldr-pages/tldr/blob/dbf4e997968223ac4772c0634aa044cac0648685/pages/linux/httpie.md#L9

I also never saw GET or POST be treated specially inside code blocks in any `README.md`, Markdown documentation, blog posts, et cetera, so I don't really think that's necessary.

However, with the above being said, I would not argue if the majority of us thinks it's better to have those wrapped in backticks.

pages.pt-BR/common/ab.md
mebeim: @andrik I agree too :+1:

pages.pt-BR/common/ab.md
waldyrious: I agree the change would need to happen in other pages as well to be consistent, but I don't think it's problematic to have this page kick-off that change and a later PR making the equivalent change to all other pages.

But if you guys either disagree with adding backticks to GET/POST etc. (I think both @andrik and @sbrl were OK with that, but please confirm), or think it's better for that change to happen entirely in a separate PR, that'd be fine by me.

andrik: @waldyrious I think this could be done in a separate PR

pages.pt-BR/common/ab.md
```diff
@@ -1,13 +1,12 @@
 # ab
-
-> Ferramenta de benchmarking do servidor Apache. A ferramenta mais simples para realizar testes de carga.
+> Ferramenta do servidor Apache para realizar benchmarking e testes de carga em servidores web.
```
waldyrious: You removed the blank line after the title, please restore it :)

waldyrious: Also, I thought we had [agreed](https://github.com/tldr-pages/tldr/pull/2941#discussion_r278220302) that the tool isn't specific to the Apache _server_. The english description we agreed on was

> Apache Benchmarking tool ...

So in Portuguese it would be something like

> Ferramenta da Apache ...

Am I missing something?

andrik: :+1:

pages.pt-BR/common/ab.md
waldyrious: Unresolving this conversation to make it more discoverable. We had some trouble re-locating it in #3114 😅

PR #2943, opened 2019-04-23
nologin: add page

lbonanomi:

This PR is in support of #2214.


pages/linux/nologin.md
```diff
@@ -0,0 +1,11 @@
...
+
+> Alternative shell that prevents a user's login.
+
+- Set user's login shell to `nologin` to prevent user logging-in:
+
+`chsh -s {{user}} nologin`
+
+- Customize message for users with the login shell of `nologin`:
+
+`echo "{{declined_login_message}}" > /etc/nologin.txt`
```
schneiderl: "/etc/nologin.txt" is probably a parameter and it can be changed by the user, right? 
If that is correct I would sugest having curly brackets around it and changing the name for something like "/path/to/output.txt"

lbonanomi: @schneiderl: Thank you! This is a thoughtful thing to ask and was an interesting question for me to try to answer. 

The source references "_PATH_NOLOGIN_TXT" at [line 63](https://github.com/mmalecki/util-linux/blob/5fdd35070ceb79589d52fa247d1b359bd2bdc471/login-utils/nologin.c#L63) of no login.c and seems to define the filename variable at [line 34](https://github.com/mmalecki/util-linux/blob/5fdd35070ceb79589d52fa247d1b359bd2bdc471/include/pathnames.h#L34) of pathnames.h, so this seems to be baked-into the distributed binary.

lbonanomi:

pages/linux/nologin.md
```diff
@@ -0,0 +1,11 @@
+# nologin
+
+> Alternative shell that prevents a user's login.
+
+- Set user's login shell to `nologin` to prevent user logging-in:
```
mebeim:
```suggestion
- Set a user's login shell to `nologin` to prevent the user from logging in:
```
Just some grammar.

pages/linux/nologin.md
mebeim: @lbonanomi so it's basically a static configuration file, right? In such case that's correct as is.

lbonanomi: @mebeim: I'm pretty sure it's static.

mebeim: Right, thank you.

pages/linux/nologin.md
```diff
@@ -0,0 +1,11 @@
+# nologin
+
+> Alternative shell that prevents a user's login.
```
sbrl: Perhaps this will read better?
```suggestion
> Alternative shell that prevents a user from logging in.
```

pages/linux/nologin.md
schneiderl: Thank you for the answer. Approving it then :)

PR #2958, opened 2019-04-29
nsenter: add page

lbonanomi:

This PR is in support of https://github.com/tldr-pages/tldr/issues/2214


pages/linux/nsenter.md
```diff
@@ -0,0 +1,16 @@
+# nsenter
+
+> Run a new command in a running process' namespace.
+> Particularly useful for docker images of chroot jails.
+
+- Run command in existing processes network namespace:
+
+`nsenter -t {{pid}} -n {{command}} {{command_arguments}}`
+
+- Run command in existing processes process namespace:
```
schneiderl: This sentence is kind of confusing. Can we find a way of rewriting it?

lbonanomi: @schneiderl: Agreed. I was excited because I finally found a use-case for it.

How about `Run a new command in an existing processes ps-table namespace`?

schneiderl: That is way better in my opinion. I say we go for it.

pages/linux/nsenter.md
```diff
@@ -0,0 +1,17 @@
...
+
+`nsenter -t {{pid}} -n {{command}} {{command_arguments}}`
+
+- Run a new command in an existing processes ps-table namespace:
+
+`nsenter -t {{pid}} -p {{command}} {{command_arguments}}`
+
+- Run command in existing processes network namespace:
+
+`nsenter -t {{pid}} -n {{command}} {{command_arguments}}`
```
pxgamer: This example is the same as the first example 🤔 Should this be removed?

lbonanomi: Well that's no good! Let's replace it with a different namespace.

PR #2959, opened 2019-04-29
autojump: add page

Aracki:
I have added [autojump](https://github.com/wting/autojump) - a tool for jumping through dirs. 

PS. `j` is a convenience wrapper function around `autojump`

### DIFF
--- a/pages/common/autojump.md
+++ b/pages/common/autojump.md
@@ -1,24 +1,24 @@
 # autojump
 
-> Quickly jump among the directories you visit the most.
-> Aliases like j or jc are provided for even less typing.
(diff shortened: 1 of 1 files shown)


sbrl:
Any reason for closing this, @Aracki?

Aracki:
@sbrl There is already `autojump` [page](https://github.com/tldr-pages/tldr/blob/master/pages/common/autojump.md). 

I was trying to find it with `tldr j`. Probably it would be a bad idea to change it to `j`.

sbrl:
Ah, I see! No problem :-)

We could always have a copy of it under `j` to avoid confusion, with a `see also` linking the 2 to each other.

PR #2960, opened 2019-04-29
package.json: update dependencies

pxgamer:
Note, this can probably be rebased and merged rather than squashing. 👍

-----

Due to the issues mentioned in PRs #2867 and #2868, I've created this new PR with the appropriate changes committed by Dependabot.

I was planning on updating [`markdownlint-cli`](https://npm.im/markdownlint-cli) to `v0.15.0` which is the latest, but it now actually enforces the tests (see the [Travis build failure](https://travis-ci.com/pxgamer/tldr/builds/109987111)) so it may be better to implement this with a future PR.

PR #2965, opened 2019-05-01
package.json: update markdownlint-cli to v0.15.0

pxgamer:
As mentioned in #2960, this updates [`markdownlint-cli`](https://npm.im/markdownlint-cli) to `v0.15.0`.

The package now actually enforces the tests. The issue that arises from this is that there is a page [`pages/linux/whiptail.md:27`](https://github.com/pxgamer/tldr/blob/da2344c7a5e289b92aa3d767363bfba4afb67e67/pages/linux/whiptail.md) that has a line with **233** characters. Whereas we restrict pages to a max of **200** characters in our linting config.

This is mentioned in previous builds, but it won't be enforced when linting (the exit code is `0`). In the new versions, the exit code is `1`, so Travis CI fails to build.

To resolve this, we could cut down or remove the `Display a multiple-choice menu` example in `whiptail` (@sbrl?), or increase/remove the line length limit?

### DIFF
--- a/.markdownlintrc
+++ b/.markdownlintrc
@@ -2,7 +2,7 @@
   "default": true,
   "MD003": { "style": "atx" },
   "MD007": { "indent": 4 },
-  "MD013": { "line_length": 200 },
+  "MD013": { "line_length": 250 },
   "MD033": false,
   "MD034": false,
   "no-hard-tabs": false,
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Simplified, community-driven man pages",
   "dependencies": {
     "glob": "7.1.3",
-    "markdownlint-cli": "0.1.0",
+    "markdownlint-cli": "0.15.0",
     "tldr-lint": "~0.0.7",
     "husky": "1.3.1"
   },
--- a/package.json
+++ b/package.json
@@ -9,9 +9,9 @@
     "husky": "1.3.1"
   },
   "scripts": {
-    "lint-markdown": "markdownlint pages/**/*.md",
+    "lint-markdown": "markdownlint pages*/**/*.md",
     "lint-tldr": "tldr-lint ./pages",
-    "test": "bash -c 'markdownlint pages/ && tldr-lint ./pages 2>&1 | tee test_result; test ${PIPESTATUS[0]} -eq 0'",
+    "test": "bash -c 'markdownlint pages*/**/*.md && tldr-lint ./pages 2>&1 | tee test_result; test ${PIPESTATUS[0]} -eq 0'",
     "build-index": "node ./scripts/build-index.js | tee pages/index.json > index.json"
   },
   "husky": {


pxgamer:
@sbrl, that seems like a nice tool. 👍 Just saying, the reason I mentioned you in the PR description was because you created the `whiptail` page that has the line limit. Any thoughts on the last paragraph?

sbrl:
Ah, I didn't read the PR description properly and only looked at the dependency update :P

Erm unfortunately, `whiptail` is just a rather verbose command - and if we cut it down to under the 200 char limit we'd probably compromise readability.

Where did the 200 char limit come from, anyway? It seems kinda arbitrary.

pxgamer:
Looks like it was added in 2016 ([`.markdownlintrc`](https://github.com/tldr-pages/tldr/blame/master/.markdownlintrc#L5)). But in the previous configs that it replaced, I can't find any references to a line limit other than [`markdown-style.rb`](https://github.com/tldr-pages/tldr/blob/5f6b09b04b4b3c871a0e5c49e1770b83d609f4f4/scripts/markdown-style.rb) which specifically excluded `MD013` completely.

mebeim:
@pxgamer you know, I was taking a look at `markdownlint-cli` yesterday to try and figure out if it was actually configurable, and it's a shame that there's not an option to just print errors instead of returning `1`. I feel like PR should be made to `markdownlint-cli` to add such option. I would suggest appending a simple `|| true` in the lint command in the [`package.json`](https://github.com/tldr-pages/tldr/blob/master/package.json) to make it work like we'd like:

    "lint-markdown": "markdownlint pages/**/*.md || true",

pxgamer:
I feel like that defeats the purpose of using the package, unless it was just used as a local command. :thinking: Perhaps it would be better to just set `MD013` to `false`. We do want to know if there are errors in markdown, so that we can update them when reviewing PRs. Otherwise the Travis build might succeed even if there are tabs or the wrong heading formatting (I'm not sure how much we check with `tldr-lint`).

mebeim:
@pxgamer oh, I totally missed the fact that you could set a certain error to `false`. That's way better, I agree!

pxgamer:
@agnivade, @waldyrious, @schneiderl and @sbrl.

What are your opinions on disabling the `MD013` (line-length) check?

sbrl:
I'd agree that disabling it is probably the best option here.

pxgamer:
Could anyone push a commit to disable that option for me? As I'm away from my PC this weekend. Or we can leave this until Monday when I'll be back. :+1:

mebeim:
Done :+1: 

I also noticed only the english pages were being checked with the `test` or `lint-markdown` script and updated those two lines in the `package.json` as well.

pxgamer:
Cheers @mebeim. I think this is ready to go. :thinking:

agnivade:
Could we increase the limit to 250 ? I would feel safer if we had some limit instead of an unbounded line length.

schneiderl:
> Could we increase the limit to 250 ? I would feel safer if we had some limit instead of an unbounded line length.

I personally think this is the best option here. So that if we face this same issue again in the future we discuss it deeply just to make sure it does make sense to have a page so big.

mebeim:
Seems reasonable to me. I updated the linter config to 250 max length.

waldyrious:
Sorry I couldn't comment earlier. Just for future reference, I personally feel it would be more to the spirit of tldr to keep limits (page size, line length, etc.) short, and find ways to work within them, even if that means compromising on accuracy. Even 200 characters feel excessive to me for a single line :(

mebeim:
Well, I agree to some extent. I'd say limits are useful, but readability and usefulness of an example are to consider too. Most of the times an example can be rewritten to stay in the limits, or maybe even removed because it's too complicated, but I don't necessarily see exceptions like the one in `whiptail` as a problem.

PR #2966, opened 2019-05-01
J

Aracki:
I have added [j](https://github.com/wting/autojump) (copy of `autojump`) and links to each other.


Aracki:
@sbrl Here I made another PR about `autojump`. Can you please take a look?

mebeim:
I'm not sure about this PR:

- `j` is not a command itself. It's just a shortcut for `autojump`, and as a consequence the examples are also duplicated.
- `autojump` has more than one shortcut, if we create a page for one of them, does that also mean we should create a page for each of the others? Wouldn't really make much sense IMHO.
- The description of `autojump` taking up four lines doesn't look that pretty. I wouldn't suggest that addition.

sbrl:
Ah, good points @mebeim. Not actually having used `autojump` before, I thought that `j` was a command that autojump provides.

Aracki:
OK no problem, I will close it.

mebeim:
@sbrl `j` is indeed a command provided by `autojump`.

sbrl:
@mebeim If it's a command provided by `autojump` an  not an alias, should it not have it's own page?

mebeim:
@sbrl sorry for the misunderstanding, I meant to say that `j` is a command provided by `autojump` as a short alias to itself. So yes, basically an alias.

PR #2967, opened 2019-05-02
github-label-sync: add page

pxgamer:

### DIFF
--- a/pages/common/github-label-sync.md
+++ b/pages/common/github-label-sync.md
@@ -11,7 +11,7 @@
 
 `github-label-sync --access-token {{token}} --labels {{url|path/to/json_file}} {{repository_name}}`
 
-- Don't actually synchronise labels (i.e. do a dry run):
+- Perform a dry run instead of actually synchronising labels:
 
 `github-label-sync --access-token {{token}} --dry-run {{repository_name}}`
 
--- a/pages/common/github-label-sync.md
+++ b/pages/common/github-label-sync.md
@@ -1,0 +1,24 @@
+# github-label-sync
+
+> A command line interface for synchronising GitHub labels.
+> Homepage: <https://npmjs.com/package/github-label-sync>.
+
+- Synchronise labels using a local "labels.json" file:
+
+`github-label-sync --access-token {{token}} {{repository_name}}`
+
+- Synchronise labels using a specific labels JSON file:
+
+`github-label-sync --access-token {{token}} --labels {{url|path/to/json_file}} {{repository_name}}`
+
+- Synchonise labels with a dry run:
+
+`github-label-sync --access-token {{token}} --dry-run {{repository_name}}`
+
+- Keep labels that aren't in "labels.json":
+
+`github-label-sync --access-token {{token}} --allow-added-labels {{repository_name}}`
+
+- Synchronise using a `GITHUB_ACCESS_TOKEN` environment:
+
+`github-label-sync {{repository_name}}`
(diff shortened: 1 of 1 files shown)


pages/common/github-label-sync.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Synchronise labels using a local "labels.json" file:
+
+`github-label-sync --access-token {{token}} {{repository_name}}`
+
+- Synchronise labels using a specific labels JSON file:
+
+`github-label-sync --access-token {{token}} --labels {{url|path/to/json_file}} {{repository_name}}`
+
+- Synchonise labels with a dry run:
```
sbrl: Perhaps this will read better?
```suggestion
- Don't acutally synchonise labels (i.e. do a dry run):
```

pages/common/github-label-sync.md
```diff
@@ -0,0 +1,24 @@
...
+
+`github-label-sync --access-token {{token}} --dry-run {{repository_name}}`
+
+- Keep labels that aren't in "labels.json":
```
sbrl: Perhaps this will read better?
```suggestion
- Keep labels that aren't in `labels.json`:
```

Is this right? I can't remember whether filenames should be in backticks in descriptions or not

pages/common/github-label-sync.md
```diff
@@ -0,0 +1,24 @@
...
+
+`github-label-sync --access-token {{token}} --allow-added-labels {{repository_name}}`
+
+- Synchronise using a `GITHUB_ACCESS_TOKEN` environment:
```
sbrl: Perhaps this will read better?
```suggestion
- Synchronise using the `GITHUB_ACCESS_TOKEN` environment variable:
```

pages/common/github-label-sync.md
pxgamer: I've applied the other changes, I haven't yet applied this as I never know which they're meant to use.

pages/common/github-label-sync.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Synchronise labels using a local `labels.json` file:
+
+`github-label-sync --access-token {{token}} {{repository_name}}`
+
+- Synchronise labels using a specific labels JSON file:
+
+`github-label-sync --access-token {{token}} --labels {{url|path/to/json_file}} {{repository_name}}`
+
+- Don't actually synchronise labels (i.e. do a dry run):
```
agnivade: I am away from laptop atm. But how do we usually write dry run descriptions for other pages ? IMO mentioning the dry run part before may be better - Perform a dry run instead of actually synchronising labels.

pxgamer: We have a few ways looking at previous pages:
- Dry-run - ...
- Perform a dry-run...
- Do a dry-run...

It does sound better with `Perform a dry run` as you said. 👍

Is it worth standardising this across pages? Or not really?

waldyrious: I think it would be more intuitive to explicitly say something like "Simulate action X" for dry-run actions (I'm not opposed to actually using the "dry-run" term in addition to that wording).

pxgamer: D'oh. 😋 I should leave things before committing. I'll wait until we decide on the wording.

sbrl:
Are we ok to merge, @agnivade?

agnivade:
@waldyrious does it look good ?

waldyrious:
> @waldyrious does it look good ?

Sure, the current wording LGTM 👍

PR #2973, opened 2019-05-03
ipcmk: add page

lbonanomi:

### DIFF
--- a/pages/linux/ipcmk.md
+++ b/pages/linux/ipcmk.md
@@ -1,0 +1,19 @@
+# ipcmk
+
+> Create IPC (Inter-process Communication) resources.
+
+- Create shared memory segment:
+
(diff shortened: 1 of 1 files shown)


pages/linux/ipcmk.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Create a semaphore:
+
+`ipcmk --semaphore {{element_size}}`
+
+- Create a message queue:
+
+`ipcmk --queue`
+
+- Create a shared memory segment with permissions other than 0644:
```
agnivade: I think it might be better to say - "Create a shared memory segment with specific permissions (default is 0644)."

Atleast that's what I understand.

lbonanomi: I prefer this phrasing as well, will patch when I'm at a computer.

Thank you!

PR #2978, opened 2019-05-05
pactl: add page

terminalnode:


terminalnode:
> We usually write our pages in 3rd person. So let us remove all occurences of you/your.

Fixed :+1:

pages/linux/pactl.md
```diff
@@ -0,0 +1,23 @@
+# pactl
+
+> Control a running PulseAudio sound server.
+
+- List all sinks (or other types - sinks are outputs and sink-inputs are active audio streams):
+
+`pactl list {{sinks}} short`
```
waldyrious: How many types can there be? If only a few, I'd suggest listing them in the token using the pipe syntax, e.g. `{{sinks|sink-inputs}}`.

pages/linux/pactl.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Change the default sink (output) to 1 (the number can be retrieved via the list command):
```
waldyrious: Please wrap "list" in backticks since it's a specific keyword for this command.

pages/linux/pactl.md
```diff
@@ -0,0 +1,23 @@
...
+
+`pactl set-default-sink {{1}}`
+
+- Move a sink-input 627 to sink 1:
```
waldyrious:
```suggestion
- Move sink-input 627 to sink 1:
```

pages/linux/pactl.md
terminalnode: Eight types I think. cards, clients, modules, samples, sink-inputs, sinks, source-outputs, sources.

Though for someone reading the tldr I think sinks, sink-inputs, sources and source-outputs would be the most interesting.

I'm not entirely sure what you can do with clients, modules and samples to be honest, they're probably useful for people who know what they're doing but I don't think they do much for an average user.

pages/linux/pactl.md
terminalnode: Inside the description you mean? It makes sense, but I can't find anything about this is the style guidelines.

Come to think of it, would it perhaps be better to change "via the `list` command" to "via the `list` subcommand"?

pages/linux/pactl.md
waldyrious: I'm worrying that if we present a shorter list like `{{sinks|sink-inputs|sources|source-outputs}}`, it might induce people in error, as if we're suggesting no other values are possible (and it's not that compact either). I guess it makes more sense to keep this example as it is, but let me know if you think otherwise.

pages/linux/pactl.md
waldyrious: Yeah, there have been a couple discussions about usage of backticks and requests to have a guideline added to the documentation. I'll add it to my list, but I'm afraid it may take me a while to get to it :/

In any case, yes, that's what I meant. And I agree with your suggestion to use "subcommand" instead. 👍

terminalnode:
@agnivade I just learned of a very useful option where you can use `@DEFAULT_SINK@` in place of an ID when doing operations. Also on closer inspection I'm not sure this tool is linux specific but should probably work on other platforms where pulseaudio works.

With that in mind I'd like to push two new commits:
1) Changing the location to pages/common
2) Changing the last example (currently toggling mute on sink 1) to the quote below, showing of the ever useful `@DEFAULT_SINK@` keyword instead.
> - Toggle mute on default sink (using special name `@DEFAULT_SINK@`):
>
> `pactl set-sink-mute {{@DEFAULT_SINK@}} toggle

sbrl:
I think our policy on platform-specific stuff is that if it works on more than 1 platform, it belongs in the `common` directory - and then we can create platform-specific copies if functionality differs on some platforms.

PR #2981, opened 2019-05-07
gopass: add page

abacao:

### DIFF
--- a/pages/linux/gopass.md
+++ b/pages/linux/gopass.md
@@ -12,7 +12,7 @@
 
 - Mount a shared git store:
 
-`gopass mounts add {{store_name}} {{git_repo_url}}``
+`gopass mounts add {{store_name}} {{git_repo_url}}`
 
 - Search interactively using a keyword:
 
--- a/pages/linux/gopass.md
+++ b/pages/linux/gopass.md
@@ -2,7 +2,7 @@
 
 > Standard Unix Password Manager for Teams. Written in Go.
 
-- Startup config:
+- Initialise the configuration settings:
 
 `gopass init`
 
--- a/pages/linux/gopass.md
+++ b/pages/linux/gopass.md
@@ -1,6 +1,7 @@
 # gopass
 
 > Standard Unix Password Manager for Teams. Written in Go.
+> Homepage: <https://www.gopass.pw>.
 
 - Initialise the configuration settings:
 
--- a/pages/linux/gopass.md
+++ b/pages/linux/gopass.md
@@ -6,7 +6,7 @@
 
 `gopass init`
 
-- Show all "Keystores":
+- Show all "keystores":
 
 `gopass mounts`
 
@@ -14,7 +14,7 @@
 
 `gopass mounts add keystore_name location_git_repo`
 
-- Search interactive using a keyword:
+- Search interactively using a keyword:
 
 `gopass show keyword`
 
--- a/pages/linux/gopass.md
+++ b/pages/linux/gopass.md
@@ -6,29 +6,29 @@
 
 `gopass init`
 
-- Show all "keystores":
+- Show all stores:
 
 `gopass mounts`
 
-- Mount a shared git "keystore":
+- Mount a shared git store:
 
-`gopass mounts add keystore_name location_git_repo`
+`gopass mounts add {{store_name}} {{git_repo_url}}``
 
 - Search interactively using a keyword:
 
-`gopass show keyword`
+`gopass show {{keyword}}`
 
 - Search using a keyword:
 
-`gopass find keyword`
+`gopass find {{keyword}}`
 
-- Sync all mounted "keystores":
+- Sync all mounted stores:
 
 `gopass sync`
 
-- Show particular password entry:
+- Show a particular password entry:
 
-`gopass keystore_name/directory01/dir02/email@email.com`
+`gopass {{store_name}}/{{path/to/directory}}/{{email@email.com}}`
 
 - Create a new entry:
 
--- a/pages/linux/gopass.md
+++ b/pages/linux/gopass.md
@@ -7,6 +7,10 @@
 
 `gopass init`
 
+- Create a new entry:
+
+`gopass new`
+
 - Show all stores:
 
 `gopass mounts`
@@ -29,8 +33,4 @@
 
 - Show a particular password entry:
 
-`gopass {{store_name}}/{{path/to/directory}}/{{email@email.com}}`
-
-- Create a new entry:
-
-`gopass new`
+`gopass {{store_name|path/to/directory|email@email.com}}`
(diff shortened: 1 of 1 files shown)


pages/linux/gopass.md
```diff
@@ -0,0 +1,35 @@
+# gopass
+
+> Standard Unix Password Manager for Teams. Written in Go.
+
+- Startup config:
+
+`gopass init`
+
+- Show all "Keystores":
```
schneiderl: Should we have "Keystores" in uppercase? If so I think we should have it uppercase in all file.

I personally dont think this should be in uppercase.

pages/linux/gopass.md
```diff
@@ -0,0 +1,35 @@
...
+
+`gopass mounts`
+
+- Mount a shared git "keystore":
+
+`gopass mounts add keystore_name location_git_repo`
+
+- Search interactive using a keyword:
```
schneiderl: Shouldn't it be "Search interactively..."? as "Search interactive..." sounds a little odd.

pages/linux/gopass.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Search using a keyword:
+
+`gopass find keyword`
+
+- Sync all mounted "keystores":
+
+`gopass sync`
+
+- Show particular password entry:
```
schneiderl: Show this entry based on what? What is used to do the search? 
I think it would be good to clarify that on the description.

pages/linux/gopass.md
abacao: :+1:

pages/linux/gopass.md
abacao: "keystores" will have a name you want them to have. Doesn't make sense for me to be all uppercase. I can substitute for `keystores`

pages/linux/gopass.md
abacao: This command shows one entry in a specific location. That location if composed by keystore_name/..../entry.
Between the keystore and the entry, you can have tons of sub-folders.

pages/linux/gopass.md
schneiderl: IMHO It is better that way.

pages/linux/gopass.md
schneiderl: oh, understood. seems fine the way it is then.

pages/linux/gopass.md
abacao: :+1:

pages/linux/gopass.md
abacao: :+1:

abacao:
@schneiderl Please re-review

pages/linux/gopass.md
```diff
@@ -0,0 +1,35 @@
+# gopass
+
+> Standard Unix Password Manager for Teams. Written in Go.
+
+- Startup config:
```
sbrl: Perhaps this will read better?
```suggestion
- Initialise the configuration settings:
```

pages/linux/gopass.md
```diff
@@ -0,0 +1,35 @@
...
+
+`gopass init`
+
+- Show all "keystores":
```
sbrl: Not sure the quotes are needed here - same with the below example.

pages/linux/gopass.md
```diff
@@ -0,0 +1,35 @@
...
+
+`gopass mounts`
+
+- Mount a shared git "keystore":
+
+`gopass mounts add keystore_name location_git_repo`
```
sbrl: Perhaps this will read better?
```suggestion
`gopass mounts add {{keystore_name}} {{git_repo_url}}`
```

pages/linux/gopass.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Search interactively using a keyword:
+
+`gopass show keyword`
```
sbrl: Is `keyword` something that the user can change? If so, could you wrap it in `{{` and `}}` please as I've done in the suggestion above? Same with the below example.

pages/linux/gopass.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Search using a keyword:
+
+`gopass find keyword`
+
+- Sync all mounted "keystores":
+
+`gopass sync`
+
+- Show particular password entry:
```
sbrl: Perhaps this will read better?
```suggestion
- Show a particular password entry:
```

pages/linux/gopass.md
```diff
@@ -0,0 +1,35 @@
...
+
+`gopass keystore_name/directory01/dir02/email@email.com`
```
sbrl: Can the user change any of these arguments? Perhaps we could wrap them in tokens as I've done in some of the above examples.

Also, could we use `path/to/directory` please instead of `directory01/dir02`? This'll make it more consistent with other pre-existing pages.

pages/linux/gopass.md
abacao: I wasn't aware of the official name for it. Keystore, store, safe...
It seems it is "store". I will update it accordingly.

pages/linux/gopass.md
abacao: :+1:

pages/linux/gopass.md
abacao: :+1:

pages/linux/gopass.md
abacao: :+1:

abacao:
I will use your suggestion on the URL. About the other commands, I and starting to use this and those were my initial findings (I may suggest new ones in sometime).
I had more to add but I saw somewhere that the max was 8 entries.
I want to better understand the program before providing other entries.

abacao:
Do you need anything more from my side?

abacao:
Please review @sbrl @mebeim  :)

pages/linux/gopass.md
```diff
@@ -0,0 +1,36 @@
...
+
+`gopass find {{keyword}}`
+
+- Sync all mounted stores:
+
+`gopass sync`
+
+- Show a particular password entry:
+
+`gopass {{store_name}}/{{path/to/directory}}/{{email@email.com}}`
```
mebeim:
```suggestion
`gopass {{store_name|path/to/directory|email@email.com}}`
```
Better formatting.

pages/linux/gopass.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Create a new entry:
```
mebeim: I'd suggest moving this to the top as the second example since it looks more common/basic than others. Having it as the last example is counter-intuitive. Do you agree?

abacao:
Please review @sbrl @pxgamer @schneiderl @mebeim

PR #2982, opened 2019-05-07
transmission-remote: add page

terminalnode:

Some thoughts on this PR:
- The trickiest thing to figure out with this command is how to select torrents. The option -t is quite flexible. I tried to illustrate different ways in which -t can be used to select individual torrents, ranges such as torrent 1-10, combinations of the two as well as the option "all".
- If `{{hostname}}` isn't specified it's assumed to be localhost. I'd like to illustrate this somehow since a lot of people will be connecting to a local daemon, but I'm not sure how best to illustrate it without it looking like a mistake in the page.
- The order of the arguments when dealing with `-t` and `-a` is very important. This is made clear with `-t` I think, but less so for `-a`. For example in the first example if -a and -w are flipped the default download directory is changed permanently, but in the example the download directory is only specified for the new torrent.

### DIFF
--- a/pages/common/transmission-remote.md
+++ b/pages/common/transmission-remote.md
@@ -8,7 +8,7 @@
 
 - Change the default download directory:
 
-`transmission-remote {{hostname}} -w {{/path/to/download/to}}`
+`transmission-remote {{hostname}} -w {{path/to/download_directory}}`
 
 - List all torrents:
 
--- a/pages/common/transmission-remote.md
+++ b/pages/common/transmission-remote.md
@@ -4,7 +4,7 @@
 
 - Add a torrent file or magnet link to Transmission and download to a specified directory:
 
-`transmission-remote {{hostname}} -a {{torrent|url}} -w {{/path/to/download/to}}`
+`transmission-remote {{hostname}} -a {{torrent|url}} -w {{path/to/download_directory}}`
 
 - Change the default download directory:
 
--- a/pages/common/transmission-remote.md
+++ b/pages/common/transmission-remote.md
@@ -1,0 +1,31 @@
+# transmission-remote
+
+> Remote control utility for transmission-daemon and transmission.
+
+- Add a torrent file or magnet link to Transmission and download to a specified directory:
+
+`transmission-remote {{hostname}} -a {{torrent|url}} -w {{/path/to/download/to}}`
+
+- Change the default download directory:
+
+`transmission-remote {{hostname}} -w {{/path/to/download/to}}`
+
+- List all torrents:
+
+`transmission-remote {{hostname}} --list`
+
+- Start torrent 1 and 2, stop torrent 3:
+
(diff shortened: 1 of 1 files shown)


terminalnode:
For the record there is a tldr entry on `transmission-cli`, [which has been deprecated](https://github.com/transmission/transmission/commit/950387ab5a443629598f93c057f41150707866ab) in favor of `transmission-remote`:

> Limited to a single torrent at a time, transmission-cli is deprecated and exists primarily to support older hardware dependent upon it. In almost all instances, transmission-remote should be used instead.

pxgamer:
Thanks for the info about `transmission-cli`, shows how long it's been since I've used it.

I wonder if I should add a note to the description in the other page such as:
> This tool has been deprecated, please see `transmission-remote`.

pages/common/transmission-remote.md
```diff
@@ -0,0 +1,31 @@
+# transmission-remote
+
+> Remote control utility for transmission-daemon and transmission.
+
+- Add a torrent file or magnet link to Transmission and download to a specified directory:
+
+`transmission-remote {{hostname}} -a {{torrent|url}} -w {{/path/to/download/to}}`
```
pxgamer:
```suggestion
`transmission-remote {{hostname}} -a {{torrent|url}} -w {{path/to/download_directory}}`
```

pages/common/transmission-remote.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Change the default download directory:
+
+`transmission-remote {{hostname}} -w {{/path/to/download/to}}`
```
pxgamer:
```suggestion
`transmission-remote {{hostname}} -w {{path/to/download_directory}}`
```

pages/common/transmission-remote.md
```diff
@@ -0,0 +1,31 @@
...
+
+`transmission-remote {{hostname}} -t {{1}} --remove -t {{2}} --remove-and-delete`
+
+- Stop all torrents:
+
+`transmission-remote {{hostname}} -t {{all}} --stop`
+
+- Move torrents 1-10 and 15-20 to a new folder (folder will be created if it does not exist):
+
+`transmission-remote {{hostname}} -t {{"1-10,15-20"}} --move {{/new/path}}`
```
pxgamer:
```suggestion
`transmission-remote {{hostname}} -t {{"1-10,15-20"}} --move {{path/to/new_directory}}`
```

terminalnode:
All requested changes commited. :)

terminalnode:
> Thanks for the info about `transmission-cli`, shows how long it's been since I've used it.
> 
> I wonder if I should add a note to the description in the other page such as:
> 
> > This tool has been deprecated, please see `transmission-remote`.

I think that's a good idea, as a user of the program that's very much the kind of information I'd want there. I think I saw a discussion on another page, can't remember which, where something like that was suggested so I don't think it's unprecedented.

pxgamer:
I'll open a PR later after Endgame. 😋

PR #2987, opened 2019-05-08
pamixer: add page

terminalnode:

This tool might be too simple to really warrant a tldr page, but some examples can't hurt right? Especially since, at least on my system, there isn't actually a man page available (just the --help option).

I'm not 100% sure if this works on other operating systems than linux, but since PulseAudio does I think it does and thus it's placed in commons.


pages/common/pamixer.md
```diff
@@ -0,0 +1,35 @@
+# pamixer
+
+> A simple command-line mixer for PulseAudio.
+
+- List all sinks and sources with corresponding IDs:
```
sbrl: Perhaps this will read better?
```suggestion
- List all sinks and sources with their corresponding IDs:
```

pages/common/pamixer.md
```diff
@@ -0,0 +1,35 @@
...
+
+`pamixer --list-sinks --list-sources`
+
+- Set volume to 75% on the default sink:
```
sbrl: Perhaps this will read better?
```suggestion
- Set the volume to 75% on the default sink:
```

pages/common/pamixer.md
```diff
@@ -0,0 +1,35 @@
...
+
+`pamixer --set-volume {{75}}`
+
+- Toggle mute on a sink other than the default:
+
+`pamixer --toggle-mute --sink {{ID}}`
+
+- Increase volume on default sink by 5%:
```
sbrl: Perhaps this will read better?
```suggestion
- Increase the volume on default sink by 5%:
```

pages/common/pamixer.md
```diff
@@ -0,0 +1,35 @@
...
+
+`pamixer --increase {{5}}`
+
+- Decrease volume on a source by 5%:
```
sbrl: Perhaps this will read better?
```suggestion
- Decrease the volume on a source by 5%:
```

pages/common/pamixer.md
```diff
@@ -0,0 +1,35 @@
...
+
+`pamixer --decrease {{5}} --source {{ID}}`
+
+- Use the allow boost option to increase/decrease/set volume above 100%:
```
sbrl: Perhaps this will read better?
```suggestion
- Use the allow boost option to increase, decrease, or set the volume above 100%:
```

pages/common/pamixer.md
```diff
@@ -0,0 +1,35 @@
...
+
+`pamixer --set-volume {{105}} --allow-boost`
+
+- Mute the default sink (if already muted it will stay muted):
```
sbrl: Perhaps this will read better?
```suggestion
- Mute the default sink (if it's already muted it will stay muted):
```

pages/common/pamixer.md
```diff
@@ -0,0 +1,35 @@
...
+
+`pamixer --mute`
+
+- Unmute the default sink (if already unmuted it will stay unmuted):
```
sbrl: Perhaps this will read better?
```suggestion
- Unmute the default sink (if it's already unmuted it will stay unmuted):
```

terminalnode:
> Thanks for the new page, @terminalnode! I've left some comments below for you to review.

These all look good to me as well. All commited.

terminalnode:
> I see all your examples are concerning volume, but there's no mention of other more useful options like `--sink` and `--source`. I think some of the examples can be removed in favor of more useful options or even condensed.
> 
> For example, the last two examples could be rewritten as:
> 
> ```
> - Mute the default sink (use `--unmute` instead to unmute):
> 
> `pamixer --mute`
> ```

I agree, condensing those to one example would be better. It was something I considered but I don't think I've seen it in any other tldr pages and so opted not to. I'm pushing a new commit for this right after posting this.

As for the `--sink` and `--source` option, as far as I know they don't do anything on their own but are only used to specify which sink/source the actions are to be performed on and there are examples of this in the page. For example in the increase example no sink/source is specified and the description says that it acts on the default sink, then in the next example which is decrease the source is specified via  `--source`.

Perhaps something like this would be more illustrative?
- Perform a set of actions on a specified sink (use --source to perform them on a source instead):
`pamixer --sink {{ID}} {{[options]}}`

- Increase the volume of the default sink by 5% (use --default-source to increase the volume of the default source instead):
`pamixer --increase {{5}}`

(I'm not sure if there's some convention with regards to placeholders like {{[options]}}, but you get the idea)

mebeim:
@terminalnode thank you for the quick response and for clarifying the meaning of those options, I misunderstood their function while reading the help page of the command. I think the page looks good as it is now, no need to overcomplicate 😅

pages/common/pamixer.md
```diff
@@ -0,0 +1,31 @@
+# pamixer
+
+> A simple command-line mixer for PulseAudio.
```
mebeim:
```suggestion
> A simple command-line mixer for PulseAudio.
> Homepage: <https://github.com/cdemoulins/pamixer>.
```
Let's add the homepage while we're at it.

PR #2990, opened 2019-05-08
column: fix and reorder -s example

waldyrious:
The `-s` option only works if `-t` is passed as well (as the description indicates!)
And because of that, it should be shown after the `-t` example.

### DIFF
--- a/pages/common/column.md
+++ b/pages/common/column.md
@@ -7,13 +7,13 @@
 
 `printf "header1 header2\nbar foo\n" | column -c {{30}}`
 
-- Specify column delimiter character for the -t option (e.g. "," for csv); default is whitespace:
-
-`printf "header1,header2\nbar,foo\n" | column -s{{,}}`
-
 - Split columns automatically and auto-align in a tabular format:
 
 `printf "header1 header2\nbar foo\n" | column -t`
+
+- Specify column delimiter character for the -t option (e.g. "," for csv); default is whitespace:
+
+`printf "header1,header2\nbar,foo\n" | column -t -s{{,}}`
 
 - Fill columns before filling rows:
 


waldyrious:
It seems like this page would benefit from some clarification, by the way. Without the auto-detection of the number of columns (using `-t`), it seems to assume each full line (regardless of whitespace) is an entire cell, and arranges the cells into a grid:

```sh
# With no line breaks, there is a single "cell":
❯ echo -e "a b c d" | column
a b c d

# This single cell is not wrapped if the available width is constrained:
❯ echo -e "a b c d" | column -c 1
a b c d

# ...nor broken apart if there's plenty of space:
❯ echo -e "a b c d" | column -c 100
a b c d

# With one line break, there are two "cells":
❯ echo -e "a b\nc d" | column
a b     c d

# This time, the cells are wrapped when the available width is reduced:
❯ echo -e "a b\nc d" | column -c 5
a b
c d

# ...but not broken apart further:
❯ echo -e "a b\nc d" | column -c 1
a b
c d

# With line breaks between each element, each is considered its own cell:
❯ echo -e "a\nb\nc\nd" | column
a       b       c       d

# These cells can be wrapped to multiple lines if space is limited:
❯ echo -e "a\nb\nc\nd" | column -c 20
a       b
c       d

❯ echo -e "a\nb\nc\nd" | column -c 1
a
b
c
d
```

Additionally, as some of the examples above may lead you to suspect, it also seems to give each column a minimum length of 8 characters:

```
❯ echo -e "a\nb\nc\nd" | column -c 15
a
b
c
d

❯ echo -e "a\nb\nc\nd" | column -c 16
a       c
b       d

❯ echo -e "a\nb\nc\nd" | column -c 31
a       c
b       d

❯ echo -e "a\nb\nc\nd" | column -c 32
a       b       c       d
```

Can you guys confirm this behavior? I'm using the version of `column` available on macOS (last updated in 2004).

waldyrious:
[This page](https://www.howtoforge.com/linux-column-command/) has some useful examples that may help clarify the tool's behavior. It seems to me that they're using a more feature-rich version of `column`. Mine only allows the options `-c`, `-s`, `-t` and `-x`.

waldyrious:
Paging @kidpixo from the original PR which introduced this page, #1053, as well as @agnivade who participated in the discussion there.

kidpixo:
sorry, don't have currently time for this, it's in my TODO.

waldyrious:
No problem @kidpixo :) I was just letting you know in case you wanted to chime in.

sbrl:
Recorded a demo:

[asciicast](https://asciinema.org/a/245272)

waldyrious:
@sbrl thanks. Did I understand correctly that your tests replicated my results exactly? It would have been easier if you just said so :P

So I suppose this is really how `column` works everywhere. I'll open an issue for reviewing and clarifying this page. As for this PR, I believe its changes are correct regardless of the ambiguity in the other examples, so it isn't affected by these findings.

kidpixo:
I forgot to mention I'm also using Mac and the BSD programs version here differ often significantly from GNU version.

waldyrious:
> I'll open an issue for reviewing and clarifying this page.

Done in #3016.

mebeim:
I can confirm the behavior @waldyrious. That's such a strange tool to use, I got really confused while running those commands :')

waldyrious:
Thanks @mebeim :) Hopefully we can make this page better reflect the behavior model of the command so that it's not unexpected to our readers :)

sbrl:
I think so, but I wasn't exactly sure - hence the recording :P

PR #2999, opened 2019-05-09
ssh: Remove real domains to remain neutral

Lappro:
Real domains should not be used as to alleviate bias and prevent people accidentally running commands against real sites.


andrik:
Hi @Lappro, thank you for your submission.

Well noted, we should not use real domains on the examples.

LGTM

PR #3000, opened 2019-05-09
diff: clarified and elaborated examples

terminalnode:

The current version of the page doesn't do a great job of explaining how the order of the arguments affect the output, this PR tries to remedy that by changing all instances of `{{file1}}` to `{{old_file}}` and all instances of `{{file2}}` to `{{new_file}}` as well as elaborating on the explanations for `diff {{old_file}} {{new_file}}` and `diff -r {{directory1}} {{directory2}}`.

`{{directory1}}` and `{{directory2}}` retain their old names since diffs between directories is a little different. With just the `-r` option it does list differences in files similar to when called with no options but a lot of differences in file structure is just showed as is, for example with output such as "Only in directory2: file1".

The examples themselves are great and left unchanged.

### DIFF
--- a/pages/common/diff.md
+++ b/pages/common/diff.md
@@ -2,26 +2,26 @@
 
 > Compare files and directories.
 
-- Compare files:
+- Compare files (lists changes to turn `old_file` into `new_file`):
 
-`diff {{file1}} {{file2}}`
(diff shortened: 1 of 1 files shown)


bl-ue:
Yeah, great change.  Also it's PR 3,000 🥇 😃

PR #3001, opened 2019-05-09
mesg: add page

lbonanomi:

### DIFF
--- a/pages/linux/mesg.md
+++ b/pages/linux/mesg.md
@@ -1,7 +1,7 @@
 # mesg
 
 > Check or set a terminal's ability to receive messages from other users, usually from the write command.
-> See also `write`
+> See also `write`.
 
 - Check terminal's openness to write messages:
 
--- a/pages/linux/mesg.md
+++ b/pages/linux/mesg.md
@@ -1,6 +1,7 @@
 # mesg
 
 > Check or set a terminal's ability to receive messages from other users, usually from the write command.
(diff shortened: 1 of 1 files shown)


pages/linux/mesg.md
```diff
@@ -0,0 +1,15 @@
+# mesg
+
+> Check or set a terminal's ability to receive messages from other users, usually from the write command.
+
```
sbrl: Maybe we could add an additional like below the description line like this?

> See also `write`.

lbonanomi: Would a reciprocal link from write to mesg be appropriate?

sbrl: Good idea! Looks like you forget the trailing full stop though

lbonanomi: @sbrl I've spent 15 minutes wringing my hands about this: should a change to [write.md](https://github.com/tldr-pages/tldr/blob/master/pages/linux/write.md) be in a separate PR? I realize this is a silly concern, but I'm keen to get it right.

pxgamer: @lbonanomi, I would create it as a separate PR. :+1:

sbrl: Not a silly question, @lbonanomi! Generally, edits to each command should each have their own PRs. The exception here is if you're making the same change to many pages, such as in #2896 for example.

PR #3004, opened 2019-05-10
gox: add page

zQueal:
- [✔️] The page (if new), does not already exist in the repo.
- [✔️] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [✔️] The page has 8 or fewer examples.
- [✔️] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [✔️] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).

### DIFF
--- a/pages/common/gox.md
+++ b/pages/common/gox.md
@@ -14,6 +14,6 @@
 
 `gox -os="{{os}}"`
 
-- Compile current directory for `{{os}}` running `{{arch}}`:
+- Compile current directory for a single operating system and architecture combination:
 
 `gox -osarch="{{os}}/{{arch}}"`
--- a/pages/common/gox.md
+++ b/pages/common/gox.md
@@ -1,6 +1,7 @@
 # gox
 
 > A tool for cross-compiling Go programs.
+> Homepage: https://github.com/mitchellh/gox.
 
 - Compile Go program in the current directory for all operating systems and architecture combinations:
 
--- a/pages/common/gox.md
+++ b/pages/common/gox.md
@@ -1,6 +1,6 @@
 # gox
 
-> A dead simple, no frills Go cross compile tool
+> A dead simple, no frills Go cross compile tool.
 
 - Compile current Go directory for all OS and arch:
 
--- a/pages/common/gox.md
+++ b/pages/common/gox.md
@@ -10,7 +10,7 @@
 
 `gox {{url_1}} {{url_2}}`
 
-- Compile current directory for `{{os}}`:
+- Compile current directory for a particular operating system:
 
 `gox -os="{{os}}"`
 
--- a/pages/common/gox.md
+++ b/pages/common/gox.md
@@ -1,6 +1,6 @@
 # gox
 
-> A tool for Go cross compilation.
+> A tool for cross-compiling Go programs.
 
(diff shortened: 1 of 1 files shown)


schneiderl:
@Aracki you are right. We should always avoid adjectives as they almost sound like an ad. 
We should always try to be as neutral as possible.

zQueal:
> I would change the title to: `A tool for Go cross compilation`. I think the practice is to avoid adjectives such as `dead simple`, `ultra fast`, etc.
> 
> Beside that, PR LGTM.

Updated as per your suggestion.

pages/common/gox.md
```diff
@@ -0,0 +1,19 @@
+# gox
+
+> A tool for Go cross compilation.
```
sbrl: If I understand it right, maybe this would read better?

```suggestion
> A tool for cross-compiling Go programs.
```

pages/common/gox.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Compile current Go directory for all OS and arch:
```
sbrl: I assume Go programs are laid out in a specific directory structure?

Perhaps this will read better?
```suggestion
- Compile Go program in the current directory for all operating systems and architecture combinations:
```

pages/common/gox.md
```diff
@@ -0,0 +1,19 @@
...
+
+`gox`
+
+- Compile from URL:
```
sbrl: Perhaps this will read better?
```suggestion
- Download and compile a Go program from a remote URL:
```

pages/common/gox.md
```diff
@@ -0,0 +1,19 @@
...
+
+`gox {{url_1}} {{url_2}}`
+
+- Compile current directory for `{{os}}`:
```
sbrl: Perhaps this will read better?
```suggestion
- Compile current directory for a particular operating system:
```

pages/common/gox.md
```diff
@@ -0,0 +1,19 @@
...
+
+`gox -os="{{os}}"`
```
sbrl: Are you missing an extra dash there?

pages/common/gox.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Compile current directory for `{{os}}` running `{{arch}}`:
```
sbrl: Perhaps this will read better?
```suggestion
- Compile current directory for a single operating system and architecture combination:
```

pages/common/gox.md
zQueal: Negative. As per the `gox --help` a single dash is listed, but both seem to play nice. I added single dashes because that's how it's displayed in the help.

pages/common/gox.md
```diff
@@ -0,0 +1,19 @@
+# gox
+
+> A tool for cross-compiling Go programs.
```
agnivade: Please add the home page url in a new line after this.

sbrl: Like this:

```markdown
> Homepage: <https://example.com>.
```

Issue #2989, opened 2019-05-08
Add @pxgamer as an organisation owner


sbrl:
Hi, @pxgamer! You've been an active tldr-pages org member for over 6 months.
Thanks for sticking around this far and helping out!
According to [COMMUNITY-ROLES.md](https://github.com/tldr-pages/tldr/blob/master/COMMUNITY-ROLES.md), you're now eligible for becoming an owner in the organization.

That means you will, from now on, be part of the team responsible for performing role changes (like this one!) in the community.
Before performing such role transitions, make sure to review the [COMMUNITY-ROLES.md](https://github.com/tldr-pages/tldr/blob/master/COMMUNITY-ROLES.md) document.

Is that OK with you? Let us know!

Either way, thanks so much for all the work you've done so far. You rock!

pxgamer:
Hi @sbrl, thanks for the offer! I'm really enjoying being a part of the tldr pages community. 🎉

It would be great to continue participating as an owner. 👍

sbrl:
Role changed! You're now an owner:

![image](https://user-images.githubusercontent.com/9929737/57387645-a0f26880-71ae-11e9-8f2d-6eb050d173ed.png)

waldyrious:
> Feel free to close this issue :-)

@sbrl please don't induce him in error ;) the final step for this issue to be closed is a PR updating `MAINTAINERS.md`, as described [here](https://github.com/tldr-pages/tldr/blob/master/COMMUNITY-ROLES.md#adding-new-organization-owners).

pxgamer:
Whoops. 🤔 Opened #2997, just a quick question. Is the `[26 August 2018](https://github.com/tldr-pages/tldr/issues/2258) — present` meant to use this issue (and date) as the reference? Or the original issue.

waldyrious:
I've been thinking about that recently. Ideally we could have a all the dates preserved for historical record, but I haven't thought of a format that wouldn't be too verbose. Any ideas welcome :)

waldyrious:
Maybe, since we have a `from-to` date range in each entry, we could keep the previous entries in the corresponding sections, de-emphasised somehow (either using strikethrough or blockquote in the old entries, or by making the current entries bold). WDYT?

pxgamer:
Hmm. I'm not really that sure 🤔

sbrl:
Oops! I totally forgot about that. I really should double-check the instructions :P

And yeah, that sounds like a good idea. I like that!

[...]

PR #3006, opened 2019-05-10
MAINTAINERS: rework page to keep historical record

waldyrious:
See discussion starting [here](https://github.com/tldr-pages/tldr/issues/2989#issuecomment-490796319) for context and motivation.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,6 +1,8 @@
 # Maintainers
 
 This file contains a list of the maintainers of the tldr-pages project.
+Note: only the people marked with **bold** are currently in the indicated role.
+The other entries are kept for historical record.
 
 There are three types of maintainers, as described in
 [COMMUNITY-ROLES.md](https://github.com/tldr-pages/tldr/blob/master/COMMUNITY-ROLES.md#when-to-change-roles):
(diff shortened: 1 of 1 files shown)


waldyrious:
In a later PR I intend to reconstruct the past history of role changes in order to full in the log for all sections. Any assistance in doing so is welcome!

andrik:
@waldyrious what about a timeline?

I know markdown doesn't support it, but we can use an image to represent that.

schneiderl:
@andrik I'm not really in favor of a image as it creates a bunch of maintenance need. 

The changes proposed in this PR are, IMHO, the way to go. As it documents the past while keeping the current points of contact highligted.

waldyrious:
@andrik I'd much rather have the information in textual form (searchable, versionable, etc.) Not opposed to adding a timeline if there's a simple way to update it; otherwise, I agree with @schneiderl that it would be too burdensome to keep this page updated.

waldyrious:
> If only there was a way to grey things out like lo-colour in the terminal.....

Well, we could use blockquote, but I tested and couldn't find a way to play nicely with the list formatting.

Option 1, list items inside blockquote:
- foo
- bar
> - baz
> - qux

Option 2, blockquote inside list items:
- foo
- bar
- > baz
- > qux

waldyrious:
That said, we can look at some alternatives:

### Option 1: Bold and normal

This is the one currently implemented in this PR

- **Person One ([@personone]())**: [1 January 2001]() — present
- **Person Two ([@persontwo]())**: [2 February 2002]() — present
- Person Three ([@personthree]()): [3 March 2003]() — [13 March 2013]()
- Person Four ([@personfour]()): [4 April 2004]() — [14 March 2014]()

### Option 2: Blockquote without list

Person One ([@personone]()): [1 January 2001]() — present
Person Two ([@persontwo]()): [2 February 2002]() — present

> Person Three ([@personthree]()): [3 March 2003]() — [13 March 2013]()
> Person Four ([@personfour]()): [4 April 2004]() — [14 March 2014]()

### Option 3: Using the `<details>` element

- Person One ([@personone]()): [1 January 2001]() — present
- Person Two ([@persontwo]()): [2 February 2002]() — present

<details><summary>Past members</summary>

- Person Three ([@personthree]()): [3 March 2003]() — [13 March 2013]()
- Person Four ([@personfour]()): [4 April 2004]() — [14 March 2014]()
</details>

My opinion is that the first option is the simpler to implement and nicer-looking, but let me know what you guys think.

schneiderl:
@waldyrious Agree with you.

sbrl:
Hrm, I'm not sure about blockquote. The details looks good actually - I'm evenly split between bold and details.

agnivade:
No blockquote please. I am also evenly split between bold and details.

waldyrious:
@pxgamer, @andrik, what do you think?

pxgamer:
I think option 1 (bold) seems the best. It also will work across pretty much all markdown editors and previewers, whereas `details` may not. Having `details` is also maybe a little less neat in the source. And the spacing between markdown and HTML elements can sometimes cause issues.

However, if people prefer using details, that's fine with me.

mebeim:
I think option 1 is the best both for simplicity and compatibility.

andrik:
Hi guys,

Great work @waldyrious 

I vote for number one too.

waldyrious:
Great, so it looks like we have consensus. I'll go ahead and merge this to unblock @pxgamer's PR.

PR #3007, opened 2019-05-10
disown: add -h example

RH-sdavey:

### DIFF
--- a/pages/linux/disown.md
+++ b/pages/linux/disown.md
@@ -15,6 +15,6 @@
 
 `disown -a`
 
-- Keep job (do not disown it), but mark it with SIGHUP so its not removed on shell exit:
+- Keep job (do not disown it), but mark it so that no future SIGHUP is received on shell exit:
 
 `disown -h %{{job_number}}`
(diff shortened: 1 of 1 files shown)


pages/linux/disown.md
```diff
@@ -14,3 +14,7 @@
 - Disown all jobs:
 
 `disown -a`
+
+- Disown job from current shell (SIGHUP not sent to job on shell exit):
```
sbrl: How is this different from a regular disown? I'm not sure I understand.

RH-sdavey: Sorry, my wording was completely off (long Friday), 
the -h flag will not remove the job from the jobs table, but will rather keep the job in the job table but mark it so it does not receive any future SIGHUP signal from a shell exit. 
Wording has been changed now, hope its better, thanks

RH-sdavey: from bash man page/built in commands/
 If the -h option is given, each jobspec is not removed from the table, but is marked so that SIGHUP is not sent to the job if the shell receives a SIGHUP

sbrl: Ah, I see! No problem. The new wording looks much better :-)

Useful addition!

PR #3010, opened 2019-05-10
multiple pages: standardize domain examples

schneiderl:
This PR proposes the standardization of domain examples in tldr-pages. 

As most pages already use example.com as example to domains I have decided to keep that and change only the pages that do not follow that.

### DIFF
--- a/pages/common/irssi.md
+++ b/pages/common/irssi.md
@@ -4,7 +4,7 @@
 
 - Open irssi and connect to a server with a nickname:
 
-`irssi -n {{nickname}} -c {{server.example.com}}`
+`irssi -n {{nickname}} -c {{irc.example.com}}`
 
 - Open irssi and connect with a specific server on a given port:
 
--- a/pages/common/irssi.md
+++ b/pages/common/irssi.md
@@ -4,11 +4,11 @@
 
 - Open irssi and connect to a server with a nickname:
 
-`irssi -n {{nickname}} -c {{server.name.com}}`
+`irssi -n {{nickname}} -c {{server.example.com}}`
 
 - Open irssi and connect with a specific server on a given port:
 
-`irssi -c {{server.name.com}} -p {{port}}`
+`irssi -c {{server.example.com}} -p {{port}}`
 
 - View the help:
 
--- a/pages/common/irssi.md
+++ b/pages/common/irssi.md
@@ -8,7 +8,7 @@
 
 - Open irssi and connect with a specific server on a given port:
 
-`irssi -c {{server.example.com}} -p {{port}}`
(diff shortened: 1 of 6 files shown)


pages/common/irssi.md
```diff
@@ -4,11 +4,11 @@
 
 - Open irssi and connect to a server with a nickname:
 
-`irssi -n {{nickname}} -c {{server.name.com}}`
+`irssi -n {{nickname}} -c {{server.example.com}}`
```
waldyrious: Is there any particular reason to use a subdomain in this example?

schneiderl: I don't know the command but I believe the subdomain is needed because this is a IRC client. IRC servers will usually run under something like "irc.example.com".

schneiderl: Not that it is a must but it makes sense to have that subdomain there.

sbrl: Yeah, I'd vote to keep the subdomain here.

pxgamer:
@schneiderl, was the final commit (52c6db8ca3331ecf1f23a63d3ad7532d09d521c) intentional?

schneiderl:
> @schneiderl, was the final commit ([52c6db8](https://github.com/tldr-pages/tldr/commit/52c6db8ca3331ecf1f23a63d3ad7532d09d521c0)) intentional?

It wasn't haha. I was just testing some stuff.
Thank you for the heads up. I've reverted the changes there.

pages/common/irssi.md
waldyrious: I see, thanks for the clarifications. That said, what do you guys think about using `irc.example.com`, or `irc-server.example.com`?

sbrl: Hrm. Not a bad idea!

mebeim: I don't think it makes much difference, both `irc` or `server` (or even `ir-server`) look clear to me. Go with the one you like guys :)

sbrl:
Are we ready to merge?

agnivade:
Nice work @schneiderl  !

waldyrious:
> Are we ready to merge?

There hasn't been objection to changing `server.example.com` to `irc.example.com` as suggested [above](https://github.com/tldr-pages/tldr/pull/3010#discussion_r282967401), so I'd merge only after that change is done (or if there's agreement _not_ to make the change).

schneiderl:
Oh I haven't seen those comments. Well I'll make the change then. I agree that irc.example.com is easier to understand then server.example.com.

schneiderl:
Made the requested changes @waldyrious. Can you please re-review?

PR #3011, opened 2019-05-10
shutdown: add absolute time examples

paulgalow:

### DIFF
--- a/pages/osx/shutdown.md
+++ b/pages/osx/shutdown.md
@@ -22,6 +22,6 @@
 
 `shutdown -h {{1300}}`
 
-- Reboot on May 10th 2042 at 11:30 am:
+- Reboot on May 10th 2042 at 11:30 am (Input format: YYMMDDHHMM):
 
 `shutdown -r {{4205101130}}`
--- a/pages/osx/shutdown.md
+++ b/pages/osx/shutdown.md
@@ -18,10 +18,10 @@
 
 `shutdown -r +{{5}}`
 
-- Power off (halt) today at 23:00
+- Power off (halt) at 1:00 pm (Uses 24h clock):
 
-`shutdown -h {{2300}}`
+`shutdown -h {{1300}}`
 
-- Reboot on May 10th 2042 at 11:30
+- Reboot on May 10th 2042 at 11:30 am:
 
 `shutdown -r {{4205101130}}`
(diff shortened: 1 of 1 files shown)


pages/osx/shutdown.md
```diff
@@ -17,3 +17,11 @@
 - Reboot in 5 minutes:
 
 `shutdown -r +{{5}}`
+
+- Power off (halt) at 1:00 pm (Uses 24h clock):
+
+`shutdown -h {{1300}}`
+
+- Reboot on May 10th 2042 at 11:30 am:
```
schneiderl: IMHO it would be good to show the input format on the description so that the user does not have to "decode" the example tu understand the command.

paulgalow: That's a good point. I've now included it in the description. What do you think about that?

schneiderl: Looks good to me.

paulgalow:
Thanks. Weird isn't it? From `man shutdown`:

> Time is the time at which shutdown will bring the system down and may be the word now (indicating an immediate shutdown) or specify a future time in one of two formats: +number, or yymmddhhmm, where the year, month, and day may be defaulted to the current system values.  The first form brings the system down in number minutes and the second at the absolute time specified.

And:

> The hours and minutes in the second time format may be separated by a colon (``:'') for backward compatibility.

I'm interpreting that as: Using colons as separators is considered legacy / deprecated.

Aracki:
Squashed and merged! Thanks for the contribution @paulgalow 👍

sbrl:
It doesn't support ISO dates, does it?

paulgalow:
> It doesn't support ISO dates, does it?

I'm afraid it doesn't: 
`shutdown -r "2019-05-13 12:30:00"                   # shutdown: bad time format`
`shutdown -r 2019-05-13                                      # shutdown: bad time format`
`shutdown -r 20190513                                         # shutdown: bad time format`

sbrl:
No problem. What a total pain though! Someone really should have thought about that when implementing that command.

pxgamer:
@sbrl, yeah. 😋 I'm confused what happens when we get to the year `2100`, etc. Does it just assume that the shutdown is in the current century? 🤔

sbrl:
Good question! Maybe there'll be anothe y2k bug where servers around the world will all shutdown because the time specification was wrong :P

PR #3019, opened 2019-05-13
alias: add list all aliases

Aracki:
I have added a very useful command which will list all aliases defined.

### DIFF
--- a/pages/common/alias.md
+++ b/pages/common/alias.md
@@ -3,6 +3,10 @@
 > Creates aliases -- words that are replaced by a command string.
 > Aliases expire with the current shell session, unless they're defined in the shell's configuration file, e.g. `~/.bashrc`.
 
+- List all aliases:
+
+`alias`
+
 - Create a generic alias:
 
 `alias {{word}}="{{command}}"`
--- a/pages/common/alias.md
+++ b/pages/common/alias.md
@@ -19,10 +19,6 @@
 
 `unalias {{word}}`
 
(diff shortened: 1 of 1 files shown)


waldyrious:
> Just interested, does this do anything different to `alias -p`?

Actually, I don't have the `-p` option in my `alias` built-in command:

```
$ alias -p
alias: bad option: -p

$ man alias

BUILTIN(1)                BSD General Commands Manual               BUILTIN(1)
...
Commands marked ``No**'' under External do exist externally, but are implemented as scripts using a builtin command of the same name.

           Command       External    csh(1)    sh(1)
           ...           ...         ...       ...
           alias         No**        Yes       Yes
...
BSD                            February 23, 2005                           BSD
```

(For reference, I'm using zsh on macOS.)

pxgamer:
I use zsh too (Ubuntu and MacOS), just I'm at work so on Windows.

Aracki:
`alias -p` doesn't work on fish either.

pxgamer:
Perhaps we should remove `alias -p` and just keep this one then. 🤔

However, I don't have a huge range of shells to test on.

waldyrious:
I can test two more shells tonight at home. In the meantime, additional data from others would be nice.

Aracki:
I suggest we remove `alias -p` and leave `alias` as a first example.

sbrl:
It seems like `alias -p` is a bash-specific thing.

PR #3020, opened 2019-05-13
mas: add page

Aracki:
I have added [mas](https://github.com/mas-cli/mas) - a CLI tool for the Mac App Store.

### DIFF
--- a/pages/osx/mas.md
+++ b/pages/osx/mas.md
@@ -1,0 +1,24 @@
+# mas
+
+> Command line interface for the Mac App Store.
+> Homepage: <https://github.com/mas-cli/mas>.
+
+- Show all installed applications and their product identifiers:
+
+`mas list`
+
+- Search for an application with price included:
+
+`mas search {{application}} --price`
+
+- Install or update an application:
+
+`mas install {{product_identifier}}`
+
+- Install all pending updates:
+
+`mas upgrade`
+
+- Sign into the Mac App Store for the first time:
+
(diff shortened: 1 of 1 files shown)


pxgamer:
Hmm. Looks like this is the same command as [`brew-mas`](https://github.com/tldr-pages/tldr/blob/bbd3c4cbca2dd3fc9c2e09cd6fbf1912b41e1f6a/pages/osx/brew-mas.md) but I guess the page name is incorrect.

Wondering if we should keep this and remove the other, or rename the other and close this. 🤔

andrik:
I think the description made by @Aracki is unbiased

Maybe editing the existing page by adding new examples and updating the description would be better.

What do you think @Aracki ?

Aracki:
1. I think the title should be `mas` instead of `brew-mas`. 
2. The title description is better without adjectives like 'simple'.
3. `mas install app_name app_identifier` example is incorrect. `install` command is taking only one integer argument which is app ID(s) to install or how they call it `product_identifier`.

sbrl:
Yeah, I like yours better @Aracki.

Pages should always bear the name of the command executed. Does `brew-mas` work in the terminal? If so, then it can probably stay.

If not, then it needs renaming & editing I think

Aracki:
@sbrl `brew-mas` doesn't work. Maybe it's and old alias for that specific tool.

Aracki:
This is moved here #3032

Issue #3022, opened 2019-05-13
line.md : move to correct folder


schneiderl:
Hi,
I have just noticed that the [line.md](https://github.com/tldr-pages/tldr/blob/master/line.md) file is not where it is supposed to be. It is hanging around the project front page.

As I don't know which is the correct folder it should be under, I'm just opening this issue. 

@lbonanomi maybe you can help us here as you were the one who commited this.

lbonanomi:
*facepalm*

Under the Linux heading, please.

mebeim:
Oh well, this is awkward, LOL.

lbonanomi:
Certainly not my finest hour, that's for sure.

mebeim:
@lbonanomi no problem, that's not your fault at all, it was an oversight when reviewing the PR. I just thought it was fun.

schneiderl:
> Certainly not my finest hour, that's for sure.

Happens to all of us haha.

lbonanomi:
I feel like the collegial spirit of this org is making me a better person to work with.

PR #3023, opened 2019-05-13
line: move page

lbonanomi:
Relocating line.md to under pages/linux

This PR closes https://github.com/tldr-pages/tldr/issues/3022


sbrl:
So `line` is only available on Linux systems, and not on, say, macOS?

lbonanomi:
@sbrl I take your point, and put myself in the hands of the org staff for an editorial decision.

I took a cue to home `line` under the Linux directory because it features in the util-linux tarball. `line` doesn't appear on my MacBook, but *does* appear on the (stock) Solaris and AIX hosts at my disposal.

sbrl:
You have access to a solaris box? Awesome!

Yeah, the general rule of thumb is that if it's available on more than 1 platform, it belongs in `common/`.

This wasn't always the case though, but we changed our minds on it a while back (I think it was shortly after the new Windows platform was added).

lbonanomi:
@sbrl I'm pretty sure I can find some pages in the linux directory that appear in at least one of the UNIX-compliant OSes. Do you think this is worth an audit? As I'm sure the org owners know I have time to spare. ;)

sbrl:
Ah, that'd be great @lbonanomi if you've got some time! Any command that is supported more than 1 platform supported by tldr-pages (e.g. `linux` and `osx`) should be in `common/`.

lbonanomi:
I'm filling in [a table of commands that common to Linux, Solaris and MacOS](https://gist.github.com/lbonanomi/49668c6108e8764305e23185ec46369f) as the obvious targets present themselves. 

I don't mean to drag this out, but I thought a public start would be a good

sbrl:
Nice, @lbonanomi! I like your suggestions there to start with. Perhaps we can do all the _promote to `common`_ ones first in a single PR?

PR #3029, opened 2019-05-14
multiple italian pages: add homepages

schneiderl:
This PR proposes the copy of homepages from English pages to Italian translated pages. 
If this gets accepted I would be happy to do this for every other translated pages we have. 
(I would actually write a script for that instead of doing it by hand haha)

### DIFF
--- a/pages.it/common/7z.md
+++ b/pages.it/common/7z.md
@@ -1,4 +1,5 @@
 # 7z
+> Homepage: <https://www.7-zip.org/>.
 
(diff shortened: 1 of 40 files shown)


pxgamer:
😋 Was just about to review the `7z` one.

schneiderl:
@mebeim is homepage the proper term in italian?

mebeim:
@schneiderl yeah, "homepage" it's used in Italian too.

PR #3030, opened 2019-05-14
style-guide and PR template: add web link guidelines to checklists

### LINKED ISSUE #3027: URL of commands
I noticed that @schneiderl and @pxgamer added URL to a lot of pages, so I was wondering if it's a good idea to add one more check to the linter or just put one more checkbox on the PR message.

Ideas?

pxgamer:
Closes #3027

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,3 +7,4 @@
 - [ ] The page has 8 or fewer examples.
 - [ ] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
 - [ ] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
+- [ ] The page includes a web link (if applicable)
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,4 +7,4 @@
 - [ ] The page has 8 or fewer examples.
 - [ ] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
 - [ ] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
-- [ ] The page includes a web link in the main description (if applicable)
+- [ ] The page description includes a link to documentation or a homepage (if applicable).
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,4 +7,4 @@
 - [ ] The page has 8 or fewer examples.
 - [ ] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
 - [ ] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
-- [ ] The page includes a web link (if applicable)
+- [ ] The page includes a web link in the main description (if applicable)
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -11,6 +11,7 @@ The basic format of each page should match the following template:
 
 > Short, snappy description.
 > Preferably one line; two are acceptable if necessary.
+> Homepage: <https://example.com>.
 
 - Example description:
 


pxgamer:
@sbrl, yes. It doesn't have to be. :+1:

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,3 +7,4 @@
 - [ ] The page has 8 or fewer examples.
 - [ ] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
 - [ ] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
+- [ ] The page includes a web link (if applicable)
```
agnivade: I would further clarify that if it is a core linux util, it doesn't need a web link.

waldyrious: I actually wouldn't object to adding a link to a html rendering of the manpage (in a reasonably canonical location), or GNU manual, etc. for such commands.

waldyrious:
> It doesn't have to be.

Hmm, indeed this may not make that clear. Maybe we should try to make a decision on a standard label to use in all cases? (I was under the impression that we had talked about this before, but I couldn't track the discussion down.)

How about `More information: <https://example.com>.`? I think that covers both proper homepages, repositories, and documentation (including manpages).

sbrl:
> Hmm, indeed this may not make that clear. Maybe we should try to make a decision on a standard label to use in all cases? (I was under the impression that we had talked about this before, but I couldn't track the discussion down.)
> 
> How about `More information: <https://example.com>.`? I think that covers both proper homepages, repositories, and documentation (including manpages).

That's an interesting idea, but I rather like the differentiation between "this is the command's homepage" and "this is the CLI reference for this (sub)command" (e.g. `git merge`), and "this is the general documentation for this program".

agnivade:
I think having a free-form label text is fine. It depends on what the author wants to display in a tldr page.

.github/PULL_REQUEST_TEMPLATE.md
agnivade: I think linking to a man page would be pretty redundant. The whole reason why we started adding homepage links was that folks couldn't find further info about a command. But for core util commands, everything is just a `man <command>` away. Thus, I feel adding a homepage link for core linux utils does not make much sense.

waldyrious:
> That's an interesting idea, but I do rather like the differentiation between "this is the command's homepage" and "this is the CLI reference for this (sub)command" (e.g. git merge), and "this is the general documentation for this program".

Interestingly, that's precisely the decision I think we should make about what these links should serve for. Should they work as disambiguation to know the origin of the command (homepage)? Or as a way to get the non-tldr help content for this command (CLI reference documentation)? That decision would affect what the label should be, and it needs to be made because in many cases both are available.

I've just raised this question in #3033, please take a look there for some examples.

schneiderl:
I believe a "More Information:" label is the most generic way to go. The page author would be responsible for finding the most relevant page (that being the homepage or the CLI reference documentation).

andrik:
I agree with @schneiderl 

I think our main goal is the contributor add what he sees as the most relevant information.   

We could explore if it's better to use the project link or a specific one on a new page of best practices.

waldyrious:
> We could explore if it's better to use the project link or a specific one on a new page of best practices exploring some scenarios.

I couldn't understand what you meant here. Can you clarify?

andrik:
We could add one section to CONTRIBUTING.md maybe called Best Practices / Good and Bad Examples and provide two scenarios that embrace projects without a specific link of commands documentation, eg. [cmus](https://cmus.github.io/) and projects that have a specific link, eg. [python](https://docs.python.org/3/using/cmdline.html)

Emphasizing to use a specific link when appropriate

.github/PULL_REQUEST_TEMPLATE.md
mebeim: I agree with @agnivade here.

mebeim:
```suggestion
- [ ] The page includes a web link in the description (if applicable).
```

waldyrious: Good idea. Maybe "in the main description", just to be extra clear?

waldyrious:
Thanks for the clarification, @andrik. If we agree on a standard generic label, and use an example with a specific-looking format (e.g. `More information: <https://example.com/docs/cli.html`), then that extra section / page might not be needed. In fact, if our documentation is too verbose, people won't read it -- as a project with TLDR in the name, we ought to keep that in mind! 😄

.github/PULL_REQUEST_TEMPLATE.md
mebeim: Right, makes sense.

sbrl:
Good points. I think it depends on the command as to what the link is useful for. For example, a (relatively) obscure tool whose GitHub README both serves as a homepage and the CLI documentation all-in-one would probably use `Homepage:`.

However, a command like `git merge` would perhaps have `Documentation:` or `Reference:`, as a link to the specific CLI documentation for the (sub)command would be more useful in my opinion.

andrik:
I see @waldyrious 

How about `[ ] The page includes a web link (if applicable prefer using a specific link of documentation, (e.g. https://bundler.io/docs.html) instead of the main link of the project (e.g. https://bundler.io/)`

I know is a bit verbose but in this case, I think it's worth adding more details

mebeim:
@andrik that sounds waaaaay too long and complicated for a single checklist item (also nested parentheses, lol?). I would advise keeping it simpler.

mebeim:
I would probably go with something along the lines of:

    [ ] The page description includes a link to documentation or a homepage (if applicable).

andrik:
Hi guys, I think the suggestion of @mebeim is great. 

You agree?

waldyrious:
> Hi guys, I think the suggestion of @mebeim is great.
>
> Do you agree?

I'd say the thumbs-up reactions in his comment are a good indicator of that :)

andrik:
> > Hi guys, I think the suggestion of @mebeim is great.
> > Do you agree?
> 
> I'd say the thumbs-up reactions in his comment are a good indicator of that :)

Yes :)

I forgot to ping others @schneiderl @pxgamer @agnivade

agnivade:
Any comments to my concern raised here https://github.com/tldr-pages/tldr/pull/3030#discussion_r284037257 ?

.github/PULL_REQUEST_TEMPLATE.md
waldyrious: @agnivade if nothing else is available for a given tool, would you prefer having no link at all instead of having a link to the manpage?

agnivade: Yes, only if the tool is a core linux util. Having links to manpages for `cp`, `mv` is pretty redundant IMO.

Aracki:
Maybe the `(if applicable)` part will be enough for people to realize they don't need to add it if it's a Linux core util or something similar...?

pxgamer:
Sorry for the delay, I've applied @mebeim's suggestion (definitely sounds better).

mebeim:
@agnivade while I agree about coreutils not needing a link, I don't really know how that could be expressed in a clear and concise manner given the current phrasing. I would rather have that added to the contribution guidelines or the style guide along with all the other more specific details.

andrik:
I agree with @mebeim. Maybe it's better to address these cases during reviews

agnivade:
Sounds good. @mebeim will you take that up ?

.github/PULL_REQUEST_TEMPLATE.md
waldyrious: Well, remember that tldr pages are not necessarily consumed on a command line. There's ostera's web client, the duckduckgo integration, the pages at distrowatch, mobile clients, etc. I think _not including any link at all_ based on the assumption that the user is looking at the tldr page in a command line is doing a disservice to users. Of course, if better links exist, I fully suport using them other than the manpage.

In other words, in my view every page should have a link of some sort, even if just to confirm the provenance of the tool.

mebeim:
@agnivade sure, I'll make a PR later today.

Merging this then since we all agree on the current version.

.github/PULL_REQUEST_TEMPLATE.md
agnivade: I get your point. You want the user to readily have a link to the full documentation. I prefer to have the links only in cases where it is not dead obvious where to look for documentation. It doesn't matter whether the user is in command line or not, they can always search for "man cp" and get the manpage.

That's just my opinion. I will leave it to the community to come to a decision.

mebeim:
Giving this another thought though... I kind of lost track about who prefers what. I think we need to make a clear decision as to whether or not links should be added for *all* pages, so I'll probably open an Issue for that first. Once we make a clear decision I'll make a PR. 

What do you guys think? @agnivade @waldyrious @pxgamer @sbrl @andrik @Aracki

Aracki:
I am pretty sure the vast majority of future tldr pages will be open-source applications for which is really helpful to have links inside them.

The issue is what to do with ones which have only man pages and doesn't make a sense to put the link at all.

agnivade:
Yeah, my point was only regarding core linux util pages. We can open a separate issue, but I have already mentioned my reasons here https://github.com/tldr-pages/tldr/pull/3030#discussion_r285334966. So maybe just link that thread in the issue.

sbrl:
> The issue is what to do with ones which have only man pages

Link to an online man page :-)

.github/PULL_REQUEST_TEMPLATE.md
waldyrious: Fair enough. My position is that it's better to err on the side of offering unneeded help than not offering needed help. But let's flesh this out at #3041.

waldyrious:
> Giving this another thought though... I kind of lost track about who prefers what. I think we need to make a clear decision as to whether or not links should be added for all pages

@mebeim there was also ongoing discussion on what label to use with the link, both here and in #3033. Note that there's no hurry to merge PRs -- so please make sure a decision is reached (and all conversations are resolved) before doing so :)

mebeim:
@waldyrious you're right that's a bit to much info in this thread, I thought yours was just a reminder here. That in fact seems like another decision which deserves its own Issue, doesn't it? Would you be able to create one for that, since you're more up to date than me on this topic? 

After solving both that and #3041 I think we'll have a pretty clear and ready to implement guideline about links.

PR #3032, opened 2019-05-14
brew-mas: rename and edit page

Aracki:
As we discussed in #3020, `brew-mas` needs to be renamed to `mas`. 

I have changed the examples. `mas install` was incorrect and it's fixed now.


pages/osx/mas.md
```diff
@@ -0,0 +1,24 @@
+# mas
+
+> Command line interface for the Mac App Store.
+> Homepage: <https://github.com/mas-cli/mas>.
+
+- Show all installed applications and their product identifiers:
+
+`mas list`
+
+- Search for an application with price included:
```
sbrl: Does this include the price in the search criteria, or does it display the price alongside the results?

pages/osx/mas.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Install or update an application:
+
+`mas install {{product_identifier}}`
+
+- Install all pending updates:
+
+`mas upgrade`
+
+- Sign into the Mac App Store for the first time:
```
sbrl: Perhaps this should be the first example?

pages/osx/mas.md
Aracki: Display price alongside the results.

pages/osx/mas.md
Aracki: It sounds good.

pages/osx/mas.md
sbrl: Right. What about this then?

```suggestion
- Search for an application, displaying the price alongside the results:
```

Aracki:
What should I do regarding the conflict about `brew-mas.md` removed file?

PR #3033, opened 2019-05-15
multiple pages: add homepages

pxgamer:
This can be rebased and merged.

### DIFF
--- a/pages/common/salt-call.md
+++ b/pages/common/salt-call.md
@@ -1,6 +1,7 @@
 # salt-key
 
 > Invoke salt locally on a salt minion.
+> Homepage: <https://www.saltstack.com>.
 
 - Perform a highstate on this minion:
 
--- a/pages/common/salt-call.md
+++ b/pages/common/salt-call.md
@@ -1,7 +1,7 @@
 # salt-call
 
 > Invoke salt locally on a salt minion.
-> Homepage: <https://docs.saltstack.com/ref/cli/salt-call.html>.
+> More information: <https://docs.saltstack.com/ref/cli/salt-call.html>.
 
 - Perform a highstate on this minion:
 
--- a/pages/common/salt-call.md
+++ b/pages/common/salt-call.md
@@ -1,7 +1,7 @@
-# salt-key
+# salt-call
 
 > Invoke salt locally on a salt minion.
-> Homepage: <https://www.saltstack.com>.
+> Homepage: <https://docs.saltstack.com/ref/cli/salt-call.html>.
 
 - Perform a highstate on this minion:
 
--- a/pages/common/salt-key.md
+++ b/pages/common/salt-key.md
@@ -2,7 +2,7 @@
 
 > Manages salt minion keys on the salt master.
 > Needs to be run on the salt master, likely as root or with sudo.
-> Homepage: <https://www.saltstack.com>.
+> Homepage: <https://docs.saltstack.com/ref/cli/salt-key.html>.
 
 - List all accepted, unaccepted and rejected minion keys:
 
--- a/pages/common/salt-key.md
+++ b/pages/common/salt-key.md
@@ -2,7 +2,7 @@
 
 > Manages salt minion keys on the salt master.
 > Needs to be run on the salt master, likely as root or with sudo.
-> Homepage: <https://docs.saltstack.com/ref/cli/salt-key.html>.
+> More information: <https://docs.saltstack.com/ref/cli/salt-key.html>.
 
 - List all accepted, unaccepted and rejected minion keys:
 
--- a/pages/common/salt-key.md
+++ b/pages/common/salt-key.md
@@ -2,6 +2,7 @@
 
 > Manages salt minion keys on the salt master.
 > Needs to be run on the salt master, likely as root or with sudo.
+> Homepage: <https://www.saltstack.com>.
 
 - List all accepted, unaccepted and rejected minion keys:
 
--- a/pages/common/salt-run.md
+++ b/pages/common/salt-run.md
@@ -1,7 +1,7 @@
 # salt-run
 
 > Frontend for executing salt-runners on minions.
-> Homepage: <https://docs.saltstack.com/ref/cli/salt-run.html>.
+> More information: <https://docs.saltstack.com/ref/cli/salt-run.html>.
 
 - Show status of all minions:
 
--- a/pages/common/salt-run.md
+++ b/pages/common/salt-run.md
@@ -1,6 +1,7 @@
 # salt-run
 
 > Frontend for executing salt-runners on minions.
+> Homepage: <https://www.saltstack.com>.
 
 - Show status of all minions:
 
--- a/pages/common/salt-run.md
+++ b/pages/common/salt-run.md
@@ -1,7 +1,7 @@
 # salt-run
 
 > Frontend for executing salt-runners on minions.
-> Homepage: <https://www.saltstack.com>.
+> Homepage: <https://docs.saltstack.com/ref/cli/salt-run.html>.
 
 - Show status of all minions:
 
--- a/pages/common/salt.md
+++ b/pages/common/salt.md
@@ -1,6 +1,7 @@
 # salt
 
 > Execute commands and assert state on remote salt minions.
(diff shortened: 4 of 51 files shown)


pages/common/salt-key.md
```diff
@@ -2,6 +2,7 @@
 
 > Manages salt minion keys on the salt master.
 > Needs to be run on the salt master, likely as root or with sudo.
+> Homepage: <https://www.saltstack.com>.
```
waldyrious:
```suggestion
> Homepage: <https://docs.saltstack.com/en/latest/ref/cli/salt-key.html>.
```

pxgamer: I'll update this for `salt-call`, `salt-key` and `salt-run`.

Just wondering, should it use https://docs.saltstack.com/ref/cli/salt-key.html or include the `/en/latest` in the URL (which is automatically redirected).

pages/common/salt-call.md
```diff
@@ -1,6 +1,7 @@
 # salt-key
```
waldyrious: I realize this is not in the scope of this PR, but please fix the command name to match the page (file) name. You could do that as a separate commit.

As for the other commits, I'm not sure it's justified to have one commit per page, since they're all doing the same thing. What do you think about squashing them into one?

pages/common/salt-call.md
```diff
@@ -1,6 +1,7 @@
...
 
 > Invoke salt locally on a salt minion.
+> Homepage: <https://www.saltstack.com>.
```
waldyrious:
```suggestion
> Homepage: <https://docs.saltstack.com/en/latest/ref/cli/salt-call.html>.
```

pages/common/salt-run.md
```diff
@@ -1,6 +1,7 @@
 # salt-run
 
 > Frontend for executing salt-runners on minions.
+> Homepage: <https://www.saltstack.com>.
```
waldyrious:
```suggestion
> Homepage: <https://docs.saltstack.com/en/latest/ref/cli/salt-run.html>.
```

pages/common/salt.md
```diff
@@ -1,6 +1,7 @@
 # salt
 
 > Execute commands and assert state on remote salt minions.
+> Homepage: <https://www.saltstack.com>.
```
waldyrious:
```suggestion
> Homepage: <https://docs.saltstack.com/en/latest/ref/cli/salt.html>.
```

pages/common/sass.md
```diff
@@ -1,6 +1,7 @@
 # sass
 
 > Converts SCSS or Sass files to CSS.
+> Homepage: <https://sass-lang.com>.
```
waldyrious: According to https://sass-lang.com/documentation/cli, there are two CLIs for Sass: `dart-sass` and `ruby-sass`, both being invoked with simply `sass`. Since the ruby CLI is deprecated, I'd say we should link directly to the dart-sass page:

```suggestion
> Homepage: <https://sass-lang.com/documentation/cli/dart-sass>.
```

pages/common/salt-key.md
waldyrious: If they're equivalent, I'd go with the shorter URL.

pages/common/salt-call.md
waldyrious: Do you object to combining the link-adding commits?

waldyrious:
Please don't update a PR so quickly after comments are added, especially for such a big PR :) I was still adding comments, and some of the changes may warrant discussion first. (Yeah, I'm saying you're _too_ efficient 😄)

pxgamer:
Sorry. 😋 Will wait for the review to finish. 👍

pages/common/salt-call.md
pxgamer: Combining all of the commits? Or just the `salt` ones? 🙂

pages/common/sass.md
```diff
@@ -1,6 +1,7 @@
 # sass
 
 > Converts SCSS or Sass files to CSS.
+> Homepage: <https://sass-lang.com/documentation/cli/dart-sass>.
```
waldyrious: I'm not sure we'll want to do this (link to the CLI documentation vs. the project's homepage) on every case. I've added a comment in the main thread proposing a discussion about that.

waldyrious:
Ok, I have no further comments to add for the time being. Next time I'll use actual review comments so they're all submitted at the same time :+1:.

agnivade:
> I don't have a strong opinion on this, and I would like others to chime in so we can make a decision.

I think it depends on the tool. Mostly I would prefer a link to the CLI documentation. But some tools have a single page for overall description and the CLI documentation. So in those cases, maybe homepage is also fine.

waldyrious:
> some tools have a single page for overall description and the CLI documentation. So in those cases, maybe homepage is also fine.

Definitely, I fully agree. Even if no documentation for the CLI usage is not available, I'd support adding the home page if there's one. My goal here was to make sure we agree on a preference ranking for the various types of links we could add.

pages/common/salt-call.md
waldyrious: I was thinking all the ones that add links to the pages, so there would be one commit with the adding of links (with the list of pages in the extended commit message, like this one: ce642b6c1), and one commit with the other fixes (page title and whitespace). But that's my opinion (I think semantically they're one "single" change); if you disagree, let me know.

pages/common/salt-key.md
sbrl: Maybe this should be `Documentation:`, since it links to the CLI reference?

sbrl:
Here's my order of preference when I search for links:

 - Specific CLI (sub)command reference
 - General program documentation
 - Program homepage
 - Online man page

waldyrious:
I agree with that list @sbrl. My suggestion (and in response to [your comment above](https://github.com/tldr-pages/tldr/pull/3033#discussion_r284484185)) is that all of those would be well described with a "More information" label, as I suggested in [#3030](https://github.com/tldr-pages/tldr/pull/3030#issuecomment-492405898)

pages/common/salt-key.md
waldyrious: @sbrl replied [below](https://github.com/tldr-pages/tldr/pull/3033#issuecomment-493329994).

sbrl:
Hrm, true. I still like `Documentation:`, `Homepage:`, `Reference:`, etc. though as I feel they are more specific.

waldyrious:
Of course they are more specific :) but the generic one works perfectly well IMO, and is way easier to document (and for page authors to figure the pattern from, in case they don't read the documentation, which, let's be honest, is often :P).

pxgamer:
I'd support `More information` (d'oh. Need to re-do my multi-page PR). Even though it is less specific, it works in all contexts, is easy to read, and makes it easier for people to contribute as they don't have to decide between the wording.

sbrl:
>  in case they don't read the documentation, which, let's be honest, is often

you don't? I spend some time looking and reading to determine what each URL I link to is. In that case, maybe `More Information` is better, but I think this is one of those things that goes over my head (with respect to ASD).

pxgamer:
@sbrl, I assumed @waldyrious was meaning people not reading our docs (style guide, etc.)

waldyrious:
From the discussion above and at #3030, it looks like we can proceed with `More information:` as the label for links. (If anyone disagrees, we can create an issue similar to #3041 to make a decision on that.)

pxgamer:
@waldyrious, should I re-commit all of these and force push? Or just do a single commit that updates all of them at once?

waldyrious:
I would go with a single commit (for the link-adding changes, as discussed [above](https://github.com/tldr-pages/tldr/pull/3033#discussion_r284171627)), but that's my preference. Feel free to do multiple commits if that works better for you.

sbrl:
I still disagree the reasons I've mentioned already, but we can do it the way everyone else prefers.

pxgamer:
Is this ok to go @sbrl, @waldyrious? 🤔

sbrl:
Go for it!

pxgamer:
Not a fan of merging my own PRs 👍

PR #3037, opened 2019-05-17
lslocks: add page

RH-sdavey:
Part of #2214 

### DIFF
--- a/pages/linux/lslocks.md
+++ b/pages/linux/lslocks.md
@@ -1,6 +1,6 @@
 # lslocks
 
-> List all local system locks.
+> List local system locks.
 
 - List all local system locks:
 
--- a/pages/linux/lslocks.md
+++ b/pages/linux/lslocks.md
@@ -8,7 +8,7 @@
 
 - List locks with defined column headers:
 
-`lslocks --output PID,COMMAND,PATH`
+`lslocks --output {{PID}},{{COMMAND}},{{PATH}}`
 
 - List locks producing a raw output (no columns), and without column headers:
 
--- a/pages/linux/lslocks.md
+++ b/pages/linux/lslocks.md
@@ -12,7 +12,7 @@
 
 - List locks raw output (no columns), and no column headers:
 
-`lslocks -r -n`
+`lslocks --raw --noheadings`
 
 - Populate a csv file with the information from above command:
 
--- a/pages/linux/lslocks.md
+++ b/pages/linux/lslocks.md
@@ -8,7 +8,7 @@
 
 - List locks with defined column headers:
 
-`lslocks -o PID,COMMAND,PATH`
+`lslocks --output PID,COMMAND,PATH`
 
 - List locks raw output (no columns), and no column headers:
 
--- a/pages/linux/lslocks.md
+++ b/pages/linux/lslocks.md
@@ -1,0 +1,19 @@
+# lslocks
+
+> List all local system locks.
+
+- List all local system locks:
+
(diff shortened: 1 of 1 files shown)


pages/linux/lslocks.md
```diff
@@ -0,0 +1,19 @@
...
+
+> List all local system locks.
+
+- List all local system locks:
+
+`lslocks`
+
+- List locks with defined column headers:
+
+`lslocks -o PID,COMMAND,PATH`
```
mebeim:
```suggestion
`lslocks --output PID,COMMAND,PATH`
```
Long option names are to be preferred over short ones.

pages/linux/lslocks.md
```diff
@@ -0,0 +1,19 @@
...
+
+- List locks raw output (no columns), and no column headers:
+
+`lslocks -r -n`
```
mebeim:
```suggestion
`lslocks --raw --noheadings`
```
Long option names are to be preferred over short ones.

pages/linux/lslocks.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Populate a csv file with the information from above command:
```
mebeim: I don't think this example is needed. After all it's just the same as the previous one with an added pipe which only replaces spaces with commas (also, a CSV doesn't realy *need* to be comma-delimited, CSVs can be space-delimited too). It does not really show different features of the `lslocks` command.

I would suggest adding examples for the `--pid` and `--json` options instead, which are not mentioned. If you don't feel like or don't have the time to add examples for those too, that's fine they can be added in the future, but I wouldn't populate the page with more complex examples without having all the basic functionality covered first.

pages/linux/lslocks.md
```diff
@@ -0,0 +1,19 @@
+# lslocks
+
+> List all local system locks.
```
mebeim:
```suggestion
> List local system locks.
```
I'd suggest removing "all" here, the command has more functionalities.

pages/linux/lslocks.md
```diff
@@ -0,0 +1,19 @@
...
+
+- List all local system locks:
+
+`lslocks`
+
+- List locks with defined column headers:
+
+`lslocks -o PID,COMMAND,PATH`
+
+- List locks raw output (no columns), and no column headers:
```
mebeim:
```suggestion
- List locks producing a raw output (no columns), and without column headers:
```
Looks clearer, don't you think?

RH-sdavey: Yes, long friday and brain wasn't working 👍

pages/linux/lslocks.md
RH-sdavey: Ok, will remove this example ( I only added that as I used it myself quite recently ) and add the `--pid` and `--json` examples after work this evening. Thanks for feedback 👍

pages/linux/lslocks.md
```diff
@@ -0,0 +1,23 @@
...
+
+> List local system locks.
+
+- List all local system locks:
+
+`lslocks`
+
+- List locks with defined column headers:
+
+`lslocks --output PID,COMMAND,PATH`
```
agnivade: Should PID,COMMAND,PATH be within tokens (`{{}}`) ?

mebeim: You're right, it should.

RH-sdavey:
Changes made, thanks for merge, will add more updates/new pages later this week.

PR #3040, opened 2019-05-18
dexter: add page

Aracki:
I have added [dexter](https://github.com/gini/dexter/) - a helper tool for authenticating kubectl users with OpenID Connect.


Aracki:
@sbrl 👍

PR #3042, opened 2019-05-19
git-remote: mention how to add a new URL

waldyrious:

### DIFF
--- a/pages/common/git-remote.md
+++ b/pages/common/git-remote.md
@@ -11,7 +11,7 @@
 
 `git remote add {{remote_name}} {{remote_url}}`
 
-- Change the URL of a remote:
+- Change the URL of a remote (use `--add` to keep the existing URL):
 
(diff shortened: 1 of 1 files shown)


sbrl:
Great addition, @waldyrious! If you add a new url whilst keeping the existing one, does it pick one at random or something?

pxgamer:
@sbrl, I'm pretty sure it allows you to push to multiple repositories at once. 👍

waldyrious:
> If you add a new url whilst keeping the existing one, does it pick one at random or something?

To be honest I don't recall exactly what algorithm git uses to pick the remote URL to act on. It could prefer git-protocol URLs to https ones, or use the oldest one first, etc. I had this change lying around but it's been a while since I actually used this feature, so I'm not sure. Maybe if I figure it out I'll add a comment here for future reference :)

sbrl:
Ah, that's very interesting! I didn't know that git could do that.

PR #3047, opened 2019-05-22
arch, bzip2, command, envsubst, mesg, type, ulimit, umask, write: move to common/

lbonanomi:

These changes are an effort to address https://github.com/tldr-pages/tldr/pull/3023#issuecomment-493616339

Only pages where all examples tested and shown working in the below environments are included:

"Linux" commands are tested on RHEL 7.6
"Solaris" commands are tested on Solaris 11.3 (SPARC)
"MacOS" commands are tested on MacOS 10.14.4


lbonanomi:
@mebeim: I sincerely appreciate your rigor, I know that my enthusiasm can go running past my judgement. 

I ran through all of the examples on the pages in this PR on a Solaris host that I'm not root on and didn't build, so there is *absolutely* a chance that a local customization I'm not aware of is souring my tests.

`arch`: accepts no switches
`bzip2`: No variation across OSes running tests
`command`: page demonstrates no switches
`envsubst` pages demonstrates no switches
`mesg`: Solaris allows for '-y' and '-n' as well as the demonstrated 'y' and 'n' arguments, but I see no value in documenting this variation, do you?
`type`: You're right about the man page, but please see below code block and advise?
`ulimit`: No variation found across OSes running tests
`umask`: No variation found across OSes running tests
`write`: No variation found across OSes running tests

```bash
$ uname -vm
11.3 sun4v
$ /bin/type --help
Usage: type [ options ] name  ...
OPTIONS
  -a              Displays all uses for each name rather than the first.
  -f              Do not check for functions.
  -p              Do not check to see if name is a reserved word, a built-in, an alias, or a function. This turns off the -v option.
  -q              Quiet mode. Returns 0 if all arguments are built-ins, functions, or are programs found on the path.
  -v              For each name you specify, the shell displays a line that indicates if that name is one of the following:
                          Reserved word
                          Alias
                          Built-in
                          Undefined function
                          Function
                          Tracked alias
                          Program
```

mebeim:
@lbonanomi thank you for clarifying this in such detail. 

Regarding `mesg`,: I don't see value in that either, it's fine as is IMHO.

Regardint `type`: well there it goes another example of a misleading man-page... no options shown, but the command actually supports them. That's unfortunate.

Wit the above said, I agree with you that this PR looks fine and for what I can see these commands belong to `common/`. I'm probably going to test these commands in a VM and let you know.

PR #3048, opened 2019-05-23
chore: fix use doublequotes and $(..) notation in deploy script

mstruebing:
see: https://github.com/koalaman/shellcheck/wiki/SC2006 and https://github.com/koalaman/shellcheck/wiki/SC2086

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -5,7 +5,7 @@ set -ex
 
 function initialize {
   if [ -z "$TLDRHOME" ]; then
-    export TLDRHOME=${TRAVIS_BUILD_DIR:-`pwd`}
+    export TLDRHOME=${TRAVIS_BUILD_DIR:-$(pwd)}
   fi
   export TLDR_ARCHIVE="tldr.zip"
   export SITE_HOME="$HOME/site"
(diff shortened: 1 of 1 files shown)


agnivade:
Did you actually run `shellcheck` and these were the only errors ? If not, we should fix any others too.

mstruebing:
There was only one left: `$encrypted_973441be79af_key` is used but not defined but I assume this comes somewhere magically, maybe from travis.

agnivade:
Yes.

waldyrious:
Thanks @mstruebing. Maintenance cleanup like this is often overlooked and considered somewhat unglamorous, but truly pays off in the long run.

sbrl:
Great, thanks! I wonder if it passes shellcheck now?

pxgamer:
@sbrl, looks like it passes except for the Travis variable that @mstruebing mentioned (which is fine). 👍
```
Line 23:
  openssl aes-256-cbc -K "$encrypted_973441be79af_key" -iv "$encrypted_973441be79af_iv" -in ./scripts/id_ed25519_tldr_asset_upload.enc -out id_ed25519 -d
                          ^-- SC2154: encrypted_973441be79af_key is referenced but not assigned.
>>                                                          ^-- SC2154: encrypted_973441be79af_iv is referenced but not assigned.
```

sbrl:
Ah, awesome! I should have read it more closely :P

PR #3050, opened 2019-05-23
sw_vers: improve descriptions, add basic example

waldyrious:


waldyrious:
Thanks @sbrl. It is not a hard rule that can be applied indiscriminately, as it can both remove or [create ambiguity](https://en.wikipedia.org/wiki/Serial_comma#Creating_ambiguity), but in this case the use is fine, so thanks for doing it!

PR #3052, opened 2019-05-25
kubectl: add explain command example

spachava753:

### DIFF
--- a/pages/common/kubectl.md
+++ b/pages/common/kubectl.md
@@ -32,4 +32,4 @@
 
 - Display an explanation of the fields shown:
 
-`kubectl explain deployment`
+`kubectl explain pods.spec.containers`
--- a/pages/common/kubectl.md
+++ b/pages/common/kubectl.md
@@ -30,10 +30,6 @@
 
 `kubectl cluster-info`
 
-- For getting explanations of fields to create a yaml file
+- For getting explanations of fields:
 
 `kubectl explain deployment`
-
-- To display all of the explanations recursively
-
-`kubectl explain deployment --recursive`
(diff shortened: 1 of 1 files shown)


pages/common/kubectl.md
```diff
@@ -29,3 +29,7 @@
 - Print the address of the master and cluster services:
 
 `kubectl cluster-info`
+
+- For getting explanations of fields:
```
sbrl: Thanks for the edit, @spachava753! Perhaps this will read better?
```suggestion
- Display an explanation of the fields shown:
```

Aracki: I would say: 
`- Get the documentation of the resource and its fields`

Aracki: In this specific example, you will get an output of **_deployment_** documentation. 

For example, if you run: `kubectl explain pods.spec.containers`, then you will get the explanation of that specific field of the resource (`.containers` field)

Aracki: I would suggest:

`- Get the documentation of the resource and its fields:`

pages/common/kubectl.md
```diff
@@ -29,3 +29,7 @@
 - Print the address of the master and cluster services:
 
 `kubectl cluster-info`
+
+- Display an explanation of the fields shown:
```
Aracki: What about `Display an explanation of the specific field:` or `Display an explanation of the specific resource's field:`

pages/common/kubectl.md
sbrl: Ah, I see @Aracki. What about this instead then?

```suggestion
- Show the documentation of the resource and its fields:
```

sbrl: Erm.... never mind. Looks like it's already been changed below.

PR #3060, opened 2019-05-29
style guide: update wording of the page link

pxgamer:

### DIFF
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -11,7 +11,7 @@ The basic format of each page should match the following template:
 
 > Short, snappy description.
 > Preferably one line; two are acceptable if necessary.
-> Homepage: <https://example.com>.
+> More information: <https://example.com>.
 
 - Example description:
 


pxgamer:
I think that's what we ended up with. Correct me if I'm wrong though...

mebeim:
@pxgamer I'm not that much up to date about the topic, but I remember seeing the majority of people agree on `More information`, so I think it's ok (?)

schneiderl:
I think that ```More Information``` is the best option in this case. 

Let's just wait for others.

mebeim:
While I agree that other words might be more suitable to more precisely describe the linked resource, there's two reasons why I'd say `More information` is the way to go:

1. It's generic, so we don't have to define which link corresponds to which keyword. It is not that simple to draw the line between terms like `Documentation`, `Reference`, `Manual`, and so on. It can get confusing both for the users and the contributors.
2. It's still as easy to understand as any other keyword, and gets the job done without needing to think much about it: "you want more information? Then go here <link>". We don't really need to precisely describe each link.

sbrl:
@mebeim Very true. Here I think it's very much my high-functioning Autism speaking: "but it needs to be _precise_!". Let's go ahead with this change.

PR #3061, opened 2019-05-29
docker images: add page

Aracki:
I have added `docker images` missing command.

### DIFF
--- a/pages/common/docker-images.md
+++ b/pages/common/docker-images.md
@@ -7,11 +7,11 @@
 
 `docker images`
 
-- List all Docker images with intermediate included:
+- List all Docker images including intermediates:
 
 `docker images -a`
 
-- List all Docker images numeric IDs:
+- List the output in quiet mode (only numeric IDs):
 
 `docker images -q`
 
--- a/pages/common/docker-images.md
+++ b/pages/common/docker-images.md
@@ -3,18 +3,18 @@
 > Manage Docker images.
 > More information: <https://docs.docker.com/engine/reference/commandline/images/>.
 
-- List all docker images:
+- List all Docker images:
 
 `docker images`
 
-- List all docker images with intermediate included:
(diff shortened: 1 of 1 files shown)


pages/common/docker-images.md
```diff
@@ -0,0 +1,20 @@
...
+
+- List all Docker images with intermediate included:
+
+`docker images -a`
+
+- List all Docker images numeric IDs:
+
+`docker images -q`
+
+- List all Docker images not used by any container:
```
sbrl: Is there any way of deleting these automatically with this command?

Aracki: No. For that one would use `docker rmi`.

sbrl: Right. That  would be a separate page then.

pages/common/docker-images.md
```diff
@@ -0,0 +1,20 @@
+# docker images
+
+> Manage Docker images.
+> More information: <https://docs.docker.com/engine/reference/commandline/images/>.
+
+- List all Docker images:
+
+`docker images`
+
+- List all Docker images with intermediate included:
```
agnivade: intermediates ? Maybe better is - List all Docker images including intermediates.

pages/common/docker-images.md
```diff
@@ -0,0 +1,20 @@
...
+
+`docker images -a`
+
+- List all Docker images numeric IDs:
```
agnivade: We should try to match the `-q` with the description. Let us change it to - "List the output in quiet mode (only numeric IDs)"

Aracki: good suggestion 👍

pages/common/docker-images.md
Aracki: there are many subcommands missing in **_tldr_**, but slowly we are getting there :)

PR #3062, opened 2019-05-29
docker container: add page

Aracki:
I have added `docker container` missing command.

### DIFF
--- a/pages/common/docker-containers.md
+++ b/pages/common/docker-containers.md
@@ -9,7 +9,7 @@
 
 - Start one or more stopped containers:
 
-`docker container start {{container_name}}`
+`docker container start {{container1_name}} {{container2_name}}`
 
 - Kill one or more running containers:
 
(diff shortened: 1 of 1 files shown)


Aracki:
@pxgamer You are right. Fixed.

pages/common/docker-containers.md
```diff
@@ -0,0 +1,36 @@
+# docker container
+
+> Manage Docker containers.
+> More information: <https://docs.docker.com/engine/reference/commandline/container/>.
+
+- List currently running Docker containers:
+
+`docker container ls`
+
+- Start one or more stopped containers:
```
sbrl: How do you specify which container to start?

Aracki: @sbrl Thanks for helping me notice there is a missing argument.

pages/common/docker-containers.md
```diff
@@ -0,0 +1,36 @@
...
+
+`docker container start {{container_name}}`
```
agnivade: Perhaps we should add another container here to show that multiple containers are supported. That way we have a mix of single and multiple containers to give the idea that both are allowed.

Aracki: I have added that just in this one command (maybe it's enough).

PR #3064, opened 2019-05-30
conda: update delete an environment

Offpics:
The original command `conda remove --name {{environment_name}} --all` does not work for me on Ubuntu (it works on windows).

I tested another command for it `conda env remove -n {{environment_name}}` and it works on both Ubuntu and Windows.


pxgamer:
The existing example seems to work fine for me on Ubuntu, and is the recommended way in the Conda documentation. 🤔

PR #2857, opened 2019-03-29
cpulimit: add page

slicer69:
Add TLDR page for cpulimit command.

- [ ] The page (if new), does not already exist in the repo.
- [ ] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [ ] The page has 8 or fewer examples.
- [ ] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [ ] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).

PR #3066, opened 2019-05-30
cpulimit: add page

sbrl:
Follow-on from #2857, since it didn't let me alter the branch for some reason.


sbrl:
Oops. I think something went rather wrong. Let me close this one and try again.

PR #3070, opened 2019-05-31
rails-db: add page

nelsonfigueroa:

### DIFF
--- a/pages/common/rails-db.md
+++ b/pages/common/rails-db.md
@@ -2,6 +2,10 @@
 
 > Various database-related subcommands for Ruby on Rails.
 
+- Create databases, load the schema, and initialize with seed data:
+
+`rails db:setup`
+
 - Access the database console:
 
 `rails db`
@@ -29,7 +33,3 @@
 - Fill the current database with data defined in db/seeds.rb:
 
 `rails db:seed`
-
-- Create databases, load the schema, and initialize with seed data:
-
-`rails db:setup`
--- a/pages/common/rails-db.md
+++ b/pages/common/rails-db.md
@@ -26,7 +26,7 @@
 
 `rails db:rollback`
 
-- Fill database with data defined in db/seeds.rb:
+- Fill the current database with data defined in db/seeds.rb:
 
 `rails db:seed`
 
--- a/pages/common/rails-db.md
+++ b/pages/common/rails-db.md
@@ -6,7 +6,7 @@
 
 `rails db`
 
-- Create databases defined in current environment:
+- Create the databases defined in the current environment:
 
 `rails db:create`
 
--- a/pages/common/rails-db.md
+++ b/pages/common/rails-db.md
@@ -10,7 +10,7 @@
 
 `rails db:create`
 
-- Destroy databases defined in current environment:
+- Destroy the databases defined in the current environment:
 
 `rails db:drop`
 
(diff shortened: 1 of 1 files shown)


pages/common/rails-db.md
```diff
@@ -0,0 +1,35 @@
+# rails db
+
+> Rails database actions.
```
sbrl: Perhaps this will read better?
```suggestion
> Various database-related subcommands for Ruby on Rails.
```

pages/common/rails-db.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Access database console:
```
sbrl: Perhaps this will read better?
```suggestion
- Access the database console:
```

pages/common/rails-db.md
```diff
@@ -0,0 +1,35 @@
...
+
+`rails db`
+
+- Create databases defined in current environment:
```
sbrl: Perhaps this will read better?
```suggestion
- Create the databases defined in the current environment:
```

Also, define "current environment"? Is this an environment variable, a configuration file, or something else?

pages/common/rails-db.md
```diff
@@ -0,0 +1,35 @@
...
+
+`rails db:create`
+
+- Destroy databases defined in current environment:
```
sbrl: Perhaps this will read better?
```suggestion
- Destroy the databases defined in the current environment:
```

pages/common/rails-db.md
```diff
@@ -0,0 +1,35 @@
...
+
+`rails db:drop`
+
+- Run pending migrations:
+
+`rails db:migrate`
+
+- View status of each migration file:
```
sbrl: Perhaps this will read better?
```suggestion
- View the status of each migration file:
```

pages/common/rails-db.md
```diff
@@ -0,0 +1,35 @@
...
+
+`rails db:migrate:status`
+
+- Rollback the last migration:
+
+`rails db:rollback`
+
+- Fill database with data defined in db/seeds.rb:
```
sbrl: Perhaps this will read better?
```suggestion
- Fill the database with data defined in db/seeds.rb:
```

If _Ruby on Rails_ supports multiple databases as you've alluded to above, how does this command know which database to seed?

pages/common/rails-db.md
```diff
@@ -0,0 +1,35 @@
...
+
+`rails db:seed`
+
+- Create databases, load the schema, and initialize with seed data:
```
sbrl: Generally we like tldr-pages to have the examples in the order you're  likely to use them. This command looks like it might be useful first - so maybe it should be at the top?

pages/common/rails-db.md
nelsonfigueroa: Depends on the environment variable `RAILS_ENV`. By default, it's set to `development`. Perhaps I should have included this information for clarity.

pages/common/rails-db.md
nelsonfigueroa: Makes sense to me, I'm okay with moving this to the top.

pages/common/rails-db.md
sbrl: Ah, I see. Maybe saying something like this would be helpful then?

```suggestion
- Fill the current database with data defined in db/seeds.rb:
```

PR #3071, opened 2019-06-01
nnn: add page

chrillebile:

### DIFF
--- a/pages/common/nnn.md
+++ b/pages/common/nnn.md
@@ -3,9 +3,9 @@
 > Interactive terminal file manager and disk usage nalyzer.
 > More information: <https://github.com/jarun/nnn>.
 
-- Open file manager at current directory, optional: at given directory from current:
+- Open the current directory (or specify one as the first argument):
 
-`nnn {{path/to/directory}}`
+`nnn`
 
 - Open an existing bookmark:
 
--- a/pages/common/nnn.md
+++ b/pages/common/nnn.md
@@ -1,6 +1,6 @@
 # nnn
 
-> Interactive terminal file manager and fisk analyzer.
+> Interactive terminal file manager and disk usage nalyzer.
 > More information: <https://github.com/jarun/nnn>.
 
 - Open file manager at current directory, optional: at given directory from current:
--- a/pages/common/nnn.md
+++ b/pages/common/nnn.md
@@ -1,13 +1,13 @@
 # nnn
 
-> Interactive terminal file manager and disk usage nalyzer.
+> Interactive terminal file manager and disk usage analyzer.
 > More information: <https://github.com/jarun/nnn>.
 
 - Open the current directory (or specify one as the first argument):
 
 `nnn`
 
-- Open an existing bookmark:
+- Open an existing bookmark (defined in the `NNN_BMS` environment variable):
 
 `nnn -b {{bookmark_name}}`
 
@@ -23,9 +23,9 @@
 
 `nnn -l`
 
-- Copy selected file path to file_out, or stdout if file_out='-':
+- Copy the selected file path to a specific output file (stdout if output file is `-`):
 
-`nnn -p {{file_out}}`
+`nnn -p {{path/to/output_file}}`
 
 - Use substring matching for filters instead of regex:
 
--- a/pages/common/nnn.md
+++ b/pages/common/nnn.md
@@ -1,0 +1,52 @@
+# nnn
+
+> The missing terminal file manager for X.
+> Read more at https://github.com/jarun/nnn.
+
+- Specify bookmark key to open:
+
+`nnn -b {{key}}`
+
+- Show hidden files:
+
+`nnn -d`
+
+- Use exiftool instead of mediainfo:
+
+`nnn -e`
+
+- Start in navigate-as-you-type mode:
+
+`nnn -i`
+
+- Start in light mode (fewer details):
+
+`nnn -l`
+
+- Use version compare to sort files:
+
+`nnn -n`
+
+- Copy (or pick) selection to file, or stdout if file='-':
+
+`nnn -p {{file}}`
+
+- Use substring match for filters instead of regex:
+
+`nnn -s`
+
+- Start in disk usage analyzer mode:
+
+`nnn -S`
+
+- Show version and exit:
+
+`nnn -v`
+
+- Wild load - entries unsorted on directory load:
+
+`nnn -w`
+
+- Show program help and exit:
+
+`nnn -h`
--- a/pages/common/nnn.md
+++ b/pages/common/nnn.md
@@ -1,32 +1,36 @@
 # nnn
 
-> The missing terminal file manager for X.
-> Read more at https://github.com/jarun/nnn.
+> Interactive terminal file manager and fisk analyzer.
+> More information: <https://github.com/jarun/nnn>.
 
-- Specify bookmark key to open:
+- Open file manager at current directory, optional: at given directory from current:
 
-`nnn -b {{key}}`
+`nnn {{path/to/directory}}`
+
+- Open an existing bookmark:
+
+`nnn -b {{bookmark_name}}`
 
 - Show hidden files:
 
 `nnn -d`
 
+- Start in disk usage analyzer mode:
+
+`nnn -S`
+
 - Start in light mode (fewer details):
 
 `nnn -l`
 
-- Use version compare to sort files:
+- Copy selected file path to file_out, or stdout if file_out='-':
 
-`nnn -n`
+`nnn -p {{file_out}}`
 
-- Copy (or pick) selection to file, or stdout if file='-':
-
-`nnn -p {{file}}`
-
-- Use substring match for filters instead of regex:
+- Use substring matching for filters instead of regex:
 
 `nnn -s`
 
-- Start in disk usage analyzer mode:
+- Use version compare to sort files:
 
-`nnn -S`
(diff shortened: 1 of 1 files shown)


pages/common/nnn.md
```diff
@@ -0,0 +1,32 @@
+# nnn
+
+> The missing terminal file manager for X.
+> Read more at https://github.com/jarun/nnn.
+
+- Specify bookmark key to open:
```
mebeim:
```suggestion
- Open an existing bookmark:
```
I'd say this reads better.

pages/common/nnn.md
```diff
@@ -0,0 +1,32 @@
...
+
+`nnn -b {{key}}`
```
mebeim:
```suggestion
`nnn -b {{bookmark_name}}`
```
Let's make it clearer what the parameter means.

pages/common/nnn.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Start in light mode (fewer details):
+
+`nnn -l`
+
+- Use version compare to sort files:
+
+`nnn -n`
+
+- Copy (or pick) selection to file, or stdout if file='-':
```
mebeim: I'm not really sure what this command does, could you provide an example usage? It does not seem to do what's described here (at least for what I tried).

pages/common/nnn.md
```diff
@@ -0,0 +1,32 @@
...
+
+`nnn -p {{file}}`
+
+- Use substring match for filters instead of regex:
```
mebeim:
```suggestion
- Use substring matching for filters instead of regex:
```

pages/common/nnn.md
```diff
@@ -0,0 +1,36 @@
+# nnn
+
+> Interactive terminal file manager and fisk analyzer.
```
mebeim:
```suggestion
> Interactive terminal file manager and disk usage analyzer.
```
Whoops, a typo (also, I forgot the important word "usage", since it's not a _disk analyzer_).

pages/common/nnn.md
```diff
@@ -0,0 +1,36 @@
...
+> More information: <https://github.com/jarun/nnn>.
+
+- Open file manager at current directory, optional: at given directory from current:
```
mebeim:
```suggestion
- Open the current directory (or specify one as the first argument):
```
This doesn't feel quite right. I'd reword like this.

pages/common/nnn.md
```diff
@@ -0,0 +1,36 @@
...
+
+`nnn {{path/to/directory}}`
```
mebeim:
```suggestion
`nnn`
```
Typically the first example is always the *simplest* usage of the command.

pages/common/nnn.md
chrillebile: It just copies the selected file path and outputs it to stdout or a specifyd file.

try: `nnn -p -` and then select an file and press enter

chrillebile: example of commando + output after I've selected the file "7zr.md"

image

pages/common/nnn.md
```diff
@@ -0,0 +1,36 @@
+# nnn
+
+> Interactive terminal file manager and disk usage nalyzer.
```
agnivade: typo at nalyzer

pages/common/nnn.md
```diff
@@ -0,0 +1,36 @@
...
+
+`nnn -S`
+
+- Start in light mode (fewer details):
+
+`nnn -l`
+
+- Copy selected file path to file_out, or stdout if file_out='-':
+
+`nnn -p {{file_out}}`
```
agnivade: Just say file_path instead of file_out. And maybe change the latter part of the description to - "or stdout if file_path is '-'"

pages/common/nnn.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Use substring matching for filters instead of regex:
+
+`nnn -s`
+
+- Use version compare to sort files:
```
agnivade: This is not clear to me. From where is it getting the version ?

pages/common/nnn.md
```diff
@@ -0,0 +1,36 @@
+# nnn
+
+> Interactive terminal file manager and disk usage nalyzer.
+> More information: <https://github.com/jarun/nnn>.
+
+- Open the current directory (or specify one as the first argument):
+
+`nnn`
+
+- Open an existing bookmark:
```
agnivade: How do we create a bookmark in the first place ? Maybe this isn't that important compared to other options ? If so, then let us leave it out.

pages/common/nnn.md
mebeim: Good point, at a quick look from building the command I thought it was meant for ordering files based on the version string in their names, for example `abc-1.10` vs `abc-1.2`, but _it does not look like so_ (order should be `1.1`, `1.2`, `1.10` in such case):

    $ touch abc-1.1
    $ touch abc-1.2
    $ touch abc-1.10
    $ nnn

    2019-06-07 12:43 644      0B abc-1.1
    2019-06-07 12:43 644      0B abc-1.10
    2019-06-07 12:43 644      0B abc-1.2

    $ nnn -s

    2019-06-07 12:43 644      0B abc-1.1
    2019-06-07 12:43 644      0B abc-1.10
    2019-06-07 12:43 644      0B abc-1.2

Some explanation would be helpful indeed :thinking:

pages/common/nnn.md
mebeim: Bookmarks are defined in the `NNN_BMS` environment variable, for example in your `.bashrc` you could add:

    export NNN_BMS='d:~/Documents;m:/home/user/Music;c:~/Downloads/Google Chrome'

pages/common/nnn.md
mebeim: I'd suggest the following:

    - Copy the selected file path to a specific output file (stdout if `out_file` is `-`):

    `nnn -p {{file_out}}`

And would also suggest leaving `file_out` as is, since the path being copied is not the one of that file, but the one selected by the user.

mebeim: Example:

Peek 2019-06-07 12-58

pages/common/nnn.md
mebeim: @chrillebile thank you for the explanation!

pages/common/nnn.md
agnivade: Oh yeah. Then we should follow the usual convention I think - path/to/output_file. Hows that sound ?

pages/common/nnn.md
agnivade: Hmm .. needs extra explanation. I think we can leave this one out.

pages/common/nnn.md
mebeim: @agnivade you're right (edited my comment above just to avoid duplicating code).

pages/common/nnn.md
mebeim: I think this example is ok to have, after all it's probably one of the most used and useful ones. I would suggest editing its description to:

    - Open an existing bookmark (defined in the `NNN_BMS` environment variable):

agnivade: Sure

pages/common/nnn.md
agnivade: I don't know. We need clarity on how the sorting is done in the first place. If someone can explain that, that would be great.

sbrl: To me, that looks like a string-based comparison sort - i.e. it compares each character one at a time. On the flip-side, a _natural_ sort would look like this:

```
1.1
1.2
1.10
```

This uses the terminology used in the PHP documentation.

mebeim: @sbrl yes, we came to the same conclusion. What we were trying to understand was what actually is the difference between `nnn` and `nnn -s`...

sbrl: Ah, I see. Is there a manual that we can refer to at all, or a developer we can @mention and ask?

mebeim: This is the GH project, https://github.com/jarun/nnn 

Maybe @jarun could be so kind to help us figure this out?

jarun: Hey guys,

Thanks for considering `nnn`. Please take a look at the program help from master:

```
usage: nnn [-a] [-b key] [-d] [-f] [-H] [-i] [-n] [-o]
           [-p file] [-r] [-s] [-S] [-t] [-v] [-h] [PATH]

The missing terminal file manager for X.

positional args:
  PATH   start dir [default: current dir]

optional args:
 -a      use access time
 -b key  open bookmark key
 -d      detail mode
 -f      run filter as cmd on prompt key
 -H      show hidden files
 -i      nav-as-you-type mode
 -n      version sort
 -o      press Enter to open files
 -p file selection file (stdout if '-')
 -r      show cp, mv progress on Linux
 -s      string filters [default: regex]
 -S      du mode
 -t      disable dir auto-select
 -v      show version
 -h      show help
```

I am suggesting master because there would be some changes in the upcoming release.

The switch to version sort is `-n`. The option `-s` is to use plain text strings instead of regex at the filter prompt.

jarun:
Please update to nnn v2.8 which is the latest.

jarun:
Also, @pxgamer please let us know if you think this doesn't belong to tldr; (it's been a while). In that case we can close the PR.

agnivade:
Let's just close it for now. @chrillebile, please feel to reopen this whenever you get to it. Thanks.

PR #3076, opened 2019-06-03
etckeeper: add page

rbuchberger:

### DIFF
--- a/pages/linux/etckeeper.md
+++ b/pages/linux/etckeeper.md
@@ -1,9 +1,9 @@
 # etckeeper
 
 > Track system configuration files in git.
-> More information: http://etckeeper.branchable.com/.
+> More information: <http://etckeeper.branchable.com/>.
 
-- First time setup (run from /etc):
+- Set up a git repo and perform various setup tasks (run from /etc):
 
 `sudo etckeeper init`
 
--- a/pages/linux/etckeeper.md
+++ b/pages/linux/etckeeper.md
@@ -1,0 +1,24 @@
+# etckeeper
+
+> Track system configuration files in git.
+> More information: http://etckeeper.branchable.com/.
+
+- Initialize: Run once from /etc to initialize a new repo, and add package manager hooks for automatic commits when installing and removing packages:
+
+`sudo etckeeper init`
+
+- Commit all changes in /etc:
+
+`sudo etckeeper commit {{message}}`
+
+- Run arbitrary git commands:
+
+`sudo etckeeper vcs {{command}}`
+
+- Check if there are uncommitted changes:
+
+`if (sudo etckeeper unclean); then echo dirty; else echo clean; fi`
+
+- Destroy existing repo and stop tracking changes:
+
+`sudo etckeeper uninit`
--- a/pages/linux/etckeeper.md
+++ b/pages/linux/etckeeper.md
@@ -3,7 +3,7 @@
 > Track system configuration files in git.
 > More information: http://etckeeper.branchable.com/.
 
-- Initialize: Run once from /etc to initialize a new repo, and add package manager hooks for automatic commits when installing and removing packages:
+- First time setup (run from /etc):
 
 `sudo etckeeper init`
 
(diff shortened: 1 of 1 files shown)


pages/linux/etckeeper.md
```diff
@@ -0,0 +1,24 @@
...
+
+`sudo etckeeper commit {{message}}`
+
+- Run arbitrary git commands:
+
+`sudo etckeeper vcs {{command}}`
+
+- Check if there are uncommitted changes:
+
+`if (sudo etckeeper unclean); then echo dirty; else echo clean; fi`
```
Aracki: I don't know if it's a good practice to keep this function here. Other commands seem good to me.

rbuchberger: Ok. I added it because  `etckeeper unclean` just returns a boolean, but you can do the same thing with `sudo etckeeper git status`. Should I change it to that or remove it entirely?

agnivade: Just keep `etckeeper unclean` and explain what it does. It is upto the user to build commands on top of it.

pages/linux/etckeeper.md
```diff
@@ -0,0 +1,24 @@
+# etckeeper
+
+> Track system configuration files in git.
+> More information: http://etckeeper.branchable.com/.
+
+- Initialize: Run once from /etc to initialize a new repo, and add package manager hooks for automatic commits when installing and removing packages:
```
agnivade: I think we should keep it short. No need to have a prefix "Initialize: ". And does the user need to add the package manager hooks or is it done automatically ?

rbuchberger: It's done automatically for most common package managers (though I'm having a hard time finding a list of which package managers exactly). I'll shorten it.

pages/linux/etckeeper.md
```diff
@@ -0,0 +1,24 @@
...
+
+`sudo etckeeper init`
+
+- Commit all changes in /etc:
+
+`sudo etckeeper commit {{message}}`
+
+- Run arbitrary git commands:
+
+`sudo etckeeper vcs {{command}}`
```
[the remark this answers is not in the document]
rbuchberger: Should I wrap it in curly braces? 

`sudo etckeeper vcs status` or `sudo etckeeper vcs {{status}}` to signify that the status part can be replaced with any git command?

pages/linux/etckeeper.md
sbrl: Yeah. Different shells will support different if statement syntax, so it's probably best to keep just the command itself and explain that the exit code is either 0 or 1 depending on the result.

pages/linux/etckeeper.md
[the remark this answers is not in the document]
agnivade: The latter.

pages/linux/etckeeper.md
```diff
@@ -0,0 +1,24 @@
+# etckeeper
+
+> Track system configuration files in git.
+> More information: http://etckeeper.branchable.com/.
```
agnivade: There is a `<>` around the link

pages/linux/etckeeper.md
```diff
@@ -0,0 +1,24 @@
...
+
+- First time setup (run from /etc):
```
agnivade: How about - "Initialize a git repo (run from /etc)"

rbuchberger: It does a bit more than that though --  here's the relevant documentation:

> The etckeeper init command initialises an /etc/.git/ repository. If you installed etckeeper from a package, this was probably automatically performed during the package installation. If not, your first step is to run it by hand:
>
>`etckeeper init`
>
>The etckeeper init command is careful to never overwrite existing files or directories in /etc. It will create a .gitignore if one doesn't already exist (or update content inside a "managed by etckeeper" comment block), sets up pre-commit hooks if they don't already exist, and so on. It does not commit any files, but does git add all interesting files for an initial commit later.
>
>Now you might want to run git status to check that it includes all the right files, and none of the wrong files. And you can edit the .gitignore and so forth. Once you're ready, it's time to commit:

So it creates an empty repo, sets up a gitignore, sets up various hooks, and adds interesting files for the first commit. `git init` just creates an empty repo; would that be a bit confusing?

agnivade: I see. How about - "Perform various housekeeping tasks to setup a git repo (run from /etc)" ? I think that gives an indication that basically it sets up a git repo and performs other tasks.

rbuchberger: Sounds great! New commit inbound.

PR #3078, opened 2019-06-03
screen: add kill from inside a screen example

digitalhabitat:

### DIFF
--- a/pages/common/screen.md
+++ b/pages/common/screen.md
@@ -26,10 +26,10 @@
 
 `Ctrl + A, D`
 
+- Kill the current screen session:
+
+`Ctrl + A, K`
+
 - Kill a detached screen:
 
 `screen -X -S {{session_name}} quit`
-
-- Kill from inside a screen:
-
-`Ctrl + A, K`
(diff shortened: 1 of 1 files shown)


sbrl:
Hey, thanks for the edit @digitalhabitat! Unfortunately, we have a limit of 8 examples per page. Is there an example that less commonly used that we could remove to bring us back under the limit?

Also, perhaps `- Kill the current screen session:` would be an alternative wording that would improve clarity a bit?

digitalhabitat:
Hi @sbrl. I'm pretty new at using screen but once I discover this option I wished I knew it earlier. The original method I used was ```- Kill a detached screen:```  but I my experience I wanted to just kill the current session. I do like your alternative wording.

sbrl:
@digitalhabitat I agree. How about we remove that one then to bring us under the limit then?

_(Also, @pxgamer / @agnivade?)_

agnivade:
If I am counting it correctly, this example just takes it to 8. Not more than that.

Also, please move it one example up.

pxgamer:
@agnivade, looks like you're right. 👍

sbrl:
Oops! I can't count apparently then :P

PR #3080, opened 2019-06-04
pacman4console: add page

mtpiercey:


pages/linux/pacman4console.md
```diff
@@ -0,0 +1,20 @@
+# pacman4console
+
+> A text-based console game inspired by the original Pacman.
+> You play as Pacman (C); avoid the ghosts (&'s); collect all the pellets (*'s and .'s).
```
sbrl: Perhaps this will read better? Also, perhaps we could add the repository url too?
```suggestion
> You play as Pacman (the `C`), avoid the ghosts (the `&`s), and collect all the pellets (the `*`s and `.`s).
> More information: <https://github.com/YoctoForBeaglebone/pacman4console>.
```

pages/linux/pacman4console.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Start a game (at level 1 by default):
```
sbrl: Perhaps this will read better?
```suggestion
- Start a game at level 1:
```

pages/linux/pacman4console.md
```diff
@@ -0,0 +1,20 @@
...
+
+`pacman4console`
+
+- Start a game on a certain level (there are nine official levels):
+
+`pacman4console --level={{#}}`
```
sbrl: Perhaps we can be a bit more explicit here?
```suggestion
`pacman4console --level={{level_number}}`
```

pages/linux/pacman4console.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Start the pacman4console editor with a level text-file:
```
sbrl: I'm not sure about this example, as it uses a different command. tldr pages should only contain examples for a single specific command. Usually, we split multiple related commands into their own pages, including something liek this as the 2nd description line:

```
> See also: `other_command_name`.
```

What do you think? If we're going to keep this, perhaps this will read better?
```suggestion
- Start the pacman4console level editor, saving to a specified text file:
```

pages/linux/pacman4console.md
```diff
@@ -0,0 +1,20 @@
...
+
+`pacmanforconsoleedit {{level_file}}`
+
+- Start a non-standard level (like a custom one):
```
sbrl: Perhaps this will read better?
```suggestion
- Play a custom level:
```

Also, could `level_file` on the line below be `path/to/level_file` please? Then it'll be more consistent with our existing pages :-)

pages/linux/pacman4console.md
mtpiercey: I know ```pacman4consoleedit``` _is_ a different command, though I just thought that since it installs with ```pacman4console``` by default, and there's only one command for it, it might be worth bundling them (since AFAIK there's only one command for using ```pacman4consoleedit```. 

However, it definitely makes sense to separate them and have a ```> See also:```, since otherwise a ```tldr``` search for ```pacman4consoleedit``` would come out empty (though again, there's only one command for using it, so...)

Honestly, I don't know what to do about this.

sbrl: Hrm. Perhaps a 2nd opinion would be useful. @pxgamer / @mebeim / @agnivade, what do you think?

agnivade: We do have some commands where other related commands are used in the same page. But only when they are very closely related. This does seem to be the case.

@mtpiercey - btw, the command is wrongly spelled as `pacmanforconsoleedit`. Note the `for` instead of `4`.

mtpiercey: OK, I fixed the typo; thanks for noticing that.

pages/linux/pacman4console.md
```diff
@@ -0,0 +1,21 @@
+# pacman4console
+
+> A text-based console game inspired by the original Pacman.
+> You play as Pacman (the `C`), avoid the ghosts (the `&`s), and collect all the pellets (the `*`s and `.`s).
```
mebeim: Do we really need to have this second line of description? It seems overly specific and kind of unnecessary to me. :thinking:

mtpiercey: Well, seeing as the game *is* based on Pacman, and the characters used to represent Pacman, the ghosts, and the pellets are about as similar to their original forms as possible, I could see how this could be spurious.

mebeim: Yep, I think the game is already intuitive enough, no need to explain it.

mtpiercey: Alright; point taken (since with the addition of the repo's link, people could look it up and get that information there).

PR #3084, opened 2019-06-04
git-cherry-pick: add example for -n option

waldyrious:
Also format command in main description with backticks.

### DIFF
--- a/pages/common/git-cherry-pick.md
+++ b/pages/common/git-cherry-pick.md
@@ -1,7 +1,7 @@
 # git cherry-pick
 
 > Apply the changes introduced by existing commits to the current branch.
-> To apply changes to another branch, first use git-checkout to switch to the desired branch.
+> To apply changes to another branch, first use `git checkout` to switch to the desired branch.
 > More information: <https://git-scm.com/docs/git-cherry-pick>.
 
 - Apply a commit to the current branch:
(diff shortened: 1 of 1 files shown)


pages/common/git-cherry-pick.md
```diff
@@ -1,7 +1,7 @@
 # git cherry-pick
 
 > Apply the changes introduced by existing commits to the current branch.
-> To apply changes to another branch, first use git-checkout to switch to the desired branch.
+> To apply changes to another branch, first use `git checkout` to switch to the desired branch.
```
waldyrious: I'm not sure this line should be here at all, actually. It sounds obvious to me, and not particularly related to the `cherry-pick` subcommand, but maybe I'm getting too used to git's tough love 😄

Any opinions on whether to keep it or remove it?

waldyrious: History laughs in my face. Here's a quote from a comment I wrote [back in 2016](https://github.com/tldr-pages/tldr/pull/1196#issuecomment-266944024), in the PR that introduced this page:

> Would it make sense to add another example, say, to cherry pick into a branch that isn't the current one?

To be fair, though, I think that might have been asking for the ability to transport a commit to some place without first going there, but if that's not possible, then I'm not sure the extra line helps much. Thoughts?

sbrl: Hrm, I'm not sure. The change made here is a good one - but I think we should keep this line, as tldr-pages is aimed at beginners - and this is likely to be a common question with respect to the `cherry-pick` subcommand.

waldyrious: Sure, then. Thanks for the input :)

PR #3090, opened 2019-06-05
git commit: add example for committing specific files

Aracki:
This command was missing, but it's very important one.

### DIFF
--- a/pages/common/git-commit.md
+++ b/pages/common/git-commit.md
@@ -1,7 +1,7 @@
 # git commit
 
 > Commit files to the repository.
-> Homepage: <https://git-scm.com/docs/git-commit>.
+> More information: <https://git-scm.com/docs/git-commit>.
 
 - Commit staged files to the repository with a message:
 
@@ -14,3 +14,7 @@
 - Replace the last commit with currently staged changes:
 
 `git commit --amend`
+
+- Commit only specific files or folders:
+
+`git commit -m {{message}} {{path/to/my/file.ext}}`
--- a/pages/common/git-commit.md
+++ b/pages/common/git-commit.md
@@ -15,6 +15,6 @@
 
 `git commit --amend`
 
-- Commit only specific files or folders:
+- Commit only specific (already staged) files:
 
-`git commit -m {{message}} {{path/to/my/file}} {{path/to/my/second_file}}`
+`git commit {{path/to/my/file1}} {{path/to/my/file2}}`
--- a/pages/common/git-commit.md
+++ b/pages/common/git-commit.md
@@ -17,4 +17,4 @@
 
 - Commit only specific files or folders:
 
-`git commit -m {{message}} {{path/to/my/file.ext}}`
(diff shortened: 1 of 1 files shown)


pages/common/git-commit.md
```diff
@@ -14,3 +14,7 @@
 - Replace the last commit with currently staged changes:
 
 `git commit --amend`
+
+- Commit only specific files or folders:
+
+`git commit -m {{message}} {{path/to/my/file.ext}}`
```
waldyrious: Maybe add two files, or a glob expression?

Also, how does this interact with the staged changes? E.g. you stage file A but commit file B, do both A and B get committed, or just B? Or if you stage part of file B, does this commit the entirety of B or just the staged parts?

Aracki: You cannot commit files that are not staged!

If you stage for example A and B and do `git commit -m "msg" A` then only file A will be commit.

Aracki: I would add second file for example.

waldyrious: > You cannot commit files that are not staged!

Well, you can with `commit -a`, so the behavior I described wouldn't be that out of place, IMO.

So what you're saying is that this can serve as a filter for the staged changes? E.g. if I stage file A, file B and half of file C, then `git commit A C` will commit only file a and half of file C, leaving file B staged and ready for another commit?

pages/common/git-commit.md
```diff
@@ -14,3 +14,7 @@
 - Replace the last commit with currently staged changes:
 
 `git commit --amend`
+
+- Commit only specific files or folders:
+
+`git commit -m {{message}} {{path/to/my/file}} {{path/to/my/second_file}}`
```
waldyrious: Maybe clearer:

```suggestion
`git commit -m {{message}} {{path/to/file1}} {{path/to/file2}}`
```

agnivade: I would remove the `-m`. It is already specified above and doesn't need to be used again. This example is clearer without it.

pages/common/git-commit.md
mebeim: I would specify that files have to be staged though, something like:

> Commit only specific (already staged) files or directories:

Also: directories, not folders :P

Aracki: @waldyrious afaik `-a` flag will just do `add` command before commit. But I agree with you 👍 

@mebeim Probably: `Commit only specific (already staged) files` is good enough?

mebeim: Yeah @Aracki I think that sounds good :)

waldyrious: This is clearer, yes. Now if this works with [`commit --patch`](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--p), that means I'll have a feature I've been yearning for a while! Multi-level staging :P

waldyrious: @Aracki I'm leaving this conversation "unresolved" so it doesn't get collapsed, since it contains important details that may inform future readers.

PR #3093, opened 2019-06-07
sshuttle: add exclude example

sbrl:
It's that time again when I have internet troubles - and I break out _sshuttle_ to fix them. Now that I've done some more advanced networking at University (and I've used sshuttle in a different context), I've realised that this page could use some tweaking.


sbrl:
Thanks, @mebeim! Just done some more testing & revised it a bit more. That should be the last change :-)

PR #3100, opened 2019-06-09
Translate 60+ common pages in Italian.

mebeim:
Had this branch sitting around for a while. There should be 63 new common pages translated in italian (if I counted them correctly). The reason you see "81 files changed" is because I've also added all the missing homepages. 

NB: This should obviously be rebased rather than squashed.


schneiderl:
Wow 🤙

PR #3103, opened 2019-06-11
faketime: add page

Nijin22:

### DIFF
--- a/pages/linux/faketime.md
+++ b/pages/linux/faketime.md
@@ -3,10 +3,10 @@
 > Fake the system time for a given command.
 > More information: <https://manpages.ubuntu.com/manpages/trusty/man1/faketime.1.html>.
 
-- Fake the time to this evening, before printing the result of {{date}}:
+- Fake the time to this evening, before printing the result of `date`:
 
-`faketime 'today 23:30' date`
+`faketime '{{today 23:30}}' date`
 
-- Open a new {{bash}} shell, which uses yesterday as the current date:
+- Open a new `bash` shell, which uses yesterday as the current date:
 
-`faketime 'yesterday' bash`
+`faketime '{{yesterday}}' bash`
--- a/pages/linux/faketime.md
+++ b/pages/linux/faketime.md
@@ -3,10 +3,10 @@
 > Fake the system time for a given command.
 > More information: <https://manpages.ubuntu.com/manpages/trusty/man1/faketime.1.html>.
 
-- Fake the time to this evening, before printing the result of `date`:
+- Fake the time to this evening, before printing the result of {{date}}:
 
 `faketime 'today 23:30' date`
 
-- Open a new bash shell, which uses yesterday as the current date:
+- Open a new {{bash}} shell, which uses yesterday as the current date:
 
 `faketime 'yesterday' bash`
--- a/pages/linux/faketime.md
+++ b/pages/linux/faketime.md
@@ -1,0 +1,11 @@
+# faketime
+
+> Fake this computers time for a given command.
+
+- Fake the time to this evening, before printing the result of `date`:
+
+`faketime 'today 23:30' date`
+
+- Open a new bash shell, which uses yesterday as the current date:
+
+`faketime 'yesterday' bash`
--- a/pages/linux/faketime.md
+++ b/pages/linux/faketime.md
@@ -5,8 +5,12 @@
 
 - Fake the time to this evening, before printing the result of `date`:
 
-`faketime '{{today 23:30}}' date`
+`faketime '{{today 23:30}}' {{date}}`
 
 - Open a new `bash` shell, which uses yesterday as the current date:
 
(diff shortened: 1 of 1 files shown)


Nijin22:
Feedback by @tldr-bot implemented with a amended commit.

I'm unsure if that already triggered the bot again. Is there anything else I need to do?

pages/linux/faketime.md
```diff
@@ -0,0 +1,11 @@
+# faketime
+
+> Fake this computers time for a given command.
```
agnivade: Please also add a line for `Homepage: <>`. And how about `Fake the system time for a given command` ?

Nijin22: What exactly would you suggest for the homepage? There is a [GitHub repo for the library](https://github.com/wolfcw/libfaketime), but nothing for the `faketime` tool itself, except for [online manpages](http://manpages.ubuntu.com/manpages/trusty/man1/faketime.1.html).

agnivade: I think the manpage is fine.

agnivade: Duh .. thanks @mebeim

pages/linux/faketime.md
```diff
@@ -0,0 +1,12 @@
+# faketime
+
+> Fake the system time for a given command.
+> More information: <https://manpages.ubuntu.com/manpages/trusty/man1/faketime.1.html>.
+
+- Fake the time to this evening, before printing the result of `date`:
+
+`faketime 'today 23:30' date`
```
agnivade: date and bash should be within tokens (`{{}}`)

agnivade: Sorry, not in the descriptions. In the examples. https://github.com/tldr-pages/tldr/blob/master/contributing-guides/style-guide.md

Nijin22: No problem, should be fixed in 46f1eca.

pxgamer: @nijin22, as @agnivade mentioned, this should be around the examples, i.e.

```suggestion
`faketime '{{today 23:30}}' date`
```

and
```
`faketime '{{yesterday}}' bash`
```

Nijin22: Oh, I see. This makes a lot more sense to me. Thanks!

Fixed in c90c2c67

agnivade: Oh my apologies. @pxgamer - AFAIU - the commands `date` and `bash` are variable. And so is the time set. So shouldn't both of them be in tokens ? i.e. - `faketime '{{today 23:30}}' {{date}}` ?

pxgamer: I hadn't really thought about that @agnivade 🤔 Yes, probably. 👍

Nijin22: Hmm, I would disagree. I'd argue that they are variable, in the sense that they can be switched out for something else, but then it would no longer fit the description of the example. ('output the value of date' and 'open a bash shell') and therefore shouldn't be marked that way. 

Similar, how we didn't tokenize the `ps` in  `docker ps` [in the docker.md file](https://github.com/tldr-pages/tldr/blob/db795994730108131d36e7a50b67378e79e27c10/pages/common/docker.md) as it would no longer list the running containers but do something else instead.

Certainly not a black-or-white decisions, so I'm okay with whatever @agnivade prefers - just wanted to tell you my thoughts first. Should I put them in tokens?

agnivade: Right. @Nijin22 - Really sorry about the multiple back and forth here. Could you please add tokens around date and bash too like I mentioned above ? Thanks.

Nijin22: A though I had: How about keeping the current 2 examples as-is and adding a third one:

Run any possible command with a faked time:

    faketime '{{tomorrow 11:30}}' {{my_custom_program}}

This would highlight the fact that the last parameter is, in fact, customizable without the issues of having variable parameters changing the explanation of the `date` or `bash` example.

Nijin22: Actually, I'm no longer sure if my "doesn't fit the example description" argument is valid anymore. As changing the specified faked-date would also change the meaning.

I'll tokenize `date` and `bash` and will also add the custom-program example.

schneiderl:
Thank you for the new command @Nijin22 and congrats for you first contribution to tldr-pages 👍

pages/linux/faketime.md
agnivade: > Similar, how we didn't tokenize the ps in docker ps in the docker.md file as it would no longer list the running containers but do something else instead.

There is a difference there. `docker ps` is the command itself. But for `faketime {{time}} {{command}}`, command can be anything. It is similar to why `iostat {{2}}` has 2 in tokens but still says "show statistics every 2 seconds". Think of tokens like any user-supplied parameters.

agnivade:
Appreciate your patience with the review process @Nijin22

PR #3106, opened 2019-06-13
remove outdated android clients

baruchiro:


baruchiro:
But maybe keep the reference somewhere for direct Developers to fix the app

agnivade:
From what I see, the app does not seem to be maintained anymore. There are pending bugs which are not fixed. While that is certainly not good, is that reason to remove it from the list ? There are bugs in many clients, and the author might be late in fixing them. What is the criteria to keep clients on the list ?

I think it might be better to just add a word `(unmaintained)` beside the client to indicate that it is not maintained.

pxgamer:
@agnivade, that does make slightly more sense. 👍

PR #3107, opened 2019-06-13
cowsay: simplify example

mebeim:
No need for an `echo` + pipe for the third example.

### DIFF
--- a/pages/common/cowsay.md
+++ b/pages/common/cowsay.md
@@ -13,7 +13,7 @@
 
 - Print an ASCII dragon saying "Hello!":
 
-`echo "Hello!" | cowsay -f dragon`
+`cowsay -f dragon "Hello!"`
 
 - Print a stoned thinking ASCII cow:
 
--- a/pages/common/cowsay.md
+++ b/pages/common/cowsay.md
@@ -7,6 +7,10 @@
 
 `cowsay "Hello world!"`
 
+- Use text from standard input for the balloon:
+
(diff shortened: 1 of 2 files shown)


agnivade:
I think the intention was to show that cowsay accepts text from the standard input also. Because the previous examples already show text as a third parameter. ping @fluxw42 (the original author of the page)

mebeim:
@agnivade Oh, I see. I got confused by this one because I thought the only way to use another character was to use stdin. If that's the case I think it would be better to add another example. I could add it to this PR, what do you think?

agnivade:
Sure, SGTM.

mebeim:
Done. I think this should be rebased rather than squashed to avoid confusion with the commit message.

pages/common/cowsay.md
```diff
@@ -18,3 +18,7 @@
 - Print a stoned thinking ASCII cow:
 
 `cowthink -s "I'm just a cow, not a great thinker ..."`
+
+- Use text from standard input for the balloon:
```
agnivade: I think you can make this the 2nd example.

mebeim: I was wondering about moving it up... will do :)

agnivade: @mebeim - Did you commit your changes ? I don't see them.

mebeim: @agnivade oh sorry, I committed without adding any change :laughing: should be ok now.

PR #3120, opened 2019-06-21
qpdf: add merge files example

dgsb:

### DIFF
--- a/pages/common/qpdf.md
+++ b/pages/common/qpdf.md
@@ -11,7 +11,6 @@
 
 `qpdf --empty --pages {{file1.pdf}} {{file2.pdf}} {{file3.pdf}} -- {{output.pdf}}`
 
-
 - Merge (concatenate) given pages from a list of PDF files and save the result as a new PDF:
 
 `qpdf --empty --pages {{file1.pdf}} {{1,6-8}} --pages {{file2.pdf}} {{3,4,5}} -- {{output.pdf}}`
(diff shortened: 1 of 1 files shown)


pages/common/qpdf.md
```diff
@@ -11,6 +11,10 @@
 
 `qpdf --empty --pages {{file1.pdf}} {{1,6-8}} --pages {{file2.pdf}} {{3,4,5}} -- {{output.pdf}}`
 
+- Merge all the pages of a list of PDF files and save the result as another one:
```
schneiderl: "Another one" might be confusing since you could be talking about lists or pdf files. 
Something like "save the result to a pdf" file would read better IMHO.

mebeim: What about:

    Merge all the pages of a list of PDF files and save the result as a new PDF:

agnivade: Then let us do it for all the descriptions.

Also, this should move one example above, as this is a more general case than the one above.

pages/common/qpdf.md
```diff
@@ -11,6 +11,10 @@
```
agnivade: Now this description becomes slightly misleading because of the new one. We should change it to - "Merge (concatenate) given pages from a list of PDF files and save the result as a new PDF"

dgsb:
Thanks for your reviews, I've done the required changes.

PR #3122, opened 2019-06-21
echo: add one more example

Aracki:
I have added one more example to `echo` page.

### DIFF
--- a/pages/common/echo.md
+++ b/pages/common/echo.md
@@ -14,6 +14,10 @@
 
 `echo -n {{"Hello World"}}`
 
+- Append a message to the file:
+
+`echo {{"Hello World"}} > {{file.txt}}`
+
 - Enable interpretation of backslash escapes (special characters):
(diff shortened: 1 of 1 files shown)


pages/common/echo.md
```diff
@@ -14,6 +14,10 @@
 
 `echo -n {{"Hello World"}}`
 
+- Append a message to the file:
+
+`echo {{"Hello World"}} > {{file.txt}}`
```
mebeim:
```suggestion
`echo {{"Hello World"}} >> {{file.txt}}`
```
Careful! That command overwrites the file :')

Also, I'd suggest `to a file` instead of `to the file`.

Aracki: I forgot about that. Thanks, @mebeim.

sbrl: Ah, yeah. `to a file` does sound better.

PR #3126, opened 2019-06-23
rename: fix 'rename' command has two different versions

### LINKED ISSUE #3125: The 'rename' command has multiple different versions
There are 2 versions of the `rename` command, a perl version for debian-based distros and the util-linux version which all other distros use. [source](https://unix.stackexchange.com/questions/275254/why-is-the-rename-utility-on-debian-ubuntu-different-than-the-one-on-other-distr)
The page is written for the perl version.

This can cause problems (which it did for me because I'm using fedora) for users with the util-linux version because it doesn't use Perl expressions.
Perl: [man-page](http://manpages.ubuntu.com/manpages/trusty/man1/prename.1.html)
util-linux: [man-page](http://man7.org/linux/man-pages/man1/rename.1.html)

I can't think of a solution for this problem... :/

principis:
Fixes #3125 
Moved rename to the linux folder and added explanation for rename.ul.
Added rename.ul

Let me know if any changes are necessary! :)


pages/linux/rename.md
```diff
@@ -1,6 +1,7 @@
 # rename
 
 > Renames multiple files.
+> This is the perl version of `rename`, which is the default on Debian. See `rename.ul` for the stand-alone version.
```
waldyrious: Maybe make this more explicit? Also capitalize "Perl".

```suggestion
> This is the Perl version of `rename`, which is the default on Debian. See `rename.ul` for the util-linux version.
```

And perhaps to equalize the length of both lines, we could instead consider something like this:

> Rename multiple files using Perl regular expressions.
> For the C-based command included in util-linux, see `rename.ul`.

principis: >Rename multiple files using Perl regular expressions.
For the C-based command included in util-linux, see rename.ul.

This looks a bit strange to me. Changing the first sentence is fine but the `C-based` probably wont mean anything to most people.

waldyrious: Yeah, I was unsure as well. In any case, the current version sounds good to me.

pages/linux/rename.ul.md
```diff
@@ -0,0 +1,29 @@
+# rename.ul
+
+> Renames multiple files.
+> This is the util-linux version of `rename`. See `rename` for the perl version which is default on Debian.
+> Warning: The renaming has no safeguards and will overwrite files without any questions.
```
sbrl: Perhaps this will read better?
```suggestion
> Warning: This command has no safeguards and will overwrite files without prompting!
```

principis: Much better, thanks!

principis: @sbrl  the exclamation mark broke it again, I forgot. :/ Do I need to change it or does the test need to be changed?

principis:
@sbrl is fine for me. I still need to make some changes according to what's discussed in #3125 but I'm currently short on time. I'll finish it friday or saturday. (Rename the files, and add perl-rename)

pages/linux/rename.md
```diff
@@ -1,6 +1,7 @@
 # rename
 
 > Renames multiple files.
+> This is the Perl version of `rename`, which is the default on Debian. See `rename.ul` for the util-linux version.
```
mebeim:
```suggestion
> This is the Perl version of `rename`, which is the default on Debian. See `rename.ul` for the util-linux version.
> Warning: This command has no safeguards and will overwrite files without prompting.
```
I'd keep the same warning on both pages.

pages/linux/rename.ul.md
```diff
@@ -0,0 +1,29 @@
+# rename.ul
+
+> Renames multiple files.
+> This is the util-linux version of `rename`. See `rename` for the perl version which is default on Debian.
```
mebeim:
```suggestion
> This is the util-linux version of `rename`. See `rename` for the Perl version which is default on Debian.
```
Typo.

pages/linux/rename.ul.md
```diff
@@ -0,0 +1,29 @@
...
+
+`rename -vn {{foo}} {{bar}} {{*}}`
+
+- Do not overwrite existing files:
+
+`rename -o {{foo}} {{bar}} {{*}}`
+
+- Change file extensions:
+
+`rename {{.ext}} {{.bak}} *.foo`
```
mebeim:
```suggestion
`rename {{.ext}} {{.bak}} {{*.ext}}`
```
This command doesn't make much sense, it changes `.ext` to `.bak`, but the files selected have a `.foo` extension which basically nullifies the purpose of the command.

pages/linux/rename.ul.md
```diff
@@ -0,0 +1,29 @@
...
+
+- Prepend all files with 'foo':
```
mebeim:
```suggestion
- Prepend "foo" to all filenames in the current directory:
```
I think it would be clearer worded this way, do you agree?

pages/linux/rename.ul.md
```diff
@@ -0,0 +1,29 @@
...
+
+`rename '' 'foo' {{*}}`
+
+- Rename 'foo1', ..., 'foo10', ..., 'foo100' to 'foo001', ..., 'foo010', ..., 'foo100':
```
mebeim: This command and description is somewhat confusing, and the command is also wrong (the first one should be `rename foo foo0 foo?`, notice only one zero).

I would suggest the following:

    > Rename a group of increasingly numbered files zero-padding the numbers up to 3 digits:

    `rename {{foo}} {{foo0}} {{foo?}} && rename {{foo}} {{foo0}} {{foo??}}`

pages/linux/rename.md
```diff
@@ -1,6 +1,7 @@
 # rename
 
 > Renames multiple files.
```
mebeim:
```suggestion
> Rename multiple files.
```
While we're at it, let's also correct this and make it impersonal (really strange that the man page uses third person). Same goes for `rename.ul`.

pages/linux/rename.ul.md
mebeim: Do we really need to have it say "which is default on Debian"? I wouldn't say that's really necessary. What do you think?

pages/linux/rename.ul.md
mebeim: By the way, this becomes even cooler with bash tokens:

    rename bar{,0,?} && rename bar{,0,??}

principis: The command is correct. ;)
```
>ls
foo  foo1  foo10  foo100  foo11  foo12  foo2  foo3  foo4  foo5  foo99
>rename foo foo00 foo?
>ls
foo  foo001  foo002  foo003  foo004  foo005  foo10  foo100  foo11  foo12  foo99
>rename foo foo0 foo??
>ls
foo  foo001  foo002  foo003  foo004  foo005  foo010  foo011  foo012  foo099  foo100
```
I'll change it to your suggestion, it's easier to understand. I don't know if I should add the bash tokens? I have never heard of it :o

pages/linux/rename.md
principis: The Perl version prompts before overwriting. You can force it with `-f` so the warning is inapplicable.

pages/linux/rename.ul.md
mebeim: Hmm yeah that shorter example is less clear, let's just go with the long one.

pages/linux/rename.ul.md
principis: Removed it ;)

pages/linux/rename.md
principis: Fixed, thanks!

pages/linux/rename.ul.md
principis: Woops thanks :)

pages/linux/rename.ul.md
principis: Thanks!

principis:
@mebeim thanks for all the fixes. I just resolved everything because I changed a lot to what was discussed in #3125 

I was thinking about an example for prename/perl-rename but couldn't immediately find something. According to the man page (prename/perl-rename), it shouldn't overwrite files without asking but in my testing it did... Can someone confirm?

pages/linux/rename.md
```diff
@@ -0,0 +1,30 @@
...
+
+- Change file extensions:
+
+`rename {{.ext}} {{.bak}} {{*.ext}}`
+
+- Prepend "foo" to all filenames in the current directory:
+
+`rename {{''}} {{'foo'}} {{*}}`
+
+- Rename 'foo1', ..., 'foo10', ..., 'foo100' to 'foo001', ..., 'foo010', ..., 'foo100':
```
mebeim:
```suggestion
- Rename a group of increasingly numbered files zero-padding the numbers up to 3 digits:
```

mebeim:
By the way, I `prename` does indeed rename without a warning on my machine (Debian 9.9).

mebeim:
Ping @principis could you please apply my last unresolved suggestion so that this can finally be merged?

principis:
@mebeim I'm so sorry, I've had terrible internet the past few weeks because I'm not at home... (~35 KB/s) and few time. I thought I still wanted to make some changes, but it's difficult to do anything from here so merging is fine. Sorry for the delay.

mebeim:
@principis no problem at all! Thank you again for your work :smile:

PR #3128, opened 2019-06-24
iptables

vastpeng:

### DIFF
--- a/pages/common/iptables.md
+++ b/pages/common/iptables.md
@@ -1,0 +1,21 @@
+# iptables
+
+> Monitor incoming and outgoing traffic to a server and filter it.
+> More infomation: <https://linux.die.net/man/8/iptables>.
+
+- Enable/Disable INPUT/OUTPUT connections on specific port:
+
+`iptables -A {{INPUT/OUTPUT}} -p tcp --dport {{port}} -j ACCEPT`
+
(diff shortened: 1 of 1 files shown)


Aracki:
1. This should not be in `common/` folder as macOS doesn't support that command.
2. There is an `iptables` page for linux platform. (you can try `tldr iptables -p linux`).

sbrl:
Hrm, I'm not sure that immediately closing a PR is the right approach to take, @Aracki. While the PR may seem redundant to you, it's usually best to work out the details first.

sbrl:
@vastpeng It does appear that there is already a page for `iptables` in `linux` as @Aracki mentions. Since you've placed this in `common/`, are you seeing this command available in OSes other than Linux? I don't have anything other than Linux to test with (and I certainly don't have a Solaris box to hand), so I'm not certain.

If it turns out that `iptables` is linux-specific, perhaps we could merge this page with that one and pic  the best / most useful examples from each?

Also, it looks like you're having some issues with the linter. We've got the style we prefer documented over [here](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#markdown-format), and a few good examples are the pages for `tar`, `eyeD3`, and `ln`. If you need any assistance, just ask :smiley_cat:

PR #3129, opened 2019-06-24
perl: Add PCRE one liner example

hello-woof:

### DIFF
--- a/pages/common/perl.md
+++ b/pages/common/perl.md
@@ -31,6 +31,6 @@
 
 `perl -p0e 's/{{foo\nbar}}/{{foobar}}/g' {{input_file}} > {{output_file}}`
 
-- Run a PCRE on a piped input, print out first capture group:
+- Run a regular expression on stdin, printing out first capture group for each line:
 
 `echo "$INPUT" | perl -nle 'if (/.*({{foo}}).*/) {print "$1"; last;}'`
--- a/pages/common/perl.md
+++ b/pages/common/perl.md
@@ -31,6 +31,6 @@
 
 `perl -p0e 's/{{foo\nbar}}/{{foobar}}/g' {{input_file}} > {{output_file}}`
 
-- Run a PCRE on a piped input, print out first capture group
+- Run a PCRE on a piped input, print out first capture group:
 
 `echo "$INPUT" | perl -nle 'if (/.*({{foo}}).*/) {print "$1"; last;}'`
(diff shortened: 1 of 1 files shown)


hello-woof:
Any suggestions on how to reword the description of the command are welcome.

pages/common/perl.md
```diff
@@ -30,3 +30,7 @@
 - Run a multi-line find/replace expression on a file, and save the result in another file:
 
 `perl -p0e 's/{{foo\nbar}}/{{foobar}}/g' {{input_file}} > {{output_file}}`
+
+- Run a PCRE on a piped input, print out first capture group:
```
sbrl: Perhaps this will read better?
```suggestion
- Run a regular expression on stdin, printing out first capture group for each line:
```

pxgamer: @sbrl, as per the conversation in #3114, should this remain as stdin?

sbrl: @pxgamer Good shout. I'd say so personally, but I'm not sure what the general consensus is.

mebeim: I'd personally say that "standard input" sounds better here. Otherwise, if using "stdin", since #3114 is already merged I'd say let's enclose it in backticks. We can always change it later if we want, but let's keep it consistent.

waldyrious: We could simply say "input from another command". "Standard input" is about as opaque as "stdin" for those who don't know about the concept.

pages/common/perl.md
```diff
@@ -30,3 +30,7 @@
 - Run a multi-line find/replace expression on a file, and save the result in another file:
 
 `perl -p0e 's/{{foo\nbar}}/{{foobar}}/g' {{input_file}} > {{output_file}}`
+
+- Run a regular expression on stdin, printing out first capture group for each line:
+
+`echo "$INPUT" | perl -nle 'if (/.*({{foo}}).*/) {print "$1"; last;}'`
```
waldyrious: Can we use something more basic than `echo "$INPUT"`? E.g. `cat {{somefile.txt}}` or something else along those lines.

Also, it's not immediately clear to me what specific feature this example intends to demonstrate. Is it the `$1`? Or maybe the `{...}` construct? The `/.* ... .*/` pattern? The `-n` option, or the `-l` one?

We should be careful not to overload a single example with multiple new concepts. Examples should work as building blocks that users can compose, not necessarily ready-to-use "incantations" that serve a particular purpose but fail to teach the user about the mechanics of the command.

pages/common/perl.md
sbrl: Hrm. That's true @waldyrious, but I'd also be wary of using non-standard terminology. It's probably a good thing to get users used to words and phrases like `stdin` and `standard input`, since they are used elsewhere.

In this case though, I think it does sound better to use `input from another command`.

PR #3136, opened 2019-06-25
tldr: Fix broken example

### LINKED ISSUE #3135: tldr's own entry has a broken example.
TLDR's own entry has one problematic example:

`tldr -p linux tar`

* The `-p` flag should be replaced with the `-o` flag
* There is no linux entry for `tar`. 

If you run that example, you get this:

```bash

~
❯ tldr -p linux tar
usage: tldr [-h] [-u] [-o {linux,osx,sunos}] [-s SOURCE] [-c] command [command ...]
tldr: error: unrecognized arguments: -p

```
If you then replace -p with -o, you get the following:

```

~
❯ tldr -o linux tar
`tar` documentation is not available. Consider contributing Pull Request to https://github.com/tldr-pages/tldr
```

rbuchberger:

fix #3135 

I chose tree because it's Linux specific, but distro agnostic. There might be a better choice for an example command, but I don't think it really matters too much. Sadly, cowsay is under common and figlet just isn't as funny.

Might it be better to remove that example entirely? The -o flag doesn't understand `windows` or `common` either.

### DIFF
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -7,9 +7,9 @@
 
 `tldr {{command}}`
 
-- Show the tar tldr page for linux:
+- Show the tree tldr page for linux:
 
-`tldr -p {{linux}} {{tar}}`
+`tldr -o {{linux}} {{tree}}`
 
 - Get help for a git subcommand:
 


mebeim:
Hi @rbuchberger, thank you for the PR.

The `-p` option is part of the official [client specification](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md), which *should* be followed by clients. The `-o` option is not. Which client are you using? This page reflects the specification, so you're most probably using a client which is not following it. I would suggest opening an issue on their repository. 

As per the `tar` command, that's really just a placeholder for a "common" command. It's the same as shown in the screenshot in the readme (and GitHub homepage) of this repository. Which command to use as an example is not really important.

I'm closing this, let me know if you've got any question or doubt.

mebeim:
@rbuchberger no problem, thanks for taking the time to report it. I'd suggest transfering #3135 to the appropriate repo. I could do that for you if you want.

waldyrious:
> I'd suggest transfering #3135 to the appropriate repo then.

There was already https://github.com/tldr-pages/tldr-python-client/issues/76, which was fixed in https://github.com/tldr-pages/tldr-python-client/pull/77.

(Btw, I suggest that we continue the conversation in #3135 to avoid dispersing the thread.)

mebeim:
Sure.

PR #3139, opened 2019-06-25
pigz: added folder compression method

mtonsmann:

### DIFF
--- a/pages/common/pigz.md
+++ b/pages/common/pigz.md
@@ -15,7 +15,7 @@
 
 `pigz -0 -p{{4}} {{filename}}`
 
-- Compress a folder using tar:
+- Compress a directory using tar:
 
 `tar cf - {{folder}} | pigz > {{filename}}.tar.gz`
 
--- a/pages/common/pigz.md
+++ b/pages/common/pigz.md
@@ -17,7 +17,7 @@
 
 - Compress a directory using tar:
 
-`tar cf - {{folder}} | pigz > {{filename}}.tar.gz`
+`tar cf - {{path/to/directory}} | pigz > {{filename}}.tar.gz`
 
 - Decompress a file:
 
--- a/pages/common/pigz.md
+++ b/pages/common/pigz.md
@@ -15,6 +15,10 @@
 
 `pigz -0 -p{{4}} {{filename}}`
 
+- Compress a folder using tar:
+
+`tar cf - {{folder}} | pigz > {{filename}}.tar.gz`
+
 - Decompress a file:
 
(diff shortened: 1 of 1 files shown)


pages/common/pigz.md
```diff
@@ -15,6 +15,10 @@
 
 `pigz -0 -p{{4}} {{filename}}`
 
+- Compress a folder using tar:
+
+`tar cf - {{folder}} | pigz > {{filename}}.tar.gz`
```
mebeim: I don't quite understand this command, is this any different from `tar -czf folder.tar.gz folder`? It looks like reinventing the wheel.

mtonsmann: https://github.com/madler/pigz is a parallelized implementation of gzip, and thus significantly faster at compressing/decompressing large files on multi-core machines.

The existing tldr page for pigz is useful but does not include the de-facto method for compressing folders, which is to tar them first.

mebeim: I see. Thank you for the explanation. So the decompression would still have to be made with `tar`, am I right? I mean, you'd either have to do `pigz -d archive.tar.gz && tar -xf archive.tar` or `tar -xzf archive.tar.gz`, right?

Also, sorry, forgot to mention: it's better to say `directory` instead of `folder`, could you please change that?

pages/common/pigz.md
```diff
@@ -15,6 +15,10 @@
 
 `pigz -0 -p{{4}} {{filename}}`
 
+- Compress a folder using tar:
```
mebeim:
```suggestion
- Compress a directory using tar:
```

pages/common/pigz.md
mebeim:
```suggestion
`tar cf - {{path/to/directory}} | pigz > {{filename}}.tar.gz`
```

pages/common/pigz.md
mebeim:
```suggestion
- Compress a directory using tar:
```

pages/common/pigz.md
mebeim:
```suggestion
`tar cf - {{path/to/directory}} | pigz > {{filename}}.tar.gz`
```

mtonsmann: That's right, and it's worth noting that `pigz` doesn't use multiple cores for decompression. However, it does use them for other calculations which can result in a performance improvement over gzip, [see here](https://stackoverflow.com/a/51688385).

The one-liner to do so would be:
`pigz -dc archive.tar.gz > tar -xf -`

Issue #3111, opened 2019-06-16
Fix broken badges with statistics


Aracki:
I suggest we remove the following two badges. 	

* `GitHub contributors` is not needed as you can see that in the top bar.
* `Number of files` seems broken and nevertheless I think we do not need this (maybe the number of pages by OS would be better statistic)

waldyrious:
I agree with removing the number of files badge, Although I'd prefer if we could replace it with another service providing a count of pages in the repo or in a specific folder. I think it was @pxgamer who mentioned that we could even contribute new badges to shields.io, so even if such a badge doesn't exist, maybe we could create it.

As for the contributors badge, I think it's worth keeping it since it matches well with the ones counting merged PRs and issue resolution time. Essentially they're signals of community activity, and the fact that we chose to include those numbers in the README also reflects our project's focus on welcoming contributions.

sbrl:
That's a really good argument for keeping the page count badge, actually.

On Gitter, we considered a few approaches, all centred around the new [dynamic badge](https://shields.io/#dynamic-badge) feature of shields.io:

1. Reconfigure @tldr-bot to emit statistics.
2. Alter the build process to emit a `stats.json` containing various interesting statistics.

Personally, I really like the 2nd option here. It could even be used to make a 'status panel' that shows various things like page counts per platform, translation percentages, etc.

Aracki:
@waldyrious Why there is **406** contributors showed in the badge and **741** in the top bar?

pxgamer:
Anon users maybe? You can do:

**GitHub Contributors (current badge)**
![Contributors](https://img.shields.io/github/contributors/tldr-pages/tldr.svg)
**With anonymous contributions**
![Contributors with Anon](https://img.shields.io/github/contributors-anon/tldr-pages/tldr.svg)

Maybe GitHub has merged some duplicates or something?

waldyrious:
It would indeed be nice to have a more concrete idea of the reason for the discrepancy. That would be relevant e.g. for #1076.

mebeim:
For the number of files badge, the second option proposed by @sbrl seems like the way to go IMHO. I think having some kind of statistics about the project would be cool and helpful, and that would of course make the file count badge trivial to generate from shields.io's dynamic badges.

[...]

PR #3141, opened 2019-06-26
README: remove broken file count badge

mebeim:
Temporary patch while we wait for #3111 to be properly solved, since having a broken badge is just not useful.


Aracki:
👍

schneiderl:
👍

PR #3147, opened 2019-06-28
pickle: add page

pxgamer:

### DIFF
--- a/pages/common/pickle.md
+++ b/pages/common/pickle.md
@@ -1,0 +1,20 @@
+# pickle
+
+> A PHP extension installer based on Composer.
+> More information: <https://github.com/FriendsOfPHP/pickle>.
+
+- Install a specific PHP extension:
+
+`pickle install {{memcache}}`
+
+- Convert an existing PECL extension configuration to a Pickle configuration file:
+
+`pickle convert {{path/to/directory}}`
+
+- Validate a PECL extension:
+
+`pickle validate {{path/to/directory}}`
+
+- Package a PECL extension for releasing:
+
+`pickle release {{path/to/directory}}`
--- a/pages/common/pickle.md
+++ b/pages/common/pickle.md
@@ -15,6 +15,6 @@
 
 `pickle validate {{path/to/directory}}`
 
-- Package a PECL extension for releasing:
+- Package a PECL extension for release:
 
 `pickle release {{path/to/directory}}`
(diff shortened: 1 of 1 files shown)


pages/common/pickle.md
```diff
@@ -0,0 +1,20 @@
+# pickle
+
+> A PHP extension installer based on Composer.
+> More information: <https://github.com/FriendsOfPHP/pickle>.
+
+- Install a specific PHP extension:
+
+`pickle install {{memcache}}`
```
andrik:
```suggestion
`pickle install {{extension_name}}`
```

pxgamer: I put `memcache` as a definitive example. But it does look a bit odd. 🤔

In PHP, extension names can be referenced as:
- `memcache`
- `php-memcache` (won't work here, but used in package managers)
- `ext-memcache` (won't work here, but used in Composer)

So I thought having an example would be clearer.

mebeim: Maybe something like this? :thinking:

    - Install a specific PHP extension (don't include `php-` or `ext-` in the name):

    `pickle install {{extension_name}}`

agnivade: Using a concrete name definitely encouraged and in the spirit of tldr pages. I don't see any issues with it.

pages/common/pickle.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Convert an existing PECL extension configuration to a Pickle configuration file:
+
+`pickle convert {{path/to/directory}}`
+
+- Validate a PECL extension:
+
+`pickle validate {{path/to/directory}}`
+
+- Package a PECL extension for releasing:
```
agnivade: "for release" sounds better to me. But not sure if the other one is objectively correct.

pxgamer: `for release` sounds better to me too, so I think I'll go with that :+1:

agnivade:
Hmm .. that's a good point @andrik. Traditionally, our concrete examples have been numbers, IPs, mount locations etc. And for website names, we chose `example.com` to avoid preference to any specific site. In this case, it does seem that we are unnecessarily choosing `memcache`. Maybe it is good to be generic after all.

I didn't give much thought on this initially. But on a deeper look, I would have to agree to @andrik's point.

mebeim:
Right, I too agree with @andrik .

PR #3149, opened 2019-06-29
kops: add page

Aracki:
I have added [kops](https://github.com/kubernetes/kops/) - k8s cluster management tool.

### DIFF
--- a/pages/common/kops.md
+++ b/pages/common/kops.md
@@ -1,0 +1,24 @@
+# kops
+
+> A tool that create, destroy, upgrade and maintain production-grade, highly available, Kubernetes clusters from the command line.
+> More informations: <https://github.com/kubernetes/kops/>.
+
+- Create a cluster from the configuration specification:
+
+`kops create cluster -f {{my-cluster.yaml}}`
+
+- Create secret from secret spec file:
+
+`kops create -f {{secret.yaml}}`
+
+- Export configurations from a cluster:
+
+`kops export kubecfg {{my-cluster}}`
+
+- Get the cluster configuration as yaml:
+
+`kops get cluster {{my-cluster}} -o yaml`
+
+- Delete cluster:
+
+`kops delete cluster {{my-cluster}} --yes`
--- a/pages/common/kops.md
+++ b/pages/common/kops.md
@@ -5,20 +5,20 @@
 
 - Create a cluster from the configuration specification:
 
-`kops create cluster -f {{my-cluster.yaml}}`
+`kops create cluster -f {{cluster_name.yaml}}`
 
-- Create secret from secret spec file:
+- Create a new ssh public key:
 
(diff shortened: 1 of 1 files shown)


pages/common/kops.md
```diff
@@ -0,0 +1,24 @@
+# kops
+
+> A tool that create, destroy, upgrade and maintain production-grade, highly available, Kubernetes clusters from the command line.
```
mebeim:
```suggestion
> A tool to create, destroy, upgrade and maintain Kubernetes clusters from the command line.
```
Do we really need all those adjectives? This sounds way simpler and more concise to me.

pages/common/kops.md
```diff
@@ -0,0 +1,24 @@
...
+> More informations: <https://github.com/kubernetes/kops/>.
+
+- Create a cluster from the configuration specification:
+
+`kops create cluster -f {{my-cluster.yaml}}`
+
+- Create secret from secret spec file:
```
mebeim:
```suggestion
- Create a secret from secret spec file:
```

pages/common/kops.md
```diff
@@ -0,0 +1,24 @@
...
+
+`kops create -f {{secret.yaml}}`
+
+- Export configurations from a cluster:
```
mebeim: Export to where exactly? Could this be a bit clearer?

pages/common/kops.md
```diff
@@ -0,0 +1,24 @@
...
+
+`kops export kubecfg {{my-cluster}}`
+
+- Get the cluster configuration as yaml:
+
+`kops get cluster {{my-cluster}} -o yaml`
+
+- Delete cluster:
```
mebeim:
```suggestion
- Delete a cluster:
```

pages/common/kops.md
```diff
@@ -0,0 +1,24 @@
...
+
+`kops create cluster -f {{my-cluster.yaml}}`
+
+- Create secret from secret spec file:
+
+`kops create -f {{secret.yaml}}`
+
+- Export configurations from a cluster:
+
+`kops export kubecfg {{my-cluster}}`
```
mebeim:
```suggestion
`kops export kubecfg {{cluster_name}}`
```
Could we have all these tokens use an underscore instead of a dash? Just to be consistent with other pages. Also, let's make it impersonal `cluster_name` vs `my_cluster`.

Same applies to every token in the page.

pages/common/kops.md
Aracki: I will change this example with the better one!

pages/common/kops.md
Aracki: You are right, this is missing. Will be fixed!

pages/common/kops.md
Aracki: 👍 makes sense

PR #3154, opened 2019-06-30
entr: simplify example

mebeim:
The first example of `entr` uses `ag -l` to list all files recursively. `find` does the same thing if launched without arguments.


mebeim:
Uh oh 

> ag and ack offer many advantages over utilities such as find(1) or grep(1) in that they recognize files by their contents and are smart enough to skip directories such as .git

Guess `ag -l` is actually better for this, nevermind!

PR #3155, opened 2019-07-01
HTTPie: add common host header command

josefglatz:
Scenario:

You often must check if a HTTP request to a (new) webserver is working, before the DNS records are pointing to this webserver. The generic example is helpful, but such a specific command example helps ongoing and sporadic webdevs/devops.


sbrl:
Hey, @josefglatz!

I'm not entirely sure how this example differs from the one before it. Could you elaborate please?

josefglatz:
It is a specific example for the one above which is a common scenario in hosting „scene“.

Please decide if such more specific example should be go into it. If not i‘m totally fine when you want to close it without merge 😇

andrik:
If this is a common scenario, I'd suggest merging the examples into only one. 

To me, it seems to be the same scenario discussed in #3147 about using a specific name (concrete example) or a generic one (abstract example)

Aracki:
I would replace `X-MyHeader` with `Host`. 

Everybody knows about `Host` header so it's better example.

sbrl:
Ah, I see! In that case I'd go with @Aracki's suggestion and edit the previous example, adding a concrete example.

Aracki:
@josefglatz we would be happy if you continue with the PR 👍

PR #3164, opened 2019-07-02
awk: fix error

gary-lgy:

### DIFF
--- a/pages/common/awk.md
+++ b/pages/common/awk.md
@@ -11,7 +11,7 @@
 
 `awk '/{{something}}/ {print $2}' {{filename}}`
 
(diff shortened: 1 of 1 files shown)


agnivade:
Have you actually run the command ? ;) It does print the last column

`NF` is the no. of columns. So `{print NF}` does what you are saying. But `$NF` will print the last column.

gary-lgy:
Ah, my bad. `$NF` indeed prints the last column. I tried running the command but missed the `$`, and a quick `man` page search for `NF` led me to the wrong conclusion.
Anyway, sorry for the trouble and thanks for responding quickly!

Issue #3161, opened 2019-07-02
How to change language using a client?


liuhaojava:
How to Change Language?

waldyrious:
According to the [client spec](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#language), it is expected that clients will determine what language to show based on the environment variables such as `LANG`.

@sbrl, wouldn't it make sense to also recommend a `--lang` or `--language` [CLI argument](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#arguments) to switch the language, just like we have `--platform` to override the auto-detected operating system?

(By the way, maybe we should define versioning for the spec, so that clients can indicate to users exactly what version they support and remain compliant even when future changes to the spec are introduced. What do you guys think?)

mebeim:
Hi @liuhaojava, it depends on which client you are currently using. Not many clients support changing language or identifying language at all as of now. The only one which I am aware of is [`elecprog/tldr`](https://github.com/elecprog/tldr).

mebeim:
Also, aren't we already versioning the client spec? I see it marked as v1.0.

sbrl:
@waldyrious Thinking about it, it absolutely would (though you can always do `LANG=abc tldr example`, but that feels a bit weird). Have any clients actually added multi-language support yet? We've got quite a selection of translated pages to test with now.

I'll open a PR to update the client spec.

Maybe it's worth @mentioning client authors in the PR to make them aware of it.

agnivade:
@liuhaojava  - I am going to close this now as your question has been answered. If you want details on a specific client, please feel free to open an issue in that client's repo. Thanks.

PR #3168, opened 2019-07-03
Client spec v1.2: new --language flag, BCP47 to POSIX and other clarifications

sbrl:
As per the discussion in #3161, this PR adds a new required `--language` CLI argument to the spec. Once clients have good support for it, we may consider updating the tldr page for tldr (though we should be careful to mention there that pages are returned in the system language by default).

Ping @pepa65 ([tldr-bash-client](https://gitlab.com/pepa65/tldr-bash-client), has language support already), @elecprog ([elecprog/tldr](https://github.com/elecprog/tldr), also has language support), @felixonmars ([tldr-python-client](https://github.com/tldr-pages/tldr-python-client)), @agnivade ([tldr-node-client](https://github.com/tldr-pages/tldr-node-client)) - please feel free to @mention anyone else who maintains a client that I've missed (I'm probably missing a bunch).

### DIFF
--- a/CLIENT-SPECIFICATION.md
+++ b/CLIENT-SPECIFICATION.md
@@ -98,18 +98,16 @@ Command name    | Mapped name     | Filename
 
 
 ### Translations
-Other directories sit alongside the main `pages` directory, and contain translations of the master versions of every page - though pages MAY NOT have a translation available for a given language yet. Furthermore, a given language MAY NOT have a folder yet either. The format of these directories is `pages.{language-tag}`, where `{language-tag}` is a [BCP 47](https://tools.ietf.org/html/bcp47) conforming tag  in the form of `<language>-<region>`, where:
+Other directories sit alongside the main `pages` directory, and contain translations of the master versions of every page - though pages MAY NOT have a translation available for a given language yet. Furthermore, a given language MAY NOT have a folder yet either. The format of these directories is `pages.<locale>`, where `<locale>` is a [POSIX Locale Name](https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html#Locale-Names) in the form of `<language>_<country>`, where:
 
  - `<language>` is the shortest [ISO 639](https://en.wikipedia.org/wiki/ISO_639) language code for the chosen language (see [here](https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes) for a complete list).
- - `<region>` is the two-letter [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) region code for the chosen region (see [here](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements) for a complete list).
+ - `<country>` is the two-letter [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) country code for the chosen region (see [here](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements) for a complete list).
 
 Some examples:
 
- - Chinese (Taiwan): `pages.zh-TW`.
- - Portuguese (Brazil): `pages.pt-BR`.
- - German (Germany): `pages.de-DE`.
-
-You can check the validity of BCP 47 tags [here](http://schneegans.de/lv/).
+ - Chinese (Taiwan): `pages.zh_TW`.
+ - Portuguese (Brazil): `pages.pt_BR`.
+ - Italian: `pages.it`.
 
 The structure inside these translation folders is identical to that of the main `pages` folder.
 
@@ -176,9 +174,10 @@ Additionally, clients MAY automatically update the cache on a regular basis.
 
 ## Changelog
  - [v1.2, July 3rd 2019](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md) (#3168)
-   - Added new `-L, --language` recommended command-line argument.
-   - Reworded the language section also encouraging the use of configuration files for language.
- - [v1.1, April 1st 2019](https://github.com/tldr-pages/tldr/blob/fbdc06b7425f92cc0d4fc9a5cfc5860ef017251e/CLIENT-SPECIFICATION.md) (#2859)
+   - Addition of a new `-L, --language` recommended command-line argument.
+   - Rewording of the language section also encouraging the use of configuration files for language.
+   - Shift from BCP-47 to POSIX style locale tags, with consequent **deprecation of previous versions of the spec**.
+ - [v1.1, April 1st 2019](https://github.com/tldr-pages/tldr/blob/fbdc06b7425f92cc0d4fc9a5cfc5860ef017251e/CLIENT-SPECIFICATION.md) (deprecated) (#2859)
    - Clarified platform section.
- - [v1.0, January 23rd 2019](https://github.com/tldr-pages/tldr/blob/f5be8a2614a455288f26e42953efeb8cb3bc50b0/CLIENT-SPECIFICATION.md) (#2706)
+ - [v1.0, January 23rd 2019](https://github.com/tldr-pages/tldr/blob/f5be8a2614a455288f26e42953efeb8cb3bc50b0/CLIENT-SPECIFICATION.md) (deprecated) (#2706)
    - Initial release.
--- a/CLIENT-SPECIFICATION.md
+++ b/CLIENT-SPECIFICATION.md
@@ -186,14 +186,12 @@ The [`LANGUAGE` environment variable](https://www.gnu.org/software/gettext/manua
 The [`LC_MESSAGES` environment variable](https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html) MAY be present. If the client itself is localized and this environment variable is present, it MUST use its value in order to determine the language in which interface text is shown (separately from the language used for pages). In absence of `LC_MESSAGES`, then `LANG` and `LANGUAGE` MUST be used for this purpose instead.
 
 
-Other Considerations
---------------------
+Caching
+-------
 
-This section contains a number of other items that don't neatly fit into any of the sections defined above.
+If appropriate, it is RECOMMENDED that clients implement a cache of pages. If implemented, clients MUST download the archive either from **[http://tldr.sh/assets/tldr.zip](http://tldr.sh/assets/tldr.zip)** or [https://raw.githubusercontent.com/tldr-pages/tldr-pages.github.io/master/assets/tldr.zip](https://raw.githubusercontent.com/tldr-pages/tldr-pages.github.io/master/assets/tldr.zip) (which is pointed by the first link).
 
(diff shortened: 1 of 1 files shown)


sbrl:
Also, should we perhaps add in a clause that says that clients should mention what version of this spec they implement in their readmes?

CLIENT-SPECIFICATION.md
```diff
@@ -38,6 +42,7 @@ Argument			| Required?	| Meaning
 `-p`, `--platform`	| Yes		| Specifies the platform to be used to perform the action (either listing or searching). If this option is specified, the selected platform MUST be checked first instead of the current platform as described below.
 `-u`, `--update`	| Conditional	| Updates the offline cache of pages. MUST be implemented if cache is supported.
 `-l`, `--list`		| No		| Lists all the pages in the current platform to the standard output. If the special platform `all` is specified a list of all pages in all platforms MUST be displayed.
+`--language`		| Yes		| Specifies the preferred language for the page returned. Overrides the `LANG` environment variable - see the [section on translations](#translations) for more information.
```
agnivade: Ahh .. `-l` is already taken. Have you thought of choosing some other single letter to have a single letter flag ? I don't know .. maybe -e, -g ?

mebeim:
As far as I am concerned, having a language flag is not *necessary*. All localized command line programs I've seen use environment variables to detect the language. See for example `man man`. We should ask clients to automatically detect language (as we already do), with an additional encouragement to make it configurable for the user (via configuration files or command-line options), but I would not suggest it as a must have feature.

CLIENT-SPECIFICATION.md
mebeim:
```suggestion
`-L, --language`		| No		| Specifies the preferred language for the page returned. Overrides other language detection mechanisms. See the [language section](#language) for more information.
```

CLIENT-SPECIFICATION.md
```diff
@@ -1,15 +1,19 @@
...
 
- - **Current Specification Version:** 1.1
+ - **Current Specification Version:** 1.2
 
 This document contains the official specification for tldr-pages clients. It is _not_ a specification of the format of the pages themselves - only a specification of how a user should be able to interface with an official client.
 
 The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
 
 ## Changelog
```
mebeim: Since this changelog is going to get longer and longer, I would suggest moving this section at the end of the file, referencing it in the first paragraph, like this (bold text added by me):

> This document contains the official specification for tldr-pages clients. It is _not_ a specification of the format of the pages themselves - only a specification of how a user should be able to interface with an official client. **For a list of previous versions of the specification, see the [changelog section](#Changelog) below.**

I would also suggest to link to specific versions of the file along with PR numbers, and have the list in reverse order (most recent first), like this (see source of this comment for the Markdown):

 - [v1.2, July 3rd 2019](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md) (#3168)
   - Added new `-L, --language` recommended command-line argument.

 - [v1.1, April 1st 2019](https://github.com/tldr-pages/tldr/blob/fbdc06b7425f92cc0d4fc9a5cfc5860ef017251e/CLIENT-SPECIFICATION.md) (#2859)
   - Clarified platform section.

 - [v1.0, January 23rd 2019](https://github.com/tldr-pages/tldr/blob/f5be8a2614a455288f26e42953efeb8cb3bc50b0/CLIENT-SPECIFICATION.md) (#2706)
   - Initial release.

sbrl: Done :D

CLIENT-SPECIFICATION.md
sbrl: Ah, good idea. Not sure what to choose that would make sense, actually.

sbrl:
Great point, @mebeim. Ideally we want to encourage the user to use the environment variable, since that's a generally accepted standard.

Should we keep the `--language` after all then? :thinking:

mebeim:
@sbrl yes like I said I think we should encourage to have such flag or a config file, but not make it a must have feature.

agnivade:
in progress

principis:
> - please feel free to @mention anyone else who maintains a client that I've missed (I'm probably missing a bunch).

@sbrl tldr-sharp had language support since 1.2.0 ;)
Good to see that it'll be included in the spec. :)

mebeim:
As promised in [#3183 (comment)](https://github.com/tldr-pages/tldr/issues/3183#issuecomment-512705073), I've integrated in this PR the changes regarding language and language tags (or, more precisely, locale tags), and I've changed the title accordingly. Since I believe that having fewer versions is better to avoid confusion, I also integrated some other changes.

I think this PR should be rebased in order to maintain the meaningfulness of the commits, but it could be also squashed: in such case please *use the summary I'm adding below*. Forgive the force-pushing, you can still see the commit history clearly [from the relative tab](https://github.com/tldr-pages/tldr/pull/3168/commits).

Here's a summary of what happens in version v1.2 of the client specification:

- Addition of the recommended --language command-line option.
- Shift from BCP-47 tags to POSIX locale names, with *deprecation of previous spec versions*.
- Overall reformatting of the document: use of `---` and `===` to define section headers and more spacing, better spacing in tables using spaces instead of tabs.
- Rewording and reordering of the language section, also encouraging the use of configuration files for language.
- Rewording of the term "arguments" into "options`: the values described in the table are command-line options, "argument" is a more general term.
- Conversion of the "Other Consideration" section into "Caching" with additional clarifications: since the only thing being discussed under this section was caching.

CLIENT-SPECIFICATION.md
```diff
@@ -1,45 +1,49 @@
...
+-------------------|-------------|----------
+`-v`, `--version`  | Yes         | Shows the current version of the client, and the version of this specification that it implements.
+`-p`, `--platform` | Yes         | Specifies the platform to be used to perform the action (either listing or searching) as an argument. If this option is specified, the selected platform MUST be checked first instead of the current platform as described below.
+`-u`, `--update`   | Conditional | Updates the offline cache of pages. MUST be implemented if cache is supported.
+`-l`, `--list`     | No          | Lists all the pages in the current platform to the standard output. If the special platform `all` is specified a list of all pages in all platforms MUST be displayed.
+`-L`, `--language` | No          | Specifies the preferred language for the page returned. Overrides other language detection mechanisms. See the [language section](#language) for more information.
 
-Clients MAY NOT implement the long form of an argument (`--update`, `--version`, `--list`, `--platform`).
+Clients MAY choose to onlu implement the short version of an option, ignoring the long form.
```
agnivade: only

CLIENT-SPECIFICATION.md
```diff
@@ -148,32 +159,53 @@ Steps #3 and #4 may be done in either order.
 
 It is possible that due to this page resolution logic, the client may show a page which does not belong to the host platform because a page can reside in `common`, and not be present on the host platform. Clients must not assume that a given command is always executable on the host platform.
 
-#### If a page isn't found
+#### If a page isn not found
```
agnivade: is

agnivade:
Just gave some minor comments. Looks good otherwise.

> and the change of the spec before updating the client spec as to not create confusion.

Which other spec did you mean ?

mebeim:
@agnivade sorry, bad wording, not really a "spec", I was talking about CONTRIBUTING.md, we still talk about BCP-47 there. I will update it along with folder names in another PR.

https://github.com/tldr-pages/tldr/blob/4445f03f3f7e0b1d5761bca44e6658d2f00b3579/CONTRIBUTING.md#L78

agnivade:
Got it. Thanks.

principis:
Some thought I just had, what is more important, platform or language?
Not sure if this should be added to the spec.
But for example there is a page "X" and we have our default language nl-BE. LANGUAGE=fr:de:en and the platform is linux.
So what I'm doing now is checking every language for the Linux platform if it doesn't exist, look for the "common" platform. If that doesn't exist look for other platform and show a warning.

But maybe we should look for Linux and common first for each language before going to the next language?

mebeim:
@principis the lookup should be like this:

    $ LANGUAGE="it_IT:it:en_US:en" tldr some-page

    pages.it_IT -> does not exist
    pages.it/linux/some-page.md -> does not exist
    pages.it/common/some-page.md -> does not exist
    pages.en_US -> does not exist
    pages/linux/some-page.md -> does not exist
    pages/common/some-page.md -> FOUND!

    Display pages/common/some-page.md

sbrl:
Hrm, that is indeed an issue. Looks like we've got 2 options:

 - Language-first: Choose a page from another platform and prefer pages in the specified language
 - Platform-first: Choose a page in another language in order to keep the platform the same.

I'd probably say platform-first is the better option:

```
pages.it/linux/some-page.md -> does not exist
pages/linux/some-page.md -> does not exist
pages.it/common/some-page.md -> does not exist
pages/common/some-page.md -> FOUND!
```

mebeim:
@sbrl yes, that's what I was thinking. Platform first makes more sense. We cannot do much except having `common` pages only if the page is the same on all systems. It's up to the client developers to choose which of the two methods makes more sense.

principis:
_(So I wrote a long explanation but then had to leave and now it's gone...)_
IMO platform is more important, I've currently implemented it like @sbrl mentioned. (With a fallback to English). But maybe this doesn't need to be in the spec? I did spend some time how to implement it correctly and consistent, so it could be added in the spec as a suggestion.

I think language-first really shouldn't be implemented (depends on how you do it). It could for example show a `windows` page on the linux platform because it's not in your first selected language, for example `it`, but in your second, `en`, there is a page for `linux`. 

Platform-first doesn't have this problem.

mebeim:
@principis yes I agree that platform first is really the only one which makes sense, and I also agree that it should be in the spec at least as a "SHOULD" or "RECOMMENDED" clause. I will add it to this PR as soon as I can. Language-first could cause problems for users in situations like the one you just explained.

sbrl:
@mebeim is working on this @probot.

mebeim:
Welp, I almost forgot about this, thank you @probot :sweat_smile: 

I pushed the last commit applying @principis suggestions. After this, it should be pretty much ready to merge. Let me know what you guys think.

mebeim:
Ping @agnivade, looks fine to merge for me. Any comments?

sbrl:
Apparently I can't approve my own PR, but this looks fine to me :D

sbrl:
We should have some way of @mentioning all client authors at once.

principis:
> Welp, I almost forgot about this, thank you @probot sweat_smile
> 
> I pushed the last commit applying @principis suggestions. After this, it should be pretty much ready to merge. Let me know what you guys think.

(I really need to fix my notifications... Didn't see this.) it's too late but still, looks good, thanks!

PR #3173, opened 2019-07-05
tsc: add page

waldyrious:

### DIFF
--- a/pages/common/tsc.md
+++ b/pages/common/tsc.md
@@ -15,14 +15,14 @@
 
 `tsc --outFile {{output.js}} {{input.ts}}`
 
-- Get command line options and files from a file:
-
-`tsc @{{args.txt}}`
-
-- Compile a TypeScript project if outdated:
+- Compile all `.ts` files of a TypeScript project defined in a `tsconfig.json` file:
 
 `tsc --build {{tsconfig.json}}`
 
-- Type-check JavaScript files and output only the errors:
+- Run the compiler using command line options and arguments fetched from a text file:
+
+`tsc @{{args.txt}}`
+
+- Type-check multiple JavaScript files, and output only the errors:
 
 `tsc --allowJs --checkJs --noEmit {{src/**/*.js}}`
--- a/pages/common/tsc.md
+++ b/pages/common/tsc.md
@@ -1,0 +1,28 @@
+# tsc
+
+> TypeScript compiler.
+> More information: <https://www.typescriptlang.org/docs/handbook/compiler-options.html>.
+
+- Compile a TypeScript file `hello.ts` into a JavaScript file `hello.js`:
+
+`tsc {{foo.ts}}`
+
+- Compile a TypeScript file into JavaScript using a specific target syntax (default is `ES3`):
+
+`tsc --target {{ES5|ES2015|ES2016|ES2017|ES2018|ESNEXT}} {{foo.ts}}`
+
+- Compile a TypeScript file into a JavaScript file with a custom name:
+
+`tsc --outFile {{foo.js}} {{bar.ts}}`
+
+- Get command line options and files from a file:
+
+`tsc @{{args.txt}}`
+
+- Compile a TypeScript project if outdated:
+
+`tsc --build {{tsconfig.json}}`
+
+- Type-check JavaScript files and output only the errors:
+
+`tsc --allowJs --checkJs --noEmit {{src/**/*.js}}`
--- a/pages/common/tsc.md
+++ b/pages/common/tsc.md
@@ -3,17 +3,17 @@
 > TypeScript compiler.
 > More information: <https://www.typescriptlang.org/docs/handbook/compiler-options.html>.
 
-- Compile a TypeScript file `hello.ts` into a JavaScript file `hello.js`:
+- Compile a TypeScript file `foobar.ts` into a JavaScript file `foobar.js`:
 
-`tsc {{foo.ts}}`
+`tsc {{foobar.ts}}`
 
 - Compile a TypeScript file into JavaScript using a specific target syntax (default is `ES3`):
 
-`tsc --target {{ES5|ES2015|ES2016|ES2017|ES2018|ESNEXT}} {{foo.ts}}`
+`tsc --target {{ES5|ES2015|ES2016|ES2017|ES2018|ESNEXT}} {{foobar.ts}}`
 
 - Compile a TypeScript file into a JavaScript file with a custom name:
 
-`tsc --outFile {{foo.js}} {{bar.ts}}`
+`tsc --outFile {{output.js}} {{input.ts}}`
 
 - Get command line options and files from a file:
 


waldyrious:
For future reference, here are the options this command accepts, and their meaning. I'm not sure this is exhaustive or fully up-to-date; but in any case it shouldn't be too far from the current UI:

```
--allowJs                              Allow javascript files to be compiled.
--allowUnreachableCode                 Do not report errors on unreachable code.
--allowUnusedLabels                    Do not report errors on unused labels.
--alwaysStrict                         Parse in strict mode and emit "use strict" for each source file.
--baseUrl                              Base directory to resolve non-absolute module names.
--build, -b                            Build one or more projects and their dependencies, if out of date
--declaration, -d                      Generates corresponding '.d.ts' file.
--declarationMap                       Generates a sourcemap for each corresponding '.d.ts' file.
--esModuleInterop                      Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'.
--experimentalDecorators               Enables experimental support for ES7 decorators.
--forceConsistentCasingInFileNames     Disallow inconsistently-cased references to the same file.
--init                                 Initializes a TypeScript project and creates a tsconfig.json file.
--lib                                  Specify library files to be included in the compilation: 'es5' 'es6' 'es2015' 'es7'...
--mapRoot LOCATION                     Specify the location where debugger should locate map files instead of generated locations.
--maxNodeModuleJsDepth                 The maximum dependency depth to search under node_modules and load JavaScript files
--noEmit                               Do not emit outputs.
--noEmitOnError                        Do not emit outputs if any errors were reported.
--noFallthroughCasesInSwitch           Report errors for fallthrough cases in switch statement.
--noImplicitAny                        Raise error on expressions and declarations with an implied 'any' type.
--noImplicitReturns                    Report error when not all code paths in function return a value.
--noImplicitThis                       Raise error on 'this' expressions with an implied 'any' type.
--noImplicitUseStrict                  Do not emit 'use strict' directives in module output.
--noUnusedLocals                       Report errors on unused locals.
--noUnusedParameters                   Report errors on unused parameters.
--outDir DIRECTORY                     Redirect output structure to the directory.
--outFile FILE                         Concatenate and emit output to single file.
--preserveConstEnums                   Do not erase const enum declarations in generated code.
--pretty                               Stylize errors and messages using color and context (experimental).
--project DIRECTORY, -p DIRECTORY      Compile the project in the given directory.
--project CONFIG_FILE, -p CONFIG_FILE  Compile the project given the path to its configuration file.
--removeComments                       Do not emit comments to output.
--rootDir LOCATION                     Specify the root directory of input files. Use to control the output directory structure with --outDir.
--skipLibCheck                         Skip type checking of declaration files.
--sourceMap                            Generates corresponding '.map' file.
--sourceRoot LOCATION                  Specify the location where debugger should locate TypeScript files instead of source locations.
--strict                               Enable all strict type-checking options.
--strictBindCallApply                  Enable strict 'bind', 'call', and 'apply' methods on functions.
--strictFunctionTypes                  Enable strict checking of function types.
--strictNullChecks                     Enable strict null checks.
--strictPropertyInitialization         Enable strict checking of property initialization in classes.
--suppressImplicitAnyIndexErrors       Suppress noImplicitAny errors for indexing objects lacking index signatures.
--traceResolution                      Enable tracing of the name resolution process.
--types                                Type declaration files to be included in compilation.
--target VERSION, -t VERSION           Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'.
--version, -v                          Print the compiler's version.
--watch, -w                            Watch input files.
@<file>                                Get command line options and files from a file.
```

pages/common/tsc.md
```diff
@@ -0,0 +1,28 @@
+# tsc
+
+> TypeScript compiler.
+> More information: <https://www.typescriptlang.org/docs/handbook/compiler-options.html>.
+
+- Compile a TypeScript file `hello.ts` into a JavaScript file `hello.js`:
```
agnivade: Perhaps change these to foo.ts and foo.js too, to be consistent with the examples ?

pages/common/tsc.md
```diff
@@ -0,0 +1,28 @@
...
+
+`tsc {{foo.ts}}`
+
+- Compile a TypeScript file into JavaScript using a specific target syntax (default is `ES3`):
+
+`tsc --target {{ES5|ES2015|ES2016|ES2017|ES2018|ESNEXT}} {{foo.ts}}`
+
+- Compile a TypeScript file into a JavaScript file with a custom name:
+
+`tsc --outFile {{foo.js}} {{bar.ts}}`
```
agnivade: Does --outFile {{bar.js}} {{foo.ts}} sound better ? So that the foo -> bar direction is maintained.

pages/common/tsc.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Get command line options and files from a file:
+
+`tsc @{{args.txt}}`
+
+- Compile a TypeScript project if outdated:
```
agnivade: What does "if outdated" mean ?

pages/common/tsc.md
waldyrious: Sure, makes sense.

pages/common/tsc.md
waldyrious: I don't know this command that well, to be honest; this particular example was pretty much copied verbatim from the documentation.

I take it to mean something similar to other compilers, i.e. only source files that have changed since the last compilation (perhaps validated via some hash or a cache, not sure) will be re-compiled.

waldyrious: The [documentation](https://www.typescriptlang.org/docs/handbook/project-references.html#build-mode-for-typescript) doesn't seem to help much in clarifying what they mean, so I'll assume it's indeed the common compiler behavior I described above.

pages/common/tsc.md
waldyrious: Ah, my bad. Will update.

pages/common/tsc.md
waldyrious: Actually, since this is precisely an example to show usage of custom names, let's go with something more explicit: "input.ts" and "output.js".

agnivade: Sure

pages/common/tsc.md
agnivade: Ah I see. I was confused because you mentioned "project" as a whole. Whereas this means, compile only the files in the project which have changed. Perhaps clarify that ?

sbrl: I guess it checks the mtime of the files involved, and only recompiles if the source files have an mtime that's after the destination file's mtime?

waldyrious: I have no idea how the check is implemented, but yeah, it's per-file. I'll reword to clarify.

waldyrious: @agnivade, @sbrl: I went back and forth and ended up rephrasing the example differently. I would say the "if outdated" part is not the main point to understand about the `--build` option, and it was getting the description too long, so I removed that bit. Here's the diff, from  608a8dc:

```diff
-- Compile a TypeScript project if outdated:
+- Compile all `.ts` files of a TypeScript project defined in a `tsconfig.json` file:
 
 `tsc --build {{tsconfig.json}}
```

WDYT?

sbrl: Nice!

agnivade: Agree

PR #3181, opened 2019-07-07
rg: add filename search

Mordavolt:

The next line in the document is very similar to the PR one, but I'd argue that the use case is different, and there is no way to merge these two examples.

### DIFF
--- a/pages/common/rg.md
+++ b/pages/common/rg.md
@@ -19,6 +19,10 @@
 
 `rg {{pattern}} {{set_of_subdirs}}`
 
+- Search for a file with a name matching a pattern (e.g., `*.csv`):
+
+`rg -g {{pattern}} --files`
+
 - Search for a pattern in files matching a glob (e.g., `README.*`):
 
 `rg {{pattern}} -g {{glob}}`


pages/common/rg.md
```diff
@@ -19,6 +19,10 @@
 
 `rg {{pattern}} {{set_of_subdirs}}`
 
+- Search for a file with a name matching a pattern (e.g., `*.csv`):
+
+`rg -g {{pattern}} --files`
+
 - Search for a pattern in files matching a glob (e.g., `README.*`):
 
 `rg {{pattern}} -g {{glob}}`
```
waldyrious: It's not clear how the example added above differs from this one. Isn't that behavior equivalent to using the `find` command? Or does it support more complex regex patterns that simple globs?

In any case, I think it may be better to either merge these two examples, or synchronize the descriptions (and command structure) to make it clearer how they relate to each other.

waldyrious:
Thanks @Mordavolt! I refer to this page all the time so it's nice to see improvements to it. I've left a comment inline, please take a look.

pages/common/rg.md
Mordavolt: Basically, I feel this use case is justified as a separate case because it is a very iconic and standalone use case. "Find a file by filename or part of filename".
Essentially I feel that these are 3 distinct use cases, even though they are achieved by a combination of the same params:
- `rg -g {{pattern}} --files`
Find all files that match the filename pattern.
- `rg {{pattern}} -g {{glob}}`
Find all occurrences of the pattern in named files.
- `rg --files-with-matches {{pattern}}`
Find all files where we did find a match for the pattern.

I understand that it might make more sense to merge then, but as I've expressed in the original PR message, I don't think that is possible. Maybe someone else has some idea on what would be the best way to merge them?

waldyrious: It's not imperative to merge them :) but it would at least be desirable to use similar terminology. For instance, does the `-g` support more advanced regex syntax than just regular glob patterns? In that case it would make sense to change the second example (the existing one) to use `{{pattern}}` as well, instead of `{{glob}}`. (The option name seems to suggest that it's not the case, though.)

If, as I suspect, the patterns that `-g` allows are also supported by `find`, then I'd say this particular task (searching by filename alone rather than content) is better suited to the latter command. Tldr-pages don't need to be exhaustive, and should instead focus on the core functionality that the command provides. Bumping into the page size limit for an example that is not unique nor the main functionality of a command may not be the best use of the limited space.

Of course, this is just my opinion. I'm open to being convinced otherwise! :)

sbrl:
Hey, @Mordavolt! Any reason for closing this?

Mordavolt:
> Hey, @Mordavolt! Any reason for closing this?

I've succumbed to this comment:
> If, as I suspect, the patterns that -g allows are also supported by find, then I'd say this particular task (searching by filename alone rather than content) is better suited to the latter command. 

This is exactly the way I've used the command, could never quite figure out the `find` syntax. For a couple of weeks was trying to compose a nice response, but realised I must be misusing ripgrep, so the whole PR fell apart. I'll continue using `rg` this way, but now I doubt other people should too.

waldyrious:
@Mordavolt I appreciate you taking the time to tell us about your reasoning! For what it's worth, I use tldr all the time for precisely that use case — including (perhaps embarrassingly often) to confirm the syntax of commands like `grep` and `find`. I find it helpful to ensure I'm staying true to the philosophy of using the right tool for the job ("do one thing and do it well") and composing such specialized tools together.

In any case, thanks for the PR!

PR #3185, opened 2019-07-10
Travis: separate building from deployment

### LINKED ISSUE #3148: Automated deploy of updated ZIP archive failing
See the latest build [here](https://travis-ci.org/tldr-pages/tldr/builds/551699390#L215), and also [previous builds](https://travis-ci.org/tldr-pages/tldr/builds) of the master branch. They seem to randomly fail to deploy the ZIP archive because this particular command fails:

https://github.com/tldr-pages/tldr/blob/241155bfd72de97c9b22ebddb7e66ac2dfa87cf9/scripts/deploy.sh#L36

For example:

    (adding: pages.it/common/clockwork-cli.md (deflated 51%)
    (adding: pages.it/common/arch.md (deflated 28%)
    (adding: pages.it/common/alex.md (deflated 42%)
    (stored 0%) <------------------------------- this seems suspicious
    Script failed with status 1

NOTE: this is a problem in the deploy step, the build is still working fine (in fact Travis says "error", not "failed"). We should check what's wrong.

mebeim:
Fixes #3148.

I finally figured this out. After looking at the output of some failed builds, it seems that the build always crashes around ~1100 lines of output: for some reason Travis chokes on the very long output of `zip -r pages* ...`. The immediate fix for this was to add `-q` to make the `zip` command quiet. This also means there is no need to use `travis_retry` (or any sort of retry logic at all).

Other than this, as previously discussed in #3148 I also updated the build process to be more consistent. To summarize, the changes are the following:

 1. Added `-q` to the `zip` command to create the archive.
 2. Moved the actual building of the index and the archive from `deploy.sh` to a new `build.sh` script.
 3. Added a `script:` step to `.travis.yml` to run `build.sh` after `npm test`.
 4. Specified a clone depth of 1 in `.travis.yml` to make things faster since we don't need to do any git operation during the build.

---

NOTE: for obvious reasons, I could not actually test the deployment of the updated ZIP archive. The code that deploys stuff to `tldr-pages.github.io` in `deploy.sh` is basically left untouched, so it should all work. Just to be safe though, to anyone who merges this: please **keep an eye on the build when this is merged** and make sure everything works fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,13 @@ language: node_js
 node_js: stable
 cache: false
 
+git:
+  depth: 1
+
+script:
+  - npm test
+  - bash scripts/build.sh
+
 after_failure:
   - python scripts/send_to_bot.py < test_result
 
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+# This script is executed by Travis CI for every successful push (on any branch, PR or not).
+set -ex
+
+function initialize {
+  if [ -z "$TLDRHOME" ]; then
+    export TLDRHOME=${TRAVIS_BUILD_DIR:-$(pwd)}
+  fi
+
+  export TLDR_ARCHIVE="tldr.zip"
+}
+
+function build_index {
+  npm run build-index
+  echo "Pages index succesfully built."
+}
+
+function build_archive {
+  rm -f "$TLDR_ARCHIVE"
+  cd "$TLDRHOME/"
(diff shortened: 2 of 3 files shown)


agnivade:
LGTM. Can you update the commit message with the text in the PR body, so that this information is there later for somebody who wants to see what changed ?

Thanks for doing this.

mebeim:
@agnivade don't know why I didn't think about that, good idea. Done.

pxgamer:
Well, looks like it built successfully 🤔
https://travis-ci.org/tldr-pages/tldr/builds/557282684

agnivade:
Yeah. I checked the zip file in the tldr-pages repo. Seems like it is fine.

PR #3190, opened 2019-07-12
gixy: add page

Aracki:
I have added [gixy](https://github.com/yandex/gixy) - a tool for analyzing nginx configuration files.

### DIFF
--- a/pages/common/gixy.md
+++ b/pages/common/gixy.md
@@ -1,0 +1,20 @@
+# gixy
+
+> Analyze Nginx configuration.
+> More information: <https://github.com/yandex/gixy>.
+
+- Analyze nginx configuration (default path: /etc/nginx/nginx.conf):
+
+`gixy`
+
+- Analyze nginx configuration but skip specific tests: 
+
+`gixy --skips {{http_splitting}}`
+
+- Analyze nginx configuration with the specific level:
+
+gixy {{-l|-ll|-lll}}
+
+- Analyze nginx configuration on the specific path:
+
+`gixy {{config_path}}`
--- a/pages/common/gixy.md
+++ b/pages/common/gixy.md
@@ -17,4 +17,4 @@
 
 - Analyze nginx configuration on the specific path:
 
-`gixy {{config_path}}`
+`gixy {{path/to/configuration_file}}`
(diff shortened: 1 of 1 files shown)


pages/common/gixy.md
```diff
@@ -0,0 +1,20 @@
+# gixy
+
+> Analyze Nginx configuration.
+> More information: <https://github.com/yandex/gixy>.
+
+- Analyze nginx configuration (default path: /etc/nginx/nginx.conf):
```
pxgamer: Should Nginx be stylised the same everywhere (i.e. Nginx, NGINX, nginx) for consistency?

pages/common/gixy.md
```diff
@@ -0,0 +1,20 @@
...
+
+`gixy`
+
+- Analyze nginx configuration but skip specific tests:
+
+`gixy --skips {{http_splitting}}`
+
+- Analyze nginx configuration with the specific level:
```
pxgamer: Specific level of? :slightly_smiling_face:

pages/common/gixy.md
```diff
@@ -0,0 +1,20 @@
...
+
+`gixy {{-l|-ll|-lll}}`
+
+- Analyze nginx configuration on the specific path:
+
+`gixy {{config_path}}`
```
pxgamer:
```suggestion
`gixy {{path/to/configuration_file}}`
```
Unless it's a config directory, in which case `configuration_directory` :slightly_smiling_face:

pages/common/gixy.md
```diff
@@ -0,0 +1,20 @@
+# gixy
+
+> Analyze nginx configuration.
```
sbrl: Perhaps this will read better?
```suggestion
> Analyze nginx configuration files.
```

Aracki: Singular is correct in this case.

sbrl: That depends. Many configurations will have multiple nested configuration files in directories such as `sites-available` and `conf.d`. In addition, if we used the singular here we'd need to reword it to be `Analyze an nginx configuration file` to add an article, though this would imply that it can only check a single configuration file, @Aracki

Aracki: Thanks, @sbrl for noticing this. I checked it, and it can check multiple files. So I modify the title and last example.

PR #3075, opened 2019-06-03
Refactor: update english and italian pages' links' description.

mebeim:
Followup on #3060.

This PR updates the wording of description links of all English and Italian pages to `More information: <...>` (italian `Maggiori informazioni: <...>`).

If anybody wants to take care of other languages, here's the script I used:

    sed -i 's/> .\+:.\+</> More information: </' pages/**/*

Pinging @schneiderl and @andrik in case one of you is interested.

PR #3193, opened 2019-07-12
portuguese pages: changed "More information" token

schneiderl:
This PR changes the "Página Oficial" token in order to match the english page tokens. 

It is similar to changing from "Official page" to "More Information".

### DIFF
--- a/pages.pt-BR/common/7z.md
+++ b/pages.pt-BR/common/7z.md
@@ -1,7 +1,7 @@
 # 7z
 
 > Um compactador de arquivos com alta taxa de compressão.
-> Página oficial: <https://www.7-zip.org/>.
+> Mais informações: <https://www.7-zip.org/>.
 
 - Compactar um arquivo ou diretório:
 
--- a/pages.pt-BR/common/7za.md
+++ b/pages.pt-BR/common/7za.md
@@ -2,7 +2,7 @@
 
 > Um compactador de arquivos com alta taxa de compressão.
 > Versão compacta do `7z`, com suporte para menos tipos de arquivamento/compressão.
-> Página oficial: <https://www.7-zip.org/>.
+> Mais informações: <https://www.7-zip.org/>.
 
 - Compactar um arquivo ou diretório:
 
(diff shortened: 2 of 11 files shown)


pxgamer:
I'd say this could be squashed as they are all related. But rebasing would be fine too as the messages are consistent and neat.

schneiderl:
According to the maintainers guide:

>  instead the PR author took the time to craft individual, informative messages for each commit, then use the Rebase and merge method, to honor that work and preserve the history of the changes. For less clear-cut cases, a simple heuristic you can follow is that if there are more "dirty" commits than "clean" commits, then prefer squash, else do a rebase.


That being said, I think rebasing is fine here.

waldyrious:
For future reference, those guidelines are meant to apply to distinct, unrelated changes, e.g. adding a new page, moving another to the correct platform folder, fixing a typo in the README, etc.

An extreme example of where the rule wouldn't apply is, for instance, a PR fixing a file that used a dash (—) instead on a hyphen (-) in the example descriptions, with the PR having one commit per each line: "Fix dash to hyphen in first example of page-foo.md", "Fix dash to hyphen in second example of page-foo.md", and so on. In that case, it would make more sense to make all those changes in a single step, since they're essentially the same change. That applies here as well.

Perhaps a good heuristic to use as a rule of thumb is to think whether it would make sense to revert only one of those commits while keeping the others; for this case, should we decide to go back from "Mais informações" to "Página oficial", then the logical change would be to revert all these commits at once, rather than only a subset of the commits.

Does that make sense to you guys? Is there a way the guidelines could be better phrased to convey these principles?

waldyrious:
Oh, and by the way, just as a matter of principle, maintainers should generally avoid merging their own PRs. We can skip the 2 maintainers approvals rule for trivial changes, but at least it should be a different person hitting the merge button if possible.

mebeim:
@waldyrious don't think you could rephrase the guidelines better without making them harder to follow. I'd say trivial refactoring tasks should most of the times fall under the squashing hammer, like for example what I did for the Italian and English pages in #3075 (it really wouldn't have made sense to have hundreds of commits for that).

mebeim:
Also please don't rush merging something right away if it's not all clear @schneiderl. I should have said it earlier (my bad I guess), but in this case I would have advised to squash this using a commit message similar to mine in #3075 for consistency, like `Refactor: reword Portuguese pages' links' descriptions.`

sbrl:
I think it's a case-by-case basis. In this case, I'd squash it into a single commit personally (but keeping the extended description), but if the changes were more substantial I'd opt for a rebase.

PR #3197, opened 2019-07-16
trash: add page

Fealthas:
for trash-cli

https://github.com/andreafrancia/trash-cli

### DIFF
--- a/pages/linux/trash.md
+++ b/pages/linux/trash.md
@@ -1,0 +1,23 @@
+# trash
+
+> A CLI trashcan/recycling bin
+
+- Delete a file (send to trash):
+
+`trash-put {{path/to/file}}`
+
+- List files in trash:
+
+`trash-list`
+
+- Restore file from trash:
+
+'trash-restore'
+
+- Empty trash:
+
+'trash-empty'
+
+- Empty trash older than {{X}} days:
+
+'trash-empty {{X}}'
--- a/pages/linux/trash.md
+++ b/pages/linux/trash.md
@@ -1,6 +1,6 @@
 # trash
 
-> A CLI trashcan/recycling bin
+> A CLI trashcan/recycling bin.
 
 - Delete a file (send to trash):
 
--- a/pages/linux/trash.md
+++ b/pages/linux/trash.md
@@ -7,10 +7,6 @@
 
 `trash {{path/to/file}}`
 
-or
-
-`trash-put {{path/to/file}}`
-
 - List files in trash:
 
 `trash-list`
--- a/pages/linux/trash.md
+++ b/pages/linux/trash.md
@@ -1,6 +1,7 @@
 # trash
 
 > A CLI trashcan/recycling bin.
+> More information: <https://github.com/andreafrancia/trash-cli>.
 
 - Delete a file (send to trash):
 
--- a/pages/linux/trash.md
+++ b/pages/linux/trash.md
@@ -1,9 +1,13 @@
 # trash
 
-> A CLI trashcan/recycling bin.
+> A CLI for managing your trashcan / recycling bin.
 > More information: <https://github.com/andreafrancia/trash-cli>.
 
 - Delete a file (send to trash):
+
+`trash {{path/to/file}}`
+
+or
 
 `trash-put {{path/to/file}}`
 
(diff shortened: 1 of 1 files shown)


pages/linux/trash.md
```diff
@@ -0,0 +1,23 @@
...
+
+`trash-put {{path/to/file}}`
+
+- List files in trash:
+
+`trash-list`
+
+- Restore file from trash:
+
+`trash-restore`
```
schneiderl: Doesn't that expect a parameter that says which is the file to be restored?

pages/linux/trash.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Empty trash:
+
+`trash-empty`
+
+- Empty trash older than {{X}} days:
```
schneiderl: Maybe this can be rephrased. I think it is better to be clear that files on the trash that are older than X days will be deleted. With this description it feels like the trash bins will be deleted.

pages/linux/trash.md
Fealthas: Yes, but it seems kinda useless to me. 
Trash-list doesn't show the item id.
If you do not pass a parameter, it just lists all the files and waits for you to input a number, corresponding to the item to restore(actually shows numbers with trash-restore).

Fealthas: Or actually no, it doesn't take parameters at all.

Fealthas: Just running the command seems to be the only way to use it.

pages/linux/trash.md
Fealthas: Changed, it. Take a look.

pages/linux/trash.md
schneiderl: Oh thats good then. We can leave it as it is.

pages/linux/trash.md
schneiderl: Reads better now in my opinion. Thank you for your time 👍

pages/linux/trash.md
```diff
@@ -0,0 +1,23 @@
...
+
+`trash-restore`
+
+- Empty trash:
+
+`trash-empty`
+
+- Empty trash, keeping files trashed less than {{X}} days ago:
+
+`trash-empty {{X}}`
```
pxgamer: Perhaps use a concrete example, such as `10`:
```suggestion
`trash-empty {{10}}`
```
And then update the description to:
`Empty trash, keeping files trashed less than 10 days ago:`

pages/linux/trash.md
```diff
@@ -0,0 +1,23 @@
+# trash
+
+> A CLI trashcan/recycling bin.
```
agnivade: Perhaps - "A CLI for managing trash can / recycling bin" sounds better ?

schneiderl: Add the homepage. 
You have alread linked it in your PR description. 
```suggestion
> A CLI trashcan/recycling bin.
> More information: <https://github.com/andreafrancia/trash-cli>.
```

pages/linux/trash.md
Fealthas: fixed

pages/linux/trash.md
Fealthas: fixed

Fealthas:
Should I put just trash or trash-put?
Syntax checker seems to not like the "or".
I guess trash-put is unnecessary to know, as trash just links to trash-put IIRC, but it could be confusing if it is not documented.

schneiderl:
@Fealthas As far as I understood they do the same thing. 
tldr is all about keeping it simple so i believe only one example should be listed. 

I would vote for "trash" because it looks clean.

pages/linux/trash.md
```diff
@@ -0,0 +1,28 @@
+# trash
+
+> A CLI for managing your trashcan / recycling bin.
+> More information: <https://github.com/andreafrancia/trash-cli>.
+
+- Delete a file (send to trash):
+
+`trash {{path/to/file}}`
+
+or
```
sbrl: I think this is messing up the linter. Perhaps you could remove this?

While the `trash-put` might do the same thing, I'd probably suggest that it has it's own page, as generally speaking each command will have it's own page.

Fealthas: fixed

sbrl:
Sorry for the delay.

Are we ready to merge then, @schneiderl?

schneiderl:
Sorry for the delay @Fealthas 
I have been a little busy the last few days.

Thank you for your contribution and feel welcome to keep contributing to tldr.

PR #3198, opened 2019-07-17
kubectl: edit page

### LINKED ISSUE #3195: kubectl page: Make it more concise
I have a few suggestions to kubectl pages, but I'm not sure if it makes sense for everyone or just for me, so here I go:

use:  
```
 # More information: <https://kubernetes.io/docs/reference/kubectl/cheatsheet/>.
``` 
instead of
```
# More information: <https://kubernetes.io/docs/tasks/tools/install-kubectl>.
``` 

one command: 
```
# List all information about a resource (pods, deployments, services, ingresses, etc.) with more information:
kubectl get pods -o wide
``` 
instead of 2 commands:
```
# List all pods in all namespaces:
kubectl get pods --all-namespaces

# List all pods with more information (such as node name):
kubectl get pods -o wide
``` 

one command: 
```
# Display Resource (CPU/Memory/Storage) usage of nodes or pods
kubectl top node
``` 
instead of two:
```
# Show metrics for all nodes:
kubectl top node

# Show metrics for all pods in the default namespace:
kubectl top pod
```  

and I'd like to see two more commands:
```
# Print the logs for a container in a pod or specified resource.
kubectl logs my-pod 

# Run command in existing pod
kubectl exec my-pod -- ls / 
```   

If it makes sense, I can proceed and submit a pr.

xild:
Fixes #3195

### DIFF
--- a/pages/common/kubectl.md
+++ b/pages/common/kubectl.md
@@ -26,3 +26,9 @@
 - Display an explanation of the specific field:
 
 `kubectl explain pods.spec.containers`
+
+- Print the logs for a container in a pod or specified resource: 
+
+`kubectl logs my-pod`
+
+
--- a/pages/common/kubectl.md
+++ b/pages/common/kubectl.md
@@ -3,9 +3,9 @@
 > Command line interface for running commands against Kubernetes clusters.
 > More information: <https://kubernetes.io/docs/reference/kubectl/>.
 
-- List all information about a resource (pods, deployments, services, ingresses, etc.) with more details:
+- List all information about a resource with more details:
 
-`kubectl get pods -o wide`
+`kubectl get {{pod|service|deployment|ingress|...}} -o wide`
 
 - Update specified pod with the label 'unhealthy' and the value 'true':
 
@@ -17,20 +17,20 @@
 
 - Display resource (CPU/Memory/Storage) usage of nodes or pods:
 
-`kubectl top pod`
+`kubectl top {{pod|node}}`
 
 - Print the address of the master and cluster services:
 
 `kubectl cluster-info`
 
-- Display an explanation of the specific field:
+- Display an explanation of a specific field:
 
-`kubectl explain pods.spec.containers`
+`kubectl explain {{pods.spec.containers}}`
 
 - Print the logs for a container in a pod or specified resource:
 
-`kubectl logs my-pod`
+`kubectl logs {{pod_name}}`
 
 - Run command in an existing pod:
 
-`kubectl exec my-pod -- ls /`
+`kubectl exec {{pod_name}} -- {{ls /}}`
--- a/pages/common/kubectl.md
+++ b/pages/common/kubectl.md
@@ -1,7 +1,7 @@
 # kubectl
 
 > Command line interface for running commands against Kubernetes clusters.
-> More information: <https://kubernetes.io/docs/tasks/tools/install-kubectl>.
+> More information: <https://kubernetes.io/docs/reference/kubectl/>.
 
 - List all pods in all namespaces:
 
--- a/pages/common/kubectl.md
+++ b/pages/common/kubectl.md
@@ -27,11 +27,10 @@
 
 `kubectl explain pods.spec.containers`
 
-- Print the logs for a container in a pod or specified resource: 
+- Print the logs for a container in a pod or specified resource:
 
 `kubectl logs my-pod`
 
-- Run command in an existing pod
+- Run command in an existing pod:
 
 `kubectl exec my-pod -- ls /`
-
--- a/pages/common/kubectl.md
+++ b/pages/common/kubectl.md
@@ -15,7 +15,7 @@
 
 `kubectl get all`
 
-- Display resource (CPU/Memory/Storage) usage of nodes or pods
+- Display resource (CPU/Memory/Storage) usage of nodes or pods:
 
 `kubectl top pod`
 
--- a/pages/common/kubectl.md
+++ b/pages/common/kubectl.md
@@ -15,11 +15,7 @@
 
 `kubectl get all`
 
-- Show metrics for all nodes:
-
-`kubectl top node`
-
-- Show metrics for all pods in the default namespace:
+- Display resource (CPU/Memory/Storage) usage of nodes or pods
 
 `kubectl top pod`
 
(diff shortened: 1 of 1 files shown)


agnivade:
@Aracki do you deal with k8s ?

pages/common/kubectl.md
```diff
@@ -19,11 +15,7 @@
...
 `kubectl get all`
 
-- Show metrics for all nodes:
-
-`kubectl top node`
-
-- Show metrics for all pods in the default namespace:
+- Display resource (CPU/Memory/Storage) usage of nodes or pods:
 
 `kubectl top pod`
```
Aracki: Instead of pod, I would put `{{pod|node}}`

pages/common/kubectl.md
```diff
@@ -1,13 +1,9 @@
 # kubectl
 
 > Command line interface for running commands against Kubernetes clusters.
-> More information: <https://kubernetes.io/docs/tasks/tools/install-kubectl>.
+> More information: <https://kubernetes.io/docs/reference/kubectl/>.
```
Aracki: good one :)

pages/common/kubectl.md
```diff
@@ -1,13 +1,9 @@
...
 
-- List all pods in all namespaces:
-
-`kubectl get pods --all-namespaces`
-
-- List all pods with more information (such as node name):
+- List all information about a resource (pods, deployments, services, ingresses, etc.) with more details:
 
 `kubectl get pods -o wide`
```
Aracki: In this case, if we want to put a generic object instead of a pod (which I think is a good suggestion), we would need to change pods with `{{pod|deployment|services|ingress|...}}`.

@agnivade can you just help us with what is the right syntax to use in this case: if we want to specify multiple objects in `{{}}` but there are a potentially unlimited number of them. Is it ok to put `|...` in the end?

pages/common/kubectl.md
Aracki: these are only two options for `top` command

pages/common/kubectl.md
```diff
@@ -34,3 +26,11 @@
 - Display an explanation of the specific field:
 
 `kubectl explain pods.spec.containers`
+
+- Print the logs for a container in a pod or specified resource:
+
+`kubectl logs my-pod`
```
Aracki: I would suggest changing `my-pod` to `pod_name` just to follow the general convention in tldr-pages.

pages/common/kubectl.md
```diff
@@ -34,3 +26,11 @@
...
+
+- Run command in an existing pod:
+
+`kubectl exec my-pod -- ls /`
```
Aracki: Same here ^^

Aracki:
@agnivade Thanks, you can always ping me for "devops" kinda pages 👍

pages/common/kubectl.md
mebeim: I would also suggest to wrap it: `{{pod_name}}`.

pages/common/kubectl.md
mebeim: Same here for wrapping (lol)

agnivade: Just to be clear, please wrap both {{my_pod}} and {{ls /}}

pages/common/kubectl.md
```diff
@@ -34,3 +26,11 @@
 - Display an explanation of the specific field:
```
agnivade: of _a_ specific field

pages/common/kubectl.md
```diff
@@ -34,3 +26,11 @@
...
 
 `kubectl explain pods.spec.containers`
```
agnivade: Wrap pods.spec.containers in tokens.

pages/common/kubectl.md
```diff
@@ -1,13 +1,9 @@
...
 > Command line interface for running commands against Kubernetes clusters.
-> More information: <https://kubernetes.io/docs/tasks/tools/install-kubectl>.
+> More information: <https://kubernetes.io/docs/reference/kubectl/>.
 
-- List all pods in all namespaces:
-
-`kubectl get pods --all-namespaces`
-
-- List all pods with more information (such as node name):
+- List all information about a resource (pods, deployments, services, ingresses, etc.) with more details:
```
agnivade: We can remove (pods, deployments ..) from the description as we are going to mention those in the example

pages/common/kubectl.md
agnivade: Yes that is correct. See the page for tshark. Although we are using the unicode ellipsis there instead of 3 dots. I think we later reverted our decision to use 3 dots again instead of ellipsis. Does anyone remember ? @sbrl @mebeim @waldyrious 

btw - the example has pods, so is saying `pod` equivalent to `pods` ?

sbrl: Yeah, I do believe we decided to use ASCII characters for maximum compatibility, though I can't remember the ultimate reason.

Aracki: @agnivade `pods` & `pod` are the equivalents. In this example I would use singular.

pages/common/kubectl.md
Aracki: Agree ^

pages/common/kubectl.md
Aracki: I agree on that.

xild:
Thanks for the comments, folks. I will work on it tonight.

pages/common/kubectl.md
waldyrious: > I think we later reverted our decision to use 3 dots again instead of ellipsis. Does anyone remember ? `@sbrl` `@mebeim` `@waldyrious`

> Yeah, I do believe we decided to use ASCII characters for maximum compatibility, though I can't remember the ultimate reason.

We actually never wrapped the discussion. The first change to use the ellipsis character was done in #2745, and the unfinished discussion about whether to use ASCII or Unicode is [here](https://github.com/tldr-pages/tldr/pull/2808#discussion_r261263925). Maybe let's try to finalize the discussion there?

agnivade: Agree. I suggest to revert back to 3 dots.

mebeim: I'm fine with switching to ASCII 3 dots.

xild:
Tonight took a bit too long 🤷‍♂  but it came. 

@Aracki

Aracki:
Great one! Thanks @xild 👍

PR #3203, opened 2019-07-22
git-grep: add new command

Aracki:
I have added `git grep` missing command which is very useful. [Source](https://stackoverflow.com/questions/7151311/using-git-how-could-i-search-for-a-string-across-all-branches)


mebeim:
Ok never mind about changing the message, I did it when squashing.

sbrl:
Nice addition!

PR #3208, opened 2019-07-26
7z: new example: extract archive to stdout

madacol:

### DIFF
--- a/pages/common/7z.md
+++ b/pages/common/7z.md
@@ -19,6 +19,10 @@
 
 `7z x {{archived.7z}} -o{{path/to/output}}`
 
+- Extract an archive to stdout:
+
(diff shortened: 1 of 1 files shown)


madacol:
> Thanks, @madacol smiley_cat
> 
> Hrm, wouldn't this concatenate all the files though thinking

Haven't tested it, but I agree, it should concatenate it all.

Yet It is very useful when it has one big file, because It allows to pipe it for post-processing, for example, you can peek what's inside without decompressing all of it.
`7z x huge_file.zip -so | head` it's practically instantaneous no matter how huge the file is

mebeim:
> `7z x huge_file.zip -so | head` it's practically instantaneous no matter how huge the file is

Didn't think about this at all! Cool.

PR #3211, opened 2019-07-29
sphinx-build: add page

NightMachinary:

### DIFF
--- a/pages/common/sphinx-build.md
+++ b/pages/common/sphinx-build.md
@@ -1,0 +1,16 @@
+# sphinx-build
+
+> Generates documentation from the files in sourcedir and places it in the outputdir.
+> More information: http://www.sphinx-doc.org/en/master/man/sphinx-build.html.
+
+- Build documentation as "epub" and put it in "builddir":
+
+`sphinx-build -b {{epub}} {{path/to/sourcedir}} {{path/to/builddir}}`
+
+- Build documentations intended for readthedocs.io:
+
+`pip install -U sphinx sphinx-rtd-theme ; sphinx-build -b {{html}} {{path/to/docs_dir}} {{path/to/builddir}}`
+
+- Common builders:
+
+`html dirhtml singlehtml htmlhelp qthelp devhelp epub applehelp latex man texinfo text gettext doctest linkcheck xml pseudoxml`
--- a/pages/common/sphinx-build.md
+++ b/pages/common/sphinx-build.md
@@ -1,9 +1,12 @@
 # sphinx-build
 
 > Sphinx documentation generator.
-> Can be used to build documentations intended for readthedocs.io (needs Python package sphinx-rtd-theme).
 > More information: <http://www.sphinx-doc.org/en/master/man/sphinx-build.html>.
 
 - Build documentation:
 
 `sphinx-build -b {{html|epub|text|latex|man|...}} {{path/to/source_dir}} {{path/to/build_dir}}`
+
+- Build documentations intended for readthedocs.io (requires the sphinx-rtd-theme pip package):
+
+`sphinx-build -b {{html}} {{path/to/docs_dir}} {{path/to/build_dir}}`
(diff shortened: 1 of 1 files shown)


pages/common/sphinx-build.md
```diff
@@ -0,0 +1,16 @@
+# sphinx-build
+
+> Generates documentation from the files in sourcedir and places it in the outputdir.
```
mebeim:
```suggestion
> Generates documentation from the files in sourcedir and places it in the outputdir.
> Sphinx Python documentation generator.
```
I would suggest rephrasing this to give a simpler and clearer description.

pages/common/sphinx-build.md
```diff
@@ -0,0 +1,16 @@
...
+> More information: http://www.sphinx-doc.org/en/master/man/sphinx-build.html.
```
mebeim:
```suggestion
> More information: <http://www.sphinx-doc.org/en/master/man/sphinx-build.html>.
```
Missing angle brackets around URL.

pages/common/sphinx-build.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Build documentation as "epub" and put it in "builddir":
```
mebeim:
```suggestion
- Build documentation for a Python project:
```
More concise and general, we can already see source dir, build dir and builder in the example (see my next comment)

pages/common/sphinx-build.md
```diff
@@ -0,0 +1,16 @@
...
+
+`sphinx-build -b {{epub}} {{path/to/sourcedir}} {{path/to/builddir}}`
```
mebeim:
```suggestion
`sphinx-build -b {{html|epub|text|...}} {{path/to/source_dir}} {{path/to/build_dir}}`
```
Let's use underscores for multi-word token. Also, since there are multiple builders, we can list some of them in the token, as we usually do in cases like this one.

pages/common/sphinx-build.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Build documentations intended for readthedocs.io:
+
+`pip install -U sphinx sphinx-rtd-theme ; sphinx-build -b {{html}} {{path/to/docs_dir}} {{path/to/builddir}}`
```
mebeim:
```suggestion
`pip install -U sphinx sphinx-rtd-theme ; sphinx-build -b {{html}} {{path/to/docs_dir}} {{path/to/builddir}}`
```
I don't think there's the need to also show how to install the `sphinx-rtd-theme` package. It should be added in the description of the example, like this:

    - Build documentations intended for readthedocs.io (requires the sphinx-rtd-theme pip package):

pages/common/sphinx-build.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Common builders:
+
+`html dirhtml singlehtml htmlhelp qthelp devhelp epub applehelp latex man texinfo text gettext doctest linkcheck xml pseudoxml`
```
mebeim: This isn't really a command or example. Let's just remove it altogether.

pages/common/sphinx-build.md
NightMachinary: I don’t think sphinx is specific to Python? It’s a general documentation generator, which uses Python to do so. So just `Sphinx documentation generator`?

pages/common/sphinx-build.md
NightMachinary: Again, let’s drop the `for a Python project`.

NightMachinary:
Is everything ok now? :-)

pages/common/sphinx-build.md
mebeim: You're right. It can be used for multiple languages, not only Python. Sorry about that.

mebeim:
Hi @NightMachinary, I don't think the second line of description is necessary. It was really better as an example. I would suggest removing it and keep the below example instead:

    - Build documentations intended for readthedocs.io (requires the sphinx-rtd-theme pip package):

    `sphinx-build -b {{html}} {{path/to/docs_dir}} {{path/to/build_dir}}`

NightMachinary:
@mebeim I’ve fixed it.

sbrl:
Done, @mebeim :smiley_cat:

PR #3216, opened 2019-08-01
Included file creation with no text.

russdb:
Want to recreate the touch command from linux? Here is how!

### DIFF
--- a/pages/windows/cmd.md
+++ b/pages/windows/cmd.md
@@ -32,4 +32,8 @@
 
 - Force output to use unicode encoding:
 
-`cmd /u`
+`cmd /u`  
+
+- Create file with no text inside, similar to "touch" command on linux distros:  
+
+`cmd:>filename.ext`


agnivade:
This exceeds our limit of 8 examples. Not sure why you ticked that checkbox. In any case, this seems to be an esoteric example usage of cmd. In tldr pages, we mainly focus on the basic and most widely used examples of a command.

mebeim:
Hi @russdb, thank you for the PR, but as @agnivade said pages should focus on the most used commands. The example you added is more of an "hack" rather than the intended use of the `cmd` command. The "correct" way to create an empty file on Windows would probably be `copy NUL file.txt` or `type NUL > file.txt`. The fact that it also exceeds our limit of 8 examples doesn't help either, replacing one of the existing examples with this one wouldn't make much sense. I therefore think we should close this PR.

russdb:
Sounds good

PR #3219, opened 2019-08-02
envoy: add page

pxgamer:

### DIFF
--- a/pages/common/envoy.md
+++ b/pages/common/envoy.md
@@ -1,0 +1,28 @@
+# envoy
+
+> A PHP-based task manager for Laravel remote servers.
+> More information: <https://laravel.com/docs/envoy>.
+
+- Initialise a configuration file:
+
+`envoy init {{host_name}}`
+
+- Run a task:
+
+`envoy run {{task_name}}`
+
+- Run a task from a specific project:
+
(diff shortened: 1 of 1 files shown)


pxgamer:
@agnivade, ah. I hadn't heard of that before. Naming projects is so crowded these days. _sigh_

Aracki:
I agree with @agnivade, but what do we do in such cases?

agnivade:
We likely need to come up with a modified pages architecture and an index format which can handle this. I don't know what exactly that should be. Needs some thought.

mebeim:
@sbrl well, there still isn't... but I would argue that creating a "disambiguation" page would be the most simple option. Anyway we'll see what to do when this happens.

sbrl:
Ah, that's a good idea - I hadn't even thought of that! Maybe something like this?

```markdown
# envoy

> Disambiguation page.

- Show information about envoy, the PHP command:

`tldr envoy_php`

- Show information about envoy, the proxy:

`tldr enoy_proxy`
```

mebeim:
@sbrl yep, exactly what I was thinking about :slightly_smiling_face:

agnivade:
PR looks good. Merging.

Aracki:
Thats a good practice for such cases @sbrl thanks 👍

PR #3224, opened 2019-08-05
genie: add page

cerebrate:

### DIFF
--- a/pages/linux/genie.md
+++ b/pages/linux/genie.md
@@ -1,0 +1,17 @@
+# genie
+
+> Set up and use a "bottle" namespace to run systemd under WSL (Windows Subsystem for Linux).
+> To run these from Windows rather than an already-running distribution, precede them with `wsl`.
+> Further information available at: <https://github.com/arkane-systems/genie>
+
+- Initialize the bottle (run once, at start):
+
+`genie -i`
+
+- Run a login shell inside the bottle.
+
+`genie -s`
+
+- Run a specified command inside the bottle:
+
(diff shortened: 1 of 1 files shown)


pages/linux/genie.md
```diff
@@ -0,0 +1,17 @@
+# genie
+
+> Set up and use a "bottle" namespace to run systemd under WSL (Windows Subsystem for Linux).
+> To run these from Windows rather than an already-running distribution, precede them with `wsl`.
+> Further information available at: <https://github.com/arkane-systems/genie>.
```
principis: See the [style guide](https://github.com/tldr-pages/tldr/blob/master/contributing-guides/style-guide.md)

```suggestion
> More information: <https://github.com/arkane-systems/genie>.
```

cerebrate:
@sbrl I thought about that. I placed it in the `linux` platform folder because it is a Linux app itself that has to be run from inside the WSL Linux environment, and so it seems to me that it should be accessible by default to the tldr client on Linux.

_That said_, a common use case for it is being run indirectly from Windows using commands like `wsl genie -i` or `wsl genie -s`, so I can see a case for putting a slightly modified version of this page in the `windows` platform folder indexed under that combination, as `wsl-genie.md`, in addition to this one. I don't think it would be a good replacement, though, due to the above, and I don't know what y'all would think of the duplication.

sbrl:
Yeah, it's a tough one. I agree, we should keep this version in _Linux_ because you execute it in a Linux environment.

PR #3226, opened 2019-08-06
phpbu: add page

pxgamer:


pages/common/phpbu.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Run backups using a specific configuration file:
+
+`phpbu --configuration={{path/to/configuration_file.xml}}`
+
+- Only run the specified backups:
+
+`phpbu --limit={{backup_task_name}}`
+
+- Don't actually create the backups:
```
schneiderl: Maybe rephrase this? 
This description does not say what the command does. If fact, it says exactly the opposite.

Aracki: but the command itself says `simulate`. I think this description can be left as it is :/

schneiderl: In my opinion you should not need to look at the example to understand the description. 
But if you guys feel that this is the best way i'll go with that.

andrik: I agree with @schneiderl

Aracki: Or "Simulate actions phpbu would have permormed"... I am ok with both examples...

pxgamer:
I'm not sure it's really for backing up the code (although you could), more for backing up anything. It supports a wide range of databases, etc.

sbrl:
Ah, I see! That makes much more sense :D

PR #3233, opened 2019-08-12
rofi: added dmenu option, added homepage

CoffeeVector:

### DIFF
--- a/pages/linux/rofi.md
+++ b/pages/linux/rofi.md
@@ -1,6 +1,7 @@
 # rofi
 
 > An application launcher and window switcher.
+> More Information: <https://github.com/davatorium/rofi>.
 
 - Show the list of apps:
 
@@ -13,3 +14,7 @@
 - Switch between windows:
 
 `rofi -show window`
+
+- Pipe a list of items to stdin and print the selected item to stdout:
+
+`printf "{{Choice1\nChoice2\nChoice3}}" | rofi -dmenu`


schneiderl:
It might just be me but this dmenu option description doesn't really say anything to me. 
Maybe we can rephrase it to make more clear what the command actually does?

CoffeeVector:
- If we pipe a list of items to stdin, the selected item will be printed (or piped into a different process) to stdout:

Would this be more clear?

schneiderl:
reads better imho @CoffeeVector

pages/linux/rofi.md
```diff
@@ -13,3 +14,7 @@
 - Switch between windows:
 
 `rofi -show window`
+
+- If we pipe a list of items to stdin, the selected item will be printed (or piped into a different process) to stdout:
```
sbrl: Sorry to be a pain, but I'm not so sure about this. I think it's the `If` here that's bothering me. What about this?

```suggestion
- Pipe a list of items to stdin and print the selected item stdout:
```

I think we can assume that the reader will know that they can send the standard output of one command into another as the standard input.

Gorock:
I looked a bit into the manpages of Dmenu and rofi to see how this option could be worded.

Personally i would word it like dmenu, which is as follows:

>  dmenu is a dynamic menu for X, which reads a list of newline-separated items from stdin.  When the  user  selects an item and presses Return, their choice is printed to stdout and dmenu terminates.  Entering text will
>        narrow the items to those matching the tokens in the input.


tldr:
rofi reads a list of newline-seperated items from stdin. When the user selects an item, their choice is printed to stdout and rofi terminates.

pages/linux/rofi.md
agnivade: It is not clear to me how the item is selected. The stdin input is just a list of lines.

CoffeeVector: Rofi will open a ui thing that'll let the user select one

agnivade: Ok, then we should clarify that. Perhaps say "select an item and then print it to console" or something to that effect.

pages/linux/rofi.md
```diff
@@ -13,3 +14,7 @@
 - Switch between windows:
 
 `rofi -show window`
+
+- Pipe a list of items to stdin and print the selected item stdout:
```
mebeim:
```suggestion
- Pipe a list of items to stdin and print the selected item to stdout:
```

Typo.

pages/linux/rofi.md
```diff
@@ -13,3 +14,7 @@
...
+
+`echo "{{Something\nOr\nOther}}" | rofi -dmenu`
```
mebeim:
```suggestion
`echo -e "{{Coiche1\nChoice2\nChoice3}}" | rofi -dmenu`
```
`echo` needs `-e` to interpret newlines `\n`. Also, I think using `ChoiceN` for the items is clearer.

PR #3235, opened 2019-08-13
chflags: add missing 'no' to clear hidden flag

vszakats:


schneiderl:
Whops, you have already signed it. 
Thanks a lot once again.

PR #3237, opened 2019-08-14
bash: add `--` option.

KsRyY:

### DIFF
--- a/pages/common/bash.md
+++ b/pages/common/bash.md
@@ -28,6 +28,10 @@
 
 `bash -s`
 
+- Pass arguments behind to the script (can be combined with `-s` to pass arguments to commands/scripts from `stdin`):
+
+`bash {{file.sh}} --`
+
 - Print the version information of bash (use `echo $BASH_VERSION` to show just the version string):
 
 `bash --version`
--- a/pages/common/bash.md
+++ b/pages/common/bash.md
@@ -30,7 +30,7 @@
 
 - Stop bash argument processing and pass all subsequent arguments directly to the script being executed:
 
-`bash -c {{ls}} -- -lA`
+`bash {{path/to/script}} -- {{arguments}}`
 
 - Print the version information of bash (use `echo $BASH_VERSION` to show just the version string):
 
--- a/pages/common/bash.md
+++ b/pages/common/bash.md
@@ -28,7 +28,7 @@
 
 `bash -s`
 
-- Stop argument processing bash and pass all the following arguments directly to the command:
+- Stop bash argument processing and pass all subsequent arguments directly to the script being executed:
 
 `bash -c {{ls}} -- -lA`
 
--- a/pages/common/bash.md
+++ b/pages/common/bash.md
@@ -30,7 +30,7 @@
 
 - Stop argument processing bash and pass all the following arguments directly to the command:
 
-`bash {{file.sh}} --`
+`bash -c {{ls}} -- -lA`
 
 - Print the version information of bash (use `echo $BASH_VERSION` to show just the version string):
 
--- a/pages/common/bash.md
+++ b/pages/common/bash.md
@@ -28,7 +28,7 @@
 
 `bash -s`
 
-- Stop further argument parsing for bash and pass all the following arguments directly to the command:
+- Stop argument processing bash and pass all the following arguments directly to the command:
 
 `bash {{file.sh}} --`
 
--- a/pages/common/bash.md
+++ b/pages/common/bash.md
@@ -28,7 +28,7 @@
 
 `bash -s`
 
-- Pass arguments behind to the script (can be combined with `-s` to pass arguments to commands/scripts from `stdin`):
+- Stop further argument parsing for bash and pass all the following arguments directly to the command:
 
 `bash {{file.sh}} --`
 


pages/common/bash.md
```diff
@@ -28,6 +28,10 @@
 
 `bash -s`
 
+- Pass arguments behind to the script (can be combined with `-s` to pass arguments to commands/scripts from `stdin`):
+
+`bash {{file.sh}} --`
```
mfrw: In my opinion, `--` disables further options parsing and whatever is given after `-` or `--` is considered as filenames or arguments to the script itself and not to bash.
Could we have a concise example about that.

pages/common/bash.md
```diff
@@ -28,6 +28,10 @@
 
 `bash -s`
 
+- Pass arguments behind to the script (can be combined with `-s` to pass arguments to commands/scripts from `stdin`):
```
mfrw: This seems a little vague IMHO, How about we start like:
`Stop processing options for bash....`

KsRyY: Ya, this is better. I am going to change this.

KsRyY: I've changed the description, is the new one better?

pages/common/bash.md
```diff
@@ -28,6 +28,10 @@
 
 `bash -s`
 
+- Stop argument processing bash and pass all the following arguments directly to the command:
```
sbrl: Perhaps this will read better?
```suggestion
- Stop bash argument processing and pass all subsequent arguments directly to the script being executed:
```

pages/common/bash.md
```diff
@@ -28,6 +28,10 @@
...
+
+`bash -c {{ls}} -- -lA`
```
sbrl: We've already got a `bash -c` example above. Perhaps this could be just a regular `"{{path/to/script.sh}}"` instead?

pages/common/bash.md
KsRyY: Yes. I would commit your suggested change.

pages/common/bash.md
sbrl: I think you might have missed this comment @KsRyY?

mebeim:
Isn't this just a different version of the second example?

    bash -c "{{command}}"

If so, I would suggest modifying that example instead:

    bash -c "{{command --with arguments}}"

agnivade:
Agree with @mebeim. We don't need an extra example for this.

pages/common/bash.md
KsRyY: yes. making the changes.

KsRyY:
Given a better example without opverlap with `-c` options

pages/common/bash.md
```diff
@@ -28,6 +28,10 @@
 
 `bash -s`
 
+- Stop bash argument processing and pass all subsequent arguments directly to the script being executed:
+
+`bash {{path/to/script}} -- {{arguments}}`
```
agnivade: path/to/file.sh to be consistent

mebeim: This is still the same as `bash -c "..."`, and it also seems to be an unneeded example, at least based on my experiments with it:

Here, `-c` and `-i` are `bash` options, bu they still get ignored by `bash` and get passed to `x.sh`:

image

agnivade: Ohh I see, I hadn't read the man page. But now it is clear. Yes you are right, I think we need to adjust our description of `-c` to properly reflect this. 

> -c        If the -c option is present, then commands are read from the first non-option argument command_string.  If there are argu‐
                 ments after the command_string, the first argument is assigned to $0 and any remaining arguments are assigned to the posi‐
                 tional parameters.  The assignment to $0 sets the name of the shell, which is used in warning and error messages.

mebeim: image

Don't think the description for the `-c` command needs any change. If anything, like I said in my first review above, I would add `--with args`, making the example:

    bash -c "{{command --with args}}"

sbrl: After thinking about it, I'm guessing that the idea with the `--` one is to avoid nasty issues with quotes. Hrm :thinking:

agnivade: When I read the description, I thought it is only used for executing commands like echo and date, but did not realize that this option was actually about consuming arguments properly.

Updating the example is also necessary. But the description should clarify when should we use "bash -c "{{something}}"" vs "bash {{something}}". The current description just says "execute a command" vs "execute a file", which is why I thought `-c` is only for executing oneliner commands.

mebeim: @agnivade  it **is** only for commands. In my last example, `bash -c "./x.sh ..."` only works because `x.sh` is marked executable and has the proper shebang (note the `./` prefix, since the current directory is not in my `$PATH`). Something like `bash -c "x.sh ..."` would not work (even if the file is marked as executable), because the current folder is not in the `$PATH`.

In other words, `-c` is **not** to execute files *and* commands, it's only to execute bash commands, for example:

    $ bash -c "if [ 0 -eq 0 ]; then echo 1; else echo 2; fi"
    1

agnivade: I see. Thanks. Then is there a need to add `--` to the `-c` example ? Because if the command is enclosed within quotes then it is already safe from intereference with the other bash arguments.

mebeim: @sbrl granted that, but I don't really see the point since these two invocations are the same:

    $ bash ./script.sh arg1 arg2 arg3
    $ bash ./script.sh -- arg1 arg2 arg3

This PR is basically just a "special case" for the second example really.

mebeim: @agnivade IMHO no, I see no need to add `--` for the `-c` example.

KsRyY: @mebeim @agnivade @sbrl But this option is necessary when you want to pass arguments to a script that is piped into bash, for example:
``` bash
$ curl -fsSL https://git.io/n-install | bash -s -- {{arugments}}
```
If you don't add the `--` option then curl will recieve the arguments, not the script that is piped into bash.

KsRyY: The problem is I am not sure if I could use this curl command as the example.

mebeim: As I already said, that does not seem to be needed for me. And besides, this is a *very different* example than the one you are proposing to add in this PR.

image

What version of bash are you using?

sbrl: Hrm. It does seem to be a special use-case, so perhaps it's not a good fit for tldr-pages, since we prefer to focus on the most common use-cases. I'm not sure. @waldyrious?

waldyrious: Waaay behind schedule :sweat_smile:, but reading the entire discussion (and special thanks to @mebeim for the thorough investigation), I agree that we would be entering tricky waters that are not a good fit for tldr's beginner-focused approach. Especially since the page is already sizeable with the other examples.

waldyrious:
Hey @KsRyY -- although it's been a while since this PR was created, and it was ultimately closed without being merged, I thought I'd wrap the thread up with a word of appreciation for your contribution and patience. Cheers! :)

PR #3242, opened 2019-08-16
tldr: add cache control and search example

KsRyY:

### DIFF
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -15,11 +15,11 @@
 
 `tldr {{git checkout}}`
 
-- Update the local cache and rebuild index
+- Update the local cache and rebuild index:
 
 `tldr -u`
 
-- Clear the local cache
+- Clear the local cache:
 
 `tldr -c`
 
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -15,14 +15,6 @@
 
 `tldr {{git checkout}}`
 
-- Update the local cache and rebuild index:
-
-`tldr -u`
-
-- Clear the local cache:
-
-`tldr -c`
-
 - Search keywords in the index and show match result:
 
 `tldr -s {{keywords}}`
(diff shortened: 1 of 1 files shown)


pxgamer:
I was thinking about this, and wasn't sure whether this page should just follow the [Client Spec](https://github.com/tldr-pages/tldr/blob/dec9fb400bb7bac4778a752567f58be4815eb668/CLIENT-SPECIFICATION.md#command-line-interface). 🤔 Or if it should just match the [`tldr-node`](https://github.com/tldr-pages/tldr-node-client) client. I assume additional options were previously excluded as they may not exist in all clients?

andrik:
Oh, I didn't think about that @pxgamer. You are right, but I would maintain `-u` example, even though it is optional in the spec

KsRyY:
So I might add an notice that not every client can do that.

pages/common/tldr.md
```diff
@@ -14,3 +14,15 @@
 - Get help for a git subcommand:
 
 `tldr {{git checkout}}`
+
+- Update the local cache and rebuild index:
+
+`tldr -u`
+
+- Clear the local cache:
```
sbrl: I'm not sure about adding these last 2 examples, as they are not in the [official client specification](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md).

KsRyY: So I would remove it. Applying.

KsRyY: Changes committed to `patch-5`.

KsRyY:
@sbrl @pxgamer removed these cache control examples. Thank you for pointing our my error.

mebeim:
As @agnivade said, this page should contain only options that are listed as mandatory in the client specification, so that it can be compatible with all conforming clients. Appreciate the PR @KsRyY, but I think it should be closed unfortunately.

PR #3246, opened 2019-08-18
git-switch: add page

sbrl:
Part 1 of several from GitHub's blog post about git v2.23!

Original source: https://github.blog/2019-08-16-highlights-from-git-2-23/?utm_campaign=1565911976&utm_medium=social&utm_source=twitter&utm_content=1565911976

- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #3247, opened 2019-08-18
git-restore: add page

sbrl:
Part 2! Original PR is #3246.

### DIFF
--- a/pages/common/git-restore.md
+++ b/pages/common/git-restore.md
@@ -2,7 +2,7 @@
 
 > Restore working tree files. Requires git version 2.23+.
 > See also `git checkout`.
-> More information: <https://git-scm.com/docs/git-switch/>.
+> More information: <https://git-scm.com/docs/git-restore/>.
 
 - Restore a file that was accidentally deleted:
 
--- a/pages/common/git-restore.md
+++ b/pages/common/git-restore.md
@@ -1,0 +1,17 @@
+# git restore
+
+> Restore working tree files. Requires git version 2.23+.
+> See also `git checkout`.
+> More information: <https://git-scm.com/docs/git-switch/>.
+
+- Restore a file that was accidentally deleted:
+
+`git restore {{path/to/file}}`
+
+- Restore a file to an earlier version:
+
+`git restore --source {{commit}} {{path/to/file}}`
+
+- Restore all the files in the current working tree to the last committed version:
+
+`git restore .`
--- a/pages/common/git-restore.md
+++ b/pages/common/git-restore.md
@@ -4,7 +4,7 @@
 > See also `git checkout`.
 > More information: <https://git-scm.com/docs/git-restore/>.
 
-- Restore a file from the current HEAD that was previously deleted:
+- Restore a deleted file from the contents of the current commit (HEAD):
 
 `git restore {{path/to/file}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/git-restore.md
```diff
@@ -0,0 +1,17 @@
+# git restore
+
+> Restore working tree files. Requires git version 2.23+.
+> See also `git checkout`.
+> More information: <https://git-scm.com/docs/git-restore/>.
+
+- Restore a file that was accidentally deleted:
```
waldyrious: I don't think the "accidentally" is needed here :)

Also, maybe specify that this fetches the file contents from the current commit / HEAD?

pages/common/git-restore.md
```diff
@@ -0,0 +1,17 @@
...
+
+`git restore {{path/to/file}}`
+
+- Restore a file to an earlier version:
```
waldyrious: The other version is not necessarily earlier. It could simply be from a different branch. Maybe say that explicitly.

pages/common/git-restore.md
```diff
@@ -0,0 +1,17 @@
...
+
+`git restore --source {{commit}} {{path/to/file}}`
+
+- Restore all the files in the current working tree to the last committed version:
```
waldyrious: Alternatively:

> Undo any uncommitted changes to tracked files.

I'm not sure which version I prefer. Yours is certainly more explicit. I'll wait to see what others have to say :)

mebeim: Does this use `HEAD` as "last committed version"? Too bad I don't have the latest Git version so I cannot test it out. If this uses `HEAD`, saying "the last committed version" could be unclear and I would suggest using @waldyrious' version.

agnivade: I prefer the shorter version. It succinctly explains what the command does without losing any information.

waldyrious: > Does this use HEAD as "last committed version"?

Yes, just tested — it defaults to HEAD if no revision is specified (what else would it do, anyway?).

> If this uses HEAD, saying "the last committed version" could be unclear

Well, true, but for most cases this will hold. But yeah, if you've checked out, say, a tag, that typically won't be the latest commit in the branch. Granted, one usually won't make changes in a non-latest commit of a branch, but it's certainly possible, so I see your point.

mebeim: Yeah I'm totally nitpicking here, but whatever, that was a long way to say I prefer the shortest version 😆

sbrl: I've taken the shorter one, but added a bit to try and make it clearer what it is restoring to. Does this work?

pages/common/git-restore.md
```diff
@@ -0,0 +1,17 @@
+# git restore
+
+> Restore working tree files. Requires git version 2.23+.
+> See also `git checkout`.
+> More information: <https://git-scm.com/docs/git-restore/>.
+
+- Restore a file from the current HEAD that was previously deleted:
```
waldyrious: How about something like this?

```suggestion
- Restore a deleted file from the contents of the current commit (HEAD):
```

pages/common/git-restore.md
waldyrious: @sblr why ask a question then mark the conversation as resolved? :P

But yes, I agree with your changes. :+1:

pages/common/git-restore.md
sbrl: Sure!

PR #3248, opened 2019-08-19
Update git-switch.md

sbrl:
Changes left-over from #3246

### DIFF
--- a/pages/common/git-switch.md
+++ b/pages/common/git-switch.md
@@ -16,7 +16,7 @@
 
 `git switch --create {{branch_name}} {{commit}}`
 
-- Switch to a branch and update all submodules to match the target branch:
+- Switch to a branch and update all submodules to match:
 
 `git switch --recurse-submodules {{branch_name}}`
 
--- a/pages/common/git-switch.md
+++ b/pages/common/git-switch.md
@@ -12,14 +12,14 @@
 
 `git switch --create {{branch_name}}`
 
-- Create a new branch based on an existing commit:
+- Create a new branch based on an existing commit and switch to it:
 
 `git switch --create {{branch_name}} {{commit}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/git-switch.md
```diff
@@ -12,14 +12,14 @@
 
 `git switch --create {{branch_name}}`
 
-- Create a new branch based on an existing commit:
+- Create a new branch based on an existing commit and switch to it:
```
waldyrious: Maybe add a comma after "commit", to make the sentence flow a bit better?

pages/common/git-switch.md
```diff
@@ -12,14 +12,14 @@
...
 
 `git switch --create {{branch_name}} {{commit}}`
 
-- Update all submodules to match the target branch:
+- Switch to a branch and update all submodules to match the target branch:
```
waldyrious: Maybe "...update all submodules to match it"? Or "...update all submodules to the versions it specifies"?

pages/common/git-switch.md
sbrl: Hrm, that sound  odd to me. I'll add it though if you think it's a good idea.

pages/common/git-switch.md
sbrl: changed.

pages/common/git-switch.md
waldyrious: Sure, let's see what others think.

agnivade: Not a grammar person, but I think it is fine without the comma.

waldyrious: It is fine :) I meant it as a minor suggestion, not as a correction. Since both of you prefer the no-comma version, let's go with that then. 👍

PR #3254, opened 2019-08-23
Transition from BCP47 to POSIX

### LINKED ISSUE #3183: Language tags: POSIX or BCP-47?
This issue just popped up from [this thread on the `tldr-node-client` repo](https://github.com/tldr-pages/tldr-node-client/pull/270#discussion_r300885553), from which I'm quoting the relevant comments:

> @mebeim: Our client specification does not follow [the POSIX locale specification](https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html#Locale-Names) for language environment variables. As a result of this, the `LANG` environment var is basically never going to be right, and should be converted first. 
> [...]
> This is an issue I've been thinking about for quite a while. I really think we should update the client spec to follow the POSIX locale specification (basically just use `ll_CC` instead of `ll-CC`). It would make it really easier for clients to deal with languages.

> @pxgamer: I was thinking this earlier too. I know there was a reason that we chose the current language code format though, can't remember why.

> @mebeim: the only issue here is the dash - versus the underscore _. I don't remember a real reason for this choice, if I recall right 🤔.

> @agnivade: It's here - https://gitter.im/tldr-pages/tldr?at=5c387be59289cc7aa757c1ad. You commented yourself @mebeim :) 
>
> tldr is GNU uses _ and BCP-47 uses -. They have different standards. And we decided to go with BCP-47.

> @mebeim: Thanks for the link @agnivade makes a lot more sense now, we were looking at BCP-47 and not POSIX. That's a bummer since now that I think about it the POSIX version is probably the only one you'll find used around when it comes to programming and environment variables.

---

Currently TLDR pages uses a simplified version of the [IETF BCP 47 language tags](https://tools.ietf.org/rfc/bcp/bcp47.txt) (we only use the language and country fields). This is reflected in the client specification. In reality though, BCP 47 tags are quite uncommon, and on any POSIX compliant system [POSIX language tags](https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html#Locale-Names) are used instead. 

To put it simply, what all of this boils down to is the following:

 - We are currently using BCP 47 `ll-CC` tags (e.g. `pages.pt-BR`).
 - Most (if not all) operating systems use POSIX `ll_CC` tags. 

While the only thing that changes is a dash (`-`) versus an underscore (`_`), this seems to makes things unnecessarily more complicated for clients and in general any programmatic operation that has to deal with language tags.

I am therefore opening this issue to suggest switching to POSIX tags now that we can since that virtually no client is supporting languages already, and this might be the last chance to make such a change without much trouble.

If we choose to go with this, we would have to:

1. Rename pages folders (currently only `pages.pt-BR`).
2. Update the "Translations" section of the contributing guide.
3. Update the "Language" section of the client specification.
4. Deprecate any previous client specification (currently only two versions) version since specification-wise this is a breaking change. For this point I'd suggest to simply add `(deprecated)` alongside previous client spec versions in the "Changelog" section of the document.

I'd suggest the changes to the client spec into #3168 to limit the number of different versions of the spec.

What do you guys think?

mebeim:
Closes #3183 :tada: 

Changes are pretty straightforward:

 - Rename `pages.pt-BR` to `pages.pt_BR`.
 - Update [`CONTRIBUTING.md`](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) changing BCP47 tags to POSIX (mostly copy-pasted from [`CLIENT-SPECIFICATION.md`](https://github.com/tldr-pages/tldr/blob/d9e1d592e5f7074a5459875ef06a4d4841659ced/CLIENT-SPECIFICATION.md#L113)).

Please rebase rather than squashing when merging.

PS: this should also "unlock" [tldr-node-client#270](https://github.com/tldr-pages/tldr-node-client/pull/270).


CONTRIBUTING.md
```diff
@@ -75,21 +75,19 @@ refer to the [style guide](contributing-guides/style-guide.md).
 
 Translation of pages can be done by simply creating the corresponding page within the appropriate language-specific directory, creating that as well if it does not already exist.
 
-Language specific directories must follow the pattern `pages.<language_tag>`, where `<language_tag>` is a [BCP 47](https://tools.ietf.org/html/bcp47) conforming tag in the form of `<language>[-<region>]`, where:
+Language specific directories must follow the pattern `pages.<locale>`, where `<locale>` is a [POSIX Locale Name](https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html#Locale-Names) in the form of `<language>_<country>`, where:
```
agnivade: Isn't the `_<country>` optional according to your text below ?

mebeim: @agnivade hmm, yes it is, are you suggesting the addition of square brackets around `_<country>`?

agnivade: Right.

mebeim: Done :+1:

mebeim:
(fixed a typo `examples of valid BCP-47 tags` -> `examples of valid POSIX tags`)

PR #3258, opened 2019-09-03
jest: copyedit, add one command

garyking:
I've copyedited the Jest page, and added one command, for testing specific files.

### DIFF
--- a/pages/common/jest.md
+++ b/pages/common/jest.md
@@ -13,11 +13,11 @@
 
 - Run tests whose names match the regex pattern:
 
-`jest -t {{spec_name}}`
+`jest --testNamePattern {{spec_name}}`
 
 - Run tests related to uncommitted files:
 
-`jest -o`
+`jest --onlyChanged`
 
 - Watch files for changes and re-run related tests:
 
--- a/pages/common/jest.md
+++ b/pages/common/jest.md
@@ -7,6 +7,10 @@
 
 `jest`
 
+- Run tests from the test files that match the regex patterns:
+
+`jest {{test_file1}} {{test_file2}}`
+
 - Run tests whose names match the regex pattern:
 
 `jest -t {{spec_name}}`
--- a/pages/common/jest.md
+++ b/pages/common/jest.md
@@ -13,7 +13,7 @@
 
 - Run tests related to uncommitted files:
 
-`jest --onlyChanged`
+`jest -o`
 
 - Watch files for changes and re-run related tests:
 
--- a/pages/common/jest.md
+++ b/pages/common/jest.md
@@ -7,15 +7,15 @@
 
 `jest`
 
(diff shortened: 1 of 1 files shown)


pages/common/jest.md
```diff
@@ -7,15 +7,19 @@
...
+
+- Run tests whose names match the regex pattern:
 
 `jest -t {{spec_name}}`
 
-- Run only tests relating to uncommitted files:
+- Run tests related to uncommitted files:
 
-`jest --onlyChanged`
+`jest -o`
```
pxgamer: I think it seems better to keep this as the long option as there isn't anything in the new description that matches the `o`. Or update the description to mention it `[O]nly run tests related to uncommitted files:` as before. 😕

Aracki: Yep maybe it's better to keep longer flags.

pages/common/jest.md
```diff
@@ -7,15 +7,19 @@
 
 `jest`
 
-- Run only tests with a name that matches the regex pattern:
+- Run tests from the test files that match the regex patterns:
+
+`jest {{test_file1}} {{test_file2}}`
```
agnivade: Since we are mentioning the word regex, it might be useful to use a regex like `file*.js`.

pages/common/jest.md
garyking: I don't mind either way, but I think it should be consistent. The long name for `-t` is `--testNamePattern`, which I think is too long. Can I have the short name first, then the long name below it?

So:

```
- Run tests whose names match the regex pattern:

`jest -t {{spec_name}}`
`jest --testNamePattern {{spec_name}}`
```

pages/common/jest.md
garyking: Perhaps `jest {{*.test.js}}` as the example then?

sbrl: If it's a regular expression,  don't you mean `file.*.js`, @agnivade?

pages/common/jest.md
pxgamer: I think it's fine to just use the long name as users can always run `--help` to find the short name? We don't currently support multiple versions of a flag in the spec afaik.

garyking: Sure, just the long name is fine, then.

pages/common/jest.md
agnivade: Yes, I didn't focus on the details :P Just wanted to get my point across. @garyking - what you say is fine.

sbrl: `jest {{*.test.js}}` is still a glob. It would need to be `jest {{.*.test.js}}` to be a regular expression.

garyking: I'll just leave it as `test_file1` and `test_file2`, just like the `spec_name`.

garyking:
Alright, I've changed to using the long names for options.

PR #3265, opened 2019-09-15
nextflow: add page

ewels:
- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #3270, opened 2019-09-17
nextflow: Move to common

ewels:
Nextflow isn't linux-only, so it should probably be under `common`. See https://github.com/tldr-pages/tldr/pull/3265#issuecomment-531966166 for discussion. From the docs:

> **Requirements**
> Nextflow can be used on any POSIX compatible system (Linux, OS X, etc). It requires Bash 3.2 (or later) and Java 8 (or later, up to 11) to be installed.
>
> Windows systems may be supported using a POSIX compatibility layer like Cygwin (unverified) or, alternatively, installing it into a Linux VM using virtualization software like VirtualBox or VMware.


sbrl:
Merging now, since this is a previously discussed change that's fairly trivial.

PR #3282, opened 2019-10-01
ul: add page

johncurcio:

### DIFF
--- a/pages/linux/ul.md
+++ b/pages/linux/ul.md
@@ -1,7 +1,7 @@
 # ul
 
 > Performs the underlining of a text.
-> To underline a text is necessary to underline each character separately.
+> Each character in a given string must be underlined separately.
 
 - Display the contents of the file with underlines where applicable:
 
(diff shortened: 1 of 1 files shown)


andrik:
Hi @johncurcio 

Thank you for your submission. Looks like linter founded some issues, would you please fix and push again?

pages/linux/ul.md
```diff
@@ -0,0 +1,12 @@
+# ul
+
+> Performs the underlining of a text.
+> To underline a text is necessary to underline each character separately.
```
sbrl: What precisely do you mean by this?

johncurcio: If you want to underline the character H you can write something like ``_\bH`` in a file or using ``echo`` and call ``ul``. However, if you want to underline Hello, you can surround the whole word with underline ``_\bHello_\b``, you need to do this for each character ``_\bH_\be_\bl_\bl_\bo``.

sbrl: Ah, I see. Perhaps we could reword this to make it clearer. What about this?
```suggestion
> Each character in a given string must be underlined separately.
```

PR #3283, opened 2019-10-01
runuser: add page

hieuhtr:

### DIFF
--- a/pages/linux/runuser.md
+++ b/pages/linux/runuser.md
@@ -14,7 +14,7 @@
 
 `runuser {{user}} -l`
 
-- Specify a [s]hell for running instead of default shell (also works for login):
+- Specify a shell for running instead of the default shell (also works for login):
 
 `runuser {{user}} -s {{/bin/sh}}`
 
--- a/pages/linux/runuser.md
+++ b/pages/linux/runuser.md
@@ -14,7 +14,7 @@
 
 `runuser {{user}} -l`
 
-- Specific a shell for running or login instead of default shell:
+- Specify a [s]hell for running instead of default shell (also works for login):
 
 `runuser {{user}} -l -s /bin/sh`
 
--- a/pages/linux/runuser.md
+++ b/pages/linux/runuser.md
@@ -1,6 +1,6 @@
 # runuser
 
-> Run commands as a specific user and group IDs. Does not ask for password. Only useful when run as the root user.
+> Run commands as a specific user and group without asking for password (needs root privileges).
 
 - Run command as a different user:
 
--- a/pages/linux/runuser.md
+++ b/pages/linux/runuser.md
@@ -16,7 +16,7 @@
 
 - Specify a [s]hell for running instead of default shell (also works for login):
 
-`runuser {{user}} -l -s /bin/sh`
+`runuser {{user}} -s {{/bin/sh}}`
 
 - Preserve the entire environment of root (only if `--login` is not specified):
 
--- a/pages/linux/runuser.md
+++ b/pages/linux/runuser.md
@@ -18,6 +18,6 @@
 
 `runuser {{user}} -l -s /bin/sh`
 
-- Preserve the entire environment of root (if running by root). Ignore when having option `--login`:
+- Preserve the entire environment of root (only if `--login` is not specified):
 
 `runuser {{user}} --preserve-environment -c '{{command}}'`
--- a/pages/linux/runuser.md
+++ b/pages/linux/runuser.md
@@ -1,7 +1,7 @@
 # runuser
 
 > Run commands as a specific user and group IDs.
-> Do not ask for password. Only useful when run as root user.
+> Does not ask for password. Only useful when run as the root user.
 
 - Run 'command' under substitute 'user':
 
--- a/pages/linux/runuser.md
+++ b/pages/linux/runuser.md
@@ -1,13 +1,12 @@
 # runuser
 
-> Run commands as a specific user and group IDs.
-> Does not ask for password. Only useful when run as the root user.
+> Run commands as a specific user and group IDs. Does not ask for password. Only useful when run as the root user.
 
-- Run 'command' under substitute 'user':
+- Run command as a different user:
 
 `runuser {{user}} -c '{{command}}'`
 
-- Run 'command' under substitute 'user' and 'group':
+- Run command as a different user and group:
 
 `runuser {{user}} -g {{group}} -c '{{command}}'`
 
@@ -15,7 +14,7 @@
 
 `runuser {{user}} -l`
 
-- Specific a 'shell' for running or login instead of default shell:
+- Specific a shell for running or login instead of default shell:
 
 `runuser {{user}} -l -s /bin/sh`
 
--- a/pages/linux/runuser.md
+++ b/pages/linux/runuser.md
@@ -1,0 +1,25 @@
+# runuser
+
+> Run commands as a specific user and group IDs
+> Do not ask for password. Only useful when run as root user.
+
+- Run 'command' under substitute 'user':
+
+`runuser {{user}} -c '{{command}}'`
+
+- Run 'command' under substitute 'user' and 'group':
+
+`runuser {{user}} -g {{group}} -c '{{command}}'`
+
+- Start a login shell as a specific user:
+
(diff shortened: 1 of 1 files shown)


hieuhtr:
@mebeim @agnivade 
Sorry, but could you please help me review my PR? Thanks in advance

mebeim:
@hieuhtr we will do for sure, don't worry :) we're just a bit over-encumbered right now with the amount of submitted PRs in such a short period of time.

pages/linux/runuser.md
```diff
@@ -0,0 +1,24 @@
+# runuser
+
+> Run commands as a specific user and group IDs.
```
sbrl: Perhaps this will read better?
```suggestion
> Run commands as specific users and group IDs.
```

pages/linux/runuser.md
```diff
@@ -0,0 +1,24 @@
...
+> Do not ask for password. Only useful when run as root user.
```
sbrl: Perhaps this will read better?
```suggestion
> Does not ask for password. Only useful when run as the root user.
```

pages/linux/runuser.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Run 'command' under substitute 'user':
```
sbrl: Perhaps this will read better?
```suggestion
- Run a command under a substitute user:
```

pages/linux/runuser.md
```diff
@@ -0,0 +1,24 @@
...
+
+`runuser {{user}} -c '{{command}}'`
+
+- Run 'command' under substitute 'user' and 'group':
```
sbrl: Perhaps this will read better?
```suggestion
- Run a command under a substitute user and group:
```

pages/linux/runuser.md
```diff
@@ -0,0 +1,24 @@
...
+
+`runuser {{user}} -g {{group}} -c '{{command}}'`
+
+- Start a login shell as a specific user:
+
+`runuser {{user}} -l`
+
+- Specific a 'shell' for running or login instead of default shell:
```
sbrl: Perhaps this will read better?
```suggestion
- Start a specific shell instead of the default shell:
```

pages/linux/runuser.md
```diff
@@ -0,0 +1,24 @@
...
+
+`runuser {{user}} -l -s /bin/sh`
+
+- Preserve the entire environment of root (if running by root). Ignore when having option `--login`:
```
sbrl: Perhaps this will read better?
```suggestion
- Inherit the current environment (ignored when using the `--login` option):
```

pages/linux/runuser.md
hieuhtr: `runuser` only run as a specific user at one time, not users, disagree.

pages/linux/runuser.md
hieuhtr: Agree, thanks for reviewing

pages/linux/runuser.md
hieuhtr: `runuser` can run many commands or even some scripts, not only one command. Disagree.

pages/linux/runuser.md
hieuhtr: Same as above, `runuser` can run many commands or even some scripts, not only one command. Disagree.

pages/linux/runuser.md
hieuhtr: `for running or login` is describing exactly the feature of `runuser`. Disagree.

pages/linux/runuser.md
hieuhtr: The original sentence is describing exactly the concept of **preserve** env from `runuser` or even `su` philosophy, we cannot use **inherit** in this case. Disagree.

hieuhtr:
> @hieuhtr we will do for sure, don't worry :) we're just a bit over-encumbered right now with the amount of submitted PRs in such a short period of time.
> 
> At first glance I would suggest to make the description only one line (joining the two together in a single phrase) and change "under substitute" with "as a different". Also, single quotes around "command", "user" and "group" are not needed.

Updated @mebeim, thanks for reviewing.

pages/linux/runuser.md
```diff
@@ -0,0 +1,23 @@
+# runuser
+
+> Run commands as a specific user and group IDs. Does not ask for password. Only useful when run as the root user.
```
mebeim:
```suggestion
> Run commands as a specific user and group without asking for password (needs root privileges).
```

Now that I take a closer look at it, I think this can be simplified in a shorter sentence. What do you think?

pages/linux/runuser.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Run command as a different user and group:
+
+`runuser {{user}} -g {{group}} -c '{{command}}'`
+
+- Start a login shell as a specific user:
+
+`runuser {{user}} -l`
+
+- Specific a shell for running or login instead of default shell:
```
mebeim:
```suggestion
- Specify a [s]hell for running instead of default shell (also works for login):
```
Typo. Also, since the command we show only does one of the two things, let's reword to say it *can* also be used for login.

pages/linux/runuser.md
```diff
@@ -0,0 +1,23 @@
...
+
+`runuser {{user}} -l -s /bin/sh`
+
+- Preserve the entire environment of root (if running by root). Ignore when having option `--login`:
```
mebeim:
```suggestion
- Preserve the entire environment of root (only if `--login` is not specified):
```
Simplified a bit, since the command can basically only be run by root.

pages/linux/runuser.md
```diff
@@ -0,0 +1,23 @@
...
+
+`runuser {{user}} -g {{group}} -c '{{command}}'`
+
+- Start a login shell as a specific user:
+
+`runuser {{user}} -l`
+
+- Specific a shell for running or login instead of default shell:
+
+`runuser {{user}} -l -s /bin/sh`
```
mebeim:
```suggestion
`runuser {{user}} -s {{/bin/sh}}`
```
Wrap shell path in tokens and remove `-l` as per the previous comment.

pages/linux/runuser.md
hieuhtr: Agree, thanks for reviewing.

pages/linux/runuser.md
hieuhtr: Agree, thanks for reviewing.

pages/linux/runuser.md
hieuhtr: Great. Thanks for reviewing.

pages/linux/runuser.md
hieuhtr: Drop `-l` is OK, no need to have this option, thanks for reviewing.

pages/linux/runuser.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Run command as a different user and group:
+
+`runuser {{user}} -g {{group}} -c '{{command}}'`
+
+- Start a login shell as a specific user:
+
+`runuser {{user}} -l`
+
+- Specify a [s]hell for running instead of default shell (also works for login):
```
mebeim:
```suggestion
- Specify a shell for running instead of the default shell (also works for login):
```
Sorry missed the article there, also my bad didn't mean to use `[s]`, copy-paste error. :sweat_smile:

hieuhtr: oh, I think it's format style from our guideline 🙏 ...

hieuhtr: Btw, it's OK

pages/linux/runuser.md
```diff
@@ -0,0 +1,23 @@
+# runuser
+
+> Run commands as a specific user and group without asking for password (needs root privileges).
```
sbrl: If it needs root privileges, perhaps we could prefix the commands themselves with `sudo` instead of mentioning it in the description?

This is what we've done in other pages.

mebeim: Hmm, this is intended to be literally run as root though, since it does not ask for password. Running it with sudo defeats the purpose of the command, for that you have `su`.

sbrl: Ah, I see! In that case this might be fine then.

PR #3291, opened 2019-10-01
rpm: add install package example

mira01:

### DIFF
--- a/pages/linux/rpm.md
+++ b/pages/linux/rpm.md
@@ -10,6 +10,10 @@
 
 `rpm -qa '{{mariadb*}}'`
 
+- Forcibly install a package regardless of currently installed versions:
+
+`rpm -U {{some.rpm}} --force`
+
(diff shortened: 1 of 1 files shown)


pages/linux/rpm.md
```diff
@@ -10,6 +10,10 @@
 
 `rpm -qa '{{mariadb*}}'`
 
+- Install rpm package regardless of currently installed versions:
```
mebeim:
```suggestion
- Forcibly install a package regardless of currently installed versions:
```
I think it reads clearer this way.

pages/linux/rpm.md
```diff
@@ -10,6 +10,10 @@
...
+
+`rpm -U {{some.rpm}} --force`
```
agnivade: some.rpm is really a poor name. While we are here, let's change it to package_name.rpm here and the penultimate one.

PR #3303, opened 2019-10-02
iverilog: add page

AshishShenoy:

### DIFF
--- a/pages/common/iverilog.md
+++ b/pages/common/iverilog.md
@@ -17,7 +17,7 @@
 
 - Compile using Verilog library files from a different path:
 
-`iverilog {{source.v}} -o {{executable}} -I{{library_path}}`
+`iverilog {{source.v}} -o {{executable}} -I{{path/to/library_directory}}`
 
 - Preprocess Verilog code without compiling:
 
--- a/pages/common/iverilog.md
+++ b/pages/common/iverilog.md
@@ -1,0 +1,24 @@
+# iverilog
+
+> Preprocesses and compiles Verilog HDL (IEEE-1364) code, into executable programs for simulation.
+> More information: <http://iverilog.icarus.com/>.
+
+- Compile a source file into an executable:
+
+`iverilog {{source.v}} -o {{executable}}`
+
+- Display all warnings in output:
+
+`iverilog {{source.v}} -Wall -o {{executable}}`
+
+- Compile and run explicitly using the VVP runtime:
+
+`iverilog -o {{execuable}} -tvvp {{source.v}}`
+
+- Compile using Verilog library files from a different path:
+
+`iverilog {{source.v}} -o {{executable}} -I{{library_path}}`
+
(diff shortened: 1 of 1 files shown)


pages/common/iverilog.md
```diff
@@ -0,0 +1,24 @@
+# iverilog
+
+> Preprocesses and compiles Verilog HDL (IEEE-1364) code, into executable programs for simulation.
+> More information: <http://iverilog.icarus.com/>.
+
+- Compile a source file into an executable:
+
+`iverilog {{source.v}} -o {{executable}}`
+
+- Display all warnings in output:
```
sbrl: I think `in output` might be a bit confusing here. What about this?
```suggestion
- Also display all warnings:
```

pages/common/iverilog.md
```diff
@@ -0,0 +1,24 @@
...
+
+`iverilog {{source.v}} -Wall -o {{executable}}`
+
+- Compile and run explicitly using the VVP runtime:
+
+`iverilog -o {{execuable}} -tvvp {{source.v}}`
+
+- Compile using Verilog library files from a different path:
+
+`iverilog {{source.v}} -o {{executable}} -I{{library_path}}`
```
sbrl: Perhaps we could have a specific example here? It would bring it more inline with pre-existing pages.
```suggestion
`iverilog {{source.v}} -o {{executable}} -I{{path/to/library_directory}}`
```

Aracki: Do `-I` need to be stick to the `{{}}`?

pages/common/iverilog.md
```diff
@@ -0,0 +1,24 @@
...
+
+`iverilog {{source.v}} -Wall -o {{executable}}`
+
+- Compile and run explicitly using the VVP runtime:
+
+`iverilog -o {{execuable}} -tvvp {{source.v}}`
+
+- Compile using Verilog library files from a different path:
+
+`iverilog {{source.v}} -o {{executable}} -I{{path/to/library_directory}}`
```
schneiderl: Isn't that "-I{{path/to/library_directory}}" missing a space?

mebeim: Almost any program which takes single letter options only optionally requires a space between the option letter and the value. For example:

image

The variant without space is very commonly used in compilers for `-I` (include path) and `-L` (library path). See for example our pages for [`gcc`](https://github.com/tldr-pages/tldr/tree/master/pages/common/gcc.md), [`g++`](https://github.com/tldr-pages/tldr/tree/master/pages/common/g++.md), [`clang`](https://github.com/tldr-pages/tldr/blob/master/pages/common/clang.md], and so on:

    - Include libraries from a different path:

    `gcc {{source.c}} -o {{executable}} -I{{header_path}} -L{{library_path}} -l{{library_name}}`

schneiderl: Thank you @mebeim 

That being said, the PR looks good to me.

PR #3304, opened 2019-10-02
sdiff: add page

nsuthison:

### DIFF
--- a/pages/common/sdiff.md
+++ b/pages/common/sdiff.md
@@ -2,7 +2,7 @@
 
 > Compare the differences and/or merge between two files.
 
-- Compare between two files:
+- Compare 2 files:
 
 `sdiff {{path/to/first_file}} {{path/to/second_file}}`
 
--- a/pages/common/sdiff.md
+++ b/pages/common/sdiff.md
@@ -1,6 +1,6 @@
 # sdiff
 
-> Compare the differences and/or merge between two files.
+> Compare the differences between and optionally merge 2 files.
 
 - Compare 2 files:
 
--- a/pages/common/sdiff.md
+++ b/pages/common/sdiff.md
@@ -1,0 +1,23 @@
+# sdiff
+
+> Compare the differences and/or merge between two files.
+
+- Compare between two files:
+
+`sdiff {{path/to/first_file}} {{path/to/second_file}}`
+
+- Compare between two files, ignore all tabs and white space:
+
+`sdiff -W {{path/to/first_file}} {{path/to/second_file}}`
+
+- Compare between two files, ignore white space at the end of the line:
+
+`sdiff -Z {{path/to/first_file}} {{path/to/second_file}}`
+
+- Compare between two files, case insensitive:
+
+`sdiff -i {{path/to/first_file}} {{path/to/second_file}}`
+
+- Compare between two files then merge and create new file:
+
+`sdiff -o {{path/to/new_merge_file}} {{path/to/first_file}} {{path/to/second_file}}`
--- a/pages/common/sdiff.md
+++ b/pages/common/sdiff.md
@@ -4,20 +4,20 @@
 
 - Compare 2 files:
 
-`sdiff {{path/to/first_file}} {{path/to/second_file}}`
+`sdiff {{path/to/file1}} {{path/to/file2}}`
 
 - Compare 2 files, ignoring all tabs and whitespace:
 
-`sdiff -W {{path/to/first_file}} {{path/to/second_file}}`
+`sdiff -W {{path/to/file1}} {{path/to/file2}}`
 
 - Compare 2 files, ignoring whitespace at the end of lines:
 
-`sdiff -Z {{path/to/first_file}} {{path/to/second_file}}`
+`sdiff -Z {{path/to/file1}} {{path/to/file2}}`
 
 - Compare 2 files in a case-insensitive manner:
 
-`sdiff -i {{path/to/first_file}} {{path/to/second_file}}`
+`sdiff -i {{path/to/file1}} {{path/to/file2}}`
 
 - Compare and then merge, writing the output to a new file:
 
-`sdiff -o {{path/to/merged_file}} {{path/to/first_file}} {{path/to/second_file}}`
+`sdiff -o {{path/to/merged_file}} {{path/to/file1}} {{path/to/file2}}`
--- a/pages/common/sdiff.md
+++ b/pages/common/sdiff.md
@@ -10,7 +10,7 @@
 
 `sdiff -W {{path/to/first_file}} {{path/to/second_file}}`
 
-- Compare between two files, ignore white space at the end of the line:
+- Compare 2 files, ignoring whitespace at the end of lines:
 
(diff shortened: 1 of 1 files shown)


pages/common/sdiff.md
```diff
@@ -0,0 +1,23 @@
+# sdiff
+
+> Compare the differences and/or merge between two files.
+
+- Compare between two files:
```
sbrl: Perhaps this will read better?
```suggestion
- Compare 2 files:
```

pages/common/sdiff.md
```diff
@@ -0,0 +1,23 @@
...
+
+`sdiff {{path/to/first_file}} {{path/to/second_file}}`
+
+- Compare between two files, ignore all tabs and white space:
```
sbrl: Perhaps this will read better?
```suggestion
- Compare 2 files, ignoring all tabs and whitespace:
```

pages/common/sdiff.md
```diff
@@ -0,0 +1,23 @@
...
+
+`sdiff -W {{path/to/first_file}} {{path/to/second_file}}`
+
+- Compare between two files, ignore white space at the end of the line:
```
sbrl: Perhaps this will read better?
```suggestion
- Compare 2 files, ignoring whitespace at the end of lines:
```

pages/common/sdiff.md
```diff
@@ -0,0 +1,23 @@
...
+
+`sdiff -Z {{path/to/first_file}} {{path/to/second_file}}`
+
+- Compare between two files, case insensitive:
```
sbrl: Perhaps this will read better?
```suggestion
- Case insensitively compare 2 files:
```

pages/common/sdiff.md
```diff
@@ -0,0 +1,23 @@
...
+
+`sdiff -i {{path/to/first_file}} {{path/to/second_file}}`
+
+- Compare between two files then merge and create new file:
```
sbrl: Perhaps this will read better?
```suggestion
- Compare 2 files and then merge, creating a new file:
```

nsuthison:
@sbrl Already fix grammar as your comments. Thank you for your checking and suggestion! :)

pages/common/sdiff.md
```diff
@@ -0,0 +1,23 @@
+# sdiff
+
+> Compare the differences and/or merge between two files.
```
sbrl: Ah, looks like I missed the description What about this?
```suggestion
> Compare the differences between and optionally merge 2 files.
```

pages/common/sdiff.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Compare 2 files:
+
+`sdiff {{path/to/first_file}} {{path/to/second_file}}`
```
agnivade: Let's change first_file to file1 and second_file to file2 everywhere to make things short and simple.

pages/common/sdiff.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Compare 2 files, ignoring all tabs and whitespace:
+
+`sdiff -W {{path/to/first_file}} {{path/to/second_file}}`
+
+- Compare 2 files, ignoring whitespace at the end of lines:
+
+`sdiff -Z {{path/to/first_file}} {{path/to/second_file}}`
+
+- Case insensitively compare 2 files:
```
agnivade: Compare in a case-insensitive manner:

pages/common/sdiff.md
```diff
@@ -0,0 +1,23 @@
...
+
+`sdiff -i {{path/to/first_file}} {{path/to/second_file}}`
+
+- Compare 2 files and then merge, creating a new file:
```
agnivade: Compare and then merge the output to a new file

pages/common/sdiff.md
```diff
@@ -0,0 +1,23 @@
...
+
+`sdiff -o {{path/to/new_merge_file}} {{path/to/first_file}} {{path/to/second_file}}`
```
agnivade: path/to/merged_file

pages/common/sdiff.md
nsuthison: I'm using file1 and file 2 at first and then think first_file and second_file might be better. What do you think? @sbrl

pages/common/sdiff.md
nsuthison: I just change this as @sbrl suggest. Personally I like 'Compare 2 files in a case-insensitive manner'.

What do you think ?

pages/common/sdiff.md
nsuthison: I also just change this by the suggestion from @sbrl . I personally like the one that he suggest more.

pages/common/sdiff.md
nsuthison: I also quite not sure on this, 'new' will communicate clearly to the reader that this is a new file that create after merging.

What do you think? @agnivade

pages/common/sdiff.md
agnivade: Oh I see. "Case insensitively compare" sounded a bit odd to me.

Actually "case insensitively" is just used twice in the repo. Rest are all "case insensitive <verb>". Not a big deal either way.

pages/common/sdiff.md
agnivade: @sbrl suggested to avoid the repeating "and", "then". My suggestion is slightly different.

pages/common/sdiff.md
agnivade: It must always be a new file which gets created right. By definition, if you "create" something, that is new. So I don't think there is any ambiguity here.

pages/common/sdiff.md
nsuthison: Agree, I'll commit this suggestion.

pages/common/sdiff.md
nsuthison: I agree. I'll use 'Compare 2 files in a case-insensitive manner' then :D

pages/common/sdiff.md
nsuthison: Agree, I'll change it.

pages/common/sdiff.md
nsuthison: Quite not sure, for me both is ok. Let ask for third opinion.

@sbrl what do you think?

sbrl: Ah, I see. I'm not quite sure about the `merge the output to` bit specifically there. What about this?
```suggestion
- Compare and then merge, writing the output to a new file:
```

nsuthison: I like this one, do you agree with this? @agnivade

agnivade: I see your point. It doesn't really merge the output, rather it merges the output of the comparisons of the 2 files. However, I really wanted to avoid the comma in the sentence, which kinda stymies the flow. But I don't see a better way. Let's go with your one.

pages/common/sdiff.md
nsuthison: I'll change as your suggest since I also use file1 and file2 in another PR.

Issue #3085, opened 2019-06-04
page request: sa


StewAlexanderACC:
[SA Man Page](http://man7.org/linux/man-pages/man8/sa.8.html)

The "sa" command prints a summary of the commands executed by every user:
```
support@ntopng:/$ sa
    8973    2843.78re       2.48cp         0avio      4076k
     162       3.18re       0.99cp         0avio     31328k   redis-server*
      42       1.12re       0.87cp         0avio     13604k   apt-get
       3       0.20re       0.13cp         0avio     30560k   apt-check
       2       0.49re       0.10cp         0avio    174912k   ZMQbg/49*
      82      10.54re       0.09cp         0avio      5403k   ***other*
      90       0.07re       0.06cp         0avio      4406k   find
      31       0.25re       0.05cp         0avio    153394k   salt-minion*
       2       0.05re       0.04cp         0avio     58152k   gdbus
      36    2617.51re       0.04cp         0avio         0k   kworker/dying*
      57       0.04re       0.03cp         0avio     10246k   lsb_release
      13       0.03re       0.02cp         0avio     10899k   apt-cache
       2       0.03re       0.01cp         0avio      1462k   tripwire
```
(Which includes commands run by services such as cron, salt and zabbix)

* Using ***sa -u*** shows the commands by all users (including many kernel related users such as cron jobs etc),  which produces a rather long totally inclusive list.
```
support    0.04 cpu     4178k mem      0 io sa
support    0.00 cpu     1685k mem      0 io clear
[...]
```
* Better, typically, to grep for the commands run by a specific user using "***sa -u |grep <username>***" like so:
```
support@ntopng:/$ sa -u |grep -i support
root       0.10 cpu    21504k mem      0 io hwe-support-sta
support    0.00 cpu     1861k mem      0 io locale-check
support    0.00 cpu     5336k mem      0 io bash            *
support    0.00 cpu     1867k mem      0 io basename
support    0.00 cpu     1867k mem      0 io dirname
support    0.00 cpu     1157k mem      0 io lesspipe        *
support    0.00 cpu     1157k mem      0 io lesspipe
support    0.00 cpu     5696k mem      0 io bash            *
support@ntopng:/$
```
* Using "***sa -c***" displays the highest percentage of users, useful to find who or what is doing the most on the server:
```
support@ntopng:/$ sa -c
   10404  100.00%    4769.34re  100.00%       2.91cp  100.00%         0avio      4188k
[...]

PR #3315, opened 2019-10-03
sa: add page

glantonb:
Added sa per request from StewAlexanderACC req #3085

### DIFF
--- a/pages/linux/sa.md
+++ b/pages/linux/sa.md
@@ -1,0 +1,18 @@
+# sa
+
+> Summarizes accounting information. Part of the acct package.\
+> Shows commands called by users, including basic info on CPU time spent processing and I/O rates.
+
+- Summary of commands by every user:
+
+`sudo sa`
+
+
+- executables invoked by every user:
+
+`sudo sa --print-users`
+
+
+- List resources used recently per user
+
+`sudo sa --user-summary`
--- a/pages/linux/sa.md
+++ b/pages/linux/sa.md
@@ -3,7 +3,7 @@
 > Summarizes accounting information. Part of the acct package.
 > Shows commands called by users, including basic info on CPU time spent processing and I/O rates.
 
-- Display executable invocations by every user:
+- Display executable invocations by every user (username not displayed):
 
 `sudo sa`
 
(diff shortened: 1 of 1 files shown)


pages/linux/sa.md
```diff
@@ -0,0 +1,16 @@
+# sa
+
+> Summarizes accounting information. Part of the acct package.
+> Shows commands called by users, including basic info on CPU time spent processing and I/O rates.
+
+- Display executable invocations by every user:
+
+`sudo sa`
+
+- Display executable invocations by every user with username:
```
sbrl: Specifically what does this command do differently to the previous one? I'm not sure I understand the description here.

glantonb: Invoking `sa` does not resolve user-ids into readable usernames.

You think it is preferable to modify the text in the very first example - the lack of usernames might surprise users coming to the tool for the first time. In the name of a shorter learning curve a slightly higher bump at the start might be appropriate?

sbrl: That sounds like a good plan, yeah. Let's do that.

pages/linux/sa.md
```diff
@@ -0,0 +1,16 @@
+# sa
+
+> Summarizes accounting information. Part of the acct package.
+> Shows commands called by users, including basic info on CPU time spent processing and I/O rates.
+
+- Display executable invocations by every user (username not displayed):
```
sbrl: Perhaps this will read better?
```suggestion
- Display executable invocations per user (username not displayed):
```

pages/linux/sa.md
```diff
@@ -0,0 +1,16 @@
...
+
+`sudo sa`
+
+- Display executable invocations by every user with username:
```
sbrl: Perhaps this will read better?
```suggestion
- Display executable invocations per user, showing responsible usernames:
```

pages/linux/sa.md
```diff
@@ -0,0 +1,16 @@
+# sa
+
+> Summarizes accounting information. Part of the acct package.
```
schneiderl: Is really important to say that it is a part of the acct package?

sbrl: Hrm, I'm not sure. If we change this then we'll have to go back over the others that are also part of `acct` that have been contributed recently.

schneiderl: In my opinion, the package does not really matter. 

Not a problem for it to be there, though. 
I think we can merge this as it is.

PR #3322, opened 2019-10-04
mcookie: add page

nsuthison:

### DIFF
--- a/pages/linux/mcookie.md
+++ b/pages/linux/mcookie.md
@@ -8,12 +8,12 @@
 
 - Generate a random number, using the contents of a file as a seed for the randomness:
 
-`mcookie -f {{path/to/file1}} -f {{path/to/file2}}`
+`mcookie --file {{path/to/file1}} --file {{path/to/file2}}`
 
 - Generate a random number, using a specific number of bytes from a file as a seed for the randomness:
 
-`mcookie -f {{path/to/file1}} -m {{number_of_bytes}} -f {{path/to/file2}} -m {{number_of_bytes}}`
+`mcookie --file {{path/to/file1}} --max-size {{number_of_bytes}} --file {{path/to/file2}} --max-size {{number_of_bytes}}`
 
 - Print the detials of the randomness such as origin and seed from each sources:
 
-`mcookie -v`
+`mcookie --verbose`
--- a/pages/linux/mcookie.md
+++ b/pages/linux/mcookie.md
@@ -12,7 +12,7 @@
 
 - Generate random number, read specific number of bytes from destination file and use it as a seed for the randomness:
 
-`mcookie -f {{path/to/file1}} -m {{number_of_bytes}} -f {{path/to/file2}} -m {{number_of_bytes}}`s
+`mcookie -f {{path/to/file1}} -m {{number_of_bytes}} -f {{path/to/file2}} -m {{number_of_bytes}}`
 
 - Print the detials of the randomness such as origin and seed from each sources:
 
--- a/pages/linux/mcookie.md
+++ b/pages/linux/mcookie.md
@@ -1,6 +1,6 @@
 # mcookie
 
-> Generate a 128-bit random hexadecimal number.
+> Generates random 128 bit hexadecimal numbers.
 
 - Generate a random number:
 
--- a/pages/linux/mcookie.md
+++ b/pages/linux/mcookie.md
@@ -14,6 +14,6 @@
 
 `mcookie --file {{path/to/file1}} --max-size {{number_of_bytes}} --file {{path/to/file2}} --max-size {{number_of_bytes}}`
 
-- Print the detials of the randomness such as origin and seed from each sources:
+- Print the details of the randomness used, such as the origin and seed for each source:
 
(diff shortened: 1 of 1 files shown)


pages/linux/mcookie.md
```diff
@@ -0,0 +1,19 @@
+# mcookie
+
+> Generate a 128-bit random hexadecimal number.
+
+- Generate random number:
```
sbrl: Perhaps this will read better?
```suggestion
- Generate a random number:
```

pages/linux/mcookie.md
```diff
@@ -0,0 +1,19 @@
+# mcookie
+
+> Generate a 128-bit random hexadecimal number.
```
sbrl: Perhaps this will read better?
```suggestion
> Generates random 128 bit numbers in hexadecimal.
```

pages/linux/mcookie.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Generate random number, use destination file as a seed for the randomness:
+
+`mcookie -f {{path/to/file1}} -f {{path/to/file2}}`
+
+- Generate random number, read specific number of bytes from destination file and use it as a seed for the randomness:
+
+`mcookie -f {{path/to/file1}} -m {{number_of_bytes}} -f {{path/to/file2}} -m {{number_of_bytes}}`
+
+- Print the detials of the randomness such as origin and seed from each sources:
```
sbrl: Perhaps this will read better?
```suggestion
- Print the details of the randomness used, such as the origin and seed for each source of randomness:
```

pages/linux/mcookie.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Generate random number:
+
+`mcookie`
+
+- Generate random number, use destination file as a seed for the randomness:
+
+`mcookie -f {{path/to/file1}} -f {{path/to/file2}}`
+
+- Generate random number, read specific number of bytes from destination file and use it as a seed for the randomness:
```
sbrl: Perhaps this will read better?
```suggestion
- Generate a random number, using a specific number of bytes from a file as a seed for the randomness:
```

pages/linux/mcookie.md
```diff
@@ -0,0 +1,19 @@
+# mcookie
+
+> Generate a 128-bit random hexadecimal number.
+
+- Generate random number:
+
+`mcookie`
+
+- Generate random number, use destination file as a seed for the randomness:
```
sbrl: Perhaps this will read better?
```suggestion
- Generate a random number, using the contents of a file as a seed for the randomness:
```

Also, are there long forms of these options? In tldr-pages we prefer to use long forms of arguments to improve clarity.

nsuthison: > In tldr-pages we prefer to use long forms of arguments to improve clarity.

Ok, I'll change it right away.

pages/linux/mcookie.md
nsuthison: @sbrl What about this?

- Print the details of the randomness used, such as the origin and seed for each source:

pages/linux/mcookie.md
nsuthison: @sbrl What about this ?

Generates random 128 bit hexadecimal number.

sbrl: Almost, @nsuthison. You're missing an article there.
```suggestion
> Generates random 128 bit hexadecimal numbers.
```
It either has to be plural (`numbers`), or include an article (`a random number`).

pages/linux/mcookie.md
sbrl: Sounds good to me!

nsuthison: Thank you a lot! :D

pages/linux/mcookie.md
```diff
@@ -0,0 +1,19 @@
+# mcookie
+
+> Generates random 128 bit hexadecimal numbers.
+
+- Generate a random number:
+
+`mcookie`
+
+- Generate a random number, using the contents of a file as a seed for the randomness:
```
agnivade: I think we can just use one file in these examples, to keep it simple.

nsuthison: I quite not sure about this. Let us have the third opinion please. @sbrl

schneiderl: Well having 2 files surely explains better what the command does. But the whole idea behind tldr is to exemplify in the most simple way the most common form of use of the commands we maintain.

I would prefer to have only one file on the example as more than one could cause some confusion.

nsuthison: Ok, I'll change it to have just only one file. Thank you for your comments! @agnivade @schneiderl

sbrl: Hmmmm. I think, in this instance, if someone wants to specify multiple files, it should be fairly intuitive to use `--file` multiple times. To this end, I think it's ok if we don't show this off specifically here.

It sounds like an advanced use-case, which generally isn't a good fit for a tldr page. If it's something that users of this command would need regularly, then we should probably reconsider.

PR #3325, opened 2019-10-04
exec: correct and clarify example descriptions.

mebeim:
The `-c` example description was wrong, and two others were strangely worded.

### DIFF
--- a/pages/common/exec.md
+++ b/pages/common/exec.md
@@ -6,14 +6,14 @@
 
 `exec {{command -with -flags}}`
 
-- Replace with the specified command using initialized environment variables:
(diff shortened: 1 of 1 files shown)


pages/common/exec.md
```diff
@@ -6,14 +6,14 @@
 
 `exec {{command -with -flags}}`
 
-- Replace with the specified command using initialized environment variables:
+- Replace with the specified command clearing environment variables:
```
agnivade: Should there be a comma after "command" ?

mebeim: Hmm I was kind of unsure, but ok, added.

PR #3336, opened 2019-10-06
tar: simplify extraction examples

toonn:
Dropped the examples for extraction of specific compressed archives, tar
has had autodetection of the compression for over two decades.

Removed the dashes. No versions of tar require them and all they do is
cost you POSIX compatibility. (POSIX tar did *not* allow for the
dashes.)

### DIFF
--- a/pages/common/tar.md
+++ b/pages/common/tar.md
@@ -6,32 +6,28 @@
 
 - Create an archive from files:
 
-`tar -cf {{target.tar}} {{file1 file2 file3}}`
+`tar cf {{target.tar}} {{file1 file2 file3}}`
 
 - Create a gzipped archive:
 
-`tar -czf {{target.tar.gz}} {{file1 file2 file3}}`
+`tar czf {{target.tar.gz}} {{file1 file2 file3}}`
+
+- Extract a (compressed) archive in the current directory:
+
+`tar xf {{source.tar[.gz|.bz2|.xz]}}`
 
 - Extract an archive in a target directory:
 
-`tar -xf {{source.tar}} -C {{directory}}`
-
-- Extract a gzipped archive in the current directory:
-
-`tar -xzf {{source.tar.gz}}`
-
-- Extract a bzipped archive in the current directory:
-
-`tar -xjf {{source.tar.bz2}}`
+`tar xf {{source.tar}} -C {{directory}}`
 
 - Create a compressed archive, using archive suffix to determine the compression program:
 
-`tar -caf {{target.tar.xz}} {{file1 file2 file3}}`
+`tar caf {{target.tar.xz}} {{file1 file2 file3}}`
 
 - List the contents of a tar file:
 
-`tar -tvf {{source.tar}}`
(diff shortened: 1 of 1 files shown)


pages/common/tar.md
```diff
@@ -6,32 +6,28 @@
...
-- Extract a bzipped archive in the current directory:
+- Extract an archive in a target directory:
 
-`tar -xjf {{source.tar.bz2}}`
+`tar xf {{source.tar}} -C {{directory}}`
 
 - Create a compressed archive, using archive suffix to determine the compression program:
 
-`tar -caf {{target.tar.xz}} {{file1 file2 file3}}`
+`tar caf {{target.tar.xz}} {{file1 file2 file3}}`
```
agnivade: We can get rid of the `a` here too.

toonn: Are you sure? The gnutar manpage states you need to enable `--auto-compress` for creation if you want this detection. The bsdtar manpage doesn't even mention it, so maybe it *does* autodetect the compression you want.

Otoh, the `a` option *is* specific to gnutar so I'd be in favor of dropping it for explicit gzip compression choice with `z`. (The other compression formats should be looked up in the manpage, though only xz is noteworthy imo, since bzip2 use has been mostly supplanted by it.)

toonn: Yeah so, I tested GNU tar and it definitely needs the `a` flag for this functionality. How are you checking the tarball is compressed? `file(1)` reports the difference between files created with the same invocation of tar with/without `a` flag.

pages/common/tar.md
```diff
@@ -6,32 +6,28 @@
...
 
 - Create a gzipped archive:
 
-`tar -czf {{target.tar.gz}} {{file1 file2 file3}}`
+`tar czf {{target.tar.gz}} {{file1 file2 file3}}`
 
-- Extract an archive in a target directory:
-
-`tar -xf {{source.tar}} -C {{directory}}`
+- Extract a (compressed) archive in the current directory:
```
sbrl: Perhaps this will read better?
```suggestion
- Extract a (compressed) archive into the current directory:
```

Otherwise, it's implying that the compressed archive to extract has to be in the current directory to perform the extraction.

pages/common/tar.md
```diff
@@ -6,32 +6,28 @@
...
 
-- Extract a gzipped archive in the current directory:
+`tar xf {{source.tar[.gz|.bz2|.xz]}}`
 
-`tar -xzf {{source.tar.gz}}`
-
-- Extract a bzipped archive in the current directory:
+- Extract an archive in a target directory:
```
sbrl: Perhaps this will read better?
```suggestion
- Extract an archive into a target directory:
```

pages/common/tar.md
agnivade: Sorry, you are right. I didn't test it properly enough.

pages/common/tar.md
toonn: Not a new page, just trying to improve upon it. Applied your suggestions.

PR #3338, opened 2019-10-06
fish: add example for environmental variables

samruddhisomani:

### DIFF
--- a/pages/common/fish.md
+++ b/pages/common/fish.md
@@ -23,3 +23,7 @@
 - Display version information and exit:
 
 `fish --version`
+
+- Set and export universal environment variables  
+
+`set -Ux {{variable_name}} {{variable_value}}`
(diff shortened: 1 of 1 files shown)


pages/common/fish.md
```diff
@@ -23,3 +23,7 @@
 - Display version information and exit:
 
 `fish --version`
+
+- Set and export universal environment variables:
```
agnivade: What are universal environment variables ? Is this a commonly used option ? Our scope in tldr is to mainly keep the pages to only use the common options for a user to get started.

samruddhisomani: Universal variables persist across restarts, similar to adding it to the .bashrc in bash. For most use cases for new users of fish shell and/or command line in general, I feel like this is the way users would want to set environmental variables. 

Does `Set and export environmental variables that persist across restarts` explain the intent better?

agnivade: That's much better, thanks !

PR #3342, opened 2019-10-06
k8s-unused-secret-detector: add page

Aracki:
I've added [k8s-unused-secret-detector](https://github.com/dtan4/k8s-unused-secret-detector) - simple tool that detects unused secrets in k8s clusters.


Aracki:
@mebeim Yeah I suppose :D

PR #3349, opened 2019-10-07
dockerd: add page

nsuthison:

### DIFF
--- a/pages/linux/dockerd.md
+++ b/pages/linux/dockerd.md
@@ -1,0 +1,20 @@
+# dockerd
+
+> Starting docker daemon, persistent process that manages docker containers.
+> More information: <https://docs.docker.com/engine/reference/commandline/dockerd/>.
+
+- Run docker daemon and config it to listening to specific sockets:
+
+`dockerd --host {{socket1}} --host {{socket2}}`
+
+- Run docker daemon with specific daemon PID file:
+
+`dockerd --pidfile {{path/to/pid_file}}`
+
+- Run docker daemon in debug mode:
+
+`dockerd --debug={{boolean}}`
+
+- Run docker daemon, set it to log only specific log level or more severe events:
+
+`dockerd --log-level={{log_level}}`
--- a/pages/linux/dockerd.md
+++ b/pages/linux/dockerd.md
@@ -9,7 +9,7 @@
 
 - Run docker daemon and config it to listen to specific sockets(unix,tcp):
 
-`dockerd --host unix://{{path/to/*.sock}} --host tcp://{{IP}}`
+`dockerd --host unix://{{path/to/*.sock}} --host tcp://{{ip}}`
 
 - Run with specific daemon PID file:
 
--- a/pages/linux/dockerd.md
+++ b/pages/linux/dockerd.md
@@ -1,6 +1,6 @@
 # dockerd
 
-> Starting docker daemon, persistent process that manages docker containers.
+> A persistent process to start and manage docker containers.
 > More information: <https://docs.docker.com/engine/reference/commandline/dockerd/>.
 
 - Run docker daemon and config it to listening to specific sockets:
--- a/pages/linux/dockerd.md
+++ b/pages/linux/dockerd.md
@@ -3,18 +3,22 @@
 > A persistent process to start and manage docker containers.
 > More information: <https://docs.docker.com/engine/reference/commandline/dockerd/>.
 
-- Run docker daemon and config it to listening to specific sockets:
+- Run docker daemon:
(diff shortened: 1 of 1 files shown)


pages/linux/dockerd.md
```diff
@@ -0,0 +1,20 @@
+# dockerd
+
+> Starting docker daemon, persistent process that manages docker containers.
+> More information: <https://docs.docker.com/engine/reference/commandline/dockerd/>.
+
+- Run docker daemon and config it to listening to specific sockets:
+
+`dockerd --host {{socket1}} --host {{socket2}}`
```
nsuthison: I quite not sure how to make this example to be clear. Docker daemon can listen to unix, tcp and also https socket(or port). To name the token as socket1, socket2 is it clear enough?

I need your suggestion please. @sbrl

pages/linux/dockerd.md
```diff
@@ -0,0 +1,20 @@
+# dockerd
+
+> Starting docker daemon, persistent process that manages docker containers.
```
schneiderl: I feel like this is not really clear. 
Perhaps this will read better? 
```suggestion
> A persistent process to start and manage docker containers.
```

nsuthison: I feel like we still need to point out that this command is going to start docker daemon. What about

Starting docker daemon, a persistent process to start and manage docker containers.

pages/linux/dockerd.md
nsuthison: Do you have any suggestion about this one ? @schneiderl

pages/linux/dockerd.md
agnivade: All of the descriptions mention "Run docker daemon", so I think we can leave it out from the heading. Managing docker containers is good enough to explain what is happening at a broad level.

nsuthison: Good point, I'll change it.

pages/linux/dockerd.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Run docker daemon with specific daemon PID file:
+
+`dockerd --pidfile {{path/to/pid_file}}`
+
+- Run docker daemon in debug mode:
+
+`dockerd --debug={{boolean}}`
+
+- Run docker daemon, set it to log only specific log level or more severe events:
```
agnivade: A log_level always logs events equal to that or more severe. So we can say "Set a specific log level"

pages/linux/dockerd.md
```diff
@@ -0,0 +1,20 @@
...
+
+`dockerd --log-level={{log_level}}`
```
agnivade: log_level -> debug|info|warn|error|fatal

pages/linux/dockerd.md
```diff
@@ -0,0 +1,20 @@
+# dockerd
+
+> A persistent process to start and manage docker containers.
+> More information: <https://docs.docker.com/engine/reference/commandline/dockerd/>.
+
+- Run docker daemon and config it to listening to specific sockets:
+
+`dockerd --host {{socket1}} --host {{socket2}}`
+
+- Run docker daemon with specific daemon PID file:
```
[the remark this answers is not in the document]
nsuthison: Yeah, you are right for both point.

pages/linux/dockerd.md
```diff
@@ -0,0 +1,24 @@
...
+> A persistent process to start and manage docker containers.
+> More information: <https://docs.docker.com/engine/reference/commandline/dockerd/>.
+
+- Run docker daemon:
+
+`dockerd`
+
+- Run docker daemon and config it to listen to specific sockets(unix,tcp):
+
+`dockerd --host unix://{{path/to/*.sock}} --host tcp://{{IP}}`
```
agnivade: Let's use an actual file. For eg. {{path/to/tmp.sock}}. Also let's change IP to ip.

nsuthison: - Could you please explain why do you want to token name to be some specific file instead of *.sock when it can be any sock file?
- I already change from IP > ip.

agnivade: Please see https://github.com/tldr-pages/tldr/blob/master/contributing-guides/style-guide.md#token-syntax (last point). In short, a token already signifies a variable value, something that the user should change when running the real command. path/to/*.sock is not a valid file path. The concept that it can be any file is already expressed by putting the value in tokens (`{{}}`). After that we can choose any specific value to show what it can be. That's the reason to choose tmp.sock.

nsuthison: Done.

PR #3351, opened 2019-10-07
ceph: add page

alimalek71:

### DIFF
--- a/pages/linux/ceph.md
+++ b/pages/linux/ceph.md
@@ -5,7 +5,7 @@
 
 - Check or watch cluster health:
 
-`ceph {{status | -w}}`
+`ceph status`
 
 - Check cluster usage stats:
 
@@ -17,16 +17,16 @@
 
 - Create a storage pool:
 
-`ceph osd pool create {pool_name} {page_number}`
+`ceph osd pool create {{pool_name}} {{page_number}}`
 
 - Delete a storage pool:
 
-`ceph osd pool delete {pool_name} [--yes-i-really-really-mean-it]`
+`ceph osd pool delete {{pool_name}}`
 
 - Rename a storage pool:
 
-`ceph osd pool rename {current-pool-name} {new-pool-name}`
+`ceph osd pool rename {{current-pool-name}} {{new-pool-name}}`
 
 - Self-repair pool storage:
 
-`ceph pg repair {pool_name}`
+`ceph pg repair {{pool_name}}`
--- a/pages/linux/ceph.md
+++ b/pages/linux/ceph.md
@@ -1,0 +1,32 @@
+# ceph
+
+> Ceph is a free-software storage platform, implements object storage on a single distributed computer cluster, and provides interfaces for object, block and file-level storage.
+> More information: <https://ceph.io>.
+
+- Check or watch cluster health:
+
+`ceph {{status | -w}}`
+
+- Check cluster usage stats:
+  
+`ceph df`
+
+- Get the statistics for the placement groups in your cluster:
+
+`ceph pg dump [--format {{plain | json}}]`
+
+- Create a storage pool:
+
+`ceph osd pool create {pool_name} {page_number}`
+
+- Delete a storage pool:
+
+`ceph osd pool delete {pool_name} [--yes-i-really-really-mean-it]`
+
+- Rename a storage pool:
+
+`ceph osd pool rename {current-pool-name} {new-pool-name}`
+
+- Self-repairing pool storage:
+
+`ceph pg repair {pool_name}`
--- a/pages/linux/ceph.md
+++ b/pages/linux/ceph.md
@@ -11,7 +11,7 @@
 
 `ceph df`
 
-- Get the statistics for the placement groups in your cluster:
+- Get the statistics for the placement groups in a cluster:
 
 `ceph pg dump --format {{plain}}`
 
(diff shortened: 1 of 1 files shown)


pages/linux/ceph.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Delete a storage pool:
+
+`ceph osd pool delete {pool_name} [--yes-i-really-really-mean-it]`
+
+- Rename a storage pool:
+
+`ceph osd pool rename {current-pool-name} {new-pool-name}`
+
+- Self-repairing pool storage:
```
schneiderl: Does this read better? I feel like we should always use verbs to describe what a command does.

```suggestion
- Self-repair pool storage:
```

pages/linux/ceph.md
```diff
@@ -0,0 +1,32 @@
+# ceph
+
+> Ceph is a free-software storage platform, implements object storage on a single distributed computer cluster, and provides interfaces for object, block and file-level storage.
```
schneiderl: We tend to omit adjectives ("free-software", "open source", "top quality") in the main description. Stick to describing what it is and what it does.

pages/linux/ceph.md
alimalek71: I agree.

pages/linux/ceph.md
```diff
@@ -0,0 +1,32 @@
+# ceph
+
+> Ceph implements object storage on a single distributed computer cluster, and provides interfaces for object, block and file-level storage.
+> More information: <https://ceph.io>.
+
+- Check or watch cluster health:
+
+`ceph {{status | -w}}`
```
sbrl: What's the difference between `status` and `-w` here? If they do the same thing, I think I'd prefer `status`, since it's more clear.

pages/linux/ceph.md
```diff
@@ -0,0 +1,32 @@
...
+
+`ceph df`
+
+- Get the statistics for the placement groups in your cluster:
+
+`ceph pg dump [--format {{plain | json}}]`
+
+- Create a storage pool:
+
+`ceph osd pool create {pool_name} {page_number}`
```
sbrl: Tokens need double `{{` and `}}`, not single.

pages/linux/ceph.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Delete a storage pool:
+
+`ceph osd pool delete {pool_name} [--yes-i-really-really-mean-it]`
```
sbrl: Same here. Also, not sure the `[` and `]` are needed there.

pages/linux/ceph.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Rename a storage pool:
+
+`ceph osd pool rename {current-pool-name} {new-pool-name}`
```
sbrl: Same thing with the tokens here. `{{` and `}}`, not `{` and `}`. Also the example below.

pages/linux/ceph.md
```diff
@@ -0,0 +1,32 @@
...
+
+`ceph {{status | -w}}`
+
+- Check cluster usage stats:
+
+`ceph df`
+
+- Get the statistics for the placement groups in your cluster:
+
+`ceph pg dump [--format {{plain | json}}]`
```
sbrl: The syntax make this look quite complicated. What about sticking with something like this?
```suggestion
`ceph pg dump --format {{plain}}`
```

This way, we can indicate that `plain` can be changed, but since JSON output is probably an advanced use-case, we can avoid complicating the page by mentioning it directly.

pages/linux/ceph.md
alimalek71: Actually They both doing something with a little different. The `ceph -w` run in watch mode but `ceph status` write once the status. for more clarity, I'll remove the `-w`.

pages/linux/ceph.md
```diff
@@ -0,0 +1,32 @@
+# ceph
+
+> Ceph implements object storage on a single distributed computer cluster, and provides interfaces for object, block and file-level storage.
```
agnivade: How about - "A tool provide object storage and maintenance on a distributed computer cluster" ?

pages/linux/ceph.md
```diff
@@ -0,0 +1,32 @@
...
+
+`ceph osd pool create {{pool_name}} {{page_number}}`
+
+- Delete a storage pool:
+
+`ceph osd pool delete {{pool_name}}`
+
+- Rename a storage pool:
+
+`ceph osd pool rename {{current-pool-name}} {{new-pool-name}}`
```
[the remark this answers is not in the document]
alimalek71: current_name, new_name ?!

pages/linux/ceph.md
alimalek71: For more simplicity, it is better not to mention `distributed` and I rather say about the main responsibility of Ceph which is handling object storage. How about `Ceph uniquely delivers object, block, and file storage in one unified system.`?

pages/linux/ceph.md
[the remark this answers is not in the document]
agnivade: Works.

pages/linux/ceph.md
agnivade: First off, we can remove "Ceph" from the description because it is mentioned in the title. Secondly, there are still too many words in the heading. Is there a way to combine "object, block and file" into a single thing ?

alimalek71: For combining these words we can simply use `storage system` and my suggested description is `A unified storage system`. It's a general description of Ceph type tools. What about this?

agnivade: SGTM !

PR #1137, opened 2016-10-30
ionic: add page

jobinlawrance:

PR #3352, opened 2019-10-07

schneiderl:
Hi
On multiple occasions ([1](https://github.com/tldr-pages/tldr/pull/3004#pullrequestreview-236040219) [2](https://github.com/tldr-pages/tldr/pull/3020#issuecomment-492260939) [3](https://github.com/tldr-pages/tldr/pull/1137#discussion_r85696118) ) I've seen us talking about omitting adjectives in our command descriptions. 
And, to be honest, it always bothered me a little. 

This PR rephrases a few pages to get rid of some adjectives that slipped in.

If this PR gets approved I intent to update other pages that also have adjectives on the descriptions.

### DIFF
--- a/pages/common/fd.md
+++ b/pages/common/fd.md
@@ -1,6 +1,6 @@
 # fd
 
-> A alternative to `find`.
+> An alternative to `find`.
 > More information: <https://github.com/sharkdp/fd>.
 
 - Find files matching the given pattern in the current directory:
--- a/pages/common/fd.md
+++ b/pages/common/fd.md
@@ -1,6 +1,7 @@
 # fd
 
 > An alternative to `find`.
+> Aims to be faster and easier to use than `find`.
 > More information: <https://github.com/sharkdp/fd>.
 
 - Find files matching the given pattern in the current directory:
--- a/pages/common/fd.md
+++ b/pages/common/fd.md
@@ -1,6 +1,6 @@
 # fd
 
-> A simple, fast and user-friendly alternative to `find`.
+> A alternative to `find`.
 > More information: <https://github.com/sharkdp/fd>.
 
 - Find files matching the given pattern in the current directory:
--- a/pages/common/hugo.md
+++ b/pages/common/hugo.md
@@ -1,6 +1,6 @@
 # hugo
 
-> Hugo is a fast and flexible static site generator.
+> Hugo is a static site generator.
 > More information: <https://gohugo.io>.
 
 - Create a new Hugo site:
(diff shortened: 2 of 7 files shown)


pages/common/fd.md
```diff
@@ -1,6 +1,6 @@
 # fd
 
-> A simple, fast and user-friendly alternative to `find`.
+> A alternative to `find`.
```
andrik:
```suggestion
> An alternative to `find`.
```

pages/common/fd.md
```diff
@@ -1,6 +1,6 @@
 # fd
 
-> A simple, fast and user-friendly alternative to `find`.
+> An alternative to `find`.
```
sbrl: I guess removing them is a good thing to make the pages more impartial, but this now leaves me asking "why would I use it? What's different about it compared to `find`?"

pages/common/hugo.md
```diff
@@ -1,6 +1,6 @@
 # hugo
 
-> Hugo is a fast and flexible static site generator.
+> Hugo is a static site generator.
```
sbrl: Again, this leaves me wondering about many aspects of Hugo. What about this?
```suggestion
> Template-based static site generator. Uses modules, components, and themes.
```

pages/common/ipfs.md
```diff
@@ -1,7 +1,7 @@
 # ipfs
 
 > Inter Planetary File System.
-> A peer-to-peer hypermedia protocol to make the web faster, safer, and more open.
+> A peer-to-peer hypermedia protocol.
```
sbrl: I think we can include maybe a few of these adjectives, if we qualify them correctly. What about this?
```suggestion
> A peer-to-peer hypermedia protocol. Aims to make the web more open.
```

pages/common/rg.md
```diff
@@ -1,6 +1,6 @@
 # ripgrep
 
-> A fast command-line search tool.
+> A command-line search tool.
```
sbrl: Another instance of adjective removal resulting in a lack of detail, in my opinion. What about this?
```suggestion
> A recursive line-oriented CLI search tool.
> Aims to be a faster alternative to `grep`.
```

pages/common/fd.md
schneiderl: I see your concern and agree with all of your comments. It's a matter of rephrasing the descriptions properly. For this specific case, though, I have no idea of how to phrase it differently.

sbrl: Hrm. Taking a look at the README, what about this?
```suggestion
> An alternative to `find`.
> Aims to be faster and easier to use than `find`.
```

schneiderl: Looks good.

PR #3353, opened 2019-10-07
setxkbmap: add complex example and available values examples

PrimaMateria:

### DIFF
--- a/pages/linux/setxkbmap.md
+++ b/pages/linux/setxkbmap.md
@@ -10,6 +10,10 @@
 
 `setxkbmap -layout us,de -variant ,qwerty -option 'grp:alt_caps_toggle'`
 
+- Get help:
+
+`setxkbmap -help`
+
 - List all layouts:
 
 `localectl list-x11-keymap-layouts`
--- a/pages/linux/setxkbmap.md
+++ b/pages/linux/setxkbmap.md
@@ -12,7 +12,7 @@
 
 - List all layouts:
 
-`localectl list-x11-keymap-layouts - gives you layouts`
+`localectl list-x11-keymap-layouts`
 
 - List variants for the layout:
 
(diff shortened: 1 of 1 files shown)


PrimaMateria:
Hi @schneiderl 
yes, I thought that this is quite obvious command. But on the second thought `-h` and `--help` don't work. Should I put it back?

schneiderl:
@PrimaMateria I agree with you that it is obvious but given that `-h` and `--help` don't work I say we put it back.

pages/linux/setxkbmap.md
```diff
@@ -6,6 +6,22 @@
 
 `setxkbmap {{fr}}`
 
+- Set multiple keyboard layouts, their variants and switching option:
+
+`setxkbmap -layout us,de -variant ,qwerty -option 'grp:alt_caps_toggle'`
```
sbrl: We should wrap the bits that users can change in `{{` and `}}`.

PrimaMateria: I added curly braces to all arguments that makes sense for user to change. But I was bit unsure if I need to set it to all changeable values or not. For example `grep :grp` or `grep {{:grp}}` in the last example.

sbrl: Cool, thanks!

Really, it's only the values that makes sense for the user to change in the contact of your command example.

pages/linux/setxkbmap.md
```diff
@@ -6,6 +6,22 @@
...
+
+`setxkbmap -layout {{us,de}} -variant {{,qwerty}} -option {{'grp:alt_caps_toggle'}}`
+
 - Get help:
 
 `setxkbmap -help`
+
+- List all layouts:
+
+`localectl list-x11-keymap-layouts`
```
agnivade: These seem to be a different command. Is this fine ?

PR #3357, opened 2019-10-08
htpasswd: add direct generation of username/password

ursbraem:

### DIFF
--- a/pages/common/htpasswd.md
+++ b/pages/common/htpasswd.md
@@ -22,3 +22,7 @@
 - Verify user password:
 
 `htpasswd -v {{path/to/file}} {{user_name}}`
+
+- Generate username/password string (md5) for manual use:
+
+`htpasswd -nbm {{user_name}} {{password}}`
--- a/pages/common/htpasswd.md
+++ b/pages/common/htpasswd.md
@@ -23,6 +23,6 @@
 
 `htpasswd -v {{path/to/file}} {{user_name}}`
 
-- Generate username/password string (md5) for manual use:
+- Display a string with username (plain text) and password (md5):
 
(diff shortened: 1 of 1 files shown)


pages/common/htpasswd.md
```diff
@@ -22,3 +22,7 @@
 - Verify user password:
 
 `htpasswd -v {{path/to/file}} {{user_name}}`
+
+- Generate username/password string (md5) for manual use:
```
schneiderl: I'm not sure I understood this description. Does the command concatenate usarname+password then hashes it?

ursbraem: No, it will output the line that is normally written into the htpasswd file, e.g:

`myusername:$apr1$MlN143435.A9jJJQrd17ABD9LIt.`

Context: I use this all the time, as my server's control panel demands complex passwords, but you don't want complex passwords all the time (eg if it's not security related but just to prevent a site from being reached by everybody pre-launch). So I generate the content of htpassword file by hand and paste it.

Maybe better like this?:

`Generate username (plain text) and password (md5) to paste into passwd file`

schneiderl: This does read much better to me. Thank you for the explanation 👍

agnivade: `Output string` is not grammatically correct. May be "Display a string .."

ursbraem: Or "return a string"

ursbraem:
@schneiderl ok with the wording?

schneiderl:
@ursbraem I'm sorry for the (very) late answer. I was on vacation over the past two weeks and very busy before that. 
But the wording looks good indeed.

PR #3367, opened 2019-10-10
README: remove deprecated rust client

keawade:
I think this should probably be removed because it is no longer possible to install via `cargo install tldr` as the `tldr` crate has been [yanked from the crates.io package registry](https://crates.io/crates/tldr).

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,13 +112,8 @@ You can access these pages on your computer using one of the following clients:
   `devtools::install_github('kirillseva/tldrrr')`
 - [Ruby client](https://github.com/YellowApple/tldrb):
   `gem install tldrb`
-- Rust clients:
-    - [rust-tldr](https://github.com/rilut/rust-tldr)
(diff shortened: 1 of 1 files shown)


agnivade:
@dbrgn for further comments.

dbrgn:
Yep, the `tldr` crate is unmaintained. The author offered me to take over the `tldr` name, but I think I'll stick with `tealdeer`.

agnivade:
Do you have a new crate named `tealdeer` that we can use ?

dbrgn:
@agnivade hm, I don't quite understand. My implementation (https://github.com/dbrgn/tealdeer/) is still active and it's also published on crates.io: https://crates.io/search?q=tealdeer

The other implementation (https://crates.io/crates/tldr) is inactive and was yanked from the crates.io index.

agnivade:
Nvm, I didn't read the PR completely.

PR #3371, opened 2019-10-10
firefox: add extra flag options

scttrgd:
also added note that profile selection is not available on windows

### DIFF
--- a/pages/common/firefox.md
+++ b/pages/common/firefox.md
@@ -27,6 +27,14 @@
 
 `firefox --headless --screenshot {{path/to/output_file.png}} {{https://example.com/}}`
 
-- Use a specific profile directory to allow multiple separate instances of Firefox to run at once:
+- Use a specific profile directory to allow multiple separate instances of Firefox to run at once (not available for Windows):
 
 `firefox --new-instance --profile {{path/to/directory}} {{https://example.com/}}`
+
+- Create a new profile in the default directory without starting Firefox:
+
+`firefox --CreateProfile {{profile_name}}`
+
+- Set Firefox as default browser:
+
+`firefox --setDefaultBrowser
--- a/pages/common/firefox.md
+++ b/pages/common/firefox.md
@@ -27,10 +27,6 @@
 
 `firefox --headless --screenshot {{path/to/output_file.png}} {{https://example.com/}}`
 
-- Create a new profile in the default directory without starting Firefox:
-
-`firefox --CreateProfile {{profile_name}}`
-
 - Use a specific profile to allow multiple separate instances of Firefox to run at once:
 
 `firefox --profile {{path/to/directory}} {{https://example.com/}}`
--- a/pages/common/firefox.md
+++ b/pages/common/firefox.md
@@ -37,4 +37,4 @@
 
 - Set Firefox as default browser:
 
-`firefox --setDefaultBrowser
+`firefox --setDefaultBrowser`
--- a/pages/common/firefox.md
+++ b/pages/common/firefox.md
@@ -31,6 +31,10 @@
 
 `firefox --CreateProfile {{profile_name}}`
 
+- Use a specific profile to allow multiple separate instances of Firefox to run at once:
+
+`firefox --profile {{path/to/directory}} {{https://example.com/}}`
+
 - Set Firefox as the default browser:
 
 `firefox --setDefaultBrowser`
--- a/pages/common/firefox.md
+++ b/pages/common/firefox.md
@@ -31,6 +31,6 @@
 
 `firefox --CreateProfile {{profile_name}}`
 
-- Set Firefox as default browser:
+- Set Firefox as the default browser:
 
 `firefox --setDefaultBrowser`
--- a/pages/common/firefox.md
+++ b/pages/common/firefox.md
@@ -27,10 +27,6 @@
 
 `firefox --headless --screenshot {{path/to/output_file.png}} {{https://example.com/}}`
 
-- Use a specific profile directory to allow multiple separate instances of Firefox to run at once (not available for Windows):
-
-`firefox --new-instance --profile {{path/to/directory}} {{https://example.com/}}`
-
 - Create a new profile in the default directory without starting Firefox:
 
 `firefox --CreateProfile {{profile_name}}`


pages/common/firefox.md
```diff
@@ -27,6 +27,14 @@
 
 `firefox --headless --screenshot {{path/to/output_file.png}} {{https://example.com/}}`
 
-- Use a specific profile directory to allow multiple separate instances of Firefox to run at once:
+- Use a specific profile directory to allow multiple separate instances of Firefox to run at once (not available for Windows):
```
schneiderl: Well now this is interesting. I feel like we shouldn't have an example that is not available to all platforms in the `common` directory.
If thats the case we have two options:
1. Keep this `common` page, while creating a new page under `windows` but without this example. 
2. Delete this page, recreate it under `linux` (and maybe macOS too?) while also creating a new page under `windows` without this example. 

What's the way to go here, guys?

schneiderl: Well now this is interesting. I feel like we shouldn't have an example that is not available to all platforms in the common directory.
If that's the case we have two options:

Keep this `common` page, while creating a new page under `windows` but without this example.
Delete this page, recreate it under `linux` (and maybe `macOS` too?) while also creating a new page under `windows` without this example.
What's the way to go here, guys?

sbrl: Wait, this doesn't work for Windows? That _is_ surprising. What version of Firefox is that? If it's an old version it won't work, since it's a recent thing.

Also, I'd suggest we create a new page under Windows and keep the common one if we need to go down that route.

scttrgd: I was digging through the [firefox cli docs](https://developer.mozilla.org/en-US/docs/Mozilla/Command_Line_Options#-new-instance), I'm not actually running the windows version. However I have just noticed that the functionality may be being removed as functionality, so it might be worth removing the flag in its entirety?

sbrl: It is? Do you have a reference for that? It would be highly unfortunate if that was the case, @scttrgd

scttrgd: [This is the reference to it being removed as supported](https://bugzilla.mozilla.org/show_bug.cgi?id=855899#ct-37). But continuing to dig through conversations I'm not actually seeing it being removed, which is a little annoying. I've probably missed something from [this google group conversation](url). Apologies for muddying the waters here...

sbrl: Ah, I see. Thanks, @scttrgd! Reading into that, I see this:

> The new remoting does everything based on profile. When starting Firefox we
do normal profile selection, which includes considering any -P and
--profile command line arguments. Once we've selected a profile we attempt
to find an existing Firefox instance using that profile. If one is found we
send it our command line arguments and quit. If not continue start up. 

So to this end, we can instead remove the `--new-instance` flag, as in a future version of Firefox it will no longer be needed. However, I would prefer to make that change to tldr-pages once the change you linked to lands in stable.

pages/common/firefox.md
```diff
@@ -27,6 +27,10 @@
 
 `firefox --headless --screenshot {{path/to/output_file.png}} {{https://example.com/}}`
 
-- Use a specific profile directory to allow multiple separate instances of Firefox to run at once:
+- Create a new profile in the default directory without starting Firefox:
 
-`firefox --new-instance --profile {{path/to/directory}} {{https://example.com/}}`
+`firefox --CreateProfile {{profile_name}}`
```
sbrl: Hold on. This wasn't quite what I was trying to suggest. What about just removing the `--new-instance`, and leaving the `--profile`?

Also, I'm not seeing any documentation for `--CreateProfile` - and the casing is different to the other arguments int he rest of the page. Could you clarify where this is from please?

pages/common/firefox.md
```diff
@@ -27,6 +27,10 @@
...
+
+- Set Firefox as default browser:
```
sbrl: Perhaps this will read better?
```suggestion
- Set Firefox as the default browser:
```

pages/common/firefox.md
scttrgd: Here's the [specific documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Command_Line_Options#-CreateProfile_profile_name) for --CreateProfile. I'm not sure why it's different casing, but there are a few others like [-setDefaultBrowser](https://developer.mozilla.org/en-US/docs/Mozilla/Command_Line_Options#-setDefaultBrowser) may be an old flag from Thunderbird as they share `--CreateProfile` but that doesn't explain the others...

scttrgd: > Hold on. This wasn't quite what I was trying to suggest. What about just removing the `--new-instance`, and leaving the `--profile`?

My bad, I'll fix that now (double checked and it does still function correctly with just `--profile`)

sbrl: Thanks :-)

scttrgd:
I personally rarely use the `--search` flag, but I obviously don't speak for everyone. Any objections to removing that flag?

sbrl:
I mean I'm not a huge user of the CLI for Firefox. I'd tend towards removing the most complex and least frequently used command myself. Commands like that are generally useful only in scripting, which isn't generally

pages/common/firefox.md
```diff
@@ -27,6 +27,14 @@
 
 `firefox --headless --screenshot {{path/to/output_file.png}} {{https://example.com/}}`
 
-- Use a specific profile directory to allow multiple separate instances of Firefox to run at once:
+- Use a specific profile directory to allow multiple separate instances of Firefox to run at once (not available for Windows):
 
 `firefox --new-instance --profile {{path/to/directory}} {{https://example.com/}}`
+
+- Create a new profile in the default directory without starting Firefox:
```
zdroid: This seems the least useful to me. Most of the time you use only one profile, especially as a beginner. How about removing this to resolve the issue?

sbrl: Hrm, yeah. Firefox should create a new profile automatically if you specify the profile directory.

@zdroid: Only 1 instance of Firefox per profile may be started at any given time. To this end, if you do headless CLI Firefox stuff you usually need to specify a different directory for it to use instead.

zdroid: So to conclude, do you agree with me? I'll commit the change if you ultimately do.

sbrl: I do ultimately agree, but for different reasons, which I felt is an important distinction.

zdroid:
There has been no response for 2 months, so I committed the changes myself.

PR #3372, opened 2019-10-10
whatis: add osx page

Aracki:
Simple page added. http://www.linfo.org/whatis.html


sbrl:
We've already got a page for `whatis` here: https://github.com/tldr-pages/tldr/blob/master/pages/linux/whatis.md

Perhaps you could merge yours with that one?

Also, if `whatis` is available on multiple platforms, we should move it to `common` I guess.

Aracki:
I've just investigated. I suggest I move it to `osx` dir. It seems it has totally different flags and different outputs.

sbrl:
Cool. Does the macOS version contain any other interesting flags we could add to the page?

Aracki:
@sbrl Linux version has multiple flags and different outputs. macOS version has no flags at all according to `man whatis`.

pages/osx/whatis.md
```diff
@@ -0,0 +1,12 @@
+# whatis
+
+> Tool that searches a set of database files containing short descriptions of system commands for keywords.
+> More information: <http://www.linfo.org/whatis.html>.
+
+- Search for informations about keyword:
```
agnivade: `informations` is grammatically incorrect. Let's also change to "about _a_ keyword"

Aracki: thanks @agnivade I didn't know `information` is non-countable word. I will make another PR for the fix.

PR #3373, opened 2019-10-10
info: add page

Aracki:
Even though `man` is more popular these days, I've added old command [info](https://en.wikipedia.org/wiki/Info_(Unix)).

### DIFF
--- a/pages/common/info.md
+++ b/pages/common/info.md
@@ -3,11 +3,11 @@
 > Reads documentation stored in the info format.
 > More information: <https://en.wikipedia.org/wiki/Info_(Unix)>.
 
-- Start reading top-level dir menu:
+- Start reading top-level directory menu:
 
 `info`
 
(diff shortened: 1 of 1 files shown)


pages/common/info.md
```diff
@@ -0,0 +1,16 @@
+# info
+
+> Reads documentation stored in the info format.
+> More information: <https://en.wikipedia.org/wiki/Info_(Unix)>.
+
+- Start reading top-level dir menu:
```
agnivade: expand dir to directory for clarity

pages/common/info.md
```diff
@@ -0,0 +1,16 @@
...
+
+`info`
+
+- Start reading at given menu item node from top-level dir:
```
agnivade: Same as above

pages/common/info.md
Aracki: done! 👍

PR #3341, opened 2019-10-06
k8sec: add page

Aracki:
I've added [k8sec](https://github.com/dtan4/k8sec) - CLI tool for managing k8s secrets easily.

PR #3378, opened 2019-10-11
k8sec: fix replacement of theses

Aracki:
Small fix. In #3341 I've messed up by replacing thesis about encoding/decoding.

### DIFF
--- a/pages/common/k8sec.md
+++ b/pages/common/k8sec.md
@@ -7,7 +7,7 @@
 
 `k8sec list`
 
-- List a specific secret and decode it with base64:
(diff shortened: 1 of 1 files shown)


pages/common/k8sec.md
```diff
@@ -7,7 +7,7 @@
 
 `k8sec list`
 
-- List a specific secret and decode it with base64:
+- List a specific secret as base64-encoded string:
```
sbrl: Perhaps this will read better?
```suggestion
- List a specific secret as a base64-encoded string:
```

PR #3380, opened 2019-10-11
aptitude: add search term and hold examples

junjuew:

### DIFF
--- a/pages/linux/aptitude.md
+++ b/pages/linux/aptitude.md
@@ -14,6 +14,10 @@
 
 `aptitude search {{package}}`
 
+- Search for an installed package. `?installed` is an aptitude search term:
+
+`aptitude search '?installed({{package}})'`
+
 - Remove a package and all packages depending on it:
 
 `aptitude remove {{package}}`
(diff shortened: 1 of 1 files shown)


pages/linux/aptitude.md
```diff
@@ -14,6 +14,10 @@
 
 `aptitude search {{package}}`
 
+- Search for an installed package. `?installed` is an aptitude search term:
```
sbrl: Perhaps this will read better?
```suggestion
- Search for an installed package (`?installed` is an aptitude search query keyword):
```

junjuew: The parenthesis indeed makes it read better. Thank you for the suggestion. I've changed the separate sentence to be in the parenthesis.

The word 'search term' is used in the aptitude manual. The intention of keeping it here is to make it easier for people to search and land on its reference page: https://www.debian.org/doc/manuals/aptitude/ch02s04s05.en.html.

sbrl: Ah, I see! I wasn't aware. In that case, perhaps we should keep it as search term then.

junjuew: Sounds good. I'll go ahead and mark this as resolved then.

Issue #749, opened 2016-01-29
page request: csvkit (for common args)


oxguy3:
I recently contributed tldr pages for a lot of the tools in [csvkit](https://csvkit.readthedocs.org/en/0.9.1/cli.html), a command-line utility belt for dealing with CSV files. Since the pages got merged and I've actually been using them in practice, I'm noticing that I really wish I had more information.

Every tool in csvkit has its own set of arguments and options, but there's also [a lot of options common to all commands](https://csvkit.readthedocs.org/en/0.9.1/common_arguments.html). Many of these options tend to be very useful very often, but there wasn't really space to cover them on each individual tldr page.

So here's my idea: There's no "csvkit" command, but we could make a "csvkit.md" tldr page. This page could be used exclusively for documenting the shared arguments. On the pages for each individual command, the note "Included in csvkit." could be replaced with "See `tldr csvkit` for further information."

This would allow these very useful arguments to be documented without overfilling any one tldr page. But it'd be a bit of a deviation from normal tldr-pages operating procedures, so I figured I'd ask before making a PR. What do you guys think?

oxguy3:
The arguments I'd probably want to include in csvkit.md would be `-d` (specify weird delimiter), `-t` (specify delimiter is tab character, overrides `-d`), `-q` (specify weird quote character), and `-H` (denote that there is no header row)

waldyrious:
I agree with your thinking :+1:

waldyrious:
Fixed by @GuilhermeJSilva in #3386! 🎉

PR #702, opened 2016-01-21
Added most of csvkit

oxguy3:
Added most of the tools from [csvkit](https://csvkit.readthedocs.org), a command-line toolbelt for dealing with CSVs and other tabular data formats. Specifically, I added:

* csvclean
* csvcut
* csvformat
* csvgrep
* csvlook
* csvpy
* csvsort
* csvstat
* in2csv

(still missing from csvkit: csvstack, csvjson, csvsql, sql2csv)

PR #701, opened 2016-01-21
Added csvkit, other misc tools

oxguy3:
Added most of the tools from [csvkit](https://csvkit.readthedocs.org), a command-line toolbelt for dealing with CSVs and other tabular data formats. Specifically, I added:
* csvclean
* csvcut
* csvformat
* csvgrep
* csvlook
* csvpy
* csvsort
* csvstat
* in2csv

(still missing from csvkit: csvstack, csvjson, csvsql, sql2csv)

I also added a few other random pages:

* archey (system information tool for OS X)
* dhcpwn (DHCP vulnerability tester)
* lwp-request (command-line HTTP client)
* mmv (mass file move/rename tool)
* redshift (screen color-temperature adjuster)
* watch (monitors output of a command)
* xz (compression tool)
* youtube-dl (downloader for various video-hosting websites)

PR #3386, opened 2019-10-12
csvkit: add file

GuilhermeJSilva:

#749

### DIFF
--- a/pages/common/csvkit.md
+++ b/pages/common/csvkit.md
@@ -1,6 +1,7 @@
 # csvkit
 
 > Manipulation toolkit for CSV files.
+> For more information see commands: `csvclean`, `csvcut`, `csvformat`, `csvgrep`, `csvlook`, `csvpy`, `csvsort`, `csvstat`.
 > More information: <https://csvkit.readthedocs.io/en/0.9.1/cli.html>.
 
 - Run a command on a CSV file with the non default delimiter:
--- a/pages/common/csvkit.md
+++ b/pages/common/csvkit.md
@@ -4,15 +4,15 @@
 > For more information see commands: `csvclean`, `csvcut`, `csvformat`, `csvgrep`, `csvlook`, `csvpy`, `csvsort`, `csvstat`.
 > More information: <https://csvkit.readthedocs.io/en/0.9.1/cli.html>.
 
-- Run a command on a CSV file with the non default delimiter:
+- Run a command on a CSV file with a custom delimiter:
 
 `{{cmd}} -d {{delimiter}} {{filename.csv}}`
 
-- Run a command on a CSV file with a tab as a delimiter (overides -d):
+- Run a command on a CSV file with a tab as a delimiter (overrides -d):
 
 `{{cmd}} -t {{filename.csv}}`
 
-- Run a command on a CSV file with the non default quote character:
+- Run a command on a CSV file with a custom quote character:
 
 `{{cmd}} -q {{quote_char}} {{filename.csv}}`
 
--- a/pages/common/csvkit.md
+++ b/pages/common/csvkit.md
@@ -1,7 +1,7 @@
 # csvkit
 
 > Manipulation toolkit for CSV files.
-> For more information see commands: `csvclean`, `csvcut`, `csvformat`, `csvgrep`, `csvlook`, `csvpy`, `csvsort`, `csvstat`.
+> See the individual commands: `csvclean`, `csvcut`, `csvformat`, `csvgrep`, `csvlook`, `csvpy`, `csvsort`, `csvstat`.
 > More information: <https://csvkit.readthedocs.io/en/0.9.1/cli.html>.
 
 - Run a command on a CSV file with a custom delimiter:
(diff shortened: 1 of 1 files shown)


waldyrious:
Hey @GuilhermeJSilva, thanks for the PR! Please note that `csvkit` doesn't seem to be a single command, but a collection of commands that probably should each have their own page. Please take a look at PR #702 for some context, and let me know what you think.

GuilhermeJSilva:
Hey, in the #759 the request suggests a single page with all options, because would otherwise crowd every page with the same options, however I did not add the info to each command page to show that more information is available elsewhere as it is suggested in the aforementioned issue.

waldyrious:
Good point. I'd suggest adding a second line to the main description suggesting the user to look up the individual pages.

By the way, we should do the same in similar commands like `git` and `docker` — but that's for another PR :)

waldyrious:
Hey Guilherme, I've asked the other community members about the Travis failures. That seems to be a bug. Please follow the discussion [here](https://gitter.im/tldr-pages/tldr?at=5da1dffe894dee56e52e2644).

pages/common/csvkit.md
```diff
@@ -0,0 +1,21 @@
+# csvkit
+
+> Manipulation toolkit for CSV files.
+> For more information see commands: `csvclean`, `csvcut`, `csvformat`, `csvgrep`, `csvlook`, `csvpy`, `csvsort`, `csvstat`.
+> More information: <https://csvkit.readthedocs.io/en/0.9.1/cli.html>.
+
+- Run a command on a CSV file with the non default delimiter:
```
sbrl: Perhaps this will read better?
```suggestion
- Run a command on a CSV file with a custom default delimiter:
```

Also, this seems like a more advanced usage of the command. Perhaps we could include a basic one first? Generally, the examples on a tldr-page should go in the order that you might use them in.

Also, what does the `csvkit` command actually do? You mention running it on a CSV file, but don't mention what it would actually do to the CSV file.

pages/common/csvkit.md
```diff
@@ -0,0 +1,21 @@
...
+
+`{{cmd}} -d {{delimiter}} {{filename.csv}}`
```
sbrl: I'm a bit confused on `{{cmd}}` here. Generally, a each command should have it's own tldr page. If there are multiple related commands, a description line is fine - but perhaps we could put the command name here?
```suggestion
`csvkit -d {{delimiter}} {{filename.csv}}`
```

The same goes for the examples below.

pages/common/csvkit.md
```diff
@@ -0,0 +1,21 @@
...
+
+- Run a command on a CSV file with a tab as a delimiter (overides -d):
```
sbrl: Perhaps this will read better?
```suggestion
- Run a command on a CSV file with a tab as a delimiter (overrides `-d`):
```

pages/common/csvkit.md
GuilhermeJSilva: `csvkit` is not a command, but a kit of several commands which have several common options this page serves as a reference to all of them as it was suggested in the issue #749

pages/common/csvkit.md
GuilhermeJSilva: Corrected

pages/common/csvkit.md
sbrl: Ah, I see. In that case it's probably fine :-)

sbrl:
Ping @waldyrious on this one I think.

waldyrious:
Oh, thanks @mebeim. I was quite sure we had a discussion about the csvkit beyong #701 and #702, but failed to resurface #749.

IMO it does make sense to aggregate common options to various subcommands in the main page, unless there are options specific to a root `csvkit`c command which we should document instead. Basically, I agree with the plan @oxguy3 proposed in #749.

An aside, for additional context: I often try to use Homebrew to directly install tools that come with a larger package (e.g. ruby's `gem`, or rust's `cargo`), and I've found it frustrating that Homebrew does not make an explicit connection between them, requiring manual tweaks such as https://github.com/Homebrew/brew/pull/5678 or https://github.com/Homebrew/brew/pull/5769. IMO we should strive to be helpful to people looking for both the specific and the generic (possible better-known) tool names, rather than attempt a one-page-one-command rule.

pages/common/csvkit.md
```diff
@@ -0,0 +1,21 @@
+# csvkit
+
+> Manipulation toolkit for CSV files.
+> For more information see commands: `csvclean`, `csvcut`, `csvformat`, `csvgrep`, `csvlook`, `csvpy`, `csvsort`, `csvstat`.
```
mebeim:
```suggestion
> See the individual commands: `csvclean`, `csvcut`, `csvformat`, `csvgrep`, `csvlook`, `csvpy`, `csvsort`, `csvstat`.
```

sbrl:
Are we ok to merge, @waldyrious?

waldyrious:
> Are we ok to merge, @waldyrious?

Yes! Apologies for the delay in getting to this. Thanks for the contribution, @GuilhermeJSilva! 🎉

PR #3389, opened 2019-10-12
hdiutil: add page

dpinela:

### DIFF
--- a/pages/osx/hdiutil.md
+++ b/pages/osx/hdiutil.md
@@ -1,0 +1,19 @@
+# hdiutil
+
+> Utility to create and manage disk images.
+
+- Mount an image:
+
+`hdiutil attach {{filename}}`
+
+- Unmount an image:
+
+`hdiutil detach /Volumes/{{volumename}}`
+
+- List mounted images:
+
+`hdiutil info`
+
+- Create an ISO image from the contents of a directory:
+
+`hdiutil makehybrid -o {{filename}} {{dirname}}`


waldyrious:
Nice contribution, thanks! I have nothing to point out in the contents, but maybe consider adding a URL, if there's one we could use. Otherwise, LGTM! 👍

dpinela:
> Nice contribution, thanks! I have nothing to point out in the contents, but maybe consider adding a URL, if there's one we could use. Otherwise, LGTM! 👍

This is a built-in utility in macOS. I could add a link to the man page (though I can't find them on Apple's website) but none of the others seemed to have it, so I did the same.

waldyrious:
Yeah, many pages were created before we introduced the URL field. But sure, I suspected that a link wouldn't be available. Apple used to host manpages in the past but for some reason they discontinued that practice :/

Given that, I think this is ready to go. Thanks!

pages/osx/hdiutil.md
```diff
@@ -0,0 +1,19 @@
+# hdiutil
+
+> Utility to create and manage disk images.
+
+- Mount an image:
+
+`hdiutil attach {{filename}}`
```
agnivade: All of the tokens should be snake_case, like file_name, volume_name.

pages/osx/hdiutil.md
```diff
@@ -0,0 +1,19 @@
...
+
+`hdiutil detach /Volumes/{{volumename}}`
+
+- List mounted images:
+
+`hdiutil info`
+
+- Create an ISO image from the contents of a directory:
+
+`hdiutil makehybrid -o {{filename}} {{dirname}}`
```
agnivade: dirname should be path/to/directory

pages/osx/hdiutil.md
sbrl: I think `filename` is ok as-is - I believe we've done `path/to/filename` before though.

agnivade: Yeah ok, I see a lot of `filename`s. But we should change dirname to path/to/directory though. There are no other `dirname`s

PR #3400, opened 2019-10-12
reboot: add page (for macOS)

dpinela:

### DIFF
--- a/pages/osx/reboot.md
+++ b/pages/osx/reboot.md
@@ -6,6 +6,6 @@
 
 `sudo reboot`
 
-- Reboot immediately without shutting down gracefully:
+- Reboot immediately without gracefully shutting down:
 
 `sudo reboot -q`
--- a/pages/osx/reboot.md
+++ b/pages/osx/reboot.md
@@ -1,0 +1,11 @@
+# reboot
+
+> Reboot the system.
+
+- Reboot immediately:
+
+`sudo reboot`
+
+- Reboot immediately without shutting down gracefully:
+
+`sudo reboot -q`


dpinela:
> Thanks, @dpinela!
> 
> I assume the flags on macOS' `reboot` are different to Linux?

Yes, they're very different. This is what the macOS man page says:

```
     The options are as follows:

     -l      The halt or reboot is not recorded in the system log.  This option is
             intended for applications such as shutdown(8), that call reboot or halt
             and log this themselves.

     -n      The file system cache is not flushed.  This option should probably not
             be used.

     -q      The system is halted or restarted quickly and ungracefully, and only the
             flushing of the file system cache is performed (if the -n option is not
             specified).  This option should probably not be used.

     -u      The system is halted up until the point of removing system power, but
             waits before removing power for 5 minutes so that an external UPS (unin-
             terruptible power supply) can forcibly remove power.  This simulates a
             dirty shutdown to permit a later automatic power on. OS X uses this mode
             automatically with supported UPSs in emergency shutdowns.
```

dpinela:
Would you like me to reword this one like we did for the linux page (https://github.com/tldr-pages/tldr/pull/3401#discussion_r334254166), so they're consistent?

waldyrious:
> Would you like me to reword this one like we did for the linux page (#3401 (comment)), so they're consistent?

Yes, that would be appreciated. 👍

dpinela:
> > Would you like me to reword this one like we did for the linux page (#3401 (comment)), so they're consistent?
> 
> Yes, that would be appreciated. 👍

It is done.

agnivade:
This is probably the only PR with 5 approvals.

PR #3403, opened 2019-10-12
afplay: add page

dpinela:

### DIFF
--- a/pages/osx/afplay.md
+++ b/pages/osx/afplay.md
@@ -1,0 +1,19 @@
+# afplay
+
+> Command-line audio player.
+
+- Play a sound file (waits until playback ends):
+
+`afplay {{path/to/file}}`
+
+- Play a sound file at 2x speed:
+
+`afplay -r 2 {{path/to/file}}`
+
+- Play a sound file at half speed:
+
+`afplay -r 0.5 {{path/to/file}}`
+
+- Play the first N seconds of a sound file:
+
+`afplay -t {{seconds}} {{path/to/file}}`
--- a/pages/osx/afplay.md
+++ b/pages/osx/afplay.md
@@ -6,14 +6,14 @@
 
 `afplay {{path/to/file}}`
 
-- Play a sound file at 2x speed:
+- Play a sound file at 2x speed (playback rate):
 
-`afplay -r 2 {{path/to/file}}`
+`afplay --rate 2 {{path/to/file}}`
 
 - Play a sound file at half speed:
 
-`afplay -r 0.5 {{path/to/file}}`
+`afplay --rate 0.5 {{path/to/file}}`
 
 - Play the first N seconds of a sound file:
 
-`afplay -t {{seconds}} {{path/to/file}}`
+`afplay --time {{seconds}} {{path/to/file}}`


pages/osx/afplay.md
```diff
@@ -0,0 +1,19 @@
+# afplay
+
+> Command-line audio player.
+
+- Play a sound file (waits until playback ends):
+
+`afplay {{path/to/file}}`
+
+- Play a sound file at 2x speed:
```
waldyrious: Maybe mention "reproduction speed" or "rate", or some other word starting with "r", to help users develop mnemonics to learn the command.

pages/osx/afplay.md
```diff
@@ -0,0 +1,19 @@
...
+
+`afplay -r 2 {{path/to/file}}`
+
+- Play a sound file at half speed:
+
+`afplay -r 0.5 {{path/to/file}}`
+
+- Play the first N seconds of a sound file:
```
waldyrious: Do you know what the "t" stands for? Maybe we could incorporate that into the description.

dpinela: Going by the usage text, it seems to stand for "time":

```
  {-t | --time} TIME
    play for TIME seconds
```

I don't know how I'd rewrite the description to include the word "time" without sounding (no pun intended) contorted. "Play a sound file for a maximum amount of time" maybe? "Limit playback time"? I wrote it this way because it seems like the most straightforward to understand.

(The current wording already omits one detail: the time limit is applied after the rate adjustment, so if you play at 2x speed and use `-t N` it will still play sound for N seconds, but get twice as far along into the file. I did not mention it in the description because I couldn't come up with a succinct enough wording to explain that.)

pages/osx/afplay.md
dpinela: Interesting idea, but all of the alternatives I can think of sound worse to me. I don't think I've ever seen "reproduction" used to mean audio playback. "rate" makes sense semantically but I think "speed" will be easier to understand for a novice user.

waldyrious: Sure. You're correct that it's not always possible/desirable. One strategy we use often is parenthesis — so here it could be e.g. `at 2x speed (playback rate)`, or `at 2x the normal rate (speed)`, or some other variant. WDYT?

pages/osx/afplay.md
waldyrious: I see what you mean -- it's a common struggle we face as we balance succinctness, correctness and beginner-friendliness. I actually like something like `Play a sound file for a limited time` or `Play a sound file up to the given timestamp (in seconds)`, but I agree the current wording is simpler. Another strategy that is often handy is using the spelled-out version of the option, so in this case `--time`. Do you think any of these could improve the example?

(And yeah, don't worry about the nuance of this option's interaction with the rate one.)

pages/osx/afplay.md
dpinela: "playback rate" is almost there, though even that still sounds a bit weird to me. We could also use the longform `--rate` option, like you suggested below for `--time`.

pages/osx/afplay.md
dpinela: Using `--time` and keeping the description as-is sounds good to me.

pages/osx/afplay.md
agnivade: Given that the option itself is `rate`, not `speed`, I am fine with using it.

pages/osx/afplay.md
```diff
@@ -0,0 +1,19 @@
+# afplay
+
+> Command-line audio player.
```
einverne: Is there any links we can add here, any official documents?
You can add a new line here.

`> More information: <link here>.`

pxgamer: Apparently this is a built-in command, so I can't find any links other than the Man page.

pxgamer: Apparently this is a built-in command, so I can't find any links other than the Man page.

agnivade:
ping @waldyrious

pages/osx/afplay.md
einverne: Okay.

sbrl:
Merging, since @waldyrious' comments appear to have been addressed, and this has been sitting around for a while.

waldyrious:
Sorry guys for dropping the ball on this one (and thanks for picking it up!)

Nice work @dpinela! 🎉

pages/osx/afplay.md
waldyrious: For future reference, "playback rate" and the full option name were added in [`cd068a4`](https://github.com/tldr-pages/tldr/pull/3403/commits/cd068a4b2cc341d3d602050e40dbec482956d4d0).

pages/osx/afplay.md
waldyrious: For future reference, the expanded option name `--time` was added in [`cd068a4`](https://github.com/tldr-pages/tldr/pull/3403/commits/cd068a4b2cc341d3d602050e40dbec482956d4d0).

PR #3412, opened 2019-10-13
pr: add page

monst12:
New page for command `pr` of GNU Core Utilities as listed in #2213.
I would be very happy about feedback.

### DIFF
--- a/pages/common/pr.md
+++ b/pages/common/pr.md
@@ -1,0 +1,24 @@
+# pr
+
+> Paginate or columnate files for printing.
+> More information: <https://www.gnu.org/software/coreutils/manual/html_node/pr-invocation.html>.
+
+- Print multiple files with a default header and footer:
+
+`pr {{file1 file2 file3}}`
+
+- Print all files in parallel, one in each column without a header or footer:
+
+`pr -m -t {{file1 file2 file3}}`
+
+- Print all files beginning at 2 up to 5 with a given page length (including header and footer):
+
+`pr +{{2}}:{{5}} -l {{page_length}} {{file1 file2 file3}}`
+
+- Print all files with an offset each line and a truncating custom page width:
+
+`pr -o {{offset}} -W {{width}} {{file1 file2 file3}}`
+
+- Print files with numbered lines, a custom centered header and date format:
+
+`pr -n -h "{{header}}" -D "{{%Y.%m.%d - %H:%M:%S}}" {{file1 file2 file3}}`
--- a/pages/common/pr.md
+++ b/pages/common/pr.md
@@ -9,9 +9,9 @@
 
 - Print all files in parallel, one in each column without a header or footer:
 
-`pr -m -t {{file1 file2 file3}}`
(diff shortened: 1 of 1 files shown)


pages/common/pr.md
```diff
@@ -0,0 +1,24 @@
+# pr
+
+> Paginate or columnate files for printing.
+> More information: <https://www.gnu.org/software/coreutils/manual/html_node/pr-invocation.html>.
+
+- Print multiple files with a default header and footer:
+
+`pr {{file1 file2 file3}}`
+
+- Print all files in parallel, one in each column without a header or footer:
```
agnivade: parallel is technically the wrong word here. Perhaps better may be "Print all files together, .."

pages/common/pr.md
```diff
@@ -0,0 +1,24 @@
...
+
+`pr -m -T {{file1 file2 file3}}`
+
+- Print all files beginning at page 2 up to page 5 with a given page length (including header and footer):
+
+`pr +{{2}}:{{5}} -l {{page_length}} {{file1 file2 file3}}`
+
+- Print all files with an offset each line and a truncating custom page width:
```
agnivade: with an offset _for_ each line ..

And we can get rid of "all files". It is implicit after being mentioned 2 times above.

pages/common/pr.md
```diff
@@ -0,0 +1,24 @@
...
+
+`pr -o {{offset}} -W {{width}} {{file1 file2 file3}}`
+
+- Print files with numbered lines, a custom centered header and date format:
```
agnivade: I think we can split this into 2 separate examples - one for custom header, one for date format. And same applies as above: we can get rid of "files"

pages/common/pr.md
```diff
@@ -0,0 +1,28 @@
+# pr
+
+> Paginate or columnate files for printing.
+> More information: <https://www.gnu.org/software/coreutils/manual/html_node/pr-invocation.html>.
+
+- Print multiple files with a default header and footer:
+
+`pr {{file1 file2 file3}}`
```
sbrl: I'd be tempted to use `{{file1}} {{file2}} {{file3}}` instead of `{{file1 file2 file3}}` here.

schneiderl: Reads better and I think that's how we deal with this in the majority of the pages.

monst12: I used the tar page as reference for mutliple files input, but I agree that it would look nicer with separated tokens.

sbrl: Ah, interesting @monst12! Perhaps we should update the `tar` page then.

sbrl:
Are we good to merge, @monst12?

pages/common/pr.md
sbrl: Sure, go ahead! A PR would be great @monst12 :D

sbrl:
Merging, as @agnivade's comments appear to have been addressed, wee have a high volume of PRs at the moment, and this one has been sitting around for a while.

Issue #1109, opened 2016-10-12
Define the brand name of the project


zdroid:
I noticed many names are used for this project, so I think it should be clarified. For example, you can find _TL;DR pages_, _tldr_, _tldr pages_ and _TLDR pages_ in various documents and website.

My suggestion is to use and _tldr_ for the name of npm package and _TLDR pages_ or _tldr pages_ for everything else. How does that sound?

waldyrious:
I normally don't like when brands attempt to enforce specific capitalization styling of the brand name, so I'd preferif we avoided the capitalized version, and the pedantic version with the semicolon. I suppose I'd either go with **tldr pages** or **tldr-pages**, but I'm open to hear other arguments.

zdroid:
Agree @waldyrious, I like that version more. *TLDR pages* sounds a bit too administrative. *tldr-pages* sounds cool in dev matter of cool.

ostera:
`tldr` has my vote.

agnivade:
I like **tldr pages**. Contrasts nicely with **man pages**.

waldyrious:
If we are willing to deviate a little more from the current name, we could consider rebranding as tldr.sh (which is akin to how many projects nowadays now use .js to indicate their context), or even go as far as ditching the hard-to-pronounce (and somewhat jargon-y) "tldr" altogether and pick something like "miniman" or "lil' man" :P just some fresh/wild (crazy?) ideas to spice up the discussion, not something I'm advocating at the moment.

Getting back down to earth, I suspect that "tldr-pages" may be better than "tldr pages" as it clearly identifies as a single word, in the absence of capitalization or formatting. But it's not a super-strong preference.

zdroid:
I agree with @waldyrious. `tldr-pages` seems the best at this moment.

zdroid:
Any news on this? Branding is an important part of the project.

waldyrious:
I'd like to hear @agnivade and @ostera's responses to my previous comment.

ostera:
I reeeeally need to setup tldr.sh

agnivade:
My preference is still "tldr pages", but I am okay with "tldr.sh" too.

waldyrious:
If tldr.sh is ok with you guys, I'd vote for it too.

[...]

PR #3413, opened 2019-10-13
Fixed wording a bit

SimonVerhoeven:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,8 +20,8 @@
 [license-url]: https://github.com/tldr-pages/tldr/blob/master/LICENSE.md
 [license-image]: https://img.shields.io/github/license/tldr-pages/tldr.svg
 
-Contributions to the tldr-pages project are [most welcome](GOVERNANCE.md)!
-All `tldr` pages are stored in Markdown right here on GitHub.
+Contributions to the TLDR-pages project are [most welcome](GOVERNANCE.md)!
+All `TLDR` pages are stored in Markdown right here on GitHub.
 Just open an issue or send a pull request and we'll incorporate it as soon as possible.
 To get started, please [sign](https://cla-assistant.io/tldr-pages/tldr) the
 [Contributor License Agreement](https://gist.github.com/waldyrious/e50feec13683e565769fbd58ce503d4e).
@@ -30,14 +30,14 @@ To get started, please [sign](https://cla-assistant.io/tldr-pages/tldr) the
 
 ## Guidelines
 
-The basic format of a `tldr` page is a set of concrete usage examples.
+The basic format of a `TLDR` page is a set of concrete usage examples.
 Here are a few guidelines to get started:
 
 1. Try to keep pages at around 5 examples. Pages can be longer if needed, but don't exceed 8 examples.
    Remember, it's OK if the page doesn't cover everything; that's what `man` is for.
 2. When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness.
    For example, commands that require `sudo` should include it directly in the examples.
-3. Try to incorporate the spelled-out version of single-letter options in the example's description.
+3. Try to incorporate the spelt-out version of single-letter options in the example's description.
    The goal is to allow people to *understand* the syntax of the commands, not just *memorize* it.
 4. Introduce options gradually, starting with the simplest command invocations,
    and using more complex examples progressively.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -20,8 +20,8 @@
 [license-url]: https://github.com/tldr-pages/tldr/blob/master/LICENSE.md
 [license-image]: https://img.shields.io/github/license/tldr-pages/tldr.svg
 
-Contributions to the tldr-pages project are [most welcome](GOVERNANCE.md)!
-All `tldr` pages are stored in Markdown right here on GitHub.
+Contributions to the TLDR-pages project are [most welcome](GOVERNANCE.md)!
```
schneiderl: I'm unsure about this. 
Should we have this in uppercase? @sbrl @mebeim @waldyrious

mebeim: Not really sure.. but this suggested change is inconsistent: it only changes some occurrences of the word, not all of them. :thinking:. I don't personally have a preference between the two by the way.

schneiderl: I think I have an opinion now.
"TLDR" is the term while "tldr" is the project name. When we're talking about the project, I believe we should keep it lower case.

mebeim: That could be a good point.

sbrl: I believe that a while ago it was decided that `tldr-pages` is the official wording for referring to the tldr-pages project.

CONTRIBUTING.md
```diff
@@ -30,14 +30,14 @@ To get started, please [sign](https://cla-assistant.io/tldr-pages/tldr) the
 
 ## Guidelines
 
-The basic format of a `tldr` page is a set of concrete usage examples.
+The basic format of a `TLDR` page is a set of concrete usage examples.
```
sbrl: Again here. `tldr` should stay as `tldr`, not `TLDR`.

CONTRIBUTING.md
```diff
@@ -91,7 +91,7 @@ Some examples of valid locale tags:
 
 ### Default language for newly added pages
 
-The default language used for pages is English (US). Pages written in English are stored in the default `pages` directory (notice the absence of a specific language tag). Although not strictly required, if you'd like to add a new page in a different language, please consider creating the English page too.
```
sbrl: I'm not sure this change makes it clearer. The `used` is important, as it indicates that the language is used on a pages.

CONTRIBUTING.md
```diff
@@ -116,7 +116,7 @@ Examples:
 
 ## Licensing
 
-`tldr` is licensed under the [MIT license](https://github.com/tldr-pages/tldr/blob/master/LICENSE.md).
+`TLDR` is licensed under the [MIT license](https://github.com/tldr-pages/tldr/blob/master/LICENSE.md).
```
sbrl: And again here. This should stay as-is.

CONTRIBUTING.md
```diff
@@ -30,14 +30,14 @@ To get started, please [sign](https://cla-assistant.io/tldr-pages/tldr) the
...
-The basic format of a `tldr` page is a set of concrete usage examples.
+The basic format of a `TLDR` page is a set of concrete usage examples.
 Here are a few guidelines to get started:
 
 1. Try to keep pages at around 5 examples. Pages can be longer if needed, but don't exceed 8 examples.
    Remember, it's OK if the page doesn't cover everything; that's what `man` is for.
 2. When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness.
    For example, commands that require `sudo` should include it directly in the examples.
-3. Try to incorporate the spelled-out version of single-letter options in the example's description.
+3. Try to incorporate the spelt-out version of single-letter options in the example's description.
```
sbrl: Nice catch.

agnivade: Except this change, let's revert all the others.

CONTRIBUTING.md
waldyrious: > I believe that a while ago it was decided that tldr-pages is the official wording for referring to the tldr-pages project.

IIRC we didn't end up making a final decision. We just recently restarted discussion in #1109, but let me know if there are other discussions I'm missing.

In any case, I'm pretty sure all-caps TLDR was never widely supported in such discussions, so this particular bit should be reverted.

CONTRIBUTING.md
waldyrious: It's not actually incorrect, it's just [one of the valid variants](https://en.wiktionary.org/wiki/spelled) for the past participle of "spell".

waldyrious:
Considering the comments left inline, it looks like none of the proposed changes are mergeable. Thanks for taking the time to contribute, @SimonVerhoeven, but regrettably we'll have to close the PR unmerged.

PR #3423, opened 2019-10-15
Refactor: code clean up

tatiana-scda:
replace temporary variable with query and replace error code with exception

### DIFF
--- a/scripts/pdf/render.py
+++ b/scripts/pdf/render.py
@@ -22,8 +22,7 @@ def main(loc, colorscheme):
 
     #Checking correctness of path
     if not os.path.isdir(loc):
-        print("Invalid directory. Please try again!", file = sys.stderr)
-        sys.exit(1)
+        raise ValueError("Invalid directory. Please try again!", file=sys.stderr)
 
     #Writing names of all directories inside 'pages' to a list
     for os_dir in os.listdir(loc):
--- a/scripts/send_to_bot.py
+++ b/scripts/send_to_bot.py
@@ -18,11 +18,6 @@ def post_comment(pr_id, comment_body):
   if f.getcode() != 200:
     print(f.read())
 
-
-# Get the environment variables
-PR_NUMBER = os.environ.get('TRAVIS_PULL_REQUEST')
-BUILD_ID = os.environ.get('TRAVIS_BUILD_ID')
-
 # Read the test result output from stdin
 test_result = sys.stdin.read().strip()
 # Populate the template text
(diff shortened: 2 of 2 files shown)


scripts/send_to_bot.py
```diff
@@ -18,11 +18,6 @@ def post_comment(pr_id, comment_body):
   if f.getcode() != 200:
     print(f.read())
 
-
-# Get the environment variables
-PR_NUMBER = os.environ.get('TRAVIS_PULL_REQUEST')
-BUILD_ID = os.environ.get('TRAVIS_BUILD_ID')
-
```
sbrl: I'm not sure that moving these to be inline is a good idea. The general idea here is to separate concerns. If in the future we wanted to change the way we got those values, or needed them in more places than one, you'd have a problem if they weren't separated.

Additionally, it's useful for reference to have them all in one place.

scripts/send_to_bot.py
```diff
@@ -34,8 +29,8 @@ def post_comment(pr_id, comment_body):
...
 "\n```\n"
 "Please fix the error(s) and push again."
-).format(build_id=BUILD_ID, comment_body=test_result)
+).format(build_id=os.environ.get('TRAVIS_BUILD_ID'), comment_body=test_result)
 
 # If it's a PR, post a comment on it
-if PR_NUMBER != "false":
-  post_comment(PR_NUMBER, comment)
+if os.environ.get('TRAVIS_PULL_REQUEST') != "false":
+  post_comment(os.environ.get('TRAVIS_PULL_REQUEST'), comment)
```
sbrl: Save as above. Here we're duplicating effort, and if we need to change the variable name later it's more effort to change.

scripts/pdf/render.py
```diff
@@ -22,8 +22,7 @@ def main(loc, colorscheme):
 
     #Checking correctness of path
     if not os.path.isdir(loc):
-        print("Invalid directory. Please try again!", file = sys.stderr)
-        sys.exit(1)
+        raise ValueError("Invalid directory. Please try again!", file=sys.stderr)
```
sbrl: Not sure here. This might be semantically more correct, but by doing this we generate a Python stack trace, and the output is much more complicated - leading it to be less user-friendly.

agnivade:
Agree with @sbrl's observations. This doesn't seem worth changing. The `ValueError` change is still debatable, but it's a one-off script used rarely. Showing a one-line error is fine.

PR #3435, opened 2019-10-17
tmux: remove named sessions and add ways to split pane into 2 terminals

robbyrenz:

### DIFF
--- a/pages/common/tmux.md
+++ b/pages/common/tmux.md
@@ -7,9 +7,13 @@
 
 `tmux`
 
-- Start a new named tmux session:
+- Horizontally split the current pane into two terminals:
 
-`tmux new -s {{name}}`
+`Ctrl + B, "`
+
+- Vertically split the current pane into two terminals:
+
+`Ctrl + B, %`
 
 - List sessions:
 
@@ -19,17 +23,13 @@
 
 `tmux a`
 
-- Attach to a named session:
-
-`tmux a -t {{name}}`
-
 - Detach from session:
 
 `Ctrl + B, D`
 
 - Kill session:
 
-`tmux kill-session -t {{name}}`
+`tmux kill-session`
 
 - Kill session when attached:
 


pages/common/tmux.md
```diff
@@ -19,17 +23,13 @@
...
-`tmux a -t {{name}}`
-
 - Detach from session:
 
 `Ctrl + B, D`
 
 - Kill session:
 
-`tmux kill-session -t {{name}}`
+`tmux kill-session`
```
einverne: hi @robbyrenz , how about change the command description to `Kill the given session` while not change the command?

pages/common/tmux.md
```diff
@@ -19,17 +23,13 @@
 
 `tmux a`
 
-- Attach to a named session:
-
-`tmux a -t {{name}}`
-
```
einverne: I think this command should keep. `tmux a -t {{name}}` is a quite useful command in daily use.

sbrl: Seconded. I think we really ought to keep the `Start a new named tmux session` one too.

pages/common/tmux.md
robbyrenz: Okay, sure, I'll make the changes, thanks for the feedback.

pages/common/tmux.md
robbyrenz: Okay, thanks for the feedback. Unfortunately, if I do retain the above two commands, I will not be able to add any more of my examples as this file has already exceeded the limit of 8 examples. Is it okay if I still add my examples as I believe that knowing how to split the pane into two terminals is a vital skill to know.

sbrl: Hrm, I see. That is an awkward one. The thing is that we generally try to keep pages shorter, because they aren't meant to be an _exhaustive_ list of commands etc.

A keyboard shortcut sounds like the kind of thing that would fit well on a cheat sheet.

robbyrenz: Yes, I understand, so I guess I should just close the pull request?

agnivade: Yes

PR #3442, opened 2019-10-18
Formatting Chinese translation

### LINKED ISSUE #3426: Formatting quality of Chinese translation
This is an issue but also a reminder for Chinese translators, that many items come with low formatting quality. Note that this issue is not about the translation quality but only about the copywriting in Chinese.

For example:

1. Whether there should be a space between English words and Chinese characters. (Right answer: yes)

2. Which symbols should be used in a Chinese sentence, full-width or half-width. (Right answer: only use half-width in commands or full English sentences or terms, otherwise use full-width)

In current translations, it looks like there is no rule about copywriting at all and thus different symbols are used in one file mistakenly. Here are some examples:

1. 
Wrong:
`Mas是一个简单的 Mac App Store 命令行界面.`

Correct:
`Mas 是一个简单的 Mac App Store 命令行界面。`

Added the space between "Mas" and "是"; Changed "." to "。".

2. 
Wrong:
`显示有关指定软件的信息（版本，安装路径，依赖关系等):`

Correct:
`显示有关指定软件的信息（版本，安装路径，依赖关系等）：`

Changed the bracket and colon to full width.

I recommend you follow a copywriting standard, like this one https://github.com/sparanoid/chinese-copywriting-guidelines, to unify the format.

einverne:
Try to fix #3426

I am using Pangu to format all translations

    vim *.md
    
and then

    :argdo Pangu | update
    :wq

Then change all period to `.`

    find . -type -f -exec sed -i "s/。/./g" {} \;


einverne:
Tried to update trailing colon with:

    find . -type f -exec sed -i "s/：$/:/g" {} \;

mebeim:
After looking more carefully I spotted some more errors:

1. I see you replaced ASCII parentheses (`(` and `)`) with their full-width Unicode alternative (`（` and `）`). On some lines though, I see that only the opening parentheses are replaced, but not the closing ones, like for example in `pages.zh/common/wget.md` ([line 7](https://github.com/tldr-pages/tldr/pull/3442/files#diff-516fbe9c1922dfaa6254053c325a649fR7)). This should be fixed.

2. I also see that *not all* the parentheses are replaced, like for example in `pages.zh/common/vim.md` ([line 16](https://github.com/tldr-pages/tldr/pull/3442/files#diff-b37f2306c16016e14d4b15c304c54562R16)). Seems that every occurrence of `):` has not been replaced. Is this intended? If not, this should also be fixed.

einverne:
Oh, thank you for pointing out. It may be the problem of format script. Let me modify these missing mistakes.

mebeim:
Hey guys, it's been a while since the last comment, it seems that the mentioned people are currently unavailable. I would get this merged though, it looks fine hand hasn't much to do with Chinese translation/spelling and similar, just simple punctuation and spacing fixes.

PR #3453, opened 2019-10-21
wine: add page

icsd12015:

### DIFF
--- a/pages/linux/wine.md
+++ b/pages/linux/wine.md
@@ -1,6 +1,6 @@
 # wine
 
-> Run Windows programs on Unix
+> Run Windows programs on Unix.
 > More information: <https://wiki.winehq.org/List_of_Commands>.
 
 - Run a program with args:
@@ -27,6 +27,6 @@
 
 `wine msiexec /i {{package}}`
 
-- Run GUI configuration tool for Wine
+- Run GUI configuration tool for Wine:
 
 `wine winecfg`
--- a/pages/linux/wine.md
+++ b/pages/linux/wine.md
@@ -5,11 +5,11 @@
 
 - Run ipconfig.exe program:
 
-`wine ipconfig /all
+`wine ipconfig /all`
 
 - Run cmd.exe in background:
 
-`wine start cmd 
+`wine start cmd`
 
 - Run Windows-like Package Manager:
 
--- a/pages/linux/wine.md
+++ b/pages/linux/wine.md
@@ -1,0 +1,32 @@
+# wine
+
+> Run Windows programs on Unix
+> More information: <https://wiki.winehq.org/List_of_Commands>.
+
+- Run a program with args:
+
+`wine {{program}} {{program args}}`
+
+- Start a program:
+
+`wine start {{program}} {{program args}}`
+
+- Run Windows command line:
+
+`wine cmd`
+
+- Run Windows-like Task Manager:
+
+`wine taskmgr`
+
+- Run Windows-like Package Manager:
+
+`wine uninstaller`
+
+- Install MSI packages:
+
+`wine msiexec /i {{package}}`
+
+- Run GUI configuration tool for Wine
+
+`wine winecfg`
--- a/pages/linux/wine.md
+++ b/pages/linux/wine.md
@@ -3,13 +3,13 @@
 > Run Windows programs on Unix.
 > More information: <https://wiki.winehq.org/>.
 
-- Run a program with args:
+- Run ipconfig.exe program:
 
-`wine {{program}} {{program args}}`
+`wine ipconfig /all
 
-- Run a program with args in background:
+- Run cmd.exe in background:
 
-`wine start {{program}} {{program args}}`
+`wine start cmd 
 
 - Run Windows-like Package Manager:
 
--- a/pages/linux/wine.md
+++ b/pages/linux/wine.md
@@ -1,7 +1,7 @@
 # wine
 
 > Run Windows programs on Unix.
-> More information: <https://wiki.winehq.org/List_of_Commands>.
+> More information: <https://www.winehq.org/>.
 
 - Run a program with args:
 
--- a/pages/linux/wine.md
+++ b/pages/linux/wine.md
@@ -7,17 +7,9 @@
 
 `wine {{program}} {{program args}}`
 
-- Start a program:
+- Run a program with args in background:
 
 `wine start {{program}} {{program args}}`
-
-- Run Windows command line:
-
-`wine cmd`
-
-- Run Windows-like Task Manager:
-
-`wine taskmgr`
 
 - Run Windows-like Package Manager:
 
@@ -26,7 +18,3 @@
 - Install MSI packages:
 
 `wine msiexec /i {{package}}`
-
-- Run GUI configuration tool for Wine:
-
-`wine winecfg`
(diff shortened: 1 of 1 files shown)


pages/linux/wine.md
```diff
@@ -0,0 +1,32 @@
+# wine
+
+> Run Windows programs on Unix.
+> More information: <https://wiki.winehq.org/List_of_Commands>.
```
schneiderl: Is this the best link for us to use? 
It doesn't seem to be directly talkin about Wine. It actually talks about wine subcommands. 
Maybe we can use the official documentation 
```suggestion
> More information: <https://www.winehq.org/documentation>.
```

or the main page

```suggestion
> More information: <https://www.winehq.org/>.
```

pages/linux/wine.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Run a program with args:
+
+`wine {{program}} {{program args}}`
+
+- Start a program:
```
agnivade: I guess the difference between start and run is that start spawns a process in the background ? We should clarify that, because otherwise it is not clear.

pages/linux/wine.md
```diff
@@ -0,0 +1,32 @@
+# wine
+
+> Run Windows programs on Unix.
+> More information: <https://wiki.winehq.org/List_of_Commands>.
+
+- Run a program with args:
+
+`wine {{program}} {{program args}}`
```
agnivade: I think we should have the msiexec example here. And remove all the examples for cmd, taskmgr. All of them are the same repetition of a program with/without args. Typically, we do not use extra examples (we only have a space of 8!) to show the same thing.

C4vla: I included those 4 sub-programs (cmd, taskmgr, msiexec and uninstaller), because i believe they are the most important, the most helpful of all. I believe that with those examples, the user has a good knowledge, for a basic usage of the tool. At least from my experience with the tool, those are all i ever need to remember, for anything i wanted to do. Although, if you find the page excessive, i agree that cmd, taskmgr  and winecfg can be omitted. But i believe uninstaller should stay, because  the cmd and tasmgr are referring to the real Windows cmd.exe and taskmgr.exe under System32 but the "Add or remove programs wizard" utility is appwiz.cpl so its a bit confusing as it is replaced with the uninstaller designation in wine.

icsd12015: included those 4 sub-programs (cmd, taskmgr, msiexec and uninstaller), because i believe they are the most important, the most helpful of all. I believe that with those examples, the user has a good knowledge, for a basic usage of the tool. At least from my experience with the tool, those are all i ever need to remember, for anything i wanted to do. Although, if you find the page excessive, i agree that cmd, taskmgr and winecfg can be omitted. But i believe uninstaller should stay, because the cmd and tasmgr are referring to the real Windows cmd.exe and taskmgr.exe under System32 but the "Add or remove programs wizard" utility is appwiz.cpl so its a bit confusing as it is replaced with the uninstaller designation in wine.

pages/linux/wine.md
icsd12015: I agree with your note. I will update the page accordingly.

pages/linux/wine.md
icsd12015: I agree. the main page is more appropriate reference for help with the tool. I will update the page.

pages/linux/wine.md
agnivade: > Although, if you find the page excessive, i agree that cmd, taskmgr and winecfg can be omitted.

Sure let's do that.

And can we also use `cmd` with some args as an example for "Run a program with args" ? That should remove one more example. Then I am good with this.

icsd12015: Oh yes, good suggestion.

agnivade: @icsd12015 - Not sure if you have fixed this or yet to do it, but I don't see the change that I requested. Please let me know if you need clarification.

C4vla: I thought that my last [commit ](https://github.com/tldr-pages/tldr/blob/cef11004eaa32e3cf98ee17871a0de292be4b295/pages/linux/wine.md) fulfilled your request.  If you not please do specify why.

sbrl:
Thanks for the approval, @C4vla! Unfortunately, it won't count against the 2 approvals required, as you need to have contributor privileges or better for it to count. As [our roles document outlines](https://github.com/tldr-pages/tldr/blob/master/COMMUNITY-ROLES.md) however, once you have submitted 5 non-trivial pull requests you are eligible to be made a contributor :-)

pages/linux/wine.md
agnivade: Sure, I wanted:

> And can we also use cmd with some args as an example for "Run a program with args" ?

I still see this: `wine {{program}} {{program args}}`

sbrl:
t

schneiderl:
ping @agnivade

zdroid:
@sbrl @schneiderl The problem presented by @agnivade is that there are no examples of arguments specified. Any idea how to fix that? (It also makes this PR mergable)

schneiderl:
@zdroid Well, it seems to have a more "generic" approach in this case. Calling it "program" and "program args" makes it more wide open but I don't think that is how we write pages here in tldr. Maybe changing "program" to "cmd" and relevant args would be more concise with the pages that were previously written in tldr.

sbrl:
I'd suggest `command` and `arguments`.

agnivade:
I just wanted the first example to be wine {{cmd}} {{some args}}.

C4vla:
Hi all,
sorry for my absence i was on a trip with internet access only via phone and i didn't check on Github during this time.
@agnivade i understand your request and i agree that example of a command with args should be added to the page.
What do you think about adding also a reference to the host file system which is mounted by default on Z:\ ?  I'm thinking something like `wine 'Z:/usr/share/windows-resources/binaries/nc.exe' '-lvp 4444'`
alternatively `ipconfig.exe` of System32 could be used as an example like this  `wine ipconfig /all` or this `wine ipconfig.exe /all`. 
Please advice if you have a better idea for this and about tell me what you think of my options in order to make the preferred changes.

agnivade:
I don't think there is any need to complicate this any more than needed. I would have preferred something with `cmd`, but `ipconfig /all` is fine too. Also, if you can think of a simple background command, then we can use that as the second example too.

icsd12015:
I added the ipconfig example and for start sub command i added cmd which pops a new cmd windows. Althought i haven't think yet of any (common) program for demonstration to start in background with command line args :/

pages/linux/wine.md
```diff
@@ -3,13 +3,13 @@
 > Run Windows programs on Unix.
 > More information: <https://wiki.winehq.org/>.
 
-- Run a program with args:
+- Run ipconfig.exe program:
```
zdroid: I think what others meant is something like this:

```
Run a program with args:
wine {{ipconfig}} {{/all}}
```

agnivade: Yes, that's what I mentioned here: https://github.com/tldr-pages/tldr/pull/3453#issuecomment-565712705. Sorry if that was confusing.

icsd12015: Oh yes my mistake!  I corrected my syntax

pages/linux/wine.md
icsd12015: Sorry i made the correct pr about this issue. I didn't understand it at first.

PR #3458, opened 2019-10-21
qmv, qcp: add page

neizod:

Personal note:
- I add 2 commands at once because the interface is quite the similar, and they comes with `renameutils` package.
- `qmv` has 5 examples, while `qcp` has only 3 since the ability to copy directories do not work now.
- What I'm not sure the most is do these 2 commands should go to `linux/` instead of `common/`? On my Ubuntu I can just run `apt install renameutils` and done, but I don't know about OS X or Windows (or others). I see that there is a source code but I don't have other platform to compile and try it.

### DIFF
--- a/pages/common/qcp.md
+++ b/pages/common/qcp.md
@@ -1,0 +1,16 @@
+# qcp
+
+> Copy files with your favorite text editor.
+> More information: <https://www.nongnu.org/renameutils/>
+
+- Copy a single file (invoke an editor with source on the left, target on the right):
+
+`qcp {{source_file}}`
+
+- Copy multiple JPG files:
+
+`qcp {{*.jpg}}`
+
+- Copy files, but swap pattern in the editor -- target on the left and source on the right:
+
+`qcp --option swap {{IMG-*.jpg}}`
--- a/pages/common/qcp.md
+++ b/pages/common/qcp.md
@@ -1,7 +1,7 @@
 # qcp
 
 > Copy files with your favorite text editor.
-> More information: <https://www.nongnu.org/renameutils/>
+> More information: <https://www.nongnu.org/renameutils/>.
 
 - Copy a single file (invoke an editor with source on the left, target on the right):
 
--- a/pages/common/qcp.md
+++ b/pages/common/qcp.md
@@ -11,6 +11,6 @@
 
 `qcp {{*.jpg}}`
 
-- Copy files, but swap pattern in the editor -- target on the left and source on the right:
+- Copy files, but swap the positions of the source and the target in the editor:
 
 `qcp --option swap {{*.jpg}}`
--- a/pages/common/qcp.md
+++ b/pages/common/qcp.md
@@ -13,4 +13,4 @@
 
 - Copy files, but swap pattern in the editor -- target on the left and source on the right:
 
-`qcp --option swap {{IMG-*.jpg}}`
+`qcp --option swap {{*.jpg}}`
--- a/pages/common/qmv.md
+++ b/pages/common/qmv.md
@@ -11,13 +11,13 @@
 
 `qmv {{*.jpg}}`
 
-- Move directories:
+- Move 3 directories:
 
-`qmv -d {{directory/}} {{another_directory/}} {{the_other_directory/}}`
+`qmv -d {{path/to/dir_1}} {{path/to/dir_2}} {{path/to/dir_3}}`
 
 - Move files/directories inside a directory:
 
-`qmv -R {{directory/}}`
+`qmv -R {{path/to/directory}}`
 
 - Move files, but swap pattern in the editor -- target on the left and source on the right:
 
--- a/pages/common/qmv.md
+++ b/pages/common/qmv.md
@@ -13,11 +13,11 @@
 
 - Move directories:
 
-`qmv --directory {{my_directory/}} {{another_directory/}} {{the_other_directory/}}`
+`qmv -d {{directory/}} {{another_directory/}} {{the_other_directory/}}`
 
 - Move files/directories inside a directory:
 
-`qmv --recursive {{my_directory/}}`
+`qmv -R {{directory/}}`
 
 - Move files, but swap pattern in the editor -- target on the left and source on the right:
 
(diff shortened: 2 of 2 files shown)


pages/common/qmv.md
```diff
@@ -0,0 +1,24 @@
...
+
+`qmv {{*.jpg}}`
+
+- Move directories:
+
+`qmv --directory {{my_directory/}} {{another_directory/}} {{the_other_directory/}}`
+
+- Move files/directories inside a directory:
+
+`qmv --recursive {{my_directory/}}`
```
einverne: How about `qmv -R {{directory}}`?

pages/common/qcp.md
```diff
@@ -0,0 +1,16 @@
...
+
+`qcp {{source_file}}`
+
+- Copy multiple JPG files:
+
+`qcp {{*.jpg}}`
+
+- Copy files, but swap pattern in the editor -- target on the left and source on the right:
+
+`qcp --option swap {{IMG-*.jpg}}`
```
einverne: How about remove the preceding `IMG-`, I think `IMG-` does not need here.

neizod: My bad here, I just happen to remove `IMG-` from `qmv.md`, but forget to remove it from `qcp.md`. Already fix in f9224f2409e35677935e22ed0d19d666a2b182af.

pages/common/qmv.md
```diff
@@ -0,0 +1,24 @@
...
+
+`qmv {{source_file}}`
+
+- Move multiple JPG files:
+
+`qmv {{*.jpg}}`
+
+- Move directories:
+
+`qmv --directory {{my_directory/}} {{another_directory/}} {{the_other_directory/}}`
```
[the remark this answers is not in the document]
neizod: I personally like short option too! But since I never contribute to TLDR before, I read the guild line and see that long option is more preferable. So I PR the long option.

I've change it to short option in 166264880efc5de6535bc741e6df7f4f10e5020f since I recall that `ls -d` also work the same way (and the short option is already de facto standard in this case, I think).

I also change `my_directory` to just `directory`.

pages/common/qmv.md
neizod: Same as above, I already use short option in 166264880efc5de6535bc741e6df7f4f10e5020f.

By the way, do I need to remove trailing slash after directory name?

neizod:
Hi @eiverne, thanks for comment. I have review your comments and push new commits.

pages/common/qmv.md
einverne: I have tried the command, it seems with or without trailing slash, the result is same. So let's remove it? Or let more people to review this command.

neizod: Trailing slash is just my personal style (since I observe for a long time that almost all command treat directory with or without argument the same way). I put it there to make it obvious that the argument is a directory.

neizod:
Today I test these 2 commands on other platform on my friends machines.

- __OS X__: I can `brew install renameutils` and then `export EDITOR=vi` and it just work like Linux.
- __Windows 10__: I try to build it with Visual Studio C++, but I fail to locate the right workloads (download all VS workloads would require ~20GB space on HDD). After unsuccessfully building for an hour or two, I gave up on this one.

agnivade:
That's fine, as long as it works on both osx and linux, it can be in common.

pages/common/qcp.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Copy a single file (invoke an editor with source on the left, target on the right):
+
+`qcp {{source_file}}`
+
+- Copy multiple JPG files:
+
+`qcp {{*.jpg}}`
+
+- Copy files, but swap pattern in the editor -- target on the left and source on the right:
```
agnivade: @waldyrious @sbrl - Is "--" the right way to express a pause in a sentence ?

pages/common/qmv.md
```diff
@@ -0,0 +1,24 @@
...
+
+`qmv {{source_file}}`
+
+- Move multiple JPG files:
+
+`qmv {{*.jpg}}`
+
+- Move directories:
+
+`qmv -d {{directory/}} {{another_directory/}} {{the_other_directory/}}`
```
agnivade: path/to/directory, path/to/another_directory, path/to/target_directory

pages/common/qmv.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Move files/directories inside a directory:
+
+`qmv -R {{directory/}}`
```
agnivade: path/to/directory

neizod: also fix in bb2f8e0566cad4f47c037ca682480b136aa70ded.

pages/common/qmv.md
neizod: There are no target directory there, if the command get 3 arguments, then the 3 of them will be move (telling targets in the editor later).

But I get the idea, fix in bb2f8e0566cad4f47c037ca682480b136aa70ded.

pages/common/qcp.md
neizod: Or should I rephrase the whole sentence, like this:

    - Copy files, but swap position of source and target in the editor:

neizod:
Hi @agnivade, I have resolve some of your reviews. I left the last one waiting for more comments since I'm not a native speaker.

pages/common/qcp.md
sbrl: Sort of, @agnivade. Personally I'd use a single dash ` - ` there, not a double one - or, of course, ....`editor, with the target`..... instead.

neizod: @sbrl thanks for correcting them.

pages/common/qmv.md
sbrl: Ah, I think @agnivade is asking where do the directories get moved to, @neizod?

sbrl:
Also, no need to force-push @neizod. We've got a special button for that when merging, and it helps us to see what's changed when reviewing better :-)

agnivade:
Yep

PR #3464, opened 2019-10-22
ceph: add page

amitech:

### DIFF
--- a/pages/common/ceph.md
+++ b/pages/common/ceph.md
@@ -1,0 +1,19 @@
+# ceph
+
+> Object, block and file-based storage platform.
+
+- Check cluster health:
+
+`ceph status`
+
+- Check cluster usage stats:
+
+`ceph df`
+
+- Create a storage pool:
+
+`ceph osd pool create`
+
+- Delete a storage pool:
+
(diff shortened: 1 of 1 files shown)


pages/common/ceph.md
```diff
@@ -0,0 +1,19 @@
...
+
+`ceph status`
+
+- Check cluster usage stats:
+
+`ceph df`
+
+- Create a storage pool:
+
+`ceph osd pool create`
```
Syphdias: I'm pretty sure this will fail without a pool name

pages/common/ceph.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Delete a storage pool:
+
+`ceph osd pool delete`
```
Syphdias: As with `ceph osd pool create` this needs an argument. And there are at least two other hoops to jump though since this a a dangerous action is usually not daily business.
This is the reason why I think you should replace it with something else. This is not something you need to know when you start off with `ceph`. I'd suggest something related to `ceph auth` instead.

amitech:
I will create a PR for moving this page to common.

PR #3472, opened 2019-10-24
move ceph.md from linux to common pages

amitech:


andrik:
Hi @amitech 

Thank you for your PR

I looked the Ceph documentation and it look likes the main focus of the project is on Linux. I found a [thread](https://forums.freebsd.org/threads/ceph-on-freebsd.67673/) on FreeBSD forum saying that the client is well supported but not the server. So I'm not sure if it is right moment to move the command to common directory. Let's see what others think about it.

sbrl:
I'm not sure we have a platform for bsd yet at this time, though adding one is as trivial as creating a directory and writing a page.

If `ceph` is supported on more than one platform, then it should probably go in `common`. if it's supported on another platform we don't yet have a directory for, I'd feel more comfortable if we had at least 1 page for that platform before putting this in `common`.

agnivade:
If it does not work on osx, then there's no point of moving this.

amitech:
I am closing it. If I see it is working for multiple platforms then can re-open.

PR #3475, opened 2019-10-25
rpcinfo: add page

pxgamer:

### DIFF
--- a/pages/windows/rpcinfo.md
+++ b/pages/windows/rpcinfo.md
@@ -1,0 +1,20 @@
+# rpcinfo
+
+> List programs via RPC on remote computers.
+> More information: <https://docs.microsoft.com/windows-server/administration/windows-commands/rpcinfo>.
+
+- List all programs registered on the local computer:
+
+`rpcinfo`
+
+- List all programs registered on a remote computer:
+
+`rpcinfo /p {{computer_name}}`
+
+- Call a specific program on a remote computer using TCP:
+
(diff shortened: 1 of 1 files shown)


pages/windows/rpcinfo.md
```diff
@@ -0,0 +1,20 @@
...
+
+- List all programs registered on the local computer:
+
+`rpcinfo`
+
+- List all programs registered on a remote computer:
+
+`rpcinfo /p {{computer_name}}`
+
+- Call a specific program on a remote computer using TCP:
```
agnivade: What's the default protocol ? We should mention that in parentheses.

mebeim: No default as far as I can tell from the Microsoft doc.

agnivade: Well, there has to be _some_ default protocol which is used. Strange that it's not mentioned in the docs.

pxgamer: I can't find any information on this at all, so just installing it to test. 👍

mebeim: @agnivade the point is that from the doc it seems that you cannot connect to a remote computer if you don't specify the protocol. If you look at the doc, you can see `[/t <Node Program> [<version>]]`, there is no `<Node Program>` outside square brackets that are already including `/t` or `/u`.

agnivade: Oh I see ! Okay with me then

PR #3498, opened 2019-10-29
docker-compose: add --rmi all --volumes

ninearif:
add "--rmi all --volumes" parameters to docker-compose down command, in order to match with description (by default, docker-compose didn't remove volumes).

- [/ ] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [ /] The page has 8 or fewer examples.
- [ /] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [ /] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [ /] The page description includes a link to documentation or a homepage (if applicable).

### DIFF
--- a/pages/common/docker-compose.md
+++ b/pages/common/docker-compose.md
@@ -21,7 +21,7 @@
 
 - Stop and remove all containers, networks, images, and volumes:
 
-`docker-compose down`
+`docker-compose down --rmi all --volumes`
 
 - Follow logs for all containers:
 


Aracki:
What about only `docker-compose down` method?

ninearif:
> What about only `docker-compose down` method?

By default, the only things removed are:

- **Containers** for services defined in the Compose file
- **Networks** defined in the networks section of the Compose file
- The default network, if one is used

Networks and volumes defined as **external** are _never removed_.
According to : https://docs.docker.com/compose/reference/down/

So you need to put **-v** or **--volumes** to tell docker to remove volumes that defined in Compose file

PR #3502, opened 2019-10-29
wc: add omit filename example

stsourlidakis:

### DIFF
--- a/pages/common/wc.md
+++ b/pages/common/wc.md
@@ -18,6 +18,6 @@
 
 `wc -m {{file}}`
 
-- Omit the filename from the output
+- Omit the filename from the output:
 
 `wc -l < {{file}}`
--- a/pages/common/wc.md
+++ b/pages/common/wc.md
@@ -18,6 +18,6 @@
 
 `wc -m {{file}}`
 
-- Omit the filename from the output:
+- Count lines in a file, omitting the filename from the output:
 
 `wc -l < {{file}}`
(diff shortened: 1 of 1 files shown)


pages/common/wc.md
```diff
@@ -17,3 +17,7 @@
 - Count characters in file (taking multi-byte character sets into account):
 
 `wc -m {{file}}`
+
+- Omit the filename from the output:
```
schneiderl: Maybe this will read better?

```suggestion
- Count words in a file, omitting the filename from the output:
```

stsourlidakis:
My example intents to explain how to omit the filename and not what the `-l` flag does, maybe I should remove the flag to make it clearer.
Ommiting the filename is useful when `wc` is used as a subcommand e.g. `echo "The file contains $(wc -l < file) lines"`

agnivade:
I understand, but this is not really in the scope of tldr pages. The tldr project is more of an explainer than a cheatsheet -- in other words, the underlying mission is not to merely provide copy-pastable snippets (like, say, commandlinefu.com) but rather to allow readers to understand the language of the commands, so that the logic behind the options is evident and they can then derive new commands from the provided examples, or compose several simple examples into more complex ones that are useful to them.

stsourlidakis:
I understand, feel free to close the PR.
Thank you both for taking time to review this!

agnivade:
No problem.

PR #3503, opened 2019-10-29
clip: add examples, fix typos

NNRepos:
added important examples and fixed old examples. Also removed spaces, which are generally a bad idea in batch.

### DIFF
--- a/pages/windows/clip.md
+++ b/pages/windows/clip.md
@@ -5,8 +5,16 @@
 
 - Pipe command line output to the Windows clipboard:
 
-`{{dir}} | clip`
+`dir|clip`
 
 - Copy the contents of a file to the Windows clipboard:
 
-`clip < {{path/to/file}}`
+`clip<path/to/file.ext`
+
+- Copy text with trailing newline to the Windows clipboard:
+
+`echo some text|clip`
+
+- Copy text without trailing newline to the Windows clipboard:
+
+`echo|set /p=some text|clip`
--- a/pages/windows/clip.md
+++ b/pages/windows/clip.md
@@ -5,16 +5,16 @@
 
 - Pipe command line output to the Windows clipboard:
 
-`dir|clip`
+`{{dir}} | clip`
 
 - Copy the contents of a file to the Windows clipboard:
 
-`clip<path/to/file.ext`
+`clip < {{path/to/file.ext}}`
 
-- Copy text with trailing newline to the Windows clipboard:
+- Copy text with a trailing newline to the Windows clipboard:
 
-`echo some text|clip`
+`echo {{some text}} | clip`
 
-- Copy text without trailing newline to the Windows clipboard:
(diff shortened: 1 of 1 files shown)


pages/windows/clip.md
```diff
@@ -5,8 +5,16 @@
 
 - Pipe command line output to the Windows clipboard:
 
-`{{dir}} | clip`
+`dir|clip`
```
sbrl: Could this change be reverted please? The spacing helps to aid clarity in the command, and the `{{` `}}` token syntax is actually a tldr-specific extension to markdown that indicates that a user can change a value. It isn't rendered in the final page that's shown to the user, as I'm sure you've noticed.

pages/windows/clip.md
```diff
@@ -5,8 +5,16 @@
...
 
 - Copy the contents of a file to the Windows clipboard:
 
-`clip < {{path/to/file}}`
+`clip<path/to/file.ext`
```
sbrl: Again, here the lack of spacing makes it harder to revert, and see also my comment about the token syntax `{{` `}}`.

pages/windows/clip.md
```diff
@@ -5,8 +5,16 @@
...
+
+- Copy text with trailing newline to the Windows clipboard:
+
+`echo some text|clip`
```
sbrl: Perhaps this will read better?
```suggestion
`echo "{{some text}}" | clip`
```

pages/windows/clip.md
```diff
@@ -5,8 +5,16 @@
...
 
-`{{dir}} | clip`
+`dir|clip`
 
 - Copy the contents of a file to the Windows clipboard:
 
-`clip < {{path/to/file}}`
+`clip<path/to/file.ext`
+
+- Copy text with trailing newline to the Windows clipboard:
```
sbrl: Perhaps this will read better?
```suggestion
- Copy text with a trailing newline to the Windows clipboard:
```

pages/windows/clip.md
```diff
@@ -5,8 +5,16 @@
...
+
+`echo some text|clip`
+
+- Copy text without trailing newline to the Windows clipboard:
```
sbrl: Perhaps this will read better?
```suggestion
- Copy text without a trailing newline to the Windows clipboard:
```

pages/windows/clip.md
```diff
@@ -5,8 +5,16 @@
...
+
+`echo|set /p=some text|clip`
```
sbrl: Perhaps this will read better?
```suggestion
`echo | set /p={{some text}} | clip`
```

Also, is the `echo | ` required in this example? Seems a little odd to me. Don't have a windows machine to test it on.

pages/windows/clip.md
NNRepos: I'll add the curly brackets, however, echoing with quotes prints the quotes as well, which is not the intended result.

PR #3508, opened 2019-10-30
ethtool: add page

borland502:
Established the entrry, and added the case I needed to use the command for

### DIFF
--- a/pages/linux/ethtool.md
+++ b/pages/linux/ethtool.md
@@ -5,4 +5,4 @@
 
 - Set the link speed, duplex mode, and parameter autonegotiation for a given interface:
 
-`ethtool -s eth0 speed {{10|100|1000}} duplex {{half|full}} autoneg {{on|off}}`
+`ethtool -s {{eth0}} speed {{10|100|1000}} duplex {{half|full}} autoneg {{on|off}}`
--- a/pages/linux/ethtool.md
+++ b/pages/linux/ethtool.md
@@ -5,4 +5,4 @@
 
 - Set the link speed, duplex mode, and parameter autonegotiation for a given interface:
 
-`ethtool -s {{interface}} speed {{10/100/1000}} duplex {{half/full}} autoneg {{on/off}}`
+`ethtool -s eth0 speed {{10|100|1000}} duplex {{half|full}} autoneg {{on|off}}`
--- a/pages/linux/ethtool.md
+++ b/pages/linux/ethtool.md
@@ -3,6 +3,6 @@
 > Display and modify Network Interface Controller (NIC) parameters.
 > More information: <http://man7.org/linux/man-pages/man8/ethtool.8.html>.
 
-- Set the link speed, duplex, and speed autonegotiation:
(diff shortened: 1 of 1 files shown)


mebeim:
Hi @borland502, I can see from [the man page](https://linux.die.net/man/8/ethtool) that this command actually has a wide variety of options and usages, while you are only suggesting one example. Are there any common usages/examples that you can think of? If so, it would be helpful to add them! Just a suggestion though, the new page is ok anyway.

pages/linux/ethtool.md
```diff
@@ -0,0 +1,7 @@
+# ethtool
+
+> Displays and modifies Network Interface Controller (NIC) parameters.
```
mebeim:
```suggestion
> Display and modify Network Interface Controller (NIC) parameters.
> More information: <http://man7.org/linux/man-pages/man8/ethtool.8.html>.
```
Let's stay impersonal, and while we're at it, also add a useful "more information" link.

pages/linux/ethtool.md
```diff
@@ -0,0 +1,7 @@
...
+
+- Set the link speed, duplex, and speed autonegotiation:
```
mebeim:
```suggestion
- Set the link speed, duplex mode, and parameter autonegotiation for a given interface:
```
Let's clarify here. Duplex "mode" is more correct, also the autonegotiation is not just for the speed. Let's also mention that this affects the selected interface only.

borland502:
> Hi @borland502, thank you for the contribution 🙂. I can see from [the man page](https://linux.die.net/man/8/ethtool) that this command actually has a wide variety of options and usages, while you are only suggesting one example. Are there any other common/useful examples that you know? If so, it would be helpful to add them! Just a suggestion though, the new page is ok anyway.

Thanks.  There are a ton, but I'm more comfortable adding what I've actually used -- I guess on the theory that I at least have a sample of one for a popular use case.  I'll make the changes you requested.

pages/linux/ethtool.md
```diff
@@ -0,0 +1,7 @@
...
+
+`ethtool -s {{interface}} speed {{10/100/1000}} duplex {{half/full}} autoneg {{on/off}}`
```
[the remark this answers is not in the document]
borland502: Curious though.  How widespread is eth0 though now?  I'm only using a few distros, but I see the new fangled enp1s0f0 name more often.

pages/linux/ethtool.md
```diff
@@ -0,0 +1,8 @@
+# ethtool
+
+> Display and modify Network Interface Controller (NIC) parameters.
+> More information: <http://man7.org/linux/man-pages/man8/ethtool.8.html>.
+
+- Set the link speed, duplex mode, and parameter autonegotiation for a given interface:
+
+`ethtool -s eth0 speed {{10|100|1000}} duplex {{half|full}} autoneg {{on|off}}`
```
mebeim:
```suggestion
`ethtool -s {{eth0}} speed {{10|100|1000}} duplex {{half|full}} autoneg {{on|off}}`
```
Keep curly braces around editable fields.

PR #3509, opened 2019-10-30
findstr: edit-page

### LINKED ISSUE #3504: page change request: windows/findstr
The 2 most common usages of findstr are not in the .md file:
findstr -sinr (recurSive, Insensitive, Numbered, Regex search), and
{{something}}|findstr.
I recommend removing /i, /n, /r which are rarely used by themselves.

NNRepos:
added more useful examples and removed some old ones. fixes #3504

### DIFF
--- a/pages/windows/findstr.md
+++ b/pages/windows/findstr.md
@@ -11,14 +11,6 @@
 
 `findstr /s "{{query}}" *`
 
-- Find strings using a case-insensitive search:
-
-`findstr /i "{{query}}" *"`
-
-- Find strings in all files using regular expressions:
-
-`findstr /r "{{expression}}" *`
-
 - Find a literal string (containing spaces) in all text files:
 
 `findstr /c:"{{query}}" *.txt`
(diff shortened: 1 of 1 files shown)


NNRepos:
> I disagree with the removal of those two examples. The page still has 9 examples, there is no need to remove both of them. Furthermore, those are basic and useful usages of the command that are worth being mentioned, just as `-r` and `-i` is mentioned for the `grep` command. IMHO those are two very useful (and very frequently used) options. I would be more inclined to remove `/c`, rather than `/i` or `/r`.

I haven't completely removed them - I've grouped them into one example. I think it's more concise, and also emphasizes that opts can be grouped.

mebeim:
I noticed, still that's not how TLDR pages are meant to be written IMHO. Of course someone could infer the behavior of single command-line options from a long and more complex command, but I've never seen this done. I'd like to see what other more experienced maintainers think. Ping @waldyrious @agnivade :eyes:

agnivade:
I agree with @mebeim. We always introduce options one by one and let user build them up and combine them as they get familiarized with it. Adding a new option if you think it is very common is fine. But not at the expense of making a page more complicated.

I would suggest you to choose something that you think can be removed and use that space for `{{dir}} | findstr "{{query}}"`

NNRepos:
@agnivade  your wish is my command

pages/windows/findstr.md
```diff
@@ -23,9 +23,9 @@
 
 `findstr /c:"{{query}}" *.txt`
 
-- Find only lines that match the query e[x]actly:
+- Find space-separated string(s) in a piped command's output:
```
agnivade: Let's make this the 2nd example.

NNRepos:
@agnivade good to merge?

PR #3519, opened 2019-10-31
dnf: add examples

Iniesta8:

### DIFF
--- a/pages/linux/dnf.md
+++ b/pages/linux/dnf.md
@@ -6,10 +6,6 @@
 
 `sudo dnf install {{package}}`
 
-- Install a new package and assume yes to all questions:
-
-`sudo dnf -y install {{package}}`
-
 - Remove a package:
 
 `sudo dnf remove {{package}}`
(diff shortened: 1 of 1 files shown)


Iniesta8:
I've also removed the `sudo dnf install -y` example because it's trivial.

sbrl:
I'm on the fence about the example removal there. On the one hand, it is trivial. Just because it's trivial doesn't mean to say that we should remove it! tldr-pages is aimed at beginners, and those to forget how a command works (amongst others).

On the other hand, assuming yes to all queries could be considered advanced and somewhat dangerous in nature. To that end, I'd prefer to remove it.

I'm not sure, what do you think?

/cc @agnivade

einverne:
I've noticed that there is `install` and `install -y` in same page. How about merge that into one?

Iniesta8:
The -y parameter also automatically installs dependencies of the desired package without the user having to confirm again. This can definitely lead to problems. In addition, the -y parameter of the install command is much less interesting than the other examples in the page.

agnivade:
Given that the `yum` page has this, I think we should keep it since we have space for more examples.

Iniesta8:
@agnivade : Ok, i've added the example with -y again. Ready to merge?

pages/linux/dnf.md
```diff
@@ -17,3 +17,15 @@
...
+
+- List installed packages:
+
+`dnf list --installed`
+
+- Search packages via keyword:
+
+`sudo dnf search {{keyword}}`
+
+- Find what package provides the given value:
```
sbrl: Perhaps this will read better?
```suggestion
- Find which packages provide a given value:
```

PR #3543, opened 2019-11-09
scanimage: add page

dmfay:

### DIFF
--- a/pages/linux/scanimage.md
+++ b/pages/linux/scanimage.md
@@ -1,8 +1,9 @@
 # scanimage
 
 > Scan images with the Scanner Access Now Easy API.
+> More information: <http://sane-project.org/>.
 
-- List available scanners to ensure it is connected and recognized:
+- List available scanners to ensure the target device is connected and recognized:
 
 `scanimage -L`
 
--- a/pages/linux/scanimage.md
+++ b/pages/linux/scanimage.md
@@ -1,7 +1,7 @@
 # scanimage
 
 > Scan images with the Scanner Access Now Easy API.
-> More information: <http://sane-project.org/>.
+> More information: <http://sane-project.org/man/scanimage.1.html>.
 
 - List available scanners to ensure the target device is connected and recognized:
 
--- a/pages/linux/scanimage.md
+++ b/pages/linux/scanimage.md
@@ -2,7 +2,7 @@
 
 > Scan images with the Scanner Access Now Easy API.
 
-- List available scanners to ensure yours is connected and recognized:
+- List available scanners to ensure it is connected and recognized:
 
 `scanimage -L`
 
--- a/pages/linux/scanimage.md
+++ b/pages/linux/scanimage.md
@@ -2,7 +2,7 @@
 
 > Scan images with the Scanner Access Now Easy API.
 
-- Make sure your scanner is recognized:
+- List available scanners to ensure yours is connected and recognized:
 
 `scanimage -L`
 
--- a/pages/linux/scanimage.md
+++ b/pages/linux/scanimage.md
@@ -1,0 +1,11 @@
+# scanimage
+
+> Scan images with the Scanner Access Now Easy API.
+
+- Make sure your scanner is recognized:
+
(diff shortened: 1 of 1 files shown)


pages/linux/scanimage.md
```diff
@@ -0,0 +1,11 @@
+# scanimage
+
+> Scan images with the Scanner Access Now Easy API.
+
+- Make sure your scanner is recognized:
```
agnivade: What does this mean ?

pages/linux/scanimage.md
```diff
@@ -0,0 +1,11 @@
...
+
+`scanimage -L`
+
+- Scan an image and save it to a file:
+
+`scanimage --format={{pnm|tiff|png|jpeg}} > {{new_image}}`
```
agnivade: path/to/new_image

pages/linux/scanimage.md
dmfay: The operating system has to know that you have a scanner attached, and, more to the point, that it _is_ a scanner in an appropriate state. "Recognized" seems appropriate.

agnivade: I was trying to understand how does `-L` map to the description. From the man [page](https://linux.die.net/man/1/scanimage), it seems that `-L` lists the devices. Let's make that the description ?

agnivade:
And btw, you don't need to force-push, it prevents us from seeing what changed between commits and we have a button to squash commits anyways while merging.

mebeim:
Is there an home page that we can link in the description?

pages/linux/scanimage.md
```diff
@@ -0,0 +1,12 @@
+# scanimage
+
+> Scan images with the Scanner Access Now Easy API.
+> More information: <http://sane-project.org/>.
```
mebeim:
```suggestion
> More information: <http://sane-project.org/man/scanimage.1.html>.
```
Better to use a more specific link since it's available.

dmfay:
I'm actually going to push back a little: if you have `scanimage` installed and the tldr is insufficient, `man scanimage` is right there. I've seen some other tldr items linking to documentation or online manpages instead of the home page, but there isn't (and can't be) a consistent convention.

Personally I don't agree that linking to the documentation -- especially in the case of manpages, which the user will already have installed -- is an improvement over linking to the home page. It's not inconceivable that I might want to reach the issue tracker or mailing list instead, and it's much more convenient to get there from the home page than from deep in the docs.

sbrl:
I believe it's been discussed before about which URLs should be used, @dmfay - though I can't find where (maybe we should have created a wiki page about it :sweat_smile:):

1. The specific documentation for the command-line tool itself
2. The general documentation for the project it's part of
3. The landing page website for the project
4. An online man page (optional; debatable as to whether this should be included)

.....though this is off the top of my head. Point is, that we prefer more specific links to less-specific links, so help users find what they're looking for faster.

dmfay:
Outside the specific point regarding manpages I would expect that more specific links help a subset of users find what they're looking for faster at the expense of hindering another subset. But I expect you've probably been over that too -- I'll go ahead and accept the change.

waldyrious:
@dmfay, thanks for the thoughtful comments regarding the usefulness of linking to manpages vs. homepages.

If it helps, the reasoning behind linking to more complete CLI documentation (including manpages, although that hasn't been unanimous in past discussions) is that the primary goal of tldr is to provide _usage documentation_ for CLI tools, rather than general information about them (other than the short sentence used in the main description).

As the name suggests, we focus on a compact subset of usage patterns in the interests of conciseness and approachability. Linking to a _documentation_ page, therefore, complements that mission by allowing the user to obtain the rest of the content that we've had to trim down from the tldr-pages entry.

That's why the link is labeled "More information" rather than "Project homepage", by the way. In fact, we would have gone with something more specific like "Complete documentation", but that wouldn't work for tools that don't have online docs but do have a website, or for those that have the documentation embedded in their website in a way that's not easily linkable.

Hope this makes sense to you! And thanks again for the contribution! 🙂

PR #3545, opened 2019-11-09
androguard: add page

Rudloff:

### DIFF
--- a/pages/common/androguard.md
+++ b/pages/common/androguard.md
@@ -1,0 +1,16 @@
+# androguard
+
+> Python tool to play with Android files.
+> More information: https://github.com/androguard/androguard.
+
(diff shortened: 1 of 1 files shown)


pages/common/androguard.md
```diff
@@ -0,0 +1,16 @@
+# androguard
+
+> Python tool to play with Android files.
```
schneiderl: I would prefer to state that it is a reverse engineering tool as "a tool to play" seem a little vague. 
Maybe something like
```suggestion
> Reverse engineering tool for Android applications. Written in Python.
```

What do you think?

Rudloff: I reused the Debian description, but I agree yours is better :+1:

PR #3562, opened 2019-11-12
logcat: add page

Rudloff:

This is an Android-only tool, so I assumed `linux` was the correct directory?

### DIFF
--- a/pages/linux/logcat.md
+++ b/pages/linux/logcat.md
@@ -1,0 +1,17 @@
+# logcat
+
+> Dump a log of system messages.
+> Native Android CLI tool.
+> More information: <https://developer.android.com/studio/command-line/logcat>.
+
+- Display system logs:
+
+`logcat`
+
+- Write system logs to a file:
+
+`logcat -f {{path/to/file}}`
+
+- Display lines that match a regex:
+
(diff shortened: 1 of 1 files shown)


sbrl:
Do you mean to say that it is a command only available on Linux for interacting with Android devices, or that it only runs on Android devices? If it's the latter, then we *may* need a new platform folder.

Rudloff:
This command is only available on Android devices.
(It can be called from a computer connected to the device with `adb logcat` though.)

sbrl:
Ah, I see @Rudloff. In that case, I think we might need to tweak this a bit. I'd suggest that it is indeed on Linux that it's used most often. In that case, we should treat this as a sub-page of `adb` - just like `git merge` is a sub-page of git. This would mean that we'd need to do the following:

 - Rename the page to `adb-logcat.md`
 - Prefix all instances of `logcat` with `adb ` in the page itself

If the opposite is true (i.e. it's used on Android directly more often), then we may need to create a new platform folder.

What do you think?

Rudloff:
Hum, I guess it is more often used with adb.

agnivade:
Yeah, let's do what @sbrl said.

PR #3563, opened 2019-11-12
dexdump: add page

Rudloff:

I could not find an official documentation page.

### DIFF
--- a/pages/common/dexdump.md
+++ b/pages/common/dexdump.md
@@ -1,0 +1,19 @@
+# dexdump
+
+> Display information about Android DEX files.
+
+- Extract classes and methods from an APK file:
+
+`dexdump {{path/to/file.apk}}`
+
+- Display DEX files header information:
+
+`dexdump -f {{path/to/file.apk}}`
+
+- Display the dis-assembled output of executable sections:
+
+`dexdump -d {{path/to/file.apk}}`
+
+- Output results to a file:
+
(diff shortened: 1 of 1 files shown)


pages/common/dexdump.md
```diff
@@ -0,0 +1,19 @@
+# dexdump
+
+> Display information about Android DEX files.
+
```
einverne: Is there any document link we can put here?

Rudloff: I could not find any official documentation :slightly_frowning_face:

pages/common/dexdump.md
```diff
@@ -0,0 +1,19 @@
...
+- Extract classes and methods from an APK file:
+
+`dexdump {{path/to/file.apk}}`
+
+- Display DEX files header information:
```
agnivade: Just wanted to clarify whether apk files and dex files are one and the same ? Then perhaps we should mention apk file here rather than dex files. Perhaps: "Display header information of a DEX/apk file"

Rudloff: An APK file is an archive that can contain several DEX files.
I tried to clarify that.

pages/common/dexdump.md
einverne: Okay.

bl-ue: Looks like this conversation never got resolved.  I'll do that now.  Btw @Rudloff it's okay to use non-official links, at least as of 2021. 😉

PR #3568, opened 2019-11-14
robo: add page

prudloff-insite:

### DIFF
--- a/pages/common/robo.md
+++ b/pages/common/robo.md
@@ -3,6 +3,10 @@
 > PHP task runner.
 > More information: <https://robo.li/>.
 
+- List available commands:
+
+`robo list`
+
 - Run a specific command:
 
 `robo {{foo}}`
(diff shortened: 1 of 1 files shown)


pages/common/robo.md
```diff
@@ -0,0 +1,12 @@
+# robo
+
+> PHP task runner.
+> More information: <https://robo.li/>.
+
+- Run a specific command:
```
sbrl: Just want to clarify: Where does it find the tasks that it runs?

prudloff-insite: Tasks are defined in a file called `RoboFile.php` in the current folder.
I am not sure there is a way to specify another path for this file.

sbrl: Ah, I see. Thanks for clarifying :-)

prudloff-insite:
`grumphp list` will list all the commands.

sbrl:
Ah, cool @prudloff-insite. Perhaps we could add another example for that then please? It does sound useful enough to be included - probably before the current first example actually.

sbrl:
Thanks, @prudloff-insite :smiley_cat:

PR #3569, opened 2019-11-14
pm2: update page

nire0510:
New commands: save, resurrect & monit

### DIFF
--- a/pages/common/pm2.md
+++ b/pages/common/pm2.md
@@ -23,3 +23,15 @@
 - Restart a process:
 
 `pm2 restart {{myapp}}`
+
+- Dump all processes for resurrecting them later:
+
+`pm2 save`
+
+- Resurrect previously dumped processes:
+
+`pm2 resurrect`
+
+- Launch monitoring:
+
+`pm2 monit`


pages/common/pm2.md
```diff
@@ -23,3 +23,15 @@
 - Restart a process:
 
 `pm2 restart {{myapp}}`
+
+- Dump all processes for resurrecting them later:
```
sbrl: This doesn't make grammatical sense. Perhaps we could do something like this instead?
```suggestion
- Dump all processes for later resurrecting:
```
Also, off choice of words. Usually commands like this use freeze / thaw or something like that instead. "Resurrecting" makes it sound like the process has died previously, which doesn't appear to technically be the case here.

nire0510: Hi @sbrl , I don't mind accepting your suggestion, but I actually took the description from their own help file, **as is** 🤔. What do you suggest?

agnivade: We usually prefer to change the descriptions to make it more suitable to newcomers. But I am curious to know @sbrl why do you think this is grammatically incorrect. It looks fine to me, or perhaps "... to resurrect them later" is also okay.

sbrl: Taking another look, it does actually make sense here. It still sounds weird though - but I guess this is fine as-is.

mebeim:
Merging since previous comments have all been addressed.

mebeim:
Ah, probably should have changed the commit message to be clearer. My bad :\

PR #3572, opened 2019-11-15
Remove duplicated Korean pages

mebeim:
Some common Korean pages were duplicated in the top level `pages.ko` folder.

NOTE: this was also [making the build silently fail to generate `index.json`](https://travis-ci.org/tldr-pages/tldr/jobs/612185741#L230) (damn bash pipes...), though that error is fixed in an upcoming PR that I will submit soon. In the meantime, this should also make `npm build-index` run correctly.


agnivade:
I will merge this right away as this is masking real failures in the CI.

PR #3573, opened 2019-11-15
Add trailing newline to Korean pages

mebeim:
All common Korean pages were missing a trailing newline.

Issue #2865, opened 2019-04-03
Removal of duplicated index.json from the generated ZIP archive, up-to-date clients tracking


mebeim:
As stated in #2862, I'm opening this to keep track of clients' compliance with the small change to the ZIP generation, that is, adding a top-level `/index.json` as replacement for `/pages/index.json`. Right now, both files are present since some clients rely on `/pages/index.json` to work, but the goal is to completely remove it and only keep the top-level `/index.js`.

As a consequence of this, I'm going to open issues on repositories of clients which rely on `/pages/index.json` being in the generated ZIP to alert them of the new change. Look at the bottom to see what I'll be mostly copy-pasting around for this purpose. 

---

## List of up-to-date clients

Checked box == up-to-date.

- [ ] [anoopengineer/tldr](https://github.com/anoopengineer/tldr): opened [issue #2](https://github.com/anoopengineer/tldr/issues/2).
- [x] [BrainMaestro/tldr-php](https://github.com/BrainMaestro/tldr-php): already ok.
- [x] [cs1707/tldr-alfred](https://github.com/cs1707/tldr-alfred): already ok.
- [x] [dbrgn/tealdeer](https://github.com/dbrgn/tealdeer): opened [issue #80](https://github.com/dbrgn/tealdeer/issues/80).
- [ ] DistroWatch: **lacks source/repo**, unknown status.
- [x] [edgurgel/tldr_elixir_client](https://github.com/edgurgel/tldr_elixir_client): already ok.
- [x] [elecprog/tldr](https://github.com/elecprog/tldr): already ok.
- [x] [ex0dus-0x/tldr-bot](https://github.com/ex0dus-0x/tldr-bot): already ok.
- [x] [freesuraj/TLDR](https://github.com/freesuraj/TLDR): already ok.
- [x] [fuerbringer/tldr-telegram](https://github.com/fuerbringer/tldr-telegram): already ok.
- [x] [gianasista/tldr-viewer](https://github.com/gianasista/tldr-viewer): already ok.
- [x] [hidroh/tldroid](https://github.com/hidroh/tldroid): already ok.
- [x] [hill/tldr-chrome](https://github.com/hill/tldr-chrome): already ok.
- [x] [hterkelsen/tldr](https://github.com/hterkelsen/tldr): already ok.
- [x] [isacikgoz/tldr](https://github.com/isacikgoz/tldr): already ok.
[...]

PR #3576, opened 2019-11-15
Update build process, add additional PR build checks

mebeim:
Finally, after some<sup>[1](https://github.com/tldr-pages/tldr/pull/3572)</sup> little<sup>[2](https://github.com/tldr-pages/tldr/pull/3573)</sup> hiccups<sup>[3](https://github.com/tldr-pages/tldr/pull/3575)</sup> here<sup>[4](https://gitter.im/tldr-pages/tldr?at=5dcebb6792a84f79fe70b7ce)</sup> and there<sup>[5](https://gitter.im/tldr-pages/tldr?at=5dcec56f35889012b1131762)</sup>, I'm able to open this PR.

Commit messages alone should already provide a good changelog, and I've commented the code where needed, so make sure to check it out! I tested this on my fork for what I could, and it *looks fine*. I could not do a test run to make tldr-bot generate comments, but that can be done once this is merged.

The main reason for this PR was to add a script to check pull requests for possible problems, like duplicated, copied or misplaced pages. This is accomplished through the new [`check-pr.sh`](https://github.com/tldr-pages/tldr/blob/8af498e0d3b73c8d82982d0b156e7fb355634ed6/scripts/check-pr.sh) script (see the script itself for a better description of what it checks). 

Since the PR check would need to contact tldr-bot to add a comment, and since it only makes sense to add such comment when the PR is opened (not on every push to the branch), this also required [an update to tldr-bot](https://github.com/tldr-pages/tldr-bot/pull/6), which should be merged soon. The [`send-to-bot.py`](https://github.com/tldr-pages/tldr/blob/8af498e0d3b73c8d82982d0b156e7fb355634ed6/scripts/send-to-bot.py) script was updated to support this additional functionality and also refactored.

<s>It's important to note that since I've updated `send-to-bot.py` to use the new endpoints `/comment` and `/comment/once` to talk to tldr-bot, **this PR depends on https://github.com/tldr-pages/tldr-bot/pull/6**, so please don't merge until that is merged first!</s> Merged!

Other remarkable changes are:

 - Changing `npm test` to run a real bash script (`test.sh`) to make it cleaner and avoid bash piping errors.
 - Bumping `markdownlint-cli` in the `package.json` to version 0.19.0, since version 0.15.0 had a bug that made it fail to recognize some errors.


package.json
```diff
@@ -4,15 +4,15 @@
   "description": "Simplified, community-driven man pages",
   "dependencies": {
     "glob": "7.1.3",
-    "markdownlint-cli": "0.15.0",
+    "markdownlint-cli": "0.19.0",
```
mebeim: Version 0.15.0 had a bug that didn't recognize some errors.

scripts/build.sh
```diff
@@ -13,6 +13,7 @@ function initialize {
 
 function build_index {
   npm run build-index
+  cp index.json pages/
```
mebeim: The `npm run build-index` command was changed to only generate `index.json` in the current directory. The build script duplicates it where needed. This line will be removed once #2865 is fixed.

scripts/check-pr.sh
```diff
@@ -0,0 +1,119 @@
...
+    local file1="${entry[1]}"
+    local file2="${entry[2]}"
+
+    case "$change" in
+      C*) # file2 is a copy of file1
+        local percentage=${change#C}
+        percentage=${percentage#0}
+        percentage=${percentage#0}
+
+        printf "\x2d $msg" "$file2" "$file1" "$percentage"
```
mebeim: The `\x2d` here is just a little trick to make `printf` correctly print a hyphen (`-`) when it's the first character of the string, instead of interpreting it as a command line option.

scripts/test.sh
```diff
@@ -0,0 +1,54 @@
...
+# It runs some basic tests on pages. If the build is also a PR, additional
+# checks are run through the check-pr script, and any message or error is sent
+# to tldr-bot to be commented on the PR.
+#
+# NOTE: must be run from the repository root directory to correctly work!
+# NOTE: `set -e` is applied conditionally only if needed.
+
+# Default test.
+function default_test {
+  markdownlint pages*/**/*.md
```
mebeim: This is the only thing that `npm test` will end up doing when normally run. 

On Travis PR builds, the other two functions below are run instead.

scripts/test.sh
```diff
@@ -0,0 +1,54 @@
...
+}
+
+# Special test for Travis CI pull request builds.
+# Runs default_test collecting errors for tldr-bot.
+function pr_test {
+  errs=`default_test 2>&1`
+
+  if [ -n "$errs" ]; then
+    echo -e "Test failed!\n$errs\n" >&2
+    echo 'Sending errors to tldr-bot.' >&2
```
agnivade: Refactor these 2 lines in a function and call them. You are using them in pr_test and pr_check.

scripts/test.sh
```diff
@@ -0,0 +1,54 @@
...
+    echo -n "$msgs" | python3 scripts/send_to_bot.py check
+  fi
+}
+
+###################################
+# MAIN
+###################################
+
+if [ "$TRAVIS" = "true" ] && [ "$TRAVIS_REPO_SLUG" = "tldr-pages/tldr" ] && [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+  pr_check
```
agnivade: I think if you refactor the 2 lines I suggested above, you don't need separate functions for these. Just have functions for default_test (probably can rename it to test now), and the send_to_bot.

scripts/check-pr.sh
```diff
@@ -0,0 +1,119 @@
...
+  local other
+
+  readarray -td'/' parts < <(echo -n "$page")
+
+  local platform=${parts[0]}
+  local file=${parts[1]}
+
+  case "$platform" in
+    common) # check if page already exists in other platforms
+      for other in `ls pages/`; do
```
agnivade: Let's compute `ls pages/` from before and reuse it. This is being executed repeteadly.

scripts/send_to_bot.py
```diff
@@ -1,41 +1,94 @@
...
+
+def post_comment(pr_id, body, once):
+  if once:
+    endpoint = BOT_URL + '/comment/once'
+  else:
+    endpoint = BOT_URL + '/comment'
+
+  headers = {'Content-Type': 'application/json'}
+  data = json.dumps({'pr_id': pr_id, 'body': body})
+  req = urllib.request.Request(endpoint, data.encode(), headers)
```
agnivade: Now I remember why I was talking about not using `requests` in the tldr bot script. It was here that I did not use `requests` :stuck_out_tongue_closed_eyes:

scripts/test.sh
mebeim: I'm not sure if it makes sense to refactor this in a function. It would add a lot of boilerplate code just for a simple command. The two messages are different and the two invocations of the python script are different too, this would mean creating a function that would need to be called with two arguments and have additional logic inside, it would make the code harder to read IMHO.

scripts/test.sh
mebeim: I think this would hurt readability. I split the code in two functions to clearly define what is a test (that can fail the build) and what is just a check (that can generate warnings but not fail the build).

scripts/test.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# Special test for Travis CI pull request builds.
+# Runs default_test collecting errors for tldr-bot.
+function pr_test {
+  errs=`default_test 2>&1`
+
+  if [ -n "$errs" ]; then
+    echo -e "Test failed!\n$errs\n" >&2
+    echo 'Sending errors to tldr-bot.' >&2
+    echo -n "$errs" | python3 scripts/send_to_bot.py error
```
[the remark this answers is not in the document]
mebeim: Does it make sense for a script to have one single option which is also required? The way it is implemented now, it works much like a subcommand. Maybe changing the name of the two actions would be enough to make it clear, what do you think?

scripts/test.sh
```diff
@@ -0,0 +1,54 @@
...
+  markdownlint pages*/**/*.md
+  tldr-lint ./pages
+}
+
+# Special test for Travis CI pull request builds.
+# Runs default_test collecting errors for tldr-bot.
+function pr_test {
+  errs=`default_test 2>&1`
+
+  if [ -n "$errs" ]; then
```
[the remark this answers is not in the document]
mebeim: Does this:

```bash
function f {
    res=`cmd`

    if [ $res != "err" ]; then
        return 0
    fi 

    echo "error!"
    exit 1
}
```

... really read better/clearer than this?

```bash
function f {
    res=`cmd`

    if [ $res = "err" ]; then
        echo "error!"
        exit 1
    fi
}
```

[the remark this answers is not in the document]
agnivade: Well, this is what it would look like:

```
  if [ -z "$errs" ]; then
   return 0
 fi

  echo -e "Test failed!\n$errs\n" >&2
  echo 'Sending errors to tldr-bot.' >&2
  echo -n "$errs" | python3 scripts/send_to_bot.py error
  exit 1
```

So yeah, I think it is clearer. It is always good to unindent the main case, and indent the shorter, return case.

scripts/test.sh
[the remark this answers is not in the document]
agnivade: Ok, let's go forward with better names then.

scripts/test.sh
agnivade: >  this would mean creating a function that would need to be called with two arguments and have additional logic inside,

Yep, there would be 2 arguments, but I fail to see where you would need additional logic ? It will just be `echo -n "$msg" | python3 scripts/send_to_bot.py $command` right ?

scripts/test.sh
agnivade: The default_test function will still be there. You will just be unwrapping pr_test and pr_check. When I read the code, it wasn't immediately clear to me that `pr_test` wraps `default_test`. Hence I suggested this. If you still want to do it, I would suggest to use names like `test` and `wrapped_test` or something similar.

scripts/test.sh
mebeim: Well, to refactor L24 and L25 into a function they should be the same as L37 and L38. In particular L24 and L37 differ, one says "sending errors" and one says "sending check results". 

To refactor both L24 and L25 into a function would mean adding additional logic in the function. Otherwise, refactoring only L25 basically means creating a function to substitute one single line of code... which seems like over-engineering the thing IMHO.

scripts/test.sh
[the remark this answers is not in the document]
mebeim: That's not the "main case" though, it's an exception, that's why it seemed unnatural to me.

I mean, it's like:

    if (ok)
        return Ok;
    return Err;

I've never seen this pattern in C, Go, Python or whichever other language.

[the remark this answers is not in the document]
agnivade: By "main case", I meant the case with the main logic. It can be the error case or the no-error case. And it is very much a pattern in Go I can assure you ;) I am not too strongly attached to it though. So do go ahead with the existing way if you feel more comfortable with it.

scripts/test.sh
agnivade: Oh, my bad, I missed that the log messages were different. You can parameterize that too.

But if you feel that's complicating things too much, let's use better names like I mentioned [here](https://github.com/tldr-pages/tldr/pull/3576/files#r347075482).

scripts/test.sh
mebeim: I avoided naming the function exactly `test` because that would have shadowed the `test` command and it seemed like a bad idea :confused: 

`default_test` is the first name that popped into my mind, probably not *the best* name though, but right now I can't think of something else that's better other than just `test`... I've got to think harder :sweat_smile:

mebeim:
I just pushed some more commits. I noticed that the `.travis.yml` file still had the `after_failure` part that is not needed anymore. I also renamed the python script using hyphens instead of underscores (like every other script).

Any thoughts [here](https://github.com/tldr-pages/tldr/pull/3576#discussion_r346944887) and [here](https://github.com/tldr-pages/tldr/pull/3576#discussion_r346944394) guys? @waldyrious @sbrl @pxgamer I don't really have good ideas about how to rename the functions.

scripts/test.sh
```diff
@@ -0,0 +1,54 @@
...
+# This script is executed by Travis CI for every successful push (on any branch, PR or not).
+# It runs some basic tests on pages. If the build is also a PR, additional
+# checks are run through the check-pr script, and any message or error is sent
+# to tldr-bot to be commented on the PR.
+#
+# NOTE: must be run from the repository root directory to correctly work!
+# NOTE: `set -e` is applied conditionally only if needed.
+
+# Default test.
+function default_test {
```
sbrl: I'd suggest using the following function syntax instead:

```bash
default_test() {
    # ....
}
```

It's more portable :-)

scripts/test.sh
sbrl: I think it's probably fine, but we could have a function like this:

```bash
make_comment() {
    type="${1}" # either "error" or "check"
    message = "${2}"
    echo -n "${message}" | python3 scripts/send_to_bot.py "${type}"
}
```

The 2nd param is optional - we could leave it as taking stdin I guess.

scripts/test.sh
```diff
@@ -0,0 +1,54 @@
...
+  markdownlint pages*/**/*.md
+  tldr-lint ./pages
+}
+
+# Special test for Travis CI pull request builds.
+# Runs default_test collecting errors for tldr-bot.
+function pr_test {
+  errs=`default_test 2>&1`
```
sbrl: Use `errs="$(default_test 2>&1)";` here. The backtick syntax is not obvious.

scripts/test.sh
sbrl: I think I'd prefer them separate personally.

As for a name, what about `default_test` -> `run_tests` and `pr_test` -> `run_tests_travis`?

I'd argue that the 2nd function doesn't have any additional checks at all - it's just a wrapper for handling the error messages differently.

scripts/test.sh
mebeim: We're using the `function fname { ... }` syntax everywhere though. How is this more portable?

scripts/test.sh
mebeim: Does it really? I've used it all around in the rest of the scripts. Are you suggesting to use `$(cmd)` everywhere instead of <code>`cmd`</code>?

scripts/test.sh
mebeim: I like the names, and yes that's exactly the purpose of the second function.

mebeim: @agnivade what do you think about naming the functions `run_tests`, `run_tests_travis_pr` and `run_checks_travis_pr`?

scripts/test.sh
agnivade: Some interesting discussion here: https://stackoverflow.com/questions/9449778/what-is-the-benefit-of-using-instead-of-backticks-in-shell-scripts

scripts/test.sh
sbrl: Just looked it up, and I must be mistaken in this instance.

Ref https://ryanstutorials.net/bash-scripting-tutorial/bash-functions.php

scripts/test.sh
sbrl: Basically `$()` is actually POSIX

mebeim: I see, I shall refactor all the code to use `$()` then. Thanks guys.

scripts/test.sh
agnivade: Sounds good to me. Is there a reason to add "_travis_" ? Seems to me that "run_tests_pr" is sufficient.

scripts/test.sh
agnivade: This is basically refactoring one line. Which is what @mebeim mentioned here
> Otherwise, refactoring only L25 basically means creating a function to substitute one single line of code... which seems like over-engineering the thing IMHO.

mebeim: All I'm saying is that moving that command into a function would mean writing something like this:

```bash
function send_to_bot {
	echo -n "$1" | python3 scripts/send_to_bot.py $0
}

# ...

send_to_bot error "$errs"
```

instead of just:

```bash
echo -n "$errs" | python3 scripts/send_to_bot.py error
```

As I said it looks like over-engineering the thing, so I would not advise this unless you guys have a concrete example that makes more sense.

agnivade: I am with you @mebeim on this one :) . I was just explaining what you already mentioned to @sbrl.

mebeim: OK, I should have addressed all remaining comments, leaving this open for @sbrl to read.

mebeim:
Aaand we just tested out that the test correctly detects errors :sweat_smile:. PR #3583 was merged a while ago (after I opened this) and the pages have some errors. I'll fix them in another PR.

scripts/test.sh
sbrl: No particular reason, except to make it more explanatory.

sbrl:
Merged #3633, @mebeim :-)

mebeim:
Ah, no need to do anything, seems like this can just be merged with a rebase as is.

waldyrious:
I haven't performed a thorough review myself, but I've followed the conversation, and I trust @mebeim, @sbrl and @agnivade (and the actual code, thanks to Travis!), so I'll go ahead and merge! 😃

waldyrious:
@mebeim would you like to combine some of the commits or are you happy with keeping them as they are?

mebeim:
@waldyrious I'm fine with the current commits, thanks for asking :)

PR #3579, opened 2019-11-16
multiple pages(bzip2 to cake) : Add Korean Translation

Proscream:


Proscream:
@mebeim Of course it can be translated. But in Korea, some nouns like playbook are better to call in English. It is not called as something else, but as a person's name is called as an English pronunciation. So it seems hard to translate everything.

mebeim:
@Proscream no problem. If something cannot be translated or it's better to say it in English then that's ok, not everything needs to be translated.

pages.ko/common/c99.md
```diff
@@ -0,0 +1,19 @@
+# c99
+
+> ISO C 표준에 따라 C 프로그램을 컴파일한다.
+
+- 원본 파일 컴파일 및 실행 파일 생성:
+
+`c99 {{.c파일}}`
```
mebeim:
```suggestion
`c99 {{.c파일}}`
```
Does it make sense to have this say `.c파일`? What is usually used for something like this in manual pages in Korean? It would be very helpful if you could provide a reference (if you know of one). 

If this feels awkward/unnatural and it reads better in English, we can keep it in English. I just want to make sure that's clear.

Proscream: @mebeim Perhaps this page will be used by programmers. And since most Korean programmers code in English, the instructions and the terms in the computer are used in English. "file.c" may be more like "파일명.c." ".c파일" means a file of type ".c"

mebeim: @Proscream I see, so in this case `.c파일` does not make sense for sure. `파일명.c` is better, but if you think English works best then let's use English for this kind of tokens.

sbrl:
Merging, since @mebeim's comments appear to have been addressed.

PR #3588, opened 2019-11-18
Correctly store language and platform mappings together

agnivade:
Previously they were stored separately in their own arrays.
This made them disconnected, and impossible to know if a page
belonged to which platform and language combination.

### DIFF
--- a/scripts/build-index.js
+++ b/scripts/build-index.js
@@ -22,15 +22,20 @@ function buildPagesIndex(files) {
     let language = parseLanguage(file);
 
     if (index[page]) {
-      if (!index[page].platform.includes(os)) {
-        index[page].platform.push(os);
+      let targets = index[page].targets;
+      let needsPush = true;
+      for (const target of targets) {
+        if (target.platform === os && target.language === language) {
+          needsPush = false;
+          continue
+        }
       }
-
-      if (!index[page].language.includes(language)) {
-        index[page].language.push(language);
+      if (needsPush) {
+        targets.push({"os": os, "language": language})
+        index[page].targets = targets;
       }
     } else {
-      index[page] = {name: page, platform: [os], language: [language]};
+      index[page] = {name: page, targets: [{"os": os, "language": language}]};
     }
 
     return index;
@@ -43,8 +48,7 @@ function buildPagesIndex(files) {
       .map(function(page) {
         return {
           name: page,
-          platform: obj[page]["platform"],
-          language: obj[page]["language"]
+          targets: obj[page]["targets"]
         };
       });
 }
--- a/scripts/build-index.js
+++ b/scripts/build-index.js
@@ -22,6 +22,12 @@ function buildPagesIndex(files) {
     let language = parseLanguage(file);
 
     if (index[page]) {
+      if (!index[page].platform.includes(os)) {
+        index[page].platform.push(os);
+      }
+      if (!index[page].language.includes(language)) {
+        index[page].language.push(language);
+      }
       let targets = index[page].targets;
       let needsPush = true;
       for (const target of targets) {
@@ -35,7 +41,11 @@ function buildPagesIndex(files) {
         index[page].targets = targets;
       }
     } else {
-      index[page] = {name: page, targets: [{"os": os, "language": language}]};
+      index[page] = {name: page,
+        platform: [os],
+        language: [language],
+        targets: [{"os": os, "language": language}]
+      };
     }
 
     return index;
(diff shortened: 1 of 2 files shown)


agnivade:
@mebeim - PTAL

mebeim:
I just noticed that we have `parsePlatform`, `parsePagename` and `parseLanguage`, which are all basically doing the same `pagefile.split(/\//)` over and over again. We could squash those into a simpler function and do something like `[language, platform, page] = parsePath(file)`. What do you think, should I open another PR?

scripts/build-index.js
```diff
@@ -25,12 +25,27 @@ function buildPagesIndex(files) {
       if (!index[page].platform.includes(os)) {
         index[page].platform.push(os);
       }
-
       if (!index[page].language.includes(language)) {
```
mebeim: I would really suggest adding an empty line above L28, L31, L33, and L39. Having this all condensed makes it kind of hard to read.

scripts/build-index.js
```diff
@@ -25,12 +25,27 @@ function buildPagesIndex(files) {
...
+          continue
+        }
+      }
+      if (needsPush) {
+        targets.push({"os": os, "language": language})
+        index[page].targets = targets;
+      }
     } else {
-      index[page] = {name: page, platform: [os], language: [language]};
+      index[page] = {name: page,
```
mebeim:
```suggestion
      index[page] = {
        name: page,
```
Probably an oversight.

scripts/build-index.js
```diff
@@ -25,12 +25,27 @@ function buildPagesIndex(files) {
...
         index[page].language.push(language);
       }
+      let targets = index[page].targets;
+      let needsPush = true;
+      for (const target of targets) {
+        if (target.platform === os && target.language === language) {
+          needsPush = false;
+          continue
+        }
+      }
```
mebeim: This could all be simplified as:

```js
needsPush = targets.some(t => t.platform != os || t.language != language);
```

scripts/build-index.js
```diff
@@ -25,12 +25,27 @@ function buildPagesIndex(files) {
...
+      if (needsPush) {
+        targets.push({"os": os, "language": language})
+        index[page].targets = targets;
```
mebeim:
```suggestion
        index[page].targets = targets;
```

This line isn't necessary.

agnivade: Yes, I saw that. I wasn't too sure about what JS does internally. Thanks for confirming.

agnivade:
> What do you think, should I open another PR?

Sure !

scripts/build-index.js
```diff
@@ -29,8 +29,19 @@ function buildPagesIndex(files) {
       if (!index[page].language.includes(language)) {
         index[page].language.push(language);
       }
+
+      const targets = index[page].targets;
+      const exist = targets.some((t) => {return t.platform === os && t.language === language});
```
mebeim: No need for `()`, `{}` and `return` here. It can just be `t ⇒ expr`. Just checking if this is intended, if you want to keep it like this it works as well, it's just longer.

scripts/build-index.js
```diff
@@ -29,8 +29,19 @@ function buildPagesIndex(files) {
...
+      if (!exist) {
+        targets.push({"os": os, "language": language})
+      }
     } else {
-      index[page] = {name: page, platform: [os], language: [language]};
+      index[page] = {
+        name: page,
+        platform: [os],
+        language: [language],
+        targets: [{"os": os, "language": language}]
```
mebeim: Now that I think about it this could be rewritten in ES6 style as:

```suggestion
        targets: [{os, language}]
```

What do you think?

scripts/build-index.js
```diff
@@ -44,7 +55,8 @@ function buildPagesIndex(files) {
         return {
           name: page,
           platform: obj[page]["platform"],
-          language: obj[page]["language"]
+          language: obj[page]["language"],
```
mebeim: Also here I see that `["key"]` is used while in the rest of the script the dot is used instead (`.key`). Should these lines be refactored to use the dot?

scripts/build-index.js
mebeim: Also, the variable should be called `exists`.

agnivade: Yeah, it is intended. I find the `()` to be more readable. Will change the variable.

scripts/build-index.js
agnivade: Ooh, I had forgotten about that feature. Nice.

scripts/build-index.js
agnivade: I don't mind.

sbrl: Personally I'd refactor, but it isn't a big deal.

PR #3590, opened 2019-11-18
bluetoothctl: add page

riscie:

### DIFF
--- a/pages/linux/bluetoothctl.md
+++ b/pages/linux/bluetoothctl.md
@@ -1,0 +1,27 @@
+# bluetoothctl
+
+> Handling bluetooth devices from the shell
+
+- Enter the bluetoothctl shell:
+
+`bluetoothctl`
+
+- List devices:
+
+`bluetoothctl -- devices`
+
+- Pair a device:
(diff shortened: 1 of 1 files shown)


pages/linux/bluetoothctl.md
```diff
@@ -0,0 +1,27 @@
...
+
+`bluetoothctl`
+
+- List devices:
+
+`bluetoothctl -- devices`
+
+- Pair a device:
+
+`bluetoothctl -- pair {{MAC address}}`
```
agnivade: Let's change "MAC address" to "mac_address" here and everywhere else.

riscie:
> LGTM. Thank you @riscie .
> 
> By the way, is there any document link we can put here?

@einverne the [official site](http://www.bluez.org) does not have the cli documented. We could link to the ubuntu or arch wiki, but that seems kind of wrong too. That's why I left it out.

pages/linux/bluetoothctl.md
riscie: Done

PR #3598, opened 2019-11-20
cmake: improve out-of-source build example

waldyrious:
This trick was first introduced in #1386. I can't find any discussion about why the `-H.` was needed; I do vaguely recall wondering about that back when I first came across this trick, and my impression was that the trick didn't work without it; but I just tested the command without the `-H` and it seems to work:

<details>

- Create a temporary folder and enter it:
  ```sh
  cd $(mktemp -d)
  ```

- Create the source file:
  ```sh
  echo '#include<iostream>
  
  int main(int argc, char *argv[]){
    std::cout << "Hello World!" << std::endl;
    return 0;
  }' > helloworld.cpp
  ```

- Create the project file:
  ```sh
  echo 'cmake_minimum_required(VERSION 2.8)
  project (hello)
  add_executable(hello helloworld.cpp)' > CMakeLists.txt
  ```

- Compile the project:
  ```
  cmake -B foobar && make -C $_
  ```
- Run the executable:
  ```
  ./foobar/hello #Hello World!
  ```
</details>

(Note that this was on a macOS machine; I will test under Linux later on.)

In any case, it does seem weird that the `-H` option would be needed for this command to work, because it's been an alias of the `--help` option for [at least a decade and a half](https://github.com/Kitware/CMake/commit/d54e7a68885e43c53a8929a57b50b34b2be03d2d#diff-c73e705212ebf865a952d011da13cd5dR567)!

Besides removing the need for the `-H` argument, the example can even be simplified further to avoid typing the folder name twice, by using the `$_` trick (that change was inspired by [this UnixToolTip tweet](https://twitter.com/UnixToolTip/status/1196483245778571266)).

The result is a much cleaner example! 🎉

### DIFF
--- a/pages/common/cmake.md
+++ b/pages/common/cmake.md
@@ -10,7 +10,7 @@
 
 - Generate a Makefile and use it to compile a project in a separate "build" directory (out-of-source build):
 
-`cmake -H. -B{{build}} && make -C {{build}}`
+`cmake -B {{build}} && make -C $_`
 
 - Run cmake in interactive mode (it will ask for each variable, instead of using defaults):
 


waldyrious:
Unfortunately my Linux tests are not looking good :( on one machine `cmake -B foobar` works, but the `$_` trick fails -- the shell seems to think that the value of the argument is `__bp_preexec_invoke_exec` rather than `foobar`. I found [this post](https://unix.stackexchange.com/questions/413144/bash-sleep-outputs-bp-preexec-invoke-exec) after a quick search but it wasn't very enlightening.

On another machine, using an older CMake (2.8), the first command only works if I do provide `-H.` _and_ add no space between `-B` and `foobar`. That in turn breaks the make invocation, which assigns `-Bfoobar` instead of `foobar` to `$_`.

Any ideas? :disappointed:

waldyrious:
Alright, after some investigation and a bunch of testing, I managed to narrow down the change to [CMake 3.13](https://cmake.org/cmake/help/v3.13/release/3.13.html#command-line), which introduced dedicated `-B` and `-S` options to specify the build and source directories, respectively. Versions prior to this do support the `cmake -B build_dir` formulation, but only if the build directory already exists; otherwise they have to use the `cmake -H. -Bbuild_dir` form to work.

Now, 3.13 is a fairly recent version of CMake (about a year old), so we probably need to wait a bit longer until it propagates to the various package repositories. (At the time of writing, Repology.org [reports](https://repology.org/project/cmake/information) 143 repositories with versions older than 3.13, and 96 with 3.13 or above.) That said, it's good news that we will eventually be able to simplify this example further :)

waldyrious:
Regarding the `$_` issue, it turns out there's an alternative that seems to work across all systems I've managed to test it on: `!#:2` instead of `$_`.

According to [this post](https://stackoverflow.com/a/25497459/266309) on StackOverflow and the official [documentation](https://www.gnu.org/software/bash/manual/html_node/Word-Designators.html), it refers to the word at index 2 (counting from 0) of whatever comes before the `!#` marker in the command line.

As an example, in `cmake -B foobar && make -C !#:2`, the first few words would be 0) `cmake`, 1) `-B` and 2) `foobar`, thus resulting in the command `cmake -B foobar && make -C foobar`.

Of course, as described above, CMake versions older than 3.13 require `-B` to be used with no space between the option and its argument, which would result in `-Bfoobar` as the second word of `cmake -H. -Bfoobar`. We _could_ use [yet another bash trick](https://www.gnu.org/software/bash/manual/html_node/Modifiers.html) and do `cmake -H. -Bfoobar && make -C !#:2:s/-B//`, but IMO that crosses the simplicity threshold and becomes too cryptic for use in a tldr page.

waldyrious:
Concluding the comments above, it is clear that this PR should not be merged for now; I'll close it for the time being, and maybe we can revisit the change in the future, when CMake 3.13 is more widespread.

sbrl:
On a Linux shell, that `$` in `$_` will need to be escaped, @waldyrious. This can be done in 2 ways:

```
echo '$_'; # With single quotes
echo \$_; # With a backslash
```

I wonder if this would work.

waldyrious:
On one of the Linux systems I tested this on, plain $_ worked (it just wasn't able to extract `foobar` from `-Bfoobar`, which was needed due to the CMake version available there).

On the other system (the one that produced the `__bp_preexec_invoke_exec` issue) I tried both escaping methods, but neither seemed to work: they simply passed the exact string "$_" to the second command, rather than substituting it with the appropriate argument from the first command. Are you able to reproduce a different behavior?

sbrl:
That's odd. I haven't actually tried myself - perhaps it's a compile-time enable feature or something?

waldyrious:
I have no idea, but in any case we can't use the trick yet, so I don't think it's worth investigating this further at this time.

PR #3600, opened 2019-11-21
git-show: add example for showing file contents

waldyrious:
This is a handy option which I have to look up how to do every now and then — most recently when seeking old versions of the Homebrew formula for cmake in the context of #3598.

### DIFF
--- a/pages/common/git-show.md
+++ b/pages/common/git-show.md
@@ -23,6 +23,6 @@
 
 `git show --oneline -s {{commit}}`
 
-- Show the contents of a file as it was at a given branch, tag or commit:
+- Show the contents of a file as it was at a given revision (e.g. branch, tag or commit):
 
-`git show {{commit}}:{{path/to/file}}`
+`git show {{revision}}:{{path/to/file}}`
--- a/pages/common/git-show.md
+++ b/pages/common/git-show.md
@@ -22,3 +22,7 @@
 - Show a commit's hash and message in a single line, suppressing the diff output:
 
 `git show --oneline -s {{commit}}`
+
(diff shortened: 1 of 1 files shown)


pages/common/git-show.md
```diff
@@ -22,3 +22,7 @@
 - Show a commit's hash and message in a single line, suppressing the diff output:
 
 `git show --oneline -s {{commit}}`
+
+- Show the contents of a file as it was at a given branch, tag or commit:
+
+`git show {{commit}}:{{path/to/file}}`
```
agnivade: Should it be {{branch|tag|commit}} then ?

waldyrious: I pondered the same. Technically any commit-ish [revision specification](https://git-scm.com/docs/gitrevisions) works here, not just branches, tags and commit hashes. I had thought about using `{{reference}}` as a generic term, but maybe `{{revision}}` is clearer. WDYT?

```
- Show the contents of a file as it was at a given revision (e.g. branch, tag or commit):

`git show {{revision}}:{{path/to/file}}`
```

agnivade: Sounds good.

sbrl: What about `{{commit_ish}}`? I think the official git docs use this language.

waldyrious: They use lots of different terms, including "revision" (the page I linked to is called "gitrevisions" 😄). So I'd rather go with that since it's more intuitive IMO.

sbrl: Sure, that works for me.

waldyrious: Updated. :+1:

PR #3603, opened 2019-11-22
Fix whitespace in Korean pages

mebeim:
Yet another Korean pages formatting PR. Trivial formatting change that can be merged right away.

---

Removed all unneeded whitespace with this script:

    for p in pages.ko/**/*; do echo $p; cat $p > tmp; awk '{$1=$1};1' tmp > $p; done; rm tmp

Plus some more minor manual fix for double unneeded newlines.


Proscream:
Thank you for your interest in Korean translation.

Issue #2986, opened 2019-05-07
Define a brand image and logo


sbrl:
GitHub have recently added a new feature to the repository settings known as _social preview_:

![image](https://user-images.githubusercontent.com/9929737/57339427-7dd1a580-7129-11e9-8978-86f5ef0c8a94.png)

[Here's the associated blog post](https://github.blog/2019-04-17-custom-open-graph-images-for-repositories/).

Should tldr-pages have a social preview image? If so, what image or screenshot should we use?

sbrl:
Yeah, my design skills are somewhat limited too. I guess we could use the [image on the homepage](https://raw.githubusercontent.com/tldr-pages/tldr/master/screenshot.png) as an interim solution?

agnivade:
Interesting. We have never really thought of any brand image or logo for tldr. I have zero design skills, but if somebody does, I am game for providing input.

waldyrious:
There's a nice graphic in [this blog post](https://laravel-news.com/tldr-pages) about tldr pages. I'm not sure if it's more useful than the screenshot in our homepage and README, but it is certainly less busy and easier to read. Maybe the post's author, @paulredmond, can offer it for us to use?

paulredmond:
> There's a nice graphic in this blog post about tldr pages. I'm not sure if it's more useful than the screenshot in our homepage and README, but it is certainly less busy and easier to read. Maybe the post's author, @paulredmond, can offer it for us to use?

We use https://thenounproject.com/ for icons and Eric Barnes owns the copyright for content on Laravel-news.com. My advice when considering using icons for branding/marks be careful to read the TOS in how you can use them.

Aracki:
I think the graphic @waldyrious mentioned is more appropriate than the picture with a lot of commands. No one will actually try to read something from social preview image?

I can help with a design/redesign.

waldyrious:
Thanks for the clarification and advice @paulredmond. I couldn't find that specific icon in the Noun Project, but in any case I agree that if we're to use this for our own branding, we might want to design the icon ourselves. @Aracki if you could make a proposal, that'd be great :) something simple, similar to the graphic linked above would work well IMO.

Aracki:
![image](https://user-images.githubusercontent.com/6129986/61576955-29f81b00-aae1-11e9-8fca-ae9cd8bcf02f.png)

Proposals here ^^

[...]

PR #3606, opened 2019-11-22
add logo

sbrl:
As part of #2986, we've now got a social preview and logo for tldr-pages.

I've set the social preview already, but GH doesn't have a native setting like GitLab does for logos - so you have to commit the file to your repository.

We should also consider adding the logo to the README.


waldyrious:
Looks really great :)

Some screenshots below.

|Organization page|
|-----------------|
|Screenshot 2019-11-24 at 19 19 35|

|Profile of an org member|
|------------------------|
|Screenshot 2019-11-24 at 19 18 55|

|Tweet|
|-----|
|Screenshot 2019-11-24 at 19 30 39

All of the above worked nicely, as expected. However, when I attempted to make a Facebook post to test the larger image, it didn't seem to have been picked up:
Screenshot 2019-11-24 at 19 32 30

Did you set that too, @sbrl? Or just the logo?

waldyrious:
Also, I'm wondering if we could change the `:books:` in the repository tagline with the actual Unicode character, 📚. That would probably improve the appearance of the tagline in the social preview cards, as shown above.

sbrl:
Ah, I forgot about the org-level logo! No, that wasn't me.

I'll update the description now.

Hrm, not sure where FB is getting that image from. Is it cached or something?

pxgamer:
Looks like the Facebook cache it their end. I've forced a refresh through the Developer panel, so all good now. :+1:

waldyrious:
Great, I can attest that the new image is now shown when making a Facebook post:

Screenshot from 2019-11-24 23-39-06

As for the emoji, I suspect you're right @sbrl. I used the macOS emoji insertion widget (`Ctrl+Cmd+Space`) which sometimes enters custom codepoints rather than the Unicode standard emoji characters. I do mean the actual unicode character 📚 (`U+1F4DA`). Can you try it out in the repo description and see if it stops appearing as `:books:` in the social preview?

pxgamer:
I've updated it to use the unicode character, seems to look good according to https://metatags.io, I'll refresh the Facebook developer dashboard again. 👍

schneiderl:
Looks great. 😄

sbrl:
Oh, no! Looks like GH automatically translate unicode chars into their emojis in the comments. That's why I got confused :P

PR #3609, opened 2019-11-22
a2disconf, a2dismod, a2dissite, a2enconf, a2enmod, a2ensite, a2query: add page

Rudloff:

### DIFF
--- a/pages/linux/a2disconf.md
+++ b/pages/linux/a2disconf.md
@@ -1,7 +1,7 @@
 # a2disconf
 
 > Disable an Apache configuration file on Debian-based OSes.
-> More information: <https://httpd.apache.org/>.
+> More information: <https://manpages.debian.org/buster/apache2/a2disconf.8.en.html>.
 
 - Disable a configuration file:
 
--- a/pages/linux/a2disconf.md
+++ b/pages/linux/a2disconf.md
@@ -1,6 +1,6 @@
 # a2disconf
 
-> Disable an Apache configuration file.
+> Disable an Apache configuration file on Debian-based OSes.
 > More information: <https://httpd.apache.org/>.
 
 - Disable a configuration file:
--- a/pages/linux/a2disconf.md
+++ b/pages/linux/a2disconf.md
@@ -9,4 +9,4 @@
 
 - Don't show informative messages:
 
-`sudo a2disconf -q {{configuration_file}}`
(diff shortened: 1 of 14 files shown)


pages/linux/a2disconf.md
```diff
@@ -0,0 +1,12 @@
+# a2disconf
+
+> Disable an Apache configuration file on Debian-based OSes.
+> More information: <https://httpd.apache.org/>.
```
sbrl: Is there a more specific link to the documentation about this command? Same for the other pages.

pages/linux/a2disconf.md
```diff
@@ -0,0 +1,12 @@
...
+
+- Disable a configuration file:
+
+`sudo a2disconf {{configuration_file}}`
```
sbrl: Perhaps this will read better?
```suggestion
`sudo a2disconf {{path/to/configuration_file}}`
```

....same for the all other instances.

pages/linux/a2disconf.md
```diff
@@ -0,0 +1,12 @@
...
+
+- Don't show informative messages:
+
+`sudo a2disconf -q {{configuration_file}}`
```
sbrl: Is there a long form of the `-q` option at all? In tldr-pages we prefer to use long options where possible.

Same goes for the other instances of short options.

pages/linux/a2disconf.md
Rudloff: Since this command was created for Debian, we could use their man page: https://manpages.debian.org/buster/apache2/a2disconf.8.en.html

pages/linux/a2disconf.md
Rudloff: This argument is not a path, but only a filename. It will then look for a file with that name in `/etc/apache2/conf-available/`.

pages/linux/a2disconf.md
Rudloff: I replaced it with `--quiet`.
`a2query` does not seem to have any long option though.

pages/linux/a2disconf.md
sbrl: Ah, that'd be perfect! That's do that.

pages/linux/a2disconf.md
sbrl: Ah, I see. Let's leave it as-is then

PR #3619, opened 2019-11-24
Add SVG logo and update PNG render to match

### LINKED ISSUE #2986: Define a brand image and logo

waldyrious:
Finally I managed to work on cleaning up the SVG of the logo for upload. Inkscape adds a _lot_ of cruft -- I trimmed the file down from 5.7 KiB to a mere 773 bytes!

Part of the simplification included properly calibrating the drawing, so that the geometry is now fully symmetric. I also tweaked the dark blue color ever-so-slightly from `#202a4a` to `#203050`, which IMO works better with the button color -- but happy to revert if you guys prefer the original blue.

The final file is so small that it can be entirely pasted below, and is even quite human-readable!
```svg
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="600" height="600">
  <defs>
    <linearGradient id="a">
      <stop offset="0" stop-color="#97d6d2"/>
      <stop offset="1" stop-color="#4d9182"/>
    </linearGradient>
    <linearGradient xlink:href="#a" id="b" x1="10%" y1="10%" x2="90%" y2="90%"/>
    <linearGradient xlink:href="#a" id="c" x1="90%" y1="90%" x2="10%" y2="10%"/>
  </defs>
  <rect width="100%" height="100%" x="0" y="0" fill="#203050"/>
  <rect width="460" height="460" x="70" y="70" rx="50" ry="50" fill="url(#b)"/>
  <rect width="380" height="380" x="110" y="110" rx="40" ry="40" fill="url(#c)"/>
  <text x="115" y="390" font-size="300" font-family="Sometype Mono" fill="#203050">
    $_
  </text>
</svg>
```

Closes #2986.


waldyrious:
I will add the full-size image as well (both SVG and PNG), for completeness. Please don't merge until then.

waldyrious:
Update: I have now added full-size banner / social preview the as well.

waldyrious:
Per @mebeim's [suggestion in #2986](https://github.com/tldr-pages/tldr/issues/2986#issuecomment-552253236), here are two versions of the logo using regular vs bold font:

regular &emsp; bold

regular &emsp; bold

regular &emsp; bold

I think the regular version is clearer and easier to read, especially at smaller sizes, where the `$` can become a bit too heavy an hard to discern. But let me know what you guys think!

waldyrious:
> Shall we merge this?

Yeah, I think the design itself has been discussed and agreed upon in #2986; since no objections have been raised in this PR, it should be safe to merge (and as always, any changes can be proposed in further PRs).

waldyrious:
By the way, who manages the [tldr-bot](https://github.com/tldr-bot) account? We could update its image as suggested [here](https://github.com/tldr-pages/tldr/issues/2986#issuecomment-549096181). The reaction was positive in that comment, but I can refine the design if you guys have suggestions. Otherwise, I can provide a high-res version of the image (and maybe add it to the [tldr-bot repo](https://github.com/tldr-pages/tldr-bot) as well).

agnivade:
> By the way, who manages the tldr-bot account?

That's me. I will update it.

waldyrious:
> Waldir, can you provide the image for the bot with a transparent background ? That looks better than a white background.

Yes, sure! I hadn't produced a finished version yet because I was unsure of the community's opinion, but I will provide it tonight when I get back home, as a PR to the tldr-bot repo.

agnivade:
Sure, go ahead.

waldyrious:
I noticed that the current image may not have sufficient padding to serve as a good icon in some places, especially those that make a round cut-out of the logo. I prepared a version with a bit more padding, which works better when cropped into a circle:

&emsp;&emsp;&emsp;

What do you guys think? Shall I replace the files in the PR?

mebeim:
@waldyrious what about just keeping the SVG version in the repo so that it can just be adapted for any occurrence where it's needed? I really like the round version though.

waldyrious:
> what about just keeping the SVG version in the repo

I'm not sure I follow. Isn't adding the SVG version to the repo precisely what this PR is doing already? Do you mean we should remove the PNG version? Please clarify.

mebeim:
@waldyrious yes, I meant removing the PNGs and just keeping the SVGs.

waldyrious:
I think the PNG was meant to be used as the source for the repository social preview configuration. Quoting @sbrl from #3606:

> GH doesn't have a native setting [...] for logos - so you have to commit the file to your repository.

So if we remove the PNG I think that may stop working.

But that aside, I don't see what we would gain by removing the PNG. Note that it is merely a convenience rendering of the SVG file, so both files should look exactly the same. (Of course, we can grab the SVG and make changes if we need to adapt the logo to some specific end, but that does not interfere with the storage of an in-sync rendering of the vector logo as a raster image.)

My question is simply how much padding the logo should have _by default_, regardless of the file format. I think it makes more sense to have sufficient padding for the logo to work well if circle-cropped, so what I was suggesting was to change both the SVG and the PNG files in this PR to have the extra padding.

mebeim:
@waldyrious GitHub does not need any file inside the repo for the social preview. Having just an SVG with *a good amount* of padding so that it can be easily rendered and cropped as needed seems like the simplest choice to me.

agnivade:
We may want to use the png for other things too. It doesn't hurt to keep it IMO.

mebeim:
If you guys think it's better to keep PNGs then let's keep them. Mine was just a suggestion.

waldyrious:
Any objections to adding more padding to the images as suggested above? I'll keep the square background so the icon can adapt to any shape.

sbrl:
I'd keep a copy with more padding, and keep the PNGs.

While some platforms need the extra padding, others don't.

waldyrious:
> I'd keep a copy with more padding

By copy do you mean having two SVG files, one with extra padding and one without (and two PNGs to match them)? If so, I think that's ooverkill, especially for files that will be at the root of the repo. Cropping an image too remove excess padding is quite easy; the same cannot be said of extending it. So my proposal would be too keep a single SVG and a single PNG, with sufficient padding to fit within any common shape for icons (square, circle, [squircle](https://en.wikipedia.org/wiki/Squircle), etc).

sbrl:
Sure, that works @waldyrious

waldyrious:
Ok, I've updated the images to have the extra padding. I also defaulted the font in the SVG files to monospace, so they at least bear some resemblance to the target appearance when rendered in systems without the fonts installed. On that note, I'm considering embedding the font in the SVG file, or adding the font files to this repo so the SVG files can refer to them directly. But that can be done as a separate optimization step, so I'll do it in a later PR. This one should be ready to merge :)

PR #3620, opened 2019-11-25
git-status: improve main description & add example

waldyrious:

### DIFF
--- a/pages/common/git-status.md
+++ b/pages/common/git-status.md
@@ -1,6 +1,7 @@
 # git status
 
-> Show the index (changed files).
+> Show the changes to files in a git repository.
+> Lists changed, added and deleted files compared to the currently checked-out commit.
 > More information: <https://git-scm.com/docs/git-status>.
 
 - Show changed files which are not yet added for commit:
@@ -10,3 +11,7 @@
 - Give output in short format:
 
 `git status -s`
+
+- Don't show untracked files in the output:
+
+`git status --untracked-files=no`


pages/common/git-status.md
```diff
@@ -10,3 +11,7 @@
 - Give output in short format:
 
 `git status -s`
+
+- Don't show untracked files in the output:
```
agnivade: Frankly, I have never used it and untracked files in the output never really bothered me. Since there is extra headroom for more examples, I am OK, but do we really need it ?

waldyrious: It is particularly helpful when handling projects that generate a lot of files in a build process, or clone submodules, that sort of thing. I need this every now and then, usually at a point when the repo in question ends up with hundreds of untracked files, which obscure the actual modified files.

I think it's worth adding not just because the page has lots of room, but also because its short form is quite unintuitive (`-u no` rather than, say, `-U`) so it's likely that people (e.g. me 😄) will forget the exact syntax.

agnivade: Build artifacts should be added in `.gitignore`. And submodules should be only a one time thing until a .gitsubmodules file get generated.

einverne: I never used this option either. But good to know it exist. I agree with @agnivade , all files do not need to track should be in `.gitignore`.

waldyrious: Sure, `.gitignore` is the ideal place for such files, but sometimes that's not feasible, e.g. for large third-party repos that one may be just submitting a one-off contribution to.

As an example, the last time I had the need to do this was while working on an issue for a formula in the Homebrew repo, and IIRC checking out multiple snapshots of the repo during my experiments resulted in the working directory having files that git assumed not to be tracked, but that simply were added in more recent versions of the repo. In such a high-traffic project, that amounted to dozens or even hundreds of files, but those files should not be in a `.gitignore` file.

So again, I agree this is not necessarily a very common need, but there _are_ cases where it is useful, and since the page is so short, I don't think there's any downside in having the example present for quick reference.

PR #3482, opened 2019-10-26
gedit: add page

nolasconapoleao:
- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #3622, opened 2019-11-25
gedit: streamline page

waldyrious:
- Remove the second line of the description, which didn't add useful info (as discussed [previously](https://github.com/tldr-pages/tldr/pull/3540#discussion_r349377075) in the PR adding the Portuguese version of the page)
- Reorder two examples to provide a more logical sequence
- Tweak some descriptions for clarity and correctness

/cc @nolasconapoleao and @mebeim (OP and reviewer of the PR #3482, which introduced this page).

### DIFF
--- a/pages/linux/gedit.md
+++ b/pages/linux/gedit.md
@@ -1,7 +1,6 @@
 # gedit
 
-> Text editor for the GNOME Desktop.
-> It can be used to create and edit all kinds of text files.
+> Text editor of the GNOME Desktop project.
 
 - Open a text file:
 
@@ -11,10 +10,10 @@
 
 `gedit {{file1 file2 ...}}`
 
-- Display list of supported encodings:
-
-`gedit --list-encodings`
-
-- Open a text file with the specified encoding:
+- Open a text file with a specific encoding:
 
 `gedit --encoding={{UTF-8}} {{path/to/file}}`
+
+- Display a list of the supported encodings:
+
+`gedit --list-encodings`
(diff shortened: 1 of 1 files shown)


pages/linux/gedit.md
```diff
@@ -11,10 +10,10 @@
...
 `gedit {{file1 file2 ...}}`
 
-- Display list of supported encodings:
+- Open a text file with a specific encoding:
 
-`gedit --list-encodings`
+`gedit --encoding={{UTF-8}} {{path/to/file}}`
 
-- Open a text file with the specified encoding:
+- Display a list of the supported encodings:
```
sbrl: If we want to use `the` here, it might sound better like this:
```suggestion
- Display a list of the encodings supported:
```

waldyrious: Makes sense, thanks! That was clearly my Portuguese creeping in! 😄

waldyrious: After re-reading both variants, I think I prefer the the-less version:

```suggestion
- Display a list of supported encodings:
```

WDYT?

mebeim: The current version ("supported encodings") sounds better than @sbrl's suggestion to me. Removing the article altogether though also seems like a good idea.

sbrl: Yeah, let's remove the article.

waldyrious: Done :)

waldyrious:
When merging, please fix the commit message that's currently saying "to make to provide" and should instead be just "to provide".

PR #3623, opened 2019-11-25
mlr: update page

vapniks:
Add example of --prepipe option and filter verb with -S option (omitting this option can cause errors which are hard to track down if you forget).

### DIFF
--- a/pages/common/mlr.md
+++ b/pages/common/mlr.md
@@ -26,3 +26,7 @@
 - Receive JSON and format the output as vertical JSON:
 
 `echo '{"hello":"world", "foo":"bar"}' | mlr --ijson --ojson --jvstack cat`
+
+- Filter lines of compressed CSV file (-S ensures numbers are treated as strings to prevent errors):
+
+mlr --prepipe 'gunzip' --csv filter -S '${{fieldName}} =~ "{{regexp}}"' {{example.csv.gz}}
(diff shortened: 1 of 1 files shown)


pages/common/mlr.md
```diff
@@ -26,3 +26,7 @@
 - Receive JSON and format the output as vertical JSON:
 
 `echo '{"hello":"world", "foo":"bar"}' | mlr --ijson --ojson --jvstack cat`
+
+- Filter lines of compressed CSV file (-S ensures numbers are treated as strings to prevent errors):
```
sbrl: Perhaps this will read better?
```suggestion
- Filter lines of a compressed CSV file (`-S` ensures numbers are treated as strings to prevent errors):
```

vapniks: yes, that's better

agnivade: I would suggest something like: Filter lines of a compressed CSV file, treating numbers as strings

sbrl: Marking as unresolved, since the change hasn't actually been made yet.

agnivade: @vapniks, what do you think of my suggested change ? It gets rid of the parentheses and reads smoother IMO.

vapniks: I just added a new commit for that change.

PR #3630, opened 2019-11-27
e2label: add page

einverne:


pages/linux/e2label.md
```diff
@@ -0,0 +1,7 @@
+# e2label
+
+> Change the label on an ext2/ext3/ext4 filesystem.
+
+- Change the volume label on the ext2/ext3/ext4 filesystem:
```
sbrl: Perhaps this will read better?
```suggestion
- Change the volume label on a specific partition:
```

pages/linux/e2label.md
```diff
@@ -0,0 +1,7 @@
+# e2label
+
+> Change the label on an ext2/ext3/ext4 filesystem.
+
+- Change the volume label on a specific partition:
+
+`e2label {{/dev/sda1}} {{"my-label"}}`
```
agnivade: What do you think about changing it my-label to label_name ? That matches with the other token names that we have.

einverne: Done.

pages/linux/e2label.md
zdroid: I would still change this to "a specific ext partition" or perhaps "a specific ext-type partition".

einverne: @zdroid done.

zdroid:
Should we change the `.gitignore` to exclude the files I deleted? (especially `package-lock.json`, as it was failing the Travis build)

agnivade:
I think we should commit the package-lock.json file, as it helps to lock down the dependency hashes.

zdroid:
@agnivade On the contrary, it causes the problems with Travis build.

pxgamer:
It looks like that issue was because that lock file was outdated (using `markdownlint-cli @ 0.15.0`, and 6af2c9b9b3af3e74b68a25cfdef7cf3633b12d85 updated our dependencies to require `0.19.0`.

agnivade:
Yes, that's right. Let's get that file committed. package-lock.json should be a part of every Node project.

PR #3583, opened 2019-11-17
multiple pages : Add Korean Translation

JJByun:
chsh.md
cksum.md
clamscan.md
clang.md
clear.md
clementine.md
cloc.md
clockwork-cli.md

- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #3633, opened 2019-11-27
Add missing trailing newline to some Korean pages

mebeim:
Missing newlines in commands added in #3583. Well, well, well, what can I say. We really need to merge #3576 after this :sweat_smile:


Proscream:
Oops. I thought I looked at the pr in detail, but I didn't catch it.

PR #3644, opened 2019-12-05
git-ls-tree, git-rev-list, git-rev-parse: add pages

BeLi4L:

### DIFF
--- a/pages/common/git-ls-tree.md
+++ b/pages/common/git-ls-tree.md
@@ -11,6 +11,6 @@
 
 `git ls-tree -r {{commit_hash}}`
 
-- List the contents of the tree {{N}} commits ago and show only filenames:
+- List only the filenames of the tree on a commit:
 
-`git ls-tree --name-only {{HEAD}}~{{N}}`
+`git ls-tree --name-only {{commit_hash}}`
--- a/pages/common/git-ls-tree.md
+++ b/pages/common/git-ls-tree.md
@@ -5,8 +5,12 @@
 
 - List the contents of the tree on a branch:
 
-`git ls-tree {{master}}`
+`git ls-tree {{branch_name}}`
 
-- List the contents of the tree on a commit, recurse into subtrees, and show only filenames:
+- List the contents of the tree on a commit and recurse into subtrees:
 
-`git ls-tree -r --name-only {{c39ab84d}}`
+`git ls-tree -r {{commit_hash}}`
+
+- List the contents of the tree on a commit and show only filenames:
+
+`git ls-tree --name-only {{HEAD~3}}`
--- a/pages/common/git-ls-tree.md
+++ b/pages/common/git-ls-tree.md
@@ -7,7 +7,7 @@
 
 `git ls-tree {{branch_name}}`
 
-- List the contents of the tree on a commit and recurse into subtrees:
+- List the contents of the tree on a commit, recursing into subtrees:
 
 `git ls-tree -r {{commit_hash}}`
 
--- a/pages/common/git-ls-tree.md
+++ b/pages/common/git-ls-tree.md
@@ -11,6 +11,6 @@
 
 `git ls-tree -r {{commit_hash}}`
 
-- List the contents of the tree on a commit and show only filenames:
+- List the contents of the tree {{N}} commits ago and show only filenames:
 
-`git ls-tree --name-only {{HEAD~3}}`
+`git ls-tree --name-only {{HEAD}}~{{N}}`
--- a/pages/common/git-ls-tree.md
+++ b/pages/common/git-ls-tree.md
@@ -1,0 +1,12 @@
+# git ls-tree
+
+> List the contents of a tree object.
+> More information: <https://git-scm.com/docs/git-ls-tree>.
+
+- List the contents of the tree on a branch:
+
+`git ls-tree {{master}}`
+
+- List the contents of the tree on a commit, recurse into subtrees, and show only filenames:
+
+`git ls-tree -r --name-only {{c39ab84d}}`
--- a/pages/common/git-rev-list.md
+++ b/pages/common/git-rev-list.md
@@ -3,7 +3,7 @@
 > Lists commits in reverse chronological order.
 > More information: <https://git-scm.com/docs/git-rev-list>.
 
-- List all commits on current branch:
+- List all commits on the current branch:
 
 `git rev-list {{HEAD}}`
 
--- a/pages/common/git-rev-list.md
+++ b/pages/common/git-rev-list.md
@@ -1,6 +1,6 @@
 # git rev-list
 
-> Lists commits in reverse chronological order.
+> List revisions (commits) in reverse chronological order.
 > More information: <https://git-scm.com/docs/git-rev-list>.
 
 - List all commits on the current branch:
--- a/pages/common/git-rev-list.md
+++ b/pages/common/git-rev-list.md
@@ -1,0 +1,16 @@
+# git rev-list
+
+> Lists commits in reverse chronological order.
+> More information: <https://git-scm.com/docs/git-rev-list>.
+
+- List all commits on current branch:
+
+`git rev-list {{HEAD}}`
+
+- List commits more recent than a specific date, on a specific branch:
+
+`git rev-list --since={{'2019-12-01 00:00:00'}} {{master}}`
+
+- List all merge commits on a specific commit:
+
+`git rev-list --merges {{commit}}`
--- a/pages/common/git-rev-parse.md
+++ b/pages/common/git-rev-parse.md
@@ -3,10 +3,14 @@
 > Parses a revision.
 > More information: <https://git-scm.com/docs/git-rev-parse>.
 
-- Get the absolute path to the root directory:
+- Get the commit hash of a branch:
 
-`git rev-parse --show-toplevel`
+`git rev-parse {{branch_name}}`
 
 - Get the current branch name:
 
 `git rev-parse --abbrev-ref {{HEAD}}`
+
+- Get the absolute path to the root directory:
+
+`git rev-parse --show-toplevel`
--- a/pages/common/git-rev-parse.md
+++ b/pages/common/git-rev-parse.md
@@ -1,0 +1,12 @@
+# git rev-parse
+
+> Parses a revision.
+> More information: <https://git-scm.com/docs/git-rev-parse>.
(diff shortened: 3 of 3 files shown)


pages/common/git-ls-tree.md
```diff
@@ -0,0 +1,12 @@
...
+> List the contents of a tree object.
+> More information: <https://git-scm.com/docs/git-ls-tree>.
+
+- List the contents of the tree on a branch:
+
+`git ls-tree {{master}}`
+
+- List the contents of the tree on a commit, recurse into subtrees, and show only filenames:
+
+`git ls-tree -r --name-only {{c39ab84d}}`
```
waldyrious: I'd prefer using a descriptive name here, like `{{commit_hash}}`.

Also, we usually try to avoid introducing too many new options in a single example, to make sure the pages remain beginner-friendly.

I would suggest splitting this example into two: one for the `-r` option, and one for the `--name-only` option (maybe use a different object for the third example, e.g. a tag or a relative reference like `HEAD~3`.

pages/common/git-ls-tree.md
```diff
@@ -0,0 +1,12 @@
+# git ls-tree
+
+> List the contents of a tree object.
+> More information: <https://git-scm.com/docs/git-ls-tree>.
+
+- List the contents of the tree on a branch:
+
+`git ls-tree {{master}}`
```
waldyrious: I'd have a slight preference to use `{{branch_name}}` instead of `{{master}}` here, but I'm open to being convinced otherwise.

BeLi4L: Fixed in 35035345650e07534a454774f0aa987ca60c6124

pages/common/git-ls-tree.md
BeLi4L: Sure ;)

Done in 35035345650e07534a454774f0aa987ca60c6124

pages/common/git-ls-tree.md
```diff
@@ -5,8 +5,12 @@
...
 
-- List the contents of the tree on a commit, recurse into subtrees, and show only filenames:
+- List the contents of the tree on a commit and recurse into subtrees:
 
-`git ls-tree -r --name-only {{c39ab84d}}`
+`git ls-tree -r {{commit_hash}}`
+
+- List the contents of the tree on a commit and show only filenames:
+
+`git ls-tree --name-only {{HEAD~3}}`
```
zdroid: I would change `HEAD~3` to `HEAD~N` and note it shows the situation N commits ago or something like that.

BeLi4L: 👍 

a2c3e613135cec75a31027ea1e7dc6535dd1c46a

zdroid: I would rather not put it as {{{HEAD}}~{{N}}. Maybe just HEAD~{{N}}?

/cc @waldyrious what would be the proper syntax here?

waldyrious: @BeLi4L maybe wait for a consensus to emerge in these discussions before applying the changes, otherwise you may need to do a lot of back and forth 😅 

As for the syntax, I'm afraid we're making things too complex. The syntax for relative references is not something that this page is responsible for introducing. My suggestion is to simply include the `3` in the command description, and use `{{HEAD~3}}` as the token, rather than generalize with `N`. WDYT?

pages/common/git-ls-tree.md
```diff
@@ -0,0 +1,16 @@
+# git ls-tree
+
+> List the contents of a tree object.
+> More information: <https://git-scm.com/docs/git-ls-tree>.
+
+- List the contents of the tree on a branch:
+
+`git ls-tree {{branch_name}}`
+
+- List the contents of the tree on a commit and recurse into subtrees:
```
waldyrious:
```suggestion
- List the contents of the tree on a commit, recursing into subtrees:
```

pages/common/git-ls-tree.md
waldyrious: Or maybe we could bypass this conundrum altogether by using another type of reference, like a tag (e.g. `{{v1.2.3}}`.

pages/common/git-rev-list.md
```diff
@@ -0,0 +1,16 @@
+# git rev-list
+
+> Lists commits in reverse chronological order.
```
waldyrious: This description may make it seem like the "rev" in the command name stands for "reverse". I'd suggest incorporating "revision" in the sentence, to make sure we avoid conveying that misconception.

pages/common/git-rev-list.md
```diff
@@ -0,0 +1,16 @@
...
+> More information: <https://git-scm.com/docs/git-rev-list>.
+
+- List all commits on current branch:
```
waldyrious:
```suggestion
- List all commits on the current branch:
```

pages/common/git-rev-list.md
```diff
@@ -0,0 +1,16 @@
...
+
+`git rev-list {{HEAD}}`
```
waldyrious: `HEAD` is assumed by default, right? I'd simply remove the token from the command.

pages/common/git-rev-parse.md
```diff
@@ -0,0 +1,12 @@
+# git rev-parse
+
+> Parses a revision.
```
waldyrious: Can we use something more descriptive than "parse"? Maybe something relating to exposing/displaying information (or "metadata") about a revision, WDYT?

pages/common/git-rev-parse.md
```diff
@@ -0,0 +1,12 @@
...
+> More information: <https://git-scm.com/docs/git-rev-parse>.
+
+- Get the absolute path to the root directory:
```
waldyrious: Not sure this should be the first example since the root directory doesn't typically change between revisions. WDYT about moving it to the end of the page?

pages/common/git-rev-parse.md
```diff
@@ -0,0 +1,12 @@
...
+
+`git rev-parse --show-toplevel`
+
+- Get the current branch name:
+
+`git rev-parse --abbrev-ref {{HEAD}}`
```
waldyrious: As mentioned in another comment, if HEAD is assumed by default, maybe we can simply omit it from the command.

On this note, I'd say this page is missing an example for addressing a specific (i.e. non-head) revision. Can you please add one?

pages/common/git-ls-tree.md
zdroid: `{{HEAD~3}}` simply seems too hard coded for an example. Maybe it could be given without braces.

pages/common/git-rev-list.md
BeLi4L: Unfortunately there's no default :/

pages/common/git-rev-parse.md
BeLi4L: I agree it's not descriptive enough. The git documentation is even more obscure: `Pick out and massage parameters` (see https://git-scm.com/docs/git-rev-parse)

If you have a better wording, I'm open to suggestions ;)

pages/common/git-rev-parse.md
BeLi4L: HEAD isn't the default :crying_cat_face: 

And sure, I'll add another example

BeLi4L: Done in 3923a26b0f1ffe04e37f7fb476d1add4d92bdf87

pages/common/git-rev-parse.md
BeLi4L: Done in 3923a26b0f1ffe04e37f7fb476d1add4d92bdf87

pages/common/git-rev-list.md
BeLi4L: Good point. See 3dc463b5a70d58485f5e7a87387563bc9302e854

pages/common/git-ls-tree.md
BeLi4L: Or we could simply keep `{{commit_hash}}`.

People will probably try any ref anyway, what do you think?

pages/common/git-rev-list.md
waldyrious: Thanks for confirming.

pages/common/git-rev-parse.md
waldyrious: Geez, sometimes it feels like they do it on purpose to make things hard to understand! 😅

Would `Display metadata related to specific revisions` work for you?

pages/common/git-rev-parse.md
waldyrious: 🙏 Thanks!

pages/common/git-rev-parse.md
BeLi4L: Sure, see 98a60c0d11b1013b8cb92f0720abb20674fc9b75

pages/common/git-ls-tree.md
waldyrious: Well, since we already have an example using a branch name, and one using a commit hash, I think a different type of reference for this one would be useful. I get the reluctance regarding the potential complexity of using a relative reference, but I wonder if you guys have similar reservations about using a tag as I suggested above.

BeLi4L: All these solutions are OK for me, just tell me what you prefer ;)

waldyrious: My preference is using a tag. Looking forward to @zdroid's thoughts.

zdroid: A `{{commit_hash}}` sounds the best to me out of all proposed solutions. I say go with it.

BeLi4L: Are you OK with d3cccfffc26bea964548c255666f92162627ac73?

waldyrious: @zdroid I'd like to know what you dislike about using a tag, in light of my comments above.

sbrl: Isn't the term commit-ish used in the man pages?

BeLi4L: Any news on this? I can use `commit-ish` if you're all OK with that

zdroid: The current version is fine with me. Maybe @sbrl wants to use "commit-ish" instead, both work for me.

sbrl: I don't mind. Either work for me.

PR #3516, opened 2019-10-30
ocamlc: add page

mira01:
- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).

PR #3646, opened 2019-12-05
ocamlc: add page (closes #3516)

zdroid:
~~I'm not sure why this fails the trailing newline test, as I wrote this in Atom with enabled editorconfig.~~ It's not this page making the problem

Also it seems no-reply email doesn't add co-authors the way the real email does. I found @mira01's email in API, but since it isn't public, I would rather not add it publicly.


sbrl:
Can't you edit the original pull request as a maintainer now, @zdroid? Why the new PR?

Also, if you're going to create a new PR I recommend creating a new branch based on the existing branch @mira01 has created. That way you preserve authorship information.

zdroid:
@sbrl You're completely right, I forgot about that option. :hushed:

PR #3654, opened 2019-12-10
Rework PDF script

zdroid:
Highlights:
- matched the coding style for all files in `scripts/pdf/` with the settings provided in `.editorconfig`
- removed duplicate CSS, by using `basic.css` as basic CSS (:laughing:) and themes as layers on top of it
- matched CSS and HTML with general standards for those
- added PDF-related files to `.gitignore`

There may be more to do:
- remove the color themes as they really seem unesthetic
- add a note on how to make own styles

(although I'd keep the last two for some future PR)


scripts/pdf/render.py
```diff
@@ -15,108 +15,111 @@
...
 
-    oslist.sort()
+  oslist.sort()
 
-    #Required strings to create intermediate HTML files
-    header = "<html><head><link rel=stylesheet type=text/css href=" + colorscheme + ".css></head><body>\n"
-    footer = "</body></html>"
-    title_content = "<h1 class=titlemain>tldr pages</h1><h4 class=titlesub>Simplified and community driven man pages</h4></body></html>"
+  # Required strings to create intermediate HTML files
+  header = '<!DOCTYPE html><html><head><link rel="stylesheet" href="basic.css">'
```
agnivade: Wait, so we weren't using the basic.css at all ? But the file seems to be already there ?

zdroid: We were, but it was treated as a color scheme. So every scheme was included completely, as a file for itself. That way we had a lot of duplicate styles in 3 files. Another solution as I said in the PR description is to keep it as one file, but to remove our color schemes as they're really bad-looking.

agnivade: Nope, this looks fine. I will review when I have some time.

sbrl:
Nice one, @zdroid!

Suggestion: Create a `requirements.txt` file, so that you can do `pip3 install -r requirements.txt`

zdroid:
@sbrl here you go!

scripts/pdf/render.py
agnivade: While we are here, let's also add a "meta charset=utf-8"

PR #3657, opened 2019-12-15
Rework npm configuration

zdroid:
This PR does two things:

1. adds`package-lock.json`

2. removes all repo-related information from `package.json` for multiple reasons
   * none of this is actually required
   * `tldr` package name is already used by tldr's node client
   * provided version is misleading
   * package metadata is unsuitable for future changes (e.g. description, repository, license, homepage)


package.json
```diff
@@ -19,9 +16,5 @@
       "pre-commit": "npm test"
     }
   },
-  "repository": "tldr-pages/tldr",
```
agnivade: I am not sure why these are not required. Seemed helpful to me.

zdroid: What is its use? If someone needs any metadata, it's all available in tldr-node-client's package.json.

agnivade: This metadata is for this repo. I agree that the name clashes with the node client. In that case, I think we can change it to tldr-pages, and remove the version. And keep everything else. Even if there are no downstream consumers of this package, keeping this metadata doesn't do any harm.

I would prefer to keep it. That said, if you feel strongly about it, and others agree, I won't be stopping you :).

zdroid: We have all that information already stored in node client's [`package.json`](https://github.com/tldr-pages/tldr-node-client/blob/master/package.json). It's also publicly available from there, so we're just keeping duplicate metadata. That was my main motivation.

package.json
```diff
@@ -1,7 +1,4 @@
 {
-  "name": "tldr",
-  "version": "1.0.0",
-  "description": "Simplified, community-driven man pages",
```
sbrl: Not sure that these should be removed. Perhaps we could update them to be different to the tldr-node-client?

I believe removing these will cause npm to get upset and warn you.

zdroid: It won't, since neither is required. I tested this on my own PC and it seems no warning are given. :open_mouth:

sbrl: Ah, that must have changed then. I'm still uneasy about removing them though.

pxgamer: I think as long as `private` is set to `true`, then these won't flag an error. They're only required for publishing to a package repository afaik.

sbrl: True @pxgamer, but I think the metadata here is valuable to keep - even if we rewrite it.

zdroid:
Any new thoughts on this? Since we've changed the license, there is even more incorrect information in that file than before.

zdroid:
I split this PR in two, so the `package-lock.json` part can be merged ASAP. #3733

agnivade:
Nothing new. We can change the license now. But my old comment still stands.

package.json
sbrl: I agree with @agnivade here. It's not harmful to keep the information.

zdroid: It's not harmful if the information is correct. However, here at least 3 tags have incorrect information: `description`, `version`, `license`.

sbrl: True, so let's update them instead of deleting them?

zdroid: I'm for deleting `version` in any case, unless we are going to officially publish the pages on npm.

zdroid:
Let's review the reasons for changing the data once again:
1. `tldr` name is already used by our node client
2. `version` is essentially useless, as this package has never been versioned nor published
3. `description` and `license` are outdated
4. `author` and `repository` are also useless for private packages
5. `homepage` points to tldr-pages.github.io instead of tldr.sh

All the problematic "missing" metadata can easily be found in our tldr-node-client's [`package.json`](https://github.com/tldr-pages/tldr-node-client/blob/master/package.json). You can see we are just keeping duplicate information (most of it being incorrect).

package.json
mebeim: Agree with @sbrl let's fix what's wrong and delete what cannot be fixed. I agree that the `version` field does not make sense in our case, but the others can stay.

mebeim:
@zdroid here's my take on your points:

1. I think we can remove the `name` since as you said it conflicts with the Node CLI, and it's not useful anyway for this kind of package that is not published.
2. True, `version` seems useless, we can remove that IMHO.
3. Since `description` and `license` are outdated, we should update those, but not remove them.
4. We could remove `author` as the project has multiple authors, we should probably hear what @rprieto has to say since he's the creator of the project and that's his name. We should keep `repository` as it does no harm and provides the repo slug.
5. The `homepage` field should indeed be updated to point to tldr.sh.

zdroid:
I've reverted some of the changes. I used https://docs.npmjs.com/files/package.json#license to write the contents of `license` field.

rprieto:
@mebeim @zdroid Hi, thanks for the ping! Your changes look good.

As for the `author` field, I'm very grateful many of you have stepped up and worked hard to transform an idea into a well maintained project! Please feel free to update it to anything more meaningful!

PR #3653, opened 2019-12-10
magick: add page

redreceipt:
- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #3661, opened 2019-12-17
magick: Fix title & add better description

sbrl:
Ref #3653.

I've also updated the description, since I didn't get a chance to comment on it. `magick` is actually only available in ImageMagick v7+, which isn't yet available on Ubuntu (why? it's been out long enough).

### DIFF
--- a/pages/common/magick.md
+++ b/pages/common/magick.md
@@ -1,6 +1,7 @@
-# ImageMagick
+# magick
 
 > Create, edit, compose, or convert bitmap images.
+> ImageMagick version 7+. See `convert` for versions 6 and below.
 > More information: <https://imagemagick.org/>.
 
 - Convert file type:


pages/common/magick.md
```diff
@@ -1,6 +1,7 @@
-# ImageMagick
+# magick
 
 > Create, edit, compose, or convert bitmap images.
+> ImageMagick version 7+. See `convert` for versions 6 and below.
```
zdroid: Perhaps this should be `tldr convert` or `convert tldr page`, since I guess . What do you think?

sbrl: I'm not sure what you're getting at. This isn't an example - it's a description line with a see also - e.g. like `arch`, `lsmod`, etc.

zdroid: Ah okay, nevermind. I understood what you meant now.

PR #3668, opened 2019-12-20
pdfimages: add page

dcervenkov:

Not sure what the last point means, the rest should be OK.

### DIFF
--- a/pages/common/pdfimages.md
+++ b/pages/common/pdfimages.md
@@ -4,16 +4,16 @@
 
 - Extract all images from a PDF file and save them as PNGs:
 
-`pdfimages -png {{PDF-file}} {{image-root}}`
+`pdfimages -png {{path/to/file.pdf}} {{filename_prefix}}`
 
 - Extract images from pages 3 to 5:
 
-`pdfimages -f {{3}} -l {{5}} {{PDF-file}} {{image-root}}`
+`pdfimages -f {{3}} -l {{5}} {{path/to/file.pdf}} {{filename_prefix}}`
 
 - Extract images from a PDF file and include the page number in the output filenames:
 
-`pdfimages -p {{PDF-file}} {{image-root}}`
+`pdfimages -p {{path/to/file.pdf}} {{filename_prefix}}`
 
 - List information about all the images in a PDF file:
 
-`pdfimages -list {{PDF-file}}`
+`pdfimages -list {{path/to/file.pdf}}`
--- a/pages/common/pdfimages.md
+++ b/pages/common/pdfimages.md
@@ -1,0 +1,20 @@
+# pdfimages
+
+> Utility for extracting images from PDFs.
+
+- Extract all images from a PDF file and save them in PNG format:
+
+`pdfimages -png {{PDF-file}} {{image-root}}`
+
+- Extract images from a PDF file from pages 3-5:
+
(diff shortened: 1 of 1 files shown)


pages/common/pdfimages.md
```diff
@@ -0,0 +1,20 @@
+# pdfimages
+
+> Utility for extracting images from PDFs.
+
+- Extract all images from a PDF file and save them in PNG format:
```
sbrl: Perhaps this will read better?
```suggestion
- Extract all images from a PDF file and save them as PNGs:
```

pages/common/pdfimages.md
```diff
@@ -0,0 +1,20 @@
...
+
+`pdfimages -png {{PDF-file}} {{image-root}}`
+
+- Extract images from a PDF file from pages 3-5:
```
sbrl: Perhaps this will read better?
```suggestion
- Extract images from pages 3 to 5:
```

pages/common/pdfimages.md
```diff
@@ -0,0 +1,20 @@
...
+
+`pdfimages -f 3 -l 5 {{PDF-file}} {{image-root}}`
```
sbrl: Perhaps the numbers could be in `{{tokens}}`?
```suggestion
`pdfimages -f {{3}} -l {{5}} {{PDF-file}} {{image-root}}`
```

pages/common/pdfimages.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Extract images from a PDF file and include the page number in their names:
```
sbrl: Perhaps this will read better?
```suggestion
- Extract images from a PDF file and include the page number in the output filenames:
```

pages/common/pdfimages.md
```diff
@@ -0,0 +1,20 @@
...
+
+`pdfimages -p {{PDF-file}} {{image-root}}`
```
sbrl: Perhaps this will read better?
```suggestion
`pdfimages -p {{path/to/file.pdf}} {{path/to/output_directory}}`
```

Same goes for the other examples.

pages/common/pdfimages.md
zdroid: I disagree with you on this one, but :+1: for the other suggestions.

sbrl: Why do you disagree/

zdroid: Oh... I thought it was related to PDF files and not the generated files. :+1: for this one too.

pages/common/pdfimages.md
dcervenkov: Thanks for all the suggestions! I've committed all of them, except this one. The last parameter here is not a path to the output directory, but rather a name root, that all of the resulting filenames will share. I.e., running the command
```
pdfimages -png file.pdf test
```
will produce files
```
test-001.png
test-002.png
...
```
rather than saving them in a directory named `test`.

I agree it is not readily understandable from the token `image-root`, but I can't think of anything better and that's the term the program and its man page uses.

sbrl: Ah, I see! Thanks for clarifying. What about `filename_prefix` then?

dcervenkov: Good idea; implemented.

PR #3673, opened 2019-12-20
README: Fix description of C client

sbrl:
Ref https://github.com/tldr-pages/tldr-cpp-client/issues/20


zdroid:
Once the repository gets renamed, we should also change the link to it. What do you think about doing both in this PR?

sbrl:
Not sure about that @zdroid. It sounded complicated to do the repo rename - we should do that in a separate PR.

zdroid:
It's not really complicated at all. There are no conflicts with any other repositories, so GitHub will make a repository redirect without any problem.

sbrl:
@zdroid: https://github.com/tldr-pages/tldr-cpp-client/issues/20#issuecomment-246769596

agnivade:
Let's merge this for now.

PR #3681, opened 2019-12-22
Add Spanish as an example of valid locale tag

ivanhercaz:
This PR is to add Spanish as an example of valid locale tag. Do you know any other file in which I should add it?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,6 +88,7 @@ Some examples of valid locale tags:
  - Chinese: `zh`.
  - Chinese (Singapore): `zh_SG`.
  - Portuguese (Brazil): `pt_BR`.
+ - Spanish: `es`.
 
 ### Default language for newly added pages
 


zdroid:
I don't think adding new language tags is required anywhere, since both `CONTRIBUTING.md` and `CLIENT-SPECIFICATION.md` contain only a list of examples.

ivanhercaz:
> I don't think adding new language tags is required anywhere, since both `CONTRIBUTING.md` and `CLIENT-SPECIFICATION.md` contain only a list of examples.

Ups! All right. So we could close this PR if you know there is consensus about it. And if someone has another opinion, it could be reopened. Anyway, I am agree with have it or not in the list, so I haven't problems with any decision :smile:.

agnivade:
Thank you @ivanhercaz. Yes, those are just examples. Every language does not need to be mentioned explicitly. I will close this.

PR #3684, opened 2019-12-22
clear: add -x, -T and -V examples

ivanhercaz:

I edited the page to add the examples for `-x`, `-T` and `-V`.

### DIFF
--- a/pages/common/clear.md
+++ b/pages/common/clear.md
@@ -14,6 +14,6 @@
 
 `clear -T {type_of_terminal}`
 
-- Show the version of `ncurses` used by `clear`.
+- Show the version of `ncurses` used by `clear`:
 
 `clear -V`
--- a/pages/common/clear.md
+++ b/pages/common/clear.md
@@ -10,7 +10,7 @@
 
 `clear -x`
 
-- Indicate the type of terminal to clean (normally this option is unnecessary because it is taken from the environment variable `TERM`):
+- Indicate the type of terminal to clean (defaults to the value of the environment variable `TERM`):
 
 `clear -T {{type_of_terminal}}`
 
--- a/pages/common/clear.md
+++ b/pages/common/clear.md
@@ -2,7 +2,7 @@
 
 > Clears the screen of the terminal.
 
-- Clear the screen (equivalent to typing Control-L when using the bash shell):
+- Clear the screen (equivalent to pressing Control-L in Bash shell):
 
 `clear`
 
--- a/pages/common/clear.md
+++ b/pages/common/clear.md
@@ -12,7 +12,7 @@
 
 - Indicate the type of terminal to clean (normally this option is unnecessary because it is taken from the environment variable `TERM`):
 
-`clear -T {type_of_terminal}`
+`clear -T {{type_of_terminal}}`
 
 - Show the version of `ncurses` used by `clear`:
 
--- a/pages/common/clear.md
+++ b/pages/common/clear.md
@@ -5,3 +5,15 @@
 - Clear the screen (equivalent to typing Control-L when using the bash shell):
 
 `clear`
+
+- Clear the screen but keep the terminal's scrollback buffer:
+
+`clear -x`
+
+- Indicate the type of terminal to clean (normally this option is unnecessary because it is taken from the environment variable `TERM`):
+
+`clear -T {type_of_terminal}`
+
+- Show the version of `ncurses` used by `clear`.
+
+`clear -V`


pages/common/clear.md
```diff
@@ -5,3 +5,15 @@
 - Clear the screen (equivalent to typing Control-L when using the bash shell):
```
zdroid: (Unrelated to this PR, but I caught it now)
```suggestion
- Clear the screen (equivalent to pressing Control-L in Bash shell):
```

ivanhercaz:
> The [build](https://travis-ci.org/tldr-pages/tldr/builds/628441386)
> for this PR failed with the following error(s):
> 
> ```
> pages/common/clear.md:17: TLDR005 Example descriptions should end in a colon
> ```
> 

Fixed with 9d35ee4!

pages/common/clear.md
```diff
@@ -5,3 +5,15 @@
...
 
 `clear`
+
+- Clear the screen but keep the terminal's scrollback buffer:
```
zdroid: I'd replace `terminal's scrollback buffer` with something simpler.

pages/common/clear.md
```diff
@@ -5,3 +5,15 @@
...
+
+`clear -x`
+
+- Indicate the type of terminal to clean (normally this option is unnecessary because it is taken from the environment variable `TERM`):
+
+`clear -T {type_of_terminal}`
+
+- Show the version of `ncurses` used by `clear`.
```
zdroid: Is this really necessary? Seems unneeded to absolute majority of the users.

pages/common/clear.md
ivanhercaz: What would you propose? Now I don't know something simpler to explain it, but I will think about it.

pages/common/clear.md
ivanhercaz: Yes, that's true. But it is a command with just a few parameters. Do you think it would be necessary to remove it?

zdroid: You can wait for the opinions of others, but I'd keep only the useful examples. :wink:

pages/common/clear.md
zdroid: Sounds way too technical, but understandable. If we don't come up with anything, this can stay, it's not a problem.

ivanhercaz: Yes, I am agree with your argument. May be... "Clear the screen and allow to scroll up to the previously used commands" or "Clear the screen and allow to scroll up to the the previous typed commands"?

pages/common/clear.md
ivanhercaz: As we are waiting for more opinions in the above issue, we can wait.

pages/common/clear.md
sbrl: I'd argue that we should keep it as-is. I'm not sure that it's any simpler if we change the language, and we'd risk being ambiguous if we were to change it.

pages/common/clear.md
sbrl: There's no harm in including it, I guess - since there isn't anything else we can include on the  page.

pages/common/clear.md
```diff
@@ -2,6 +2,18 @@
...
 
 `clear`
+
+- Clear the screen but keep the terminal's scrollback buffer:
+
+`clear -x`
+
+- Indicate the type of terminal to clean (normally this option is unnecessary because it is taken from the environment variable `TERM`):
+
+`clear -T {type_of_terminal}`
```
sbrl: I think you meant to have double `{{}}` here?
```suggestion
`clear -T {{type_of_terminal}}`
```

pages/common/clear.md
```diff
@@ -2,6 +2,18 @@
...
-- Clear the screen (equivalent to typing Control-L when using the bash shell):
+- Clear the screen (equivalent to pressing Control-L in Bash shell):
 
 `clear`
+
+- Clear the screen but keep the terminal's scrollback buffer:
+
+`clear -x`
+
+- Indicate the type of terminal to clean (normally this option is unnecessary because it is taken from the environment variable `TERM`):
```
sbrl: Perhaps this will read better?
```suggestion
- Indicate the type of terminal to clean (defaults to the value of the environment variable `TERM`):
```

pages/common/clear.md
ivanhercaz: Nice! Thank you for your comment, @sbrl. I am going to mark it as resolved.

pages/common/clear.md
ivanhercaz: So could it being marked as resolved? Or should we wait for more opinions about this?

ivanhercaz:
> Just curious, in what situation will we use this options?

Hi @einverne! 
  - `clear` will be used anytime the user want to clean the terminal. I use it a lot.
  - `clear -x` is a really useful option to clean but keep the history of commands written in that terminal. I admit that I knew about it a few time ago but I really think this option might be really useful.
  - `clear -T`. I haven't read or seen any situation in which it can be used because `clear` uses the environment variable `TERM` to know the terminal type. I understand there could be situations in which this env variable haven't been configured or something similar.
  - `clear -V` seems just an option to know a bit of information about it. It may be for bug and report reasons, but the first time I used it was to confirm its usage for the written of this tldr page.

pages/common/clear.md
ivanhercaz: Ups! I little slip. I will fixed it.

ivanhercaz:
@sbrl, I commit the changes for both modifications you suggested.

pages/common/clear.md
andrik: I think we could maintain the example and if we have problems exceeding the examples limit, then this example would be the first one to be removed.

PR #3687, opened 2019-12-24
Change main license to CC-BY 4.0

### LINKED ISSUE #1076: Re-License from MIT to ?
This can of worms was opened in #1048 perhaps we should relicense with a documentation specific license rather than the current MIT license. This issue applies only to tldr pages, not the tldr clients.

Some potential options: 
- [CC0 (Public Domain)](http://choosealicense.com/licenses/cc0-1.0/)
- [CC-BY-4.0](http://choosealicense.com/licenses/cc-by-4.0/)
- [CC-BY-SA-4.0 (Share Alike)](http://choosealicense.com/licenses/cc-by-sa-4.0/)
- Stick with [MIT](http://choosealicense.com/licenses/mit/)

Any re-licensure could be accomplished reaching out to all previous committers to get them to take 30 seconds to agree to the [Contributor License Agreement wizard proposed in #1048](https://github.com/tldr-pages/tldr/issues/1048).  If original contributors can't be reached/don't agree we would just have to re-write/replace the examples they wrote.

waldyrious:
This change has been a long time coming, but we can finally wrap up the long-running discussion at #1076!

This change is justified by the following:

1. The MIT license is appropriate for open source software, but the vast majority of content in this repository is open content prose;
2. The best-known licenses for open content text are the Creative Commons ones;
3. The Creative Commons license that most closely approximates the properties of the MIT license is CC-BY (Attribution) 4.0;
4. The community has voiced overall support for this change in #1076; 
5. The MIT license allows for relicensing content subject to it.

Note: The software part of this repository, i.e. the contents of the `scripts/` directory, remains under the MIT license.

Fixes #1076 (at long last! 🎉)

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,18 +1,23 @@
-## The MIT License (MIT)
+Copyright © 2014—present the [TLDR team](https://github.com/orgs/tldr-pages/people) and [contributors](https://github.com/tldr-pages/tldr/graphs/contributors).
 
-Copyright (c) 2014 the [TLDR team](https://github.com/orgs/tldr-pages/people) and [contributors](https://github.com/tldr-pages/tldr/graphs/contributors)
+This work is licensed under a [**Creative Commons Attribution 4.0 International License**](http://creativecommons.org/licenses/by/4.0/).
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
+The contents of the [scripts/](https://github.com/tldr-pages/tldr/tree/master/scripts) directory
+are licensed under the MIT license:
+
+> ## The MIT License
+>
+> Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
 in the Software without restriction, including without limitation the rights
 to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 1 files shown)


zdroid:
@sbrl It probably will recognize the license, as per https://choosealicense.com/licenses/cc-by-4.0/.

zdroid:
Wouldn't it be better to have LICENSE files separate, i.e. just put MIT into `scripts/`?

waldyrious:
> Wouldn't it be better to have LICENSE files separate, i.e. just put MIT into `scripts/`?

Kind of. It would make the license file(s) slightly simpler, but at the same time more complex by requiring two of them rather than one, and it wouldn't make it immediately obvious that there's non CC-BY content in the repo unless someone's paying close attention (which we might reasonably assume may be an issue since license files are easily overlooked). I don't think that change is worth the potential licensing confusion, but I'm happy to hear others' opinions.

agnivade:
I think having a single LICENSE file is fine.

zdroid:
To support my idea: someone using scripts is most likely also involved in the project and informed about it, since it's for power-users. Also they have to go through the directory (and thus see the LICENSE file) to reach the scripts.

PR #3699, opened 2019-12-27
tic: add page

jedahan:


sbrl:
Looks like I have `tic` installed - though I don't remember installing it myself :P

What about adding an example for the `-c` option, which only checks for errors but doesn't actually change anything? Or the `-D` option, which prints the database locations and then exits?

Either way, if you'd rather not we can always merge this PR and then create another one :-)

ivanhercaz:
It seems to be fine for me, including the new examples added after the suggestion made by @sbrl. This PR has the two required approvals and could be merged, but I prefer to wait until sbrl, as pending reviewer, check it.

Issue #3639, opened 2019-12-02
Having OS-specific teams


zdroid:
Since this project is scaling quickly, it may be a good idea to have OS-specific teams in the maintainers. It'd allow easier reviewing by mentioning qualified people for each task.

The result could be them pinged any time a page is added under an OS-specific folder. Related: #1068

sbrl:
Interesting idea, @zdroid! I'd suggest that the primary divide at the moment is actually languages and translation, which we have #2793 currently tracking some improvements (but I think we're back to the drawing board on what to do about it though). Quite a few commands are actually cross-platform these days. It would be great if there was a way to quickly @mention people who have previously contributed to translations for a given language.

zdroid:
There is a divide, at least on Windows and UNIX-based operating systems. Commands definitely aren't the same, far from it. Additionally, some more niche environments like Sun OS (or BSD if we add it in the future) cannot be followed that easily by all people.

sbrl:
Very true, @zdroid. Perhaps we could solve a wider issue with @mentioning groups of people here though. Perhaps a bot where you can click a link to be added to a named group of people, and then use something like `|group_name|` would automatically generate the a comment @mentioning everyone in that group, perhaps?

Also of interest may be #1068.

ivanhercaz:
Both ideas, languages and OS teams, are really interesting and useful. If this idea progress it would be a pleasure to contribute as a member of a Linux group, as well of the Spanish group.

zdroid:
@sbrl Doesn't GitHub do that automatically, though?

sbrl:
I'm not aware of any mechanism by which it does @zdroid. There are teams, but IIRC this requires that you're an org member first before you can be made a team member?

zdroid:
I cannot test that, since I'm not an owner in any organization with outside collaborators. Did you try making a team and putting outside collaborators in?

sbrl:
My point is that it would be cool if there was a mechanism by which users could join the teams themselves @zdroid - rather than having to make a request to join.

zdroid:
Don't we have a condition for each collaborator to get 5 PRs merged first? Even if we had a bot, then you also need to add a mechanism to regulate the people in the team.

sbrl:
True, but a bot that comments on a PR to say how many previous PRs someone has merged would be nice.

zdroid:
[...]

PR #3703, opened 2019-12-28
git-merge: add Spanish translation

ivanhercaz:

### DIFF
--- a/pages.es/common/git-merge.md
+++ b/pages.es/common/git-merge.md
@@ -1,0 +1,20 @@
+# git merge
+
+> Fusiona ramas.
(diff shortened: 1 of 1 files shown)


ivanhercaz:
> Don't speak spanish, but approving anyway since I don't think we have any other spanish-speaking contributors...

One plan that I have is to invite some Spanish-speakers mates that I know they love to write documentation. We need to have more than one collaborator reviewing a language. I am really sure there are things that I write that might be improved.

> (perhaps we should compile a list at some point, but I think there's an issue already open about groups of people to @mention).

Yes, there is #3639 open. It is specific for OS teams, but we talked about the language teams too.

PR #3709, opened 2019-12-29
cd: add Spanish translation

ivanhercaz:

### DIFF
--- a/pages.es/common/cd.md
+++ b/pages.es/common/cd.md
@@ -1,0 +1,19 @@
+# cd
+
+> Cambiar el directorio de trabajo.
+
+- Accede al directorio especificado:
+
+`cd {{ruta/al/directorio}}`
+
+- Accede al directorio *home* del usuario actual:
+
+`cd`
+
+- Accede al directorio padre del directorio actual:
+
+`cd ..`
+
+- Accede al directorio elegido previamente:
+
+`cd -`


pages.es/common/cd.md
```diff
@@ -0,0 +1,19 @@
+# cd
+
+> Cambiar el directorio de trabajo.
+
+- Accede al directorio especificado:
+
+`cd {{ruta/al/directorio}}`
+
+- Accede al directorio *home* del usuario actual:
```
zdroid: I'd rather remove the italic marks around `home`.

ivanhercaz: I write "home" with italic marks because in Spanish is necessary to mark in italic the words in another language (it is a rule of the official Spanish). As `home` is known as `home` in whatever language, I think that it should be in italic at least.

Is there any reason why you think that despite this rule they should be retired?

zdroid: It's fine then. Don't you translate home/ directory's name though?

sbrl: I think this is referring to the home directory here @zdroid, which is a special concept / name.

ivanhercaz: Yes, I am referring to `home`. I speak from Spanish-speaker perspective and trying to follow as possible the grammar-spelling rules. In the case of `home` directory in all the Linux distributions I have used in Spanish, it is always `home`.

zdroid: Okay, so we keep the italicized version?

ivanhercaz: I am waiting your decision about this kind of stuff.

sbrl: I say that we keep the italics here.

PR #3712, opened 2019-12-29
qcp, qmv: small tweaks

waldyrious:
These changes are a follow-up to #3458. Besides the self-explanatory clarification changes, I also restored the fully-spelled `--recursive` option instead of `-R` (reverting the change discussed [here](https://github.com/tldr-pages/tldr/pull/3458#discussion_r337400925)), since the word "recursive" is not used in the example's description, so the single letter option's meaning isn't obvious (unlike the `-d` option which was changed from `--directory` following [this discussion](https://github.com/tldr-pages/tldr/pull/3458#discussion_r337399770)).

/cc @einverne and @neizod.

### DIFF
--- a/pages/common/qmv.md
+++ b/pages/common/qmv.md
@@ -1,9 +1,9 @@
 # qmv
 
-> Move files and directories with your favorite text editor.
+> Move files and directories using the default text editor to define the filenames.
 > More information: <https://www.nongnu.org/renameutils/>.
 
-- Move a single file (invoke an editor with source on the left, target on the right):
+- Move a single file (open an editor with the source filename on the left and the target filename on the right):
 
 `qmv {{source_file}}`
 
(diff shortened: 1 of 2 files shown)


pages/common/qmv.md
```diff
@@ -1,24 +1,24 @@
...
 `qmv {{*.jpg}}`
 
-- Move 3 directories:
+- Move multiple directories:
 
-`qmv -d {{path/to/dir_1}} {{path/to/dir_2}} {{path/to/dir_3}}`
+`qmv -d {{path/to/directory1}} {{path/to/directory2}} {{path/to/directory3}}`
 
-- Move files/directories inside a directory:
+- Move all files and directories inside a directory:
```
einverne: Make sense. But for me, it is still uncertain whether should I commit the short option or commit the full-length option. So under this situation, I can see that if we describe the command very clear, or use the option in our description, we can use the short version in command line. So how about 

```suggestion
- Move all files and directories inside a directory recursively:
```
In this way, we can indicate that `-R` stands for `recursive`.

waldyrious: Yes, your suggestion is precisely what the [contribution guideline \#3](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines) recommends:

> Try to incorporate the spelled-out version of single-letter options in the example's description. The goal is to allow people to _understand_ the syntax of the commands, not just _memorize_ it.

Maybe that wording isn't clear enough? Is there any change to it that would make it more explicit in your eyes?

zdroid: I'm neutral on this, both seem to work.

einverne: I am just a little confused about `-R` or `--recursive` choice. These changes are clear to me.

waldyrious: > These changes are clear to me.

@einverne I'm not sure if you were responding to my comment, but I wanted to know if the contribution guidelines (specifically the entry I quoted above) do clarify when we are supposed to use the abbreviated vs. the spelled-out version of a command option. If it's still not clear, please let me know what's confusing so we can improve the guideline. 🙂

einverne: @waldyrious I understand, the spelled-out version is preferred if we cannot incorporate it into the description.

PR #3716, opened 2019-12-30
Avoid title case

zdroid:
Makes titles consistent through the project.

### DIFF
--- a/CLIENT-SPECIFICATION.md
+++ b/CLIENT-SPECIFICATION.md
@@ -1,5 +1,4 @@
-tldr-pages client specification
-===============================
+# tldr pages client specification
 
 **Current Specification Version:** 1.2
 
@@ -8,8 +7,7 @@ This document contains the official specification for tldr-pages clients. It is 
 The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
 
-Terminology
------------
+## Terminology
 
 In order to aid the understanding of this specification document, a number of terms will be defined in this section.
 
@@ -26,8 +24,7 @@ If a page is common across multiple platforms, but slightly differently on a giv
 For example, if the command `foo` is common to `mac`, `windows`, and `linux` but functions differently on `windows`, then the main page will be stored in `common`, and a copy will be placed in `windows` that's altered to match the different functionality.
 
 
-Command-Line Interface
-----------------------
+## Command-line interface
 
 This section describes the standardised command-line interface (CLI) for clients implementing one. Clients that do not provide a CLI can ignore this section.
 
@@ -57,7 +54,7 @@ tldr --version
 tldr -l
 ```
 
-### Page Names
+### Page names
 
 The first argument that does not start with a dash (`-`), MUST be considered the page name.
 
@@ -83,8 +80,7 @@ tldr --platform osx bash
 ```
 
 
-Directory Structure
--------------------
+## Directory structure
 
 This section documents the directory structure that contains the pages themselves.
 
@@ -124,14 +120,12 @@ Some examples:
 The structure inside these translation folders is identical to that of the main `pages` folder.
 
 
-Page Structure
---------------
+## Page structure
 
 Although this specification is about the interface that clients must provide, it is also worth noting that pages are written in standard [CommonMark](https://commonmark.org/), which the exception of the non-standard `{{` and `}}` syntax, which surrounds values in an example that users may edit. Clients MUST NOT break if the page format is changed within the _CommonMark_ specification.
 
 
-Page Resolution
----------------
+## Page resolution
 
 This section defines the algorithm by which a client can decide which page a user has requested.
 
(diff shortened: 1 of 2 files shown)


sbrl:
These are, umm, titles @zdroid. Shouldn't they be in Title Case?

/cc @waldyrious

zdroid:
@sbrl In my opinion, title cases are neither stylistic nor useful. This especially applies to multi-language projects and projects unrelated to journalism or literature.

Note: almost all of the project's documentation avoids title case, except for those few places that I corrected.

zdroid:
Any opinions on this @agnivade @pxgamer?

pxgamer:
LGTM, I much prefer this way of having headings. But was there a reason for them being this way originally?

zdroid:
Not much really, @pxgamer. The rest of the project uses sentence case solely.

agnivade:
I am not aware of the finer points of English language. Let's wait for @waldyrious' thoughts on this.

ivanhercaz:
Title case is something that it isn't very standardized and it is applied in different ways. See [_title case_](https://en.wikipedia.org/wiki/Letter_case#Title_case) in the _Letter case_ page of the English Wikipedia. But the most interesting part of the page is the one dedicated to the [_heading and publication titles](https://en.wikipedia.org/wiki/Letter_case#Headings_and_publication_titles):

> In English-language publications, various conventions are used for the capitalisation of words in publication titles and headlines, including chapter and section headings. The rules differ substantially between individual house styles.
>
> The convention followed by many British publishers (including scientific publishers, like Nature, magazines, like The Economist and New Scientist, and newspapers, like The Guardian and The Times) and also U.S. newspapers, is sentence-style capitalisation in headlines, i.e. capitalisation follows the same rules that apply for sentences. This convention is usually called sentence case. It may also be applied to publication titles, especially in bibliographic references and library catalogues. An example of a global publisher whose English-language house style prescribes sentence-case titles and headings is the International Organization for Standardization.

From the [English Wikipedia](https://en.wikipedia.org), by its collaborators, licensed under a [Creative Commons Attribution-ShareAlike 4.0 license](https://creativecommons.org/licenses/by-sa/4.0/).

mebeim:
Title Case is a pretty standard and probably the most used format for section headers. I don't see a good reason to change this to be honest. I'm personally ok with Title Case where it makes sense to be applied. With that being said, it's not a rule of the English language for sure, so I wouldn't argue if everybody else prefers "Sentence case" (i.e. capitalization of the first letter of the first word only).

As per the markdown formatting, I prefer explicit lines below the text (matching the length of the title) where possible for level 1 and level 2 headings, because it makes it much clearer to distinguish them from body text or other lower level headers, both while editing and also while reading from a text viewer that does not render markdown (or from the terminal).

zdroid:
@mebeim Please refer to the rest of project's docs. I've basically just made the style of writing same everywhere. Nothing I've changed is new, it's the same style we're using everywhere.

Note on title case: it's avoided by most of the international projects for being too confusing for non-native speakers. This project is quite international and multilingual, so I'd avoid it at all costs as general preference.

mebeim:
@zdroid good point on Title Case, makes sense. As per the markdown formatting I see that the underline style for headers is only applied in the client spec. While it makes sense to standardize it, I would like to hear what others think, for example if anybody else agrees with my comment above:

> I would suggest keeping explicit lines below the text (matching the length of the title) where possible for level 1 and level 2 headings, since it makes it much clearer to distinguish them from body text or other lower level headers, both while editing and also while reading from a text viewer that does not render markdown (or from the terminal).

Maybe we should open an Issue for this.

zdroid:
I prefer the current:
```markdown
# Heading 1

Text


## Heading 2

Text
```

to what you're suggesting (if I understood you correctly):
```markdown
# Heading 1


Text


## Heading 2


Text
```

It looks better when heading and text under it are grouped.

mebeim:
@zdroid sorry if I was not clear enough.

What I'm suggesting is this:

```
Heading level 1
===============

Text.


Heading level 2
---------------

Text.
```

That's what I meant with "keeping explicit lines matching the length of the title".

zdroid:
It makes it easier to read if you are reading it as plain text instead of parsed Markdown. On the other side, it makes editing harder, which is why `#` and `##` are preferred over dashes.

ivanhercaz:
> @zdroid sorry if I was not clear enough.
> 
> What I was talking about is the following:
> 
> ```gfm
> Heading level 1
> ===============
> 
> Text.
> 
> 
> Heading level 2
> ---------------
> 
> Text.
> ```
> 
> That's what I meant with "keeping explicit lines matching the length of the title". As I said, IMHO it makes the document much easier to read.

About it, I prefer to use `#` and `##` over `====` and `----`.

sbrl:
I agree with @ivanhercaz, but I do question as to whether this is really necessary? Surely this is a very minor thing with respect to other areas of the project.

zdroid:
@sbrl perhaps it isn't, but keep in mind this is a technical, documentation-oriented project. If anyone is going to be pedantic about documentation, then it has to be us. :P

zdroid:
@sbrl Still waiting for your final say :laughing:

sbrl:
Go ahead and merge if you like, but I don't really see the need for this?

I'm still against it.

waldyrious:
@sbrl I'm so sorry that I wasn't able to join the discussion in due time. For the record, I do agree with both sentence case and with [atx-style](http://www.aaronsw.com/2002/atx/intro) headings. The latter, for the reasons succinctly [described](https://github.com/google/styleguide/blob/gh-pages/docguide/style.md#atx-style-headings) in Google's documentation style guide. The former, for a few reasons: 

- First, as was mentioned above, because it's an English tradition that doesn't map well to other languages;
- Second, because it doesn't offer a consistent, unambiguous standard that could be followed, due to the complexities of the English language;
- Third, and most importantly IMO, because `Title Case`, like `ALL CAPS`, was created as a form of typographic emphasis that is no longer necessary now that we have rich text capabilities in most written media, especially digital ones. (In fact, [many have argued](https://law.stackexchange.com/questions/743/do-disclaimers-need-to-be-capitalized) that the all-caps disclaimers in licenses, for example, are an archaic tradition that these days only contributes to make "legalese" more intimidating and less readable — because all that copyright law demands is that these sections should be made prominent —conspicuous, to use the actual term— in the license text.)

Anyway — apologies again for the absence; I hope these justifications make a little more sense to you, and ease the nuisance of this being merged against your wishes.

Issue #2799, opened 2019-02-23
Should we allow more than 8 examples?


mfrw:
@sbrl @waldyrious @agnivade @mebeim certain pages are already kind of full. How should we prioritize what 8 examples stay and what don't?
Also, does it make sense to break the rule of 8 examples, if the utility is useful?
e.g: #2798

waldyrious:
This discussed several times if I recall correctly; the thread that originated the current limit was in PR #1149. You may also find relevant remarks if you search for ["commandlinefu"](https://github.com/tldr-pages/tldr/search?q=commandlinefu+is%3Apr&type=Issues).

What we've repeatedly agreed is that the main purpose of the tldr-pages project (it's in the name!) is to provide a short summary, and never exhaustive coverage of the tools. This is even item \#1 in our guidelines:

> Try to keep pages at around 5 examples. Pages can be longer if needed, but don't exceed 8 examples. Remember, it's OK if the page doesn't cover everything; that's what `man` is for.

Of course, for more complex commands this will be challenging, but it is a challenge we should embrace to help us formulate the most concise quick-start for each tool. In fact, the most complex tools are _precisely_ the ones more in need of a quick intro resource.

> How should we prioritize what 8 examples stay and what don't?

I'd use the 80-20 rule here: even if what we can cover in a tldr-page is only about 20% of what the command can do, we can strive to cover the 20% most common invocations that will cover 80% of the use cases.

Additionally, we can balance that strategy with one of picking examples that illustrate the tool's usage in composable ways, so that people can use the tldr-pages examples as building blocks to figure out how to build specific more complex invocations they may need.

So rather than offering complete, ready-made examples that can be used directly, we should leave that goal to commandlinefu, and instead focus on "teaching users how to fish" by making short guides that _explain_ how each tool works.

mebeim:
I agree with @agnivade.

I think focusing on having a short and useful set of most common examples to get started with a command is the main goal. Eight examples already feel like enough for that purpose. If a command is more complex and requires more than just a TLDR page, then there's `man` that can answer all your questions. 

TLDR is for when you're in a hurry or when you need to perform a trivial task, but you don't remember the syntax of the command. More complex commands could also use examples that combine multiple options to create a more complete page, as already said.

Taking #2798 as an example: most of the examples in the file proposed in that PR are actually kind of redundant, and some are not as commonly used as others. In such cases I think we should carefully analyze which one makes more sense to be in the page and which one doesn't.

[...]

PR #3718, opened 2019-12-31
Added a merging audo and video command

Ray6464:
I need this command again and again but keep forgetting.

     Not really, the page is a bit long, however people using this tool don't really care as long as it's well organized and to the point.
- [ x ] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [ x ] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).

### DIFF
--- a/pages/common/ffmpeg.md
+++ b/pages/common/ffmpeg.md
@@ -34,3 +34,8 @@
 - Convert MP4 video to VP9 codec. For the best quality, use a CRF value (recommended range 15-35) and -b:video MUST be 0:
 
 `ffmpeg -i {{input_video}}.mp4 -codec:video libvpx-vp9 -crf {{30}} -b:video 0 -codec:audio libopus -vbr on -threads {{number_of_threads}} {{output_video}}.webm`
+
+- Merge audio with video. This command takes in a video file, and an audio file and give out a merged file with both audio and video.
+
+`ffmpeg -i video.mp4 -i audio.wav \
+-c:v copy -c:a aac -strict experimental output.mp4`
--- a/pages/common/ffmpeg.md
+++ b/pages/common/ffmpeg.md
@@ -35,7 +35,6 @@
 
 `ffmpeg -i {{input_video}}.mp4 -codec:video libvpx-vp9 -crf {{30}} -b:video 0 -codec:audio libopus -vbr on -threads {{number_of_threads}} {{output_video}}.webm`
 
-- Merge audio with video. This command takes in a video file, and an audio file and give out a merged file with both audio and video.
+- Merge audio with video. This command takes in a video file, and an audio file and give out a merged file with both audio and video:
 
-`ffmpeg -i video.mp4 -i audio.wav \
--c:v copy -c:a aac -strict experimental output.mp4`
+`ffmpeg -i video.mp4 -i audio.wav -c:v copy -c:a aac -strict experimental output.mp4`
--- a/pages/common/ffmpeg.md
+++ b/pages/common/ffmpeg.md
@@ -37,4 +37,4 @@
 
 - Merge audio with video. This command takes in a video file, and an audio file and give out a merged file with both audio and video:
 
-`ffmpeg -i video.mp4 -i audio.wav -c:v copy -c:a aac -strict experimental output.mp4`
+`ffmpeg -i {{video}}.mp4 -i {{audio}}.wav -c:v copy -c:a aac -strict experimental {{output}}.mp4`


pages/common/ffmpeg.md
```diff
@@ -34,3 +34,7 @@
 - Convert MP4 video to VP9 codec. For the best quality, use a CRF value (recommended range 15-35) and -b:video MUST be 0:
 
 `ffmpeg -i {{input_video}}.mp4 -codec:video libvpx-vp9 -crf {{30}} -b:video 0 -codec:audio libopus -vbr on -threads {{number_of_threads}} {{output_video}}.webm`
+
+- Merge audio with video. This command takes in a video file, and an audio file and give out a merged file with both audio and video:
+
+`ffmpeg -i video.mp4 -i audio.wav -c:v copy -c:a aac -strict experimental output.mp4`
```
sbrl: Perhaps we could put the things that can be changed in tokens?
```suggestion
`ffmpeg -i {{video}}.mp4 -i {{audio}}.wav -c:v copy -c:a aac -strict experimental {{output}}.mp4`
```

Ray6464:
> Hey, thanks for the edit @Ray6464! I've left a comment below for you to review.
> 
> In addition, it looks like we've got too many examples on this page. Which one would you suggest is the least useful?
> 
> Finally, it looks like you haven't yet signed the CLA. We need you to sign this in order to be able to merge your PR. Note that the email address you've committed with has to be present in your GitHub account settings.

I don't know anything about video editing, however the "Convert MP4 video to VP9 codec" seems useless for me for the most part, it has 4 "unknowns" in it which suggests that if the user really using ffmpeg for that kind of stuff they should just read man pages instead or tldr, I mean 4 variables? how in the world is that tldr.

I also suggest that tldr should keep both smaller and longer versions of files that works when tldr is used with a "-L" flag. I mean I idea behind tldr is not keeping the docs small but the examples small and simple right? Keeping the docs as small as 8 examples is a huge setback without a clear philosophy of what example goes in and what goes out.

TL;DR: My commit in, lengthy commits out!

sbrl:
My knowledge of video editing is more limited than I'd like it to be, but VP9 is the latest WebM codec (which is great for embedding videos in web pages, since it tends to generate smaller files), so I think that example is useful. However, I certainly agree that it can be simplified. For example if a specific value must be zero, then there's no need to mention it in the description.

Regarding shorter and longer versions of pages, that's probably best suggested in a separate issue. Perhaps you could open one?

zdroid:
I also agree this example is useful. Perhaps we could change maximum number of examples to 10.

Ray6464:
> My knowledge of video editing is more limited than I'd like it to be, but VP9 is the latest WebM codec (which is great for embedding videos in web pages, since it tends to generate smaller files), so I think that example is useful. However, I certainly agree that it can be simplified. For example if a specific value must be zero, then there's no need to mention it in the description.
> 
> Regarding shorter and longer versions of pages, that's probably best suggested in a separate issue. Perhaps you could open one?

How about this, if the example is really useful it's allowed through the 8-example limit, than later if we have a over-flow problem we can break the overflowing pages into small pages with an 8-limit and larger pages.

Also keep in mind that the 8-limit is will eventually kill this tool, it's very limiting; Instead of 8-limits, let the 8-limit be a rule of a loose thumb and lets be very strict about what goes in and what comes out. How about that?

zdroid:
I would definitely keep the limit, @Ray6464. Our pages are supposed to be short and concise. On the other side, I'd prefer 10 examples instead of arbitrarily chosen 8.

Ray6464:
> I would definitely keep the limit, @Ray6464. Our pages are supposed to be short and concise. On the other side, I'd prefer 10 examples instead of arbitrarily chosen 8.

Works for me!

sbrl:
Changing the limit on the number of examples is not a small change - it should probably be discussed in a separate issue. I'm sure there was a reason for the existing limit - I'm just not sure what it is.

zdroid:
It definitely does. Do you have some formed text for the matter, so you can open the issue yourself?

sbrl:
Define `formed text`

zdroid:
That means: do you have some worthy thoughts or comments about it to write down in the (supposed) issue's description?

pxgamer:
@sbrl/@zdroid, https://github.com/tldr-pages/tldr/issues/2799

sbrl:
Ah, thanks @pxgamer! I did have a feeling that we discussed it before.

zdroid:
:frowning_woman:

PR #3723, opened 2020-01-01
Update the OS list in README

zdroid:


README.md
```diff
@@ -46,7 +46,7 @@ How about:
 ![tldr screenshot](images/screenshot.png)
 
 This repository is just that: an ever-growing collection of examples
-for the most common UNIX / Linux / macOS / SunOS commands.
+for the most common UNIX, Linux, OS X, SunOS and Windows commands.
```
sbrl: This is not quite correct. The new name for OS X is macOS - this should be changed.

ivanhercaz:
> Hrm, looks like this was merged before I got a chance to review.
> 
> I do recommend with documentation changes like the README, guidelines, etc. leaving them open for longer to allow all members to comment.

All right! Excuse me the inconvenience, @sbrl.  I will keep this in mind for the next PR of the same type.

README.md
zdroid: Hmm, that's true. I must have been out of date with this, as this has changed back in 2016. :confused:

PR #3728, opened 2020-01-02
git-show: add example for --stat option

waldyrious:
As the title says :)

Also reword two descriptions to avoid misleading people regarding when the commit hash is shown.

### DIFF
--- a/pages/common/git-show.md
+++ b/pages/common/git-show.md
@@ -23,7 +23,7 @@
 
 `git show --oneline -s {{commit}}`
 
-- Show only a list of the files changed in a commit:
+- Show only the list of the files changed in a commit:
 
 `git show --stat {{commit}}`
 
--- a/pages/common/git-show.md
+++ b/pages/common/git-show.md
@@ -3,7 +3,7 @@
 > Show various types of git objects (commits, tags, etc.).
 > More information: <https://git-scm.com/docs/git-show>.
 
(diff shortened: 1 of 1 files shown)


pages/common/git-show.md
```diff
@@ -19,10 +19,14 @@
 
 `git show {{branch}}~{{3}}`
 
-- Show a commit's hash and message in a single line, suppressing the diff output:
+- Show a commit's message in a single line, suppressing the diff output:
```
principis: The short hash is shown in the output, so why change this?
```
git show --oneline -s bf3503d258865b5344128e8ffccb34435a3e8b6c
bf3503d Fixes
```

waldyrious: As I mentioned in the PR description:

> to avoid misleading people regarding when the commit hash is shown.

The hash was added to the first example to make it clearer that it's shown by default, rather than specifically in this example.

principis: Ok :)

pages/common/git-show.md
```diff
@@ -19,10 +19,14 @@
...
 
 `git show --oneline -s {{commit}}`
 
+- Show only a list of the files changed in a commit:
```
zdroid:
```suggestion
- Show only the list of the files changed in a commit:
```
should be definite article there AFAIK

waldyrious: Makes sense 👍

PR #3731, opened 2020-01-04
dig : Add Korean Translation

Proscream:

### DIFF
--- a/pages.ko/common/dig.md
+++ b/pages.ko/common/dig.md
@@ -1,0 +1,31 @@
+# dig
+
+> DNS 조회 유틸리티.
+
+- 호스트이름과 관련된 IP 조회하기 (A records):
+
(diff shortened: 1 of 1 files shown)


Proscream:
@zdroid  In the text, I thought 'record' was the same as the file format or extension (like `avi or mp4`), and the reason why I didn't translate the example of the site(`example.com`) is because even in Korea, all the website links are used in English.(like Korea's largest website, `www.naver.com`) :)

Proscream:
To explain in more detail, for example, in the case of mp4 (mpeg-4), if I don't write this down as mp4 and translate the Moving Picture Experts Group 4 into Korean, it will be difficult to think of the extension mp4 after seeing the above translation.

zdroid:
Makes sense.

PR #3512, opened 2019-10-30
tar: Add French translation

jabesq:
Signed-off-by: Hugo D. (jabesq) <jabesq@gmail.com>

- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #3732, opened 2020-01-04
tar: add French translation

mebeim:
Opening this as #3512 has been sitting there inactive for a little bit too much. All suggestions have been applied, so it should be ready to merge.


ivanhercaz:
Thank you for taking care of this pending PR, @mebeim.

PR #3734, opened 2020-01-04
Add extra installation instructions to clients which OS packaged versions

ivanhercaz:
This tldr client written in Haskell is [available as a Debian package](https://packages.debian.org/sid/tldr) maintained by the Debian Haskell Group and Clint Adams, so it can be installed with the different commands Debian has to manage packages. The README already gives alternative installation instructions for some clients (one written in Go and one in Python).

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ You can access these pages on your computer using one of the following clients:
     [App Store](https://itunes.apple.com/us/app/tldt-pages/id1071725095?ls=1&mt=8)
 - [Haskell client](https://github.com/psibi/tldr-hs):
   `stack install tldr`
-  or `apt-get install tldr` on Debian
+  or `apt-get install tldr` on Debian and Ubuntu
 - [Node.js client](https://github.com/tldr-pages/tldr-node-client):
   `npm install -g tldr`
 - [OCaml client](https://github.com/RosalesJ/tldr-ocaml): `opam install tldr`
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ You can access these pages on your computer using one of the following clients:
     [App Store](https://itunes.apple.com/us/app/tldt-pages/id1071725095?ls=1&mt=8)
 - [Haskell client](https://github.com/psibi/tldr-hs):
   `stack install tldr`
+  or `apt-get install tldr` on Debian
 - [Node.js client](https://github.com/tldr-pages/tldr-node-client):
   `npm install -g tldr`
 - [OCaml client](https://github.com/RosalesJ/tldr-ocaml): `opam install tldr`
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ You can access these pages on your computer using one of the following clients:
   - [tldr-python-client](https://github.com/tldr-pages/tldr-python-client):
     `pip install tldr` or `pacman -S tldr` on Arch Linux
   - [tldr.py](https://github.com/lord63/tldr.py):
-    `pip install tldr.py`
+    `pip install tldr.py` or `apt-get install tldr-py` on Ubuntu
 - [R client](https://github.com/kirillseva/tldrrr):
   `devtools::install_github('kirillseva/tldrrr')`
 - [Ruby client](https://github.com/YellowApple/tldrb):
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -96,6 +96,7 @@ You can access these pages on your computer using one of the following clients:
     [App Store](https://itunes.apple.com/us/app/tldt-pages/id1071725095?ls=1&mt=8)
 - [Haskell client](https://github.com/psibi/tldr-hs):
   `stack install tldr`
+  or `apt-get install tldr` on Debian
```
zdroid: Does this work on Ubuntu too? (I guess it does)
It should be noted in some way. :p

ivanhercaz: @zdroid, I search in [Ubuntu Packages](https://packages.ubuntu.com/search?keywords=Tldr&searchon=names&suite=eoan&section=all) and yes, in both correspond with the Haskell client. As soon as possible I will note this and the [one for the Python client](https://packages.ubuntu.com/eoan/tldr-py), although it seems a packaged version of lord63/tldr.py, which its last commit was at November 27, 2017.

I will change the title of the PR too.

ivanhercaz: @zdroid, look at the new changes in which I added the improvement you suggested.

ivanhercaz:
@zdroid, look at the new changes in which I added the improvement you suggested.

ivanhercaz:
> Personally I think that the original way was more universal, which is why it was used. It works cross-platform.

I understand you, but I think it is very useful to have packaged versions for different systems. Install them are easier, at least in my opinion, for uses who don't work with the package managers of the different languages.

I also propose it because I saw there are already one with installation instructions for Arch Linux.

pxgamer:
Fair enough. :+1: I have no problem with this being merged.

README.md
sbrl: Ubuntu is based on Debian, so it's perhaps best to mention the base distribution here - since there are lots of different Debian-based distributions - we wouldn't have space to mention them all :P

README.md
```diff
@@ -96,6 +96,7 @@ You can access these pages on your computer using one of the following clients:
     [App Store](https://itunes.apple.com/us/app/tldt-pages/id1071725095?ls=1&mt=8)
 - [Haskell client](https://github.com/psibi/tldr-hs):
   `stack install tldr`
+  or `apt-get install tldr` on Debian and Ubuntu
```
sbrl: Why mention `Ubuntu` specifically here? It implies that other Debian-based distros (e.g. Linux Mint or [Deepin](https://www.deepin.org/en/)) are excluded. How about this?
```suggestion
  or `apt-get install tldr` on Debian-based distributions such as Ubuntu
```

README.md
```diff
@@ -107,7 +108,7 @@ You can access these pages on your computer using one of the following clients:
   - [tldr-python-client](https://github.com/tldr-pages/tldr-python-client):
     `pip install tldr` or `pacman -S tldr` on Arch Linux
   - [tldr.py](https://github.com/lord63/tldr.py):
-    `pip install tldr.py`
+    `pip install tldr.py` or `apt-get install tldr-py` on Ubuntu
```
sbrl: Does the `apt-get` command work on other Debian-based distributions? If so, it would be worth using `Debian-based distributions` here instead.
```suggestion
    `pip install tldr.py` or `apt-get install tldr-py` on Debian-based distributions
```

README.md
zdroid:
```suggestion
  or `apt-get install tldr` on Debian-based distributions
```
is enough

README.md
ivanhercaz: That's true. I will commit your suggestions @sbrl.

README.md
ivanhercaz: That's true. I will commit your suggestions @sbrl, @zdroid.

zdroid:
@pxgamer is this ready for merge?

pxgamer:
@zdroid, LGTM. :+1:

ivanhercaz:
Ups, excuse me everyone @sbrl, @pxgamer and @zdroid, I merged this PR that I opened. I understand that it is recommended to not merge our own PRs, so excuse me, it was a mistake.

pxgamer:
That's fine IMO, @ivanhercaz. It was a pretty trivial change. :+1:

zdroid:
@ivanhercaz You cannot go worse than me, making an issue proposing myself as an organisation member, accepting it, making a PR and merging the PR. :smile:

PR #3737, opened 2020-01-05
Add organization member: @ivanhercaz

### LINKED ISSUE #3736: Add organization member: @ivanhercaz
Hi, @ivanhercaz! After joining as a collaborator in the repository, you have been regularly performing maintenance tasks.
Thank you for that!
According to [COMMUNITY-ROLES.md](https://github.com/tldr-pages/tldr/blob/master/COMMUNITY-ROLES.md), you've now met the thresholds to be effectively considered an active maintainer of the project (#3700 #3714 #3713 #3720 #3722 #3723 #3729 #3728 #3735 #3731).
To publicly acknowledge that fact, we'd like to add you to the tldr-pages organization.

If you accept the invitation, we ask you to make your membership public, and (in case you don't already) start hanging out in our Gitter chat room.
You'll now be one of the public faces of the tldr-pages project.

How does that sound? Are you up for it?

zdroid:
Closes #3736.

It also fixes some petty issues of improper dashes. ~In case Ivan rejects (he better not :laughing:), I'll keep that part only.~ He accepted.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -30,19 +30,19 @@ If you are an owner of the organization, you should be able to see an automated 
 - **Lucas Schneider ([@schneiderl](https://github.com/schneiderl))**:
   [11 April 2019](https://github.com/tldr-pages/tldr/issues/2898) — present
 - **Andrik Albuquerque ([@andrik](https://github.com/andrik))**:
-  [9 May 2019](https://github.com/tldr-pages/tldr/issues/2988) - present
+  [9 May 2019](https://github.com/tldr-pages/tldr/issues/2988) — present
 - **Ivan Aracki ([@Aracki](https://github.com/Aracki))**:
-  [9 May 2019](https://github.com/tldr-pages/tldr/issues/2988) - present
+  [9 May 2019](https://github.com/tldr-pages/tldr/issues/2988) — present
 - **Ein Verne ([@einverne](https://github.com/einverne))**:
-  [29 October 2019](https://github.com/tldr-pages/tldr/issues/3488) - present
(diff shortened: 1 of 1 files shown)


MAINTAINERS.md
```diff
@@ -30,19 +30,19 @@ If you are an owner of the organization, you should be able to see an automated
 - **Lucas Schneider ([@schneiderl](https://github.com/schneiderl))**:
   [11 April 2019](https://github.com/tldr-pages/tldr/issues/2898) — present
 - **Andrik Albuquerque ([@andrik](https://github.com/andrik))**:
-  [9 May 2019](https://github.com/tldr-pages/tldr/issues/2988) - present
+  [9 May 2019](https://github.com/tldr-pages/tldr/issues/2988) — present
```
sbrl: We should keep the ASCII dashes here I think, @zdroid. This would be consistent with our existing approach to the pages themselves, where we prefer ASCII quotes instead of smart ones, for instance.

zdroid: This file isn't supposed to be viewed as plain text, but as processed Markdown. It's also provided the file will be viewed on GitHub's web interface, which supports this style. Also, it's not appropriate to use hyphens for this kind of indication.

mebeim: I think it's ok to use em dashes here. I didn't notice them from the raw text, but it makes a difference when viewing the rendered document, em dashes look way better IMHO.

sbrl: Hrm, I guess that makes sense. This is still markdown, it's just a case of being dependent on where it's going to be viewed.

zdroid:
Welcome! :tada:

PR #3739, opened 2020-01-06
numlockx: add page

sbrl:
To solve that really annoying problem where numlockx is always turned off when you boot Ubuntu (sorry wayland users, you'll have to keep looking).

### DIFF
--- a/pages/linux/numlockx.md
+++ b/pages/linux/numlockx.md
@@ -1,0 +1,20 @@
+# numlockx
+
+> Control the number lock key status in X11 sessions.
+> More information: <http://www.mike-devlin.com/linux/README-numlockx.htm>.
+
+- Show the current number lock status:
+
+`numlockx status`
+
+- Turn the number lock on:
+
+`numlockx on`
+
+- Turn it off:
+
+`numlockx off`
+
+- Toggle the current state:
+
+`numlockx toggle`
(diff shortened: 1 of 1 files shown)


pages/linux/numlockx.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Show the current number lock status:
+
+`numlockx status`
+
+- Turn the number lock on:
+
+`numlockx on`
+
+- Turn it off:
```
zdroid:
```suggestion
- Turn the number lock off:
```
let's stay consistent

sbrl: Do we need to specify that it's the number lock here, since we've already said so in the previous example? It's not like it's going to be anything else. On other tldr pages we have generally tended to be more concise in later examples.

For example, if command `foo` always `bars` a `baz`, we would first say `Bar a specfied baz:`. In later examples we might simply say `Also do X:`.

zdroid: This way it's more suitable for scrolling and various display formats. The way you've written it, someone has to go a few lines up to understand what the line is about.

sbrl: Ok I guess.

mebeim:
Merging since this already has four approvals and @zdroid's commend has been addressed.

PR #3744, opened 2020-01-09
top: update page

einverne:

### DIFF
--- a/pages/linux/top.md
+++ b/pages/linux/top.md
@@ -14,6 +14,10 @@
 
 `top -u {{user_name}}`
 
+- Show individual threads of given process:
+
+`top -Hp {{process_id}}`
+
 - Show only the processes with the given PID(s), passed as a comma-separated list. (Normally you wouldn't know PIDs off hand. This example picks the PIDs from the process name):
 
 `top -p $(pgrep -d ',' {{process_name}})`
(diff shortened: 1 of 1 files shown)


Aracki:
Is it possible to achieve this on other platforms? @einverne

einverne:
> Is it possible to achieve this on other platforms? @einverne

Sorry @Aracki , I don't have a mac, so I cannot test it. We can keep watch, and let some work under mac to verify it.

Aracki:
I have tried with osx and it's not working. I was wondering what about windows.

pages/linux/top.md
```diff
@@ -14,6 +14,10 @@
 
 `top -u {{user_name}}`
 
+- Show individual threads of given process:
```
zdroid:
```suggestion
- Show the individual threads of a given process:
```

sbrl:
Hrm, if it doesn't work on macOS then we may have to have a macOS-specific copy of the page with the  example removed.

mebeim:
Not sure what's the problem here, the page is already under `linux`, the fact that this does not work on macOS is not a problem, there already is [`pages/osx/top.md`](https://github.com/tldr-pages/tldr/blob/master/pages/osx/top.md). LGTM.

sbrl:
Ah, I wasn't aware that we already had a macOS-specific version of the page. That makes it all ok then :P

PR #3339, opened 2019-10-06
Translate git commands to Italian - part 2 

glenacota:
Ten more Italian translations to git commands. 🇮🇹 
Related issue: #3317

- [x] The pages are new
- [x] The pages are in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The pages has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #3727, opened 2020-01-02
[WIP] git* commands: add Spanish translation

ivanhercaz:
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [ ] The page description includes a link to documentation or a homepage (if applicable). **Note**. This one will depend of each command, so I will mark it when I have all the commands translated.

This PR is a work in progress with the purpose to translate to Spanish all the `git*` commands. There are 49 untranslated commands if I am not wrong, so if you think I should translate just a batch of 10 or 20 commands, warn me to keep in mind and stop to work in this PR when I reach the recommended limit for a batch in the PR.

PR #3746, opened 2020-01-10
Translate git commands to Italian - part 2 (revived)

mebeim:
Just a copy of #3339, with my suggested changes applied. Unfortunately I was not able to apply changes on the PR itself, and since it's been sitting there for quite some time, I closed it to open this one. All commits also have @glenacota as author.

Just count my "auto-approve" on this since the PR is basically not mine.

### DIFF
--- a/pages.it/common/git-am.md
+++ b/pages.it/common/git-am.md
@@ -1,6 +1,6 @@
 # git am
 
-> Applica file di patch. Utile quando si ricevono commits via email.
+> Applica file di patch. Utile quando si ricevono commit via email.
 > Vedi anche `git format-patch` per generare file di patch.
 > Maggiori informazioni: <https://git-scm.com/docs/git-am>.
 
--- a/pages.it/common/git-clone.md
+++ b/pages.it/common/git-clone.md
@@ -1,0 +1,24 @@
+# git clone
+
+> Clona un repository esistente.
+> Maggiori informazioni: <https://git-scm.com/docs/git-clone>.
+
+- Clona un repository remoto esistente:
+
+`git clone {{url_repository_remoto}}`
+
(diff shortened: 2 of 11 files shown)


ivanhercaz:
I labeled this PR as `page edit` because there are commits with changes in pages that already exist. I think the `mass changes` label i self-explanatory.

zdroid:
Grazie mille @mebeim

mebeim:
@ivanhercaz we use the mass changes label when the PR edits a significant number of pages, this is not the case though. I know an addition of several translation is technically speaking is a mass change, but it's not necessary to label it as such.

zdroid:
Judging by the past use of the label, it seems we use it for both purposes.

mebeim:
@zdroid there are some PRs in which this is applied for the same reason I explained above, which is the translation of multiple pages, but I would really try to avoid it. Labels are meant to make searching easier, this only confuses things as translations are not a "mass change" of existing content.

waldyrious:
@mebeim just to be clear, mass-change PRs are not intended for any regular page edits that happen to be done all at once — for those, we'd prefer separate PRs. The typical use case is for edits that apply across multiple pages but are closely related to each other, such as a typo fix or a formatting change, but also adding multiple pages from the same source (#702 comes to mind). In that regard, touching multiple pages as part of a translation round sounds like a valid use case for the label.

Perhaps the word "mass" is too loaded for the actual meaning of the tag; maybe "bulk changes" would work better? Or is your concern about the "change" as opposed to "addition"? In that case, we could go for the more explicit "multiple pages". Let me know what you think!

/cc @ivanhercaz, due to [this comment](https://github.com/tldr-pages/tldr/pull/3727#issuecomment-573467508).

PR #3754, opened 2020-01-13
Fix tldr-bot messages

mebeim:
GitHub wraps text in comments even when a single newline is encountered. This caused @tldr-bot messages to be formatted like this:

> The [build](https://travis-ci.org/tldr-pages/tldr/builds/635578923)
> for this PR failed with the following error(s):
> ...

Which does not look good. They should instead be formatted like this:

> The [build](https://travis-ci.org/tldr-pages/tldr/builds/635578923) for this PR failed with the following error(s):
> ...

This PR removes the unneeded newlines in the script.


zdroid:
I'm merging this to see if the fix is working.

PR #3755, opened 2020-01-14
pkgutil: add page (osx)

notpeter:

### DIFF
--- a/pages/osx/pkgutil.md
+++ b/pages/osx/pkgutil.md
@@ -10,6 +10,6 @@
 
 `pkgutil --check-signature {{filename.pkg}}`
 
-- List all the files for an installed package (e.g. Microsoft Word):
+- List all the files for an installed package (e.g. Microsoft Word) given its ID:
 
 `pkgutil --files {{com.microsoft.Word}}`
--- a/pages/osx/pkgutil.md
+++ b/pages/osx/pkgutil.md
@@ -1,0 +1,15 @@
+# pkgutil
+
+> Query and manipulate Mac OS X Installer packages and receipts.
+
+- List package ids for all installed packages:
+
+`pkgutil --pkgs`
+
+- Verify cryptographic signatures of a package file:
+
+`pkgutil --check-signature {{filename.pkg}}`
+
+- List all the files for an installed package:
+
+`pkgutil --files {{com.microsoft.Word}}`
--- a/pages/osx/pkgutil.md
+++ b/pages/osx/pkgutil.md
@@ -10,6 +10,6 @@
 
 `pkgutil --check-signature {{filename.pkg}}`
 
-- List all the files for an installed package:
+- List all the files for an installed package (e.g. Microsoft Word):
 
 `pkgutil --files {{com.microsoft.Word}}`
--- a/pages/osx/pkgutil.md
+++ b/pages/osx/pkgutil.md
@@ -2,7 +2,7 @@
 
 > Query and manipulate Mac OS X Installer packages and receipts.
 
-- List package ids for all installed packages:
+- List package IDs for all installed packages:
 
 `pkgutil --pkgs`
 
--- a/pages/osx/pkgutil.md
+++ b/pages/osx/pkgutil.md
@@ -10,6 +10,6 @@
 
 `pkgutil --check-signature {{filename.pkg}}`
 
-- List all the files for an installed package (e.g. Microsoft Word) given its ID:
+- List all the files for an installed package given its ID:
 
 `pkgutil --files {{com.microsoft.Word}}`


pages/osx/pkgutil.md
```diff
@@ -0,0 +1,15 @@
+# pkgutil
+
+> Query and manipulate Mac OS X Installer packages and receipts.
+
+- List package ids for all installed packages:
```
zdroid:
```suggestion
- List package IDs for all installed packages:
```

pages/osx/pkgutil.md
```diff
@@ -0,0 +1,15 @@
+# pkgutil
+
+> Query and manipulate Mac OS X Installer packages and receipts.
```
zdroid: Is the program's name really Mac OS X Installer? I can't find the name anywhere.

pages/osx/pkgutil.md
```diff
@@ -0,0 +1,15 @@
...
+
+- List package ids for all installed packages:
+
+`pkgutil --pkgs`
+
+- Verify cryptographic signatures of a package file:
+
+`pkgutil --check-signature {{filename.pkg}}`
+
+- List all the files for an installed package:
```
zdroid: either `for Microsoft Word` here or change it to something general in the line under

pages/osx/pkgutil.md
notpeter: Yes. image

pages/osx/pkgutil.md
```diff
@@ -10,6 +10,6 @@
 
 `pkgutil --check-signature {{filename.pkg}}`
 
-- List all the files for an installed package:
+- List all the files for an installed package (e.g. Microsoft Word):
```
zdroid: I'd rather wait for the opinions of others on this example. Seems good to me, though.

notpeter:
Yes, these are in what I believe the likely order of potential utility. No, I don't think providing a regex filtering example is appropriate. 

More importantly the feedback I've received is basically bikeshedding:
* add "Microsoft Word"
* add "given it's ID"
* consider reordering
* consider adding an example

This sort of feedback in PRs is what caused me to [stop contributing](https://github.com/tldr-pages/tldr/commits?author=notpeter) despite having spent dozens of hours writing examples and having authored nearly a hundred commands from scratch.

You may disagree, but having three people suggest changes, two of whom withhold their PR approval is a terrible way to encourage project participation. I saddened to see that 3+ years later and maintainers of this project are still contributor-hostile gatekeepers. Especially in response to new content which meets all style/quality guidelines. @zdroid specifically: don't request a change and then  say  "I'd rather wait for the opinions of others on this example" when I act on that change. It's a waste of everyones time. 

I'd highly recommend a much faster path to merge. Responding to three rounds of feedback over multiple days is pointless.  Things like changing "ids" to "IDs" could've been silently done at merge time. The health of the community is more important than any particular merge.

While I appreciate your contributions/time to this project @zdroid please ask yourself three questions when commenting on a PR:

* Is this my comment identifying error or expressing my opinion? Is that clear in my language?
* Is this suggesting a change or requiring a change? If the former, shouldn't I "approve"?   
* Will my comments make someone more or less likely to contribute in the future ?

Feel free to merge. I'm done with this PR. I won't be commenting further.

pages/osx/pkgutil.md
mebeim: It's ok to not explicitly mention the package in the command description and use a real value in the user token of the example command. We in fact do encourage explicit examples over more general tokens when possible. This seems as is (currently without the added parentheses).

mebeim:
I'm sorry to hear that @notpeter. 

Due to the nature of the project, there is no single point of truth when it comes to suggesting changes and reviewing PRs, and some proposed changes might be more subjective than others. Other than requiring more time to merge, I don't really see the harm of having multiple people suggest changes to a PR. I personally always try to get the original author's attention on proposed changes instead of applying them myself because they most often know better than me what the tool is about (and I think most of the other reviewers do the same for the same reason). Authors and maintainers are equally valuable contributors to the project.

We are not *demanding* anything, all contributions (including maintenance) are voluntary work. Reviewing a PR as "needs changes" shouldn't be seen merely as withholding an approval. Moreover, it should not be seen as a negative critique to a contribution: that is *never* our intention. The ultimate goal we all want to accomplish is to make examples as clear and correct as possible. If you are unable or don't feel like addressing review comments we can either wait (if the problem is time, then there is no rush) or sort it out by ourselves, but of course this can only happen if you let us know explicitly. I would like to clarify that I did not mean to impose my opinion or changes. My intention was to ask rather than to demand something. 

About the ordering, I was only asking for a confirmation since I do not use the tool (thanks for confirming), as it sometimes happens that when writing a page one may forget about re-ordering (we've had countless PRs re-ordering previously written pages exactly for this reason). As per the regexp example, may I ask why you don't see it as appropriate? If you don't want to comment any further as you said, then no problem, I will investigate this myself.

Granted, this was not a complicated page, but I nonetheless think that requesting intervention from the original author is better than applying changes unconditionally. Most of the times it is just a matter of a couple of clicks in the GitHub web UI anyway. Tidying up a page before merging is way simpler than having to open subsequent PRs for the same page (which one could even forget to do), and that's also why I was asking about the additional example rather than waiting for the PR to be merged creating a new one afterwards. Even if it's a minor detail like the capitalization of a word, there is no need to get things merged in record time. Optimizing for time is clearly not the goal here, we would just end up with a bunch of unpolished pages.

With the above being said, I thank you for the honest feedback, and I'll keep your suggestions in mind for future reviews.

PS: looks good to merge to me now that my comment was addressed.

sbrl:
Merging, since everyone's comments have been addressed.

PR #3478, opened 2019-10-25
ab, ack: add Tamil translation

MohamedSabthar:
- [ ] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #3761, opened 2020-01-15
ab, ack: add Tamil translation (revived)

mebeim:
Just a copy of #3478, with the suggested changes applied. Unfortunately it was not possible to apply changes on the PR itself, and since it's been sitting there for quite some time, I closed it to open this one. All commits have @MohamedSabthar as author.


zdroid:
Your revival capabilities are astounding. Keep up the good work!

PR #3770, opened 2020-01-18
cmus: update command

syrofoam:
Mutch needed kb shortcuts.

### DIFF
--- a/pages/linux/cmus.md
+++ b/pages/linux/cmus.md
@@ -11,6 +11,14 @@
 
 `:add {{path/to/file_or_directory}}`
 
+- next song:
+
+`b`
+
+- previous song:
+
+`z`
+
 - Pause/unpause current song:
 
 `c`
--- a/pages/linux/cmus.md
+++ b/pages/linux/cmus.md
@@ -4,13 +4,9 @@
 > Use arrow keys to navigate, `<enter/return>` to select, and numbers 1-8 switch between different views.
 > More information: <https://cmus.github.io/>.
 
-- Open cmus from specified directory:
-
-`cmus {{path/to/directory}}`
-
 - Add file/directory to library:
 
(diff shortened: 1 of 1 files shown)


pages/linux/cmus.md
```diff
@@ -11,6 +12,14 @@
```
zdroid: Your addition itself is good, but as we're here, I suggest to change `file/directory` above to `file or directory`.

pages/linux/cmus.md
```diff
@@ -24,7 +24,7 @@
 
 `s`
 
-- Turn volume up/down and seeking
+- Turn volume up/down and seeking:
```
zdroid: I can't understand the meaning behind this, and it also seems `seeking` doesn't agree gramatically with the rest of the sentence. What's the context of `seeking` here?

pages/linux/cmus.md
```diff
@@ -2,14 +2,19 @@
 
 > Commandline Music Player.
```
sbrl: Perhaps this will read better?
```suggestion
> Command-line music player.
```

pages/linux/cmus.md
agnivade: The up/down arrow keys are for volume. And the left/right keys are for seeking.

I would separate the two into separate examples.

pages/linux/cmus.md
```diff
@@ -2,14 +2,19 @@
...
 > Use arrow keys to navigate, `<enter/return>` to select, and numbers 1-8 switch between different views.
+> More information: <https://cmus.github.io/>.
 
-- Open cmus from specified directory:
+- Add file/directory to library:
```
agnivade: The first example should always invoke the command. Otherwise, one wouldn't know where to enter the commands.

PR #3771, opened 2020-01-20
Update nano.md

nire0510:
Start nano in terminal with {filename} at line {line}, column {column}.

### DIFF
--- a/pages/common/nano.md
+++ b/pages/common/nano.md
@@ -1,6 +1,6 @@
 # nano
 
-> Simple, easy to use editor. An enhanced, free Pico clone.
+> Simple, easy to use command-line text editor. An enhanced, free Pico clone.
 > More information: <https://nano-editor.org>.
 
 - Start nano in terminal with a specified filename:
--- a/pages/common/nano.md
+++ b/pages/common/nano.md
@@ -5,7 +5,7 @@
 
 - Start nano in terminal with a specified filename:
 
-`nano {{filename}}`
+`nano {{path/to/file}}`
 
 - Open a file positioning the cursor at the specified line and column:
 
--- a/pages/common/nano.md
+++ b/pages/common/nano.md
@@ -7,6 +7,10 @@
 
 `nano {{filename}}`
 
+- Start nano in terminal with {filename} at line {line}, column {column}:
+
+`nano +{{line}},{{column}} {{filename}}`
+
 - Enable smooth scrolling:
 
 `nano -S {{filename}}`
--- a/pages/common/nano.md
+++ b/pages/common/nano.md
@@ -3,7 +3,7 @@
 > Simple, easy to use editor. An enhanced, free Pico clone.
 > More information: <https://nano-editor.org>.
 
-- Start nano in terminal with {filename}:
+- Start nano in terminal with a specified filename:
 
 `nano {{filename}}`
 
--- a/pages/common/nano.md
+++ b/pages/common/nano.md
@@ -9,7 +9,7 @@
 
 - Start nano in terminal with filename at specified line and column:
 
-`nano +{{line}},{{column}} {{filename}}`
+`nano +{{line}},{{column}} {{path/to/file}}`
 
 - Enable smooth scrolling:
 
--- a/pages/common/nano.md
+++ b/pages/common/nano.md
@@ -3,7 +3,7 @@
 > Simple, easy to use command-line text editor. An enhanced, free Pico clone.
 > More information: <https://nano-editor.org>.
 
-- Start nano in terminal with a specified filename:
+- Open a specific file:
 
 `nano {{path/to/file}}`
 
--- a/pages/common/nano.md
+++ b/pages/common/nano.md
@@ -7,7 +7,7 @@
 
 `nano {{filename}}`
 
-- Start nano in terminal with filename at specified line and column:
+- Open a file positioning the cursor at the specified line and column:
 
 `nano +{{line}},{{column}} {{path/to/file}}`
 
--- a/pages/common/nano.md
+++ b/pages/common/nano.md
@@ -7,7 +7,7 @@
 
 `nano {{filename}}`
 
-- Start nano in terminal with filename at line,column:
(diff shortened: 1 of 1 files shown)


pages/common/nano.md
```diff
@@ -7,6 +7,10 @@
 
 `nano {{filename}}`
 
+- Start nano in terminal with {filename} at line {line}, column {column}:
```
schneiderl: Let`s remove these curly brackets from the example description. 
Having them only on the example is enough because the example itself makes it pretty intuitive.

nire0510: NP, done (please note there are curly brackets on first example, as well)

pages/common/nano.md
```diff
@@ -7,7 +7,7 @@
 
 `nano {{filename}}`
 
-- Start nano in terminal with {filename} at line {line}, column {column}:
+- Start nano in terminal with filename at line,column:
```
zdroid: Perhaps `at specified line and column` will read better?

schneiderl: I agree with @zdroid's suggestion.

schneiderl:
Since we are already here I took the freedom to also remove the curly brackets from the first description.

pages/common/nano.md
```diff
@@ -3,10 +3,14 @@
 > Simple, easy to use editor. An enhanced, free Pico clone.
 > More information: <https://nano-editor.org>.
 
-- Start nano in terminal with {filename}:
+- Start nano in terminal with a specified filename:
```
mebeim:
```suggestion
- Start nano opening a specific file:
```

pages/common/nano.md
```diff
@@ -3,10 +3,14 @@
...
 
 `nano {{filename}}`
```
mebeim:
```suggestion
`nano {{path/to/file}}`
```

pages/common/nano.md
```diff
@@ -3,10 +3,14 @@
...
 
+- Start nano in terminal with filename at specified line and column:
```
mebeim:
```suggestion
- Start nano opening a file and positioning the cursor at the specified line and column:
```

pages/common/nano.md
```diff
@@ -3,10 +3,14 @@
...
+
+`nano +{{line}},{{column}} {{filename}}`
```
mebeim:
```suggestion
`nano +{{line}},{{column}} {{path/to/file}}`
```

pages/common/nano.md
```diff
@@ -3,10 +3,14 @@
 > Simple, easy to use editor. An enhanced, free Pico clone.
```
mebeim:
```suggestion
> Simple, easy to use command-line text editor. An enhanced, free Pico clone.
```

pages/common/nano.md
pxgamer: Thinking about this, it probably isn't necessary to mention nano in each example. So they could probably just be:

- Open a specific file:
- Open a file with the cursor at a specific line:
- etc.

mebeim: @pxgamer true, updated.

pages/common/nano.md
zdroid: `- Open a file starting at the specified line and column:`

Perhaps this could be more clear, but I'm unsure. The suggestion you made is quite long.

mebeim: I don't think that's too long. It's medium sized. Just "starting at the specified line and column" is strangely worded, the command doesn't "start" anything.

pages/common/nano.md
```diff
@@ -1,11 +1,15 @@
...
 
-- Start nano in terminal with {filename}:
+- Open a specific file:
 
-`nano {{filename}}`
+`nano {{path/to/file}}`
+
+- Open a file positioning the cursor at the specified line and column:
+
+`nano +{{line}},{{column}} {{path/to/file}}`
```
zdroid:
```suggestion
`nano +{{line}},{{column}} {{/path/to/file}}`
```

pages/common/nano.md
```diff
@@ -1,11 +1,15 @@
...
 
-> Simple, easy to use editor. An enhanced, free Pico clone.
+> Simple, easy to use command-line text editor. An enhanced, free Pico clone.
 > More information: <https://nano-editor.org>.
 
-- Start nano in terminal with {filename}:
+- Open a specific file:
 
-`nano {{filename}}`
+`nano {{path/to/file}}`
```
zdroid:
```suggestion
`nano {{/path/to/file}}`
```

pages/common/nano.md
```diff
@@ -1,11 +1,15 @@
 # nano
 
-> Simple, easy to use editor. An enhanced, free Pico clone.
+> Simple, easy to use command-line text editor. An enhanced, free Pico clone.
 > More information: <https://nano-editor.org>.
 
-- Start nano in terminal with {filename}:
+- Open a specific file:
```
zdroid:
```suggestion
- Open a file:
```

pages/common/nano.md
mebeim: This should not be added. The path does not need to be absolute and this seems to imply it.

pages/common/nano.md
mebeim: Same here.

pages/common/nano.md
mebeim: I don't think this is an improvement, it's correct to say that the file is the one specified, and not just "a file".

zdroid: Okay, I edited it.

mebeim: Still no improvement though. "Open a specific file" sounds better to me than "Open the specified file". It describes the general operation, using "a" instead of "the". We use "a specific" a lot more than "the specified" through all pages.

zdroid: Perhaps you're right. I judged this according to some recent edits, so my view is probably skewed.

mebeim:
Merging since all comments have been addressed and there already are 4 approvals.

PR #3778, opened 2020-01-22
git-add: add git add interactively for all changed files

ryanolsonx:

### DIFF
--- a/pages/common/git-add.md
+++ b/pages/common/git-add.md
@@ -19,6 +19,10 @@
 
 `git add -f`
 
-- Add parts of a file interactively:
+- Interactively stage parts of changes to your tracked files:
+
+`git add -p`
+
+- Interactively stage parts of a file:
 
 `git add -p {{path/to/file}}`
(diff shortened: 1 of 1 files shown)


pages/common/git-add.md
```diff
@@ -19,6 +19,10 @@
 
 `git add -f`
 
-- Add parts of a file interactively:
+- Interactively stage hunks of your tracked files:
```
zdroid: Isn't `parts` a better word than `hunks` in this context? It's easier to understand.

ryanolsonx: That's a fair point.  I switched to hunks because that's what terminology git likes to use.  I'll switch to parts again.  It seems more clear to someone not as familiar with git.

pages/common/git-add.md
```diff
@@ -19,6 +19,10 @@
 
 `git add -f`
 
-- Add parts of a file interactively:
+- Interactively stage parts of changes to your tracked files:
```
agnivade: I would keep the description short and similar to the previous one. Can we change this to - "Interactively stage parts of files", and make the next one "Interactively stage parts of a given file" ?

PR #3783, opened 2020-01-22
vim: helpful additions

### LINKED ISSUE #3779: TLDR for Vim has too many tips about things inside Vim
When I look at `tldr vim`, I see that it has many tips that tell you how to do things in Vim, rather than using the `vim` command itself.  I feel like there are more tips that we can add that are really useful. It's already at its max (of 8) items.

What should we do?

ryanolsonx:

Closes #3779

### DIFF
--- a/pages/common/vim.md
+++ b/pages/common/vim.md
@@ -8,6 +8,10 @@
 
 `vim {{path/to/file}}`
 
+- View Vim's help manual:
+
+`:help<Enter>`
+
 - Save a file:
 
 `:write<Enter>`
@@ -19,10 +23,6 @@
 - Open a file at a specified line number:
 
 `vim +{{line_number}} {{path/to/file}}`
-
-- View Vim's help manual:
-
-`:help<Enter>`
 
 - Undo the last operation:
 
--- a/pages/common/vim.md
+++ b/pages/common/vim.md
@@ -1,20 +1,28 @@
 # vim
 
-> Vi IMproved, a programmer's text editor, provides several modes for different kinds of text manipulation.
+> Vim (Vi IMproved), a programmer's command-line text editor, provides several modes for different kinds of text manipulation.
 > Pressing `i` enters edit mode. `<Esc>` goes back to normal mode, which doesn't allow regular text insertion.
 > More information: <https://www.vim.org>.
 
 - Open a file:
 
-`vim {{file}}`
+`vim {{path/to/file}}`
 
-- Enter text editing mode (insert mode):
+- Save a file:
 
-`<Esc>i`
+`:write`
 
-- Copy ("yank") or cut ("delete") the current line (paste it with `P`):
+- Quit without saving:
 
-`<Esc>{{yy|dd}}`
+`<Esc>:quit!<Enter>`
+
+- Open a file at a specified line number:
+
+`vim +{{line_number}} {{path/to/file}}`
+
+- View help for setting (such as {{set number}}):
+
+`<Esc>:help '{{setting_name}}'`
 
 - Undo the last operation:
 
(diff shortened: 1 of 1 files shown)


pages/common/vim.md
```diff
@@ -1,16 +1,28 @@
 # vim
 
-> Vi IMproved, a programmer's text editor, provides several modes for different kinds of text manipulation.
+> Vim (Vi IMproved), a programmer's command line text editor, provides several modes for different kinds of text manipulation.
```
zdroid: `command-line` here

pages/common/vim.md
```diff
@@ -1,16 +1,28 @@
...
+`<Esc>:q!<Enter>`
+
+- Open a file at a specified line number:
+
+`vim +{{line number}} {{path/to/file}}`
+
+- View help for setting (such as {{set number}}):
 
-`<Esc>i`
+`<Esc>:help '{{setting name}}'`
```
zdroid:
```suggestion
`<Esc>:help '{{setting_name}}'`
```

pages/common/vim.md
```diff
@@ -1,16 +1,28 @@
...
+`vim {{path/to/file}}`
 
-- Enter text editing mode (insert mode):
+- Quit without saving:
+
+`<Esc>:q!<Enter>`
+
+- Open a file at a specified line number:
+
+`vim +{{line number}} {{path/to/file}}`
```
zdroid:
```suggestion
`vim +{{line_number}} {{path/to/file}}`
```

ryanolsonx:
Thanks for the requested changes.  I'll fix all of those up.

After adding my tips, Vim now has 10 tips.  What do we do about that?

zdroid:
2 of the examples have to be removed, as 8 is the limit. For example, `Open without your custom configurations` and `View help for setting (such as {{set number}}):` appear less useful than the other examples. However, you should wait for other opinions on removing those.

ryanolsonx:
I would say that finding vim documentation (using `:help`) is probably one of the most important things to learn when looking at the tldr.

I can agree that the custom configurations one isn't as important.

zdroid:
You may be quite right, as I'm not a vim user at all, so I have no reliable knowledge on that. I'm glad we agreed on one example so far, so you have to find one more to remove.

ryanolsonx:
I have my proposed vim.md file all set now.

pages/common/vim.md
```diff
@@ -1,20 +1,28 @@
...
 - Open a file:
 
-`vim {{file}}`
+`vim {{path/to/file}}`
 
-- Enter text editing mode (insert mode):
+- Save a file:
 
-`<Esc>i`
+`:write`
```
zergov: Would it be better to omit the `<Esc>` key like this command, for all commands in this file? 
I feel like they should all include it or not include the `<Esc>` at all.

I feel like putting the `<Esc>`key in those definitions assumes that the user is currently in a mode other than `normal`. I feel like only the actual `:commands` should be there.

What do you think?

ryanolsonx: I agree with you.  I'll make the change in my PR.

pages/common/vim.md
```diff
@@ -1,37 +1,37 @@
...
+`:quit!<Enter>`
 
-- Undo the last operation:
+- Open a file at a specified line number:
 
-`<Esc>u`
+`vim +{{line_number}} {{path/to/file}}`
 
-- Search for a pattern in the file (press `n`/`N` to go to next/previous match):
+- View Vim's help manual:
```
sbrl: Perhaps this should be the second example?

ryanolsonx: Great idea. I'll change that in the morning.

ryanolsonx:
Is this ready for merge?  Can I get another reviewer's approval?

ryanolsonx:
@sbrl Thank you!

PR #3788, opened 2020-01-23
efibootmgr: add page

sp1ritCS:

As you might have already seen, I included a specific example on creating a new boot option. I believe that in this case It is simpler to understand it with an extra example (because of the weird fat \ instead of / and it is possible that adding the UEFI Shell as a Bootoption.

### DIFF
--- a/pages/linux/efibootbgr.md
+++ b/pages/linux/efibootbgr.md
@@ -1,15 +1,15 @@
 # efibootmgr
 
 > Manipulate the UEFI Boot Manager (the Bootoptions).
-> 
-> More Information: https://linux.die.net/man/8/efibootmgr
+>
+> More Information: https://linux.die.net/man/8/efibootmgr .
 
 - List the current settings / bootnums:
 
 `efibootmgr`
 
 - List the Filepaths:
-  
+
 `efibootmgr -v`
 
 - Add UEFI Shell v2 as a boot option:
--- a/pages/linux/efibootbgr.md
+++ b/pages/linux/efibootbgr.md
@@ -18,6 +18,6 @@
 
 `sudo efibootmgr -o {{0002,0008,0001,0005}}`
 
-- Delete Bootoption:
+- Delete a boot option:
 
 `sudo efibootmgr -b {{bootnum}} -B`
--- a/pages/linux/efibootbgr.md
+++ b/pages/linux/efibootbgr.md
@@ -14,7 +14,7 @@
 
 `sudo efibootmgr -c -d /dev/sda1 -l \EFI\tools\Shell.efi -L "UEFI Shell"`
 
-- Change the current Bootorder (you can get the Bootnums from the first command):
+- Change the current boot order (you can get the Bootnums from the first command):
 
 `sudo efibootmgr -o {{0002,0008,0001,0005}}`
 
--- a/pages/linux/efibootbgr.md
+++ b/pages/linux/efibootbgr.md
@@ -4,7 +4,7 @@
 
 - List the current settings:
 
-`efibootmgr `(`-v` for Filepaths)
+`efibootmgr -v for Filepaths`
 
 - Create new Bootoption:
 
--- a/pages/linux/efibootbgr.md
+++ b/pages/linux/efibootbgr.md
@@ -1,0 +1,23 @@
+# efibootmgr
+
+> Manipulate the UEFI Boot Manager (the Bootoptions).
+
+- List the current settings:
+
+`efibootmgr `(`-v` for Filepaths)
+
+- Create new Bootoption:
+
+`sudo efibootmgr -c -d {{esp_blockdevice}} -l {{\path\to\file.efi}} -L "{{Label}}"`
+
+- > Specific Example: Adding UEFI Shell v2 as a Boot Options:
+
+`sudo efibootmgr -c -d /dev/sda1 -l \EFI\tools\Shell.efi -L "UEFI Shell"`
+
+- Change the current Bootorder (you can get the Bootnums from the first command):
+
+`sudo efibootmgr -o {{0002,0008,0001,0005}}`
+
+- Delete Bootoption:
+
+`sudo efibootmgr -b {{bootnum}} -B`
--- a/pages/linux/efibootbgr.md
+++ b/pages/linux/efibootbgr.md
@@ -6,7 +6,7 @@
 
 `efibootmgr -v for Filepaths`
 
-- Create new Bootoption:
+- Create a new boot option:
 
 `sudo efibootmgr -c -d {{esp_blockdevice}} -l {{\path\to\file.efi}} -L "{{Label}}"`
 
--- a/pages/linux/efibootbgr.md
+++ b/pages/linux/efibootbgr.md
@@ -10,7 +10,7 @@
 
 `sudo efibootmgr -c -d {{esp_blockdevice}} -l {{\path\to\file.efi}} -L "{{Label}}"`
 
-- > Specific Example: Adding UEFI Shell v2 as a Boot Options:
+- Add UEFI Shell v2 as a boot option:
 
 `sudo efibootmgr -c -d /dev/sda1 -l \EFI\tools\Shell.efi -L "UEFI Shell"`
 
--- a/pages/linux/efibootbgr.md
+++ b/pages/linux/efibootbgr.md
@@ -1,23 +1,25 @@
 # efibootmgr
 
 > Manipulate the UEFI Boot Manager (the Bootoptions).
+> 
+> More Information: https://linux.die.net/man/8/efibootmgr
 
-- List the current settings:
+- List the current settings / bootnums:
 
-`efibootmgr -v for Filepaths`
+`efibootmgr`
 
-- Create a new boot option:
-
-`sudo efibootmgr -c -d {{esp_blockdevice}} -l {{\path\to\file.efi}} -L "{{Label}}"`
+- List the Filepaths
+  
+`efibootmgr -v`
 
 - Add UEFI Shell v2 as a boot option:
 
-`sudo efibootmgr -c -d /dev/sda1 -l \EFI\tools\Shell.efi -L "UEFI Shell"`
+`sudo efibootmgr -c -d {{/dev/sda1}} -l {{\EFI\tools\Shell.efi}} -L "{{UEFI Shell}}"`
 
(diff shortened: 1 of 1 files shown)


pages/linux/efibootbgr.md
```diff
@@ -0,0 +1,23 @@
...
+
+- List the current settings:
+
+`efibootmgr `(`-v` for Filepaths)
+
+- Create new Bootoption:
+
+`sudo efibootmgr -c -d {{esp_blockdevice}} -l {{\path\to\file.efi}} -L "{{Label}}"`
+
+- > Specific Example: Adding UEFI Shell v2 as a Boot Options:
```
ivanhercaz:
```suggestion
- Add UEFI Shell v2 as a boot option:
```
It isn't necessary to specify "Specific example" nor ">". In addition, "Boot Options" should be in singular and in lowercase.

pages/linux/efibootbgr.md
```diff
@@ -0,0 +1,23 @@
+# efibootmgr
+
+> Manipulate the UEFI Boot Manager (the Bootoptions).
+
+- List the current settings:
+
+`efibootmgr `(`-v` for Filepaths)
+
+- Create new Bootoption:
```
ivanhercaz:
```suggestion
- Create new boot option:
```

"Boot option" should be in lowercase.

pages/linux/efibootbgr.md
```diff
@@ -0,0 +1,23 @@
+# efibootmgr
+
+> Manipulate the UEFI Boot Manager (the Bootoptions).
+
+- List the current settings:
+
+`efibootmgr `(`-v` for Filepaths)
```
ivanhercaz: The above command doesn't work for me, it needs another `, and even if I add it I the next traceback (syntax error) before the settings of my boot manager:

```
bash: command substitution: línea 2: error sintáctico: no se esperaba el final del fichero
bash: command substitution: línea 1: error sintáctico cerca del elemento inesperado `)'
bash: command substitution: línea 1: ` for Filepaths)'
```

The suggestion I do below works nice for me. It might be useful to have the feedback of another reviewers to test this.
```suggestion
`efibootmgr -v for Filepaths`
```

pages/linux/efibootbgr.md
```diff
@@ -0,0 +1,23 @@
...
+
+- > Specific Example: Adding UEFI Shell v2 as a Boot Options:
+
+`sudo efibootmgr -c -d /dev/sda1 -l \EFI\tools\Shell.efi -L "UEFI Shell"`
+
+- Change the current Bootorder (you can get the Bootnums from the first command):
+
+`sudo efibootmgr -o {{0002,0008,0001,0005}}`
+
+- Delete Bootoption:
```
ivanhercaz:
```suggestion
- Delete boot option:
```

"Bootoption" should be in lowercase.

pages/linux/efibootbgr.md
ivanhercaz: This one [has been reported](https://github.com/tldr-pages/tldr/pull/3788#issuecomment-577631625) tldr-bot.

ivanhercaz:
I forgot it... Thank you for your contribution, @sp1ritCS!

sp1ritCS:
> I forgot it... Thank you for your contribution, @sp1ritCS!

Yeah, no problem. I have no issues with your changes. One other thing im unsure about is that you have to know what your ESP Blockdevice is and where it is mounted. Otherwise somebody might assume that its the root partition and create a /EFI there.

pages/linux/efibootbgr.md
```diff
@@ -0,0 +1,23 @@
...
+
+> Manipulate the UEFI Boot Manager (the Bootoptions).
+
+- List the current settings:
+
+`efibootmgr `(`-v` for Filepaths)
+
+- Create new Bootoption:
+
+`sudo efibootmgr -c -d {{esp_blockdevice}} -l {{\path\to\file.efi}} -L "{{Label}}"`
```
sbrl: I'm a bit confused - doesn't Linux use forward slashes `/` in paths - perhaps this will read better?
```suggestion
`sudo efibootmgr -c -d {{esp_blockdevice}} -l {{/path/to/file.efi}} -L "{{Label}}"`
```

pages/linux/efibootbgr.md
sbrl: Marking as unresolved, since I can't see any changes to the file itself.

pages/linux/efibootbgr.md
sp1ritCS: The ESP Blockdevice has a FAT (dos) Filesystem, and FAT Filesystems are ("officially") using \.
I have no clue if you could use / as well, because I have never done it. And `efibootmgr -v` also gives out file paths like  `Boot0001* GRUB	HD(1,GPT,[UUID],0x800,0x20000)/File(\EFI\GRUB\grubx64.efi)`.

sbrl: Ah, I see. So this isn't a regular file path - it's a special ESP / boot filepath. That makes more sense then :-)

pages/linux/efibootbgr.md
```diff
@@ -0,0 +1,23 @@
+# efibootmgr
+
+> Manipulate the UEFI Boot Manager (the Bootoptions).
+
+- List the current settings:
+
+`efibootmgr -v for Filepaths`
```
agnivade: Please use just the simple 'efibootmgr' as the first example. Make `-v` to be the last example.

pages/linux/efibootbgr.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Create a new boot option:
+
+`sudo efibootmgr -c -d {{esp_blockdevice}} -l {{\path\to\file.efi}} -L "{{Label}}"`
```
agnivade: The convention for path tokens we follow is "path/to/file.efi". And please give a concrete example for block device - like "{{/dev/sda1}}" and same for label.

pages/linux/efibootbgr.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Add UEFI Shell v2 as a boot option:
```
agnivade: This is the same as above. We can just make the concrete examples part of that itself. Let's remove this one.

pages/linux/efibootbgr.md
```diff
@@ -0,0 +1,23 @@
...
+
+`sudo efibootmgr -c -d /dev/sda1 -l \EFI\tools\Shell.efi -L "UEFI Shell"`
+
+- Change the current boot order (you can get the Bootnums from the first command):
```
agnivade: Let's mention in the first example that it shows the boot numbers. Then we don't need to mention that here.

pages/linux/efibootbgr.md
```diff
@@ -0,0 +1,23 @@
...
+
+`sudo efibootmgr -o {{0002,0008,0001,0005}}`
+
+- Delete a boot option:
+
+`sudo efibootmgr -b {{bootnum}} -B`
```
agnivade: Let's use the more descriptive "--delete-bootnum" instead of "-B" and use a concrete number like {{002}} for bootnum.

pages/linux/efibootbgr.md
```diff
@@ -0,0 +1,23 @@
+# efibootmgr
+
+> Manipulate the UEFI Boot Manager (the Bootoptions).
+
```
agnivade: Also, let's add a "More Information" link to the man page.

pages/linux/efibootbgr.md
sp1ritCS: The backslashes are there, because the ESP is a FAT Filesystem, and "officially" FAT uses \. `efibootmgr -v` is displaying the Paths using \, the Manpage is using \. And UEFI Shell also displays \.
We have had the same thing somewhere in this PR already.

agnivade: The `/` convention is something we decided to use irrespective of what platform the page is for.

sbrl: I think I remember pointing this one out before @agnivade before - in this case it's not a regular file path. It's an EFI file path, which _must_ use backward slashes - otherwise it doesn't work IIRC.

So in this case I think we might be able to make an exception.

pages/linux/efibootbgr.md
```diff
@@ -0,0 +1,25 @@
+# efibootmgr
+
+> Manipulate the UEFI Boot Manager (the Bootoptions).
+>
```
sbrl: Looks like this blank line is throwing off the linter. Perhaps we can remove it?

pages/linux/efibootbgr.md
zdroid: Can you explain this more clearly? I don't quite understand you.

sbrl: I think what @agnivade means here is that this example is identical to the previous one in functionality, but unlike the example that appears directly before this one - this example contains concrete values instead of using the `{{token_syntax}}`. Hence, we can instead edit the previous example to use concrete values inside the `{{token_syntax}}` - and remove this example.

agnivade: That's exactly it.

sbrl: Not sure what's happened here, but this example doesn't look identical to the one before it anymore to me. I'mconfused?

/cc @agnivade

agnivade: This commit changed it https://github.com/tldr-pages/tldr/pull/3788/commits/f27372da2f82749602aedd8fed59cb563c80544a

pages/linux/efibootbgr.md
```diff
@@ -0,0 +1,24 @@
+# efibootmgr
+
+> Manipulate the UEFI Boot Manager (the Bootoptions).
+> More information: https://linux.die.net/man/8/efibootmgr.
+
+- List the current settings / bootnums:
+
+`efibootmgr`
+
+- List the Filepaths:
```
agnivade, line 10: Should Filepaths be capitalized ? I can make the change myself, but just wanted to check.

sp1ritCS: As a German, I capitalize a lot of words, that the english language doesn't capitalize. 🙃

Thinking about it, I shouldn't be capitalized

ivanhercaz: I think the suggestion made by @agnivade is right and due it was a little change and @sp1ritCS seems to be agree, I performed the change.

PR #3790, opened 2020-01-23
youtube-dl: add download the best audio+video

mirorauhala:
With this command the user can download the best audio and video separately. After it's done downloading, it will spawn ffmpeg to merge the two files together. Handy when downloading stuff from youtube as they host video and audio separate.

### DIFF
--- a/pages/common/youtube-dl.md
+++ b/pages/common/youtube-dl.md
@@ -21,7 +21,7 @@
 
 - Download the best quality audio and video and merge them:
 
-`youtube-dl -f {{bestvideo+bestaudio}} {{url}}`
+`youtube-dl -f bestvideo+bestaudio {{url}}`
 
 - Download video(s) as MP4 files with custom filenames:
 
--- a/pages/common/youtube-dl.md
+++ b/pages/common/youtube-dl.md
@@ -19,6 +19,10 @@
 
 `youtube-dl -x --audio-format {{mp3}} {{url}}`
 
+- Download the best quality audio and video and merge them:
+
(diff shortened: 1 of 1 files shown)


pages/common/youtube-dl.md
```diff
@@ -19,6 +19,10 @@
 
 `youtube-dl -x --audio-format {{mp3}} {{url}}`
 
+- Download the best quality audio and video and merge them:
+
+`youtube-dl -f {{bestvideo+bestaudio}} {{url}}`
```
zdroid: I'm really not sure how is this supposed to work. Have you meant to put something else in place of `{{bestvideo+bestaudio}}`?

pxgamer: This works because `bestvideo` and `bestaudio` are the actual values. 👍 ([Docs](https://github.com/ytdl-org/youtube-dl#format-selection))

zdroid: Shouldn't the braces be removed then?

ivanhercaz: I am bit confused. I think the braces should be removed, as @zdroid commented.

mirorauhala: I removed the braces.

sbrl: I'm not sure that they should, @zdroid, @ivanhercaz. `bestvideo+bestaudio` can be replaced with other values.

zdroid: They can, but the example was directed at this. If there are other available values, then `{{value1|value2|value3}}` syntax should have been used instead.

PR #3795, opened 2020-01-24
tldr: fix typo. 

Mgallimore88:
Adds missing hyphon from tldr tldr example page: 
tldr {{git checkout}} should read 
tldr {{git-checkout}}

### DIFF
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -13,4 +13,4 @@
 
 - Get help for a git subcommand:
 
-`tldr {{git checkout}}`
+`tldr {{git-checkout}}`


pages/common/tldr.md
```diff
@@ -13,4 +13,4 @@
 
 - Get help for a git subcommand:
 
-`tldr {{git checkout}}`
+`tldr {{git-checkout}}`
```
agnivade: Actually, it also works without the hyphen. But both are fine.

Mgallimore88: Not on my setup. I'm using Ubuntu 19.10 and bash. Typing 
`$tldr git checkout`
I get the error
`$Invalid` argument `checkout'
whereas
`$tldr git-checkout`
works. 
If the hyphon and no hyphon both work then lets keep the hyphon.

agnivade: That is probably because of the particular client you are using. If you use the node client, it works fine.

zdroid: Hyphen version is universal, space isn't, so it's better to stay this way anyway.

sbrl: @Mgallimore88: It should work in it's original form. Which client are you using? The [client spec](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md) does specify this as a requirement.

Perhaps we shoudl revert this PR then? I didn't notice this at first @agnivade

agnivade: Oh it's in the spec actually. But is it illegal to pass "git-checkout" ? If not, I'll just let it be. Not a huge deal either way.

sbrl: I don't think it's illegal. I believe the spec allows it both ways.

PR #3800, opened 2020-01-25
openrc, rc-status, rc-service, rc-update: add pages

sbrl:
I noticed that although we have pretty extensive pages for systemd and sysvinit, we don't yet have any have any pages for OpenRC.

This PR adds pages for the full suite of public-facing OpenRC commands.

### DIFF
--- a/pages/linux/rc-status.md
+++ b/pages/linux/rc-status.md
@@ -1,0 +1,29 @@
+# rc-status
+
+> Show status info about runlevels
+> Part of OpenRC - see also `openrc`.
+
+- Show a summary of services and their status:
+
+`rc-status`
+
+- Include services in all runlevels in the summary:
+
+`rc-status --all`
+
+- List services that have crashed:
+
+`rc-status --crashed`
+
+- List manually started services:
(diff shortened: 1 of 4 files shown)


pages/linux/rc-status.md
```diff
@@ -0,0 +1,32 @@
+# rc-status
+
+> Show status info about runlevels.
+> Part of OpenRC - see also `openrc`.
```
zdroid:
```suggestion
> See also `openrc`.
```

pages/linux/rc-update.md
```diff
@@ -0,0 +1,20 @@
+# rc-update
+
+> Add and remove OpenRC services to and from runlevels.
+> See also: `openrc`.
```
zdroid:
```suggestion
> See also `openrc`.
```

pages/linux/rc-status.md
sbrl: Hrm. I thought explicitly indicating that it's part of OpenRC was a good idea.

PR #3802, opened 2020-01-25
rsync: add -rauL option

laura-dietz:
This probably catches 99% of rsync usages and it its easy to remember:  `rsync -Laura`

### DIFF
--- a/pages/common/rsync.md
+++ b/pages/common/rsync.md
@@ -11,7 +11,7 @@
 
 `rsync {{remote_host}}:{{path/to/remote_file}} {{path/to/local_directory}}`
 
-- Transfer options for most common usage (handles soft-links, directories, access rights, and previously transferred files):
+- Transfer any file or directory, in archive (to preserve attributes), resolving contained softlinks, and ignoring already transferred files (unless newer):
 
 `rsync -Laura {{remote_host}}:{{path/to/remote_file}} {{path/to/local_directory}}`
 
--- a/pages/common/rsync.md
+++ b/pages/common/rsync.md
@@ -11,6 +11,10 @@
 
 `rsync {{remote_host}}:{{path/to/remote_file}} {{path/to/local_directory}}`
 
+- Worry-free file transfer (handles soft-links, directories, access rights, and previously transferred files):
+
+`rsync -Laura {{from}} {{to}}`
+
 - Transfer file in archive (to preserve attributes) and compressed (zipped) mode with verbose and human-readable progress:
 
 `rsync -azvhP {{path/to/local_file}} {{remote_host}}:{{path/to/remote_directory}}`
--- a/pages/common/rsync.md
+++ b/pages/common/rsync.md
@@ -11,7 +11,7 @@
 
 `rsync {{remote_host}}:{{path/to/remote_file}} {{path/to/local_directory}}`
 
-- Transfer any file or directory, in archive (to preserve attributes), resolving contained softlinks, and ignoring already transferred files (unless newer):
+- Transfer any file or directory \[r\], in archive \[a\] to preserve attributes, resolving contained softlinks \[L\], and ignoring already transferred files \[u\] (unless newer):
 
 `rsync -Laura {{remote_host}}:{{path/to/remote_file}} {{path/to/local_directory}}`
 
--- a/pages/common/rsync.md
+++ b/pages/common/rsync.md
@@ -13,7 +13,7 @@
 
 - Worry-free file transfer (handles soft-links, directories, access rights, and previously transferred files):
 
-`rsync -Laura {{from}} {{to}}`
+`rsync -Laura {{remote_host}}:{{path/to/remote_file}} {{path/to/local_directory}}`
 
 - Transfer file in archive (to preserve attributes) and compressed (zipped) mode with verbose and human-readable progress:
 
--- a/pages/common/rsync.md
+++ b/pages/common/rsync.md
@@ -11,11 +11,7 @@
 
 `rsync {{remote_host}}:{{path/to/remote_file}} {{path/to/local_directory}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/rsync.md
```diff
@@ -11,6 +11,10 @@
 
 `rsync {{remote_host}}:{{path/to/remote_file}} {{path/to/local_directory}}`
 
+- Worry-free file transfer (handles soft-links, directories, access rights, and previously transferred files):
+
+`rsync -Laura {{from}} {{to}}`
```
zdroid: `{{from}}` and `{{too}}` should be made to match other examples, maybe as `{{/from/file}}` and `{{/to/file}}`, but I'm not sure.

zdroid:
You do not need to make a new PR to fix an error. You can commit directly to the one with the failed build.

pages/common/rsync.md
laura-dietz: ok.

laura-dietz:
I think the `-ru` example is superseded by `-Laura` and can be dropped. 

```rsync -ru {{remote_host}}:{{path/to/remote_directory}} {{path/to/local_directory}}```

**Comment by agnivade** [comment]
Ok, let's do that then. Also, to maintain the tone of tldr pages, I would suggest to drop "Worry-free" and actually explain what "Laura" does, by matching each of the option to the action it performs.

**Comment by laura-dietz** [comment]
> Ok, let's do that then. Also, to maintain the tone of tldr pages,

done and done (sorry for the delay)

**Comment by agnivade** [comment]
Could you also 

> explain what "Laura" does, by matching each of the option to the action it performs.

Thanks.

**Comment by laura-dietz** [comment]
I could, but then it would become a multi-line comment that would break the linguistic style of the other examples.  Are you sure this is what you want?

**Comment by agnivade** [comment]
If that's the case, then I'm afraid we would have to drop this. Our main objective here is to explain each option to the user, and let the user build up complex commands by themselves. This is slightly different from cheat sheet like approaches where a user is just presented with a list of commands that they can simply cargo-cult.

But if it doesn't become too long, I'm okay with it. It's just that our first priority is keeping the explanations simple.

**Comment by laura-dietz** [comment]
Maybe this is what you are looking for. I was following the style of -azvhP. 

I wanted add this to tldr because it is my single most useful rsync combination (I use it a lot). I wish it would be rsync's default, although I understand why its not. 

Several of my CS students struggle with rsync because it never does what they expect. As a consequence they are forgetting to copy directories and I get useless softlinks that point to nirvana. Rsync's man page is overwhelming and the "worry-free options" are scattered across. While many stack-overflow posts mention to -a, most of them forget about softlinks, leading to problems that are much harder to detect.

Thanks for your effort and patience.

File: pages/common/rsync.md [hunk 2]
```diff
@@ -11,6 +11,10 @@
 
 `rsync {{remote_host}}:{{path/to/remote_file}} {{path/to/local_directory}}`
 
+- Transfer any file or directory, in archive (to preserve attributes), resolving contained softlinks, and ignoring already transferred files (unless newer):
```
agnivade: I'll explain why the explanation for `azvhP` works and this can be improved.

From "in archive (to preserve attributes) and compressed (zipped) mode with verbose and human-readable progress", I can gather a is for archive, z is for zipped, v is for verbose, h is for human-readable and p is for progress.

Here I can make out archive, but nothing else.

There is another convention that we follow in tough situations like this. It is to wrap the letters with third brackets. For example: [a]rchive, [r]ecursively.

Is there a possibility we can change the description to use this convention and spell out each letter in "Laura" in the description ?

laura-dietz: done.  Let me know if you want to squash my commits. 

Also let me know if you think this pun is too silly to tldr.

agnivade: The letters should be part of the words themselves. I can do that myself. But what is the second `a` for ? I just see one for archive.

pages/common/rsync.md
```diff
@@ -11,7 +11,7 @@
 
 `rsync {{remote_host}}:{{path/to/remote_file}} {{path/to/local_directory}}`
 
-- Transfer any file or directory, in archive (to preserve attributes), resolving contained softlinks, and ignoring already transferred files (unless newer):
+- Transfer any file or directory \[r\], in archive \[a\] to preserve attributes, resolving contained softlinks \[L\], and ignoring already transferred files \[u\] (unless newer):
```
ivanhercaz: I have the same question than @agnivade: has any function the second `a`? I mean, if `a` is "to preserve attributes", this command could just `-Laur` instead of `-Laura`, unless the last `a` means something.

laura-dietz:
The second a completes the pun.

As I said, if you find the pun too silly for tldr, then just close it.

ivanhercaz:
> As I said, if you find the pun too silly for tldr, then just close it.

I think we should add this interesting option, but I really have doubts about if it would be necessary to add with the pun (`-Laura`) or without it (`-Laur`).

pages/common/rsync.md
sbrl: I think it's supposed to be a menomic, to make it easy to remember. Why the backslashes here?

ivanhercaz:
I decide to merge it because there is enough approvals (four!) and the change of the name fits with the guidelines and even with the use of a mnemonic.

laura-dietz:
I hope you treat your future contributors with more respect.

agnivade:
@laura-dietz - My apologies. I sincerely was trying to help because you had already put so much time and effort into trying to get this option in, and really hated asking you to make yet another round of changes. So I went ahead and made the changes myself. This is something we usually do at tldr, and is part of our maintainer guidelines.

Only now, did I realize that `-Laura` is a pun with your name which is actually Laura. So sorry, I hadn't noticed this. I just thought that `-Laura` is easy to remember and that's why you put it. So to keep all the letters without repetition, I changed it to `rauL` which is still easy to remember.

I hope that clarifies my intent. And sorry again if I have caused any disrespect.

PR #3815, opened 2020-01-28
firewall-cmd: add zone change and arbitrary ports

lzap:
I do this a lot with my test servers - put an interface into trusted
zone. The syntax is a bit weird, therefore I always tend to dig in the
man page.

The patch also adds arbitrary port naming because the syntax is also a
bit weird and hard to remember. Cheers!

### DIFF
--- a/pages/linux/firewall-cmd.md
+++ b/pages/linux/firewall-cmd.md
@@ -10,6 +10,10 @@
 
 `firewall-cmd --list-all`
 
+- Permanently move the interface into the block zone, effectively blocking all communication:
+
+`firewall-cmd --permanent --zone={{block}} --change-interface={{enp1s0}}`
+
 - Permanently open the port for a service in the specified zone (like port `443` when in the `public` zone):
 
 `firewall-cmd --permanent --zone={{public}} --add-service={{https}}`
@@ -17,6 +21,10 @@
 - Permanently close the port for a service in the specified zone (like port `80` when in the `public` zone):
 
 `firewall-cmd --permanent --zone={{public}} --remove-service={{http}}`
+
+- Permanently open two arbitrary ports in the specified zone:
+
+`firewall-cmd --permanent --zone={{public}} --add-port={{25565/tcp}} --add-port={{19132/udp}}`
 
 - Reload firewalld to force rule changes to take effect:
 


pages/linux/firewall-cmd.md
```diff
@@ -18,6 +22,10 @@
 
 `firewall-cmd --permanent --zone={{public}} --remove-service={{http}}`
 
+- Permanently open two arbitrary ports in the specified zone (Minecraft service):
```
zdroid: What do you mean by `Minecraft service`?

pages/linux/firewall-cmd.md
```diff
@@ -10,6 +10,10 @@
 
 `firewall-cmd --list-all`
 
+- Permanently move the interface into the trusted zone effectively allowing all communication:
```
zdroid:
```suggestion
- Permanently move the interface into the trusted zone, effectively allowing all communication:
```
Maybe this could be made even more clear.

pages/linux/firewall-cmd.md
lzap: I wanted to provide an example of a service that needs both UDP and TCP protocol, thought it's popular and fun. Let's find a different one then.

pages/linux/firewall-cmd.md
lzap: Sure.

lzap:
Well I could also drop the "minecraft service" remark completely, it's irrelevant. Ports as examples are there, I guess that's enough. I've rebased the changes. Going to sign the CLA now.

pages/linux/firewall-cmd.md
```diff
@@ -10,6 +10,10 @@
 
 `firewall-cmd --list-all`
 
+- Permanently move the interface into the trusted zone, effectively allowing all communication:
```
sbrl: Ooh errr. This example sounds dangerous - are you sure it's a good fit for the tldr-page?

zdroid: What does it do exactly?

sbrl: It looks like it disables a the firewall completely for a given network interface.

zdroid: I agree with you, if that's the case. In any case, you should request changes here so the PR can be put on hold.

lzap: Good point, I can change to a different zone. Perhaps the default zone `block` is a good candidate?

lzap:
Right, well I haven't remove it since I value the example being present in there as it is one of the most important commands of this tool (you usually want to move your interface to external/internal for NAT or trusted or public).

Therefore I changed it to be safe by default, `block` is a default zone which actually does the opposite than `trusted`.

If you insist I can remove, but this looks like a good proposal to me.

lzap:
> since most servers only have a single network interface

Lucky you! :-)

PR #3816, opened 2020-01-28
nmcli: add more server examples

lzap:

I know this expands the page up to maximum of 8 examples, but the
problem here is that NetworkManager is used both on clients (laptops,
PCs) mainly for WiFi connection management but also on servers to manage
ethernet, bond or VLAN connections.

It's one of the quite complex commands with lots of options, I
handpicked three new examples I struggle quite a lot: static IP
configurations for both IPv4 and IPv6. These options are inconsistent
(ip4 vs ipv4) and I always need to dig in the man page.

To hit the 8 examples limit, I propose to delete the last exapmle which
is IMHO rarely used (`ip` is preferred) and it renders the very similar
output as the very first example (the only difference is that it lists
unmanaged devices).

Cheers!

### DIFF
--- a/pages/linux/nmcli-connection.md
+++ b/pages/linux/nmcli-connection.md
@@ -14,14 +14,14 @@
 
 `nmcli connection down uuid {{uuid}}`
 
-- Create a static IPv4-only connection:
+- Create an auto-configured dual stack connection:
 
-`nmcli connection add ifname {{interface_name}} type {{ethernet}} ip4 {{10.0.0.7/8}} gw4 {{10.0.0.1}} ipv4.dns {{10.0.0.1}} ipv6.method {{ignore}}`
+`nmcli connection add ifname {{interface_name}} type {{ethernet}} ipv4.method {{auto}} ipv6.method {{auto}}`
 
 - Create a static IPv6-only connection:
 
 `nmcli connection add ifname {{interface_name}} type {{ethernet}} ip6 {{2001:db8::2/64}} gw6 {{2001:db8::1}} ipv6.dns {{2001:db8::1}} ipv4.method {{ignore}}`
 
-- Create an auto-configured dual stack connection:
+- Create a static IPv4-only connection:
 
-`nmcli connection add ifname {{interface_name}} type {{ethernet}} ipv4.method {{auto}} ipv6.method {{auto}}`
+`nmcli connection add ifname {{interface_name}} type {{ethernet}} ip4 {{10.0.0.7/8}} gw4 {{10.0.0.1}} ipv4.dns {{10.0.0.1}} ipv6.method {{ignore}}`
--- a/pages/linux/nmcli-connection.md
+++ b/pages/linux/nmcli-connection.md
@@ -1,0 +1,27 @@
+# nmcli connection
+
+> Connection management with NetworkManager.
+
+- List all NetworkManager connections (shows name, uuid, type and device):
+
+`nmcli connection`
+
+- Activate a connection by specifying an uuid:
+
+`nmcli connection up uuid {{uuid}}`
+
+- Deactivate a connection:
+
+`nmcli connection down uuid {{uuid}}`
+
+- Create a static IPv4-only connection:
+
+`nmcli connection add ifname {{interface_name}} type {{ethernet}} ip4 {{10.0.0.7/8}} gw4 {{10.0.0.1}} ipv4.dns {{10.0.0.1}} ipv6.method {{ignore}}`
+
+- Create a static IPv6-only connection:
+
+`nmcli connection add ifname {{interface_name}} type {{ethernet}} ip6 {{2001:db8::2/64}} gw6 {{2001:db8::1}} ipv6.dns {{2001:db8::1}} ipv4.method {{ignore}}`
+
+- Create an auto-configured dual stack connection:
+
+`nmcli connection add ifname {{interface_name}} type {{ethernet}} ipv4.method {{auto}} ipv6.method {{auto}}`
--- a/pages/linux/nmcli-device.md
+++ b/pages/linux/nmcli-device.md
@@ -1,0 +1,15 @@
+# nmcli device
+
+> Hardware device management with NetworkManager.
+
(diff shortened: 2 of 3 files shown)


zdroid:
Your examples are way more informative, I support this totally. :smile:
Also, please sign the CLA.

pages/linux/nmcli.md
```diff
@@ -22,6 +22,14 @@
 
 `nmcli connection down uuid {{uuid}}`
 
-- Print statuses of network interfaces:
+- Create a static IPv4-only connection:
 
-`nmcli device status`
+`nmcli c add con-name enp1s0 ifname enp1s0 type ethernet ip4 10.0.0.7/8 gw4 10.0.0.1 ipv4.dns 10.0.0.1 ipv6.method ignore`
```
sbrl: I'm pretty sure there's a long form of the `c` here. Even though it makes the line even longer, perhaps we could use it here please?
```suggestion
`nmcli connection add con-name enp1s0 ifname enp1s0 type ethernet ip4 10.0.0.7/8 gw4 10.0.0.1 ipv4.dns 10.0.0.1 ipv6.method ignore`
```

lzap: I wanted to avoid putting more options as it's long enough. Ignore simply does not configure IPv6.

lzap:
Apologies for the delay. I've incorporated the change. On top of that:

* Removed `con-name` from the examples as it's optional flag and NM will figure out a reasonable connection name.
* Reordered options so they are in the same order.
* Broken nmcli into subcommands
* Put the `device` list command back since there is enough room now.

pages/linux/nmcli-connection.md
```diff
@@ -0,0 +1,27 @@
+# nmcli
```
agnivade: nmcli connection

pages/linux/nmcli-device.md
```diff
@@ -0,0 +1,15 @@
+# nmcli
```
agnivade: nmcli device

pages/linux/nmcli-connection.md
lzap: Good catch, rebased.

pages/linux/nmcli-device.md
lzap: Good catch, rebased.

pages/linux/nmcli.md
sbrl: Ah, I see. Considering IPv6 is the future version of the Internet Protocol, I'm reluctant to go ahead with the `ipv6.method {{ignore}}` bit here unless there's a convincing and common use-case.

This said, it looks like you've made some pretty extensive changes since I last reviewed, so I'll do a fresh review.

pages/linux/nmcli-connection.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Activate a connection by specifying an uuid:
+
+`nmcli connection up uuid {{uuid}}`
+
+- Deactivate a connection:
+
+`nmcli connection down uuid {{uuid}}`
+
+- Create a static IPv4-only connection:
```
sbrl: Perhaps we could alter the order of these 3 examples to be more similar to the order people might use them in. What about reversing it, like this?

 - Create an auto-configured.....
 - Static IPv6
 - Static IPv4

pages/linux/nmcli-device.md
```diff
@@ -0,0 +1,15 @@
+# nmcli device
+
+> Hardware device management with NetworkManager.
+
+- Print statuses of network interfaces:
```
sbrl: Perhaps this will read better?
```suggestion
- Print the statuses of all network interfaces:
```

pages/linux/nmcli-device.md
```diff
@@ -0,0 +1,15 @@
...
+
+`nmcli device status`
+
+- Print the available Wi-Fi access points:
+
+`nmcli device wifi`
+
+- Connect to the Wi-Fi network with a specified name and password:
+
+`nmcli device wifi connect {{name}} password {{password}}`
```
sbrl: Perhaps we could tweak this slightly to be more intuitive?
```suggestion
`nmcli device wifi connect {{ssid}} password {{password}}`
```

pages/linux/nmcli.md
```diff
@@ -2,26 +2,18 @@
...
+`nmcli --help`
 
-- Connect to the Wi-Fi network with a specified name and password:
+- Call help on a command:
 
-`nmcli device wifi connect {{name}} password {{password}}`
+`nmcli {{command}} --help`
 
-- Activate a connection by specifying an uuid:
+- Execute nmcli command:
```
sbrl: Perhaps this will read better?
```suggestion
- Execute an `nmcli` command:
```

pages/linux/nmcli-connection.md
lzap: I am coming from enterprise datacenters world, I can tell you that IPv6 is not a thing there yet. And at homes, that's disaster. But sure, won't hurt, shaked it a bit.

pages/linux/nmcli-device.md
lzap: Yeah.

pages/linux/nmcli-device.md
lzap: Done.

pages/linux/nmcli.md
lzap: Yup.

lzap:
> thanks so much for the edits here!

Bad on my side, I have about 200+ PRs to review on my plate these days. Slowly working it through :-)

I will add new commits, we tend to rebase as github now shows changes anyway even if you rebase. But sure, if that's your thing! Pushed.

pages/linux/nmcli-connection.md
sbrl: I see. With IPv6 being the future, it seems like a sensible idea to emphasise that in tldr-pages.

sbrl:
Merging, since @ivanhercaz's comment appears to be addressed.

PR #3821, opened 2020-01-30
Pages/f3

sbrl:
(Almost) the rest of the f3* commands! There's still `f3brew`, but I'm not sure what it's for - and I don't have a fake flash drive to test it on :P


zdroid:
It took me a while to realize there were two newlines at the end of the files. :confused:

sbrl:
Thanks, @zdroid! I must have missed that.

PR #3831, opened 2020-02-02
yabai: add page

BosEriko:

### DIFF
--- a/pages/osx/yabai.md
+++ b/pages/osx/yabai.md
@@ -1,6 +1,6 @@
 # yabai
 
-> A tiling window manager for macOS based on binary space partitioning
+> A tiling window manager for macOS based on binary space partitioning.
 > More information: <https://github.com/koekeishiya/yabai>.
 
 - Set the layout to bsp:
--- a/pages/osx/yabai.md
+++ b/pages/osx/yabai.md
@@ -1,0 +1,24 @@
+# yabai
+
+> A tiling window manager for macOS based on binary space partitioning
+> More information: <https://github.com/koekeishiya/yabai>.
+
+- Set the layout to bsp:
+
+`yabai -m config layout bsp`
+
+- Set the window gap:
+
+`yabai -m config window_gap 10`
+
+- Enable opacity:
+
+`yabai -m config window_opacity on`
+
+- Disable window shadow:
+
+`yabai -m config window_shadow off`
+
+- Enable status bar:
+
+`yabai -m config status_bar on`
--- a/pages/osx/yabai.md
+++ b/pages/osx/yabai.md
@@ -5,11 +5,11 @@
 
 - Set the layout to bsp:
 
-`yabai -m config layout bsp`
+`yabai -m config layout {{bsp}}`
 
-- Set the window gap:
+- Set the window gap to 10pt:
 
-`yabai -m config window_gap 10`
+`yabai -m config window_gap {{10}}`
 
 - Enable opacity:
 


pages/osx/yabai.md
```diff
@@ -0,0 +1,24 @@
+# yabai
+
+> A tiling window manager for macOS based on binary space partitioning.
+> More information: <https://github.com/koekeishiya/yabai>.
+
+- Set the layout to bsp:
+
+`yabai -m config layout bsp`
```
zdroid:
```suggestion
`yabai -m config layout {{bsp}}`
```

pages/osx/yabai.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Set the window gap:
+
+`yabai -m config window_gap 10`
```
zdroid:
```suggestion
`yabai -m config window_gap {{10}}`
```

pages/osx/yabai.md
```diff
@@ -0,0 +1,24 @@
+# yabai
+
+> A tiling window manager for macOS based on binary space partitioning.
+> More information: <https://github.com/koekeishiya/yabai>.
+
+- Set the layout to bsp:
+
+`yabai -m config layout bsp`
+
+- Set the window gap:
```
zdroid:
```suggestion
- Set the window gap to 10:
```

Maybe even `10 pixels` (?) if that's the correct unit of measure here.

BosEriko: I've updated it to `to 10pt` instead. Based on the looks it's definitely pixels but on Yabai's documentation they didn't mention it as pixels, they did mention it as 10pt, though.

I've also updated the other lines based on your comments. Thanks!

pages/osx/yabai.md
```diff
@@ -5,11 +5,11 @@
 
 - Set the layout to bsp:
 
-`yabai -m config layout bsp`
+`yabai -m config layout {{bsp}}`
 
-- Set the window gap:
+- Set the window gap to 10pt:
```
zdroid: Is it pt or px? Nevertheless, it should be in this form:
```suggestion
- Set the window gap to 10 pt:
```

mebeim: The unit is usually not spaced (see any program that takes values like `10px` `10G`, `10K`). I would leave it as is.

zdroid: @mebeim Right, I deleted that part.

BosEriko: It seems to be like it's pixels but on their documentation, they call it pt. That's why I went with pt. Shall I proceed with using px for the sake of unity between instructions here in tldr?

mebeim: I think just saying "pixels" is the best choice here. After all, you said they call it "pixels" in their documentation too, right?

mebeim: @BosEriko I can't find where exactly they use `pt`. Are you talking about [this](https://github.com/koekeishiya/yabai/blob/master/doc/yabai.asciidoc) documentation?

BosEriko: I found it here.
https://github.com/koekeishiya/yabai/wiki/Configuration#padding-and-gaps

mebeim: That's strange. I suppose they say `pt` to emphasize the fact that display scaling could result in different real pixel values. I don't see any mention of "pixels" in the docs though, I think we can leave it as is.

sbrl: No idea what this tool is, but a `pt` and a `px` are not actually the same thing. `px` is a pixel, and `pt` is a point - which may be a different size to a pixel IIRC.

zdroid: You are correct on using `pt` here. Thanks for contributing!

pages/osx/yabai.md
```diff
@@ -0,0 +1,24 @@
...
+
+`yabai -m config layout {{bsp}}`
+
+- Set the window gap to 10pt:
+
+`yabai -m config window_gap {{10}}`
+
+- Enable opacity:
+
+`yabai -m config window_opacity on`
```
agnivade: on and off should be tokens (`{{}}`)

zdroid: It could be, but only if `enable` was changed to `enable/disable`.

agnivade: Enable/disable would need it to be {{on|off}}. But if it's just enable, then I think it should just be {{on}}

PR #3835, opened 2020-02-04
mysqldump: harmonize the examples

brunob:
- always use `-p` for password
- use short hand syntax for restoring a database (like the one used for restoring all databases)

### DIFF
--- a/pages/common/mysqldump.md
+++ b/pages/common/mysqldump.md
@@ -3,18 +3,18 @@
 > Backups MySQL databases.
 > More information: <https://dev.mysql.com/doc/refman/en/mysqldump.html>.
 
-- Create a backup, user will be prompted for a password:
+- Create a backup (user will be prompted for a password):
 
-`mysqldump -u {{user}} -p {{database_name}} -r {{filename.sql}}`
+`mysqldump --user {{user}} --password {{database_name}} -r {{filename.sql}}`
 
-- Restore a backup, user will be prompted for a password:
+- Restore a backup (user will be prompted for a password):
 
-`mysql -u {{user}} -p {{database_name}} < {{filename.sql}}`
+`mysqldump --user {{user}} --password {{database_name}} < {{path/to/file.sql}}`
 
 - Backup all databases redirecting the output to a file (user will be prompted for a password):
 
-`mysqldump -u {{user}} -p --all-databases > {{filename.sql}}`
+`mysqldump --user {{user}} --password --all-databases > {{filename.sql}}`
 
 - Restore all databases from a backup (user will be prompted for a password):
 
-`mysql -u {{user}} -p < {{filename.sql}}`
+`mysqldump --user {{user}} --password < {{filename.sql}}`
--- a/pages/common/mysqldump.md
+++ b/pages/common/mysqldump.md
@@ -5,7 +5,7 @@
 
 - Create a backup (user will be prompted for a password):
 
-`mysqldump --user {{user}} --password {{database_name}} -r {{filename.sql}}`
+`mysqldump --user {{user}} --password {{database_name}} -r {{path/to/file.sql}}`
 
 - Restore a backup (user will be prompted for a password):
 
@@ -13,8 +13,8 @@
 
 - Backup all databases redirecting the output to a file (user will be prompted for a password):
 
-`mysqldump --user {{user}} --password --all-databases > {{filename.sql}}`
+`mysqldump --user {{user}} --password --all-databases > {{path/to/file.sql}}`
 
 - Restore all databases from a backup (user will be prompted for a password):
 
-`mysqldump --user {{user}} --password < {{filename.sql}}`
+`mysqldump --user {{user}} --password < {{path/to/file.sql}}`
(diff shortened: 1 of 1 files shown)


agnivade:
Our official stance is to always use an expanded option if the short hand option is not clear enough. In this case, -p seems to be clear enough. But I also see that `mysql` page uses `--password`. If we are doing this, we should also update the mysql page.

pages/common/mysqldump.md
```diff
@@ -5,11 +5,11 @@
 
 - Create a backup, user will be prompted for a password:
 
-`mysqldump -u {{user}} --password {{database_name}} -r {{filename.sql}}`
+`mysqldump -u {{user}} -p {{database_name}} -r {{filename.sql}}`
 
 - Restore a backup, user will be prompted for a password:
 
-`mysql -u {{user}} --password -e "source {{filename.sql}}" {{database_name}}`
+`mysql -u {{user}} -p {{database_name}} < {{filename.sql}}`
```
sbrl: I'd probably argue there that it's ambiguous as to whether the `{{database_name}}` applies to the `-p` or not if we shorten it. I'd suggest that we stick with the longer one in this instance.

mebeim: Agree with @sbrl here. Also, I just noticed that the command is `mysql` and not `mysqldump`, that looks like an error, care to fix that too please?

mebeim: I don't have a preference between `-p` and `--password`... but command in this example is `mysql` and not `mysqldump`, that looks like an error, care to fix that too please?

ivanhercaz: Ups! @mebeim is right! @sbrl, I think it might be solved editing your own suggestion.

sbrl: I can certainly generate a suggestion. My original comment doesn't have one though.
```suggestion
`mysqldump --user {{user}} --password {{database_name}} < {{path/to/file.sql}}`
```

brunob:
thx for the merge :)

pages/common/mysqldump.md
```diff
@@ -3,18 +3,18 @@
...
+- Create a backup (user will be prompted for a password):
 
-`mysqldump -u {{user}} -p {{database_name}} -r {{filename.sql}}`
+`mysqldump --user {{user}} --password {{database_name}} -r {{filename.sql}}`
 
-- Restore a backup, user will be prompted for a password:
+- Restore a backup (user will be prompted for a password):
 
-`mysql -u {{user}} -p {{database_name}} < {{filename.sql}}`
+`mysqldump --user {{user}} --password {{database_name}} < {{path/to/file.sql}}`
```
brunob: I have a doubt about this one, tested it now and it doesn't wok, are you sure that mysqldump (not mysql) allow this syntax to import a dump ?

mebeim: Oh, really? I assumed that was just a typo on the command name :thinking: so you're saying this is actually a `mysql` command?

brunob: yes this is a mysql command :p

mebeim: Does it actually restore a backup though? Because from what I can see on [the man page](https://linux.die.net/man/1/mysql) it seems like that just reads an SQL script to execute on the given database:

> You can execute SQL statements in a script file (batch file) like this:
>
>     shell> mysql db_name < script.sql > output.tab

brunob: yes, my original patch restore a dump cf `mysql -u {{user}} -p {{database_name}} < {{filename.sql}}`

mebeim: Thank you, I'll move that to the `mysql` page then since it's a different command.

brunob: I understand, but it seems a bit pitty since i was proposing to refine an item on mysqldump page, not to remove it :\

People that will look to mysqldump will loose this info that was available on this page.

mebeim: Well that's a different command though. If `mysqldump` can't restore backups I would suggest to change the description of this page to something like:

> Backup MySQL databases to be later restored through the `mysql` command.

What do you think?

brunob: Or keep an item about dump restoring without any command but instead a mention pointing to mysql page ?

mebeim: I'm afraid that'd go against the format of the pages. On the other hand, some pages have mention of other "companion" pages in their description.

schneiderl: On this discussion: We can use the `see also` that is used in multiple pages around tldr (eg. [1](pages/linux/sinfo.md) ). 
Something like `See also mysql for more information on restoring databases`

sbrl: Full example:

```markdown
> See also `mysql` for restoring databases.
```

mebeim: See #3912.

pages/common/mysqldump.md
```diff
@@ -3,18 +3,18 @@
 > Backups MySQL databases.
 > More information: <https://dev.mysql.com/doc/refman/en/mysqldump.html>.
 
-- Create a backup, user will be prompted for a password:
+- Create a backup (user will be prompted for a password):
```
waldyrious, line 6: ❤️

PR #3837, opened 2020-02-05
tree: ignore specific directories

pavelshtanko:
Ignoring of the specific directories is very important when you're going to print the tree of the repositories these days. 
I expected this information in `tldr` but didn't find it, so decided to add it. 

### DIFF
--- a/pages/linux/tree.md
+++ b/pages/linux/tree.md
@@ -2,15 +2,15 @@
 
 > Show the contents of the current directory as a tree.
 
-- Show files and directories up to 'num' levels of depth (where 1 means the current directory):
+- Print files and directories up to 'num' levels of depth (where 1 means the current directory):
 
 `tree -L {{num}}`
 
-- Show directories only:
+- Print directories only:
 
 `tree -d`
 
-- Show hidden files too:
+- Print hidden files too:
 
 `tree -a`
 
--- a/pages/linux/tree.md
+++ b/pages/linux/tree.md
@@ -29,3 +29,7 @@
 - Ignore entries that match a wildcard (glob) pattern:
 
 `tree -I {{*.txt}}`
+
+- Print the tree with ignoring of specific directories:
+
+`tree -I 'vendor|.git|node_modules'`
--- a/pages/linux/tree.md
+++ b/pages/linux/tree.md
@@ -32,4 +32,4 @@
 
 - Print the tree ignoring the given directories:
 
-`tree -I 'vendor|.git|node_modules'`
+`tree -I '{{directory_name1|directory_name2}}'`
--- a/pages/linux/tree.md
+++ b/pages/linux/tree.md
@@ -30,6 +30,6 @@
 
 `tree -I {{*.txt}}`
 
-- Print the tree with ignoring of specific directories:
+- Print the tree ignoring the given directories:
 
 `tree -I 'vendor|.git|node_modules'`
--- a/pages/osx/tree.md
+++ b/pages/osx/tree.md
@@ -30,6 +30,6 @@
 
 `tree -P {{directory_name}} --matchdirs --prune`
 
-- Print the tree with ignoring of specific directories:
+- Print the tree ignoring the given directories:
 
 `tree -I 'vendor|.git|node_modules'`
--- a/pages/osx/tree.md
+++ b/pages/osx/tree.md
@@ -29,3 +29,7 @@
 - Find directories within the tree hierarchy, pruning out directories that aren't ancestors of the wanted one:
 
 `tree -P {{directory_name}} --matchdirs --prune`
+
+- Print the tree with ignoring of specific directories:
+
+`tree -I 'vendor|.git|node_modules'`
--- a/pages/osx/tree.md
+++ b/pages/osx/tree.md
@@ -2,15 +2,15 @@
 
 > Show the contents of the current directory as a tree.
 
-- Show files and directories up to 'num' levels of depth (where 1 means the current directory):
+- Print files and directories up to 'num' levels of depth (where 1 means the current directory):
 
 `tree -L {{num}}`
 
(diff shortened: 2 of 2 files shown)


pages/linux/tree.md
```diff
@@ -29,3 +29,7 @@
 - Ignore entries that match a wildcard (glob) pattern:
 
 `tree -I {{*.txt}}`
+
+- Print the tree with ignoring of specific directories:
```
schneiderl: I like your addition but I feel like it could be worded differently.
Maybe:
```suggestion
- Print the tree ignoring the given directories:
```

pavelshtanko: nice! Will fix

pavelshtanko: Fixed

schneiderl:
Also, while we are here, maybe we could reword other examples on this pages. 
Examples like 

`Show files and directories up to 'num' levels of depth (where 1 means the current directory):`.
I think it makes more sense to use the wording you've used for the added example, something like:
`Print files and directories up to 'num' levels of depth (where 1 means the current directory):`

That should be applied to all examples on all pages. 
Do you guys agree with that?

pavelshtanko:
I also noticed this, will do

pavelshtanko:
@zdroid  Thanks for the feedback! 

Basically, when we are talking about terminals\consoles it's always a `Print`. 

From the beginning of time: 
C's `printf`, Fortran's `PRINT` till the golang's `log.Println\fmt.Print`, I think creators of these languages know the thing. 

So I would like to keep the `Print`. But I'm open to suggestions and not that rigid with this decision.

zdroid:
I know that aswell, but programming and documentation approaches are different in this case. Most of the commands of which I've read the docs (including their tldr pages) use `show` or `display` instead of `print`.

pavelshtanko:
@zdroid To be honest they are using all of the verbs:
image

zdroid:
It depends on the group of the commands you're viewing. Compare:
https://github.com/tldr-pages/tldr/search?q=print&unscoped_q=print
https://github.com/tldr-pages/tldr/search?q=print&unscoped_q=show

pavelshtanko:
Can you elaborate on it? What is the difference between groups? (btw these two link leads to the same search results)

Both `show` and `print` are equaly presented everywhere.

pavelshtanko:
Anyway, what others think?

ivanhercaz:
I want to give my perspective as Spanish speaker that translates this two words. In Spanish, `show` is "mostrar" and `print` is "imprimir", however I translate both according to the context and the most common translation I use for both words is "mostrar", not "imprimir", because I understand that in Spanish sometimes the word "imprimir" might be hard to relate with a screen context and not with a paper in which you print a content. This is one of the reasons why I hope there are more Spanish speakers in order to discuss this kind of issues.

What I want to express is that we should use the easier way to understand the text: what does the command do? Print/Show? For me, bot are synonyms in this context because both are displaying something in the output of the terminal.

pavelshtanko:
Indeed, both are the same in current context

zdroid:
@pavelshtanko Sorry, I put a wrong link for `show`. Check it now.

pavelshtanko:
So, what is your proposal for now? Change to `show`? What others think? 
@schneiderl @ivanhercaz

zdroid:
I just wanted to put this into another perspective. Both options are good, but the decision should definitely wait for more voices.

pavelshtanko:
I agree with you, but looks like we already have four (sorry, including me, which do not count) for `Print`. 

I'm personally ok with any of decisions, I like `show` not less than `print`. These two are my favorite words in the world)

ivanhercaz:
@pavelshtanko, I really have not problems with any of two. I usually work more in Spanish translations than in the creation of new pages, so my perspective is from the translator more than the creator of new pages. Both words seems nice to me because I treat them with the same meaning for the Spanish translations. In the case of English probably I will use `show` because it seems more natural to me, but I don't have problems reading `print`.

Yes, I have messed up too much :rofl: But what I want to say it is that I will be agree with any of the words in this context.

schneiderl:
I think I do prefer `print`, as it seems more natural in this context. 
Also, I've liked your sense of humor @pavelshtanko 😄

pavelshtanko:
So then... what should I do? Merge is still blocked :(

pages/osx/tree.md
```diff
@@ -22,10 +22,14 @@
...
 `tree -P '{{*.txt}}' --prune`
 
-- Find directories within the tree hierarchy, pruning out directories that aren't ancestors of the wanted one:
+- Print directories within the tree hierarchy, using the wildcard (glob) pattern, and pruning out directories that aren't ancestors of the wanted one:
 
 `tree -P {{directory_name}} --matchdirs --prune`
+
+- Print the tree ignoring the given directories:
+
+`tree -I 'vendor|.git|node_modules'`
```
zdroid: Is this a legal syntax? Or it's just one directory at the time? Anyway, it should be inside `{{}}`.

(Is `vendor|.git|node_modules` legal, or it should be `{{vendor|.git|node_modules}}` instead)?

schneiderl: It should be `{{vendor|.git|node_modules}}` since it is a parameter as any other.

zdroid: Maybe it is, but I don't know the specifics of this command. That's why I asked it in the first place.

mebeim: It's legal syntax, it means to ignore any name that is `vendor` OR `.git` OR `node_modules`.

pavelshtanko:
Nice finding! I didn't know about this rule. 
Anyway, switched to more generic parameters instead of specific. I think it makes more sense.

ivanhercaz:
Merged it! Thank you very much for your contribution.

PR #3841, opened 2020-02-05
multiple pages: change user_name to username

schneiderl:
This PR changes the token `user_name` to `username` in multiple files. 

**Why?**
To me `user_name` is just weird to read.
Also: Most pages use the token `username` so just to have it concise between pages.


schneiderl:
I should probably have one commit per file.
Sorry about that.

mebeim:
> I should probably have one commit per file.

Nah, it's no problem for simple mass changes like this one. One single commit is actually easier to keep track of.

PR #3842, opened 2020-02-05
pyenv-virtualenv: add activate/deactivate examples

kintoandar:

### DIFF
--- a/pages/common/pyenv-virtualenv.md
+++ b/pages/common/pyenv-virtualenv.md
@@ -5,8 +5,16 @@
 
 - Create a new Python 3.6.6 virtual environment:
 
-`pyenv virtualenv {{3.6.6}} {{my-virtual-env-3.6.6}}`
+`pyenv virtualenv {{3.6.6}} {{virtualenv_name}}`
 
 - List all existing virtual environments:
 
 `pyenv virtualenvs`
+
+- Activate a virtual environment:
+
+`pyenv activate {{virtualenv_name}}`
+
+- Deactivate the virtual environment:
+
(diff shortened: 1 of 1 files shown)


pages/common/pyenv-virtualenv.md
```diff
@@ -10,3 +10,11 @@
 - List all existing virtual environments:
 
 `pyenv virtualenvs`
+
+- Activate a virtual environment:
+
+`pyenv activate {{my-virtual-env-3.6.6}}`
```
zdroid: Maybe `{{my-virtual-env-3.6.6}}` should be changed here, as well as the one in the first example.

kintoandar: Agree, pinning a version in an example does not age well. May I suggest replacing `my-virtual-env-3.6.6` with `example`?

ivanhercaz: Or even `virtualenv_name`.

schneiderl: I agree with @ivanhercaz, `virtualenv_name` will do.

kintoandar: Changed 👍

ivanhercaz:
Thank you for the fix, @kintoandar. Now, after you applied the suggestion made by @zdroid, it looks good to me, but it is necessary he review it again to approve it.

Just a note, you have not to make a force push, because then, in cases like the edits of one page, we squash and merge it in just one commit with the title of the PR. So don't worry about it.

Again, thank you very much for your contribution.

PR #3846, opened 2020-02-10

schneiderl:
This is a follow up to #3352 .

If you have a better way to phrase those sentences, please propose changes so that we can discuss them.

### DIFF
--- a/pages/common/bmaptool.md
+++ b/pages/common/bmaptool.md
@@ -1,6 +1,6 @@
 # bmaptool
 
-> Create or Copy blockmaps intelligently (and therefore faster than `cp` or `dd`).
+> Create or Copy blockmaps intelligently (designed to be faster than `cp` or `dd`).
 > More information: <https://source.tizen.org/documentation/reference/bmaptool>.
 
 - Create a blockmap from image file:
--- a/pages/common/bmaptool.md
+++ b/pages/common/bmaptool.md
@@ -1,6 +1,6 @@
 # bmaptool
 
-> Create or Copy blockmaps intelligently (designed to be faster than `cp` or `dd`).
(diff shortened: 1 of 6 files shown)


pages/common/fnm.md
```diff
@@ -1,6 +1,6 @@
 # fnm
 
-> Fast Node.js version manager.
+> A Node.js version manager.
```
zdroid:
```suggestion
> Fast Node.js version manager.
```
This change should not be included. `nvm` is simply a node version manager, while `fvm` is exactly what it says.

schneiderl: but.....what? 😄  
What is different from this case to any other cases on this PR?

zdroid: If you changed this, then `nvm` and `fvm` would have essentially the same description, although nvm's [description](https://github.com/tldr-pages/tldr/blob/master/pages/common/nvm.md) is much longer.

sbrl: I think we should keep this one as-is, as the word `fast` is in `fvm`'s name.

pages/common/bmaptool.md
```diff
@@ -1,6 +1,6 @@
 # bmaptool
 
-> Create or Copy blockmaps intelligently (and therefore faster than `cp` or `dd`).
+> Create or Copy blockmaps intelligently (designed to be faster than `cp` or `dd`).
```
sbrl: Perhaps this will read better?
```suggestion
> Create or copy block maps intelligently (designed to be faster than `cp` or `dd`).
```

pages/common/fnm.md
schneiderl: alright, makes sense 😄

PR #3848, opened 2020-02-11
awk: simplify loop/fix missing filename

vivekkalyan:

### DIFF
--- a/pages/common/awk.md
+++ b/pages/common/awk.md
@@ -29,7 +29,7 @@
 
 - Print all values starting from the third column:
 
-`awk '{ s = ""; for (i=3; i <= NF; i++) s = s $i " "; print s }'`
+`awk '{for (i=3; i <= NF; i++) printf $i""FS; print""}'`
 
 - Print different values based on conditions:
 
--- a/pages/common/awk.md
+++ b/pages/common/awk.md
@@ -29,8 +29,8 @@
 
 - Print all values starting from the third column:
 
-`awk '{for (i=3; i <= NF; i++) printf $i""FS; print""}'`
+`awk '{for (i=3; i <= NF; i++) printf $i""FS; print""}' {{filename}}`
 
 - Print different values based on conditions:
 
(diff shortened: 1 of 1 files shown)


vivekkalyan:
two of the examples are missing the filename portion of the command, and for the loop manipulation assigning a variable adds unnecessary complication imo and is harder to compose with other options in awk.

pages/common/awk.md
```diff
@@ -29,8 +29,8 @@
 
 - Print all values starting from the third column:
 
-`awk '{ s = ""; for (i=3; i <= NF; i++) s = s $i " "; print s }'`
+`awk '{for (i=3; i <= NF; i++) printf $i""FS; print""}' {{filename}}`
```
sbrl: Perhaps this will read better?
```suggestion
`awk '{for (i=3; i <= NF; i++) printf $i""FS; print""}' {{path/to/filename}}`
```

pages/common/awk.md
```diff
@@ -29,8 +29,8 @@
...
 
 - Print different values based on conditions:
 
-`awk '{if ($1 == "foo") print "Exact match foo"; else if ($1 ~ "bar") print "Partial match bar"; else print "Baz"}'`
+`awk '{if ($1 == "foo") print "Exact match foo"; else if ($1 ~ "bar") print "Partial match bar"; else print "Baz"}' {{filename}}`
```
sbrl: Perhaps this will read better?
```suggestion
`awk '{if ($1 == "foo") print "Exact match foo"; else if ($1 ~ "bar") print "Partial match bar"; else print "Baz"}' {{path/to/filename}}`
```

zdroid: Wouldn't both of your suggestions make this example even longer?

vivekkalyan: i kept the naming consistent with the rest of the `awk.md` file thus using `{{filename}}`. If you suggest to change it to path to file, then maybe the entire file should be changed?

sbrl: @zdroid: Yeah, but then it would be more consistent with other pages.

@vivekkrish: Ah, I didn't notice that. Perhaps it might be worth leaving it as-is in this case for now since the examples are quite long.

PR #3849, opened 2020-02-12
tcptraceroute: add page

mt-empty:


pages/common/tcptraceroute.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Set initial and maximum TTL:
+
+`tcptraceroute {{host}} -f {{first_ttl} -m {{max_ttl}}`
+
+- Specify wait time and number of probes for each hop:
+
+`tcptraceroute {{host}} -w {{wait_time}} -q {{number_of_queries}}`
+
+- Specify interface with HTTP(port 80) as the destination port:
```
zdroid:
```suggestion
- Specify interface with HTTP (port 80) as the destination port:
```

mt-empty: Fixed typo.

pages/linux/tcptraceroute.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Set initial and maximum TTL:
+
+`tcptraceroute {{host}} -f {{first_ttl} -m {{max_ttl}}`
+
+- Specify wait time and number of probes for each hop:
+
+`tcptraceroute {{host}} -w {{wait_time}} -q {{number_of_queries}}`
+
+- Specify interface with HTTP (port 80) as the destination port:
```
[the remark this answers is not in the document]
mt-empty: You could always type port x instead, where x is a port number, or you can type http, https, ftp , ssh... . Though I'm not sure if a simple name can be used for every single port.

I think removing it might be simpler --> `tcptraceroute {{host}} -i {{interface}}`

What do you think?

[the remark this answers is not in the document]
sbrl: That would probably be a good idea. Since we've specified the port number in a different example, we probably don't need to include it in this one too.

einverne:
Just take a look at the man page, it shows `tcptraceroute` is equivalent to `traceroute -T`. It means a lot of explanations can be quite similar to the `traceroute`. Is it worth to mention in the description?

pages/linux/tcptraceroute.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Specify the destination port and packet length in bytes:
+
+`tcptraceroute {{host}} {{destination_port}} {{packet_length}}`
+
+- Specify the local source port and source address:
+
+`tcptraceroute {{host}} -p {{source_port}} -s {{source_address}}`
+
+- Set the initial and maximum TTL:
```
agnivade: Let's specify the unit also.

EDIT: Here the description mentions initial, but the flag mentions "first". We should be consistent and use the same thing everywhere. I am leaning towards using first in the description.

sbrl: Oops! I didn't catch this.

mt-empty: I was following the man pages. If that's not preferred, then I will change line 24 as well: 

- Specify the wait time and number of ~probes for each~ queries per hop:

agnivade: Yeah, we usually want to simplify the language and make it consistent with our overall theme and tone. If we just copy from the man page, then there's no point. :)

Your new change sounds good as well.

PR #3852, opened 2020-02-12
ctags: add page

thezeroalpha:

### DIFF
--- a/pages/common/ctags.md
+++ b/pages/common/ctags.md
@@ -1,7 +1,7 @@
 # ctags
 
 > Generates an index (or tag) file of language objects found in source files for many popular programming languages.
-> More information: <https://ctags.io/>
+> More information: <https://ctags.io/>.
 
 - Generate tags for a single file, and save them in a file named "tags" in the current directory:
 
--- a/pages/common/ctags.md
+++ b/pages/common/ctags.md
@@ -3,14 +3,14 @@
 > Generates an index (or tag) file of language objects found in source files for many popular programming languages.
 > More information: <https://ctags.io/>.
 
-- Generate tags for a single file, and save them in a file named "tags" in the current directory:
+- Generate tags for a single file, and output them to a file named "tags" in the current directory, overwriting the file if it exists:
 
 `ctags {{path/to/file}}`
 
-- Generate tags for all files in the current directory and save them in a file with a custom filename:
+- Generate tags for all files in the current directory, and output them to a specific file, overwriting the file if it exists:
 
 `ctags -f {{filename}} *`
 
-- Generate tags for all files in the current directory and all subdirectories (-R, --recurse):
+- Generate tags for all files in the current directory and all subdirectories:
 
-`ctags -R`
+`ctags --recurse`
--- a/pages/common/ctags.md
+++ b/pages/common/ctags.md
@@ -1,0 +1,16 @@
+# ctags
+
+> Generates an index (or tag) file of language objects found in source files for many popular programming languages.
+> More information: <https://ctags.io/>
+
+- Generate tags for a single file, and save them in a file named "tags" in the current directory:
+
+`ctags {{path/to/file}}`
+
+- Generate tags for all files in the current directory and save them in a file with a custom filename:
+
+`ctags -f {{filename}} *`
+
+- Generate tags for all files in the current directory and all subdirectories (-R, --recurse):
+
+`ctags -R`


pages/common/ctags.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Generate tags for a single file, and save them in a file named "tags" in the current directory:
+
+`ctags {{path/to/file}}`
+
+- Generate tags for all files in the current directory and save them in a file with a custom filename:
+
+`ctags -f {{filename}} *`
+
+- Generate tags for all files in the current directory and all subdirectories (-R, --recurse):
```
zdroid:
```suggestion
- Generate tags for all files in the current directory and all subdirectories:
```

pages/common/ctags.md
```diff
@@ -0,0 +1,16 @@
...
+
+`ctags -R`
```
zdroid:
```suggestion
`ctags --recurse`
```
It's not `--recursive`, but `--recurse`, right?

pages/common/ctags.md
```diff
@@ -0,0 +1,16 @@
+# ctags
+
+> Generates an index (or tag) file of language objects found in source files for many popular programming languages.
+> More information: <https://ctags.io/>
+
+- Generate tags for a single file, and save them in a file named "tags" in the current directory:
+
+`ctags {{path/to/file}}`
+
+- Generate tags for all files in the current directory and save them in a file with a custom filename:
```
zdroid: This needs to be rewritten.

pages/common/ctags.md
```diff
@@ -0,0 +1,16 @@
+# ctags
+
+> Generates an index (or tag) file of language objects found in source files for many popular programming languages.
+> More information: <https://ctags.io/>
+
+- Generate tags for a single file, and save them in a file named "tags" in the current directory:
```
zdroid: To be rewritten, similar to the other example.

pages/common/ctags.md
thezeroalpha: Yes, it's `--recurse`. My thought process with this example was that it's much more common for people to use the short form (`-R`) in both scripts and commands, so I figured it would be better to give that form as the example, while explaining the full meaning in the description. 

Do you think it's better to just give the long form in the example, and not mention the short form?

pages/common/ctags.md
thezeroalpha: I assume you mean line 10 specifically; what would you suggest? I.e., what's unclear, or what should be explained better?

pages/common/ctags.md
zdroid: In this project it's preferred to use the longer version if it makes things more clear. I'm neutral on that, but I'd leave it here up for the discussion.

pages/common/ctags.md
zdroid: For example like this. I edited my original comment to include a suggestion. It isn't unclear, but I feel it's too long to explain something not so complicated.

pages/common/ctags.md
sbrl: I guess it depends on how often this command is used.

We could potentially use the short version if we use the word `recurse` in the description, but I'd probably say that it's best to use the long-form argument here to be consistent with other pages.

pages/common/ctags.md
sbrl: Perhaps `to a` instead of `in a` there too? I do agree it reads a little better.

pages/common/ctags.md
thezeroalpha: Sure, I know that's the convention for tldr pages, which is why I was debating adding it. Another reason for including the short option that I considered is that usually, you can guess the short option from the long option (e.g. for the `--recursive` option for other tools, the short option will usually be `-r`) . Here that's not the case, as the short option is not `-r`, but `-R`.

Still, if the general agreement is that it doesn't add any value to have it in the tldr page, I'll remove it.

pages/common/ctags.md
thezeroalpha: Agreed, what do you think of phrasing it as "...and output them to a specific file, overwriting the file if it exists"? I thought it might be useful to also mention that the file would be overwritten, as normally a user would expect the program to warn/prompt them in case a file would be overwritten, but ctags doesn't do that.

pages/common/ctags.md
sbrl: I see. Generally we'll use a short option if we can somehow include the letter in the description text (e.g. `- [R]ecurse over foo:`).

I think in this instance we should use the long option.

pages/common/ctags.md
sbrl: Sounds good to me!

PR #3859, opened 2020-02-15
Tar: Add extraction of file without folder structure

Firesphere:
Most commonly used with e.g. Solr, which has these instructions for installation, but might come in handy for other extraction operations as well.

### DIFF
--- a/pages/common/tar.md
+++ b/pages/common/tar.md
@@ -32,6 +32,6 @@
 
 `tar xf {{source.tar}} --wildcards {{"*.html"}}`
 
-- Extract a specific file without the folder structure
+- Extract a file without making the folder structure:
 
 `tar xf {{source.tar}} {{source.tar}}/path/to/binary.sh --strip-components={{amount-of-folders-deep}}`
--- a/pages/common/tar.md
+++ b/pages/common/tar.md
@@ -32,6 +32,6 @@
 
 `tar xf {{source.tar}} --wildcards {{"*.html"}}`
 
-- Extract a file without making the folder structure:
+- Extract a file or folder the folder structure in the archive:
 
-`tar xf {{source.tar}} {{source.tar}}/path/to/binary.sh --strip-components={{amount-of-folders-deep}}`
+`tar xf {{source.tar}} {{source.tar/path/to/extract}} --strip-components={{depth_to_strip}}`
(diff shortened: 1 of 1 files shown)


pages/common/tar.md
```diff
@@ -31,3 +31,7 @@
 - Extract files matching a pattern:
 
 `tar xf {{source.tar}} --wildcards {{"*.html"}}`
+
+- Extract a specific file without the folder structure
```
zdroid:
```suggestion
- Extract a file without making the folder structure:
```

pages/common/tar.md
```diff
@@ -31,3 +31,7 @@
...
+
+`tar xf {{source.tar}} {{source.tar}}/path/to/binary.sh --strip-components={{amount-of-folders-deep}}`
```
zdroid: I don't understand why is `{{source.tar}}/path/to/binary.sh` present here. Anyway, if it stays, it should be put inside `{{}}`.

Firesphere: Yes, the whole example should be between the braces.

pages/common/tar.md
sbrl: That still doesn't really make sense @zdroid. What about this?
```suggestion
- Extract a specific file without preserving the folder structure:
```

Firesphere:
Apologies for the delay. I've been quite busy. I'll try to look at the feedback this week!

pages/common/tar.md
agnivade: I think the `strip-components` could be explained better. I would say something in the lines of - "Extract a file without creating the given level of directories".

And also, I think just mentioning `--strip-components` is enough. There's no need to extract a single file. The most common usage is to just skip a directory.

Lastly, we use snake_case in our tokens. But we can just shorten it to - {{depth}}

Firesphere:
Sorry stale bot, some virus got in the way. I'll try to get to it soon!

pages/common/tar.md
```diff
@@ -31,3 +31,7 @@
 - Extract files matching a pattern:
 
 `tar xf {{source.tar}} --wildcards {{"*.html"}}`
+
+- Extract a file or folder the folder structure in the archive:
```
agnivade: The description is incorrect. There have been many suggested alternatives in the review. Let's go with one of them.

Firesphere: Agreed, I thought I put that in there, but apparently not. Updated now.

PR #3862, opened 2020-02-17
Add further at example

tpo:
I have to look that particular kind of command often, but seldom enough that I forget in between. The added example show the following two additional at features:

* passing the command in via pipe
* executing on a specified date

### DIFF
--- a/pages/linux/at.md
+++ b/pages/linux/at.md
@@ -14,6 +14,6 @@
 
 `at {{hh:mm:ss}} -f {{path/to/file}}`
 
-- Reboot at 11pm on February 18th
+- Reboot at 11pm on February 18th:
 
-`echo "reboot" | at 11pm Feb 18
+`echo "reboot" | at {{11pm}} {{Feb 18}}
--- a/pages/linux/at.md
+++ b/pages/linux/at.md
@@ -16,4 +16,4 @@
 
 - Reboot at 11pm on February 18th:
 
-`echo "reboot" | at {{11pm}} {{Feb 18}}
+`echo "reboot" | at {{11pm}} {{Feb 18}}`
--- a/pages/linux/at.md
+++ b/pages/linux/at.md
@@ -13,3 +13,7 @@
 - Execute a script at the given time:
 
 `at {{hh:mm:ss}} -f {{path/to/file}}`
+
+- Reboot at 11pm on February 18th
+
+`echo "reboot" | at 11pm Feb 18
--- a/pages/linux/at.md
+++ b/pages/linux/at.md
@@ -14,6 +14,6 @@
 
 `at {{hh:mm:ss}} -f {{path/to/file}}`
 
-- Reboot at 11pm on February 18th:
+- Reboot at 11pm on February 18th (needs to be executed as root):
 
 `echo "reboot" | at {{11pm}} {{Feb 18}}`


zdroid:
I edited the file to solve the issues presented by the linter bot. Also, I put `11pm` and `Feb 18` inside the braces, because those are not hard-coded, but variables free to be changed.

Also, please sign the CLA.

tpo:
> I edited the file to solve the issues presented by the linter bot. Also, I put `11pm` and `Feb 18` inside the braces, because those are not hard-coded, but variables free to be changed.

Woah, thanks a lot  @zdroid !
 
> Also, please sign the CLA.

tpo:
> Also, please sign the CLA.

I am unable to do that, since cla-assistant.io wants me to give it rights to access all kinds of things (private Gists? What the fuck?). Is cla-assistant.io yet another of those applications created by Satan that is free but requires you to sign over your soul to him?

Whatever, I herewith authorize you, Zlatan to do whatever you wish with my contribution here (aka public domain). Thanks!

sbrl:
That's really odd that @claassistantio is requiring you to grant access to all those things.

@agnivade / @pxgamer, thoughts?

I don't remember having to do that myself.

> I'd be very happy if you could push it upstream or such.

This is upstream - once this PR is merged your contribution will be part of tldr-pages :-)

pages/linux/at.md
```diff
@@ -13,3 +13,7 @@
 - Execute a script at the given time:
 
 `at {{hh:mm:ss}} -f {{path/to/file}}`
+
+- Reboot at 11pm on February 18th:
+
+`echo "reboot" | at {{11pm}} {{Feb 18}}`
```
sbrl: Wouldn't this need to be `sudo`?
```suggestion
`echo "reboot" | sudo at {{11pm}} {{Feb 18}}`
```

zdroid: Perhaps putting the braces around `reboot` is not needed, since it doesn't say "execute a command `reboot`", but just "reboot".

tpo: * `reboot` really *is* a shell command (thanks for fixing up my markup!)
* `sudo` might be needed depending on whether one is root or not. I'd think it's self evident that the `reboot` command (and thus the `at` command) needs to be executed as root, so I'd not add the example with `sudo`

sbrl: Right. Note that your markup isn't actually fixed - the above is just a suggestion that you need to agree to by clicking the "commit suggestion" button.

I've updated my suggestion above to remove the `sudo` and make the command more generic. Perhaps we could alter the description for this command too?

My concern is that `reboot` is a specific command that does require `sudo`, so the example wouldn't necessarily work out-of-the-box - depending on who the reader is logged into their terminal as. It is for this reason that I'd prefer a command that does not require `sudo` privileges for the example.

agnivade:
@sbrl - I think I remember the assistant asking me for such permissions. It's a github app and I think the permissions are a bit coarse, so it does ask a bit more than it needs.

By upstream, OP meant for somebody to push it on their behalf.

pages/linux/at.md
tpo: Rebooting a machine on off-office hours is by far the most common use for `at` I have and I suspect it might be quite common. So I think it makes sense to use "reboot" as an example.

sbrl: I see, @tpo. I'm still concerned that the example might not work for some users though. For example, if I'm logged in as a regular user and copy-paste the command (as a beginner might do), it will fail because the `at` command was not preceded by `sudo`.

tpo: See https://github.com/tpo/tldr/commit/06b6cae8570c6f0d9de2aa1f1363bc2f676e8702

You are supposing that making that `sudo` will somehow fail less. I think that belief is unsubstanciated, as sudo isn't available by default on Linux systems and setting up sudo is not trivial. So I suggest to just mention the fact that one needs to be root.

sbrl: > sudo isn't available by default on Linux systems

Erm what kind of Linux system are you using? On Ubuntu, Arch, Centos, and Fedora it's definitely installed and configured by default as far as I can tell.

Thanks for the edit - I guess that could be ok for now (though another alternative to `reboot` might be simply `{{command}}`, which would be consistent with some other pages.

tpo: I'm on Debian.

We have already discussed `{{command}}`.

tpo:
> Unfortunately we're unable to merge until you sign the CLA though confused

See https://github.com/tldr-pages/tldr/pull/3862#issuecomment-587138737

zdroid:
@sbrl can't org owner force the merge?

sbrl:
Yeah @zdroid, I can. The thing is that the CLA is there to manage the process of signing the agreement, so we can programmatically tell who has signed it.

By force-merging this, @tpo will show as having not signed the CLA. It's for this reason that I would rather not merge this without the OP having signed the CLA. I think we had another instance of this recently too.

/cc @waldyrious, @pxgamer, @agnivade - am I being too cautious here?

agnivade:
I think since OP has given @zdroid permission to take over the content, we can simply have Zlatan open a new PR, and then merge it. That way we get the job done, and yet include commits for only members who have signed the CLA.

PR #3863, opened 2020-02-17

schneiderl:
This is a follow up to #3352 and #3846.

If you have a better way to phrase those sentences, please propose changes so that we can discuss them.

### DIFF
--- a/pages/common/jq.md
+++ b/pages/common/jq.md
@@ -1,6 +1,6 @@
 # jq
 
-> A command-line JSON processor.
+> A command-line JSON processor that uses a domain-specific language.
 > More information: <https://stedolan.github.io/jq>.
 
 - Output a JSON file, in pretty-print format:
--- a/pages/common/jq.md
+++ b/pages/common/jq.md
@@ -1,6 +1,6 @@
 # jq
 
-> A lightweight and flexible command-line JSON processor.
+> A command-line JSON processor.
 > More information: <https://stedolan.github.io/jq>.
 
 - Output a JSON file, in pretty-print format:
--- a/pages/common/pass.md
+++ b/pages/common/pass.md
@@ -1,6 +1,6 @@
 # pass
 
-> A tool to store and read passwords or other sensitive data easily.
+> Tool for storing and reading passwords or other sensitive data.
 > All data is GPG-encrypted, and managed with a git repository.
 > More information: <https://www.passwordstore.org>.
 
--- a/pages/common/pass.md
+++ b/pages/common/pass.md
@@ -1,6 +1,6 @@
 # pass
 
-> Safely store and read passwords or other sensitive data easily.
+> A tool to store and read passwords or other sensitive data easily.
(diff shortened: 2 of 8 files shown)


zdroid:
Should this be rebased and merged or squashed and merged?

pages/common/jq.md
```diff
@@ -1,6 +1,6 @@
 # jq
 
-> A lightweight and flexible command-line JSON processor.
+> A command-line JSON processor.
```
sbrl: Perhaps we could include some additional information here?
```suggestion
> A command-line JSON processor that uses a domain-specific language.
```

pages/common/pass.md
```diff
@@ -1,6 +1,6 @@
 # pass
 
-> Safely store and read passwords or other sensitive data easily.
+> A tool to store and read passwords or other sensitive data easily.
```
sbrl: Perhaps this will read better?
```suggestion
> Tool for storing and reading passwords or other sensitive data.
```

pages/common/sops.md
```diff
@@ -1,7 +1,7 @@
 # sops
 
 > SOPS: Secrets OPerationS.
-> Simple and flexible tool for managing secrets.
+> A tool for managing secrets.
```
sbrl: Perhaps this will read better?
```suggestion
> Tool for managing secrets.
```

pages/linux/prt-get.md
```diff
@@ -1,6 +1,6 @@
 # prt-get
 
-> The advanced CRUX package manager.
+> A CRUX package manager.
```
sbrl: This implies that there is more than 1 package manager. Is this the case? If not, we should keep using `The` instead.

pages/linux/tlp.md
```diff
@@ -1,6 +1,6 @@
 # tlp
 
-> Advanced power management for Linux. See `tlp-stat` page for additional information.
+> Power management for Linux. See `tlp-stat` page for additional information.
```
sbrl: I'd perhaps suggest that in this case we should keep the `advanced`. It differentiates the tool from existing basic power management features built-in to one's operating system. It suggests that additional more advanced features are provided by the command, which is the whole aim of the command itself.

pages/common/pass.md
schneiderl: nice

pages/linux/tlp.md
schneiderl: While writing this I've also thought of that. I'm changing it because I feel neutral about this.

pages/linux/prt-get.md
schneiderl: I've done some research and didn't find any other package manager.
I'll wait for a few hours to give others some time to read this (maybe someone knows if there is another package manager) and then I'll change it.

schneiderl:
Changed `prt-get` condition after some research because I could not find any other package manager for CRUX.

PR #3866, opened 2020-02-22
wormhole: add page

vakili:


sbrl:
Good point, @zdroid.

@vakili, do you think we could have a single copy in `common` please? Pages that apply to multiple platforms should go in `common` to avoid duplication. If there are platform-specific differences, we can make an altered copy in that platform's folder, which will override the copy in `common`.

zdroid:
According to https://magic-wormhole.readthedocs.io/en/latest/welcome.html#installation, this works on Windows aswell. Definitely should be in common.

vakili:
Thank you for the feedback. I have moved the page to common.

pages/common/wormhole.md
```diff
@@ -0,0 +1,16 @@
+# wormhole
+
+> Get things from one computer to another, safely.
+> More information: <https://magic-wormhole.readthedocs.io/en/latest/>.
+
+- Send a file:
+
+`wormhole send {{file}}`
```
sbrl: How does it know where to send it? Also, can we tweak this to make it more consistent with other pages please?
```suggestion
`wormhole send {{path/to/file}}`
```

pages/common/wormhole.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Send a file:
+
+`wormhole send {{path/to/file}}`
+
+- Receive a file:
+
+`wormhole receive {{wormhole_code}}`
+
+- Send raw text:
```
agnivade: Where is the text taken from ?

PR #3867, opened 2020-02-23
tar: add a relative paths example

pshirshov:

### DIFF
--- a/pages/common/tar.md
+++ b/pages/common/tar.md
@@ -12,7 +12,7 @@
 
 `tar czf {{target.tar.gz}} {{file1}} {{file2}} {{file3}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/tar.md
```diff
@@ -12,6 +12,10 @@
 
 `tar czf {{target.tar.gz}} {{file1}} {{file2}} {{file3}}`
 
+- Create a gzipped archive from directory omitting leading path:
```
zdroid: What is `omitting leading path` supposed to mean? It's pretty unclear here, so it should be changed.

pshirshov: I've updated it, is it better now?

PR #3869, opened 2020-02-23
passwd: add base command

Aracki:
Add missing & most commonly used `passwd` example.

### DIFF
--- a/pages/common/passwd.md
+++ b/pages/common/passwd.md
@@ -2,7 +2,7 @@
 
 > Passwd is a tool used to change a user's password.
 
-- Change the password interactively:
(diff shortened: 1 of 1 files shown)


pages/common/passwd.md
```diff
@@ -2,6 +2,10 @@
 
 > Passwd is a tool used to change a user's password.
 
+- Change the password interactively:
```
mebeim:
```suggestion
- Change the password of the current user interactively:
```
Let's specify that's the password of the *current* user.

Aracki: It makes sense. Thanks for the suggestion.

PR #3871, opened 2020-02-25
vault: add command

Aracki:
Really important one that was missing. I thought it was better to use `field_name` instead of already used in previous commands `value`.

### DIFF
--- a/pages/common/vault.md
+++ b/pages/common/vault.md
@@ -23,6 +23,10 @@
 
 `vault read secret/{{hello}}`
 
+- Read a specific field from the value:
+
+`vault read -field={{field_name}} secret/{{hello}}`
+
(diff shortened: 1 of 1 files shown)


pages/common/vault.md
```diff
@@ -23,6 +23,10 @@
 
 `vault read secret/{{hello}}`
 
+- Read a specific field from the value:
+
+`vault read -field={{field_name}} secret/{{hello}}`
```
sbrl: Is there a dash missing here?
```suggestion
`vault read --field={{field_name}} secret/{{hello}}`
```

Aracki: Not at all. Official docs: https://www.vaultproject.io/docs/commands/read/#output-options.

PR #3872, opened 2020-02-25
snap: add update package

jxu:

I am surprised this isn't in here yet as it is one of the most important commands for a package manager. This is very different from updating all packages with `snap refresh`. You may be able to guess the syntax for updating one package if you know apt but the tldr pages are not about guessing.

### DIFF
--- a/pages/linux/snap.md
+++ b/pages/linux/snap.md
@@ -11,6 +11,10 @@
 
 `snap install {{package_name}}`
 
+- Update a package:
+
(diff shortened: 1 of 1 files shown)


jxu:
I tend to use snap for large programs that aren't updated as frequently in apt. Example: arduino was 2 years out of date in normal apt packages. Each snap is independent and may be 100 MB and a 1 minute install. If I want to update vlc I don't want to through updating gimp, arduino, etc.

PR #3883, opened 2020-02-27
tldr: update the description

zdroid:

### DIFF
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -1,6 +1,6 @@
 # tldr
 
-> Simplified man pages.
+> Command-line client for tldr pages: simplified and community-driven man pages.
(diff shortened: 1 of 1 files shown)


mebeim:
This is different from the other PR that was about to be merged. It reads:

> Command-line client for tldr pages.
> Displays simplified and community-driven man pages.

While previously it read:

> Command-line client for tldr pages.

I don't really think the second line is needed.

zdroid:
@mebeim Read @agnivade's comment in the original discussion.
Also, previously it read `Simplified man pages`, not `Command-line client for tldr pages`. That's why we agreed to change it in the first place.

mebeim:
@zdroid ah, missed that. I guess it makes sense. Thank you!

PR #3884, opened 2020-02-27
git-lfs: add fetch and checkout subcommands

sbrl:
Oops! I accidentally committed this to master instead of opening a PR. Hopefully all should be fixed now, thanks to a sneaky ninja force-push.

Anyway, I've found this pair of commands extremely useful, as git lfs doesn't always like to automatically check out files.

### DIFF
--- a/pages/common/git-lfs.md
+++ b/pages/common/git-lfs.md
@@ -26,3 +26,11 @@
 - Push all LFS objects to the remote server (useful if errors are encountered):
 
 `git lfs push --all {{remote_name}} {{branch_name}}`
+
+- Fetch all LFS objects:
+
+`git lfs fetch`
+
+- Checkout all Git LFS objects:
+
+`git lfs checkout`
--- a/pages/common/git-lfs.md
+++ b/pages/common/git-lfs.md
@@ -23,11 +23,11 @@
 
 `git lfs ls-files`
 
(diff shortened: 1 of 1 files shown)


pages/common/git-lfs.md
```diff
@@ -26,3 +26,11 @@
 - Push all LFS objects to the remote server (useful if errors are encountered):
 
 `git lfs push --all {{remote_name}} {{branch_name}}`
+
+- Fetch all LFS objects:
+
+`git lfs fetch`
+
+- Checkout all Git LFS objects:
```
zdroid: Is there a technical difference between "Git LFS objects" and "LFS objects"? As far as I can tell, "objects" are already referring to "Git objects" in the repository.

sbrl: Git LFS Objects == LFS Objects, but Git Objects ≠ Git LFS Objects as far as I know.

Git LFS objects are stored separately to regular Git objects, with a reference to the LFS Object as a regular Git object - which gets replaced by `git-lfs` on checkout automatically (or, at least, it should - but aren't always, hence the edit).

zdroid: Okay, so can the examples be changed to use only one of the terms "Git LFS objects" and "LFS objects", in order to avoid ambiguity?

PR #3894, opened 2020-03-06
deno: add page

DarylPinto:

### DIFF
--- a/pages/common/deno.md
+++ b/pages/common/deno.md
@@ -5,20 +5,20 @@
 
 - Run a JavaScript or TypeScript file:
 
-`deno file.ts`
+`deno {{path/to/file.ts}}`
 
 - Start a REPL (interactive shell):
 
 `deno`
 
-- Run a file with access to the network:
+- Run a file with network access enabled:
 
-`deno --allow-net server.ts`
+`deno --allow-net {{path/to/file.ts}}`
 
 - Run a file from a URL:
 
-`deno https://deno.land/std/examples/welcome.ts`
+`deno {{https://deno.land/std/examples/welcome.ts}}`
 
 - Install an executable script from a URL:
 
-`deno install --allow-net --allow-read file_server https://deno.land/std/http/file_server.ts`
+`deno install --allow-net --allow-read {{file_server}} {{https://deno.land/std/http/file_server.ts}}`
--- a/pages/common/deno.md
+++ b/pages/common/deno.md
@@ -1,0 +1,25 @@
+# deno
+
+> A secure runtime for JavaScript and TypeScript
+> More information: <https://deno.land/>.
+
+- Run a JavaScript or TypeScript file:
+
+`deno file.ts`
+
+- Start a REPL (interactive shell):
+
+`deno`
+
+- Run a file with access to the network:
+
+`deno --allow-net server.ts`
+
+- Run a file from a URL:
+
+`deno https://deno.land/std/examples/welcome.ts`
+
+- Install an executable script from a URL:
+
+`deno install --allow-net --allow-read file_server https://deno.land/std/http/file_server.ts`
+
--- a/pages/common/deno.md
+++ b/pages/common/deno.md
@@ -1,6 +1,6 @@
 # deno
 
-> A secure runtime for JavaScript and TypeScript
+> A secure runtime for JavaScript and TypeScript.
 > More information: <https://deno.land/>.
 
 - Run a JavaScript or TypeScript file:
@@ -22,4 +22,3 @@
 - Install an executable script from a URL:
 
 `deno install --allow-net --allow-read file_server https://deno.land/std/http/file_server.ts`
-


pages/common/deno.md
```diff
@@ -0,0 +1,25 @@
+# deno
+
+> A secure runtime for JavaScript and TypeScript
+> More information: <https://deno.land/>.
+
+- Run a JavaScript or TypeScript file:
+
+`deno file.ts`
+
+- Start a REPL (interactive shell):
```
zdroid:
```suggestion
- Start a REPL (interactive shell):
```

pages/common/deno.md
```diff
@@ -0,0 +1,25 @@
...
+
+`deno`
+
+- Run a file with access to the network:
```
zdroid:
```suggestion
- Run a file with the access to the network:
```

pages/common/deno.md
```diff
@@ -0,0 +1,25 @@
+# deno
+
+> A secure runtime for JavaScript and TypeScript
+> More information: <https://deno.land/>.
+
+- Run a JavaScript or TypeScript file:
+
+`deno file.ts`
```
zdroid:
```suggestion
`deno {{path/to/file}}`
```

pages/common/deno.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Start a REPL (interactive shell):
+
+`deno`
+
+- Run a file with access to the network:
+
+`deno --allow-net server.ts`
```
zdroid:
```suggestion
`deno --allow-net {{path/to/file}}`
```

pages/common/deno.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Run a file from a URL:
+
+`deno https://deno.land/std/examples/welcome.ts`
```
zdroid:
```suggestion
`deno {{file_url}}`
```

I'm not sure on this one, maybe it should be some other name for the parameter.

pages/common/deno.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Install an executable script from a URL:
+
+`deno install --allow-net --allow-read file_server https://deno.land/std/http/file_server.ts`
```
zdroid: I don't know what is `file_server` here exactly (a file?), but this should be changed just as suggested for the previous examples.

pages/common/deno.md
DarylPinto: While I understand the intent of this suggestion, I don't think it matches the format of other similar commands. For example, see: 

- `tldr node`
- `tldr python`
- `tldr rustc`

pages/common/deno.md
DarylPinto: I don't think this suggestion sounds quite right. What do you think of: "Run a file with network access"?

pages/common/deno.md
DarylPinto: See my comment here: https://github.com/tldr-pages/tldr/pull/3894/files#r389159180

pages/common/deno.md
zdroid: You mean this should have the `.ts` extension? Keep in mind that JS / TS extension is not required for a file to be recognized as such.

pages/common/deno.md
DarylPinto: `flie_server` here is the name of the exectuable after installing. After running the command, `file_server` is now installed on the user's machine and can be run from anywhere.

zdroid: Is that a name you can choose yourself? If yes, then it should be inside `{{ }}`.

DarylPinto:
@zdroid I feel like changing all the references into `{{path/to/file}}` (or similar) takes a lot away from the beauty of `tldr`'s scannability and ease of understanding at a glance.

The README of this repository is a perfect example of how using snippets that can be pasted as real commands allows for an instant understanding of real world use cases:

preview

I would like to see what others think before moving forward as changes like these that obfuscate file names and URLs seem more akin to `man` pages or similar than to `tldr` itself.

Personally I feel like the original proposal is the clearest and easiest to understand, which at the end of the day is the purpose of `tldr`, isn't it?

zdroid:
@DarylPinto You are right about that. It's just been a general policy of the project so far, to use path/to/file, that's why I suggested it. Let's wait for the input of other collaborators.

pages/common/deno.md
sbrl: True @zdroid, but it can also be used as a hint. I suggest this:
```suggestion
`deno {{path/to/file.ts}}`
```

pages/common/deno.md
sbrl: What about this?
```suggestion
`deno https://example.com/std/examples/welcome.ts`
```

sbrl:
Hey, @DarylPinto! I think you've got confused there. The `{{token_syntax}}` is actually present on all pages. For example, check out [the source to the `tar` command's page](https://raw.githubusercontent.com/tldr-pages/tldr/master/pages/common/tar.md).

tldr clients actually handle the syntax specially, and so don't display it to the user. For example, if you use the Node.js reference client, you'll find that it doesn't display the syntax but instead colours it differently.

pages/common/deno.md
mebeim: This is correct as is, maybe a clearer version could be:

    - Run a file with network access enabled:

which is indeed what the command does.

pages/common/deno.md
mebeim: We usually use full URLs where possible. Changing the URL to `example.com` doesn't make much sense since it becomes invalid. The provided URL is already ok, it just needs to be enclosed in curly braces `{{ }}`.

pages/common/deno.md
mebeim:
```suggestion
`deno install --allow-net --allow-read {{file_server}} {{https://deno.land/std/http/file_server.ts}}`
```
Just adding braces is fine here.

mebeim:
Hey, thank you for the page @DarylPinto :)

We tend to use `{{path/to/file.ext}}` where applicable for new pages (see the [style guide](https://github.com/tldr-pages/tldr/blob/master/contributing-guides/style-guide.md)).  The fact you don't see it in `tar` is because there are multiple filenames in a single command line and that would become unreadable. The fact that you don't see it in other programming language related pages is because either (1) the command line has multiple files again or (2) the decision to use `{{path/to/file.ext}}` was made after those pages were created.

Nonetheless, changing `deno {{file.ts}}` to `deno {{path/to/file.ts}}` doesn't hurt readability.

pages/common/deno.md
zdroid: That's good, too. My main concern was about path-to-file format, not the extension.

DarylPinto:
@sbrl @mebeim Ah yes I was mistaken; thank you for the clarification! I've added changes from the feedback in your various comments. Please have a look and see if the page is stylistically correct now.

PR #3902, opened 2020-03-10
Add date RFC 3339

a-b:
RFC 3339 widely used in logging and automation

### DIFF
--- a/pages/common/date.md
+++ b/pages/common/date.md
@@ -25,5 +25,3 @@
 - Display RFC-3339, great for logging and timestamps YYYY-MM-DD hh:mm:ss TZ
 
 `date ---rfc-3339=s`
-
-
--- a/pages/common/date.md
+++ b/pages/common/date.md
@@ -22,6 +22,6 @@
 
 `date -d "{{2018-09-01 00:00}}" +%s --utc`
 
-- Display RFC-3339, great for logging and timestamps YYYY-MM-DD hh:mm:ss TZ
+- Display the current date using the RFC-3339 format (`YYYY-MM-DD hh:mm:ss TZ`):
 
 `date --rfc-3339=s`
--- a/pages/common/date.md
+++ b/pages/common/date.md
@@ -21,3 +21,9 @@
 - Convert a specific date to the Unix timestamp format:
 
 `date -d "{{2018-09-01 00:00}}" +%s --utc`
+
+- Display RFC-3339, great for logging and timestamps YYYY-MM-DD hh:mm:ss TZ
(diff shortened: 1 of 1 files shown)


pages/common/date.md
```diff
@@ -21,3 +21,9 @@
 - Convert a specific date to the Unix timestamp format:
 
 `date -d "{{2018-09-01 00:00}}" +%s --utc`
+
+- Display RFC-3339, great for logging and timestamps YYYY-MM-DD hh:mm:ss TZ
+
+`date ---rfc-3339=s`
```
sbrl: This example doesn't work for me on Ubuntu 19.10. What system are you using?
```
$ date ---rfc-3339=s
date: unrecognised option '---rfc-3339=s'
Try 'date --help' for more information.
```

pages/common/date.md
```diff
@@ -21,3 +21,9 @@
 - Convert a specific date to the Unix timestamp format:
 
 `date -d "{{2018-09-01 00:00}}" +%s --utc`
+
+- Display RFC-3339, great for logging and timestamps YYYY-MM-DD hh:mm:ss TZ
```
zdroid:
```suggestion
- Display RFC-3339 logging with timestamps:
```
Much simpler this way

pages/common/date.md
zdroid: It should be with `--`, not `---`. I edited the file to correct those pesky errors.

pages/common/date.md
mebeim: Hrm, @zdroid this does not convey the same meaning. "Display RFC-3339 logging" does not make much sense, the `date` command has nothing to do with logging *per se*, and neither does the RFC-3339, it's just a date format. Other than that, keeping the date format makes sense IMHO, that way you don't have to try the command to know the format. I would suggest the following:

```suggestion
- Display the current date using the RFC-3339 format (`YYYY-MM-DD hh:mm:ss TZ`):
```

Which is also the same way other examples in the page are worded.

zdroid: Yes, you're right. I tried to rephrase the original, but I realized it has nothing to do with logging after running the command.

pages/common/date.md
sbrl: Ah, I see. Thanks @zdroid :-)

zdroid:
I've edited this myself, since there was no resonse.

a-b:
Thanks @mebeim!

sbrl:
oops, I forgot to edit the commit message

PR #3904, opened 2020-03-10
uncrustify: add page

sbrl:
Ran across a use-case for this when patching some badly formatted C++ code. I could have done with this page this afternoon :P


zdroid:
@sbrl you need some editorconfig-ish implementation in your editor. Trailing whitespace should be cut out automatically.

sbrl:
Oops, I must have missed that. I actually use gedit (or sometimes Atom, but not in this case) for writing pages, and then I usually copy/paste into the GH web interface.

PR #3907, opened 2020-03-11
ssh-agent: add page

nibalizer:
- [x ] The page (if new), does not already exist in the repo.
- [x ] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x ] The page has 8 or fewer examples.
- [x ] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x ] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x ] The page description includes a link to documentation or a homepage (if applicable).

### DIFF
--- a/pages/common/ssh-agent.md
+++ b/pages/common/ssh-agent.md
@@ -1,0 +1,18 @@
+# ssh-agent
+
+> Spawn an ssh-agent process.
+> The ssh-agent will hold your ssh key decrypted in memory until evicted.
+> Add and manage keys with `ssh-add`.
+
+- Start an SSH Agent for the current shell:
+
+`eval $(ssh-agent)`
+
+- Kill the current agent:
+
+`ssh-agent -k`
+
+- Manually connect to an existing agent (rare):
+
+`export SSH_AUTH_SOCK={{ssh-agent socket}}`
+`export SSH_AGENT_PID={{ssh-agent pid}}`
--- a/pages/common/ssh-agent.md
+++ b/pages/common/ssh-agent.md
@@ -8,7 +8,7 @@
 
 `eval $(ssh-agent)`
 
-- Kill the current agent:
+- Kill the currently running agent:
 
 `ssh-agent -k`
 
--- a/pages/common/ssh-agent.md
+++ b/pages/common/ssh-agent.md
@@ -1,8 +1,8 @@
 # ssh-agent
 
-> Spawn an ssh-agent process.
-> The ssh-agent will hold your ssh key decrypted in memory until evicted.
-> Add and manage keys with `ssh-add`.
+> Spawn an SSH Agent process.
+> An SSH Agent holds ssh keys decrypted in memory until removed or the process is killed.
+> See also `ssh-add`, which can add and manage keys held by an SSH Agent.
 
 - Start an SSH Agent for the current shell:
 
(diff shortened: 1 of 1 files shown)


pages/common/ssh-agent.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Kill the current agent
+
+`ssh-agent -k`
+
+- Manually connect to an existing agent (rare)
+
+`
+export SSH_AUTH_SOCK={{ssh-agent socket}}
+export SSH_AGENT_PID={{ssh-agent pid}}
```
zdroid: It should be one command per example.

nibalizer: I think you need to have both of these to get the effect you want.

sbrl: Indeed you do, @nibalizer. I think that the first example is sufficient though - but we could tweak the page description I guess.

pages/common/ssh-agent.md
```diff
@@ -0,0 +1,18 @@
+# ssh-agent
+
+> Spawn an ssh-agent process.
+> The ssh-agent will hold your ssh key decrypted in memory until evicted.
+> Add and manage keys with `ssh-add`.
```
sbrl: Perhaps this will read better?
```suggestion
> Spawn an SSH Agent process.
> An SSH Agent holds ssh keys decrypted in memory until removed or the process is killed.
> See also `ssh-add`, which can add and manage keys held by an SSH Agent.
```

pages/common/ssh-agent.md
```diff
@@ -0,0 +1,18 @@
...
+
+- Start an SSH Agent for the current shell:
+
+`eval $(ssh-agent)`
+
+- Kill the current agent:
```
sbrl: Perhaps this will read better?
```suggestion
- Kill the currently running agent:
```

pages/common/ssh-agent.md
```diff
@@ -0,0 +1,18 @@
...
+
+`ssh-agent -k`
+
+- Manually connect to an existing agent (rare):
+
+`export SSH_AUTH_SOCK={{ssh-agent socket}}`
+`export SSH_AGENT_PID={{ssh-agent pid}}`
```
sbrl: Just a suggestion per my previous comment.
```suggestion
`export SSH_AUTH_SOCK={{ssh_agent_socket}} SSH_AGENT_PID={{ssh_agent_pid}}`
```

nibalizer: We can do this if you want, I don't think exporting two variables on one line is clearer though.

sbrl: Sorry for aking so lonig to get back to you.

True, but what @zdroid was trying to say is that in tldr pages we only have 1 line of command for each example. In addition, I'm not really sure of the usefulness of this example as it is, since the `eval` one does this for you.

tldr pages aren't supposed to be _exhaustive_ - just include the most commonly used and most useful examples. So given the either existing example, I'm not sure we need this example unless you've got a good use-cas for not using the `eval` one.

nibalizer: Ok I will remove it. We can always add it back if that makes sense or a user asks for it!

sbrl:
Hey! Next time, perhaps you could avoid force-pushing please? It makes it harder to tell what's changed when reviewing, and we've got a special button that allows us to squash easily on merge :smiley_cat:

nibalizer:
@sbrl hey sorry - I won't force push next time. Thanks!

PR #3908, opened 2020-03-11
sed: minor syntax fix

cforno12:

Comments:
sed takes in the second argument as a filename. So it read 's/{{find}}/{{replace}}/g' as a filename.


pages/osx/sed.md
```diff
@@ -12,7 +12,7 @@
 
 - Replace all occurrences of a string in a file, overwriting the file (i.e. in-place):
 
-`sed -i '' 's/{{find}}/{{replace}}/g' {{filename}}`
+`sed -i 's/{{find}}/{{replace}}/g' {{filename}}`
```
sbrl: I'm not sure this is correct. From the `sed` manual:
```
-i[SUFFIX], --in-place[=SUFFIX]

              edit files in place (makes backup if SUFFIX supplied)
```

...so the `''` here is specifying a suffix of nothing to overwrite the file.

If anything, we would perhaps want the long-form of the option, like this:

```suggestion
`sed --in-place='' 's/{{find}}/{{replace}}/g' {{filename}}`
```

cforno12:
Please let me know if you want me to clean up and squash these commits.

Issue #3909, opened 2020-03-11
trash: add trash-rm commands


jxu:
Use case: permanently deleting file (removing from trash) 

Right now `tldr trash` lists commands for `trash` and related functions: `trash-list`, `trash-restore`, `trash-empty`, but not `trash-rm`

From trash-rm man page:

```
EXAMPLE
       $ trash-rm foo          # Removes all files with name 'foo' in trash can
       $ trash-rm '*.o'        # Removes all files ending with '.o' in trash can
       $ trash-rm /full/path   # Removes all files removed with exactly '/full/path' as original location
```

schneiderl:
Thanks for requesting this change! Do you want to have a go yourself?

Pages are best maintained by people who understand the command they are about.

If you do, take a quick read of [CONTRIBUTING.md](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md), which guides you through the process of creating or changing a page here. Then, just simply send a pull request to this repository - and if you need any assistance, just ask :-)

jxu:
Ok I went ahead and made a PR.

PR #3910, opened 2020-03-12
trash: add trash-rm examples

jxu:

#3909 

Does not include the more complex and powerful pattern-matching command ex. `trash-rm '*.o'` for simplicity. Let me know if this should be included. We have room for one more example.

### DIFF
--- a/pages/linux/trash.md
+++ b/pages/linux/trash.md
@@ -27,6 +27,6 @@
 
 `trash-rm foo`
 
-- Remove all files with exactly '/full/path' as original location:
+- Remove all files with '/full/path' as original location:
 
 `trash-rm /full/path`
--- a/pages/linux/trash.md
+++ b/pages/linux/trash.md
@@ -23,7 +23,7 @@
 
 `trash-empty {{10}}`
 
-- Remove all files named 'foo' from trash:
+- Remove all files named 'foo' from the trash:
 
 `trash-rm foo`
 
--- a/pages/linux/trash.md
+++ b/pages/linux/trash.md
@@ -22,3 +22,11 @@
 - Empty trash, keeping files trashed less than {{10}} days ago:
 
 `trash-empty {{10}}`
+
+- Remove all files matching name 'foo' from trash:
+
+`trash-rm foo`
+
+- Remove all files with exactly '/full/path' as original location:
+
+`trash-rm /full/path`
--- a/pages/linux/trash.md
+++ b/pages/linux/trash.md
@@ -27,6 +27,6 @@
 
 `trash-rm foo`
 
-- Remove all files with '/full/path' as original location:
+- Remove all files with a given original location:
 
 `trash-rm /full/path`
--- a/pages/linux/trash.md
+++ b/pages/linux/trash.md
@@ -23,7 +23,7 @@
 
 `trash-empty {{10}}`
 
-- Remove all files matching name 'foo' from trash:
+- Remove all files named 'foo' from trash:
 
 `trash-rm foo`
 


pages/linux/trash.md
```diff
@@ -22,3 +22,11 @@
 - Empty trash, keeping files trashed less than {{10}} days ago:
 
 `trash-empty {{10}}`
+
+- Remove all files matching name 'foo' from trash:
```
schneiderl: In this case stuff like "barfoobar" would also be removed from trash? Or only "foo" exactly?

jxu: Looks like exact name match from the source, since matching is done with fnmatch, unix wildcard style. I tested it to confirm.

https://github.com/andreafrancia/trash-cli/blob/5abecd53e1d84f2a5fd3fc60d2f5d71e518826c5/trashcli/rm.py#L84

jxu: I'm not sure if it works for folders.

pages/linux/trash.md
```diff
@@ -22,3 +22,11 @@
...
+
+`trash-rm foo`
+
+- Remove all files with exactly '/full/path' as original location:
```
zdroid: `exactly` is not really needed here

pages/linux/trash.md
zdroid:
```suggestion
- Remove all files named 'foo' from trash:
```

pages/linux/trash.md
jxu: ok. I used the language from manpage examples

pages/linux/trash.md
```diff
@@ -22,3 +22,11 @@
 - Empty trash, keeping files trashed less than {{10}} days ago:
 
 `trash-empty {{10}}`
+
+- Remove all files named 'foo' from trash:
```
sbrl: Perhaps this will read better?
```suggestion
- Remove all files named 'foo' from the trash:
```

pages/linux/trash.md
```diff
@@ -22,3 +22,11 @@
...
+
+`trash-rm foo`
+
+- Remove all files with '/full/path' as original location:
```
sbrl: Perhaps this will read better?
```suggestion
- Remove all files with a given original location:
```

jxu:
@sbrl this is a good question. `trash-cli` installs all the little scripts. `trash` is the main program which does the same as `trash-put`. However the `trash` manpages are pretty sparse and actually `man trash` doesn't work (you must use `man trash-put`). So for now I think the `trash` commands are simple enough that they can all be put into one tldr page, but I have no objections to making them separate pages.

sbrl:
Ah, I see @jxu. Let's keep them all on the same page for now then.

Once my other comments are addressed, I can approve it 🙂

(writing this from the brand new GitHub mobile app, so if you've made them already I apologise for missing them)

jxu:
Ok. There are a lot of commits so it may be worth squashing them. However I do not know if that messes up the history on this PR.

owenvoke:
@jxu, no need to squash them. 👍 We can do that when we merge the PR. 🙂

sbrl:
Merging, since @zdroid's comments appear to have been addressed.

pages/linux/trash.md
```diff
@@ -22,3 +22,11 @@
...
 
 `trash-empty {{10}}`
+
+- Remove all files named 'foo' from the trash:
+
+`trash-rm foo`
+
+- Remove all files with a given original location:
+
+`trash-rm /full/path`
```
agnivade, line 32: Shouldn't this be in a token (`{{}}`) ?

sbrl: How did I manage to miss this? Opening another PR.

PR #3911, opened 2020-03-12
git-update-index: add page

yukuku:

### DIFF
--- a/pages/common/git-update-index.md
+++ b/pages/common/git-update-index.md
@@ -3,6 +3,6 @@
 > Git command for manipulating the index.
 > More information: <https://git-scm.com/docs/git-update-index>.
 
-- Pretend that a modified file is unchanged. (`git status` will not show this as changed.):
+- Pretend that a modified file is unchanged (`git status` will not show this as changed):
 
 `git update-index --skip-worktree {{modified_file}}`
--- a/pages/common/git-update-index.md
+++ b/pages/common/git-update-index.md
@@ -3,6 +3,6 @@
 > Git command for manipulating the index.
 > More information: <https://git-scm.com/docs/git-update-index>.
 
-- Pretend that a modified file is unchanged. (`git status` will not show this as changed.)
+- Pretend that a modified file is unchanged. (`git status` will not show this as changed.):
 
 `git update-index --skip-worktree {{modified_file}}`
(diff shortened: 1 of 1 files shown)


pages/common/git-update-index.md
```diff
@@ -3,6 +3,6 @@
 > Git command for manipulating the index.
 > More information: <https://git-scm.com/docs/git-update-index>.
 
-- Pretend that a modified file is unchanged. (`git status` will not show this as changed.)
+- Pretend that a modified file is unchanged. (`git status` will not show this as changed.):
```
zdroid:
```suggestion
- Pretend that a modified file is unchanged (`git status` will not show this as changed):
```

yukuku:
I've used this several times when dealing with repositories that I don't regularly maintain but has `.gitignore` that does not exclude temporary/build files that are produced on my machine. 

e.g.

1. A repository does not exclude the `out/` directory but when I build the project, the intermediate files are stored by default in `out/` directory. I'm not familiar enough with this project so I don't want to add `out/` to `.gitignore` an push that.

2. A repository has an instruction to patch certain lines in order for a hack to work, or in order to enable debug mode, or to increase network connection time out. The change modifies the source code but only for local developer's machine, it is not supposed to be commited.

pages/common/git-update-index.md
```diff
@@ -0,0 +1,8 @@
+# git update-index
+
+> Git command for manipulating the index.
+> More information: <https://git-scm.com/docs/git-update-index>.
+
+- Pretend that a modified file is unchanged (`git status` will not show this as changed):
+
+`git update-index --skip-worktree {{modified_file}}`
```
einverne:
```suggestion
`git update-index --skip-worktree {{path/to/modified_file}}`
```

Keep consistent with other pages.

sbrl:
Wow, that's such a very cool use-case @yukuku - thanks so much! I'll remember that one - it would have made a number of situations much easier :D

waldyrious:
Hey @yukuku and @sbrl — just thought I'd mention that for the use case of a missing `.gitignore` in a project, you can easily set a repo-specific gitignore file by editing the `.git/info/exclude` file within the repo. That's more flexible than manually updating the index. Check the [official docs](https://git-scm.com/docs/gitignore) and [this StackOverflow Q&A](https://stackoverflow.com/questions/22906851/when-would-you-use-git-info-exclude-instead-of-gitignore-to-exclude-files) for more info.

PR #3913, opened 2020-03-13
lpr: add page

rightbrace:
- [*] The page (if new), does not already exist in the repo.
- [*] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [*] The page has 8 or fewer examples.
- [*] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [*] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).

### DIFF
--- a/pages/common/lpr.md
+++ b/pages/common/lpr.md
@@ -1,6 +1,8 @@
 # lpr
 
 > CUPS tool for printing files.
+> See also `lpstat` and `lpadmin` for listing and configuring printers.
+> More information: <http://www.CUPS.org>.
 
 - Print a file to the default printer:
 
--- a/pages/common/lpr.md
+++ b/pages/common/lpr.md
@@ -6,9 +6,9 @@
 
 `lpr {{path/to/file}}`
 
-- Print a certain number of copies:
+- Print 2 copies:
 
-`lpr -# {{copies}} {{path/to/file}}`
+`lpr -# {{2}} {{path/to/file}}`
 
 - Print to a named printer:
 
@@ -16,16 +16,16 @@
 
 - Print double sided in portrait:
 
-`lpr -o sides=two-sided-long-edge {{path/to/file}}`
+`lpr -o sides={{two-sided-long-edge}} {{path/to/file}}`
 
 - Print double sided in landscape:
 
-`lpr -o sides=two-sided-short-edge {{path/to/file}}`
+`lpr -o sides={{two-sided-short-edge}} {{path/to/file}}`
 
-- Set page size (a4, letter, legal, etc):
+- Set page size (more options may be available depending on setup):
 
-`lpr -o media={{paper_type}} {{path/to/file}}`
+`lpr -o media={{a4|letter|legal}} {{path/to/file}}`
 
-- Print multiple pages per sheet (2, 4, 6, 9, or 16):
+- Print multiple pages per sheet:
 
-`lpr -o number-up={{pages_per_sheet}} {{path/to/file}}`
+`lpr -o number-up={{2|4|6|9|16}} {{path/to/file}}`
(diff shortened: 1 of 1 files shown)


pages/common/lpr.md
```diff
@@ -0,0 +1,31 @@
+# lpr
+
+> CUPS tool for printing files.
+
+- Print a file to the default printer:
+
+`lpr {{path/to/file}}`
+
+- Print a certain number of copies:
```
agnivade: Let's use a concrete number here like 2

pages/common/lpr.md
```diff
@@ -0,0 +1,31 @@
...
+
+`lpr -o sides=two-sided-long-edge {{path/to/file}}`
+
+- Print double sided in landscape:
+
+`lpr -o sides=two-sided-short-edge {{path/to/file}}`
+
+- Set page size (a4, letter, legal, etc):
+
+`lpr -o media={{paper_type}} {{path/to/file}}`
```
agnivade: Instead of paper_type, we can do a4|letter|legal. And change the description to "Set page size"

pages/common/lpr.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Print multiple pages per sheet (2, 4, 6, 9, or 16):
```
agnivade: Are these the only options ? Then we can do the same `|` operator thing mentioned above.

pages/common/lpr.md
```diff
@@ -0,0 +1,31 @@
...
+
+`lpr -# {{copies}} {{path/to/file}}`
+
+- Print to a named printer:
+
+`lpr -P {{printer}} {{path/to/file}}`
+
+- Print double sided in portrait:
+
+`lpr -o sides=two-sided-long-edge {{path/to/file}}`
```
agnivade: two-sided-long-edge in tokens (`{{}}`)

pages/common/lpr.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Print double sided in landscape:
+
+`lpr -o sides=two-sided-short-edge {{path/to/file}}`
```
agnivade: Same as above

pages/common/lpr.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Print a file to the default printer:
+
+`lpr {{path/to/file}}`
+
+- Print a certain number of copies:
+
+`lpr -# {{copies}} {{path/to/file}}`
+
+- Print to a named printer:
```
agnivade: How does one list the printers ?

rightbrace: It relies on a different tool, so I am not sure if it is appropriate to include here? `lpstat -a` will do it, and is from the same set of tools, but not the same command.

schneiderl: We can always include a "see also" which points to lpstat. Something like [this](https://github.com/tldr-pages/tldr/blob/75d109448a21c8bd82e6cc1f2912d8b0d2a8d631/pages/linux/sinfo.md).

rightbrace: Added, along with lpadmin (as it seems to be of equivalent pertinence). Also, a link to cups.org for completeness, in case just writing CUPS isn't enough

pages/common/lpr.md
sbrl: I think there are a large number of different paper sizes, so `{{a4|letter|legal|...}}` might be a better choice.

einverne:
It seems Travis CI is dead. How can I trigger it manually?

agnivade:
There seems to be something funny with this PR. Last commit travis ran on was this - https://github.com/tldr-pages/tldr/commit/804d13b09da7a87afe16aa3f892149015058179a.

@rightbrace - could you make a dummy change, and then revert that change with another commit ? Just to rerun travis.

rightbrace:
@agnivade Done

agnivade:
Awesome, thanks !

PR #3924, opened 2020-03-19
build-system: Add in auditjs to scan the dependencies in node_modules

timlevett:
Hi there,

Not your typical PR, but thought it could be helpful. We thought it would be helpful to scan your node project to make sure you are not vulnerable. I plugged it into your `npm test`. Welcome feedback, be safe.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -3,6 +3,11 @@
 # This script is executed by Travis CI when a PR is merged (i.e. in the `deploy` step).
 set -ex
 
+# Scan TLDR for vulnerabilities
+function scan_for_vulns {
+  npx auditjs@latest ossi
+}
+
 function initialize {
   if [ -z "$TLDRHOME" ]; then
     export TLDRHOME=${TRAVIS_BUILD_DIR:-$(pwd)}
@@ -43,4 +48,5 @@ function upload_assets {
 ###################################
 
 initialize
+scan_for_vulns
 upload_assets
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -39,6 +39,12 @@ function run_checks_pr {
   fi
 }
 
+# Scan TLDR for vulnerabilities
+function scan_for_vulns {
+  npm install -g auditjs
+  auditjs ossi
+}
+
 ###################################
 # MAIN
 ###################################
@@ -50,5 +56,7 @@ else
   set -e
   run_tests
 fi
+
+scan_for_vulns
 
 echo 'Test ran succesfully!'
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -41,8 +41,7 @@ function run_checks_pr {
 
 # Scan TLDR for vulnerabilities
 function scan_for_vulns {
-  npm install -g auditjs
-  auditjs ossi
+  npx auditjs@latest ossi
 }
 
 ###################################
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -39,11 +39,6 @@ function run_checks_pr {
   fi
 }
 
-# Scan TLDR for vulnerabilities
-function scan_for_vulns {
-  npx auditjs@latest ossi
-}
-
 ###################################
 # MAIN
 ###################################
@@ -55,7 +50,5 @@ else
   set -e
   run_tests
 fi
-
-scan_for_vulns
 
 echo 'Test ran succesfully!'
--- a/package.json
+++ b/package.json
@@ -6,9 +6,9 @@
   "homepage": "https://tldr.sh/",
   "dependencies": {
     "glob": "7.1.3",
+    "husky": "^2.2.0",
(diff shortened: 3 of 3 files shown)


scripts/test.sh
```diff
@@ -39,6 +39,12 @@ function run_checks_pr {
   fi
 }
 
+# Scan TLDR for vulnerabilities
+function scan_for_vulns {
+  npm install -g auditjs
```
DarthHater: Instead of installing globally, you could just do `npx auditjs@latest ossi` and it will effectively do the same thing.

timlevett: good idea, i'll change that @DarthHater

timlevett: Done in https://github.com/tldr-pages/tldr/pull/3924/commits/36a864a1e2670069f425888e13c805bcac4e3c00

DarthHater:
@sbrl , it will exit non zero if it finds any issues, for sure!

If you can't clean an issue up you can whitelist an issue: https://github.com/sonatype-nexus-community/auditjs#whitelisting

Also if you strictly want to run this informationally you can do `npx auditjs@latest ossi || true`, which will ignore the exit of non zero

timlevett:
> Hey there, @timlevett - thanks for the PR!
> 
> Does this make Travis fail if vulnerabilities are detected? If so, this could negatively impact merging PRs that just add pages and don't touch any node stuff.

Hi @sbrl, it most likely will. I just tested locally by adding in `npm install lodash@4.17.11` and running a commit (which triggers the post commit hook) and it did fail. That is a good point about triggering when a new vuln is detected to individuals that are just adding some markdown. What do you suggest? I think it is good to keep tabs on these things, but also don't want to be a negative impact on being able to merge community PRs. One idea I had was we could change it to only run on master if you wanted. Anything else you can think of?

timlevett:
>Also if you strictly want to run this informationally you can do npx auditjs@latest ossi || true, which will ignore the exit of non zero

Good work around @DarthHater !

sbrl:
Ah, I see. Thanks for the clarification @timlevett. Perhaps only running on `master` might bee a better approach, yeah. Then I guess we could still merge PRs, but get the notification if it fails?

/cc @pxgamer, @agnivade, @waldyrious

agnivade:
Running only on master sounds good.

timlevett:
Thank you for the feedback @sbrl and @agnivade! I've moved the check from the "test" script to the "deploy" script since deploy only runs on merge to master. One thing that wasn't clear in your feedback is do you want the deploy to fail if it does find something? Right now it would, but its easy to make it just informative (as described above by @DarthHater).

sbrl:
Hrm. There is a good case for having the deploy fail on `master` if there are linter errors I think. Let's wait to see what @agnivade and others think.

agnivade:
If we make it passive, how would we be notified ? I think it would be nice to not have the deploy fail. But we also want to be notified in a clear manner.

DarthHater:
How would you like to be notified @agnivade ? Asking because I work on the project, and I might be able to help answer some questions, offer suggestions. 

At a certain point, having a build fail before it deploys because it found a vulnerability in a third party library is a best practice. It forces a review of the actual vulnerability (which can be ignored via whitelisting), but it requires someone to essentially sign off and go "Yep, it's ok, we can live with this".

That said, it requires changing your process at least a bit, to shift security left.

agnivade:
> How would you like to be notified @agnivade ? 

A comment on a Github issue, or perhaps an auto-creation of a GitHub issue stating that a vulnerability has been found would be ideal.

Yes, it's good to have the build fail. But while introducing a new process, I would want to start off with being passive rather than being active, and see how it goes. If the build fails, and there's nobody around to handle it, important updates might delay getting pushed.

DarthHater:
@agnivade for that, check out: https://github.com/apps/sonatype-depshield

It'll be a lil bit noisier than AuditJS, but it'll do exactly what you are describing. It also I believe is a bit better at reducing false positives than AuditJS might be (that's the word the bird told me once).

agnivade:
Awesome, that looks great. On a different note though, I am wondering the value add of adding something like this in this repo. The "build" for this repo is just a bunch of markdown files. But it maybe much more useful in the node client repo where we actually ship code.

In any case, this seems very helpful and we will take a look at it. Thanks !

scripts/deploy.sh
```diff
@@ -3,6 +3,11 @@
 # This script is executed by Travis CI when a PR is merged (i.e. in the `deploy` step).
 set -ex
 
+# Scan TLDR for vulnerabilities
```
zdroid, line 6:
```suggestion
# Scan tldr for vulnerabilities
```
Just a minor suggestion

agnivade:
From the conversation above, I believe we won't be merging this PR and investigate something like  https://github.com/apps/sonatype-depshield instead for our purposes.

PR #3925, opened 2020-03-22
zoxide: add page

robcowie:

### DIFF
--- a/pages/common/zoxide.md
+++ b/pages/common/zoxide.md
@@ -26,4 +26,4 @@
 
 - Generate shell configuration for command aliases (`z`, `za`, `zi`, `zq`, `zr`):
 
-`zoxide init {{shell}}`
+`zoxide init {{bash|fish|zsh}}`
--- a/pages/common/zoxide.md
+++ b/pages/common/zoxide.md
@@ -1,6 +1,7 @@
 # zoxide
 
-> Keep track of the directories you use most frequently, and use a ranking algorithm to navigate to the best match.
+> Keep track of the most frequently used directories.
+> Uses a ranking algorithm to navigate to the best match.
 > More information: <https://github.com/ajeetdsouza/zoxide>.
 
 - Go to the highest-ranked directory that contains "foo" in the name:
--- a/pages/common/zoxide.md
+++ b/pages/common/zoxide.md
@@ -11,7 +11,7 @@
 
 `zoxide query {{foo}} {{bar}}`
 
-- Interactive directory search (requires [fzf](https://github.com/junegunn/fzf)):
+- Start an interactive directory search (requires `fzf`):
 
 `zoxide query --interactive`
 
--- a/pages/common/zoxide.md
+++ b/pages/common/zoxide.md
@@ -1,0 +1,28 @@
+# zoxide
+
+> Keep track of the directories you use most frequently, and use a ranking algorithm to navigate to the best match.
+> More information: <https://github.com/ajeetdsouza/zoxide>.
+
+- Go to the highest-ranked directory that contains "foo" in the name:
+
+`zoxide query {{foo}}`
+
+- Go to the highest-ranked directory that contains "foo" and then "bar":
+
+`zoxide query {{foo}} {{bar}}`
+
+- Interactive directory search (requires [fzf](https://github.com/junegunn/fzf)):
+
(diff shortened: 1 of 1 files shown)


pages/common/zoxide.md
```diff
@@ -0,0 +1,28 @@
+# zoxide
+
+> Keep track of the directories you use most frequently, and use a ranking algorithm to navigate to the best match.
```
sbrl: Perhaps this will read better?
```suggestion
> Keep track of the most frequently used directories.
> Uses a ranking algorithm to navigate to the best match.
```

pages/common/zoxide.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Go to the highest-ranked directory that contains "foo" in the name:
+
+`zoxide query {{foo}}`
+
+- Go to the highest-ranked directory that contains "foo" and then "bar":
+
+`zoxide query {{foo}} {{bar}}`
+
+- Interactive directory search (requires [fzf](https://github.com/junegunn/fzf)):
```
sbrl: Perhaps we could tweak this to make it more consistent with other pages? We already have a page for `fzf`, so we can reference it here.
```suggestion
- Start an interactive directory search (requires `fzf`):
```

pages/common/zoxide.md
```diff
@@ -0,0 +1,28 @@
...
+
+`zoxide query --interactive`
+
+- Add a directory or increment it's rank:
```
sbrl: Perhaps this will read better?
```suggestion
- Add a directory or increment its rank:
```

`it's` ->  `it is`. `its` doesn't have an apostrophe if it's possessive :-)

pages/common/zoxide.md
```diff
@@ -0,0 +1,28 @@
...
+
+`zoxide add {{path/to/dir}}`
+
+- Remove a directory from `zoxide`'s database:
+
+`zoxide remove {{path/to/dir}}`
+
+- Generate shell configuration for command aliases (`z`, `za`, `zi`, `zq`, `zr`):
+
+`zoxide init {{shell}}`
```
sbrl: What about this?
```suggestion
`zoxide init {{bash|zsh}}`
```

We can also include other shells here too, if they are supported.

pages/common/zoxide.md
robcowie: Absolutely correct, very lazy of me!

pages/common/zoxide.md
robcowie: Done

robcowie:
Comments addressed, thanks.

pages/common/zoxide.md
```diff
@@ -0,0 +1,29 @@
...
+
+> Keep track of the most frequently used directories.
+> Uses a ranking algorithm to navigate to the best match.
+> More information: <https://github.com/ajeetdsouza/zoxide>.
+
+- Go to the highest-ranked directory that contains "foo" in the name:
+
+`zoxide query {{foo}}`
+
+- Go to the highest-ranked directory that contains "foo" and then "bar":
```
zdroid: Does it have to be in that order?

pages/common/zoxide.md
```diff
@@ -0,0 +1,29 @@
+# zoxide
+
+> Keep track of the most frequently used directories.
+> Uses a ranking algorithm to navigate to the best match.
```
einverne:
```suggestion
> Use a ranking algorithm to navigate to the best match.
```

I'm not sure weather should we use `s` here. I remember I read some guideline before. But now I cannot found out.
@zdroid @sbrl  What do you think here?

pages/common/zoxide.md
robcowie: After a little experimentation it turns out yes, each of those terms must occur in the path _in that order_.

pages/common/zoxide.md
robcowie: My choice would be for `Uses a`. The alternative imperative sentence reads too much like a command to the user, rather than a description of the tool.

zdroid: `Uses`, since it's talking about what program itself does, not what you do with the program.

einverne: It is okay then. I mark this as resolved.

PR #3930, opened 2020-03-24
ip address: add page

sbrl:
We've needed subpages for the `ip` command for ages in my opinion, so with the current IRL situation affecting the world I thought I'd use this as a welcome distraction.

As it happens the man page has some quite helpful examples at the bottom - this page is a curation and adjustment thereof.

By doing `man ip` you can see a list of all the subcommands if you're interested.

Edit: it's probably worth knowing that `ip address` is actually really powerful - it's just not possible to demonstrate all the crazy stuff it lets you do in a single tldr page :P

### DIFF
--- a/pages/linux/ip-address.md
+++ b/pages/linux/ip-address.md
@@ -1,0 +1,27 @@
+# ip address
+
+> IP Address management subcommand.
+
+- List network interfaces and their associated IP addresses:
+
+`ip address`
+
+- Filter to show only active network interfaces:
+
+`ip address show up`
+
+- Display information about a specific network interface:
+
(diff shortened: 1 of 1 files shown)


pages/linux/ip-address.md
```diff
@@ -0,0 +1,27 @@
...
+
+`ip address`
+
+- Filter to show only active network interfaces:
+
+`ip address show up`
+
+- Display information about a specific network interface:
+
+`ip address show dev {{network_interface}}`
```
agnivade, line 15: might be good to mention a concrete network interface like eth0.

PR #3948, opened 2020-03-28
pidof: add page

qhyun2:

### DIFF
--- a/pages/linux/pidof.md
+++ b/pages/linux/pidof.md
@@ -1,0 +1,19 @@
+# pidof
+
+> Gets the id of a process using it's name.
+
+- List all process ids with given name:
+
+`pidof {{bash}}`
+
+- List a single process id with given name:
+
+`pidof -s {{bash}}`
+
+- List process ids including scripts with given name:
+
+`pidof -x {{script.py}}`
+
+- Kill all processes with given name:
+
+``kill `pidof {{name}}` ``
--- a/pages/linux/pidof.md
+++ b/pages/linux/pidof.md
@@ -16,4 +16,4 @@
 
 - Kill all processes with given name:
 
-``kill `pidof {{name}}` ``
+`kill "$(pidof {{name}})" `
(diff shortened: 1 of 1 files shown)


pages/linux/pidof.md
```diff
@@ -0,0 +1,19 @@
+# pidof
+
+> Gets the id of a process using it's name.
```
zdroid:
```suggestion
> Gets the ID of a process using its name.
```

id should be changed to ID in all other instances, too.

qhyun2: Is there a rule against using apostrophes? I think my grammar is correct. Also how should I handle having back ticks ( ` ) inside a code block?

zdroid: About back ticks, I am not sure. /cc @agnivade @pxgamer

Here it is `its` because that word represents a pronoun. You can search for `its vs it's` on the Internet to find when to use each of them.

sbrl: `it's`/`its` is a special case. While one usually includes an apostophe to indicate possessiveness, you in fact omit it with `its`. The apostrophe in this specific instance indicates a contraction of `it is`.

Also, `id` → `ID` doesn't feel required to me. It could be either - it doesn't really matter.

pages/linux/pidof.md
```diff
@@ -0,0 +1,19 @@
...
+
+`pidof -s {{bash}}`
+
+- List process ids including scripts with given name:
+
+`pidof -x {{script.py}}`
+
+- Kill all processes with given name:
+
+``kill `pidof {{name}}` ``
```
sbrl: The single backtick syntax here isn't really good practice on the command line any more. Instead, use the `$(command)` syntax instead:
```suggestion
`kill "$(pidof {{name}})" `
```

pages/linux/pidof.md
zdroid: It's a norm to write `ID` uppercase, as it's an acronym. `id` is a term from psychoanalysis and is unrelated to tech.

sbrl:
Thanks, @qhyun2 :smiley_cat:

PR #3952, opened 2020-03-30
transcrypt: add page

rlemasquerier:

### DIFF
--- a/pages/common/transcrypt.md
+++ b/pages/common/transcrypt.md
@@ -1,7 +1,7 @@
 # transcrypt
 
-> Transparently encrypt files within a git repository
-> More information: <https://github.com/elasticdog/transcrypt>
+> Transparently encrypt files within a git repository.
+> More information: <https://github.com/elasticdog/transcrypt>.
 
 - Initialize an unconfigured repository:
 
--- a/pages/common/transcrypt.md
+++ b/pages/common/transcrypt.md
@@ -1,0 +1,24 @@
+# transcrypt
+
+> Transparently encrypt files within a git repository
+> More information: <https://github.com/elasticdog/transcrypt>
+
+- Initialize an unconfigured repository:
+
+`transcrypt`
+
+- List the currently encrypted files:
+
+`git ls-crypt`
+
+- Display the credentials of a configured repository:
+
+`transcrypt --display`
+
+- Initialize and decrypt a fresh clone of a configured repository:
+
+`transcrypt -c <cipher> -p '<password>'`
+
+- Rekey to change the encryption cipher or password:
+
+`transcrypt --rekey`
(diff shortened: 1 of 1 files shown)


pages/common/transcrypt.md
```diff
@@ -0,0 +1,24 @@
...
+
+`git ls-crypt`
+
+- Display the credentials of a configured repository:
+
+`transcrypt --display`
+
+- Initialize and decrypt a fresh clone of a configured repository:
+
+`transcrypt -c <cipher> -p '<password>'`
```
sbrl: Are there any long forms of these `-c` and `-p` arguments? We prefer long-form arguments in tldr pages.

Also, is there a way to get it to prompt for a password or use a password file instead? Providing the password directly on the command line is rather insecure, since it will be stored in the shell history.

rlemasquerier: @sbrl 
- Yes, there are long forms for those :) 
- And also yes: users can simply use `transcrypt` to enter the cipher and the password via the prompt (it is the exact same command as the one to initially setup transcrypt actually! ).

Should I remove this command from the page, and better explain that the first one can also be used to decrypt a repo ?

Thank you for your review !

sbrl: - Cool, If we could change the examples to use the long forms of the arguments instead, that'd be great!
 - Right. Perhaps we could keep specifying the cipher here and just remove the password bit?

It seems odd that it requires you to explicitly specify the cipher used for encryption. I'd have thought that it would store that information somewhere.

Many thanks for the new page!

zdroid: I agree with removing the password part if it can be done with prompt instead.

rlemasquerier: @sbrl Changes applied !
> It seems odd that it requires you to explicitly specify the cipher used for encryption. I'd have thought that it would store that information somewhere.

True... It seems like it is considered as part of the "credentials"

pages/common/transcrypt.md
```diff
@@ -0,0 +1,24 @@
...
+
+`git ls-crypt`
+
+- Display the credentials of a configured repository:
+
+`transcrypt --display`
+
+- Initialize and decrypt a fresh clone of a configured repository:
+
+`transcrypt --cipher=<cipher>`
```
sbrl, line 20: Perhaps we could correct this to use the `{{token_syntax}}`?
```suggestion
`transcrypt --cipher={{cipher}}`
```

rlemasquerier: Definitely yes, commiting your suggestion :)

rlemasquerier:
> Just another comment

Committed your suggestion :)

PR #3956, opened 2020-03-31
evil-winrm: add page

dhondta:

### DIFF
--- a/pages/common/evil-winrm.md
+++ b/pages/common/evil-winrm.md
@@ -1,0 +1,38 @@
+# evil-winrm
+
+> Windows Remote Management (WinRM) shell for pentesting.
+> More information: <https://github.com/Hackplayers/evil-winrm>.
+
+- Connect to a victim host:
+
+`evil-winrm --ip {{ip}} --user {{user}} --password {{password}}`
+
+- Connect to a victim host, passing the password hash:
+
+`evil-winrm --ip {{ip}} --user {{user}} --hash {{nt_hash}}`
+
+- Connect to a victim host, specifying folders for scripts and executables:
+
+`evil-winrm --ip {{ip}} --user {{user}} --password {{password}} --scripts {{path/to/scripts}} --executables {{path/to/executables}}`
+
+- Connect to a victim host, using SSL:
+
+`evil-winrm --ip {{ip}} --user {{user}} --password {{password}} --ssl --pub-key {{path/to/pubkey}} --priv-key {{path/to/privkey}}`
+
+> Once connected, we get a PowerShell prompt on the victim host.
+
+- Upload a file to the victim machine:
+
+`*Evil-WinRM* PS C:\Users\user\Documents> upload {{path/to/local/file}} {{path/to/remote/file}}`
+
+- Get the list of loaded PowerShell functions:
+
+`*Evil-WinRM* PS C:\Users\user\Documents> menu`
+
+- Load a PowerShell script from the `--scripts` folder:
+
+`*Evil-WinRM* PS C:\Users\user\Documents> {{script.ps1}}`
+
+- Invoke a binary on the victim machine from the `--executables` folder:
+
+`*Evil-WinRM* PS C:\Users\user\Documents> Invoke-Binary {{binary.exe}}`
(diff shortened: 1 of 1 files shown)


pages/common/evil-winrm.md
```diff
@@ -0,0 +1,38 @@
...
+
+- Connect to a victim host, specifying folders for scripts and executables:
+
+`evil-winrm --ip {{ip}} --user {{user}} --password {{password}} --scripts {{path/to/scripts}} --executables {{path/to/executables}}`
+
+- Connect to a victim host, using SSL:
+
+`evil-winrm --ip {{ip}} --user {{user}} --password {{password}} --ssl --pub-key {{path/to/pubkey}} --priv-key {{path/to/privkey}}`
+
+> Once connected, we get a PowerShell prompt on the victim host.
```
zdroid: This is problematic. Why not put it in the top description as well?

Also, I would suggest to change `victim host` to something else in the whole document. Perhaps just `host`.

zdroid:
It's supposed to be before the *More information* label. I edited it myself. Well done otherwise. :+1:

pages/common/evil-winrm.md
sbrl: Not sure I understand what you're getting at here @zdroid.

pages/common/evil-winrm.md
```diff
@@ -0,0 +1,37 @@
...
+
+`evil-winrm --ip {{ip}} --user {{user}} --password {{password}} --scripts {{path/to/scripts}} --executables {{path/to/executables}}`
+
+- Connect to a host, using SSL:
+
+`evil-winrm --ip {{ip}} --user {{user}} --password {{password}} --ssl --pub-key {{path/to/pubkey}} --priv-key {{path/to/privkey}}`
+
+- Upload a file to the host:
+
+`*Evil-WinRM* PS C:\Users\user\Documents> upload {{path/to/local/file}} {{path/to/remote/file}}`
```
sbrl, line 25: Is there any reason why this is different to the previous example? If this is a shell command for `evil-winrm`, perhaps we can represent that more concisely, like this?
```suggestion
`PS > upload {{path/to/local/file}} {{path/to/remote/file}}`
```

Same goes  for the other examples if I'm understanding this correctly.

pages/common/evil-winrm.md
```diff
@@ -0,0 +1,37 @@
...
+
+- Get the list of loaded PowerShell functions:
```
sbrl, line 27: Perhaps this will read better?
```suggestion
- Get a list of loaded PowerShell functions:
```

pages/common/evil-winrm.md
dhondta: That's up to you. My idea is to show the prompt like the user will see when she gets a shell. I've put the Windows path as well as it is where the user lands when Evil-WinRM can connect to the target.
Also, putting `PS > ` confuses with a normal PowerShell prompt, no ? What do you think about this ?

sbrl: Ah, I see. The prompt is a bit long though. What about this?
```suggestion
`*Evil-WinRM* PS > upload {{path/to/local/file}} {{path/to/remote/file}}`
```

sbrl:
Hey! Next time, perhaps you could avoid force-pushing please? It makes it harder to tell what's changed when reviewing, and we've got a special button that allows us to squash easily on merge :smiley_cat:

pages/common/evil-winrm.md
```diff
@@ -0,0 +1,37 @@
...
+
+- Connect to a host:
+
+`evil-winrm --ip {{ip}} --user {{user}} --password {{password}}`
+
+- Connect to a host, passing the password hash:
+
+`evil-winrm --ip {{ip}} --user {{user}} --hash {{nt_hash}}`
+
+- Connect to a host, specifying folders for scripts and executables:
```
agnivade, line 15: specifying _directories_

sbrl: Ah, I didn't even notice that.

And good point - @dhondta, perhaps you could open another PR please?

dhondta:
> This seems like a windows only tool ? If so, it should be in the windows directory.

Very good remark, but you can see it on the [repo](https://github.com/Hackplayers/evil-winrm) ; it is a Python script supported on both Linux and Windows (see *Features* of the README).

PR #3964, opened 2020-04-04
john: add page

dhondta:

### DIFF
--- a/pages/common/john.md
+++ b/pages/common/john.md
@@ -1,0 +1,36 @@
+# john
+
+> Password cracker.
+> More information: <https://www.openwall.com/john/>.
+
+- Crack password hashes:
+
+`john {{path/to/hashes.txt}}`
+
+- Show passwords cracked:
+
+`john --show {{path/to/hashes.txt}}`
+
+- Display user's cracked password by user identifier from multiple files:
+
+`john --show --users={{0}} {{path/to/hashes*}} {{path/to/other/hashes*}}`
+
+- Crack password hashes, using a custom wordlist:
+
+`john --wordlist={{path/to/wordlist.txt}} {{path/to/hashes.txt}}`
+
+- Crack password hashes, using a specific hash format:
+
+`john --format={{md5crypt|...}} {{path/to/hashes.txt}}`
+
+- List available hash formats:
+
+`john --list=formats`
+
+- Crack password hashes, enabling word mangling rules:
+
+`john --rules {{path/to/hashes.txt}}`
+
+- Restore an interrupted cracking session:
+
+`john --restore`
--- a/pages/common/john.md
+++ b/pages/common/john.md
@@ -11,26 +11,26 @@
 
 `john --show {{path/to/hashes.txt}}`
 
-- Display user's cracked password by user identifier from multiple files:
+- Display users' cracked passwords by user identifier from multiple files:
 
-`john --show --users={{0}} {{path/to/hashes*}} {{path/to/other/hashes*}}`
+`john --show --users={{user_ids}} {{path/to/hashes*}} {{path/to/other/hashes*}}`
 
 - Crack password hashes, using a custom wordlist:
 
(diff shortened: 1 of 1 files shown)


pages/common/john.md
```diff
@@ -0,0 +1,36 @@
...
+
+`john {{path/to/hashes.txt}}`
+
+- Show passwords cracked:
+
+`john --show {{path/to/hashes.txt}}`
+
+- Display user's cracked password by user identifier from multiple files:
+
+`john --show --users={{0}} {{path/to/hashes*}} {{path/to/other/hashes*}}`
```
zdroid, line 16: What does `0` mean here? Also, I think `*` is redundant after `hashes*`.

dhondta: `0` is the user identifier.

For the path, `*` has its importance ; it emphasizes that the wildcard can be used to designate any file starting with "`hashes`"...

zdroid: I understand it's an identifier, but it's not explained in the example's description. Maybe `user_id` instead of `0` would be better?

If the hashes file can have any name, then I'd suggest to drop the wildcard anyway.

sbrl: By the sounds of things, a wildcard is a very commonly used option there, so perhaps keeping it would be a good idea.

I do agree though that `{{user_id}}` instead of `{{0}}` might aid clarity in this situation though.

pages/common/john.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Crack password hashes, using a custom wordlist:
+
+`john --wordlist={{path/to/wordlist.txt}} {{path/to/hashes.txt}}`
+
+- Crack password hashes, using a specific hash format:
+
+`john --format={{md5crypt|...}} {{path/to/hashes.txt}}`
+
+- List available hash formats:
```
sbrl, line 26: Perhaps it would make more logical sense if this example was before the `Crack password hashes, using a specific hash format` example?

pages/common/john.md
```diff
@@ -0,0 +1,36 @@
...
+
+`john --list=formats`
+
+- Crack password hashes, enabling word mangling rules:
+
+`john --rules {{path/to/hashes.txt}}`
+
+- Restore an interrupted cracking session:
```
sbrl, line 34: How does it know where to restore from?

zdroid: Probably has its own local cache/config.

pages/common/john.md
dhondta: OK, I inverted the examples.

pages/common/john.md
dhondta: For this one, I modified the example to something more specific, using a state file.

pages/common/john.md
dhondta: I've let the wildcard and I changed the `{{0}}` to `{{user_ids}}`. Note the `user_idS` as the option is well `--userS`...

pages/common/john.md
```diff
@@ -0,0 +1,36 @@
...
+
+`john --format={{md5crypt|...}} {{path/to/hashes.txt}}`
+
+- Crack password hashes, enabling word mangling rules:
+
+`john --rules {{path/to/hashes.txt}}`
+
+- Restore an interrupted cracking session from a state file, e.g. `mycrack.rec`:
+
+`john --restore={{mycrack}}`
```
agnivade, line 36: If we are specifying the extension in the description, we should use the same in the example.

pages/common/john.md
```diff
@@ -0,0 +1,36 @@
...
+
+`john --wordlist={{path/to/wordlist.txt}} {{path/to/hashes.txt}}`
+
+- List available hash formats:
+
+`john --list=formats`
+
+- Crack password hashes, using a specific hash format:
+
+`john --format={{md5crypt|...}} {{path/to/hashes.txt}}`
```
agnivade, line 28: Let's remove the `...` from here, since we are already showing the available hash formats in the previous command.

PR #3969, opened 2020-04-06
hydra: add page

dhondta:

### DIFF
--- a/pages/common/hydra.md
+++ b/pages/common/hydra.md
@@ -1,6 +1,7 @@
 # hydra
 
 > Online password guessing tool.
+> It supports various protocols: AFP, CVS, FTP, HTTP(S), POP3, RDP, RTSP, SIP, SMB, SMTP, SNMP, SSH, Telnet, VNC, XMPP, ...
 > More information: <https://github.com/vanhauser-thc/thc-hydra>.
 
 - Start Hydra's wizard:
@@ -11,11 +12,11 @@
 
 `hydra -l {{username}} -P {{path/to/wordlist.txt}} {{host_ip}} {{ssh}}`
 
-- Guess Telnet credentials using a list of usernames and a single password, specifying a non-standard port:
+- Guess Telnet credentials using a list of usernames and a single password, specifying a non-standard port and IPv6:
 
-`hydra -L {{path/to/usernames.txt}} -p {{password}} -s {{port}} {{host_ip}} {{telnet}}`
+`hydra -L {{path/to/usernames.txt}} -p {{password}} -s {{port}} -6 {{host_ip}} {{telnet}}`
 
-- Guess FTP credentials using usernames and passwords lists, specifying the number of tasks:
+- Guess FTP credentials using usernames and passwords lists, specifying the number of threads:
 
 `hydra -L {{path/to/usernames.txt}} -P {{path/to/wordlist.txt}} -t {{n_tasks}} {{host_ip}} {{ftp}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/hydra.md
```diff
@@ -0,0 +1,36 @@
...
+> Online password guessing tool.
+> More information: <https://github.com/vanhauser-thc/thc-hydra>.
+
+- Start Hydra's wizard:
+
+`hydra-wizard`
+
+- Guess SSH credentials using a given username and a list of passwords:
+
+`hydra -l {{username}} -P {{path/to/wordlist.txt}} {{host_ip}} {{ssh}}`
```
sbrl, line 12: Are there any long forms of these options?

pages/common/hydra.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Guess Telnet credentials using a list of usernames and a single password, specifying a non-standard port:
+
+`hydra -L {{path/to/usernames.txt}} -p {{password}} -s {{port}} {{host_ip}} {{telnet}}`
+
+- Guess FTP credentials using usernames and passwords lists, specifying the number of tasks:
```
sbrl, line 18: Do you mean concurrent threads here?

pages/common/hydra.md
```diff
@@ -0,0 +1,36 @@
...
+
+`hydra -L {{path/to/usernames.txt}} -P {{path/to/wordlist.txt}} -t {{n_tasks}} {{host_ip}} {{ftp}}`
+
+- Guess MySQL credentials using a username and a passwords list, exiting when a username/password pair is found:
+
+`hydra -l {{username}} -P {{path/to/wordlist.txt}} -f {{host_ip}} {{mysql}}`
+
+- Guess RDP credentials using a username and a passwords list, showing each attempts:
```
sbrl, line 26: Perhaps this will read better?
```suggestion
- Guess RDP credentials using a username and a passwords list, showing each attempt:
```

pages/common/hydra.md
dhondta: No, all the options are in short form.

dhondta:
I made some consequent changes, adding a line at the beginning and demonstrating some other options.
Also, for those who know this tool, there is a much more complex example that would have been interesting with HTTP but it would have been a very long one, hence I ignored it. For this, the user can refer to the documentation (available at the link in *More information*).

pages/common/hydra.md
```diff
@@ -1,6 +1,7 @@
 # hydra
 
 > Online password guessing tool.
+> It supports various protocols: AFP, CVS, FTP, HTTP(S), POP3, RDP, RTSP, SIP, SMB, SMTP, SNMP, SSH, Telnet, VNC, XMPP, ...
```
zdroid, line 4: Maybe we can shorten this list to include only the notable ones, like HTTP, POP3, SMTP, SSH, Telnet?

dhondta: I would let it as is ; I already used this tool multiples times in the past and only realized after reading its full help that it supported so many different protocols. As all of the aforementioned protocols are very common, I would not shorten the list if you agree...

zdroid: It seems good this way too. I understand now that other protocols are usable too, and that may be something that users are looking for.
@sbrl what do you say?

sbrl: What about we say something like this?
```suggestion
> Protocols supported include FTP, HTTP(S), SMTP, SNMP, XMPP, SSH, and more.
```

dhondta: OK, done.

PR #3980, opened 2020-04-10
ssh-keyscan: add known_hosts example

sbrl:
Just discovered this whilst configuring my cluster.

Ref https://www.techrepublic.com/article/how-to-easily-add-an-ssh-fingerprint-to-your-knownhosts-file-in-linux/

### DIFF
--- a/pages/common/ssh-keyscan.md
+++ b/pages/common/ssh-keyscan.md
@@ -16,4 +16,4 @@
 
 - Manually update the ssh known_hosts file with the fingerprint of a given host:
 
-`ssh-keyscan -H {{host}} >> ~/.ssh/known_hosts
+`ssh-keyscan -H {{host}} >> ~/.ssh/known_hosts`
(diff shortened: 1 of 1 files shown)


pages/common/ssh-keyscan.md
```diff
@@ -13,3 +13,7 @@
 - Retrieve certain types of public ssh keys of a remote host:
 
 `ssh-keyscan -t {{rsa,dsa,ecdsa,ed25519}} {{host}}`
+
+- Manually update the ssh known_hosts file with the fingerprint of a given host:
```
zdroid, line 17: Should be `SSH` instead of `ssh` in this example, and in the whole document as well.

sbrl: Perhaps, but I just matched the existing page structure :P

sbrl: Though actually, here we are almost referring to `ssh` the command-line command, but SSH the protocol. In this case, I'd actually argue that it does not make sense to use capital letters.

zdroid: Right,  I missed that.

sbrl: Updated

PR #3986, opened 2020-04-15
mkfs.ext4: add page

MikeLindner:


MikeLindner:
Cleaned up the grammar.  volume-label is the name used in the man page however.

pages/linux/mkfs.ext4.md
```diff
@@ -0,0 +1,11 @@
+# mkfs.ext4
+
+> Creates an ext4 filesystem inside a partition.
+
+- Create an ext4 filesystem inside partition 1 on device b (`sdb1`):
+
+`mkfs.ext4 {{/dev/sdb1}}`
```
sbrl, line 7: I'm pretty sure this command requires root privileges. On other pages like this, we've prefixed the examples with `sudo` to signify this. Perhaps we could do the same here?
```suggestion
`sudo mkfs.ext4 {{/dev/sdb1}}`
```

pages/linux/mkfs.ext4.md
```diff
@@ -0,0 +1,11 @@
...
+
+- Create an ext4 filesystem with a volume-label:
+
+`mkfs.ext4 -L {{volume_label}} {{/dev/sdb1}}`
```
sbrl, line 11: Same here.
```suggestion
`sudo mkfs.ext4 -L {{volume_label}} {{/dev/sdb1}}`
```

MikeLindner: Hi,
Thanks for the feedback!
I agree It definitely needs root access, but it's generally not explicitly stated in doco when a command does - running the command as non root will alert the user to that.
For example it's usually used when building systems, this is done as root so the sudo is redundant.
Hope this clears up my thinking!
Once I've worked through the process of adding one page I'll be adding heaps for you :)  I'm currently writing a book on Linux commands, so will enjoy contributing to TLDR (big fan of cht.sh) while I do.
Thanks!
Michael.
```

michael@chelsea ~> tldr mkfs
mkfs
Build a Linux filesystem on a hard disk partition.
This command is deprecated in favor of filesystem specific mkfs.
<type>
 utils.

 - Build a Linux ext2 filesystem on a partition:
   mkfs {{path/to/partition}}

 - Build a filesystem of a specified type:
   mkfs -t {{ext4}} {{path/to/partition}}

 - Build a filesystem of a specified type and check for bad blocks:
   mkfs -c -t {{ntfs}} {{path/to/partition}}




```

MikeLindner: I think if you asked 90% of Linux sysadmins they'd agree that "if you don't know it needs to be run as root you shouldn't be running it" :D

zdroid: In tldr it's a practice to add `sudo` so less experienced users can use it easily on the first run. :wink:

zdroid: sudo should be added here too.

MikeLindner:
Hey, are we good to go now?

MikeLindner:
Sweet!

pages/linux/mkfs.ext4.md
sbrl: Very true @MikeLindner, but in this instance it's almost always going to require root. Of course, more experienced users should definitely avoid `sudo`ing things just because, but for less experienced users I think it's best in this instance to include `sudo`, as most common use-cases will require it.

PR #3990, opened 2020-04-15
john: apply suggestions

sbrl:
Ref #3964.

/cc @agnivade, @dhondta

### DIFF
--- a/pages/common/john.md
+++ b/pages/common/john.md
@@ -1,7 +1,8 @@
 # john
 
 > Password cracker.
-> More information: <https://www.openwall.com/john/>.
+> More in
+ion: <https://www.openwall.com/john/>.
 
 - Crack password hashes:
 
@@ -25,7 +26,7 @@
 
 - Crack password hashes, using a specific hash format:
 
-`john --format={{md5crypt|...}} {{path/to/hashes.txt}}`
+`john --format={{hash_format}} {{path/to/hashes.txt}}`
 
 - Crack password hashes, enabling word mangling rules:
 
@@ -33,4 +34,4 @@
 
 - Restore an interrupted cracking session from a state file, e.g. `mycrack.rec`:
 
-`john --restore={{mycrack}}`
+`john --restore={{path/to/mycrack.rec}}`
--- a/pages/common/john.md
+++ b/pages/common/john.md
@@ -25,7 +25,7 @@
 
 - Crack password hashes, using a specific hash format:
 
-`john --format={{hash_format}} {{path/to/hashes.txt}}`
+`john --format={{md5crypt}} {{path/to/hashes.txt}}`
 
 - Crack password hashes, enabling word mangling rules:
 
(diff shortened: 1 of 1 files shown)


pages/common/john.md
```diff
@@ -25,12 +26,12 @@
 
 - Crack password hashes, using a specific hash format:
 
-`john --format={{md5crypt|...}} {{path/to/hashes.txt}}`
+`john --format={{hash_format}} {{path/to/hashes.txt}}`
```
zdroid, line 28: What's the reason behind this change?

agnivade: I actually just wanted to remove the `...` and not change the concrete example. My reasoning was that we use the `|` format to show alternative options, but only when they are complete. If all options cannot be specified, then it's simple to just use one option, as it's implicit that it can be replaced by other options.

zdroid: Why not just put `md5crypt` here instead?

agnivade: That's what I am suggesting to do here :) Perhaps the way I put it across might have meant something else.

pages/common/john.md
```diff
@@ -25,12 +25,12 @@
...
```
zdroid, line 28:
```suggestion
`john --format={{md5crypt}} {{path/to/hashes.txt}}`
```

dhondta: @zdroid I think @agnivade  is right ; `hash_format` is more generic and anyway, another example is provided for listing possible hash formats. So, specifying a particular hash format seems confusing, I think.

pages/common/john.md
sbrl: Ah, I see. I must have misinterpreted. I'm somewhat cautious of including `md5crypt` on it's own here, sincce md5 is known to be a weak hashing algorithm - and this is a password-based tool. But perhaps it will be ok because this is a password cracking tool, rather than a hashing.

pages/common/john.md
sbrl: @dhondta so you're saying that we should leave it as `hash_format`? I'd agree with that, but just want to make sure :P

dhondta: @sbrl Yes, that's indeed what I'm saying. Let us leave it as `hash_format`.

agnivade: But that's not what I said :) So your statement "agnivade is right" is not right :laughing: 

The reason I suggested to use `md5crypt` is because we prefer concrete examples rather than generic ones to clarify the command usage, and in fact, it is part of our style guide.

I am wondering why you think specifying a hash format can be confusing ? Yes, there is an example to list hash formats, and a concrete example can stay too.

The discussion whether md5crypt is an appropriate example or not can of course be debated, and if needed something else can be used. But following our usual guidelines, I am in favor of using a concrete example.

zdroid: @agnivade This looks like a case when two rights make one wrong. :laughing:

I'm for md5crypt being put here, as there is `--format` preceding it.

sbrl: Oops, I totally forgot that I'm the commit author here lol

PR updated

PR #3992, opened 2020-04-16
youtube-dl: add subtitle options

DrHyde:
This adds examples for getting the list of available subtitle languages and downloading subtitles in a chosen language.

The page already had the maximum eight examples, so I removed two that I thought were far less important. The addition and removal are in separate commits in case you think that it would be better to break the no more than 8 rule.


zdroid:
I do agree your 2 examples are more important than the previous ones, but let's see what do others think.

PR #4000, opened 2020-04-19
at: add notify-send example

zdroid:
Closes #3862.

### DIFF
--- a/pages/linux/at.md
+++ b/pages/linux/at.md
@@ -16,4 +16,4 @@
 
 - Display a system notification at 11pm on February 18th:
 
-`echo "notify-send {{Wake up!}}" | at {{11pm}} {{Feb 18}}`
+`echo "notify-send '{{Wake up!}}'" | at {{11pm}} {{Feb 18}}`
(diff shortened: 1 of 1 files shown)


zdroid:
There was. You can re-read the previous discussion at #3862. I understood that I should make a new PR to avoid CLA conflicts.

sbrl:
Ah, I see @zdroid - thanks!

That still leaves the  issue of `reboot` though and the duplicated functionality.

zdroid:
@sbrl what is so problematic with it though? It's a use case many could need.

sbrl:
To summarise my earlier comments, I see 2 issues:

1. The `shutdown` command already allows a reboot to be scheduled.
2. The `reboot` command as an example requires `sudo`, so the command as-is wouldn't work unless you are root (e.g. you `sudo` the `at`).

zdroid:
Do you have any commands to replace it in mind? For example, it could be `notify-send` in Linux.

sbrl:
Yeah, `notify-send` could work - especially since this page is in the `linux` folder.

Other candidates could include a gzip, or something similar.

zdroid:
@sbrl Is it good now?

PR #4001, opened 2020-04-19
fgrep: add page

cesarsotovalero:

### DIFF
--- a/pages/common/fgrep.md
+++ b/pages/common/fgrep.md
@@ -15,7 +15,7 @@
 
 `fgrep -c {{search_string}} {{path/to/file}}`
 
-- Show line number of file with the line matched:
+- Show the line number in the file along with the line matched:
 
 `fgrep -n {{search_string}} {{path/to/file}}`
 
--- a/pages/common/fgrep.md
+++ b/pages/common/fgrep.md
@@ -1,0 +1,28 @@
+# fgrep
+
+> Matches patterns in files.
+> Supports simple patterns and regular expressions.
+
+- Search for an exact string in a file:
+
+`fgrep {{search_string}} {{path/to/file}}`
+
+- Search only lines that match entirely in files: 
+
+`fgrep -x {{path/to/file1}} {{path/to/file2}}`
+
+-Count of number of lines that match the given string in a file:
+
+`fgrep -c {{search_string}} {{path/to/file}}`
+
+- Shows line number of file with the line matched:
+
+`fgrep -n {{search_string}} {{path/to/file}}`
+
+- Display all lines except those that contain the regular expression:
+
+`fgrep -v {{^regex$}} {{path/to/file}}`
+
+- Search the file names that match the regular expression:
+
+`fgrep -l {{^regex$}} {{path/to/file1}} {{path/to/file2}}`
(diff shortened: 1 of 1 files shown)


zdroid:
I edited this myself to fix the errors pointed out by Travis. Other than that, it's all good.

cesarsotovalero:
Thanks @zdroid

pages/common/fgrep.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Search only lines that match entirely in files:
+
+`fgrep -x {{path/to/file1}} {{path/to/file2}}`
+
+- Count the number of lines that match the given string in a file:
+
+`fgrep -c {{search_string}} {{path/to/file}}`
+
+- Show line number of file with the line matched:
```
sbrl, line 18: Perhaps this will read better?
```suggestion
- Show the line number in the file along with the line matched:
```

pages/common/fgrep.md
```diff
@@ -0,0 +1,28 @@
...
+
+`fgrep -n {{search_string}} {{path/to/file}}`
+
+- Display all lines except those that contain the regular expression:
```
sbrl, line 22: Perhaps this will read better?
```suggestion
- Display all lines except those that contain the given regular expression:
```

pages/common/fgrep.md
```diff
@@ -0,0 +1,28 @@
...
+
+`fgrep -v {{^regex$}} {{path/to/file}}`
+
+- Search the file names that match the regular expression:
```
sbrl, line 26: I think this might be a bit misleading. Here's an extract from the man page:

```
       -l, --files-with-matches
              Suppress normal output; instead print the  name  of  each  input
              file  from  which  output would normally have been printed.  The
              scanning will stop on the first match.
```

In your explanation, it sounds like it searches the filenames themselves, instead of their content. Speaking of, you include a regex there. I thought you needed to include the `--perl-regexp` (`-P`) flag to activate this feature?

Anyway, how about we reword this - perhaps this will read better?
```suggestion
- Display filenames whose content matchs the regular expression at least once:
```

cesarsotovalero:
@sbrl: I agree with your changes. Thanks.

sbrl:
Thanks, @cesarsotovalero! GH's been weird today, so I think part of a comment got lost there. I also wanted to ask about long forms of arguments. In tldr pages we prefer long forms of arguments to aid clarify. For example, we prefer `--foo` instead of `-f`.

Looking at the `fgrep` man page, it would appear that long forms of options do exist. Perhaps we could update this page to use the long forms of options instead of the shorter ones?

cesarsotovalero:
@sbrl, I disagree with this.

> In tldr pages we prefer long forms of arguments to aid clarify

The purpose of each command is supposed to be given in the description above each command. Short commands are better because prevents users from typing or copying or reading large commands. Also, tldr is supposed to be as simple as possible. That's my opinion as a user.

If you look at https://github.com/tldr-pages/tldr/blob/master/pages/common/grep.md you notice that the abbreviated form is used there.

So, I propose `fgrep` as it is right now.

WDYT?

sbrl:
Ah, I see @cesarsotovalero. On most other pages, long forms of arguments are preferred and used, so using long-forms of arguments would make it more consistent with other pages.

However, as you mention the `grep` page does indeed use the short forms of arguments, so maybe it's ok to leave this with the short forms too.

PR #4025, opened 2020-05-07
Add subcommand section to contributing guidelines

### LINKED ISSUE #4003: Proposal: support subcommand
Command tool like `go` has powerful subcommand tools like `go test` `go build` which are totally different tools but having same prefix, it would be better tldr distinguishing them well.

fejx:
The ability to add subcommands is not documented anywhere. This is my proposal for a documentation.

Closes #4003 (Although the issue is both about subcommands and the `go` command.)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,6 +71,20 @@ As a quick reference, the format of each page should match the following templat
 For more detailed page formatting guidelines,
 refer to the [style guide](contributing-guides/style-guide.md).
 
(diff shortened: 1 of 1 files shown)


sbrl:
Hey, thanks @fejx! I believe it's documented in the [client specification](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md), but it does make sense to have it documented in the contributing guide too. Thanks!

PR #4028, opened 2020-05-07
reportbug: add page

Rudloff:

### DIFF
--- a/pages/linux/bugreport.md
+++ b/pages/linux/bugreport.md
@@ -1,6 +1,6 @@
 # bugreport
 
-> File a bug report on Debian.
+> Bug report tool of Debian distribution.
 > More information: <https://manpages.debian.org/buster/reportbug/reportbug.1.en.html>.
 
 - Generate a bug report about a specific package, then send it by e-mail:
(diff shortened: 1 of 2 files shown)


pages/linux/bugreport.md
```diff
@@ -0,0 +1,20 @@
+# bugreport
+
+> File a bug report on Debian.
```
andrik, line 3:
```suggestion
> Bug report tool of Debian distribution.
```

pages/linux/bugreport.md
```diff
@@ -0,0 +1,20 @@
...
+
+`reportbug {{package}}`
+
+- File a report about a package that does not exist yet (to request it, for example):
+
+`reportbug wnpp`
+
+- Report a bug that is not about a specific package (general problem, infrastructure, etc.):
+
+`reportbug other`
```
andrik, line 16: I read the documentation and it looks like to me that these two examples are referring to pseudo-package, right? Could we merge them?

Rudloff: `reportbug other` will list all the pseudopackages and you can select `wnpp` from there so I guess it is a bit redundant.
Do you think we can just drop the `wnpp` example?

andrik: Yeah, in that case, I think we can drop it.

PR #4035, opened 2020-05-10
Harmonize project name

waldyrious:
Follow-up of #1109, in particular of [this comment](https://github.com/tldr-pages/tldr/issues/1109#issuecomment-624346491).

I've already updated/renamed the wiki pages (see [here](https://github.com/tldr-pages/tldr/wiki/Home/_history), [here](https://github.com/tldr-pages/tldr/wiki/Minimum-specifications-for-tldr-pages-command-line-clients/_history) and [here](https://github.com/tldr-pages/tldr/wiki/tldr-pages-clients/_history)).

If I may add: now that I went through all the places where we used "tldr-pages" (with hyphen) and replaced them with "tldr pages" (with a space), I am now convinced that removing the hyphen is a the wrong direction to do the normalization, as it makes all the phrases where the project name appears more ambiguous. A good brand/name should achieve, if anything, the opposite!

Considering that the discussion in #1109 actually never produced a clear decision regarding the hyphenization of the project name (it was merely a no-consensus, as described [here](https://github.com/tldr-pages/tldr/issues/1109#issuecomment-568769222)), I'm inclined to make the reverse change in this PR, i.e. normalize everything towards tldr-pages. I'll wait for others to chime in before doing so, but please review the diff of commit ca3cca5dd to inform your comments.


agnivade:
I don't have any strong opinions. tldr-pages looks fine to me.

waldyrious:
Marked this as a draft to ensure it isn't merged until we make a decision. @zdroid does your approval mean you are OK with me changing to tldr-pages (with hyphen), as I mentioned above?

zdroid:
I'm approving any change that standardizes the project's name. Either work fine. Perhaps `tldr-pages` makes it more clear we are talking about the organization.

ivanhercaz:
@waldyrious, I am totally agree with changing to `tldr-pages` (with hyphen), at least it is what has more sense for me. As @zdroid indicated in the previous message, `tldr pages` seems like a plural of a tldr page, while `tldr-pages` sounds like a real project name!

waldyrious:
Alright, I'll make the change to `tldr-pages` then 👍

waldyrious:
This is now ready for review, but I'd like your input about related changes that I held off making yet: should we also rename the following items?
  - tldr.pdf --> tldr-pages.pdf
  - tldr.zip --> tldr-pages.zip
  - lint-tldr --> lint-tldr-pages (package.json script)

I suppose they should mostly be safe except perhaps the tldr.zip filename, but do let me know if I'm missing something regarding the other two.

Also: when merging, please use the rebase method so that the formatting tweaks commit isn't mixed up with the name standardization one.

zdroid:
`tldr-pages.zip` can be kept for the new pages structure.

zdroid:
@waldyrious tldr-pages.github.io should be updated too. I tried to do a part of the job in https://github.com/tldr-pages/tldr-pages.github.io/commit/43a767b528a82f1d50c1603ec58319824f6a1f13.

mebeim:
@waldyrious indeed the ZIP filename is pretty hard to rename. Many clients use it and renaming it would definitely break those. We could keep it as is and it also kind of makes sense to me since the client programs are always named `tldr`, not `tldr-pages` anyway, so it seems consistent with that.

waldyrious:
@zdroid, @mebeim, @sbrl no objections to including the following renames in this PR, then?

- `tldr.pdf` --> `tldr-pages.pdf`
- `lint-tldr` --> `lint-tldr-pages` (package.json script)

zdroid:
@waldyrious Those includes are alright.

mebeim:
@waldyrous oh sorry I slipped over them, those are fine to me :)

waldyrious:
Alright, I've added the two renames mentioned above. Ready to merge at your discretion!

zdroid:
Geronimo!

Issue #4032, opened 2020-05-09
ExTldr: new client for tldr-pages


ivanhercaz:
Dear friends of tldr-pages!
It is a pleasure to regards you again since a lot of time I have been inactive. I have been very busy, and I still busy, so my contributions to `tldr` suffered a drastic reduction. I will try to contribute as soon as possible with some commands I have in mind.

Between the stuff I am doing, I am learning Elixir for a project, but before to apply it I need to test my skills and improve them. I was curious about the existence of a Elixir client for tldr-pages and I found [TLDR Elixir Client](https://github.com/edgurgel/tldr_elixir_client), but it is archived and it isn't updated since 6 years ago. So I decided to contribute to tldr-pages with a Elixir client and...

[ExTldr](https://github.com/ivanhercaz/extldr) arrived! At this moment it is a very basic client, it just supports search a command using `extldr operating_system command`, but I will work in the multilingual support (ivanhercaz/extldr#1) as soon as possible. It just has one requirements, in addition to the one of download it: it is necessary to have Erlang installed (not Elixir).

Well, besides to announce you this new client, I would like to know if it could be added to the client list and what should we do with the current Elixir client in the list.

Again, it is a great pleasure to regard all of you. I hope everyone is nice and no one is affected of this horrible pandemic everyone is suffering around the world.

Huge hugs for everyone and good health!

**P.S.** Of course, I think it is not necessary to say but, it is more than welcome all suggestions or recommendations :heart: 
**P.S. (2)** v0.2.0 released, now ExTldr detect automatically the operating system of the user.

sbrl:
Hey there, @ivanhercaz! Thanks so much for the new client! Of course we can add it to the client list - it sounds like you're going to be putting lots of effort into developing and maintaining it (and, indeed, you have already!).

There are 2 places if I recall correctly that you can add your client to:

 - The wiki (no pull request required; can't remember if the list is still there or not)
 - The README (would you like to open a PR?)

zdroid:
Nice work. I agree with @sbrl's suggestions, you should definitely submit a PR to add it to the client's list. :smile:

ivanhercaz:
Thank you very much for your comments, @sbrl and @zdroid!

> it sounds like you're going to be putting lots of effort into developing and maintaining it (and, indeed, you have already!)

[...]

PR #4039, opened 2020-05-11
readme: add ExTldr client

ivanhercaz:
As I commented in #4032, I developed a new Elixir client for tldr-pages, although it is still in progress, it works fine at this moment and, as you can see in ivanhercaz/extldr#6, the main priority is to accomplish with the clients specification of tldr-pages. I am not going to extend much explaining the client and the aims, because there are already [open issues in the ExTldr repository](https://github.com/ivanhercaz/extldr/issues) and I explained practically everything in the issue mentioned at the beginning of this already long paragraph.

About this PR, three points about the changes I performed:

1. I rename the existing, but archived and apparently inactive, Elixir client.
2. I created a group of Elixir clients and then I added ExTldr.
3. I didn't specify instructions for ExTldr because to use it you just have to download the release and there isn't a package for any platform available.

As I said in the issue in which I introduced ExTldr to tldr-pages, I will be very thankfully of any feedback you give me :heart:


ivanhercaz:
In addition, yesterday [I edit the "Other console clients"](https://github.com/tldr-pages/tldr/wiki/tldr-pages-clients#other-console-clients) section to add ExTldr.

PR #4041, opened 2020-05-12
unshadow: add page

ldorigo:


ldorigo:
Note: there's only two examples (which are really the same) because it's an extremely simple utility that does only one thing. Still, I think it would be nice for it to have a page.

pages/linux/unshadow.md
```diff
@@ -0,0 +1,11 @@
+# unshadow
+
+> Utility provided by the package "john", to combine /etc/passwd and /etc/shadow into a file that can be processed by John the Ripper. More information: <https://www.openwall.com/john/>.
```
andrik, line 3:
```suggestion
> Utility provided by John the Ripper project to obtain the traditional Unix password file if the system uses shadow passwords. More information: <https://www.openwall.com/john/>.
```

sbrl: Perhaps the `More information` could be on the next line, like this @andrik?
```suggestion
> Utility provided by the John the Ripper project to obtain the traditional Unix password file if the system uses shadow passwords.
> More information: <https://www.openwall.com/john/>.
```

pages/linux/unshadow.md
```diff
@@ -0,0 +1,12 @@
+# unshadow
+
+> Utility provided by the John the Ripper project to obtain the traditional Unix password file if the system uses shadow passwords.
+> More information: <https://www.openwall.com/john/>.
+
+- Combine the /etc/shadow and /etc/passwd of the current system:
```
agnivade, line 6: I think these 2 examples should be just one. We always use tokens for options in commands. The value inside them can be concrete or generic. I had made the same comment some time earlier too. This is what I'd suggest:

Combine any two shadow and password files:

`sudo unshadow {{/etc/passwd}} {{/etc/shadow}}`.

PR #4045, opened 2020-05-14
pkg-config: add page

hojjatabdollahi:

### DIFF
--- a/pages/linux/pkg-config.md
+++ b/pages/linux/pkg-config.md
@@ -2,7 +2,6 @@
 
 > Return metainformation about installed libraries.
 
-
 - Get the list of libraries and their dependencies:
 
 `pkg-config --libs {{library1 library2 ...}}`
@@ -10,7 +9,3 @@
 - Get the list of libraries, their dependencies, and proper cflags for using with gcc:
 
 `pkg-config --cflags --libs {{library1 library2 ...}}`
-
-- Compile your code with `libwebkit2gtk-4.0`:
-
-``c++ main.c `pkg-config --cflags --libs webkit2gtk-4.0` ``
--- a/pages/linux/pkg-config.md
+++ b/pages/linux/pkg-config.md
@@ -1,0 +1,16 @@
+# pkg-config
+
+> Return metainformation about installed libraries.
+
+
+- Get the list of libraries and their dependencies:
+
+`pkg-config --libs {{library1 library2 ...}}`
+
+- Get the list of libraries, their dependencies, and proper cflags for using with gcc:
+
+`pkg-config --cflags --libs {{library1 library2 ...}}`
+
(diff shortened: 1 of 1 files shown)


pages/linux/pkg-config.md
```diff
@@ -0,0 +1,11 @@
+# pkg-config
+
+> Return metainformation about installed libraries.
+
+- Get the list of libraries and their dependencies:
+
+`pkg-config --libs {{library1 library2 ...}}`
+
+- Get the list of libraries, their dependencies, and proper cflags for using with gcc:
```
zdroid, line 9:
```suggestion
- Get the list of libraries, their dependencies, and proper cflags for gcc:
```

pages/linux/pkg-config.md
```diff
@@ -0,0 +1,11 @@
+# pkg-config
+
+> Return metainformation about installed libraries.
```
andrik, line 3:
```suggestion
> Tool to provide the details of installed libraries for compiling applications
```

zdroid: @andrik Maybe this is even better?
```suggestion
> Provide the details of installed libraries for compiling applications.
```

andrik: @zdroid I updated your suggestion adding `More information` section

hojjatabdollahi:
Hi @andrik and @zdroid,
Thanks for your comments. I agree with both of you. My original description was from the man pages. But this description is better, especially for the beginners.

hojjatabdollahi:
@andrik , @zdroid 
Originally I wanted to add an example with a compiler as well, but there was some error. Do you think I should try and add it one more time?

- Compile your code with `libwebkit2gtk-4.0`:
``c++ main.c `pkg-config --cflags --libs webkit2gtk-4.0` ``

sbrl:
Hey, @hojjatabdollahi! Thanks for the new page.

That example looks good. What do you mean by error? If you mean when trying to edit, try using the GitHub web interface.

Also, perhaps we could adjust that example there slightly. What about this?
```markdown
- Compile your code with libwebkit2gtk-4.0:

`c++ main.c "$(pkg-config --cflags --libs webkit2gtk-4.0)"`
```

The `$()` syntax is newer and more robust than the older backtick syntax.

hojjatabdollahi:
@sbrl 
Thank you. This might solve the "check error", I had to use double backticks and mix them with single backticks and I think the check failed because of that.
However, the double quotations in your example will result in an error, but the same works without the double qoutations. Should I go ahead and add it to the page?
- Compile your code with libwebkit2gtk-4.0:

`c++ main.c "$(pkg-config --cflags --libs webkit2gtk-4.0)"`

sbrl:
Go ahead, @hojjatabdollahi :smiley_cat:

PR #4058, opened 2020-05-21
File location fix

### LINKED ISSUE #3763: Token inconsistency: filename or file_name
The objective of this issue is to discuss a decision about the syntax of the token for the filenames in the examples written in the tldr-pages. As soon as there is a decision, I could mass change the discarded token to the chosen one.

In this case I address the `{{filename}}` and `{{file_name}}` token. To know the current situation I run the next script to know the current state in the English pages:

```bash
grep -o "{{file_name}}" pages/*/*.md && grep -o "{{file_name}}" pages/*/*.md | wc -l
```

For `{{filename}}` there are 170 results in English pages ([gist with the results](https://gist.github.com/ivanhercaz/8bea611aaa13713baddd4fe732d759c6#file-result_filename) and for `{{file_name}}` there are 18 results ([gist with the results](https://gist.github.com/ivanhercaz/8bea611aaa13713baddd4fe732d759c6#file-result_file_name).

**I prefer to use `{{filename}}`**, not because it is the token more used, but because, if I am not wrong, in English it is correct to use "filename" when we are talking about the name of the files in a computing context ([Wordreference definition](https://www.wordreference.com/definition/filename)).

What do you think about it?

fejx:
Closes #3763


fejx:
Woops, was a bit to fast with typing and clicking 😅

PR #4062, opened 2020-05-21
julia: add -E example, fix link

zdroid:
The link previously pointed to language's homepage, but not the REPL. @waldyrious @owenvoke  What do you think, is this link actually more useful?

I also added another example.

### DIFF
--- a/pages/common/julia.md
+++ b/pages/common/julia.md
@@ -1,7 +1,7 @@
 # julia
 
 > A high-level, high-performance dynamic programming language for technical computing.
-> More information: <https://julialang.org>.
+> More information: <https://docs.julialang.org/en/v1/manual/getting-started/>.
 
 - Start a Julia REPL session:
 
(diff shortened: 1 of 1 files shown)


pages/common/julia.md
```diff
@@ -1,7 +1,7 @@
 # julia
 
 > A high-level, high-performance dynamic programming language for technical computing.
-> More information: <https://julialang.org>.
+> More information: <https://docs.julialang.org/en/v1/manual/getting-started/>.
```
waldyrious, line 4: A better link would be https://docs.julialang.org/en/v1/stdlib/REPL/.

zdroid: But it doesn't help with any of the `julia`'s CLI subcommands (such as evaluation, parallelization, disabling fast math). That's why I went for this link instead.

waldyrious: Good point. The link I suggested is more about the REPL module than the CLI interface. It's a pity that the URL is very generic, but that's a minor nitpick. Let's keep your link. 👍

PR #4067, opened 2020-05-26
wc: add pipe example

jassler:

### DIFF
--- a/pages/common/wc.md
+++ b/pages/common/wc.md
@@ -17,3 +17,7 @@
 - Count characters in file (taking multi-byte character sets into account):
 
 `wc -m {{file}}`
+
+- Use standard input to count lines, words and characters (bytes) in that order:
+
+`find . | wc`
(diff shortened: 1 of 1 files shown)


pages/common/wc.md
```diff
@@ -17,3 +17,7 @@
 - Count characters in file (taking multi-byte character sets into account):
 
 `wc -m {{file}}`
+
+- Use standard input to count lines, words and characters (bytes) in that order:
+
+`find . | wc`
```
sbrl, line 23: Perhaps we could add some token syntax here?
```suggestion
`{{find .}} | wc`
```

jassler: Thanks for the suggestion, I like it!

Some other pipe examples (eg. in [grep](https://github.com/tldr-pages/tldr/blob/master/pages/common/grep.md)) leave the command itself red. Would it be more consistent then to only tokenize the dot? And / or replace `.` with `path/to/folder`?

`find {{path/to/folder}} | wc`

PR #4073, opened 2020-05-27
trash: fix token syntax

sbrl:
Ref #3910


zdroid:
Travis is certainly broken. :confused:

PR #4074, opened 2020-05-27
convert: fix delay example - fixes #4064

### LINKED ISSUE #4064: convert example doesn’t use the right unit for delay
The page about `convert` states:

> - Create a gif from a series of images with 100ms delay between them:
>
> `convert {{image1.png}} {{image2.png}} {{image3.png}} -delay {{100}} {{animation.gif}}`

But the option `-delay` is in ticks. The base for convert is 100 ticks-per-seconds so a delay of 1 is equivalent to 1/100 of a second.

So the actual example set a delay of 1s, 100ms would correspond to `-delay {{10}}`.

[Documentation of `-delay`](https://imagemagick.org/script/command-line-options.php#delay)

sbrl:
Fixes #4064

### DIFF
--- a/pages/common/convert.md
+++ b/pages/common/convert.md
@@ -25,7 +25,7 @@
 
 - Create a gif from a series of images with 100ms delay between them:
 
-`convert {{image1.png}} {{image2.png}} {{image3.png}} -delay {{100}} {{animation.gif}}`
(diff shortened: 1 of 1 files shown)


sbrl:
That's weird, I think Travis has had a hiccup. Merging manually, since this has to be ok with 4 approvals :P

PR #4084, opened 2020-05-31
Update project motto and improve README

### LINKED ISSUE #1192: Standardize the project description
Following up on #1109 (unifying the name of the project), I'm putting forward a proposal to unify the project description as well. Currently we have at least three variants:

- "Simplified and community-driven man pages" (in the [repo description](https://github.com/tldr-pages/tldr) and the [website sub-heading](http://tldr-pages.github.io/))
- "A community effort to simplify the beloved man pages with practical examples." (in the [website description](http://tldr-pages.github.io/))
- "A collection of simplified and community-driven man pages." (in the [README file](https://github.com/tldr-pages/tldr/blob/master/README.md#tldr))

I'm proposing `"A community-curated collection of simplified and example-driven man pages."` which is combines the best aspects of all three IMO. The goal is to use the same phrasing everywhere, kind of as a slogan.

Any thoughts? I'm wondering about "curated" (maybe "maintained" instead?) and "example-driven" (maybe "example-based" instead?).

waldyrious:
Fixes #1192 by editing the README, the `tldr.md` page, and the banner. We'll also want to update the project description under the repository title:

Screenshot 2020-05-31 at 19 21 58

While editing the banner contents, I took the liberty to rework its layout a bit, as its proportions were stretched when the main title changed from "TL;DR" to "tldr-pages" in #4035.

I also took the opportunity to restructure the README a bit to make it more approachable and consistent. **Check the [rendered version](https://github.com/tldr-pages/tldr/blob/project-slogan/README.md)** and let me know what you think!


README.md
```diff
@@ -39,18 +40,27 @@ Maybe it doesn't help that the first option explained in `man tar` is:
    and usually not even then as the default block size of 20 records (10240 bytes) is very common.
 ```
 
-Surely people could benefit from simplified man pages
-focused on practical examples.
+There seems to be room for simpler help pages, focused on practical examples.
```
agnivade, line 43: Totally a light suggestion, but I kinda liked the previous phrasing. Especially the "Surely people could benefit from.." part.

README.md
```diff
@@ -13,24 +14,24 @@
 [gitter-image]: https://badges.gitter.im/tldr-pages/tldr.svg
 [prs-merged-url]: https://github.com/tldr-pages/tldr/pulls?q=is:pr+is:merged
 [prs-merged-image]: https://img.shields.io/github/issues-pr-closed-raw/tldr-pages/tldr.svg?label=merged+PRs
-[issuestats-url]: http://isitmaintained.com/project/tldr-pages/tldr
-[issuestats-image]: http://isitmaintained.com/badge/resolution/tldr-pages/tldr.svg
+[issuestats-url]: https://isitmaintained.com/project/tldr-pages/tldr
+[issuestats-image]: https://isitmaintained.com/badge/resolution/tldr-pages/tldr.svg
```
sbrl, line 18: I'm not sure that the time to issue resolution is really valuable here. Perhaps we should remove it, but that sounds like a separate issue to the one being discussed here.

README.md
```diff
@@ -39,18 +40,27 @@ Maybe it doesn't help that the first option explained in `man tar` is:
...
-## Clients
+A popular and convenient way to access these pages on your computer
+is to install the [Node.js client](https://github.com/tldr-pages/tldr-node-client),
+which is supported by the tldr-pages project maintainers:
 
-You can access these pages on your computer using one of the following clients:
+    npm install -g tldr
+
+That way you can write `tldr tar` in the terminal to show the tldr page for `tar`,
+just like you would write `man tar` to show its manpage.
```
andrik, line 60: Hi @waldyrious , what about?

```suggestion
That way you can write `tldr tar` in the terminal and you will see the most common situations where such a command is used. Isn't that great?
```

README.md
agnivade: It's separate though. Let's do it in a different PR. And yes, I agree it's not valuable.

waldyrious: For context, it was meant as a concrete signal to visitors that contributions are welcomed in this project, and regularly integrated, as a way to encourage people to get involved. That's also the purpose of the "number of contributors" and "merged PRs" badges.

We can keep that badge only in CONTRIBUTING.md, as we've done with the CLA badge for example. If we do so, would we want to move all three? I think they sort of work as a group.

README.md
waldyrious: Can you explain a bit why you prefer this version?

Note that we already explain in the previous section what a tldr page is, so it's redundant to add yet another take at that here; and besides, with your suggestion we lose the parallel with manpages, which we [agreed in #1192](https://github.com/tldr-pages/tldr/issues/1192#issuecomment-629808084) that we should keep.

README.md
waldyrious: Just to provide some context: I removed the "Surely..." to avoid making the phrasing of this section too repetitive, since the previous sentence was changed to start with "It certainly..."

```diff
-Maybe it doesn't help that the first option explained in `man tar` is:
+It certainly doesn't help that the first option explained in `man tar` is:
```

... which in turn was motivated by the inclusion of "Maybe" in

```diff
-New to the command-line world? Or just a little rusty?
+Maybe you are new to the command-line world? Or just a little rusty?
```

... which was written that way to make it flow better with the introductory paragraph I added to the top of the section:

> The **tldr-pages** project is a collection of community-maintained help pages
for command-line tools, that aims to be a simpler, more approachable complement
to traditional [man pages](https://en.wikipedia.org/wiki/Man_page).

So we can change this line back to what it was, but the preceding paragraphs would need to be adapted as well to avoid making the whole thing awkward or repetitive. Happy to hear any suggestions :)

README.md
agnivade: The problem is that the metrics used by isitmaintained are incorrect and not what we would use to judge "activity" in our org. They measure how fast issues are closed, and what percentage of issues are open. Both of which are skewed in our case, because we have issues which are several years old. Mostly because those are discussions which we never got around to implement. That has got nothing to do with merging regular PRs.

If there is some other website that shows the contribution graph, then I am happy to include that.

README.md
agnivade: IMO, it doesn't look awkward to me. But I am totally unopinionated on this. It is a minor thing and I will defer to you.

README.md
andrik: Hi @waldyrious I totally forgot about the parallel with manpages and didn't pay attention that tldr was explained before

my bad :grimacing:

README.md
waldyrious: That's a good point, but I'm not sure the first metric (the only one we use — how fast issues are closed) is skewed in the statistical sense, considering that they use the median rather than the mean, and that they ignore issues older than 6 months, as described [here](https://isitmaintained.com/project/tldr-pages/tldr).

However, it may be skewed in the sense that IsItMaintainer counts both issues and PRs, rater than just PRs, and we tend to be more forgiving with long-lived issues\* than most open source projects, adopting an [eventualist](https://meta.wikimedia.org/wiki/Eventualism) philosophy that doesn't strive for closing issues as fast as possible, but instead favors actual resolution of the original issue (even though some some of us are sometimes more impatient — I'm looking at you, @zdroid! 😜).

Indeed a PR resolution metric would be more useful for the message we want to convey here. They don't seem to support that, and I didn't find alternatives that do, so yeah, maybe we should just remove the badge.

<sub>\* sometimes PRs too, but that's much rarer and probably statistically insignificant.</sub>

README.md
waldyrious: I appreciate your understanding. My preference is to either keep the set of changes I mentioned above, or adjust them all with alternative phrasings that would avoid the issues I pointed out above. Let me know what you guys think.

README.md
waldyrious: For you Golangers: I found two Go-based projects ([bitbadger](https://github.com/Pixep/bitbadger) and [nisekoi](https://github.com/UnsafePointer/nisekoi)) that seem to focus precisely on average PR merging time, but I suppose they'd need some massaging to feed an endpoint that would generate data we could pass to, say, a [dynamic shields.io badge](https://shields.io/#dynamic-badge). Do any of those look like something we could adapt without too much work?

README.md
agnivade: Sounds good. No objections from me. Thanks for your work Waldir.

README.md
agnivade: FYI, we are called "Gophers" :stuck_out_tongue:. I don't know, personally I don't find any value in badges like these. To me, if I want to see if a project is active - I just look at the last commit data, and if the issue tracker is active or not. Both of which are perfectly fine in our case. I am fine with average PR merging time, but wouldn't bother about spending too much effort on that.

waldyrious: Ok, I'll remove the badge in an upcoming PR.

sbrl: Sounds good! Sorry for not getting back sooner with this. Let's fiddle with this badge in a future PR. For now I'll approve this PR.

zdroid:
I think that with a few tweaks (e.g. to the clients list), we could easily submit this readme to https://github.com/matiassingers/awesome-readme.

zdroid:
@sbrl @agnivade @owenvoke tldr repository description should be changed too: https://github.com/tldr-pages/tldr

agnivade:
Done. Thanks.

zdroid:
@waldyrious Which description did you have in mind to be put in repo's description, the one in the banner image, or the one in the readme text (as it is currently in the repo's description)?

waldyrious:
> @waldyrious Which description did you have in mind to be put in repo's description, the one in the banner image, or the one in the readme text (as it is currently in the repo's description)?

Good point, @zdroid. I would indeed suggest using the newly adopted motto in all short descriptions, i.e. `collaborative cheatsheets for console commands`, or at most the longer version mentioned in #1192, `collaborative collection of compact cheatsheets for console commands`. The one I used in the README (`A collection of community-maintained help pages for command-line tools`) was written that way in order to flow better with the sentence and paragraph it's embedded into, rather than stand alone as an independent sentence.

IMO, using the latter option as the repo/org description —although I fully agree with its contents, as the person who wrote it!— goes against the goal of #1192, which was to standardize the short description we use in various places. That said, this is not a dogmatic position; I'm happy to be convinced otherwise (even if that means changing our motto again!). Let me know what you guys think.

zdroid:
I'd prefer `collaborative cheatsheets for console commands` used for that, as it is in banner.

agnivade:
Done that too.

README.md
waldyrious: Update: I've opened #4103 to remove the badge.

PR #4088, opened 2020-06-03
unrar: extract to another directory example

d9k:
- [v] The page has 8 or fewer examples.
- [v] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [v] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).

### DIFF
--- a/pages/common/unrar.md
+++ b/pages/common/unrar.md
@@ -6,6 +6,10 @@
 
 `unrar x {{compressed.rar}}`
 
+- Extract files to existing path with original directory structure:
+
+`unrar x {{compressed.rar}} {{path/to/extract}}`
+
 - Extract files into current directory, losing directory structure in the archive:
 
 `unrar e {{compressed.rar}}`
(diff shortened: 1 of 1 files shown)


pages/common/unrar.md
```diff
@@ -6,6 +6,10 @@
 
 `unrar x {{compressed.rar}}`
 
+- Extract files to existing path with original directory structure:
```
schneiderl, line 9: I'm unsure I understood this description. It is kind of confusing to me. Would you mind to explain?
As far as I understand this command extracts the content of a compressed file into a directory. What do you mean by "with original directory structure"?

schneiderl: Oh, I think I get it now. 

Does this mean that the file would be extracted and a new directory would be created?
So the final result would be something like this:
`c:/exctracted_path/extracted_files`
instead of this:
`c:/exctracted_files`

Is my understanding correct?

zdroid, line 9:
```suggestion
- Extract files to a specified path with the original directory structure:
```

zdroid:
@schneiderl @einverne I guess this can be merged now.

PR #4093, opened 2020-06-07
exiv2: add page

jn64:

### DIFF
--- a/pages/common/exiv2.md
+++ b/pages/common/exiv2.md
@@ -5,24 +5,24 @@
 
 - Print a summary of the image Exif metadata:
 
-`exiv2 {{filename}}`
+`exiv2 {{path/to/file}}`
 
 - Print all metadata (Exif, IPTC, XMP) with interpreted values:
 
-`exiv2 -P kt {{filename}}`
+`exiv2 -P kt {{path/to/file}}`
 
 - Print all metadata with raw values:
 
-`exiv2 -P kv {{filename}}`
+`exiv2 -P kv {{path/to/file}}`
 
 - Delete all metadata from an image:
 
-`exiv2 -d a {{filename}}`
+`exiv2 -d a {{path/to/file}}`
 
 - Delete all metadata, preserving the file timestamp:
 
-`exiv2 -d a -k {{filename}}`
+`exiv2 -d a -k {{path/to/file}}`
 
 - Rename the file, prepending the date and time from metadata (not from the file timestamp):
 
-`exiv2 -r {{'%Y%m%d_%H%M%S_:basename:'}} {{filename}}`
+`exiv2 -r {{'%Y%m%d_%H%M%S_:basename:'}} {{path/to/file}}`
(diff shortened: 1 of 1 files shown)


pages/common/exiv2.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Print all metadata (Exif, IPTC, XMP) with interpreted values:
+
+`exiv2 -P kt {{filename}}`
+
+- Print all metadata with raw values:
+
+`exiv2 -P kv {{filename}}`
+
+- Delete all metadata from image:
```
zdroid, line 18:
```suggestion
- Delete all metadata from an image:
```

pages/common/exiv2.md
```diff
@@ -0,0 +1,28 @@
...
+
+`exiv2 -d a {{filename}}`
+
+- Delete all metadata, preserving the file timestamp:
+
+`exiv2 -d a -k {{filename}}`
+
+- Rename the file, prepending the date and time from metadata (not file timestamp):
```
zdroid, line 26:
```suggestion
- Rename the file, prepending the date and time from metadata (not from the file timestamp):
```

pages/common/exiv2.md
```diff
@@ -0,0 +1,28 @@
+# exiv2
+
+> Image metadata manipulation tool.
+> More information: <https://www.exiv2.org/manpage.html>.
+
+- Print a summary of the image Exif metadata:
+
+`exiv2 {{filename}}`
```
sbrl, line 8: Perhaps this will read better? It would be more consistent with other pages.
```suggestion
`exiv2 {{path/to/filename}}`
```

Same goes for the other examples.

pages/common/exiv2.md
```diff
@@ -0,0 +1,28 @@
...
+
+`exiv2 -d a {{filename}}`
+
+- Delete all metadata, preserving the file timestamp:
+
+`exiv2 -d a -k {{filename}}`
+
+- Rename the file, prepending the date and time from metadata (not from the file timestamp):
+
+`exiv2 -r '%Y%m%d_%H%M%S_:basename:' {{filename}}`
```
sbrl, line 28: If the user can edit the value of the `-r` flag, perhaps it could go in `{{token_syntax}}`?
```suggestion
`exiv2 -r '{{%Y%m%d_%H%M%S_:basename:}}' {{filename}}`
```

Also, I notice that a number of these example have short arguments (e.g. `-d`, `-r`). In tldr pages, we generally prefer the long-form arguments (e.g. `--foo`), as it aids clarity. Do any of these arguments have long forms that we could use instead?

pages/common/exiv2.md
jn64: That seems overly verbose and also technically incorrect. It could be path/to/file but not path/to/filename

pages/common/exiv2.md
jn64: Yes there're long options, but I thought the convention was to use short options if they were fairly self-explanatory / unambiguous from the description? Looking at the pages for grep, sort, ls, git-commit...

PR #4096, opened 2020-06-09
grep: regexp is used by default

szszszsz:
grep searches for regexp by default on Linux, and the '-F' switch is required for the exact string. 
Corrected by replacing `search_string` with `regexp`, or adding the mentioned switch.

Please review / confirm.

### DIFF
--- a/pages/common/grep.md
+++ b/pages/common/grep.md
@@ -3,34 +3,34 @@
 > Matches patterns in input text.
 > Supports simple patterns and regular expressions.
 
+- Search for a pattern within a file:
+
+`grep {{search_pattern}} {{path/to/file}}`
+
 - Search for an exact string:
 
-`grep {{search_string}} {{path/to/file}}`
+`grep -F {{exact_string}} {{path/to/file}}`
 
-- Search in case-insensitive mode:
+- Search for a pattern recursively in the current directory, ignoring non-text files:
 
-`grep -i {{search_string}} {{path/to/file}}`
+`grep -RI {{search_pattern}} .`
 
-- Search recursively (ignoring non-text files) in current directory for an exact string:
+- Use extended regular expressions (supporting `?`, `+`, `{}`, `()` and `|`), in case-insensitive mode:
 
-`grep -RI {{search_string}} .`
-
-- Use extended regular expressions (supporting `?`, `+`, `{}`, `()` and `|`):
-
-`grep -E {{^regex$}} {{path/to/file}}`
+`grep -Ei {{search_pattern}} {{path/to/file}}`
 
 - Print 3 lines of [C]ontext around, [B]efore, or [A]fter each match:
 
-`grep -{{C|B|A}} 3 {{search_string}} {{path/to/file}}`
+`grep -{{C|B|A}} 3 {{search_pattern}} {{path/to/file}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/grep.md
```diff
@@ -3,34 +3,38 @@
...
+
 - Search for an exact string:
 
-`grep {{search_string}} {{path/to/file}}`
+`grep -F {{search_string}} {{path/to/file}}`
 
 - Search in case-insensitive mode:
 
-`grep -i {{search_string}} {{path/to/file}}`
+`grep -i {{^regex$}} {{path/to/file}}`
```
einverne, line 16: If you put `^` and `$` here, I think it only filter the line exactly match the `regex` here.

zdroid: I don't think it's actually necessary to replace all `search_string`s with regexes. That way someone could get a wrong impression that you must use a regex with grep.

zdroid: Is this how regexes were displayed inside `{{}}` in the other pages?

szszszsz:
Note: it seems page has now 9 examples (I have added one perhaps unnecessarily).

pages/common/grep.md
```diff
@@ -3,34 +3,38 @@
...
 - Search recursively (ignoring non-text files) in current directory for an exact string:
 
-`grep -RI {{search_string}} .`
+`grep -RIF {{search_string}} .`
 
 - Use extended regular expressions (supporting `?`, `+`, `{}`, `()` and `|`):
 
 `grep -E {{^regex$}} {{path/to/file}}`
 
 - Print 3 lines of [C]ontext around, [B]efore, or [A]fter each match:
```
zdroid, line 26: My suggestion is to remove this one. The example after that already shows filename and line numbers, which is enough context to find the match.

sbrl: I don't know. I find the lines either side as context much more useful myself.

agnivade: I agree. I almost always grep with a context. Would suggest keeping this one.

waldyrious:
Hey @szszszsz — thanks for the PR! I agree with your original premise, but I'd suggest changing `{{search_string}}` to `{{search_pattern}}` instead of to `{{^regex$}}`, for the reasons already raised by others in the comments.

Also, I think we only need one example for an exact string match, so I'd keep the first one (maybe using a `{{exact_string}}` parameter to further drive the point home) and would remove the "exact" from the description of the second one (which also avoids overloading the example with three options).

Finally, given our space constraints, maybe we could combine the case-insensitive example with the extended regex one.

With these changes, I'd say the PR would be OK to merge.

What do you think?

waldyrious:
@agnivade, @einverne, @zdroid, WDYT about the changes I proposed above? @sbrl has already reacted to my suggestions with agreement.

zdroid:
I agree with your suggestions as well. ;)

einverne:
> @agnivade, @einverne, @zdroid, WDYT about the changes I proposed above? @sbrl has already reacted to my suggestions with agreement.

I agree with your suggestions.

waldyrious:
Alright, thanks all for the input. I'll make the changes so we can merge the PR.

waldyrious:
Updated the PR. I kept @szszszsz's authorship of the commit.

Issue #4092, opened 2020-06-07
Clarify language resolution algorithm


MasterOdin:
It's a bit unclear to me on the language resolution. Looking at the spec, it has:

> Pages can be written in multiple languages. If a client has access to environment variables, several standard ones exist to specify the language in which a client should operate. If not, then clients MUST make reasonable assumptions based on the information provided by the environment in which they operate (e.g. consulting navigator.languages in a browser, etc.). If possible, it is RECOMMENDED to also make language configurable, as to not only rely on the environment. Clients SHOULD therefore offer options to configure or override the language using configuration files or command line options (like -L, --language as suggested in the arguments section above).
> 
> The LANG environment variable, if present, MUST be used to determine the language of pages to display.

From this, there's some amount of ambiguity (to me) on how clients should handle when `LANG` is set to a language in which a page does not exist (or even any pages exist). From a strict interpretation, it would mean that if `LANG=cz` for example, then no pages would get shown, unless they configured the client in some other way (e.g. through `--language flag`). However, a looser intrepretation (which the node client takes), is to always check if the page exists in English, after checking other languages, even if explicitly using the `--language` flag. For example, the following two return the same english page:
```
LANG=fr node bin/tldr ansible galaxy
node bin/tldr --language fr ansible galaxy
```

It would be good to clarify that either clients should always fallback to english if a page does not have a page in the requested locale, or vice versa, as currently, it leaves room for clients to have different behavior imo.

Personally, I think the clients should always fallback to English given how sparse the translations are, and would suggest adding the following text to the section:

> Regardless of selected language through environment variables, tldr should always attempt to fallback to English if the page does not exist in the requested languages. It is SUGGESTED that the client tell the user that the page does not exist in their requested language however if it was not English. If the client supports a command line argument for language, the client MUST only attempt to show the page in that language.

sbrl:
Hey there! That's a great observation. I kind of assumed that people would have English at least somewhere in their `LANG` setting - but in hindsight that's a silly assumption. That wording sounds ok. What about this though?

```
Regardless of selected language through environment variables, clients SHOULD always attempt to fallback to English if the page does not exist in the requested languages. It is SUGGESTED that the client tell the user that the page does not exist in their requested language however if it was not English. If the client supports a command line argument for language, the client MUST only attempt to show the page in that language (though it is SUGGESTED that clients notify the user that a page is available in other languages if present).
```

zdroid:
[...]

PR #4101, opened 2020-06-11
CLIENT-SPECIFICATION: clarify fallback to english

sbrl:
As per the conversation in #4092, I've opened this PR.

I've changed the wording again a bit in order to  comply with [RFC 2119](https://tools.ietf.org/html/rfc2119) and [RFC 6919](https://tools.ietf.org/html/rfc6919).


CLIENT-SPECIFICATION.md
```diff
@@ -179,6 +179,8 @@ The [`LANG` environment variable](https://www.gnu.org/software/gettext/manual/ht
...
 
+Regardless of the language selected through the above environment variables, clients SHOULD always attempt to fallback to English if the page does not exist in the requested languages. In this case clients REALLY SHOULD tell the user that the page does not exist in their requested language however if it was not English. If the client supports a command line argument for language, the client MUST only attempt to show the page in that language (clients OUGHT TO notify the user that a page is available in other languages if present).
```
zdroid, line 182:
```suggestion
Regardless of the language selected through the above environment variables, clients SHOULD always attempt to fallback to English if the page does not exist in the requested languages. In this case clients REALLY SHOULD tell the user that the page does not exist in their requested language however if it was not English. If the client supports a command-line argument for language, the client MUST only attempt to show the page in that language (clients OUGHT TO notify the user that a page is available in other languages if present).
```
`command-line` is the form we use in the other places.

principis:
I'm a bit confused. According to [gettext](https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html), `LANGUAGE` may be preferred over `LANG`, but the spec specifies `LANG` should be preferred over `LANGUAGE`.

Also, the note at the bottom of the Language section is a bit confusing:
  
    Note that it is highly RECOMMENDED to give precedence to the platform first, and then the language. In other words, look for a platform under each language, before falling back to the next preferred language. This ensures a meaningful and correct page resolution.

Especially this part: _look for a platform under each language, before falling back to the next preferred language_ 
I think this should be: _before falling back to the next `platform`_

This ties in to the whole problem of selecting a page by platform and falling back to English.
`clients SHOULD always attempt to fallback to English if the page does not exist in the requested languages`

For example:
A user searches for `some-page`, is on the `Linux` platform and the variables are set to `LANG=nl_BE;LANGUAGE=nl:fr:de`

Let's assume that `some-page` exists like this:
```
pages/common/some-page.md
pages.fr/windows/some-page.md
pages/windows/some-page.md
```

Does the client need to select the French windows page or the English common page? In my opinion, if English isn't set as a preferred language, it should always be searched for (and the user should be notified if the client shows an English page), thus showing the English common page. Showing a windows page to a Linux user isn't really useful most of the time. But there may be users who can't read English very well.

MasterOdin:
Pulling my last comment https://github.com/tldr-pages/tldr/issues/4092#issuecomment-642717320 here:

> My intention with the first sentence was to mandate the fallback to English, and would advocate heavily for upgrading "SHOULD" to "MUST" (I realize I used should in my original draft, apologies).

I think there are already too many optional bits to the language spec, and I think this one should be mandatory as it provides a consistent behavior across clients that do language (namely that they fallback to English).

zdroid:
@sbrl I edited two minor things. Other than that, the proposal is all good.

columbarius:
I was suggested to engage in this discussion. Since this PR is about the specific formulation of the specification I added my suggestion to the   https://github.com/tldr-pages/tldr/issues/4092#issuecomment-647021447.

zdroid:
This needs review by org owners, i.e. @agnivade @owenvoke @mebeim.

sbrl:
@MasterOdin I completely agree about the fallback there. I'll edit.

CLIENT-SPECIFICATION.md
```diff
@@ -179,6 +179,8 @@ The [`LANG` environment variable](https://www.gnu.org/software/gettext/manual/ht
...
 
+Regardless of the language selected through the above environment variables, clients MUST always attempt to fallback to English if the page does not exist in the requested languages. In this case clients REALLY SHOULD tell the user that the page does not exist in their requested language however if it was not English. If the client supports a command-line argument for language, the client MUST only attempt to show the page in that language (clients OUGHT TO notify the user that a page is available in other languages if present).
+
```
MasterOdin, line 183: Taking from https://github.com/tldr-pages/tldr/issues/4092#issuecomment-647036499. Could be useful to include a table (similar to the platform & language section) to give examples of different variable configurations to make things as explicit as possible.

```suggestion
LANGUAGE | LANG | Result
---------|------|-------
it,cz,de | cz   | it,cz,de,en
it,de,fr | cz   | it,de,fr,cz,en
  --     | it   | it,en
it,cz    | --   | it,cz,en
  --     | --   | en

Note: `LANG` or `LANGUAGE` may contain a value of `C` or `POSIX` and this should be treated as ignored.
```

columbarius: Should'nt the 4th case in this table have just '"en"' as the Result, since `LANGUAGE` should be ignored, if `LANG` is not set, or set to `"C"`, as described [here](https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html#The-LANGUAGE-variable)?

MasterOdin: Yup, updated. All the more reason to have a table as reading those docs is a pain.

columbarius: Thanks for updating and adding this table. It's my first time collaborating here and i'm still learning how things are done.

agnivade:
I haven't been following this discussion closely. I'll settle for whatever is decided here.

CLIENT-SPECIFICATION.md
```diff
@@ -179,6 +179,16 @@ The [`LANG` environment variable](https://www.gnu.org/software/gettext/manual/ht
 
 The [`LANGUAGE` environment variable](https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html) specifies a priority list of languages that a user wishes to read in. In absence of `LANG` and if this environment variable is present, a client MUST use the defined priority list to decide on the language of pages to display. If a page is not available in the user's preferred language, then a client MUST respect the user's priority list defined in the `LANGUAGE` variable, and MAY choose to notify the user that a page in their chosen language couldn't be found (perhaps along with a link to the [translations section of the contributing guide](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#translations)).
```
sbrl, line 180: With the updates below, this paragraph might need a bit of a rethink. It seems to be contradicting the below behaviour with the `LANG` and `LANGUAGE` environment variables.

(also, if `LANG` needs to be set to something other than an empty string or `C` to enable the `LANGUAGE` env  variable, what does one set it to?0

principis: > With the changes made below, this paragraph might need a bit of a rethink. It seems to be contradicting the below behaviour with the `LANG` and `LANGUAGE` environment variables.
> 
> (also, if `LANG` needs to be set to something other than an empty string or `C` to enable the `LANGUAGE` env variable, what does one set it to?0

LANG should be set to your primary language, for example `en_US.UTF-8` or `nl_BE.UTF-8`

sbrl: Ah, I see - thanks @principis!

Anyway, I think this paragraph needs updating to reflect the actual behaviour with respect to the GNU spec.

MasterOdin:
```suggestion
The [`LANGUAGE` environment variable](https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html) specifies a priority list of languages that a user wishes to read in. The `LANG` environment variable MUST be present for `LANGUAGE` to be used, and then `LANGUAGE` takes precendence over `LANG`. If `LANG` is set to a language not in `LANGUAGE`, then `LANG` should be appended to the end of the priority list. If a page is not available in the user's preferred language, then a client MUST respect the user's priority list defined in the `LANGUAGE` variable, and MAY choose to notify the user that a page in their chosen language couldn't be found (perhaps along with a link to the [translations section of the contributing guide](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#translations)).
```

CLIENT-SPECIFICATION.md
```diff
@@ -179,6 +179,16 @@ The [`LANG` environment variable](https://www.gnu.org/software/gettext/manual/ht
...
 
+Regardless of the language selected through the above environment variables, clients MUST always attempt to fallback to English if the page does not exist in the requested languages. In this case clients REALLY SHOULD tell the user that the page does not exist in their requested language however if it was not English. If the client supports a command-line argument for language, the client MUST only attempt to show the page in that language (clients OUGHT TO notify the user that a page is available in other languages if present).
```
mebeim, line 182: I'd suggest:

```suggestion
Regardless of the language selected through the above environment variables, clients MUST always attempt to fallback to English if the page does not exist in the requested languages. In this case clients SHOULD tell the user that the page does not exist in their requested language if it was not English. If the client supports a command-line argument for language, the client MUST only attempt to show the page in that language (clients MAY notify the user that a page is available in other languages if present).

```

- `REALLY SHOULD` -> `SHOULD`
- `OUGHT TO` -> `MAY`
- "however" not needed as that sentence already implies the exception.
- Added one line of spacing between text and the table below.

CLIENT-SPECIFICATION.md
```diff
@@ -179,6 +179,16 @@ The [`LANG` environment variable](https://www.gnu.org/software/gettext/manual/ht
...
+LANGUAGE | LANG | Result
+---------|------|-------
+it,cz,de | cz   | it,cz,de,en
+it,de,fr | cz   | it,de,fr,cz,en
+  --     | it   | it,en
+it,cz    | --   | en
+  --     | --   | en
+
+Note: `LANG` or `LANGUAGE` may contain a value of `C` or `POSIX` and this should be treated as ignored.
```
mebeim, line 191:
```suggestion
Note: `LANG` or `LANGUAGE` may contain the values `C` or `POSIX`, which should be ignored.

```

Clearer, also added one line of padding.

CLIENT-SPECIFICATION.md
```diff
@@ -179,6 +179,16 @@ The [`LANG` environment variable](https://www.gnu.org/software/gettext/manual/ht
...
 
+Regardless of the language selected through the above environment variables, clients MUST always attempt to fallback to English if the page does not exist in the requested languages. In this case clients REALLY SHOULD tell the user that the page does not exist in their requested language however if it was not English. If the client supports a command-line argument for language, the client MUST only attempt to show the page in that language (clients OUGHT TO notify the user that a page is available in other languages if present).
+LANGUAGE | LANG | Result
+---------|------|-------
+it,cz,de | cz   | it,cz,de,en
```
mebeim, line 185:
```suggestion
`it:cz:de` | `cz`   | `it`, `cz`, `de`, `en`
```

I'd suggest changing this table to have correct values for `LANGUAGE` (using `:`) and also using `code blocks` like the above suggestion.

CLIENT-SPECIFICATION.md
mebeim: I agree with @MasterOdin's suggestion here :+1:

sbrl:
Updates complete!

CLIENT-SPECIFICATION.md
```diff
@@ -177,7 +177,19 @@ Pages can be written in multiple languages. If a client has access to environmen
...
+
+Regardless of the language selected through the above environment variables, clients MUST always attempt to fallback to English if the page does not exist in the requested languages. In this case clients SHOULD tell the user that the page does not exist in their requested language if it was not English. If the client supports a command-line argument for language, the client MUST only attempt to show the page in that language (clients MAY notify the user that a page is available in other languages if present).
+
+LANGUAGE	| LANG	| Result
+------------|-------|----------
+`it:cz:de`	| `cz`	| `it`, `cz`, `de`, `en`
+`it:de:fr`	| `cz`	| `it`, `de`, `fr`, `cz`, `en`
+--			| `it`	| `it`, `en`
+`it,cz`		| --	| `en`
```
mebeim, line 189:
```suggestion
`it:cz`		| --	| `en`
```

Missed this.

mebeim: Actually I'll just apply it myself since it's just a typo.

mebeim:
Re-reading the env var paragraph, I don't think that explicitly requiring `LANG` to be set is needed here. The [linked GNU gettext reference](https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html) does not actually say that `LANGUAGE` is to be ignored if `LANG` is not set. 

This paragraph is quite complex. I'll try to reword it when I have time to make it as simple and clear as possible, but in a nutshell I think we should just stick to the priority list of [this documentation page](https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html#Locale-Environment-Variables), simply giving `LANGUAGE` priority over `LANG`.

columbarius:
> Re-reading the env var paragraph, I don't think that explicitly requiring `LANG` to be set is needed here. The [linked GNU gettext reference](https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html) does not actually say that `LANGUAGE` is to be ignored if `LANG` is not set.


As i read the Note:-Part LANG must be present to enable the LANGUAGE variable. I tested it with `gettext --usage` and found exactly this behaviour:
``` bash
% echo $LANG                          
de_DE.UTF-8
% gettext --usage
gettext: Unbekannte Option »--usage«
»gettext --help« gibt weitere Informationen.
% LANGUAGE=it_IT:de_DE gettext --usage
gettext: Unbekannte Option »--usage«
Usare 'gettext --help' per maggiori informazioni.
% env -u LANG LANGUAGE=it_IT:de_DE gettext --usage
gettext: unrecognized option '--usage'
Try 'gettext --help' for more information.
% LANG=en gettext --version
gettext (GNU gettext-runtime) 0.20.2
Copyright (C) 1995-2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Written by Ulrich Drepper.
````
You may try this on your system.

**Comment by mebeim** [comment]
@columbarius Ok I see. Then the bottom line is: respect `$LANG` when available, followed by the list in `$LANGUAGE` if present, otherwise use English. I still believe the current paragraph is quite complex so I'll reword it.

**Comment by sbrl** [comment]
Thanks for spotting the typo @mebeim! I guess I'll hold off on merging until you've re-worded it?

**Comment by columbarius** [comment]
@mebeim Yes, respect `$LANG` when available, followed by the list in `$LANGUAGE` which prioritizes the value in `$LANG`, which is the confusing part of this standard.  If `$LANG` and `$LANGUAGE` is set, respect`$LANGUAGE` before `$LANG`, but if `$LANG` is not set, ignore `$LANGUAGE`. 

My [example with the gettext command](https://github.com/tldr-pages/tldr/pull/4101#issuecomment-650645755), shows, that this is used for stdout, but stderr just uses `$LANG` and ignores `$LANGUAGE`, but I don't think, we have to use that behaviour.

File: CLIENT-SPECIFICATION.md [hunk 8]
```diff
@@ -173,29 +173,37 @@ If multiple versions of a page were found for different platforms, then a client
...
+In order to determine the display language, a client MUST:
 
-Regardless of the language selected through the above environment variables, clients MUST always attempt to fallback to English if the page does not exist in the requested languages. In this case clients SHOULD tell the user that the page does not exist in their requested language if it was not English. If the client supports a command-line argument for language, the client MUST only attempt to show the page in that language (clients MAY notify the user that a page is available in other languages if present).
+1. Check the value of `LANG`.
```
sbrl, line 182: What about this?
```suggestion
1. Check the value of `LANG`. If not set, then skip to step 5.
```

zdroid: I think it's not what is intended, since `LANGUAGE` (step 2) can contain the information, too.

MasterOdin: Looking at https://github.com/tldr-pages/tldr/pull/4101#issuecomment-650645755, which demonstrates different values of `LANG` and `LANGUAGE` for `gettext`, it would seem that `LANG` must be set for `LANGUAGE` to have any effect.

mebeim: @sbrl @MasterOdin ops, you're right guys, thanks for spotting this.

zdroid:
Since virtually every team member agreed with this, I guess we can merge it now?

sbrl:
Sounds good to me @zdroid - go ahead :D

PR #4105, opened 2020-06-12
multipass: add page

monsieurlo:


pages/linux/multipass.md
```diff
@@ -0,0 +1,36 @@
+# multipass
+
+> Creates a Ubuntu virtual machine using native hypervisors.
+> More information: <https://multipass.run/>.
+
+- List the aliases that can be used to launch an instance:
+
+`multipass find`
+
+- Launch a new instance, set its name, use a cloud-init configuration file:
```
zdroid, line 10:
```suggestion
- Launch a new instance and set its name using a cloud-init configuration file:
```

monsieurlo: - The 3 files are similar but there's none for sunos. So should it go to common anyway?
- "using" is not appropriate: the name is set using the -n flag while the config file (which does not define the name) is set using --cloud-init.

zdroid: - Yep. `common/` is for 2 or more platforms, especially for 3. :smile:
-  Then I probably didn't understand the text here. Can you try to rephrase what you meant more clearly?

monsieurlo: - Moved to common :)
- I don't really know how to make it clearer... Here's from the official [doc](https://multipass.run/docs/launch-command):
  - -n, --name <name>    Name for the instance.
  - --cloud-init <file>  Path to a user-data cloud-init configuration

I can remove that --cloud-init flag, but that would be a pity since it's a pretty useful one

jn64: If I understand correctly they are just unrelated options, so:

> Launch a new instance, set its name, _and_ use a cloud-init configuration file:

monsieurlo: You understand correctly :)
Applied as-is, only removed the comma before the "and".

pages/common/multipass.md
```diff
@@ -0,0 +1,36 @@
+# multipass
+
+> Creates a Ubuntu virtual machine using native hypervisors.
```
andrik, line 3: Hi @monsieurlo what about?

```suggestion
> CLI to manage Ubuntu virtual machines using native hypervisors.
```

monsieurlo: Hi, I just pushed the modification. (Sorry for CLA thing, I completely forgot in the meantime I had changed my gitconfig locally :p )

PR #4110, opened 2020-06-17
Tweak badge colors

waldyrious:
Just a few adjustments to the badges to use more appropriate colors:
- gitter: use a deep pink to match the platform's brand
- PRs merged: switch from yellow to green, to convey the intended positiveness
- license: switch from the dull grey to a move vivid blue

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,15 +7,15 @@
 [![license][license-image]][license-url]
 
 [gitter-url]: https://gitter.im/tldr-pages/tldr
-[gitter-image]: https://badges.gitter.im/tldr-pages/tldr.svg
+[gitter-image]: https://img.shields.io/badge/chat-on_gitter-deeppink
 [prs-merged-url]: https://github.com/tldr-pages/tldr/pulls?q=is:pr+is:merged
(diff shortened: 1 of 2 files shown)


zdroid:
TBH I don't like red in this context, it looks like something is failing (e.g. a build). Green was better for Gitter.

andrik:
I agree with @zdroid 

@waldyrious what about a lighter pink or even purple

waldyrious:
Maybe one of these?
1. 
2. 
3. 

My preference would be 2 (which is closer to gitter's brand color), but I'd be OK with any of these.

zdroid:
@waldyrious any of those is better. :smile: I like the last one (purple) the most .

andrik:
@waldyrious great work

Like you, I go on option 2.

waldyrious:
I've made the change to option 2 above, and took the opportunity to combine both commits into one.

zdroid:
> and took the opportunity to combine both commits into one.

@waldyrious May the force-push be with you!

Approved and merged.

sbrl:
Thanks @waldyrious! I love bright colours myself (anyone who's seen my website will know that lol) - so great to see some more vivid colours here :D

PR #4118, opened 2020-06-19
du: change --max-depth to -d

AZn5ReD:
Faster to type


zdroid:
I believe our general policy is to use the long form because it's more clear and easier to remember. E.g. if you want to change depth, you know it's `--depth` instead of thinking whether it's `-d`, `-D`, `-de` or whatever. What do you think?

AZn5ReD:
That's true, my contributions is out of the guidelines.
Thanks for pointing that out !  :+1:

PR #4121, opened 2020-06-21
Update zstd.md

Cyan4973:
Fix minor inaccuracies in `zstd` page:

### DIFF
--- a/pages/common/zstd.md
+++ b/pages/common/zstd.md
@@ -15,10 +15,10 @@
 
 `zstd -dc {{file}}.zst`
 
-- Compress a file, while specifing the compression level (0 being worst, 19 best, and 3 default):
+- Compress a file, while specifing the compression level (1=Fastest, 19=Strongest, 3=Default):
 
 `zstd -{{level}} {{file}}`
 
-- Use more memory (both when compressing and decompressing) to achieve a higher compression ratio:
+- Unlock higher compression levels (up to 22) using more memory (both for compression and decompression):
 
 `zstd --ultra -{{level}} {{file}}`
--- a/pages/common/zstd.md
+++ b/pages/common/zstd.md
@@ -15,7 +15,7 @@
 
 `zstd -dc {{file}}.zst`
 
-- Compress a file, while specifing the compression level (1=Fastest, 19=Strongest, 3=Default):
+- Compress a file specifying the compression level, where 1=fastest, 19=slowest and 3=default:
 
 `zstd -{{level}} {{file}}`
 


pages/common/zstd.md
```diff
@@ -15,10 +15,10 @@
 
 `zstd -dc {{file}}.zst`
 
-- Compress a file, while specifing the compression level (0 being worst, 19 best, and 3 default):
+- Compress a file, while specifing the compression level (1=Fastest, 19=Strongest, 3=Default):
```
zdroid, line 18:
```suggestion
- Compress a file, while specifing the compression level (1=Fastest, 19=Strongest, 3=Default):
```
Using Fastest and Strongest seems not appropriate for this example. How are those two comparable? I think I also advised against "1=Something, 2=Something, 3=Something" syntax in the original PR, and suggested the current one instead.

I would rather not change this line.

sbrl:
Looks good after @zdroid's comment is addressed

pages/common/zstd.md
Cyan4973: I disagree with the suggestion. "worst" and "best" are just abstract notions which do not convey any useful property.

zdroid: Well their documentation says it's the compression level. Changing the compression level changes the compression speed, not the other way around. So it ought to be "best" and "worst" (file quality), or something similar. Fastest and strongest doesn't make much sense in that context.

Now again, I am not saying that your suggestions are incorrect. They just won't be understood by most of tldr's users, which is what we want to avoid.

mebeim: I think we should do it like [the `gzip` page does ](https://github.com/tldr-pages/tldr/blob/master/pages/common/gzip.md), so either:

 1. `(1=fastest, 19=slowest, 3=default)`
 2. `(1=worst, 19=best, 3=default)`

Yes, "worst" and "best" are abstract notions, but contextualized to this specific command they are quite obviously indicating the worst or best possible compression level.

Also, I would remove the `, while` and just say `Compress a file specifying [...]`.

Cyan4973:
Sentence updated, following @mebeim suggestion

pages/common/zstd.md
zdroid: @mebeim do you want it to be inside parens (as you proposed), or outside them, like @Cyan4973 has done in the last change?

mebeim: @zdroid I think option 3 above is the clearest, so like this:

Compress a file specifying the compression level, where 1=fastest (worst), 19=slowest (best) and 3=default:

PR #4125, opened 2020-06-24
Update to use GitHub Actions for CI

### LINKED ISSUE #4097: Consider moving to GitHub Actions
It might be worth moving to GitHub Actions as has been discussed [previously on Gitter](https://gitter.im/tldr-pages/tldr?at=5ebc6bdd40e3122a8817d764). We've already moved the [Python client](https://github.com/tldr-pages/tldr-python-client) over to GH Actions in https://github.com/tldr-pages/tldr-python-client/pull/105. And it would also possibly make it easier to merge the [tldr-bot](https://github.com/tldr-pages/tldr-bot) code into the repository which would probably be faster (using the built in integrations).

I've started work on this and there are only a [few more variables](https://gitter.im/tldr-pages/tldr?at=5edf969d24a3382d5d56f48c) that are required for it to work.
https://github.com/tldr-pages/tldr/compare/master...owenvoke:feature/gh-actions?expand=1

- `TRAVIS_PULL_REQUEST`, this provides the id of the pull request and is empty when triggered by a non-PR. I've had a look and most things just say to parse the `GITHUB_REF` env variable instead.
- `TRAVIS_COMMIT_RANGE`, this provides the range of commits included in the PR or push. It's used only in the [deploy script commit messages](https://github.com/owenvoke/tldr/blob/100385f491ba133b3ee1a08191c65040e5674650/scripts/deploy.sh#L35).

---

Secrets can be added in [the settings](https://github.com/tldr-pages/tldr/settings/secrets), and we'd need to add three as far as I can tell. It's also worth noting that GitHub Actions doesn't allow lowercase env variable names, they are case-insensitive so the script would need to be updated. :+1:

- [`encrypted_973441be79af_key`](https://github.com/tldr-pages/tldr/blob/master/scripts/deploy.sh#L23) - used for decrypting the website repository deploy key
- [`encrypted_973441be79af_iv`](https://github.com/tldr-pages/tldr/blob/master/scripts/deploy.sh#L23) - used for decrypting the website repository deploy key
- `TRAVIS_BOT_GITHUB_TOKEN` - which looks like it's not being used (I think this was from [this script](https://github.com/tldr-pages/tldr/blame/2b0bc0be91019685476b1d13c23e5da831b0f839/scripts/post_pr_comment.py))

---

It may also be good to, once we've updated to use GitHub Actions, add some [scheduled workflows](https://help.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events-schedule) for tasks such as opening issues about contributor/maintainer/owner actions. See https://github.com/tldr-pages/tldr/issues/4113#issuecomment-647539108 for some more information.

owenvoke:
We need to add the website's deploy key to the [GitHub Action secrets](https://github.com/tldr-pages/tldr/settings/secrets). @agnivade, it looks like you added that deploy key, do you happen to still have it? If not, I guess we can add a new one? 🤔

Once we've added the deploy key there with a key of `DEPLOY_KEY`, then this should all (🤞) work as expected.

---

One other thing we need to remember to do (if this gets merged) is to remote the [Travis CI required status check in the protected branch settings](https://github.com/tldr-pages/tldr/settings/branches).

Closes #4097


zdroid:
> One other thing we need to remember to do (if this gets merged) is to **remote** the [Travis CI required status check in the protected branch settings](https://github.com/tldr-pages/tldr/settings/branches).

Did you mean **remove**?

Also, I think it'd be the best to just generate a new deploy key.

agnivade:
I don't remember very clearly now, but there is nothing I have for a deploy key. All that I have is a personal access token for tldr-bot which is used to comment on PRs. I think we can just generate a new key for this one.

owenvoke:
Great, I'll set up a new key later today. :+1:

owenvoke:
Deploy key added, all good to go. 🤞 Can anyone see any issues with this?

zdroid:
Works so far :smile:

mebeim:
Okay, so it seems to work fine, except that it seems that the "Deploy" step is being skipped ([see log here](https://github.com/tldr-pages/tldr/runs/811947276?check_suite_focus=true)):

screenshot

I am not sure why this happens since the check looked ok:

https://github.com/tldr-pages/tldr/blob/68f8e470d1c81100d95f4d5a1872711ae24733a9/.github/workflows/ci.yml#L30

zdroid:
@mebeim It should be `github.ref` instead of `github.branch`, that's why. I tried to fix it in https://github.com/tldr-pages/tldr/commit/d8e6c7e11fc4747b0296667232a2406f469a97ee, but it doesn't work: https://github.com/tldr-pages/tldr/runs/811991894. :/

mebeim:
@zdroid no, `github.ref` is not the branch name, it's the complete REF, for example `refs/pull/1234/merge` for PR `1234`.

zdroid:
This is what the reference says:

> github.ref | string | The branch or tag ref that triggered the workflow run.

Yeah, my mistake, I'm sorry for that. I got used to GitHub automatically making a new branch when I edit some file.

mebeim:
@zdroid you're right. It does not mention `branch`. I wonder where we could get that information from. It almost [looks like](https://stackoverflow.com/questions/58033366/how-to-get-current-branch-within-github-actions) there's no good way of doing it other than parsing the ref manually, though I'm not 100% sure. That's annoying.

I guess we could be better off creating two separate workflows, one for testing and one for deployment, since you can [specify the branch name](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-multiple-events-with-activity-types-or-configuration) in the `.yml`.

owenvoke:
Ah, I could have sworn there was a `github.branch` reference. 🤔

mebeim:
@owenvoke hmm, are we sure it's always going to be `refs/heads/master` though? If so, then I think it's fine.

PR #4126, opened 2020-06-24
unset: clarify example with no options

jn64:

### DIFF
--- a/pages/linux/unset.md
+++ b/pages/linux/unset.md
@@ -2,9 +2,13 @@
 
 > Remove shell variables or functions.
 
+- Remove the variable or function foo (only removes functions if the variable doesn't exist):
+
+`unset {{foo}}`
+
 - Remove the variables foo and bar:
 
-`unset {{foo}} {{bar}}`
+`unset -v {{foo}} {{bar}}`
 
 - Remove the function my_func:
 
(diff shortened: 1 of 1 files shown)


pages/linux/unset.md
```diff
@@ -2,9 +2,13 @@
 
 > Remove shell variables or functions.
 
+- Remove the variable or function foo (only removes functions if the variable doesn't exist):
```
sbrl, line 5: The bit in  brackets confuses me. Could you elaborate please?

zdroid: Functions and variables can have the same name, which is why that explanation is required.

jn64: If there is a variable foo, `unset foo` will remove it. If there is no variable foo, `unset foo` will look for a function foo and remove it.

The important point is that `unset foo` will _not_ remove a function foo if there is also a variable foo.

Here is another way it could be phrased, if you prefer:

> Remove the variable foo, or if the variable doesn't exist, remove the function foo

sbrl: Thanks, @jn64!

Yeah, that explanation looks clearer to me. Let's change it to that.

sbrl:
```suggestion
- Remove the variable `foo`, or if the variable doesn't exist, remove the function `foo`:
```

Issue #3953, opened 2020-03-30
Let's document! Git edition


sbrl:
For those not in the know, tldr-pages is about more than just commands. Those commands that are too expansive to fit on 1  page often have multiple subcommands - which we can document on their own pages. For example, try these:

```bash
tldr git merge
tldr git add
tldr ip address
```

In the case of git, it has quite a number of subcommands (139 ❗). Let's document them all!

- [x] [`add`](https://git-scm.com/docs/git-add) (#159)
- [x] [`am`](https://git-scm.com/docs/git-am) (#2352)
- [x] [`annotate`](https://git-scm.com/docs/git-annotate) (#5096)
- [x] [`apply`](https://git-scm.com/docs/git-apply) (#4252)
- [ ] [`archimport`](https://git-scm.com/docs/git-archimport) - this is for importing projects that use the [GNU Arch SCM](https://www.gnu.org/software/gnu-arch/), which is deprecated in favor of [GNU Bazaar](http://bazaar.canonical.com/)...should we still add this page?
- [x] [`archive`](https://git-scm.com/docs/git-archive) (#4362)
- [x] [`bisect`](https://git-scm.com/docs/git-bisect) (#1227)
- [x] [`blame`](https://git-scm.com/docs/git-blame) (#366)
- [x] [`branch`](https://git-scm.com/docs/git-branch) (#159)
- [x] [`bugreport`](https://git-scm.com/docs/git-bugreport) (#5097)
- [x] [`bundle`](https://git-scm.com/docs/git-bundle) (#4364)
- [x] [`cat-file`](https://git-scm.com/docs/git-cat-file) (#4492)
- [x] [`check-attr`](https://git-scm.com/docs/git-check-attr) (#5098)
- [x] [`check-ignore`](https://git-scm.com/docs/git-check-ignore) (#3113)
- [x] [`check-mailmap`](https://git-scm.com/docs/git-check-mailmap) (#5099)
- [x] [`check-ref-format`](https://git-scm.com/docs/git-check-ref-format) (#5100)
- [x] [`checkout`](https://git-scm.com/docs/git-checkout) (#159)
- [x] [`checkout-index`](https://git-scm.com/docs/git-checkout-index) (#5107)
- [x] [`cherry`](https://git-scm.com/docs/git-cherry) (#4473)
- [x] [`cherry-pick`](https://git-scm.com/docs/git-cherry-pick) (#1196)
[...]

PR #4127, opened 2020-06-25
git-difftool: add page

jn64:

for #3953

### DIFF
--- a/pages/common/git-difftool.md
+++ b/pages/common/git-difftool.md
@@ -1,6 +1,6 @@
 # git difftool
 
-> Show changes to files using external diff tools. Takes the same options and arguments as git diff.
+> Show file changes using external diff tools. Accepts the same options and arguments as git diff.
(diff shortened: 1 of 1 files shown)


pages/common/git-difftool.md
```diff
@@ -0,0 +1,20 @@
+# git difftool
+
+> Show changes to files using external diff tools. Takes the same options and arguments as git diff.
```
zdroid, line 3:
```suggestion
> Show file changes using external diff tools. Takes the same options and arguments as git diff.
```
What does the second sentence mean, though?

jn64: It uses the same options and arguments as `git diff`..?

If `git diff --foo bar` is valid, then `git difftool --foo bar` is also valid.

zdroid: I think `Accepts` would be more clear than `takes` then. Just my take.

PR #4138, opened 2020-06-30
history: list last n commands from the history

SirGlorg:

### DIFF
--- a/pages/common/history.md
+++ b/pages/common/history.md
@@ -6,9 +6,9 @@
 
 `history`
 
-- Display last 20 commands from the history list:
+- Display the last 20 commands:
 
-`history 20`
+`history {{20}}`
 
 - Clear the commands history list (only for current `bash` shell):
 
--- a/pages/common/history.md
+++ b/pages/common/history.md
@@ -6,6 +6,10 @@
 
 `history`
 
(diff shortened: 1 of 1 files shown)


pages/common/history.md
```diff
@@ -6,6 +6,10 @@
 
 `history`
 
+- Display last 20 commands from the history list:
+
+`history 20`
```
zdroid, line 11:
```suggestion
- Display the last 20 commands:

`history {{20}}`
```

einverne:
I have tested this command under Ubuntu, and I cannot make this `display the last 20 commands` work. Did I do something wrong?

zdroid:
@einverne for me `history 20` works fine. Did you miss to remove `{{ }}`?

SirGlorg:
@einverne https://www.gnu.org/software/bash/manual/html_node/Bash-History-Builtins.html documentation supports the `n` argument. I tested on macOs, also works. Can you check the `man history` or `history -h` and check, if Ubuntu's implementation happen to miss this option? I'd be surprised, but it does sometimes (like the `find` command, which I hate with passion for its varying behavior on different distributions).

einverne:
Maybe some problems with zsh, I just tested on another server it works.

PR #4144, opened 2020-07-02
pdftk: Avoid non-ascii characters

davidhusz:
The ellipsis character (U+2026) generates an encoding error in the web client (tldr.ostera.io): the UTF-8 input is interpreted as Latin-1, resulting in garbled output.


bl-ue:
>Unicode is great, but tldr-pages prefers ascii for the time being.

I know this is old, but I want to add: except for translations, right?

PR #4172, opened 2020-07-10
jq: Add example for constructing new JSON object

matt-watson90:
This example covers the functionality which `jq` provides to transform JSON input in to another JSON object. I've found this to be a common use case for `jq`.

This does bring the number of examples up to 9. In my experience this new example is a more common use case than the last two examples, so would propose that we remove one of those two.

### DIFF
--- a/pages/common/jq.md
+++ b/pages/common/jq.md
@@ -27,7 +27,7 @@
 
 `cat {{file.json}} | jq 'map(.{{key_name}})'`
 
-- Output the value of multiple keys as a new JSON object (assuming the input JSON has keys `key_name` and `other_key_name`):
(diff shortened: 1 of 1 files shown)


pages/common/jq.md
```diff
@@ -27,6 +27,10 @@
 
 `cat {{file.json}} | jq 'map(.{{key_name}})'`
 
+- Output the value of multiple keys as a new JSON object (assuming the input JSON has keys `key_name` and `other_key_name`):
```
sbrl, line 30: Perhaps this will read better?
```suggestion
- Output the value of multiple keys as a new JSON object (assuming the input JSON has the keys `key_name` and `other_key_name`):
```

matt-watson90:
@sbrl I have applied the suggestion :)

PR #4173, opened 2020-07-10
Update youtube-dl.md

tminei:
Adding download playlist as mp3.


sbrl:
Hey there, @tminei! Looks like you're having linter trouble. I suspect that you've got some rogue whiteespace that's getting in the way here.

PR #4174, opened 2020-07-10
youtube-dl: add download playlist example

tminei:
Adding download playlist as mp3.

### DIFF
--- a/pages/common/youtube-dl.md
+++ b/pages/common/youtube-dl.md
@@ -27,10 +27,6 @@
 
 `youtube-dl --format {{mp4}} -o {{"%(title)s by %(uploader)s on %(upload_date)s in %(playlist)s.%(ext)s"}} {{url}}`
 
-- List available subtitles:
-
-`youtube-dl --list-subs {{https://www.youtube.com/watch?v=Mwa0_nE9H7A}}`
-
 - Download a particular language's subtitles along with the video:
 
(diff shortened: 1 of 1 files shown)


tminei:
> There are already 8 examples, so one has to go. Which one is less informative than yours?
Hello, i think this example less useful, because there is already an example of quality choice

Download the best quality audio and video and merge them:
```
youtube-dl -f bestvideo+bestaudio {{url}}
```

zdroid:
I'd rather remove list formats or list subtitles examples. Those are not really needed for most use cases.

tminei:
> 
> 
> I think they used to be useful, but with the other great examples on this page I think the list formats might be ok to be removed.

Yeah, I guess so. I think you need to add the --all key to tldr to see the extended version of the manual with all the examples, and without that key to show 8 main examples, because 8 examples are quite small, and watching man is not always convenient, because of the curved style of its pages.

tminei:
I prefer to listen to YouTube music from the playlist, but I don't always have stable Internet and I would like to save the playlist to my device in mp3 format. I think that most Internet users don't have a high-speed connection, and for many of them my manual will be useful.

zdroid:
@sbrl So, which one to remove, list subtitles or list formats? I'd say list subtitles.

tminei:
> which one to remove, list subtitles or list formats? I'd say list subtitles.

I think formats are more important than subtitles.

zdroid:
I applied those changes @tminei. @sbrl you may wanna review this now

PR #4179, opened 2020-07-13
Added vboxmanage extpack command

giper45:


pages/linux/vboxmanage.md
```diff
@@ -27,3 +27,7 @@
 - Shutdown down the virtual machine without saving its state:
 
 `VBoxManage controlvm {{name|uuid}} poweroff`
+
+- Update VBox extension packs:
+
+`VBoxManage extpack install --replace Oracle_VM_VirtualBox_Extension_Pack-<version>.vbox-extpack`
```
sbrl, line 33: Is there anything that the user can change here? If so, perhaps we could enclose it in the `{{token_syntax}}`?

giper45: Replice is optional optional.
Oracle_VM... is the name if the extension pack downloaded by virtualbox site. Usually it contains suggested name, but it could be considered a variabile name so maybe it could  be better to use token syntax

sbrl: Ah, I see. Let's use the token syntax here then. I also note the `<version>` there too.

giper45: Version is included in extension name. Right version shoudl be: 
`VBoxManage extpack install --replace {{VboxExtensionFileNamePack}}`

PR #4188, opened 2020-07-15
uptime: add page

Bailey-1:
Added uptime.md to OSX platform since OSX doesn't support the arguments listed in common/uptime.md such as `--version`, `--pretty` and `--since`.

### DIFF
--- a/pages/osx/uptime.md
+++ b/pages/osx/uptime.md
@@ -1,0 +1,7 @@
+# uptime
+
+> Tell how long the system has been running and other information.
+
+- Print current time, uptime, number of logged-in users and other information:
+
+`uptime`


Bailey-1:
I don't know why the check hasn't run or isn't being listed under the checks tab. I don't think I did anything wrong but please let me know if I did. Thanks.

sahilister:
@Bailey-1 , switch indent size to 4 in your file instead of default 2.

I found that GitHub GUI editor defaults to 2 indent spaces and the CI here expects 4 indent spaces.

Bailey-1:
Hi @sbrl,
For some reason, uptime on OSX doesn't support any of the arguments that the version of uptime does on Linux such as `--version`, `--pretty` and `--since`. So I think it makes sense to create a seperate uptime.md file in `osx` which excludes the additional arguments since they still work great and are relevant in Linux but unfortunately do not work in OSX.

The man page for the command literally just lists the plain 'uptime' command and nothing more.
Screenshot 2020-07-16 at 23 05 32

Thanks.

owenvoke:
Regarding the CI, it looks like this may have been caused by the GitHub Actions downtime. Unfortunately I can't find a way to manually start the CI.

You could amend your commit and then force push it which should cause the CI to be triggered, or we can do this for you? 🤔

Bailey-1:
Hi @owenvoke,
That fixed it. Thanks.

Issue #4195, opened 2020-07-16
Typo: ack 


GermanS:
Hello

To show list all valid types for "ack" command should be 
`ack --help-types`
instead of
`ack --help=types`

here is the link to documentation
https://beyondgrep.com/documentation/

sbrl:
Fixed :-)

PR #4196, opened 2020-07-16
ack: fix typo

GermanS:
Hello

Please make a review for  GI #4195


sbrl:
Hey there, @GermanS! It would appear that according to the man page both are equally accepted:

```
       --help-types, --help=types
           Print all known types.
```

However, in other places in the man page confusingly `--help-types` seems to be preferred:

```
....
      In order to see all currently defined types, use --help-types, e.g.
       ack --type-set backup:ext:bak --type-add perl:ext:perl --help-types
....
```

So I think that `--help-types` is probably the one to prefer here, despite both of them being equally valid, as it's meaning is more clear.

PR #4197, opened 2020-07-16
pwd, ls, mkdir, rmdir, touch, rm, cd: add Traditional Chinese translation

fanfly:


fanfly:
I create pages written in Traditional Chinese, since Traditional Chinese characters, which are mainly used in Taiwan, are quite different from Simplified Chinese characters, which are mainly used in mainland China.
I would like to check if it is proper to create a new folder "pages.zh_TW", while the folder "pages.zh" already exists.

zdroid:
It actually may be a better idea to split these two, since we already have pt and pt_BR.

sbrl:
Oh sorry  @fanfly! I completely missed your comment yesterday - I was reviewing lots of PRs. Indeed, I think it's a great idea to split these into a separate language folder :D

PR #4204, opened 2020-07-17
git-diff: remove 9th example

### LINKED ISSUE #3753: git-diff: exceed the number of examples
In 6a85525 I translated [`git-diff`](https://github.com/tldr-pages/tldr/blob/master/pages/common/git-diff.md), a page where are 9 examples for the command. Should we remove one of them? Maybe.. the fifth? I refer to the one dedicated to show just the filenames of the changed files:

```markdown
- Show only names of changed files since a given commit:
git diff --name-only {{commit}}
```
This issue is very relevant for me due I am translating all the `git` commands to Spanish (#3727), so I need to know if I should keep this in the Spanish translation or if we should change the English page to remove one and then commit another change removing it for the translation I did.

zdroid:
Closes #3753.

I removed the example I found the least useful. The last two are common use cases.

### DIFF
--- a/pages.es/common/git-diff.md
+++ b/pages.es/common/git-diff.md
@@ -27,10 +27,6 @@
 
 `git diff --summary {{commit}}`
 
-- Crea un archivo parche:
-
-`git diff > {{archivo_objetivo}}.patch`
-
 - Compara un único archivo entre dos ramas o commits:
 
 `git diff {{rama_1}}..{{rama_2}} [--] {{ruta/del/archivo}}`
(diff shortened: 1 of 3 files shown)


zdroid:
@sbrl I was just thinking which example is going to be used the least. I for one have never needed a patch file, while I've needed pretty much everything else to fix something in Git.

sbrl:
Depends which repos you contribute to, I guess. If it's GH, then you're all good. If not however, you might need to send patch files around (e.g. I believe development of the Linux Kernel works like this, if I'm not mistaken).

We can remove if you like though.

zdroid:
@sbrl I'd still go with removing it, but if you have some better idea, you can edit the PR directly. @agnivade said "I'm okay with removing the 5th or any of the last 2 examples" in the previous PR, so it's not a problem.

PR #4208, opened 2020-07-18
pkgutil: add --expand-full

notpeter:


pages/osx/pkgutil.md
```diff
@@ -13,3 +13,7 @@
 - List all the files for an installed package given its ID:
 
 `pkgutil --files {{com.microsoft.Word}}`
+
+- Extract the contents of a packge file into a directory:
+
+`pkgutil  --expand-full {{filename.pkg}} {{directory}}`
```
sbrl, line 19: Perhaps this would be more consistent with other pages?
```suggestion
`pkgutil  --expand-full {{path/to/filename.pkg}} {{path/to/directory}}`
```

pages/osx/pkgutil.md
```diff
@@ -13,3 +13,7 @@
 - List all the files for an installed package given its ID:
 
 `pkgutil --files {{com.microsoft.Word}}`
+
+- Extract the contents of a packge file into a directory:
```
sbrl, line 17: Typo?
```suggestion
- Extract the contents of a package file into a directory:
```

zdroid:
Now there are 9 instead of 8 examples, that's one too much. Which one should go?

notpeter:
> Perhaps this would be more consistent with other pages?
> `pkgutil  --expand-full {{path/to/filename.pkg}} {{path/to/directory}}`

@sbrl I know this is in the style guide, but I personally find it less clear when there are 2+ parameters. Whatevs though, I fixed it here and in the earlier examples on L11 so it's "consistent". Also, "more consistent" is relative: there are nearly 300 examples of it in other pages: `{{filename` without pathing: `grep  -R '{{filename' . `

----
> Now there are 9 instead of 8 examples, that's one too much. Which one should go?

@zdroid What are you talking about? There are 4 examples.

zdroid:
@notpeter My previous comment looks like a bug, I deleted it now. It was intended for an other issue, not this one.

PR #4210, opened 2020-07-19
grpcurl: example fix

zkosanovic:
My OCD is tingling, I think the brackets should go this way (around the quotation marks).


zkosanovic:
Ahh, so the curly braces denote the optional part? Then I guess I should leave them, but apply the same logic to the body (`-d` parameter).

zkosanovic:
Nvm 😄

PR #4216, opened 2020-07-22
ps: Added sort example

ufobat:

### DIFF
--- a/pages/common/ps.md
+++ b/pages/common/ps.md
@@ -25,3 +25,7 @@
 - Get the parent pid of a process:
 
 `ps -o ppid= -p {{pid}}`
+
+- Sort processes according to memory consumption:
+
+`ps --sort size`
--- a/pages/common/ps.md
+++ b/pages/common/ps.md
@@ -26,6 +26,6 @@
 
 `ps -o ppid= -p {{pid}}`
 
-- Sort processes according to memory consumption:
+- Sort processes by memory consumption:
 
 `ps --sort size`


pages/common/ps.md
```diff
@@ -25,3 +25,7 @@
 - Get the parent pid of a process:
 
 `ps -o ppid= -p {{pid}}`
+
+- Sort processes according to memory consumption:
```
zdroid, line 29:
```suggestion
- Sort processes by memory consumption:
```

pages/common/ps.md
```diff
@@ -25,3 +25,7 @@
...
+
+`ps --sort size`
```
einverne, line 31: This command shows some different on my computer(Ubuntu 18.04 with Guake terminal)

```
➜ ps --sort size
  PID TTY          TIME CMD
17359 pts/30   00:00:00 ps
13722 pts/30   00:00:01 zsh
```

Is this what we expected?

PS. I have a little suggestion, we can sort memory here, how about the cpu usage and other?

ufobat: I am happy that you ask.

`ps --sort=size` sorts it, but doesn't print the size. There are different memory size definitions in for the output of ps. The paramters `-F` will also display the size. (There are more parameters that disalay the size or the size in a different defintion, as well). With the parameter `-o` you can specify that you want to see the key youre sorting for, but only this one:
```
> ps --sort size -o size 
 SIZE
 1020
 3040
```

Of course you can sort by different keys as well. `man ps` gives a huge amount of different sort keys in this section:

```STANDARD FORMAT SPECIFIERS
       Here are the different keywords that may be used to control the output format (e.g., with option
       -o) or to sort the selected processes with the GNU-style --sort option.
```

I would be happy to add more sort keys to my example, but I don't in what way. (Because `-F` might again not display the value)

Any advice?

einverne: hi @ufobat Thank you for your adequate explanation, I just curious when will you use `ps --sort size` and what information you want to check? 
By the way, this example looks good to me.

ufobat: My PC does not have a huge amount of memory, sometimes it starts to do some swapping. In that case I want to figure out which processes are consuming the most memory so I can decide to whether I can terminate them. So I am actually doing `ps aux --sort size | tail -n 15`

einverne: Thank you @ufobat , this use case is much helpful, I personally use the `top`, but I think `ps aux --sort size` may be more efficient.

PR #4219, opened 2020-07-22
netselect: add page

sahilister:


pages/linux/netselect.md
```diff
@@ -0,0 +1,24 @@
+# netselect
+
+> Speed test for choosing a fast network server.
+> More information: <https://github.com/apenwarr/netselect> .
+
+- List available options:
```
sbrl, line 6: Perhaps this would be best as the last option? Generally we put examples in the order you might use  them - so I'd assume that the user would want to use example #2 first instead of this one.

sahilister:
@sbrl , "fastest" here meaning server with the lowest latency.

pages/linux/netselect.md
sahilister: Right, I'll change the order.

pages/linux/netselect.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Display nameserver resolution and statistics:
+
+`sudo netselect -vv {{host_1}} {{host_2}}`
+
+- Define maximum TTL (time to live):
+
+`sudo netselect -m {{10}} {{host_1}} {{host_2}}`
+
+- Print this many fastest servers:
```
einverne, line 18: This description is a little bit confusing, what does the option `-s` do here?

sahilister: `-s` option prints  "x" fastest servers amongst "n" servers tested. 

"x" here is the number entered with `-s`.

pages/linux/netselect.md
```diff
@@ -0,0 +1,24 @@
...
+
+`sudo netselect -s {{2}} {{host_1}} {{host_2}} {{host_3}}`
```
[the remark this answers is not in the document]
sahilister: This seems nice, thanks @einverne

pages/linux/netselect.md
```diff
@@ -0,0 +1,24 @@
+# netselect
+
+> Speed test for choosing a fast network server.
+> More information: <https://github.com/apenwarr/netselect> .
+
+- Choose the fastest server:
```
sbrl, line 6: Thanks for the clarification, and sorry it's taken me so long to get back to you. Perhaps this will read better?
```suggestion
- Choose the server with the lowest latency:
```

sahilister: This seems on point. 

Thanks!

PR #4220, opened 2020-07-22
node: add --inspect-brk example

tkolleh:

### DIFF
--- a/pages/common/node.md
+++ b/pages/common/node.md
@@ -5,7 +5,7 @@
 
 - Run a JavaScript file:
 
-`node {{file}}.js`
+`node {{path/to/file}}`
 
 - Start a REPL (interactive shell):
 
@@ -19,6 +19,6 @@
 
 `node -p "{{process.versions}}"`
 
-- Enable Node's inspector agent and pause execution until a debugger is connected. Source code is fully compiled:
+- Enable Node's inspector agent and pause execution until a debugger is connected. Prevent delayed (lazy) parsing of source code:
 
-`node --no-lazy --inspect-brk {{file}}.js`
+`node --no-lazy --inspect-brk {{path/to/file}}`
--- a/pages/common/node.md
+++ b/pages/common/node.md
@@ -18,3 +18,7 @@
 - Evaluate and print result, useful to see node's dependencies versions:
 
 `node -p "{{process.versions}}"`
+
+- Enable Node's inspector agent and pause execution until a debugger is connected. Source code is fully compiled:
+
+`node --no-lazy --inspect-brk {{file}}.js`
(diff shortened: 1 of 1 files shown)


pages/common/node.md
```diff
@@ -18,3 +18,7 @@
 - Evaluate and print result, useful to see node's dependencies versions:
 
 `node -p "{{process.versions}}"`
+
+- Enable Node's inspector agent and pause execution until a debugger is connected. Source code is fully compiled:
+
+`node --no-lazy --inspect-brk {{file}}.js`
```
sbrl, line 24: This would improve consistency with other pages :-)
```suggestion
`node --no-lazy --inspect-brk {{path/to/file.js}}`
```

Also, files don't necessarily _have_ to end in `.js` :P

tkolleh: Would you accept the same changes to the other file path reference? If so i'll update the document for consistency. 

I agree with the recommended changes. My commit used `{{file}}.js` to remain consistent with the document. Reason being, since its the [accepted (prod)](https://github.com/tldr-pages/tldr/commit/856c3d1535bfa95fdd757937c3d9911bef292d8c) approach - reuse what's there. Yes, no filename is required e.g. `node {{file}}` is accurate.

pages/common/node.md
```diff
@@ -18,3 +18,7 @@
 - Evaluate and print result, useful to see node's dependencies versions:
 
 `node -p "{{process.versions}}"`
+
+- Enable Node's inspector agent and pause execution until a debugger is connected. Source code is fully compiled:
```
agnivade, line 22: What does "Source code is fully compiled" mean in this context? Can we explain it a bit more?

pages/common/node.md
tkolleh: See. Latest commit.

pages/common/node.md
tkolleh: Node.js parses the functions inside JavaScript files on first access. The option prevents the delayed parsing and ensures that breakpoints can be validated before running the code. I've clarified this in the latest commit.

agnivade: Sure, but is this something somebody needs to know to use the command? Feels like it makes the description unncessarily long without adding much value.

tkolleh: It is not required for debugging. I would recommend no lazy and VSCode defaults to using it because breakpoints don't work in source code areas that haven't been seen (parsed) by Node.js. Since this isn't something found in the man page `man node` I thought to add it.

agnivade: Ah okay, now I understand that this is tied to the --no-lazy option. Yeah okay, this makes sense. But let's change the description to include this in a single sentence. Otherwise it reads like something extra and disconnected from the `--no-lazy` option.

tkolleh: See latest commit.

PR #4221, opened 2020-07-23
git-stash: Add command to display changes in a stash

postrational:

### DIFF
--- a/pages/common/git-stash.md
+++ b/pages/common/git-stash.md
@@ -19,7 +19,11 @@
 
 `git stash list`
 
-- Apply a stash (default is the latest, named stash@{0}):
+- Display changes in a stash (default is the latest, named stash@{0}):
+
+`git stash show -p {{optional_stash_name_or_commit}}`
+
+- Apply a stash (default is stash@{0}):
 
 `git stash apply {{optional_stash_name_or_commit}}`
 
--- a/pages/common/git-stash.md
+++ b/pages/common/git-stash.md
@@ -23,13 +23,9 @@
 
 `git stash show -p {{optional_stash_name_or_commit}}`
 
-- Apply a stash (default is stash@{0}):
+- Apply a stash (default is stash@{0}), use `stash pop` to apply the stash and remove it from the stash list:
 
 `git stash apply {{optional_stash_name_or_commit}}`
-
-- Apply a stash (default is stash@{0}), and remove it from the stash list if applying doesn't cause conflicts:
-
-`git stash pop {{optional_stash_name}}`
 
 - Drop a stash (default is stash@{0}):
 


postrational:
Personally I use `tldr` to look for non-obvious command. I think that `stash show -p` is such a command, and I'd love to see it here.

How about combining `stash apply` and `stash pop` into one example? I added a commit with this approach.

agnivade:
I think that is essentially trying to cram everything to fit the limit. It is indeed a challenge accomodating everyone's favorite option on a tldr page. That's why we only use the most common ones that someone will normally use in their day to day usage of the command.

A simple litmus test can be that among the 3 commands - stash apply, stash pop and stash show, which are the two you use most frequently. For me personally, it is only apply and pop. I don't need to actually see a stashed commit, because usually I know what is stashed by seeing the commit message.

einverne (commented):
Thank you @postrational for this new patch. The command to show `changes in a stash` is useful. But I think the `git stash pop` command also important. It is hard to choose. I would like to keep the origin one.

postrational:
Well, I never forget what `git stash apply` or `git stash pop` do, or how to use them. But I always forget the `-p` switch to `git show` and have to look it up. tldr seems like the perfect place to store this reminder.

agnivade (changes requested):
> Well, I never forget what git stash apply or git stash pop do, or how to use them. But I always forget the -p switch to git show and have to look it up. tldr seems like the perfect place to store this reminder.

That's a fair point. But I think we are forgetting the viewpoint of beginners who have heard of "git stash" the first time and trying to understand what that command does. At a very basic level, tldr pages should contain the most used commands in day to day usage starting from the the simplest form to an advanced invocation. If there is still space left after that, we can accomodate other commands which aren't that easy to remember.

But from a priority viewpoint, `stash pop` is far more commonly used than `stash show -p`. Yes, it might be easy to remember, but that's only because you have learnt and used the command for so long that you don't have to think. But a tldr page is for all levels of people who might be trying to familiarize with the command for the first time. Thinking from that perspective, we should keep `stash pop` IMO.

Nevertheless, thank you for spending your time to send the change. Please do send PRs for any other page that you think needs improvement. :)

einverne (approved):
The new patch is the best solution, follow the guideline and introduce the new thing. Great thanks @postrational .

postrational:
Is there any chance this PR will be merged?

agnivade:
Hi @postrational - I have laid out my opinion about the PR above. However, it seems like others are in favor. So I think it should be good to merge.

navarroaxel:
Yeah, it's a complex command but I think the changes are good. Looks like this is up to you @agnivade :grin:

agnivade:
I still stand by my original opinion. It seems to go against the spirit of tldr pages to include a more complex command just because it's hard to remember, and exclude easy ones because they are simple.

PR #4246, opened 2020-08-03
Client spec: require exiting with non-zero code on missing page

### LINKED ISSUE #4227: tldr should not return 0 when a TLDR page is not found
## Description

When `tldr` doesn't find a TLDR page, it should not return 0.

## Steps to reproduce

Feed `tldr` with a trainspotting-powered argument, observe its return code.

## Expected result

`tldr` does not return 0

## Actual restult

`tldr` return 0 whatever you feed it with.

## Example 

```
$ tldr Choose a life. Choose a job. Choose a career. Choose a family.  2> /dev/null && echo true          
true
```

## Why would this been addressed

For instance, when a page is not found in your [$LANG language](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#language), this would allow to fallback to `en` (which supposedly has more references, being [the default language](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#default-language-for-newly-added-pages)) , such as:

```shell
tldr ss || tldr --language en ss
```
### Actual result

```
$ tldr ss || tldr --language en ss
`ss` documentation is not available. Consider contributing Pull Request to https://github.com/tldr-pages/tldr
```
### Expected result

```
$ tldr ss || tldr --language en ss

  ss

  Utility to investigate sockets.
[...]
```

MasterOdin:
This updates the specification to include a recommendation on exiting with a non-zero exit code in addition to showing an error message, which is helpful for scripts and things that may be consuming this tool.

---------------------

Edit by @sbrl: Fixes #4227

### DIFF
--- a/CLIENT-SPECIFICATION.md
+++ b/CLIENT-SPECIFICATION.md
@@ -164,7 +164,7 @@ If a page cannot be found in _any_ platform, then it is RECOMMENDED that clients
 https://github.com/tldr-pages/tldr/issues/new?title=page%20request:%20{command_name}
 ```
 
-where `{command_name}` is the name of the command that was not found. It is RECOMMENDED that clients exit with a non-zero exit code in addition to showing the above message.
+where `{command_name}` is the name of the command that was not found. Clients that have control over their exit code on the command line (i.e. clients that provide a CLI) MUST exit with a non-zero exit code in addition to showing the above message.
 
 #### If multiple versions of a page were found
 
--- a/CLIENT-SPECIFICATION.md
+++ b/CLIENT-SPECIFICATION.md
@@ -221,6 +221,9 @@ Caching SHOULD be done according to the user's language configuration (if any), 
 
 ## Changelog
 
+ - [v1.4, August 13th 2020](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md) (#4246)
+   - Add requirement for CLI clients to use non-zero exit code on failing to find a page.
+
  - [v1.3, June 11th 2020](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md) (#4101)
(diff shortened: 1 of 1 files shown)


CLIENT-SPECIFICATION.md
```diff
@@ -164,7 +164,7 @@ If a page cannot be found in _any_ platform, then it is RECOMMENDED that clients
 https://github.com/tldr-pages/tldr/issues/new?title=page%20request:%20{command_name}
 ```
 
-where `{command_name}` is the name of the command that was not found.
+where `{command_name}` is the name of the command that was not found. It is RECOMMENDED that clients exit with a non-zero exit code in addition to showing the above message.
```
sbrl, line 167: I think we should word this more strongly. What about this?
```suggestion
where `{command_name}` is the name of the command that was not found. Clients that have control over their exit code on the command line (i.e. clients that provide a CLI) MUST exit with a non-zero exit code in addition to showing the above message.
```

This helps to provide a consistent scripting interface (e.g. for tools that detect whether pages exist or not).

MasterOdin: I agree with making it mandatory, though I had made it just recommended as I'm unsure how many clients would need changing to follow this. If no one has objections, I'll apply this change tomorrow.

sbrl: Right - good point. Still, I think having a requirement is a good idea. Please can you bump the version of the spec too? This might be considered a 'breaking' change to the spec, since we require something new.

MasterOdin: Committed changes, and bumped the version of the specification at the bottom.

sbrl:
Are we good to merge this?

waldyrious (approved):
LGTM — thanks for taking this on!

PR #4263, opened 2020-08-14
Update crontab.md

tminei:

### DIFF
--- a/pages/common/crontab.md
+++ b/pages/common/crontab.md
@@ -2,7 +2,7 @@
 
 > Schedule cron jobs to run on a time interval for the current user.
 > Job definition format: "(min) (hour) (day_of_month) (month) (day_of_week) command_to_execute".
-> To make no mistake, use [crontab guru](https://crontab.guru).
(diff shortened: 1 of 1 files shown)


pages/common/crontab.md
```diff
@@ -2,6 +2,7 @@
 
 > Schedule cron jobs to run on a time interval for the current user.
 > Job definition format: "(min) (hour) (day_of_month) (month) (day_of_week) command_to_execute".
+> To avoid mistake, use [crontab guru](https://crontab.guru).
```
agnivade, line 5: While this is helpful, I am not sure content like this falls within the scope of tldr pages. This is more appropriate in other developer content related websites IMO.

Kreyren: I would suggest placing it in tldr pages as it just makes everything so much easier.

zdroid: I think it's not our job to promote any specific tool, especially not on other tool's page. If crontab guru has a CLI tool, it can be made into its own page.

PR #4266, opened 2020-08-15
java: add debug info

maxandersen:

### DIFF
--- a/pages/common/java.md
+++ b/pages/common/java.md
@@ -11,6 +11,10 @@
 
 `java -jar {{filename.jar}}`
 
+- Execute a .jar program with debug waiting to connect on port 5005: 
+
+`java -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005 -jar {{filename.jar}}`
+
 - Display JDK, JRE and HotSpot versions:
 
 `java -version`
--- a/pages/common/java.md
+++ b/pages/common/java.md
@@ -11,7 +11,7 @@
 
 `java -jar {{filename.jar}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/java.md
```diff
@@ -11,6 +11,10 @@
 
 `java -jar {{filename.jar}}`
 
+- Execute a .jar program with debug waiting to connect on port 5005:
```
agnivade, line 14: I think this is a fairly advanced usage of the command, which unfortunately makes it out of scope for a tldr page.

maxandersen: its the one thing you would need when using it from command line.

perl and gcc has info on how to do use and enable debug. 

unfortunately in java the line is just quite long and verbose; but thats why it qualifies to be in tldr as noone remembers that by heart imo.

sbrl: /cc @agnivade - I think this one has slipped through the net

maxandersen:
signed the cla - anything else I need to do ?

pages/common/java.md
agnivade: Well, I think I still respectfully disagree that this needs to be added. But if you think that this is fine, then I have no objections.

navarroaxel (approved):

pages/common/java.md
navarroaxel, line 14: I think the dot in `.jar` is not well received for the linter
```suggestion
- Execute a jar file program with debug waiting to connect on port 5005: 
```

navarroaxel:
I fixed a trialing space in the example description

PR #4271, opened 2020-08-16
ix: Add page

Kreyren:
I hope to provide this page to be used as a quick reference for a new users who are not confident in CS yet to provide the expected outputs for diagnostics.

	- Can be used crossplatform -> Concluded that `common/` is the place for it

### DIFF
--- a/pages/common/ix.md
+++ b/pages/common/ix.md
@@ -1,0 +1,16 @@
+# ix
+
+> An alias used for ix.io pastebin service.
+> More information: <http://ix.io/>.
+
+- Define an alias for ix.io:
+
+`alias ix="curl -F 'f:1=<-' ix.io"`
+
+- Add the alias in bashrc and re-source it:
+
+`printf '%s\n' "alias ix=\"curl -F 'f:1=<-' ix.io\"" >> "$HOME/.bashrc" && source "$HOME/.bashrc"`
+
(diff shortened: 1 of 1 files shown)


einverne (commented):
Thank you @Kreyren . The command `ix` looks cool. But I am not sure it is a real cli because you can alias it to anything you like. So I leave it for others to review.

owenvoke:
@einverne, I agree with your review. 🤔 It seems like it can be aliased to anything and isn't really a CLI tool.

Does seem pretty neat as a command though. 👍🏻

waldyrious:
Hello, @Kreyren, and thanks for contributing this page! Unfortunately, and as the others alluded to above, I'm afraid it does not fit the goals of the tldr pages project. The first two entries are essentially installation instructions, which we don't typically cover in tldr pages (not in actual examples, that is; a short note in the main description is OK if it's concise, e.g. "available in the X repository", or "installable using the Y tool"). The only actual usage example —the third one in the page— is quite trivial, so it's arguable that a tldr page isn't needed for it.

And finally, this is not a command-line tool, but more of a neat trick to interact with a web service. This is not by itself a deal-breaker, but coupled with the above, it adds extra weight to the argument that it doesn't fit this project. Thank you, regardless, for taking the time to submit it. :)

PR #4287, opened 2020-08-23
adb-shell: add page

navarroaxel:

### DIFF
--- a/pages/common/adb-shell.md
+++ b/pages/common/adb-shell.md
@@ -7,9 +7,9 @@
 
 `adb shell`
 
-- Reset dangerous permissions for an application:
+- Revert all runtime permissions to their default:
 
-`adb shell pm reset-permissions {{package}}`
+`adb shell pm reset-permissions`
 
 - Revoke a dangerous permission for an application:
 
--- a/pages/common/adb-shell.md
+++ b/pages/common/adb-shell.md
@@ -1,0 +1,32 @@
+# adb shell
+
+> Android Debug Bridge Shell: Run remote shell command with an Android emulator instance or connected Android devices.
+> More information: <https://developer.android.com/studio/command-line/adb>.
+
+- Start a remote interactive shell in the emulator/device:
+
+`adb shell`
+
+- Reset dangerous permissions to an application:
+
+`adb shell pm reset-permissions {{package}}`
+
+- Revoke a dangerous permission to an application:
+
+`adb shell pm revoke {{package}} {{permission}}`
+
+- Trigger a key event:
+
+`adb shell input keyevent {{keycode}}`
+
+- Clear the data from an emulator/device's application:
+
+`adb shell pm clear {{package}}`
+
+- Start an activity on emulator/device:
+
+`adb shell am start -n {{package}}/{{activity}}`
+
+- Start home activity on emulator/device:
+
+`adb shell am start -W -c android.intent.category.HOME -a android.intent.action.MAIN`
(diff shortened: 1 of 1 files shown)


pages/common/adb-shell.md
```diff
@@ -0,0 +1,32 @@
+# adb shell
+
+> Android Debug Bridge Shell: Run remote shell command with an Android emulator instance or connected Android devices.
```
sbrl, line 3: Perhaps this would be more consistent with other existing pages?
```suggestion
> Android Debug Bridge Shell: Run remote shell commands on an Android emulator instance or connected Android devices.
```

pages/common/adb-shell.md
```diff
@@ -0,0 +1,32 @@
...
+> More information: <https://developer.android.com/studio/command-line/adb>.
+
+- Start a remote interactive shell in the emulator/device:
```
sbrl, line 6: Perhaps this would be more consistent with other existing pages?
```suggestion
- Start a remote interactive shell on the emulator/device:
```

pages/common/adb-shell.md
```diff
@@ -0,0 +1,32 @@
...
+
+`adb shell`
+
+- Reset dangerous permissions to an application:
```
sbrl, line 10: I'm not quite sure what you mean here - it's a little unclear. Do you mean `for` instead of `to` here?
```suggestion
- Reset dangerous permissions for an application:
```

pages/common/adb-shell.md
```diff
@@ -0,0 +1,32 @@
...
+
+`adb shell pm reset-permissions {{package}}`
+
+- Revoke a dangerous permission to an application:
```
sbrl, line 14: Same here?
```suggestion
- Revoke a dangerous permission for an application:
```

pages/common/adb-shell.md
```diff
@@ -0,0 +1,32 @@
...
+
+`adb shell pm revoke {{package}} {{permission}}`
+
+- Trigger a key event:
+
+`adb shell input keyevent {{keycode}}`
+
+- Clear the data from an emulator/device's application:
```
sbrl, line 22: Perhaps this will read better?
```suggestion
- Clear the data of an application on an emulator or device:
```

pages/common/adb-shell.md
```diff
@@ -0,0 +1,32 @@
...
+
+`adb shell pm clear {{package}}`
+
+- Start an activity on emulator/device:
+
+`adb shell am start -n {{package}}/{{activity}}`
+
+- Start home activity on emulator/device:
```
sbrl, line 30: Perhaps this will read better?
```suggestion
- Start the home activity on an emulator or device:
```

Also, is this the main home screen, or the home screen of a specific application? If it's the latter, then perhaps something like this would be better?
```suggestion
- Start the home activity for an application on an emulator or device:
```

Finally, are there any long forms of the arguments used here? We generally prefer long forms of arguments in tldr pages.

sbrl (changes requested):
Thanks for the new page, @navarroaxel! I've left some comments below for you to review.

pages/common/adb-shell.md
navarroaxel: Is the Android home.

adb has only short forms for most arguments

navarroaxel:
I used the shorthands to run `am`, `pm` and `input` commands using the `adb shell`.
But if the user starts an interactive shell in the android emulator/device with `adb shell`, the user can use `am` ([activity manager](https://developer.android.com/studio/command-line/adb#am)), `pm` ([package manager](https://developer.android.com/studio/command-line/adb#pm)) and `input` directly inside the android shell itself.

Should we add this pages too?

sbrl:
Pages for those subcommands would be cool! But I think this page is fine as-is, so maybe those could be a separate PR?

PR #4299, opened 2020-08-27
jq: correct description of example

johnsmith2077:
From the latest [jq manual](https://stedolan.github.io/jq/manual/#Basicfilters), it's said that when `use '.[]' on an object, and it will return all the values of the object`, but not all key-value pairs from objects


einverne (approved):
LGTM.

PR #4296, opened 2020-08-27
resize2fs: add page

samuel-w:
- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [ ] The page description includes a link to documentation or a homepage (if applicable).

Relates to #4250

PR #4305, opened 2020-08-31
e4defrag: fix grammar

samuel-w:
Trivial fix, based on this discussion: https://github.com/tldr-pages/tldr/pull/4296#discussion_r479833047


sbrl (approved):
Thanks so much for coming back and adjusting this :D

PR #4309, opened 2020-09-01
recsel: add page

BonfaceKilz:

### DIFF
--- a/pages/common/recsel.md
+++ b/pages/common/recsel.md
@@ -1,0 +1,16 @@
+# recsel
+
+> Print records from a human-editable, plain text database called a recfile.
+> More information <https://www.gnu.org/software/recutils/manual/recutils.html>.
+
+- Extract name and version field from the output of guix:
+
+`guix package -s malloc | recsel -p name,version`
+
+- Use "~" to match a string with a given regular expression:
+
+`guix package -s python | recsel -e "name  ~ 'numpy$'"`
+
+- Use a predicate to match a name and a version:
+
+`guix search python | recsel -e "name ~ 'numpy$' && version ~ '^1\.1.*'"`
--- a/pages/common/recsel.md
+++ b/pages/common/recsel.md
@@ -1,16 +1,16 @@
 # recsel
 
-> Print records from a human-editable, plain text database called a recfile.
+> Print records from a recfile: a human-editable, plain text database.
 > More information <https://www.gnu.org/software/recutils/manual/recutils.html>.
 
-- Extract name and version field from the output of guix:
+- Extract name and version field:
 
-`guix package -s malloc | recsel -p name,version`
+`recsel -p name,version {{data.rec}}`
 
 - Use "~" to match a string with a given regular expression:
 
-`guix package -s python | recsel -e "name  ~ 'numpy$'"`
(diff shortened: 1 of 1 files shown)


pages/common/recsel.md
```diff
@@ -0,0 +1,16 @@
+# recsel
+
+> Print records from a human-editable, plain text database called a recfile.
+> More information <https://www.gnu.org/software/recutils/manual/recutils.html>.
```
sbrl, line 4: Perhaps this will read better?
```suggestion
> Print records from a recfile: a human-editable, plain text database.
> See also `guix`.
> More information <https://www.gnu.org/software/recutils/manual/recutils.html>.
```

pages/common/recsel.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Extract name and version field from the output of guix:
```
sbrl, line 6: Perhaps this will read better?
```suggestion
- Extract name and version field from the output of `guix`:
```

pages/common/recsel.md
```diff
@@ -0,0 +1,16 @@
...
+
+`guix package -s malloc | recsel -p name,version`
```
sbrl, line 8: What can be changed here? We should wrap it in the `{{token_syntax}}` so that tldr clients know and can highlight it as such.

Same goes for all the other examples.

Also, generally we don't showcase the usage of more than 1 command on a page. e.g. on this page we would _only_ have examples of the `recsel` command. If possible, it would be great to refactor this page a bit there, but if not it should be fine. This seems like an exception if 1 command can't be used without the other :-)

sbrl, line 8: What can be changed here? We should wrap it in the `{{token_syntax}}` so that tldr clients know and can highlight it as such.

Same goes for all the other examples.

Also, generally we don't showcase the usage of more than 1 command on a page. e.g. on this page we would _only_ have examples of the `recsel` command. If possible, it would be great to refactor this page a bit there, but if not it should be fine. This seems like an exception if 1 command can't be used without the other :-)

sbrl (changes requested):
Thanks for the new page, @BonfaceKilz! I've left some comments below for you to review.

pages/common/recsel.md
BonfaceKilz: I agree with that :+1: I'll add commands for `guix` later; I noticed there are no tldr pages for that.

BonfaceKilz: I've removed the `guix` examples, so now the command uses a "{{data.rec}}" which I think is more clearer. So we won't need *See also `guix`*

pages/common/recsel.md
BonfaceKilz: I've removed all references from guix and used a *{{data.rec}}* instead.

BonfaceKilz: So I've removed all references for `guix` since you can definitely use `recsel` without `guix`.

BonfaceKilz:
@sbrl I've updated the PR. Wdyt? Also, thanks for reviewing this!

sbrl (approved):
Looks great! Thanks, @BonfaceKilz :D

I've just added a missing colon I noticed too.

Now we just need a 2nd reviewer to give the ok.

navarroaxel (approved):

owenvoke (approved):

PR #4325, opened 2020-09-07
sort: add --general-numeric-sort example

hntee:

### DIFF
--- a/pages/common/sort.md
+++ b/pages/common/sort.md
@@ -29,3 +29,7 @@
 - Sort human-readable numbers (in this case the 5th field of `ls -lh`):
 
 `ls -lh | sort -h -k 5`
+
+- Sort numbers with exponents:
+
+`sort -g {{filename}}`
--- a/pages/common/sort.md
+++ b/pages/common/sort.md
@@ -5,19 +5,19 @@
 
 - Sort a file in ascending order:
 
-`sort {{filename}}`
+`sort {{path/to/file}}`
 
 - Sort a file in descending order:
 
-`sort -r {{filename}}`
+`sort -r {{path/to/file}}`
 
 - Sort a file in case-insensitive way:
 
-`sort --ignore-case {{filename}}`
+`sort --ignore-case {{path/to/file}}`
 
 - Sort a file using numeric rather than alphabetic order:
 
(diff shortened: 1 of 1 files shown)


pages/common/sort.md
```diff
@@ -29,3 +29,7 @@
 - Sort human-readable numbers (in this case the 5th field of `ls -lh`):
 
 `ls -lh | sort -h -k 5`
+
+- Sort numbers with exponents:
+
+`sort -g {{filename}}`
```
navarroaxel, line 35: we prefer the long argument:
```suggestion
`sort --general-numeric-sort {{path/to/file}}`
```
await confirmation before apply

navarroaxel (commented):
Hey @hntee thanks for your contribution! I think this page is some outdated respect the style guide.

Could you add the link to the documentation in the page description?
```
> Sort lines of text files.
> More information: <https://www.gnu.org/software/coreutils/manual/html_node/sort-invocation.html>.
```

hntee:
Hi @navarroaxel , thanks for your reply.
I have changed the parameter to the longer one and added the link.

Another question: since the above descriptions are all `{{filename}}`, is it better to modify all of these to `{{path/to/file}}`?

navarroaxel:
> Hi @navarroaxel , thanks for your reply.
> I have changed the parameter to the longer one and added the link.

@hntee thanks!

> Another question: since the above descriptions are all `{{filename}}`, is it better to modify all of these to `{{path/to/file}}`?

I had that in mind, and also fix some arguments like `-r` to `--reverse` but I want the confirmation from @sbrl before apply those changes. Sometime there are exceptions for the rule.

sbrl (approved):
Thanks, @hntee!

Yeah since we're here, if you wouldn't mind adjusting all the other `{{filename}}` instances to `{{path/to/filename}}` that would be great!

Generally it's only pages like `diff` for example that are exceptions.

hntee:
Hi @navarroaxel @sbrl , I have modified all the {{filename}} instances to {{path/to/filename}}

navarroaxel (approved):

sbrl:
Thanks, @hntee :smiley_cat:

PR #4332, opened 2020-09-11
tree: add -C, --du and --prune examples

navarroaxel:

### DIFF
--- a/pages/linux/tree.md
+++ b/pages/linux/tree.md
@@ -19,7 +19,7 @@
 
 `tree -i -f`
 
-- Print the size of each node next to it, in human-readable format, with directories displaying their cumulative size (as in the `du` command):
+- Print the size of each file and the cumulative size of each directory, in human-readable format:
 
 `tree -s -h --du`
 
--- a/pages/linux/tree.md
+++ b/pages/linux/tree.md
@@ -1,6 +1,7 @@
 # tree
 
 > Show the contents of the current directory as a tree.
+> More information: <http://mama.indstate.edu/users/ice/tree/>.
 
 - Print files and directories up to 'num' levels of depth (where 1 means the current directory):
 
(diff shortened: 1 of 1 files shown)


navarroaxel:
Sync the linux/tree.md page with https://github.com/navarroaxel/tldr/blob/master/pages/osx/tree.md

pages/linux/tree.md
```diff
@@ -10,25 +11,25 @@
...
 
-`tree -a`
+`tree -a -C`
 
 - Print the tree without indentation lines, showing the full path instead (use `-N` to not escape whitespace and special characters):
 
 `tree -i -f`
 
-- Print the size of each node next to it, in human-readable format:
+- Print the size of each node next to it, in human-readable format, with directories displaying their cumulative size (as in the `du` command):
```
sbrl, line 22: Wow, this is a pretty long description. Is this syncing with another copy of this page, or is it original?

If the latter, then if we could shorten this a bit and make it more concise that'd be great :-)

Same goes for the other particularly long one further down.

navarroaxel: This commit extended the description https://github.com/tldr-pages/tldr/commit/c294bdc4 in September 2016.

navarroaxel: @sbrl This commit extended the description https://github.com/tldr-pages/tldr/commit/c294bdc4 in September 2016.

pages/linux/tree.md
```diff
@@ -10,25 +11,25 @@
...
+`tree -s -h --du`
 
-- Filter the tree using a wildcard (glob) pattern:
+- Print files within the tree hierarchy, using a wildcard (glob) pattern, and pruning out directories that don't contain matching files:
 
-`tree -P {{*.txt}}`
+`tree -P '{{*.txt}}' --prune`
 
-- Ignore entries that match a wildcard (glob) pattern:
+- Print directories within the tree hierarchy, using the wildcard (glob) pattern, and pruning out directories that aren't ancestors of the wanted one:
```
mebeim, line 30: I understand you wanting to add more command line options, to this page, but this is rather advanced usage. Think of it this way: is this in the top 8 most common usages of `tree` you can think of? I would suggest keeping the original, or at best elaborating it while still keeping it simple.

pages/linux/tree.md
mebeim: I would simplify it somewhere around this:

```suggestion
- Print the size of each file and the cumulative size of each directory, in human-readable format:
```
What do you think?

pages/linux/tree.md
navarroaxel: I'm just syncing this page with https://github.com/navarroaxel/tldr/blob/master/pages/osx/tree.md
Do you think we should shrink the osx page too?

mebeim: @navarroaxel okay, that's a good point, thank you for linking. I see that the OSX page has different because the command actually supports different options (see [relevant commit](https://github.com/tldr-pages/tldr/commit/31a07858ca2d010625499fb9e942f9c1371b339a) where this was introduced). Is this example compatible with the Linux version of the tool? If yes, then you are right that it makes sense to sync them, and we can work on a PR to simplify both later.

sbrl (approved):

pages/linux/tree.md
navarroaxel: Yeah, I've checked and the same options applies to the Linux `tree`, and all these arguments are explained in the `more information` link in this page.

navarroaxel: @mebeim Yeah, I've checked and the same options applies to the Linux `tree`, and all these arguments are explained in the `more information` link in this page, the same link than `osx` page.

mebeim: @navarroaxel thanks a lot

mebeim (approved):

PR #4342, opened 2020-09-14
tldr: add --update example

samarth261:

### DIFF
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -14,3 +14,7 @@
 - Get help for a git subcommand:
 
 `tldr {{git-checkout}}`
+
+- Update the local tldr pages:
+
+`tldr --update`
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -17,4 +17,4 @@
 
 - Update local pages (if the client supports caching):
 
-`tldr --update`
+`tldr -u`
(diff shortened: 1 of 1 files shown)


pages/common/tldr.md
```diff
@@ -14,3 +14,7 @@
 - Get help for a git subcommand:
 
 `tldr {{git-checkout}}`
+
+- Update the local tldr pages:
```
mebeim, line 18:
```suggestion
- Update local pages (if the client supports caching):
```
Simplify, use "pages" instead of "tldr pages". Also mention that client needs to support caching in order for the `-u` option to be available.

pages/common/tldr.md
```diff
@@ -14,3 +14,7 @@
...
+
+`tldr --update`
```
mebeim, line 20:
```suggestion
`tldr -u`
```
Use short version since long version is not only optional to support for clients.

mebeim (commented):
Hi @samarth261, thanks for your contribution.

The `-u/--update` command line flag is not mandatory for all TLDR clients compliant to [our specification](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md) (see the [arguments section](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#arguments)). Since we would not want to document something wrong, we would prefer not adding such example.

samarth261:
Okay, got it. Thanks for the review.

mebeim (approved):
Thank you!

navarroaxel (approved):

PR #4345, opened 2020-09-16
chore: remove locale from page urls

owenvoke:
This just updates all Microsoft docs links to not use `en-us`, meaning that they will redirect to the user's locale.

This applies to most of the Windows pages, although there was also one in `common`.

### DIFF
--- a/pages/common/dotnet-publish.md
+++ b/pages/common/dotnet-publish.md
@@ -1,7 +1,7 @@
 # dotnet publish
 
 > Publish a .NET application and its dependencies to a folder for deployment to a hosting system.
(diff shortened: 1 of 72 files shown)


navarroaxel (approved):

sbrl (approved):
Oh wow, awesome work @owenvoke! Very cool. Did you use a script to detect which ones needed altering?

owenvoke:
I just used `rg -l docs.microsoft.com/en-us windows/ | xargs sed 's/docs.microsoft.com\/en-us/docs.microsoft.com/g' -`

Or something to that effect.

PR #4347, opened 2020-09-17
brew: use the term "formulae" for consistency

quackduck:
Change "formulas" to "formulae" because that is what's used in the rest of the page.
Another thing - are you gonna be changing the name "osx" to "macos" or something because Big Sur will no longer officially be OS _X_ (No longer version 10)

### DIFF
--- a/pages/osx/brew.md
+++ b/pages/osx/brew.md
@@ -3,7 +3,7 @@
 > Package manager for macOS.
 > More information: <https://brew.sh>.
 
-- Search for available formulas and casks:
+- Search for available formulae and casks:
 
 `brew search {{text}}`
 


mebeim (approved):
Thank you for the contribution :)

> are you gonna be changing the name "osx" to "macos" or something because Big Sur will no longer officially be OS X

Yep, we already know about this, but unfortunately it's impossible to change the name of the folders without possibly breaking existing clients. If and when we decide to do this, we will have to deprecate the "osx" folder, add a symlink folder called "macos" and then remove the "osx" folder when enough time has passed.

navarroaxel (approved):

PR #4352, opened 2020-09-23
ip: add brief listing examples

ecklm:

### DIFF
--- a/pages/linux/ip.md
+++ b/pages/linux/ip.md
@@ -4,7 +4,7 @@
 
 - List interfaces with detailed info:
 
-`ip a`
+`ip address`
 
 - List interfaces with brief network layer info:
 
@@ -16,11 +16,11 @@
 
 - Display the routing table:
 
-`ip r`
+`ip route`
 
 - Show neighbors (ARP table):
 
-`ip n`
+`ip neighbour`
 
 - Make an interface up/down:
 
--- a/pages/linux/ip.md
+++ b/pages/linux/ip.md
@@ -6,6 +6,14 @@
 
 `ip a`
 
+- List interfaces with brief network layer info:
+
+`ip -brief address`
+
+- List interfaces with brief link layer info:
+
+`ip -brief link`
+
 - Display the routing table:
 
 `ip r`


navarroaxel (changes requested):
hey @ecklm, thanks for your submission! but `ip a` is the most basic usage, please keep that example and add `ip -brief a` as a new one.

ecklm:
> hey @ecklm, thanks for your submission! but `ip a` is the most basic usage, please keep that example and add `ip -brief a` as a new one.

There you are: 720bd6c816449eee68b0cf8d9395cda6bed9e8e5

navarroaxel (approved):

pages/linux/ip.md
```diff
@@ -6,6 +6,14 @@
 
 `ip a`
 
+- List interfaces with brief network layer info:
+
+`ip -brief a`
```
sbrl, line 11: Perhaps we could use the full name instead of just `a` here? We generally prefer that in tldr pages as it aids clarity.

I know that the rest of the page follows the trend of shortening everything, but it would be awesome if we could change that while we're at it :-)
```suggestion
`ip -brief address`
```

ecklm: I actually _do_ agree with you on this, I just tried to follow the trend. I'll apply it soon.

ecklm: I have pushed it. Also added a refactor patch of the other examples in the file.

navarroaxel (approved):

pages/linux/ip.md
```diff
@@ -4,15 +4,23 @@
```
navarroaxel, line 4: Plase the 🍒 of the page, add a link to the man page:
```suggestion
> More information: <https://manpages.debian.org/stretch/iproute2/ip.8.en.html>.

```

navarroaxel (approved):

pages/linux/ip.md
ecklm: I'm reluctant about adding this link. This link holds the documentation of the `ip` command _relevant to this specific Debian version_, no more, no less. For a RedHat or an ArchLinux user, it doesn't necessarily contain up to date information, therefore is misleading.

If you are okay with it, I'd just add the `man ip` command as the source for more information, as that is the place that holds all the details available on the user's distribution.

navarroaxel: :thinking: No, the `map {{command}}` is a well known option, but I didn't find the offical manpage for `ip` with the latest version. Could you find a better manpage link?

ecklm: `ip address` is also a well-known command, we still document it here.

My point is that every URL that leads to _a_ man page can be misleading because it is not related to the version the user uses. I might have an iproute package 3 versions newer than a CentOS user. It doesn't make sense for them to browse what the newest package knows, because they can only work with what's available on their distro.

Therefore I say, _if_ we include any "More info" note here, the best possibility is to refer to the man command. If you don't want the `man` command here because it is trivial, then simply don't include any link here.

sbrl: Hrm. I think we should avoid linking to the man page provided by a specific distribution - unless the command is specific to that distro. linux.die.net is fine for generics, but in this case I'd have to agree with @ecklm I think.

sbrl (approved):
Thanks, @ecklm!

ecklm:
Can someone merge it then?

PR #4362, opened 2020-09-26
git-archive: add page

glenacota:

Related to issue #3953

### DIFF
--- a/pages/common/git-archive.md
+++ b/pages/common/git-archive.md
@@ -3,22 +3,22 @@
 > Create an archive of files from a named tree.
 > More information: <https://git-scm.com/docs/git-archive>.
 
-- Create a tar archive from the contents of the latest commit on the current branch
+- Create a tar archive from the contents of the latest commit on the current branch:
 
 `git archive --verbose -o {{target.tar}} HEAD`
 
-- Create a tar archive from the contents of a specific directory
+- Create a tar archive from the contents of a specific directory:
 
 `git archive -o {{target.tar}} HEAD:{{directory}}`
 
-- Same as above, but change the prefix of all files in the archive
+- Same as above, but change the prefix of all files in the archive:
 
 `git archive -o {{target.tar}} --prefix={{prefix}} HEAD:{{directory}}`
 
-- Create a zip archive from the latest commit and print it to standard output
+- Create a zip archive from the latest commit and print it to standard output:
 
 `git archive --format=zip HEAD`
 
-- Create a zip archive from the latest commit on a specific branch
+- Create a zip archive from the latest commit on a specific branch:
 
 `git archive -o {{target.zip}} {{branch_name}}`
--- a/pages/common/git-archive.md
+++ b/pages/common/git-archive.md
@@ -3,6 +3,18 @@
 > Create an archive of files from a named tree.
 > More information: <https://git-scm.com/docs/git-archive>.
 
+- Create a tar archive from the contents of the current HEAD and print it to standard output:
+
+`git archive --verbose HEAD`
+
+- Create a zip archive from the current HEAD and print it to standard output:
+
+`git archive --verbose --format=zip HEAD`
+
+- Same as above, but write the zip archive to file:
+
+`git archive --verbose --output={{target.zip}} HEAD`
+
 - Create a tar archive from the contents of the latest commit on the current branch:
 
 `git archive --verbose --output={{target.tar}} HEAD`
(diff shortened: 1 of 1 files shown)


pages/common/git-archive.md
```diff
@@ -0,0 +1,24 @@
+# git archive
+
+> Create an archive of files from a named tree.
+> More information: <https://git-scm.com/docs/git-archive>.
+
+- Create a tar archive from the contents of the latest commit on the current branch:
+
+`git archive --verbose -o {{target.tar}} HEAD`
```
navarroaxel, line 8: We prefer the long argumento format:
```suggestion
`git archive --verbose --output={{path/to/file.tar}} HEAD`
```

pages/common/git-archive.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Create a tar archive from the contents of a specific directory:
+
+`git archive -o {{target.tar}} HEAD:{{directory}}`
```
navarroaxel, line 12:
```suggestion
`git archive --output={{path/to/file.tar}} HEAD:{{path/to/directory}}`
```

pages/common/git-archive.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Same as above, but change the prefix of all files in the archive:
+
+`git archive -o {{target.tar}} --prefix={{prefix}} HEAD:{{directory}}`
```
navarroaxel, line 16:
```suggestion
`git archive --output={{path/to/file.tar}} --prefix={{prefix}} HEAD:{{path/to/directory}}`
```

pages/common/git-archive.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Create a tar archive from the contents of the latest commit on the current branch:
+
+`git archive --verbose -o {{target.tar}} HEAD`
+
+- Create a tar archive from the contents of a specific directory:
+
+`git archive -o {{target.tar}} HEAD:{{directory}}`
+
+- Same as above, but change the prefix of all files in the archive:
```
navarroaxel, line 14: I'm not sure about this description, maybe the example is clearer if we say `preffix/` to include all the files inside a specific folder

pages/common/git-archive.md
```diff
@@ -0,0 +1,24 @@
...
+
+`git archive -o {{target.tar}} --prefix={{prefix}} HEAD:{{directory}}`
+
+- Create a zip archive from the latest commit and print it to standard output:
+
+`git archive --format=zip HEAD`
+
+- Create a zip archive from the latest commit on a specific branch:
+
+`git archive -o {{target.zip}} {{branch_name}}`
```
navarroaxel, line 24: Is this a typo? or `git` change the file type to `zip` using the extension?
```suggestion
`git archive --output={{path/to/file.tar}} {{branch_name}}`
```

pages/common/git-archive.md
```diff
@@ -0,0 +1,24 @@
+# git archive
+
+> Create an archive of files from a named tree.
+> More information: <https://git-scm.com/docs/git-archive>.
+
+- Create a tar archive from the contents of the latest commit on the current branch:
```
navarroaxel, line 6: Maybe an extra example (at the first position) saying that the output is written to the `stdout` is better.

navarroaxel (changes requested):
Hey @glenacota, thanks for your contribution. I left some comments.

pages/common/git-archive.md
glenacota: Will use the long argument :)
As for the file name, I used the same convention used for the `tar` command, e.g., `tar cf {{target.tar}} {{file1}} {{file2}} {{file3}}`

pages/common/git-archive.md
glenacota: yes, it's inferred. From the docs:
> If this option is not given, and the output file is specified, the format is inferred from the filename if possible (e.g. writing to "foo.zip" makes the output to be in the zip format)

glenacota:
Thanks for the review @navarroaxel ! Please have a look at the latest changes. I only kept `{{target.tar}}` instead of `{{path/to/target.tar}}` because (1) it's also `{{target.tar}}` for the `tar` command, (2) it keeps the example shorter. What do you think?

owenvoke (approved):
LGTM after @navarroaxel's comments have been addressed. 👍🏻

navarroaxel:
Ref: #3953

glenacota:
Thanks, @owenvoke! @navarroaxel, there is still 1 requested change pending, but I'm not sure what it refers to. Is it the `target.tar` one? Is it your final decision to stick to go for `path/to/file.tar`, and then update the `tar` page consistently afterwards?

pages/common/git-archive.md
sbrl: Hey! Yeah, filenames can be an awkward one. Generally, the rule of thumb is that we prefer including the `path/to/` prefix. However, if you've got a number of files in a row (e.g. see the page for the `diff` command), then we drop the `path/to/` prefix to avoid the example from being unnecessarily verbose. In this case though I think we can include it.

sbrl:
Hey! I've left a comment about the filename issue above.

navarroaxel (approved):

PR #4365, opened 2020-09-27
xephyr: add page

fxdave:

### DIFF
--- a/pages/common/xephyr.md
+++ b/pages/common/xephyr.md
@@ -1,11 +1,11 @@
 # Xephyr
 
-> A nested X server that runs as an X application
+> A nested X server that runs as an X application.
 
-- Create a black window with display ID ":2"
+- Create a black window with display ID ":2":
 
 `Xephyr -br -ac -noreset -screen 800x600 :2`
 
-- Start BSPWM on the new screen
+- Start BSPWM on the new screen:
 
 `DISPLAY=:2 bspwm`
--- a/pages/common/xephyr.md
+++ b/pages/common/xephyr.md
@@ -8,4 +8,4 @@
 
 - Start BSPWM on the new screen:
 
-`DISPLAY=:2 bspwm`
+`DISPLAY=:2 {{command_name}}`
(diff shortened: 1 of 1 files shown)


pages/common/xephyr.md
```diff
@@ -0,0 +1,11 @@
...
+
+> A nested X server that runs as an X application.
+
+- Create a black window with display ID ":2":
+
+`Xephyr -br -ac -noreset -screen 800x600 :2`
+
+- Start BSPWM on the new screen:
+
+`DISPLAY=:2 bspwm`
```
navarroaxel, line 11: `bspwm` is another windows manager and should have his own page.

navarroaxel (changes requested):

pages/common/xephyr.md
fxdave: `DISPLAY=:2 bspwm` is an example of how to start an X application to the Xephyr server. 
Do you think it is unnecessary? Usually, you wouldn't use DISPLAY=:2 on startup. So neither BSPWM nor Xephyr would contain this example.

By the mentality, "When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness. For example, commands that require sudo should include it directly in the examples", I thought there is a need to show people how to use that black screen that Xephyr gave.

Should I remove it?

sbrl: Hrm, this is a tricky one. Perahaps it would be best if we use a more generic example here. What about this?
```suggestion
`DISPLAY=:2 {{command_name}}`
```
The example description would also need updating.

pages/common/xephyr.md
```diff
@@ -0,0 +1,11 @@
+# Xephyr
+
+> A nested X server that runs as an X application.
+
+- Create a black window with display ID ":2":
+
+`Xephyr -br -ac -noreset -screen 800x600 :2`
```
owenvoke, line 7: I guess this could have token syntax for the resolution and id. 👍🏻
```suggestion
`Xephyr -br -ac -noreset -screen {{800x600}} {{:2}}`
```

owenvoke (changes requested):

owenvoke (approved):

navarroaxel (approved):

PR #4378, opened 2020-09-30
openvpn: add page

bergentroll:

### DIFF
--- a/pages/common/openvpn.md
+++ b/pages/common/openvpn.md
@@ -1,0 +1,16 @@
+# openvpn
+
+> OpenVPN client and daemon binary.
+> More information: <https://openvpn.net/>.
+
+- Connect to server with config file:
+
+`sudo openvpn {{client.conf}}`
+
+- Try to set up an insecure peer-to-peer tunnel on bob.example.com host:
+
+`sudo openvpn --remote {{alice.example.com}} --dev tun1 --ifconfig {{10.4.0.1}} {{10.4.0.2}}`
+
+- Connect to the awaiting bob.example.com host:
+
+`sudo openvpn --remote {{bob.example.com}} --dev tun1 --ifconfig {{10.4.0.2}} {{10.4.0.1}}`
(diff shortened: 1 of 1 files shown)


pages/common/openvpn.md
```diff
@@ -0,0 +1,16 @@
+# openvpn
+
+> OpenVPN client and daemon binary.
+> More information: <https://openvpn.net/>.
+
+- Connect to server with config file:
```
sbrl, line 6: Perhaps this will read better?
```suggestion
- Connect to server using a config file:
```

pages/common/openvpn.md
```diff
@@ -0,0 +1,16 @@
...
+
+`sudo openvpn {{client.conf}}`
```
sbrl, line 8: Perhaps this would be more consistent?
```suggestion
`sudo openvpn {{path/to/client.conf}}`
```

pages/common/openvpn.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Try to set up an insecure peer-to-peer tunnel on bob.example.com host:
```
sbrl, line 10: Hrm, if this is insecure I'm reluctant to include it. We should always promote good security practices where possible.

sbrl (changes requested):
Thanks for the new page, @bergentroll! I've left some comments below for you to review.

pages/common/openvpn.md
bergentroll: Well, It is actually based on man page examples. I'll add simple secured version. Therefore I see no reason to eliminate this, because it described as insecure, it is simple and it gives a point to play with. And it is secure in a trusted network.

sbrl: Ah, I see. Including the secure version definitely helps.

sbrl (approved):
Thanks, @bergentroll!

navarroaxel (approved):

PR #4397, opened 2020-10-01
nano: add new command for backup

Ray6464:

### DIFF
--- a/pages/common/nano.md
+++ b/pages/common/nano.md
@@ -18,3 +18,7 @@
 - Indent new lines to the previous lines' indentation:
 
 `nano -i {{filename}}`
+
+- Before saving, back up the previous version of the file with name = "currentName+tilde" (useful for quick edits)
+
+`nano -B {{filename}}`
--- a/pages/common/nano.md
+++ b/pages/common/nano.md
@@ -19,6 +19,6 @@
 
 `nano -i {{filename}}`
 
-- Before saving, back up the previous version of the file with name = "currentName+tilde" (useful for quick edits)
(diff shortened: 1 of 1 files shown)


pages/common/nano.md
```diff
@@ -18,3 +18,7 @@
 - Indent new lines to the previous lines' indentation:
 
 `nano -i {{filename}}`
+
+- Before saving, back up the previous version of the file with name = "currentName+tilde" (useful for quick edits)
```
navarroaxel, line 22: Maybe this is better:
```suggestion
- Back up the previous version as `{{filename}}~` when save:
```
or
```suggestion
- Before saving, back up the previous version as `{{filename}}~`:
```

sbrl: Maybe `on save` instead of `when save` there @navarroaxel? Other than that, either is fine :+1:

Ray6464: If we're going that way, how about this?:
```suggestion
- Before saving, back up the previous version of the file with name = "currentName+tilde" (useful for quick edits)
+ Before modification, backup separately as `{{currentFileName}}~`
```

navarroaxel: @sbrl ?

sbrl: That would work too! Let's do that. Although, we prefer `snake_case` to `pascalCase`:
```suggestion
- Before modification, backup separately as `{{current_file_name}}~`:
```

navarroaxel (approved):

sbrl (approved):
Thanks, @Ray6464!

PR #4398, opened 2020-10-01
service: move --status-all example

Ray6464:
- [ x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).

### DIFF
--- a/pages/linux/service.md
+++ b/pages/linux/service.md
@@ -3,22 +3,18 @@
 > Manage services by running init scripts.
 > The full script path should be omitted (/etc/init.d/ is assumed).
 
-- Start/Stop/Restart/Reload service (start/stop should always be available):
-
-`service {{init_script}} {{start|stop|restart|reload}}`
-
-- Do a full restart (runs script twice with start and stop):
-
-`service {{init_script}} --full-restart`
-
-- Show the current status of a service:
-
-`service {{init_script}} status`
-
-- List the status of all services:
+- List the name and status of all services:
 
 `service --status-all`
 
-- Restart network-manager service
+- Start/Stop/Restart/Reload service (start/stop should always be available):
 
-`service network-manager restart`
+`service {{service_name}} {{start|stop|restart|reload}}`
+
+- Do a full restart (runs script twice with start and stop):
+
+`service {{service_name}} --full-restart`
+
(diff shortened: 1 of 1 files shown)


pages/linux/service.md
```diff
@@ -18,3 +18,7 @@
 - List the status of all services:
 
 `service --status-all`
+
+- Restart network-manager service
```
navarroaxel, line 22: This example is covered with the first example

sbrl (commented):
Hey there, @Ray6464 - thanks for the edit! That example seems be the same as the first example on this page though. Perhaps you could edit the example to make it new / unique?

Also, don't forget the special `{{token_syntax}}` we use in tldr pages to indicate that something can be changed by the user :slightly_smiling_face:

pages/linux/service.md
Ray6464: Yes, however, this example is an **example of** the first example. I'm hoping, people who are looking at it for the first time will find it helpful.

sbrl: This is the kind of problem that  the `{{token_syntax}}` is supposed to solve. Here's what the existing page renders as:



Note the cyan text, that indicates that something can be changed.

I'm not convinced that duplicate example will help here - another principle that we follow here at tldr-pages is keeping pages concise.

Perhaps we could instead improve this page by changing `init_script` to `service_name`? Perhaps that would aid clarity without making the page unnecessarily verbose.

Ray6464: The idea of swapping out `service_name` with `init_script` sounds good. We should also add the `service --status-all` example first instead of last, with the title `List the status of all services:` swapped out for `List the name and status of all services`. Understand that the kind of audience that uses tldr-pages is often the kind who is avoiding the un-natural reading ambiance of man pages.

sbrl: Sounds good, @Ray6464.

Ray6464: Made some changes to the initial PR.

sbrl (approved):
Thanks, @Ray6464!

navarroaxel (approved):

navarroaxel (approved):

PR #4399, opened 2020-10-01
sed: add 11q example

Ray6464:
- [ ] The page has 8 or fewer examples. (it was over 8 already + filled with a ton of examples that are not tldr)

### DIFF
--- a/pages/common/sed.md
+++ b/pages/common/sed.md
@@ -26,10 +26,6 @@
 
 `sed 11q {{filename}}`
 
-- Print only text between n-th line till the next empty line:
-
-`sed -n '{{n}},/^$/p' {{filename}}`
-
 - Apply multiple find-replace expressions to a file:
 
 `sed -e 's/{{find}}/{{replace}}/' -e 's/{{find}}/{{replace}}/' {{filename}}`
@@ -37,7 +33,3 @@
 - Replace separator / by any other character not used in the find or replace patterns, e.g., #:
 
 `sed 's#{{find}}#{{replace}}#' {{filename}}`
-
-- Print only the n-th line of a file:
-
-`sed '{{n}}q;d' {{filename}}`
--- a/pages/common/sed.md
+++ b/pages/common/sed.md
@@ -22,7 +22,7 @@
 
 `sed '/{{line_pattern}}/d' {{filename}}`
 
-- Print the first 11 lines of a file
+- Print the first 11 lines of a file:
 
 `sed 11q {{filename}}`
 
--- a/pages/common/sed.md
+++ b/pages/common/sed.md
@@ -22,6 +22,10 @@
 
 `sed '/{{line_pattern}}/d' {{filename}}`
 
+- Print the first 11 lines of a file
+
+`sed 11q {{filename}}`
+
 - Print only text between n-th line till the next empty line:
 
 `sed -n '{{n}},/^$/p' {{filename}}`


pages/common/sed.md
```diff
@@ -22,6 +22,10 @@
 
 `sed '/{{line_pattern}}/d' {{filename}}`
 
+- Print the first 11 lines of a file
```
navarroaxel, line 25: This example is covered with the last example.

navarroaxel (changes requested):
Hey @Ray6464 thanks for your submission, but the limit is 8 examples per command.

sbrl (commented):
Hey there, @Ray6464! Unfortunately we have a limit of 8 example per page. Is there a less useful example we could remove instead?

Also, don't forget the `{{token_syntax}}` to indicate what the user can change :slightly_smiling_face:

Ray6464:
> Hey there, @Ray6464! Unfortunately we have a limit of 8 example per page. Is there a less useful example we could remove instead?
> 
> Also, don't forget the `{{token_syntax}}` to indicate what the user can change slightly_smiling_face

I am hoping for you guys to figure it out; in my opinion, examples (1) and (2) are very redundant so keep (1) and remove (2) to make room for my PR.
    1) Replace all occurrences of an extended regular expression in a file, and print the result:
    2) Replace all occurrences of a string in a file, overwriting the file (i.e. in-place):

sbrl:
Hrm, I'm inclined to disagree. One oft he goals of tldr-pages is to help the beginner, so I would argue that the first 2 examples are important to help new users to understand thee syntax of the command.

If you want to print the first N lines of a file, wouldn't  `head` be the more appropriate command to use here?

Ray6464:
> Hrm, I'm inclined to disagree. One oft he goals of tldr-pages is to help the beginner, so I would argue that the first 2 examples are important to help new users to understand thee syntax of the command.
> 
> If you want to print the first N lines of a file, wouldn't `head` be the more appropriate command to use here?

I agree with you on the 1st point, however `head` is not a better tool for this for the following reasons:
1) Head is part of the GNUcoreutils package, a malware that has infected almost every single server in existence.
2) Sed is meant to stand on its own.

I'll say either drop the example or make a file length exception of programs like `sed`.

sbrl:
That's a bit extreme to say that the GNU coreutils is malware. Despite that, I do see your point about your `sed` example. However, it's always awkward with commands such as `sed`, as there is so much functionality to fit into such a small page. The purpose of tldr-pages is to distil this down into 8 examples or less, and demonstrate only the most useful functionality. To this end, I'm afraid we can't lift the 8 example limit.

Ray6464:
> That's a bit extreme to say that the GNU coreutils is malware. Despite that, I do see your point about your `sed` example. However, it's always awkward with commands such as `sed`, as there is so much functionality to fit into such a small page. The purpose of tldr-pages is to distil this down into 8 examples or less, and demonstrate only the most useful functionality. To this end, I'm afraid we can't lift the 8 example limit.

How about this, let's argue that 'Print only the n-th line of a file' example is requndant after examples like 'Print the first 11 lines of a file' and 'Print only text between n-th line till the next empty line', so let's remove the `Print only the n-th line of a file` example with this one.

Also, note that after a few iterations the examples will probably be factored into examples of search, edit, delete, etc. and currently we're giving multiple examples of similar operations. TLDR page of Sed is looking almost similar to Man page like examples.

sbrl:
I think we need another opinion on this.

/cc @agnivade, or maybe @owenvoke perhaps?

agnivade:
Yes, we do have some pages which exceed the limit of 8. But that doesn't create a precedent to add yet more examples to it. I'd argue to actually remove one to bring it to 8. :)

This sed page does indeed have a lot more than we actually need. I think that's primarily because sed is such a huge command that can be used in so many ways. It might appear that we have multiple examples of similar operations, but there are subtle differences between each of them and it's mainly geared towards beginners trying to understand the command. So we carefully start with simple things and go towards more complex ones.

But the primary usage of sed that I have seen is in editing text. So I'd say all the removals should be the print examples. From the 3:

- Print only the n-th line of a file:
- Print only text between n-th line till the next empty line:
- Print first 11 lines:

I'd say 2nd is very advanced and less common than the other 2. So that's an easy one to reject. Among 1 and 3, it's a bit hard, but I'd say 1 would be slightly less common as a use-case than 3.

My suggestion would be to remove 1 and 2, and bring down the total count to 8.

Ray6464:
> My suggestion would be to remove 1 and 2, and bring down the total count to 8.

I'm up for this!

pages/common/sed.md
navarroaxel:
```suggestion
- Print the first 11 lines of a file:
```

navarroaxel (changes requested):

navarroaxel (approved):
LGTM

owenvoke (approved):

sbrl (approved):

PR #4400, opened 2020-10-01
nmap: add -O example

Ray6464:
- [ ] The page has 8 or fewer examples. (needs disscussion)

### DIFF
--- a/pages/common/nmap.md
+++ b/pages/common/nmap.md
@@ -8,10 +8,6 @@
 
 `nmap -O {{ip_or_hostname}}`
 
-- Scan a block of IP addresses, and additionally attempt to guess the operating system on any hosts found:
-
-`sudo nmap -sS -O {{192.168.1.0}}/{{24}}`
-
 - Try to determine whether the specified hosts are up and what are their names:
 
 `nmap -sn {{ip_or_hostname}} {{optional_another_address}}`
--- a/pages/common/nmap.md
+++ b/pages/common/nmap.md
@@ -8,7 +8,7 @@
 
 `nmap -O {{ip_or_hostname}}`
 
-- To scan the Network for devices (including routers) and get info on their operating system:
+- Scan a block of IP addresses, and additionally attempt to guess the operating system on any hosts found:
 
 `sudo nmap -sS -O 192.168.1.0/24`
 
--- a/pages/common/nmap.md
+++ b/pages/common/nmap.md
@@ -6,7 +6,7 @@
 
 - To check if the IP address `192.168.1.101` is up, with it’s operating system info. It may show RTTVAR logs (for TCP) for a long time:
 
-`nmap -O 192.168.1.101`
+`nmap -O {{ip_or_hostname}}`
 
 - To scan the Network for devices (including routers) and get info on their operating system:
 
--- a/pages/common/nmap.md
+++ b/pages/common/nmap.md
@@ -4,8 +4,6 @@
 > Some features only activate when Nmap is run with privileges.
 > More information: <https://nmap.org>.
 
-- If Nmap throws a root privileges error use it with `sudo` or `admin rights cmd`
-
 - To check if the IP address `192.168.1.101` is up, with it’s operating system info. It may show RTTVAR logs (for TCP) for a long time:
 
 `nmap -O 192.168.1.101`
--- a/pages/common/nmap.md
+++ b/pages/common/nmap.md
@@ -10,7 +10,7 @@
 
 - Scan a block of IP addresses, and additionally attempt to guess the operating system on any hosts found:
 
-`sudo nmap -sS -O 192.168.1.0/24`
+`sudo nmap -sS -O {{192.168.1.0}}/{{24}}`
 
 - Try to determine whether the specified hosts are up and what are their names:
 
--- a/pages/common/nmap.md
+++ b/pages/common/nmap.md
@@ -4,7 +4,7 @@
 > Some features only activate when Nmap is run with privileges.
 > More information: <https://nmap.org>.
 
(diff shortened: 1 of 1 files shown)


pages/common/nmap.md
```diff
@@ -4,6 +4,20 @@
 > Some features only activate when Nmap is run with privileges.
 > More information: <https://nmap.org>.
 
+# nmap
```
navarroaxel, line 7: Two headers?

pages/common/nmap.md
```diff
@@ -4,6 +4,20 @@
...
+> network scanning tool
+> More information: <https://nmap.org/>.
+
+- If Nmap throws a root privileges error use it with `sudo` or `admin rights cmd`
```
navarroaxel, line 11: We don't include this kind of message in our pages.

navarroaxel (changes requested):
Hey @Ray6464 thanks for your submission, I left some comments.

pages/common/nmap.md
Ray6464: fixed

pages/common/nmap.md
sbrl: Generally our format is to have a description, followed by an example.

Also, we try to keep pages focused on the command in question, and avoid explaining generic shell concepts such as `sudo`. We do have a convention to prefix commands that require sudo with `sudo` in the example, but in this case as far as I know `nmap` doesn't require `sudo` to function (it just takes longer).

To that end, I'd suggest we remove this line.

pages/common/nmap.md
```diff
@@ -4,6 +4,16 @@
 > Some features only activate when Nmap is run with privileges.
 > More information: <https://nmap.org>.
 
+- If Nmap throws a root privileges error use it with `sudo` or `admin rights cmd`
+
+- To check if the IP address `192.168.1.101` is up, with it’s operating system info. It may show RTTVAR logs (for TCP) for a long time:
```
sbrl, line 9: Perhaps this will read better?
```suggestion
- Check if a given IP address is up, and attempt to get the remote operating system:
```

pages/common/nmap.md
```diff
@@ -4,6 +4,16 @@
...
+
+`nmap -O 192.168.1.101`
+
+- To scan the Network for devices (including routers) and get info on their operating system:
```
sbrl, line 13: Perhaps this will read better?
```suggestion
- Scan a block of IP addresses, and additionally attempt to guess the operating system on any hosts found:
```

pages/common/nmap.md
```diff
@@ -4,6 +4,16 @@
 > Some features only activate when Nmap is run with privileges.
 > More information: <https://nmap.org>.
 
+- If Nmap throws a root privileges error use it with `sudo` or `admin rights cmd`
+
+- To check if the IP address `192.168.1.101` is up, with it’s operating system info. It may show RTTVAR logs (for TCP) for a long time:
+
+`nmap -O 192.168.1.101`
```
sbrl, line 11: Missing `{{token_syntax}}`?
```suggestion
`nmap -O {{ip_or_hostname}}`
```

pages/common/nmap.md
```diff
@@ -4,6 +4,16 @@
...
+
+- To scan the Network for devices (including routers) and get info on their operating system:
+
+`sudo nmap -sS -O 192.168.1.0/24`
```
sbrl, line 15: Missing token syntax
```suggestion
`sudo nmap -sS -O {{192.168.1.0/24}}`
```

sbrl (changes requested):
Thanks for the new page, @Ray6464! I've left some comments below for you to review.

pages/common/nmap.md
Ray6464: Done!

pages/common/nmap.md
Ray6464: This is better:
```suggestion
- Check if an IP address is up, with OS info:
```

pages/common/nmap.md
Ray6464: Check this out:
```suggestion
`sudo nmap -sS -O {{IP/port}}`
```

pages/common/nmap.md
sbrl: Hrm, I'm not keen on shortening `OS info` there, and it's also ambiguous what it does with it. What about this?
```suggestion
- Check if an IP address is up, and guess the remote host's operating system:
```

pages/common/nmap.md
sbrl: Hrm. It's not actually a port - it's a CIDR block number. We could do this?
```suggestion
`sudo nmap -sS -O {{ip_address}}/{{CIDR_block}}`
```
.....but it's not obvious what you're supposed to put there.  To that end, I think a concrete example would be best for this example:
```suggestion
`sudo nmap -sS -O {{192.168.1.0}}/{{24}}`
```

sbrl (approved):
Thanks, @Ray6464 :slightly_smiling_face:

navarroaxel:
Oh, but we can't merge this command, has 9 examples.

sbrl:
Oh, oops! I didn't notice that @navarroaxel. Great catch! Is there one we can remove?

Ray6464:
We have 2 examples in this PR, both closely related, we can remove one without much harm.

navarroaxel:
> We have 2 examples in this PR, both closely related, we can remove one without much harm.

@Ray6464 Which 2? Could you please send the edition?

Ray6464:
>1) `- Check if an IP address is up, and guess the remote host's operating system:`
>2) `- Scan a block of IP addresses, and additionally attempt to guess the operating system on any hosts found:`

I added 2 examples, just remove one! They are similar!

sbrl:
I think we should remove example 2 from that list.

navarroaxel (approved):

PR #4422, opened 2020-10-02
openssl-prime: add page

IacobusKopiirefuto:
Added two command for prime numbers. One generates a prime number, the other checks if given number is prime.

**Rational for exceeding 8 examples limit**:
I'm aware that this tldr page already has 8 examples, however all current examples are related to certificates. Direct generation of prime numbers using cryptographically secure methods is also a useful, even if it is just a niche function of openssl.

 
### DIFF
--- a/pages/common/openssl.md
+++ b/pages/common/openssl.md
@@ -34,11 +34,3 @@
 - Display the complete certificate chain of an HTTPS server:
 
 `openssl s_client -connect {{host}}:443 -showcerts </dev/null`
-
-- Generate a 2048bit prime number in hexadecimal format:
-
-`openssl prime -generate -bits 2048 -hex`
-
-- Check if given number is prime:
-
-`openssl prime {{number}}`
--- a/pages/common/openssl.md
+++ b/pages/common/openssl.md
@@ -34,3 +34,11 @@
 - Display the complete certificate chain of an HTTPS server:
 
 `openssl s_client -connect {{host}}:443 -showcerts </dev/null`
+
+- Generate a 2048bit prime number in hexadecimal format:
+
+`openssl prime -generate -bits 2048 -hex`
+
+- Check if given number is prime:
+
+`openssl prime {{number}}`
--- a/pages/common/openssl-genrsa.md
+++ b/pages/common/openssl-genrsa.md
@@ -1,7 +1,7 @@
 # openssl genrsa
 
 > OpenSSL command to generate RSA private keys.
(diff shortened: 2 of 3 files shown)


sbrl:
Hey there, @IacobusKopiirefuto! Ah, I totally see the predicament here. What about writing a page dedicated to the `prime` subcommand? We've done this before for other commands - for example `git merge`, and `docker ps`.

There are far too many uses for `openssl` to document them all in a single tldr page, so I suggest we refactor all the existing examples out into dedicated pages for each of the subcommands, and rewrite the `openssl` page to be a generic page for getting started with the `openssl` CLI tool.

IacobusKopiirefuto:
> What about writing a page dedicated to the `prime` subcommand? We've done this before for other commands - for example `git merge`, and `docker ps`.

@sbrl the problem is that `prime` subcommands doesn't have really that much options to fill its own page. It would be possible, but then it would just sort of duplicate its own manpage `man openssl prime`.

It might make more sense either to:
1. Create a page dedicated to `s_client` or `x509`. They might have more interesting option that could be mentioned. This would create space for the two `prime` examples on the main `openssl` page.
2. Exceed the tldr example limit just by one and only include the command for generating primes `openssl prime -generate -bits 2048 -hex`, as this command is probably more useful from the two commands that I wanted to include (the other was just checking if number is prime or not).

sbrl:
Ah, I see. While that's not necessarily a problem (eventually it would be awesome to have pages for all the openssl subcommands :D), I can see your point there.

Let's go with Option 1 there - it sounds like a good idea. Feel free to either add those pages to this PR or to open a new one for them after this one is merged.

pages/common/openssl.md
```diff
@@ -34,3 +34,11 @@
...
 
 `openssl s_client -connect {{host}}:443 -showcerts </dev/null`
+
+- Generate a 2048bit prime number in hexadecimal format:
+
+`openssl prime -generate -bits 2048 -hex`
+
+- Check if given number is prime:
+
+`openssl prime {{number}}`
```
navarroaxel, line 44:
```suggestion
```

navarroaxel (changes requested):

navarroaxel (approved):

sbrl (approved):
Just tweaked the grammar a bit, and I think it's good to go!

Thanks to @IacobusKopiirefuto for the original PR, and @navarroaxel for tidying it up :slightly_smiling_face:

sbrl:
Uh oh, looks like there's a conflict preventing a merge :thinking:

PR #4425, opened 2020-10-02
legit: add page

nath1as:

### DIFF
--- a/pages/linux/legit.md
+++ b/pages/linux/legit.md
@@ -3,20 +3,20 @@
 > Legit is a complementary command-line interface for Git
 > More information: <https://frostming.github.io/legit>.
 
-- Switches to branch. Stashes and restores unstaged changes.
+- Switches to branch. Stashes and restores unstaged changes:
 `git switch <branch>`
 
-- Synchronizes current branch. Auto-merge/rebase, un/stash.
+- Synchronizes current branch. Auto-merge/rebase, un/stash:
 `git sync`
 
-- Publishes branch to remote server.
+- Publishes branch to remote server:
 `git publish {{branch}}`
 
-- Removes branch from remote server.
+- Removes branch from remote server:
 `git unpublish {{branch}}`
 
-- Nice & pretty list of branches + publication status.
+- Nice & pretty list of branches + publication status:
 `git branches {{wildcard pattern}}`
 
-- Removes the last commit from history.
+- Removes the last commit from history:
 `git undo {{--hard}}`
--- a/pages/linux/legit.md
+++ b/pages/linux/legit.md
@@ -7,7 +7,7 @@
 
 `git switch {{target_branch}}`
 
-- Synchronize current branch. Auto-merge/rebase, un/stash:
+- Synchronize current branch, automatically merging or rebasing, and stashing and unstashing:
 
 `git sync`
 
--- a/pages/linux/legit.md
+++ b/pages/linux/legit.md
@@ -1,22 +1,28 @@
 # legit
 
-> Legit is a complementary command-line interface for Git
+> Legit is a complementary command-line interface for Git.
 > More information: <https://frostming.github.io/legit>.
 
 - Switch to branch. Stash and restore unstaged changes:
+
 `git switch <branch>`
 
 - Synchronize current branch. Auto-merge/rebase, un/stash:
+
 `git sync`
 
 - Publish branch to remote server:
+
 `git publish {{branch}}`
 
 - Remove branch from remote server:
+
 `git unpublish {{branch}}`
 
 - List of branches + publication status:
+
 `git branches {{wildcard pattern}}`
 
 - Remove the last commit from history:
+
 `git undo {{--hard}}`
--- a/pages/linux/legit.md
+++ b/pages/linux/legit.md
@@ -19,9 +19,9 @@
 
 `git unpublish {{branch_name}}`
 
-- List of branches + publication status:
+- List all branches and their publication status:
 
-`git branches {{wildcard pattern}}`
+`git branches {{glob_pattern}}`
 
 - Remove the last commit from the history:
 
--- a/pages/linux/legit.md
+++ b/pages/linux/legit.md
@@ -1,6 +1,6 @@
 # legit
 
-> Legit is a complementary command-line interface for Git.
+> Complementary command-line interface for Git.
 > More information: <https://frostming.github.io/legit>.
 
 - Switch to branch. Stash and restore unstaged changes:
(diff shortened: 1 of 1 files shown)


pages/linux/legit.md
```diff
@@ -0,0 +1,28 @@
+# legit
+
+> Legit is a complementary command-line interface for Git.
```
sbrl, line 3: Perhaps this will read better?
```suggestion
> Complementary command-line interface for Git.
```

pages/linux/legit.md
```diff
@@ -0,0 +1,28 @@
...
+> More information: <https://frostming.github.io/legit>.
+
+- Switch to branch. Stash and restore unstaged changes:
+
+`git switch <branch>`
```
sbrl, line 8: I'm confused. These examples show the `git` command, when this page is about `legit`?

tldr pages should always be named after the command they are about, and avoid talking about other commands if possible (although there are rare exceptions).

sbrl (changes requested):
Thanks for the new page, @nath1as! I've left some comments below for you to review.

pages/linux/legit.md
nath1as: you install legit that extends the functionalities of git, so you can write git undo and it 'uncommits' your commit and things like that
do you recommend I rename the file to git-legit, or something like that?

nath1as: all the git commands from legit also work if you write legit undo, legit switch etc. 
so the other option is to  change them into that

sbrl: Ah, I see. In that case, we need to have 1 page per subcommand. For example, you mention the `git undo` command, so that would be a candidate for a page.

In other words, we have 1 page for the `git` command itself, which shows the basic usage. Then, we have 1 page for each subcommand - e.g. `git pull`, `git merge`, etc.

I propose we close this PR and open a new one for each subcommand's page.

nath1as: @sbrl that would be weird in this case, as they have no additional flags, its just git undo and that would be the only thing on that page ...
I think its better to just rename the git part to legit, as it is its own program, which you can connect into git to extend it, but if you need to remember the commands, you would be better off tldr legit (at least I would)

sbrl: Ah, I see @nath1as. In that case, we should keep this page then.

However, I'd suggest that creating pages for each individual subcommand would be a good idea.

I've also just noticed that `git switch` is a core git command, so is it also a `legit` extension too? There seems to be a name conflict there.

nath1as: legit was made before ```git switch``` so it's now  a 'conflict'
I would use ```tldr legit``` to remember which of the legit extensions I have extending my git, and how to use them, if it is separated by pages I would have to know the subcommands in advance which is what I would like to know...
This is the totality of the legit help, so I'm not sure what would be shown on individual pages. 
````
Usage: legit [OPTIONS] COMMAND [ARGS]...

  legit command-line interface

Options:
  --version    Show the version and exit.
  --verbose    Enables verbose mode.
  --fake       Show but do not invoke git commands.
  --install    Install legit git aliases.
  --uninstall  Uninstall legit git aliases.
  --config     Edit legit configuration file.
  -h, --help   Show this message and exit.

Usage Examples:
Switch to specific branch:
$ legit switch <branch>

Sync current branch with remote:
$ legit sync

Sync current code with a specific remote branch:
$ legit sync <branch>

Publish current branch to remote:
$ legit publish

Publish a specific branch to remote:
$ legit publish <branch>

Unpublish a specific branch from remote:
$ legit unpublish <branch>

List branches matching wildcard pattern:
$ legit branches [<wildcard pattern>]

Commands:
  switch     Switches to specified branch.
  sync       Synchronizes the given branch with remote.
  publish    Publishes specified branch to the remote.
  unpublish  Removes specified branch from the remote.
  undo       Removes the last commit from history.
  branches   Displays a list of branches.
```

sbrl: Ah, I see - that makes sense. We should still have pages for the individual subcommands though - we have pages for `arch` for example - so they wouldn't be out of place :P

pages/linux/legit.md
```diff
@@ -0,0 +1,28 @@
+# legit
+
+> Complementary command-line interface for Git.
+> More information: <https://frostming.github.io/legit>.
+
+- Switch to branch. Stash and restore unstaged changes:
```
sbrl, line 6: Perhaps this will read better?
```suggestion
- Switch to a specified branch, stashing and restoring unstaged changes:
```

pages/linux/legit.md
```diff
@@ -0,0 +1,28 @@
...
+
+`git switch <branch>`
```
sbrl, line 8: Perhaps we could use the tldr-pages `{{token_syntax}}` here?
```suggestion
`git switch {{target_branch}}`
```

pages/linux/legit.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Synchronize current branch. Auto-merge/rebase, un/stash:
```
sbrl, line 10: Perhaps this will read better?
```suggestion
- Synchronize current branch, automatically merging or rebasing, and stashing and unstashing:
```

pages/linux/legit.md
```diff
@@ -0,0 +1,28 @@
...
+
+`git sync`
+
+- Publish branch to remote server:
+
+`git publish {{branch}}`
```
sbrl, line 16: Perhaps this will read better?
```suggestion
- Publish a specified branch to the remote server:

`git publish {{branch_name}}`
```

pages/linux/legit.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Remove branch from remote server:
+
+`git unpublish {{branch}}`
```
sbrl, line 20: Perhaps this will read better?
```suggestion
- Remove a branch from the remote server:

`git unpublish {{branch_name}}`
```

pages/linux/legit.md
```diff
@@ -0,0 +1,28 @@
...
+
+- List of branches + publication status:
+
+`git branches {{wildcard pattern}}`
```
sbrl, line 24: Perhaps this will read better?
```suggestion
- List all branches and their publication status:

`git branches {{glob_pattern}}`
```

pages/linux/legit.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Remove the last commit from history:
```
sbrl, line 26: Perhaps this will read better?
```suggestion
- Remove the last commit from the history:
```

sbrl (changes requested):
Just got a few more things before we can get to merging. Sorry for complicating the review process, @nath1as!

sbrl (approved):
Thanks, @nath1as!

navarroaxel (approved):

PR #4426, opened 2020-10-02
git-show-ref: add page

sanehab:

### DIFF
--- a/pages/common/git-show-ref.md
+++ b/pages/common/git-show-ref.md
@@ -15,6 +15,6 @@
 
 `git show-ref --tags`
 
-- Verify that a certainr reference exists:
+- Verify that a given reference exists:
 
 `git show-ref --verify {{path/to/ref}}`
--- a/pages/common/git-show-ref.md
+++ b/pages/common/git-show-ref.md
@@ -1,0 +1,20 @@
+# git show-ref
+
+> Git command for listing references.
+> More information: <https://git-scm.com/docs/git-show-ref>.
+
+- Show all refs in the repository:
+
+`git show-ref`
+
+- Show only heads references:
+
+`git show-ref --heads`
+
+- Show only tags references:
+
+`git show-ref --tags`
+
+- Verify that a certainr reference exists:
+
+`git show-ref --verify {{path/to/ref}}`


pages/common/git-show-ref.md
```diff
@@ -0,0 +1,20 @@
+# git show-ref
+
+> Git command for listing references.
+> More information: <https://git-scm.com/docs/git-show-ref>.
+
+- Show all refs in the repository:
+
+`git show-ref`
+
+- Show only heads references:
```
sbrl, line 10: Perhaps this will read better?
```suggestion
- Show only references with an associated head:
```

pages/common/git-show-ref.md
```diff
@@ -0,0 +1,20 @@
...
+
+`git show-ref --heads`
+
+- Show only tags references:
```
sbrl, line 14: Perhaps this will read better?
```suggestion
- Show only references with an associated tag:
```

pages/common/git-show-ref.md
```diff
@@ -0,0 +1,20 @@
...
+
+`git show-ref --tags`
+
+- Verify that a certainr reference exists:
```
sbrl, line 18: typo?
```suggestion
- Verify that a given reference exists:
```

sbrl (changes requested):
Thanks for the new page, @sanehab! I've left some comments below for you to review.

pages/common/git-show-ref.md
sanehab: I don't think so, because of two reasons:
1- This is the language used by git itself in different git pages, you can see for example here https://git-scm.herokuapp.com/book/zh-tw/v2/Git-Internals-Git-References that they use ```head references```.
2- a ```ref with an associated head``` might give the wrong impression that a ref might be associated with more than just a head or that it can be empty, but both of these can not happen.

pages/common/git-show-ref.md
sanehab: Yea this is a typo, and i like given over certain, since the reference is indeed given through the arguments.

pages/common/git-show-ref.md
sanehab: same reasons for heads

pages/common/git-show-ref.md
sbrl: Ah, I see. Let's leave it as it is then.

pages/common/git-show-ref.md
sbrl: I see. It's just it reads oddly, is all.

sbrl (approved):
Thanks, @sanehab

navarroaxel (approved):

PR #4427, opened 2020-10-02
cryfs, df, dfc, fls, ipfs, duperemove, edquote, lsattr, diskutil: consistently use "filesystem"

Hoi15A:
- [X] The pages follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).

In https://github.com/tldr-pages/tldr/pull/4379#discussion_r498534894 I mentioned making a PR to use "filesystem" consistently instead of "file system" in `df`. What I've done instead is do it for all occurrences I could find (which to my surprise also made me realise that I managed to miss 2 inside `dfc` after having said that I fixed it which is my bad).

I ran `grep -Hni "file system" */*` inside the pages directory to get a list of all files containing the words "file system" and those (with two exceptions) are the pages I've changed in this PR.

In [ipfs](https://github.com/tldr-pages/tldr/blame/master/pages/common/ipfs.md#L3) I left "Inter Planetary File System" as is as that's a name and is how its written in the [ipfs docs](https://docs.ipfs.io/).
And in [mount (windows)](https://github.com/tldr-pages/tldr/blame/master/pages/windows/mount.md#L3) I left "Network File System" as it is also a name.

### DIFF
--- a/pages/common/cryfs.md
+++ b/pages/common/cryfs.md
@@ -3,11 +3,11 @@
 > A cryptographic filesystem for the cloud.
 > More information: <https://www.cryfs.org/>.
 
-- Mount an encrypted file system. The initialization wizard will be started on the first execution:
+- Mount an encrypted filesystem. The initialization wizard will be started on the first execution:
 
 `cryfs {{path/to/cipher_dir}} {{path/to/mount_point}}`
 
-- Unmount an encrypted file system:
+- Unmount an encrypted filesystem:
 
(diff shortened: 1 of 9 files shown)


navarroaxel (approved):

schneiderl (approved):
As this is already a thing on tldr pages, I think it looks great. 
Thank you for your contribution.

PR #4437, opened 2020-10-03
Checkout all branches in GitHub Actions

JamieMagee:
By default, `actions/checkout@v2` only checks out the latest commit of the current branch. Unfortunately, this causes the `git diff` check in `check-pr.sh` to fail, as `master` branch isn't available.

This change checks out all commits of all branches, which allows `git diff` to work correctly.

Ideally, I'd like to use the `ref` setting of `actions/checkout@v2`, but that only supports a single ref so it's not possible to check out the PR branch _and_ `master` branch at the same time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,8 @@ jobs:
 
     - name: Checkout
       uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
 
     - name: Set up PR environment
       if: github.event.number != null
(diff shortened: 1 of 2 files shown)


sbrl (commented):
This looks ok to me, though admittedly I don't have much experience with GitHub actions.

/cc @owenvoke, who setup our GH actions IIRC.

owenvoke (approved):
Ugh, has that been happening all this time? 😬 Not sure how I missed that, this looks great!

sbrl (approved):

Issue #3689, opened 2019-12-24
Reach parity with cheat cheatsheets


waldyrious:
At the time of writing, the [cheat repository](https://github.com/cheat/cheatsheets) has 39 pages that we don't seem to have yet in tldr-pages:

- [ ] [ansi](https://github.com/cheat/cheatsheets/blob/master/ansi)
- [ ] [apparmor](https://github.com/cheat/cheatsheets/blob/master/apparmor)
- [ ] [asciiart](https://github.com/cheat/cheatsheets/blob/master/asciiart)
- [ ] [asterisk](https://github.com/cheat/cheatsheets/blob/master/asterisk)
- [ ] [cheat](https://github.com/cheat/cheatsheets/blob/master/cheat)
- [ ] [cpdf](https://github.com/cheat/cheatsheets/blob/master/cpdf)
- [ ] [cups](https://github.com/cheat/cheatsheets/blob/master/cups)
- [ ] [deb](https://github.com/cheat/cheatsheets/blob/master/deb)
- [ ] [distcc](https://github.com/cheat/cheatsheets/blob/master/distcc)
- [ ] [gyb](https://github.com/cheat/cheatsheets/blob/master/gyb)
- [ ] [hardware-info](https://github.com/cheat/cheatsheets/blob/master/hardware-info)
- [ ] [indent](https://github.com/cheat/cheatsheets/blob/master/indent)
- [ ] [iwconfig](https://github.com/cheat/cheatsheets/blob/master/iwconfig)
- [ ] [lib](https://github.com/cheat/cheatsheets/blob/master/lib)
- [ ] [lvm](https://github.com/cheat/cheatsheets/blob/master/lvm)
- [ ] [markdown](https://github.com/cheat/cheatsheets/blob/master/markdown)
- [ ] [nova](https://github.com/cheat/cheatsheets/blob/master/nova)
- [ ] [ntp](https://github.com/cheat/cheatsheets/blob/master/ntp)
- [ ] [org-mode](https://github.com/cheat/cheatsheets/blob/master/org-mode)
- [ ] [pkgtools](https://github.com/cheat/cheatsheets/blob/master/pkgtools)
- [ ] [r2](https://github.com/cheat/cheatsheets/blob/master/r2)
- [ ] [rcs](https://github.com/cheat/cheatsheets/blob/master/rcs)
- [ ] [readline](https://github.com/cheat/cheatsheets/blob/master/readline)
[...]

PR #4443, opened 2020-10-03
indent: add page

staceybellerose:

Relates to #3689

### DIFF
--- a/pages/common/indent.md
+++ b/pages/common/indent.md
@@ -1,0 +1,16 @@
+# indent
+
+> Change the appearance of a C/C++ program by inserting or deleting whitespace.
+> More information: <https://www.gnu.org/software/indent/>.
+
+- Format C/C++ source according to the Linux style guide:
+
+`indent -linux {{source1.c}} > {{source1.out}}`
+
+- Format C/C++ source according to the GNU style:
+
+`indent -gnu {{source2.c}} > {{source2.out}}`
+
+- Format C/C++ source according to the style of Kernigan & Ritchie (K&R), no tabs, 3 spaces per indent, and wrap lines at 120 characters:
+
+`indent -kr -i3 -nut -l120  {{source3.c}} > {{source3.out}}`
--- a/pages/common/indent.md
+++ b/pages/common/indent.md
@@ -3,14 +3,14 @@
 > Change the appearance of a C/C++ program by inserting or deleting whitespace.
 > More information: <https://www.gnu.org/software/indent/>.
 
-- Format C/C++ source according to the Linux style guide:
+- Format C/C++ source according to the Linux style guide, automatically back up the original files, and replace with the indented versions:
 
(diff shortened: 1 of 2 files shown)


pages/common/indent.md
```diff
@@ -0,0 +1,16 @@
+# indent
+
+> Change the appearance of a C/C++ program by inserting or deleting whitespace.
+> More information: <https://www.gnu.org/software/indent/>.
+
+- Format C/C++ source according to the Linux style guide:
+
+`indent -linux {{source1.c}} > {{source1.out}}`
```
navarroaxel, line 8: Please use the long argument format. And we use a path to file placeholder.
```suggestion
`indent --linux-style {{path/to/source.c}} > {{path/to/indented_source.c}}`
```
I read the docs, but I feel that we should keep the `.c` extension for the output file.

pages/common/indent.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Format C/C++ source according to the GNU style:
+
+`indent -gnu {{source2.c}} > {{source2.out}}`
```
navarroaxel, line 12: The docs says this command needs the `--standard-output` argument to write to `stdout`, or the `--output-file` (`-o`) to don't replace the original file.
https://www.gnu.org/software/indent/manual/indent.html#IDX3

```suggestion
`indent --gnu-style {{path/to/source.c}} > {{path/to/indented_source.c}}`
```

navarroaxel (changes requested):
Hey @staceybellerose Thanks for your submission, I left some comments.

Also, the docs say:
```
In the first form, one or more input files are specified. indent makes a backup copy of each file, and the original file is replaced with its indented version.
```
Could you please add an example with more than 1 file or the file of a specific directory? Is a space-separated list ?

pages/common/indent.md
staceybellerose: Updated to use long argument format and path to file placeholders.

pages/common/indent.md
staceybellerose: Corrected the examples to use -o argument.

staceybellerose:
> Hey @staceybellerose Thanks for your submission, I left some comments.
> 
> Also, the docs say:
> 
> ```
> In the first form, one or more input files are specified. indent makes a backup copy of each file, and the original file is replaced with its indented version.
> ```
> 
> Could you please add an example with more than 1 file or the file of a specific directory? Is a space-separated list ?

I've updated the first example to use the first form, with multiple files specified on the command line.

staceybellerose:
When testing on my Mac, I noticed that it doesn't use the GNU version of indent, but the FreeBSD version, so that has been included as well.

navarroaxel (approved):

sbrl (approved):
Wow, those are some pretty long lines, but in some cases it can't be helped.

Thanks, @staceybellerose!

PR #3170, opened 2019-07-04
hub: add hub pr usage for checkout someone created pull-request

conao3:
Hi! Thanks for such a great project!

I add `hub pr` usege for checkout someone created pull-request.

Cited from `hub` official readme https://hub.github.com/#maintainer

best,

- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #4472, opened 2020-10-04
git-pr: Remove page (en, es)

Amorymeltzer:
Added in #2674 (en) and #3727 (es) but sadly `git-pr` is not part of `git`, rather [`hub`](https://hub.github.com/hub.1.html#new-commands-provided-by-hub).  The en page for `hub` has an example for `pr` documented as of #3170, but there's no es page for hub (cc @ivanhercaz).


navarroaxel (approved):

owenvoke (changes requested):
I think that this shouldn't be applied. The `git-pr` page that I added was a reference to Git PR from [`git-extras`](https://github.com/tj/git-extras/blob/master/bin/git-pr) (which is not related to Hub). 🤔

I think what would make more sense is to add a link to these pages that reference:
https://github.com/tj/git-extras/blob/master/Commands.md#git-pr

i.e.
```md
> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-pr>.
```

Amorymeltzer:
@owenvoke Indeed!  My mistake — I use them both and forget which does which.

Is it desirable for `tldr` to document `git-extras` commands?  If so, there are *plenty* to be added!  Is the header sufficient to clarify that these aren't built-in git commands?

owenvoke:
In my opinion, yes! I think a lot of them are really useful to add to tldr pages. 👍🏻

ivanhercaz:
> Added in #2674 (en) and #3727 (es) but sadly `git-pr` is not part of `git`, rather [`hub`](https://hub.github.com/hub.1.html#new-commands-provided-by-hub). The en page for `hub` has an example for `pr` documented as of #3170, but there's no es page for hub (cc @ivanhercaz).
> 
> * [x]  The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).

Excuse me a lot! I am really busy with work tasks and I have not been enough time to check all the GitHub notifications, and even more considering the activity of tldr-pages.

sbrl (changes requested):
Just because they aren't core git subcommands, doesn't mean we shouldn't document them!

tldr-pages has a limitation that the name of a page  must exactly match the command being documented, so as log as you can access the `pr` extension git command by typing `git pr {{arguments_here}}`, we should keep it as-is.

Amorymeltzer:
Looks like a consensus to me!  Closing this then; I've opened #4838 to do @owenvoke's [suggestion](https://github.com/tldr-pages/tldr/pull/4472#pullrequestreview-501655737)

PR #4479, opened 2020-10-04
htpdate: add page

antoniobusrod:

Since I have only used `htpdate` command in Linux systems, I have not placed in common.

### References

* [htpdate(8) - Linux man page](https://linux.die.net/man/8/htpdate)
* [Keene Enterprises :: Open Source :: HTP](http://www.rkeene.org/oss/htp/)
* [ntpdate(8) - Linux man page](https://linux.die.net/man/8/ntpdate)

### DIFF
--- a/pages/linux/htpdate.md
+++ b/pages/linux/htpdate.md
@@ -6,7 +6,7 @@
 
 - Synchronize date and time:
 
-`htpdate {{host}}`
+`sudo htpdate {{host}}`
 
 - Perform simulation of syncronization, without any action:
 
@@ -14,8 +14,8 @@
 
 - Compensate the systematisch clock drift:
 
-`htpdate -x {{host}}`
+`sudo htpdate -x {{host}}`
 
 - Set time immediate after the syncronization:
 
-`htpdate -s {{host}}`
+`sudo htpdate -s {{host}}`
--- a/pages/linux/htpdate.md
+++ b/pages/linux/htpdate.md
@@ -11,3 +11,11 @@
 - Perform simulation of syncronization, without any action:
 
 `htpdate -q {{host}}`
+
+- Compensate the systematisch clock drift:
+
+`htpdate -x {{host}}`
+
+- Set time immediate after the syncronization:
+
+`htpdate -s {{host}}`
--- a/pages/linux/htpdate.md
+++ b/pages/linux/htpdate.md
@@ -6,8 +6,8 @@
 
 - Synchronize date and time:
 
-`sudo htpdate {{host}}`
+`htpdate {{host}}`
 
 - Perform simulation of syncronization, without any action:
 
-`sudo htpdate -q {{host}}`
+`htpdate -q {{host}}`
--- a/pages/linux/htpdate.md
+++ b/pages/linux/htpdate.md
@@ -2,12 +2,12 @@
 
 > Synchronize local date and time via HTTP headers from web servers.
 > Requires administrator privileges.
-> More information: <http://www.rkeene.org/oss/htp/>
+> More information: <http://www.rkeene.org/oss/htp/>.
 
-- Synchronize date and time
+- Synchronize date and time:
 
 `sudo htpdate {{host}}`
 
-- Perform simulation of syncronization, without any action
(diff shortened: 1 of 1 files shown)


pages/linux/htpdate.md
```diff
@@ -0,0 +1,13 @@
...
+> Requires administrator privileges.
+> More information: <http://www.rkeene.org/oss/htp/>.
+
+- Synchronize date and time:
+
+`sudo htpdate {{host}}`
+
+- Perform simulation of syncronization, without any action:
+
+`sudo htpdate -q {{host}}`
```
navarroaxel, line 13: The help says:
```-q    query only, don't make time changes (default)
```
So, is this the default behavior?
```suggestion
`htpdate -q {{host}}`
```

pages/linux/htpdate.md
```diff
@@ -0,0 +1,13 @@
+# htpdate
+
+> Synchronize local date and time via HTTP headers from web servers.
+> Requires administrator privileges.
+> More information: <http://www.rkeene.org/oss/htp/>.
+
+- Synchronize date and time:
+
+`sudo htpdate {{host}}`
```
navarroaxel, line 9: We don't use `sudo` in the pages.
```suggestion
`htpdate {{host}}`
```

navarroaxel (changes requested):
Hey @antoniobusrod, thanks for your submission. I left some comments.

Reading the help page `-x`, `-s` and `-p` looks good arguments to document here. Maybe you can add examples for these.

antoniobusrod:
> Reading the help page -x, -s and -p looks good arguments to document here. Maybe you can add examples for these.

I have explained `-x` and `-s` options.

However, I left unexplained `-p` option since it is recommended to be used with caution and it only has effect in daemon mode, which is not explained in `tldr`.

>        -p     Precision  (in  milliseconds) specifies the operating accuracy of htpdate. Internally htpdate uses a
>              different algorithm to detect a time offset, when precision is specified. Precision only has  effect
>              in daemon mode. Use with caution.

pages/linux/htpdate.md
sbrl: Erm @navarroaxel, that's not quite true. We always prefix a command with `sudo` if root privileges are required in order to execute said command.

sbrl:
Left a comment above. Are root privileges required to execute the command? If so, the examples for which root privilages are required should be prefixed with `sudo`.

navarroaxel:
base on https://github.com/tldr-pages/tldr/pull/4479#issuecomment-704376695 @antoniobusrod should we restore the `sudo` is all examples? Or some commands don't require sudo?

antoniobusrod:
> base on [#4479 (comment)](https://github.com/tldr-pages/tldr/pull/4479#issuecomment-704376695) @antoniobusrod should we restore the `sudo` is all examples? Or some commands don't require sudo?

After reading the available source code, I have seen that the command changes with `seteuid` the effective user ID of the calling process just before setting the date.

However, it is true that some options, like `-q` does not require sudo permissions. Moreover, for the options that require administrator permissions, `htpdate` returns a descriptive error as it follows.

```
$ htpdate -s google.com
Only root can change time
```

IMO I would modify documentation without `sudo` to just remind that administrator privileges are required.

```
Synchronize date and time (requires administrator privileges):

`htpdate {{host}}`
```

sbrl:
Right. We don't have to prefix _all_ the examples with `sudo` - just the ones that require admin privileges. I think it reads much better and is more concise if we prefix the examples that require root with `sudo`, rather than bloat the descriptions with it - it makes the page overall much easier to read.

pages/linux/htpdate.md
```diff
@@ -0,0 +1,21 @@
+# htpdate
+
+> Synchronize local date and time via HTTP headers from web servers.
+> Requires administrator privileges.
```
navarroaxel, line 4: I think we don't need this anymore, right @sbrl ?
```suggestion
```

sbrl (approved):
Thanks, @antoniobusrod :smiley

navarroaxel (approved):

Issue #4250, opened 2020-08-04
page request: e2fsprogs


samuel-w:
Currently this is what we have:
- [x] badblocks
- [x] blkid
- [x] chattr
- [x] debugfs
- [x] dumpe2fs
- [x] e2freefrag
- [x] e2fsck
- [x] e2image
- [x] e2label
- [x] e2undo
- [x] e4defrag
- [x] filefrag
- [x] findfs
- [ ] ~findsuper~ ([deprecated](https://kernel.googlesource.com/pub/scm/fs/ext2/e2fsprogs/+/v1.42.7/debian/changelog#2881))
- [x] logsave
- [x] lsattr
- [x] mke2fs
- [x] resize2fs
- [x] tune2fs

sbrl:
Thanks for the list, @samuel-w! They all look very useful!

bl-ue:
2 left!  @teeteejo @amitech @samuel-w any time/familiarity with the commands to add pages for `e2undo` and `findsuper`?

bl-ue:
Thank you @navarroaxel for `e2undo`!  Can anyone do `findsuper`?

navarroaxel:
Looks like `findsuper` was dropped.
https://github.com/tytso/e2fsprogs/blob/master/misc/findsuper.c
In e2fsprogs 1.19
https://kernel.googlesource.com/pub/scm/fs/ext2/e2fsprogs/+/v1.42.7/debian/changelog

bl-ue:
Looks like you're right @navarroaxel! [Here's the specific line.](https://kernel.googlesource.com/pub/scm/fs/ext2/e2fsprogs/+/v1.42.7/debian/changelog#2881)

In this case, would we add a page, and say it was deprecated?  Or still add a page?

navarroaxel:
It's a similar scenario like `dep` https://github.com/golang/dep, this is deprecated and archived and was removed from tldr pages.

bl-ue:
Okay.  And `findsuper` was deprecated long before dep was even started.  So I'll check that one off, and say we're done!

bl-ue:
Thank you @navarroaxel! ❤️

PR #4486, opened 2020-10-04
logsave: add page

teeteejo:

### DIFF
--- a/pages/linux/logsave.md
+++ b/pages/linux/logsave.md
@@ -3,15 +3,15 @@
 > Save the output of a command in a logfile.
 > More information: <https://linux.die.net/man/8/logsave>.
 
-- Execute command with specified argument(s) and save its output to logfile:
+- Execute command with specified argument(s) and save its output to log file:
 
-`logsave {{logfile}} {{command}}`
+`logsave {{path/to/logfile}} {{command}}`
 
-- Take input from standard input and save it in logfile:
+- Take input from standard input and save it in a log file:
 
 `logsave {{logfile}} -`
 
-- Append the output to logfile, instead of replacing its current contents:
+- Append the output to a log file, instead of replacing its current contents:
 
 `logsave -a {{logfile}} {{command}}`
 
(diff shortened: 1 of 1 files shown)


pages/linux/logsave.md
```diff
@@ -0,0 +1,20 @@
+# logsave
+
+> Save the output of a command in a logfile.
+> More information: <https://linux.die.net/man/8/logsave>.
+
+- Execute command with specified argument(s) and save its output to logfile:
+
+`logsave {{logfile}} {{command}}`
```
schneiderl, line 8: Perhaps it would be good to include some example arguments here. 
I understand that the {{command}} tag would include the command plus the arguments but maybe it would be good to make that explicit.

schneiderl: well....maybe that would just make it more confusing. I'm going to approve the PR as I'm not completely sure if we should or shouldn't add these explicit arguments. I'll let you and the other reviewers decide on adding this or not.

schneiderl (approved):

schneiderl:
I've left a minor comment for you to take a look at. It looks great overall, thank you for the PR and be welcome to the project. 
Feel welcome to contribute again.

pages/linux/logsave.md
teeteejo: Thank you @schneiderl for reviewing it. I was debating the same, but as the description clearly says "command with specified argument(s)", I hope this is implied? Just thinking if it would confuse further if we include the arguments for the command? Or should I change `{{command}}` to `{{command_with_arguments}}` or something? Thanks

schneiderl: I was thinking about something like `{{command argument}}`. But I think you are right. Adding this would only make it more confusing as it seems like you would be just passing the arguments (at least on the way I've put it). 
I believe we can leave this as is. 
Once again thank you for your contribution.

teeteejo: Thanks a lot again.

navarroaxel (approved):

sbrl (changes requested):
Thanks for the new page, @teeteejo! I've left some comments below for you to review.

sbrl (approved):
Thanks, @teeteejo

teeteejo:
Thank you. This relates to #4250

PR #4506, opened 2020-10-05
fail2ban-client: add page

antoniobusrod:

Since I have only used `fail2ban-client` command in Linux systems, I have not placed in common. I have seen that `fail2ban-client` can be installed in Mac with [MacPorts](https://ports.macports.org/port/fail2ban/summary), but since Mac installation is not supported in [Wiki installation documentation page](https://github.com/fail2ban/fail2ban/wiki/How-to-install-or-upgrade-fail2ban-manually), I have just included it in `linux` directory.

Moreover, I have documented the main options I usually use, for example, to manage `nginx-http-auth` jail.

### References
* [fail2ban commands](https://www.fail2ban.org/wiki/index.php/Commands)
* [GitHub repository](https://github.com/fail2ban/fail2ban/)
* [fail2ban-client(1) - Linux man page](https://linux.die.net/man/1/fail2ban-client)

### DIFF
--- a/pages/linux/fail2ban-client.md
+++ b/pages/linux/fail2ban-client.md
@@ -1,0 +1,16 @@
+# fail2ban-client
+
+> Configure and control fail2ban server
+> More information: <https://github.com/fail2ban/fail2ban/>
+
+- Gets the current status of the jail service:
+
+`fail2ban-client status {{jail}}`
+
+- Manually unban IP from the jail service:
+
+`fail2ban-client set {{jail}} unbanip {{ip}}`
+
+- Verify fail2ban server is alive:
+
+`fail2ban-client ping`
--- a/pages/linux/fail2ban-client.md
+++ b/pages/linux/fail2ban-client.md
@@ -1,7 +1,7 @@
 # fail2ban-client
 
 > Configure and control fail2ban server.
-> More information: <https://github.com/fail2ban/fail2ban/>.
+> More information: <https://github.com/fail2ban/fail2ban>.
 
 - Retrieve current status of the jail service:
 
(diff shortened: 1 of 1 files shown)


owenvoke (approved):
LGTM! Thanks @antoniobusrod.

pages/linux/fail2ban-client.md
```diff
@@ -0,0 +1,16 @@
+# fail2ban-client
+
+> Configure and control fail2ban server.
+> More information: <https://github.com/fail2ban/fail2ban/>.
+
+- Retrieve current status of the jail service:
+
+`fail2ban-client status {{jail}}`
+
+- Manually unban IP from the jail service:
```
navarroaxel, line 10: I think the `manually` is not needed:
```suggestion
- Unban IP from the jail service:
```
I'm sure @sbrl won't like the `unban` as a verb, maybe he knows a good synonym.

navarroaxel (changes requested):

pages/linux/fail2ban-client.md
owenvoke: I feel like this is probably fine (to use _unban_) as it's the terminology used by Fail2ban. 🤷🏻

sbrl: Unban is fine as a verb here, I think. However, I think we're missing an article. What  about this?
```suggestion
- Unban an IP from the jail service:
```

However, I like @owenvoke's description much better, so I think we should go with that instead.
```suggestion
- Remove the specified IP from the jail service's ban list:
```

navarroaxel (approved):

PR #4510, opened 2020-10-05
aws-kinesis: add page

mearns:

### DIFF
--- a/pages/common/aws-kinesis.md
+++ b/pages/common/aws-kinesis.md
@@ -1,0 +1,28 @@
+# aws kinesis
+
+> Official AWS CLI for Amazon Kinesis streaming data services.
+> More information: <https://aws.amazon.com/cli>.
+
+- Show all streams in the account:
+
+`aws kinesis list-streams`
+
+- Write one record to a Kinesis stream:
+
+`aws kinesis put-record --stream-name {{streamname}} --partition-key {{partition_key}} --data {{base64_encoded_message}}`
+
+- Write a record to a Kinesis stream with inline base64 encoding:
+
+`aws kinesis put-record --stream-name {{streamname}} --partition-key {{partition_key}} --data "$( echo "{{my raw message}}" | base64 )"`
+
+- List the shards available on a stream:
+
+`aws kinesis list-shards --stream-name {{streamname}}`
+
+- Get a shard iterator for reading from the oldest message in a stream's shard:
+
+`aws kinesis get-shard-iterator --shard-iterator-type TRIM_HORIZON --stream-name {{streamname}} --shard-id {{shard_id}}`
+
+- Read records from a shard, using a shard iterator (e.g., retrieved with get-shard-iterator):
+
+`aws kinesis get-records --shard-iterator {{AAAAAAAAAAEVPE8Ghmz0wTmWN4+A0agl...TJ9i4Dp}}`
--- a/pages/common/aws-kinesis.md
+++ b/pages/common/aws-kinesis.md
@@ -25,4 +25,4 @@
 
 - Read records from a shard, using a shard iterator:
 
-`aws kinesis get-records --shard-iterator {{AAAAAAAAAAEVPE8Ghmz0wTmWN4+A0agl...TJ9i4Dp}}`
+`aws kinesis get-records --shard-iterator {{shard_iterator}}`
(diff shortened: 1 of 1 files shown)


pages/common/aws-kinesis.md
```diff
@@ -0,0 +1,28 @@
...
+
+- List the shards available on a stream:
+
+`aws kinesis list-shards --stream-name {{streamname}}`
+
+- Get a shard iterator for reading from the oldest message in a stream's shard:
+
+`aws kinesis get-shard-iterator --shard-iterator-type TRIM_HORIZON --stream-name {{streamname}} --shard-id {{shard_id}}`
+
+- Read records from a shard, using a shard iterator (e.g., retrieved with get-shard-iterator):
```
agnivade, line 26: IMO, there is no need to mention the e.g. part. The earlier example clearly mentions "Get a shard iterator". So it's sufficient to just say "using a given shard iterator"

pages/common/aws-kinesis.md
```diff
@@ -0,0 +1,28 @@
...
+
+`aws kinesis get-records --shard-iterator {{AAAAAAAAAAEVPE8Ghmz0wTmWN4+A0agl...TJ9i4Dp}}`
```
agnivade, line 28: I see that you have given a concrete example for shard iterator, but other examples are generic. Any reason for this discrepancy?

agnivade (commented):
Thanks for the PR. Please take a look at the comments.

pages/common/aws-kinesis.md
mearns: Makes sense, removing in 595a3c6a.

pages/common/aws-kinesis.md
mearns: The shard iterator is an opaque value, and the response from `get-shard-iterator` is a JSON document like the following:

```json
{
    "ShardIterator": "AAAAAAAAAAGNPjLxaLMm/PQ4Ka/Ki3ULAff087oBFTXxNSIr7ekWhpAc9LzVHTrUAZQ/injqkfMyMw1z7ftO2y/PDEuIUgpONu14nGEaftcRuKPV5WhvbV7LFLWL5Qn0PLWrTNGFhLeceJ1lNRWST+TFrhXzMVhC1trd6YVO1tFEcb5FAu5iLAkUZfP1o6oe4bWJDMST6u3OazfQHaf4YgxEHCO8Vdwj"
}
```

Given this, I thought it was beneficial to show what the expected shard-iterator value looks like, so that, for instance, it would be clear that the entire JSON document should _not_ be given in this parameter.

But I'm not at all married to this; if you think it is sufficient to use a generic value like `{{shard-iterator}}`, I'm fine with that.

agnivade: I think it should be fine to give a generic value lik `{{shard_iterator}}`. Thanks!

mearns: You got it. Updated in d623b8f8.

agnivade (approved):
Thanks!

pages/common/aws-kinesis.md
```diff
@@ -0,0 +1,28 @@
...
+> Official AWS CLI for Amazon Kinesis streaming data services.
+> More information: <https://docs.aws.amazon.com/cli/latest/reference/kinesis/index.html#cli-aws-kinesis>.
+
+- Show all streams in the account:
+
+`aws kinesis list-streams`
+
+- Write one record to a Kinesis stream:
+
+`aws kinesis put-record --stream-name {{streamname}} --partition-key {{partition_key}} --data {{base64_encoded_message}}`
```
navarroaxel, line 12: You can use just `name` because the argument has `-name` is easy to read:
```suggestion
`aws kinesis put-record --stream-name {{name}} --partition-key {{partition_key}} --data {{base64_encoded_message}}`
```
Or
```suggestion
`aws kinesis put-record --stream-name {{stream_name}} --partition-key {{partition_key}} --data {{base64_encoded_message}}`
```
The same for the other examples with `--stream-name`

navarroaxel (changes requested):

pages/common/aws-kinesis.md
mearns: Good call, updated in 669be42a. I applied the same to the other parameters, like changing `--partition-key {{partition_key}}` to `--partition-key {{key}}`. I can change those back if you prefer.

navarroaxel (approved):

PR #4513, opened 2020-10-05
vim: add :wq, in case of emergency for beginners.

juniorRubyist:
> But the one added seems necessary.

### DIFF
--- a/pages/common/vim.md
+++ b/pages/common/vim.md
@@ -3,8 +3,7 @@
 > Vim (Vi IMproved), a command-line text editor, provides several modes for different kinds of text manipulation.
 > Pressing `i` enters edit mode. `<Esc>` goes back to normal mode, which doesn't allow regular text insertion.
 > More information: <https://www.vim.org>.
-
-**Important:** If you accidentally open Vim and have never used it before, _do not_ press anything. Type `:q!` and press enter to quit.
+> **Important:** If you accidentally open Vim and have never used it before, _do not_ press anything. Type `:q!` and press enter to quit.
 
 - Open a file:
 
--- a/pages/common/vim.md
+++ b/pages/common/vim.md
@@ -20,7 +20,7 @@
 
 - Quit without saving:
 
-`:quit!<Enter>` or `:q!<Enter>`
+`:quit!<Enter>`
 
 - Save and quit:
 
--- a/pages/common/vim.md
+++ b/pages/common/vim.md
@@ -4,7 +4,7 @@
 > Pressing `i` enters edit mode. `<Esc>` goes back to normal mode, which doesn't allow regular text insertion.
 > More information: <https://www.vim.org>.
 
-**Important:** If you accidentally open Vim and have never used it before, _do not_ press anything. Type `:q!` and hit <kbd>Enter</kbd> to quit.
+**Important:** If you accidentally open Vim and have never used it before, _do not_ press anything. Type `:q!` and press enter to quit.
 
 - Open a file:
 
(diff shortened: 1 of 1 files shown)


pages/common/vim.md
```diff
@@ -18,7 +20,11 @@
 
 - Quit without saving:
 
-`:quit!<Enter>`
+`:quit!<Enter>` or `:q!<Enter>`
```
navarroaxel, line 23: We prefer the long format to be easier for beginners.

navarroaxel (changes requested):

pages/common/vim.md
juniorRubyist: That's a fair point.

pages/common/vim.md
```diff
@@ -4,6 +4,8 @@
 > Pressing `i` enters edit mode. `<Esc>` goes back to normal mode, which doesn't allow regular text insertion.
 > More information: <https://www.vim.org>.
 
+**Important:** If you accidentally open Vim and have never used it before, _do not_ press anything. Type `:q!` and hit <kbd>Enter</kbd> to quit.
```
sbrl, line 7: Normally, I would suggest that this is not part of the standard format for tldr pages, and should be removed. However, this is a very unusual page, and it adds value. HTML support though isn't in the [client spec](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md), so it'll have to be removed. (CommonMark support is in the spec though, hence why I support leaving this in).
```suggestion
**Important:** If you accidentally open Vim and have never used it before, _do not_ press anything. Type `:q!` and press enter to quit.
```

Hopefully we can prevent situations such as [this one](https://www.pixelstech.net/images/fun/normal/IMG-20170805-WA0000.jpg) :P ([source](https://www.pixelstech.net/fun/412-Exit-Vim))

sbrl (changes requested):
Thanks for the edit, @juniorRubyist ! I've left some comments below for you to review.

Also, we should wait for feedback from others before merging this due to the unusual situation here.

juniorRubyist:
@navarroaxel @sbrl I have implemented your suggestions. Let me know if any other changes may be needed.

sbrl (commented):
Looks ok to me! We should get some additional opinions before merging though, so let's wait a bit to give the others time to review.

pages/common/vim.md
```diff
@@ -4,6 +4,8 @@
 > Pressing `i` enters edit mode. `<Esc>` goes back to normal mode, which doesn't allow regular text insertion.
 > More information: <https://www.vim.org>.
 
+**Important:** If you accidentally open Vim and have never used it before, _do not_ press anything. Type `:q!` and press enter to quit.
```
sbrl, line 7: Oops! I forgot to return to this PR. Sorry about that! Upon further investigation, it looks like this line is causing the build to fail, so we probably need to remove it.

navarroaxel, line 7:
```suggestion
> **Important:** If you accidentally open Vim and have never used it before, _do not_ press anything. Type `:q!` and press enter to quit.
```

navarroaxel (changes requested):

bl-ue:
@juniorRubyist thank you very much for this PR, however, I don't think we need another example as the 3rd and 4th examples already show how to save and quit, respectively.  We try to keep a maximum of 8 examples.

bl-ue:
@juniorRubyist Just a note: the title of this PR reads "add :wq, in case of emergency for beginners", but in the event that a user is accidentally stuck in vim and wants out, it's very unlikely they want to write the files (even if they didn't change them it will still update the date on the file) 😉

PR #4520, opened 2020-10-06
git-mv: add it translation

glenacota:

### DIFF
--- a/pages.it/common/git-mv.md
+++ b/pages.it/common/git-mv.md
@@ -9,7 +9,7 @@
 
 - Rinomina i file e aggiungi l'operazione al commit successivo:
 
-`git mv {{percorso/al/file}} {{percorso/al/file_rinominato}}`
+`git mv {{file}} {{file_rinominato}}`
 
 - Sposta sovrascrivendo eventuali file esistenti nel percorso di destinazione:
 
--- a/pages.it/common/git-mv.md
+++ b/pages.it/common/git-mv.md
@@ -11,6 +11,6 @@
 
 `git mv {{percorso/al/file}} {{percorso/al/file_rinominato}}`
 
-- Sovrascrivi un file forzatamente se già presente nel percorso di destinazione:
+- Sposta sovrascrivendo eventuali file esistenti nel percorso di destinazione:
 
 `git mv --force {{percorso/al/file}} {{nuovo/percorso/al/file}}`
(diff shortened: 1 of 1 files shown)


pages.it/common/git-mv.md
```diff
@@ -0,0 +1,16 @@
...
+> Sposta o rinomina file e aggiorna l'indice Git.
+> Maggiori informazioni: <https://git-scm.com/docs/git-mv>.
+
+- Sposta i file nella repository e aggiungi l'operazione al commit successivo:
+
+`git mv {{percorso/al/file}} {{nuovo/percorso/al/file}}`
+
+- Rinomina i file e aggiungi l'operazione al commit successivo:
+
+`git mv {{percorso/al/file}} {{percorso/al/file_rinominato}}`
```
navarroaxel, line 12: This is not the 2nd example in the English page:
```suggestion
`git mv {{file}} {{file_rinominato}}`
```

navarroaxel (changes requested):
Hey @glenacota, thanks for your translation. Only 1 comment.

pages.it/common/git-mv.md
glenacota: In my understanding, it actually is. The second example in the English page refers to a filename, which it's a path/to/file. You might think of a mapping `filename` -> `path/to/file` (i.e., `percorso/al/file`), and `new_filename` -> `path/to/renamed_file` (i.e., `percorso/al/file_rinominato`).

But if you think that is more clear stressing the name change only, I'll update my translation :)

navarroaxel: In Linux terms, the 1st and 2nd examples are identical, but I think the idea is only say `as user you can rename a file within the same folder`. You can await the review from another collaborator cc @sbrl @owenvoke

sbrl: Hrm, I don't know italian so I can't really comment here.

navarroaxel: me neither, but I think @glenacota is not talking about something specific to it language, right?

glenacota: > me neither, but I think @glenacota is not talking about something specific to it language, right?

exactly, I have the same doubt for the English version. But for the sake of this PR, I can translate the token literally ;)

schneiderl (approved):
LGTM

navarroaxel (approved):

PR #4528, opened 2020-10-06
basename: change description and add directory example

gutjuri:
The previous description was wrong, because `basename {{path/to/directory/}}` does not return the non-directory portion of the specified path (which would be the empty string), but `directory`.

### DIFF
--- a/pages/common/basename.md
+++ b/pages/common/basename.md
@@ -1,10 +1,14 @@
 # basename
 
-> Returns non-directory portion of a pathname.
+> Removes leading directory portions from a path.
 
 - Show only the file name from a path:
 
 `basename {{path/to/file}}`
+
+- Show only the rightmost directory name from a path:
+
+`basename {{path/to/directory/}}`
 
 - Show only the file name from a path, with a suffix removed:
 
--- a/pages/common/basename.md
+++ b/pages/common/basename.md
@@ -1,6 +1,6 @@
 # basename
 
-> Removes leading directory portions from a path.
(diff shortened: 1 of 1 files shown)


pages/common/basename.md
```diff
@@ -1,11 +1,15 @@
...
-> Returns non-directory portion of a pathname.
+> Removes leading directory portions from a path.
 
 - Show only the file name from a path:
 
 `basename {{path/to/file}}`
 
+- Show only the rightmost directory name from a path:
+
+`basename {{path/to/directory/}}`
```
navarroaxel, line 11: In my tests: the trailing `/` is not needed.
```suggestion
`basename {{path/to/directory}}`
```

pages/common/basename.md
```diff
@@ -1,11 +1,15 @@
 # basename
 
-> Returns non-directory portion of a pathname.
+> Removes leading directory portions from a path.
```
navarroaxel, line 3:
```suggestion
> Remove leading directory portions from a path.
```

navarroaxel (changes requested):

pages/common/basename.md
gutjuri: It is indeed not needed, but I found it well-suited for an example because the trailing `/` is removed in basename's output (which a novice user might not expect).

sbrl: Let's keep this as-is then.

sbrl (approved):
Thanks, @gutjuri :smiley_cat:

PR #4531, opened 2020-10-06
scoop: add cleanup example

LukBukkit:
_Please see below, I want to discuss which example should be removed or moved to it's own page._

Hello,

I've been using scoop for more than a year now and sometime I was beginning to wonder why my SSD, where `scoop` was installed, has been filling up so fast. At this time, I didn't know, that scoop doesn't automatically removes old versions of installed software. To do this you have to run the `scoop cleanup *` command manually. Just recently I discoverd, that by using the `-k` option, also the download cache of old software would be cleared. That's the reason why I think, it's worth adding this example. 

But there are now eight examples in this file. This is according to your guidelines not allowed. I would propose to move the two `bucket` commands into the separate file `scoop-bucket.md`, because both of them could be important for using this application. If you agree with my idea, please let me know and I'll implement it by creating an additional commit.

### DIFF
--- a/pages/windows/scoop.md
+++ b/pages/windows/scoop.md
@@ -34,3 +34,7 @@
 - Add a bucket by its alias or a Git repository URL:
 
 `scoop bucket add {{bucket}}`
+
+- Remove old versions of all packages and clear the download cache:
+
+`scoop cleanup -k *`
--- a/pages/windows/scoop.md
+++ b/pages/windows/scoop.md
@@ -27,14 +27,6 @@
 
 `scoop search {{package}}`
 
-- List all known buckets (a bucket is an application repository):
-
-`scoop bucket known`
-
-- Add a bucket by its alias or a Git repository URL:
-
-`scoop bucket add {{bucket}}`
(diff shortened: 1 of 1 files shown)


sbrl:
Hey there, @LukBukkit! Thanks for the edit. Wow, this looks like an important addition to the page. That sounds like a good plan - let's do that. I suggest we remove the `bucket` examples in this PR, and then you open a new PR for a `scoop-bucket` page :-)

LukBukkit:
Hello @sbrl. Thanks for your fast answer, I'm removing the two bucket examples in this pull request and created another one for the new `scoop-bucket.md` page, which includes more and updated examples.

navarroaxel (approved):

owenvoke (approved):
LGTM! Much better to split the bucket commands. Scoop has so many useful commands. 👍🏻

sbrl (approved):
Thanks, @LukBukkit!

Haha you managed to catch me before I'd finished reviewing the other PRs :P

PR #4459, opened 2020-10-04
(chore) replace master/slave with inclusive language

qvissak:
In an effort to promote more inclusive language, replacing the words master/slave where applicable (other than in the names of repository/branch links). [GitHub is changing the default branch name](https://github.com/github/renaming) from `master` to `main`, so these examples will be more accurate in the future with the new terminology.

- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #4012, opened 2020-04-30
git-subtree: add page

einverne:
Try to add more git command.

- [ ] The page (if new), does not already exist in the repo.
- [ ] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [ ] The page has 8 or fewer examples.
- [ ] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [ ] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [ ] The page description includes a link to documentation or a homepage (if applicable).

PR #4533, opened 2020-10-06
git-*: use inclusive language

sbrl:
Suggestions for improving on #4459, which I reverted in #4532.

/cc everyone involved: @owenvoke, @mebeim, @navarroaxel, @qvissak, @agnivade.

### DIFF
--- a/pages.es/common/git-rebase.md
+++ b/pages.es/common/git-rebase.md
@@ -34,4 +34,4 @@
 
 - Resuelve automáticamente cualquier conflicto favoreciendo la versión de la rama en la que se esta trabajando (en este caso la palabra `theirs` tiene un significado invertido):
 
-`git rebase -X theirs {{branch_name}}`
+`git rebase -X theirs {{rama_de_reorganización}}`
--- a/pages.es/common/git-rebase.md
+++ b/pages.es/common/git-rebase.md
@@ -4,7 +4,7 @@
 > Se utiliza comúnmente para "mover" una rama entera a otra base, ya que crea copias de los commits en una nueva ubicación.
 > Más información: <https://git-scm.com/docs/git-rebase>.
 
-- Rebasa la rama actual en lo más alto de una rama:
+- Reorganiza la rama actual en lo más alto de otra rama:
 
 `git rebase {{branch_name}}`
 
--- a/pages.es/common/git-rebase.md
+++ b/pages.es/common/git-rebase.md
@@ -4,9 +4,9 @@
 > Se utiliza comúnmente para "mover" una rama entera a otra base, ya que crea copias de los commits en una nueva ubicación.
 > Más información: <https://git-scm.com/docs/git-rebase>.
 
-- Rebasa la rama actual en lo más alto de la rama master:
+- Rebasa la rama actual en lo más alto de la rama branch_name:
 
-`git rebase {{master}}`
+`git rebase {{branch_name}}`
 
 - Inicia un rebase interactivo que permite reordenar los commits, omitirlos, combinarlos o modificarlos:
 
@@ -34,4 +34,4 @@
 
 - Resuelve automáticamente cualquier conflicto favoreciendo la versión de la rama en la que se esta trabajando (en este caso la palabra `theirs` tiene un significado invertido):
 
-`git rebase -X theirs {{master}}`
+`git rebase -X theirs {{branch_name}}`
--- a/pages.es/common/git-rebase.md
+++ b/pages.es/common/git-rebase.md
@@ -6,7 +6,7 @@
 
 - Reorganiza la rama actual en lo más alto de otra rama:
 
-`git rebase {{branch_name}}`
+`git rebase {{rama_de_reorganización}}`
 
 - Inicia un rebase interactivo que permite reordenar los commits, omitirlos, combinarlos o modificarlos:
 
--- a/pages.es/common/git-rebase.md
+++ b/pages.es/common/git-rebase.md
@@ -4,7 +4,7 @@
 > Se utiliza comúnmente para "mover" una rama entera a otra base, ya que crea copias de los commits en una nueva ubicación.
 > Más información: <https://git-scm.com/docs/git-rebase>.
 
-- Rebasa la rama actual en lo más alto de la rama branch_name:
+- Rebasa la rama actual en lo más alto de una rama:
 
 `git rebase {{branch_name}}`
 
--- a/pages.es/common/git-rev-list.md
+++ b/pages.es/common/git-rev-list.md
@@ -7,7 +7,7 @@
 
 `git rev-list {{HEAD}}`
 
-- Muestra los commits más reciente a partir de una fecha y una rama específica:
+- Muestra los commits más recientes a partir de una fecha y una rama específica:
 
 `git rev-list --since={{'2019-12-01 00:00:00'}} {{nombre_de_rama}}`
 
--- a/pages.es/common/git-rev-list.md
+++ b/pages.es/common/git-rev-list.md
@@ -9,7 +9,7 @@
 
 - Muestra los commits más reciente a partir de una fecha y una rama específica:
 
-`git rev-list --since={{'2019-12-01 00:00:00'}} {{branch_name}}`
+`git rev-list --since={{'2019-12-01 00:00:00'}} {{nombre_de_rama}}`
 
 - Muestra todos los commits fusionados en un commit específico:
 
--- a/pages.es/common/git-rev-list.md
+++ b/pages.es/common/git-rev-list.md
@@ -9,7 +9,7 @@
 
 - Muestra los commits más reciente a partir de una fecha y una rama específica:
 
-`git rev-list --since={{'2019-12-01 00:00:00'}} {{master}}`
+`git rev-list --since={{'2019-12-01 00:00:00'}} {{branch_name}}`
 
 - Muestra todos los commits fusionados en un commit específico:
 
--- a/pages.es/common/git-revert.md
+++ b/pages.es/common/git-revert.md
@@ -13,7 +13,7 @@
 
 - Revierte múltiples commits:
 
-`git revert {{master~5..master~2}}`
+`git revert {{branch_name~5..branch_name~2}}`
 
 - No crea nuevos commits, solo cambia el árbol de trabajo:
 
--- a/pages.es/common/git-revert.md
+++ b/pages.es/common/git-revert.md
@@ -13,7 +13,7 @@
 
 - Revierte múltiples commits:
 
-`git revert {{branch_name~5..branch_name~2}}`
+`git revert {{rama~5..rama~2}}`
 
 - No crea nuevos commits, solo cambia el árbol de trabajo:
 
--- a/pages/common/git-rebase.md
+++ b/pages/common/git-rebase.md
@@ -6,7 +6,7 @@
 
 - Rebase the current branch on top of the another specified branch:
 
-`git rebase {{branch_name}}`
+`git rebase {{new_base_branch}}`
 
 - Start an interactive rebase, which allows the commits to be reordered, omitted, combined or modified:
 
--- a/pages/common/git-rebase.md
+++ b/pages/common/git-rebase.md
@@ -4,7 +4,7 @@
 > Commonly used to "move" an entire branch to another base, creating copies of the commits in the new location.
 > More information: <https://git-scm.com/docs/git-rebase>.
 
-- Rebase the current branch on top of the another specified branch:
+- Rebase the current branch on top of another specified branch:
 
 `git rebase {{new_base_branch}}`
 
--- a/pages/common/git-rebase.md
+++ b/pages/common/git-rebase.md
@@ -4,9 +4,9 @@
 > Commonly used to "move" an entire branch to another base, creating copies of the commits in the new location.
 > More information: <https://git-scm.com/docs/git-rebase>.
 
-- Rebase the current branch on top of the master branch:
+- Rebase the current branch on top of the another specified branch:
 
-`git rebase {{master}}`
+`git rebase {{branch_name}}`
 
 - Start an interactive rebase, which allows the commits to be reordered, omitted, combined or modified:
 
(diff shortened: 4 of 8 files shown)


pages.es/common/git-rebase.md
```diff
@@ -4,9 +4,9 @@
 > Se utiliza comúnmente para "mover" una rama entera a otra base, ya que crea copias de los commits en una nueva ubicación.
 > Más información: <https://git-scm.com/docs/git-rebase>.
 
-- Rebasa la rama actual en lo más alto de la rama master:
+- Rebasa la rama actual en lo más alto de la rama branch_name:
```
mebeim, line 7:
```suggestion
- Rebasa la rama actual en lo más alto de una rama:
```
Change "the branch branch_name" to "a branch".

mebeim (changes requested):
Minor nitpick for the one spanish page, otherwise looks good to me.

pages.es/common/git-rebase.md
sbrl: Thanks! I forgot to mention that the `branch_name` needed translating here, since I don't speak the language haha

pages.es/common/git-rebase.md
```diff
@@ -4,9 +4,9 @@
 > Se utiliza comúnmente para "mover" una rama entera a otra base, ya que crea copias de los commits en una nueva ubicación.
 > Más información: <https://git-scm.com/docs/git-rebase>.
 
-- Rebasa la rama actual en lo más alto de la rama master:
+- Rebasa la rama actual en lo más alto de una rama:
```
navarroaxel, line 7: I'm from Argentina but I like to read docs in English :grimacing:.  `Rebasa` is the worst Spanish translation for `rebase` lol. The [official git](https://git-scm.com/book/es/v2/Ramificaciones-en-Git-Reorganizar-el-Trabajo-Realizado) use `reorganización`.
```suggestion
- Reorganiza la rama actual en lo más alto de otra rama:
```

qvissak (approved):

pages.es/common/git-rebase.md
```diff
@@ -4,9 +4,9 @@
...
 
-`git rebase {{master}}`
+`git rebase {{branch_name}}`
```
waldyrious, line 9: It's unclear whether `branch_name` refers to the current branch or the target branch. We might want to use a more explicit token name here.

pages.es/common/git-rebase.md
```diff
@@ -34,4 +34,4 @@
 
 - Resuelve automáticamente cualquier conflicto favoreciendo la versión de la rama en la que se esta trabajando (en este caso la palabra `theirs` tiene un significado invertido):
 
-`git rebase -X theirs {{master}}`
+`git rebase -X theirs {{branch_name}}`
```
waldyrious, line 37: I'm wondering if we shouldn't translate "branch_name" to Spanish, as there's no need for it to be in English.

pages/common/git-rebase.md
```diff
@@ -4,9 +4,9 @@
 > Commonly used to "move" an entire branch to another base, creating copies of the commits in the new location.
 > More information: <https://git-scm.com/docs/git-rebase>.
 
-- Rebase the current branch on top of the master branch:
+- Rebase the current branch on top of the another specified branch:
 
-`git rebase {{master}}`
+`git rebase {{branch_name}}`
```
waldyrious, line 9: As mentioned above, 

> It's unclear whether `branch_name` refers to the current branch or the target branch. We might want to use a more explicit token name here.

pages/common/git-subtree.md
```diff
@@ -5,19 +5,19 @@
...
-`git subtree add --prefix={{path/to/directory/}} --squash {{repository_url}} {{master}}`
+`git subtree add --prefix={{path/to/directory/}} --squash {{repository_url}} {{branch_name}}`
 
 - Update subtree repository to its latest commit:
 
-`git subtree pull --prefix={{path/to/directory/}} {{repository_url}} {{master}}`
+`git subtree pull --prefix={{path/to/directory/}} {{repository_url}} {{branch_name}}`
 
-- Merge a subtree repository into master:
+- Merge a subtree repository into a specified branch:
```
waldyrious, line 14: I'm not sure if the last parameter refers to the source or the target branch. As the [author of this page](https://github.com/tldr-pages/tldr/pull/4012), could you comment, @einverne?

pages/common/git-subtree.md
```diff
@@ -5,19 +5,19 @@
 
 - Add a git repository as a subtree:
 
-`git subtree add --prefix={{path/to/directory/}} --squash {{repository_url}} {{master}}`
+`git subtree add --prefix={{path/to/directory/}} --squash {{repository_url}} {{branch_name}}`
```
waldyrious, line 8: Is the final parameter mandatory? Maybe we should omit it in most examples if it defaults to the primary branch, which I'd suppose is what most people would want.

pages.es/common/git-rebase.md
mebeim: You're definitely right, the token should be translated.

pages/common/git-rebase.md
navarroaxel:
```suggestion
`git rebase {{base_branch}}`
```

pages.es/common/git-rebase.md
navarroaxel:
```suggestion
`git rebase {{rama_base}}`
```

pages.es/common/git-rebase.md
navarroaxel:
```suggestion
`git rebase -X theirs {{base_branch}}`
```

pages.es/common/git-rebase.md
waldyrious: In English, I'd go with `target_branch` or `new_base_branch`. That's because the very notion of rebasing implies two bases, the old one and the new one, so merely base_branch would still be ambiguous as it could refer to either base. How would you suggest we express this concept in Spanish?

pages.es/common/git-rebase.md
waldyrious: I don't understand your suggestion, @navarroaxel — can you clarify the reasoning for it?

pages/common/git-rebase.md
waldyrious: I'll have to quote myself again (the two discussions are the same, only in two different languages):

> I'd go with `target_branch` or `new_base_branch`. That's because the very notion of rebasing implies two bases, the old one and the new one, so merely `base_branch` would still be ambiguous as it could refer to either base.

pages.es/common/git-rebase.md
navarroaxel: Re-reading again [the docs](https://git-scm.com/book/es/v2/Ramificaciones-en-Git-Reorganizar-el-Trabajo-Realizado) they use the term:
```suggestion
`git rebase {{rama_de_reorganización}}`
```

pages/common/git-rebase.md
navarroaxel: Reading [the docs](https://git-scm.com/book/en/v2/Git-Branching-Rebasing#_more_interesting_rebases:~:text=More%20Interesting%20Rebases), in `More Interesting Rebases` section, they use the term:
```suggestion
`git rebase {{rebase_target_branch}}`
```

pages/common/git-subtree.md
qvissak: I would say that is outside the scope of this PR.

qvissak:
To combat ambiguity, this is why in my original PR I used `main` since this is the new official terminology by [GitHub](https://github.com/github/renaming), and the direction that [GitLab](https://gitlab.com/gitlab-org/gitlab/-/issues/222204) and [BitBucket](https://jira.atlassian.com/browse/BCLOUD-20212) will likely take. BitBucket, however, is waiting for `git init` to change the default branch from `master` to `main`.

I think by definition, the word `main` is clear. No need to guess what `base_branch`, `target_branch`, `rebase_target_branch`, etc. means.

pages.es/common/git-rebase.md
sbrl: Yeah, we should definitely translate here. I just inserted it as a suggestion as I unfortunately don't know the language

pages/common/git-rebase.md
sbrl: Great point, @waldyrious. `new_base_branch` sounds the most clear to me, so how about we go with that?

pages/common/git-subtree.md
sbrl: It doesn't hurt to review it now I don't think, @qvissak.

Anyway, it looks like only the repository URL is optional according to the man page @waldyrious, as I see the following in the usage summary at the top:

```
git subtree add   -P <prefix> <commit>
git subtree add   -P <prefix> <repository> <ref>
```

You'd have thought it would be the commit ref that's optional, but it's the repository url instead :woman_shrugging:

Interestingly we mention here `branch_name`, but it's actually more like a `commit_ish` according to the man page. That said, I havevn't actually used this command personally, so I wouldn't know.

pages/common/git-subtree.md
sbrl: Looking at the man page, it details this:

> Merge recent changes up to <commit> into the <prefix> subtree.

So I think that it's actually going in the opposite direction to that of the description here.

Again, I haven't actually used the command before, so I'm not sure. Let's wait for @einverne's response.

einverne: According to the document, it says:

> git subtree merge -P <prefix> <commit>
> Merge recent changes up to <commit> into the <prefix> subtree.

What the `subtree merge` do is to merge the latest commit to the subtree.

sbrl: Ohhh I see. Thanks for the insight here, @einverne! So could we reword to `{{subtree_branch_name}}` here then perhaps?

sbrl:
Oh I see, @qvissak! I haven't actually seen that being the new official terminology yet - as I don't create new repositories very often - hence the confusion :P

Either way, I think replacing the concrete example here of a branch name with something more descriptive will aid clarity too.

pages/common/git-subtree.md
einverne: @sbrl Yes, and the origin description may not be accurate. How about

> Merge recent changes up to {{latest_subtree_commit}} into the subtree.

pages/common/git-rebase.md
sbrl: Applied `new_base_branch`, so marking this as resolved. Feel free to unresolve if this isn't the case.

pages/common/git-subtree.md
sbrl: Applied suggestion, @einverne.

Is this better?

pages/common/git-subtree.md
sbrl: Marking as resolved, but feel free to unresolve if this isn't the case.

pages.es/common/git-rebase.md
waldyrious: @navarroaxel would `{{nombre_de_rama}}` or simply `{{rama}}` work here?

pages.es/common/git-rev-list.md
```diff
@@ -9,7 +9,7 @@
 
 - Muestra los commits más reciente a partir de una fecha y una rama específica:
 
-`git rev-list --since={{'2019-12-01 00:00:00'}} {{master}}`
+`git rev-list --since={{'2019-12-01 00:00:00'}} {{branch_name}}`
```
waldyrious, line 12: Same as [above](https://github.com/tldr-pages/tldr/pull/4533/files#r500594375) — this should be translated.

pages.es/common/git-revert.md
```diff
@@ -13,7 +13,7 @@
 
 - Revierte múltiples commits:
 
-`git revert {{master~5..master~2}}`
+`git revert {{branch_name~5..branch_name~2}}`
```
waldyrious, line 16: Same as [above](https://github.com/tldr-pages/tldr/pull/4533/files#r500594375) — these should be translated.

pages.es/common/git-rebase.md
navarroaxel: I think rama is OK:
```suggestion
`git rebase -X theirs {{rama}}`
```
but if we use `rama_de_reorganización` above, this is another option:
```suggestion
`git rebase -X theirs {{rama_de_reorganización}}`
```
If you now English `theirs {{rama}}` has a lot of sense, otherwise `theirs {{rama_de_reorganización}}` is better.

pages.es/common/git-rev-list.md
navarroaxel:
```suggestion
`git rev-list --since={{'2019-12-01 00:00:00'}} {{nombre_de_rama}}`
```

pages.es/common/git-revert.md
navarroaxel:
```suggestion
`git revert {{rama~5..rama~2}}`
```

mebeim (approved):
I've applied @navarroaxel suggestions for pages under `pages.es`. LGTM now!

pages/common/git-rebase.md
```diff
@@ -4,9 +4,9 @@
 > Commonly used to "move" an entire branch to another base, creating copies of the commits in the new location.
 > More information: <https://git-scm.com/docs/git-rebase>.
 
-- Rebase the current branch on top of the master branch:
+- Rebase the current branch on top of the another specified branch:
```
waldyrious, line 7:
```suggestion
- Rebase the current branch on top of another specified branch:
```

pages.es/common/git-rev-list.md
```diff
@@ -9,7 +9,7 @@
 
 - Muestra los commits más reciente a partir de una fecha y una rama específica:
```
waldyrious, line 10: This is not strictly related to this PR, but while we're at it, we should fix the typo:
```suggestion
- Muestra los commits más recientes a partir de una fecha y una rama específica:
```

@navarroaxel can you confirm please?

waldyrious (changes requested):
Just a couple of additional tweaks and should be good to go!

pages.es/common/git-rev-list.md
navarroaxel: Yep, it's good.

sbrl:
Suggestions applied, @waldyrious :-)

waldyrious (approved):
Alright, let's go! 🚀🚀

PR #4554, opened 2020-10-07
scoop-bucket: add page

LukBukkit:

Hello,

this is the follow-up pull request for #4531.
I'm creating a new `scoop-bucket.md` page to keep the eight example limit on the `scoop.md` page. 

If you compare the two old bucket examples on the `scoop.md` page to this page, you may see that I've added three options.
1. I guess it's useful to see, which buckets you are currently using: `scoop bucket list`
2. I've had to split up the `scoop bucket add` command, because the syntax has changed.
If you want to add an unknown bucket, you have to set a name and a repository url.
3. To complete the lifecycle of buckets, there's also an example how to remove them.

### DIFF
--- a/pages/windows/scoop-bucket.md
+++ b/pages/windows/scoop-bucket.md
@@ -1,7 +1,7 @@
 # scoop bucket
 
-> A bucket is a Git repository containing files which describe how scoop installs applications.
-> If Scoop doesn't know where the bucket is located we have to specify its repository location.
+> Manage buckets: Git repositories containing files which describe how scoop installs applications.
+> If Scoop doesn't know where the bucket is located its repository location must be specified.
 > More information: <https://github.com/lukesampson/scoop/wiki/Buckets>.
 
 - List all buckets currently in use:
--- a/pages/windows/scoop-bucket.md
+++ b/pages/windows/scoop-bucket.md
@@ -1,6 +1,6 @@
 # scoop bucket
 
-> A bucket is Git repository containing files which describe how scoop installs applications.
+> A bucket is a Git repository containing files which describe how scoop installs applications.
 > If Scoop doesn't know where the bucket is located we have to specify its repository location.
 > More information: <https://github.com/lukesampson/scoop/wiki/Buckets>.
 
--- a/pages/windows/scoop-bucket.md
+++ b/pages/windows/scoop-bucket.md
@@ -1,7 +1,7 @@
 # scoop bucket
 
-> A bucket is an application repository.
-> Some buckets are already known to scoop, so you don't have to provide an url.
+> A bucket is Git repository containing files which describe how scoop installs applications.
+> If Scoop doesn't know where the bucket is located we have to specify its repository location.
 > More information: <https://github.com/lukesampson/scoop/wiki/Buckets>.
 
 - List all buckets currently in use:
@@ -16,9 +16,9 @@
 
 `scoop bucket add {{name}}`
 
-- Add an unknown bucket by its name and Git repository url:
+- Add an unknown bucket by its name and Git repository URL:
 
-`scoop bucket add {{name}} {{repository}}`
+`scoop bucket add {{name}} {{https://example.com/repository.git}}`
 
 - Remove a bucket by its name:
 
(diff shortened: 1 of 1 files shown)


pages/windows/scoop-bucket.md
```diff
@@ -0,0 +1,25 @@
+# scoop bucket
+
+> A bucket is an application repository.
```
navarroaxel, line 3: Based on the wiki page:
```suggestion
> A bucket is a Git repository with a manifest which describe how to install an application.
```
too verbose?

pages/windows/scoop-bucket.md
```diff
@@ -0,0 +1,25 @@
...
+> Some buckets are already known to scoop, so you don't have to provide an url.
```
navarroaxel, line 4: Is this better?
```suggestion
> if Scoop doesn't know where the bucket is located we should specify its repository location.
```

navarroaxel (changes requested):
Thanks @LukBukkit, I left some comments.

pages/windows/scoop-bucket.md
```diff
@@ -0,0 +1,25 @@
...
+
+`scoop bucket known`
+
+- Add a known bucket by its name:
+
+`scoop bucket add {{name}}`
+
+- Add an unknown bucket by its name and Git repository url:
+
+`scoop bucket add {{name}} {{repository}}`
```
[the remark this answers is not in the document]
LukBukkit: Yeah this looks indeed a lot better.
According to my understanding you can choose the name freely. For a lot of the repositories (take a look at [this list](https://scoop.netlify.app/buckets/) and view some GitHub pages of repositories) the repository URL is somewhat simillar to the suggested name, but isn't not always equals, so I would choose the second option.

pages/windows/scoop-bucket.md
LukBukkit: This also a good idea, but I would put it a little bit more general. Maybe the term `manifest` could confuse new users. 
```
A bucket is Git repository containing files which describe how scoop installs applications.
```

pages/windows/scoop-bucket.md
LukBukkit: Yeah this also does sound a lot better. Just one small remark: Maybe we can replace the `should` with a `have to`. By using `should` this does sound more like a suggestion, rather than an instruction.

LukBukkit:
Thanks for these helpful comments. I reacted to them and pushed a new commit (I guess you can squash it a later or I do) with your suggested changes or a small variation of it, which we can discuss above.

pages/windows/scoop-bucket.md
```diff
@@ -0,0 +1,25 @@
+# scoop bucket
+
+> A bucket is Git repository containing files which describe how scoop installs applications.
```
navarroaxel, line 3: Just a typo
```suggestion
> A bucket is a Git repository containing files which describe how scoop installs applications.
```

navarroaxel (changes requested):

pages/windows/scoop-bucket.md
LukBukkit: You're correct. Thanks :D

navarroaxel (approved):

sbrl:
Yep, we have a button to squash when we merge @LukBukkit :-)

pages/windows/scoop-bucket.md
```diff
@@ -0,0 +1,25 @@
+# scoop bucket
+
+> A bucket is a Git repository containing files which describe how scoop installs applications.
+> If Scoop doesn't know where the bucket is located we have to specify its repository location.
```
sbrl, line 4: Perhaps this will read better?
```suggestion
> Manage buckets: Git repositories containing files which describe how scoop installs applications.
> If Scoop doesn't know where the bucket is located its repository location must be specified.
```

navarroaxel (approved):

pages/windows/scoop-bucket.md
LukBukkit: Yeah, that's even better

sbrl (approved):
Thanks, @LukBukkit :slightly_smiling_face:

sbrl (approved):

PR #4557, opened 2020-10-07
dotnet: add German translation

Abrynos:


pages.de/common/dotnet.md
```diff
@@ -0,0 +1,20 @@
+# dotnet
+
+> Plattformübergreifende Kommandozeilenandwendungen für .NET Core.
+> Mehr Informationen: <https://docs.microsoft.com/de-de/dotnet/core/tools/>.
+
+- Ein neues .NET Projekt initialisieren:
+
+`dotnet new {{template_name}}`
```
navarroaxel, line 8: You can translate the `template_name` here to German.

pages.de/common/dotnet.md
```diff
@@ -0,0 +1,20 @@
+# dotnet
+
+> Plattformübergreifende Kommandozeilenandwendungen für .NET Core.
+> Mehr Informationen: <https://docs.microsoft.com/de-de/dotnet/core/tools/>.
```
navarroaxel, line 4: We don't use i18n in MSFT URLs because he choose your language automatically. Should we do the same in non-English pages? cc @owenvoke @schneiderl 
```suggestion
> Mehr Informationen: <https://docs.microsoft.com/dotnet/core/tools/>.
```

sbrl: Yeah, that definitely seems like a good idea.

owenvoke: Yeah, I think it's best across all pages with MS links.

pages.de/common/dotnet.md
gutjuri: "name" in german is "name" and I would consider "template" to be common knowledge.

Abrynos: @gutjuri aparently had the same train of thought as me

navarroaxel (approved):

owenvoke (approved):

PR #4623, opened 2020-10-10
git-stash: add French translation

Hyftar:

This PR adds a French translation of the `git stash` command.

### DIFF
--- a/pages.fr/common/git-stash.md
+++ b/pages.fr/common/git-stash.md
@@ -1,0 +1,36 @@
+# git stash
+
+> Mettre les changements locaux de Git dans un endroit temporaire (Le stash).
+> Plus d'informations: <https://git-scm.com/docs/git-stash>.
+
+- Mettre les changements courants (sauf ceux provenant de fichiers qui ne sont pas suivi) dans un *stash*:
+
+`git stash [push -m {{message_optionnel_pour_le_stash}}]`
+
+- Mettre les changements courants (incluant les fichiers qui ne sont pas suivi) dans un *stash*:
+
+`git stash -u`
+
+- Selectionner interactivement les parties des fichiers changer à mettre dans un *stash*:
+
+`git stash -p`
+
+- Lister tout les *stash* (Montre le nom du *stash*, la branche reliée et le message):
+
+`git stash list`
+
+- Appliquer un *stash* à la tête git (le dernier stash est celui choisi par défaut, aussi appellé `stash@{0}`):
+
+`git stash apply {{nom_du_stash_ciblé_optionnellement}}`
+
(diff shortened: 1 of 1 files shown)


pages.fr/common/git-stash.md
```diff
@@ -0,0 +1,36 @@
+# git stash
+
+> Mettre les changements locaux de Git dans un endroit temporaire (Le stash).
+> Plus d'informations: <https://git-scm.com/docs/git-stash>.
+
+- Mettre les changements courants (sauf ceux provenant de fichiers qui ne sont pas suivi) dans un *stash*:
+
+`git stash [push -m {{optional_stash_message}}]`
```
schneiderl, line 8: Is this example parameter already translated? I've quickly thrown it into a translator and it seems to be already translated but also it looks exactly like English so just making sure.

Hyftar: Oops, my bad I completely missed the parameter. It is indeed still English.

navarroaxel (approved):
LGTM @Hyftar, please don't `push --force` to the PR's branch, it's harder to us to review your changes and track the suggestions. We squash before merge. :grin:

pages.fr/common/git-stash.md
```diff
@@ -0,0 +1,36 @@
+# git stash
+
+> Mettre les changements locaux de Git dans un endroit temporaire (Le stash).
+> Plus d'informations: <https://git-scm.com/docs/git-stash>.
+
+- Mettre les changements courants (sauf ceux provenant de fichiers qui ne sont pas suivi) dans un *stash*:
```
sbrl, line 6: Any particular reason `*stash*` is in italics here? Just curious, since this isn't the case in the original page.

I'm unsure as to whether this will be rendered correctly - I can't remember whether we've done this elsewhere or not (I don't remember any specific examples off-hand).

Hyftar: Because there isn't a universal French equivalent for stash so I used the English word but since it's still in English I have to put it in italics.

sbrl: Ah, I see. This isn't technically against the client spec, so this should be ok. I'd just like another opinion  before we go ahead though, if you don't mind waiting on this PR for a little while.

schneiderl (approved):

pages.fr/common/git-stash.md
schneiderl: @sbrl If I'm not completely mistaking, we already do this in some pt_BR pages so I believe this is fine.

sbrl: Oh, sounds good! Just wanted to check. In that case, it's fine as-is.

sbrl (approved):

PR #4624, opened 2020-10-10
yes: add French translation, add common usage example

Hyftar:

This PR adds a French translation of the `yes` command and a new usage example to illustrate a common use.

### DIFF
--- a/pages/common/yes.md
+++ b/pages/common/yes.md
@@ -1,6 +1,7 @@
 # yes
 
 > Output something repeatedly.
+> This command is commonly used to answer yes to every prompt by install commands (such as apt-get).
 
 - Repeatedly output "message":
 
@@ -9,3 +10,7 @@
 - Repeatedly output "y":
 
 `yes`
+
+- Agree to everything prompted by the `sudo apt-get` command:
+
+`sudo apt-get install {{program}} | yes`
--- a/pages/common/yes.md
+++ b/pages/common/yes.md
@@ -13,4 +13,4 @@
 
 - Agree to everything prompted by the `sudo apt-get` command:
 
-`sudo apt-get install {{program}} | yes`
+`yes | sudo apt-get install {{program}}`
--- a/pages/common/yes.md
+++ b/pages/common/yes.md
@@ -3,7 +3,7 @@
 > Output something repeatedly.
 > This command is commonly used to answer yes to every prompt by install commands (such as apt-get).
 
-- Repeatedly output "message":
+- Repeatedly output "message" :
 
 `yes {{message}}`
 
@@ -11,6 +11,6 @@
 
 `yes`
 
-- Agree to everything prompted by the `sudo apt-get` command:
+- Accept everything prompted by the `apt-get` command:
 
 `yes | sudo apt-get install {{program}}`
--- a/pages/common/yes.md
+++ b/pages/common/yes.md
@@ -3,7 +3,7 @@
 > Output something repeatedly.
 > This command is commonly used to answer yes to every prompt by install commands (such as apt-get).
 
-- Repeatedly output "message" :
+- Repeatedly output "message":
 
 `yes {{message}}`
 
--- a/pages.fr/linux/yes.md
+++ b/pages.fr/linux/yes.md
@@ -14,4 +14,4 @@
 
 - Accepter tout ce qui est demandé par la commande `sudo apt-get`:
 
-`sudo apt-get install {{program}} | yes`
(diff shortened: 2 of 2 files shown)


schneiderl:
Hi,
I believe it would be better to have those changes separated into two different commits and two different PRs. 
The changes have no relation between them: one is changing an English page and the other is adding a French page.

No worries though, the changes are small so I believe we can continue with the review process within this pull request. Just take that in my for future PRs.

Also, thank you for your contribution, it looks great to me (even though I don't speak French). 
Be welcome to keep contributing to tldr, it is much appreciated.

schneiderl (approved):

SethiPandi:
Just my opinion, I think it's great to have an example that uses `yes` via a pipe (`|`), but is `apt-get` really the best example for this?

`apt-get` already has a built-in `-y` argument which is much safer to use than piping from `yes`.

That makes the example a bit redundant, and very unsafe compared to the built-in `-y` argument.
If the point of this example is to say yes automatically, even to unsafe operations, it might be good to document that it will do so?

```
-y, --yes, --assume-yes
           Automatic yes to prompts; assume "yes" as answer to all prompts and
           run non-interactively. If an undesirable situation, such as
           changing a held package, trying to install a unauthenticated
           package or removing an essential package occurs then apt-get will
           abort. Configuration Item: APT::Get::Assume-Yes.
```

Hyftar:
I agree that `apt-get` might not be the safest choice, it's just the first that came to my mind. If you have another command that we could use that would be safer for the average user I would gladly change it in my examples.

SethiPandi:
I've been giving it some thought, but I can't really think of a good example for this that wouldn't be unsafe, or that there is a built-in argument to handle that case. In cases where there isn't a `-y` or equivalent, it still might not be the best idea to spam `y` without contextualizing the prompts.

I was just _especially_ hesitant on this particular example as I'm wary it might encourage poor practices to users that are new to `apt-get` if they aren't aware of the `-y` argument yet. (Which is also why I thought documented that in the example might be nice.)

If anyone else has a nice example, that'd be great, otherwise I feel just a note on the matter might suffice personally.

---

I actually tested this your example a bit locally though and just want to check, are you sure the command isn't meant to be:

```bash
yes | sudo apt-get install {{program}}
```
> This responds `y` to all interactive prompts. I'm succesfully able to install software this way.

Not:
```bash
sudo apt-get install {{program}} | yes
```
> This just hides the command input and spams `y` until killed. The software does not get installed.

Another test I did was with `npm init`.

```bash
yes "" | npm init
```
> This responds with nothing to all prompts, and generates the default `package.json` file.

Not:
```bash
npm init | yes ""
```
> This just hides the command input and spams `y` until killed. The `package.json` does not get created.

Hyftar:
Oh you are indeed right, I'll fix both points shortly. Thanks you for your attention to details!

sbrl:
Perhaps such an example would be best put on the page for `yes`?

sbrl (commented):
Oh oops! I  see what those comments were about now.

I don't see anything wrong with this, since there isn't much else to the `yes` command and this is one of it's intended use-cases. `apt-get` does have the `-y` argument though, so I'm unsure as to whether it's the best example. If there isn't another alternative though this will be fine.

Deuchnord (changes requested):
Nice translation!

sbrl (approved):

sbrl:
Also, next time, perhaps you could avoid force-pushing please? It makes it harder to tell what's changed when reviewing, and we've got a special button that allows us to squash easily on merge :smiley_cat:

Hyftar:
> next time, perhaps you could avoid force-

I only forced push my last commit because it was missing a file. 🙂

sbrl:
Oh, no worries then :-)

pages/common/yes.md
```diff
@@ -3,14 +3,14 @@
 > Output something repeatedly.
 > This command is commonly used to answer yes to every prompt by install commands (such as apt-get).
 
-- Repeatedly output "message":
+- Repeatedly output "message" :
```
Deuchnord, line 6: I think you have added a space that is not required here ^^

navarroaxel:
```suggestion
- Repeatedly output "message":
```

owenvoke (approved):
I guess people are more likely to use the `-y` flag for `apt-get`, but makes sense as an example of the usage. 👍🏻 Nice work.

sbrl (approved):

PR #4628, opened 2020-10-10
module: add page

cbohn4:

As a short summary, the module command allows for loading in different programs, or modules, into the user environment. This is useful in the case of HPC systems that may have numerous pieces of software that do not need to be loaded to run a job.

### DIFF
--- a/pages/linux/module.md
+++ b/pages/linux/module.md
@@ -1,0 +1,28 @@
+# module
+
+> Modify a users' environment using the module command.
+> More information: <https://modules.readthedocs.io/en/latest/module.html>.
+
+- Display available modules:
+
+`module avail`
+
+- Search for a module by name:
+
+`module spider {{module_name}}`
+
+- Load a module:
+
+`module load {{module_name}}`
+
+- Display loaded modules:
+
+`module list`
+
+- Unload a specific loaded module:
+
+`module {{module_name}}`
+
+- Unload all loaded modules:
+
+`module purge`
--- a/pages/linux/module.md
+++ b/pages/linux/module.md
@@ -1,7 +1,7 @@
 # module
 
 > Modify a users' environment using the module command.
-> More information: <https://modules.readthedocs.io/en/latest/module.html>.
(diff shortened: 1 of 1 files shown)


pages/linux/module.md
```diff
@@ -0,0 +1,28 @@
...
+> Modify a users' environment using the module command.
+> More information: <https://modules.readthedocs.io/en/latest/module.html>.
+
+- Display available modules:
+
+`module avail`
+
+- Search for a module by name:
+
+`module spider {{module_name}}`
```
navarroaxel, line 12: I can't find the `spider` subcommand in the more info link that you provided. Are you talking about `Lmod` ? <https://lmod.readthedocs.io/en/latest/>

navarroaxel (changes requested):
Hi @cbohn4, I have a question.

pages/linux/module.md
cbohn4: Yeah, that would be it. [https://lmod.readthedocs.io/en/latest/010_user.html](https://lmod.readthedocs.io/en/latest/010_user.html) might be a better link specific to the lmod module command in terms of function rather than the docs for the full lmod ecosystem?. Would it be better practice in the TLDR repository to include the page more specific to the ecosystem or to `module` itself?

sbrl: Is the whole page about lmod, or just this example? I'm somewhat confused here (despite the fact that I use the `module` on my University's HPC).

cbohn4: The page @navarroaxel posted is for all of Lmod itself. The link in my previous comment is on the module portion itself user side rather than the background that app specialists would handle like the module files. For the sake of `module` in general use, the link I sent, I believe, would be better, but either would be appropriate.

sbrl: Ah, I see. I'm still somewhat confused, but let's go with the one you think would be more appropriate here.

cbohn4: Sounds good.

owenvoke (approved):

navarroaxel (approved):

PR #4668, opened 2020-10-12
title: add pt_BR translation

isrmicha:
adding pt_BR translation to windows/title.md

### DIFF
--- a/pages.pt_BR/windows/title.md
+++ b/pages.pt_BR/windows/title.md
@@ -1,0 +1,8 @@
+# title
+
+> Exibe o título do prompt de comando.
+> Mais informações: <https://docs.microsoft.com/windows-server/administration/windows-commands/title>.
+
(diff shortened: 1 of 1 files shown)


navarroaxel (approved):

pages.pt_BR/windows/title.md
```diff
@@ -0,0 +1,8 @@
+# title
+
+> Exibe o título do prompt de comando.
```
schneiderl, line 3: It seems to me that the command description and the examples are incoherent: 
In the description, we say it shows the title, and we only have an example to change the prompt title. 
Perhaps it would be good to also add an example on how to show the prompt title. This would need to be done on both files: English and Portuguese. Perhaps you want to tackle this? 

Since this is a translation request I will approve it as this is not a translation problem.

schneiderl (approved):

PR #4691, opened 2020-10-14
docker: add German translation

Abrynos:


navarroaxel (changes requested):
Hey @Abrynos, thanks for your submission.

Can you translate the variables like `{{container_name}}` and `{{image}}` for make this easier for non-English speakers?

Remember that for translations you can send more than 1 translated page in 1 PR, just keep 1 commit per file and no more than 10 pages. This doesn't apply to the primary «English» pages.

Abrynos:
Well... "name" is also the german word for "name" and we use the words "container" and "image" in german as well.

If we translated "image" verbatim we would end up with "Bild", which may be a correct translation for the word, but is completely wrong in the context of docker, as it has the meaning of "picture" and is not associated with being a template for something else. Same problem occurs if we verbatim-translate "container". I've never heard a university professor use any form of german translation. (Also the word "container" is already part of the german language used to describe containers on ships).

navarroaxel (approved):
Thanks for the clarification @Abrynos, I don't know the language but the format looks good.

sbrl (approved):

Issue #2680, opened 2019-01-04
page request: iex/elixir


shmink:
iex being the interactive terminal for the elixir language. But there is also `elixir` as well they work very similarly only seeming to have a few differences. Might be able to put both into the one page? I'll look into doing this myself in future.

https://www.mankier.com/1/iex

sbrl:
Not sure that it would work with our current structure having just 1 page. If you'd like to attempt it yourself, that'd be great :D

bl-ue:
`iex` and `elixir` were added by #4658 and #4693.  Thank you @Schoonology!

PR #4693, opened 2020-10-14
elixir: add page

Schoonology:
Though the `elixir` command has multiple uses and purposes, the included examples are the simplest of them. This addresses the second half of #2680, and likely closes it.

### DIFF
--- a/pages/common/elixir.md
+++ b/pages/common/elixir.md
@@ -1,0 +1,12 @@
+# elixir
+
+> Elixir programming language interpreter.
+> More information: <https://elixir-lang.org/>.
+
+- Run an Elixir file:
+
+`elixir {{path/to/file}}`
+
+- Evaluate Elixir by passing it in the command:
+
+`elixir -e "{{code}}"`
--- a/pages/common/elixir.md
+++ b/pages/common/elixir.md
@@ -1,7 +1,7 @@
 # elixir
 
 > Elixir programming language interpreter.
-> More information: <https://elixir-lang.org/>.
+> More information: <https://elixir-lang.org>.
 
 - Run an Elixir file:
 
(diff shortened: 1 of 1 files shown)


pages/common/elixir.md
```diff
@@ -0,0 +1,12 @@
+# elixir
+
+> Elixir programming language interpreter.
+> More information: <https://elixir-lang.org/>.
```
navarroaxel, line 4:
```suggestion
> More information: <https://elixir-lang.org>.
```

navarroaxel (changes requested):
Thanks @Schoonology, just 1 change.

pages/common/elixir.md
```diff
@@ -0,0 +1,12 @@
...
+
+- Run an Elixir file:
+
+`elixir {{path/to/file}}`
+
+- Evaluate Elixir by passing it in the command:
```
sbrl, line 10: Perhaps this will read better?
```suggestion
- Evaluate Elixir code by passing it as an argument:
```

navarroaxel: @sbrl he copied the example description from the `node` page

sbrl: Ah, I see. I think this makes more sense though. Maybe we should update the `node` page in a separate PR :thinking:

Schoonology: I did. It feels like there are _many_ ways we could word this, and that it might be better to focus on what someone using `tldr` to go from one language to another might expect.

I'm more than happy to change this, certainly if we want to move toward a more canonical expression for these interpreters (e.g. `ruby`, `node`, `python`).

navarroaxel: @Schoonology please apply the suggestion from @sbrl and send another PR to edit the node page with this grammar.

owenvoke (approved):

navarroaxel (approved):

PR #4703, opened 2020-10-16
git-commit: clarify amend example

andrealmeid:
`git commit -a ` amends (or, in a more simple way, "adds") current staged changes on the last commit. The way is currently written gives the impression that changes on the last commit will be deleted and replaced by those new changes.

I understand why this was written in this way, since in `git commit --help` there is:

```
--amend

    Replace the tip of the current branch by creating a new commit. [...]
```   
But notice that "replace" here is referring to the commit, that will be replaced by a new commit that has the combination of changes of the old commit and the new added changes.

---

### DIFF
--- a/pages/common/git-commit.md
+++ b/pages/common/git-commit.md
@@ -11,7 +11,7 @@
 
 `git commit -a -m {{message}}`
 
-- Update the last commit by adding the currently staged changes, changing commit's hash:
+- Update the last commit by adding the currently staged changes, changing the commit's hash:
 
 `git commit --amend`
 
--- a/pages/common/git-commit.md
+++ b/pages/common/git-commit.md
@@ -11,7 +11,7 @@
 
 `git commit -a -m {{message}}`
 
-- Replace the last commit with currently staged changes:
+- Update the last commit by adding the currently staged changes, changing commit's hash:
 
 `git commit --amend`
 


pages/common/git-commit.md
```diff
@@ -11,7 +11,7 @@
 
 `git commit -a -m {{message}}`
 
-- Replace the last commit with currently staged changes:
+- Add currently staged changes on the last commit:
```
sbrl, line 14: Hrm, I see your point. However, the new description here implies that commits can be changed without altering their hash value. What about this?
```suggestion
- Update the last commit by adding the currently staged changes (this changes the commit's hash):
```

andrealmeid:
@sbrl I did some changes based on your suggestion, please have a look

navarroaxel:
@andrealmeid I think I'm seeing the same text that you send the first time.

andrealmeid:
@navarroaxel ops, ironically, I messed with my git commit :P have a look now

pages/common/git-commit.md
```diff
@@ -11,7 +11,7 @@
 
 `git commit -a -m {{message}}`
 
-- Replace the last commit with currently staged changes:
+- Update the last commit by adding the currently staged changes, changing commit's hash:
```
sbrl, line 14: This looks better, but there's a missing article here. Perhaps this would read better?
```suggestion
- Update the last commit by adding the currently staged changes, changing the commit's hash:
```

andrealmeid: @sbrl it should be good now

navarroaxel (approved):

sbrl (approved):

Issue #4685, opened 2020-10-14
page request: zmv


hrai:
A page for zsh's utility function - [zmv](http://zsh.sourceforge.net/Doc/Release/User-Contributions.html)

navarroaxel:
Thanks for requesting this page! Do you want to have a go yourself?

Pages are best created by people who understand the command they are about.

If you do, take a quick read of CONTRIBUTING.md, which guides you through the process of creating a new page here. Then, just simply send a pull request to this repository - and if you need any assistance, just ask :-)

waldyrious:
Fixed in #4705! 🎉

PR #4705, opened 2020-10-16
zmv: add page

hrai:

Ref: #4685

### DIFF
--- a/pages/common/zmv.md
+++ b/pages/common/zmv.md
@@ -1,6 +1,7 @@
 # zmv
 
 > Move or rename files matching a specified extended glob pattern.
+> See also `zcp` and `zln`.
 > More information: <http://zsh.sourceforge.net/Doc/Release/User-Contributions.html>.
 
 - Move files using a regex-like pattern:
@@ -18,11 +19,3 @@
 - Verbosely print each action as it's being executed:
 
 `zmv -v '{{(*).log}}' '{{$1.txt}}'`
-
-- Force `cp` (copy) regardless of the name of the function. Same as `zcp`:
-
-`zmv -C '{{(*).log}}' '{{$1.txt}}'`
-
-- Force `ln` (link) regardless of the name of the function. Same as `zln`:
-
-`zmv -L '{{(*).log}}' '{{$1.txt}}'`
--- a/pages/common/zmv.md
+++ b/pages/common/zmv.md
@@ -1,6 +1,6 @@
 # zmv
 
-> Move or rename files matching a specified regex pattern.
+> Move or rename files matching a specified extended glob pattern.
 > More information: <http://zsh.sourceforge.net/Doc/Release/User-Contributions.html>.
 
 - Below renames `foo.log` to `foo.txt`, and so on:
--- a/pages/common/zmv.md
+++ b/pages/common/zmv.md
@@ -7,7 +7,7 @@
 
 `zmv '{{(*).log}}' '{{$1.txt}}'`
 
-- Preview the result, without making any actual changes:
+- Preview the result of a move, without making any actual changes:
 
 `zmv -n '{{(*).log}}' '{{$1.txt}}'`
 
--- a/pages/common/zmv.md
+++ b/pages/common/zmv.md
@@ -3,7 +3,7 @@
 > Move or rename files matching a specified extended glob pattern.
 > More information: <http://zsh.sourceforge.net/Doc/Release/User-Contributions.html>.
 
-- Below renames `foo.log` to `foo.txt`, and so on:
+- Move files using a regex-like pattern:
 
 `zmv '{{(*).log}}' '{{$1.txt}}'`
 
--- a/pages/common/zmv.md
+++ b/pages/common/zmv.md
@@ -15,7 +15,7 @@
 
 `zmv -i '{{(*).log}}' '{{$1.txt}}'`
 
-- Verbosely print each command as it’s being executed:
+- Verbosely print each action as it's being executed:
 
 `zmv -v '{{(*).log}}' '{{$1.txt}}'`
 
--- a/pages/common/zmv.md
+++ b/pages/common/zmv.md
@@ -3,7 +3,7 @@
 > Move or rename files matching the regex pattern to corresponding files having names of the form given by destination.
 > More information: <http://zsh.sourceforge.net/Doc/Release/User-Contributions.html>.
 
-- Below renames ‘foo.lis’ to ‘foo.txt’, and so on:
+- Below renames `foo.lis` to `foo.txt`, and so on:
 
 `zmv '{{(*).log}}' '{{$1.txt}}'`
 
--- a/pages/common/zmv.md
+++ b/pages/common/zmv.md
@@ -9,7 +9,7 @@
 
 - No execution: print what would happen, but don’t do it:
 
-`zmv -n '(*).lis' '$1.txt'`
+`zmv -n '{{(*).log}}' '{{$1.txt}}'`
 
 - Interactive: show each line to be executed and ask the user whether to execute it:
 
--- a/pages/common/zmv.md
+++ b/pages/common/zmv.md
@@ -1,4 +1,4 @@
-# zsh
+# zmv
 
 > Move (usually, rename) files matching the regex pattern to corresponding files having names of the form given by destination.
 > More information: <http://zsh.sourceforge.net/Doc/Release/User-Contributions.html>.
--- a/pages/common/zmv.md
+++ b/pages/common/zmv.md
@@ -7,7 +7,7 @@
 
 `zmv '{{(*).log}}' '{{$1.txt}}'`
 
-- Preview the command result without any changes to the files:
+- Preview the result, without making any actual changes:
 
 `zmv -n '{{(*).log}}' '{{$1.txt}}'`
 
--- a/pages/common/zmv.md
+++ b/pages/common/zmv.md
@@ -13,4 +13,4 @@
 
 - Interactive: show each line to be executed and ask the user whether to execute it:
 
-`zmv -i '(*).lis' '$1.txt'`
+`zmv -i '{{(*).log}}' '{{$1.txt}}'`
--- a/pages/common/zmv.md
+++ b/pages/common/zmv.md
@@ -1,6 +1,6 @@
 # zmv
 
-> Move or rename files matching the regex pattern to corresponding files having names of the form given by destination.
+> Move or rename files matching a specified regex pattern.
 > More information: <http://zsh.sourceforge.net/Doc/Release/User-Contributions.html>.
 
 - Below renames `foo.log` to `foo.txt`, and so on:
--- a/pages/common/zmv.md
+++ b/pages/common/zmv.md
@@ -3,7 +3,7 @@
 > Move or rename files matching the regex pattern to corresponding files having names of the form given by destination.
 > More information: <http://zsh.sourceforge.net/Doc/Release/User-Contributions.html>.
 
-- Below renames `foo.lis` to `foo.txt`, and so on:
+- Below renames `foo.log` to `foo.txt`, and so on:
 
 `zmv '{{(*).log}}' '{{$1.txt}}'`
 
--- a/pages/common/zmv.md
+++ b/pages/common/zmv.md
@@ -7,10 +7,10 @@
 
 `zmv '{{(*).log}}' '{{$1.txt}}'`
 
-- No execution: print what would happen, but don’t do it:
+- Preview the command result without any changes to the files:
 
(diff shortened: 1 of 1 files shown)


pages/common/zmv.md
```diff
@@ -0,0 +1,16 @@
+# zsh
```
navarroaxel, line 1:
```suggestion
# zmv
```

pages/common/zmv.md
```diff
@@ -0,0 +1,16 @@
...
+
+> Move (usually, rename) files matching the regex pattern to corresponding files having names of the form given by destination.
```
navarroaxel, line 3:
```suggestion
> Move or rename files matching the regex pattern to corresponding files having names of the form given by destination.
```

pages/common/zmv.md
```diff
@@ -0,0 +1,16 @@
...
+> More information: <http://zsh.sourceforge.net/Doc/Release/User-Contributions.html>.
+
+- Below renames ‘foo.lis’ to ‘foo.txt’, and so on:
+
+`zmv '(*).lis' '$1.txt'`
+
+- No execution: print what would happen, but don’t do it:
```
navarroaxel, line 10:
```suggestion
- Display the command result without do changes in the filesystem:
```

pages/common/zmv.md
```diff
@@ -0,0 +1,16 @@
...
+
+`zmv -n '(*).lis' '$1.txt'`
+
+- Interactive: show each line to be executed and ask the user whether to execute it:
```
navarroaxel, line 14:
```suggestion
- Interactively move files, with a prompt before every change:
```

pages/common/zmv.md
```diff
@@ -0,0 +1,16 @@
+# zsh
+
+> Move (usually, rename) files matching the regex pattern to corresponding files having names of the form given by destination.
+> More information: <http://zsh.sourceforge.net/Doc/Release/User-Contributions.html>.
+
+- Below renames ‘foo.lis’ to ‘foo.txt’, and so on:
```
navarroaxel, line 6: Maybe this is better:
```suggestion
- Change the extension for all `*.log` files to `.txt`, keeping the same name:
```

pages/common/zmv.md
```diff
@@ -0,0 +1,16 @@
...
+
+`zmv '(*).lis' '$1.txt'`
```
navarroaxel, line 8: The arguments should be between `{}`:
```suggestion
`zmv '{{(*).log}}' '{{$1.txt}}'`
```

pages/common/zmv.md
```diff
@@ -0,0 +1,16 @@
...
+
+- No execution: print what would happen, but don’t do it:
+
+`zmv -n '(*).lis' '$1.txt'`
+
+- Interactive: show each line to be executed and ask the user whether to execute it:
+
+`zmv -i '(*).lis' '$1.txt'`
```
navarroaxel, line 16:
```suggestion
`zmv -i '{{(*).log}}' '{{$1.txt}}'`
```

pages/common/zmv.md
```diff
@@ -0,0 +1,16 @@
...
+> Move (usually, rename) files matching the regex pattern to corresponding files having names of the form given by destination.
+> More information: <http://zsh.sourceforge.net/Doc/Release/User-Contributions.html>.
+
+- Below renames ‘foo.lis’ to ‘foo.txt’, and so on:
+
+`zmv '(*).lis' '$1.txt'`
+
+- No execution: print what would happen, but don’t do it:
+
+`zmv -n '(*).lis' '$1.txt'`
```
navarroaxel, line 12:
```suggestion
`zmv -n '{{(*).log}}' '{{$1.txt}}'`
```

navarroaxel (changes requested):
Hey @hrai welcome and thanks for your submission. I left some comments.

I think we should add extra examples for `-C`, `-L`, `-M` and `-v` because these are useful options.

pages/common/zmv.md
hrai: It's not only an extension rename. We can change other parts of file name so let's keep it generic.

hrai:
> Hey @hrai welcome and thanks for your submission. I left some comments.
> 
> I think we should add extra examples for `-C`, `-L`, `-M` and `-v` because these are useful options.

@navarroaxel I've updated the md.

pages/common/zmv.md
navarroaxel: But your example can be dedicated for just change extensions.

hrai: I suppose the example is simple but a bit restrictive.

It lets you define `capturing group` like in regex and use it in the destination file name. That's where I see it shine.

Eg - `zmv '(*) - (*).mp3' '$2 - $1.mp3'` renames `HeartBreaker - Mariah Carey.mp3` to `Mariah Carey - HeartBreaker.mp3`

pages/common/zmv.md
```diff
@@ -0,0 +1,28 @@
+# zmv
+
+> Move or rename files matching the regex pattern to corresponding files having names of the form given by destination.
```
sbrl, line 3: Perhaps this will read better?
```suggestion
> Move or rename files matching a specified regex pattern.
```

pages/common/zmv.md
```diff
@@ -0,0 +1,28 @@
...
+> More information: <http://zsh.sourceforge.net/Doc/Release/User-Contributions.html>.
+
+- Below renames `foo.log` to `foo.txt`, and so on:
```
sbrl, line 6: Is there any way  that we can describe precisely what this is doing, rather than giving an example of how the example functions here?

pages/common/zmv.md
```diff
@@ -0,0 +1,28 @@
...
+
+`zmv '{{(*).log}}' '{{$1.txt}}'`
+
+- Preview the command result without any changes to the files:
```
sbrl, line 10: Perhaps this will read better?
```suggestion
- Preview the result, without making any actual changes:
```

pages/common/zmv.md
```diff
@@ -0,0 +1,28 @@
...
+
+`zmv -n '{{(*).log}}' '{{$1.txt}}'`
+
+- Interactively move files, with a prompt before every change:
+
+`zmv -i '{{(*).log}}' '{{$1.txt}}'`
+
+- Verbosely print each command as it’s being executed:
```
sbrl, line 18: We generally prefer simple quotes rather than smart quotes.
```suggestion
- Verbosely print each action as it's being executed:
```

pages/common/zmv.md
```diff
@@ -0,0 +1,28 @@
...
+
+`zmv -v '{{(*).log}}' '{{$1.txt}}'`
+
+- Force `cp` (copy) regardless of the name of the function. Same as `zcp`:
```
sbrl, line 22: I'm confused. What does this actually do? Same for the example below.

sbrl (changes requested):

pages/common/zmv.md
hrai: I think examples are better than full-on descriptions but if there are any specific changes you'd like, I'm happy to incorporate that. I'm also wary of not adding too much text as `tldr` :P

pages/common/zmv.md
hrai: `zmv` moves/renames the files deleting the old files.

If you want to `copy` the files, you can use this flag (keeps both old and new).
If you want to `create` links, you can use the flag below (keeps both old and new).

pages/common/zmv.md
sbrl: Hrm. What about this?
```suggestion
- Change extensions in filenames using a pattern:
```

hrai: It may give the users an impression that it can only be used for file ext change. 

How about the below example?

`zmv '{{(*) - (*).txt}}' '{{$2 - $1.txt}}'`

This renames `test - item.txt` to `item - test.txt`

This is how I use it mostly and that's where I see its strength.

navarroaxel (approved):

pages/common/zmv.md
hrai: A better example is here https://github.com/tldr-pages/tldr/pull/4705#discussion_r507267446

hrai:
@sbrl I've addressed your comments.

sbrl (commented):
Thanks, @hrai!

/cc @owenvoke or @mebeim perhaps for another opinion on the first example. I'm unsure about it.

pages/common/zmv.md
waldyrious: The first example should be the simplest (see [guideline #4](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines)). We can add a second example with more advanced regex, to showcase the power of the tool, but the first one should be short and easy to understand at a glance. I think we can keep the current example and just change the description, maybe to something like this:

```
- Move files using a regex-like pattern:

`zmv '{{(*).log}}' '{{$1.txt}}'`
```

WDYT?

pages/common/zmv.md
```diff
@@ -0,0 +1,28 @@
+# zmv
+
+> Move or rename files matching a specified regex pattern.
+> More information: <http://zsh.sourceforge.net/Doc/Release/User-Contributions.html>.
+
+- Below renames `foo.log` to `foo.txt`, and so on:
+
+`zmv '{{(*).log}}' '{{$1.txt}}'`
+
+- Preview the result, without making any actual changes:
```
waldyrious, line 10: Just to be a little more explicit:

```suggestion
- Preview the result of a move, without making any actual changes:
```

pages/common/zmv.md
waldyrious: This seems like an overloading of the tool's purpose, and muddies the waters regarding the dedicated tools that exist to perform these actions. I would suggest omitting these two examples (`-C` and `-L`) and adding a "See also" note in the main description pointing to `zcp` and `zln` instead.

pages/common/zmv.md
```diff
@@ -0,0 +1,28 @@
+# zmv
+
+> Move or rename files matching a specified regex pattern.
```
waldyrious, line 3: Hmm, it doesn't seem like "regex" is the right terminology. From [this page](https://filipe.kiss.ink/zmv-zsh-rename/):

> First of all, let me get this out of the way: `zmv` does not operate using regular expressions, but it uses [glob](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Glob-Operators) operators to match the files. It's the same pattern you are used to use in the CLI environment but, unfortunately, we can't do things like `[0-9]{4}` to match any sequence of 4 numbers. We actually need to repeat the pattern four times — i.e. `[0-9][0-9][0-9][0-9]` - to get the same result.

And from the [official documentation](http://zsh.sourceforge.net/Doc/Release/User-Contributions.html#Other-Functions), linked below:

> The pattern is always treated as an EXTENDED_GLOB pattern.

So maybe let's change "regex pattern" to "extended glob pattern"?

navarroaxel:
```suggestion
> Move or rename files matching a specified extended glob pattern.
```

pages/common/zmv.md
```diff
@@ -0,0 +1,28 @@
...
+> More information: <http://zsh.sourceforge.net/Doc/Release/User-Contributions.html>.
+
+- Below renames `foo.log` to `foo.txt`, and so on:
+
+`zmv '{{(*).log}}' '{{$1.txt}}'`
```
navarroaxel, line 8:
```suggestion
- Move files using a regex-like pattern:

`zmv '{{(*).log}}' '{{$1.txt}}'`
```

pages/common/zmv.md
```diff
@@ -0,0 +1,28 @@
+# zmv
+
+> Move or rename files matching a specified extended glob pattern.
```
navarroaxel, line 3:
```suggestion
> Move or rename files matching a specified extended glob pattern.
> See also `zcp` and `zln`.
```

pages/common/zmv.md
```diff
@@ -0,0 +1,28 @@
...
+
+`zmv -v '{{(*).log}}' '{{$1.txt}}'`
+
+- Force `cp` (copy) regardless of the name of the function. Same as `zcp`:
+
+`zmv -C '{{(*).log}}' '{{$1.txt}}'`
+
+- Force `ln` (link) regardless of the name of the function. Same as `zln`:
+
+`zmv -L '{{(*).log}}' '{{$1.txt}}'`
```
navarroaxel, line 28:
```suggestion
```

navarroaxel (changes requested):

sbrl (approved):
Approving as @navarroaxel has applied the changes

navarroaxel (approved):

PR #4708, opened 2020-10-16
yum: correct update command

apm1467:

The original command `yum update` upgrades installed packages. `yum check-update` is the correct command which synchronizes list of packages and versions available.

### DIFF
--- a/pages/linux/yum.md
+++ b/pages/linux/yum.md
@@ -18,7 +18,7 @@
 
 `yum remove {{package}}`
 
-- Get a list of available updates for installed packages:
+- Display available updates for installed packages:
 
 `yum check-update`
 
--- a/pages/linux/yum.md
+++ b/pages/linux/yum.md
@@ -1,6 +1,7 @@
 # yum
 
 > Package management utility for RHEL, Fedora, and CentOS (for older versions).
+> More information: <https://man7.org/linux/man-pages/man8/yum.8.html>.
 
 - Install a new package:
 
--- a/pages/linux/yum.md
+++ b/pages/linux/yum.md
@@ -4,7 +4,7 @@
 
 - Synchronize list of packages and versions available. This should be run first, before running subsequent yum commands:
 
-`yum update`
+`yum check-update`
 
 - Install a new package:
 
--- a/pages/linux/yum.md
+++ b/pages/linux/yum.md
@@ -2,10 +2,6 @@
 
 > Package management utility for RHEL, Fedora, and CentOS (for older versions).
 
-- Synchronize list of packages and versions available. This should be run first, before running subsequent yum commands:
-
-`yum check-update`
-
 - Install a new package:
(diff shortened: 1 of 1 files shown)


navarroaxel:
@apm1467 Please sign the `license/cla`. Also, I don't agree with the example description: `This should be run first, before running subsequent yum commands`.

I can run `yum upgade` without run `yum update` or `yum check-update`.

apm1467:
@navarroaxel You are right. It is normally not necessary to run `yum check-update` at all. People get the habit from `brew` or `apt-get` that they should always run `update` before doing other things. This is not how `yum` works.

Do you think it is appropriate to remove this whole command?

navarroaxel:
@apm1467 Indeed not, the `yum` page has 6 examples and the limit is 8. So, you can fix the example description and add the `check-update` example if you want. Also you can rearrange the examples if you think another order is better to learn this command.

And, you can add pages for the subcommands like `yum-update.md` to explore the subcommand options with more detail.

pages/linux/yum.md
```diff
@@ -22,6 +18,10 @@
 
 `yum remove {{package}}`
 
+- Get a list of available updates for installed packages:
```
navarroaxel, line 21:
```suggestion
- Display available updates for installed packages:
```

navarroaxel (changes requested):

pages/linux/yum.md
```diff
@@ -2,10 +2,6 @@
 
 > Package management utility for RHEL, Fedora, and CentOS (for older versions).
```
navarroaxel, line 3: Could you add a link to a web man page? I found this:
```suggestion
> Package management utility for RHEL, Fedora, and CentOS (for older versions).
> More information: <https://man7.org/linux/man-pages/man8/yum.8.html>.
```
Do you know a better one?

navarroaxel (changes requested):
1 comment. Let me know if you're done with the editions, or if you're planning more.

pages/linux/yum.md
apm1467: I looked for similar pages like [`apt`](https://github.com/tldr-pages/tldr/blob/cd3a80bffbd3f6c26c39e9513b5e072f58c9ddfa/pages/linux/apt.md) and [`apk`](https://github.com/tldr-pages/tldr/blob/5ef55e6fc44ddb75672469a9403e49702bc27753/pages/linux/apk.md), and they do not have a web man page link. I think the one you found is decent.

apm1467:
@navarroaxel Thank you very much for suggestions and help. I'm done with editing.

navarroaxel (approved):
LGTM, thanks @apm1467.

sbrl (approved):
Thanks, @apm1467! That's interesting that `yum` doesn't need you to update the repo metadata first.

PR #4711, opened 2020-10-16
jq: add an example of skipping non-JSON lines

vforvova:

Hi 👋🏼

I add a new example that I find useful in my day-to-day work as it helps me to deal with reading JSON log streams with occasionally appearing non-JSON lines. For example, if you read a Node.js logger which produces JSON entries and from time to time you receive some non-JSON log entries from the Node.js itself.

Although the page has 9 examples already I decided to try it out anyway. I understand that the aim is to make the output as easy and beginner-friendly as possible, so I won't get hard feelings if the PR ends up closed. 🤗

### DIFF
--- a/pages/common/jq.md
+++ b/pages/common/jq.md
@@ -38,3 +38,7 @@
 - Output the value of a given key to a string (and disable JSON output):
 
 `cat {{file.json}} | jq --raw-output '"some text: \(.{{key_name}})"'`
+
+- Output the value of a given key from a JSON parsable lines, and skip if parsing is not successful:
+
+`cat {{file.json}} | jq --raw-input 'fromjson? | .{{key_name}}'`
--- a/pages/common/jq.md
+++ b/pages/common/jq.md
@@ -15,13 +15,9 @@
 
 `jq --slurp . {{file.json}}`
 
-- Output the first element in a JSON file:
+- Output a value under given path. The path can consist of array indices (wrapped with `[]`) and object keys:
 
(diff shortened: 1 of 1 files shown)


navarroaxel (changes requested):
Hi @vforvova, welcome and thanks for your submission. I didn't know that `jq` can skip invalid JSON lines :scream_cat:.

We should reduce the page to 8 examples :thinking: I think we can start merging the example 4 (`.[0]`) and 5 (`[0].key_name`). Could you try?

vforvova:
@navarroaxel sure. Will try to make something meaningful out of it 🤔

vforvova:
@navarroaxel please take a look, I am not sure if it is good.

sbrl (commented):
While the example you're  adding is great @vforvova, it seems like it's an advanced use-case that's  designed to solve a specific problem. It's always challenging with pages like this to keep the example number down ( `awk` and `sed` also come to mind), but in tldr-pages it's also worth keeping the beginner in mind. Keeping the examples simple is helpful or the beginner to understand the basic syntax.

However, it looks like a great command to add to a more general jq cheat sheet.

navarroaxel:
Sorry @vforvova but after talk about this your example looks advanced for this pages and not common to use. Please, I invite you to continue contributing in this project.

vforvova:
@navarroaxel @sbrl hey guys, thanks for the feedback! It is fair enough and I am fine 😊 I guess it is ok to close the PR then, should it be me?

PR #4721, opened 2020-10-17
cat: add Russian translation

NikonP:


pages.ru/common/cat.md
```diff
@@ -0,0 +1,23 @@
+# cat
+
+> Выводит и объединяет файлы.
```
navarroaxel, line 3: The gTranslator use `present continuous` when translate your page. I think the page should use `imperative mood`, but I don't speak Russian.

navarroaxel (approved):
LGTM, just 1 question.

pages.ru/common/cat.md
NikonP: I guess there is no mistake, `выводит` can mean both action in the process (printing) and a verb (just print). I could use something like `напечатать и объединить файлы` but this would rather mean the printing process, and not that the utility outputs the contents of the file

navarroaxel: np, Leave as-is.

sbrl (approved):

PR #4726, opened 2020-10-17
git-log: add German translation

CraxyTM:

Hi,

I've added a German translation for the git-log page.

### DIFF
--- a/pages.de/common/git-log.md
+++ b/pages.de/common/git-log.md
@@ -3,34 +3,34 @@
 > Zeigt die Commit-Historie an.
 > Mehr Informationen: <https://git-scm.com/docs/git-log>.
 
-- Zeigt die Sequenz der Commits des Git-Repository im aktuellen Ordner, beginnend mit dem aktuellen, an. In umgekehrter chronologischer Reihenfolge:
+- Zeige die Sequenz der Commits des Git-Repository im aktuellen Ordner, beginnend mit dem aktuellen, an. In umgekehrter chronologischer Reihenfolge:
 
 `git log`
 
-- Zeigt die Historie einer bestimmten Datei oder eines Verzeichnisses, inklusive Unterschiede, an:
+- Zeige die Historie einer bestimmten Datei oder eines Verzeichnisses, inklusive Unterschiede, an:
 
 `git log -p {{pfad/zu/datei_oder_ordner}}`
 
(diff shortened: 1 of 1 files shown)


pages.de/common/git-log.md
```diff
@@ -0,0 +1,36 @@
+# git log
+
+> Zeigt die Commit-Historie an.
```
navarroaxel, line 3: The verbs should be in `present indicative`, not `present continuous`. I don't speak German but I should check.

navarroaxel (approved):
Hi @CraxyTM, welcome and thanks for your submission. I left 1 question.

sbrl (approved):

I don't know what the `present indicative` is, so I can't comment on that.

pages.de/common/git-log.md
navarroaxel: @LukBukkit, based on your PR #4793. Should we replace `Zeigt` by `Zeige` here?

LukBukkit: I'm sorry for my late answer. 
Yeah, I would prefer the usage of `Zeige` as it is the [imperative of `Zeigen`](https://www.duden.de/rechtschreibung/zeigen#grammatik) and so it does sound more targeted to the user. A lot of other German translations are also using the imperative:
* [`7z`](https://github.com/tldr-pages/tldr/blob/master/pages.de/common/7z.md)
* [`bash`](https://github.com/tldr-pages/tldr/blob/master/pages.de/common/bash.md)
* [`bat`](https://github.com/tldr-pages/tldr/blob/master/pages.de/common/bat.md)

But not every translated pages does, for example the page [`git`](https://github.com/tldr-pages/tldr/blob/master/pages.de/common/git.md) also uses `Zeigt`.

navarroaxel: @CraxyTM Could you please apply this change?

sbrl: Hrm. If other translations are using the imperative, then it's probably fine to use the imperative here too? I mean I don't speak German and I have no idea what the imperative, present simple, or indicative are, so I'm not sure can comment here.

CraxyTM: I replaced "Zeigt" with "Zeige" as requested by @navarroaxel

PR #4749, opened 2020-10-18
smbmap: add page

dhondta:

### DIFF
--- a/pages/linux/smbmap.md
+++ b/pages/linux/smbmap.md
@@ -3,27 +3,27 @@
 > SMB enumeration tool.
 > More information: <https://github.com/ShawnDEvans/smbmap>.
 
-- Enumerate SMB shares and permissions on a host:
+- Display SMB shares and permissions on a host:
 
 `smbmap -u {{username}} -p {{password}} -H {{ip}}`
 
-- Enumerate SMB shares and permissions on a host, specifying the domain and passing the password NTLM hash:
+- Display SMB shares and permissions on a host, specifying the domain and passing the password NTLM hash:
 
 `smbmap -u {{username}} -p {{ntlm_hash}} -d {{domain}} -H {{ip}}`
 
-- Enumerate SMB shares and list a single level of directories and files:
+- Display SMB shares and list a single level of directories and files:
 
 `smbmap -u {{username}} -p {{password}} -H {{ip}} -r`
 
-- Enumerate SMB shares and recursively list a defined number of levels of directories and files:
+- Display SMB shares and recursively list a defined number of levels of directories and files:
 
 `smbmap -u {{username}} -p {{password}} -H {{ip}} -R --depth {{3}}`
 
-- Enumerate SMB shares and recursively list directories and files, downloading the files matching a regular expression:
+- Display SMB shares and recursively list directories and files, downloading the files matching a regular expression:
 
 `smbmap -u {{username}} -p {{password}} -H {{ip}} -R -A {{pattern}}`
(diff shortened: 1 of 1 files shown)


pages/linux/smbmap.md
```diff
@@ -0,0 +1,36 @@
+# smbmap
+
+> SMB enumeration tool.
+> More information: <https://github.com/ShawnDEvans/smbmap>.
+
+- Enumerate SMB shares and permissions on a host:
```
navarroaxel, line 6: I prefer the verb display, if this command print information in the terminal:
```suggestion
- Display SMB shares and permissions on a host:
```
The same for the others examples.

navarroaxel (changes requested):

pages/linux/smbmap.md
dhondta: OK, adapted.

navarroaxel (approved):

pages/linux/smbmap.md
```diff
@@ -0,0 +1,36 @@
+# smbmap
+
+> SMB enumeration tool.
+> More information: <https://github.com/ShawnDEvans/smbmap>.
+
+- Display SMB shares and permissions on a host:
+
+`smbmap -u {{username}} -p {{password}} -H {{ip}}`
```
sbrl, line 8: Specifying a password like this will mean that the password ends up in the user's shell history. Is there another way we can pass the password to avoid this? Same goes for all the other examples.

pages/linux/smbmap.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Display SMB shares and recursively list directories and files, downloading the files matching a regular expression:
+
+`smbmap -u {{username}} -p {{password}} -H {{ip}} -R -A {{pattern}}`
+
+- Display SMB shares and recursively list directories and files, searching for file content matching a regular expression:
+
+`smbmap -u {{username}} -p {{password}} -H {{ip}} -R -F {{pattern}}`
+
+- Execute an OS command on a remote system:
```
sbrl, line 30: Perhaps this will read better?
```suggestion
- Execute a shell command on a remote system:
```

pages/linux/smbmap.md
```diff
@@ -0,0 +1,36 @@
...
+
+`smbmap -u {{username}} -p {{password}} -H {{ip}} -x {{command}}`
+
+- Upload a file on a remote system:
```
sbrl, line 34: Perhaps this will read better?
```suggestion
- Upload a file to a remote system:
```

sbrl (changes requested):

pages/linux/smbmap.md
dhondta: This tool is typically used during penetration tests. And indeed, you should carefully manage your history while using cleartext passwords in command lines. That's the way it is used in general.
But this is a very good remark ; looking at the source of the tool reveals that you can either use `-p {{password}}` (to specify the password in the command line) or `--prompt` (to prompt for a password using the `getpass.getpass()` function. I will adapt at least one example with the prompt option.

dhondta: I adapted one example. Good for you ?

sbrl: Heya! Ah, I see. That looks much better! Can this be reflected across the other examples please?

dhondta: Done.

sbrl (approved):
Thanks, @dhondta!

PR #4754, opened 2020-10-18
README: add fast-tldr Haskell client

gutjuri:
I modified an existing Haskell client heavily in order to have the fastest tldr-client. It does pretty good, it's a bit faster than the fastest other client I could find [Tealdeer](https://github.com/dbrgn/tealdeer). 

It follows the minimum requirements. It also follows large parts of the full requirements for tldr-clients, though I still have to assess to what extend.

I'd also appreciate if you'd add it to your [Wiki](https://github.com/tldr-pages/tldr/wiki/tldr-pages-clients) :)


sbrl (approved):
Thanks, @gutjuri!

You're welcome to edit the wiki too. Do you not have permission to do so? I don't remember how the wiki works, so just want to check

gutjuri:
> Do you not have permission to do so?

I don't see an option for that in the web interface... might be restricted to maintainers or something like that ..

navarroaxel (approved):

Issue #4746, opened 2020-10-18
Token inconsistencies: enforce snake_case in token


glenacota:
The [token syntax guideline](https://github.com/tldr-pages/tldr/blob/master/contributing-guides/style-guide.md#token-syntax) for multi-word tokens is to use the_snake_case_form. However, in some pages, words in tokens are separated by whitespace (e.g., buku, gist, man, [git-send-email](https://github.com/tldr-pages/tldr/pull/4737#discussion_r507013716), ...) and in others dashes (e.g., dexter, git-ls-remote, ...).

I can work on a bulk PR to enforce the snake_case syntax guideline, maybe creating one commit for each wrongful separator.

(this Issue might be linked to [this old one)](https://github.com/tldr-pages/tldr/issues/2509#issuecomment-435521621)

glenacota:
BTW, I'm not sure this is the right place where to ask, but as I'm very fond of this project and lately I've got some more time to contribute to it, I'd like to ask the org owners how does it work to apply for a [regular contributor position](https://github.com/tldr-pages/tldr/blob/master/COMMUNITY-ROLES.md)  :)

seenukarthi:
@glenacota there is [gitter room](https://gitter.im/tldr-pages/tldr) I think general discussion can be done there.

sbrl:
Hey there, @glenacota! We'd love to have you onboard. According to our [community roles](https://github.com/tldr-pages/tldr/blob/master/COMMUNITY-ROLES.md) document, you need to have 5 non-trivial PRs merged. From the pull request history, I notice you have the following non-trivial PRs merged: #4364, #4362, #2464, #2458, #2452, and #2440. This qualifies you to become a contributor according to the aforementioned document, so I'll open a new issue about that and @mention you :-)

sbrl:
And to answer the original question in this issue, a PR would be appreciated! Though do be careful to check each one: there may be some instances of concrete examples where the space is essential to the page. And also yes, this does seem related to that issue

sbrl:
Opened #4763

glenacota:
issue solved with https://github.com/tldr-pages/tldr/pull/4788

PR #4788, opened 2020-10-20
Enforce snake_case in token

glenacota:
Changes in this PR attempt to solve issue #4746, using the snake case in tokens where it is not.

~- [ ] The page is in the correct platform directory (`common/`, `linux/`, etc.)~
~- [ ] The page has 8 or fewer examples.~
~- [ ] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).~
~- [ ] The page description includes a link to documentation or a homepage (if applicable).~

### DIFF
--- a/pages/common/nohup.md
+++ b/pages/common/nohup.md
@@ -4,4 +4,4 @@
 
 - Run process that can live beyond the terminal:
 
-`nohup {{command_with_options}}`
+`nohup {{command}} {{command_options}}`
--- a/pages/common/nohup.md
+++ b/pages/common/nohup.md
@@ -4,4 +4,4 @@
 
 - Run process that can live beyond the terminal:
 
-`nohup {{command options}}`
+`nohup {{command_with_options}}`
--- a/pages/windows/azcopy.md
+++ b/pages/windows/azcopy.md
@@ -9,19 +9,19 @@
 
 - Upload a local file:
 
-`azcopy copy '{{path/to/source/file}}' 'https://{{storage-account-name}}.blob.core.windows.net/{{container-name}}/{{blob-name}}'`
+`azcopy copy '{{path/to/source/file}}' 'https://{{storage_account_name}}.blob.core.windows.net/{{container_name}}/{{blob_name}}'`
 
 - Upload files with `.txt` and `.jpg` extensions:
 
-`azcopy copy '{{path/to/source}}' 'https://{{storage-account-name}}.blob.core.windows.net/{{container-name}}' --include-pattern '{{*.txt;*.jpg}}'`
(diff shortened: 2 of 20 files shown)


pages/common/nohup.md
```diff
@@ -4,4 +4,4 @@
 
 - Run process that can live beyond the terminal:
 
-`nohup {{command options}}`
+`nohup {{command_with_options}}`
```
navarroaxel, line 7: Yeah, is definitively better. What about this is option?
```suggestion
`nohup {{command}} {{command_options}}`
```

pages/windows/azcopy.md
```diff
@@ -9,19 +9,19 @@
 
 - Upload a local file:
 
-`azcopy copy '{{path/to/source/file}}' 'https://{{storage-account-name}}.blob.core.windows.net/{{container-name}}/{{blob-name}}'`
+`azcopy copy '{{path/to/source/file}}' 'https://{{storage_account_name}}.blob.core.windows.net/{{container_name}}/{{blob_name}}'`
```
navarroaxel, line 12: Maybe this was made because `_` is not valid in URLs. But we don't need that the example URL works :man_shrugging:

navarroaxel (approved):

pages/common/nohup.md
sbrl: Either would work.

pages/windows/azcopy.md
sbrl: It isn't? I didn't know that. In that case, we should revert this one.

sbrl (commented):
Looks ok to   me after @navarroaxel's comments are addressed

pages/common/nohup.md
glenacota: I found in the codebase three `{{command}} {{command_arguments}}` , one `{{command}} {{args}}`, and a couple `{{args}}`. Just to add some more doubts :)

pages/windows/azcopy.md
glenacota: Underscores in URLs are valid, e.g., https://en.wikipedia.org/wiki/Main_Page. I don't see any special reason not to use them :)

navarroaxel: I mean for the subdomain, you can't use underscore (`_`) but you can use hyphens (`-`)

glenacota: You're right, but then only the first token would require the hyphen. Of course, it'd be very ugly to use hyphens and underscores in the same example and we should stick to one :) IMHO: validity shouldn't be an issue with tokens

sbrl: Generally we prefer underscores in token syntax elsewhere, so for consistency I think underscores might be the better choice here.

navarroaxel (approved):

sbrl (approved):
Thanks, @glenacota!

PR #4809, opened 2020-10-22
dotnet-build: fix compile in release mode example

Ryzhehvost:
@navarroaxel sorry for doing those as separate files, I was already working on this one when I've seen your message.

### DIFF
--- a/pages.ru/common/dotnet-build.md
+++ b/pages.ru/common/dotnet-build.md
@@ -1,0 +1,32 @@
+# dotnet build
+
+> Собирает приложение .NET и все его зависимости.
+> Больше информации: <https://docs.microsoft.com/dotnet/core/tools/dotnet-build>.
+
+- Скомпилировать проект или решение в текущей директории:
+
+`dotnet build`
+
+- Скомпилировать проект или решение .NET в режиме debug:
+
+`dotnet build {{путь/до/проекта или решения}}`
+
+- Скомпилировать в режиме release:
+
+`dotnet build`
+
+- Скомпилировать без восстановления зависимостей:
+
+`dotnet build --no-restore`
+
+- Скомпилировать с заданным уровнем детализации выводимой информации:
+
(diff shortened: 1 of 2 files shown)


pages.ru/common/dotnet-build.md
```diff
@@ -0,0 +1,32 @@
...
+
+`dotnet build`
+
+- Скомпилировать проект или решение .NET в режиме debug:
+
+`dotnet build {{путь/до/проекта_или_решения}}`
+
+- Скомпилировать в режиме release:
+
+`dotnet build`
```
navarroaxel, line 16: As author of the English page: there is a bug :upside_down_face:. This is the correct example for the description:
```suggestion
`dotnet build --configuration Release`
```
Can you include the fix to the English page in this PR?

navarroaxel (changes requested):
Hi @Ryzhehvost, I left some comments.

pages.ru/common/dotnet-build.md
Ryzhehvost: Sure, just fixed it in both original and translation.

navarroaxel (approved):

pages/common/dotnet-build.md
```diff
@@ -13,7 +13,7 @@
 
 - Compile in release mode:
 
-`dotnet build`
+`dotnet build --configuration Release`
```
sbrl, line 16: Missing token syntax?
```suggestion
`dotnet build --configuration {{Release}}`
```

Ryzhehvost: Are you sure about it? `Release` is one of possible configurations, so it's more like `dotnet build --configuration {{Configuration}}`, but then we need to change description too. @navarroaxel what do you think?

navarroaxel: I agree with @sbrl because we had this format in other pages. We have rules, but we have exceptions too. And the argument option should be between `{{}}`.

sbrl (approved):
Thanks, @Ryzhehvost!

PR #4811, opened 2020-10-22
7z: add Japanese translation

Miyuutsu:


navarroaxel (approved):

sbrl (approved):

Feel free to open 1 big PR if you're translating lots of pages, with 1 commit per page  :-)

rinsuki:
This translate looks like machine translation, it is acceptable in (license of) this repository?

e.g. in japanese, `道/に/ファイルまたはディレクトリ` is wrong translate (it looks translated from `path/to/file_or_directory`).
correct translate is like this: `ファイルまたはフォルダ/への/パス` or simply `path/to/file_or_directory`

Miyuutsu:
I can definitely say it was not machine translation, but I do not have experience with translating this nature of thing. I was not certain whether it should have been translated literally as I did so or if I should have done it naturally. If I have time later I will cross-reference with the 7-Zip website.

sbrl:
Unfortunately I do not have the pleasure of knowing multiple human languages, but I would suggest that it should be translated to whatever makes the most sense and reads best in the target language, while being sensitive of any techincal meaning :-)

Pull requests are of course accepted to update and improve translations - it's an iterative process of continuous improvement :slightly_smiling_face:

Miyuutsu:
Japanese is only a secondary language for me, and the reason I chose 7z to translate to begin with was because 7-Zip has a Japanese website that I could reference if I had any direct issues with syntax and such. Sadly, as such, I'm unsure which way would be better. I think I would leave it as is personally, but it looks like maybe rinsuki uses Japanese as a primary language, so if they think it would be best to be more natural I would go with it.

PR #4818, opened 2020-10-22
docker-exec: add new example

alimalek71:

### DIFF
--- a/pages/common/docker-exec.md
+++ b/pages/common/docker-exec.md
@@ -5,24 +5,24 @@
 
 - Enter an interactive shell session on an already-running container:
 
-`docker exec -it {{container_name}} {{/bin/bash}}`
+`docker exec --interactive --tty {{container_name}} {{/bin/bash}}`
 
 - Run a command in the background (detached) on a running container:
 
-`docker exec -d {{container_name}} {{command}}`
+`docker exec --detach {{container_name}} {{command}}`
 
 - Select the working directory for a given command to execute into:
 
-`docker exec -it -w {{path/to/directory}} {{container_name}} {{command}}`
+`docker exec --interactive -tty --workdir {{path/to/directory}} {{container_name}} {{command}}`
 
 - Run a command in background on existing container but keep `stdin` open:
 
-`docker exec -i -d {{container_name}} {{command}}`
+`docker exec --interactive --detach {{container_name}} {{command}}`
 
 - Set an environment variable in a running bash session:
 
-`docker exec -it -e {{variable_name}}={{value}} {{container_name}} {{/bin/bash}}`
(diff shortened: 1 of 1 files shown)


pages/common/docker-exec.md
```diff
@@ -22,3 +22,7 @@
 - Set an environment variable in a running bash session:
 
 `docker exec -it -e {{variable_name}}={{value}} {{container_name}} {{/bin/bash}}`
+
+- Execute command as a specific user:
```
navarroaxel, line 26:
```suggestion
- Run a command as a specific user:
```

pages/common/docker-exec.md
```diff
@@ -22,3 +22,7 @@
...
+
+`docker exec -u {{user}} {{container_name}} {{command}}`
```
navarroaxel, line 28: Should the user be already created in the container, or is created by docker automatically?
```suggestion
`docker exec --user {{user}} {{container_name}} {{command}}`
```

navarroaxel (changes requested):
Thanks @alimalek71 I left some comments.

pages/common/docker-exec.md
alimalek71: I choose `execute` just because of `exec` in the command btw I agree with the `run` verb.

pages/common/docker-exec.md
alimalek71: Nice question, The user must be determined in the Dockerfile. Just can pick one of the existing users in the container.

alimalek71: And by this change, I think it's better to change all shortened options to the original form.

pages/common/docker-exec.md
sbrl: Either would work here. I'm include to use execute, but run works too

navarroaxel: @sbrl Should we add the condition that the user should be already added to the container? Otherwsie we get this error message: `unable to find user XXX: no matching entries in passwd file`

alimalek71: @navarroaxel @sbrl Could I change the example to this:
- Run a command as root user:
`docker exec --user {{root}} {{container_name}} {{command}}`

alimalek71: I take back my words :)
As one of `tldr` user of I prefer the simple general example

alimalek71: Also, I applied the requested change to code but now I do not resolve the conversation to reach finalize the sentence.

alimalek71: what is the decision?

navarroaxel (approved):

sbrl (approved):
Thanks, @alimalek71!

PR #4856, opened 2020-10-25
udisktctl: add page

eimanip:

### DIFF
--- a/pages/linux/udisksctl.md
+++ b/pages/linux/udisksctl.md
@@ -15,11 +15,11 @@
 
 `udisksctl info --block-device {{/dev/sdXN}}`
 
-- Mount a device and prints the mount point:
+- Mount a device partition and prints the mount point:
 
 `udisksctl mount --block-device {{/dev/sdXN}}`
 
-- Unmount a device:
+- Unmount a device partition:
 
 `udisksctl unmount --block-device {{/dev/sdXN}}`
 
--- a/pages/linux/udisksctl.md
+++ b/pages/linux/udisksctl.md
@@ -11,6 +11,10 @@
 
 `udisksctl info --block-device {{/dev/sdX}}`
 
+- Show detailed information about a device partition:
+
+`udisksctl info --block-device {{/dev/sdXN}}`
+
 - Mount a device and prints the mount point:
 
 `udisksctl mount --block-device {{/dev/sdXN}}`
--- a/pages/linux/udisksctl.md
+++ b/pages/linux/udisksctl.md
@@ -1,4 +1,4 @@
-# udisktctl
+# udisksctl
 
 > A command-line program used to interact with the udisksd daemon process.
 > More information: <http://storaged.org/doc/udisks2-api/latest/udisksctl.1.html>.
--- a/pages/linux/udisktctl.md
+++ b/pages/linux/udisktctl.md
@@ -13,11 +13,11 @@
 
 - Mount a device and prints the mount point:
 
-`udisksctl mount --block-device {{/dev/device_name}}`
+`udisksctl mount --block-device {{/dev/sdXN}}`
 
 - Unmount a device:
 
-`udisksctl unmount --block-device {{/dev/device_name}}`
+`udisksctl unmount --block-device {{/dev/sdXN}}`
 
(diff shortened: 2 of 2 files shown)


pages/linux/udisktctl.md
```diff
@@ -0,0 +1,24 @@
+# udisktctl
+
+> A command-line program used to interact with the udisksd daemon process.
+> More information: <https://manpages.ubuntu.com/manpages/focal/man1/udisksctl.1.html>.
```
navarroaxel, line 4: I prefer the official site based on Arch Linux https://www.archlinux.org/packages/extra/x86_64/udisks2/
```suggestion
> More information: <http://storaged.org/doc/udisks2-api/latest/udisksctl.1.html>.
```

pages/linux/udisktctl.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Show high-level information about disk drives and block devices:
+
+`udisksctl status`
+
+- Show detailed information about a device:
+
+`udisksctl info --block-device {{/dev/device_name}}`
```
navarroaxel, line 12: If is for block device, so:
```suggestion
`udisksctl info --block-device {{/dev/sdX}}`
```

pages/linux/udisktctl.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Mount a device and prints the mount point:
+
+`udisksctl mount --block-device {{/dev/device_name}}`
```
navarroaxel, line 16:
```suggestion
`udisksctl mount --block-device {{/dev/sdX}}`
```

pages/linux/udisktctl.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Unmount a device:
+
+`udisksctl unmount --block-device {{/dev/device_name}}`
```
navarroaxel, line 20: Could be 1 partition or all the partitions?
```suggestion
`udisksctl unmount --block-device {{/dev/sdXN}}`
```

navarroaxel (changes requested):
Hey @eimanip, I left some comments.

pages/linux/udisktctl.md
eimanip: Hey, thanks for mentioning this. I checked it myself. The partition path should be passed here. The same goes for the `mount`. I fixed both of them in a separate commit.

pages/linux/udisktctl.md
eimanip: Fixed

pages/linux/udisktctl.md
navarroaxel: Is `partition` better than `device` when we aren't talking about a full device?

eimanip:
Hey @navarroaxel,
Thanks for your suggestions. I resolved all of them. Also, I realized that there was a typo in the file name. I fixed it too.

pages/linux/udisksctl.md
```diff
@@ -0,0 +1,24 @@
+# udisksctl
+
+> A command-line program used to interact with the udisksd daemon process.
+> More information: <http://storaged.org/doc/udisks2-api/latest/udisksctl.1.html>.
+
+- Show high-level information about disk drives and block devices:
+
+`udisksctl status`
+
+- Show detailed information about a device:
```
navarroaxel, line 10: This command works with specific partitions too. Maybe we should add an extra example for that.

pages/linux/udisktctl.md
eimanip: You're right, however, its man page used the term `device` for both mount/unmount options. But as I tested it now, it can only mount partitions. I'm not sure which term suits best here. Yet, are you suggesting to change it to `device`?

pages/linux/udisksctl.md
eimanip: Added another sample for `device partition`

pages/linux/udisktctl.md
eimanip: I went with `device partition` term to be more specific about the command.

navarroaxel (approved):

Proscream (approved):

PR #4161, opened 2020-07-07
tar: Add brackets around letters of keywords and remove one example

sawshep:
- [x] The page has 8 or fewer examples.
~~^ There are 9 examples on tar~~
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).

This fixes #3876, in order to better fit contributing guideline 3. This may get shot down by the bot due to the fact that some of the lines no longer start with capitals due to the letters in brackets.

PR #4872, opened 2020-10-27
tar: create bz2 archive

axmz:
bz2 seems to be a common method, especially when smaller archive size needs to be obtained.
that is why i think it should be added as well.

note: there will be 10 examples, which exceeds recommended 8 per command.
note: other languages might need to be updated.
note: example image from tldr.sh might need to be updated.

### DIFF
--- a/pages/common/tar.md
+++ b/pages/common/tar.md
@@ -12,6 +12,10 @@
 
 `tar czf {{target.tar.gz}} {{file1}} {{file2}} {{file3}}`
 
+- Create a bzip2 archive:
+
+`tar cjf {{target.tar.bz2}} {{file1}} {{file2}} {{file3}}`
+
 - Create a gzipped archive from a directory using relative paths:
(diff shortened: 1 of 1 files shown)


navarroaxel (changes requested):
Hi @axmz, welcome and thanks for your submission. But the rule is up to 8 examples for tldr pages, and `tar` has 9 at the moment :upside_down_face:. I invite you to edit this page to reduce to 8 examples. There is a PR #4161 trying to reduce to 8 examples.

Maybe you can try that, or contribute with other commands.

sbrl (changes requested):
Thanks for the edit, @axmz! Unfortunately we have a limit of 8 example per page.

This example looks like it's a bit similar to the `tar caf` example further down, since with `caf` if you suffix the archive name with `.tar.bz2`, it creates a bz2 tar archive?

PR #4877, opened 2020-10-27
scrcpy: add --serial example

reinhart1010:

The `--serial` command is important to users who have more than one Android devices connected. Since each `scrcpy` command can only be able to mirror one device, they will need to specify the serial ID or the IP address of the connected device for mirroring.

The examples are taken straight from [scrcpy's README file](https://github.com/Genymobile/scrcpy#multi-devices), but you're welcome to replace `0123456789abcdef` to something more random, mimicking the actual device ID.

### DIFF
--- a/pages/common/scrcpy.md
+++ b/pages/common/scrcpy.md
@@ -7,7 +7,7 @@
 
 `scrcpy`
 
-- Display a mirror of a specific device based on its serial ID or IP address, as shown in the "adb devices" command:
+- Display a mirror of a specific device based on its ID or IP address. You can find it under the "adb devices" command:
 
 `scrcpy --serial {{0123456789abcdef|192.168.0.1:5555}}`
 
--- a/pages/common/scrcpy.md
+++ b/pages/common/scrcpy.md
@@ -7,6 +7,10 @@
 
 `scrcpy`
 
+- Display a mirror of a specific device based on its serial ID or IP address, as shown in the "adb devices" command:
+
+`scrcpy --serial {{0123456789abcdef|192.168.0.1:5555}}`
+
 - Start display in fullscreen mode:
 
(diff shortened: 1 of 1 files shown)


navarroaxel:
I use `scrcpy` for work but how can I connect using the IP:5555 of my Android phone? maybe need an advanced config with the Android device.

reinhart1010:
There are two ways to connect via TCP/IP. One requires passing commands to Android's Terminal (which requires root access). Another one is by running `adb tcpip 5555` to the currently connected device (via USB).

The device will be able to connect by running `adb connect <IP>:5555`

See https://stackoverflow.com/questions/2604727 for details.

pages/common/scrcpy.md
```diff
@@ -7,6 +7,10 @@
 
 `scrcpy`
 
+- Display a mirror of a specific device based on its ID or IP address. You can find it under the "adb devices" command:
```
sbrl, line 10: Perhaps this will read better?
```suggestion
- Display a mirror of a specific device based on its ID or IP address (find it under the `adb devices` command):
```

Generally we try to avoid the 2nd person (e.g. `you`, `your`, etc) in tldr-pages :slightly_smiling_face:

sbrl (changes requested):

pages/common/scrcpy.md
reinhart1010: Thnks for you feedback. I was unsure to use the backticks on the command description, so I use "..." instead to avoid conflicts with tldr clients.

navarroaxel (approved):

sbrl (approved):
Thanks, @reinhart1010!

Yeah, backticks are allowed in descriptions :-)

PR #4953, opened 2020-11-13
tmux: add window management commands and small improvements

nuno-agostinho:

Window management commands are important to correctly work with tmux. I replaced two redundant commands with two window management commands.

Please note I reverted the full command names used by @ryanolsonx. This is a subjective choice, but I think the shorter command names are sufficient to understand what they do and preferable to a `tldr` (people can check more information on the `man` page anyway). I can revert this specific commit if you prefer.

### DIFF
--- a/pages/common/tmux.md
+++ b/pages/common/tmux.md
@@ -19,18 +19,18 @@
 
 `tmux attach`
 
-- Attach to a named session:
-
-`tmux attach -t {{name}}`
-
 - Detach from the current session (with prefix Ctrl-B):
 
 `Ctrl-B d`
 
+- Create a new window (with prefix Ctrl-B):
+
+`Ctrl-B c`
+
+- Switch between sessions and windows (with prefix Ctrl-B):
+
+`Ctrl-B w`
+
 - Kill a session by name:
 
 `tmux kill-session -t {{name}}`
-
-- Kill the current session (with prefix Ctrl-B):
-
-`Ctrl-B :kill-session<Enter>`
--- a/pages/common/tmux.md
+++ b/pages/common/tmux.md
@@ -19,15 +19,15 @@
 
 `tmux attach`
 
-- Detach from the current session (with prefix Ctrl-B):
+- Detach from the current session (inside a tmux session):
 
 `Ctrl-B d`
 
-- Create a new window (with prefix Ctrl-B):
+- Create a new window (inside a tmux session):
 
 `Ctrl-B c`
 
-- Switch between sessions and windows (with prefix Ctrl-B):
+- Switch between sessions and windows (inside a tmux session):
 
 `Ctrl-B w`
 
--- a/pages/common/tmux.md
+++ b/pages/common/tmux.md
@@ -9,7 +9,7 @@
 
 - Start a new named session:
 
-`tmux new-session -s {{name}}`
+`tmux new -s {{name}}`
 
 - List existing sessions:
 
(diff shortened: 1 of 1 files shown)


pages/common/tmux.md
```diff
@@ -9,28 +9,28 @@
...
+`tmux new -s {{name}}`
 
 - List existing sessions:
 
 `tmux ls`
 
 - Attach to the most recently used session:
 
-`tmux attach-session`
+`tmux attach`
```
sbrl, line 20: Maybe this example could be one for attaching to a named session instead? I don't see that mentioned here, and it would be nice to include that.

sbrl (commented):
Thanks for the edit! Just got 1 comment below for you to review :-)

pages/common/tmux.md
nuno-agostinho: My thought process was that, after you attach to the most recent session, you can change the currently selected session with the following controls. Anyway, to attach to a named session, one would do `tmux attach -t {{name}}`. I can put this one instead, as you prefer.

ryanolsonx:
> Please note I reverted the full command names used by @ryanolsonx

If you look at guidelines for the project, you're supposed to use the longer names.

nuno-agostinho:
> If you look at guidelines for the project, you're supposed to use the longer names.

Before posting the commit, I read the [guidelines](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines) and [Style guide](https://github.com/tldr-pages/tldr/blob/master/contributing-guides/style-guide.md) to make sure if there was a reason to use the complete command, but there is none, unless I am missing something.

I would argue that `tldr` should make examples of the usage of the commands as recommended by the [official tmux guide](https://github.com/tmux/tmux/wiki/Getting-Started). They use `tmux new` and `tmux attach` throughout the document.

ryanolsonx:
Check out guideline #3
> Try to incorporate the spelled-out version of single-letter options in the example's description. The goal is to allow people to understand the syntax of the commands, not just memorize it.

Either way I'm good with it as long as maintainers are good with it. I just remember that was a huge piece of feedback when I made some changes.

nuno-agostinho:
> Try to incorporate the spelled-out version of single-letter options in the example's description. The goal is to allow people to understand the syntax of the commands, not just memorize it.

I think that guideline does not apply in this case because:

1 ) There are no single-letter options being used in these commands;
2) It only recommends to spell out the command in the example description, not the command itself;
3) The shorter command name is sufficiently explanatory with the example description so it allows "people to understand the syntax of the commands, not just memorize it".

If the members of `tldr` agree that it is important to spell-out the commands that have shorter forms (regardless of their description), that should be made explicit in the guidelines.

Personally, I think we should use the commands as they appear in tmux's official tutorial. But to me it is fine either way and I completely understand why you did it like that. :)

pages/common/tmux.md
sbrl: Oh I see! I didn't notice that. I don't use tmux myself, so I apologise. Let's leave this as-is.

sbrl (approved):
Thanks, @nuno-agostinho!

sbrl:
Regarding the shortened names, the guidelines are there because we generally prefer the longer forms of arguments to aid clarity. For example, we prefer `--foo` over `-f`. In this case, I'm guessing that if you use tmux regularly, you'd be typing some of these a lot, hence we may lean on a shorter form here.

Furthermore, the shortened forms here still make logical sense - in that they aren't one-letter words. To that end, I think this edit is fine as-is.

However, thanks for bringing that up, @ryanolsonx. Great catch!

navarroaxel (approved):

PR #4956, opened 2020-11-16
python: include example using pip

nuno-agostinho:

### DIFF
--- a/pages/common/python.md
+++ b/pages/common/python.md
@@ -23,6 +23,10 @@
 
 `python -m {{module}} {{arguments}}`
 
+- Install a package:
+
+`python -m pip install {{package_name}}`
+
 - Interactively debug a Python script:
 
 `python -m pdb {{script.py}}`
--- a/pages/common/python.md
+++ b/pages/common/python.md
@@ -23,7 +23,7 @@
 
 `python -m {{module}} {{arguments}}`
 
-- Install a package:
+- Install a package using pip:
 
 `python -m pip install {{package_name}}`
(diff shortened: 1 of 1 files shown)


sbrl (commented):
Thanks for the edit! However, isn't there a `pip`/`pip3` command too? Which one is the preferred mechanism for calling `pip`? I'm hesitant to accept this, because it makes it difficult to draw a line between documenting Python and Python modules you install.

nuno-agostinho:
> However, isn't there a `pip`/`pip3` command too? Which one is the preferred mechanism for calling `pip`? I'm hesitant to accept this, because it makes it difficult to draw a line between documenting Python and Python modules you install.

I understand your worries, but it is the recommended method found in the official Python documentation: https://docs.python.org/3/installing

pages/common/python.md
```diff
@@ -23,6 +23,10 @@
 
 `python -m {{module}} {{arguments}}`
 
+- Install a package:
```
sbrl, line 26: I see. In that case, I guess it's ok.

Let's make this a little more explicit though, just to be on the safe side:
```suggestion
- Install a package using pip:
```

I'm not particularly experienced with Python, so a 2nd opinion from someone who is would be appreciated.

nuno-agostinho: That seems all right, committed! :)

navarroaxel (approved):

sbrl (approved):
Thanks, @nuno-agostinho!

PR #4957, opened 2020-11-16
R: fix command case and add examples

nuno-agostinho:


pages/common/R.md
```diff
@@ -0,0 +1,32 @@
...
+
+`R --version`
+
+- Start R in vanilla mode (i.e. blank session that doesn't save workspace at the end):
+
+`R --vanilla`
+
+- Execute a file:
+
+`R -f {{file.R}}`
```
sbrl, line 20: Perhaps this will read better?
```suggestion
`R -f {{path/to/file.R}}`
```

Also, is the file extension commonly `R` in uppercase? That's rather unusual.

pages/common/R.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Execute R expression and exit:
```
sbrl, line 22: Perhaps this will read better?
```suggestion
- Execute an R expression and then exit:
```

pages/common/R.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Start an R interactive shell (REPL):
+
+`R`
+
+- Check R version:
+
+`R --version`
+
+- Start R in vanilla mode (i.e. blank session that doesn't save workspace at the end):
```
sbrl, line 14: Perhaps this will read better?
```suggestion
- Start R in vanilla mode (i.e. a blank session that doesn't save the workspace at the end):
```

pages/common/R.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Execute a file:
+
+`R -f {{file.R}}`
+
+- Execute R expression and exit:
+
+`R -e {{expr}}`
+
+- Run R through a debugger:
```
sbrl, line 26: Perhaps this will read better?
```suggestion
- Run R with a debugger:
```

pages/common/R.md
```diff
@@ -0,0 +1,32 @@
...
+
+`R -d {{debugger}}`
+
+- Check R packages from package sources:
+
+`R CMD check {{package_source}}`
```
sbrl, line 32: Is this a file/directory path, or something else?

sbrl (changes requested):

pages/common/R.md
nuno-agostinho: R files are usually written with the uppercase extension. For instance, these are some R scripts from R source code (note that they end in uppercase R): https://github.com/wch/r-source/tree/5a156a0865362bb8381dcd69ac335f5174a4f60c/src/library/base/R

pages/common/R.md
nuno-agostinho: It is either a directory or a `tar` archive. The official `man` documentation reads:

> Usage: R CMD check [options] pkgs
>
> Check R packages from package sources, which can be directories or package 'tar' archives with extension '.tar.gz', '.tar.bz2', '.tar.xz' or '.tgz'.

nuno-agostinho: Also, to create the mentioned `tar` archive, one would need to use `R CMD build {{/path/to/package_source}}`. This would create ther `tar` archive to be used for R CMD check.

Alternatively, it is possible to run `R CMD check` directly on the directory without first creating the archive. As this command can be run directly, I am not sure if we should include the `R CMD build {{/path/to/package_source}}` in this page.

pages/common/R.md
sbrl: Thanks for clarifying!

pages/common/R.md
sbrl: Ah, I see. What about this then?
```suggestion
`R CMD check {{path/to/package_source}}`
```

nuno-agostinho: Seems good!

sbrl (approved):
Thanks, @nuno-agostinho!

navarroaxel (approved):

PR #4958, opened 2020-11-16
crontab: add file example

thomaswuensche:


pages/common/crontab.md
```diff
@@ -11,6 +11,10 @@
 
 `sudo crontab -e -u {{user}}`
 
+- Edit the crontab file by using the contents of a file:
```
navarroaxel, line 14: The doc says the content of that file is called `specifications` so I think this is better:
```suggestion
- Edit the crontab file by using specifications of a specific file:
```

navarroaxel (changes requested):
Thanks @thomaswuensche just a comment.

pages/common/crontab.md
thomaswuensche: thanks for the input @navarroaxel !

specifications sounds good to me, but i would suggest using separate instead of specific for more clarity.

```suggestion
- Edit the crontab file by using specifications of a separate file:
```

What do you think about including sth about the crontab file being overwritten?

sbrl: Hrm, I'm not convinced by either description here. The original description is somewhat confusing, because it's unclear as to whether you're interactively editing the crontab, or whether you're non-interactively replacing the crontab with another file.

@navarroaxel's suggestion here is also confusing, because I'm not sure that `specifications` is really the right word in this scenario.

Looking at the `man` page, it says this:

> The  first  form  of this command is used to install a new crontab from
>        some named file or standard  input  if  the  pseudo-filename  ``-''  is
>        given.


With this in mind, what about this?
```suggestion
- Replace the current crontab with the contents of the given file:
```

thomaswuensche: good point @sbrl. that's what i meant by overwriting, but non-interactively replacing probably says it the best.
Unless @navarroaxel feels strongly about using the term specifications, i would go ahead and commit your suggestion.

navarroaxel (approved):

sbrl (approved):
Thanks, @thomaswuensche!

PR #4966, opened 2020-11-18
lpr: add page-ranges example

aaaawwWWWwwwwWWW:

### DIFF
--- a/pages/common/lpr.md
+++ b/pages/common/lpr.md
@@ -32,10 +32,6 @@
 
 `lpr -o number-up={{2|4|6|9|16}} {{path/to/file}}`
 
-- Print only page 2:
+Print either a single page (e.g. 2) or a range of pages (e.g. 2–16):
 
-`lpr -o page-ranges={{2}} {{path/to/file}}`
-
-- Print from page 2 to 16:
-
-`lpr -o page-ranges={{2-16}} {{path/to/file}}`
+`lpr -o page-ranges={{2|2-16}} {{path/to/file}}`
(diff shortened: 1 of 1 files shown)


navarroaxel (changes requested):
Hi @aaaawwWWWwwwwWWW, welcome and thanks for your submission. But our rule is up to 8 examples per page and this has 9 examples.

Which example do you think we can remove?

aaaawwWWWwwwwWWW:
hi. it's hard to tell because i personally only use the first 2 examples in the page and the 2 examples i committed. The rest seems useful also; but the having 2 examples for printing in double sides seems unnecessary mostly taking into consideration their commands in the example are the same. So i think turning that into a single example?

aaaawwWWWwwwwWWW:
or i actually could transform my example in a single example too as in

Print either a single page (2) or a range of pages (2-16):
`lpr -o page-ranges={{2|2-16}} {{path/to/file}}`

navarroaxel (approved):

aaaawwWWWwwwwWWW:
oh i didn't saw you approved the changes and  made 2 other changes. that double-sided example was wrong so i fixed it and also turned it into a single example so another can be added in the future.

navarroaxel:
@aaaawwWWWwwwwWWW Don't worry. I reviewed the new changes and I think your changes are good, I made the mistake to not re-count the example count.

sbrl (approved):
Thanks, @aaaawwWWWwwwwWWW!

PR #4968, opened 2020-11-18
vimdiff: cleanup and update page

hodapp512:
More specifically, this updates description, removes more advanced info, and adds commonly used commands

### DIFF
--- a/pages/common/vimdiff.md
+++ b/pages/common/vimdiff.md
@@ -1,16 +1,36 @@
 # vimdiff
 
-> Open up to four files in vim and show the differences between them.
-> See `vim` documentation for working with files and navigating within a window.
+> Open up two or more files in vim and show the differences between them.
+> More information: https://www.vim.org
 
-- Open two files and show the differences (up to four files can be compared):
+- Open two files and show the differences:
 
 `vimdiff {{file1}} {{file2}}`
 
-- Open two files using a horizontal window split instead of the default vertical split:
+- Move the cursor to the window on the left|right:
 
-`vimdiff -o {{file1}} {{file2}}`
+`Ctrl + w {{h|l}}`
 
-- Move the cursor to the window on the left|right|up|down:
+- Jump to next difference
 
-`Ctrl + w {{h|l|k|j}}`
+`[c`
+
+- Jump to previous difference
+
+`]c`
+
+- Obtain difference from other window to current window
+
+`do`
+
+- Put difference from current window to other window
+
+`dp`
+
+- Update the diff highlighting and folds
+
+`:diffupdate`
+
+- Toggle code fold
+
+`za`
--- a/pages/common/vimdiff.md
+++ b/pages/common/vimdiff.md
@@ -1,7 +1,7 @@
 # vimdiff
 
 > Open up two or more files in vim and show the differences between them.
-> More information: https://www.vim.org
+> More information: https://www.vim.org.
 
 - Open two files and show the differences:
 
@@ -11,26 +11,26 @@
 
 `Ctrl + w {{h|l}}`
 
-- Jump to next difference
+- Jump to next difference:
 
 `[c`
 
-- Jump to previous difference
+- Jump to previous difference:
 
 `]c`
 
-- Obtain difference from other window to current window
+- Obtain difference from other window to current window:
 
 `do`
 
-- Put difference from current window to other window
+- Put difference from current window to other window:
 
 `dp`
 
-- Update the diff highlighting and folds
+- Update the diff highlighting and folds:
 
 `:diffupdate`
 
-- Toggle code fold
+- Toggle code fold:
 
 `za`
--- a/pages/common/vimdiff.md
+++ b/pages/common/vimdiff.md
@@ -2,7 +2,7 @@
 
 > Open up two or more files in vim and show the differences between them.
 > See also `vim`.
-> More information: https://www.vim.org.
+> More information: <https://www.vim.org>.
 
 - Open two files and show the differences:
 
--- a/pages/common/vimdiff.md
+++ b/pages/common/vimdiff.md
@@ -1,6 +1,7 @@
 # vimdiff
 
 > Open up two or more files in vim and show the differences between them.
+> See also `vim`.
 > More information: https://www.vim.org.
 
 - Open two files and show the differences:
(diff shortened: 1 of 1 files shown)


pages/common/vimdiff.md
```diff
@@ -1,16 +1,36 @@
 # vimdiff
 
-> Open up to four files in vim and show the differences between them.
-> See `vim` documentation for working with files and navigating within a window.
+> Open up two or more files in vim and show the differences between them.
+> More information: https://www.vim.org.
```
sbrl, line 4: Maybe we could add a see also here too?
```suggestion
> See also `vim`.
> More information: https://www.vim.org.
```

pages/common/vimdiff.md
```diff
@@ -1,16 +1,36 @@
...
-`vimdiff -o {{file1}} {{file2}}`
+`Ctrl + w {{h|l}}`
 
-- Move the cursor to the window on the left|right|up|down:
+- Jump to next difference:
 
-`Ctrl + w {{h|l|k|j}}`
+`[c`
+
+- Jump to previous difference:
```
sbrl, line 18: Perhaps this will read better?
```suggestion
- Jump to the previous difference:
```

pages/common/vimdiff.md
```diff
@@ -1,16 +1,36 @@
...
 `vimdiff {{file1}} {{file2}}`
 
-- Open two files using a horizontal window split instead of the default vertical split:
+- Move the cursor to the window on the left|right:
 
-`vimdiff -o {{file1}} {{file2}}`
+`Ctrl + w {{h|l}}`
 
-- Move the cursor to the window on the left|right|up|down:
+- Jump to next difference:
```
sbrl, line 14: Perhaps this will read better?
```suggestion
- Jump to the next difference:
```

pages/common/vimdiff.md
```diff
@@ -1,16 +1,36 @@
...
 
-`Ctrl + w {{h|l|k|j}}`
+`[c`
+
+- Jump to previous difference:
+
+`]c`
+
+- Obtain difference from other window to current window:
```
sbrl, line 22: Perhaps this will read better?
```suggestion
- Copy differences from another window to the current window:
```

pages/common/vimdiff.md
```diff
@@ -1,16 +1,36 @@
...
+
+`do`
+
+- Put difference from current window to other window:
```
sbrl, line 26: Perhaps this will read better?
```suggestion
- Copy the differences from the current window to another window:
```

pages/common/vimdiff.md
```diff
@@ -1,16 +1,36 @@
...
+
+`dp`
+
+- Update the diff highlighting and folds:
+
+`:diffupdate`
+
+- Toggle code fold:
```
sbrl, line 34: Perhaps this will read better?
```suggestion
- Toggle code folding:
```

sbrl (changes requested):

pages/common/vimdiff.md
hodapp512: This would actually be incorrect- only the highlighted difference is copied/obtained, and "another" is a bit confusing as this tldr assumes only two windows, so "other" would be a bit more clear than "another". My edit takes these points into consideration.

pages/common/vimdiff.md
hodapp512: See my above response.

pages/common/vimdiff.md
hodapp512: This would also be incorrect since it only toggles the highlighted code fold, not all folds. I've updated this line in my new commit to make that more clear however.

hodapp512:
@sbrl I've made edits based off your suggestions. I've left comments above explaining if I've diverged from your edits in any significant way. Thanks!

pages/common/vimdiff.md
sbrl: Not sure I'm following you, but I'm not a vim user lol :P

Anyway, the point I was trying to get at here is that `Obtain` sounds very odd indeed here. Obtain generally means to "gain possession of something", which I somewhat doubt is the meaning you're after here. Is there a way we can reword this at all?

sbrl:
Thanks for the edits, @hodapp512! I've replied to your above comment.

pages/common/vimdiff.md
hodapp512: Sorry I might not have been clear... I agree, "obtain" does sound a bit odd even though technically that word was [selected by the Vim author](http://vimdoc.sourceforge.net/htmldoc/diff.html#do) to match with the shortcut letter "o", so I have changed the description it mostly based on your suggestion since tldr should be intuitive for beginners. You can see that change in [my most recent update to the PR](https://github.com/tldr-pages/tldr/pull/4968/files).

sbrl: Ah, I see. I didn't realise that `Obtain` was a technical term. In that case, what about this?

```suggestion
- Obtain differences from other window in the current window:
```

hodapp512: Unfortunately, that is still incorrect. This command only grabs one difference, not all of them.

Do you not like [my most recent commit wording](https://github.com/tldr-pages/tldr/pull/4968/files) on line 23?
`- Copy the highlighted difference from the other window to the current window:`

If you like the technical term, I'm happy to modify it to:
`- Obtain the highlighted difference from the other window to the current window:`

sbrl: Ah, I didn't notice that you'd changed it. Thanks!

I think it's fine with the new version.

sbrl (approved):
Thanks, @hodapp512!

pages/common/vimdiff.md
navarroaxel:
```suggestion
> More information: <https://www.vim.org>.
```

navarroaxel (changes requested):

pages/common/vimdiff.md
```diff
@@ -1,16 +1,37 @@
 # vimdiff
 
-> Open up to four files in vim and show the differences between them.
-> See `vim` documentation for working with files and navigating within a window.
+> Open up two or more files in vim and show the differences between them.
+> See also `vim`.
+> More information: https://www.vim.org.
```
navarroaxel, line 5:
```suggestion
> More information: <https://www.vim.org>.
```

navarroaxel (approved):

PR #4987, opened 2020-11-24
find: add space to -exec example to avoid error

mojth:
Add space before backslash in line 31. Without the space you would get the error message: missing argument to exec

- [x ] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).

### DIFF
--- a/pages/common/find.md
+++ b/pages/common/find.md
@@ -28,7 +28,7 @@
 
 - Run a command for each file (use `{}` within the command to access the filename):
 
-`find {{root_path}} -name '{{*.ext}}' -exec {{wc -l {} }}\;`
+`find {{root_path}} -name '{{*.ext}}' -exec {{wc -l {} }} \;`
 
 - Find files modified in the last 7 days, and delete them:
 


sbrl (approved):
Wow, great catch @mojth!

Force-merging sincce this is a trivial change.

pages/common/find.md
```diff
@@ -28,7 +28,7 @@
 
 - Run a command for each file (use `{}` within the command to access the filename):
 
-`find {{root_path}} -name '{{*.ext}}' -exec {{wc -l {} }}\;`
+`find {{root_path}} -name '{{*.ext}}' -exec {{wc -l {} }} \;`
```
sbrl, line 31: Oh oops! I've just noticced that we now have a double space here - one inside the closing `}}` of the tldr-pages `{{token_syntax}}`, and the other now outside. I think this was fine as it was:

Selection_2020-11-27_01:52:10_001_2dc5b7d

sbrl (changes requested):
See above

mojth:
If you copy and paste the example and then replace the "{{}}", you may miss the space. At least that's what happened to me. In my opinion, a space outside of "{{}}" is more visible. But I'm not married to the idea. If you oppose this change, I would be fine too.

sbrl:
Ah, I see @mojth. Generally we expect clients to parse out the `{{token_syntax}}` though. The reason it was done like that is because a triple `}}}` there may confuse in terms of syntax :P

sbrl:
I'm going to close this PR now, but I can always reopen if we want to continue the discussion.

PR #4993, opened 2020-11-27
pdf script: format & add datetime

sbrl:
This is an update to the PDF generation script to add a timestamp to the cover page, which fixes #4970.

It also formats the code to be easier to read.

PR #4996, opened 2020-11-27
ci: add PDF generation to deployment

### LINKED ISSUE #4969: Update the PDF version?
Latest pdf version is 4 years old?

owenvoke:
This adds PDF generation as a build step in the CI. It's currently only generating a Solarized Dark version, however we could upload a `tldr-book-dark.pdf` if we wanted to. 🤷🏻

You can view the  generated output on my test run here: https://github.com/owenvoke/tldr/runs/1462999918

And download the zipped build artifact containing the PDF here: https://github.com/owenvoke/tldr/suites/1570682479/artifacts/28298606

Personally I'm not sure a PDF is worth the wait time (3 minutes for that build step), which will only increase. But if people actually use it, then I guess this seems ok. The only other thing I'm not sure about is if multiple commits go to the master branch, which PDF would be uploaded last.

Closes #4969, and it would probably be good to wait until after #4993 is merged. 👍🏻

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,19 @@ jobs:
     - name: Build
       run: bash scripts/build.sh
 
+    - name: Setup Python for PDF generation
+      if: github.repository == 'tldr-pages/tldr' && github.ref == 'refs/heads/master'
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+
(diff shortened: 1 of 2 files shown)


navarroaxel (approved):

sbrl (approved):
Nice one! Looks great to me.

Now we just need to merge #4993.

PR #5032, opened 2020-12-16
kubectl: add kubectl describe in header

makusu2:
https://github.com/tldr-pages/tldr/pull/4494 added `kubectl describe` help, but the main page still doesn't mention it. Add a note about its usage.

~~This is an important command, but it exceeds the suggested limit of 8 examples (do subcommands count as "examples"?). We could instead replace the `kubectl get` example with `kubectl {{get|describe}}`, but I'm really not a fan of that, since they're two different subcommands.~~

### DIFF
--- a/pages/common/kubectl.md
+++ b/pages/common/kubectl.md
@@ -3,9 +3,13 @@
 > Command line interface for running commands against Kubernetes clusters.
 > More information: <https://kubernetes.io/docs/reference/kubectl/>.
 
-- List all information about a resource with more details:
+- List information about a resource with more details:
 
 `kubectl get {{pod|service|deployment|ingress|...}} -o wide`
+
+- Describe information about a resource with all details:
+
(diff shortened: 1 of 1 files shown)


navarroaxel (changes requested):
We can't accept more than 8 examples, but maybe you can add a description like:
```
> See also `kubectl describe` and other pages for additional information.
```
in the header like the `hg.md` page has.

makusu2:
> We can't accept more than 8 examples, but maybe you can add a description like:
> 
> ```
> > See also `kubectl describe` and other pages for additional information.
> ```
> 
> in the header like the `hg.md` page has.

Sure, I can do that. I've kept my removal of "all" from the `get` example, since that's just incorrect (that's what `describe` is for).

navarroaxel (approved):

sbrl (approved):
Thanks, @makusu2!

PR #4861, opened 2020-10-25
badblocks, ddrescue, fdisk, fsck, ioping, smartctl, wipefs: change /dev/sda into /dev/sdX

Agno94:
- [X] The page (if new), does not already exist in the repo.
- [X] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [X] The page has 8 or fewer examples.
- [X] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [X] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [X] The page description includes a link to documentation or a homepage (if applicable).

PR #5036, opened 2020-12-18
contributing-guides/style-guide: add block device rules

Agno94:
Change to style guide
Add new rules as discussed in PR #4861

> @Agno94 I invite you to collaborate sending a PR for add this rule to the https://github.com/tldr-pages/tldr/blob/master/contributing-guides/style-guide.md file.

_Originally posted by @navarroaxel in https://github.com/tldr-pages/tldr/issues/4861#issuecomment-718686605_

### DIFF
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -64,6 +64,9 @@ Keep the following guidelines in mind when choosing tokens:
    But in a command like `wc -l {{file}}`, using `{{file}}` (without extension) is sufficient.
 7. If the example is clearer with an actual value rather than a generic placeholder, use the actual value.
    For example, use `iostat {{2}}` rather than `iostat {{interval_in_secs}}`.
+8. For any reference to *block devices* use the format `{{/dev/sdX}}` instead of `/dev/sda`.
+   For example use `ddrescue --force --no-scrape {{/dev/sdX}} {{/dev/sdY}}` instead of `ddrescue --force --no-scrape /dev/sda /dev/sdb`.
+   Use `{{/dev/sdXY}}` for partition instead of `/dev/sda2` and `{{/dev/nvmeXnNpM}}` instead of `/dev/nvme0n0p1` only for SSD-specific command.
 
 In general, tokens should make it as intuitive as possible
 to figure out how to use the command and fill it in with values.
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -65,7 +65,7 @@ Keep the following guidelines in mind when choosing tokens:
 7. If the example is clearer with an actual value rather than a generic placeholder, use the actual value.
    For example, use `iostat {{2}}` rather than `iostat {{interval_in_secs}}`.
 8. If a command performs irreversible changes to a file system or to user's devices, then write every example in a way that they cannot be unmindfully copy-pasted by the user.
-   For example, instead of `ddrescue --force --no-scrape /dev/sda /dev/sdb` write `ddrescue --force --no-scrape {{/dev/sdX}} {{/dev/sdY}}` and use the `{{/dev/sdXY}}` placeholder for *block devices* instead of `/dev/sda`.
+   For example, instead of `ddrescue --force --no-scrape /dev/sda /dev/sdb` write `ddrescue --force --no-scrape {{/dev/sdX}} {{/dev/sdY}}` and use the `{{/dev/sdXY}}` placeholder for *block devices* instead of `/dev/sda1`.
 
 In general, tokens should make it as intuitive as possible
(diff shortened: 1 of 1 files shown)


waldyrious:
Hmm, I'm a bit on the fence on this one, for two reasons: first, I'm not sure this recommendation is general enough to warrant adding a new entry to the style guide — it only affects 7 out of the hundreds of pages we have so far, as illustrated in #4861. And second, I honestly feel the recommended placeholders are more likely to confuse than the concrete examples of e.g. sda and sdb. @navarroaxel, could you expand on your reasoning for sdX and sdY being better choices?

navarroaxel:
@waldyrious I believe we should avoid dangerous examples where the user can made changes to the main device or partition on their own computer.
Sometimes a new line is copied with the example and the command is pasted and executed without confirmation of any kind, even the sudo prompt and admin privileges because we already include the `sudo` within the example.
And the `sdX` and `sdXN`are used as placeholders in the Arch Wiki and some GNU man pages that I can find search if you need them.

waldyrious:
I see @navarroaxel, thanks for the detailed context. I think that if we are to include this entry in the guidelines, we should justify referring to the things you alluded to in your comment. That said, I still have some reservations:

1) am still unsure that the new placeholders will provide more good than harm, since we don't know how many people do use the examples as-is and end up doing something unintended as a result, versus how many would be confused by the sdX/sdY, etc. nomenclature, as opposed to the more familiar sda/sdb naming scheme (I know I would be in the second group and not in the first, though of course, I'm just a single data point).
2) I reiterate my first point about this rule being rather specific compared to the other ones in the style guide; so while the intention is commendable, the style guide is not meant to be exhaustive, so we must weigh the inclusion of this rule in light of the reduced number of where it will apply.

Any thoughts about these points?

bl-ue:
@waldyrious what about doing something like `{{/dev/sd<X>}}`?  That would make it pretty obvious to the user to substitute and impossible for commands to use that parameter, possibly pasted directly from the tldr page.

bl-ue:
But why are we even discussing this when @navarroaxel approved and merged the PR that changed `sda -> {{sdX}}` (#4861) and in his comment (https://github.com/tldr-pages/tldr/pull/4861#issuecomment-718686605) invited @Agno94 to open this PR?

Agno94:
>     2\. I reiterate my first point about this rule being rather specific compared to the other ones in the style guide; so while the intention is commendable, the style guide is not meant to be exhaustive, so we must weigh the inclusion of this rule in light of the reduced number of where it will apply.
> 
> 
> Any thoughts about these points?

Maybe the general rules for the style guide can be to avoid writing command examples in a way that might cause harm to the user PC. And this specific case could be just and example.

bl-ue:
I like this change, and I vote for keeping it.  @waldyrious @navarroaxel can we merge?

waldyrious:
> Maybe the general rules for the style guide can be to avoid writing command examples in a way that might cause harm to the user PC. And this specific case could be just and example.

@Agno94 first of all, my apologies for the long delay. As for your suggestion quoted above, yeah, I would definitely agree with a more general rule as you suggest! And reading the thread so far, it seems to me that nobody would object to that. Feel free to make the change whenever's convenient to you.

marchersimon (approved):

Agno94:
> As for your suggestion quoted above, yeah, I would definitely agree with a more general rule as you suggest!

Sorry for the delay. I updated the pull request with a new general rule. Please have a look at the content: English is not my first language.

contributing-guides/style-guide.md
```diff
@@ -64,6 +64,8 @@ Keep the following guidelines in mind when choosing tokens:
    But in a command like `wc -l {{file}}`, using `{{file}}` (without extension) is sufficient.
 7. If the example is clearer with an actual value rather than a generic placeholder, use the actual value.
    For example, use `iostat {{2}}` rather than `iostat {{interval_in_secs}}`.
+8. If a command performs irreversible changes to a file system or to user's devices, then write every example in a way that they cannot be unmindfully copy-pasted by the user.
+   For example, instead of `ddrescue --force --no-scrape /dev/sda /dev/sdb` write `ddrescue --force --no-scrape {{/dev/sdX}} {{/dev/sdY}}` and use the `{{/dev/sdXY}}` placeholder for *block devices* instead of `/dev/sda`.
```
marchersimon, line 68: Did you mean `/dev/sda1` at the end of the sentence?

marchersimon (approved):

contributing-guides/style-guide.md
navarroaxel: Yep, sounds like a partition.

navarroaxel (approved):

bl-ue:
🚀

PR #5040, opened 2020-12-21
sort: add output file example

thomaswuensche:

---

This PR aims to add an example for redirecting the sorted output to a file. I especially like the ability to use the `--output` flag to modify a file in-place, replacing the contents with the sorted lines.
Maybe even using the shorter version `sort -o path/to/file path/to/file`, since `-o` is a common flag, but the long version follows the guidelines and is probably easier to understand.

However, the sort page already has 8 examples. My commit suggestion removes an example that in my opinion refers to a more specific usecase. The possibility to sort by a field is also outlined in the 7th example.

But of course this is open for discussion and i'm interested to hear your feedback!

### DIFF
--- a/pages/common/sort.md
+++ b/pages/common/sort.md
@@ -19,9 +19,9 @@
 
 `sort -n {{path/to/file}}`
 
-- Sort the passwd file by the 3rd field, numerically:
+- Sort a file, printing the output to the specified output file (can be used to sort a file in-place):
 
-`sort -t: -k 3n /etc/passwd`
+`sort --output={{path/to/file}} {{path/to/file}}`
 
 - Sort a file preserving only unique lines:
 
--- a/pages/common/sort.md
+++ b/pages/common/sort.md
@@ -19,17 +19,17 @@
 
 `sort -n {{path/to/file}}`
 
-- Sort a file, printing the output to the specified output file (can be used to sort a file in-place):
+- Sort the passwd file by the 3rd field, numerically:
 
-`sort --output={{path/to/file}} {{path/to/file}}`
+`sort -t: -k 3n /etc/passwd`
 
 - Sort a file preserving only unique lines:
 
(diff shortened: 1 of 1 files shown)


pages/common/sort.md
```diff
@@ -19,9 +19,9 @@
 
 `sort -n {{path/to/file}}`
 
-- Sort the passwd file by the 3rd field, numerically:
+- Sort a file, printing the output to the specified output file (can be used to sort a file in-place):
 
-`sort -t: -k 3n /etc/passwd`
+`sort --output={{path/to/file}} {{path/to/file}}`
```
einverne, line 24: I'm not sure it is ok to replace the origin case, in my opinion, sort by field is used often.

sbrl (commented):
I don't see anything wrong with this. I like the `-t:` from the old example though as the field separator, which I feel is valuable - potentially more valuable than the 7th example. Perhaps we could replace the 7th example with the one you removed?

bl-ue:
I agree with @sbrl.  @thomaswuensche, will you please keep the 5th example and instead replace the 7th example with yours?

thomaswuensche:
Thanks for the feedback everybody! Hope all of you enjoyed the holidays, that’s also why i’m a bit late to respond.

I’ve commited the updated page to replace the 7th example.

One thing i liked about this example though was that it shows piping the output of a command to `sort`. However, this is more of a general concept, so feel free to merge the PR if you don’t think it is that useful.

bl-ue (approved):

bl-ue:
@thomaswuensche

>One thing i liked about this example though was that it shows piping the output of a command to sort

I thought the same thing, but I decided that it was the only example that wasn't all about `sort`

> However, this is more of a general concept

Good point.

navarroaxel (approved):

PR #5060, opened 2020-12-31
nix-shell, nix-env: add page and fix manual link

nagromLobo:

### DIFF
--- a/pages/common/nix-shell.md
+++ b/pages/common/nix-shell.md
@@ -1,0 +1,28 @@
+# nix-shell
+
+> Start an interactive shell based on a Nix expression.
+> More information: <https://nixos.org/manual/nix/stable/#sec-nix-shell>.
+
+- Start with nix expression in `shell.nix` or `default.nix` in the current directory:
+
+`nix-shell`
+
+- Run shell command in non-interactive shell and exit:
+
+`nix-shell --run "{{command}} {{command_arguments}}"`
+
+- Start with expression in `default.nix` in the current directory:
+
+`nix-shell {{default.nix}}`
+
+- Start with packages loaded from nixpkgs:
+
+`nix-shell --packages {{package_name_1}} {{package_name_2}}`
+
+- Start with packages loaded from specific nixpkgs revision:
+
+`nix-shell --packages {{package_names}} -I nixpkgs={{https://github.com/NixOS/nixpkgs/archive/nixpkgs_revision.tar.gz}}`
+
+- Evaluate rest of file in specific interpreter, for use in `#!-scripts` (see <https://nixos.org/manual/nix/stable/#use-as-a-interpreter>):
+
+`nix-shell -i {{interpreter}} --packages {{package_names}}`
--- a/pages/common/nix-env.md
+++ b/pages/common/nix-env.md
@@ -1,7 +1,7 @@
 # nix-env
 
 > Manipulate or query Nix user environments.
-> More information: <https://nixos.org/releases/nix/latest/manual#sec-nix-env>.
+> More information: <https://nixos.org/manual/nix/stable/#sec-nix-env>.
 
 - List all installed packages:
 


pages/common/nix-shell.md
```diff
@@ -0,0 +1,28 @@
+# nix-shell
+
+> Start an interactive shell based on a Nix expression.
+> More information: <https://nixos.org/manual/nix/stable/#sec-nix-shell>.
+
+- Start with nix expression in ./shell.nix or ./default.nix:
```
navarroaxel, line 6: Maybe this is better:
```suggestion
- Start with nix expression in `shell.nix` or `default.nix` in the current directory:
```

pages/common/nix-shell.md
```diff
@@ -0,0 +1,28 @@
...
+
+`nix-shell`
+
+- Run shell command in non-interactive shell and exit:
+
+`nix-shell --run {{cmd}}`
+
+- Start with expression in default.nix in the current directory:
```
navarroaxel, line 14:
```suggestion
- Start with expression in `default.nix` in the current directory:
```

pages/common/nix-shell.md
```diff
@@ -0,0 +1,28 @@
...
+
+`nix-shell default.nix`
```
navarroaxel, line 16: We add the token syntax to this kind of arguments:
```suggestion
`nix-shell {{default.nix}}`
```

pages/common/nix-shell.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Start with packages loaded from nixpkgs:
+
+`nix-shell -p {{pkg_name_1}} {{pkg_name_2}}`
```
navarroaxel, line 20: Please use long argument format:
```suggestion
`nix-shell --packages {{package_name1}} {{package_name2}}`
```

pages/common/nix-shell.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Start with packages loaded from specific nixpkgs revision:
+
+`nix-shell -p {{pkg_names}} -I nixpkgs=https://github.com/NixOS/nixpkgs-channels/archive/{{nixpkgs_revision_hash}}.tar.gz`
```
navarroaxel, line 24: I'm not sure about this URL.
```suggestion
`nix-shell --packages {{package_names}} -I nixpkgs={{https://github.com/NixOS/nixpkgs-channels/archive/{{nixpkgs_revision_hash}}.tar.gz}}`
```

pages/common/nix-shell.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Evaluate rest of file in specific interpreter, for use in #!-scripts (see <https://nixos.org/manual/nix/stable/#use-as-a-interpreter>):
```
navarroaxel, line 26:
```suggestion
- Evaluate rest of file in specific interpreter, for use in `#!-scripts` (see <https://nixos.org/manual/nix/stable/#use-as-a-interpreter>):
```

pages/common/nix-shell.md
```diff
@@ -0,0 +1,28 @@
...
+
+`nix-shell -i {{interpreter}} -p {{pkg_names}}`
```
navarroaxel, line 28:
```suggestion
`nix-shell -i {{interpreter}} --packages {{package_names}}`
```

pages/common/nix-shell.md
```diff
@@ -0,0 +1,28 @@
...
+> Start an interactive shell based on a Nix expression.
+> More information: <https://nixos.org/manual/nix/stable/#sec-nix-shell>.
+
+- Start with nix expression in ./shell.nix or ./default.nix:
+
+`nix-shell`
+
+- Run shell command in non-interactive shell and exit:
+
+`nix-shell --run {{cmd}}`
```
navarroaxel, line 12: How should I type this command for add `command_arguments`?
```suggestion
`nix-shell --run {{command}} {{command_arguments}}`
```

navarroaxel (changes requested):

pages/common/nix-shell.md
nagromLobo: Looks like this is a deprecated URL, thank you! I pulled it out of the official nix-manual, but it looks like it may need to be updated there too. 

Your suggested change has a token in a token, is that allowed?

pages/common/nix-shell.md
```diff
@@ -0,0 +1,28 @@
...
+
+`nix-shell {{default.nix}}`
+
+- Start with packages loaded from nixpkgs:
+
+`nix-shell --packages {{package_name_1}} {{package_name_2}}`
+
+- Start with packages loaded from specific nixpkgs revision:
+
+`nix-shell --packages {{package_names}} -I nixpkgs={{https://github.com/NixOS/nixpkgs/archive/{{nixpkgs_revision_hash}}.tar.gz}}`
```
navarroaxel, line 24: Oh, we have nested `{{`:
```suggestion
`nix-shell --packages {{package_names}} -I nixpkgs={{https://github.com/NixOS/nixpkgs/archive/nixpkgs_revision_hash.tar.gz}}`
```

navarroaxel: Could we use something more generic? Or the packages only comes from the `nixpkgs` repository?
```suggestion
`nix-shell --packages {{package_names}} -I nixpkgs={{https://example.com/nix_package.tar.gz}}`
```

nagromLobo: I think we could, and they could come from a tarball with a single top-level directory or a path to a local file. I'm inclined to say that this example is most valuable if it references `nixpkgs` directly, because that seems like the more common use-case by far, and if you are doing something else, you probably already have a more advanced understanding of nix / this option.

nagromLobo: I updated the placeholder to be `nixpkgs_revision` instead of `nixpkgs_revision_hash` to not rule out other references, let me know what you think?

navarroaxel (approved):

bl-ue (approved):

Issue #5046, opened 2020-12-23
Update to tldr-lint v0.0.9


owenvoke:
We currently can't update as we need to clean up some of the pages. The new release now does the following:
- Validates the spelling of "More information: " (`TLDR016`)
- Validates that the URL uses angle brackets (`TLDR017`)
- Validates that there is only a single link per page (`TLDR018`)
- Validates that there are a maximum of 8 examples per page (`TLDR019`)

Currently when running `tldr-lint pages/`, it returns the following which needs to be resolved:

<details>
<summary>List of pages that need to be addressed</summary>

```text
pages/common/arc.md:4: TLDR016 Label for information link should be spelled exactly `More information: `
pages/common/deemix.md:5: TLDR016 Label for information link should be spelled exactly `More information: `
pages/common/eva.md:4: TLDR016 Label for information link should be spelled exactly `More information: `
pages/common/fnm.md:5: TLDR016 Label for information link should be spelled exactly `More information: `
pages/common/glow.md:4: TLDR016 Label for information link should be spelled exactly `More information: `
pages/common/glow.md:4: TLDR017 Information link should be surrounded with angle brackets
pages/common/ipython.md:4: TLDR017 Information link should be surrounded with angle brackets
pages/common/jq.md:0: TLDR019 Page should only include a maximum of 8 examples
pages/common/kotlin.md:4: TLDR016 Label for information link should be spelled exactly `More information: `
pages/common/nativefier.md:4: TLDR016 Label for information link should be spelled exactly `More information: `
pages/common/nc.md:0: TLDR019 Page should only include a maximum of 8 examples
pages/common/nmap.md:0: TLDR019 Page should only include a maximum of 8 examples
pages/common/notmuch.md:4: TLDR016 Label for information link should be spelled exactly `More information: `
pages/common/tar.md:0: TLDR019 Page should only include a maximum of 8 examples
Error: Parse error on line 5:
...irshcmdref.html>.> NOTE: 'guest_id' can...
----------------------^
Expecting 'INFORMATION_LINK', got 'DESCRIPTION_LINE'
pages/linux/efibootmgr.md:4: TLDR017 Information link should be surrounded with angle brackets
pages/linux/homeshick.md:4: TLDR016 Label for information link should be spelled exactly `More information: `
pages/linux/netselect.md:4: TLDR004 Command descriptions should end in a period
pages/linux/rofi.md:4: TLDR016 Label for information link should be spelled exactly `More information: `
[...]

PR #5077, opened 2021-01-03
nmap: remove SSL example to reduce count to 8

bl-ue:

For #5046

### DIFF
--- a/pages/common/nmap.md
+++ b/pages/common/nmap.md
@@ -32,10 +32,6 @@
 
 `nmap -sSU {{address_or_addresses}}`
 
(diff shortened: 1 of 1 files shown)


bl-ue:
I removed the SSL one because that seems like the most useless/irrelevant to me.

navarroaxel (approved):

owenvoke (approved):

waldyrious:
For future reference, this example was first introduced in #1570 (at the time, the page did have fewer than 8 examples even with this one).

bl-ue:
Hm...that's odd.  Thank you anyway.

PR #5078, opened 2021-01-03
tar: remove last example to reduce count to 8

bl-ue:

For #5046

### DIFF
--- a/pages/common/tar.md
+++ b/pages/common/tar.md
@@ -35,7 +35,3 @@
 - Extract files matching a pattern:
 
 `tar xf {{source.tar}} --wildcards "{{*.html}}"`
-
-- Extract a specific file without preserving the folder structure:
-
-`tar xf {{source.tar}} {{source.tar/path/to/extract}} --strip-components={{depth_to_strip}}`


bl-ue:
I picked the last example because it looked like a useful one, but the user has to specify the `--strip-components` option and it may not be obvious what to pass it.

navarroaxel (approved):

owenvoke (approved):

bl-ue:
@owenvoke why didn't you merge this?  Can I?

owenvoke:
Apologies, I forgot to merge.

bl-ue:
Don't apologize!  I was just wondering if we have some special procedure I was missing.

waldyrious:
For future reference, this example was initially introduced in #3859.

@bl-ue, good question about the self-merge. Our [maintenance guide](https://github.com/tldr-pages/tldr/blob/master/contributing-guides/maintainers-guide.md#ii-handling-prs) does not explicitly address this case, but technically the first rule of handling PRs would allow for a self-merge after two approvals by other maintainers. That said, IMO asking first (as you did) is the right call, since there may be indeed a reason why the merge wasn't performed.

bl-ue:
Okay, thank you @waldyrious!  I should have probably checked when that example was added before removing it :)

That's good info to have.

waldyrious:
> I should have probably checked when that example was added before removing it :)

Just to be clear, I didn't mean to criticize you for not doing that :) I just thought it would be useful to add that info, for historical reference.

bl-ue:
No worries @waldyrious! I'm not offended :)

PR #5079, opened 2021-01-04
yetris: reduce example count to 8

bl-ue:


sahilister (approved):
Thanks @bl-ue and bad slip up on my end.

navarroaxel (approved):

PR #5095, opened 2021-01-06
git-var: add page

bl-ue:

For #3953

### DIFF
--- a/pages/common/git-var.md
+++ b/pages/common/git-var.md
@@ -1,6 +1,6 @@
 # git var
 
-> Prints a Git logical variable.
+> Prints a Git logical variable's value.
 > More information: <https://git-scm.com/docs/git-var>.
 
 - Print the value of a Git logical variable:
--- a/pages/common/git-var.md
+++ b/pages/common/git-var.md
@@ -1,12 +1,13 @@
 # git var
 
 > Prints a Git logical variable's value.
+> See `git config`, which is preferred over `git var`.
 > More information: <https://git-scm.com/docs/git-var>.
 
 - Print the value of a Git logical variable:
 
 `git var {{GIT_AUTHOR_IDENT|GIT_COMMITTER_IDENT|GIT_EDITOR|GIT_PAGER}}`
 
-- [l]ist all Git logical variables (including those found in `.git/config`; deprecated in favor of `git config --list`):
+- [l]ist all Git logical variables:
 
 `git var -l`
--- a/pages/common/git-var.md
+++ b/pages/common/git-var.md
@@ -7,6 +7,6 @@
 
 `git var {{GIT_AUTHOR_IDENT|GIT_COMMITTER_IDENT|GIT_EDITOR|GIT_PAGER}}`
 
-- List all Git logical variables (including those found in `.git/config`; deprecated in favor of `git config -l`):
+- [l]ist all Git logical variables (including those found in `.git/config`; deprecated in favor of `git config --list`):
 
 `git var -l`
--- a/pages/common/git-var.md
+++ b/pages/common/git-var.md
@@ -1,0 +1,12 @@
+# git var
+
+> Prints a Git logical variable.
+> More information: <https://git-scm.com/docs/git-var>.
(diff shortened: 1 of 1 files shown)


pages/common/git-var.md
```diff
@@ -0,0 +1,12 @@
+# git var
+
+> Prints a Git logical variable.
```
sbrl, line 3: Perhaps this will read better?
```suggestion
> Prints the value of a Git logical variable.
```

sbrl (commented):
Thanks for the new page! I've left a comment below for you to review.

pages/common/git-var.md
bl-ue: Yep.  I copied it right from https://git-scm.com/docs/git-var.

schneiderl:
@bl-ue PR looks good but I have a question: in which context those commands that you have opened PRs for are used? It looks like most of the functionality can be achieved by using "git" and its subcommands.

schneiderl (approved):

pages/common/git-var.md
```diff
@@ -0,0 +1,12 @@
+# git var
+
+> Prints a Git logical variable's value.
+> More information: <https://git-scm.com/docs/git-var>.
+
+- Print the value of a Git logical variable:
+
+`git var {{GIT_AUTHOR_IDENT|GIT_COMMITTER_IDENT|GIT_EDITOR|GIT_PAGER}}`
+
+- List all Git logical variables (including those found in `.git/config`; deprecated in favor of `git config -l`):
```
navarroaxel, line 10:
```suggestion
- [l]ist all Git logical variables (including those found in `.git/config`; deprecated in favor of `git config --list`):
```

pages/common/git-var.md
```diff
@@ -0,0 +1,12 @@
...
+
+`git var -l`
```
navarroaxel, line 12: Why include a deprecated command?

navarroaxel (changes requested):

pages/common/git-var.md
bl-ue: It's not deprecated, just the listing of the config variables is deprecated.  If you run `git var -l`, you'll notice the same output as `git config -l`, plus some `CONSTANT_CASE` variables at the end.  This command is still in Git and still works.

sbrl: Oh, I see. So only half the command's functionality is deprecated?

sbrl (approved):
Looks ok to me after @navarroaxel's comment is addressed

pages/common/git-var.md
bl-ue: Yes.  Exactly.

navarroaxel: Should we remove the example @sbrl ?

bl-ue:
@schneiderl I didn't see your comment!

>@bl-ue PR looks good but I have a question: in which context those commands that you have opened PRs for are used? It looks like most of the functionality can be achieved by using "git" and its subcommands.

Yes, I know that, and it makes thing feel a little silly :D but I decided to keep doing it (1) just for the sake of completeness and (2) so that users can know a little about the command&mdash;and what to look for instead.

Do you agree?

pages/common/git-var.md
bl-ue: @sbrl @navarroaxel take a look at https://github.com/tldr-pages/tldr/commit/b680abe810d86b37a9cdbba0223cc2f2b2bac00d

schneiderl:
@bl-ue Agreed. I don't think we are the ones who decide which commands our user will use. Our job is to have everything documented as good as possible. 😄 
Thank you for the job you've been doing here over the past weeks.

pages/common/git-var.md
bl-ue: Actually I like this better:
```diff
-Prints a Git logical variable's value.
+Prints the value of a Git logical variable.
```
WDYT, @sbrl?

navarroaxel (approved):

Issue #4701, opened 2020-10-15
Let's document! GitHub CLI (gh)


antoine-amara:
I used the github client (gh) and I consult the tldr documentation and it is already very good, but I think we can improve a little bit this command documentation.

First, they're some little mistake on `gh pr view` and `gh issue view`, we can update the command to correspond to the description.

Moreover, we can create other examples to cover the entire github process from repo creation to release:

* add an example to create a repo.
* add an example to merge a PR.
* add an example to create a release.
* add an example to list all releases.

Finally, If it is ok I can propose a PR to make these changes.

After some discussions , it is better to create sub-pages for the `gh` command:

For those not in the know, tldr-pages is about more than just commands. Those commands that are too expansive to fit on 1 page often have multiple subcommands - which we can document on their own pages. For example, try these:

```
tldr git merge
tldr git add
tldr ip address
```

So, let's document `gh`, we need:

- [x] [`alias`](https://cli.github.com/manual/gh_alias) (#5125)
- [x] [`api`](https://cli.github.com/manual/gh_api) (#5127)
- [x] [`auth`](https://cli.github.com/manual/gh_auth) (#4839)
- [x] [`completion`](https://cli.github.com/manual/gh_completion) (#5134)
- [x] [`config`](https://cli.github.com/manual/gh_config) (#4837)
- [x] [`environment`](https://cli.github.com/manual/gh_help_environment) (#5131)
- [x] [`gist`](https://cli.github.com/manual/gh_gist) (#4898)
- [x] [`help`](https://cli.github.com/manual/gh_help) (#5129)
- [x] [`issue`](https://cli.github.com/manual/gh_issue) (#4836)
- [x] [`pr`](https://cli.github.com/manual/gh_pr) (#4717)
- [x] [`reference`](https://cli.github.com/manual/gh_help_reference) (#5130)
- [x] [`release`](https://cli.github.com/manual/gh_release) (#4773)
- [x] [`repo`](https://cli.github.com/manual/gh_repo) (#4897)
- [x] [`secret`](https://cli.github.com/manual/gh_secret) (#5133)

navarroaxel:
I think you should implement the same structure that the ticket #3953 to create pages for all the `gh` subcommands

antoine-amara:
After some discussions , it is better to create sub-pages for the `gh` command:

[...]

PR #5119, opened 2021-01-08
multiple pages: format technical tokens

schneiderl:
This PR updates some tokens that had the backticks missing. 
This does not tackle all the issues but it is a head start on fixing this formatting inconsistency. 

It basically updates filenames to have backticks while also fixing some other minor issues found. Please take a look at the changed files to make sure the changes make sense. 

**Also**: I've messed up the commit history but this should not be a problem. The previous commits are already merged so there will be no changes on the master branch (apart from 2 conflicts I've manually fixed).

PR #5125, opened 2021-01-10
gh-alias: add page

navarroaxel:

For: #4701

### DIFF
--- a/pages/common/gh-alias.md
+++ b/pages/common/gh-alias.md
@@ -1,13 +1,13 @@
 # gh alias
 
-> Manage GitHub command shortcuts from the command line.
+> Manage GitHub CLI command aliases from the command line.
 > More information: <https://cli.github.com/manual/gh_alias>.
 
 - Display the subcommand help:
 
 `gh alias`
 
-- List all of the aliases gh is configured to use:
+- List all of the aliases `gh` is configured to use:
 
 `gh alias list`
 
--- a/pages/common/gh-alias.md
+++ b/pages/common/gh-alias.md
@@ -11,11 +11,11 @@
 
 `gh alias list`
 
-- Set a command shortcut:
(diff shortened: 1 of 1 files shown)


pages/common/gh-alias.md
```diff
@@ -0,0 +1,24 @@
+# gh alias
+
+> Manage GitHub command shortcuts from the command line.
+> More information: <https://cli.github.com/manual/gh_alias>.
+
+- Display the subcommand help:
+
+`gh alias`
+
+- List all of the aliases gh is configured to use:
```
bl-ue, line 10: We should probably do code per #5119, WDYT?
```suggestion
- List all of the aliases `gh` is configured to use:
```

pages/common/gh-alias.md
```diff
@@ -0,0 +1,24 @@
...
+
+`gh alias list`
+
+- Set a command shortcut:
```
schneiderl, line 14: I don't think I like the use of "shortcut" in favor of using the term "alias". Aliases are well known in our context so I would prefer having them as "alias". 
But since this is just personal preference and not an issue on the page, I'm approving the PR.
If anyone agrees with the alias thing please hop in.

schneiderl (approved):

pages/common/gh-alias.md
bl-ue: I think I agree with that, once you point it out.

pages/common/gh-alias.md
```diff
@@ -0,0 +1,24 @@
+# gh alias
+
+> Manage GitHub command shortcuts from the command line.
```
bl-ue, line 3: I'm also not sure this is that clear.  Perhaps this will read better?
```suggestion
> Manage GitHub CLI command aliases from the command line.
```

bl-ue (approved):

PR #5128, opened 2021-01-10
style-guide: add backticks guide

schneiderl:
This PR proposes a set of rules to use `backticks` in the command descriptions. Currently, it feels like there are no rules and there is no document to back us up in case we have any doubts on this. 

This goes along with #5119 as this sets the set of rules and #5119 implements them.
This is an only personal preference so I'm more than welcome to receive any feedback to add or remove anything from this set of rules.

I would especially like other's opinions about if we should or should not add two types of information to that list:

1. File types, ex. `JSON`
2. Addresses, ex. `localhost`, `123.456.789.1:3030`

What are your thoughts on this?

### DIFF
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -78,7 +78,3 @@ Use backticks on the following:
 4. Commands, ex `ls`.
 5. Keyboard commands, ex. `CTRL+ALT+DEL`.
 6. Port numbers, ex. `8080`.
-
-
-
-
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -68,7 +68,6 @@ Keep the following guidelines in mind when choosing tokens:
 In general, tokens should make it as intuitive as possible
 to figure out how to use the command and fill it in with values.
 
-
 More technical wording on description lines should use the `backtick` syntax.
 Use backticks on the following:
 
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -74,4 +74,3 @@ Use backticks on the following:
 1. Paths, ex. `package.json`, `/etc/package.json`.
 2. Extensions, ex. `.dll`.
 3. Commands, ex. `ls`.
-4. Keyboard commands, ex. `CTRL+ALT+DEL`.
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -73,8 +73,8 @@ More technical wording on description lines should use the `backtick` syntax.
 Use backticks on the following:
 
 1. Filenames, ex. `package.json`.
-2. Directories, ex `/etc/ansible/hosts`.
-3. Extensions, ex `.dll`.
-4. Commands, ex `ls`.
+2. Directories, ex. `/etc/ansible/hosts`.
+3. Extensions, ex. `.dll`.
+4. Commands, ex. `ls`.
 5. Keyboard commands, ex. `CTRL+ALT+DEL`.
 6. Port numbers, ex. `8080`.
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -67,3 +67,18 @@ Keep the following guidelines in mind when choosing tokens:
 
 In general, tokens should make it as intuitive as possible
 to figure out how to use the command and fill it in with values.
+
+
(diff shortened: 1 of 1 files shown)


contributing-guides/style-guide.md
```diff
@@ -67,3 +67,14 @@ Keep the following guidelines in mind when choosing tokens:
 
 In general, tokens should make it as intuitive as possible
 to figure out how to use the command and fill it in with values.
+
+
```
navarroaxel, line 71:
```suggestion
```

contributing-guides/style-guide.md
```diff
@@ -67,3 +67,14 @@ Keep the following guidelines in mind when choosing tokens:
...
+More technical wording on description lines should use the `backtick` syntax.
+Use backticks on the following:
+
+1. Filenames, ex. `package.json`.
```
bl-ue, line 75: How about merging 1 and 2 togerther into "Paths"?

schneiderl: IMO it is not clear that filenames would be included in "paths". 
I still believe we can merge them if we have two examples: a sole filename and a full path (directory+filename). What do you think?

bl-ue: Good idea.  That must be the missing bit I was sensing.

schneiderl: done

einverne (approved):
Looks good to me

navarroaxel (approved):

bl-ue (approved):

schneiderl:
May we merge this then?

bl-ue:
I think it might be best to have the opinion of @waldyrious or @sbrl first, but if you think otherwise, I suppose so.

schneiderl:
Agreed. Let's wait for them.

contributing-guides/style-guide.md
```diff
@@ -67,3 +67,12 @@ Keep the following guidelines in mind when choosing tokens:
...
 to figure out how to use the command and fill it in with values.
+
+More technical wording on description lines should use the `backtick` syntax.
+Use backticks on the following:
+
+1. Paths, ex. `package.json`, `/etc/package.json`.
+2. Extensions, ex. `.dll`.
+3. Commands, ex. `ls`.
+4. Keyboard commands, ex. `CTRL+ALT+DEL`.
+5. Port numbers, ex. `8080`.
```
sbrl, line 78: I disagree strongly. Port numbers do not feel right in backticks.

contributing-guides/style-guide.md
```diff
@@ -67,3 +67,12 @@ Keep the following guidelines in mind when choosing tokens:
...
 In general, tokens should make it as intuitive as possible
 to figure out how to use the command and fill it in with values.
+
+More technical wording on description lines should use the `backtick` syntax.
+Use backticks on the following:
+
+1. Paths, ex. `package.json`, `/etc/package.json`.
+2. Extensions, ex. `.dll`.
+3. Commands, ex. `ls`.
+4. Keyboard commands, ex. `CTRL+ALT+DEL`.
```
sbrl, line 77: I don't think this works very well. Backticks indicate a file or command, and this is neither.

sbrl (changes requested):
Good idea, but I disagree on some points here.

contributing-guides/style-guide.md
schneiderl: Agreed. I'll update it.

contributing-guides/style-guide.md
schneiderl: Isn't it a command?

schneiderl: Actually, after some thought, I think I agree that this should not be surrounded by backticks.
Removed it from the style guide.

sbrl (approved):
This looks ok to me. I do not feel comfortable with merging this without @waldyrious weighing in though, as they are much better at commenting on this kinda documentation change.

schneiderl:
Agreed @sbrl.

schneiderl:
Should we merge this without @waldyrious approval? There has been some time without any answer from him.

sbrl:
Hrm. It has been 9 days, so I think we should be ok to merge this. Merging the  PR does not preclude further discussion though, and we can always open another PR to correct any issues.

PR #5136, opened 2021-01-11
multiple pages: normalize line endings from CRLF to LF

bl-ue:
I used [dos2unix](https://waterlan.home.xs4all.nl/dos2unix.html) to convert all files with Windows-style line endings (CRLF) to the standard Git line endings (which 90%+ of our pages are), Unix (LF).

The culprits are:
- mostly [`pages.ko`](https://github.com/tldr-pages/tldr/tree/master/pages.ko/common) (cc @Proscream)
- some in [`pages.ru`](https://github.com/tldr-pages/tldr/tree/master/pages.ru/common) (cc @Ryzhehvost, @NikonP, @subzey, @Bodyhealer)
- one in [`pages.es`](https://github.com/tldr-pages/tldr/tree/master/pages.es/common) (cc @menykuro).


MasterOdin:
It would probably make sense to add a `.gitattributes` file to force the `lf` line ending for text files.

bl-ue:
Good idea!  I'll do that next.  But we still need this PR, correct?

bl-ue:
@navarroaxel / @waldyrious would it be OK to add the `.gitattributes` files in this PR in a new commit?

bl-ue:
@MasterOdin done.

mebeim (approved):
Good job and thank you for the PR 🙂 

One note though: this could have been done in a single commit (or maybe two, one for pages and one for `.gitattributes`) with a more descriptive message (we usually do these simple typo fixes in batches), for example:

```
Refactor: normalize line endings in pages
```

Same goes for the `.gitattribute` commit message. Please, try being more descriptive if possible, simply saying "Add `.gitattributes` file" does not explain the reason of the commit. A better commit message would have been for example:

```
Add .gitattributes file to normalize line endings

This will normalize all line endings of files recognized by Git as text
to LF (Unix style).
```

We can always squash all these commits when merging and change the commit message, so this is no issue for this simple PR, but in the case of more complex PRs where separate commits matter and squashing isn't an option, remember to be descriptive.

Thank you!

---

**Note to other maintainers**: please squash and apply a more descriptive commit message when merging.

bl-ue:
@mebeim thank you for the heads up.  I can also squash and create to separate commits locally and then force-push, if you'd like.

mebeim:
@bl-ue oh, if that's no problem for you then sure, that's even better. Thank you :)

bl-ue:
@mebeim done.

.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto
```
waldyrious, line 1: Maybe add a comment above this line explaining what this does, and add a link to the relevant documentation.

waldyrious (changes requested):
Besides the comment I left inline, please rename the commit according to the PR title that I just edited. If you want, you could also add the details from the PR's opening comment as an extended commit message.

schneiderl (approved):

owenvoke (approved):
LGTM!

.gitattributes
bl-ue: @waldyrious done.

.gitattributes
```diff
@@ -0,0 +1,5 @@
+# This ensures that the line endings in any files added or modified are
+# normalized before being committed.  On Windows they will automatically
```
navarroaxel, line 2:
```suggestion
# normalized before being committed. On Windows they will automatically
```

bl-ue: Whoops! :smile:

schneiderl (approved):

waldyrious (approved):
Thanks for doing this, @bl-ue!

subzey:
Probably we could add an `.editorconfig` file also:

```ini
[*.md]
end_of_line = lf
```
or even
```ini
[*]
end_of_line = lf
```

It doesn't do much as itself, but there are [many IDEs that respects the editorconfig file](https://editorconfig.org/#pre-installed)

bl-ue:
Good idea&mdash;would you like to open a PR to add that @subzey?  If you do, please choose the **second** option as LF is generally desirable for all text files in a repo except certain cases, e.g. test data.

subzey:
@bl-ue I'm sorry, it turns out there is already https://github.com/tldr-pages/tldr/blob/master/.editorconfig with that setting.

My bad, for some reason I've missed that in the file listing :(

bl-ue:
Oh yes, my fault too. 🙂 Well the PR that added that was merged a while after the Korean pages were created.

sbrl:
Many contributors use the GitHub web interface, so it's likely some weird line endings managed to sneak in like that.

owenvoke:
Yeah, I think GitHub follows EditorConfig rules, so should be fine now. 👍🏻

PR #5139, opened 2021-01-13
lambo, lambo-new: add page

owenvoke:

### DIFF
--- a/pages/common/lambo.md
+++ b/pages/common/lambo.md
@@ -1,0 +1,36 @@
+# lambo
+
+> A super-powered `laravel new` for Laravel and Valet.
+> More information: <https://github.com/tighten/lambo>.
+
+- Create a new Laravel application:
+
+`lambo new {{name}}`
+
+- Install the application in a specific path:
+
+`lambo new --path={{path/to/directory}} {{name}}`
+
+- Include authentication scaffolding:
+
+`lambo new --auth {{name}}`
+
+- Include a specific frontend:
+
+`lambo new --{{vue|bootstrap|react}} {{name}}`
+
+- Install npm dependencies after the project has been created:
+
+`lambo new --node {{name}}`
+
+- Create a Valet site after the project has been created:
+
+`lambo new --link {{name}}`
+
+- Create a new MySQL database with the same name as the project:
+
+`lambo new --create-db --dbuser={{user}} --dbpassword={{password}} {{name}}`
+
+- Open a specific editor after the project has been created:
+
+`lambo new --editor="{{editor}}" {{name}}`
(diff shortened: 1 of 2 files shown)


bl-ue:
Is there a reason why you sometimes use `=` in flags but sometimes not?
Like this:
```
lambo new --path={{path/to/directory}} {{name}}

lambo new --auth {{name}}

**Comment by owenvoke** [comment]
@bl-ue, because the ones without an equals sign don't have a value (they are a boolean value). Whereas the ones with `=` need a value.

**Comment by bl-ue** [comment]
Seems like an obvious one now 😄

**Review by bl-ue — APPROVED** [review]

**Review by navarroaxel — COMMENTED** [review]
I think this should be move to `lambo-new.md` and create a `lambo.md` for `--help`, `edit-config`, `make-after`, `edit-after` and the simplest `lambo new` subcommands.

File: pages/common/lambo.md [hunk 1]
```diff
@@ -0,0 +1,36 @@
+# lambo
+
+> A super-powered `laravel new` for Laravel and Valet.
+> More information: <https://github.com/tighten/lambo>.
+
+- Create a new Laravel application:
+
+`lambo new {{name}}`
```
sbrl, line 8: Maybe this would aid clarity? Same goes for the rest of the examples.
```suggestion
`lambo new {{app_name}}`
```

owenvoke: Yeah, I did think about that. I was just being consistent with other pages.

sbrl: Ah, ok. Still, I think this adds enough clarity to consider breaking consistency.

I guess it doesn't really matter then.

owenvoke: I'll update this shortly. Cheers.

navarroaxel:
Hey @owenvoke, do you think this is not useful? https://github.com/tldr-pages/tldr/pull/5139#pullrequestreview-568185967

owenvoke:
Apologies @navarroaxel, I missed that comment. I'm not entirely sure that this would be necessary. 🤔

bl-ue:
I'm inclined to agree with @navarroaxel, because what about the other `lambo` subcommands like `edit-config`, `make-after `, and `edit-after`?  Just because they may not be very useful/used often doesn't mean we shouldn't document them, you know.

owenvoke:
Fair enough, I was just showing the most used commands, for reference `make-after` no longer exists as it's irrelevant, and there isn't a `--help` flag (it's displayed when no command or directory is specified). But I've added now split it into another page with the other commands.

bl-ue (approved):

navarroaxel (approved):

bl-ue:
Oh cool!  I didn't notice `lambo` _in addition to_ `lambo-new` :smile:

Thank you @owenvoke!

PR #5148, opened 2021-01-15
tldr: add render example

thomaswuensche:

### DIFF
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -9,7 +9,7 @@
 
 - Show the tar tldr page for Linux:
 
-`tldr --platform {{linux}} {{tar}}`
+`tldr -p {{linux}} {{tar}}`
 
 - Get help for a Git subcommand:
 
@@ -17,8 +17,8 @@
 
 - Update local pages (if the client supports caching):
 
-`tldr --update`
+`tldr -u`
 
-- Render a local page for testing purposes:
+- Render a local page for testing purposes (if the client supports it):
 
 `tldr --render {{path/to/page}}`
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -9,7 +9,7 @@
 
 - Show the tar tldr page for Linux:
 
-`tldr -p {{linux}} {{tar}}`
+`tldr --platform {{linux}} {{tar}}`
 
 - Get help for a Git subcommand:
 
@@ -17,4 +17,8 @@
 
 - Update local pages (if the client supports caching):
 
-`tldr -u`
+`tldr --update`
+
+- Render a local page for testing purposes:
+
+`tldr --render {{path/to/page}}`


bl-ue (approved):

pages/common/tldr.md
```diff
@@ -9,12 +9,16 @@
 
 - Show the tar tldr page for Linux:
 
-`tldr -p {{linux}} {{tar}}`
+`tldr --platform {{linux}} {{tar}}`
```
navarroaxel, line 12: Use short version since long version is only optional to support for clients.
```suggestion
`tldr -p {{linux}} {{tar}}`
```

pages/common/tldr.md
```diff
@@ -9,12 +9,16 @@
...
 
 - Get help for a Git subcommand:
 
 `tldr {{git-checkout}}`
 
 - Update local pages (if the client supports caching):
 
-`tldr -u`
+`tldr --update`
```
navarroaxel, line 20:
```suggestion
`tldr -u`
```

pages/common/tldr.md
```diff
@@ -9,12 +9,16 @@
...
 - Show the tar tldr page for Linux:
 
-`tldr -p {{linux}} {{tar}}`
+`tldr --platform {{linux}} {{tar}}`
 
 - Get help for a Git subcommand:
 
 `tldr {{git-checkout}}`
 
 - Update local pages (if the client supports caching):
```
navarroaxel, line 18: We can explain the short argument:
```suggestion
- [u]pdate local pages (if the client supports caching):
```

pages/common/tldr.md
```diff
@@ -9,12 +9,16 @@
...
 
-`tldr -u`
+`tldr --update`
+
+- Render a local page for testing purposes:
```
navarroaxel, line 22:
```suggestion
- Render a local page for testing purposes (if the client supports it):
```

navarroaxel (changes requested):
Hi @thomaswuensche, thanks for your contribution.

The -r/--render command line flag is not listed for all TLDR clients compliant to [our specification](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md) (see the [arguments section](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#arguments)), but only for those which implement it.

Also, the long version is also not mandatory. See my suggestions below.

thomaswuensche:
Thanks for the feedback @navarroaxel!
Would it be better then to also use `-r` instead of `--render`? Since it's not in the specifications, i don't know if some clients choose to only implement a short option, you probably know more about this.

pages/common/tldr.md
thomaswuensche: I prefer the original version in this case. For me, it breaks with the structure of the rest of the page

schneiderl:
@thomaswuensche We usually prefer the long versions.

pages/common/tldr.md
schneiderl: I also prefer the original version. Also, the `-u` is pretty intuitive and self explanatory IMO.

schneiderl (approved):

pages/common/tldr.md
sbrl: I'm not sure this change adds any value. The existing version looks fine.

Note also this from the client spec:

> Clients MAY choose to only implement the short version of an option, ignoring the long form.

pages/common/tldr.md
```diff
@@ -18,3 +18,7 @@
 - Update local pages (if the client supports caching):
 
 `tldr -u`
+
+- Render a local page for testing purposes (if the client supports it):
+
+`tldr --render {{path/to/page}}`
```
sbrl, line 24: This isn't a standard argument defined in the tldr [client specification](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#arguments), so unfortunately we can't include it here.

thomaswuensche: Hey @sbrl, thanks for the feedback! Then i guess this PR is no longer useful... May it rest in peace 😄

bl-ue: It most certainly will!  PRs, once closed/merged, seem to have a very peaceful existence here on GitHub! 😂

sbrl: @thomaswuensche sorry about that. We hope that you'll return with more contributions in the future :-)

sbrl:
Closing, as it has been agreed with the author that this PR is not going to be merged.

Issue #5103, opened 2021-01-07
Let's document: next-gen filesystems (btrfs/zfs)


sbrl:
Since #4250 was recently closed, I thought that maybe we could start tracking a new list of file-system related commands. This time, the theme is the next-generation filesystems [btrfs](https://btrfs.wiki.kernel.org/) and (open [zfs](https://openzfs.org/wiki/Main_Page) - please feel free to update this list if I've made a mistake.

## Btrfs

 - [ ] `btrfs subvolume`
 - [ ] `btrfs filesystem` (#5102)
 - [ ] `btrfs balance`
 - [ ] `btrfs device`
 - [ ] `btrfs scrub` (recommended monthly, according to the btrfs wiki - maybe we could mention that in the description?)
 - [ ] `btrfs check`
 - [ ] `btrfs rescue`
 - [ ] `btrfs restore`
 - [ ] `btrfs inspect-internal`
 - [ ] `btrfs property`
 - [ ] `btrfs send`
 - [ ] `btrfs receive`
 - [ ] `btrfs quota`
 - [ ] `btrfs qgroup`
 - [ ] `btrfs replace`
 - [ ] `btrfs help`
 - [ ] `btrfs version` (maybe?)

## ZFS

 - [ ] `zfs allow`
 - [ ] `zfs bookmark`
 - [ ] `zfs change-key`
 - [ ] `zfs clone`
 - [ ] `zfs create`
 - [ ] `zfs destroy`
 - [ ] `zfs diff`
 - [ ] `zfs get`
 - [ ] `zfs groupspace`
 - [ ] `zfs hold`
 - [ ] `zfs holds`
 - [ ] `zfs inherit`
 - [ ] `zfs list`
 - [ ] `zfs load-key`
 - [ ] `zfs mount`
 - [ ] `zfs program`
 - [ ] `zfs project`
 - [ ] `zfs projectspace`
 - [ ] `zfs promote`
 - [ ] `zfs receive`
 - [ ] `zfs redact`
 - [ ] `zfs release`
 - [ ] `zfs rename`
 - [ ] `zfs rollback`
 - [ ] `zfs send`
 - [ ] `zfs set`
 - [ ] `zfs share`
 - [ ] `zfs snapshot`
 - [ ] `zfs unallow`
 - [ ] `zfs unload-key`
 - [ ] `zfs unmount`
 - [ ] `zfs unshare`
 - [ ] `zfs upgrade`
 - [ ] `zfs userspace`
[...]

PR #5153, opened 2021-01-16
btrfs-subvolume: add page

sbrl:
My continuing question to explore strange new worlds..... oh wait, that's star trek.

For #5103


pages/linux/btrfs-subvolume.md
```diff
@@ -0,0 +1,28 @@
...
+
+`sudo btrfs subvolume list {{path/to/btrfs_filesystem}}`
+
+- Delete a subvolume:
+
+`sudo btrfs subvolume delete {{path/to/subvolume}}`
+
+- Create a read-only snapshot of an existing subvolume:
+
+`sudo btrfs subvolume snapshot -r {{path/to/source_subvolume}} {{path/to/target}}`
```
bl-ue, line 20: Not sure if this is correct...
```suggestion
`sudo btrfs subvolume snapshot --rootid {{path/to/source_subvolume}} {{path/to/target}}`
```

sbrl: It isn't, actually. That would be the long form of `-r`, yes - but the `-r` for `show` and not `snapshot`. For the snapshot subsubcommand, it's short for "read-only", rather than "rootid".

bl-ue: I though maybe so.

bl-ue (approved):

navarroaxel (approved):

PR #5160, opened 2021-01-18
cmake: update page

bergentroll:
Actualise examples and improve description.

### DIFF
--- a/pages/common/cmake.md
+++ b/pages/common/cmake.md
@@ -7,9 +7,9 @@
 
 `cmake {{path/to/project_directory}}`
 
-- Generate a build recipe, build type setted to `Release` with CMake variable:
+- Generate a build recipe, with build type set to `Release` with CMake variable:
 
-`cmake {{path/to/project_directory}} -D CMAKE_BUILD_TYPE=Release`
+`cmake {{path/to/project_directory}} -D {{CMAKE_BUILD_TYPE=Release}}`
 
 - Build artifacts with the generated recipe in `build_directory`:
 
@@ -21,7 +21,7 @@
 
 - Install the build artifacts using the custom prefix for paths:
 
-`cmake --install {{path/to/build_directory}} --strip --prefix {{path/to/prefix_directory}}`
+`cmake --install {{path/to/build_directory}} --strip --prefix {{path/to/directory}}`
 
 - Run a custom build target:
 
--- a/pages/common/cmake.md
+++ b/pages/common/cmake.md
@@ -1,21 +1,28 @@
 # cmake
 
-> Cross-platform build, testing and packaging automation system.
-> CMake uses its own shell-like syntax and generates recipes for many build systems.
+> Cross-platform build automation system, that generates recipes for native build systems.
 > More information: <https://cmake.org/cmake/help/latest/manual/cmake.1.html>.
 
 - Generate a build recipe in the current directory with `CMakeLists.txt` from a project directory:
 
 `cmake {{path/to/project_directory}}`
 
+- Generate a build recipe, build type setted to `Release` with CMake variable:
+
+`cmake {{path/to/project_directory}} -D CMAKE_BUILD_TYPE=Release`
+
 - Build artifacts with the generated recipe in `build_directory`:
 
 `cmake --build {{path/to/build_directory}}`
 
-- Install the build artifacts into /usr/local/ striping debugging symbols:
+- Install the build artifacts into `/usr/local/` and strip debugging symbols:
 
-`cmake --install {{path/to/build_directory}} --strip --prefix /usr/local/`
+`cmake --install {{path/to/build_directory}} --strip`
+
+- Install the build artifacts using the custom prefix for paths:
+
+`cmake --install {{path/to/build_directory}} --strip --prefix {{path/to/prefix_directory}}`
 
 - Run a custom build target:
 
-`cmake --build {{path/to/build_dir/}} --target {{target_name}}`
+`cmake --build {{path/to/build_directory}} --target {{target_name}}`
--- a/pages/common/cmake.md
+++ b/pages/common/cmake.md
@@ -11,7 +11,7 @@
 
 `cmake {{path/to/project_directory}} -D {{CMAKE_BUILD_TYPE=Release}}`
 
-- Build artifacts with the generated recipe in `build_directory`:
+- Use a generated recipe in a given directory to build artifacts:
 
 `cmake --build {{path/to/build_directory}}`
 
--- a/pages/common/cmake.md
+++ b/pages/common/cmake.md
@@ -1,16 +1,16 @@
 # cmake
 
 > Cross-platform build, testing and packaging automation system.
-> CMake uses own shell-like syntax and generates recipes for several build systems.
+> CMake uses its own shell-like syntax and generates recipes for many build systems.
 > More information: <https://cmake.org/cmake/help/latest/manual/cmake.1.html>.
 
-- Generate a build recipe in the current directory with CMakeLists.txt from a project directory:
+- Generate a build recipe in the current directory with `CMakeLists.txt` from a project directory:
 
-`cmake {{path/to/project_dir/}}`
+`cmake {{path/to/project_directory}}`
 
-- Build with the generated recipe in build_dir (artifacts go to build_dir):
+- Build artifacts with the generated recipe in `build_directory`:
 
-`cmake --build {{path/to/build_dir/}}`
+`cmake --build {{path/to/build_directory}}`
 
 - Install the build artifacts into /usr/local/ striping debugging symbols:
 
--- a/pages/common/cmake.md
+++ b/pages/common/cmake.md
@@ -1,8 +1,7 @@
 # cmake
 
 > Cross-platform build, testing and packaging automation system.
-> CMake uses own syntax inspired by shell and generates recipes for build
-> systems like Make, Ninja and Microsoft Visual Studio.
+> CMake uses own shell-like syntax and generates recipes for several build systems.
 > More information: <https://cmake.org/cmake/help/latest/manual/cmake.1.html>.
 
 - Generate a build recipe in the current directory with CMakeLists.txt from a project directory:
--- a/pages/common/cmake.md
+++ b/pages/common/cmake.md
@@ -1,13 +1,22 @@
 # cmake
 
-> Cross-platform build system generator.
-> It generates Makefiles, Visual Studio projects or others, depending on the target system.
+> Cross-platform build, testing and packaging automation system.
+> CMake uses own syntax inspired by shell and generates recipes for build
+> systems like Make, Ninja and Microsoft Visual Studio.
 > More information: <https://cmake.org/cmake/help/latest/manual/cmake.1.html>.
 
-- Generate a Makefile and use it to compile a project in the same directory as the source:
+- Generate a build recipe in the current directory with CMakeLists.txt from a project directory:
 
-`cmake && make`
+`cmake {{path/to/project_dir/}}`
 
-- Generate a Makefile and use it to compile a project in a separate "build" directory (out-of-source build):
+- Build with the generated recipe in build_dir (artifacts go to build_dir):
 
-`cmake -H. -B {{build}} && make -C {{build}}`
+`cmake --build {{path/to/build_dir/}}`
+
+- Install the project:
+
+`cmake --install {{path/to/build_dir/}}`
+
+- Run a custom build target:
+
+`cmake --build {{path/to/build_dir/}} --target {{target_name}}`
--- a/pages/common/cmake.md
+++ b/pages/common/cmake.md
@@ -12,9 +12,9 @@
 
 `cmake --build {{path/to/build_dir/}}`
 
-- Install the project:
+- Install the build artifacts into /usr/local/ striping debugging symbols:
 
-`cmake --install {{path/to/build_dir/}}`
+`cmake --install {{path/to/build_directory}} --strip --prefix /usr/local/`
 
 - Run a custom build target:
 


pages/common/cmake.md
```diff
@@ -1,13 +1,21 @@
 # cmake
 
-> Cross-platform build system generator.
-> It generates Makefiles, Visual Studio projects or others, depending on the target system.
+> Cross-platform build, testing and packaging automation system.
+> CMake uses own shell-like syntax and generates recipes for several build systems.
```
bl-ue, line 4:
```suggestion
> CMake uses its own shell-like syntax and generates recipes for many build systems.
```

pages/common/cmake.md
```diff
@@ -1,13 +1,21 @@
...
 > More information: <https://cmake.org/cmake/help/latest/manual/cmake.1.html>.
 
-- Generate a Makefile and use it to compile a project in the same directory as the source:
+- Generate a build recipe in the current directory with CMakeLists.txt from a project directory:
 
-`cmake && make`
+`cmake {{path/to/project_dir/}}`
```
bl-ue, line 9: It's better not to have a trailing slash and to use the full name for "directory."
```suggestion
`cmake {{path/to/project_directory}}`
```

pages/common/cmake.md
```diff
@@ -1,13 +1,21 @@
...
 
-- Generate a Makefile and use it to compile a project in a separate "build" directory (out-of-source build):
+- Build with the generated recipe in build_dir (artifacts go to build_dir):
```
bl-ue, line 11:
```suggestion
- Build artifacts with the generated recipe in `build_directory`:
```

pages/common/cmake.md
```diff
@@ -1,13 +1,21 @@
...
 
-`cmake -H. -B {{build}} && make -C {{build}}`
+`cmake --build {{path/to/build_dir/}}`
```
bl-ue, line 13:
```suggestion
`cmake --build {{path/to/build_directory}}`
```

pages/common/cmake.md
```diff
@@ -1,13 +1,21 @@
...
+
+- Install the project:
```
bl-ue, line 15: Could you specify where the files get installed, something like `Install the built artifacts to the system's default folders`?

pages/common/cmake.md
```diff
@@ -1,13 +1,21 @@
...
+
+`cmake --install {{path/to/build_dir/}}`
```
bl-ue, line 17:
```suggestion
`cmake --install {{path/to/build_directory}}`
```

pages/common/cmake.md
```diff
@@ -1,13 +1,21 @@
...
+
+- Run a custom build target:
+
+`cmake --build {{path/to/build_dir/}} --target {{target_name}}`
```
bl-ue, line 21:
```suggestion
`cmake --build {{path/to/build_directory}} --target {{target_name}}`
```

pages/common/cmake.md
```diff
@@ -1,13 +1,21 @@
 # cmake
 
-> Cross-platform build system generator.
-> It generates Makefiles, Visual Studio projects or others, depending on the target system.
+> Cross-platform build, testing and packaging automation system.
+> CMake uses own shell-like syntax and generates recipes for several build systems.
 > More information: <https://cmake.org/cmake/help/latest/manual/cmake.1.html>.
 
-- Generate a Makefile and use it to compile a project in the same directory as the source:
+- Generate a build recipe in the current directory with CMakeLists.txt from a project directory:
```
bl-ue, line 7:
```suggestion
- Generate a build recipe in the current directory with `CMakeLists.txt` from a project directory:
```

bl-ue (changes requested):
Hi @bergentroll, and thank you for the page update!  A page for a command like `cmake` really needs to utilize all 8 examples :)

I've left some comments below for you to review.  They're mostly just syntax adjustments, but I left a question, too.

pages/common/cmake.md
bergentroll: @bl-ue, I am not sure cause I'm not native, but I hear "-s" possessive form is unapplicable to inanimate nouns, or does it?

pages/common/cmake.md
bergentroll: @bl-ue, why? The trailing slash points that path is actually a directory, not a file. I believe it is good practice to end directory paths with slash in scripts, it is explicit and prevents errors on concatenation.

pages/common/cmake.md
bergentroll: It hardly depends on a project setup and build system imho.

pages/common/cmake.md
bl-ue: @bergentroll I actually completely agree with you, however, it's long been the convention in the project to omit the trailing slash, so I think it would be best to follow convention.

For example, see [`pages/common/7z.md`](https://github.com/tldr-pages/tldr/blob/master/pages/common/7z.md).

pages/common/cmake.md
bl-ue: I just thinking that "Install the project" probably wouldn't mean much to a new user.

pages/common/cmake.md
bergentroll: @bl-ue, well, with respect, I'd propose to revise this convention, because change will brake nothing. But I follow your decision.

pages/common/cmake.md
bl-ue: I can't remember :smile:, but I am native and I know that "its" in this sentence is correct.

pages/common/cmake.md
bl-ue: @bergentroll I'd recommend asking @sbrl and @waldyrious about that on [Gitter](https://gitter.im/tldr-pages/tldr).

pages/common/cmake.md
bergentroll: Well, I will remember this, thanks!

pages/common/cmake.md
bergentroll: I updated example to be more verbose.

pages/common/cmake.md
```diff
@@ -1,13 +1,21 @@
...
-`cmake && make`
+`cmake {{path/to/project_directory}}`
 
-- Generate a Makefile and use it to compile a project in a separate "build" directory (out-of-source build):
+- Build artifacts with the generated recipe in `build_directory`:
 
-`cmake -H. -B {{build}} && make -C {{build}}`
+`cmake --build {{path/to/build_directory}}`
+
+- Install the build artifacts into /usr/local/ striping debugging symbols:
```
bl-ue, line 15: Looks much better @bergentroll 👍🏻

Again, I believe @navarroaxel would prefer removing the trailing slash:
```suggestion
- Install the build artifacts into `/usr/local` striping debugging symbols:
```

bl-ue: Also, since CMake is emphatically "cross-platform", could you please add the path that artifacts on Windows are installed to (if any, there may not be), in a form like:
```
- Install the build artifacts into `/usr/local` (`...` on Windows) striping debugging symbols:
```

bl-ue: Oops! I noticed a typo.  I edited the above suggestion to account for that.

bergentroll: > Install the build artifacts into `/usr/local` (`...` on Windows) striping debugging symbols:

I confused. Actually the default value for Microsoft Windows is `c:/Program Files/${PROJECT_NAME}`, but the example points exactly on ’/usr/local` to introduce the `--prefix` key.

bl-ue: Then how about just changing it like this (if you apply this suggestion don't apply the one above):
```suggestion
- Install the build artifacts into `/usr/local` (`C:\Program Files\${PROJECT_NAME}` on windows), stripping debugging symbols:
```

bergentroll: So what about this?

```Markdown
Install the build artifacts into a `prefix` directory (default is `/usr/local/` for POSIX systems and `C:\Program Files\${PROJECT_NAME}\` for Microsoft Windows), stripping debugging symbols:

`cmake --install {{path/to/build_directory}} --strip --prefix {{path/to/prefix}}`
```

bl-ue: Hi, looks good, except we can make it a little shorter:
```
Install the build artifacts into a `prefix` directory (default is `/usr/local/` for POSIX systems and `C:\Program Files\${PROJECT_NAME}\` for Windows), stripping debug symbols:
```
1. "Microsoft Windows" is unnecessarily formal
2. "debug" -> "debugging" because they convey the same meaning, and I think the two "ing"s (in `stripping debugging`) sound slightly unnatural

bergentroll: I dislike the point 1, because "Windows" is to general word. But I obey.

pages/common/cmake.md
```diff
@@ -1,13 +1,21 @@
 # cmake
 
-> Cross-platform build system generator.
-> It generates Makefiles, Visual Studio projects or others, depending on the target system.
+> Cross-platform build, testing and packaging automation system.
+> CMake uses its own shell-like syntax and generates recipes for many build systems.
```
waldyrious, line 4: I'm afraid this second line like makes things too generic to be useful for the uninitiated. I like how the previous version was more concrete and offered specific examples that allow more people to understand what this is about.

I'm open to being convinced otherwise, though :)

pages/common/cmake.md
waldyrious: Sorry to interject late in the discussion, but here I would apply tldr-pages' principle of non-exhaustiveness: we should prioritize clarity and conciseness, rather than attempting to cover all bases (if that makes things too verbose, which I think it does in this case).

In this case, I think it would be acceptable to simply omitting what the default prefix is, especially since this is an example of passing a custom prefix, so the default one would not be used anyway.

waldyrious:
@bergentroll I haven't used CMake in a while, but last time I actively used it, the following two examples were pretty useful:

> - Generate a Makefile and use it to compile a project in the same directory as the source: 
>
> `cmake && make`
>
> - Generate a Makefile and use it to compile a project in a separate "build" directory (out-of-source build):
>
> `cmake -H. -B {{build}} && make -C {{build}}`

I would like to hear your thoughts about why we should remove them.

bergentroll:
@waldyrious, modern CMake have the `--build` key to wrap a building system. You able to build with make only when your build system is make. This point initially induced me to update the page.

pages/common/cmake.md
bergentroll: That is true. But also [we should follow good practices](https://github.com/tldr-pages/tldr/pull/4378#discussion_r498453046) . And there is good practice to install binaries into `/usr/local/` (not to `/usr/` or `/opt/` or whatever). I have no idea how to show the `--prefix` key, say word about good POSIX prefix and word about Microsoft Windows prefix, and also keep generality of example. I'd prefer to skip the note of defaults and set `--prefix=/usr/local/`.

pages/common/cmake.md
bergentroll: I'd prefer the prev version too. @bl-ue, @waldyrious, my friends, it is for your decision.

pages/common/cmake.md
sbrl: Hrm, this seems to have gotten out of hand. Isn't it as simple as fixing the typo here?
```suggestion
- Install the build artifacts into /usr/local/ stripping debugging symbols:
```

Not sure I understand the confusion here.

bl-ue: Not quite that simple @sbrl, we're also not sure whether or not to end the directory with a slash etc...

Here's what I think, based on all of this feedback:
```suggestion
- Install the build artifacts into the default directories, stripping debug symbols:
```
@sbrl @waldyrious @bergentroll looks OK to you?

pages/common/cmake.md
bl-ue: I say go back to the old version then. 👍🏻

bergentroll: Oh, sorry, I was dissipated, I meant other version. Initial one seems inaccurate for me. CMake does not generate build systems and projects. What about the following statement?
```Markdown
Cross-platform build automation system, that generates recipes for native build systems.
```

pages/common/cmake.md
bergentroll: @bl-ue, so example will be without prefix nevertheless?
```Markdown
`cmake --install {{path/to/build_directory}} --strip`
```

Also "default directory" is not suitable, because the prefix is only prefix, and further paths depend on project setup at the end. If it is difficult concept, we may just purge it. It is good to say about prefixes because they especially useful on maintenment tasks like building packages in chroot environments.

pages/common/cmake.md
bl-ue: Sounds good.  Go for it.

pages/common/cmake.md
bl-ue: @bergentroll there are actually net even 8 examples, so I say:
1. One example showing how to install with the _default_ prefix,
2. Another example showing how to install with a _custom_ prefix.

Probably both should have `--strip` since installing usually happens for non-debug (aka release) builds

The descriptions could be something like:
```
- Install the build artifacts into the default directories, stripping debug symbols:

`...`

- Install the build artifacts into a custom directory:

`...`
```
I wouldn't think it's neccessary to specify "stripping debug symbols" twice as a user would mentally match `--strip` with the "stripping debug symbols" in the first example and see `--strip` in the next example and mentally apply it there.

What do you think?

bergentroll:
I tried to resolve remained conversations. I also append an example to show how to set variables. I kept `{{path/to/some_directory}}` without closing slashes and regular path `/usr/local/` with a such slash.

pages/common/cmake.md
```diff
@@ -1,13 +1,28 @@
...
 > More information: <https://cmake.org/cmake/help/latest/manual/cmake.1.html>.
 
-- Generate a Makefile and use it to compile a project in the same directory as the source:
+- Generate a build recipe in the current directory with `CMakeLists.txt` from a project directory:
 
-`cmake && make`
+`cmake {{path/to/project_directory}}`
 
-- Generate a Makefile and use it to compile a project in a separate "build" directory (out-of-source build):
+- Generate a build recipe, build type setted to `Release` with CMake variable:
```
bl-ue, line 10: The past tense for "set" is "set" ;)
```suggestion
- Generate a build recipe, with build type set to `Release` with CMake variable:
```

pages/common/cmake.md
```diff
@@ -1,13 +1,28 @@
...
 
-`cmake -H. -B {{build}} && make -C {{build}}`
+`cmake {{path/to/project_directory}} -D CMAKE_BUILD_TYPE=Release`
```
bl-ue, line 12: Token syntax on variable parameters, when reasonable, is preferable, even when the description explicitly states the value of such a parameter, as in this case.
```suggestion
`cmake {{path/to/project_directory}} -D {{CMAKE_BUILD_TYPE=Release}}`
```

pages/common/cmake.md
```diff
@@ -1,13 +1,28 @@
...
+
+- Build artifacts with the generated recipe in `build_directory`:
+
+`cmake --build {{path/to/build_directory}}`
+
+- Install the build artifacts into `/usr/local/` and strip debugging symbols:
+
+`cmake --install {{path/to/build_directory}} --strip`
+
+- Install the build artifacts using the custom prefix for paths:
```
bl-ue, line 22:
```suggestion
- Install the build artifacts into a custom directory:
```

pages/common/cmake.md
```diff
@@ -1,13 +1,28 @@
...
+
+`cmake --install {{path/to/build_directory}} --strip --prefix {{path/to/prefix_directory}}`
```
bl-ue, line 24: I don't think "prefix" is necessary to include here.
```suggestion
`cmake --install {{path/to/build_directory}} --strip --prefix {{path/to/directory}}`
```

bl-ue (changes requested):
Looks good @bergentroll.  I have a few more suggestions.  👍🏻

pages/common/cmake.md
bergentroll: Indeed, thx.

pages/common/cmake.md
bergentroll: I'd like to avoid this substitution.
- "Directory" borns suggestion that artifacts will be placed in the directory. But actually usually project hierarhy will be created in the prefix dir.
- Prefix is not just «let install it here», it may cause project not actually be installed in current system.

bl-ue: @bergentroll Okay!  That's great that you've thought about it so much :D
I don't have a strong opinion.

pages/common/cmake.md
```diff
@@ -1,13 +1,28 @@
...
-`cmake && make`
+`cmake {{path/to/project_directory}}`
 
-- Generate a Makefile and use it to compile a project in a separate "build" directory (out-of-source build):
+- Generate a build recipe, with build type set to `Release` with CMake variable:
 
-`cmake -H. -B {{build}} && make -C {{build}}`
+`cmake {{path/to/project_directory}} -D {{CMAKE_BUILD_TYPE=Release}}`
+
+- Build artifacts with the generated recipe in `build_directory`:
```
sbrl, line 14: I think we might be able to be more clear with a change of wording here.
```suggestion
- Use a generated recipe in a given directory to build artifacts:
```

sbrl (commented):
Thanks for the new page! I've left a comment below for you to review.

pages/common/cmake.md
```diff
@@ -1,13 +1,28 @@
...
-> Cross-platform build system generator.
-> It generates Makefiles, Visual Studio projects or others, depending on the target system.
+> Cross-platform build automation system, that generates recipes for native build systems.
 > More information: <https://cmake.org/cmake/help/latest/manual/cmake.1.html>.
 
-- Generate a Makefile and use it to compile a project in the same directory as the source:
+- Generate a build recipe in the current directory with `CMakeLists.txt` from a project directory:
 
-`cmake && make`
+`cmake {{path/to/project_directory}}`
```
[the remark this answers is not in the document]
bergentroll: @sbrl we already shortly discussed this [upper](https://github.com/tldr-pages/tldr/pull/5160#issuecomment-762797603). Yes indeed, `cmake . && make` is not recommended in «best practice» guides.

bergentroll:
@sbrl, there is not actually a new page, just update. Thanks!

bl-ue (approved):

sbrl (approved):
Thanks, @bergentroll!

PR #5162, opened 2021-01-19
scrot: add command to display a countdown before taking a screenshot

Raizo62:


bl-ue (changes requested):
@Raizo62 thank you for the page update!  Could you please change the example to use long options (if possible)?

bl-ue (approved):

owenvoke (approved):
LGTM!

PR #5170, opened 2021-01-22
btrfs-device: add page

sbrl:
Another one for #5103

### DIFF
--- a/pages/linux/btrfs-device.md
+++ b/pages/linux/btrfs-device.md
@@ -1,0 +1,24 @@
+# btrfs device
+
+> Manage devices in a btrfs filesystem.
+> More information: <https://btrfs.wiki.kernel.org/index.php/Manpage/btrfs-device>.
+
+- Add one or more devices to a btrfs filesystem:
+
+`sudo btrfs device add {{path/to/block_device}} [{{...}}] {{path/to/btrfs_filesystem}}`
+
+- Remove a device from a btrfs filesystem:
+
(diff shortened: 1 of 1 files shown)


pages/linux/btrfs-device.md
```diff
@@ -0,0 +1,24 @@
+# btrfs device
+
+> Manage devices in a btrfs filesystem.
+> More information: <https://btrfs.wiki.kernel.org/index.php/Manpage/btrfs-device>.
+
+- Add one or more devices to a btrfs filesystem:
+
+`sudo btrfs device add {{path/to/block_device}} [{{...}}] {{path/to/btrfs_filesystem}}`
```
bl-ue, line 8: @sbrl is this official tldr-pages syntax for an arbitrary number of parameters?
What about something like this:
```suggestion
`sudo btrfs device add {{path/to/block_device1}} {{path/to/block_device2}} {{path/to/btrfs_filesystem}}`
```
?

sbrl: I think I've seen it done somewhere before, but I can't remember.

bl-ue: I like my proposal.  Can we do it?

sbrl: I'd suggest enclosing the 2nd block device in square brackets to indicate it's optional. We have done that elsewhere.

bl-ue: I've updated my comment.

bl-ue (approved):

navarroaxel (approved):

PR #5185, opened 2021-01-25
README: add link to PDF version

### LINKED ISSUE #4975: List the pdf version on the README.md
I don't think the PDF version is actually listed on the main [README.md](https://github.com/tldr-pages/tldr/blob/master/README.md). It would be good to have it there.

older-pack:
Added a reference to the PDF version at the start of the "How do I use it" section

Fixes #4975

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,8 +46,8 @@ This repository is just that: an ever-growing collection of examples
 for the most common UNIX, Linux, macOS, SunOS and Windows command-line tools.
 
 ## How do I use it?
-
-A popular and convenient way to access these pages on your computer
+If you just want to browse without installing anything, check out the [PDF version.](https://github.com/tldr-pages/tldr-pages.github.io/blob/master/assets/tldr-book.pdf)
+However, a popular and convenient way to access these pages on your computer
 is to install the [Node.js client](https://github.com/tldr-pages/tldr-node-client),
 which is supported by the tldr-pages project maintainers:
 
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ for the most common UNIX, Linux, macOS, SunOS and Windows command-line tools.
 
 ## How do I use it?
 
-If you just want to browse without installing anything, check out the [PDF version.](https://github.com/tldr-pages/tldr-pages.github.io/blob/master/assets/tldr-book.pdf)
+If you just want to browse without installing anything, check out the [PDF version](https://tldr.sh/assets/tldr-book.pdf).
 However, a popular and convenient way to access these pages on your computer
 is to install the [Node.js client](https://github.com/tldr-pages/tldr-node-client),
 which is supported by the tldr-pages project maintainers:
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ This repository is just that: an ever-growing collection of examples
 for the most common UNIX, Linux, macOS, SunOS and Windows command-line tools.
 
 ## How do I use it?
+
 If you just want to browse without installing anything, check out the [PDF version.](https://github.com/tldr-pages/tldr-pages.github.io/blob/master/assets/tldr-book.pdf)
 However, a popular and convenient way to access these pages on your computer
 is to install the [Node.js client](https://github.com/tldr-pages/tldr-node-client),
--- a/README.md
+++ b/README.md
@@ -47,8 +47,7 @@ for the most common UNIX, Linux, macOS, SunOS and Windows command-line tools.
 
 ## How do I use it?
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,8 +46,8 @@ This repository is just that: an ever-growing collection of examples
 for the most common UNIX, Linux, macOS, SunOS and Windows command-line tools.
 
 ## How do I use it?
-
-A popular and convenient way to access these pages on your computer
+If you just want to browse without installing anything, check out the [PDF version.](https://github.com/tldr-pages/tldr-pages.github.io/blob/master/assets/tldr-book.pdf)
```
bl-ue, line 49: It looks like you removed the newline that usually comes after headers...

```suggestion

If you just want to browse without installing anything, check out the [PDF version.](https://github.com/tldr-pages/tldr-pages.github.io/blob/master/assets/tldr-book.pdf)
```

bl-ue (changes requested):

README.md
bl-ue: @older-pack to apply that suggestion:
1. Click "Commit suggestion"
2. Click "Commit changes"

older-pack: Sorry about that, I'm in over my head! This is my first git pull request.

bl-ue (approved):
Thank you @older-pack!!

README.md
```diff
@@ -47,7 +47,8 @@ for the most common UNIX, Linux, macOS, SunOS and Windows command-line tools.
 
 ## How do I use it?
 
-A popular and convenient way to access these pages on your computer
+If you just want to browse without installing anything, check out the [PDF version.](https://github.com/tldr-pages/tldr-pages.github.io/blob/master/assets/tldr-book.pdf)
```
navarroaxel, line 50: Why not the `github.io` deployed URL:
```suggestion
If you just want to browse without installing anything, check out the [PDF version](https://tldr.sh/assets/tldr-book.pdf).
```

navarroaxel (approved):

schneiderl (approved):

sbrl (commented):
I'm wary of presenting the PDF version as the _first_ option under "How do I use it", as it implies this is the preferred way of using tldr pages - when in reality I would think we prefer people to use the reference Node.js client. Would it be possible to swap the two around, to mention the reference client first, followed by the PDF version, and then the community clients list?

Alternatively, we could just add it to the list of clients?

bl-ue:
I'd prefer switching them 👍🏻

older-pack:
If you'd prefer to add it to the list of clients instead that's fine by me.

I didn't intend it to be the _first_ option, I just meant it as an easy way to have a quick look for the curious.
It was easier to just drop the sentence in front of the existing paragraph. But it really doesn't matter where it's listed.

Either way suits me, switch or just add to the list of clients (or if you don't want it on the README at all that's fine too).

bl-ue:
I say switch them.  That separates the node client from the rest, making it clear that it's the preferred method.

In case you're unsure how to edit README.md in this PR, you can click on your branch:

Screen Shot 2021-02-01 at 1 43 29 PM

and then click the pencil icon on he README, or you can simply click [here](https://github.com/older-pack/tldr/edit/patch-1/README.md).

older-pack:
Sorry guys, I'm clueless here :grimacing: I've zero experience editing via git. I presumed it was similar to wiki editing where anyone can make edits on top of edits. By that logic I thought sbrl would execute his idea on top of mine. I'm getting the vibe that I'm meant to implement sbrl's proposal?

bl-ue (changes requested):
@older-pack it is very similar to the wiki editing.  You can make edits on top of edits.

See my comment https://github.com/tldr-pages/tldr/pull/5185#issuecomment-771074180.  Please switch the two (the pdf and the node.js client) around.  To edit the file as it is in this PR, see [my comment](https://github.com/tldr-pages/tldr/pull/5185#issuecomment-771074180).  In it I provide two ways to edit the file, the first was is slightly more involved but will work for any PR on any repo, the second way is much simpler (just clicking a link.)

bl-ue:
@older-pack to save you the trouble and because I thought of something else, I went ahead and made the switch for you :)

bl-ue (approved):

older-pack:
Sorry, I got back too late. The job's done. Thanks for making the edit. I'll stay away from making pull requests in future, I've no idea what's going on.

bl-ue:
No problem @older-pack! You can always take a look at the [GitHub docs for PRs](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests) in case you want to ;)

PR #5206, opened 2021-01-30
scripts/pdf/render.py: fix typo

bl-ue:
`tdlr-pages-to-PDF` → `tldr-pages-to-pdf` 😄

### DIFF
--- a/scripts/pdf/render.py
+++ b/scripts/pdf/render.py
@@ -117,7 +117,7 @@ def main(loc, colorscheme):
 if __name__ == "__main__":
 
     # Parsing the arguments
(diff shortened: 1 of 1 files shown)


bl-ue:
When mergining, please correct the commit message if necessary, I forgot to change it before I commited...

bl-ue:
I plan on reworking that script sometime.  I'm a Python expert and that file could use some more cleanup 😉

navarroaxel (approved):

sbrl (approved):
Thanks!

Yeah, some rcleanup would be appreciated. I don't think  was the one who wrote it (but if I did I don't remember), but my Python is rather rusty due to a preference for bracketed languages like JS and C# - so some assistance here would definitely be welcome.

PR #5193, opened 2021-01-27
gpg*: add Chinese translation

sandylaw:
Description: add gpg pages in pages.zh/common.

- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [ ] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

@ChungZH @KsRyY @whriedplanck

PR #5214, opened 2021-01-31
multiple pages: ensure exactly one colon at the end of example descriptions

bl-ue:
Some example descriptions had no colons (or had periods) and some, but fewer, had two colons.

I didn't touch the Chinese or Japanese pages because I've learned that there is different punctuation for those languages (#5193) and I'll open a separate PR for them.


bl-ue:
Funny. Alphabetically sorted, the first changed line has two spaces after the `-`.  I'll fix that in a new PR, but since it will be modifying a line that this PR has modified, that will cause conflicts, so I'll wait until this PR is merged.

pages.es/common/chmod.md
```diff
@@ -2,7 +2,7 @@
 
 > Cambiar los permisos de acceso de un archivo o directorio.
 
-- Otorga al [u]suario que es propietario del archivo a ejecutarlo (x).
+- Otorga al [u]suario que es propietario del archivo a ejecutarlo (x):
```
navarroaxel, line 5:
```suggestion
- Otorga al [u]suario que es propietario del archivo permiso para [x] ejecutarlo:
```

pages.fr/common/apropos.md
```diff
@@ -10,6 +10,6 @@
 
 `apropos -l {{expression_reguliere}}`
 
-- Recherche les pages qui contiennent toutes les expressions données (fonction ET)
+- Recherche les pages qui contiennent toutes les expressions données (fonction ET):
```
navarroaxel, line 13: French needs an space:
```suggestion
- Recherche les pages qui contiennent toutes les expressions données (fonction ET) :
```

bl-ue: Oops, I forgot! You're right!  But is that the thin unbreakable space?

navarroaxel: oh nop.

bl-ue: It's this.  I'll edit and apply your suggestion.

navarroaxel: Obviously the French contributors don't use that char. We should add that to the linter for the French directory.

navarroaxel (approved):

pages.fr/common/apropos.md
bl-ue: Some do, some don't.  You mean we're able to add non-English language-specific rules to the linter?

navarroaxel: sounds like a conditional rule based on the parent (of the parent) directory `pages.fr`. I think is possible to add that.

bl-ue: Hmm. That's really cool.  But this discussions should get it's own issue because if that's true, we really ought to extend some of the syntactic analysis that the linter can do to some of the other languages, because the punctuation of langs like Spanish/German are very similar to English...

bl-ue: I'm currently preparing a PR for normalizing the Chinese punctuation.  They have their own periods, commans, parentheses, semicolons, etc., and their own rules when mixing Latin characters and Chinese characters, etc., and it would be super great for that to be automated.  There's really no need for such tedious tasks to be done over and over again by a human ;)

pages.de/common/borg.md
```diff
@@ -12,7 +12,7 @@
 
 `borg create --progress {{/pfad/zum/repo_verzeichnis}}::{{Montag}} {{/pfad/zum/quell_verzeichnis}}`
 
--  Alle Archive in einem Repository auflisten
+-  Alle Archive in einem Repository auflisten:
```
sbrl, line 15: Double space
```suggestion
- Alle Archive in einem Repository auflisten:
```

sbrl (commented):
Wow, awesome work @bl-ue!

Just found one teensy thing we can also fix, and then this is ready to go!

pages.de/common/borg.md
bl-ue: @sbrl I noticed that as well - please see my comment above https://github.com/tldr-pages/tldr/pull/5214#issuecomment-770426912

schneiderl:
Given [this comment ](https://github.com/tldr-pages/tldr/pull/5214#issuecomment-770426912) I believe we can merge this PR.

schneiderl (approved):

PR #5215, opened 2021-02-01
genfstab: add page

peterbabic:

### DIFF
--- a/pages/linux/genfstab.md
+++ b/pages/linux/genfstab.md
@@ -1,7 +1,7 @@
 # genfstab
 
 > Generate output suitable for addition to an fstab file.
-> More information: <https://man.archlinux.org/man/extra/arch-install-scripts/genfstab.8.en>
+> More information: <https://man.archlinux.org/man/extra/arch-install-scripts/genfstab.8.en>.
 
 - Generate the output based on LABEL
 
--- a/pages/linux/genfstab.md
+++ b/pages/linux/genfstab.md
@@ -3,7 +3,7 @@
 > Generate output suitable for addition to an fstab file.
 > More information: <https://man.archlinux.org/man/extra/arch-install-scripts/genfstab.8>.
 
-- Generate the output based on LABEL:
+- Generate the output based on a label:
 
 `genfstab -L {{path/to/mount_point}}`
 
@@ -11,7 +11,7 @@
 
 `genfstab -U {{path/to/mount_point}}`
 
-- Generate a fstab file, requires being a superuser:
+- Generate an fstab file, requires being a superuser:
 
 `genfstab -U {{/mnt}} >> {{/mnt/etc/fstab}}`
 
--- a/pages/linux/genfstab.md
+++ b/pages/linux/genfstab.md
@@ -3,18 +3,18 @@
 > Generate output suitable for addition to an fstab file.
 > More information: <https://man.archlinux.org/man/extra/arch-install-scripts/genfstab.8>.
 
-- Generate the output based on a label:
+- Display an fstab compatible output based on a volume label:
 
 `genfstab -L {{path/to/mount_point}}`
 
-- Display `fstab` output using partitions' UUID:
+- Display an fstab compatible output based on a volume UUID:
 
 `genfstab -U {{path/to/mount_point}}`
 
-- Generate an fstab file, requires being a superuser:
+- A usual way to generate an fstab file, requires root permissions:
 
 `genfstab -U {{/mnt}} >> {{/mnt/etc/fstab}}`
 
-- Append a filesystem into fstab file to mount it automatically:
+- Append a volume into an fstab file to mount it automatically:
 
 `genfstab -U {{path/to/mount_point}} | sudo tee -a /etc/fstab`
--- a/pages/linux/genfstab.md
+++ b/pages/linux/genfstab.md
@@ -1,20 +1,20 @@
 # genfstab
 
 > Generate output suitable for addition to an fstab file.
-> More information: <https://man.archlinux.org/man/extra/arch-install-scripts/genfstab.8.en>.
+> More information: <https://man.archlinux.org/man/extra/arch-install-scripts/genfstab.8>.
 
 - Generate the output based on LABEL:
 
-`genfstab -L MOUNTPOINT`
+`genfstab -L {{path/to/mount_point}}`
 
-- Generate the output based on UUID:
+- Display `fstab` output using partitions' UUID:
 
-`genfstab -U MOUNTPOINT`
+`genfstab -U {{path/to/mount_point}}`
 
 - Generate a fstab file, requires being a superuser:
 
-`genfstab -U /mnt >> /mnt/etc/fstab`
+`genfstab -U {{/mnt}} >> {{/mnt/etc/fstab}}`
 
 - Append a filesystem into fstab file to mount it automatically:
 
-`gengstab -U MOUNTPOINT | sudo tee -a /etc/fstab`
+`genfstab -U {{path/to/mount_point}} | sudo tee -a /etc/fstab`
--- a/pages/linux/genfstab.md
+++ b/pages/linux/genfstab.md
@@ -1,0 +1,20 @@
+# genfstab
+
+> Generate output suitable for addition to an fstab file.
+> More information: <https://man.archlinux.org/man/extra/arch-install-scripts/genfstab.8.en>
+
+- Generate the output based on LABEL
+
+`genfstab -L MOUNTPOINT`
+
+- Generate the output based on UUID
+
+`genfstab -U MOUNTPOINT`
+
+- Generate a fstab file, requires being a superuser
+
+`genfstab -U /mnt >> /mnt/etc/fstab`
+
+- Append a filesystem into fstab file to mount it automatically
+
+`gengstab -U MOUNTPOINT | sudo tee -a /etc/fstab`
--- a/pages/linux/genfstab.md
+++ b/pages/linux/genfstab.md
@@ -3,18 +3,18 @@
 > Generate output suitable for addition to an fstab file.
 > More information: <https://man.archlinux.org/man/extra/arch-install-scripts/genfstab.8.en>.
 
-- Generate the output based on LABEL
+- Generate the output based on LABEL:
 
 `genfstab -L MOUNTPOINT`
 
-- Generate the output based on UUID
+- Generate the output based on UUID:
 
 `genfstab -U MOUNTPOINT`
 
-- Generate a fstab file, requires being a superuser
+- Generate a fstab file, requires being a superuser:
 
 `genfstab -U /mnt >> /mnt/etc/fstab`
 
-- Append a filesystem into fstab file to mount it automatically
+- Append a filesystem into fstab file to mount it automatically:
 
 `gengstab -U MOUNTPOINT | sudo tee -a /etc/fstab`


pages/linux/genfstab.md
```diff
@@ -0,0 +1,20 @@
+# genfstab
+
+> Generate output suitable for addition to an fstab file.
+> More information: <https://man.archlinux.org/man/extra/arch-install-scripts/genfstab.8.en>.
```
navarroaxel, line 4:
```suggestion
> More information: <https://man.archlinux.org/man/extra/arch-install-scripts/genfstab.8>.
```

pages/linux/genfstab.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Generate the output based on LABEL:
+
+`genfstab -L MOUNTPOINT`
+
+- Generate the output based on UUID:
```
navarroaxel, line 10: Maybe this is better:
```suggestion
- Display `fstab` output using partitions' UUID:
```

pages/linux/genfstab.md
```diff
@@ -0,0 +1,20 @@
...
+
+`genfstab -U MOUNTPOINT`
```
navarroaxel, line 12:
```suggestion
`genfstab -U {{path/to/mount_point}}`
```

pages/linux/genfstab.md
```diff
@@ -0,0 +1,20 @@
+# genfstab
+
+> Generate output suitable for addition to an fstab file.
+> More information: <https://man.archlinux.org/man/extra/arch-install-scripts/genfstab.8.en>.
+
+- Generate the output based on LABEL:
+
+`genfstab -L MOUNTPOINT`
```
navarroaxel, line 8:
```suggestion
`genfstab -L {{path/to/mount_point}}`
```

pages/linux/genfstab.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Generate the output based on UUID:
+
+`genfstab -U MOUNTPOINT`
+
+- Generate a fstab file, requires being a superuser:
+
+`genfstab -U /mnt >> /mnt/etc/fstab`
```
navarroaxel, line 16:
```suggestion
`genfstab -U {{/mnt}} >> {{/mnt/etc/fstab}}`
```

pages/linux/genfstab.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Append a filesystem into fstab file to mount it automatically:
+
+`gengstab -U MOUNTPOINT | sudo tee -a /etc/fstab`
```
navarroaxel, line 20:
```suggestion
`genfstab -U {{path/to/mount_point}} | sudo tee -a /etc/fstab`
```

peterbabic (commented):
add maintainer suggestions

**EDIT:** do not know what to do, still reluctantly learning Github 
**EDIT2:** somehow aded maintainer suggestions into "batch" and commited them, hopefully this was expected

pages/linux/genfstab.md
```diff
@@ -0,0 +1,20 @@
+# genfstab
+
+> Generate output suitable for addition to an fstab file.
+> More information: <https://man.archlinux.org/man/extra/arch-install-scripts/genfstab.8>.
+
+- Generate the output based on LABEL:
```
bl-ue, line 6: This is a little confusing&mdash;where does LABEL come from?

pages/linux/genfstab.md
```diff
@@ -0,0 +1,20 @@
...
+
+`genfstab -L {{path/to/mount_point}}`
+
+- Display `fstab` output using partitions' UUID:
+
+`genfstab -U {{path/to/mount_point}}`
+
+- Generate a fstab file, requires being a superuser:
```
bl-ue, line 14: Instead of giving superuser notice, you can just add `sudo` to the command (see additional suggestion below):
```suggestion
- Generate an fstab file:
```

pages/linux/genfstab.md
```diff
@@ -0,0 +1,20 @@
...
+
+`genfstab -U {{/mnt}} >> {{/mnt/etc/fstab}}`
```
bl-ue, line 16: See my suggestion above.
```suggestion
`sudo genfstab -U {{/mnt}} >> {{/mnt/etc/fstab}}`
```

bl-ue (changes requested):
Hi @peterbabic, thank you for the new page! I've left some comments below for you to review.

pages/linux/genfstab.md
peterbabic: This suggestion will fail, for sudo there is the line just below.
Thiis line in specifically used when superuser and it is the 99% of it's use cases, I would definitely not edit it in this way.

pages/linux/genfstab.md
peterbabic: An fstab? I do not understand the edit.

pages/linux/genfstab.md
peterbabic: The LABEL is a filesystem volume identifier, usually assigned (but not limited to) during the volume creation. More information: https://wiki.archlinux.org/index.php/Persistent_block_device_naming#by-label

pages/linux/genfstab.md
bl-ue: "an fstab file".  I assume "fstab" is generally pronounced "ef-es-tab"&mdash;@navarroaxel am I correct about this?

pages/linux/genfstab.md
bl-ue: By "as superuser", do you mean after having run something like `su`?

bl-ue: If it's required to run the command as superuser as you say, why not prefix it with `sudo`?  We do that in many other pages, particularly in low-level file-system utilities, such as [`partx.md`](https://github.com/tldr-pages/tldr/blob/master/pages/linux/partx.md).

peterbabic:
Unfortunately I do not agree wit any of the changes @bl-ue suggested, not meaning to be rude by any way. I have provided explanations to all of them.

pages/linux/genfstab.md
bl-ue: Then can you update either the description or the command, because it is not clear what "LABEL" has to do with the example command itself.

Maybe something like this?
```suggestion
- Generate the output based on a label:
```
and then change the example from `path/to/mount_point` to `path/to/label`?

I don't use Arch Linux, so feel free to correct me!

pages/linux/genfstab.md
peterbabic: To your First question, yes, for instance `su` is necessary to elevate privileges.

To your second question, the command will fail because of the redirection operator `>>`. Prefixing the whole line with with `sudo` will not work. To use redirection operator with elevated privileges, a different approach is necessary.

https://linuxize.com/post/bash-append-to-file/#append-to-a-file-using-the-redirection-operator-
https://www.cyberciti.biz/faq/sudo-append-data-text-to-file-on-linux-unix-macos/

bl-ue:
@peterbabic no problem! Sometimes we have difficulty communicating through a purely textual medium :)  I've responded to your comments.

pages/linux/genfstab.md
bl-ue: Ah, I see!  Makes perfect sense now.  Please disregard this suggestion.

pages/linux/genfstab.md
bl-ue: It's better English to use "an" before a word like "fstab", the way it's generally pronounced.  Now of course, people have different pronunciations but I believe "an" would be a better fit here.
```suggestion
- Generate an fstab file, requires being a superuser:
```
Please also disregard the first suggestion, as I understand the superuser situation now :)

pages/linux/genfstab.md
```diff
@@ -0,0 +1,20 @@
...
+
+`genfstab -L {{path/to/mount_point}}`
+
+- Display `fstab` output using partitions' UUID:
+
+`genfstab -U {{path/to/mount_point}}`
+
+- Generate an fstab file, requires being a superuser:
+
+`genfstab -U {{/mnt}} >> {{/mnt/etc/fstab}}`
```
bl-ue, line 16: How about this:
```suggestion
`genfstab -U {{path/to/mount_point}} >> {{/mnt/etc/fstab}}`
```

peterbabic:
@bl-ue no poblem as well! Thanks for suggestions. Did not know about the nature of "an ef-es-tab file" 😄

pages/linux/genfstab.md
peterbabic: I would rather change the description to: 

- Most usual way to generate an fstab file, requires being a superuser:

And left the code the way it is. What do you think?

bl-ue: 1. IMO, changing the description isn't necessary. I think it's good already 👍🏻 
2. `/mnt` is inconsistent it seems with the rest, that may be intentional (I'm no Arch user :smile:)

bl-ue: Actually, I think it would be better to make this example more consistent with the last one, which would solve the superuser problem, by using `sudo tee`:
```suggestion
`genfstab -U {{path/to/mount_point}} | sudo tee -a {{/mnt/etc/fstab}}`
```

bl-ue: If you're sure that `/mnt` is better, even though it's inconsistent, let it be that way. I trust you know better :)

navarroaxel: @bl-ue in this example @peterbabic tried to explain the most common use case for this command as a step in the Arch Linux installation https://wiki.archlinux.org/index.php/installation_guide#Fstab. Nobody uses `sudo tee` 😄 because the installer is a live ISO and you run it with `root` user, command by command.

peterbabic: Look, we can even discard the first and the second example. The third and the fourth are the only ones that I have personally ever needed and will serve for most people.

navarroaxel: Yeah, but the first 2 examples are useful to explain how the command works and some arguments.

peterbabic: Yes exactly, that is why I have added there. I was following the contribution guidelines.

bl-ue: Okay guys, great!  @navarroaxel cool that you're an Arch Linux guy.  That's good to know.

> Nobody uses `sudo tee` 😄 because the installer is a live ISO and you run it with root user, command by command.

😄😄😄😄😄

pages/linux/genfstab.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Generate the output based on a label:
+
+`genfstab -L {{path/to/mount_point}}`
+
+- Display `fstab` output using partitions' UUID:
+
+`genfstab -U {{path/to/mount_point}}`
+
+- Generate an fstab file, requires being a superuser:
```
bl-ue, line 14: Perhaps this will read better?
```suggestion
- Generate an fstab file, requires root permissions:
```

pages/linux/genfstab.md
```diff
@@ -0,0 +1,20 @@
+# genfstab
+
+> Generate output suitable for addition to an fstab file.
+> More information: <https://man.archlinux.org/man/extra/arch-install-scripts/genfstab.8>.
+
+- Generate the output based on a label:
```
bl-ue, line 6: Perhaps this will read better?
```suggestion
- Generate output based on a label:
```

bl-ue (changes requested):
Okay, thanks for the help @navarroaxel.  @peterbabic I have only two more comments.

peterbabic:
I have added another commit where I believe the changes are the most consistent. I have also removed backticks around `fstab` as it is word, not a code nor a path.

peterbabic:
Maybe the last thing I would check is if *volume* is the right word, as *filesystem*, *block device*, *disk*, *partition* or *disk partition* would all suit as well depending on use case.

 I would keep *volume* there.

image

bl-ue:
@peterbabic - Looks good to me, except the sentence "A usual way to generate an fstab file, requires root permissions:" is inconsistent with the rest of our pages.  If that is the most common example, then I think you should:
1. Move the example to the top, so it's what the user sees first, and
2. Change the example description to "Generate an fstab file (requires root permissions):"

pages/linux/genfstab.md
sbrl: Yeah, that's correct grammar there @bl-ue.

pages/linux/genfstab.md
sbrl: Hrm, I see the problem here. I would be inclined to go with @peterbabic's suggestion here, with something like this:

```suggestion
`genfstab -U {{path/to/mount_point}} | sudo tee -a /etc/fstab`
```

The reason for this is that by convention in other pages, we use `sudo` to indicate that a given command needs to be run as root - so I'd prefer to use that same convention here too if possible.

Also, welcome to GitHub and tldr-pages, @peterbabic!

sbrl (approved):
Looks pretty much good to go to me - I think we can probably merge this.

I've left a comment on the `requires root permissions` issue, but that doesn't necessarily need to block the merge of this PR - and I can see some significant discussion has already happened - so I'm leaning on the side of getting this merged sooner rather than later.

navarroaxel (approved):

pages/linux/genfstab.md
bl-ue: @sbrl this has already been resolved on another long thread on this same line :) - https://github.com/tldr-pages/tldr/pull/5215#discussion_r567993244

bl-ue:
@sbrl that was actually already resolved :) - https://github.com/tldr-pages/tldr/pull/5215#discussion_r569025992

bl-ue (approved):
Thank you @peterbabic, and welcome to tldr-pages!

sbrl:
Ok, whatever was done there is fine then :-)

bl-ue:
We picked not to use `sudo` because that command is normally run in a boot environment where `tee` may not be available.

PR #5217, opened 2021-02-02
rankmirrors: add page

sbrl:
Note that I've used a concrete example of a filepath here, because that's where the mirrorlist is actually located (so it saves some effort on the part of the reader in locating their mirrorlist file).

### DIFF
--- a/pages/linux/rankmirrors.md
+++ b/pages/linux/rankmirrors.md
@@ -2,6 +2,7 @@
 
 > Rank a list of Pacman mirrors by connection and opening speed.
 > Writes the new mirrorlist to stdout.
+> More information: <https://wiki.archlinux.org/index.php/mirrors>.
 
 - Rank a mirror list:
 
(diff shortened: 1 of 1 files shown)


bl-ue:
>Note that I've used a concrete example of a filepath here, because that's where the mirrorlist is actually located (so it saves some effort on the part of the reader in locating their mirrorlist file).

I actually that's best to do whenever at all possible.

bl-ue:
@sbrl do you use Arch Linux? :)

bl-ue:
I'm thinking this for an information link: https://wiki.archlinux.org/index.php/mirrors
What do you think?

sbrl:
Lol, sort of! I have an [Artix Linux](https://artixlinux.org/) laptop as my travel device. This command is rather useful, as I've encountered a number of slow mirror situations recently.

That link doesn't isn't _specifically_ about `rankmirrors`, but I think it's probably a good one to include.

navarroaxel (approved):

bl-ue (approved):

PR #5218, opened 2021-02-02
grep: refresh

bl-ue:
Inspired by #5216

I removed the stdin example so I could add the `-o` example, which I personally find very useful but less obvious.

### DIFF
--- a/pages/common/grep.md
+++ b/pages/common/grep.md
@@ -30,7 +30,7 @@
 
 - Search for lines matching a pattern, printing [o]nly the matched text:
 
-`grep -o {{search_pattern}}`
+`grep -o {{search_pattern}} {{path/to/file}}`
 
 - Search for lines that do not match (in[v]ert) a pattern:
 
--- a/pages/common/grep.md
+++ b/pages/common/grep.md
@@ -6,7 +6,7 @@
 
 - Search for a pattern within a file:
 
-`grep {{search_pattern}} {{path/to/file}}`
+`grep "{{search_pattern}}" {{path/to/file}}`
 
 - Search for an exact ([F]ixed) string:
 
@@ -14,24 +14,24 @@
 
 - Search for a pattern [R]ecursively in the current directory, showing matching line [n]umbers, [I]gnoring non-text files:
 
-`grep -RIn {{search_pattern}} {{.}}`
+`grep -RIn "{{search_pattern}}" {{.}}`
 
 - Use [E]xtended regular expressions (supporting `?`, `+`, `{}`, `()` and `|`), in case-[i]nsensitive mode:
 
-`grep -Ei {{search_pattern}} {{path/to/file}}`
+`grep -Ei "{{search_pattern}}" {{path/to/file}}`
 
 - Print 3 lines of [C]ontext around, [B]efore, or [A]fter each match:
 
-`grep -{{C|B|A}} {{3}} {{search_pattern}} {{path/to/file}}`
+`grep -{{C|B|A}} {{3}} "{{search_pattern}}" {{path/to/file}}`
 
 - Print file name [H]eaders with the corresponding line [n]umber for each match:
 
-`grep -Hn {{search_pattern}} {{path/to/file}}`
+`grep -Hn "{{search_pattern}}" {{path/to/file}}`
 
 - Search for lines matching a pattern, printing [o]nly the matched text:
 
-`grep -o {{search_pattern}} {{path/to/file}}`
+`grep -o "{{search_pattern}}" {{path/to/file}}`
 
 - Search a file for lines that do not match (in[v]ert) a pattern:
 
-`cat {{path/to/file}} | grep -v {{search_pattern}}`
+`cat {{path/to/file}} | grep -v "{{search_pattern}}`"
--- a/pages/common/grep.md
+++ b/pages/common/grep.md
@@ -10,7 +10,7 @@
 
 - Search for an exact ([F]ixed) string:
 
-`grep -F {{exact_string}} {{path/to/file}}`
+`grep -F "{{exact_string}}" {{path/to/file}}`
 
 - Search for a pattern [R]ecursively in the current directory, showing matching line [n]umbers, [I]gnoring non-text files:
 
@@ -34,4 +34,4 @@
 
 - Search a file for lines that do not match (in[v]ert) a pattern:
 
-`cat {{path/to/file}} | grep -v "{{search_pattern}}`"
+`cat {{path/to/file}} | grep -v "{{search_pattern}}"`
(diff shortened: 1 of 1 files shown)


pages/common/grep.md
```diff
@@ -2,35 +2,36 @@
...
-- Print file name with the corresponding line number for each match:
+- Print file name [H]eaders with the corresponding line [n]umber for each match:
 
 `grep -Hn {{search_pattern}} {{path/to/file}}`
 
-- Use the standard input instead of a file:
+- Search for lines matching a pattern, printing [o]nly the matched text:
 
-`cat {{path/to/file}} | grep {{search_pattern}}`
+`grep -o {{search_pattern}}`
```
navarroaxel, line 33: what about the `stdin`?

navarroaxel (changes requested):

pages/common/grep.md
bl-ue: Well, I know that's useful, but I decided it was more of a general shell concept, and that since `-o` was useful and **not** as obvious, it would be better to have it instead.

bl-ue: If you think a different example would be better to replace, I'm all for doing that.

navarroaxel: but this examples an input like {{path/to/file}} or the stdin

pages/common/grep.md
```diff
@@ -2,35 +2,36 @@
...
+> More information: <https://man7.org/linux/man-pages/man1/grep.1.html>.
 
 - Search for a pattern within a file:
 
 `grep {{search_pattern}} {{path/to/file}}`
 
-- Search for an exact string:
+- Search for an exact ([F]ixed) string:
 
 `grep -F {{exact_string}} {{path/to/file}}`
```
navarroaxel, line 13: Why you don't used the long format? `--fixed-strings`

pages/common/grep.md
bl-ue: I'm not sure what you mean&mdash;will you please clarify?

navarroaxel: If I run `$ grep -o text` the grep program waits for an input from `stdin` because I didn't specify any input file or files.

pages/common/grep.md
navarroaxel: Maybe it's a nonsense split this page for Linux and macOS. It's fine as is.

pages/common/grep.md
bl-ue: Ah I see, you think I should add an example for that?

bl-ue: Oh oops!!! I see now!  I didn't specify a file in this example, so it's defaulting to stdin, that's what you're noticing?

navarroaxel (approved):

sbrl (commented):
Hrm. Lots of great changes here! I'm on the fence about the `-o` example (as demonstrating stdin functionality I feel is important, and I'm wary of making the page slightly more complex), but all the other changes I'm on board with. Let's see what others think.

owenvoke (approved):

einverne (approved):

sbrl (commented):
Hrm, I'm on the fence about the change in example here - although the other changes are great.

On the one hand, generally speaking we try to avoid demonstrating shell syntax like `cat | command`. On the other, under the edits it's not until the last example that we let the user know that grep supports piping in via stdin at all - and even then indirectly.

bl-ue:
@sbrl can we merge this? I'm inclined to say that the `-o` option is great deal less obivous than the stdin grepping.  Anyone else have a say? @tldr-pages?

sbrl:
What I'm trying to get at here @bl-ue is that I'm not sure including non-obvious / more obscure uses us actually useful here. Particularly in a page like `grep`, I feel it's more important to demonstrate basic usage than to include non-obvious use cases. This will make the page more beginner friendly.

Having thought it over, I'm against including `-o` if it means removing the stdin example.

waldyrious:
> it's not until the last example that we let the user know that grep supports piping in via stdin at all

I would assume that most command-line tools (especially the classic ones, and those aimed specifically at dealing with text, which `grep` is both) would follow the usual Unix patterns of supporting piping.

Conversely, I find the `-o` option to be quite useful, especially when using regex over a file that contains long lines. I would agree with @bl-ue on this one that `-o` brings more utility to this page than the piping example.

pages/common/grep.md
```diff
@@ -2,35 +2,36 @@
...
-- Use the standard input instead of a file:
+- Search for lines matching a pattern, printing [o]nly the matched text:
 
-`cat {{path/to/file}} | grep {{search_pattern}}`
+`grep -o {{search_pattern}} {{path/to/file}}`
 
-- In[v]ert match for excluding specific strings:
+- Search for lines that do not match (in[v]ert) a pattern:
 
 `grep -v {{search_pattern}}`
```
waldyrious, line 37: Shall we add piping explicitly to this example? The way this is "worded", the command just hangs waiting for input. That would also allow us to demonstrate the piping behavior that is being removed above.

bl-ue, line 37:
```suggestion
- Search a file for lines that do not match (in[v]ert) a pattern:

`cat {{path/to/file}} | grep -v {{search_pattern}}`
```

bl-ue: 👍🏻  https://github.com/tldr-pages/tldr/pull/5218/files#r594572492

bl-ue:
I've figured out that I can use long options, actually.  I've identified a few discrepancies/mistakes in the BSD vs. GNU man pages, but I tested the commands and they work the same 👍🏻

sbrl (approved):
Thanks for all the hard work here, @bl-ue!

And wow, finding a mistake in the `grep` man page isn't something that happens every day! Might be worth raising an issue with the authors of the man pages, though actually finding the best place to do that would probably be an adventure in and of itself haha :P

marchersimon (approved):

bl-ue:
:smile: @sbrl the error is a funny one: GNU & BSD `grep` both accept the `--binary-files=...` option, and the GNU (Linux) man page documents it as `--binary-files`, but the BSD (macOS) page describes it as `--binary-file` (without the `s`).  However, truncating the long options so long as the truncated version could only be expanded to one option. For example, the following all expand to `--binary-files`:
```
--binary-files
--binary-file
--binary-fil
--binary-fi
--binary-f
--binary-
```
It can't get any shorter because there's another argument: `-U, --binary`.  As soon as you try something even shorter like `--binar` it gets confused
```
$ grep --binar
grep: option `--binar' is ambiguous
```

bl-ue:
Finally! 😅

marchersimon:
"There has to be a reason the others don't merge, so I better don't merge either" :smile:

bl-ue:
Exactly!  I still think the same thing sometimes, although I've learned, frequently the others don't merge because they forgot to, which in fact was the case for me (and probably the others too.)

PR #5229, opened 2021-02-03
grep: fix example description for stdin usage

sjmarshall2000:
seems like the original description was backwards in what it was trying to say

### DIFF
--- a/pages/common/grep.md
+++ b/pages/common/grep.md
@@ -27,7 +27,7 @@
 
 `grep -Hn {{search_pattern}} {{path/to/file}}`
 
-- Use the standard input instead of a file:
+- Use file instead of standard input
 
 `cat {{path/to/file}} | grep {{search_pattern}}`
 


navarroaxel:
Hi @sjmarshall2000, we have an edition for this page by @bl-ue under review #5218

bl-ue:
@sjmarshall2000 two things here&mdash;

1. The wording does seem backwards, however, the existing wording was accurate&mdash;the example was demonstrate how to pass data (in this case the output of `cat`) to `grep`'s stdin.  I agree that this wording could be improved however, but

2. in the PR that @navarroaxel referenced (#5218) I removed this example entirely! (see [the lines in question](https://github.com/tldr-pages/tldr/pull/5218/files#diff-359201f8e3abd55285b88cc6c0c85ee34c4eabc70bbdf6f609bf7f7c46f34fb5L30-R33))

bl-ue:
Closing as this PR is no longer useful.

PR #5234, opened 2021-02-05
base32, base64: fix -D flag

luebken:
Gnu Core Utils now uses -d to decode instead of -D https://github.com/coreutils/coreutils/blob/00ea4bacf6063ccc125209d5186f8f2382c6f0d4/src/basenc.c#L76

Man page on a Mac: 
```
BASE64(1)                                               User Commands                                              BASE64(1)

NAME
       base64 - base64 encode/decode data and print to standard output

SYNOPSIS
       base64 [OPTION]... [FILE]

DESCRIPTION
       Base64 encode or decode FILE, or standard input, to standard output.

       With no FILE, or when FILE is -, read standard input.

       Mandatory arguments to long options are mandatory for short options too.

       -d, --decode
              decode data
....
```

TODOs: 
- [x] check all base64 translations.
- [x] check `base32`


navarroaxel:
:thinking: You're using a GNU page for macOS. This is my output in big sure 11.1:
```
% base64 --help
Usage:  base64 [-hvDd] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -Dd, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)
```
I tested the command and `-d` and `-D` looks like synonyms. I recommend to use the long argument format `--decode` to avoid this and make the example more easy to understand.

bl-ue:
Great idea @navarroaxel.  @luebken, please replace `-d` and `-D` with `--decode`, as this will clear any confusion between OS versions of `base64`, and besides, we prefer long options when possible as they are self-explanatory.

Also, @luebken, are you planning on completing those TODOs in the PR's description?  If so, please mark this PR as a draft (Click the "Convert to draft" link under the "Reviewers" section, up at the right.)

pages/osx/base64.md
```diff
@@ -8,12 +8,12 @@
 
 - Decode a file:
 
-`base64 -D -i {{base64_file}}`
+`base64 --decode -i {{base64_file}}`
```
navarroaxel, line 11:
```suggestion
`base64 --decode --input {{base64_file}}`
```

bl-ue:
@luebken while you're at it, will you please change `-i {{file}}` to `--input={{file}}` across all translations?

luebken:
```
$ base64 --input=test.dat
base64: unrecognized option '--input=test.dat'
Try 'base64 --help' for more information.

$ base64 --help
Usage: base64 [OPTION]... [FILE]
Base64 encode or decode FILE, or standard input, to standard output.

With no FILE, or when FILE is -, read standard input.

Mandatory arguments to long options are mandatory for short options too.
  -d, --decode          decode data
  -i, --ignore-garbage  when decoding, ignore non-alphabet characters
  ...  
```
@bl-ue There is no --input option on my OS (macOS 10.14.6).

bl-ue:
Really??!?!  That's crazy. I'm on macOS 11.2, and this is what my base64 looks like:
```
$ base64 --help
Usage:	base64 [-hvDd] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -Dd, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)
```
And this is the manpage:
```
base64(1)                 BSD General Commands Manual                base64(1)

NAME
     base64 -- Encode and decode using Base64 representation

SYNOPSIS
     base64 [-h | -D | -d] [-b count] [-i input_file] [-o output_file]

DESCRIPTION
     base64 encodes and decodes Base64 data, as specified in RFC 4648.  With no options, base64 reads raw data from stdin and writes encoded data as a continuous block to stdout.

OPTIONS
     The following options are available:
     -b count
     --break=count        Insert line breaks every count characters.  Default is 0, which generates an unbroken stream.

     -d
     -D
     --decode             Decode incoming Base64 stream into binary data.

     -h
     --help               Print usage summary and exit.

     -i input_file
     --input=input_file   Read input from input_file.  Default is stdin; passing - also represents stdin.

     -o output_file
     --output=output_file
                          Write output to output_file.  Default is stdout; passing - also represents stdout.

SEE ALSO
     openssl(1), wikipedia page:
           http://en.wikipedia.org/wiki/Base64,
     RFC 4648:
           http://tools.ietf.org/html/rfc4648

Mac OS X 10.7                  February 8, 2011                  Mac OS X 10.7
```
which is from macOS 10.7???

@luebken if it's short, what does your manpage look like?

bl-ue:
Oh, I wonder if you're using [GNU base64](https://www.gnu.org/software/coreutils/manual/html_node/base64-invocation.html) because it has `--ignore-garbage` like you have...

Will you please run `which base64`?  I get `/usr/bin/base64` which is the system-provided executable.

luebken:
Appologies. I didn't check. I got the [Homebrew](https://formulae.brew.sh/formula/coreutils) version: 

```
$ which base64
/usr/local/opt/coreutils/libexec/gnubin/base64
```

bl-ue:
Aha.  So there's two different versions of this page: and [`osx`](https://github.com/tldr-pages/tldr/blob/master/pages/osx/base64.md) version and a [`common`](https://github.com/tldr-pages/tldr/blob/master/pages/common/base64.md) version.  Just expand `-i` in the osx version (and translations) to `--input=...` and in the common version, if there is `-i`, change it to `--ignore-garbage` (and the translations).

bl-ue:
@luebken are you still available for this PR?

sbrl:
If there's not activity after ~3 weeks, I think we can take over and tidy a PR up & merge - so long as the OP has signed the CLA (which is the case here).

bl-ue (approved):

navarroaxel (approved):

PR #5237, opened 2021-02-05
psalm: add page

owenvoke:

### DIFF
--- a/pages/common/psalm.md
+++ b/pages/common/psalm.md
@@ -1,0 +1,32 @@
+# psalm
+
+> A static analysis tool for finding errors in PHP applications.
+> More information: <https://psalm.dev>.
+
+- Generate a Psalm configuration:
+
+`psalm --init`
+
+- Analyse the current working directory:
+
+`psalm`
+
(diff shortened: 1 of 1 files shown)


bl-ue (approved):

pages/common/psalm.md
```diff
@@ -0,0 +1,32 @@
...
+
+`psalm --init`
+
+- Analyse the current working directory:
+
+`psalm`
+
+- Analyse a specific directory or file:
+
+`psalm {{path/to/directory_or_file}}`
```
navarroaxel, line 16: we always use `file` first.
```suggestion
`psalm {{path/to/file_or_directory}}`
```

navarroaxel (changes requested):

pages/common/psalm.md
owenvoke: Good spot, I think I originally wrote `path/to/directory`, and then added `or_file` to the end. But it sounds really odd that way. 👍🏻

navarroaxel (approved):

PR #5241, opened 2021-02-05
multiple pages: move away from die.net more information links

### LINKED ISSUE #5143: Replace die.net links with better resources where applicable
There was [a short discussion](https://gitter.im/tldr-pages/tldr?at=600008e514cec811ec998765) today on Gitter about pages which have "more information" links pointing to `die.net`. As @navarroaxel showed there is man.archlinux.org, which is for Arch Linux programs, and which could be better suited for Arch Linux specific utilities.

In general though, we have the same distro-generic set of manual pages that `die.net` offers on `man7.org`, which IMHO renders the pages a lot better (sometimes `die.net` pages have syntax errors due to special characters being treated strangely when rendered in HTML, furthermore `man7.org` pages are actually in monospace). 

Therefore I suggest we convert all "more information" links which are pointing to `die.net` to point to `man7.org` instead, and in case the package/command is distro-specific, directly point to the distribution's own manual pages if available, for example:

- [Ubuntu](http://manpages.ubuntu.com/manpages/)
- [Debian](https://manpages.debian.org/buster/man-db/index.html)
- [Arch](https://man.archlinux.org)
- etc...

bl-ue:
Replaces die.net "More information" links with man7.org when possible, otherwise the command's OS's specific man page.

Closes #5143

### DIFF
--- a/pages/common/iotop.md
+++ b/pages/common/iotop.md
@@ -1,7 +1,7 @@
 # iotop
 
 > Display a table of current I/O usage by processes or threads.
-> More information: <https://linux.die.net/man/1/iotop>.
+> More information: <https://man7.org/linux/man-pages/man8/iotop.8.html>.
 
 - Start top-like I/O monitor:
 
--- a/pages/common/sponge.md
+++ b/pages/common/sponge.md
@@ -1,7 +1,7 @@
 # sponge
 
 > Soak up the input before writing the output file.
-> More information: <https://linux.die.net/man/1/sponge>.
+> More information: <https://manpages.debian.org/moreutils/sponge.1.html>.
 
 - Append file content to the source file:
 


pages/common/iotop.md
```diff
@@ -1,7 +1,7 @@
 # iotop
 
 > Display a table of current I/O usage by processes or threads.
-> More information: <https://linux.die.net/man/1/iotop>.
+> More information: <https://manpages.debian.org/iotop/iotop.8.html>.
```
navarroaxel, line 4: I prefer this:
```suggestion
> More information: <https://man7.org/linux/man-pages/man8/iotop.8.html>.
```

bl-ue: Good to know&mdash;so there are more man7.org pages than I thought. I will be more careful with the rest of the 30 die.net pages 👍🏻

navarroaxel (approved):

bl-ue:
Sorry @navarroaxel I've got 29 more pages left, so I've just started :smile:

Will you please check the latest one, `sponge`?  I couldn't find a man7 page for it.

pages/common/sponge.md
```diff
@@ -1,7 +1,7 @@
 # sponge
 
 > Soak up the input before writing the output file.
-> More information: <https://linux.die.net/man/1/sponge>.
+> More information: <https://manpages.debian.org/moreutils/sponge.1.html>.
```
navarroaxel, line 4: maybe we can use latest, to redirect to the latest man page version automatically:
```suggestion
> More information: <https://manpages.debian.org/latest/moreutils/sponge.1.en.html>.
```

navarroaxel: Oh forget it. It's the same behavior with your URL.

bl-ue: I don't think we need this, because mitting `latest/` redirects to the latest version too.

bl-ue: On that note, the debian manpage link in the other PRs need to be updated to this too, because they have the version right in the link, like `buster`, `stretch`, etc.

bl-ue:
Actually, I'm going to close this as I'm not ready to work on it, and I'd like to start with a clean slate when I'm ready.

sbrl:
Should we delete the branch or keep it around, @bl-ue?

bl-ue:
I forgot to delete it 🤷🏻

PR #5245, opened 2021-02-06
deluser, userdel: add page

TheLastGimbus:

Hi there!

Current `deluser` is in the `pages/common/`, which I don't understand, because it works a little bit different on different OSes :thinking: 

I added the correct version to `pages/linux/`, but dunno whether I should also delete old one in `common`

I know you guys are picky about language, so 1-2 example might need re-phrasing :sweat_smile:


pages/linux/deluser.md
```diff
@@ -0,0 +1,20 @@
+# deluser
+
+> Delete user from system
```
bl-ue:
```suggestion
> Delete a user from them system.
```

pages/linux/deluser.md
```diff
@@ -0,0 +1,20 @@
...
+> All commands must be executed as root (`sudo [command]`)
```
bl-ue:
```suggestion
> Note: all commands must be executed as root.
```

pages/linux/deluser.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Remove a user:
+
+`deluser {{name}}`
+
+- Remove a user and their home directory:
+
+`deluser --remove-home {{name}}`
+
+- Remove a user and their home, but backup it's files first:
```
bl-ue, line 14:
```suggestion
- Remove a user and their home, backing up their files:
```

pages/linux/deluser.md
```diff
@@ -0,0 +1,20 @@
...
+
+`deluser --backup-to {{path/to/backup}} --remove-home {{name}}`
```
bl-ue, line 16: I assume this is a directory?
```suggestion
`deluser --backup-to {{path/to/backup_directory}} --remove-home {{username}}`
```

pages/linux/deluser.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Remove a user, and all files owned by them:
+
+`deluser --remove-all-files {{name}}
```
bl-ue: This is a cool option!

pages/linux/deluser.md
```diff
@@ -0,0 +1,20 @@
+# deluser
+
+> Delete user from system.
```
bl-ue, line 3: Perhaps this will read better?
```suggestion
> Delete a user from the system.
```

pages/linux/deluser.md
```diff
@@ -0,0 +1,20 @@
...
+> All commands must be executed as root (`sudo [command]`).
```
bl-ue, line 4:
```suggestion
> Note: all commands must be executed as root.
```

pages/linux/deluser.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Remove a user:
+
+`deluser {{name}}`
```
bl-ue, line 8: A user might be confused and think display name, e.g. so they might run `deluser John Doe` instead of `deluser johndoe` ;)
```suggestion
`deluser {{username}}`
```

pages/linux/deluser.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Remove a user and their home directory:
+
+`deluser --remove-home {{name}}`
```
bl-ue, line 12:
```suggestion
`deluser --remove-home {{username}}`
```

pages/linux/deluser.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Remove a user and their home, but backup it's files first:
+
+`deluser --backup-to {{path/to/backup}} --remove-home {{name}}`
+
+- Remove a user, and all files owned by them:
+
+`deluser --remove-all-files {{name}}`
```
bl-ue, line 20:
```suggestion
`deluser --remove-all-files {{username}}`
```

This is a cool option!! :smile:

bl-ue (changes requested):
Hello @TheLastGimbus, and thank you for the new page! I've left some comments below for you to review.

pages/linux/deluser.md
TheLastGimbus: Oh, yes, it's a directory where one `.tag.gz` file will be placed

pages/linux/deluser.md
TheLastGimbus: I didn't know how to put this well, but I would keep keep it as original (if it's not strongly against the rules)

bl-ue:
@TheLastGimbus I think `common/deluser.md` might actually be [`userdel`](https://manpages.debian.org/buster/passwd/userdel.8.en.html) (#175) which I believe is more low-level.

`deluser.md` and `userdel.md` should be in `linux/` not `common/` 🤔 

Well, they're very old 😄 #175 is very old considering that this is PR #5245!

@TheLastGimbus Will you please:
1. Delete `common/userdel.md`
2. Rename `common/deluser.md` to `linux/userdel.md`
3. Add https://man7.org/linux/man-pages/man8/userdel.8.html as a "More information" link to `linux/userdel.md`
4. Add https://manpages.debian.org/deluser.8.en.html as a "More information" link to `linux/deluser.md`

Sorry, I know that's a lot, but these are old pages and need a little cleanup.

Thank you!

pages/linux/deluser.md
bl-ue: "it's" doesn't sound right&mdash;if you're referring to the user, I would use "their".

pages/linux/deluser.md
bl-ue: Oh it's an archive?  Then I would clarify this in the description.

pages/linux/deluser.md
bl-ue: Oh, also https://github.com/tldr-pages/tldr/pull/5245#discussion_r571437570

bl-ue: How about this?
```suggestion
- Remove a user and their home, but backup their files into a `.tar.gz` file in the specified directory:
```

pages/linux/deluser.md
```diff
@@ -0,0 +1,20 @@
+# deluser
+
+> Delete a user from the system.
+> Note: all commands must be executed as root.
```
bl-ue, line 4:
```suggestion
> Note: all commands must be executed as root.
> More information: <https://manpages.debian.org/deluser.8.en.html>.
```

pages/linux/deluser.md
TheLastGimbus: I tried playing with it, and I discovered that pure `--backup` doesn't work and `--backup-to` needs to have exisitng directory (can't create a new one)

Should we add this?

pages/linux/deluser.md
TheLastGimbus: Sure, we can add this. Not all people have debian tho, and I thought that man pages are kinda too obvious

bl-ue: Well, we do that if there's no other page.

bl-ue: If there's something better than the Debian manpage go ahead and use it.  We use man7.org when possible, then I guess the Debian manpages.

pages/linux/deluser.md
bl-ue: I'm not sure what you mean&mdash;my suggestion seems good enough.  If the directory is required to exist, well, I don't think we need to state that...many commands require the specified file/directory to exist (obviously :smile:)

bl-ue:
Note to maintainers: the deleted `deluser` pages was wrong, it shouldn't have been added.  @rubenvereecken added it #175, but I think they confused it with `userdel`.

TheLastGimbus:
Okay, everything resolved and cleaned up as you wished 

> I know you guys are picky about language

...but I do agree with that very much. Better finish it once for good than fix it 3 more times later :+1:

bl-ue (approved):

pages/linux/userdel.md
```diff
@@ -1,7 +1,17 @@
...
+> More information: <https://man7.org/linux/man-pages/man8/userdel.8.html>.
 
-- Remove a user and their home directory:
+- Remove a user:
+
+`userdel {{name}}`
+
+- Remove a user along with their home directory and mail spool:
 
 `userdel -r {{name}}`
```
navarroaxel, line 13: Looks like we can use the long argument here:
```suggestion
`userdel --remove {{name}}`
```

pages/linux/deluser.md
```diff
@@ -0,0 +1,21 @@
...
+> More information: <https://manpages.debian.org/deluser.8.en.html>.
+
+- Remove a user:
+
+`deluser {{username}}`
+
+- Remove a user and their home directory:
+
+`deluser --remove-home {{username}}`
+
```
navarroaxel, line 14: Could you add an extra example for `--backup`:
```suggestion

- Remove a user and their home, but backup their files into a `.tar.gz` file in the current directory:

`deluser --backup --remove-home {{username}}`

```

pages/linux/userdel.md
```diff
@@ -1,7 +1,17 @@
...
+
+- Remove a user from a group:
+
+`userdel {{name}} {{group}}`
```
navarroaxel, line 17: We can add an example for `--root`:
```suggestion
`userdel {{name}} {{group}}`

- Remove a user in a new root directory:

`userdel --root {{path/to/new/root}} {{name}}`
```

navarroaxel (commented):
Hey @TheLastGimbus, I think we can add 2 extra examples.

pages/linux/deluser.md
bl-ue: Actually @navarroaxel, it goes to `/` not the cwd.

pages/linux/userdel.md
bl-ue: @navarroaxel that description is a little unclear...could you please improve it? (only if possible, because the man page isn't clear either)

bl-ue: Or clarify me if I'm just misreading it :smile:

navarroaxel: I tried to use a similar description from the `chroot` page.
🤔 imagine you're in a live USB and you want to remove a user from your installed Linux. So mount `/dev/sda1` into `/mnt`, then run `userdel --root /mnt user1`. This commands in like `chroot /mnt` and  `userdel user1` combined.

pages/linux/deluser.md
```diff
@@ -0,0 +1,21 @@
+# deluser
+
+> Delete a user from the system.
+> Note: all commands must be executed as root.
+> More information: <https://manpages.debian.org/deluser.8.en.html>.
```
navarroaxel, line 5: This link is not working for me.

```suggestion
> More information: <https://manpages.debian.org/adduser/deluser.8.html>.
```

pages/linux/userdel.md
bl-ue: Ah I see!  Very good explanation 👍🏻  Now how would you get that succinctly into the exmaple description?

pages/linux/deluser.md
navarroaxel: My mistake. Check the description now.

bl-ue: How about just `in /`?

pages/linux/deluser.md
bl-ue: Strange. In all browsers it redirects to https://manpages.debian.org/buster/adduser/deluser.8.en.html

But adding `adduser/` is a bit more verbose&mdash;for the better, IMO.

pages/linux/deluser.md
TheLastGimbus: > I tried playing with it, and I discovered that pure `--backup` doesn't work [...]

I got no absolutley no files in / when I was using it :shrug: Ubuntu 20.04

pages/linux/userdel.md
TheLastGimbus: Added this, but changed "new" to "other"

bl-ue (approved):

navarroaxel (approved):

PR #5248, opened 2021-02-06
arch-chroot: add page

navarroaxel:

Source code for check functionality: https://git.archlinux.org/arch-install-scripts.git/tree/arch-chroot.in

### DIFF
--- a/pages/linux/arch-chroot.md
+++ b/pages/linux/arch-chroot.md
@@ -1,0 +1,28 @@
+# arch-chroot
+
+> Enhanced `chroot` command for Arch Linux installation.
+> More information: <https://man.archlinux.org/man/arch-chroot.8>.
+
+- Start interactive shell in a new root directory:
+
+`arch-chroot {{path/to/new/root}}`
+
+- Specify user to use:
+
+`arch-chroot -u {{user}} {{path/to/new/root}}`
+
+- Run command in a new root directory:
+
+`arch-chroot {{path/to/new/root}} {{command}} {{command_arguments}}`
+
+- Generate the initial ramdisk using the preset provided by the `linux` package on new root directory:
+
+`arch-chroot {{path/to/new/root}} {{mkinitcpio --preset linux}}`
+
+- Start a sh session in a new root directory:
+
+`arch-chroot {{path/to/new/root}} {{sh}}`
+
+- Start a zsh session in a new root directory (the `zsh` package should be installed in the target):
+
+`arch-chroot {{path/to/new/root}} {{zsh}}`
--- a/pages/linux/arch-chroot.md
+++ b/pages/linux/arch-chroot.md
@@ -19,10 +19,6 @@
 
 `arch-chroot {{path/to/new/root}} {{mkinitcpio --preset linux}}`
 
-- Start a sh session in a new root directory:
-
-`arch-chroot {{path/to/new/root}} {{sh}}`
-
 - Start a zsh session in a new root directory (the `zsh` package should be installed in the target):
 
 `arch-chroot {{path/to/new/root}} {{zsh}}`
--- a/pages/linux/arch-chroot.md
+++ b/pages/linux/arch-chroot.md
@@ -15,10 +15,6 @@
 
 `arch-chroot {{path/to/new/root}} {{command}} {{command_arguments}}`
 
-- Generate the initial ramdisk using the preset provided by the `linux` package on new root directory:
-
-`arch-chroot {{path/to/new/root}} {{mkinitcpio --preset linux}}`
-
 - Specify the shell, other than the default `bash` (in this case, the `zsh` package should have been installed in the target system):
 
 `arch-chroot {{path/to/new/root}} {{zsh}}`
--- a/pages/linux/arch-chroot.md
+++ b/pages/linux/arch-chroot.md
@@ -3,15 +3,15 @@
 > Enhanced `chroot` command to help in the Arch Linux installation process.
 > More information: <https://man.archlinux.org/man/arch-chroot.8>.
 
-- Start an interactive shell in a new root directory:
+- Start an interactive shell (`bash`, by default) in a new root directory:
 
 `arch-chroot {{path/to/new/root}}`
 
-- Specify the user to run as:
+- Specify the user (other than the current user) to run the shell as:
 
 `arch-chroot -u {{user}} {{path/to/new/root}}`
 
-- Run a command in a new root directory:
+- Run a custom command (instead of the default `bash`) in the new root directory:
 
 `arch-chroot {{path/to/new/root}} {{command}} {{command_arguments}}`
 
@@ -19,6 +19,6 @@
 
 `arch-chroot {{path/to/new/root}} {{mkinitcpio --preset linux}}`
 
-- Start a zsh session in a new root directory (the `zsh` package should be installed in the target):
+- Specify the shell, other than the default `bash` (in this case, the `zsh` package should have been installed in the target system):
 
 `arch-chroot {{path/to/new/root}} {{zsh}}`
--- a/pages/linux/arch-chroot.md
+++ b/pages/linux/arch-chroot.md
@@ -3,11 +3,11 @@
 > Enhanced `chroot` command for Arch Linux installation.
 > More information: <https://man.archlinux.org/man/arch-chroot.8>.
 
-- Start interactive shell in a new root directory:
+- Start an interactive shell in a new root directory:
 
 `arch-chroot {{path/to/new/root}}`
 
-- Specify user to run as:
+- Specify the user to run as:
 
 `arch-chroot -u {{user}} {{path/to/new/root}}`
 
--- a/pages/linux/arch-chroot.md
+++ b/pages/linux/arch-chroot.md
@@ -7,11 +7,11 @@
 
 `arch-chroot {{path/to/new/root}}`
 
-- Specify user to use:
+- Specify user to run as:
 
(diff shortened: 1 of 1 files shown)


pages/linux/arch-chroot.md
```diff
@@ -0,0 +1,28 @@
+# arch-chroot
+
+> Enhanced `chroot` command for Arch Linux installation.
+> More information: <https://man.archlinux.org/man/arch-chroot.8>.
+
+- Start interactive shell in a new root directory:
+
+`arch-chroot {{path/to/new/root}}`
+
+- Specify user to use:
```
bl-ue, line 10:
```suggestion
- Specify user to run as:
```

pages/linux/arch-chroot.md
```diff
@@ -0,0 +1,28 @@
...
+
+`arch-chroot -u {{user}} {{path/to/new/root}}`
+
+- Run command in a new root directory:
```
bl-ue, line 14:
```suggestion
- Run a command in a new root directory:
```

pages/linux/arch-chroot.md
```diff
@@ -0,0 +1,28 @@
...
+
+`arch-chroot {{path/to/new/root}} {{command}} {{command_arguments}}`
+
+- Generate the initial ramdisk using the preset provided by the `linux` package on new root directory:
+
+`arch-chroot {{path/to/new/root}} {{mkinitcpio --preset linux}}`
+
+- Start a sh session in a new root directory:
```
bl-ue, line 22: To me this and the next example seem unecessary; I'd say the first example covers all of the functionality demonstrated here.  What do you think?

(also `a sh` is incorrect, at least for me, as I read 'sh' as the sequential letters 's' 'h')

bl-ue (changes requested):

pages/linux/arch-chroot.md
navarroaxel: I'm not sure about `sh` but the tip about install `zsh` first is good.

pages/linux/arch-chroot.md
navarroaxel: I think we should fix the `chroot` page

bl-ue (approved):
That's great.  It solves all my points and more beautifully in one commit! 😂

pages/linux/arch-chroot.md
```diff
@@ -0,0 +1,24 @@
+# arch-chroot
+
+> Enhanced `chroot` command for Arch Linux installation.
+> More information: <https://man.archlinux.org/man/arch-chroot.8>.
+
+- Start interactive shell in a new root directory:
+
+`arch-chroot {{path/to/new/root}}`
+
+- Specify user to run as:
```
sbrl, line 10: Perhaps this will read better?
```suggestion
- Specify the user to run as:
```

pages/linux/arch-chroot.md
```diff
@@ -0,0 +1,24 @@
+# arch-chroot
+
+> Enhanced `chroot` command for Arch Linux installation.
+> More information: <https://man.archlinux.org/man/arch-chroot.8>.
+
+- Start interactive shell in a new root directory:
```
sbrl, line 6: Perhaps this will read better?
```suggestion
- Start an interactive shell in a new root directory:
```

pages/linux/arch-chroot.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Specify user to run as:
+
+`arch-chroot -u {{user}} {{path/to/new/root}}`
+
+- Run a command in a new root directory:
+
+`arch-chroot {{path/to/new/root}} {{command}} {{command_arguments}}`
+
+- Generate the initial ramdisk using the preset provided by the `linux` package on new root directory:
```
sbrl, line 18: `on new root directory` doesn't really make sense here?

sbrl (commented):

pages/linux/arch-chroot.md
navarroaxel: @bl-ue, @owenvoke ?

pages/linux/arch-chroot.md
```diff
@@ -0,0 +1,24 @@
+# arch-chroot
+
+> Enhanced `chroot` command for Arch Linux installation.
```
waldyrious, line 3: Maybe this is clearer?
```suggestion
> Enhanced `chroot` command for Arch Linux systems.
```

navarroaxel: I think no, because this command is a helper for the Arch Linux installation process, not for users of the OS.

waldyrious: Ah, gotcha. Then I'd suggest saying precisely that:
```suggestion
> Enhanced `chroot` command to help in the Arch Linux installation process.
```

pages/linux/arch-chroot.md
bl-ue: @sbrl are you sure?  This command is about changing the root directory (`arch-chroot`).

pages/linux/arch-chroot.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Start an interactive shell in a new root directory:
+
+`arch-chroot {{path/to/new/root}}`
+
+- Specify the user to run as:
+
+`arch-chroot -u {{user}} {{path/to/new/root}}`
+
+- Run a command in a new root directory:
```
waldyrious, line 14: We should clarify that this invocation executes the single command, and doesn't enter the chroot:
```suggestion
- Run a single command in a new root directory, and exit:
```

waldyrious: Or perhaps use this opportunity to clarify the mental model of how it works:
```suggestion
- Run a custom command (instead of the default `bash`) in the new root directory:
```

WDYT @navarroaxel?

pages/linux/arch-chroot.md
```diff
@@ -0,0 +1,24 @@
+# arch-chroot
+
+> Enhanced `chroot` command to help in the Arch Linux installation process.
+> More information: <https://man.archlinux.org/man/arch-chroot.8>.
+
+- Start an interactive shell in a new root directory:
```
waldyrious, line 6: I guess an extra clarification could be useful:
```suggestion
- Start an interactive shell (`bash`, by default) in a new root directory:
```

pages/linux/arch-chroot.md
```diff
@@ -0,0 +1,24 @@
...
+
+`arch-chroot {{path/to/new/root}}`
+
+- Specify the user to run as:
```
waldyrious, line 10: Is the following correct? Would it help?
```suggestion
- Specify the user (other than the current user) to run the shell as:
```

pages/linux/arch-chroot.md
waldyrious: Since this is (IIUC) just a more specific case of the previous example, I'm not sure it's worth the trouble of adding it here. I see a few options:

- Remove this example and let the generic one above provide the template for it
- Keep it, but don't attempt to explain what `mkinitcpio --preset linux` does, and why one might want to run it here (as they do [in the Arch wiki](https://wiki.archlinux.org/index.php/chroot#Run_a_single_command_and_exit))
- Keep it and edit the description to provide some context of why one would run this command. For example:
  ```suggestion
  - Generate the initial ramdisk for the new installation using the `linux` preset:
  ```

If this is indeed a common use of `arch-chroot`, it might make sense to keep it.

pages/linux/arch-chroot.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Run a command in a new root directory:
+
+`arch-chroot {{path/to/new/root}} {{command}} {{command_arguments}}`
+
+- Generate the initial ramdisk using the preset provided by the `linux` package on new root directory:
+
+`arch-chroot {{path/to/new/root}} {{mkinitcpio --preset linux}}`
+
+- Start a zsh session in a new root directory (the `zsh` package should be installed in the target):
```
waldyrious, line 22: How about moving this example to be next to the "custom user" one above? The sequencing of the page may make more sense that way IMO:

```suggestion
- Specify the shell, other than the default `bash` (in this case, the `zsh` package should have been installed in the target system):
```

(sorry if some of my comments don't make much sense, I'm a little out of my depth here.)

pages/linux/arch-chroot.md
waldyrious: I'm honestly not sure which of the two suggestions above is more helpful 😅 do you guys any preference?

pages/linux/arch-chroot.md
navarroaxel: Well, I think the `mkinitcpio` is not commonly used in this way. So, I can remove it.

pages/linux/arch-chroot.md
```diff
@@ -0,0 +1,24 @@
...
+`arch-chroot -u {{user}} {{path/to/new/root}}`
+
+- Run a custom command (instead of the default `bash`) in the new root directory:
+
+`arch-chroot {{path/to/new/root}} {{command}} {{command_arguments}}`
+
+- Generate the initial ramdisk using the preset provided by the `linux` package on new root directory:
+
+`arch-chroot {{path/to/new/root}} {{mkinitcpio --preset linux}}`
+
```
navarroaxel, line 21:
```suggestion
```

navarroaxel (commented):
Removed specific example, and leave the generic previous example.

bl-ue (approved):

sbrl (approved):
Thanks, @navarroaxel!

PR #5258, opened 2021-02-11
read: add example using 'while'

orestisfl:
Copied from `while`

### DIFF
--- a/pages/common/read.md
+++ b/pages/common/read.md
@@ -10,10 +10,6 @@
 
 `read -a {{array}}`
 
-- Enable backspace and GNU readline hotkeys when entering input with read:
-
-`read -e {{variable}}`
-
 - Specify the number of maximum characters to be read:
 
 `read -n {{character_count}} {{variable}}`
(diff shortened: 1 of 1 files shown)


orestisfl:
Hah, I didn't count. Let me know if you would like to replace an example with this one. I think it's more intuitive to recall the syntax of a while read loop through the `read` page instead of the `while` page.

bl-ue:
In my opinion, @orestisfl, the 3rd example (`Enable backspace and GNU readline hotkeys when entering input with read`) doesn't seem particularly useful, whereas your example does.  How about you remove the 3rd example? (but keep your example at the bottom to maintain visual consistency as all of the other examples start with `read`)

bl-ue:
Also @orestisfl please sign the CLA to pass the `license/cla` check.

orestisfl:
@bl-ue up :)

bl-ue (approved):

sbrl (approved):
This seems fine to me. That readline example looks neat, but I can see how your new example would be more immediately useful in the terminal. For example, I recently cooked up this to rename a sequence of PNG images into a format that ffmpeg can understand:

```bash
i=1; while read -r filename; do mv "${filename}" "${i}.png"; i=$((i+1)); done < <(find . -iname '*.png' | sort);
```

PR #5261, opened 2021-02-11
grub-install, grub-mkconfig: add page

Managor:

### DIFF
--- a/grub-mkconfig.md
+++ b/grub-mkconfig.md
@@ -1,6 +1,7 @@
 # grub-mkconfig
 
 > Generate a GRUB configuration file.
+> More information: <https://www.gnu.org/software/grub/manual/grub/html_node/Invoking-grub_002dmkconfig.html>
 
 - Generate the configuration:
 
--- a/grub-mkconfig.md
+++ b/grub-mkconfig.md
@@ -3,6 +3,6 @@
 > Generate a GRUB configuration file.
 > More information: <https://www.gnu.org/software/grub/manual/grub/html_node/Invoking-grub_002dmkconfig.html>
 
-- Generate the configuration:
+- Generate the configuration file:
 
-`grub-mkconfig -o {{/boot/grub/grub.cfg}}`
+`grub-mkconfig --output={{/boot/grub/grub.cfg}}`
--- a/grub-mkconfig.md
+++ b/grub-mkconfig.md
@@ -1,0 +1,7 @@
+# grub-mkconfig
+
+> Generate a GRUB configuration file.
+
+- Generate the configuration:
+
+`grub-mkconfig -o /boot/grub/grub.cfg`
--- a/grub-mkconfig.md
+++ b/grub-mkconfig.md
@@ -10,3 +10,7 @@
 - Generate the configuration file:
 
 `sudo grub-mkconfig --output={{/boot/grub/grub.cfg}}`
+
+- Print the help page:
+
+`grub-mkconfig --help`
--- a/grub-mkconfig.md
+++ b/grub-mkconfig.md
@@ -3,6 +3,10 @@
 > Generate a GRUB configuration file.
 > More information: <https://www.gnu.org/software/grub/manual/grub/html_node/Invoking-grub_002dmkconfig.html>
 
+- Do a dry run and print the configuration to stdout
+
+`sudo grub-mkconfig`
+
 - Generate the configuration file:
 
-`grub-mkconfig --output={{/boot/grub/grub.cfg}}`
+`sudo grub-mkconfig --output={{/boot/grub/grub.cfg}}`
--- a/grub-mkconfig.md
+++ b/grub-mkconfig.md
@@ -5,4 +5,4 @@
 
 - Generate the configuration:
 
-`grub-mkconfig -o /boot/grub/grub.cfg`
+`grub-mkconfig -o {{/boot/grub/grub.cfg}}`
--- a/grub-mkconfig.md
+++ b/grub-mkconfig.md
@@ -3,7 +3,7 @@
 > Generate a GRUB configuration file.
 > More information: <https://www.gnu.org/software/grub/manual/grub/html_node/Invoking-grub_002dmkconfig.html>
 
-- Do a dry run and print the configuration to stdout
+- Do a dry run and print the configuration to stdout:
 
 `sudo grub-mkconfig`
 
--- a/pages/linux/grub-install.md
+++ b/pages/linux/grub-install.md
@@ -1,0 +1,12 @@
+# grub-install
+
+> Install GRUB to a device.
+> More information: <https://www.gnu.org/software/grub/manual/grub/html_node/Installing-GRUB-using-grub_002dinstall.html>
+
+- Install GRUB on a BIOS system:
+
+`grub-install --target=i386-pc /dev/sd{{X}}`
+
+- Install GRUB on an UEFI system:
+
+`grub-install --target=x86_64-efi --efi-directory={{/efi/}} --bootloader-id=GRUB`
--- a/pages/linux/grub-install.md
+++ b/pages/linux/grub-install.md
@@ -5,7 +5,7 @@
 
 - Install GRUB on a BIOS system:
 
-`grub-install --target={{i386-pc}} {{/dev/sdX}}`
+`grub-install --target={{i386-pc}} {{/path/to/device}}`
 
 - Install GRUB on an UEFI system:
 
--- a/pages/linux/grub-install.md
+++ b/pages/linux/grub-install.md
@@ -1,7 +1,7 @@
 # grub-install
 
 > Install GRUB to a device.
-> More information: <https://www.gnu.org/software/grub/manual/grub/html_node/Installing-GRUB-using-grub_002dinstall.html>
+> More information: <https://www.gnu.org/software/grub/manual/grub/html_node/Installing-GRUB-using-grub_002dinstall.html>.
 
 - Install GRUB on a BIOS system:
 
--- a/pages/linux/grub-install.md
+++ b/pages/linux/grub-install.md
@@ -5,8 +5,8 @@
 
 - Install GRUB on a BIOS system:
 
-`grub-install --target=i386-pc /dev/sd{{X}}`
+`grub-install --target={{i386-pc}} {{/dev/sdX}}`
 
 - Install GRUB on an UEFI system:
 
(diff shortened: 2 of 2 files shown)


bl-ue:
Will you please add a link for `grub-mkconfig`?

pages/linux/grub-install.md
```diff
@@ -0,0 +1,12 @@
+# grub-install
+
+> Install GRUB to a device.
+> More information: <https://www.gnu.org/software/grub/manual/grub/html_node/Installing-GRUB-using-grub_002dinstall.html>
+
+- Install GRUB on a BIOS system:
+
+`grub-install --target=i386-pc /dev/sd{{X}}`
```
bl-ue, line 8: Did you get `X` from PR #5036 :smile:?  I think se should do this:
```suggestion
`grub-install --target={{i386-pc}} {{/dev/sdX}}`
```

pages/linux/grub-install.md
```diff
@@ -0,0 +1,12 @@
...
+
+- Install GRUB on an UEFI system:
+
+`grub-install --target=x86_64-efi --efi-directory={{/efi/}} --bootloader-id=GRUB`
```
bl-ue, line 12:
```suggestion
`grub-install --target={{x86_64-efi}} --efi-directory={{/efi/}} --bootloader-id={{GRUB}}`
```

bl-ue (changes requested):

pages/linux/grub-install.md
Managor: I was thinking that since the X is the only part that changes, I'll only give that a token status.

Is anything else than `i386-pc` in active use for BIOS systems?

bl-ue: I don't know&mdash;we usually do this with user-passed values.

Managor: Alright. Someone can improve it in the future if they think that those values are practically static.

grub-mkconfig.md
```diff
@@ -0,0 +1,8 @@
+# grub-mkconfig
+
+> Generate a GRUB configuration file.
```
bl-ue, line 3:
```suggestion
> Generates GRUB configuration files.
```

grub-mkconfig.md
```diff
@@ -0,0 +1,8 @@
...
+> More information: <https://www.gnu.org/software/grub/manual/grub/html_node/Invoking-grub_002dmkconfig.html>
+
+- Generate the configuration:
```
bl-ue, line 6:
```suggestion
- Generate a configuration file:
```

grub-mkconfig.md
```diff
@@ -0,0 +1,8 @@
...
+
+`grub-mkconfig -o /boot/grub/grub.cfg`
```
bl-ue, line 8: I know, this may never change, but again, this is what we usually do.  It also just makes it clearer what files are being modified, etc.
```suggestion
`grub-mkconfig -o {{/boot/grub/grub.cfg}}`
```

bl-ue (changes requested):

grub-mkconfig.md
Managor: This is actually directly copied from the `man` page

pages/linux/grub-install.md
Managor: Actually, if we look at /pages/common/dd.md, it sets the precedent of `/dev/{{sdX}}`

From dd.md: `dd if={{file.iso}} of=/dev/{{usb_drive}} status=progress`

grub-mkconfig.md
bl-ue: I suggested that because I wanted the example itself to say that, instead of the page descript, but I didn't want them both to be the same.  I realize that this tool only is not configurable, so please disregard the above suggestion.

grub-mkconfig.md
```diff
@@ -0,0 +1,8 @@
+# grub-mkconfig
+
+> Generate a GRUB configuration file.
+> More information: <https://www.gnu.org/software/grub/manual/grub/html_node/Invoking-grub_002dmkconfig.html>
+
+- Generate the configuration:
+
+`grub-mkconfig -o {{/boot/grub/grub.cfg}}`
```
navarroaxel, line 8: We prefer the long argument form:
```suggestion
`grub-mkconfig --output={{/boot/grub/grub.cfg}}`
```

grub-mkconfig.md
```diff
@@ -0,0 +1,8 @@
+# grub-mkconfig
```
navarroaxel, line 1: Based on the command help page, we can add a first example with just:
```
grub-mkconfig
```
And the generated file is send to the `stdout`, looks like this could be used as a dry run (without side effects in the OS).

Also we can add the help just to let the user know he can more info in the CLI:
```
grub-mkconfig --help
```

pages/linux/grub-install.md
```diff
@@ -0,0 +1,12 @@
+# grub-install
```
navarroaxel, line 1: You can add an example with `--modules` because is useful is some scenarios when at boot time you get `device not found`:
```
grub-install ... --modules="{{part_gpt part_msdos}}"
```

pages/linux/grub-install.md
bl-ue: There's actually ongoing discussion about this over at #5036, so let's just keep what we have here and when #5036 gets resolved we can fix everything that depends on it, including this.

grub-mkconfig.md
bl-ue: @Managor I updated this suggestion to comply with our previous discussion.  You can apply it.

grub-mkconfig.md
bl-ue: @Managor even though `--help` is not a particularly useful thing to document, especially because we don't document it in many other pages, if there's nothing else to document and we're well below the 8-example limit, we usually add options like `--help` (if the command has it, which this one does :wink:).

pages/linux/grub-install.md
navarroaxel: I think we should review the token syntax in the `dd` page, because the entery path is the argument option and some other argument options should use `{{}}` too.
Using the `mkfs` page: we can use `{{path/to/partition}}` here, because this could be `/dev/nvme0n1` too.

bl-ue: Use `{{path/to/partition}}`.  It's more generic.  Anyway we need to get #5036 figured out soon so we can avoid this trouble in the future.

Managor: Maybe {{path/to/device}} would be better as it doesn't point to any specific partion?

grub-mkconfig.md
```diff
@@ -0,0 +1,12 @@
...
+
+> Generate a GRUB configuration file.
+> More information: <https://www.gnu.org/software/grub/manual/grub/html_node/Invoking-grub_002dmkconfig.html>
+
+- Do a dry run and print the configuration to stdout
```
bl-ue, line 6:
```suggestion
- Do a dry run and print the configuration to stdout:
```

bl-ue (changes requested):
Missing colon.

pages/linux/grub-install.md
bl-ue: Sure!

pages/linux/grub-install.md
Managor: Can you suggest a change? I've never used that option so I don't feel I have the authority to know what context it can be used in.

pages/linux/grub-install.md
```diff
@@ -0,0 +1,12 @@
...
+> Install GRUB to a device.
+> More information: <https://www.gnu.org/software/grub/manual/grub/html_node/Installing-GRUB-using-grub_002dinstall.html>.
+
+- Install GRUB on a BIOS system:
+
+`grub-install --target={{i386-pc}} {{/path/to/device}}`
+
+- Install GRUB on an UEFI system:
+
+`grub-install --target={{x86_64-efi}} --efi-directory={{/efi/}} --bootloader-id={{GRUB}}`
```
navarroaxel, line 12:
```suggestion
`grub-install --target={{x86_64-efi}} --efi-directory={{path/to/efi_directory}} --bootloader-id={{GRUB}}`

- Install GRUB pre-loading specific modules:

`grub-install --target={{x86_64-efi}} --efi-directory={{path/to/efi_directory}} --modules="{{part_gpt part_msdos}}"`
```

pages/linux/grub-install.md
```diff
@@ -0,0 +1,12 @@
+# grub-install
+
+> Install GRUB to a device.
+> More information: <https://www.gnu.org/software/grub/manual/grub/html_node/Installing-GRUB-using-grub_002dinstall.html>.
+
+- Install GRUB on a BIOS system:
+
+`grub-install --target={{i386-pc}} {{/path/to/device}}`
```
navarroaxel, line 8: The patch could be relative, like `mkfs`, don't start with `/`:
```suggestion
`grub-install --target={{i386-pc}} {{path/to/device}}`
```

navarroaxel (changes requested):

pages/linux/grub-install.md
Managor: An honest mistake

bl-ue:
Whoops! Forgot about this one.

pages/linux/grub-install.md
bl-ue: cc @navarroaxel

bl-ue (changes requested):

pages/linux/grub-install.md
bl-ue: cc @navarroaxel

navarroaxel: The example was added as the last example of this page.

navarroaxel (approved):

bl-ue (approved):

pages/linux/grub-install.md
Managor: @navarroaxel So when the path must be absolute, it is to be lead with `/`? Just noticed one in `transmission-remote` and decided to ask before opening a pull request.

sbrl: @Managor yeah, in general we use `path/to/thing` for relative paths, and `/path/to/thing` for absolute paths. Unless we're on a windows-specific page I'd guess, in which case absolute paths would probably be represented by something like `Z:\path\to\file` or something (but that hasn't yet come up as far as I'm aware).

PR #5262, opened 2021-02-12
z:  add `-c` parameter to the command description

crs1138:
Details:
---
* add the `-c` parameter to the [`z`](https://github.com/tldr-pages/tldr/blob/master/pages/common/z.md) command that enables user to go to a subdirectory of the current  directory

### DIFF
--- a/pages/common/z.md
+++ b/pages/common/z.md
@@ -26,3 +26,7 @@
 - Remove the current directory from `z`'s database:
 
 `z -x .`
+
+- Go to the directory that is a subdirectory of the current directory.
+
+`z -c {{foo}}`
--- a/pages/common/z.md
+++ b/pages/common/z.md
@@ -27,6 +27,6 @@
 
 `z -x .`
 
-- Go to the directory that is a subdirectory of the current directory.
+- Go to the directory that is a subdirectory of the current directory (equivalent of `cd`):
 
 `z -c {{foo}}`
--- a/pages/common/z.md
+++ b/pages/common/z.md
@@ -1,6 +1,6 @@
 # z
 
-> Tracks the most used (_frecent_, read frequently recent) directories and enables quickly navigating to them using string or regex patterns.
+> Tracks the most used (by frecency) directories and enables quickly navigating to them using string or regex patterns.
 > More information: <https://github.com/rupa/z>.
 
 - Go to a directory that contains "foo" in the name:
--- a/pages/common/z.md
+++ b/pages/common/z.md
@@ -1,6 +1,6 @@
 # z
 
-> Tracks the most used directories and enables quickly navigating to them using string or regex patterns.
+> Tracks the most used (_frecent_, read frequently recent) directories and enables quickly navigating to them using string or regex patterns.
 > More information: <https://github.com/rupa/z>.
 
 - Go to a directory that contains "foo" in the name:
@@ -27,6 +27,6 @@
 
 `z -x .`
 
-- Go to the directory that is a subdirectory of the current directory (equivalent of `cd`):
+- Restrict matches to subdirectories of the current directory:
 
 `z -c {{foo}}`


pages/common/z.md
```diff
@@ -26,3 +26,7 @@
 - Remove the current directory from `z`'s database:
 
 `z -x .`
+
+- Go to the directory that is a subdirectory of the current directory.
```
bl-ue, line 30: @crs1138 it appears that `z -c` is the equivalent of `cd`.  If **and only if** this is true, please apply this suggestion:
```suggestion
- Go to the directory that is a subdirectory of the current directory (equivalent of `cd`):
```

bl-ue (changes requested):

bl-ue:
@crs1138 the CLA isn't working because the email address that you used for the commit (`jpozivil@leadtech.com`) isn't added to your GitHub account.  To fix this, you can either:
1. Change the email you used for the commit† to [one of the emails attached to your GitHub account](https://github.com/settings/emails),

2. Or, preferrably (and much easier), [add the email you used for this commit (`jpozivil@leadtech.com`) to your GitHub account](https://github.com/settings/emails).

†To change the email address of the last commit, run this command:
```
git commit --amend --author="Jan Honza Pozivil <email addres that you want to use instead>"
```

bl-ue:
@crs1138 Thank you.  Now please [sign the CLA](https://cla-assistant.io/tldr-pages/tldr?pullRequest=5262).

pages/common/z.md
crs1138: Thanks for your feedback. I've applied the changes you requested. Not sure how to proceed now. Can I resubmit this PR or better to make a new one?

bl-ue (approved):
The man page says about `-c`: `restrict matches to subdirectories of the current directory`.  But it seems to me that it's just `cd`. 🤔 

LGTM!

crs1138:
Not quite it can cd much deeper with a simple syntax:
`z -c foo` can be equivalent of `cd bar/car/dar/food-whatever`. I keep looking it up frequently as I have two projects with similar tree structure and sometimes I want to stay in the project that is less frequently used. And there's no `man z` and `z --help` only lists the parameters with no explanation of what they do.

bl-ue:
Hmm...I can't figure how to use `-c` then.
```console
$ pwd
/Users/me/a
$ tree
.
└── b
    └── c
        └── d
            └── e
                └── f
                    └── g

6 directories, 0 files
$ z -c g
$ pwd
/Users/me/a
$ z g -c .
$ pwd
/Users/me/a
$
```
Also, however I installed it, I have the [manpage](https://github.com/rupa/z/blob/master/z.1).

crs1138:
Hmm, I used `brew` to install it if I'm not mistaken. I just assumed no one had the man pages. 

You gotta navigate to the `g` folder first so there's a record of that folder in `z`'s _frecent_ records.

crs1138:
> _frecent_

That stands for _frequently recent_, it's not a typo.

bl-ue:
Ah I see, that works well! Thank you.  Maybe the example should be reworded.

pages/common/z.md
```diff
@@ -1,6 +1,6 @@
 # z
 
-> Tracks the most used directories and enables quickly navigating to them using string or regex patterns.
+> Tracks the most used (_frecent_, read frequently recent) directories and enables quickly navigating to them using string or regex patterns.
```
bl-ue, line 3: Though helpful, I'm not sure tldr-pages in the place to store this info.  The more information link is enough IMO&mdash;but I can be overruled.  Thoughts @navarroaxel?
```suggestion
> Tracks the most used directories and enables quickly navigating to them using string or regex patterns.
```

sbrl:
Ah, frecent is a thing I've seen around. It's an algorithm for ranking things. Firefox switch to it a while back for ranking items in the list of suggestions when you're typing into the address bar of the browser. We should keep the word as-is.

Example further reading: https://github.com/johnsylvain/frecent

pages/common/z.md
sbrl: What about this?
```suggestion
> Tracks the most used (by frecency) directories and enables quickly navigating to them using string or regex patterns.
```

sbrl (approved):
Thanks, @crs1138!

PR #5264, opened 2021-02-13
tar: fix bzip to bzip2 in command description

robbyoconnor:


bl-ue:
Good catch @robbyoconnor! Smart for updating the translations :)

bl-ue:
I'm not quite sure how to word the PR's title&mdash;@navarroaxel is an expert on that.

bl-ue (approved):

robbyoconnor:
I worded it correctly I think

robbyoconnor:
I also checked other places and it is correct there.

bl-ue:
>I worded it correctly I think

Sorry, I wasn't being clear.  For some reason or other, we're very strict about PR titles (not sure why) and a PR title that would be perfectly fine in many popular projects usually would be simplified in tldr-pages.

At least, @navarroaxel usually does that. I'm not sure if that's part of our standard, maybe @sbrl has something to say about this?

navarroaxel:
@bl-ue lol I just try to keep the PR title simple like I saw when I started contributing here. And, this is no code, so it's normal the title doesn't follow the standard way.
I think the edited page is key, but I don't think the title is a critical stuff in the process review.

navarroaxel (approved):

robbyoconnor:
> Good catch @robbyoconnor! Smart for updating the translations :)

@bl-ue - not gonna lie -- I use `sd` and `fd-find` to do a search and replace across all `tar` pages. It's bad form to translate things like `bzip2`

bl-ue:
@robbyoconnor cool. I didn't know about those tools.  Based on the tldr page `sd` looks cool, I may start using it all the just because you referenced it! `fd-find` doesn't have a tldr page, maybe we need a page for it 🤔 

As a matter of fact, I don't use CLI tools for find/replace.  I use [`rg`](https://github.com/BurntSushi/ripgrep) because I've found it to be spectacular in every way about searching, but for replacing, I prefer the visual, intuitive, incredible UI of [VS Code](https://github.com/microsoft/vscode) which lets me find matches by a regex, select which files/directories to include/exclude, see a preview of the changes, and deselect which occurrences I'd like not to replace.

Frequently, when there's only 1 or 2 matches or it's a very common generic term like a two-lettered word (e.g. `of`), I just click the file and manually make the change as it's very little work.

Anyway I strongly recommend you use it too! :D

sbrl:
@bl-ue: You should have permission to edit the PR title. If you see an inconsistency, generally I just edit it silently :slightly_smiling_face:

bl-ue:
@sbrl I am.  Collaborators+ are.  I just wanted @navarroaxel's input because I'm not as familiar with doing so as he.

Issue #5256, opened 2021-02-08
Let's document: pacman


navarroaxel:
The `pacman` command is a Swiss army knife to manage packages in Arch Linux-based OS and we should extend the current [`pacman`](https://github.com/tldr-pages/tldr/blob/master/pages/linux/pacman.md) page in `tldr` creating specific pages for its subcommands.

The issue with `pacman` is that its subcommands are arguments in uppercase and the lowercased arguments are specific for that subcommand. If we review the `pacman -Syu` command to upgrade the system: `-y/--refresh` and `-u/--sysupgrade` are specific arguments for `pacman -S` or we can run the same command in its long form: `pacman --sync --refresh --sysupgrade`.

- [x] `-D/--database` (#5302)
- [x] `-F/--files` (#5303)
- [x] `-Q/--query` (#5269)
- [x] `-R/--remove` (#5273)
- [x] `-S/--sync` (#5268)
- [x] `-T/--deptest` (#5304)
- [x] `-U/--upgrade` (#5295)

How should we name the page for `-S/--sync` argument? Maybe `pacman-sync.md`.

bl-ue:
>How should we name the page for -S/--sync argument? Maybe pacman-sync.md.

I thought of somthing else, `pacman--sync.md`.  It's slightly uglier, but it would clearly distinguish between `pacman sync` and `pacman --sync`.

nikisix:
The git man documentation uses a single-dash separator between man `<command>-<subcommand>`.

Wouldn't the `--sync` flag be defined within the top-level `pacman.md` file?

I definitely understand `pacman-sync.md` being a thing.

bl-ue:
@nikisix in an ordinary command, yes, but with `pacman`, `-S/--sync` is just like a subcommand.  There's a lot of options for it, and the same for `-D/--database`, `-F/--files`, etc. so there's not nearly enough space in `pacman.md` to document the useful capabilities of those "subcommands" (though they're not really subcommands)

sbrl:
Since I assume that pacman has long-form options for each of the "subcommands", I think `pacman-SUBCOMMAND_NAME` is a good option. e.g. `pacman --sync` becomes `pacman-sync` for example.

A bit unusual yes, but pacman doesn't appear to follow norms established by other subcommand-based applications.

bl-ue:
[...]

PR #5268, opened 2021-02-15
pacman-sync: add page

navarroaxel:

For: #5256

### DIFF
--- a/pages/linux/pacman-sync.md
+++ b/pages/linux/pacman-sync.md
@@ -1,0 +1,36 @@
+# pacman sync
+
+> Arch Linux package manager utility.
+> More information: <https://man.archlinux.org/man/pacman.8>.
+
+- Install a new package:
+
+`pacman --sync {{package_name}}`
+
+- Synchronize and update all packages (add `--downloadonly` for download but do not update anything):
+
+`pacman --sync --refresh --sysupgrade`
+
+- Search the package database for a regular expression or keyword:
+
+`pacman --sync --search "{{search_pattern}}"`
+
+- Display information for a package:
+
+`pacman --sync --info {{package_name}}`
+
+- Install a package as non-explicitly installed:
+
+`pacman --sync --asdeps {{package_name}}`
+
+- Overwrite conflicting files during an update:
+
+`pacman --sync --refresh --sysupgrade --overwrite {{path/to/file}}`
+
+- Synchronize and update all packages except the ignored (can be used more than once):
+
+`pacman --sync --refresh --sysupgrade --ignore {{package_name}}`
+
+- Remove not installed packages and unused repositories from the cache (use `-cc` to clean all packages):
+
+`pacman --sync --clean`
--- a/pages/linux/pacman-sync.md
+++ b/pages/linux/pacman-sync.md
@@ -23,7 +23,7 @@
 
 `pacman --sync --info {{package_name}}`
 
-- Overwrite conflicting files during an update:
+- Overwrite conflicting files during a package update:
 
 `pacman --sync --refresh --sysupgrade --overwrite {{path/to/file}}`
 
--- a/pages/linux/pacman-sync.md
+++ b/pages/linux/pacman-sync.md
@@ -7,7 +7,7 @@
 
 `pacman --sync {{package_name}}`
 
-- Synchronize and update all packages (add `--downloadonly` for download but do not update anything):
+- Synchronize and update all packages (add `--downloadonly` to download only):
 
 `pacman --sync --refresh --sysupgrade`
 
--- a/pages/linux/pacman-sync.md
+++ b/pages/linux/pacman-sync.md
@@ -1,4 +1,4 @@
-# pacman sync
+# pacman --sync
 
 > Arch Linux package manager utility.
 > More information: <https://man.archlinux.org/man/pacman.8>.
--- a/pages/linux/pacman-sync.md
+++ b/pages/linux/pacman-sync.md
@@ -7,7 +7,7 @@
 
 `sudo pacman --sync {{package_name}}`
 
-- Synchronize and update all packages (add `--downloadonly` to download only):
+- Synchronize and update all packages (add `--downloadonly` to download the packages and not update them):
 
 `sudo pacman --sync --refresh --sysupgrade`
 
--- a/pages/linux/pacman-sync.md
+++ b/pages/linux/pacman-sync.md
@@ -11,6 +11,10 @@
 
 `pacman --sync --refresh --sysupgrade`
 
+- Update all packages and install a new one without prompting:
+
+`pacman --sync --refresh --sysupgrade --noconfirm {{package_name}}`
+
 - Search the package database for a regular expression or keyword:
 
 `pacman --sync --search "{{search_pattern}}"`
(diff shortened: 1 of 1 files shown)


pages/linux/pacman-sync.md
```diff
@@ -0,0 +1,36 @@
+# pacman sync
```
bl-ue, line 1: Didn't we decide to use `pacman --sync` as the title?
```suggestion
# pacman --sync
```

sbrl: I think you're thinking of the filename?

pages/linux/pacman-sync.md
```diff
@@ -0,0 +1,36 @@
...
+
+> Arch Linux package manager utility.
+> More information: <https://man.archlinux.org/man/pacman.8>.
+
+- Install a new package:
+
+`pacman --sync {{package_name}}`
+
+- Synchronize and update all packages (add `--downloadonly` for download but do not update anything):
```
bl-ue, line 10: Wouldn't it make more sense to _replace_ `--sysupgrade` with `--downloadonly`, rather than add `--downloadonly`?

pages/linux/pacman-sync.md
```diff
@@ -0,0 +1,36 @@
...
+
+`pacman --sync --refresh --sysupgrade`
+
+- Search the package database for a regular expression or keyword:
+
+`pacman --sync --search "{{search_pattern}}"`
+
+- Display information for a package:
```
bl-ue, line 18:
```suggestion
- Display information about a package:
```

bl-ue (changes requested):

pages/linux/pacman-sync.md
bl-ue: No, I'm not, actually.  Don't you think it would be better to have the title match what is actually entered in the terminal?  Currently it looks like `sync` is a normal subcommand of `pacman`.

pages/linux/pacman-sync.md
navarroaxel: Without `--sysupgrade`, `pacman` doesn't download the packages. The `--refresh` only updates the local repositories (similar to `apt update`)

bl-ue: Then, how about...
```suggestion
- Synchronize and update all packages (add `--downloadonly` to download only):
```

pages/linux/pacman-sync.md
```diff
@@ -0,0 +1,36 @@
...
+
+`pacman --sync --info {{package_name}}`
+
+- Install a package as non-explicitly installed:
+
+`pacman --sync --asdeps {{package_name}}`
+
+- Overwrite conflicting files during an update:
```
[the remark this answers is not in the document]
navarroaxel: There is no system update like `do-release-upgrade`, some arch-based OS's like `Manjaro` try to imitate that behavior.
Sometimes the user needs to "force" replace a file during a package update, the most common scenario is "change of permissions" and the Arch Linux announcements are used for this news.

pages/linux/pacman-sync.md
```diff
@@ -0,0 +1,36 @@
...
+
+`pacman --sync --refresh --sysupgrade --overwrite {{path/to/file}}`
+
+- Synchronize and update all packages except the ignored (can be used more than once):
+
+`pacman --sync --refresh --sysupgrade --ignore {{package_name}}`
+
+- Remove not installed packages and unused repositories from the cache (use `-cc` to clean all packages):
```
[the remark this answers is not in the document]
navarroaxel: You're right, the docs only said `-cc`.

pages/linux/pacman-sync.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Search the package database for a regular expression or keyword:
+
+`pacman --sync --search "{{search_pattern}}"`
+
+- Display information for a package:
+
+`pacman --sync --info {{package_name}}`
+
+- Install a package as non-explicitly installed:
```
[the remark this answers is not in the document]
navarroaxel: Maybe is advanced, but orphan packages is a common concept. If you install something as a `dependency` from nothing is an orphan, and there is a `--query` to list all the orphans.

pages/linux/pacman-sync.md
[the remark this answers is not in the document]
bl-ue: Okay, thank you.  How about changing it to 'package update', because new users might think 'system update' (like I did)
```suggestion
- Overwrite conflicting files during a package update:
```
What do you think?

pages/linux/pacman-sync.md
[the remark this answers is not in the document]
bl-ue: The man page says:

>This is useful for makepkg and other build-from-source tools that need to install dependencies before building the package.

But this isn't quite clear to me.  Could you please clarify, and then I could come up with an explanatory description targeted at beginners (from a beginner :smile:)?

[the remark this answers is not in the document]
navarroaxel: `makepkg` is used to build packages locally. Arch doens't have `PPA` repositories, you can build one, but is not recommend to the users, security stuff.
The AUR is download the build instruction and build the package locally, then install it.
`makepkg` install build dependencies to build the package.
I use this method when I want to use a package for a short time, and then clean up.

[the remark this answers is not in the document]
bl-ue: If there is another example that would work, I think you should pick it :+1: because that seems more like a power-user concept.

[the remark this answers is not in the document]
navarroaxel: Done!

pages/linux/pacman-sync.md
sbrl: Oh I see what you mean. Yes, I agree - this change would be a good idea.

bl-ue (approved):

pages/linux/pacman-sync.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Display information about a package:
+
+`pacman --sync --info {{package_name}}`
+
+- Overwrite conflicting files during a package update:
+
+`pacman --sync --refresh --sysupgrade --overwrite {{path/to/file}}`
+
+- Synchronize and update all packages except the ignored (can be used more than once):
```
sbrl, line 30: Perhaps this will read better?
```suggestion
- Synchronize and update all packages, but ignore a specific package (can be used more than once):
```

pages/linux/pacman-sync.md
```diff
@@ -0,0 +1,36 @@
+# pacman --sync
+
+> Arch Linux package manager utility.
+> More information: <https://man.archlinux.org/man/pacman.8>.
+
+- Install a new package:
+
+`sudo pacman --sync {{package_name}}`
+
+- Synchronize and update all packages (add `--downloadonly` to download only):
```
bl-ue, line 10: How about this?  Coming back to this page after a while, I read it as trying to add the string `--downloadonly` to the string "download only" (???)
```suggestion
- Synchronize and update all packages (add `--downloadonly` to download the packages and not update them):
```

bl-ue (approved):

sbrl (approved):
Thanks, @navarroaxel!

PR #5270, opened 2021-02-16
systemctl: improve mask/unmask example description

aherst:
Edit example description for Mask/Unmask to clarify meaning as per man page:

           This is a stronger version of disable, since it prohibits all
           kinds of activation of the unit, including enablement and manual
           activation.

### DIFF
--- a/pages/linux/systemctl.md
+++ b/pages/linux/systemctl.md
@@ -31,6 +31,6 @@
 
 `systemctl is-active {{unit}}`
 
-- Check if a unit is enabled:
+- Check if a unit is enabled
 
 `systemctl is-enabled {{unit}}`
--- a/pages/linux/systemctl.md
+++ b/pages/linux/systemctl.md
@@ -19,7 +19,7 @@
 
 `systemctl enable/disable {{unit}}`
 
-- Mask/Unmask a unit to prohibit enablement and manual activation:
+- Mask/unmask a unit to prevent enablement and manual activation:
 
 `systemctl mask/unmask {{unit}}`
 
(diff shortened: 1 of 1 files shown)


pages/linux/systemctl.md
```diff
@@ -19,7 +19,7 @@
 
 `systemctl enable/disable {{unit}}`
 
-- Mask/Unmask a unit, prevent it to be started on bootup:
+- Mask/Unmask a unit to prohibit enablement and manual activation:
```
bl-ue, line 22: I'm biased toward the original, simpler 'prevent'.  What do you think?
```suggestion
- Mask/unmask a unit to prevent enablement and manual activation:
```

bl-ue: I get that the manpage uses 'prohibit' but IMO simpler is better, at least for tldr-pages.

bl-ue: @aherst sorrry, did you apply this suggestion? I'm confused.

aherst: Sorry, I'm new at this and learning as I go. I thought you had applied the change and was asking if I objected. I'll figure out the right thing to do when I have some time this afternoon. Thanks for you patience.

bl-ue: Np @aherst.  No, I didn't apply the change.  Click "Commit suggestion" then "Commit changes" to commit the changes.

bl-ue: (sorry about the testing below, I needed to test if tldr-bot was working because we had some server maintenance ;))

aherst: Thanks for you help.

bl-ue (approved):

navarroaxel (approved):

PR #5272, opened 2021-02-17
tldr: change -u example to long form

SanSYS:
image

### DIFF
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -17,4 +17,4 @@
 
 - Update local pages (if the client supports caching):
 
-`tldr -u`
+`tldr --update`


navarroaxel:
Hi @SanSYS, thanks for your submission, but in our [CLIENT-SPECIFICATION](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#arguments) guide we said:
```
Clients MAY choose to only implement the *short version* of an option, ignoring the long form.
```

So, this is an issue in the tldr client that you're using and a bug should be raised in their repository.

bl-ue:
But that client should certainly support the short options!?!

@SanSYS how did you install that client?

bl-ue:
Based on the "tldr Client program" in the screenshot I'm guessing you're using the [Haskell](https://github.com/psibi/tldr-hs) client, which I believe is installed via `sudo apt install tldr`, but on the [README](https://github.com/psibi/tldr-hs/blob/master/README.md) of that client I see a lot of flags and options, including the short versions 🤔 

@SanSYS is this correct?

SanSYS:
@bl-ue 
```bash
❯ history | grep "apt install tldr"
  203  sudo apt install tldr

❯ apt show tldr     
Package: tldr
Version: 0.4.0.1-1build2
Priority: optional
Section: universe/doc
Source: haskell-tldr
Origin: Ubuntu
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Debian Haskell Group <pkg-haskell-maintainers@lists.alioth.debian.org>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 3 013 kB
Depends: libatomic1 (>= 4.8), libc6 (>= 2.29), libffi7 (>= 3.3~20180313), libgmp10, git
Conflicts: tldr-py (<< 0.7.0-2~)
Homepage: https://github.com/psibi/tldr-hs#readme
Download-Size: 584 kB
APT-Manual-Installed: yes
APT-Sources: http://ru.archive.ubuntu.com/ubuntu focal/universe amd64 Packages
Description: Haskell tldr client
 Haskell tldr client with support for updating and viewing tldr pages.
 .
 The TLDR pages are a community effort to simplify the beloved man
 pages with practical examples.  See https://tldr.sh/


```

bl-ue:
It appears then that you're using version `0.4.0.1-1build2` which is several years old.  The latest version of the Haskell client is  [v0.9.0](https://github.com/psibi/tldr-hs/releases/tag/v0.9.0). @psibi when does a new release make it's way to apt?

@SanSYS in the meantime, I suggest removing the haskell client and using the official [Python client](https://github.com/tldr-pages/tldr-python-client) by running the following:

```
sudo apt remove -y tldr
sudo apt install -y tldr-py
```

bl-ue:
We should consider amending the specification and making long options **required**, because all over our pages we emphasize using long options when possible! :thinking:

@SanSYS we should close this now since we shouldn't break the specification.

psibi:
> The latest version of the Haskell client is v0.9.0. @psibi when does a new release make it's way to apt? We've had other issues with this before.

I don't maintain the package in the Debian or the ubuntu repositories. So you would have to speak to them.

> in the meantime, I suggest removing the Haskell client and using the official Python client by running the following:

Or instead you can just download the new binaries from here: https://github.com/psibi/tldr-hs/releases/tag/v0.9.0 :-)

bl-ue:
>I don't maintain the package in the Debian or the ubuntu repositories. So you would have to speak to them.

:+1:

>Or instead you can just download the new binaries from here: https://github.com/psibi/tldr-hs/releases/tag/v0.9.0 :-)

@psibi the problem with that approach is that users may wish to just install via their package manager, and not have to use binary.  Personally, I would not mind, but not everyone is the same.

psibi:
@bl-ue Yeah, I agree. :-) I was just advertising the haskell client.  :smiley:

sbrl:
Getting Debian packages updated is a rather complicated effort. We should start with getting in touch with the haskell client autthors, and see what the status of their client is. If it's still maintained, then it's possible that the apt package has been forgotten about or there's otherwise some other blocking factor. Either way, that's something that should be discussed in a separate issue.

bl-ue:
@sbrl see the above comments. @psibi is the Haskell client maintainer and they don't publish it.  Someone else does :thinking:

SanSYS:
@bl-ue thank you for your time! i tried tldr-py - not so good as a solution on haskell )

Screenshot_20210221_211717

bl-ue:
Wow that really isn't good a good interface.  Hmm...I see from one of the error message lines that the version of the Python client you have is `0.7.0`!  The latest is [1.2.0](https://github.com/tldr-pages/tldr-python-client/releases/tag/1.2.0).  Guess you're out of luck with the package manager.

Is just adding one of the [prebuilt Haskell binaries](https://github.com/psibi/tldr-hs/releases/tag/v0.9.0)&mdash;that @psibi mentioned above&mdash;to a folder in your `$PATH` a feasible solution for you?

bl-ue:
You could also try `pip install tldr` to install 1.2.0 of the Python client from PyPI.

SanSYS:
@bl-ue ok, I installed pip, installed tldr from it and... 
I received slower tool with red colors (why red 😏?)
It's not friendly way for users, I think;
But thank you for your time!

Screenshot_20210225_114859

bl-ue:
@SanSYS

>I received slower tool

Yes, it's going to be slower, because the interpreted Python is slower than the compiled Haskell.  We're going to fix this one of these days though...

>with red colors (why red 😏?)

Why red? Because that's the default color:

https://github.com/tldr-pages/tldr-python-client/blob/e52d0e404a93dd7ea9190b5b441ac18b0cf3c3ec/tldr.py#L244

You can change all of the colors, see [here](https://github.com/tldr-pages/tldr-python-client/tree/main#configuration).

sbrl:
You could always try the reference Node.js client?

My personal favourite is the [tldr-bash-client](https://github.com/pepa65/tldr-bash-client).

bl-ue:
I'd like to close this now as we've decided it's not needed.

SanSYS:
How many tldr implementations!
somebody should write in on golang, rust, C#, F#, and brainfuck )

bl-ue:
Yes! We already have 4 Go ones, 1 Rust one, and 1 C# one :D

bl-ue:
Looking at your repos @SanSYS, I think https://github.com/SanSYS/subdscan could use a tldr page... :wink:

PR #5280, opened 2021-02-19
wol: add page

Managor:

### DIFF
--- a/pages/linux/wol.md
+++ b/pages/linux/wol.md
@@ -17,7 +17,7 @@
 
 - Send a WoL packet to a specific port:
 
-`wol --port={{port_number}}`
+`wol --port={{port_number}} {{mac_address}}`
 
 - Read hardware addresses, IP addresses/hostnames, optional ports and SecureON password from a file:
 
--- a/pages/linux/wol.md
+++ b/pages/linux/wol.md
@@ -5,19 +5,19 @@
 
 - Send a WoL packet to a device:
 
-`wol {{mac_adress}}`
+`wol {{mac_address}}`
 
 - Send a WoL packet to a device in another subnet based on its IP:
 
-`wol --ipaddr={{ip_address}} {{mac_adress}}`
+`wol --ipaddr={{ip_address}} {{mac_address}}`
 
 - Send a WoL packet to a device in another subnet based on its hostname:
 
-`wol --host={{hostname}} {{mac_adress}}`
+`wol --host={{hostname}} {{mac_address}}`
 
 - Send a WoL packet to a specific port:
 
-`wol --port={{porn_number}}`
+`wol --port={{port_number}}`
 
 - Read hardware addresses, IP addresses/hostnames, optional ports and SecureON password from a file:
 
@@ -25,4 +25,4 @@
 
 - Turn on verbose input:
 
-`wol --verbose {{mac_adress}}`
+`wol --verbose {{mac_address}}`
--- a/pages/linux/wol.md
+++ b/pages/linux/wol.md
@@ -15,7 +15,7 @@
 
 `wol --host={{hostname}} {{mac_address}}`
 
-- Send a WoL packet to a specific port:
+- Send a WoL packet to a specific port on a host:
 
 `wol --port={{port_number}} {{mac_address}}`
 
--- a/pages/linux/wol.md
+++ b/pages/linux/wol.md
@@ -23,6 +23,6 @@
 
 `wol --file={{path/to/file}}`
 
-- Turn on verbose input:
+- Turn on verbose output:
 
 `wol --verbose {{mac_address}}`
--- a/pages/linux/wol.md
+++ b/pages/linux/wol.md
@@ -19,7 +19,7 @@
 
 `wol --port={{port_number}} {{mac_address}}`
 
-- Read hardware addresses, IP addresses/hostnames, optional ports and SecureON password from a file:
+- Read hardware addresses, IP addresses/hostnames, optional ports and SecureON passwords from a file:
 
 `wol --file={{path/to/file}}`
 
--- a/pages/linux/wol.md
+++ b/pages/linux/wol.md
@@ -1,0 +1,28 @@
+# wol
+
+> Client for sending Wake-on-LAN magic packets.
+> More information: <https://sourceforge.net/projects/wake-on-lan/>.
+
+- Send a WoL packet to a device:
+
+`wol {{mac_adress}}`
+
+- Send a WoL packet to a device in another subnet based on its IP:
+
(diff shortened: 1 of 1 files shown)


Managor:
Not entirely sure if `--verbose` should be at the start or after `--file` because the change is very miniscule.

pages/linux/wol.md
```diff
@@ -0,0 +1,28 @@
+# wol
+
+> Client for sending Wake-on-LAN magic packets.
+> More information: <https://sourceforge.net/projects/wake-on-lan/>.
+
+- Send a WoL packet to a device:
+
+`wol {{mac_adress}}`
```
bl-ue, line 8:
```suggestion
`wol {{mac_address}}`
```

pages/linux/wol.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Send a WoL packet to a device in another subnet based on its IP:
+
+`wol --ipaddr={{ip_address}} {{mac_adress}}`
```
bl-ue, line 12:
```suggestion
`wol --ipaddr={{ip_address}} {{mac_address}}`
```

pages/linux/wol.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Send a WoL packet to a device in another subnet based on its hostname:
+
+`wol --host={{hostname}} {{mac_adress}}`
```
bl-ue, line 16:
```suggestion
`wol --host={{hostname}} {{mac_address}}`
```

pages/linux/wol.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Send a WoL packet to a specific port:
+
+`wol --port={{porn_number}}`
```
bl-ue, line 20:
```suggestion
`wol --port={{port_number}}`
```

pages/linux/wol.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Read hardware addresses, IP addresses/hostnames, optional ports and SecureON password from a file:
+
+`wol --file={{path/to/file}}`
+
+- Turn on verbose input:
+
+`wol --verbose {{mac_adress}}`
```
bl-ue, line 28:
```suggestion
`wol --verbose {{mac_address}}`
```

bl-ue (changes requested):
Really cool tool @Managor.  I found a few typos :-)

pages/linux/wol.md
Managor: A Freudian slip while getting the inspiration to do this half asleep.

navarroaxel (approved):

pages/linux/wol.md
```diff
@@ -0,0 +1,28 @@
...
+> Client for sending Wake-on-LAN magic packets.
+> More information: <https://sourceforge.net/projects/wake-on-lan/>.
+
+- Send a WoL packet to a device:
+
+`wol {{mac_address}}`
+
+- Send a WoL packet to a device in another subnet based on its IP:
+
+`wol --ipaddr={{ip_address}} {{mac_address}}`
```
bl-ue, line 12: This confuses me a bit.  Why would you need to specify the IP address and the MAC address?  In the first example it's quite obvious that you use the MAC address to specify what host to target, but in this one it seams that you use the IP to find the host instead.  I don't understand, then, the presence of both.  Same with the below example.

pages/linux/wol.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Send a WoL packet to a device in another subnet based on its hostname:
+
+`wol --host={{hostname}} {{mac_address}}`
+
+- Send a WoL packet to a specific port:
+
+`wol --port={{port_number}} {{mac_address}}`
+
+- Read hardware addresses, IP addresses/hostnames, optional ports and SecureON password from a file:
```
bl-ue, line 22: Is 'password' supposed to plural in this case? It if it, then use "passwords". If singular is intended, then prefix "SecureON" with "the" or "a", e.g.:
```suggestion
- Read hardware addresses, IP addresses/hostnames, optional ports and the SecureON password from a file:
```

pages/linux/wol.md
bl-ue, line 12: This confuses me a bit.  Why would you need to specify the IP address and the MAC address?  In the first example it's quite obvious that you use the MAC address to specify what host to target, but in this one it seams that you use the IP to find the host instead.  I don't understand, then, the presence of both.  Same with the below example.

bl-ue (changes requested):

pages/linux/wol.md
Managor: I'd assume each device can have its own password. I copied the description straight from `man` since it was such a neat description.

pages/linux/wol.md
navarroaxel: The mac is always required. I think `wol` send the network package to an specific IP address as recipient instead of a broadcast to all the network.

bl-ue: Hmm. What is the MAC for?

pages/linux/wol.md
bl-ue: Okay, that's great as long as it's clear.  Before merging this I'd like to definitely understand this.

Managor:
@bl-ue No idea why I can't comment directly on your question but to answer it:
The PC has no concept of IP when it's powered down. The only way to contact it is to use a MAC address. IP in this case is used to direct the packet to a specific subnet as WOL works on Layer 2.

pages/linux/wol.md
navarroaxel: From [Wikipedia](https://en.wikipedia.org/wiki/Wake-on-LAN)
```
The magic packet is a frame that is most often sent as a broadcast and that contains anywhere within its payload 6 bytes of all 255 (FF FF FF FF FF FF in hexadecimal), followed by sixteen repetitions of the target computer's 48-bit MAC address, for a total of 102 bytes.
```
Sound like when you try to wake up someone repeating their name in a `while (true)`

bl-ue: Nice. So: the MAC address is required and is always that of the target computer.  Thank you ✅

bl-ue (approved):

navarroaxel:
> @bl-ue No idea why I can't comment directly on your question but to answer it:
> The PC has no concept of IP when it's powered down. The only way to contact it is to use a MAC address. IP in this case is used to direct the packet to a specific subnet as WOL works on Layer 2.

Same issue here. I think it's a GitHub bug :bug:

bl-ue:
@navarroaxel was your 2nd laptop shutdown and you turned it on from your 1st one?

navarroaxel:
I checked the BIOS first, and I found the WoL was turned ON. Then, I tried the WoL and works lol

PR #5278, opened 2021-02-18
vercel: add page

TGIFelix:
- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #5283, opened 2021-02-19
now: add page

navarroaxel:

### DIFF
--- a/pages/common/now.md
+++ b/pages/common/now.md
@@ -1,0 +1,29 @@
+# now
+
+> Cloud platform for serverless deployment.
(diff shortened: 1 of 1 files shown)


sbrl (approved):
For reference, see #5278 and https://gitter.im/tldr-pages/tldr?at=602fdd16dc108172e7893b08.

MasterOdin (approved):
While I agree that no one should be using it, I can kind of guarantee someone will probably use it until it actually fully breaks without realizing there is the upgrade. (I just got someone to migrate from Node 6 to 12 a few weeks ago as they did not even realize there were new versions).

PR #5288, opened 2021-02-20
ci: update to use @tldr-bot for deploy

owenvoke:
Closes tldr-pages/tldr-pages.github.io#41

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -14,7 +14,7 @@ function initialize {
 
   # Configure git.
   git config --global user.email "tldrbotgithub@gmail.com"
-  git config --global user.name "tldr-bot"
+  git config --global user.name "tldr bot"
   git config --global push.default simple
   git config --global diff.zip.textconv "unzip -c -a"
 
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -13,7 +13,7 @@ function initialize {
   export SITE_REPO_SLUG="tldr-pages/tldr-pages.github.io"
 
   # Configure git.
-  git config --global user.email "tldr-pages@googlegroups.com"
+  git config --global user.email "tldrbotgithub@gmail.com"
   git config --global user.name "tldr-bot"
   git config --global push.default simple
   git config --global diff.zip.textconv "unzip -c -a"
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -13,8 +13,8 @@ function initialize {
   export SITE_REPO_SLUG="tldr-pages/tldr-pages.github.io"
 
   # Configure git.
-  git config --global user.email "actions@github.com"
-  git config --global user.name "GitHub Actions"
+  git config --global user.email "tldr-pages@googlegroups.com"
+  git config --global user.name "tldr-bot"
   git config --global push.default simple
   git config --global diff.zip.textconv "unzip -c -a"
 


bl-ue (approved):

scripts/deploy.sh
```diff
@@ -13,8 +13,8 @@ function initialize {
   export SITE_REPO_SLUG="tldr-pages/tldr-pages.github.io"
 
   # Configure git.
-  git config --global user.email "actions@github.com"
-  git config --global user.name "GitHub Actions"
+  git config --global user.email "tldrbotgithub@gmail.com"
+  git config --global user.name "tldr-bot"
```
bl-ue, line 17:
```suggestion
  git config --global user.name "tldr bot"
```

bl-ue (commented):
Not sure that it matters much, but the Git user name should probably match @tldr-bot's display name, not user name.  GitHub will automatically display `tldr-bot` if it finds that the email address matches ones of the account's emails.

navarroaxel (approved):

scripts/deploy.sh
owenvoke: Fair point. 👍🏻

agnivade (approved):

owenvoke:
I'm going to merge this as it has 3 approvals, and we'll see how it goes. 👍🏻

bl-ue:
Nice! But there aren't any contributions showing up on [its profile](https://github.com/tldr-bot).  I wonder if that's because the PAT used to push the commits is under your account @owenvoke.  Could we maybe create a new one under @tldr-bot's account and use it instead of yours? 

That would also be nice because I always see

>**owenvoke** pushed to **tldr-pages/tldr-pages.github.io**

But since it's just automated, it's not really you, and so when you actually do push commits (such as merging this PR) I don't even notice it, thinking it's the automation.

What do you think?

owenvoke:
The PAT isn't under my account, it's under the @tldr-bot account (it was created by @agnivade). It will show up as whoeever created the commit in `master` that pushed to `tldr-pages/tldr-pages.github.io`, which happened to be me.

bl-ue:
Yes, I am watching that repository.  I'm watching all of our repos, so that I get notified.

Screen Shot 2021-02-24 at 8 44 29 AM

The user who pushed to `tldr-pages/tldr@master` isn't the one who pushes to `tldr-pages/tldr-pages.github.io`.  The person who merges a PR pushes to the former, and the CI (using your token, I believe) (always) pushes to the latter.

In the referenced issue, the [screenshot](https://user-images.githubusercontent.com/478237/71452007-cc46b980-2777-11ea-8058-a9ec3829d86f.png) that @waldyrious [uploaded](https://github.com/tldr-pages/tldr-pages.github.io/issues/41#issuecomment-568941787) shows that @agnivade pushed to `tldr-pages/tldr-pages.github.io`, which means that his token was used at that time (Dec 2019).  But in this screenshot of Gitter, it shows that you uploaded it:

Screen Shot 2021-02-24 at 9 06 55 AM

Which is what makes me think that you have the token now.

owenvoke:
Ah, I don't think it's related to the token at all. I think it's actually as the GitHub Actions deploy key was added by me (see the discussion in https://github.com/tldr-pages/tldr/pull/4125). 🤔

I'm not really sure if this is much of a problem as it doesn't count towards my GitHub stats, so even if we generated a new deploy key and added it through the UI using the @tldr-bot account, I don't think it would show up there anyway. 🤔 But I'm not entirely sure.

bl-ue:
@owenvoke Nope, it's not a problem for me.   Maybe when  I become and org owner I could do that sort of stuff myself though to unburden you guys...

agnivade:
Yes, Gitter just shows the owner via the Actions deploy key I believe. But the commits made to the repo are attributed to tldr-bot. https://github.com/tldr-pages/tldr-pages.github.io/commits/master.

I am not sure exactly if there is a problem to solve here.

bl-ue:
No, there's no particular problem, it would just be nice to make it so @tldr-bot is the user who pushes commits to the `tldr-pages.github.io` repo instead of @owenvoke.

PR #5308, opened 2021-02-23
gh-ssh-key: add page

navarroaxel:

### DIFF
--- a/pages/common/gh-ssh-key.md
+++ b/pages/common/gh-ssh-key.md
@@ -7,14 +7,14 @@
 
 `gh ssh-key`
 
-- List SSH keys:
+- List SSH keys for the currently authenticated user:
 
 `gh ssh-key list`
 
-- Add an SSH key to the user:
+- Add an SSH key to the currently authenticated user's account:
 
 `gh ssh-key add {{path/to/key.pub}}`
 
-- Add an SSH key to the user with a specific title:
+- Add an SSH key to the currently authenticated user's account with a specific title:
 
 `gh ssh-key add --title {{title}} {{path/to/key.pub}}`
--- a/pages/common/gh-ssh-key.md
+++ b/pages/common/gh-ssh-key.md
@@ -1,0 +1,20 @@
+# gh ssh-key
+
+> Manage GitHub SSH keys from the command line.
+> More information: <https://cli.github.com/manual/gh_ssh-key>.
+
+- Display help:
+
+`gh ssh-key`
+
+- List SSH keys:
+
+`gh ssh-key list`
+
+- Add an SSH key to the user:
+
+`gh ssh-key add {{path/to/key.pub}}`
+
+- Add an SSH key to the user with a specific title:
+
+`gh ssh-key add --title {{title}} {{path/to/key.pub}}`


pages/common/gh-ssh-key.md
```diff
@@ -0,0 +1,20 @@
...
+
+- List SSH keys:
+
+`gh ssh-key list`
+
+- Add an SSH key to the user:
+
+`gh ssh-key add {{path/to/key.pub}}`
+
+- Add an SSH key to the user with a specific title:
```
bl-ue, line 18:
```suggestion
- Add an SSH key to your account with a custom title:
```

pages/common/gh-ssh-key.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Display help:
+
+`gh ssh-key`
+
+- List SSH keys:
+
+`gh ssh-key list`
+
+- Add an SSH key to the user:
```
bl-ue, line 14:
```suggestion
- Add an SSH key to your account:
```

pages/common/gh-ssh-key.md
```diff
@@ -0,0 +1,20 @@
+# gh ssh-key
+
+> Manage GitHub SSH keys from the command line.
+> More information: <https://cli.github.com/manual/gh_ssh-key>.
+
+- Display help:
+
+`gh ssh-key`
+
+- List SSH keys:
```
bl-ue, line 10:
```suggestion
- List SSH keys attached to your account:
```

bl-ue (changes requested):
What about this innovation? :smile:

pages/common/gh-ssh-key.md
navarroaxel: Why try to avoid her/him/your. Take a look to https://github.com/tldr-pages/tldr/blob/master/pages/common/gh-issue.md

bl-ue: I known about him/her, but I didn't know about your.  Well we seem to use it in quite a few pages...126 of 2,221 (5.67%).  Still?

bl-ue: I like 'your' because it gets the point across so clearly and without the slightest confusion.

pages/common/gh-ssh-key.md
owenvoke: Why not something like:
> List SSH keys for the currently authenticated user:

pages/common/gh-ssh-key.md
owenvoke: Hmm, I think using something like `the currently authenticated user` makes this quite readable without having `him/her/your`. I also think that `specific` works better than `custom`, but I don't mind either way. 👍🏻

owenvoke (approved):
LGTM. But we could use `the currently authenticated user` for a few of these for readability. Although I think it's fine as is.

pages/common/gh-ssh-key.md
bl-ue: Sounds good.  I'll update my suggestions.

pages/common/gh-ssh-key.md
bl-ue: Personally I prefer custom in some circumstances just because it has fewer syllables.  But I think you're right.

bl-ue (approved):

pages/common/gh-ssh-key.md
sbrl: Generally speaking we try to avoid the 2nd person in tldr-pages. I like the idea of avoid him/her too - @owenvoke's suggestion here seems like the best solution.

PR #5329, opened 2021-03-01
images: add medium font

### LINKED ISSUE #5324: Missing: SometypeMono Medium-Small-Caps
The medium-small-caps variation of SometypeMono is not put in images directory therefore banner files in svg don't display properly.
image

yutyo:

Closes #5324


Yutyo:
🦗

bl-ue (approved):
LGTM.  I think @zlatanvasovic / @waldyrious should have a look as they've been involved in branding/asset changes/updates like this one the past.

sbrl (commented):
Nice work! Could you outline the changes you've made here, and perhaps - if there's a visual difference - provide a side-by-side comparison please?

Yutyo:
> Nice work! Could you outline the changes you've made here, and perhaps - if there's a visual difference - provide a side-by-side comparison please?

Not particular visual difference. What I did is simply deleting the font attribute within svg and transforming the text to path.

waldyrious:
Thanks for working on this, @Yutyo! I am however puzzled as to why you replaced actual text with hardcoded paths. Can you explain your rationale in more detail?

Please note that the current `banner.svg` file was actually hand-tweaked to ensure the source SVG is human-readable, semantically meaningful, and compact. I would very much prefer the same properties to remain in the new file.

Yutyo:
> Thanks for working on this, @Yutyo! I am however puzzled as to why you replaced actual text with hardcoded paths. Can you explain your rationale in more detail?
> 
> Please note that the current `banner.svg` file was actually hand-tweaked to ensure the source SVG is human-readable, semantically meaningful, and compact. I would very much prefer the same properties to remain in the new file.

The problem is due to those svg tweaks not being understood by many platforms such as Wikimedia Commons, that its automatically understood as a virus or doesn't work at all.

sbrl:
I think what @Yutyo is getting at here is that the font we use may not be available on all systems?

Currently, we use this in the CSS:

```css
font-family: 'Sometype Mono';
```

Perhaps an alternative solution would be use this instead:

```css
font-family: 'Sometype Mono', 'Courier New', 'monospace';
```

This will ensure that the text will at least be displayed in monospace. You can have an arbitrarily long list of fonts in priority order here too. For example, here I specify the original `Sometype Mono` is the preference, and if that's not available the [web safe font](https://www.w3schools.com/cssref/css_websafe_fonts.asp) `Courier New` is used instead. If even that's not available, then I specify the system or browser's default monospace font, whatever that may be.

CleanMachine1:
Is this PR going ahead?

zlatanvasovic:
I am not familiar with @waldyrious's work, so I'll approve if he approves of course.

marchersimon:
I'd like to bring this forward, but I don't really know how. Ping @yutyo, @sbrl, @waldyrious

sbrl:
Hmmm, I'm not sure what else needs doing to this PR before it's mergeable. I am somewhat confused ast o why a font file is included in this PR, and I would like to be certain on the licence thereof before merging. In addition, it's unlikely that that font file would actually be used by the standalone svg unless I've missed something?

marchersimon:
Regarding the font: The [official GitHub page](https://github.com/dharmatype/Sometype-Mono) says that it is licensed under the Open Font License and its [FAQ](https://scripts.sil.org/cms/scripts/page.php?item_id=OFL-FAQ_web#c26665c2) state:

> Question: 1.1 Can I use the fonts for a book or other print publication, to create logos or other graphics or even to manufacture objects based on their outlines?

> Answer: Yes. You are very welcome to do so. Authors of fonts released under the OFL allow you to use their font software as such for any kind of design work. No additional license or permission is required, unlike with some other licenses. Some examples of these uses are: logos, posters, business cards, stationery, video titling, signage, t-shirts, personalised fabric, 3D-printed/laser-cut shapes, sculptures, rubber stamps, cookie cutters and lead type.

sbrl:
Hmm yeah, that seems the best course of action @marchersimon. Looks like there's also a conflict with `images/banner.svg`, so probably best to just merge the font file (though whether it would actually be used in the svgs is another question entirely - generally I usually suggest using Ctrl + Shift + C in inkscape to turn text into a path to the font file isn't needed. In this case it might still serve as a useful reference for someone though.

marchersimon (approved):

sbrl (approved):
Looks good to merge. Thanks everyone!

PR #5344, opened 2021-03-03
if: actually use if to test a command's success

ericbn:
Using `{{command}} && {{command1}} || {{command2}}` does not actually work like if. The difference is that in the above expression, command2 gets executed if command1 fails, which does not happen in a real if as in `if {{command}}; then {{command1}}; else {{command2}}; fi`

### DIFF
--- a/pages/common/if.md
+++ b/pages/common/if.md
@@ -4,7 +4,7 @@
 
 - Echo a different thing depending on a command's success:
 
-`{{command}} && echo "success" || echo "failure"`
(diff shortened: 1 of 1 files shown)


Yutyo (approved):
I think it looks OK now.

LGTM!

bl-ue (approved):
LGTM for the moment.  This page is one of the oldest but least touched so it needs a refresh.  I'll do that in a new PR.

PR #5347, opened 2021-03-04
arch: create macOS page and add architecture flags

jakevossen5:

### DIFF
--- a/pages/osx/arch.md
+++ b/pages/osx/arch.md
@@ -1,0 +1,12 @@
+# arch
+
+> Display the name of the system architecture, or run a command under a different architecture.
+> See also `uname`.
+
+- Display the system's architecture:
+
+`arch`
+
+- Run a command using x86_64:
+
+`arch -x86_64 {{command}}`


jakevossen5:
`arch` on macOS is different than on other systems, and the `-x86_64` option is only on macOS.

bl-ue:
Strange. Is the `-x86_64` only for M1 Macs or something?

bl-ue (approved):
Strange. Is the `-x86_64` only for M1 Macs or something?

jakevossen5:
Yeah it is kinda weird, seems like Apple just hijacked `arch` to do a bunch of other stuff. Back when 32 bit was supported, you where able to add different flags to run it in 32 bit or 64 bit mode. You can also set environment variables and a whole bunch of other stuff if you look at the man pages, but I am pretty sure 99% of people typing `tldr arch` is going to be because they are on M1 and want to run something using x86_64

einverne (approved):
Looks good.

PR #5351, opened 2021-03-04
if: modernize

bl-ue:

### DIFF
--- a/pages/common/if.md
+++ b/pages/common/if.md
@@ -19,7 +19,7 @@
 
 `if [[ -d {{path/to/directory}} ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
 
-Check if a file or directory ([e]ntry) exists:
+Check if a file or directory exists:
 
 `if [[ -e {{path/to/file_or_directory}} ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
 
--- a/pages/common/if.md
+++ b/pages/common/if.md
@@ -19,7 +19,7 @@
 
 `if [[ -d "{{path/to/directory}}" ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
 
-Check if a file or directory exists:
+- Check if a file or directory exists:
 
 `if [[ -e "{{path/to/file_or_directory}}" ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
 
--- a/pages/common/if.md
+++ b/pages/common/if.md
@@ -5,7 +5,7 @@
 
 - Executed two different commands based on a condition:
 
-`if {{condition}}; then {{echo "true"}}; else {{echo "false"}}; fi`
+`if {{command}}; then {{echo "true"}}; else {{echo "false"}}; fi`
 
 - Check if a variable is defined:
 
--- a/pages/common/if.md
+++ b/pages/common/if.md
@@ -3,25 +3,25 @@
 > Simple shell conditional.
 > See also: `test`, `[`.
 
-- Executed two different commands based on a condition:
+- Execute two different commands based on a condition:
 
 `if {{command}}; then {{echo "true"}}; else {{echo "false"}}; fi`
 
 - Check if a variable is defined:
 
-`if [[ -n {{$VARIABLE}} ]]; then {{echo "defined"}}; else {{echo "not defined"}}; fi`
+`if [[ -n "{{$VARIABLE}}" ]]; then {{echo "defined"}}; else {{echo "not defined"}}; fi`
 
 - Check if a file exists:
 
-`if [[ -f {{path/to/file}} ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
+`if [[ -f "{{path/to/file}}" ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
 
 - Check if a directory exists:
 
-`if [[ -d {{path/to/directory}} ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
+`if [[ -d "{{path/to/directory}}" ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
 
 Check if a file or directory exists:
 
-`if [[ -e {{path/to/file_or_directory}} ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
+`if [[ -e "{{path/to/file_or_directory}}" ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
 
 - List all possible conditions (`test` is an alias to `[`; both are commonly used with `if`):
 
--- a/pages/common/if.md
+++ b/pages/common/if.md
@@ -25,4 +25,4 @@ Check if a file or directory exists:
 
 - List all possible conditions (`test` is an alias to `[`; both are commonly used with `if`):
 
-`help test`
+`man [`
--- a/pages/common/if.md
+++ b/pages/common/if.md
@@ -9,7 +9,7 @@
 
 - Check if a variable is defined:
 
-`if [[ ! -z {{$VARIABLE}} ]]; then {{echo "defined"}}; else {{echo "not defined"}}; fi`
+`if [[ -n {{$VARIABLE}} ]]; then {{echo "defined"}}; else {{echo "not defined"}}; fi`
 
 - Check if a file exists:
 
--- a/pages/common/if.md
+++ b/pages/common/if.md
@@ -1,27 +1,28 @@
 # if
 
 > Simple shell conditional.
+> See also: `test`, `[`.
 
-- Echo a different thing depending on a command's success:
+- Executed two different commands based on a condition:
 
-`if {{command}}; then echo "success"; else echo "failure"; fi`
+`if {{condition}}; then {{echo "true"}}; else {{echo "false"}}; fi`
 
-- Full if syntax:
+- Check if a variable is defined:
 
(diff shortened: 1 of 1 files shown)


pages/common/if.md
```diff
@@ -1,27 +1,28 @@
...
+- Check if a directory exists:
 
-`if [[ -z $GIT_BRANCH ]]; then echo "true"; else echo "false"; fi`
+`if [[ -d {{path/to/directory}} ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
 
-- Test if a file exists:
+Check if a file or directory ([e]ntry) exists:
 
-`if [[ -e {{filename}} ]]; then echo "true"; else echo "false"; fi`
+`if [[ -e {{path/to/file_or_directory}} ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
```
ericbn, line 24: I think the`-e` here refers to "exists", more than to "entry"...

pages/common/if.md
```diff
@@ -1,27 +1,28 @@
...
 
-- If directory not exists:
+- List all possible conditions (`test` is an alias to `[`; both are commonly used with `if`):
 
-`if [[ ! -d {{path/to/directory}} ]]; then echo "true"; else echo "false"; fi`
+`help test`
```
ericbn, line 28: `help` is specific to bash. I suggest using `man [` here.

pages/common/if.md
bl-ue: Hmm...you must be right 👍🏻

pages/common/if.md
```diff
@@ -1,27 +1,28 @@
...
+`if [[ -f {{path/to/file}} ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
 
-- Test if a given variable is empty:
+- Check if a directory exists:
 
-`if [[ -z $GIT_BRANCH ]]; then echo "true"; else echo "false"; fi`
+`if [[ -d {{path/to/directory}} ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
 
-- Test if a file exists:
+Check if a file or directory ([e]ntry) exists:
```
bl-ue, line 22:
```suggestion
Check if a file or directory exists:
```

pages/common/if.md
```diff
@@ -1,27 +1,28 @@
 # if
 
 > Simple shell conditional.
+> See also: `test`, `[`.
 
-- Echo a different thing depending on a command's success:
+- Executed two different commands based on a condition:
 
-`if {{command}}; then echo "success"; else echo "failure"; fi`
```
ericbn, line 7: I would keep this, to show that if can be used with any command, not only with `[` or `test`...

pages/common/if.md
bl-ue: Oh yeah, you're right about that!  I was going to do something about that but I was distracted...:smile:
```suggestion
`man [`
```
My shell `zsh` doens't even have help :P

pages/common/if.md
bl-ue: Well, I removed it because it's nearly a duplicate of the next example :/

pages/common/if.md
```diff
@@ -1,27 +1,28 @@
...
+`if {{condition}}; then {{echo "true"}}; else {{echo "false"}}; fi`
 
-- Full if syntax:
+- Check if a variable is defined:
 
-`if {{condition}}; then echo "true"; else echo "false"; fi`
+`if [[ ! -z {{$VARIABLE}} ]]; then {{echo "defined"}}; else {{echo "not defined"}}; fi`
```
ericbn, line 12: `! -z` won't test if a variable is defined, but more specifically if the the length of the variable value is non-zero. This is the same as using `-n`, which I suggest should be used instead.

bl-ue: I didn't know about `-n`. 👍🏻   Yes, I know that it doesn't actually check if it's defined, but to a user, what's the difference between a variable that's not defined and a variable that has no value?
```suggestion
`if [[ -n {{$VARIABLE}} ]]; then {{echo "defined"}}; else {{echo "not defined"}}; fi`
```

pages/common/if.md
ericbn: Then maybe just keep this one. `if` does not run on a "condition", but rather on any command (built-in or not)...

bl-ue: Ah good point.  I should have tested that 😄

bl-ue: Fixed https://github.com/tldr-pages/tldr/pull/5351/commits/c46b7b93d6fcd420e1ce6670bc4d07ebe3865f54

pages/common/if.md
ericbn: A variable can be defined or not, and have an empty value or not. These are two different things for a shell. This works in zsh:
```
% unset FOO
% if (( ${+FOO} )); then echo "FOO is defined"; else echo "FOO is not defined"; fi
FOO is not defined
% FOO=
% if (( ${+FOO} )); then echo "FOO is defined"; else echo "FOO is not defined"; fi
FOO is defined
% if [[ -n ${FOO} ]]; then echo "FOO is nonempty"; else echo "FOO is empty"; fi
FOO is empty
% FOO=1
% if [[ -n ${FOO} ]]; then echo "FOO is nonempty"; else echo "FOO is empty"; fi
FOO is nonempty
```
I know it's tricky...  :- )

pages/common/if.md
```diff
@@ -1,27 +1,28 @@
...
+- Executed two different commands based on a condition:
 
-`if {{command}}; then echo "success"; else echo "failure"; fi`
+`if {{command}}; then {{echo "true"}}; else {{echo "false"}}; fi`
 
-- Full if syntax:
+- Check if a variable is defined:
 
-`if {{condition}}; then echo "true"; else echo "false"; fi`
+`if [[ -n {{$VARIABLE}} ]]; then {{echo "defined"}}; else {{echo "not defined"}}; fi`
```
ericbn, line 12: A variable can be defined or not, and have an empty value or not. These are two different things for a shell. This works in zsh:
```
% unset FOO
% if (( ${+FOO} )); then echo "FOO is defined"; else echo "FOO is not defined"; fi
FOO is not defined
% FOO=
% if (( ${+FOO} )); then echo "FOO is defined"; else echo "FOO is not defined"; fi
FOO is defined
% if [[ -n ${FOO} ]]; then echo "FOO is nonempty"; else echo "FOO is empty"; fi
FOO is empty
% FOO=1
% if [[ -n ${FOO} ]]; then echo "FOO is nonempty"; else echo "FOO is empty"; fi
FOO is nonempty
```
I know it's tricky...  :- )

bl-ue:
Thanks for the review @ericbn! 🙂

navarroaxel (approved):

pages/common/if.md
```diff
@@ -1,27 +1,28 @@
...
 
-- List available if conditions:
+- Check if a file exists:
 
-`help test`
+`if [[ -f {{path/to/file}} ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
```
sbrl, line 16: Same here.
```suggestion
`if [[ -f "{{path/to/file}}" ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
```

pages/common/if.md
```diff
@@ -1,27 +1,28 @@
...
 
-- Test if a given variable is empty:
+- Check if a directory exists:
 
-`if [[ -z $GIT_BRANCH ]]; then echo "true"; else echo "false"; fi`
+`if [[ -d {{path/to/directory}} ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
```
sbrl, line 20: Same here.
```suggestion
`if [[ -d "{{path/to/directory}}" ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
```

pages/common/if.md
```diff
@@ -1,27 +1,28 @@
...
 
-- Test if a file exists:
+Check if a file or directory exists:
 
-`if [[ -e {{filename}} ]]; then echo "true"; else echo "false"; fi`
+`if [[ -e {{path/to/file_or_directory}} ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
```
sbrl, line 24: and here.
```suggestion
`if [[ -e "{{path/to/file_or_directory}}" ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
```

sbrl (commented):

sbrl (approved):
Thanks, @bl-ue!

sbrl:
That's not a very helpful error message, @tldr-bot.

pages/common/if.md
```diff
@@ -1,27 +1,28 @@
...
+`if [[ -f "{{path/to/file}}" ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
 
-- Test if a given variable is empty:
+- Check if a directory exists:
 
-`if [[ -z $GIT_BRANCH ]]; then echo "true"; else echo "false"; fi`
+`if [[ -d "{{path/to/directory}}" ]]; then {{echo "true"}}; else {{echo "false"}}; fi`
 
-- Test if a file exists:
+Check if a file or directory exists:
```
bl-ue, line 22:
```suggestion
- Check if a file or directory exists:
```

bl-ue (commented):
@sbrl, indeed no it isn't! But this is why:

sbrl (approved):

PR #5355, opened 2021-03-06
deluser: sync translations with PR #5245

bl-ue:
Syncs the Italian and Korean translations of `common/deluser.md` with the English translation, see PR #5245.

cc @TheLastGimbus

### DIFF
--- a/pages.it/linux/deluser.md
+++ b/pages.it/linux/deluser.md
@@ -1,6 +1,7 @@
 # deluser
 
 > Rimuovi un account utente o un utente da un gruppo.
+> Maggiori informazioni: <https://manpages.debian.org/adduser/deluser.8.html>.
 
 - Rimuovi un utente:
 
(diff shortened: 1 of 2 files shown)


sbrl (commented):
Ummm all I see are 2 deleted files? If the translations are updated, then there's really no need to delete these. In fact, [the translations dashboard](https://lukwebsforge.github.io/tldri18n/) should mark them ass outdated if the number of examples are different.

bl-ue:
Yeah, `common/deluser.md` was moved to `linux/deluser.md`, but it had examples added.  I thought the best way to settle the situation was to just delete them, then if somebody tried to view the page in their language they'd see a not found and maybe submit a translation, rather than see an outdated page but think it's up to date, etc.. But instead I'll restore them, move them to `linux/`, and add what I can: the more information link :blush:

navarroaxel (approved):

sbrl (changes requested):
Requesting changes to prevent accidental merge until @bl-ue applies changes.

navarroaxel (approved):

sbrl (approved):
Thanks, @bl-ue!

PR #5362, opened 2021-03-06
arch, bzip2, command, gunicorn, history, htop, shasum, stty, yes: move translation to correct directories

marchersimon:
I saw in [this table](https://lukwebsforge.github.io/tldri18n/), that a few translations are in `linux` or `osx`, instead of `common`. 
I just don't know what to do with the chinese (zh) version of `arch.md`, since there are different versions of the same file in `common` and `linux`.


bl-ue (approved):
**Great** catch!  I wouldn't have thought of doing _this_. Earlier today I submitted some PRs to add more ✅s but I didn't think that actually a lot of contributors might accidently put translations in the wrong place.

I'll take a look at `arch` later.  I know there's something confusing about it.

navarroaxel (approved):

PR #5363, opened 2021-03-06
German pages: fix (valid) tldr-lint errors

bl-ue:
**(valid)** because `TLDR015` and `TLDR104`, which are found a lot on the German pages, don't apply to any language but English.


navarroaxel (approved):

sbrl (approved):
Thanks, @bl-ue!

PR #5368, opened 2021-03-07
musescore: add page

marchersimon:
There's also `mscore`, which is identical to `musescore`. Is the alias system already in place or should I just add a new seperate page?

### DIFF
--- a/pages/common/musescore.md
+++ b/pages/common/musescore.md
@@ -19,7 +19,7 @@
 
 `musescore --experimental`
 
-- Export the given (or currently opened) file to the specified output file. The file type depends on the extension given:
+- Export the given file to the specified output file. The file type depends on the extension given:
 
 `musescore --export-to {{output_file}} {{input_file}}`
 
--- a/pages/common/musescore.md
+++ b/pages/common/musescore.md
@@ -19,14 +19,14 @@
 
 `musescore --experimental`
 
-- Export the given file to the specified output file. The file type depends on the extension given:
+- Export the given file to the specified output file. The file type depends on the given extension:
 
 `musescore --export-to {{output_file}} {{input_file}}`
 
 - Print a diff between the given scores:
 
-`musescore --diff {{file1}} {{file2}}`
+`musescore --diff {{path/to/file1}} {{path/to/file2}}`
 
 - Specify a MIDI import operations file:
 
-`musescore --midi-operations {{file}}`
+`musescore --midi-operations {{path/to/file}}`
--- a/pages/common/musescore.md
+++ b/pages/common/musescore.md
@@ -23,7 +23,7 @@
 
 `musescore --export-to {{output_file}} {{input_file}}`
 
-- Print a conditioned `diff` between the given scores:
+- Print a diff between the given scores:
 
 `musescore --diff {{file1}} {{file2}}`
 
--- a/pages/common/musescore.md
+++ b/pages/common/musescore.md
@@ -1,0 +1,32 @@
+# musescore
+
+> MuseScore 3 sheet music editor.
+> More information: <https://musescore.org/en/handbook/3/command-line-options>.
+
+- Use a specific audio driver:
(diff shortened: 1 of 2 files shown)


bl-ue:
Great question!  I think this is a perfect place to test out the alias system!  Go ahead and add:

`pages/common/mscore.md`
```markdown
# mscore

> This command is an alias of `musescore`.

`tldr musescore`
```

pages/common/musescore.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Start MuseScore in debug mode:
+
+`musescore --debug`
+
+- Enable experimental features, such as layers:
+
+`musescore --experimental`
+
+- Export the given (or currently opened) file to the specified output file. The file type depends on the extension given:
```
bl-ue, line 22: "currently opened" &mdash; how does this work with the CLI?

pages/common/musescore.md
```diff
@@ -0,0 +1,32 @@
...
+
+`musescore --export-to {{output_file}} {{input_file}}`
+
+- Print a conditioned `diff` between the given scores:
```
bl-ue, line 26: What do you mean by 'conditioned diff'?

pages/common/musescore.md
marchersimon: At least that's what was at the man page. But when I tested it it doesn't work without an input file. I'll just delete that part

pages/common/musescore.md
marchersimon: `--diff` prints it in a more readable way in contrast to `--raw-diff`. Should I just delete that word or do you have a better alterantive?

bl-ue: I would just removed 'conditioned'.  I don't think it matters much to get every detail in.  Also, I think you remove backticks from 'diff' because, it's not even that much of a technical word, much less a command.

bl-ue (approved):
Fun to test the alias system!  I can't wait to try `tldr mscore`! 💛

pages/common/musescore.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Start MuseScore in debug mode:
+
+`musescore --debug`
+
+- Enable experimental features, such as layers:
+
+`musescore --experimental`
+
+- Export the given file to the specified output file. The file type depends on the extension given:
```
navarroaxel, line 22: Is this sentence wrong? @bl-ue 
```suggestion
- Export the given file to the specified output file. The file type depends on the given extension:
```

pages/common/musescore.md
```diff
@@ -0,0 +1,32 @@
...
+
+`musescore --export-to {{output_file}} {{input_file}}`
+
+- Print a diff between the given scores:
+
+`musescore --diff {{file1}} {{file2}}`
+
+- Specify a MIDI import operations file:
+
+`musescore --midi-operations {{file}}`
```
navarroaxel, line 32:
```suggestion
`musescore --midi-operations {{path/to/file}}`
```

pages/common/musescore.md
```diff
@@ -0,0 +1,32 @@
...
+
+`musescore --experimental`
+
+- Export the given file to the specified output file. The file type depends on the extension given:
+
+`musescore --export-to {{output_file}} {{input_file}}`
+
+- Print a diff between the given scores:
+
+`musescore --diff {{file1}} {{file2}}`
```
navarroaxel, line 28:
```suggestion
`musescore --diff {{path/to/file1}} {{path/to/file2}}`
```

pages/common/musescore.md
bl-ue: @navarroaxel no, both versions are correct and convey the exact same thing.  It doesn't matter which one we use.  I suppose we should pick the one that's easier for you to read because that likely means easier for many others.

navarroaxel (approved):

PR #5369, opened 2021-03-07
Portuguese (Brazil) pages: fix (valid) tldr-lint errors

bl-ue:
**(valid)** because `TLDR109`, which is only erroring on `eyeD3.md`, I would like to fix, along with other offenders of that rule, in a separate PR.


navarroaxel (approved):

sbrl (approved):

bl-ue:
What happened here? :smile:

PR #5376, opened 2021-03-07
Korean pages: fix (valid) tldr-lint errors

bl-ue:
**(valid)** because `TLDR015` and `TLDR104`, which are found a lot on the Korean pages, don't apply to any language but English.


navarroaxel (approved):

bl-ue:
When you look at the files view on PRs like this, the pages really look like a pile of garbage haha :smile:

sbrl (approved):
This looks ok, but note that Korean uses a completely different character set to Japanese.

sbrl:
This PR has conflicting files, so those will need to be addressed before we can merge.

bl-ue:
>This looks ok, but note that Korean uses a completely different character set to Japanese.

Oops! I forgot to updated my copy/pasted description

>This PR has conflicting files, so those will need to be addressed before we can merge.

I'll fix those and merge.

PR #5386, opened 2021-03-07
aurman: add page

navarroaxel:

### DIFF
--- a/pages/linux/aurman.md
+++ b/pages/linux/aurman.md
@@ -32,6 +32,6 @@
 
 `aurman --remove --recursive --nosave {{package_name}}`
 
-- Remove not installed packages and unused repositories from the cache (use two `--clean` flags to clean all packages):
+- Clear the package cache (use two `--clean` flags to clean all packages):
 
 `aurman --sync --clean`
--- a/pages/linux/aurman.md
+++ b/pages/linux/aurman.md
@@ -8,7 +8,7 @@
 
 `aurman --sync --refresh --sysupgrade`
 
-- Synchronize and update all packages without show changes of PKGBUILD files:
+- Synchronize and update all packages without show changes of `PKGBUILD` files:
 
 `aurman --sync --refresh --sysupgrade --noedit`
 
@@ -16,7 +16,7 @@
 
 `aurman --sync {{package_name}}`
 
-- Install a new package without show changes of PKGBUILD files:
+- Install a new package without show changes of `PKGBUILD` files:
 
 `aurman --sync --noedit {{package_name}}`
 
--- a/pages/linux/aurman.md
+++ b/pages/linux/aurman.md
@@ -1,0 +1,37 @@
+# aurman
+
+> An Arch Linux utility to build and install packages from the Arch User Repository.
+> See also `pacman`.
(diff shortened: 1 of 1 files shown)


pages/linux/aurman.md
```diff
@@ -0,0 +1,37 @@
...
+
+> An Arch Linux utility to build and install packages from the Arch User Repository.
+> See also `pacman`.
+> More information: <https://github.com/polygamma/aurman>.
+
+- Synchronize and update all packages:
+
+`aurman --sync --refresh --sysupgrade`
+
+- Synchronize and update all packages without show changes of PKGBUILD files:
```
bl-ue, line 11:
```suggestion
- Synchronize and update all packages without show changes of `PKGBUILD` files:
```

pages/linux/aurman.md
```diff
@@ -0,0 +1,37 @@
...
+
+`aurman --sync --refresh --sysupgrade --noedit`
+
+- Install a new package:
+
+`aurman --sync {{package_name}}`
+
+- Install a new package without show changes of PKGBUILD files:
```
bl-ue, line 19:
```suggestion
- Install a new package without show changes of `PKGBUILD` files:
```

bl-ue (commented):
Don't you think we should backtick 'PKGBUILD'?

bl-ue (approved):

pages/linux/aurman.md
```diff
@@ -0,0 +1,37 @@
...
+
+- Search the package database for a keyword from the official repositories and AUR:
+
+`aurman --sync --search {{keyword}}`
+
+- Remove a package and its dependencies:
+
+`aurman --remove --recursive --nosave {{package_name}}`
+
+- Remove not installed packages and unused repositories from the cache (use two `--clean` flags to clean all packages):
```
sbrl, line 35: `Remove not installed packages` doesn't sounds very intuitive at first here. Is there another way we could phrase this?

sbrl (approved):
Just a small suggestion :-)

pages/linux/aurman.md
marchersimon: @navarroaxel How about `Remove packages that are no longer installed from the cache as well as currently unused sync databases`? That's what the page referred to in `aurman --help` says.

navarroaxel: Close, but not exactly the AUR packages are git repositories with a `PKGBUILD` file to build your own package. So `aurman` keeps some builds and that cloned Git repos in `.cache/aurman/`, so this command:
- Remove AUR cloned Git repositories from `aurman` cache for packages that are no longer installed.
- Remove downloaded packages that are no longer installed from the `pacman` cache.
- Also, the unused sync databases are removed but is not important as the previous ones.

bl-ue: What about something else, like `Clear the package cache`?

navarroaxel: Why not, is a more general description but we can explain every step of a command.

sbrl: Sounds good!

marchersimon (approved):

PR #5388, opened 2021-03-09
pgbench: add page

andyatkinson:
Primarily based on this CloudBees blog post:
<https://www.cloudbees.com/blog/tuning-postgresql-with-pgbench/>

And the PG 10 pgbench documentation:
<https://www.postgresql.org/docs/10/pgbench.html>

### DIFF
--- a/pages/common/pgbench.md
+++ b/pages/common/pgbench.md
@@ -5,7 +5,7 @@
 
 - Initialize database example with scaling option of 50 times the default size:
 
-`pgbench -i -s 50 example`
+`pgbench --initialize --scale={{50}} {{database_name}}`
 
 - Benchmark with 10 clients, 2 worker threads, and 10,000 transactions per client:
 
--- a/pages/common/pgbench.md
+++ b/pages/common/pgbench.md
@@ -1,0 +1,12 @@
+# pgbench
+
+> pgbench — run a benchmark test on PostgreSQL
+> More information: <https://www.postgresql.org/docs/10/pgbench.html>.
+
+- Initialize database called example. Use scaling option of 50 times the default size.
+
+`pgbench -i -s 50 example`
+
+- Benchmark with 10 clients, 2 worker threads, and 10,000 transactions per client
+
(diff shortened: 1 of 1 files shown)


pages/common/pgbench.md
```diff
@@ -0,0 +1,12 @@
+# pgbench
+
+> Run a benchmark test on PostgreSQL.
+> More information: <https://www.postgresql.org/docs/10/pgbench.html>.
+
+- Initialize database example with scaling option of 50 times the default size:
+
+`pgbench -i -s 50 example`
```
bl-ue, line 8:
```suggestion
`pgbench --initialize --scale={{50}} {{database_name}}`
```

bl-ue (changes requested):
Thanks for the new page @andyatkinson!  I've left some comments below for you to review.

In general:
1. We use long options in examples when possible
2. User-entered flag/argument values get wrapped in our custom token syntax (`{{` and `}}`)
3. We try to use more generic names for objects, e.g. `path/to/file` and `database_name`, rather than `myfile.txt` or `example`.

andyatkinson:
@bl-ue Those are excellent suggestions. This project had great contribution guidelines and workflow around those contributions. The long options I read about in the guidelines but wasn't sure if they applied here, although seeing them written out as the suggestions, I like them better too.

I somehow missed the custom token syntax, but that's great. 

Using `database_name` also makes sense here as it communicates more to the reader than `example`. I actually was confused myself using `pgbench` the first time running the command, and realizing I needed to first create a database named `example`.

pages/common/pgbench.md
```diff
@@ -0,0 +1,12 @@
+# pgbench
+
+> Run a benchmark test on PostgreSQL.
+> More information: <https://www.postgresql.org/docs/10/pgbench.html>.
+
+- Initialize database example with scaling option of 50 times the default size:
```
bl-ue, line 6:
```suggestion
- Initialize a database with scaling option of 50 times the default size:
```

marchersimon (approved):
Looks good to me

PR #5392, opened 2021-03-09
Add script for finding undocumented pages more easily

marchersimon:
I have made a little bash script, which creates a text file with every currently installed command (at least all in `$PATH`) which was not found in `pages/`. It's far from perfect, but at least (for me) it works. I thought about sorting the list by relevance by taking `.bashrc` into consideration, but I'm not sure how I'd do that.

I also added the output file to `.gitignore` and referenced to the script in `CONTRIBUTING.md`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,7 @@ scripts/pdf/tldr-pages.pdf
 # Python venv for testing the PDF script
 # Create it with: python3 -m venv scripts/pdf/venv/
 scripts/pdf/venv/
+
+# List of undocumented commands
+undocumented.txt
+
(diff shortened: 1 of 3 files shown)


sbrl (commented):
Hey there, @marchersimon! Looks like a great script. How does this compare to my earlier attempt on the useful scripts and tools page, [`tldr-missing-pages`](https://github.com/tldr-pages/tldr/wiki/Useful-scripts-and-programs#find-pages-that-dont-exist)?

marchersimon:
Oh, I didn't know about that page. Your script seems to be more useful, since it searched through the history too.

PR #5406, opened 2021-03-10
pio-home: add page

marchersimon:

PR #5407, opened 2021-03-10
plattformio-home: add alias page

marchersimon:
`platformio-home` is an alias of `pio-home` (#5406)

### DIFF
--- a/pages/common/platformio-home.md
+++ b/pages/common/platformio-home.md
@@ -1,6 +1,7 @@
 # platformio home
 
 > This command is an alias of `pio home`.
+
 - View documentation for the original command:
 
 `tldr pio-home`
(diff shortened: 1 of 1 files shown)


sbrl:
Hrm, this is an interesting case. I'm on the fence about this actually, in that I'm somewhat concerned about the large number of aliases as @bl-ue mentions.

I would guess that you'd try `tldr platformio` if a subcommand doesn't work though, and then you'd see the alias message? I'm not sure I'm convinced that aliases for _every_ subcommand of an aliased parent command are really necessary.

marchersimon:
@sbrl I also think that would be too much. Just having the `platformio` alias seems fine to me.

sbrl:
I'm closing this PR, but we can always reopen if there's more to discuss.

PR #5413, opened 2021-03-11
dig: add page

stanulilic:

### DIFF
--- a/pages/linux/dig.md
+++ b/pages/linux/dig.md
@@ -1,0 +1,24 @@
+# dig
+
+> Domain Information Groper: queries and displays information from DNS nameservers.
+> More information: <https://manpages.debian.org/buster/dnsutils/dig.1.en.html>.
+
+- Lookup IP address(A records)  for a given domain. It displays technical details alongside the IP address:
+
+`dig {{example.com}}`
+
+- Get the IP addresses for a given domain only:
+
+`dig {{example.com}} +short`
+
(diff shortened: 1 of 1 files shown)


einverne (changes requested):
@stanulilic  Thank you for the new page, but the `dig` command is listed in the common directory.

bl-ue:
Wow, this is a nice page @stanulilic! In many ways, it's nicer than our current dig, but unfortunately, as @einverne (and @tldr-bot) pointed out, we already have `dig.md` [here](https://github.com/tldr-pages/tldr/blob/2c0be243475e7206df984515d471dd9a6fa44842/pages/common/dig.md).

However, there are many qualities of this page (explained acronym, more info link, extra examples) that the old page doesn't have, so my suggestion is: how about you close this PR, and open a new, merging this excellent page with that one, to the best of your ability?

Mostly I'm concerned about have as many examples as possible, and keeping as much data (acronym exp., link, etc.) as possible.

What do you think?

stanulilic:
Thanks for your responses.  I missed the existence of the `dig.md` file in the common directory when I was writing the new page. I only checked in the Linux directory as I mistakenly assumed that all the Linux commands live there.  I am new to this project, and also new to contributing to open source, so I appreciate your patience with me. 

Am very humbled and grateful that you liked my page. I will try to merge my page with the existing page and keep all qualities of the page that you like. 

Thank you very much in advance!

PR #5415, opened 2021-03-11
shift: add page

stku1985:

### DIFF
--- a/pages/common/shift.md
+++ b/pages/common/shift.md
@@ -1,7 +1,7 @@
 # shift
 
 > Bash built-in command that shifts the arguments passed to the calling function or script by a specified number of places.
-> More information: <https://www.gnu.org/software/bash/manual/bash.html>.
+> More information: <https://www.gnu.org/software/bash/manual/bash.html#index-shift>.
 
 - Move arguments by one place dropping the first argument:
 
--- a/pages/common/shift.md
+++ b/pages/common/shift.md
@@ -1,7 +1,7 @@
 # shift
 
 > Bash built-in command that shifts the arguments passed to the calling function or script by a specified number of places.
-> See <https://www.gnu.org/software/bash/manual/bash.html> for more information.
+> More information: <https://www.gnu.org/software/bash/manual/bash.html>.
 
 - Move arguments by one place dropping the first argument:
 
--- a/pages/common/shift.md
+++ b/pages/common/shift.md
@@ -1,0 +1,12 @@
+# shift
+
+> A bash built-in command that shifts the (numbered) arguments of a function or script by a specified number of places.
+> See <https://www.gnu.org/software/bash/manual/bash.html> for more information.
+
+- Move arguments by one place dropping the first argument:
+
+`shift`
+
+- Move arguments by {{n}} places dropping the first {{n}} arguments:
+
+`shift {{n}}`
--- a/pages/common/shift.md
+++ b/pages/common/shift.md
@@ -1,6 +1,6 @@
 # shift
 
-> A bash built-in command that shifts the (numbered) arguments of a function or script by a specified number of places.
+> Bash built-in command that shifts the arguments passed to the calling function or script by a specified number of places.
 > See <https://www.gnu.org/software/bash/manual/bash.html> for more information.
 
 - Move arguments by one place dropping the first argument:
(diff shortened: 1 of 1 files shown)


bl-ue:
@stku1985 we only use single-line examples, enclosed in single back ticks (\`...\` rather than \`\`\`...\`\`\`)

pages/common/shift.md
```diff
@@ -0,0 +1,22 @@
...
+> See <https://www.gnu.org/software/bash/manual/bash.html> for more information.
+
+- Shift arguments by one place; argument `$1` is dropped, argument `$2` is now `$1`:
+
+```bash
+# prints two times the same output
+echo $2
+shift
+echo $1 
+```
```
bl-ue, line 13: @stku1985 I'd do something like this:
```suggestion
- Move all arguments back by one (dropping the first argument):

`shift`
```

stku1985:
@bl-ue thanks for the quick reply. I understand that single line code examples are a community standard; however since `shift` is strictly for scripting, the examples need some context imo.

What is your suggestion?

pages/common/shift.md
stku1985: Ok I will do so. I think this reduces the intelligibility of the page, but I also understand that you want to keep consistency between pages.

bl-ue:
@stku1985 if you can, please avoid force-pushing.  It's not necessary as we automatically squash commits before merging and it can make reviewing a bit difficult 🙂

pages/common/shift.md
bl-ue: Why does it decrease the intelligibility @stku1985?

stku1985:
@bl-ue I will keep that in mind for the next time. Sorry for the confusion.

pages/common/shift.md
```diff
@@ -0,0 +1,12 @@
+# shift
+
+> A bash built-in command that shifts the (numbered) arguments of a function or script by a specified number of places.
```
bl-ue, line 3:
```suggestion
> Bash built-in command that shifts the arguments passed to the calling function or script by a specified number of places.
```

pages/common/shift.md
```diff
@@ -0,0 +1,12 @@
...
+> See <https://www.gnu.org/software/bash/manual/bash.html> for more information.
+
+- Move arguments by one place dropping the first argument:
+
+`shift`
+
+- Move arguments by {{n}} places dropping the first {{n}} arguments:
+
+`shift {{n}}`
```
bl-ue, line 12: We don't typically use token syntax inside example descriptions.  I'd do this, in this case:
```suggestion
- Move arguments by N places dropping the first N arguments:

`shift {{N}}`
```

pages/common/shift.md
```diff
@@ -0,0 +1,12 @@
+# shift
+
+> A bash built-in command that shifts the (numbered) arguments of a function or script by a specified number of places.
+> See <https://www.gnu.org/software/bash/manual/bash.html> for more information.
```
bl-ue, line 4: We have a standard syntax for more information links:
```suggestion
> More information: <https://www.gnu.org/software/bash/manual/bash.html>.
```

bl-ue (changes requested):
Thanks for the new page @stku1985!  I'm really surprised this one wasn't already added 😄

I've left a few comments below.

pages/common/shift.md
stku1985: Well I think scripting commands (such as shift, return, continue, for, ...) need some context. For experienced bash users this short example is fine; for people new to bash scripting my original version or even a whole function example would be easier to grasp. 

But don't worry; either way is fine from my point of view.

bl-ue: I agree so much with that—but that's not what tldr-pages was originally invented for.  It's really for documented syntax of a command, not really how to use it in conjunction with other commands.

bl-ue (approved):
Lovely and much needed! 😋

pages/common/shift.md
```diff
@@ -0,0 +1,12 @@
+# shift
+
+> Bash built-in command that shifts the arguments passed to the calling function or script by a specified number of places.
+> More information: <https://www.gnu.org/software/bash/manual/bash.html>.
```
marchersimon, line 4:
```suggestion
> More information: <https://www.gnu.org/software/bash/manual/bash.html#index-shift>.
```

Maybe add an anchor to the link, so the command can be found more easily.

marchersimon (approved):

PR #5419, opened 2021-03-11
bssh, bvnc, bshell: add pages

marchersimon:
This situation is actually a bit complicated. There is only one binary (`bssh`) and the other ones are just links to that one. The default behaviour of the command changes depending under which name it is called. Hower I couldn't just make two alias pages, since the arguments depend on the name it is called. I hope the way I did it is okay.

### DIFF
--- a/pages/common/bshell.md
+++ b/pages/common/bshell.md
@@ -1,0 +1,22 @@
+# bshell
+
+> A GUI tool for browsing for SSH/VNC servers on the local network.
+> See also: `bssh` and `bvnc`.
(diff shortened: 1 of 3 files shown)


bl-ue:
It doesn't matter that the different commands are just links to the same binary as long as the binary functions as a different binary when called by a different name.

They do that with `minikube` and `kubectl` (https://minikube.sigs.k8s.io/docs/handbook/kubectl/).

bl-ue (approved):

sbrl (approved):
Thanks, @marchersimon! Looks great to me :D

PR #5377, opened 2021-03-07
Bump tldr-lint to 0.0.10

bl-ue:
Closes #5255
Closes #5071

Enables the following PRs to be merged:
#5205
#5088

---

This PR is waiting for the following pages to be corrected:
```
pages/common/R.md:0: TLDR109 File name should be lowercase
pages/common/theHarvester.md:0: TLDR109 File name should be lowercase
pages/linux/eyeD3.md:0: TLDR109 File name should be lowercase
pages/osx/GetFileInfo.md:0: TLDR109 File name should be lowercase
pages/osx/SafeEjectGPU.md:0: TLDR109 File name should be lowercase
pages/osx/csshX.md:0: TLDR109 File name should be lowercase
```

Checklist:

- [x] [`R.md`](https://github.com/tldr-pages/tldr/blob/6443e6a7254dd0d9c350be2db0ee9ad7cab2d032/pages/common/R.md) → #5379
- [x] [`theHarvester.md`](https://github.com/tldr-pages/tldr/blob/6443e6a7254dd0d9c350be2db0ee9ad7cab2d032/pages/common/theHarvester.md) → #5380
- [x] [`eyeD3.md`](https://github.com/tldr-pages/tldr/blob/6443e6a7254dd0d9c350be2db0ee9ad7cab2d032/pages/linux/eyeD3.md) → #5381
- [x] [`GetFileInfo.md`](https://github.com/tldr-pages/tldr/blob/6443e6a7254dd0d9c350be2db0ee9ad7cab2d032/pages/osx/GetFileInfo.md) → #5382
- [x] [`SafeEjectGPU.md`](https://github.com/tldr-pages/tldr/blob/6443e6a7254dd0d9c350be2db0ee9ad7cab2d032/pages/osx/SafeEjectGPU.md) → #5383
- [x] [`csshX.md`](https://github.com/tldr-pages/tldr/blob/6443e6a7254dd0d9c350be2db0ee9ad7cab2d032/pages/osx/csshX.md) → #5384

PR #5327, opened 2021-03-01
CLIENT-SPECIFICATION: add lowercasing of commands

MasterOdin:
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).

In preparation of landing tldr-lint@0.0.10 and resolving #5085 and #5255, I figured it would be good to add to the client specification about the lowercasing of file names as a transparent step to end users.

Issue #5184, opened 2021-01-25
Ensure HTTPS is used for definition cache download


aleksandrs-ledovskis:
Basically this is request to re-evaluate findings from https://github.com/tldr-pages/tldr/issues/2253#issuecomment-417184655

When default `tldr` client requests definition cache it does download https://tldr-pages.github.io/assets/tldr.zip

**Problem**: The link has an unsecure/plain HTTP hop which can be seen using cURL:

```
$ curl -vL 'https://tldr-pages.github.io/assets/tldr.zip'
*   Trying 185.199.111.153...
* TCP_NODELAY set
* Connected to tldr-pages.github.io (185.199.111.153) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* SSL stuff...
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0x7ff05000d800)
> GET /assets/tldr.zip HTTP/2
> Host: tldr-pages.github.io 
> User-Agent: curl/7.54.0
> Accept: */*
>
* Connection state changed (MAX_CONCURRENT_STREAMS updated)!
< HTTP/2 301
< content-type: text/html
< server: GitHub.com
< location: http://tldr.sh/assets/tldr.zip
< x-github-request-id: 15D0:08FF:4A88CE:4FC8C2:600F05FC
< accept-ranges: bytes
< date: Mon, 25 Jan 2021 18:14:51 GMT
< via: 1.1 varnish
< age: 1182
< x-served-by: cache-bma1635-BMA
< x-cache: HIT
< x-cache-hits: 1
< x-timer: S1611598492.730806,VS0,VE1
< vary: Accept-Encoding
< x-fastly-request-id: 27a137c450f28daff2f2e5216128860de8798473
< content-length: 162
```

The problem, obviously is `location: http://tldr.sh/assets/tldr.zip` reply header.

As unsecure redirect is advised, the `tldr.zip` file can be trivially MITM'ed and poisoned by unknown party to do some nefarious command suggestion or exploit a (yet unknown) Markdown processing bug. Definition cache file is not cryptographically signed, therefore a malicious file detection on client side is impossible.

[...]

PR #5428, opened 2021-03-12
CLIENT-SPECIFICATION: v1.5

bl-ue:
Accidentally forgotten in #5327 ☹️

Also see: https://github.com/tldr-pages/tldr/pull/5377#issuecomment-794082345.  Can we do it?

### DIFF
--- a/CLIENT-SPECIFICATION.md
+++ b/CLIENT-SPECIFICATION.md
@@ -1,6 +1,6 @@
 # tldr-pages client specification
 
-**Current Specification Version:** 1.4
+**Current Specification Version:** 1.5
 
 This document contains the official specification for tldr-pages clients. It is _not_ a specification of the format of the pages themselves - only a specification of how a user should be able to interface with an official client. For a list of previous versions of the specification, see the [changelog section](#Changelog) below.
 
@@ -211,6 +211,9 @@ Caching SHOULD be done according to the user's language configuration (if any), 
 
 
 ## Changelog
+
+ - [v1.5, March 11th 2021](https://github.com/tldr-pages/tldr/blob/8505e55b568dbc394ea853d488cd07484c051e21/CLIENT-SPECIFICATION.md) ([#5327](https://github.com/tldr-pages/tldr/pull/5327))
+   - Add requirement for converting command names to lowercase before searching for them.
 
  - [v1.4, August 13th 2020](https://github.com/tldr-pages/tldr/blob/87324c6e540c10a44950b14cd9fb7d758ce4d4e0/CLIENT-SPECIFICATION.md) ([#4246](https://github.com/tldr-pages/tldr/pull/4246))
    - Add requirement for CLI clients to use non-zero exit code on failing to find a page.


owenvoke (approved):

MasterOdin:
No matter what, you're going to need two commits to update the changelog appropriately with the link to the commit. As such, I did purposely leave out the version bump in #5327 to be done in a follow-up PR as I'd prefer the commit history of:

* CLIENT-SPECIFICATION: some change
* CLIENT-SPECIFICATION: v1.5

versus

* CLIENT-SPECIFICATION: some change
* CLIENT-SPECIFICATION: fix changelog

MasterOdin (approved):

bl-ue:
>No matter what, you're going to need two commits to update the changelog appropriately with the link to the commit.

Yeah, I know that and I don't see any way around it :\

>I did purposely leave out the version bump in #5327 to be done in a follow-up PR

Ah, I didn't know that. Thank you.

I agree

>I'd prefer the commit history of:
>
>- CLIENT-SPECIFICATION: some change
>- CLIENT-SPECIFICATION: v1.5
>
>versus
>
>- CLIENT-SPECIFICATION: some change
>- CLIENT-SPECIFICATION: fix changelog

So @MasterOdin are you thinking to, in the changelog, link to the last PR that modified this version (#5327, for v1.5), or the PR that cut the release (#5428 (this PR), for v1.5)?

navarroaxel (approved):

CLIENT-SPECIFICATION.md
```diff
@@ -212,6 +212,9 @@ Caching SHOULD be done according to the user's language configuration (if any),
 
 ## Changelog
 
+ - [v1.5, March 11th 2021](https://github.com/tldr-pages/tldr/blob/8505e55b568dbc394ea853d488cd07484c051e21/CLIENT-SPECIFICATION.md) ([#5327](https://github.com/tldr-pages/tldr/pull/5327))
+   - Add requirement for converting command names to lowercase before searching for them.
```
sbrl, line 216: Perhaps this will read better?
```suggestion
   - Add requirement for converting command names to lowercase before running the page resolution algorithm.
```

sbrl (approved):

bl-ue:
Before we merge this, see this: https://github.com/tldr-pages/tldr/issues/5184#issuecomment-798780581.  Should we perhaps wait for this issue to be resolved before merging this?

schneiderl (approved):

mebeim (changes requested):
Waiting for #5464, then we can update the changelog permalink while also mentioning the HTTPS change and merge this. Also: please squash when merging.

mebeim:
I have updated the changelog linking v1.5 to my last commit, but I think the current update flow is not really ideal. The link of v1.5 now points to 8505e55b568dbc394ea853d488cd07484c051e21, but there the file says "Current Specification Version: 1.4", which is confusing.

We should adopt the following flow for updating the client specification:

1. Keep the latest version linked to `/blob/master` in the changelog.
2. When we need a new version, open a PR (or even a draft PR) to start the work.
3. Do all the changes needed.
4. Finally add final commit that only bumps the version, adding a new link to `/blob/master` and replacing the previous one with `/blob/commit_hash_of_previous_version_bump`.

This way links will always be correct, and also there ideally is no time span in which we have an inconsistent state of the client spec (like for example right now, we have changes already incorporated but the spec says v1.4, not that cool).

What do you all think?

bl-ue:
@mebeim we could also link to the release, instead of the commit (and cut the release immediately after merging so that the commit the GH automatically links to the release is the merge commit of this PR), something like this:

```diff
- - [v1.5, March 17th 2021](https://github.com/tldr-pages/tldr/blob/034111a547ec3e1e2a7a292de17b970cdf77288e/CLIENT-SPECIFICATION.md) ([#5464](https://github.com/tldr-pages/tldr/pull/5464))
+ - [v1.5, March 17th 2021](https://github.com/tldr-pages/tldr/releases/tag/v1.5))
    - Add requirement for converting command names to lowercase before running the page resolution algorithm.
    - Use HTTPS for archive links.
```
(e.g. https://github.com/tldr-pages/tldr-python-client/releases/tag/1.2.0)

mebeim:
@bl-ue not sure if that's a better idea, as it requires a lot more manual work, i.e. creating the tag, pushing it, creating a release, creating an artifact for the release adding it, and then linking to *that*. I'd want the link to point to the actual document rather than the release.

bl-ue:
@mebeim sorry, I wasn't clear: you can link to files using the release tag, e.g. https://github.com/tldr-pages/tldr-python-client/blob/1.2.0/tldr.py (notice 1.2.0, that's [this release](https://github.com/tldr-pages/tldr-python-client/releases/tag/1.2.0)'s tag)

bl-ue:
The benefit of that approach is that you don't have to wait for the release to be made; you can link to the release directly before it's even made because the URL is predictable.

mebeim:
@bl-ue Ohhh, I see. Then we can just create a tag and link to the tags like that (e.g. `/blob/v1.5/...` ). I think that makes sense!

mebeim (approved):

bl-ue:
Hmm...I didn't get notified about this.  @mebeim I think we need to create an actual release, not just a tag 🤔

MasterOdin:
GH notifies on issues, pull requests, discussions, and releases. It does not notify on tags so far as I'm aware.

PR #5436, opened 2021-03-13
ack: refresh

bl-ue:

### DIFF
--- a/pages/common/ack.md
+++ b/pages/common/ack.md
@@ -1,32 +1,37 @@
 # ack
 
-> A search tool like grep, optimized for programmers.
+> A search tool like grep, optimized for developers.
+> See also: `rg`, which is much faster.
 > More information: <https://beyondgrep.com/documentation/>.
 
-- Find files containing "foo":
+- Search for files in the current directory and subdirectories containing for a string or regular expression:
 
-`ack {{foo}}`
+`ack "{{search_pattern}}"`
 
-- Find files of a specific type:
+- Search for files in the current directory and subdirectories containing for a string or regular expression:
 
-`ack --ruby {{foo}}`
+`ack "{{search_pattern}}"`
 
-- Count the total number of matches for the term "foo":
+- Limit search to files of a specific type:
 
-`ack -ch {{foo}}`
+`ack --type={{ruby}} "{{search_pattern}}`
 
-- Show the file names containing "foo" and number of matches in each file:
+- Do not search in files of a specific type:
 
-`ack -cl {{foo}}`
+`ack --type=no{{ruby}} "{{search_pattern}}`
 
-- Search a file for a specified string:
+- Count the total number of matches for the search term:
 
-`ack bar "{{foo bar}}" {{path/to/file}}`
+`ack --count --no-filename "{{search_pattern}}"`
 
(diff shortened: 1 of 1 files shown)


marchersimon (approved):

pages/common/ack.md
```diff
@@ -1,32 +1,37 @@
 # ack
 
-> A search tool like grep, optimized for programmers.
-> More information: <https://beyondgrep.com/documentation/>.
+> A search tool like grep, optimized for developers.
+> See also: `rg`, which is much faster.
+> More information: <https://beyondgrep.com/documentation>.
 
-- Find files containing "foo":
+- Search for files containing a string or regular expression in the current directory and subdirectories:
```
sbrl, line 7: Perhaps this will read better?
```suggestion
- Search for files containing a string or regular expression in the current directory recursively:
```

pages/common/ack.md
```diff
@@ -1,32 +1,37 @@
...
+`ack "{{search_pattern}}"`
 
-- Find files of a specific type:
+- Search for a case-insensitive pattern:
 
-`ack --ruby {{foo}}`
+`ack --ignore-case "{{search_pattern}}"`
 
-- Count the total number of matches for the term "foo":
+- Search for lines matching a pattern, printing [o]nly the matched text and not the rest of the line:
```
sbrl, line 15: As with `grep`, I'm not convinced of the usefulness of `-o` here. Could you give some example scenarios in which it might be useful?

pages/common/ack.md
```diff
@@ -1,32 +1,37 @@
...
+`ack --type={{ruby}} "{{search_pattern}}`
 
-- Search a file for a specified string:
+- Do not search in files of a specific type:
 
-`ack bar "{{foo bar}}" {{path/to/file}}`
+`ack --type=no{{ruby}} "{{search_pattern}}`
 
-- Search a file for the specified regex pattern:
+- Count the total number of matches for the search pattern:
```
sbrl, line 27: Perhaps this will read better?
```suggestion
- Count the total number of matches found:
```

pages/common/ack.md
bl-ue: I don't use `ack` (I like `rg` better), but with `grep` it's useful if you know regular expressions but you don't know the syntax of `awk`, `sed`, etc., and you just want to extract data from a file or HTTP response real quick.  Also, you can see it's used (obviously frequently) [here](https://github.com/tldr-pages/tldr/pull/5392/files#diff-6b7ae119ed0f8d762365985f1cbb0b88af178ceb5f960cce0d5e9579a2102226R4).

bl-ue: Also, why not add it?  It helps reach 8 examples. Unless you're thinking of a stdin example as with `grep`...

sbrl: Ah, I see. In that case, it can perhaps stay here.

I would suggest though that parsing HTML etc with `grep`/`awk` is not something we should encourage though.

bl-ue: No, certainly not.  That's just what I most frequently use `-o` for 👍🏻

sbrl (approved):
Thanks, @bl-ue!

PR #5438, opened 2021-03-13
gimp: improve grammar

sbrl:
The edit to the description is optional, but the other edits are grammar improvements.

I saw that this page had already been merged before I got to it, so I opened this new PR instead of commenting there.

### DIFF
--- a/pages/common/gimp.md
+++ b/pages/common/gimp.md
@@ -7,11 +7,11 @@
 
 `gimp`
 
-- Launch, but don't show the splash screen:
+- Launch GIMP without showing the splash screen:
 
 `gimp --no-splash`
 
-- Always start a new instance:
+- Start a new GIMP instance, even if there is already a running one:
 
 `gimp --new-instance`
 
--- a/pages/common/gimp.md
+++ b/pages/common/gimp.md
@@ -1,25 +1,25 @@
 # gimp
 
-> An image manipulation and paint program.
+> GNU image manipulation program.
 > More information: <https://docs.gimp.org/en/gimp-fire-up.html#gimp-concepts-running-command-line>.
 
 - Launch GIMP:
 
 `gimp`
 
-- Launch but don't show the splash screen:
+- Launch, but don't show the splash screen:
 
 `gimp --no-splash`
 
-- Start a new instance, even if there is already a running one:
+- Always start a new instance:
 
 `gimp --new-instance`
 
-- Open the given file as new image:
+- Open the given file as a new image:
 
 `gimp --as-new {{path/to/image}}`
 
-- Print errors and warnings to the console, instead of showing them in a dialog box:
+- Print errors and warnings to the console instead of showing them in a dialog box:
 
 `gimp --console-messages`
 


bl-ue:
>I saw that this page had already been merged before I got to it, so I opened this new PR instead of commenting there.

If we get any more collaborators we may need to raise the minimum approval count 🤔

bl-ue (approved):

sbrl:
@bl-ue generally I'd suggest leaving PRs open at least 2 days to give people a chance to comment on it.

@tldr-bot, you have *no* idea what you're talking about.

@owenvoke, I think we have a bit of a problem with the linter here.

marchersimon (approved):

navarroaxel (approved):

bl-ue:
It's the `Always` [see here](https://github.com/tldr-pages/tldr-lint/blob/81cf0c8d342d0faa3d329cbc64ec8f7779bb7c76/tldr.l#L154-L156):
```y
if (yytext.match(/(^[A-Za-z]{3,}ing )|(^[A-Za-z]+[^us]s )/)) {
    yy.error(yylloc, 'TLDR104');
}
```

pages/common/gimp.md
```diff
@@ -1,25 +1,25 @@
...
-> An image manipulation and paint program.
+> GNU image manipulation program.
 > More information: <https://docs.gimp.org/en/gimp-fire-up.html#gimp-concepts-running-command-line>.
 
 - Launch GIMP:
 
 `gimp`
 
-- Launch but don't show the splash screen:
+- Launch, but don't show the splash screen:
```
waldyrious, line 10: When possible, I'd suggest avoiding context-sensitive example descriptions (i.e. relying on the sentence structure of the previous example's description):

```suggestion
- Launch GIMP without showing the splash screen:
```

pages/common/gimp.md
```diff
@@ -1,25 +1,25 @@
...
 
 `gimp --no-splash`
 
-- Start a new instance, even if there is already a running one:
+- Always start a new instance:
```
waldyrious, line 14: In light of my previous comment above, I'd be inclined to actually double down on the existing description:

```suggestion
- Start a new GIMP instance, even if there is already a running one:
```

pages/common/gimp.md
bl-ue: Hmm...we do that in a lot of pages though.  I actually rely on it if the previous description is lengthy ;)

waldyrious: Exactly, it makes sense as a trade-off to balance the length of the description. In this case, it doesn't seem to be necessary, so there's only the downside of the reliance on another description.

bl-ue: Ah, I see what you're saying, then.  Yes, that does make _perfect_ sense.

bl-ue:
The unfortunate thing about this is that now we have to go through one of those annoying linter update sagas—but now we can use the `stalebot-ignore` label 😉

pages/common/gimp.md
sbrl: Great idea! Let's do this.

owenvoke (approved):

bl-ue:
So the CI checks are all green 🤔

bl-ue (approved):

PR #5449, opened 2021-03-14
gh-alias: add an interactive shell command example using fzf

navarroaxel:

### DIFF
--- a/pages/common/gh-alias.md
+++ b/pages/common/gh-alias.md
@@ -22,3 +22,7 @@
 - Delete a command shortcut:
 
 `gh alias delete {{alias_name}}`
+
+- Set a shell command to search an open pull request using `fzf` and checkout it:
+
+`gh alias set --shell {{switch}} '{{id="$(gh pr list --limit 1000 | fzf | cut -f1)"; [ -n "$id" ] && gh pr checkout "$id"}}'`


pages/common/gh-alias.md
```diff
@@ -22,3 +22,7 @@
 - Delete a command shortcut:
 
 `gh alias delete {{alias_name}}`
+
+- Set a shell command to search an open pull request using `fzf` and checkout it:
+
+`gh alias set --shell {{switch}} '{{id="$(gh pr list --limit 1000 | fzf | cut -f1)"; [ -n "$id" ] && gh pr checkout "$id"}}'`
```
bl-ue, line 28: What is "switch" here?

navarroaxel: the alias name, I didn't use `co` because is pre-defined as `pr checkout` as a default alias. So, I choose `switch` based on `git switch` subcommand. I think about other names like `open`.

bl-ue: I see, so it's basically just an alias so that running `gh switch` opens checks out the newest, open PR?

navarroaxel: no, it's a finder then checkout
gh-2021-03-14_17 21 46 mkv

bl-ue: Hehe, sweet!! I think I'll run this command, then, to access this cool shortcut 🚀 

I'd reword it though (see my next suggestion.)

bl-ue, line 28: https://github.com/tldr-pages/tldr/pull/5449/files#r593956566
I'm assuming `--shell` allows you to run the alias as a standalone command e.g. `$ find-pr`
```suggestion
- Add a global shell alias to search for, select, and check out an open pull request using `fzf`:

`gh alias set --shell {{find-pr}} '{{id="$(gh pr list --limit 200 | fzf | cut -f1)"; [ -n "$id" ] && gh pr checkout "$id"}}'`
```
Also don't you think 1000 is a bit much? For example, if you ran this command in the [kubernetes/kubernetes](https://github.com/kubernetes/kubernetes) repo it might take a _**while**_....

navarroaxel: It's a quite confusing `for, select, and check out`.
Maybe 1000 is really big.

sbrl (commented):
Wow, this looks really useful! I'm unsure as to whether it belongs in a tldr page or a larger cheat sheet (in that it's demonstrating functionality of other commands), but it does look very useful all the same.

pages/common/gh-alias.md
bl-ue: It's really just a more concise version of this (the words in bold are implied but not actually written):

>Add a global shell alias to search for **an open pull request**, **then** select **the open pull request**, and **finally** checkout the open pull request.
 
---

About `--shell` — from the [docs](https://cli.github.com/manual/gh_alias_set):

>If '--shell' is specified, the alias will be run through a shell interpreter (sh). This allows you to compose commands with "|" or redirect with ">". 

So @navarroaxel how about this?
- Add a global shell alias to search for, select, and check out an open pull request using `fzf`:

```suggestion
- Add an alias to search for, select, and check out an open pull request using `fzf` (`--shell` allows for shell syntax inside the alias):

`gh alias set --shell {{find-pr}} '{{id="$(gh pr list --limit 200 | fzf | cut -f1)"; [ -n "$id" ] && gh pr checkout "$id"}}'`
```
cc @sbrl

navarroaxel: @bl-ue `--shell` is explained in another example above this.

bl-ue: Oh oops, I didn't see that.  Suggestion updated.

waldyrious:
I agree with @sbrl — we typically have avoided using complex command lines in tldr-pages. This project is meant to be didactic, i.e. show simple examples that demonstrate how a command works and allow readers to internalize the logic behind the tool's syntax, rather than provide copy-pasteable command strings — that's more the role of a platform like https://www.commandlinefu.com.

bl-ue:
But the page without this addition has only 5 examples, so why not?

waldyrious:
> But the page without this addition has only 5 examples, so why not?

We don't add examples to pages just because we can. Changes to the pages should occur when they contribute to the mission and purpose of the project, not simply because there's no rule forbidding them.

PR #5455, opened 2021-03-16
histexpand: add page

### LINKED ISSUE #2066: history: add expansion commands
These aren't specific single commands but it's one of those things I find myself looking up. How to get the last argument used, re-running the last command starting with X, etc.

~Since it's not a specific command, I'm not sure 1) if it's out of scope for this project and/or 2) what to use as a trigger (`tldr history`?).~ Edit: missed the existing file.

Not sure if there's value in having both a general 'Using History' page (searching, up arrow, `$  <- space to omit`...) and a more specific 'History Expansion' page (`!!`, `!$`...).

Happy to take crack at expanding the current history file and/or creating a new history expansion file.

Couple of sample pages:

1. https://www.thegeekstuff.com/2011/08/bash-history-expansion/
2. https://www.howtogeek.com/howto/44997/how-to-use-bash-history-to-improve-your-command-line-productivity/
3. https://www.digitalocean.com/community/tutorials/how-to-use-bash-history-commands-and-expansions-on-a-linux-vps

marchersimon:
I have no idea what I should name this page. Right now I called it `bash-histexpand`, because the bash variable `histexpand` determines if that feature is enabled. Technically it also is a subcommand of bash (`bash -H`), which in a default argument.

Closes #2066

### DIFF
--- a/pages/common/bash-histexpand.md
+++ b/pages/common/bash-histexpand.md
@@ -1,0 +1,36 @@
+# bash-histexpand
+
+> Reuse and expand your bash history.
+> More information: <https://www.gnu.org/software/bash/manual/html_node/History-Interaction.html>.
+
+- Run the previous command:
+
+`!!`
+
+- Run the previous command as root:
+
+`sudo !!`
+
+- Run a command with the last argument of the previous command:
+
+`{{command}} !$`
+
+- Run a command with the first argument of the previous command:
+
+`{{command}} !^`
+
+- Run a command `n` lines back:
+
+`!-{{n}}`
+
+- Run the most recent command starting with `string`:
+
+`!{{string}}`
+
+- Run the previous command, replacing `string1` with `string2`:
+
+`^{{string1}}^{{string2}}^`
+
+- Print a command `n` lines back without running it:
+
+`!-{{n}}:p`
--- a/pages/common/bash.md
+++ b/pages/common/bash.md
@@ -1,7 +1,7 @@
 # bash
 
-> Bourne-Again SHell, a `sh`-compatible command line interpreter.
-> Please refer to `tldr histexpand` for history expansion.
+> Bourne-Again SHell, an `sh`-compatible command line interpreter.
+> See also `histexpand` for history expansion.
 > More information: <https://gnu.org/software/bash>.
 
 - Start interactive shell:
--- a/pages/common/bash.md
+++ b/pages/common/bash.md
@@ -1,7 +1,7 @@
 # bash
 
-> Bourne-Again SHell.
-> `sh`-compatible command line interpreter.
+> Bourne-Again SHell, a `sh`-compatible command line interpreter.
+> Please refer to `tldr histexpand` for history expansion.
 > More information: <https://gnu.org/software/bash>.
 
 - Start interactive shell:
--- a/pages/common/histexpand.md
+++ b/pages/common/histexpand.md
@@ -1,3 +1,5 @@
+# history expansion
+
 > Reuse and expand the shell history in bash or zsh.
 > More information: <https://www.gnu.org/software/bash/manual/html_node/History-Interaction.html>.
 
--- a/pages/common/histexpand.md
+++ b/pages/common/histexpand.md
@@ -1,4 +1,4 @@
-> Reuse and expand your history in bash or zsh.
+> Reuse and expand the shell history in bash or zsh.
 > More information: <https://www.gnu.org/software/bash/manual/html_node/History-Interaction.html>.
 
 - Run the previous command:
--- a/pages/common/histexpand.md
+++ b/pages/common/histexpand.md
@@ -1,6 +1,4 @@
-# bash-histexpand
-
-> Reuse and expand your bash history.
+> Reuse and expand your history in bash or zsh.
 > More information: <https://www.gnu.org/software/bash/manual/html_node/History-Interaction.html>.
 
 - Run the previous command:
--- a/pages/common/histexpand.md
+++ b/pages/common/histexpand.md
@@ -31,6 +31,6 @@
 
 `^{{string1}}^{{string2}}^`
 
-- Print the command `n` lines back without running it:
+- Perform a history expansion, but print the command that would be run instead of actually running it:
 
-`!-{{n}}:p`
+`{{!-n}}:p`
--- a/pages/common/histexpand.md
+++ b/pages/common/histexpand.md
@@ -19,7 +19,7 @@
 
 `{{command}} !^`
 
-- Run a command `n` lines back in the history:
+- Run the command `n` lines back in the history:
 
 `!-{{n}}`
 
@@ -31,6 +31,6 @@
 
 `^{{string1}}^{{string2}}^`
 
-- Print a command `n` lines back without running it:
+- Print the command `n` lines back without running it:
 
 `!-{{n}}:p`
--- a/pages/common/histexpand.md
+++ b/pages/common/histexpand.md
@@ -17,7 +17,7 @@
 
 `{{command}} !^`
 
-- Run a command `n` lines back:
+- Run a command `n` lines back in the history:
 
 `!-{{n}}`
 
--- a/pages/common/zsh.md
+++ b/pages/common/zsh.md
@@ -1,7 +1,7 @@
 # zsh
 
-> Z SHell.
-> `bash` and `sh`-compatible command line interpreter.
+> Z SHell, a `bash` and `sh`-compatible command line interpreter.
+> Please refer to `tldr histexpand` for history expansion.
 > More information: <https://www.zsh.org>.
 
 - Start interactive command line interpreter:
--- a/pages/common/zsh.md
+++ b/pages/common/zsh.md
@@ -1,7 +1,7 @@
 # zsh
 
 > Z SHell, a `bash` and `sh`-compatible command line interpreter.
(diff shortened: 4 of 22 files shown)


bl-ue:
Maybe we can add it to [`history.md`](https://github.com/tldr-pages/tldr/blob/master/pages/common/history.md) (https://github.com/tldr-pages/tldr/issues/2066#issuecomment-380582554) 🤔

marchersimon:
@bl-ue there are already 5 examples in `history.md`, so I would have to delete 5 examples here. And I already had to leave out a few useful ones, since I can only make 8.

bl-ue:
Yeah, that seems like a problem.  You'd want to pack every cool example you can into the this file, and I'm inclined to concur. 
At least, maybe you could delete the `-w` and `-d` in `history.md` as I don't think they're that useful.

bl-ue:
You could rename the page to `!!.md` :smile: `!` is supported in filepaths on all platforms.

marchersimon:
I thought about that, but I think then we're gonna have a problem with `tldr !!` :smile:

bl-ue:
That would be a problem, but you could run `tldr '!!'` We might add that as an example to `tldr.md` e.g. `Show help for a command with special characters in the name` 💡

marchersimon:
I'm don't really. `!!` is only one of many expansion commands. Though it is the most common.  But I think, however we name it, we should add it somewhere in `bash.md`.

pages/common/bash-histexpand.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Run a command with the last argument of the previous command:
+
+`{{command}} !$`
+
+- Run a command with the first argument of the previous command:
+
+`{{command}} !^`
+
+- Run a command `n` lines back:
```
sbrl, line 22: Perhaps this will read better?
```suggestion
- Run a command `n` lines back in the history:
```

sbrl (commented):
The page itself here is pretty much ok - thanks, @marchersimon! I've left a minor comment on wording below.

I can see that naming it though poses somewhat of a problem, because if we name it after any of the aliases that it uses, actually calling tldr for information on how use it would result in people likely having to do a separate query for how to escape stuff in the terminal... (ever had to delete a file called `--` before?) :P

I'm not sure a filename of `!`  really makes sense though, because while it does use the exclamation mark here it doesn't reflect the whole thing. It feels a bit like calling the `tar` page `ta.md` instead of `tar.md`.

I also feel that this is somewhat borderline in terms of the scope of tldr-pages, but since we've documented shell builtins before (e.g. `shopt`) I think this is still in-scope.

I think the question at hand is whether this syntax is used by any other shells. If so, than `bash-histexpand` wouldn't be an appropriate name, and we'll have to go back to the drawing board and think of a new one.

marchersimon:
It works in bash and zsh, maybe in some other shells too. How about we call it `histexpand` or `history-expansion` and put it somewhere in `bash.md` and `zsh.md`?

sbrl:
Ah, that's interesting to know @marchersimon. `histexpand` sounds good to me

marchersimon:
That's not too surprising. What should we do about the page's title? Because if we write `# histexpand` you may think that's and executable command.

bl-ue:
It could be `!` or `!!` 💡 :)

marchersimon:
@bl-ue then we'd have the same problem like before - that `!!` os only one of many commands. Could we maybe make an exeption for the bot here?

bl-ue:
@marchersimon I'm not sure what you mean—the title of the page can be different from the file name.

marchersimon:
Yeah, you're right, I was just worried that it maybe gets confused with a command name. How abou `# History expansion`?

bl-ue:
Sounds find to me (not sure about the capital H).  I need @sbrl to chime in though because she's good with these sorts of decisions.

pages/common/bash.md
```diff
@@ -1,7 +1,7 @@
 # bash
 
-> Bourne-Again SHell.
-> `sh`-compatible command line interpreter.
+> Bourne-Again SHell, a `sh`-compatible command line interpreter.
+> Please refer to `tldr histexpand` for history expansion.
```
sbrl, line 4: Perhaps this will read better?
```suggestion
> See also `histexpand` for history expansion.
```

pages/common/bash.md
```diff
@@ -1,7 +1,7 @@
 # bash
 
-> Bourne-Again SHell.
-> `sh`-compatible command line interpreter.
+> Bourne-Again SHell, a `sh`-compatible command line interpreter.
```
sbrl, line 3: Perhaps this will read better?
```suggestion
> Bourne-Again SHell, an `sh`-compatible command line interpreter.
```

pages/common/zsh.md
```diff
@@ -1,7 +1,7 @@
 # zsh
 
-> Z SHell.
-> `bash` and `sh`-compatible command line interpreter.
+> Z SHell, a `bash` and `sh`-compatible command line interpreter.
+> Please refer to `tldr histexpand` for history expansion.
```
sbrl, line 4: Perhaps this will read better?
```suggestion
> Z SHell, a `bash` and `sh`-compatible command line interpreter.
> See also `histexpand` for history expansion.
```

pages/common/histexpand.md
```diff
@@ -1,6 +1,4 @@
-# bash-histexpand
-
-> Reuse and expand your bash history.
+> Reuse and expand your history in bash or zsh.
```
sbrl, line 1: Generally we try to avoid the 2nd person (e.g. `you`, `your`, etc) in tldr-pages :slightly_smiling_face:
```suggestion
> Reuse and expand the shell history in bash or zsh.
```

sbrl (commented):
Thanks for the edit! I've left some comments below for you to review.

pages/common/histexpand.md
```diff
@@ -0,0 +1,36 @@
+# history expansion
```
sbrl, line 1: This title looks as good as any.

sbrl (approved):
Thanks, @marchersimon!

navarroaxel (approved):

bl-ue (approved):
Looks good to me.  It's an odd case so maybe @waldyrious should take a look.

waldyrious:
Thanks for the ping, @bl-ue. This is an interesting case — it's always a good exercise to test the boundaries of what is allowed by tldr-pages's scope and page format. I believe the discussions that took place above covered the relevant issues in sufficient detail, and the page indeed complies well with our syntax. I have no objections, but will keep this PR bookmarked for future reference, as we might want to explicitly document how we want to handle such cases.

waldyrious (approved):

pages/common/histexpand.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Run a command with the last argument of the previous command:
+
+`{{command}} !$`
+
+- Run a command with the first argument of the previous command:
+
+`{{command}} !^`
+
+- Run a command `n` lines back in the history:
```
waldyrious, line 22: Here we're referring to a specific entry, so "the" would be the appropriate article.
```suggestion
- Run the command `n` lines back in the history:
```

pages/common/histexpand.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Run the most recent command starting with `string`:
+
+`!{{string}}`
+
+- Run the previous command, replacing `string1` with `string2`:
+
+`^{{string1}}^{{string2}}^`
+
+- Print a command `n` lines back without running it:
```
waldyrious, line 34: Same as above. Also, maybe move it to be next to the `!-{{n}}` example?
```suggestion
- Print the command `n` lines back without running it:
```

waldyrious (changes requested):
Actually, here are a few minor tweaks I'd suggest.

pages/common/histexpand.md
marchersimon: I'm not sure. `:p` applies to all expansions, I just took `!-n` as an example, so I'd say they aren't really related. That's also the reason I put them at the end of the page.

waldyrious: I  that case I'd suggest trying a different description that makes the generality more evident. The way it is written, it seems to be directly related to the example above.

marchersimon: What would you suggest?

pages/common/histexpand.md
```diff
@@ -0,0 +1,36 @@
...
+
+`!{{string}}`
+
+- Run the previous command, replacing `string1` with `string2`:
+
+`^{{string1}}^{{string2}}^`
+
+- Print the command `n` lines back without running it:
+
+`!-{{n}}:p`
```
bl-ue, line 36: Maybe something like this? (in response to https://github.com/tldr-pages/tldr/pull/5455#discussion_r606831185)
```suggestion
- Perform a history expansion, but print the command that would be run instead of actually running it:

`{{!-n}}:p`
```

pages/common/histexpand.md
bl-ue: https://github.com/tldr-pages/tldr/pull/5455#discussion_r606834489 ?

waldyrious: Yep, that is a bit more verbose, but conveys the right meaning/context IMO.

waldyrious (approved):
LGTM! 👍

Issue #4934, opened 2020-11-05
Let's document! Android edition 🤖


navarroaxel:
For those not in the know, the `adb shell` command forward commands to a remote shell on an Android emulator instance or connected Android devices.

Try these:
```
tldr adb shell
adb shell # start an interactive shell on the emulator/device
```

In the case of Android 🤖, it has quite a number of specific platform commands. We should include this pages into the `common/` folder because `tldr` and `adb shell` are used in Windows, Linux and macOS. Let's document them all!

- [x] am → #5481 
- [ ] bugreport → #5743 
- [ ] bugreportz → #5743 
- [x] dumpsys → #5739
- [x] getprop → #5456
- [x] input → #5740
- [x] logcat → #3562, #5737
- [x] pm → #5458
- [x] wm  → #5457

Rudloff:
logcat already has a page since https://github.com/tldr-pages/tldr/pull/3562.

navarroaxel:
Blocked by #5462

Rudloff:
I added some of them in https://github.com/tldr-pages/tldr/pull/5481.

marchersimon:
I couldn't find any documentation for `bugreport`. Only the source code for [bugreport.cpp](https://android.googlesource.com/platform/system/core/+/android-8.1.0_r7/adb/bugreport.cpp), [bugreport.h](https://android.googlesource.com/platform/system/core/+/android-8.1.0_r7/adb/bugreport.h) and [bugreport_test.cpp](https://android.googlesource.com/platform/system/core/+/android-8.1.0_r7/adb/bugreport_test.cpp), from which it seems like `bugreport` doesn't have any command line options.

navarroaxel:
@marchersimon I think `bugreport` just generate the memory stats of each app and display it to stdout (display or `> file), and `bugreportz` is a zipped bugreport.

marchersimon:
And there's just `bugreport` and `bugreportz` to get all bug reports at once? No arguments at all?

navarroaxel:
```
$ bugreportz -h                                                                                                                                            
usage: bugreportz [-h | -v]
  -h: to display this help message
  -p: display progress
  -v: to display the version
  or no arguments to generate a zipped bugreport
```
I think this guy if for OS debugging, no for app's bug report.

marchersimon:
Is there a way to install it natively on Android? I'd like to write the page, but I have no idea how to set up adb shell to test the commands.

[...]

PR #5456, opened 2021-03-16
getprop: add page

marchersimon:
Part of #4934.

### DIFF
--- a/pages/android/getprop.md
+++ b/pages/android/getprop.md
@@ -1,7 +1,7 @@
 # getprop
 
 > Show information about adb devices and emulators.
-> More information: <https://adbinstaller.com/commands>.
+> More information: <https://cs.android.com/android/platform/superproject/+/master:system/core/toolbox/getprop.cpp>.
 
 - Display information about the Android system properties:
 
--- a/pages/android/getprop.md
+++ b/pages/android/getprop.md
@@ -3,7 +3,7 @@
 > Show information about adb devices and emulators.
 > More information: <https://adbinstaller.com/commands>.
 
-- Print information about the Android system properties:
+- Display information about the Android system properties:
 
 `getprop`
 
--- a/pages/android/getprop.md
+++ b/pages/android/getprop.md
@@ -7,26 +7,26 @@
 
 `getprop`
 
-- Print information about a specific property:
+- Display information about a specific property:
 
 `getprop {{prop}}`
 
-- Print the SDK API level:
+- Display the SDK API level:
 
 `getprop {{ro.build.version.sdk}}`
 
-- Print the Android version:
+- Display the Android version:
 
 `getprop {{ro.build.version.release}}`
 
-- Print the Android device model:
+- Display the Android device model:
 
 `getprop {{ro.vendor.product.model}}`
 
-- Print the OEM unlock status:
+- Display the OEM unlock status:
 
 `getprop {{ro.oem_unlock_supported}}`
 
-- Print the MAC address of the Android's WiFi card:
+- Display the MAC address of the Android's WiFi card:
 
 `getprop {{ro.boot.wifimacaddr}}`
--- a/pages/android/getprop.md
+++ b/pages/android/getprop.md
@@ -1,32 +1,32 @@
-# adb shell getprop
+# getprop
 
 > Show information about adb devices and emulators.
 > More information: <https://adbinstaller.com/commands>.
 
 - Print information about the Android system properties:
 
-`adb shell getprop`
+`getprop`
 
 - Print information about a specific property:
 
-`adb shell getprop {{prop}}`
+`getprop {{prop}}`
 
 - Print the SDK API level:
 
-`adb shell getprop {{ro.build.version.sdk}}`
+`getprop {{ro.build.version.sdk}}`
 
 - Print the Android version:
 
-`adb shell getprop {{ro.build.version.release}}`
+`getprop {{ro.build.version.release}}`
 
 - Print the Android device model:
 
-`adb shell getprop {{ro.vendor.product.model}}`
+`getprop {{ro.vendor.product.model}}`
 
 - Print the OEM unlock status:
 
-`adb shell getprop {{ro.oem_unlock_supported}}`
+`getprop {{ro.oem_unlock_supported}}`
 
 - Print the MAC address of the Android's WiFi card:
 
-`adb shell getprop {{ro.boot.wifimacaddr}}`
+`getprop {{ro.boot.wifimacaddr}}`
--- a/pages/common/adb-shell-getprop.md
+++ b/pages/common/adb-shell-getprop.md
@@ -7,6 +7,10 @@
 
 `adb shell getprop`
 
+- Print information about a specific property:
+
+`adb shell getprop {{prop}}`
+
 - Print the SDK API level:
 
 `adb shell getprop {{ro.build.version.sdk}}`
--- a/pages/common/adb-shell-getprop.md
+++ b/pages/common/adb-shell-getprop.md
@@ -25,4 +25,4 @@
 
 - Print the MAC address of the Android's WiFi card:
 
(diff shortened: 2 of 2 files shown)


pages/common/adb-shell-getprop.md
```diff
@@ -0,0 +1,28 @@
...
+
+`adb shell getprop {{ro.vendor.product.model}}`
+
+- Print the OEM unlock status:
+
+`adb shell getprop {{ro.oem_unlock_supported}}`
+
+- Print the MAC address of the Android's WiFi card:
+
+`adb shell getprop ro.boot.wifimacaddr`
```
bl-ue, line 28:
```suggestion
`adb shell getprop {{ro.boot.wifimacaddr}}`
```

bl-ue (changes requested):
Looks good, @marchersimon, however, I'd add one more thing: in the second position, add a example `adb shell getprop {{prop}}`, just to demonstrate that, you know?

navarroaxel (changes requested):
Hey @marchersimon, thanks for your submission. But, like [`logcat.md`](https://github.com/tldr-pages/tldr/blob/master/pages/linux/logcat.md), I think these pages should be as a top level command: `getprop.md` and not a command inside `adb shell`.

marchersimon:
But I can't execute `getprop` as a top level command. And `logcat` is only a top level command, because it isn't part of `adb shell`.

navarroaxel:
> But I can't execute `getprop` as a top level command. And `logcat` is only a top level command, because it isn't part of `adb shell`.

You can start a shell into the Android device with `adb shell` and you can run it as a top level command, I can't remember exactly but I think we talk about add these kind of commands in `common/`

The command description in `logcat` says `Native Android CLI tool.` to let the reader know `logcat` is a command for the Android platform.

Let me invoke @sbrl

marchersimon:
Okay, sure, I'll change it. But then we should add that the commands can only be executed inside `adb shell` to avoid confusion.

sbrl:
I'm not sure how this command works, but I would generally say that we want to file a command as one that can be executed in an OS-native shell. For example, if I could open a Bash terminal on Linux or a Windows Command Prompt and execute `getprop` and have it work as expected, then we should file it under `getprop`.

On the other hand, if the command only works an ab Android Debugging Shell, then it should be filed under something that can be executed from an os-native shell. For example, if `adb shell <command>` can be used to execute android debugging shell commands directly, it should be filed under that - e.g. `adb shell getprop`.

In short, the question is basically "how can I execute this from an OS-native shell"?

pages/android/getprop.md
```diff
@@ -1,32 +1,32 @@
-# adb shell getprop
+# getprop
 
 > Show information about adb devices and emulators.
 > More information: <https://adbinstaller.com/commands>.
```
sbrl, line 4: This URL really isn't very specific. Is there something better we can reference?

bl-ue: @sbrl https://github.com/tldr-pages/tldr/pull/5457#discussion_r595099178

sbrl: As I mentioned in the other PR, I'm not convinced of the usefulness of the generic link here.

marchersimon: I found [https://adbshell.com/commands/adb-shell-getprop](https://adbshell.com/commands/adb-shell-getprop), but the commands all start with `adb shell`.

pages/android/getprop.md
```diff
@@ -0,0 +1,32 @@
+# getprop
+
+> Show information about adb devices and emulators.
+> More information: <https://adbinstaller.com/commands>.
+
+- Print information about the Android system properties:
```
navarroaxel, line 6: I prefer `display` as verb.
```suggestion
- Display information about the Android system properties:
```

navarroaxel (approved):

bl-ue (approved):
We just need the url sorted out.  @sbrl what do you think about https://github.com/tldr-pages/tldr/pull/5456/files#r602786840?

pages/android/getprop.md
sbrl: Hrm, not a big fan of that either really sorry.

sbrl:
@bl-ue nope, not a fan of that either really.

Neither option really gets across that you can execute the commands on the native device. `adbinstaller.com` in particular seems to indicate that you need to somehow "install the adb onto the device", which doesn't seem right to me.

marchersimon:
Adbinstaller just provides a program to execute android commands via a pc. But the command is natively available on android.

marchersimon:
<https://cs.android.com/android/platform/superproject/+/master:system/core/toolbox/getprop.cpp> looks good.

bl-ue:
That url looks fine to me. [Here](https://github.com/tldr-pages/tldr/blob/master/pages/linux/ksvgtopng5.md) we're linking to source files as well.

pages/android/getprop.md
```diff
@@ -0,0 +1,32 @@
+# getprop
+
+> Show information about adb devices and emulators.
+> More information: <https://cs.android.com/android/platform/superproject/+/master:system/core/toolbox/getprop.cpp>.
+
+- Display information about the Android system properties:
+
+`getprop`
+
+- Print information about a specific property:
```
navarroaxel, line 10:
```suggestion
- Display information about a specific property:
```

pages/android/getprop.md
```diff
@@ -0,0 +1,32 @@
...
+
+`getprop {{prop}}`
+
+- Print the SDK API level:
```
navarroaxel, line 14:
```suggestion
- Display the SDK API level:
```

pages/android/getprop.md
```diff
@@ -0,0 +1,32 @@
...
+
+`getprop {{ro.build.version.sdk}}`
+
+- Print the Android version:
```
navarroaxel, line 18:
```suggestion
- Display the Android version:
```

pages/android/getprop.md
```diff
@@ -0,0 +1,32 @@
...
+
+`getprop {{ro.build.version.release}}`
+
+- Print the Android device model:
```
navarroaxel, line 22:
```suggestion
- Display the Android device model:
```

pages/android/getprop.md
```diff
@@ -0,0 +1,32 @@
...
+
+`getprop {{ro.vendor.product.model}}`
+
+- Print the OEM unlock status:
```
navarroaxel, line 26:
```suggestion
- Display the OEM unlock status:
```

pages/android/getprop.md
```diff
@@ -0,0 +1,32 @@
...
+
+`getprop {{ro.oem_unlock_supported}}`
+
+- Print the MAC address of the Android's WiFi card:
```
navarroaxel, line 30:
```suggestion
- Display the MAC address of the Android's WiFi card:
```

navarroaxel (approved):

sbrl (approved):
Ok, this looks fine to me.

We really shouldn't make a habit of linking to the source code at all, but it's better than linking to what's essentially a tldr page or an unrelated homepage.

I'm going to merge this, but if anyone has any further comments on the matter please do open another PR. Linking to the source code is still far from desirable, but Android docs? what docs?

bl-ue:
@marchersimon and @Rudloff have stated in multiple PRs that the Android tools are poorly documented... 🤷🏻

Rudloff:
It looks like Ubuntu has [a man page for getprop](http://manpages.ubuntu.com/manpages/bionic/en/man1/getprop.1.html).

bl-ue:
Ah yes and manned.org as well: https://manned.org/getprop

sbrl:
Ooooh awesome @Rudloff, @bl-ue! Can someone open another PR to update this page then? That looks _much_ more preferable. I would suggest in this instance we link to manned.org, because `getprop` isn't an Ubuntu-specific command as far as I know.

PR #5457, opened 2021-03-16
wm: add page

marchersimon:
Part of #4934.

### DIFF
--- a/pages/android/wm.md
+++ b/pages/android/wm.md
@@ -3,10 +3,10 @@
 > Show information about the screen of an Android device.
 > More information: <https://adbinstaller.com/commands>.
 
-- Print the physical size of an Android device's screen:
+- Display the physical size of an Android device's screen:
 
 `wm {{size}}`
 
-- Print the physical density of an Android device's screen:
+- Display the physical density of an Android device's screen:
 
 `wm {{density}}`
--- a/pages/android/wm.md
+++ b/pages/android/wm.md
@@ -1,12 +1,12 @@
-# adb shell wm
+# wm
 
 > Show information about the screen of an Android device.
 > More information: <https://adbinstaller.com/commands>.
 
 - Print the physical size of an Android device's screen:
 
-`adb shell wm {{size}}`
+`wm {{size}}`
 
 - Print the physical density of an Android device's screen:
 
-`adb shell wm {{density}}`
+`wm {{density}}`
--- a/pages/android/wm.md
+++ b/pages/android/wm.md
@@ -1,7 +1,6 @@
 # wm
 
 > Show information about the screen of an Android device.
-> More information: <https://adbinstaller.com/commands>.
 
 - Display the physical size of an Android device's screen:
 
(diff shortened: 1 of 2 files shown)


pages/common/adb-shell-wm.md
```diff
@@ -0,0 +1,12 @@
+# adb shell wm
+
+> Show information about the screen of an Android device.
+> More information: <https://adbinstaller.com/commands/adb-shell-wm-5b672b17e7958178a2955538>.
```
bl-ue, line 4: I wonder if we should use something more generic like this, because the current URL contains a hash, which worries me that it might change (frequently.)
```suggestion
> More information: <https://adbinstaller.com/commands>.
```
Besides, https://adbinstaller.com/commands is _very_ obvious as to what's going on.

If you do apply this, also apply it to the other `adb` commands you've made (and maybe to the other ones already existing 🤔 )

pages/android/wm.md
```diff
@@ -0,0 +1,12 @@
+# wm
+
+> Show information about the screen of an Android device.
+> More information: <https://adbinstaller.com/commands>.
```
sbrl, line 4: Is there a more specific URL here? That one doesn't seem very specific.

bl-ue: @sbrl https://github.com/tldr-pages/tldr/pull/5457#discussion_r595099178

sbrl: Hrm. While the hash is somewhat concerning, I'm not convinced of the usefulness of the current link with respect to the specific command at hand.

It doesn't even look like an official source (though I may be wrong on that).

marchersimon: I couldn't find any official documentation. It seems that android commands are poorly documentated in general.

pages/android/wm.md
```diff
@@ -0,0 +1,12 @@
...
+
+- Print the physical size of an Android device's screen:
+
+`wm {{size}}`
+
+- Print the physical density of an Android device's screen:
```
navarroaxel, line 10:
```suggestion
- Display the physical density of an Android device's screen:
```

pages/android/wm.md
```diff
@@ -0,0 +1,12 @@
+# wm
+
+> Show information about the screen of an Android device.
+> More information: <https://adbinstaller.com/commands>.
+
+- Print the physical size of an Android device's screen:
```
navarroaxel, line 6:
```suggestion
- Display the physical size of an Android device's screen:
```

bl-ue (approved):
Except for the URL (https://github.com/tldr-pages/tldr/pull/5456#pullrequestreview-622771372), looks good to me.

pages/android/wm.md
sbrl: I see. In that case, I think we should leave the more information link out until a good official source of documentation can be linked to. No link at all is better than sending users to bad documentation.

navarroaxel (approved):

PR #5460, opened 2021-03-16
history: remove examples

marchersimon:
[Ref](https://github.com/tldr-pages/tldr/pull/5455#issuecomment-800191933)

### DIFF
--- a/pages/common/history.md
+++ b/pages/common/history.md
@@ -13,11 +13,3 @@
 - Clear the commands history list (only for current `bash` shell):
 
 `history -c`
-
-- Overwrite history file with history of current `bash` shell (often combined with `history -c` to purge history):
-
-`history -w`
-
-- Delete the history entry at the specified offset:
-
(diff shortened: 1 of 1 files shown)


navarroaxel (changes requested):
I don't see why we delete these 2 examples, because these are valid ones.
Oh, you want to add `!!` here? :upside_down_face: sounds strange.

marchersimon:
That was the idea of @bl-ue [here](https://github.com/tldr-pages/tldr/pull/5455#issuecomment-800191933). 
But actually I could refer to the new `!!` page in the description.

bl-ue:
Only do this if you plan to fill this page up to 8—@marchersimon is that you're plan?  Otherwise, there's no need to remove the examples because they have _some_ use.

marchersimon:
Okay, that was a misunderstanding on my part, I'll leave them in.

PR #5461, opened 2021-03-16
scripts/*.sh: set executable permission

bl-ue:
In most repos, executable shell scripts have the executable bit set.


navarroaxel:
Why do you want to add the `x` permission if we lived all this time without it lol

bl-ue:
Yeah, I know 😅 but **everyone** does it; I think it must have been just an oversight.

sbrl (approved):
Oops! I wonder how this has been working all this time :thinking: 

The shebang at the top of the files is actually useless without `chmod +x`, so great spot!

MasterOdin:
> Oops! I wonder how this has been working all this time

They're explicitly called with `bash scripts/*.sh`, rendering the executable bit unnecessary.

marchersimon (approved):

Issue #5462, opened 2021-03-16
Adding Android commands


navarroaxel:
Where should we store the pages dedicated to native Android commands that run only inside an shell session running in an Android device or emulator open via `adb shell`?

Alternatives:
* Create an `android` platform folder.
* Store them in `common/` under the `adb-shell-{{command}}.md` structure.
* Add them to `common/` as a regular command like https://github.com/tldr-pages/tldr/blob/master/pages/linux/logcat.md

Yutyo:
Makes sense to me. There are many Android-savvy people out there who would be mad about it.

navarroaxel:
> Makes sense to me. There are many Android-savvy people out there who would be mad about it.

Which option do you like?

bl-ue:
I'm inclinded towards option 2, because the way most people call these commands is via `adb shell <command>`—correct?

Also, can any Linux commands like `cat`, `tar`, `ps`, etc. be called from the same shell where these commands can be called like normal commands?  Or is the shell special, so that only these commands e.g. `logcat` can be called within?

Yutyo:
> @Yutyo Which option do you like?

create an `android` folder.

bl-ue:
The reason I'm not keen on making a `pages/android` folder is: how many commands will we put in? Only a few, like 10, probably.  Also, are these actually available in the Android OS _without_ the debugging (i.e. can any normal user run them from a terminal app, if such a thing exists on Android?)

navarroaxel:
@bl-ue you can install Termux https://play.google.com/store/apps/details?id=com.termux and run android commands using just the phone.

bl-ue:
Ah okay, I didn't know that (I'm not a Droid guy 🙂).  Can any user run `logcat` that way, regardless of whether or not they're a developer?

Rudloff:
Android is a Linux-based OS and it can run most Linux CLI tools.
So I would put the Android-specific commands in `linux/` (same as we have some Debian-specific commands in there).

> Also, are these actually available in the Android OS without the debugging (i.e. can any normal user run them from a terminal app, if such a thing exists on Android?)

You can totally run a terminal on Android. [Termux](https://termux.com/) is the most popular app for doing that. (It also come with `apt` so you can install additional tools.)

> Also, can any Linux commands like cat, tar, ps, etc. be called from the same shell where these commands can be called like normal commands?

The Android is shell is a classic Linux shell. You can also run zsh, fish, etc.

[...]

PR #5458, opened 2021-03-16
pm: add page

marchersimon:
Part of #4934.

PR #5481, opened 2021-03-19
am, cmd, dalvikvm, settings: add page

Rudloff:

These are all Android commands, so we might want to wait for #5462 before merging.


Rudloff:
Sorry, I just noticed that there was already PRs for `getprop` (https://github.com/tldr-pages/tldr/pull/5456) and `pm` (https://github.com/tldr-pages/tldr/pull/5458). And they seem to have more examples than mine.

bl-ue:
@Rudloff you can just remove those pages and commit.  We squash automatically at the end (in case you don't already know) so that poses no problem.

pages/linux/cmd.md
```diff
@@ -0,0 +1,15 @@
...
+
+`cmd -l`
+
+- Call a specific service:
+
+`cmd {{alarm}}`
+
+- Call a service with arguments:
+
+`cmd {{vibrator}} {{vibrate 300}}`
```
bl-ue, line 15: For clarity?
```suggestion
`cmd {{vibrator}} "{{vibrate 300}}"`
```

pages/linux/settings.md
```diff
@@ -0,0 +1,19 @@
+# settings
+
+> Get information about the Android OS.
+
+- Display a list of settings in the {{global}} namespace:
```
bl-ue, line 5: Until we have a specific rule of using `{{ }}` in the example descriptions, I'd go with backticks.
```suggestion
- Display a list of settings in the `global` namespace:
```

pages/linux/settings.md
```diff
@@ -0,0 +1,19 @@
...
+
+`settings list global`
```
bl-ue, line 7:
```suggestion
`settings list {{global}}`
```

pages/linux/settings.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Get the value of a specific setting:
+
+`settings get {{global}} {{airplane_mode_on}}`
+
+- Set the value of a setting:
+
+`settings put {{system}} {{screen_brightness}} {{42}}`
```
bl-ue, line 15: Here you're using `system`, but everywhere else you use `global`, which _could_ be a little confusing.  Could you use something else, or do you have stronger feelings about it?

bl-ue (changes requested):
Nice.  However, I have a few things:
1. In the future (if it's convenient for you), will you please separate each new page into its own PR?
2. Use more long options (if they exist)
3. `cmd` and `settings` don't have more information links 🤔

Rudloff:
> Use more long options (if they exist)

I tried but `cmd` and `am start` don't seem to have long options.

>  `cmd` and `settings` don't have more information links thinking

I could not find any official documentation page. (Android CLI tools tend to be poorly documented :slightly_frowning_face:)

pages/linux/settings.md
Rudloff: I wanted to show that `global` is not the only namespace available. What would be the best way to do that?

pages/linux/cmd.md
Rudloff: This would not work, because these are separate arguments.

I guess the correct syntax would be this?
```suggestion
`cmd {{vibrator}} {{vibrate}} {{300}}`
```

bl-ue: Ah, I see.  I read `Call a service with arguments` which tells me that the arguments and the amount of them are arbitrary—am I reading that right?

pages/linux/settings.md
bl-ue: Good idea. In that case, I'd keep this example.  How about you change the last example to use a completely different namespace, then (assuming that `system` and `global` aren't the only ones.)

pages/linux/cmd.md
Rudloff: The number of arguments depends on the service command called. Each service has its own set of commands that require different arguments.
Several examples:
```bash
cmd vibrator vibrate 300
cmd netpolicy set restrict-background true
cmd testharness enable
```

bl-ue: Okay, makes sense.  I say leave it as is, then.  I think you did the right in the first place.

bl-ue: Would two dashes work, e.. `cmd vibrator -- vibrate 300`?

Rudloff: It does not seem to work:
```
Unknown command: --
```

bl-ue: Okay, all settled.  Don't make any changes, then, and let's resolve this thread.

bl-ue (approved):

marchersimon:
Should we already move the pages to `android/` or wait for #5462 to be closed?

sbrl:
The first page in the new `android` platform has already been merged by me in #5458, so I'd suggest that if these are pages for commands that are executable in a native shell on an Android device, they belong in the Android platform folder.

marchersimon (approved):

bl-ue (approved):

PR #5482, opened 2021-03-19
Client specification: point changelog to tags, add maintainer note

mebeim:
This converts the links in the changelog of the client spec to point to version tags instead of commit hashes as suggested by @bl-ue in #5428, and also includes a maintainer note in the form of a comment for future reference when updating the changelog. 

I've already pushed the tags from v1.0 to v1.5, which should be visible to everyone [here](https://github.com/tldr-pages/tldr/tags).

### DIFF
--- a/CLIENT-SPECIFICATION.md
+++ b/CLIENT-SPECIFICATION.md
@@ -212,26 +212,36 @@ Caching SHOULD be done according to the user's language configuration (if any), 
 
 ## Changelog
 
+<!--
+Maintainer note:
+
+Keep the changelog links poiting to this document under the appropriate
+`/blob/<version-tag>/...` and also reference the PR which introduced the new
+version. Tagging of the commit with a new version tag (in the form `vX.Y`)
+should be done immediately AFTER merging the version bump, as the commit hash
+changes when merging with squash or rebase.
+-->
+
  - [v1.5, March 17th 2021](https://github.com/tldr-pages/tldr/blob/v1.5/CLIENT-SPECIFICATION.md) ([#5428](https://github.com/tldr-pages/tldr/pull/5428))
    - Add requirement for converting command names to lowercase before running the page resolution algorithm.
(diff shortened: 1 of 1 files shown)


bl-ue:
Also see my comment here: https://github.com/tldr-pages/tldr/pull/5428#issuecomment-803161951

CLIENT-SPECIFICATION.md
```diff
@@ -212,26 +212,36 @@ Caching SHOULD be done according to the user's language configuration (if any),
 
 ## Changelog
 
+<!--
+Maintainer note:
+
+Keep the changelog links poiting to this document under the appropriate
```
bl-ue, line 218:
```suggestion
Keep the changelog links pointing to this document under the appropriate
```

mebeim:
Quoting your comment from #5428 just to keep the discussion here:

> Hmm...I didn't get notified about this. @mebeim I think we need to create an actual release, not just a tag thinking
>
> I actually get GH notifications when I'm watching a repository and they create a release. For example, I was notified when https://github.com/kubernetes/kubernetes/releases/tag/v1.20.5 was released.

Makes sense. I should add this to the maintainer note in the comment then. Should we go ahead and just create a v1.5 release? Previous versions don't strictly need an actual release, but we could do that too.

bl-ue:
Yes, I think we should. 👍🏻  No one needs to be notified about previous spec versions because the changes made in v1.0-1.4 are incorporated into v1.5.

mebeim:
Done and done :+1:

bl-ue (approved):

navarroaxel (approved):

sbrl (approved):

sbrl:
Is there anything else specific that needs doing after we merge this?

bl-ue:
I don't think so, @sbrl. 👍🏻

PR #5483, opened 2021-03-20
nms: add page

goznalo-git:

Hi there, 

this is my first PR to this repository, I think it complies with all guidelines (not sure whether it is `linux` or `common`, I'd say the former, hence I've placed files there). 

I have added two pages, an English one and a Spanish translation, of the usage of `nms`, also known as `no-more-secrets`, which is a fun command providing a text decryption effect. You can find a link to its official repo in the files themselves.

Please let me know if it requires some tweaks.


bl-ue:
Nice pages @goznalo-git, and this certainly a fun tool!   It looks like this tool is available on Linux, macOS, and Windows, so the pages should definitely go in the `common/` directory.  Will you please move them there?

goznalo-git:
Hi @bl-ue, thank you for your feedback!

I just moved it as you requested.

bl-ue:
FYI—I'm working on a review.

pages/common/nms.md
```diff
@@ -0,0 +1,20 @@
+# nms
+
+> A cool (piped) text decryption effect.
+> More information: <https://github.com/bartobri/no-more-secrets>.
+
+- Decrypt "Hello, World!" after a keystroke:
+
+`echo "Hello, World! | nms"`
+
+- Decrypt the output of `ls -la` automatically:
```
bl-ue, line 10: This sort of 'automatic' behavior, while maybe not default to this command, is generally expected from most commands, so 'automatically' is more likely to confuse the user.
```suggestion
- Decrypt the output of `ls -la`:
```

pages/common/nms.md
```diff
@@ -0,0 +1,20 @@
...
+
+`ls -la | nms -a`
+
+- Decrypt the content of `secret_message.txt` automatically with red output:
```
bl-ue, line 14: We try to be generic when possible, using simply 'a file' when it will work just as well as an actual file name.
```suggestion
- Decrypt the content of a file automatically with a custom output color:
```

pages/common/nms.md
```diff
@@ -0,0 +1,20 @@
...
+
+`cat secret_message.txt | nms -a -f red`
```
bl-ue, line 16: If the list is short enough, include the whole list of valid values separated by `|` (I put the default `blue` first, otherwise thie order is that found in the man page):
```suggestion
`cat {{path/to/file}} | nms -a -f {{blue|white|yellow|black|magenta|green|red}}`
```

pages/common/nms.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Clear the screen prior to decrypt `git status` automatically:
+
+`git status | nms -a -c`
```
bl-ue, line 20: I think it would be better to just use 'command' here
```suggestion
- Clear the screen before decrypting:

`{{command}} | nms -a -c`
```

pages/common/nms.md
```diff
@@ -0,0 +1,20 @@
...
+> A cool (piped) text decryption effect.
+> More information: <https://github.com/bartobri/no-more-secrets>.
+
+- Decrypt "Hello, World!" after a keystroke:
+
+`echo "Hello, World! | nms"`
+
+- Decrypt the output of `ls -la` automatically:
+
+`ls -la | nms -a`
```
bl-ue, line 12:
```suggestion
`{{ls -la}} | nms -a`
```

bl-ue (changes requested):
Here's my review.  We can discuss points if you disagree with any of them.

Also, will you please update the translation? I would have submitted suggestioned for it as well, but since I changed the grammar, I thought it would be easier for you to do the whole thing.

bl-ue:
Sorry for all of the suggestions :) Free-form commands like this (I mean ones that simply operate on stdin) can have a wide variety of usages and it's not always easy to document them in a consistent style.

pages/common/nms.md
goznalo-git: I agree with you that people usually expect this behavior by default, and for instance in the next examples I think you're right about ignoring "automatically" in their description. However, if in the first example "nms" is used, and in this one "nms -a" is used, I would say people will be more confused if not explained what's the difference.

goznalo-git (commented):
Dear @bl-ue , don't be sorry about all the suggestions, I appreciate the time you've spent cleaning my mess haha. I knew I would have to made some changes, given that it was my first PR to this repo, but I did not expect making so many mistakes, apologies for that.

goznalo-git:
I will now translate the changes to the Spanish version to then push them too.

pages/common/nms.md
bl-ue: Ah, good point.  In that case, what do you think showing the 'automatic' behavior first and the delayed behavior later?

goznalo-git: Honestly, I think we should leave it as is. It is a problem of the command itself, as you say the default behavior should be automatically taking place, with an option to make it non automatic, however this is reversed, leading to confusion.

IMO there's no simple non-confusing way to fix it here, if we do as you say (first showcasing the automatic behavior, then not) it will a bit awkward. Still, if you insist I have no problem changing it.

It would be nice to hear other opinions as well.

bl-ue: >It would be nice to hear other opinions as well.

I agree.  Applying the other suggestions would make this one much more noticeable ;)

bl-ue:
>but I did not expect making so many mistakes, apologies for that.

They're not mistakes; please don't apologize!  It's not immediately obvious (unless you spend time pouring over past PRs) how consistent we try to get the pages.

goznalo-git:
Hello again, I've added few changes as we discussed, and also in the Spanish version. The only thing left would be for someone else to give us their opinion on the -a trigger.

pages/common/nms.md
```diff
@@ -0,0 +1,20 @@
+# nms
+
+> Command line tool that recreates the famous data decryption effect seen in the 1992 movie Sneakers from stdin.
+> More information: <https://github.com/bartobri/no-more-secrets>.
+
+- Decrypt text after a keystroke:
+
+`echo "{{Hello, World!}}" | nms`
+
+- Decrypt output, automatically:
```
sbrl, line 10: Hey there - welcome to tldr-pages! Great first page here. Reading the README for the project behind this command, the `automatically` does seem to hold meaning here. What about this instead to aid clarity?

```suggestion
- Decrypt output immediately, without waiting for a keystroke:
```

goznalo-git: Hi @sbrl, thanks for the welcome! I like your suggestion quite a lot, what about you @bl-ue? Would it settle the debate?

navarroaxel: Yeah, sounds like the user doesn't need to trigger the animation with `space`, or any key. Makes sense.

bl-ue: +1 for @sbrl's suggestion—kills all the birds with one stone :D

pages.es/common/nms.md
```diff
@@ -0,0 +1,20 @@
+# nms
+
+> Herramienta de línea de comandos que recrea el famoso efecto de desencriptado de datos de la película Sneakers (1992).
+> Más información: <https://github.com/bartobri/no-more-secrets>.
+
+- Desencriptar texto tras presionar una tecla:
+
+`echo "{{Hola, Mundo!}}" | nms`
+
+- Desencriptar output, inmediatamente, sin esperar a que una tecla sea pulsada:
```
navarroaxel, line 10: Aunque a mí y a ti nos suene mal `la salida`, hay libros de informatica en Español así :man_shrugging:  : 
```suggestion
- Desencriptar la salida inmediatamente, sin esperar a que una tecla sea pulsada:
```
`la salida` is more closer to `the exit` in English :joy:, but the Spanish books that I have read use that translation :man_facepalming:

navarroaxel (changes requested):

navarroaxel (approved):

pages.es/common/nms.md
goznalo-git: Well, if textbooks say so, who am I to disagree? 😅

navarroaxel: That's why I don't read tech stuff in Spanish.

bl-ue (approved):

bl-ue:
Welcome to tldr-pages @goznalo-git 🥇  (this is your first PR)

goznalo-git:
Won't be the last for sure ;)

PR #5485, opened 2021-03-20
poweroff, halt, reboot: add examples

marchersimon:


pages/linux/poweroff.md
```diff
@@ -1,7 +1,12 @@
...
 
 > Shutdown the system.
+> More information: <https://man7.org/linux/man-pages/man8/reboot.8.html>.
 
-- Poweroff the system:
+- Power off the system:
+
+`poweroff`
+
+- Send an ACPI hardware signal to initiate a complete and immediate shutdown:
```
navarroaxel, line 10: This new example doesn't add something new, only explain how poweroff works. I think you can work over the `--halt` vs `--poweroff` arguments to add something new, or explain `--wtmp-only`. etc.

navarroaxel (changes requested):

pages/linux/poweroff.md
marchersimon: But isn't there a difference between `poweroff` and `sudo poweroff`?

navarroaxel: I think the difference is between `halt`, `poweroff` and `shutdown` shutdown sequences.

pages/linux/halt.md
```diff
@@ -1,16 +1,24 @@
 # halt
 
-> Halt, power-off or reboot the machine.
-> More information: <https://www.freedesktop.org/software/systemd/man/halt.html>.
+> Halt the system.
+> More information: <https://man7.org/linux/man-pages/man8/reboot.8.html>.
```
navarroaxel, line 4: Why reboot here?
```suggestion
> More information: <https://www.man7.org/linux/man-pages/man8/halt.8.html>.
```

navarroaxel (changes requested):
could you add the more info link to the `shutdown` page? https://www.man7.org/linux/man-pages/man8/shutdown.8.html

bl-ue:
Based on the description in the man pages, these 3 commands appear to all be identical, and thus it would seem that they ought to be aliased to one another.

From the man pages:
> halt, poweroff, reboot may be used to halt, power-off, or reboot the machine. All three commands take the same options.

navarroaxel:
> Based on the description in the man pages, these 3 commands appear to all be identical, and thus it would seem that they ought to be aliased to one another.

@bl-ue but they have differences:
https://askubuntu.com/questions/64995/what-is-the-difference-between-shutdown-and-poweroff-commands#64999
https://www.tecmint.com/shutdown-poweroff-halt-and-reboot-commands-in-linux/

bl-ue:
Ah, okay, nevermind then.

sbrl (approved):
Thanks, @marchersimon!

pages/linux/halt.md
marchersimon: Because that url covers all three commands and I didn't know there was a url for reboot too

navarroaxel (approved):

sbrl:
Why the confused emoji, @navarroaxel? I might have missed something here?

navarroaxel:
@sbrl Because I didn't understand your comment, but it was just a bug in the GitHub app for Android :joy: 
I wondered why you quoted the answer from @bl-ue.
https://user-images.githubusercontent.com/1536740/112724431-12964580-8ef2-11eb-91aa-567d3c89de68.mp4

PR #5496, opened 2021-03-24
chmod: add more info link and update French translation

patricedenis:


bl-ue (approved):
Love changes like this ⚡

By the way, @patricedenis, you can see a nice overview of which translations need to be made (and which ones need updating) here: https://lukwebsforge.github.io/tldri18n/

patricedenis:
yes, thanks ! I was on this page, that's why I committed here !

bl-ue:
Just curious: how did you find that page? https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#translations, perhaps?

patricedenis:
yes absolutely. I was reading the contributing page and I thought I could make some first easy contributions to translate some already done pages. I need to know more the syntax of markdown as well so it's a training session !  :-)

navarroaxel (approved):

bl-ue:
I'm asking because I was afraid that the place that I put that link was too hidden and people wouldn't notice it—thankfully that's not the case though!

PR #5501, opened 2021-03-24
basename: update French translation

patricedenis:
- [ ] The page (if new), does not already exist in the repo.
- [ ] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [ ] The page has 8 or fewer examples.
- [ ] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [ ] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [ ] The page description includes a link to documentation or a homepage (if applicable).

Please consider merging this pull request.
Also, can someone explain me how can I add the "translation" label to the pull request in order to sort it better?

PR #5505, opened 2021-03-25
at, ifdown, ifup, ip-address, ip: add French translation

patricedenis:

### DIFF
--- a/pages/linux/ifdown.md
+++ b/pages/linux/ifdown.md
@@ -1,6 +1,7 @@
 # ifdown
 
 > Disable network interfaces.
+> More information: <https://linux.die.net/man/8/ifdown>.
 
 - Disable interface eth0:
 
--- a/pages.fr/common/at.md
+++ b/pages.fr/common/at.md
@@ -1,0 +1,17 @@
+# at
+
+> Planifie l'exécution d'une commande une fois à un moment donné.
+> Le service atd (ou atrun) doit être actif pour l'exécution des commandes plannifiées.
+> Plus d'information : <https://www.man7.org/linux/man-pages/man1/at.1p.html>.
+
+- Planifie l'exécution de la commande donnée dans l'entrée standard dans 5 minutes (Appuyer sur `Ctrl + D` une fois la commande inscrite) :
+
+`at now + 5 minutes`
+
+- Planifie l'exécution d'une commande depuis l'entrée standard (impression echo redirigée dans un tube) aujourd'hui à 10h00 :
+
+`echo "{{./ma_commande.sh}}" | at 1000`
+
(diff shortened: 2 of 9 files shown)


patricedenis:
@bl-ue : #5501 
> 
@patricedenis only repository collaborators and organization members can add/remove issue/PR labels. Seeing "translation" in the PR title thought makes it very easy to identify it as a translation and so it's not much work to label, but thank you for the offer!
>
OK I did'nt know that :smiley: 
>
I'm not sure if GitHub supports PR templates like it does issue templates. If it did, we could create PR templates for translations, page edits, new commands, etc., and have labels automatically added when a user creates a PR. With issue templates, you can configure which labels should automatically be applied when a user opened an issue thinking
> 

All right, I think then I may need to move to a CLI git client to continue. I'm thinking that it seems really a gain in time to apply those kind of templates. I already use git for my local files and I did know about templates but never put them in use. Have you already written some .gitmessage templates that auto detect the labels that should be used into the commit message?

bl-ue:
Actually, that was just wishfull thinking on my part.  I was refering to a GitHub-only (it's not part of plain Git 😉) feature documented [here](https://docs.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates#issue-templates).

pages/linux/ifdown.md
```diff
@@ -1,6 +1,7 @@
 # ifdown
 
 > Disable network interfaces.
+> More information: <https://linux.die.net/man/8/ifdown>.
```
sbrl, line 4: I think generally we're moving to man7.org links here, but this is a very interesting case where linux.die.net has the man page, but man7.org doesn't.

`at` and `ip` do (and I see this PR correctly links to those), but `ifup` and `ifdown` don't.

I can't remember why we prefer man7 over linux.die.net though.

sbrl (approved):

pages/linux/ifdown.md
bl-ue: #5143. Per @mebeim:

>[man7.org] IMO renders the pages a lot better (sometimes die.net pages have syntax errors due to special characters being treated strangely when rendered in HTML, furthermore man7.org pages are actually in monospace).

bl-ue: Let's just leave it for now.  I'll update it later perhaps with the rest of them (if I can find an alternative.)

bl-ue (approved):

pages/linux/ifdown.md
patricedenis: @sbrl : yes that's why I put the die.net links here because I could not find them in man7.org ^^

PR #5507, opened 2021-03-25
add new animated svg of tar command

### LINKED ISSUE #5400: `tar` example in README.md is out of date
If you run `tldr tar`, the result is different than what appears on the README. Obviously not a big deal, but maybe the screenshot should be updated / replaced with text? If you wanted to go all out you could add something to the linter whenever someone changed that file.

tomadojuice:
Closes #5400 
add animated svg of new tar command in readme


tomadojuice:
This is just an example, could you maybe give some feedback? I think the font size is a little to small. What do you think of the animation length, is it too short?

marchersimon:
@tomadojuice I think the length of the animation is fine, just the time from the empty terminal to the page showing is a bit long. Also, the small font size in combination with the green/cyan colors might make it hard to read for some. I'd use the base16 theme (`tldr -t base16`) with some more distinctive colors.

sbrl:
I would agree that the font size is too small too

bl-ue:
@tomadojuice what client are you using (just for reference)? Node.js?

tomadojuice:
@bl-ue I'm using the brew client. I read it's outdated, should I rather use node?

tomadojuice:
So I adjusted the colours to base16 (it took some time because I had to edit the SVG manually). I also changed the length of the animation a little.

bl-ue (approved):
There's two svgs though, [`images/tldr-tar.svg`](https://github.com/tldr-pages/tldr/pull/5507/files#diff-fe6b1b264cfb8721f6a8c0cc0d3144b5fd9897385c3fc5d26b4768fce5145499) and [`images/tldr.svg `](https://github.com/tldr-pages/tldr/pull/5507/files#diff-ce09bd173d8488e82dd538fc3a09c49fd8abed41d96ee38e43d31dbd69525834) 🤔

tomadojuice:
@bl-ue Oh right, I forgot to delete the old one. Thanks for noticing :)

marchersimon (approved):

sbrl:
Looks much better! Was the font size increased at all? I'm not sure I can tell any difference (though I don't have anything to compare it against on-hand).

tomadojuice:
@sbrl It's slightly increased, I cannot make it to big, since the complete output wouldn't be visible anymore. I think increasing the size of the graphic itself wouldn't be a solution either.

sbrl (approved):
Thanks, @tomadojuice!

Waples (approved):

PR #5443, opened 2021-03-14
tldr: use platform-specific command in --platform examples

Waples:
- [ ] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

Closes #5401

Issue #5401, opened 2021-03-10
Change "tldr tldr" -p option to something that is platform specific?


jakevossen5:
Currently the `tldr tldr` page has the -p example of `tldr -p {{linux}} {{tar}}`. Maybe there is a rational I don't see, but I feel like the -p option would be better explained if it showed an example of a page specific to linux in this case (the `tar` page is in `common`).

I am not sure of a command that is well known to be different on different platforms (`date` is what first came to mind).

What does everyone think?

sbrl:
Great spot. What about changing it to `tldr -p {{linux}} {{pstree}}` or something?

marchersimon:
Closing, since this example has been rewritten to include all platforms now.

PR #5508, opened 2021-03-26
tldr: edit examples

marchersimon:
I also wanted to add examples for `--theme` and `--search`, but I'm not sure if other clients also support that.


bl-ue:
@marchersimon does this have anything to do with #5401?

sbrl:
Hey there, @marchersimon! Thanks for the edit. Unfortunately, this is one of the very few pages that we can't use long-form arguments on, because they are optional in our [client spec](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md).

marchersimon:
@bl-ue not on purpose, but it seems like this addresses that issue. However there already is a PR (#5443) for that. Should we adapt the changes of that PR here and close it or should we wait for it to be merged?

@sbrl oh, sure, I'll change it

marchersimon:
Btw, why is there a `-p` option? There isn't such an option for the node client, instead the option for changing the platform is `-o`/`--os`.

sbrl:
`-p` is what's specified in the [client spec](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#arguments) - it's short for `--platform` @marchersimon

Sounds like that's a spec compliance issue with the Node.js client - great catch!

Would you like to open an issue over there about it?

marchersimon:
Sure

pages/common/tldr.md
```diff
@@ -3,15 +3,15 @@
...
-- Get typical usages of a command (hint: this is how you got here!):
+- Show the tldr page for a command (hint: this is how you got here!):
 
 `tldr {{command}}`
 
-- Show the tar tldr page for Linux:
+- Show the tldr page for `cd`, overriding the default platform:
 
-`tldr -p {{linux}} {{tar}}`
+`tldr -o {{windows}} {{cd}}`
```
bl-ue, line 12:
```suggestion
`tldr -p {{windows}} {{cd}}`
```
https://github.com/tldr-pages/tldr-node-client/pull/337

marchersimon: There was a reason, we don't use `--platform` and `--update` here, right?

bl-ue: Yep: https://github.com/tldr-pages/tldr/pull/5508#issuecomment-808312905

marchersimon: Ah, I knew there was something.

sbrl (approved):
Thanks, @marchersimon!

sbrl:
Trivia: The `windows` platform was added much later than the other platforms. It's now the 2nd newest platform after the recent `android` addition.

bl-ue (approved):

Issue #5510, opened 2021-03-26
Epic: Add links to all existing pages


waldyrious:
Since #2649, tldr pages support adding a link to the tool's homepage, and [substantial efforts to backfill links into existing pages](https://github.com/tldr-pages/tldr/pulls?q=is%3Apr+sort%3Aupdated-desc+multiple+pages%3A+add+homepages+is%3Aclosed) have been made, but many are still lacking it.

We should ensure that all pages have a link that people can follow to learn more about the tool and explore more comprehensive documentation. Pages of projects that don't have a dedicated web page should either link to their parent project's page, or we should come up with a way to explicitly indicate that they don't have a homepage.

(Note: This should be done with a separate PR for each page (or at best, multi-page PRs for related sets of pages, such as all the subcommands of a given command), so that discussion about which links to use for some pages won't block the straightforward changes to the other pages.)

Once that work is completed, we should add a check to [tldr-lint](https://github.com/tldr-pages/tldr-lint) to make this field required, thus ensuring that new pages won't regress our coverage in that regard (see #5046 for examples of link-related checks that were incorporated into tldr-lint). We should also change the template for a page in `CONTRIBUTING.md`, the style guide, and other places where that may be relevant.

navarroaxel:
Question: What we should do with pages like https://github.com/tldr-pages/tldr/blob/master/pages/linux/ksvgtopng5.md? This is an internal command/util from the KDE Plasma desktop and doesn't have a man page. Should we use the link to the git repository?

bl-ue:
I'd say the Git repository for that tool (or group of related tools), or the parent project's page (e.g. the KDE Plasma Desktop's page).  We link to Git repos (e.g. [GitHub](https://github.com/tldr-pages/tldr/search?q=more+information+github.com), [GitLab](https://github.com/tldr-pages/tldr/search?q=more+information+gitlab.com)) a lot.

marchersimon:
- [ ] android/settings
- [ ] android/cmd
- [ ] common/fortune
- [ ] common/ed
- [ ] common/where
- [ ] common/calendar
- [ ] common/sed
- [ ] common/for
- [ ] common/banner
- [ ] common/sha512sum
- [ ] common/nl
- [ ] common/jhat
- [ ] common/cd
- [ ] common/fmt
- [ ] common/batch
- [ ] common/waitress-serve
- [ ] common/echo
- [ ] common/mkfifo
- [ ] common/cmp
[...]

PR #5511, opened 2021-03-26
ksvgtopng5: add more information link

navarroaxel:

For: #5510

### DIFF
--- a/pages/linux/ksvgtopng5.md
+++ b/pages/linux/ksvgtopng5.md
@@ -1,6 +1,7 @@
 # ksvgtopng5
 
 > Convert SVG files to PNG format.
+> More information: <https://invent.kde.org/plasma/kde-cli-tools/-/blob/master/ksvgtopng/ksvgtopng.cpp>.
 
 - Convert an SVG file (should be an absolute path) to PNG:
 


bl-ue (approved):
I'm not sure what others would say about linking to a `.cpp` file, but I, for one, think it's quite a good source of extra information for the user! :)

marchersimon:
How about [http://manpages.org/ksvgtopng](http://manpages.org/ksvgtopng)? (Unfortunately there's no `https` version)

bl-ue:
That page doesn't seem to give more information than this tldr page does though, besides just feeling a bit odd.

marchersimon:
Yeah, that's right. In that case, I'd actually prefer [https://invent.kde.org/plasma/kde-cli-tools/-/tree/master/ksvgtopng](https://invent.kde.org/plasma/kde-cli-tools/-/tree/master/ksvgtopng), since it gives an overview of the project, instead of jumping directly into the source code.

sbrl (approved):
Hrm, a difficult one.

<http://manpages.org/ksvgtopng> suggests that Debian has a man page for it, but when you check <https://manpages.debian.org/> it claims it can't find one.

I think linking to the .cpp file is the best trade-off here. Linking to the directory it is in adds an extra click for the user, because said directory doesn't contain a README.

marchersimon (approved):

Issue #5517, opened 2021-03-28
Let's document: platformio


marchersimon:
- [x] [`pio`](https://docs.platformio.org/en/latest/core/userguide/index.html) → #5404
- [x] [`pio access`](https://docs.platformio.org/en/latest/core/userguide/access/index.html) → #5518 
- [x] [`pio account`](https://docs.platformio.org/en/latest/core/userguide/account/index.html) → #5410
- [x] [`pio boards`](https://docs.platformio.org/en/latest/core/userguide/cmd_boards.html) → #5409 
- [x] [`pio check`](https://docs.platformio.org/en/latest/core/userguide/cmd_check.html) → #5519 
- [ ] [`pio ci`](https://docs.platformio.org/en/latest/core/userguide/cmd_ci.html) → #6186 
- [x] [`pio debug`](https://docs.platformio.org/en/latest/core/userguide/cmd_debug.html) → #5780 
- [x] [`pio device`](https://docs.platformio.org/en/latest/core/userguide/device/index.html) → #5725
- [x] [`pio home`](https://docs.platformio.org/en/latest/core/userguide/cmd_home.html) → #5406 
- [x] [`pio lib`](https://docs.platformio.org/en/latest/core/userguide/lib/index.html) → #5728
- [x] [`pio org`](https://docs.platformio.org/en/latest/core/userguide/org/index.html) → #5525
- [x] [`pio package`](https://docs.platformio.org/en/latest/core/userguide/package/index.html) → #5713 
- [x] [`pio platform`](https://docs.platformio.org/en/latest/core/userguide/platforms/index.html) → #5727
- [x] [`pio project`](https://docs.platformio.org/en/latest/core/userguide/project/index.html) → #5408 
- [x] `pio init` → #5865
- [x] [`pio remote`](https://docs.platformio.org/en/latest/core/userguide/remote/index.html) → #5902 
- [x] [`pio run`](https://docs.platformio.org/en/latest/core/userguide/cmd_run.html) → #5827
- [x] [`pio settings`](https://docs.platformio.org/en/latest/core/userguide/cmd_settings.html) → #5522
- [x] [`pio system`](https://docs.platformio.org/en/latest/core/userguide/system/index.html) → #5523
- [x] [`pio team`](https://docs.platformio.org/en/latest/core/userguide/team/index.html) → #5524
- [x] [`pio test`](https://docs.platformio.org/en/latest/core/userguide/cmd_test.html) → #5815
[...]

PR #5525, opened 2021-03-28
pio-org: add page

marchersimon:
Part of #5517


pages/common/pio-org.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Add a user to an organization:
+
+`pio org add {{orgname}} {{username}}`
+
+- Remove a user from an organization:
+
+`pio org remove {{orgname}} {{username}}`
+
+- List all organizations and their owners, the user owns:
```
bl-ue, line 22: This command does just what I'm saying in the suggestion, right? That's what the [docs](https://docs.platformio.org/en/latest/core/userguide/org/cmd_list.html#description) say.
```suggestion
- List all organizations and their owners:
```

pages/common/pio-org.md
```diff
@@ -0,0 +1,28 @@
...
+
+`pio org list`
+
+- Update the name, email and display name of an organization:
```
bl-ue, line 26:
```suggestion
- Update the name, email or display name of an organization:
```

bl-ue (changes requested):
I've left some comments.  Also @marchersimon: I forgot to discuss this in the other PRs (now merged, so we'll need to open a new PR to correct those), but will you please change the variable names to the longer formats in this and your other open PIO PRs, i.e.

- `orgname` → `organization_name`
- `new_orgname` → `new_organization_name`
- `new_displayname` → `new_display_name`
- etc.

pages/common/pio-org.md
marchersimon: Here too, there are far more organizations, than the user owns and this suggestion would imply that it lists all there are.

marchersimon:
@bl-ue those terms don't occur in the already merged PRs, so there's no need for a new one.

pages/common/pio-org.md
bl-ue: It doesn't do that?  The docs appear to me to say that.

marchersimon: There are many mistakes and bad descriptions in the docs. But at least when I run `pio org list`, I get `You do not have any orginazations`

bl-ue: How about this?
```suggestion
- List all organizations the current user is a member of and their owners:
```

bl-ue (approved):

navarroaxel (approved):

PR #5529, opened 2021-03-29
less: add more information link

navarroaxel:

I think we should use the official website because is header to find the a link to a man page.
Another option is the official repo: https://github.com/gwsw/less.

For #5510

### DIFF
--- a/pages/common/less.md
+++ b/pages/common/less.md
@@ -1,6 +1,7 @@
 # less
 
 > Open a file for interactive reading, allowing scrolling and search.
+> More information: <https://greenwoodsoftware.com/less/>.
 
 - Open a file:
 


einverne (approved):

marchersimon (approved):

pages/common/less.md
```diff
@@ -1,6 +1,7 @@
 # less
 
 > Open a file for interactive reading, allowing scrolling and search.
+> More information: <https://greenwoodsoftware.com/less/>.
```
marchersimon, line 4: Should we remove slashes at the end or leave them? IMO, it looks better without.

navarroaxel: I prefer to avoid the redirection just by 1 slash, but there are redirects in the tldr links

bl-ue: A discussion is in order on when and how much to simplify links; it's been bugging me a lot.

PR #5530, opened 2021-03-29
settings: add more information link

marchersimon:
For #5510 
The hash in the url is unfortunate, but I couldn't find anything else, which is slightly related.

### DIFF
--- a/pages/android/settings.md
+++ b/pages/android/settings.md
@@ -1,6 +1,7 @@
 # settings
 
 > Get information about the Android OS.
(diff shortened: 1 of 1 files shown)


bl-ue (approved):
I doubt the hash will change.

sbrl (approved):
Hrm yeah, this seems like the nest we can find. It's certainly far from perfect (it prefixes with `adb shell` for one), but it seems like there's no better documentation available. Let's use this then.

navarroaxel (approved):

PR #5537, opened 2021-03-29
banner: add more information link

marchersimon:
For #5510


bl-ue (approved):
Why archlinux.org everywhere though?

marchersimon:
Because of its advandages @waldyrious pointed out in https://github.com/tldr-pages/tldr/issues/5143#issuecomment-808998349 and because I think it looks pretty nice. Also it covers a lot of pages, so our links can be a bit more uniform.

navarroaxel (approved):

PR #5552, opened 2021-03-29
rm: add more information link

marchersimon:
For #5510

### DIFF
--- a/pages.es/common/rm.md
+++ b/pages.es/common/rm.md
@@ -1,6 +1,7 @@
 # rm
 
 > Elimina archivos o directorios.
+> Más información: <https://www.gnu.org/software/coreutils/manual/html_node/rm-invocation.html>.
 
 - Elimina archivos de ubicaciones arbitrarias:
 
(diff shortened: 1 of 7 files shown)


bl-ue (approved):
Did my tool work? 🙂

marchersimon:
@bl-ue absolutely, I love it. Would be cool if it would outomatically perform a `git add`.

bl-ue:
I thought of that.  Would it be convenient for it to commit as well?

marchersimon:
I don't think so. When I do it myself I can check if I'm on the right branch and write my own commit message. Also, I have to commit only once.

bl-ue:
You can also use `git add .`, sounds like maybe you don't know that.

marchersimon:
I have many untracked files (like scripts or todolists) and to make sure I don't add anything by mistake I try to avoid this.

bl-ue:
Gotcha. Will do.

bl-ue:
Done. 🎉

yutyo (approved):
LGTM

PR #5582, opened 2021-03-30
fusermount: add page

marchersimon:


pages/common/fusermount.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Unmount a FUSE filesystem:
+
+`fusermount -u {{path/to/mount_point}}`
+
+- Unmount a FUSE filesystem, as soon as it's not busy anymore:
+
+`fusermount -z {{path/to/mount_point}}`
+
+- Print the version of `fusermount` and exit:
```
navarroaxel, line 14:
```suggestion
- Display version:
```

pages/common/fusermount.md
```diff
@@ -0,0 +1,16 @@
+# fusermount
+
+> Mount and unmount FUSE filesystems.
```
navarroaxel, line 3: I can´t find an example for mount using this command :thinking:
Maybe we should update the command description only for `unmount`.

marchersimon: There is a `-o` option which expects mount options, so it should at least be possible to mount filesystems. But I couldn't find those options anywhere.

marchersimon: Maybe those are the same options for `mount`, but that's just an asumption.

pages/common/fusermount.md
```diff
@@ -0,0 +1,16 @@
...
+> More information: <https://man.archlinux.org/man/fusermount.1>.
+
+- Unmount a FUSE filesystem:
+
+`fusermount -u {{path/to/mount_point}}`
+
+- Unmount a FUSE filesystem, as soon as it's not busy anymore:
```
bl-ue, line 10:
```suggestion
- Unmount a FUSE filesystem as soon as it becomes unused:
```

pages/common/fusermount.md
bl-ue: The page doesn't need to have all possible examples, so we can leave it for someone else to pick up later.

bl-ue: I think so.  It might even prompt a reader to say "why not?" and then fix it. 🤷🏻

navarroaxel (approved):

PR #5640, opened 2021-03-30
at, atq, atrm, batch: add more information link

bl-ue:
For #5510


marchersimon:
Should we maybe use eg. <https://man.archlinux.org/man/atq.1.en> for the `atq` pages and so on? Altough the end result is the same.

marchersimon (approved):

bl-ue:
They redirect, so I didn't do that.  I'll let others weigh in.  Definitely of course omit the `.en` ending :)

navarroaxel (approved):

PR #5646, opened 2021-03-30
windows/*: add links; sigverif: remove page

marchersimon:
For #5510

### DIFF
--- a/pages.pt_BR/windows/sigverif.md
+++ b/pages.pt_BR/windows/sigverif.md
@@ -1,8 +1,0 @@
-# sigverif
-
-> Uma interface gráfica para verificação de assinaturas de arquivos do sistema.
-> Mais informações: <https://www.thewindowsclub.com/identify-unsigned-drivers-sigverif-windows>.
-
-- Abre a interface de verificação de assinatura:
-
-`sigverif`
--- a/pages.pt_BR/windows/sigverif.md
+++ b/pages.pt_BR/windows/sigverif.md
@@ -1,6 +1,7 @@
 # sigverif
 
 > Uma interface gráfica para verificação de assinaturas de arquivos do sistema.
+> Mais informações: <https://www.thewindowsclub.com/identify-unsigned-drivers-sigverif-windows>.
 
 - Abre a interface de verificação de assinatura:
 
--- a/pages.zh/windows/iscc.md
+++ b/pages.zh/windows/iscc.md
@@ -2,6 +2,7 @@
 
 > Inno Setup 安装程序的编译器.
 > 它将 Inno Setup 脚本编译为 Windows 安装程序可执行文件.
+> 更多信息： <https://jrsoftware.org/isinfo.php>.
 
 - 编译一个 Inno Setup 脚本:
 
--- a/pages.zh/windows/sigverif.md
+++ b/pages.zh/windows/sigverif.md
@@ -1,8 +1,0 @@
-# sigverif
-
-> 用于检查系统文件的 GUI 签名验证工具.
-> 更多信息： <https://www.thewindowsclub.com/identify-unsigned-drivers-sigverif-windows>.
-
-- 打开文件签名验证界面:
-
-`sigverif`
--- a/pages.zh/windows/sigverif.md
+++ b/pages.zh/windows/sigverif.md
@@ -1,6 +1,7 @@
 # sigverif
 
 > 用于检查系统文件的 GUI 签名验证工具.
+> 更多信息： <https://www.thewindowsclub.com/identify-unsigned-drivers-sigverif-windows>.
 
 - 打开文件签名验证界面:
 
--- a/pages/windows/iscc.md
+++ b/pages/windows/iscc.md
@@ -2,6 +2,7 @@
 
 > Compiler for Inno Setup installers.
 > It compiles an Inno Setup scripts into an Windows installer executable.
+> More information: <https://jrsoftware.org/isinfo.php>.
 
 - Compile an Inno Setup script:
 
--- a/pages.zh/windows/explorer.md
+++ b/pages.zh/windows/explorer.md
@@ -1,6 +1,7 @@
 # explorer
 
 > Windows 文件资源管理器.
+> 更多信息： <https://ss64.com/nt/explorer.html>.
 
 - 打开 Windows 文件资源管理器:
 
--- a/pages/windows/explorer.md
+++ b/pages/windows/explorer.md
@@ -1,6 +1,7 @@
 # explorer
 
 > The Windows File Explorer.
+> More information: <https://ss64.com/nt/explorer.html>.
 
 - Open Windows Explorer:
 
--- a/pages/windows/sigverif.md
+++ b/pages/windows/sigverif.md
@@ -1,8 +1,0 @@
-# sigverif
-
-> A GUI signature verification tool for checking system files.
-> More information: <https://www.thewindowsclub.com/identify-unsigned-drivers-sigverif-windows>.
-
-- Open the File Signature Verification interface:
-
-`sigverif`
--- a/pages/windows/sigverif.md
+++ b/pages/windows/sigverif.md
@@ -1,6 +1,7 @@
 # sigverif
 
 > A GUI signature verification tool for checking system files.
+> More information: <https://www.thewindowsclub.com/identify-unsigned-drivers-sigverif-windows>.
 
 - Open the File Signature Verification interface:
 


pages/windows/iscc.md
```diff
@@ -2,6 +2,7 @@
 
 > Compiler for Inno Setup installers.
 > It compiles an Inno Setup scripts into an Windows installer executable.
+> More information: <https://github.com/iscc/iscc-cli>.
```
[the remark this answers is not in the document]
marchersimon: How about <https://jrsoftware.org/isinfo.php>?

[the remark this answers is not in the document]
bl-ue: I wish it could be more specific, but I think it's fine.
```suggestion
> More information: <https://jrsoftware.org/isinfo.php>.
```

pages.zh/windows/iscc.md
```diff
@@ -2,6 +2,7 @@
 
 > Inno Setup 安装程序的编译器.
 > 它将 Inno Setup 脚本编译为 Windows 安装程序可执行文件.
+> 更多信息： <https://github.com/iscc/iscc-cli>.
```
bl-ue, line 5:
```suggestion
> 更多信息： <https://jrsoftware.org/isinfo.php>.
```

pages.pt_BR/windows/sigverif.md
```diff
@@ -1,6 +1,7 @@
 # sigverif
 
 > Uma interface gráfica para verificação de assinaturas de arquivos do sistema.
+> Mais informações: <https://www.thewindowsclub.com/identify-unsigned-drivers-sigverif-windows>.
```
sbrl, line 4: This seems to be a link to a 3rd party article. Is there really no official source on this?

pages.zh/windows/sigverif.md
```diff
@@ -1,6 +1,7 @@
 # sigverif
 
 > 用于检查系统文件的 GUI 签名验证工具.
+> 更多信息： <https://www.thewindowsclub.com/identify-unsigned-drivers-sigverif-windows>.
```
sbrl, line 4: Same here.

pages.pt_BR/windows/sigverif.md
marchersimon: Unfortunately I couldn't find anything

owenvoke: I think the issue with `sigverif` is that it is a GUI tool. So it has no CLI flags, which means it's not listed on the official CLI docs for Microsoft. 🤔 Is there really much point in including a link at all for this?

marchersimon: Maybe a link to its homepage or anything else would be cool. But there isn't anything online. Is there even a point in keeping the page at all, since it has only one command in it?

owenvoke: I think so, but I think I originally added it as part of #1579, where we determined it was worth including even without any flags.

marchersimon: Should we just remove it, since nobody uses it anymore?

owenvoke: The link or the page, I think we should keep the page, but just without a link. Do we really need links for every page? 🤔

bl-ue: Well, @waldyrious was indicating that we should in [this comment](https://github.com/tldr-pages/tldr/issues/5510#issue-842041969):

>Once that work is completed, we should add a check to [tldr-lint](https://github.com/tldr-pages/tldr-lint) to make this field required, thus ensuring that new pages won't regress our coverage in that regard (see #5046 for examples of link-related checks that were incorporated into tldr-lint). We should also change the template for a page in `CONTRIBUTING.md`, the style guide, and other places where that may be relevant.

If we look at the text of the label itself, "More information," it's saying that the link is for _additional_ information, but it doens't give a hint as to _what type_ of information it is, e.g. it's not "Official documentation:", so I say if it's old and not used much _and_ there's no official page, this thewindowsclub.com page does offer what I would consider "more information;" in fact, it's a good guide on how to use the GUI tool.

I will defer to @sbrl as she has much more experience than I, but those are my thoughts.

waldyrious: I agree with @bl-ue in general terms about the appropriate contents for the "More information" links — we should obviously defer to "official" content (i.e. sourced directly from the tool's authors, but it's fine to fall back to generic content such as manpages, blog posts, etc — the only deal-breaker criterion IMO should be that the information must be actually useful and accurate.

That said, in this case I'm inclined to @marchersimon's suggestion — tldr-pages is a project to document command-line tools. While we definitely can cover interactive tools, and even those that implement a [TUI](https://en.wikipedia.org/wiki/Text-based_user_interface) (though that can be hard to do), actual [GUI](https://en.wikipedia.org/wiki/Graphical_user_interface) tools are not within the scope of the project. Note that both [`rainbowstream`](https://github.com/tldr-pages/tldr/blob/master/pages/common/rainbowstream.md) and [`rtv`](https://github.com/tldr-pages/tldr/blob/master/pages/common/rtv.md), mentioned by @agnivade in [#1579](https://github.com/tldr-pages/tldr/issues/1579#issuecomment-360449367), have a TUI, rather than a GUI. As another example, we support a page for [Inkscape](https://inkscape.org/), which is primarily a GUI program but actually supports a limited set of operations via a CLI, and it's the latter that we describe in its [tldr page](https://github.com/tldr-pages/tldr/blob/master/pages/common/inkscape.md). So yeah, I'd suggest removing the `sigverif` page as out of scope.

bl-ue: It's gone now, 0619d5f96806395c85096f30aa400ce23bb0d99d.  I must admit I hate to lose a command; it seems like if users are experimenting with all of the documented commands it would be nice for their opinion of the project to find that even the strangest unused commands are deprecated, but 🤷🏻

marchersimon: But on the other hand, it makes maintaining all the pages much harder, like with this example, where we just couldn't find a link.

bl-ue: True.  In [this comment](https://github.com/tldr-pages/tldr/issues/5510#issuecomment-811381061) @waldyrious suggested `N/A` as a value for the more information label if no link was suitable; however, there's not much point to a page with one, optionless example, no link, and only a description.

sbrl: Hrm, I see. I guess that the current link is ok then if there's no official source. The issue we need to be mindful of with blog posts though is [link rot](https://en.wikipedia.org/wiki/Link_rot) though, so any references to blog posts should only be because we've explored all other options like we have here.

@waldyrious: While agree there isn't much point in documenting commands such as `sigverif` that _only_  have a GUI, the other thing we have on tldr pages is a description of what they do. So I'd argue that any commands that only have a GUI that already have a page here in tldr-pages should stay (@bl-ue: perhaps you could hold off on deleting `sigverif` please?)

bl-ue: >(@bl-ue: perhaps you could hold off on deleting sigverif please?)

😄 @marchersimon already deleted it: 0619d5f96806395c85096f30aa400ce23bb0d99d & 0af0a81d3750a49385d2355db6ca1eb96465b3ab

marchersimon: Should I add it again?

bl-ue: Let's wait until a consensus is reached before we do that, just to avoid notification noise.

sbrl: /cc @waldyrious here.

waldyrious: @sbrl I am sympathetic to your reservations about removing valid information. That said, I think that we, as a project, must have a clear vision of what we want to cover, so that we can better focus our efforts on making a good job at that, rather than attempt to cover everything (in good [Unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy) fashion, "do one thing and do it well").

Still in Unix philosophy land, the "text as a universal interface" maxim is also relevant to us: indeed, we are essentially a project to document the usage of [console/terminal/shell](https://www.hanselman.com/blog/whats-the-difference-between-a-console-a-terminal-and-a-shell) commands. So not only we focus on CLI (and to some extent TUI) programs, rather than GUI ones; but also our content is rooted in conveying usage documentation — i.e., describing the _how_, more so than the _what_ (which is still very useful, but accessory to our main purpose).

With all that said, I believe that staying true to our mission, and preserving a well-defined focus as a bedrock to guide our operational decisions, is more important than the preservation (or creation) of information on the basis of its usefulness alone. I mean, we already reject complex pipelines and hyper-specific examples as things better served by other projects (mentioning commandlinefu.com is becoming a cliché for me, but I'll keep doing it to serve as a search term for future reference 😅). Similarly, I'd say that providing descriptions and links of programs, as a goal in itself, is a role better suited to projects like [Repology](https://repology.org/), [libraries.io](https://libraries.io/), or [Wikidata](https://www.wikidata.org/wiki/Wikidata:WikiProject_Informatics/Software) (I actually contribute software info to the latter all the time — see [here](https://www.wikidata.org/wiki/Q96265025), [here](https://www.wikidata.org/wiki/Q3156928) and [here](https://www.wikidata.org/wiki/Q106357164) for examples I've edited in this month alone).

**tl;dr**: I believe, for the reasons described above, that we should delete `sigverif` and reject further instances of GUI-only programs.

marchersimon: @waldyrious would you say, that also applies to #5735?

sbrl: Ah, I see. Thanks for the detailed explanation @waldyrious!

Let's keep the removal of GUI programs without any meaningful arguments. For example, `sigverif` as I understand it does not have any meaningful arguments.

But `firefox` for example while it is a GUI program also has a CLI.

@marchersimon: I wouldn't say so, no. I'll leave a comment there, but in that case nautilus can either be called on it's own, or with an argument to a directory - the latter of which influences it's behaviour as it changes the directory it shows.

waldyrious: Yeah — If the tool has no CLI interface in addition to their GUI (like, for instance [inkscape](https://github.com/tldr-pages/tldr/blob/master/pages/common/inkscape.md) does), then I don't think it fits within our scope.

bl-ue (approved):

bl-ue:
This comment can serve as a bookmark for a very valuable (and long) conversation: https://github.com/tldr-pages/tldr/pull/5646#discussion_r604475359

pages.pt_BR/windows/sigverif.md
waldyrious: Let's keep this conversation unresolved so it won't be collapsed by default, which makes it easier to locate things ;)

waldyrious:
If nobody gets to it first, I'd like to make an interactive rebase on this branch to turn it into two commits: one adding the links, and the other removing the `sigverif` page.

bl-ue:
@waldyrious maybe you should do that now. Looks like no one has wanted to... 🤔

waldyrious:
Thanks for the ping, @bl-ue — I'll try to get to it this weekend :)

sbrl:
Rebase complete.

sbrl (approved):

bl-ue (approved):
The windows commit should really be something like `windows/*: add links` 🤔

marchersimon:
I don't think it's that important.

navarroaxel (approved):

marchersimon:
Since it's alteady rebased, is this PR ready to be merged?

PR #5647, opened 2021-03-30
units: move to common directory

marchersimon:
All those pages where almost identical and according to GNU it should be the same everywhere.


pages.pt_BR/common/units.md
```diff
@@ -1,8 +1,7 @@
 # units
 
 > Realiza a conversão entre duas unidades de medida.
-> Digitar `search{{text}}` no console retornará uma lista de todas as unidades que contêm `{{text}}`.
```
bl-ue, line 4: >Typing `search {{text}}` in the console will return a list of all units that contain `{{text}}`.

This could merit its own example 🤔

bl-ue (approved):

pages.pt_BR/common/units.md
marchersimon: But that's not really a command, since it is used in interactive mode.

bl-ue: True, but we still document that type of thing.  Take [`vim`](https://github.com/tldr-pages/tldr/blob/master/pages/common/vim.md) or [`less`](https://github.com/tldr-pages/tldr/blob/master/pages/common/less.md), for example.

marchersimon: I added it to the english page. Not sure what we should do about the translations however.

bl-ue:
Oops @marchersimon looks like you committed some unrelated stuff in f042c15b19cd09153bf6f7f4af9e68f86ac8e51a

pages.pt_BR/common/units.md
bl-ue: Leave them.  Due to the change in the example counts, https://lukwebsforge.github.io/tldri18n/ will mark the translations as out-of-date.

bl-ue (approved):

PR #5651, opened 2021-03-31
ip-link: add page

aherst:

### DIFF
--- a/pages/linux/ip-link.md
+++ b/pages/linux/ip-link.md
@@ -23,6 +23,6 @@
 
 `ip link set {{ethN}} address {{ff:ff:ff:ff:ff:ff}}`
 
-- Change the size of the MTU for a network interface to use jumbo frames:
+- Change the MTU size for a network interface to use jumbo frames:
 
 `ip link set {{ethN}} mtu {{9000}}`
--- a/pages/linux/ip-link.md
+++ b/pages/linux/ip-link.md
@@ -1,7 +1,7 @@
 # ip link
 
 > Manage network interfaces.
-> More information: <https://github.com/dmbaturin/iproute2-cheatsheet>.
+> More information: <https://man7.org/linux/man-pages/man8/ip-link.8.html>.
 
 - Show information about all network interfaces:
 
--- a/pages/linux/ip-link.md
+++ b/pages/linux/ip-link.md
@@ -17,7 +17,7 @@
 
 - Give a meaningful name to a network interface:
 
-`ip link set {{ethN}} alias {{"LAN Interface"}}`
+`ip link set {{ethN}} alias "{{LAN Interface}}"`
 
 - Change the MAC address of a network interface:
 
--- a/pages/linux/ip-link.md
+++ b/pages/linux/ip-link.md
@@ -1,0 +1,28 @@
+# ip link
+
+> Manage network interfaces.
+> More information: <https://github.com/dmbaturin/iproute2-cheatsheet>.
+
+- Show information about all network interfaces:
+
+`ip link`
+
+- Show information about a specific network interface:
+
+`ip link show {{ethN}}`
+
+- Bring a network interface up or down:
+
+`ip link set {{ethN}} {{up|down}}`
+
+- Give a meaningful name to a network interface:
+
+`ip link set {{ethN}} alias {{"LAN Interface"}}`
+
+- Change the MAC address of a network interface:
+
+`ip link set {{ethN}} address {{ff:ff:ff:ff:ff:ff}}`
+
+- Change the size of the MTU for a network interface to use jumbo frames:
+
+`ip link set {{ethN}} mtu {{9000}}`


pages/linux/ip-link.md
```diff
@@ -0,0 +1,28 @@
+# ip link
+
+> Manage network interfaces.
+> More information: <https://github.com/dmbaturin/iproute2-cheatsheet>.
```
bl-ue, line 4: Wouldn't this be better?  It's the deployed site.
```suggestion
> More information: <https://baturin.org/docs/iproute2>.
```

bl-ue:
We've needed this page for a long time.

I like how you're useing `ethN` as opposed to `eth0` here—@sbrl should we use `N` instead of a specific number in the future, like we've decided upon in #5036?

pages/linux/ip-link.md
marchersimon: I'm actually not a fan of that either. It covers a lot more than `ip link`. How about <https://man.archlinux.org/man/ip-link.8>?

bl-ue: Hmm, that man page is actually really full, I guess we should probably use it.

aherst: What if we used https://baturin.org/docs/iproute2/#ip-link? I think it's a good intermediary step between the tldr page and the man page.

I guess I still don't understand when a link to a non-man page is appropriate.

bl-ue: Well @ahert you're not the only one learning 😄 

In this case the man page linked above is quite full of info and it even has an [examples section](https://man.archlinux.org/man/ip-link.8#EXAMPLES) at the bottom, so I think it definitely qualifies as a good "more information" page.

aherst: OK. I'll change it to point to a man page.

I'm about to submit a pull request for ip-route. I don't think the ip-route man page has a particularly good Examples section. Would linking to baturin.org be appropriate in that case?

bl-ue: Well, that man page does have an examples section, and there's also lots of explanation about the various options and values, so I'd say use the man page for that one as well.

Not all of the manpages are that good, but these ones seem to be prett informative.

marchersimon (approved):

bl-ue (approved):

pages/linux/ip-link.md
```diff
@@ -0,0 +1,28 @@
...
+
+`ip link show {{ethN}}`
+
+- Bring a network interface up or down:
+
+`ip link set {{ethN}} {{up|down}}`
+
+- Give a meaningful name to a network interface:
+
+`ip link set {{ethN}} alias {{"LAN Interface"}}`
```
navarroaxel, line 20: The quotes should be outside:
```suggestion
`ip link set {{ethN}} alias "{{LAN Interface}}"`
```

navarroaxel (changes requested):

pages/linux/ip-link.md
bl-ue: Do we have that in the guidelines?  Ought to.

navarroaxel: Maybe not.

bl-ue: Nope, I'll do that today.

aherst: The quotation marks aren't a mandatory part of the syntax. They're only required in this case because the example name contained a space. An equivalent syntax would be to escape the space with a backslash, which would appear within the braces.

bl-ue: Oh.  Can you give an example similar to `ethN` then?

aherst: Do you mean something like {{meaningful_name}}? I was trying to convey in the example that it's OK to use spaces in the name as long as the name is quoted.

bl-ue: Ah I see, I wasn't being clear.  From the man page:

>give the device a symbolic name for easy reference

So it doesn't matter what name you give it. I think `LAN Interface` is fine then and the quotes are fine as well.  Just put them outside of the token syntax.  The unspoken rule is that if an argument _can_ have spaces in it, put quotes around it, even if the specific example of the argument that you're writing doesn't have spaces in.

Makes sense?

aherst: OK. 

Thanks again for your patience walking me through the process as I'm learning.

bl-ue: @aherst no problem whatsoever! 💯

navarroaxel (approved):

pages/linux/ip-link.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Give a meaningful name to a network interface:
+
+`ip link set {{ethN}} alias "{{LAN Interface}}"`
+
+- Change the MAC address of a network interface:
+
+`ip link set {{ethN}} address {{ff:ff:ff:ff:ff:ff}}`
+
+- Change the size of the MTU for a network interface to use jumbo frames:
```
sbrl, line 26: Perhaps this will read better?
```suggestion
- Change the MTU size for a network interface to use jumbo frames:
```

sbrl (commented):
Just 1 minor comment below.

@bl-ue: In this instance I don't think that using `ethN` is really worth it, because other pages use `eth0` consistently (e.g. `ip address`), and we've using a specific example here. Furthermore, on modern systems it's actually unlikely that `eth0` is the name oft he network interface anyway (unlike the `/dev/sda1` problem), because on systems with multiple network itnerfaces the order in which the Linux kernel initialises them is undefined. Hence a new naming scheme was devised, which looks something like this: `enp7s0`

So we use `eth0` as an example because everyone will be familiar with it, but it isn't used in practice on modern systems (except the raspberry pi). And the potential for damage here is much lower - it's not the same thing as the `/dev/sda1` problem.

I would suggest that on this page we use `eth0`, not `ethN`.  

Anyway, thanks, @aherst! We've needed someone who's an expert in the `ip` command to come along and write a page like this for ages (I don't know `ip link` well enough to write it myself - though I did write `ip address` IIRC).

aherst:
I thought it would be appropriate to use ethN since it is possible to lose you network connection when ssh'ing in to make adjustments to network devices and firewalls.

bl-ue:
That would be bad 😬

marchersimon:
> I would suggest that on this page we use `eth0`, not `ethN`.

> I thought it would be appropriate to use ethN since it is possible to lose you network connection when ssh'ing in to make adjustments to network devices and firewalls.

@sbrl should we keep `ethN` in this case?

sbrl:
Ok, you've convinced me. I guess keeping `ethN` makes some sense here - it's keep it.

sbrl (approved):

PR #5654, opened 2021-03-31
scripts: add set-more-info-link.py

bl-ue:
Also located here https://gist.github.com/bl-ue/5b52c9f80dd21074c0633a2bcdeb01fb (not being updated ⚠️)

### DIFF
--- a/scripts/set-more-info-link.py
+++ b/scripts/set-more-info-link.py
@@ -34,7 +34,7 @@ labels = {
 }
 
 IGNORE_FILES = (
-    '.DS_Store'
+    '.DS_Store',
 )
 
 
--- a/scripts/set-more-info-link.py
+++ b/scripts/set-more-info-link.py
@@ -104,6 +104,9 @@ def main():
     target_paths = []
     rel_paths = []
 
+    if args.page.lower().endswith('.md'):
+        args.page = args.page[:-3]
+
     for pages_dir in pages_dirs:
         pages_dir_path = os.path.join(root, pages_dir)
         platforms = [i for i in os.listdir(
--- a/scripts/set-more-info-link.py
+++ b/scripts/set-more-info-link.py
@@ -1,0 +1,132 @@
+import argparse
+import os
+import re
+import subprocess
+import sys
+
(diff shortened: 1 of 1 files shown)


marchersimon (approved):

patricedenis:
hey, very good script to achieve #5510 
Congrats !

bl-ue:
Hi @patricedenis, that's very gratifying to hear.  Thank you very much ❤️

Waples:
`IGNORE_FILES` is a bit platform specific, also don't really get why you hardset the `TLDR_ROOT` in an env var. You can make it more flexible if running directly from the `scripts` dir or the root of the repo.

bl-ue:
@Waples thank you for your review, I really appreciate it.

>also don't really get why you hardset the `TLDR_ROOT` in an env var. You can make it more flexible if running directly from the scripts dir or the root of the repo and then let the script set that variable, if needed at all.

That's already implemented—before I opened this PR, however, it only checked `TLDR_ROOT`.  Now it uses the current directory, see [here](https://github.com/tldr-pages/tldr/pull/5654/files#diff-2042b6a043587d4f66025dd7bed56fc5aa710355fa47346e00b3f6d5bbf315f1R42-R45).

>`IGNORE_FILES` is a bit platform specific

You are definitely right, however, `.DS_Store` on my Mac immediately started causing me problems.  There are other Mac users here, like @owenvoke too, who may run this scripts.  I'd gladly add more if you know of any that typically can cause problems like that.

But, you made me notice that as the tuple `IGNORE_FILES` has only one item _and_ that item is **not** followed by a trailing comma, `IGNORE_FILES` is a string not a tuple, so it's not technically what I want (although `'.DS_Store' in '.DS_Store'` will work).  I'll fix that.

yutyo (approved):
LGTM.

Waples:
Merged already, but maybe in the near future I'll have some comments on this.

bl-ue:
Thank you @Waples, I'd really love to hear everything you think about it!!

Waples:
Will do

PR #5656, opened 2021-03-31
base64: clarify example descriptions that do not edit files inplace

makusu2:
Reading "Encode a file" made me worry that the command would edit the file inplace. Clarify that it prints the results instead.

### DIFF
--- a/pages/common/base64.md
+++ b/pages/common/base64.md
@@ -3,7 +3,7 @@
 > Encode or decode file or standard input to/from Base64, to standard output.
 > More information: <https://www.gnu.org/software/coreutils/manual/html_node/base64-invocation.html>.
 
-- Encode the contents of a file as base64 to stdout:
+- Encode the contents of a file as base64 and write the result to stdout:
 
 `base64 {{filename}}`
 
--- a/pages/common/base64.md
+++ b/pages/common/base64.md
@@ -3,11 +3,11 @@
 > Encode or decode file or standard input to/from Base64, to standard output.
 > More information: <https://www.gnu.org/software/coreutils/manual/html_node/base64-invocation.html>.
 
-- Encode a file:
+- Print a plaintext file's contents encoded in base64:
 
 `base64 {{filename}}`
 
-- Decode a file:
+- Print a base64 encoded file's contents decoded to plaintext:
 
 `base64 --decode {{filename}}`
 
--- a/pages/common/base64.md
+++ b/pages/common/base64.md
@@ -7,7 +7,7 @@
 
 `base64 {{filename}}`
 
-- Decode the base64 contents of a file to stdout:
+- Decode the base64 contents of a file and write the result to stdout:
 
 `base64 --decode {{filename}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/base64.md
```diff
@@ -3,11 +3,11 @@
 > Encode or decode file or standard input to/from Base64, to standard output.
 > More information: <https://www.gnu.org/software/coreutils/manual/html_node/base64-invocation.html>.
 
-- Encode a file:
+- Print a plaintext file's contents encoded in base64:
```
bl-ue, line 6:
```suggestion
- Encode the contents of a file as base64 to stdout:
```

pages/common/base64.md
```diff
@@ -3,11 +3,11 @@
...
 
 `base64 {{filename}}`
 
-- Decode a file:
+- Print a base64 encoded file's contents decoded to plaintext:
```
bl-ue, line 10:
```suggestion
- Decode the base64 contents of a file to stdout:
```

bl-ue (commented):
Great catch @makusu2! I completely agree with you, but I think the wording can be made more concise.

pages/common/base64.md
makusu2: Words are hard, thanks for the suggestion!

bl-ue (approved):

pages/common/base64.md
```diff
@@ -3,11 +3,11 @@
 > Encode or decode file or standard input to/from Base64, to standard output.
 > More information: <https://www.gnu.org/software/coreutils/manual/html_node/base64-invocation.html>.
 
-- Encode a file:
+- Encode the contents of a file as base64 to stdout:
```
sbrl, line 6: Perhaps this would be more grammatically correct:
```suggestion
- Encode the contents of a file as base64 and write the result to stdout:
```

pages/common/base64.md
```diff
@@ -3,11 +3,11 @@
...
 
 `base64 {{filename}}`
 
-- Decode a file:
+- Decode the base64 contents of a file to stdout:
```
sbrl, line 10: Perhaps this would be more grammatically correct?
```suggestion
- Decode the base64 contents of a file and write the result to stdout:
```

sbrl (commented):
Hey there, @makusu2! Thanks for the edit. I've left a pair of comments below for you to review.

pages/common/base64.md
makusu2: Hmm, I get what you're saying, but that introduces the original issue again. I could read this as:

1. Encode the contents of a file in-place
2. Write the contents of the new file to stdout

I like it, but I think @bl-ue 's version is the most clear.

sbrl: I'm not sure I quite understand how you're interpreting that, @makusu2. I would argue here that as it stands it is actually _more_ unclear now than if we add `and write the result to`, because we aren't indicating _where_ precisely the result is written to.

By specifically stating that we are writing the result to stdout, we avoid potential confusion that the file is being edited in place.

makusu2: Yeah, that's true. I was splitting the sentence on "and", but I agree now that it's more likely people will read it the way you're describing.

Maybe we could do 
```
Encode the contents of a file as base64, writing the result to stdout:
```

Hmm... I have no strong preference. 
But anyway, you're right, I'll apply the suggestion. Thanks!

bl-ue (approved):

marchersimon (approved):

sbrl (approved):
Thanks, @makusu2!

PR #5665, opened 2021-04-02
archinstall: add page

navarroaxel:

I read the source code and I tested it in a VM but I couldn´t find any command line argument, only the guided/interactive mode.
I add this page because is a built-in command in the Arch Linux official ISO image <https://archlinux.org/news/installation-medium-with-installer/>.

### DIFF
--- a/pages/linux/archinstall.md
+++ b/pages/linux/archinstall.md
@@ -6,3 +6,7 @@
 - Start the interactive installer:
 
 `archinstall`
+
+- Start a preset installer:
+
+`archinstall {{minimal|unattended}}`
--- a/pages/linux/archinstall.md
+++ b/pages/linux/archinstall.md
@@ -1,0 +1,8 @@
+# archinstall
+
+> Guided Arch Linux installer with a twist.
+> More information: <https://archinstall.readthedocs.io>.
+
+- Start the interactive installer:
+
+`archinstall`


bl-ue:
Hmm, strange.  What about this page: https://archinstall.readthedocs.io/en/latest/examples/binary.html, where they pass `guided` to it?  I see this, unfortunately:
>As mentioned, the binary is a bit rudimentary and only supports executing whatever is found directly under ./archinstall/examples. Anything else won’t be found.

navarroaxel:
@bl-ue in the built-in binary you can type any argument, the guided installation is always executed. To run an example you should use `python -m archinstall minimal`, do you want I add that?

The examples are just `.py` scripts, you can create a custom script but your custom one should be the entry point to use this as an API.

marchersimon (approved):

bl-ue:
If every user is guaranteed to be able to run that, than yes (without `python -m` though), otherwise no.

I'm a bit concerned however that this is too similar too the [`sigverif` condition](https://github.com/tldr-pages/tldr/pull/5646#discussion_r605128934) over at #5646 however...quoting from @waldyrious (bold mine):
>That said, in this case I'm inclined to @marchersimon's suggestion — tldr-pages is a project to document command-line tools. While we definitely can cover interactive tools, and even those that implement a TUI (though that can be hard to do), **actual GUI tools are not within the scope of the project**. Note that both rainbowstream and rtv, mentioned by @agnivade in #1579, have a TUI, rather than a GUI. As another example, we support a page for Inkscape, which is primarily a GUI program but actually supports a limited set of operations via a CLI, and it's the latter that we describe in its tldr page. So yeah, I'd suggest removing the sigverif page as out of scope.

navarroaxel:
ok, re-reading the doc page these example scripts can be modified if your compile the `archinstall` locally, so you can use `archinstall minimal` if you compiled that but is not available in the official Arch Linux ISO.
Should I add `python -m archinstall minimal` or `archinstall minimal` ?

bl-ue:
The latter.  We're not really demonstrating the `python` command usage.

bl-ue (approved):

PR #5664, opened 2021-04-01
rename pages.hbs to pages.sh

bl-ue:
See @vladimyr's message:
https://gitter.im/tldr-pages/tldr?at=60625b22cfd0b814ebac164d

PR #5658, opened 2021-04-01
coreutils*: use short more info links

vladimyr:
- [ ] The page (if new), does not already exist in the repo.
- [ ] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [ ] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [ ] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [ ] The page description includes a link to documentation or a homepage (if applicable).

GNU online documentation provides user-friendly URLs - for example, https://www.gnu.org/software/coreutils/ls which redirects to https://www.gnu.org/software/coreutils/manual/html_node/ls-invocation.html#ls-invocation. Those short forms are advertised through `coreutils` manpages (for example, [`ls`](https://manned.org/ls.1#head7)) so I've updated all accompanying `coreutils` tldr pages to follow the suit.

This is done using following script:
```sh
rg -l 'gnu.org/software/coreutils/manual/html_node/(.*?)-invocation.html' |
    xargs sed -i -E 's|(gnu.org/software/coreutils)/manual/html_node/(.*?)-invocation.html|\1/\2|g'
```

PR #5669, opened 2021-04-02
sha*sum: add more information link

marchersimon:
For #5510

### DIFF
--- a/pages.hbs/common/sha224sum.md
+++ b/pages.hbs/common/sha224sum.md
@@ -1,6 +1,7 @@
 # sha224sum
 
 > Izračunava SHA224 kriptografske kontrolne brojeve.
+> Više informacija: <https://www.gnu.org/software/coreutils/manual/html_node/sha2-utilities.html>.
 
 - Izračunaj SHA224 kontrolni broj za datoteku:
 
--- a/pages.hbs/common/sha256sum.md
+++ b/pages.hbs/common/sha256sum.md
@@ -1,6 +1,7 @@
 # sha256sum
 
 > Izračunava SHA256 kriptografske kontrolne brojeve.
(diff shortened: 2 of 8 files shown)


bl-ue (dismissed):

bl-ue:
@marchersimon please see @vladimyr's PR #5658 where he shorted the coreutils links.

marchersimon:
I tried that, but that doesn't work here, since the url is sligthly different from the other pages.

bl-ue:
Yeah, I just discovered that myself.  Well, never mind 😛

bl-ue (approved):

bl-ue:
Though the linked page is tiny (this is all of the text content, Ctrl+A - Ctrl+C):
>Previous: sha1sum invocation, Up: Summarizing files   [Contents][Index]
>
>6.7 sha2 utilities: Print or check SHA-2 digests
>
>The commands sha224sum, sha256sum, sha384sum and sha512sum compute checksums of various lengths (respectively 224, 256, 384 and 512 bits), collectively known as the SHA-2 hashes. The usage and options of these commands are precisely the same as for md5sum and sha1sum. See [md5sum invocation](https://www.gnu.org/software/coreutils/manual/html_node/md5sum-invocation.html#md5sum-invocation).

...there's that link at the end which is pretty obvious IMO.  But if the others think, maybe we should change the links to all point to that md5sum link. 🤔

marchersimon:
The problem I have with directly linking to the `md5sum` page, is that nowhere in the page is mentioned that those arguments also work for the `sha*sum` pages, which will probably confuse many and make it seem the link is wrong.

bl-ue:
@navarroaxel, @glenacota, @yutyo Someone, can we get this merged?  I'd like to incorporate it into #5664 before that gets merged.

navarroaxel (approved):

Issue #4044, opened 2020-05-14
Inactive and unmaintained clients


ivanhercaz:
In Gitter arise an issue very interesting about the maintenance of the tldr-pages clients lists, "lists" rather than "list" because we have two lists, one in the README and another one in a wiki page. The user @principis address this issue asking about if someone should clean the clients lists, and after a conversation with @sbrl and me, we agree it would be interesting to address it.

The [original message](https://gitter.im/tldr-pages/tldr?at=5ebc6917496be6031dfe86b3) of @principis is the next:
> Should someone look into cleaning up the clients list? For example the TLDR Elixir Client is 6 years old, archived and still says "(binaries not yet available)". There are probably more clients which aren't being updated and/or following the client spec.

Recently I announce that I began to work in an Elixir client. The main reason I began this project is because the TLDR Elixir Client was inactive, archived and its development does not seem to continue (check [TLDR Elixir Client repository](https://github.com/edgurgel/tldr_elixir_client)).

There are probably another clients in this situation: inactives officially, like the TLDR Elixir Client, and apparently inactive or unmaintained. This second category seem a bit more subjective, but it probably needs a review.

Some questions to think about this issue may be:

- What should be the criteria to keep a client in the list?
- Should we accept any client in the list? Or just the ones which accomplish with client specification?
- Should we consider the time without activity in the repository? By development? By important issues without activity?

ivanhercaz:
I am going to answer in a comment the above questions:

**What should be the criteria to keep a client in the list?**
I think this is the key of this issue. If we decide to clean the lists we should establish a criteria in which we would use to substantiate the decision. Now I have not a clear answer, thus I think this issue is useful.

**Should we accept any client in the list? Or just the ones which accomplish with client specification?**
Accomplish with client specification might be a good point to start. This even affect me, because ExTldr does not accomplish with the client specification and in this case it would not be in the client lists. But being honest, I think it is an objective criteria based mainly in the fact that if a client accomplish with the latest client specification or have backward compatibility, it should be in the lists.

**Should we consider the time without activity in the repository? By development? By important issues without activity?**
Sometimes the time without commits or releases is not a fact to drop something, because the reason may be that the client is already fine and it does not need any other change in its source code. In the other side, sometimes it may indicate lack of maintenance or inactivity of the project even if the repository is not archived. This is more obvious when the client repository has important issues without responses by the maintainers or possible solutions, or even pull requests stacked without being reviewed.

[...]

PR #5675, opened 2021-04-03
fakedata: add page

nicokosi:

### DIFF
--- a/pages/common/fakedata.md
+++ b/pages/common/fakedata.md
@@ -21,4 +21,4 @@
 
 - Generate data using a custom output template (the first letter of generator names must be capitalized):
 
-`echo "\{\Generator\}\}" | fakedata`
+`echo "{{\{\{Generator\}\}}}" | fakedata`
--- a/pages/common/fakedata.md
+++ b/pages/common/fakedata.md
@@ -15,7 +15,7 @@
 
 `fakedata --format {{csv|tab|sql}} {{generator}}`
 
-- Generate `n` data (defaults to 10):
+- Generate a given number of fake data items (defaults to 10):
 
 `fakedata --limit {{n}} {{generator}}`
 
--- a/pages/common/fakedata.md
+++ b/pages/common/fakedata.md
@@ -15,7 +15,7 @@
 
 `fakedata --format {{csv|tab|sql}} {{generator}}`
 
-- Generate a given number of fake data items (defaults to 10):
+- Generate a given number of data items (defaults to 10):
 
 `fakedata --limit {{n}} {{generator}}`
 
--- a/pages/common/fakedata.md
+++ b/pages/common/fakedata.md
@@ -3,15 +3,15 @@
 > CLI utility for fake data generation.
 > More information: <https://github.com/lucapette/fakedata>.
 
-- List generators (all kinds of data that can be generated):
+- List all valid generators:
 
 `fakedata --generators`
 
-- Generate data for one or several generators:
+- Generate data using one or more generators:
 
 `fakedata {{generator1}} {{generator2}}`
 
-- Generate data with a specific format:
+- Generate data with a specific output format:
 
 `fakedata --format {{csv|tab|sql}} {{generator}}`
 
--- a/pages/common/fakedata.md
+++ b/pages/common/fakedata.md
@@ -21,4 +21,4 @@
 
 - Generate data using a custom output template (the first letter of generator names must be capitalized):
 
-`echo "value: {{Generator}}" | fakedata`
+`echo "\{\Generator\}\}" | fakedata`
--- a/pages/common/fakedata.md
+++ b/pages/common/fakedata.md
@@ -15,7 +15,7 @@
 
 `fakedata --format {{csv|tab|sql}} {{generator}}`
 
-- Generate `n` data:
+- Generate `n` data (defaults to 10):
 
 `fakedata --limit {{n}} {{generator}}`
 
--- a/pages/common/fakedata.md
+++ b/pages/common/fakedata.md
@@ -19,6 +19,6 @@
 
 `fakedata --limit {{n}} {{generator}}`
 
-- Generate data using a custom output template (capitalize the first letter of generators):
+- Generate data using a custom output template (the first letter of generator names must be capitalized):
 
 `echo "value: {{Generator}}" | fakedata`
--- a/pages/common/fakedata.md
+++ b/pages/common/fakedata.md
@@ -3,7 +3,7 @@
 > CLI utility for fake data generation.
 > More information: <https://github.com/lucapette/fakedata>.
 
-- List generators (all kinds of data that can ge generated):
+- List generators (all kinds of data that can be generated):
 
 `fakedata --generators`
 
@@ -11,9 +11,9 @@
 
 `fakedata {{generator1}} {{generator2}}`
 
-- Generate data with a specific format (`csv`|`tab`|`sql`):
+- Generate data with a specific format:
 
-`fakedata --format {{format}} {{generator}}`
+`fakedata --format {{csv|tab|sql}} {{generator}}`
 
 - Generate {{n}} data:
 
@@ -21,4 +21,4 @@
 
 - Generate data using a custom template (capitalize the first letter of generators):
 
-`echo "value: {{Generator}}>" | fakedata`
+`echo "value: {{Generator}}" | fakedata`
--- a/pages/common/fakedata.md
+++ b/pages/common/fakedata.md
@@ -1,6 +1,6 @@
 # fakedata
 
-> CLI utility for fake data generation.
+> Generate fake data using a large variety of generators.
 > More information: <https://github.com/lucapette/fakedata>.
 
 - List all valid generators:
--- a/pages/common/fakedata.md
+++ b/pages/common/fakedata.md
@@ -19,6 +19,6 @@
 
 `fakedata --limit {{n}} {{generator}}`
 
-- Generate data using a custom template (capitalize the first letter of generators):
+- Generate data using a custom output template (capitalize the first letter of generators):
 
 `echo "value: {{Generator}}" | fakedata`
--- a/pages/common/fakedata.md
+++ b/pages/common/fakedata.md
@@ -15,7 +15,7 @@
 
 `fakedata --format {{csv|tab|sql}} {{generator}}`
 
-- Generate {{n}} data:
+- Generate `n` data:
 
 `fakedata --limit {{n}} {{generator}}`
 
--- a/pages/common/fakedata.md
+++ b/pages/common/fakedata.md
@@ -1,0 +1,24 @@
+# fakedata
+
+> CLI utility for fake data generation.
+> More information: <https://github.com/lucapette/fakedata>.
+
+- List generators (all kinds of data that can ge generated):
+
+`fakedata --generators`
+
+- Generate data for one or several generators:
+
+`fakedata {{generator1}} {{generator2}}`
+
+- Generate data with a specific format (`csv`|`tab`|`sql`):
+
+`fakedata --format {{format}} {{generator}}`
+
+- Generate {{n}} data:
+
+`fakedata --limit {{n}} {{generator}}`
+
+- Generate data using a custom template (capitalize the first letter of generators):
+
+`echo "value: {{Generator}}>" | fakedata`


pages/common/fakedata.md
```diff
@@ -0,0 +1,24 @@
+# fakedata
+
+> CLI utility for fake data generation.
+> More information: <https://github.com/lucapette/fakedata>.
+
+- List generators (all kinds of data that can ge generated):
```
bl-ue, line 6:
```suggestion
- List generators (all kinds of data that can be generated):
```

pages/common/fakedata.md
```diff
@@ -0,0 +1,24 @@
...
+
+`fakedata --generators`
+
+- Generate data for one or several generators:
+
+`fakedata {{generator1}} {{generator2}}`
+
+- Generate data with a specific format (`csv`|`tab`|`sql`):
+
+`fakedata --format {{format}} {{generator}}`
```
bl-ue, line 16:
```suggestion
- Generate data with a specific format:

`fakedata --format {{csv|tab|sql}} {{generator}}`
```

pages/common/fakedata.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Generate {{n}} data:
+
+`fakedata --limit {{n}} {{generator}}`
+
+- Generate data using a custom template (capitalize the first letter of generators):
+
+`echo "value: {{Generator}}>" | fakedata`
```
bl-ue, line 24: Mistake?
```suggestion
`echo "value: {{Generator}}" | fakedata`
```

pages/common/fakedata.md
nicokosi: Should I amend the commit (push --force) ? Or commit the suggestion (new commit)?

bl-ue: The latter.  No need to keep one commit because we auto squash before merging.

nicokosi:
> The [build](https://github.com/tldr-pages/tldr/actions/runs/714431845) for this PR failed with the following error(s):
> 
> ```
> pages/common/fakedata.md:24 MD047/single-trailing-newline Files should end with a single newline character
> pages/common/fakedata.md:1: TLDR009 Page should contain a newline at end of file
> ```
> 

Error seems outdated. 😇

pages/common/fakedata.md
```diff
@@ -0,0 +1,24 @@
+# fakedata
+
+> CLI utility for fake data generation.
+> More information: <https://github.com/lucapette/fakedata>.
+
+- List generators (all kinds of data that can be generated):
```
bl-ue, line 6:
```suggestion
- List all valid generators:
```

pages/common/fakedata.md
```diff
@@ -0,0 +1,24 @@
...
+
+`fakedata --generators`
+
+- Generate data for one or several generators:
```
bl-ue, line 10:
```suggestion
- Generate data using one or more generators:
```

pages/common/fakedata.md
```diff
@@ -0,0 +1,24 @@
...
+
+`fakedata {{generator1}} {{generator2}}`
+
+- Generate data with a specific format:
```
bl-ue, line 14:
```suggestion
- Generate data with a specific output format:
```

pages/common/fakedata.md
```diff
@@ -0,0 +1,24 @@
...
+
+`fakedata --format {{csv|tab|sql}} {{generator}}`
+
+- Generate {{n}} data:
+
+`fakedata --limit {{n}} {{generator}}`
+
+- Generate data using a custom template (capitalize the first letter of generators):
+
+`echo "value: {{Generator}}" | fakedata`
```
bl-ue, line 24: Odd feature.  But well worded IMHO 👍🏻

pages/common/fakedata.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Generate data with a specific output format:
+
+`fakedata --format {{csv|tab|sql}} {{generator}}`
+
+- Generate {{n}} data:
+
+`fakedata --limit {{n}} {{generator}}`
+
+- Generate data using a custom template (capitalize the first letter of generators):
```
nicokosi, line 22:
```suggestion
- Generate data using a custom output template (capitalize the first letter of generators):
```

nicokosi:
Thank you @bl-ue for the typos and the reword suggestions. 😎

pages/common/fakedata.md
```diff
@@ -0,0 +1,24 @@
+# fakedata
+
+> CLI utility for fake data generation.
```
bl-ue, line 3: That it's a CLI utility is made obvious by the fact the tldr-page (tries to) only document CLI tools.  How about this instead?
```suggestion
> Generate fake data using a large variety of generators.
```

bl-ue (approved):

pages/common/fakedata.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Generate data using one or more generators:
+
+`fakedata {{generator1}} {{generator2}}`
+
+- Generate data with a specific output format:
+
+`fakedata --format {{csv|tab|sql}} {{generator}}`
+
+- Generate {{n}} data:
```
marchersimon, line 18: n data sounds a bit weird. What is meant by one data?

bl-ue: Well you know @marchersimon 'data' is plural for 'datum', hence 'n data' = 'n datums'.  I thought about this and decided to keep it as it's perfectly valid and integral English.

marchersimon: Oh, my fault. I didn't know about `data` in that context. :laugh:

marchersimon (approved):
LGTM, however it really confused me that data is the plural of datum and I fear it may confuse others too.

bl-ue:
Let's see what others have to say. @sbrl

pages/common/fakedata.md
```diff
@@ -0,0 +1,24 @@
...
+
+`fakedata --limit {{n}} {{generator}}`
+
+- Generate data using a custom output template (capitalize the first letter of generators):
+
+`echo "value: {{Generator}}" | fakedata`
```
navarroaxel, line 24:
```suggestion
`echo "value: {{generator}}" | fakedata`
```

pages/common/fakedata.md
nicokosi: What about?
```suggestion
- Generate data and limits rows up to {{n}} (default 10):
```
(I copied/pasted the usage 😇 / I'm not a native English speaker 😇)

navarroaxel: I think this should be with back tick instead of `{{`.

pages/common/fakedata.md
nicokosi: I used "Camel case" on purpose because:
> If the generator name is a single word, then it's available as a function with the same name capitalized (example: int becomes Int). If the generator name is composed by multiple words joined by dots, then the function name is again capitalized by the first letter of the word and joined together (example: product.name becomes Product.Name).

See https://github.com/lucapette/fakedata#generators-1

What do you think of this "intentional violation"? 😇

pages/common/fakedata.md
bl-ue: >I think this should be with back tick instead of `{{`.

I thought about that, but I'm really not sure.  In many clients, wrapping it in `{{` and `}}` will highlight it, which we probably want.  @sbrl have you any opinions?

>What about? ...

@nicokosi How about this simpler version:
```suggestion
- Generate {{n}} rows of data (defaults to 10):
```

pages/common/fakedata.md
navarroaxel: :thinking: all the custom format is the custom argument option in this example, so the valid highlight is the following:
```suggestion
`echo "{{value: Generator}}" | fakedata`
```

bl-ue: I agree with it.  It's required, so definitely keep it.  @navarroaxel see the example description itself; it mentions this oddity.

bl-ue: @nicokosi maybe it would be better to change the parenthetic message to `the first letter of generator names must be capitalized`, that sounds a bit clearer.

pages/common/fakedata.md
sbrl: Erm I don't think we've ever used the `{{token_syntax}}` in descriptions before. What about this?
```suggestion
- Generate a given number of fake data items (defaults to 10):
```

sbrl:
Left a comment here @bl-ue: <https://github.com/tldr-pages/tldr/pull/5675#discussion_r606724616>

pages/common/fakedata.md
bl-ue: Actually, we have in these pages:
- [`pages/linux/uvcdynctrl.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages/linux/uvcdynctrl.md)
- [`pages/linux/cryptsetup.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages/linux/cryptsetup.md)
- [`pages/linux/i3.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages/linux/i3.md)
- [`pages/linux/trash.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages/linux/trash.md)
- [`pages/common/nano.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages/common/nano.md)
- [`pages/common/swig.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages/common/swig.md)
- [`pages/common/bedtools.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages/common/bedtools.md)
- [`pages/common/samtools.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages/common/samtools.md)
- [`pages/common/dep.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages/common/dep.md)
- [`pages/common/mosh.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages/common/mosh.md)
- [`pages/common/ghdl.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages/common/ghdl.md)
- [`pages/osx/whence.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages/osx/whence.md)
- [`pages.it/common/bedtools.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages.it/common/bedtools.md)
- [`pages.fr/common/ghdl.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages.fr/common/ghdl.md)
- [`pages.fr/common/nano.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages.fr/common/nano.md)
- [`pages.zh/linux/i3.md`](https://github.com/tldr-pages/tldr/blob/289787c7e8c1177742d23004198253154fe50c3c/pages.zh/linux/i3.md)

pages/common/fakedata.md
nicokosi: Done in commit 7373486e.

nicokosi: > @nicokosi maybe it would be better to change the parenthetic message to `the first letter of generator names must be capitalized`, that sounds a bit clearer.

Done in commit 7373486e.

pages/common/fakedata.md
nicokosi: > > I think this should be with back tick instead of `{{`.
> 
> I thought about that, but I'm really not sure. In many clients, wrapping it in `{{` and `}}` will highlight it, which we probably want. @sbrl have you any opinions?
> 
> > What about? ...
> 
> @nicokosi How about this simpler version:

Done in c687b5f9.

nicokosi:
Seems OK for me (see [current version](https://github.com/tldr-pages/tldr/pull/5675/files?short_path=add8ab0#diff-add8ab00ac320da527fe77c8075d61f98b69af0af075bd55001030f764696a87)).

nicokosi:
> Left a comment here @bl-ue: [#5675 (comment)](https://github.com/tldr-pages/tldr/pull/5675#discussion_r606724616)

Done in 1c0bcc0c.

pages/common/fakedata.md
bl-ue: @nicokosi I believe @navarroaxel is right about the extents of the token syntax; "value: " is your own custom text and there fore should be highlighted as needed. https://github.com/tldr-pages/tldr/pull/5675/files#r606702953

bl-ue:
Thank you @nicokosi.  I left a comment here: https://github.com/tldr-pages/tldr/pull/5675#discussion_r606787688

pages/common/fakedata.md
bl-ue: Actually, I see there's an issue here.  `fakedata` has a special syntax for templates where you enclose the capitalized template name in `{{` and `}}`.  Both the capitalization and the braces are required because `fakedata` uses the [Go `template` library](https://golang.org/pkg/text/template/).  The issue with that is that we at tldr-pages use that very syntax for a different purpose: that of highlighting user-entered values.

Instead of the original text here and @navarroaxel's suggestion, the appropriate would be this:
```
echo "{{value: {{Generator}}}}" | fakedata`
```
However, many clients do a blind replace of `{{` and `}}` with HTML/ANSI escape codes that highlight the text, so we might need some sort of escaping, e.g.
```
echo "{{value: \{\{Generator\}\}}}" | fakedata`
```
But I'm just not sure.  @sbrl what should we do here?

nicokosi: > Actually, I see that there's an issue here. `fakedata` has a special syntax for their templates where you enclose the capitalized template name in `{{` and `}}`. Both the capitalization and the braces are required because `fakedata` uses the [Go `template` library](https://golang.org/pkg/text/template/). The issue with that is that we at tldr-pages use that very syntax for a different purpose: that of highlighting user-entered values.
> 
> Instead of the original text here and @navarroaxel's suggestion, the appropriate would be this:
> 
> ```
> echo "{{value: {{Generator}}}}" | fakedata`
> ```
> 
> However, many clients do a blind replace of `{{` and `}}` with HTML/ANSI escape codes that highlight the text, so we might need some sort of escaping, e.g.
> 
> ```
> echo "{{value: \{\{Generator\}\}}}" | fakedata`
> ```
> 

Done in 8ce90227.

pages/common/fakedata.md
```diff
@@ -0,0 +1,24 @@
...
+
+`fakedata --format {{csv|tab|sql}} {{generator}}`
+
+- Generate a given number of fake data items (defaults to 10):
+
+`fakedata --limit {{n}} {{generator}}`
+
+- Generate data using a custom output template (the first letter of generator names must be capitalized):
+
+`echo "\{\Generator\}\}" | fakedata`
```
bl-ue, line 24: Technically, it should be like this. 
```suggestion
`echo "{{\{\{Generator\}\}}}" | fakedata`
```

bl-ue (approved):
Looks good to me.  I want to know what @sbrl has to say regarding the syntax in the last example.  Ping @sbrl

nicokosi:
Seems OK for me. @sbrl, still any blockers on your side?

pages/common/fakedata.md
```diff
@@ -0,0 +1,24 @@
...
+
+`fakedata --format {{csv|tab|sql}} {{generator}}`
+
+- Generate a given number of data items (defaults to 10):
+
+`fakedata --limit {{n}} {{generator}}`
+
+- Generate data using a custom output template (the first letter of generator names must be capitalized):
+
+`echo "{{\{\{Generator\}\}}}" | fakedata`
```
sbrl, line 24: Oh wow, this is a very awkward situation here. We don't actually define any standard way to escape this in the [client spec](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#page-structure). This is the only thing we mention:

> Although this specification is about the interface that clients must provide, it is also worth noting that pages are written in standard CommonMark, which the exception of the non-standard `{{` and `}}` syntax, which surrounds values in an example that users may edit. Clients MUST NOT break if the page format is changed within the CommonMark specification.

So the escaping with the backward slashes won't be handled correctly here. What happens if you remove the backwards slashes? How do clients handle that? I would assume that since the syntax can't be nested they would display it?

I think a client spec update is in order to properly define our non-standard token syntax.

bl-ue: Also @sbrl some clients just blindly replace `{{` and `}}` with ANSI codes/HTML to highlight, e.g. [hidroh/tldroid](https://github.com/hidroh/tldroid/blob/b33394e48ffcabb39d99ceb7c69fd56322ca8a68/app/src/main/kotlin/io/github/hidroh/tldroid/MarkdownProcessor.kt#L30-L32) (though that client it no longer maintained and we're going to remove it soon I think, #4044)

navarroaxel (approved):

bl-ue:
@navarroaxel I'm not sure this was supposed to be merged actually, see our conversation above.

PR #5678, opened 2021-04-03
set-more-info-link.py: remove leading space in Persian label

bl-ue:


navarroaxel (approved):

sbrl (approved):
Merging early because this is a trivial change.

PR #5682, opened 2021-04-03
vladimyr: add page

bl-ue:
cc @vladimyr

### DIFF
--- a/pages/common/vladimyr.md
+++ b/pages/common/vladimyr.md
@@ -1,0 +1,8 @@
+# vladimyr
+
+> Dario Vladović's personal CLI.
+> More information: <https://github.com/vladimyr/vladimyr-cli>.
+
+- Start Dario's interactive CLI:
+
+`vladimyr`


sbrl (approved):
Every command documented increases coverage, I suppose!

sbrl:
(also, @bl-ue I suggest avoiding requesting review unless you really need someone's specific opinion on an issue, because otherwise it generates lots of notification noise for people.)

vladimyr:
> Every command documented increases coverage, I suppose!

And increases distribution aka `tldr.zip` size. As much as I can't say no to being _famous_ let's think a bit about bandwidth costs here.

navarroaxel (approved):

bl-ue:
Well @vladimyr image someone install the command for some reason, maybe he wants to experiment with it (I do that a lot) and then forgets about it.  Sometime later, when he's forgotten about it, he checks `tldr vladimyr` and then he remembers what it is.  For example, a user might with a script globally install a lot of packages that have `"bin":` in their `package.json`s and he runs `for command in ... ; do tldr $command >> commands.txt; done` or something to that effect.

PR #5684, opened 2021-04-04
docker-image: add page

sbrl:
A few notes about this one:

 - `docker image` ≠ `docker images`
 - A number of subcommands of `docker image` have top-level counterparts (e.g. `docker image rm` has `docker rmi`), so I've focused on unique functionality of `docker image`

### DIFF
--- a/pages/common/docker-image.md
+++ b/pages/common/docker-image.md
@@ -12,7 +12,7 @@
 
 `sudo docker image prune`
 
-- Delete all unused images, not just those without a tag:
+- Delete all unused images (not just those without a tag):
 
 `sudo docker image prune --all`
 
--- a/pages/common/docker-image.md
+++ b/pages/common/docker-image.md
@@ -6,7 +6,7 @@
 
 - List local Docker images:
 
-`sudo docker image ls`
+`docker image ls`
 
 - Delete unused local Docker images:
 
--- a/pages/common/docker-image.md
+++ b/pages/common/docker-image.md
@@ -10,12 +10,12 @@
 
 - Delete unused local Docker images:
 
-`sudo docker image prune`
+`docker image prune`
 
 - Delete all unused images (not just those without a tag):
 
-`sudo docker image prune --all`
+`docker image prune --all`
 
 - Show the history of a local Docker image:
 
-`sudo docker image history {{image}}`
+`docker image history {{image}}`
--- a/pages/common/docker-image.md
+++ b/pages/common/docker-image.md
@@ -1,0 +1,21 @@
+# docker image
+
+> Manage Docker images.
+> See also `docker build`, `docker import`, and `docker pull`.
+> More information: <https://docs.docker.com/engine/reference/commandline/image/>.
+
+- List local Docker images:
+
+`sudo docker image ls`
+
+- Delete unused local Docker images:
+
+`sudo docker image prune`
+
+- Delete all unused images, not just those without a tag:
+
+`sudo docker image prune --all`
+
+- Show the history of a local Docker image:
+
+`sudo docker image history {{image}}`


navarroaxel (approved):

pages/common/docker-image.md
```diff
@@ -0,0 +1,21 @@
+# docker image
+
+> Manage Docker images.
+> See also `docker build`, `docker import`, and `docker pull`.
+> More information: <https://docs.docker.com/engine/reference/commandline/image/>.
+
+- List local Docker images:
+
+`sudo docker image ls`
```
nicokosi, line 9: `sudo` is not required IMHO.

pages/common/docker-image.md
```diff
@@ -0,0 +1,21 @@
...
+
+- Delete unused local Docker images:
+
+`sudo docker image prune`
+
+- Delete all unused images, not just those without a tag:
```
bl-ue, line 15: What do you think about this?  
```suggestion
- Delete all unused images (not just those without a tag):
```
The original confused me at first.

pages/common/docker-image.md
bl-ue: You're right.  It's not required.

pages/common/docker-image.md
Waples: Better

sbrl: I don't really see a need for the change, but I'll apply it anyway.

pages/common/docker-image.md
sbrl: I was not aware of this. Updating page

bl-ue: Actually @sbrl `sudo` is not required on any of the examples.  I tried them all and they work for me on macOS without `sudo`.  Have you seen [this](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user)?

pages/common/docker-image.md
```diff
@@ -0,0 +1,21 @@
...
+> See also `docker build`, `docker import`, and `docker pull`.
+> More information: <https://docs.docker.com/engine/reference/commandline/image/>.
+
+- List local Docker images:
+
+`docker image ls`
+
+- Delete unused local Docker images:
+
+`sudo docker image prune`
```
bl-ue, line 13:
```suggestion
`docker image prune`
```

pages/common/docker-image.md
```diff
@@ -0,0 +1,21 @@
...
+
+- Delete all unused images (not just those without a tag):
+
+`sudo docker image prune --all`
```
bl-ue, line 17:
```suggestion
`docker image prune --all`
```

pages/common/docker-image.md
```diff
@@ -0,0 +1,21 @@
...
+
+- Show the history of a local Docker image:
+
+`sudo docker image history {{image}}`
```
bl-ue, line 21:
```suggestion
`docker image history {{image}}`
```

bl-ue (changes requested):

pages/common/docker-image.md
bl-ue: I'll fix it anf merge now.

bl-ue (approved):

sbrl:
Erm @bl-ue why did you merge this? I would have appreciated it if you'd brought it to my attention. The `sudo` _cannot_ be removed here, because you get a permission denied error otherwise:

```
$ docker image prune --all
WARNING! This will remove all images without at least one container associated to them.
Are you sure you want to continue? [y/N] y
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.24/images/prune?filters=%7B%22dangling%22%3A%7B%22false%22%3Atrue%7D%7D: dial unix /var/run/docker.sock: connect: permission denied
```

PR #5688, opened 2021-04-04
homectl: add page

navarroaxel:

### DIFF
--- a/pages/linux/homectl.md
+++ b/pages/linux/homectl.md
@@ -7,7 +7,7 @@
 
 `homectl list`
 
-- Create a user account and the associated home directory:
+- Create a user account and their associated home directory:
 
 `sudo homectl create {{username}}`
 
--- a/pages/linux/homectl.md
+++ b/pages/linux/homectl.md
@@ -7,7 +7,7 @@
 
 `homectl list`
 
-- Create a home area (the user and the associated home directory):
+- Create a user account and the associated home directory:
 
 `sudo homectl create {{username}}`
 
@@ -19,15 +19,15 @@
 
 `sudo homectl passwd {{username}}`
 
-- Run shell or a command with access to a specific home area:
+- Run shell or a command with access to a specific home directory:
 
 `sudo homectl with {{username}} -- {{command}} {{command_arguments}}`
 
-- Lock/unlock a user home area:
+- Lock/unlock a specific home directory:
 
 `sudo homectl {{lock|unlock}} {{username}}`
 
-- Resize a home area to 100 GiB:
+- Change the disk space assigned to a specific home directory to 100 GiB:
 
 `sudo homectl resize {{username}} {{100G}}`
 
--- a/pages/linux/homectl.md
+++ b/pages/linux/homectl.md
@@ -3,7 +3,7 @@
 > Create, remove, change or inspect home directories using the systemd-homed service.
 > More information: <https://man.archlinux.org/man/homectl.1>.
 
-- List home areas:
+- List user accounts and their associated home directories:
 
 `homectl list`
 
@@ -23,7 +23,7 @@
 
 `sudo homectl with {{username}} -- {{command}} {{command_arguments}}`
 
-- Lock/unlock a specific home directory:
+- Lock or unlock a specific home directory:
 
 `sudo homectl {{lock|unlock}} {{username}}`
 
--- a/pages/linux/homectl.md
+++ b/pages/linux/homectl.md
@@ -1,7 +1,7 @@
 # homectl
 
-> Create, remove, change or inspect home directories using systemd-homed service.
-> More information: <https://wiki.archlinux.org/index.php/Systemd-homed>.
+> Create, remove, change or inspect home directories using the systemd-homed service.
+> More information: <https://man.archlinux.org/man/homectl.1>.
 
 - List home areas:
 
--- a/pages/linux/homectl.md
+++ b/pages/linux/homectl.md
@@ -19,7 +19,7 @@
 
 `sudo homectl passwd {{username}}`
 
-- Run shell or a command with access to a specific home directory:
+- Run a shell or a command with access to a specific home directory:
 
 `sudo homectl with {{username}} -- {{command}} {{command_arguments}}`
 
--- a/pages/linux/homectl.md
+++ b/pages/linux/homectl.md
@@ -1,0 +1,36 @@
+# homectl
+
+> Create, remove, change or inspect home directories using systemd-homed service.
+> More information: <https://wiki.archlinux.org/index.php/Systemd-homed>.
+
+- List home areas:
+
+`homectl list`
+
+- Create a home area (the user and the associated home directory):
+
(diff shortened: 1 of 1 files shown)


pages/linux/homectl.md
```diff
@@ -0,0 +1,36 @@
+# homectl
+
+> Create, remove, change or inspect home directories using systemd-homed service.
+> More information: <https://wiki.archlinux.org/index.php/Systemd-homed>.
```
bl-ue, line 4: I wonder if we should use the man [Arch man page](https://man.archlinux.org/man/homectl.1) for the link, instead the current link, because though it's great to link to the documentation for the whole portable user accounts business, the page barely documents `homectl`; in fact, it demonstrates the simplest invocation of the command and nothing more.

The man page, on the other hand, fully documents all of the options and their valid values, and even provides an [examples section](https://man.archlinux.org/man/homectl.1#EXAMPLES) at the bottom (which you could look at for more context).
```suggestion
> More information: <https://man.archlinux.org/man/homectl.1>.
```

pages/linux/homectl.md
```diff
@@ -0,0 +1,36 @@
+# homectl
+
+> Create, remove, change or inspect home directories using systemd-homed service.
```
bl-ue, line 3:
```suggestion
> Create, remove, change or inspect home directories using the systemd-homed service.
```

bl-ue (commented):
Looks like https://wiki.archlinux.org/index.php/Systemd-homed#userdbctl could use a page too 🤔

pages/linux/homectl.md
navarroaxel: Sometimes I wonder if we should line the same output than `man homectl`, but the Arch Wiki explains another stuff around this command.

Waples:
lgtm

pages/linux/homectl.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Change the password for a specific user:
+
+`sudo homectl passwd {{username}}`
+
+- Run shell or a command with access to a specific home area:
+
+`sudo homectl with {{username}} -- {{command}} {{command_arguments}}`
+
+- Lock/unlock a user home area:
```
navarroaxel, line 26:
```suggestion
- Lock/unlock a specific home directory:
```

pages/linux/homectl.md
```diff
@@ -0,0 +1,36 @@
...
+
+`sudo homectl {{lock|unlock}} {{username}}`
+
+- Resize a home area to 100 GiB:
```
navarroaxel, line 30:
```suggestion
- Change the disk space assigned to a specific home directory to 100 GiB:
```

pages/linux/homectl.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Change the password for a specific user:
+
+`sudo homectl passwd {{username}}`
+
+- Run shell or a command with access to a specific home directory:
+
+`sudo homectl with {{username}} -- {{command}} {{command_arguments}}`
+
+- Lock/unlock a specific home directory:
```
bl-ue, line 26:
```suggestion
- Lock or unlock a specific home directory:
```

pages/linux/homectl.md
```diff
@@ -0,0 +1,36 @@
+# homectl
+
+> Create, remove, change or inspect home directories using the systemd-homed service.
+> More information: <https://man.archlinux.org/man/homectl.1>.
+
+- List home areas:
```
bl-ue, line 6:
```suggestion
- List user accounts and their associated home directories:
```

navarroaxel: Also another data:
```bash
$   homectl list
NAME     UID  GID  STATE  REALNAME     HOME           SHELL    
carl         1000 1000 active   Carl                     /home/carl /bin/bash
```

bl-ue: Oh...but I doubt that matters...do you?

navarroaxel: nah, I think is ok

bl-ue (approved):

pages/linux/homectl.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Remove a specific user and the associated home directory:
+
+`sudo homectl remove {{username}}`
+
+- Change the password for a specific user:
+
+`sudo homectl passwd {{username}}`
+
+- Run shell or a command with access to a specific home directory:
```
marchersimon, line 22:
```suggestion
- Run a shell or a command with access to a specific home directory:
```

marchersimon (approved):

pages/linux/homectl.md
```diff
@@ -0,0 +1,36 @@
...
+> Create, remove, change or inspect home directories using the systemd-homed service.
+> More information: <https://man.archlinux.org/man/homectl.1>.
+
+- List user accounts and their associated home directories:
+
+`homectl list`
+
+- Create a user account and the associated home directory:
+
+`sudo homectl create {{username}}`
```
sbrl, line 12: This is rather misleading in  that `systemd-homed` also manages creating user accounts, rather than just managing home directories. Despite this, there's little we can do about that so we don't need to change this example at all - just wanted to voice my concerns about systemd-homed in general.

Is there provision for creating system accounts with home directories here? And how does this interact with the ability for users to login or not (e.g. via SSH)?

pages/linux/homectl.md
```diff
@@ -0,0 +1,36 @@
+# homectl
+
+> Create, remove, change or inspect home directories using the systemd-homed service.
+> More information: <https://man.archlinux.org/man/homectl.1>.
+
+- List user accounts and their associated home directories:
+
+`homectl list`
+
+- Create a user account and the associated home directory:
```
sbrl, line 10: Perhaps this will read better?
```suggestion
- Create a user account and their associated home directory:
```

sbrl (commented):
Thanks for the new page, @navarroaxel! Just a few minor comments.

In general systemd homed looks very concerning indeed in that it performs a number of functions it really shouldn't - I'd suggest that this command is falsely named in that it's more of a user account management command as opposed to a home directory management command.

Regardless, documenting it is a valuable exercise as it ensures those who are unlucky enough to find themselves on the receiving end of a system with systemd-homed installed are able to quickly understand how to administrate the system.

sbrl (approved):

pages/linux/homectl.md
navarroaxel: Oh you write so much :joy_cat:

Well, if a user is not logged in and it has `.ssh/authorized_keys`, when you try to login remotely using SSH the `systemd-homed` request for the user password because (encrypted or not) the services coulnd't read the `authorized_keys` file.
There is a workaround to associate a `authorized_keys` file outside the user's home folder here: https://wiki.archlinux.org/index.php/Systemd-homed#SSH_remote_unlocking but you still need the user password to login the account itself :upside_down_face:. And when you logout from the SSH session, the account goes to `not login` states and the password is needed again :tired_face: .

PR #5694, opened 2021-04-06
sic: add page

Waples:

### DIFF
--- a/pages/linux/sic.md
+++ b/pages/linux/sic.md
@@ -1,7 +1,7 @@
 # sic
 
-> simple irc client.
-> Part of the suckless.org tools.
+> Simple IRC client.
+> Part of the <https://suckless.org/> tools.
 > More information: <https://tools.suckless.org/sic/>.
 
 - Connect to a host with the specified nickname:
@@ -12,18 +12,14 @@
 
 `sic -h {{host}} -n {{nickname}} -k {{password}}`
 
-- Usage with highlighted alerts on mention of your username:
+- Join a channel:
 
-`sic | awk '/username/ {printf "\a"}1'`
+`:j #{{channel}}<Enter>`
 
-- Join a channel (stdin):
+- Send a message:
 
-`:j #{{channel}}`
+`:m #{{channel}}/{{user}}<Enter>`
 
-- Send a message (stdin):
+- Set default channel or user:
 
-`:m #{{channel}}/{{user}}`
-
-- Set default channel or user (stdin):
-
-`:s #{{channel}}/{{user}}`
+`:s #{{channel}}/{{user}}<Enter>`
--- a/pages/linux/sic.md
+++ b/pages/linux/sic.md
@@ -4,9 +4,13 @@
 > Part of the suckless tools.
 > More information: <https://tools.suckless.org/sic/>.
 
-- Connect to a host with the specified nickname:
+- Connect to the default host (irc.ofct.net) with the nickname set in the `$USER` environment variable:
 
-`sic -h {{irc.ofct.net}} -n {{$USER}}`
+`sic`
+
+- Connect to a given host, using a given nickname:
+
+`sic -h {{host}} -n {{nickname}}`
 
 - Connect to a given host, using a given nickname and password:
 
--- a/pages/linux/sic.md
+++ b/pages/linux/sic.md
@@ -10,7 +10,7 @@
 
 - Connect to a different host with PASS connection password:
 
-`sic -n {{nickname}} -p {{password}}`
+`sic -n {{nickname}} -k {{password}}`
 
 - Usage with highlighted alerts on mention of your username:
 
--- a/pages/linux/sic.md
+++ b/pages/linux/sic.md
@@ -1,7 +1,7 @@
 # sic
 
 > Simple IRC client.
-> Part of the <https://suckless.org/> tools.
+> Part of the suckless tools.
 > More information: <https://tools.suckless.org/sic/>.
 
 - Connect to a host with the specified nickname:
--- a/pages/linux/sic.md
+++ b/pages/linux/sic.md
@@ -1,7 +1,7 @@
 # sic
 
 > simple irc client.
-> Part of suckless.org tools.
+> Part of the suckless.org tools.
 > More information: <https://tools.suckless.org/sic/>.
 
 - Connect to host (default: irc.oftc.net) with nickname (default: $USER):
--- a/pages/linux/sic.md
+++ b/pages/linux/sic.md
@@ -18,8 +18,8 @@
 
 - Send a message:
 
-`:m #{{channel}}/{{user}}<Enter>`
+`:m #{{channel|user}}<Enter>`
 
 - Set default channel or user:
 
-`:s #{{channel}}/{{user}}<Enter>`
+`:s #{{channel|user}}<Enter>`
--- a/pages/linux/sic.md
+++ b/pages/linux/sic.md
@@ -10,7 +10,7 @@
 
 - Connect to a different host with PASS connection password:
 
-`sic -n {{nickname}} -k {{password}}`
+`sic -h {{host}} -n {{nickname}} -k {{password}}`
 
 - Usage with highlighted alerts on mention of your username:
 
--- a/pages/linux/sic.md
+++ b/pages/linux/sic.md
@@ -8,7 +8,7 @@
 
 `sic -h {{irc.ofct.net}} -n {{$USER}}`
 
-- Connect to a different host with PASS connection password:
+- Connect to a given host, using a given nickname and password:
 
 `sic -h {{host}} -n {{nickname}} -k {{password}}`
 
--- a/pages/linux/sic.md
+++ b/pages/linux/sic.md
@@ -24,6 +24,6 @@ Send a message (stdin):
 
 `:m #{{channel}}/{{user}}`
 
-Set default channel or user:
+Set default channel or user (stdin):
 
 `:s #{{channel}}/{{user}}`
--- a/pages/linux/sic.md
+++ b/pages/linux/sic.md
@@ -4,9 +4,9 @@
 > Part of the suckless.org tools.
 > More information: <https://tools.suckless.org/sic/>.
 
-- Connect to host (default: irc.oftc.net) with nickname (default: $USER):
+- Connect to a host with the specified nickname:
 
-`sic -n {{nickname}}`
+`sic -h {{irc.ofct.net}} -n {{$USER}}`
 
 - Connect to a different host with PASS connection password:
 
--- a/pages/linux/sic.md
+++ b/pages/linux/sic.md
@@ -16,7 +16,7 @@
 
 `:j #{{channel}}<Enter>`
 
-- Send a message:
+- Send a message to a channel or user:
 
 `:m #{{channel|user}}<Enter>`
 
--- a/pages/linux/sic.md
+++ b/pages/linux/sic.md
@@ -4,26 +4,26 @@
 > Part of suckless.org tools.
 > More information: <https://tools.suckless.org/sic/>.
 
-Connect to host (default: irc.oftc.net) with nickname (default: $USER):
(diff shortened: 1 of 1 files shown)


Waples:
suckless doesn't use Capital letters. Can we get a pass on `TLDR003` ?

waldyrious:
> suckless doesn't use Capital letters. Can we get a pass on TLDR003 ?

Do you mean that "sic" is meant to stand for "simple irc client", all in lowercase? I don't think we need to conform to stylistic preferences of the projects we document, as long as the content itself is accurate. We can accommodate stylization of names themselves (as we do e.g. with [LaTeX](https://github.com/tldr-pages/tldr/blob/5a40dd8341ae2b69384f4033ab8bb9b3abe9a992/pages/common/latex.md)), but the descriptions in tldr pages are meant to be useful and informative, not to represent the project's branding.

I would go with:

```
# sic
> Simple IRC client.
```

owenvoke:
I think it should be fine to just use `Simple` instead of `simple`...

pages/linux/sic.md
```diff
@@ -0,0 +1,29 @@
+# sic
+
+> simple irc client.
+> Part of suckless.org tools.
+> More information: <https://tools.suckless.org/sic/>.
+
+- Connect to host (default: irc.oftc.net) with nickname (default: $USER):
```
sbrl, line 7: You don't specify the host here to connect to, so I'm not sure the description here makes sense. What about this?
```suggestion
- Connect to irc.oftc.net with a given nickname (default: `$USER`):
```

pages/linux/sic.md
```diff
@@ -0,0 +1,29 @@
...
+
+`sic -n {{nickname}}`
+
+- Connect to a different host with PASS connection password:
+
+`sic -h {{host}} -n {{nickname}} -k {{password}}`
```
sbrl, line 13: Specifying a password on the command line like this is not very secure. Is there a way to prompt the user for a password, or read it from a file?

pages/linux/sic.md
```diff
@@ -0,0 +1,29 @@
...
+
+- Usage with highlighted alerts on mention of your username:
+
+`sic | awk '/username/ {printf "\a"}1'`
+
+- Join a channel (stdin):
```
sbrl, line 19: Not sure what you mean by `(stdin)` here, or in future examples.

pages/linux/sic.md
```diff
@@ -0,0 +1,29 @@
...
+
+`sic -n {{nickname}}`
+
+- Connect to a different host with PASS connection password:
+
+`sic -h {{host}} -n {{nickname}} -k {{password}}`
+
+- Usage with highlighted alerts on mention of your username:
+
+`sic | awk '/username/ {printf "\a"}1'`
```
sbrl, line 17: This example seems to demonstrate the usage of the `awk` command, rather than the `sic` command. Generally in tldr pages we try to keep pages focused on the command they are about.

pages/linux/sic.md
```diff
@@ -0,0 +1,29 @@
+# sic
+
+> simple irc client.
+> Part of suckless.org tools.
```
sbrl, line 4: Perhaps this will read better?
```suggestion
> Part of the suckless.org tools.
```

sbrl (changes requested):
Thanks for the new page, @Waples! I've left some comments below for you to review.

Waples:
Yes, `sic` stands literally for `simple irc client`, all in lower case @waldyrious. Fair enough, but I like to honor the title of the project/product at least. Simplistic; it's `sic`. (sometimes TLDR003 is not needed imo).

pages/linux/sic.md
Waples: Devil is in the details, the explain part mentions host and the default.
Compromise: how about `connect to default (irc.ofct.net) irc host` ? 
Though I like my version better.

pages/linux/sic.md
Waples: @sbrl No. 
have a look [here](https://tools.suckless.org/sic/) =]

pages/linux/sic.md
Waples: `sic` works on `stdin` input. Please see [usage](https://tools.suckless.org/sic/usage/) and/or the man page of `sic`.

pages/linux/sic.md
Waples: No, this example demonstrates how you can highlight a message when your username is mentioned when using `sic`.

pages/linux/sic.md
Waples: fair enough.

Waples:
@sbrl you should try out `sic`, it will make more sense ;]

pages/linux/sic.md
bl-ue: How about instead `Connect to host with a given nickname:` and instead of `nickname` in the example itself use `{{$USER}}` and show add `-h {{irc.ofct.net}}`?

pages/linux/sic.md
bl-ue: Yeah, I'm not sure if we should keep this example.

pages/linux/sic.md
bl-ue: @Waples I see what you mean—this is a TUI and therefore when you start the tool it offers a prompt for you to enter commands etc.. So just omit `(stdin)` here and in the next examples.

See other documented TUIs for reference, e.g.:
- [`rainbowstream.md`](https://github.com/tldr-pages/tldr/blob/master/pages/common/rainbowstream.md)
- [`rtv.md`](https://github.com/tldr-pages/tldr/blob/master/pages/common/rtv.md)
- [`vim.md`](https://github.com/tldr-pages/tldr/blob/master/pages/common/vim.md)

pages/linux/sic.md
bl-ue: This is unfortunate but it doesn't matter much.  It's their fault if the command is bad, we're just documenting how to use it.

Waples: Exactly.

pages/linux/sic.md
Waples: Thanks @bl-ue, i'll implement that.

pages/linux/sic.md
```diff
@@ -0,0 +1,29 @@
+# sic
+
+> simple irc client.
+> Part of the suckless.org tools.
+> More information: <https://tools.suckless.org/sic/>.
+
+- Connect to host (default: irc.oftc.net) with nickname (default: $USER):
+
+`sic -n {{nickname}}`
```
bl-ue, line 9: I'm thinking this (from https://github.com/tldr-pages/tldr/pull/5694/files#r607888085)
```suggestion
- Connect to a host with the specified nickname:

`sic -h {{irc.ofct.net}} -n {{$USER}}`
```

pages/linux/sic.md
bl-ue: See, this is what I'm thinking of: https://github.com/tldr-pages/tldr/pull/5694/files#r607929529

pages/linux/sic.md
Waples: seems legit

bl-ue:
Regarding the casing, I admit that I agree with @waldyrious's [opinion](https://github.com/tldr-pages/tldr/pull/5694#issuecomment-814006257).  As he stated, we're not required to adher to the projects particular formats, and as 'simple irc client' is merely a description, changing it to match our existing ~3000 pages is the best course of action.  Over the years, there have been many such discussions about this exemption or that, but in the end, the simple approach is chosen, because otherwise the exceptions pile up and become very difficult to manage.

Waples:
@bl-ue then what ya think about.... `Suckless: sic` / `Suckless: sic, simple irc client` ?

Waples:
simple irc client is also the name (hence sic)

bl-ue:
@Waples:

- I'd think that a user would mentally match `sic` with `Simple IRC client`, so I think that's all that's necessary
- Also, that the tool is a `suckless` tool is made obvious by the link.

What do you think?

bl-ue:
I would, doesn't seem that useful given it's on the very next line (albeit slightly dissimilar, I don't it matters much)

Waples:
fixed

bl-ue (approved):

owenvoke (approved):

Waples:
bump @sbrl

pages/linux/sic.md
```diff
@@ -0,0 +1,25 @@
+# sic
+
+> Simple IRC client.
+> Part of the suckless tools.
+> More information: <https://tools.suckless.org/sic/>.
+
+- Connect to a host with the specified nickname:
+
+`sic -h {{irc.ofct.net}} -n {{$USER}}`
```
waldyrious, line 9: Maybe we could add the most basic example before this one?

```markdown
- Connect to the default host (irc.ofct.net) with the nickname set in the `$USER` variable:

`sic`
```

Then this one could be simply:

```markdown
- Connect to a host using a given nickname:

`sic -h {{host}} -n {{nickname}}`
```

pages/linux/sic.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Connect to a different host with PASS connection password:
```
waldyrious, line 11: I'm confused — what does the uppercase `PASS` refer to here?

waldyrious:
@Waples thanks for your patience with our review comments! The page looks quite good now — I left just a few comments inline, please take a look.

pages/linux/sic.md
Waples: I like, if you submit the suggestion, i'll add it.

pages/linux/sic.md
Waples: tbh, it's a copy paste from suckless docs; don't mind to remove the PASS part, turn it into irc password or something?

pages/linux/sic.md
waldyrious, line 9: WDYT @bl-ue?

```suggestion
- Connect to the default host (irc.ofct.net) with the nickname set in the `$USER` variable:

`sic`

- Connect to a given host, using a given nickname:

`sic -h {{host}} -n {{nickname}}`
```

waldyrious: WDYT @bl-ue?

```suggestion
- Connect to the default host (irc.ofct.net) with the nickname set in the `$USER` variable:

`sic`

- Connect to a given host, using a given nickname:

`sic -h {{host}} -n {{nickname}}`
```

pages/linux/sic.md
waldyrious: If my previous edit suggestion reaches consensus, then we could use matching language here as well, for consistency:

```suggestion
- Connect to a given host, using a given nickname and password:
```

Though before we make this change, I think we need to clarify what happens in the previous example (passing `-n` but not `-p`) — does sic expect the password to be in an environment variable too (say, `$PASS`)? Does it ask for it interactively?

pages/linux/sic.md
bl-ue: Looks good to me, one thing though: ~`$USER` variable~ → **`$USER` environment variable**, right, @waldyrious?

waldyrious: Right. I pondered that, and chose not to make the description too long, but it's best to be explicit. 👍

pages/linux/sic.md
Waples: I don't use it like that normally, but just tested this out with `$PASS` set as an environment variable without `-k {{pass}}`. Does not work. So I'll commit your suggestion.

Waples:
applied suggestions

waldyrious (approved):
LGTM — thanks again @Waples for the submission and patience! Since this already has two approvals, I'll go ahead and merge after I add my own.

sbrl (commented):
Sorry for not getting back to this. @Waples: That `awk` example was indeed showing highlighted messages, but it did not do so by calling `sic` with different arguments - hence it would be a better fit for a wider cheat sheet as opposed to a compact tldr page.

No need to send bump messages to get my attention - it was in my notifications list - I just haven't had time to return to this until now.

PR #5700, opened 2021-04-06
flashrom: add page

peterbabic:

### DIFF
--- a/pages/linux/flashrom.md
+++ b/pages/linux/flashrom.md
@@ -1,24 +1,24 @@
 # flashrom
 
 > Read, write, verify and erase flash chips.
-> More information: <https://man.archlinux.org/man/flashrom.8>.
+> More information: <https://manned.org/flashrom>.
 
 - Probe the chip, ensuring the wiring is correct:
 
 `flashrom --programmer {{programmer}}`
 
-- Read flash and save it into file:
+- Read flash and save it to a file:
 
 `flashrom -p {{programmer}} --read {{path/to/file}}`
 
-- Write file into flash:
+- Write a file to the flash:
 
 `flashrom -p {{programmer}} --write {{path/to/file}}`
 
-- Verify flash against file:
+- Verify the flash against a file:
 
 `flashrom -p {{programmer}} --verify {{path/to/file}}`
 
-- Probe the chip using RaspberryPi, an example:
+- Probe the chip using RaspberryPi:
 
-`flashrom -p linux_spi:dev=/dev/spidev0.0`
+`flashrom -p {{linux_spi:dev=/dev/spidev0.0}}`
--- a/pages/linux/flashrom.md
+++ b/pages/linux/flashrom.md
@@ -5,19 +5,19 @@
 
 - Probe the chip, ensuring the wiring is correct:
 
-`flashrom -p {{programmer}}`
+`flashrom --programmer {{programmer}}`
 
 - Read flash and save it into file:
 
-`flashrom -p {{programmer}} -r {{path/to/file}}`
+`flashrom -p {{programmer}} --read {{path/to/file}}`
 
 - Write file into flash:
 
-`flashrom -p {{programmer}} -w {{path/to/file}}`
+`flashrom -p {{programmer}} --write {{path/to/file}}`
 
-- Erase flash:
+- Verify flash against file:
 
-`flashrom -p {{programmer}} -E`
+`flashrom -p {{programmer}} --verify {{path/to/file}}`
 
 - Probe the chip using RaspberryPi, an example:
 
--- a/pages/linux/flashrom.md
+++ b/pages/linux/flashrom.md
@@ -1,0 +1,24 @@
+# flashrom
+
+> Read, write, verify and erase flash chips.
+> More information: <https://man.archlinux.org/man/flashrom.8>.
+
+- Probe the chip, ensuring the wiring is correct:
+
+`flashrom -p {{programmer}}`
+
+- Read flash and save it into file:
+
+`flashrom -p {{programmer}} -r {{path/to/file}}`
+
+- Write file into flash:
+
+`flashrom -p {{programmer}} -w {{path/to/file}}`
+
+- Erase flash:
+
+`flashrom -p {{programmer}} -E`
+
+- Probe the chip using RaspberryPi, an example:
+
+`flashrom -p linux_spi:dev=/dev/spidev0.0`


schneiderl (approved):

pages/linux/flashrom.md
```diff
@@ -0,0 +1,24 @@
+# flashrom
+
+> Read, write, verify and erase flash chips.
+> More information: <https://man.archlinux.org/man/flashrom.8>.
+
+- Probe the chip, ensuring the wiring is correct:
+
+`flashrom -p {{programmer}}`
```
bl-ue, line 8:
```suggestion
`flashrom --programmer {{programmer}}`
```

pages/linux/flashrom.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Read flash and save it into file:
+
+`flashrom -p {{programmer}} -r {{path/to/file}}`
```
bl-ue, line 12:
```suggestion
`flashrom --programmer {{programmer}} --read {{path/to/file}}`
```

pages/linux/flashrom.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Write file into flash:
+
+`flashrom -p {{programmer}} -w {{path/to/file}}`
```
bl-ue, line 16:
```suggestion
`flashrom --programmer {{programmer}} --write {{path/to/file}}`
```

pages/linux/flashrom.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Erase flash:
+
+`flashrom -p {{programmer}} -E`
```
bl-ue, line 20:
```suggestion
`flashrom --programmer {{programmer}} --erase`
```

pages/linux/flashrom.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Probe the chip using RaspberryPi, an example:
+
+`flashrom -p linux_spi:dev=/dev/spidev0.0`
```
bl-ue, line 24:
```suggestion
`flashrom --programmer {{linux_spi:dev=/dev/spidev0.0}}`
```

bl-ue (commented):
Welcome back, @peterbabic!!

peterbabic:
Hello @bl-ue , while I appreiate the suggested changes and I agree with the longer version of verbs, I disagree with extending the `-p` to `--programmer`. The word after the `-p` usually gets pretty long, because contains type of the programmer, path to it and even optional parameters behind the colon. Due to this fact, and due to fact that `-p` is the only required parameter that goes with every single call of the flashrom command, mostly everyone uses just `-p`.

marchersimon:
@peterbabic the philosophy of the tldr project is to make people understand the commands and their arguments, not just memorize them. But sure, once you're comfortable using a command and understand what it does you can look up the short form.

bl-ue:
@peterbabic: as @marchersimon says, we try to help users understand the options, not just memorize them.  However, you bring up a valid point, and I think that we can probably just use `-p` for this page, since it's in every example and as you say the value for this option can get rather lengthy.

I think that we'll probably come up with a mechanism eventually that allows us to to specify both short and long options and have the tldr-clients display one or the other if you configure them to do so, but I'm not sure. 🤷🏻

peterbabic:
Thank you all for explanations. 

Some notes: from the examples, all but `--erase` which translate to to `-E` not `-e` follow the convention (the first letter of the long paramater in lowercase). Maybe erase could be replaced with `--verify` with `-v` and then the examples are quite self-explanatory. This would also provide example that would explain why `flashrom -v` does not print the version, as is common.

bl-ue:
Sounds good to me. 👍🏻 

>This would also provide example that would explain why flashrom -v does not print the version, as is common.

Great point.

pages/linux/flashrom.md
```diff
@@ -0,0 +1,24 @@
+# flashrom
+
+> Read, write, verify and erase flash chips.
+> More information: <https://man.archlinux.org/man/flashrom.8>.
```
marchersimon, line 4:
```suggestion
> More information: <https://manned.org/flashrom>.
```
Since it isn't an Arch specific command, we should use the `manned.org` page.

pages/linux/flashrom.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Probe the chip, ensuring the wiring is correct:
+
+`flashrom --programmer {{programmer}}`
+
+- Read flash and save it into file:
```
marchersimon, line 10:
```suggestion
- Read flash and save it to a file:
```

pages/linux/flashrom.md
```diff
@@ -0,0 +1,24 @@
...
+
+`flashrom -p {{programmer}} --read {{path/to/file}}`
+
+- Write file into flash:
```
marchersimon, line 14:
```suggestion
- Write a file to the flash:
```

pages/linux/flashrom.md
```diff
@@ -0,0 +1,24 @@
...
+
+`flashrom -p {{programmer}} --write {{path/to/file}}`
+
+- Verify flash against file:
```
marchersimon, line 18:
```suggestion
- Verify the flash against a file:
```

pages/linux/flashrom.md
```diff
@@ -0,0 +1,24 @@
...
+
+`flashrom -p {{programmer}} --verify {{path/to/file}}`
+
+- Probe the chip using RaspberryPi, an example:
```
marchersimon, line 22:
```suggestion
- Probe the chip using RaspberryPi:
```

pages/linux/flashrom.md
bl-ue: @peterbabic I updated the above suggestion.

pages/linux/flashrom.md
peterbabic: Agreed. For a record, I did not know about manned.org, I compared arch to linux.die.net, found out that both versions differ, assumed arch is more relevant, pasted address.

pages/linux/flashrom.md
peterbabic: I am still not sure what that handlebars do to be honest, but I agree.

marchersimon (approved):

bl-ue (approved):

bl-ue:
Great!  Thank you, @peterbabic!

peterbabic:
The pleasure's on my side :)

PR #5701, opened 2021-04-07
git-range-diff: add page

xeruf:

Helpful SO link for reference:
https://stackoverflow.com/questions/52323008/can-somebody-explain-the-usage-of-git-range-diff

### DIFF
--- a/pages/common/git-range-diff.md
+++ b/pages/common/git-range-diff.md
@@ -3,10 +3,14 @@
 > Compare two commit ranges (e.g. two versions of a branch).
 > More information: <https://git-scm.com/docs/git-range-diff>.
 
-- Diff the diffs of commit1 and commit2:
+- Diff the changes in commit1 and commit2:
 
-`git range-diff {{commit1}}~..{{commit1} {{commit2}}~..{{commit2}}`
+`git range-diff {{commit1}}^! {{commit2}}^!`
 
-- Diff the changes of ours and theirs from their common ancestor:
+- Diff the changes of ours and theirs from their common ancestor, e.g. after an interactive rebase:
 
 `git range-diff {{theirs}}...{{ours}}`
+
+- Diff the changes of two commit ranges, e.g. to check whether conflicts have been resolved appropriately when rebasing commits from `base1` to `base2`:
+
+`git range-diff {{base1}}..{{rev1}} {{base2}}..{{rev2}}`
--- a/pages/common/git-range-diff.md
+++ b/pages/common/git-range-diff.md
@@ -7,6 +7,6 @@
 
 `git range-diff {{commit1}}~..{{commit1} {{commit2}}~..{{commit2}}`
 
-- Diff the changes of ours and theirs from their common ancestor
+- Diff the changes of ours and theirs from their common ancestor:
 
 `git range-diff {{theirs}}...{{ours}}`
--- a/pages/common/git-range-diff.md
+++ b/pages/common/git-range-diff.md
@@ -1,0 +1,12 @@
+# git range-diff
+
+> Compare two commit ranges (e.g. two versions of a branch)
+> More information: <https://git-scm.com/docs/git-range-diff>
+
+- Diff the diffs of commit1 and commit2:
+
+`git range-diff {{commit1}}~..{{commit1} {{commit2}}~..{{commit2}}`
+
+- Diff the changes of ours and theirs from their common ancestor
+
+`git range-diff {{theirs}}...{{ours}}`
--- a/pages/common/git-range-diff.md
+++ b/pages/common/git-range-diff.md
@@ -1,7 +1,7 @@
 # git range-diff
 
-> Compare two commit ranges (e.g. two versions of a branch)
-> More information: <https://git-scm.com/docs/git-range-diff>
+> Compare two commit ranges (e.g. two versions of a branch).
+> More information: <https://git-scm.com/docs/git-range-diff>.
 
 - Diff the diffs of commit1 and commit2:
 
--- a/pages/common/git-range-diff.md
+++ b/pages/common/git-range-diff.md
@@ -3,9 +3,9 @@
 > Compare two commit ranges (e.g. two versions of a branch).
 > More information: <https://git-scm.com/docs/git-range-diff>.
 
-- Diff the changes in commit1 and commit2:
+- Diff the changes of two individual commits:
 
-`git range-diff {{commit1}}^! {{commit2}}^!`
+`git range-diff {{commit_1}}^! {{commit_2}}^!`
 
 - Diff the changes of ours and theirs from their common ancestor, e.g. after an interactive rebase:
 


pages/common/git-range-diff.md
```diff
@@ -0,0 +1,12 @@
+# git range-diff
+
+> Compare two commit ranges (e.g. two versions of a branch).
+> More information: <https://git-scm.com/docs/git-range-diff>.
+
+- Diff the diffs of commit1 and commit2:
+
+`git range-diff {{commit1}}~..{{commit1} {{commit2}}~..{{commit2}}`
+
+- Diff the changes of ours and theirs from their common ancestor:
```
marchersimon, line 10: I'm not sure what this command does, but I think there should be better words for `ours` and `theirs`. @bl-ue what do you think?

xerus2000: Please look at the SO question I linked for a longer explanation

bl-ue: Personally, I don't mind the usage of `ours` and `theirs` because that's official Git wording, and [we allow such exceptions](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#inclusive-language).

bl-ue: However, we don't seem to use this terminology in any of the other Git pages, so maybe we should try to move away from it.  @xerus2000 also see the [`git-diff` page](https://github.com/tldr-pages/tldr/blob/a99a8d98d86e857ab1174804cfbecf8b170f9cd7/pages/common/git-diff.md).

pages/common/git-range-diff.md
```diff
@@ -0,0 +1,12 @@
...
+
+`git range-diff {{theirs}}...{{ours}}`
```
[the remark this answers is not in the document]
xerus2000: This is very misleading.
Neither of these is "base" or "head", they are just two committishs, where the base is the common ancestor.

pages/common/git-range-diff.md
xerus2000: yep, I had a look at that.
tbh, if we want to be correct, the mentions of commit and branch in such pages should actually be replaced by `commit-ish` (see https://stackoverflow.com/a/23303550) since these are interchangeable.

bl-ue: Wow, you're right. We might need to change that.  Let's wait for @sbrl's opinion then.

pages/common/git-range-diff.md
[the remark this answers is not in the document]
bl-ue: Agreed.  ping @sbrl

pages/common/git-range-diff.md
sbrl: Heya! Sorry for the wait here. I would suggest that `ours` and `theirs` would be best used when disambiguating merges - e.g. "our commits" and "their commits". I'm not sure it really helps clarity here - I suggest changing the wording here would be the best course of action to be more specific - both as to _what_ is being referred to, and also how their are different from one another.

The man page is also singularly unhelpful:

```
       <rev1>...<rev2>
           Equivalent to passing <rev2>..<rev1> and <rev1>..<rev2>.
```

I'm unsure as how useful this example is. In theory it looks like to me that this command would be most useful when used to view the differences between 2 slightly different patches.

For example, Bob creates a patch set for project Foo to add a new feature. Bob sends the patch set to Alice, who makes some changes to the patch set with `git rebase -i` and sends them back to Bob. Bob then uses `git range-diff` to compare Alice's patch set with their own local patch set.

pages/common/git-range-diff.md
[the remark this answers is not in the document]
sbrl: I see. What about this?
```suggestion
`git range-diff {{rev_a}}...{{rev_b}}`
```

In this particular and specific instance the man page suggests that this is not a commit-ish but rather a specific revision:

```
SYNOPSIS
       git range-diff [--color=[<when>]] [--no-color] [<diff-options>]
               [--no-dual-color] [--creation-factor=<factor>]
               ( <range1> <range2> | <rev1>...<rev2> | <base> <rev1> <rev2> )

```

pages/common/git-range-diff.md
xerus2000: Yes, and checking the diff after a rebase is a core use-case of this command imo, that's why I want this example. But that still leaves the question of the wording open ;)

pages/common/git-range-diff.md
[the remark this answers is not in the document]
xerus2000: well, it definitely also accepts branch names, which are commitishs not revisions. I guess the man page only abbreviates that for simplicity.

reference: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefrevisionarevision

pages/common/git-range-diff.md
sbrl: Oh, I see! I'm not sure I totally understand how that would work though, so I'm unsure if I'm the best person to come up with wordingg here @xerus2000. Given that's the central use-case of this command though, it might make sense to base the example around that to make it more relatable and useful. Perhaps we could reword this example to talk specifically about checking a diff after a rebase?

xerus2000: Tried adding examples :)

Waples (approved):

marchersimon (approved):

pages/common/git-range-diff.md
```diff
@@ -0,0 +1,16 @@
+# git range-diff
+
+> Compare two commit ranges (e.g. two versions of a branch).
+> More information: <https://git-scm.com/docs/git-range-diff>.
+
+- Diff the changes in commit1 and commit2:
```
bl-ue, line 6: Would this work?
```suggestion
- Diff the changes between two commits:
```

xerus2000: I'm a bit wary here. Phrased like that it sounds like diff, which is what I'm trying to distinguish it from. It is not about the _changes between two commits_ but _the difference of changes within_ each commit (range).

Waples:
bump @bl-ue

Waples:
@bl-ue you gonna review this again or is it ready?

pages/common/git-range-diff.md
CleanMachine1:
```suggestion
- Diff the changes in between 2 specific commits:
```

CleanMachine1:
We should look at the existing unresolved review bl-ue has left, I have added my input on it as well

bl-ue:
@CleanMachine1 `stalebot ignore` is used for labels that we're ignoring because of stuff like waiting for linter updates, etc...

pages/common/git-range-diff.md
marchersimon: Another suggestion:
```suggestion
- Diff the changes of two specific commits:
```

CleanMachine1: @marchersimon your suggestion falls under the comment @xeruf made

pages/common/git-range-diff.md
```diff
@@ -0,0 +1,16 @@
...
+
+`git range-diff {{commit1}}^! {{commit2}}^!`
```
CleanMachine1, line 8:
```suggestion
`git range-diff {{commit_1}}^! {{commit_2}}^!`
```

This just makes it look a little less messy

CleanMachine1:
The tag is needed. It has been 30 days since latest reply

pages/common/git-range-diff.md
marchersimon: Then how about:
```suggestion
- Diff the changes made by two specific commits:
```
?

CleanMachine1: Still think between is better but still need other's opinion

sbrl: What about this then?
```suggestion
- Diff the changes with in commit within the range from `commit1` to `commit2`:
```

CleanMachine1: Yes that would work however its too long, my suggestion is still the best IMO

bl-ue: `with in`??

CleanMachine1: > Within and without should not be written with a space between with/in or with/out, according to all major dictionaries.

bl-ue: I see.  Looks like a typo @sbrl.  I assume you mean like this?
```suggestion
- Diff the changes within the range from `commit1` to `commit2`:
```

xeruf: @bl-ue you are again misunderstanding the command. There is no range created of these two commits. Only the diffs of the commits themselves are diffed.

As for between: In all suggestions so far it could be interpreted also in its time dimension, feeding the same misunderstanding as @bl-ue

xeruf: @sbrl also understood it wrongly.
The suggestions by @marchersimon were pretty good so I adapted them.
Don't submit any more suggestions unless you are sure you know what the command does.

xeruf:
Ready to merge imo

pages/common/git-range-diff.md
bl-ue: I would be great if we could find an example... 😆

bl-ue (approved):

pages/common/git-range-diff.md
```diff
@@ -0,0 +1,16 @@
...
+
+`git range-diff {{commit_1}}^! {{commit_2}}^!`
+
+- Diff the changes of ours and theirs from their common ancestor, e.g. after an interactive rebase:
+
+`git range-diff {{theirs}}...{{ours}}`
+
+- Diff the changes of two commit ranges, e.g. to check whether conflicts have been resolved appropriately when rebasing commits from `base1` to `base2`:
+
+`git range-diff {{base1}}..{{rev1}} {{base2}}..{{rev2}}`
```
CleanMachine1, line 16:
```suggestion
- Diff the changes of two commit ranges, e.g. to check whether conflicts have been resolved appropriately when rebasing commits from `base_1` to `base_2`:

`git range-diff {{base_1}}..{{rev_1}} {{base_2}}..{{rev_2}}`
```

CleanMachine1:
I'd like to dismiss simons review and get another one since its very out of date and I am not sure what his person opinion on this currently is

pages/common/git-range-diff.md
bl-ue: I'm not sure that this is necessary — the extra underscores render it a little less readble IMO.

marchersimon: I also think we should leave it. Since neither `e`, nor `v` is full-height, the number following it is quite readable.

marchersimon (approved):
LGTM, thanks @xeruf

pages/common/git-range-diff.md
CleanMachine1: However to keep standard, the underscores were used earlier

xeruf: I guess we should revert that, since git itself also uses no underscores, see for example:

    find /usr/share/man -name "*git*" | xargs gunzip -c | grep commit1

produces some results while the same with underscores doesn't

sbrl: Sometimes underscores (aka snake_case) make it more readable, and it other situations they don't.

In this case, given the number of items here I'm not convinced that this improves it.

pages/common/git-range-diff.md
sbrl: Oops, my suggestion meant to be `within each commit`.

pages/common/git-range-diff.md
CleanMachine1: feel free to ignore this then, just keeping standard since snake was used earlier

sbrl (approved):
This looks ok to me. This PR has been sitting around for a while and I think it's as good as it's going to get, so let's merge it as-is.

pages/common/git-range-diff.md
CleanMachine1: https://github.com/tldr-pages/tldr/pull/5701/files#diff-83dbb0d1c631c1f5de07d28506db0756b728f046436d4ae0524d7f05d46c2364R8

For reference to what my previous message included.

xeruf:
Thanks :) I only check-in every few weeks on such projects since I have other more important projects my plate, with PRs like this being little side-strolls ;)

bl-ue:
>with PRs like this being little side-strolls ;)

And they shouldn't be anything else 😉

PR #5711, opened 2021-04-08
package.json: Move husky install to prepare run-script

MasterOdin:

The recommended way to use husky with npm is to put the `husky install` line in the `prepare` run-script, not `postinstall`. This is principally aimed at libraries where using `postinstall` would affect downstream consumers (unless you used pinst to prune the `postinstall` line), but might as well be consistent here.


bl-ue:
By the way, @MasterOdin in case you're not aware, you can modify the labels of issues/PRs yourself.

bl-ue (approved):

marchersimon (approved):

Issue #4763, opened 2020-10-19
add contributor: @glenacota


sbrl:
Hi, @glenacota! You seem to be enjoying contributing to the tldr-pages project.
You now have had five distinct pull requests merged (#4364, #4362, #2464, #2458, #2452, and #2440)!
That qualifies you to become a collaborator in this repository, as explained in our [community roles documentation](https://github.com/tldr-pages/tldr/blob/master/COMMUNITY-ROLES.md).

As a collaborator, you will have commit access to the repository.
That means you can merge pull requests, label and close issues, and perform various other maintenance tasks that are needed here and there.
Of course, all of this is voluntary — you're welcome to contribute to the project in whatever ways suit your liking.

If you do decide to start performing maintenance tasks, though, we only ask you to get familiar with the [maintainer's guide](https://github.com/tldr-pages/tldr/blob/master/contributing-guides/maintainers-guide.md).

So, what do you say? Can we add you as a collaborator?

Either way, thanks for all your work so far!

glenacota:
Hi, @sbrl, fantastic Maintainer's guide! Sure, I'd be really happy to be added as a collaborator.

Like most of you, there are periods that I can invest more time on it (like recently) and some period less time, but `tldr` it's a community I've always loved to come back to and support!

sbrl:
Invite sent!

There's a bit of a lead time at the moment on my reading of notifications, because it's both hacktoberfest and I'm busy setting up a new laptop.

schneiderl:
Welcome @glenacota 😄 

Was the invite already accepted? If so, can we close this issue?

bl-ue:
@glenacota are you a collaborator yet?

PR #5715, opened 2021-04-09
MAINTAINERS: normalize dates, reorder entries

bl-ue:
* Format all of the dates in the format '1 January 2021'
* Add @owenvoke and @mebeim to the collaborators section as emeritus
* Sort all entries by date
* Set all of the dates to the UTC date of the opening of the linked issue
* Minor wording adjustments

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -12,82 +12,84 @@ repository collaborators, organization members, and organization owners
 In general terms, all maintainers are expected to follow the
 [Maintainer's guide](contributing-guides/maintainers-guide.md).
 
-
 ## Repository collaborators
 
-Repository collaborators have write access to this repository,
-which allows them to label, edit and close issues,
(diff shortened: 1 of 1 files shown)


navarroaxel (approved):

marchersimon (approved):

owenvoke (approved):

waldyrious:
> Set all of the dates to the UTC date of the opening of the linked issue

Shouldn't the date be the date when the actual role change occurred? Typically this would be when the issue is closed, or at least there should be a comment in it indicating that the action had been taken.

bl-ue:
Yes, usually there is, but not always.  For example: #4763.  In that issue, it's not clear when the invitation was accepted.

waldyrious:
I understand. But I'd rather go with exceptions/approximations in those cases, than make everyone off by a day or two. After all, it's always possible to dig up the correct date later on.

PR #5716, opened 2021-04-09
firebase: refresh

bl-ue:


pages/common/firebase.md
```diff
@@ -1,25 +1,28 @@
 # firebase
 
-> Firebase Command Line Interface (CLI) Tools.
-> Used to test, manage, and deploy your Firebase project from the command line.
+> Test, manage, and deploy Firebase projects from the command line.
 > More information: <https://github.com/firebase/firebase-tools>.
 
-- Login:
+- Login to https://console.firebase.google.com:
```
marchersimon, line 6: Should we maybe highlight URLs somehow? Maybe even making them a link with `<...>`?

bl-ue: Not a bad idea, I think probably the angle brackets won't get removed in most clients, but that might not matter.  @sbrl?

marchersimon: Shouldn't that end up the same as the `More information` link?

bl-ue: Yes, but I was thinking that it wouldn't be as acceptable there.  I don't think it probably matters though.
```suggestion
- Login to <https://console.firebase.google.com>:
```

marchersimon (approved):

pages/common/firebase.md
```diff
@@ -1,25 +1,28 @@
 # firebase
 
-> Firebase Command Line Interface (CLI) Tools.
-> Used to test, manage, and deploy your Firebase project from the command line.
+> Test, manage, and deploy Firebase projects from the command line.
 > More information: <https://github.com/firebase/firebase-tools>.
 
-- Login:
+- Login to <https://console.firebase.google.com>:
```
navarroaxel, line 6: can the user login to another server?

bl-ue: I don't think so, but I'm not sure.  I'll check later

sbrl (approved):

pages/common/firebase.md
marchersimon: Ping @bl-ue

bl-ue: @navarroaxel Nope, this is the only server users can login to.  But I think we should still leave it in because it's probable that people will want to know.  I would, at least.

navarroaxel (approved):

PR #5718, opened 2021-04-09
autossh: refresh

bl-ue:


pages/common/autossh.md
```diff
@@ -1,29 +1,29 @@
 # autossh
 
 > Runs, monitors and restarts SSH connections.
```
marchersimon, line 3:
```suggestion
> Run, monitori and restart SSH connections.
```
Don't we usually keep the descriptions in this format?

pages/common/autossh.md
```diff
@@ -1,29 +1,29 @@
...
-> Auto-reconnects to keep port forwarding tunnels up. Accepts all ssh flags.
+> Auto-reconnects to keep port forwarding tunnels up. Accepts all `ssh` flags.
 > More information: <https://harding.motd.ca/autossh>.
```
marchersimon, line 5:
```suggestion
> More information: <https://www.harding.motd.ca/autossh>.
```

pages/common/autossh.md
```diff
@@ -1,29 +1,29 @@
...
 
-- Open an SSH session, restarting when a monitoring port fails return data:
+- Start an SSH session, restarting when a monitoring port fails return data:
```
marchersimon, line 7:
```suggestion
- Start an SSH session, restarting when a monitoring port fails to return data:
```
Is this correct?

pages/common/autossh.md
```diff
@@ -1,29 +1,29 @@
...
 `autossh -M {{monitor_port}} -L {{local_port}}:localhost:{{remote_port}} {{user}}@{{host}}`
 
-- Fork before executing ssh (runs in the background) and don't open a remote shell:
+- Fork `autossh` into the background before executing `ssh` and don't open a remote shell:
 
-`autossh -f -M {{monitor_port}} -N {{ssh_command}}`
+`autossh -f -M {{monitor_port}} -N "{{ssh_command}}"`
 
-- Run autossh in the background, with no monitoring port, instead relying on SSH keep-alives every 10 seconds to detect failure:
+- Run in the background, with no monitoring port, instead sending SSH keep-alive packets every 10 seconds to detect failure:
```
marchersimon, line 19:
```suggestion
- Run in the background, with no monitoring port, instead send SSH keep-alive packets every 10 seconds to detect failure:
```

pages/common/autossh.md
```diff
@@ -1,29 +1,29 @@
...
-- Run autossh in the background, with no monitoring port, no remote shell, exiting if the port forward fails:
+- Run in the background, with no monitoring port and no remote shell, exiting if the port forward fails:
 
 `autossh -f -M 0 -N -o "ServerAliveInterval 10" -o "ServerAliveCountMax 3" -o ExitOnForwardFailure=yes -L {{local_port}}:localhost:{{remote_port}} {{user}}@{{host}}`
 
-- Run autossh in the background with debug output logged to a file and ssh verbose output logged to a second file:
+- Run in the background, logging `autossh` debug output and `ssh` verbose output to files:
 
-`AUTOSSH_DEBUG=1 AUTOSSH_LOGFILE={{log_file}} autossh -f -M {{monitor_port}} -v -E {{ssh_log_file}} {{ssh_command}}`
+`AUTOSSH_DEBUG=1 AUTOSSH_LOGFILE={{path/to/autossh_log_file.log}} autossh -f -M {{monitor_port}} -v -E {{path/to/ssh_log_file.log}} {{ssh_command}}`
```
marchersimon, line 29: `VARIABLE=value` doesn't work in all shells. Is this a problem?

marchersimon (changes requested):

pages/common/autossh.md
bl-ue: Yes. I didn't notice this 🤷🏻

pages/common/autossh.md
bl-ue: Wow.  I almost checked the link, but for some reason I forgot.  I'm surprised that the current one doesn't even work!!

pages/common/autossh.md
bl-ue: Yep, definitely.   Good catch!

pages/common/autossh.md
bl-ue: It doesn't?  Do you know one in particular?

pages/common/autossh.md
bl-ue: Good spot, I'd say `..., and instead send ...`.  Can you update the suggestion?

pages/common/autossh.md
marchersimon: I was about to replace it completely, when I found the working version during my search. Is this actually the home page of `autossh`? Because if not I'd prefer <https://manned.org/autossh>.

pages/common/autossh.md
marchersimon: `Fish` and maybe `dash` (didn't test it).

sbrl: Hrm, that is awlward. However unfortunately according to the man page there isn't an alternate way of specifying these parameters.

marchersimon: Those are just environmental variables right? They do exist in all shells, just the syntax is different. (For example `set variable value` in fish). However I think that users of different shells should be aware of those differences.

pages/common/autossh.md
marchersimon: Updated

pages/common/autossh.md
bl-ue: Yes they are env variables @marchersimon, but they're temporary, i.e. they only last for the duration of the command that command comes after them.  To demonstrate:
```console
$ env | grep ABC
$ ABC=3 env | grep ABC
ABC=3
$ env | grep ABC
$
```
Versus a normal variable (which looks to me what `set variable value` in Fish is):
```console
$ export ABC=3
$ env | grep ABC
ABC=3
$ env | grep ABC
ABC=3
$ env | grep ABC
ABC=3
$
```

marchersimon: I didn't know that existed. I just tested it with Fish and if you write a command afterwards it works the same as in Bash.

bl-ue: Works in `dash` and `ksh` as well.  Not `xonsh`, `tcsh`, or `csh` though.  I say we should keep it 👍🏻

sbrl (approved):
Thanks, @bl-ue!

pages/common/autossh.md
marchersimon: Ping @bl-ue

marchersimon (approved):

Issue #5116, opened 2021-01-07
Detecting dead "more information" links


sbrl:
I was inspired by #5111, and I developed another of my bash one-liners to detect broken more information links in tldr pages. Broken links are defined as having a HTTP status code lower than 200, or greater than or equal to 400. Connection issues (e.g. DNS resolution failures, TLS certificate misconfigurations, etc) are also considered dead links

Here's the one-liner:

```bash
find . -type f -iname '*.md' -print0 | xargs -0 cat | awk '/> More information/ { match($0, /<(.*)>/, arr); print(arr[1]); }' | sort | uniq | shuf | xargs -n1 -I{} bash -c 'url="{}"; code="$(curl --user-agent "curl; bash; xargs; tldr-pages-bad-url-checker (+https://github.com/tldr-pages/tldr; implemented by @sbrl)" -sSL -o /dev/null -w "%{http_code}" --fail -I "${url}")"; echo "${code} ${url}" >&2; if [[ "${code}" -lt 200 ]] || [[ "${code}" -ge 400 ]]; then echo "${url}"; fi' >/tmp/bad-urls.txt;
```

`cd` to the root of a previously-cloned version of the tldr-pages repo, and then run the above command. Replace `/tmp/bad-urls.txt` with the path to the file you'd like to populate with the dead links. Sadly it doesn't tell you which page they are on, but this could be rectified with some refactoring.

Requirements:

 - xargs
 - curl
 - find
 - awk
 - bash

Note that while `xargs` is used to (indirectly) drive curl here, I haven't added a `-P "$(nproc)"` because otherwise you'll end up hitting GitHub's rate limit very quickly.

Anyway, here's the current list of dead links I've found with this:

```
https://harding.motd.ca/autossh
https://nixos.org/releases/nix/latest/manual/#sec-nix-collect-garbage
http://www.mysqlfanboy.com/mytop-3
https://nixos.org/releases/nix/latest/manual#sec-nix-build
https://taskwarrior.org/docs/timewarrior
https://docs.microsoft.com/windows-server/administration/windows-commands/reg-flags
https://clementsam.tech/glab/
https://flameshot.js.org
https://httpie.org
https://deluge-torrent.org
https://deemix.app
https://mailutils.org/manual/html_chapter/Programs.html#frm-and-from
https://docs.blender.org/manual/en/latest/render/workflows/command_line.html
https://www.infradead.org/openconnect/manual.html
https://www.clementine-player.org
https://crates.io/
[...]

PR #5340, opened 2021-03-03
tcpdump: add page

tcpdump-examples:
- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #5724, opened 2021-04-09
*: replace dead more information links

marchersimon:
For #5116


marchersimon:
Seems like linter isn't ready for `<N/A>` yet.

pages/common/deemix.md
```diff
@@ -2,7 +2,7 @@
 
 > A barebone deezer downloader library built from the ashes of Deezloader Remix.
 > It can be used as a standalone CLI app or implemented in an UI using the API.
-> More information: <https://deemix.app>.
+> More information: <N/A>.
```
bl-ue, line 5: Wow, nothing?

pages/common/redshift.md
```diff
@@ -1,7 +1,7 @@
 # redshift
 
 > Adjust the color temperature of your screen according to your surroundings.
-> More information: <https://jonls.dk/redshift>.
+> More information: <http://jonls.dk/redshift>.
```
bl-ue, line 4: This is certainly odd.

pages/common/timew.md
```diff
@@ -1,7 +1,7 @@
 # timew
 
 > A time tracking tool used to measure the duration of activities.
-> More information: <https://taskwarrior.org/docs/timewarrior>.
+> More information: <https://manned.org/timew>.
```
bl-ue, line 4: Maybe we could use this?
```suggestion
> More information: <https://timewarrior.net/docs>.
```

pages/linux/postfix.md
```diff
@@ -2,7 +2,7 @@
 
 > Postfix mail transfer agent (MTA) control program.
 > See also `dovecot`, a mail delivery agent (MDA) that integrates with Postfix.
-> More information: <http://postfix.org>.
+> More information: <https://manned.org/postfix>.
```
bl-ue, line 5: Would this work?
```suggestion
> More information: <https://www.postfix.org>.
```

pages/common/mytop.md
```diff
@@ -1,7 +1,7 @@
 # mytop
 
 > Display MySQL server performance info like `top`.
-> More information: <http://www.mysqlfanboy.com/mytop-3>.
+> More information: <http://jeremy.zawodny.com/mysql/mytop/mytop.html>.
```
bl-ue, line 4: Though ugly-looking, this link appears to be a good one.  From the page:

>mytop was developed and is maintained by Jeremy D. Zawodny (Jeremy@Zawodny.com)

pages/common/multitail.md
```diff
@@ -1,7 +1,7 @@
 # multitail
 
 > Extension of tail.
-> More information: <https://www.vanheusden.com/multitail/examples.php>.
+> More information: <https://manned.org/multitail>.
```
bl-ue, line 4: Strange.  The man page has this very URL, but, besides being very insecure, is a 404.

pages/common/dexdump.md
```diff
@@ -1,7 +1,7 @@
 # dexdump
 
 > Display information about Android DEX files.
-> More information: <https://manpages.ubuntu.com/manpages/latest/en/man1/dexdump.1.html>.
+> More information: <https://manned.org/dexdump>.
```
bl-ue, line 4: OMG this is hilarious.  From the Ubuntu man page:
>### NAME
>```
>dexdump - Dex Tool
>```
>### DESCRIPTION
>```
>debian/out/dexdump:  error  while  loading shared libraries: libsigchain.so.0: cannot open
>shared object file: No such file or directory
>```
The manned.org one is okay though.

bl-ue (commented):
Left some mere remarks.

pages/common/deemix.md
marchersimon: There is <https://pypi.org/project/deemix/>, but there isn't anything on how to use it on there.

pages/common/timew.md
marchersimon: Yeah, sure. No idea how I didn't find that

pages/linux/postfix.md
marchersimon: Sure. I didn't think about trying that

pages/common/multitail.md
marchersimon: Seems like <vanheusden.com> was repurposed. It seems to be something historical now.

pages/common/mytop.md
marchersimon: Do you mean <http://www.mysqlfanboy.com/mytop-3>? It neither works for me on linux, nor on android.

pages/common/deemix.md
bl-ue: Yeah, it's really odd.  Here's the old one, I guess: https://notabug.org/RemixDevs/DeezloaderRemix

pages/common/mytop.md
bl-ue: No, I'm referring to the new one.  I was adding a note for future reference, in case someone thought that was some random blog post (because it looks like one).

pages/common/deemix.md
marchersimon: What does this link have to do with `deemix`?

sbrl: Hrm. I can't remember which issue it was we are discussing a not-applicable syntax, but this is incompatible with CommonMark as that is the autolinking syntax. Anyway, I'd suggest that <https://pypi.org/project/deemix/> is probably the best we've got to go with here - it's better than nothing at least.

pages/common/redshift.md
sbrl: I tweeted the developer: <https://twitter.com/SBRLabs/status/1381266346562965504?s=20>

pages/common/sqlmap.md
```diff
@@ -1,7 +1,7 @@
 # sqlmap
 
 > Detect and exploit SQL injection flaws.
-> More information: <https://sqlmap.org>.
+> More information: <http://sqlmap.org>.
```
sbrl, line 4: I've opened an issue here about this: https://github.com/sqlmapproject/sqlmap/issues/4638

pages/linux/postfix.md
```diff
@@ -2,7 +2,7 @@
 
 > Postfix mail transfer agent (MTA) control program.
 > See also `dovecot`, a mail delivery agent (MDA) that integrates with Postfix.
-> More information: <http://postfix.org>.
+> More information: <https://www.postfix.org>.
```
sbrl, line 5: This link doesn't work for me, though I suspect this is a transitory issue.

marchersimon: That's strange. @bl-ue suggested this just 2 days ago (https://github.com/tldr-pages/tldr/pull/5724#discussion_r610958600) and I'm sure the link worked. But now it doesn't work for me either.

pages/common/deemix.md
marchersimon: This site is only about the python package and doesn't include any information about how to use `deemix`. Do you think linking to a  site, where you dig a while just to find out there's nothing useful, is better than no link at all?

pages/common/redshift.md
bl-ue: Good job—hopefully they'll get back. Doesn't matter much though ;)

pages/linux/postfix.md
bl-ue: Yeah, it doesn't work for me either 🤷🏻 `http://` does tho: http://www.postfix.org/

marchersimon: I'll change it back then

pages/common/deemix.md
```diff
@@ -2,7 +2,7 @@
 
 > A barebone deezer downloader library built from the ashes of Deezloader Remix.
 > It can be used as a standalone CLI app or implemented in an UI using the API.
```
bl-ue, line 4:
```suggestion
> It can be used as a standalone CLI app or implemented in a UI using the API.
```

pages/common/redshift.md
sbrl: Yeah - haven't heard anything yet. This should not block this PR though.

pages/common/deemix.md
sbrl: Ah yeah, fair point. Let's do `More information: N/A.` then.

sbrl (approved):
Thanks, @marchersimon!

navarroaxel (approved):

sbrl:
Uh oh, @owenvoke I don't think the `pushd` and `popd` pages were actually modified in this PR. Bug?

bl-ue:
I think it's checking every single page, and I don't think it shouldn't.  For example, when we update the linter, the PRs that do that don't touch any pages usually, but we still need tldr-bot to check all of the pages using that new version of the linter and report them there.

sbrl:
True @bl-ue, but in this case we have 2 pages that are slightly different. Interestingly we don't actually even have a copy of pushd` in `linux`:

image

Perhaps the simplest solution would be to bump the similarity requirement to 75% / 80%?

bl-ue:
I like that.  To be honest, I'm not even sure what the need is for a page duplication check. 🤷🏻

pages/common/sqlmap.md
bl-ue: Fixed! https://github.com/sqlmapproject/sqlmap/issues/4638#issuecomment-824611258
```suggestion
> More information: <https://sqlmap.org>.
```

sbrl:
It's in case someone opens a PR for a page that already exists maybe, or to avoid duplication of identical pages across platforms? Historically this has been an issue.

bl-ue:
Yes, it's a problem.  The late #5340 comes to mind, but the page duplication never works for cases like that.  When are two pages for the same command, by two different people, anything alike?

pages/common/deemix.md
CleanMachine1: There is this https://deezloader.net/deemix/

But it doesn't look very official and maybe breaking copyright and everything under the sun, it seems to have cracked in some way premium APKs however this may be the real site

sbrl: Hrm. In the interests of getting this PR merged, I suggest we remove the more information line completely from this page for now, and then when the linter is updated we can add the explicit `N/A` later.
```suggestion
```

pages/common/deemix.md
```diff
@@ -1,8 +1,7 @@
 # deemix
 
 > A barebone deezer downloader library built from the ashes of Deezloader Remix.
-> It can be used as a standalone CLI app or implemented in an UI using the API.
-> More information: <https://deemix.app>.
+> It can be used as a standalone CLI app or implemented in a UI using the API.
```
marchersimon, line 4:
```suggestion
> It can be used as a standalone CLI app or implemented in an UI using the API.
> More informattion: <https://deemix.app>
```
This one works again.

marchersimon: Now we only make the `an` → `a` change here. Should we revert that too?

marchersimon: [This one](https://github.com/tldr-pages/tldr/pull/5724#discussion_r657193280) also works again

pages/common/mytop.md
bl-ue: Odd, I guess so. Hopefully it won't go down again 😬

marchersimon: I think we should keep the second one in this case, since it also seems to be more official.

pages/common/deemix.md
sbrl: Looks good to me! Let's apply this change.

pages/common/mytop.md
sbrl: Looks like the existing link works again.

marchersimon: @sbrl as I exlained above, I think we should still keep the new one since it seems to be more official.

sbrl (approved):
Are we ready to merge?

marchersimon:
I'd say so, yes

CleanMachine1 (approved):

PR #5733, opened 2021-04-11
cmus: update example description wording

thomas-m-d:
...to address pet peeve where cmus exits to music directory, not directory command was run from. i also edited the first example to better fit with my new one.

### DIFF
--- a/pages/linux/cmus.md
+++ b/pages/linux/cmus.md
@@ -3,7 +3,7 @@
 > Commandline Music Player.
 > Use arrow keys to navigate, `<enter/return>` to select, and numbers 1-8 switch between different views.
 
-- Open cmus into specified directory (note this will become your new working directory):
+- Open cmus into the specified directory (this will become your new working directory):
 
 `cmus {{path/to/directory}}`
 
--- a/pages/linux/cmus.md
+++ b/pages/linux/cmus.md
@@ -3,13 +3,9 @@
 > Commandline Music Player.
 > Use arrow keys to navigate, `<enter/return>` to select, and numbers 1-8 switch between different views.
 
-- Open cmus into specified directory (note that you will exit cmus to this directory):
+- Open cmus into specified directory (note this will become your new working directory):
 
 `cmus {{path/to/directory}}`
-
-- Open cmus into specified directory then return to original directory on exit:
-
-`cd {{path/to/directory}} && cmus && cd -`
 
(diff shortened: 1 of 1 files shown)


pages/linux/cmus.md
```diff
@@ -3,10 +3,14 @@
...
 > Use arrow keys to navigate, `<enter/return>` to select, and numbers 1-8 switch between different views.
 
-- Open cmus from specified directory:
+- Open cmus into specified directory (note that you will exit cmus to this directory):
 
 `cmus {{path/to/directory}}`
 
+- Open cmus into specified directory then return to original directory on exit:
+
+`cd {{path/to/directory}} && cmus && cd -`
```
navarroaxel, line 12: :thinking: this is not an example about the command itself, it's more a bash feature.

thomas-m-d: this is true. if this doesn't fit within tldr, i understand completely.

my case for adding this example is that cmus doesn't seem to have this functionality natively and it was a pet peeve of mine when i was using cmus. to my mind this is how `cmus {{path/to/directory}}` should work, so having an example that not only gives that functionality but lets people know it's not the default functionality could be helpful.

i'd love to hear what you and others more experienced with tldr think @navarroaxel.

marchersimon: I'd say this is something which would fit into the command description. For example:
> \> When opening `cmus` to a specific directory, it will change the working directory to that directory.

thomas-m-d: that makes perfect sense. i can make the sort of change later today.

thomas-m-d: sorry for the delay, but i just committed that change.

pages/linux/cmus.md
```diff
@@ -3,7 +3,7 @@
 > Commandline Music Player.
 > Use arrow keys to navigate, `<enter/return>` to select, and numbers 1-8 switch between different views.
 
-- Open cmus from specified directory:
+- Open cmus into specified directory (note this will become your new working directory):
```
marchersimon, line 6:
```suggestion
- Open cmus into the specified directory (this will become your new working directory):
```
Just a small suggestion

marchersimon (approved):

bl-ue (approved):
So, this PR degenerated into simply a wording update. 🙂

navarroaxel (approved):

PR #5734, opened 2021-04-11
mocp: add page

thomas-m-d:
added page for music on console (moc) terminal audio player. i have added 9 examples, as i think all are useful for a beginner. if the maintainers require it be 8 examples please let me know which you'd recommend i delete.

### DIFF
--- a/pages/linux/mocp.md
+++ b/pages/linux/mocp.md
@@ -20,7 +20,7 @@
 
 `mocp --append {{path/to/directory}}`
 
-- Clear the play queue while MOC is in background:
+- Clear the play queue while MOC is in the background:
 
 `mocp --clear`
 
--- a/pages/linux/mocp.md
+++ b/pages/linux/mocp.md
@@ -11,9 +11,6 @@
 
 `mocp {{path/to/directory}}`
 
-- Launch MOC in background:
-
-`mocp &`
 
 - Add individual song to play queue while MOC is in background:
 
--- a/pages/linux/mocp.md
+++ b/pages/linux/mocp.md
@@ -1,7 +1,7 @@
 # mocp
 
 > Music on Console (MOC) audio player.
-> More information: <http://moc.daper.net/>.
+> More information: <https://moc.daper.net/>.
 
 - Launch MOC terminal UI:
 
--- a/pages/linux/mocp.md
+++ b/pages/linux/mocp.md
@@ -7,7 +7,7 @@
 
 `mocp`
 
-- Launch MOC terminal UI in specific directory:
+- Launch the MOC terminal UI in a specific directory:
 
 `mocp {{path/to/directory}}`
 
--- a/pages/linux/mocp.md
+++ b/pages/linux/mocp.md
@@ -32,6 +32,6 @@
 
 `mocp --stop`
 
-- Close MOC while it is in backgroud:
+- Close MOC while it is in the backgroud:
 
 `mocp --exit`
--- a/pages/linux/mocp.md
+++ b/pages/linux/mocp.md
@@ -3,7 +3,7 @@
 > Music on Console (MOC) audio player.
 > More information: <https://moc.daper.net/>.
 
-- Launch MOC terminal UI:
+- Launch the MOC terminal UI:
 
 `mocp`
 
--- a/pages/linux/mocp.md
+++ b/pages/linux/mocp.md
@@ -23,7 +23,7 @@
 
 `mocp --clear`
 
-- Play first song in the play queue while MOC is in the background:
+- Play the first song in the play queue while MOC is in the background:
 
 `mocp --play`
 
--- a/pages/linux/mocp.md
+++ b/pages/linux/mocp.md
@@ -12,7 +12,7 @@
 `mocp {{path/to/directory}}`
 
 
-- Add individual song to play queue while MOC is in background:
+- Add a specific song to the play queue while MOC is in the background:
 
 `mocp --enqueue {{path/to/audio_file}}`
 
--- a/pages/linux/mocp.md
+++ b/pages/linux/mocp.md
@@ -11,7 +11,6 @@
 
 `mocp {{path/to/directory}}`
 
-
 - Add a specific song to the play queue while MOC is in the background:
 
 `mocp --enqueue {{path/to/audio_file}}`
--- a/pages/linux/mocp.md
+++ b/pages/linux/mocp.md
@@ -1,6 +1,6 @@
 # mocp
 
-> Music on Console (MOC) audio player
+> Music on Console (MOC) audio player.
 > More information: <http://moc.daper.net/>.
 
 - Launch MOC terminal UI:
--- a/pages/linux/mocp.md
+++ b/pages/linux/mocp.md
@@ -16,7 +16,7 @@
 
 `mocp --enqueue {{path/to/audio_file}}`
 
-- Add songs recursively to play queue while MOC is in background:
+- Add songs recursively to the play queue while MOC is in the background:
 
 `mocp --append {{path/to/directory}}`
 
--- a/pages/linux/mocp.md
+++ b/pages/linux/mocp.md
@@ -24,7 +24,7 @@
 
 `mocp --clear`
 
-- Play first song in play queue while MOC is in background:
+- Play first song in the play queue while MOC is in the background:
 
 `mocp --play`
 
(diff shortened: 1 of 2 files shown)


pages/linux/mocp.md
```diff
@@ -0,0 +1,40 @@
+# mocp
+
+> Music on Console (MOC) audio player.
+> More information: <http://moc.daper.net/>.
```
marchersimon, line 4:
```suggestion
> More information: <https://moc.daper.net/>.
```

pages/linux/mocp.md
```diff
@@ -0,0 +1,40 @@
...
+
+- Launch MOC terminal UI:
```
marchersimon, line 6:
```suggestion
- Launch the MOC terminal UI:
```

pages/linux/mocp.md
```diff
@@ -0,0 +1,40 @@
...
+
+`mocp`
+
+- Launch MOC terminal UI in specific directory:
```
marchersimon, line 10:
```suggestion
- Launch the MOC terminal UI in a specific directory:
```

pages/linux/mocp.md
```diff
@@ -0,0 +1,40 @@
...
+
+`mocp {{path/to/directory}}`
+
+- Launch MOC in background:
+
+`mocp &`
```
marchersimon, line 16: This isn't really `mocp`-specific, so we usually don't list it in the tldr pages. Also, removing it would solve the problem, that a page shouldn't have more than 8 examples.

pages/linux/mocp.md
```diff
@@ -0,0 +1,40 @@
...
+
+- Add individual song to play queue while MOC is in background:
```
marchersimon, line 18:
```suggestion
- Add a specific song to the play queue while MOC is in the background:
```

pages/linux/mocp.md
```diff
@@ -0,0 +1,40 @@
...
+
+`mocp --enqueue {{path/to/audio_file}}`
+
+- Add songs recursively to play queue while MOC is in background:
```
marchersimon, line 22:
```suggestion
- Add songs recursively to the play queue while MOC is in the background:
```

pages/linux/mocp.md
```diff
@@ -0,0 +1,40 @@
...
+
+`mocp --append {{path/to/directory}}`
+
+- Clear the play queue while MOC is in background:
```
marchersimon, line 26:
```suggestion
- Clear the play queue while MOC is in the background:
```

pages/linux/mocp.md
```diff
@@ -0,0 +1,40 @@
...
+
+`mocp --clear`
+
+- Play first song in play queue while MOC is in background:
```
marchersimon, line 30:
```suggestion
- Play first song in the play queue while MOC is in the background:
```

pages/linux/mocp.md
```diff
@@ -0,0 +1,40 @@
...
+
+`mocp --play`
+
+- Stop any playing songs while MOC is in background:
```
marchersimon, line 34:
```suggestion
- Stop any playing songs while MOC is in the background:
```

pages/linux/mocp.md
```diff
@@ -0,0 +1,40 @@
...
+
+`mocp --stop`
+
+- Close MOC while it is in backgroud:
```
marchersimon, line 38:
```suggestion
- Close MOC while it is in the backgroud:
```

marchersimon (changes requested):
Thanks @thomas-m-d. I left a few suggestions below. To summarize:
* `command &` isn't command specific, so we don't list them in our pages
* Example descriptions should be full sentences
* We always prefer `https` over `http`

bl-ue:
@thomas-m-d in the future, when applying suggestions, go into the Files Changed tab and hit 'Add suggestion to bacth' on each suggestion and then commit them all at once, so that it generates one commit rather than multiple, which in this case caused @tldr-bot to throw up a lot of copies of the same error.

thomas-m-d:
> Thanks @thomas-m-d. I left a few suggestions below. To summarize:
> 
>     * `command &` isn't command specific, so we don't list them in our pages
> 
>     * Example descriptions should be full sentences
> 
>     * We always prefer `https` over `http`

that all makes sense. also, i noticed the commit suggestion option in this pr but just read your comment about batching them from my other pr. sorry for the mass of commits here, but now i know how to do it better! thanks :)

pages/linux/mocp.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Add songs recursively to the play queue while MOC is in the background:
+
+`mocp --append {{path/to/directory}}`
+
+- Clear the play queue while MOC is in the background:
+
+`mocp --clear`
+
+- Play first song in the play queue while MOC is in the background:
```
marchersimon, line 26:
```suggestion
- Play the first song in the play queue while MOC is in the background:
```
Almost missed.

marchersimon (approved):

pages/linux/mocp.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Launch the MOC terminal UI:
+
+`mocp`
+
+- Launch the MOC terminal UI in a specific directory:
+
+`mocp {{path/to/directory}}`
+
+- Add a specific song to the play queue while MOC is in the background:
```
navarroaxel, line 14: Hi @tom-on-the-internet.
I don't understand if this command starts `moc` in the background. is the terminal UI only launched in foreground when is executed without arguments?

marchersimon: @navarroaxel I think you tagged the wrong person here.

To your question: You're right. In order to use arguments like `--stop` or `--play`, there has to be a running `MOC`server either in the background or in another shell. If you run any of those if there's no running server you get an error.

marchersimon:
@thomas-m-d two more things I noticed:
* The homepage of MOC doesn't really provide any information on how to use the command. I think <https://manned.org/mocp> would be more helpful here.
* A command to start `mocp` in the background would be helpful, but instead of using `mocp &`, you could use `mocp --server`. However, you'd have to remove one of the examples, to not exceed 8. (I'd suggest `--clear`, since it's not as important as the others.)

thomas-m-d:
> @thomas-m-d two more things I noticed:
> 
>     * The homepage of MOC doesn't really provide any information on how to use the command. I think https://manned.org/mocp would be more helpful here.
> 
>     * A command to start `mocp` in the background would be helpful, but instead of using `mocp &`, you could use `mocp --server`. However, you'd have to remove one of the examples, to not exceed 8. (I'd suggest `--clear`, since it's not as important as the others.)

the first change makes sense to me.

the second change makes sense too, but it also looks like a good way of addressing @navarroaxel's question about how the tui vs background use of moc works. that was my goal with including `mocp &`, so i'd love to hear their feedback to see if adding a `mocp --server` example between the first two examples (which launch the tui) and the remaining examples (which work with moc in the background) do enough to address their concern.

i'll hold off on making any commits until i hear back from @navarroaxel.

navarroaxel:
@thomas-m-d I think the `--server` example is needed, you can merge the `--play` and `--stop` examples.

thomas-m-d:
changes have been committed. let me know how they look.

pages/linux/mocp.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Clear the play queue while MOC is in the background:
+
+`mocp --clear`
+
+- Play or stop the currently queued song while MOC is in the background:
+
+`mocp --{{play|stop}}`
+
+- Stop the MOC server while it is in the backgroud:
```
bl-ue, line 34: I like the shorter way this sounds, but if anyone disagrees we can skip it.
```suggestion
- Stop the MOC server while it's in the backgroud:
```

thomas-m-d: shorter works for me. i'll give it a bit in case anyone objects. otherwise i'll make the suggestion.

marchersimon (approved):

bl-ue (approved):

navarroaxel (approved):

pages/linux/mocp.md
bl-ue: Oops @navarroaxel did you see this convo?

navarroaxel: Oh, I saw your approve and I didn't check this. Can you add this change to #5748 ?

navarroaxel: @bl-ue Oh, I saw your approve and I didn't check this. Can you add this change to #5748 ?

PR #5735, opened 2021-04-11
nautilus: add page

thomas-m-d:
added page for nautilus, gnome's default file explorer which is also known as gnome files.


pages/linux/nautilus.md
```diff
@@ -0,0 +1,13 @@
+# nautilus
+
+> Default file explorer for Gnome desktop environment.
+> Also known as Gnome Files.
+> More information: <https://github.com/GNOME/nautilus>.
```
marchersimon, line 5:
```suggestion
> More information: <https://manned.org/nautilus>.
```
Altough the Github repo is more official, it doesn't give any information on how to use `nautilus` in the command line. A man page is problably more helpful if you want to find out more about the command.

pages/linux/nautilus.md
```diff
@@ -0,0 +1,13 @@
...
+
+- Launch Nautilus as current user:
```
marchersimon, line 7:
```suggestion
- Launch Nautilus:
```
I'd say that it's obvious that nautilus is launched as the current user.

marchersimon (changes requested):
Thanks for the page @thomas-m-d

pages/linux/nautilus.md
thomas-m-d: good to know. i didn't know if the 'more info' was for more info on the full program, or just the terminal command. sounds like it's the terminal command, so i'll keep that in mind going forward.

pages/linux/nautilus.md
thomas-m-d: makes sense to me. i added it to mirror the second example, but it should be pretty obvious for this example.

pages/linux/nautilus.md
marchersimon: It depends. If the official page covers all of the command line options it should be preffered. Otherwise we use `manned.org` or a distro specific man page, if the command is distro specific.

thomas-m-d (commented):
looks good to me.

thomas-m-d:
> Thanks for the page @thomas-m-d

i'm still learning github, so let me know if there's anything i need to do to resolve/approve your requested changes.

marchersimon:
To apply my changes, you can click `Commit suggestion` under each suggestion (you have to unresolve the conversation before that). You can also go to the `Files changed` tab, click `Add suggestion to batch` for each suggestion and then click `Apply suggestions from batch`. In that way you can commit all suggestions in one commit, which would be beneficial if there are many suggestions.

marchersimon (approved):
LGTM

pages/linux/nautilus.md
```diff
@@ -0,0 +1,13 @@
...
+> Also known as Gnome Files.
+> More information: <https://manned.org/nautilus>.
+
+- Launch Nautilus:
+
+`nautilus`
+
+- Launch Nautilus as root user:
+
+`sudo nautilus`
```
bl-ue, line 13: I'm not sure about this example, because it seems that sudo launches a command as root.  @sbrl

marchersimon: It shows, that Nautilus can be launched as root, which can be good to know. This isn't possible for Dolphin for example.

bl-ue: 👍🏻

pages/linux/nautilus.md
```diff
@@ -0,0 +1,13 @@
+# nautilus
+
+> Default file explorer for Gnome desktop environment.
```
bl-ue, line 3:
```suggestion
> Default file explorer for GNOME desktop environment.
```

pages/linux/nautilus.md
```diff
@@ -0,0 +1,13 @@
...
+> Also known as Gnome Files.
```
bl-ue, line 4:
```suggestion
> Also known as GNOME Files.
```

bl-ue (commented):
Why isn't 'Gnome' capitalized?

thomas-m-d:
> Why isn't 'Gnome' capitalized?

good point. capitalizations committed.

sbrl:
Hey there, @thomas-m-d! Welcome to tldr pages. You can also tell Nautilus which directory to show like this:

```bash
nautilus path/to/file_or_directory
```

....so maybe we could include this as an example at the bottom of the page? What about adding this?

```markdown
- Display the contents of a given directory or the location of a given file:

`nautilus {{path/to/file_or_directory}}`
```

thomas-m-d:
> Hey there, @thomas-m-d! Welcome to tldr pages. You can also tell Nautilus which directory to show like this:
> 
> ```shell
> nautilus path/to/file_or_directory
> ```
> 
> ....so maybe we could include this as an example at the bottom of the page? What about adding this?
> 
> ```gfm
> - Display the contents of a given directory or the location of a given file:
> 
> `nautilus {{path/to/file_or_directory}}`
> ```
> 
> **Edit:** With respect to your question [here](https://github.com/tldr-pages/tldr/pull/5646#discussion_r611565072) @marchersimon, I think this page still has value because the behaviour of Nautilus changes depending on the arguments you call it with.

i've just committed that example in. i've changed the description text to be consistent with the other 2 examples, but feel free to edit it as you see fit.

as for whether this is an entry that belongs in tldr, i'll defer to the maintainers. honestly, i saw nautilus and steam (and to a lesser degree my cmus edit) as low hanging fruit i might as well address while i'm creating a mocp entry. they're popular programs so i figured there's probably a reason they aren't included already (they're gui-based programs, their options are too simple, etc.). but the effort was low enough that i figured it was worth doing in case they do fit within tldr.

anyways, that's where i'm coming from, so i understand completely if this or my steam pr don't get accepted.

pages/linux/nautilus.md
```diff
@@ -0,0 +1,17 @@
...
+
+- Launch Nautilus:
+
+`nautilus`
+
+- Launch Nautilus as root user:
+
+`sudo nautilus`
+
+- Launch Nautilus and display a given directory:
```
marchersimon, line 15:
```suggestion
- Launch Nautilus and display a specific directory:
```
This is the syntax we typaically use.

pages/linux/nautilus.md
```diff
@@ -0,0 +1,17 @@
...
+
+`nautilus {{path/to/file_or_directory}}`
```
marchersimon, line 17: Actually I'd split those two in two different examples. Because when you specify a directory, nautilus opens at that location, but when you specify a file, nautilus opens the parent folder and just highlights the file.

marchersimon (commented):
I left some suggestions below. Regarding your pages: IMO, even a page for a command with very few command line options can be useful to show, that it has only few options. I've often looked up some command, just to find out there aren't any options.

navarroaxel (commented):
I think you can add more examples like:
```
- Launch with a specific file or directory selected:

`nautilus --select {{path/to/file_or_directory}}`

- Launch in a separated window:

`nautilus --new-window`

- Close all instances:

`nautilus --quit`

- Display help:

`nautilus --help`
```

marchersimon:
@navarroaxel I'm not sure about the first one, since `nautilus --select file` does the same as `nautilus file`. `nautilus --select directory` can be useful though.

navarroaxel:
> @navarroaxel I'm not sure about the first one, since `nautilus --select file` does the same as `nautilus file`. `nautilus --select directory` can be useful though.

@marchersimon :thinking: Dolphin has the same argument `--select`.
```
Select specified URI in parent folder.
```
The man page says `URI` instead of `path`, Maybe `--select` is a safer way to handle this, or can hace more URIs than `file:///`

marchersimon:
How about we add `--select` as you suggested and use `nautilus {{path/to/directory}}`, instead of `nautilus {{path/to/file_or_directory}}`?

thomas-m-d:
> How about we add `--select` as you suggested and use `nautilus {{path/to/directory}}`, instead of `nautilus {{path/to/file_or_directory}}`?

i'll make this change later today.

thomas-m-d:
changes made. thanks for the suggestions, @navarroaxel. it's a much better page now!

navarroaxel (approved):

einverne (approved):
Looks good to me.

pages/linux/nautilus.md
```diff
@@ -0,0 +1,33 @@
...
+
+- Launch Nautilus with a specific file or directory selected:
+
+`nautilus --select {{path/to/file_or_directory}}`
+
+- Launch Nautilus in a separated window:
+
+`nautilus --new-window`
+
+- Close all instances:
```
bl-ue, line 27:
```suggestion
- Close all Nautilus instances:
```

bl-ue (commented):
One minor comment before merging 😉

thomas-m-d:
> One minor comment before merging wink

comment committed :)

bl-ue (approved):
Nice!  Thank you @thomas-m-d! 🚀

thomas-m-d:
> Nice! Thank you @thomas-m-d! rocket

thank you yourself @bl-ue (and @navarroaxel & @marchersimon).

PR #5740, opened 2021-04-12
input: add page

marchersimon:
There isn't really any documentation for this command. There's only a list of keycodes [here](https://developer.android.com/reference/android/view/KeyEvent.html). What I used to get the infos was <https://stackoverflow.com/questions/7789826/adb-shell-input-events>.

For #4934

### DIFF
--- a/pages/android/input.md
+++ b/pages/android/input.md
@@ -1,0 +1,24 @@
+# input
+
+> Send event codes or touchscreen gestures to an Android device.
+> Event codes can be inputting single characters, ending calls, volume up, etc.
+
+- Send an event code for a single character to an Android device:
+
+`input keyevent {{event_code}}`
+
+- Send a text to an Android device (`%s` represents spaces):
+
+`input text "{{text}}"`
+
+- Send a single tap to an Android device:
+
+`input tap {{x_pos}} {{y_pos}}`
+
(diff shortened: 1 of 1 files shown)


bl-ue (approved):

marchersimon:
Is <https://developer.android.com/reference/android/view/KeyEvent.html#constants_1> okay for the more information link. It doesn't provide anything on `input` itself, but it has a list of all event codes.

navarroaxel:
@marchersimon 

> Is https://developer.android.com/reference/android/view/KeyEvent.html#constants_1 okay for the more information link? It doesn't provide anything on `input` itself, but it has a list of all event codes.

@marchersimon but this are the key constants not the command itself :thinking:.

marchersimon:
That's all I could find. There was a thread about a documentation for the command [here](https://stackoverflow.com/questions/7789826/adb-shell-input-events), but they all just pointed to the link for the key constants.

bl-ue:
To me, it seems fine.  I wish there was a good page for it but ¯\_(ツ)_/¯

bl-ue (approved):
To me, it seems fine.  I wish there was a good page for it but ¯\_(ツ)_/¯

navarroaxel (approved):

Issue #5110, opened 2021-01-07
Rename `master` branch to `main`


schneiderl:
According to our [Inclusive Language](https://github.com/tldr-pages/tldr/blob/0c9f1209d1e7192b76d992d7f947e76d190b6ed9/CONTRIBUTING.md#inclusive-language) section on the Contributing Guide, we should always try to use inclusive language. 
According to the Guide, 

> "primary"/"secondary" instead of "master"/"slave"

I believe that if that is in our Contributing Guide, we should follow it in our main repositories at least. That includes how we name our branches. 
Our "main" branches are still named "master" in all the repositories from the organization.
![image](https://user-images.githubusercontent.com/11352223/103914733-0a93ea00-50e9-11eb-854b-f9ad35bdb24f.png)

Should we think of renaming this? 

Also, if we do rename, I would suggest "main" instead of "primary" following [what is being done by GitHub](https://github.com/github/renaming).

sbrl:
Great point. Yeah, I think `main` is fast becoming the new standard. Is there any word on whether Git itself will be changing its default too, and if so in what version?

@owenvoke, will this result in an impact on any of our GitHub actions or automation?

@agnivade: would this have an impact on the tldr-bot at all?

schneiderl:
The only thing I could find on Git is that they are really cautious about renamings of the sort. 

See: [this](https://git-scm.com/book/en/v2/Git-Branching-Branch-Management) under **Changing the master branch name** section.

Either way, this section only shows how one should proceed while changing this. I could not find anything about making "main" standard.

bl-ue:
@sbrl @schneiderl See the Git project's official [statement](https://sfconservancy.org/news/2020/jun/23/gitbranchname/) and [code change](https://lore.kernel.org/git/pull.656.v4.git.1593009996.gitgitgadget@gmail.com/). I found that at GH's [Blog Post](https://github.blog/changelog/2020-07-22-github-actions-better-support-for-alternative-default-branch-names/) -> [Repo](https://github.com/github/renaming/).

agnivade:
Don't think so it's going to cause issues with tldr-bot.

owenvoke:
So, the only required changes for GitHub Actions would be a couple of references to `master` in the `if` blocks. We do however reference `master` in quite a few other places.

I've created a branch to show the differences that would be required: https://github.com/tldr-pages/tldr/compare/master...owenvoke:feature/main

[...]

PR #5744, opened 2021-04-13
tracepath: add page

peterbabic:

### DIFF
--- a/pages/linux/tracepath.md
+++ b/pages/linux/tracepath.md
@@ -1,0 +1,28 @@
+# tracepath
+
+> Trace path to a network host discovering MTU along this path.
+> More information: <https://manned.org/tracepath>.
+
+- Trace the path to a host:
+
+`tracepath {{host}}`
+
+- Print numerical IP addresses only:
+
+`tracepath -n {{host}}`
+
+- Print both hostnames and numerical IP addresses:
+
+`tracepath -b {{host}}`
+
+- Specify maximum TTL (number of hops):
+
+`tracepath -m {{max_hops}} {{host}}`
+
+- Specify the initial destination port:
+
+`tracepath -p {{destination_port}} {{host}} `
+
+- Specify the initial packet length:
+
+`tracepath -l {{packet_length}} {{host}}`
--- a/pages/linux/tracepath.md
+++ b/pages/linux/tracepath.md
@@ -19,7 +19,7 @@
 
 `tracepath -m {{max_hops}} {{host}}`
 
-- Specify the initial destination port:
+- Specify the initial destination port, preferably 33434:
 
 `tracepath -p {{destination_port}} {{host}} `
 
--- a/pages/linux/tracepath.md
+++ b/pages/linux/tracepath.md
@@ -5,7 +5,7 @@
 
 - A preferred way to trace the path to a host:
 
-`tracepath -p 33434 {{host}}`
+`tracepath -p {{33434}} {{host}}`
 
 - Specify the initial destination port, useful with non-standard firewall settings:
 
--- a/pages/linux/tracepath.md
+++ b/pages/linux/tracepath.md
@@ -3,9 +3,13 @@
 > Trace the path to a network host discovering MTU along this path.
 > More information: <https://manned.org/tracepath>.
 
-- Trace the path to a host:
+- A preferred way to trace the path to a host:
 
-`tracepath {{host}}`
+`tracepath -p 33434 {{host}}`
+
+- Specify the initial destination port, useful with non-standard firewall settings:
+
+`tracepath -p {{destination_port}} {{host}} `
 
 - Print numerical IP addresses only:
 
@@ -18,10 +22,6 @@
 - Specify a maximum TTL (number of hops):
 
 `tracepath -m {{max_hops}} {{host}}`
-
-- Specify the initial destination port, preferably 33434:
-
-`tracepath -p {{destination_port}} {{host}} `
 
 - Specify the initial packet length (defaults to 65535 for IPv4 and 128000 for IPv6):
 
--- a/pages/linux/tracepath.md
+++ b/pages/linux/tracepath.md
@@ -1,6 +1,6 @@
 # tracepath
 
-> Trace path to a network host discovering MTU along this path.
+> Trace the path to a network host discovering MTU along this path.
 > More information: <https://manned.org/tracepath>.
 
 - Trace the path to a host:
@@ -15,7 +15,7 @@
 
 `tracepath -b {{host}}`
 
-- Specify maximum TTL (number of hops):
+- Specify a maximum TTL (number of hops):
 
 `tracepath -m {{max_hops}} {{host}}`
 
@@ -23,6 +23,6 @@
 
 `tracepath -p {{destination_port}} {{host}} `
 
-- Specify the initial packet length:
+- Specify the initial packet length (defaults to 65535 for IPv4 and 128000 for IPv6):
 
 `tracepath -l {{packet_length}} {{host}}`
--- a/pages/linux/tracepath.md
+++ b/pages/linux/tracepath.md
@@ -11,10 +11,6 @@
 
 `tracepath -p {{destination_port}} {{host}} `
 
-- Print numerical IP addresses only:
-
-`tracepath -n {{host}}`
-
 - Print both hostnames and numerical IP addresses:
 
 `tracepath -b {{host}}`
@@ -26,3 +22,7 @@
 - Specify the initial packet length (defaults to 65535 for IPv4 and 128000 for IPv6):
 
 `tracepath -l {{packet_length}} {{host}}`
+
+- Use only IPv6 addresses:
+
+`tracepath -6 {{host}}`


marchersimon:
@peterbabic please make PRs to `tldr/master`, not `tldr/main`. We are currently switching from `master` to `main`, but I don't think it's ready yet.

bl-ue:
@marchersimon you're correct; the `master` branch is being auto-mirrored to `main`.  However, we're working on changing that. @peterbabic track #5110 for updates.

FYI @marchersimon, in case you're interested you can change the base branch yourself since you have write access to the tldr-pages repos ;)

peterbabic:
@marchersimon and @bl-ue I was maybe trying to be too little too progressive :)

bl-ue:
Branch looks good. 👍🏻

pages/linux/tracepath.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Print both hostnames and numerical IP addresses:
+
+`tracepath -b {{host}}`
+
+- Specify maximum TTL (number of hops):
+
+`tracepath -m {{max_hops}} {{host}}`
+
+- Specify the initial destination port:
```
marchersimon, line 22: Is it initial, because the destination port can change?

pages/linux/tracepath.md
```diff
@@ -0,0 +1,28 @@
...
+
+`tracepath -p {{destination_port}} {{host}} `
+
+- Specify the initial packet length:
```
marchersimon, line 26:
```suggestion
- Specify the initial packet length (defaults to 65535 for IPv4 or 128000 for IPv6):
```

marchersimon (changes requested):

pages/linux/tracepath.md
bl-ue: I'd say 'and' here @marchersimon, `...for IPv4 and 128000...`.  It's more common to do so. :shr

pages/linux/tracepath.md
peterbabic: Good question. I am not sure at all. It say initial everywhere I look. Should it be tested?

bl-ue: If it's convenient for you @peterbabic then please do.  Otherwise, don't go out of your way—this is what the man page says and if it needs clarification someone can come along and fix it.  We frequently get PRs to clarify this or that 👍🏻

marchersimon: My theory is, that for example if you send a request to a https port, but the destination has http only, it gets redirected, so that the destination port isn't the intended port

peterbabic: Sounds plausible, I try to test this in the morning.

Waples: Option p with tracepath command will allow to set destination port to be used. Whenever we would need to set a fixed port at the destination probe, we can use this option. It generally limits the usage of the port options.

peterbabic: - `tracepath peterbabic.dev` returns `no reply Too many hops` after TTL is reached
- `traceroute peterbabic.dev` returns the path
- `mtr -w peterbabic.dev` return the path

Looking at https://man.archlinux.org/man/core/traceroute/traceroute.8.en#default it provides a hint.
Consequentially:

- `tracepath -p 33434 peterbabic.dev` returns the path

The host probed in the examples contains post with further analysis.

What to do with this information?

peterbabic: I have added a new commit, suggesting a an initial port related change (I did not find how can I add a review like you all do).

bl-ue: >What to do with this information?

Great, thank you @Waples and @peterbabic for your work.  I'm not sure what the best thing to do here is, but probably what I'd do for now it leave it.  It's fine as-is, I think, and is at least as good as the man page.  @peterbabic I think your commit is right, but I have a comment about the formatting of that information.

>I did not find how can I add a review like you all do

It's really simple: (1) go to the "Files Changed" tab, (2) hover over a line, (3) click (and drag, if you want) the blue plus icon, (4) type your comment.

pages/linux/tracepath.md
```diff
@@ -0,0 +1,28 @@
...
+
+`tracepath -b {{host}}`
+
+- Specify a maximum TTL (number of hops):
+
+`tracepath -m {{max_hops}} {{host}}`
+
+- Specify the initial destination port, preferably 33434:
+
+`tracepath -p {{destination_port}} {{host}} `
```
bl-ue, line 24: Why not just put the port in the example?
```suggestion
- Specify the initial destination port:

`tracepath -p {{destination_port}} {{33434}} `
```

peterbabic: Note that the man page with the hint I linked above is for `traceroute` command, not the `tracepath` discussed here (hopefully everyone noticed). These two commands are only loosely related and from my research there is no comprehensive, easily accessible explanation around the internet about this topic.

I believe there are three scenarios:
1. without port - not useful, not sure what it even does tbh
2. with arbitrary port - useful for testing firewall settings
3. with port 33434 - useful for finding path (what this command should do)

Maybe we could transform it in a way that case 1. would not be even mentioned, since it is not very useful as is (at least on my machine).
Both, case 2. and 3. should then be documented as two separate use cases. This would provide me as, a semi-advanced user of this command, the most usage details - exactly what I look for when I use tldr. Not entirely sure about how the beginner user would benefit. 

What do you think?

bl-ue: I agree.  Remove this example and document cases 2 & 3. 👍🏻

peterbabic: @bl-ue I considered the "comments" as you guided me, but the new comment push is much better aligned with my workflow. I will keep pushing commits, unless someone objects. See https://github.com/tldr-pages/tldr/pull/5744/commits/714d8d6d7be51d80b4b4db331ffb43c54f82c748

pages/linux/tracepath.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Print both hostnames and numerical IP addresses:
+
+`tracepath -b {{host}}`
+
+- Specify a maximum TTL (number of hops):
+
+`tracepath -m {{max_hops}} {{host}}`
+
+- Specify the initial packet length (defaults to 65535 for IPv4 and 128000 for IPv6):
```
peterbabic, line 26: Since this got included, shouldn't we also explain ho to use IPv6 (it is with `-6` parameter).

How is the IPv6 integration among other commands overall? I do not feel like beginner users are not actively seeking it.

There could probably be a configuration option of sort to prevent IPv4 (default, at least for now), IPv6 or both to solve this.

peterbabic, line 26: Since this got included, shouldn't we also explain ho to use IPv6 (it is with `-6` parameter).

How is the IPv6 integration among other commands overall? I do not feel like beginner users are actively seeking it, but might be mistaken.

There could probably be a configuration option of sort to prefer IPv4 (default, at least for now), IPv6 or both to solve this.

pages/linux/tracepath.md
```diff
@@ -0,0 +1,28 @@
...
+
+`tracepath -p 33434 {{host}}`
+
+- Specify the initial destination port, useful with non-standard firewall settings:
+
+`tracepath -p {{destination_port}} {{host}} `
+
+- Print numerical IP addresses only:
+
+`tracepath -n {{host}}`
```
peterbabic, line 16: On a second thought, I think this is not that much useful option, compared to the `-b`. 

It could potentially be removed or replaced with `-6`, see https://github.com/tldr-pages/tldr/pull/5744/files#r613359397

pages/linux/tracepath.md
Waples: Aggreed.

bl-ue: I don't think it matters a whole lot. 🤷🏻

pages/linux/tracepath.md
```diff
@@ -0,0 +1,28 @@
+# tracepath
+
+> Trace the path to a network host discovering MTU along this path.
+> More information: <https://manned.org/tracepath>.
+
+- A preferred way to trace the path to a host:
+
+`tracepath -p 33434 {{host}}`
```
marchersimon, line 8:
```suggestion
`tracepath -p {{33434}} {{host}}`
```

peterbabic: I have read https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#page-structure where this is mentioned.

Is there some other publicly available discussion about the {{handlebars}} subject, please? Since there is no very specific keyword associated, it is hard to search it.

bl-ue (approved):

pages/linux/tracepath.md
marchersimon: Not that I'm aware of. Maybe @bl-ue or @waldyrious know of something

marchersimon (approved):

pages/linux/tracepath.md
bl-ue: We call it [token syntax](https://github.com/tldr-pages/tldr/blob/a108e73ce3cff3500e106f172db011a66dcfa4a5/contributing-guides/style-guide.md#token-syntax), and we usually put it on any value that the user has a choice in.

Waples (approved):

PR #5745, opened 2021-04-13
sfill: add page

mkcn:

### DIFF
--- a/pages/linux/sfill.md
+++ b/pages/linux/sfill.md
@@ -1,0 +1,20 @@
+# sfill
+
+> Secure overwrite free space and inodes of the partition where the specified directory resides.
+> More information: <http://manpages.ubuntu.com/manpages/xenial/man1/sfill.1.html>.
+
+- Overwrite free space and inodes of an USB with 38 writes (slow but secure):
+
+`sfill {{/media/user/mounted_usb_directory}}`
+
+- Overwrite free space and inodes of an USB with 2 writes (fast but less secure) and show status:
+
+`sfill -l -v {{/media/user/mounted_usb_directory}}`
+
+- Overwrite only free spece of an USB:
+
+`sfill -I {{/media/user/mounted_usb_directory}}`
+
+- Overwrite only free inodes of an USB:
+
+`sfill -i {{/media/user/mounted_usb_directory}}`
--- a/pages/linux/sfill.md
+++ b/pages/linux/sfill.md
@@ -17,4 +17,4 @@
 
 - Overwrite only free inodes of an USB:
 
-`sfill -i {{/media/user/mounted_usb_directory}}`
+`sfill -i {{/path/to/mounted_usb_directory}}`
--- a/pages/linux/sfill.md
+++ b/pages/linux/sfill.md
@@ -1,6 +1,6 @@
 # sfill
 
-> Secure overwrite free space and inodes of the partition where the specified directory resides.
+> Securely overwrite the free space and inodes of the partition where the specified directory resides.
 > More information: <https://manned.org/sfill>.
 
 - Overwrite free space and inodes of an USB with 38 writes (slow but secure):
--- a/pages/linux/sfill.md
+++ b/pages/linux/sfill.md
@@ -13,7 +13,7 @@
 
 - Overwrite only free spece of an USB:
 
-`sfill -I {{/media/user/mounted_usb_directory}}`
+`sfill -I {{/path/to/mounted_usb_directory}}`
 
 - Overwrite only free inodes of an USB:
 
--- a/pages/linux/sfill.md
+++ b/pages/linux/sfill.md
@@ -3,22 +3,22 @@
 > Securely overwrite the free space and inodes of the partition where the specified directory resides.
 > More information: <https://manned.org/sfill>.
 
-- Overwrite free space and inodes of an USB with 38 writes (slow but secure):
+- Overwrite free space and inodes of a USB with 38 writes (slow but secure):
 
 `sfill {{/path/to/mounted_usb_directory}}`
 
-- Overwrite free space and inodes of an USB with 6 writes (fast but less secure) and show status:
+- Overwrite free space and inodes of a USB with 6 writes (fast but less secure) and show status:
 
 `sfill -l -v {{/path/to/mounted_usb_directory}}`
 
-- Overwrite free space and inodes of an USB with 1 write (very fast but insecure) and show status:
+- Overwrite free space and inodes of a USB with 1 write (very fast but insecure) and show status:
 
 `sfill -ll -v {{/path/to/mounted_usb_directory}}`
 
-- Overwrite only free spece of an USB:
+- Overwrite only free spece of a USB:
 
 `sfill -I {{/path/to/mounted_usb_directory}}`
 
-- Overwrite only free inodes of an USB:
+- Overwrite only free inodes of a USB:
 
 `sfill -i {{/path/to/mounted_usb_directory}}`
--- a/pages/linux/sfill.md
+++ b/pages/linux/sfill.md
@@ -1,7 +1,7 @@
 # sfill
 
 > Secure overwrite free space and inodes of the partition where the specified directory resides.
-> More information: <http://manpages.ubuntu.com/manpages/xenial/man1/sfill.1.html>.
+> More information: <https://manned.org/sfill>.
 
 - Overwrite free space and inodes of an USB with 38 writes (slow but secure):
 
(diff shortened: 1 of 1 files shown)


pages/linux/sfill.md
```diff
@@ -0,0 +1,20 @@
+# sfill
+
+> Secure overwrite free space and inodes of the partition where the specified directory resides.
+> More information: <http://manpages.ubuntu.com/manpages/xenial/man1/sfill.1.html>.
```
bl-ue, line 4:
```suggestion
> More information: <https://manned.org/sfill>.
```

pages/linux/sfill.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Overwrite free space and inodes of an USB with 38 writes (slow but secure):
+
+`sfill {{/media/user/mounted_usb_directory}}`
```
bl-ue, line 8:
```suggestion
`sfill {{/path/to/mounted_usb_directory}}`
```

pages/linux/sfill.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Overwrite free space and inodes of an USB with 2 writes (fast but less secure) and show status:
+
+`sfill -l -v {{/media/user/mounted_usb_directory}}`
```
bl-ue, line 12:
```suggestion
`sfill -l -v {{/path/to/mounted_usb_directory}}`
```

pages/linux/sfill.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Overwrite only free spece of an USB:
+
+`sfill -I {{/media/user/mounted_usb_directory}}`
```
bl-ue, line 16:
```suggestion
`sfill -I {{/path/to/mounted_usb_directory}}`
```

pages/linux/sfill.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Overwrite only free inodes of an USB:
+
+`sfill -i {{/media/user/mounted_usb_directory}}`
```
bl-ue, line 20:
```suggestion
`sfill -i {{/path/to/mounted_usb_directory}}`
```

bl-ue (commented):
Thank you for the new page, @mkcn!  I've left a few commits for you to review.

pages/linux/sfill.md
mkcn: Hi @bl-ue, 
thanks for the suggestion, please consider this generic link may a bit more dangerous because if an inexperienced user try to use this command with the wrong folder (e.g. /home/user/my_custom_usb_folder/ where the USB is not currently mounted) the command will **be executed on the partition where the specified directory resides** which is the main one, during the process `sfill` creates a file as large as the free space available and some distro will slow down or even freeze as long as the process is not completed.

Specify "/media/user/mounted_usb_directory" I believe is bit less risky and more intuitive because:

- it is the default path of mounted USB for many (maybe all) distros
- when the USB is not mounted the folder does not exist 

Given this, if you still prefer the `/path/to/mounted_usb_directory` I will change it

pages/linux/sfill.md
mkcn: https://github.com/tldr-pages/tldr/pull/5745#discussion_r612804101

pages/linux/sfill.md
mkcn: https://github.com/tldr-pages/tldr/pull/5745#discussion_r612804101

pages/linux/sfill.md
mkcn: https://github.com/tldr-pages/tldr/pull/5745#discussion_r612804101

pages/linux/sfill.md
bl-ue: @mkcn You raise some good points.  We normally like to make it like an English phrase `e.g. path to file`, but I'm sure we can make an exception here.  Maybe @sbrl will have some suggestions.

Waples: I'm going with @bl-ue 's change, you can mount USBs anywhere so `path to mounted_usb_dir` is better imo. Not every system uses `/media/${USER}/${MountPoint}`, that's more with Desktop Environments and also as said before, not everytime the same on every system.

pages/linux/sfill.md
```diff
@@ -0,0 +1,20 @@
+# sfill
+
+> Secure overwrite free space and inodes of the partition where the specified directory resides.
+> More information: <https://manned.org/sfill>.
+
+- Overwrite free space and inodes of an USB with 38 writes (slow but secure):
```
bl-ue, line 6: Is `38` customizable here?

pages/linux/sfill.md
sbrl: I would suggest that consistency would be the best option here. Some systems mount to `/mnt` by default, while others prefer `/media/username`. So either `/absolute/path/to/mount_point` or `path/to/mount_mount` depending on whether an absolute path is required or not.

pages/linux/sfill.md
Waples: By default:
```bash
       The secure data deletion process of sfill goes like this:

       *      1 pass with 0xff

       *      5 random passes. /dev/urandom is used for a secure RNG if
              available.

       *      27 passes with special values defined by Peter Gutmann.

       *      5 random passes. /dev/urandom is used for a secure RNG if
              available.
```

`1+5+27+5 = 38`

When using the `-l` flag:
```bash
       -l     lessens the security. Only two passes are written: one mode with
              0xff and a final mode with random values.
```
So that makes `6`.

bl-ue (approved):

pages/linux/sfill.md
mkcn: Correct, 38 is default and not strictly customizable. I [fixed](545bd8b7f9dc6b33680e91426fea1a7fe917fd7a) the number of writes with '-l' option and added  a new example for the '-ll' mode.

Waples (dismissed):

pages/linux/sfill.md
```diff
@@ -0,0 +1,24 @@
+# sfill
+
+> Secure overwrite free space and inodes of the partition where the specified directory resides.
```
sbrl, line 3: Perhaps this will read better?
```suggestion
> Securely overwrite the free space and inodes of the partition where the specified directory resides.
```

pages/linux/sfill.md
```diff
@@ -0,0 +1,24 @@
...
+> More information: <https://manned.org/sfill>.
+
+- Overwrite free space and inodes of an USB with 38 writes (slow but secure):
```
sbrl, line 6: Why is a "USB" mentioned specifically here? Does it not work with other kinds of disks?/

sbrl (changes requested):

pages/linux/sfill.md
Waples: maybe without _*the*_ in _the free space_

pages/linux/sfill.md
Waples: It also works for other (kinds of) disks, I think people mainly use this to securly clean up USB drives.

pages/linux/sfill.md
bl-ue: @Waples I'm not sure about that—it helps `of the partition` later on.

pages/linux/sfill.md
mkcn: Correct, mainly with USBs or external disks as far as i know. The main description ("..of the partition where the specified directory resides..") is indeed more generic but I preferred to use a more useful example.

bl-ue: I say it should be `a USB` because I think it's generally pronouced as the three consecutive letters `U S B`.  Thoughts, anyone?  This might also be a regional thing, not sure.
```suggestion
- Overwrite free space and inodes of a USB with 38 writes (slow but secure):
```

mkcn: @bl-ue you're right, 'a USB' is the most correct form. Fixed.

sbrl: Hrm. I would prefer `a disk` to "a USB". It's more correct to how the command actually works and sounds better too.

mkcn: Changed

pages/linux/sfill.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Overwrite free space and inodes of a USB with 6 writes (fast but less secure) and show status:
+
+`sfill -l -v {{/path/to/mounted_usb_directory}}`
+
+- Overwrite free space and inodes of a USB with 1 write (very fast but insecure) and show status:
+
+`sfill -ll -v {{/path/to/mounted_usb_directory}}`
+
+- Overwrite only free spece of a USB:
```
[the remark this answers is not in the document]
mkcn: Fixed

sbrl (approved):
Thanks, @mkcn!

PR #5714, opened 2021-04-08
bash, dash, fish, ksh, rbash, sh, zsh: refresh

bl-ue:

PR #5746, opened 2021-04-13
shellcheck: refresh

bl-ue:


pages/common/shellcheck.md
```diff
@@ -1,20 +1,33 @@
...
 - Check a shell script:
 
-`shellcheck {{file.sh}}`
+`shellcheck {{path/to/script.sh}}`
 
-- Override script's shebang:
+- Check a shell script interpreting it as the specified shell dialect (overrides the shebang at the top of the script):
 
-`shellcheck --shell {{sh|bash|ksh}} {{file.sh}}`
+`shellcheck --shell {{sh|bash|dash|ksh}} {{path/to/file.sh}}`
```
einverne, line 13: To keep consistent, should it be

```suggestion
`shellcheck --shell {{sh|bash|dash|ksh}} {{path/to/script.sh}}`
```

pages/common/shellcheck.md
```diff
@@ -1,20 +1,33 @@
...
 
-- Ignore certain errors:
+- Ignore one or more error types:
 
-`shellcheck --exclude {{SC1009}} {{file.sh}}`
+`shellcheck --exclude {{SC1009,SC1073}} {{path/to/file.sh}}`
```
einverne, line 17: Same here.

einverne (commented):
A little suggestion to keep consistent.

pages/common/shellcheck.md
Waples: consistency, `path/to/script.sh` or `path/to/file.sh` @einverne , in your other comment you say script :P

pages/common/shellcheck.md
navarroaxel: Do you use `Hey dude, run this sh script`? I say `sh file` and `bash script` so I vote for `file.sh` :rofl:

bl-ue: @einverne you're right—good catch.

You think we should use `file.sh`?  I'm fine with that—we should also change it in #5714, as well, then. @sbrl

sbrl: It really doesn't matter. I suggest using whichever is more clear on a case-by-case basis - this isn't something that needs standardising or worrying too much over consistency.

einverne (approved):

marchersimon (approved):

navarroaxel (approved):

PR #5178, opened 2021-01-24
brew-bundle, caffeinate, open and tmutil: add German translation

GETandSELECT:
- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #5754, opened 2021-04-14
chore(deps): update tldr-lint to 0.0.11

MasterOdin:

This updates the tldr-lint package to its newest version, which includes a new flag that allows for ignoring specific TLDR codes. This can be used to allow the linter to run against non-english folders, ignoring out the TDLR errors that are english specific.


bl-ue:
@MasterOdin we need to disable the following errors:

- ~`TLDR003`: Descriptions should start with a capital letter — _In other languages, some nouns appear at the beginning of sentences, and the nouns sometimes happen to be a technical name such as a command name_~
   ^^ I'm not sure about this, because maybe we should wrap those technical nouns in backticks

- `TLDR005`: Example descriptions should end in a colon with no trailing characters — _Chinese has it's own colon and period, and we're working on normalizing all of that to the correct, **Chinese** punctuation; definitely have a look at #5240_

- `TLDR004`: Command descriptions should end in a period - _same as `TLDR005`_

bl-ue:
There's a queer line in that output:
```
Description ends in' , '. Consider writing your sentence on one line.
```
Shouldn't we be using `yy.error` or something of that nature [here](https://github.com/tldr-pages/tldr-lint/blob/d47d9ccfe8c7850d972da971ad5189a8fe5cd07d/tldr.l#L136)?

The caveats of the current approach are:
- There is no space between `in` and `'` (minor)
- The name of the file can't be shown (major)

MasterOdin:
@bl-ue it might be for https://github.com/tldr-pages/tldr/blob/master/pages.fr/linux/ip.md which has the first description line spread out over two lines. I would agree it would probably be good though to move this into an error or something that's better caught and can show the file name and line number.

bl-ue:
I just fixed that ;) https://github.com/tldr-pages/tldr/commit/d232c298c2170e72b82797007be4632d3f44a104

package-lock.json
```diff
@@ -5,12 +5,11 @@
   "packages": {
     "": {
       "name": "tldr-pages",
-      "hasInstallScript": true,
```
bl-ue, line 8: Out of curiosity, why did this get removed?

bl-ue (approved):
Finally, [this](https://github.com/tldr-pages/tldr/pull/5178#issuecomment-766375668) will be accomplished 😅

package-lock.json
MasterOdin: Yes. npm 7 saves some meta-data about the run-scripts in the package.json file within the lock file. When I moved husky from postinstall to prepare in #5711, that altered the package-lock.json file for this line technically, but since I manually editing the file through GH, that change was not propagated.

bl-ue (approved):
Works!
```bash
$ for p in pages*; do tldr-lint --ignore "TLDR003,TLDR004,TLDR005,TLDR015,TLDR104" $p; done
$
```

Waples (approved):

Issue #1960, opened 2018-01-30
Rename "common" directory to "unix"?


colinmorris:
According to comments on #1930, pages listed under `common/` must be available (natively) on all platforms, including Windows.

"For historical reasons" (presumably the fact that [the Windows platform was added fairly recently](https://github.com/tldr-pages/tldr/issues/1579)), the commands currently under `common/` don't all meet that requirement. 

In fact, if someone did clean up `common/` to enforce that requirement, I think it would go from 532 commands, to... less than 5? I don't have much Windows experience, but based on some searching, it seems like the few commands it has in common with the other OSes (`ping`, `more`, `sort`, `find`) differ significantly in their usage. The only command I could find in `common/` whose examples would run on Windows was `whoami`.

The other three platforms (linux, OSX, SunOS), are all unix-based, and have a lot of commands in common for that reason. A directory encompassing those 3 platforms (i.e. what `common/` used to refer to), is useful, and saves a lot of duplication of content.

Windows has no common ancestry with them. A directory for commands common to all 4 OSes isn't very useful. (Though if someone wants to create it *in addition to* renaming `common` -> `unix/`, I wouldn't particularly object. It'd just be very small.)

pxgamer:
I slightly disagree with renaming it due to the massive backwards compatibility issue. I do think though that there could be a `unix-common` directory (which has been mentioned before by @agnivade).

A lot of the commands in `common` I know run on Windows. More than just `whoami`. :thinking:

colinmorris:
What's the backwards compatibility issue?

> A lot of the commands in common I know run on Windows. More than just whoami.

Could you give some examples? Just curious - like I said, I'm not very familiar with Windows.

pxgamer:
Most clients use the `common` directory as the common location for finding pages. Suddenly renaming it would probably break quite a few of them. I guess the other thing would be that a lot of them do work on Windows, and therefore wouldn't be available on Windows clients without specifying `--os unix`.

Regarding the commands that run on Windows, there are a load of compilers (clang, cmake, make), editors (nano, atom, sublime, etc.), archive tools (7z) shells (bash, sh) and general tools (cp, mv, curl, wget, apm, composer, node, etc.) that all work on Windows using `cmd` if they are installed.

In fact, I use a lot of them on a daily basis through Git for Windows' `usr/bin` directory.

colinmorris:
[...]

PR #5759, opened 2021-04-14
Reference dirs builtin in pushd/popd

ambirdsall-gogo:

I left the platform box unchecked: I would have thought these belonged in `common`, but have only (so far) edited the `pushd` and `popd` pages for the platform I'm currently on. I believe that the behavior is standardized across *nix platforms—for all I know, it may even work the same on windows—but only have a macOS computer handy to verify the behavior at the moment. I do have a linux computer I can check after work if desired, but didn't want to add those pages without first verifying. Since this PR edits multiple commands, I wasn't sure how to follow the recommended "$COMMAND: description" format.

~The only other potential complication I see is that, being a builtin, its implementation is shell-specific. The listed flags work the same on bash and zsh, at least.~ [removing the new `pages/osx/dirs.md` makes that line a bit superfluous] Guidance and suggested changes welcome.


ambirdsall-gogo:
Marking as draft; I noticed a few things and decided to clean this up a little before submitting it for consideration to merge. Notably, there already is a `pages/common/dirs.md` with most of the usages I documented, so I deleted the `pages/osx/dirs.md`; since this then becomes a more tightly-scoped cross-reference addition, I figure I can expand it more safely across different platforms' `pushd`/`popd` pages: if I understand correctly, all platforms with `pushd` and `popd` also provide a related `dirs` builtin, so simply documenting the connection has less verification overhead. Once I've done so, I'll mark the PR as ready for review again.

Waples:
Can confirm: `pushd` and `popd` works the same in `bash`, `zsh` and `fish` shell, in 3 different *nix systems (arch linux, centos, rpiOs). Maybe a move to `pages/common/` @bl-ue ?

bl-ue:
Probably.  `pushd` and `popd` also surprisingly happen to be on Windows in `cmd.exe` as well, and have roughly the same functionality, although they accept different options, and there is no `dirs` there.  So let's leave `pushd.md` and `popd.md` in `pages/windows`, and merge the rest into `pages/common` (because you can have different versions of a page in different platform directories, and the one in the specific platform will be chosen instead of common if possible 😉)

Waples:
Fair enough ^^

bl-ue:
You're unfortunately right; we have long had this problem :/ (#1960)

bl-ue:
@ambirdsall-gogo FYI—there's no need to force-push in PRs because we auto-squash all of the commits into one just before merging.

Waples (approved):
lgtm

pages/common/pushd.md
```diff
@@ -1,7 +1,7 @@
 # pushd
 
 > Place a directory on a stack so it can be accessed later.
-> See also `popd` to switch back to original directory.
+> See also `popd` to switch back to original directory and, on UNIX-like operating systems, `dirs` to display directory stack contents.
```
bl-ue, line 4:
```suggestion
> See also `popd` to switch back to original directory, and `dirs` to display directory stack contents.
```

pages/common/popd.md
```diff
@@ -1,6 +1,7 @@
 # popd
 
 > Remove a directory placed on the directory stack via the pushd shell built-in.
+> See also `pushd` to place a directory on the stack and, on UNIX-like operating systems, `dirs` to display directory stack contents.
```
bl-ue, line 4:
```suggestion
> See also `pushd` to place a directory on the stack, and `dirs` to display directory stack contents.
```

bl-ue (commented):
I don't think pointing out that `dirs` is unix-only is necessary, because these pages will not even be shown on non-unix systems.

bl-ue:
As a matter of fact, all three of these commands are shell builtins, so it doesn't _really_ depend on the platform.  For examplee, mingw and cygwin use the unix-based versions rather than the Windows ones (I believe), though they are runtime-ishs only for Windows.

pages/common/pushd.md
ambirdsall-gogo: I added that with the windows user in mind: I could see someone running `tldr pushd`, following up with `tldr dirs`, and then getting frustrated when they try to use the nonexistent-on-windows `dirs` command. But I'm happy to go with your judgement here if you think the simpler description is better overall.

bl-ue: Right, I agree.  But since we already have Windows pages for [`pushd`](https://github.com/tldr-pages/tldr/blob/de637c416d65d4b849e01677caad1faea3e435bc/pages/windows/pushd.md) and [`popd`](https://github.com/tldr-pages/tldr/blob/de637c416d65d4b849e01677caad1faea3e435bc/pages/windows/popd.md), clients on Windows will display that page rather than the ones in `common/`, so they won't even see the pointer to `dirs`.

bl-ue (approved):
Good catch @ambirdsall-gogo!

sbrl:
> Sounds reasonable; I'm happy to add that to this PR. If I understand the page resolution correctly, windows users running tldr dirs would still get the pages/common/dirs.md output, which could be confusing. Doubt this is the only command where that's a risk, though.

Not quite correct. If you are on Windows and you type `tldr pushd` for example, if a `pushd` page exists in the `windows` platform, it will be chosen _first_ over any potential copy in the `common` platform. Furthermore, if a copy of `pushd` exists in the `osx` platform too, then both the `windows` _and_ the `common` copy would have to not exist for it to show.

sbrl (approved):
Thanks, @ambirdsall-gogo !

PR #5760, opened 2021-04-14
CONTRIBUTING: improve wording about example count limit

FantasyCookie17:
This was inspired by @bl-ue's comment https://github.com/jedisct1/minisign/issues/98#issuecomment-819805803.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,6 +31,8 @@ The basic format of a `tldr` page is a set of concrete usage examples.
 Here are a few guidelines to get started:
 
 1. Try to keep pages at around 5 examples. Pages can be longer if needed, but don't exceed 8 examples.
+   If specifying less than 5 makes more sense due to, for example, the command's scope being quite narrow, it is also 
+   possible to specify less, however, at least 1 example should always be present.
    Remember, it's OK if the page doesn't cover everything; that's what `man` is for.
 2. When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness.
    For example, commands that require `sudo` should include it directly in the examples.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,9 +30,8 @@ To get started, please [sign](https://cla-assistant.io/tldr-pages/tldr) the
 The basic format of a `tldr` page is a set of concrete usage examples.
 Here are a few guidelines to get started:
 
-1. Try to keep pages at around 5 examples. Pages can be longer if needed or shorter where that makes more sense, but 
-   specify one example at least and 8 examples at most.
-   Remember it's OK if the page doesn't cover everything; that's what `man` is for.
+1. Try to keep pages at around 5 examples. Pages can be longer or shorter when appropriate, but don't exceed 8 examples.
+   Remember, it's OK if the page doesn't cover everything; that's what `man` is for.
 2. When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness.
    For example, commands that require `sudo` should include it directly in the examples.
 3. Try to incorporate the spelled-out version of single-letter options in the example's description.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -31,6 +31,8 @@ The basic format of a `tldr` page is a set of concrete usage examples.
 Here are a few guidelines to get started:
 
 1. Try to keep pages at around 5 examples. Pages can be longer if needed, but don't exceed 8 examples.
+   If specifying less than 5 makes more sense due to, for example, the command's scope being quite narrow, it is also 
+   possible to specify less, however, at least 1 example should always be present.
```
bl-ue, line 35: I'd put it slightly differently...
```suggestion
   If specifying less than 5 makes more sense due to, for example, the command's scope being quite narrow, it's fine to due so; however, at least 1 example should always be present.
```

FantasyCookie17: Yeah, that sounds a bit better indeed. More compact and such. One moment…

bl-ue (approved):
Thank you @FantasyCookie17! ⚡

waldyrious:
Thank you for contributing to the documentation! I am not convinced that this change is warranted, though. As @bl-ue pointed out [here](https://github.com/jedisct1/minisign/issues/98#issuecomment-819805803), pages with only one example are a tiny minority in our corpus. Historically, we are much more likely to bump into a higher limit than into a lower one. With that in mind, changing the guideline just to make the minimum explicit doesn't seem like a net positive, especially since the proposed phrasing is also more complex to read. (Also, I don't understand why the reference to the maximum of 8 examples was removed.)

I'm open to be convinced that this change should be done, but at the moment, I'm afraid my position is against it.

FantasyCookie17:
> Also, I don't understand why the reference to the maximum of 8 examples was removed.

I think that was an accident on my side when applying @bl-ue's suggestion…

Anyway, my latest commit should fix both that, and reduce the complexity of the rule. However, I wouldn't mind if it ends up not getting accepted. I just thought it might be helpful since @bl-ue said "that should be updated".

CONTRIBUTING.md
```diff
@@ -30,8 +30,9 @@ To get started, please [sign](https://cla-assistant.io/tldr-pages/tldr) the
 The basic format of a `tldr` page is a set of concrete usage examples.
 Here are a few guidelines to get started:
 
-1. Try to keep pages at around 5 examples. Pages can be longer if needed, but don't exceed 8 examples.
-   Remember, it's OK if the page doesn't cover everything; that's what `man` is for.
+1. Try to keep pages at around 5 examples. Pages can be longer if needed or shorter where that makes more sense, but 
+   specify one example at least and 8 examples at most.
```
waldyrious, line 34: This is better! I still think that mentioning a minimum number of examples is unwarranted here, especially since we say above: "The basic format of a `tldr` page is a set of concrete usage examples."

How about this?

```suggestion
1. Try to keep pages at around 5 examples. Pages can be longer or shorter when appropriate, but don't exceed  8 examples.
```

CONTRIBUTING.md
```diff
@@ -30,8 +30,9 @@ To get started, please [sign](https://cla-assistant.io/tldr-pages/tldr) the
...
+   Remember it's OK if the page doesn't cover everything; that's what `man` is for.
```
waldyrious, line 35: Why remove the comma? I think it reads better with it.

FantasyCookie17: I probably accidentally pressed `d` in my editor or so… I wasn't intending to remove it. I'll re-add it.

CONTRIBUTING.md
FantasyCookie17: That sounds nice. I'll commit that.

waldyrious (approved):
LGTM, thanks @FantasyCookie17!

Waples (approved):

schneiderl (approved):

marchersimon (approved):

navarroaxel (approved):

schneiderl:
lots of approvals. merging it. thanks, everyone.

PR #5776, opened 2021-04-17
rusnapshot: add page

danyeet:

### DIFF
--- a/pages/linux/rusnapshot.md
+++ b/pages/linux/rusnapshot.md
@@ -5,7 +5,7 @@
 
 - Create a snapshot using a config file:
 
-`sudo rusnapshot -c {{/path/to/config.toml}} --cr`
+`sudo rusnapshot --config {{/path/to/config.toml}} --cr`
 
 - List created snapshots:
 
--- a/pages/linux/rusnapshot.md
+++ b/pages/linux/rusnapshot.md
@@ -9,20 +9,20 @@
 
 - List created snapshots:
 
-`sudo rusnapshot --config {{path/to/config.toml}} --list`
+`sudo rusnapshot -c {{path/to/config.toml}} --list`
 
 - Delete a snapshot by ID or the name of the snapshot:
 
-`sudo rusnapshot --config {{path/to/config.toml}} --del --id {{snapshot_id}}`
+`sudo rusnapshot -c {{path/to/config.toml}} --del --id {{snapshot_id}}`
 
 - Delete all `hourly` snapshots:
 
-`sudo rusnapshot --config {{path/to/config.toml}} --list --keep {{0}} --clean --kind {{hourly}}`
+`sudo rusnapshot -c {{path/to/config.toml}} --list --keep {{0}} --clean --kind {{hourly}}`
 
 - Create a read-write snapshot:
 
-`sudo rusnapshot --config {{path/to/config.toml}} --cr --rw`
+`sudo rusnapshot -c {{path/to/config.toml}} --cr --rw`
 
 - Restore a snapshot:
 
-`sudo rusnapshot --config {{path/to/config.toml}} --id {{snapshot_id}} --restore`
+`sudo rusnapshot -c {{path/to/config.toml}} --id {{snapshot_id}} --restore`
--- a/pages/linux/rusnapshot.md
+++ b/pages/linux/rusnapshot.md
@@ -11,7 +11,7 @@
 
 `sudo rusnapshot -c {{/path/to/config.toml}} -l`
 
-- Delete snapshot by id or the name of the snapshot:
+- Delete a snapshot by ID or the name of the snapshot:
 
 `sudo rusnapshot -c {{/path/to/config.toml}} --del --id {{snapshot_id}}`
 
--- a/pages/linux/rusnapshot.md
+++ b/pages/linux/rusnapshot.md
@@ -1,6 +1,6 @@
 # rusnapshot
 
-> BTRFS snapshotting utility made in rust.
+> BTRFS snapshotting utility made in Rust.
 > More information: <https://github.com/Edu4rdSHL/rusnapshot>.
 
 - Create a snapshot using a config file:
--- a/pages/linux/rusnapshot.md
+++ b/pages/linux/rusnapshot.md
@@ -3,18 +3,18 @@
 > BTRFS snapshotting utility made in rust.
 > For more information: <https://github.com/Edu4rdSHL/rusnapshot>.
 
-- Create a snapshot using a config file 
+- Create a snapshot using a config file:
 
 `sudo rusnapshot -c {{/path/to/config.toml}} --cr`
 
-- List created snapshots 
+- List created snapshots:
 
 `sudo rusnapshot -c {{/path/to/config.toml}} -l`
 
-- Delete snapshot by id or the name of the snapshot
+- Delete snapshot by id or the name of the snapshot:
 
 `sudo rusnapshot -c {{/path/to/config.toml}} --del --id idprovidedbyrusnapshot`
 
-- Delete all snapshots of [kind] three-hours
+- Delete all snapshots of [kind] three-hours:
 
 `sudo rusnapshot -c {{/path/to/config.toml}}  -l -k 0 --clean --kind three-hours`
--- a/pages/linux/rusnapshot.md
+++ b/pages/linux/rusnapshot.md
@@ -5,7 +5,7 @@
 
 - Create a snapshot using a config file:
 
-sudo rusnapshot --config {{path/to/config.toml}} --cr
+`sudo rusnapshot --config {{path/to/config.toml}} --cr`
 
 - List created snapshots:
 
--- a/pages/linux/rusnapshot.md
+++ b/pages/linux/rusnapshot.md
@@ -5,7 +5,7 @@
 
 - Create a snapshot using a config file:
 
-`sudo rusnapshot --config {{/path/to/config.toml}} --cr`
+`sudo rusnapshot --config {{path/to/config.toml}} --cr`
 
 - List created snapshots:
 
--- a/pages/linux/rusnapshot.md
+++ b/pages/linux/rusnapshot.md
@@ -1,6 +1,8 @@
 # rusnapshot
+
 > BTRFS snapshotting utility made in rust.
 > For more information: <https://github.com/Edu4rdSHL/rusnapshot>.
+
 - Create a snapshot using a config file 
 
 `sudo rusnapshot -c {{/path/to/config.toml}} --cr`
--- a/pages/linux/rusnapshot.md
+++ b/pages/linux/rusnapshot.md
@@ -15,6 +15,6 @@
 
 `sudo rusnapshot --config {{/path/to/config.toml}} --del --id {{snapshot_id}}`
 
-- Delete all snapshots of [kind] hourly:
+- Delete all hourly snapshots:
 
 `sudo rusnapshot --config {{/path/to/config.toml}} --list --keep {{0}} --clean --kind {{hourly}}`
--- a/pages/linux/rusnapshot.md
+++ b/pages/linux/rusnapshot.md
@@ -1,7 +1,7 @@
 # rusnapshot
 
 > BTRFS snapshotting utility made in rust.
-> For more information: <https://github.com/Edu4rdSHL/rusnapshot>.
+> More information: <https://github.com/Edu4rdSHL/rusnapshot>.
 
 - Create a snapshot using a config file:
 
--- a/pages/linux/rusnapshot.md
+++ b/pages/linux/rusnapshot.md
@@ -9,12 +9,12 @@
 
 - List created snapshots:
 
-`sudo rusnapshot -c {{/path/to/config.toml}} -l`
+`sudo rusnapshot --config {{/path/to/config.toml}} --list`
 
 - Delete a snapshot by ID or the name of the snapshot:
 
-`sudo rusnapshot -c {{/path/to/config.toml}} --del --id {{snapshot_id}}`
+`sudo rusnapshot --config {{/path/to/config.toml}} --del --id {{snapshot_id}}`
 
 - Delete all snapshots of [kind] three-hours:
 
-`sudo rusnapshot -c {{/path/to/config.toml}}  -l -k 0 --clean --kind three-hours`
+`sudo rusnapshot --config {{/path/to/config.toml}}  -l --keep 0 --clean --kind three-hours`
--- a/pages/linux/rusnapshot.md
+++ b/pages/linux/rusnapshot.md
@@ -17,4 +17,4 @@
 
 - Delete all snapshots of [kind] three-hours:
 
-`sudo rusnapshot --config {{/path/to/config.toml}}  -l --keep 0 --clean --kind three-hours`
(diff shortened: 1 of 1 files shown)


pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,20 @@
+# rusnapshot
+
+> BTRFS snapshotting utility made in rust.
```
marchersimon, line 3:
```suggestion
> BTRFS snapshotting utility made in Rust.
```

pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,20 @@
...
+> For more information: <https://github.com/Edu4rdSHL/rusnapshot>.
```
marchersimon, line 4:
```suggestion
> More information: <https://github.com/Edu4rdSHL/rusnapshot>.
```

pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Create a snapshot using a config file:
+
+`sudo rusnapshot -c {{/path/to/config.toml}} --cr`
```
marchersimon, line 8: Is there a long variant of `-c`, `-l`, or `-k`?

pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,20 @@
...
+
+- List created snapshots:
+
+`sudo rusnapshot -c {{/path/to/config.toml}} -l`
+
+- Delete snapshot by id or the name of the snapshot:
```
marchersimon, line 14:
```suggestion
- Delete a snapshot by id or the name of the snapshot:
```

pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,20 @@
...
+
+`sudo rusnapshot -c {{/path/to/config.toml}} --del --id idprovidedbyrusnapshot`
```
marchersimon, line 16:
```suggestion
`sudo rusnapshot -c {{/path/to/config.toml}} --del --id {{snapshot_id}}`
```

pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Delete all snapshots of [kind] three-hours:
```
marchersimon, line 18: I don't really understand what you mean by `[kind]`. Is `three-hours` a special kind of snapshot?

marchersimon (changes requested):
Thanks @danyeet. I left a few suggestions below. Also, if `rusnapshot` also works with relative paths, please delete all leading slashes.

pages/linux/rusnapshot.md
danyeet: Yes, for -l is --list, for -k is --keep and for -c is --config.

pages/linux/rusnapshot.md
Edu4rdSHL: Hi, rusnapshot dev here. `kind` is just an identifier for the snapshot, you can put any string that you want. Why it exists? Well, the tool is able to perform automatic cleanups of snapshots, you can use the `--keep` option to keep only the last X snapshots, but see the following scenario: I took hourly, weekly and monthly snapshots of the same subvolumes, without an identifier like `--kind` it's almost impossible to difference between them and the tool will treat all the snapshots equally. To solve this problem you can use  `--kind hourly` for the hourly snapshots, `--kind weekly` for the weekly snapshots and `--kind montly` for the monthly snapshots, that way when the tool performs the cleanups, it's able to "see" that there are three different "kind" of snapshots and that only need to clean the ones that match the current "kind", leaving the others intact.

I hope that I have been clear, let me know if you have any additional question.

Regards,
Ed

Edu4rdSHL:
Regarding to slashes, we are talking about btrfs subvolumes. The tool can actually works with relative paths but it's **highly** recommended to use full qualified paths, the reason is that the tool is also able to restore snapshots, so if you use relative paths and then try restoring the snapshot it will ends up creating the subvolume in the current directory what may or may not be correct. If the user wants to restore to an specific directory he can use `--from {{desired-directory-path}}` while using `--restore`.

So, I prefer to keep the full qualified paths.

Regards,
Ed

pages/linux/rusnapshot.md
bl-ue: Nice! then @danyeet, let's replace the short options with the long options.  In case you don't know, we prefer the long options to the short ones because since they're normally English words, they're essentially self-explanatory, thus providing documentation in addition to that in the example description.  Makes sense?
 ```suggestion
`sudo rusnapshot --config {{/path/to/config.toml}} --cr`
```

**Comment by bl-ue** [comment]
Hi @Edu4rdSHL, thank you very much for your input!  I'm perfectly fine with that—if the most common usages of a command are with absolute paths, then by all means that's what we'll do.

What about the path to the config file, though? In this page, it's `/path/to/config.toml`, which is absolute—but normally that type of thing works fine with being relative.  Is that the case here, or is it advisable to use an absolute path even for the config file?

File: pages/linux/rusnapshot.md [hunk 7]
```diff
@@ -0,0 +1,20 @@
+# rusnapshot
+
+> BTRFS snapshotting utility made in Rust.
```
bl-ue, line 3: I like prefer the verb `written` to `made` when discussing the software stack.  Cool that it's in Rust! 🚀
```suggestion
> BTRFS snapshotting utility written in Rust.
```

pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,20 @@
...
+> More information: <https://github.com/Edu4rdSHL/rusnapshot>.
+
+- Create a snapshot using a config file:
+
+`sudo rusnapshot -c {{/path/to/config.toml}} --cr`
+
+- List created snapshots:
+
+`sudo rusnapshot -c {{/path/to/config.toml}} -l`
```
bl-ue, line 12:
```suggestion
`sudo rusnapshot --config {{/path/to/config.toml}} --list`
```

pages/linux/rusnapshot.md
danyeet: Yeah, fully agree, just to be honest I had to double check if there were long options in the first place :D.

pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Delete a snapshot by ID or the name of the snapshot:
+
+`sudo rusnapshot -c {{/path/to/config.toml}} --del --id {{snapshot_id}}`
```
bl-ue, line 16:
```suggestion
`sudo rusnapshot --config {{/path/to/config.toml}} --del --id {{snapshot_id}}`
```

pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Delete all snapshots of [kind] three-hours:
+
+`sudo rusnapshot -c {{/path/to/config.toml}}  -l -k 0 --clean --kind three-hours`
```
bl-ue, line 20:
```suggestion
`sudo rusnapshot --config {{/path/to/config.toml}}  --list --keep 0 --clean --kind {{three-hours}}`
```

pages/linux/rusnapshot.md
danyeet: I'll fix it right at it.
Also yeah rust is very cool.

pages/linux/rusnapshot.md
danyeet: I think I made most of the changes 
I'll add the brackets for the variable

danyeet: Please take a look again if you find something I'll edit it right at it but I think for the most part is fine.

pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,20 @@
...
+
+`sudo rusnapshot --config {{/path/to/config.toml}} --list`
+
+- Delete a snapshot by ID or the name of the snapshot:
+
+`sudo rusnapshot --config {{/path/to/config.toml}} --del --id {{snapshot_id}}`
+
+- Delete all snapshots of [kind] three-hours:
+
+`sudo rusnapshot --config {{/path/to/config.toml}}  --list --keep 0 --clean --kind {{three-hours}}`
```
bl-ue, line 20:
```suggestion
`sudo rusnapshot --config {{/path/to/config.toml}}  --list --keep {{0}} --clean --kind {{three-hours}}`
```

pages/linux/rusnapshot.md
bl-ue: Maybe we should call it something like `hourly` then.  @danyeet what do you think?

danyeet: No problem with that - the idea is to show that it's a time interval so I'll change it no problem.

marchersimon: Also I'd rephrase the description a little. Maybe `- Delete all hourly snapshots:`.

danyeet: Yea that would be better. Just to clarify something - the time interval convention is just an example. You could maybe make something else, like backup before update, backup before restart and etc. We just thought it would be the easiest one to explain.

bl-ue: I'm inclined to prefer the first (`Delete all snapshots of kind hourly`) because `hourly` is just a name, with no meaning otherwise.  It could easily be `abc`, whereupon your suggestion @marchersimon would not make as much sense (`Delete all abc snapshots:`)

danyeet: Maybe just reformat hourly to [hourly] or {{hourly}} ? 
I mean I can see both sides, I am fine with both but abstracting it a bit might not be a bad idea :).

marchersimon: > the time interval convention is just an example

Ah, I tought it was all about time intervals. In that case `Delete all snapshots of kind hourly` is more universal. However in descriptions we use `` `hourly` ``, instead of `{{hourly}}` for variables.

danyeet: Gotcha.

Edu4rdSHL:
Hey @bl-ue, sorry for the delay. Yes, the path to the configuration file can be relative without any problem and doesn't really matter if it's relative/full.

pages/linux/rusnapshot.md
Edu4rdSHL: I have to clarify that the `--kind` option is not just for deleting, the deletion only happens if you use the `--clean` and `--keep` but the option can be used while creating new snapshots without the need of deleting previous ones. As I said initially, it's more of an identifier than anything else.

pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,20 @@
+# rusnapshot
+
+> BTRFS snapshotting utility written in Rust.
+> More information: <https://github.com/Edu4rdSHL/rusnapshot>.
+
+- Create a snapshot using a config file:
+
+`sudo rusnapshot --config {{/path/to/config.toml}} --cr`
```
bl-ue, line 8:
```suggestion
`sudo rusnapshot --config {{path/to/config.toml}} --cr`
```

pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,20 @@
...
+
+- List created snapshots:
+
+`sudo rusnapshot --config {{/path/to/config.toml}} --list`
```
bl-ue, line 12:
```suggestion
`sudo rusnapshot --config {{path/to/config.toml}} --list`
```

pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Delete a snapshot by ID or the name of the snapshot:
+
+`sudo rusnapshot --config {{/path/to/config.toml}} --del --id {{snapshot_id}}`
```
bl-ue, line 16:
```suggestion
`sudo rusnapshot --config {{path/to/config.toml}} --del --id {{snapshot_id}}`
```

pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Delete all `hourly` snapshots:
+
+`sudo rusnapshot --config {{/path/to/config.toml}} --list --keep {{0}} --clean --kind {{hourly}}`
```
bl-ue, line 20:
```suggestion
`sudo rusnapshot --config {{path/to/config.toml}} --list --keep {{0}} --clean --kind {{hourly}}`
```

danyeet:
I'll add two more examples as @Edu4rdSHL and I think are crucial.

bl-ue (dismissed):
Thank you @danyeet (and @Edu4rdSHL)! 🚀

danyeet:
After messing up three times in a row, I have added two more examples. So the first one is creating a read-write snapshot as the utility creates a read-only by default. The second one is restoring the created snapshot to a subvolume. `--from` specifies where.

marchersimon (approved):

bl-ue (approved):

danyeet:
Thank you for working it out with me and @Edu4rdSHL - it's great we can popularize this rust piece of art already haha.

Edu4rdSHL:
We should clarify that `--from` is only required if you want to restore the snaphsot to an specific place and not the one from where the snapshot was originally taken. If only `--restore --id {{ID}}` is provided, it will restore the snapshot to the original directory.

danyeet:
Since in the last example `--from` is not a strict requirement, should we omit it, or it's fine as is ?

Edu4rdSHL:
My vote is that we must remove `--from` from the restoration example as it's only required under very specific scenarios.

Edu4rdSHL (approved):

marchersimon (approved):

pages/linux/rusnapshot.md
```diff
@@ -0,0 +1,28 @@
...
+
+`sudo rusnapshot --config {{path/to/config.toml}} --list --keep {{0}} --clean --kind {{hourly}}`
+
+- Create a read-write snapshot:
+
+`sudo rusnapshot --config {{path/to/config.toml}} --cr --rw`
+
+- Restore a snapshot:
+
+`sudo rusnapshot --config {{path/to/config.toml}} --id {{snapshot_id}} --restore`
```
bl-ue, line 28: Since the `--config` is always present, how about we clear it a little and shorten all but the first occurence?

```suggestion
`sudo rusnapshot -c {{path/to/config.toml}} --list`

- Delete a snapshot by ID or the name of the snapshot:

`sudo rusnapshot -c {{path/to/config.toml}} --del --id {{snapshot_id}}`

- Delete all `hourly` snapshots:

`sudo rusnapshot -c {{path/to/config.toml}} --list --keep {{0}} --clean --kind {{hourly}}`

- Create a read-write snapshot:

`sudo rusnapshot -c {{path/to/config.toml}} --cr --rw`

- Restore a snapshot:

`sudo rusnapshot -c {{path/to/config.toml}} --id {{snapshot_id}} --restore`
```

danyeet:
Well definitely it looks way nicer. Testing it locally with tealdeer - looks very nice

bl-ue (approved):

Edu4rdSHL (approved):
Thank you @danyeet for creating the PR and thanks to @bl-ue & @marchersimon for the reviews, LGTM.

danyeet:
Thank you very much. Look forward to more contributions from me in the future. :rocket:

PR #5777, opened 2021-04-17
cut, grep, tee, wc: add Spanish translations

goznalo-git:

Hi there, I'm back with more translations. This time it's the turn of utterly important commands, all related to processing standard input. 

I've got some comments about these pages:
- All pages link to the corresponding "More information" website. I should note that these links are not present in the original (English) versions, we should include them. But there are more suggestions about these original versions, which is why I delayed changing them until discussing them here.
- As discussed in previous PRs, variables (e.g. patron) have been written without accents.
- Regarding the `cut` command: some of the featured examples are straight up wrong, as they don't feature any standard input or files. This should be changed, if you agree I will do so.
- Regarding the `tee` command: the last example was so far-fetched that I decided to remove it in the translated file. Again, if you agree we should do the same in the original one.
- Regarding the `wc` command: the previous-to-last option (`-m`) is IMO not relevant at all. I don't even know its use. I think we should change it for another case of wc after a pipe.

Thanks for your attention, I'd love to hear your comments about it.


pages.es/common/cut.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Corta todo desde el tercer caracter hasta el final de cada línea:
+
+`cut -c {{3-}}`
+
+- Corta el quinto campo de cada línea, usando los dos puntos como delimitadores de campos (por defecto el delimitador es tab):
+
+`cut -d'{{:}}' -f{{5}}`
+
+- Corta el segundo y décimo campo de cada línea, usando los punto y comas como delimitadores:
```
navarroaxel, line 22: El plural `punto y coma` es irregular, se mantiene invariable.
```suggestion
- Corta el segundo y décimo campo de cada línea, usando los punto y coma como delimitadores:
```

pages.es/common/grep.md
```diff
@@ -0,0 +1,37 @@
...
+
+- Busca un patrón dentro de un archivo:
+
+`grep {{patron}} {{ruta/al/archivo}}`
+
+- Busca un patrón exacto:
+
+`grep -F {{patron_exacto}} {{ruta/al/archivo}}`
+
+- Busca un patrón [R]ecursivamente en el directorio actual, mostrando los correspondientes [n]úmeros de línea, [I]gnorando archivos que no son de texto:
```
navarroaxel, line 15: Supongo que la gente que usa la terminal sabe que es un archivo binario, o lo buscará y aprenderá :man_shrugging:
```suggestion
- Busca un patrón [R]ecursivamente en el directorio actual, mostrando los correspondientes [n]úmeros de línea, [I]gnorando archivos binarios:
```

pages.es/common/grep.md
```diff
@@ -0,0 +1,37 @@
+# grep
```
navarroaxel, line 1: I think you translate an old version of the `grep` page, if you use a client that supports cache, like the python client you should use `tldr -u` to update the local cache:

pages.es/common/cut.md
```diff
@@ -0,0 +1,28 @@
+# cut
+
+> Corta campos provenientes de la entrada estándar o de archivos.
```
navarroaxel, line 3: Verifiqué y en realidad `cut out` sería `recortar`:
```suggestion
> Recorta campos provenientes de la entrada estándar o de archivos.
```

marchersimon:
Thanks @goznalo-git.

> these links are not present in the original (English) versions

I think you may have an outdated fork, since we added links to all four commands some time ago.
`cut` → <https://www.gnu.org/software/coreutils/cut>
`tee` → <https://www.gnu.org/software/coreutils/tee>
`wc` → <https://www.gnu.org/software/coreutils/wc>
`grep` → <https://man7.org/linux/man-pages/man1/grep.1.html>
(However I think we should replace the last one, with <https://www.gnu.org/software/grep/manual/>)

> Regarding the `cut` command: some of the featured examples are straight up wrong, as they don't feature any standard input or files. This should be changed, if you agree I will do so.

> Regarding the `tee` command: the last example was so far-fetched that I decided to remove it in the translated file. Again, if you agree we should do the same in the original one.

Sure, however maybe this should be done in a separate PR. @bl-ue what do you think?

> Regarding the `wc` command: the previous-to-last option (`-m`) is IMO not relevant at all. I don't even know its use. I think we should change it for another case of wc after a pipe.

If you're working with UTF-8 files, a character can have more than one Byte and in this case you want to know the number of characters, not the number of bytes.

pages.es/common/wc.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Cuenta las palabras en un archivo:
+
+`wc -w {{ruta/al/archivo}}`
+
+- Cuenta caracteres (bytes) en un archivo:
+
+`wc -c {{ruta/al/archivo}}`
+
+- Cuenta caracteres en un archivo (tomando sets de caracteres multi-byte en cuenta):
```
navarroaxel, line 18: Perhaps this is better
```suggestion
- Cuenta caracteres en un archivo (considerando los caracteres de varios bytes):
```

pages.es/common/wc.md
```diff
@@ -0,0 +1,24 @@
+# wc
+
+> Cuenta líneas, palabras, caracteres, etc.
```
navarroaxel, line 3: We update this in the main page:
```suggestion
> Cuenta líneas, palabras, o bytes.
```

pages.es/common/wc.md
```diff
@@ -0,0 +1,24 @@
...
+> Más información: <https://www.man7.org/linux/man-pages/man1/wc.1.html>
```
navarroaxel, line 4:
```suggestion
> Más información: <https://www.gnu.org/software/coreutils/wc>
```

navarroaxel (changes requested):
Hey @goznalo-git thanks! I made some comments, make sure to use the latest version of the English pages please.

bl-ue:
Btw I'm working a reply 😉

goznalo-git:
You're right guys, my fork was a bit outdated... Sorry about that. I will go over your replies asap, thanks!

navarroaxel:
> > Regarding the `wc` command: the previous-to-last option (`-m`) is IMO not relevant at all. I don't even know its use. I think we should change it for another case of wc after a pipe.
> 
> If you're working with UTF-8 files, a character can have more than one Byte and in this case you want to know the number of characters, not the number of bytes.

@marchersimon the `-m` or `--chars` try to count real chars, I tested it with this file content:
```
'⸮ⴭā
```
And I got this output :man_shrugging: :
```
$ wc -c unicode.txt 
10 unicode.txt
$ wc -m unicode.txt 
5 unicode.txt
```

bl-ue:
>- All pages link to the corresponding "More information" website. I should note that these links are not present in the original (English) versions, we should include them. But there are more suggestions about these original versions, which is why I delayed changing them until discussing them here.

Yes, we should add links.  Since #5143, however, we've decided to use <https://manned.org> for links instead of <https://man7.org> because it has some [advantages](https://github.com/tldr-pages/tldr/issues/5143#issuecomment-811640009), so will you please change the links to them.  Also, please add them to the English pages (and in fact all translations) as well.

We only use <https://manned.org> as a fallback, so if you can find something like the gnu coreutils page that @navarroaxel suggested use it instead.

I highly recommend using [`set-more-info-link.py`](https://github.com/tldr-pages/tldr/blob/ce747d2f46f40836209afcd06898073ddabbc520/scripts/set-more-info-link.py).  Usage
```
python3 scripts/set-more-info-link.py -p common/cut.md "https://manned.org/cut"
```

>- Regarding the wc command: the previous-to-last option (-m) is IMO not relevant at all. I don't even know its use. I think we should change it for another case of wc after a pipe.

As @marchersimon said:

>If you're working with UTF-8 files, a character can have more than one Byte and in this case you want to know the number of characters, not the number of bytes.

That's correct.  For exampe, emoji or Chinese characters occupy **two or more bytes per visible character**:
```console
$ node
Welcome to Node.js v12.21.0.
Type ".help" for more information.
> '🏇🏻'.length
4

```

bl-ue:
@navarroaxel I think the `5` is because of the 4 visible characters **+ the newline** ;)

navarroaxel:
@bl-ue I don't like `cut out` as verb to explain `cut` in the English page. The man page uses `remove`, `select` and `print` to explain the text processor. What do you think?

goznalo-git:
Just pushed the change of links to those suggested by @marchersimon, the `gnu.org/software/coreutils` ones.

pages.es/common/grep.md
```diff
@@ -0,0 +1,37 @@
...
+
+> Encuentra coincidencias en el texto introducido.
+> Soporta patrones simples y expresiones regulares.
+> Más información: <https://www.gnu.org/software/coreutils/grep/manual>
```
marchersimon, line 5:
```suggestion
> Más información: <https://www.gnu.org/software/grep/manual/>
```

marchersimon: Actually I think <https://www.gnu.org/software/grep/manual/grep.html> would be even better than <https://www.gnu.org/software/grep/manual/>.

pages.es/common/tee.md
```diff
@@ -0,0 +1,16 @@
+# tee
+
+> Lee desde la entrada estándar y escribe a la salida estándar a la vez que a archivos o comandos.
+> Más información: <https://www.gnu.org/software/coreutils/tee>
+
+- Copia la entrada estándar al archivo, reemplazando su contenido, y también ala salida estándar:
```
navarroaxel, line 6:
```suggestion
- Copia la entrada estándar al archivo, reemplazando su contenido, y también a la salida estándar:
```

pages.es/common/cut.md
```diff
@@ -0,0 +1,28 @@
+# cut
+
+> Recorta campos provenientes de la entrada estándar o de archivos.
+> Más información: <https://www.gnu.org/software/coreutils/cut>
+
+- Corta los primeros 16 caracteres de cada línea de la entrada estándar:
```
navarroaxel, line 6: In the all examples too:
```suggestion
- Recorta los primeros 16 caracteres de cada línea de la entrada estándar:
```

navarroaxel (changes requested):

marchersimon (approved):

navarroaxel (approved):

PR #5780, opened 2021-04-17
pio-debug, piodebuggdb: add page

marchersimon:
For #5517

### DIFF
--- a/pages/common/pio-debug.md
+++ b/pages/common/pio-debug.md
@@ -1,0 +1,24 @@
+# pio debug
+
+> Debug PlatformIO projects.
+> More information: <https://docs.platformio.org/en/latest/core/userguide/cmd_debug.html>.
+
+- Debug the PlatformIO project in the working directory:
+
+`pio debug`
+
+- Debug a specific PlatformIO project:
+
+`pio debug --project-dir {{path/to/platformio_project}}`
+
+- Debug a specific environment:
+
+`pio debug --environment {{environment}}`
+
+- Debug a PlatformIO project using a specific configuration file:
+
+`pio debug --project-conf {{path/to/platformio.ini}}`
+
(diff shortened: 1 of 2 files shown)


pages/common/pio-debug.md
```diff
@@ -0,0 +1,24 @@
+# pio debug
+
+> Debug PlatformIO projects.
+> More information: <https://docs.platformio.org/en/latest/core/userguide/cmd_debug.html>.
+
+- Debug the PlatformIO project in the working directory:
```
bl-ue, line 6:
```suggestion
- Debug the PlatformIO project in the current directory:
```

pages/common/piodebuggdb.md
```diff
@@ -0,0 +1,7 @@
+# piodebuggdb
+
+> This command is an alias of `pio debug --interface={{gdb}}`.
```
bl-ue, line 3: 👌🏻

bl-ue (approved):

pages/common/piodebuggdb.md
marchersimon: I just saw that the node client doesn't render `{{ }}` correctly, if it's not used where it expects it to be. 😢

bl-ue: Actually, let's remove the tokens.  We're just stating _what_ `piodebuggdb` is an alias for—and the user can't change that, hence there's no need to the token syntax.
```suggestion
> This command is an alias of `pio debug --interface=gdb`.
```

marchersimon: You're right. I just like the colorful output `{{ }}` creates.

bl-ue: Yeah, I do to :)

Waples (approved):

PR #5282, opened 2021-02-19
speedtest: add page

marchersimon:
speedtest: add page

PR #5300, opened 2021-02-21
speedtest: convert to alias page

bl-ue:
- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] **:star: The page (if an alias), follow the recommended [alias page style](???????????). ← ?????**
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

Please, maintainers, see what you think about this proposal.  Also see #5299.

**Note:** please wait for more than 2 approvals on this PR.  I'd like as many maintainers as possible to see it.

---

Also, what labels should I apply?  Do we need a new label?  I'll pick `new command` for right now.

PR #5781, opened 2021-04-17
speedtest, speedtest-cli: refresh

marchersimon:
This command really brings up nostalgia 😄 (#5282)


bl-ue:
What about this, though: https://github.com/tldr-pages/tldr/pull/5300#issuecomment-782950264

marchersimon:
I didn't know that PR existed 😄
However, both commands do have the exast same arguments and both came with the same package from the AUR.

Waples (approved):

bl-ue (approved):

navarroaxel:
Why we don't link the official CLI https://www.speedtest.net/apps/cli? :thinking:

marchersimon:
@navarroaxel because it doesn't give any information on its usage. <https://github.com/sivel/speedtest-cli> is the official GitHub page (even if it doesn't look like it) and it documents the command quite well.

owenvoke:
They're still different commands. 🤔 That hasn't changed since #5300, although I think `speedtest` should be moved from `linux` to `common`. 👍🏻

marchersimon:
@owenvoke they aren't. They both come with the same package and their binaries (python scripts) are the exact same.

bl-ue:
For reference (not sure if this is useful): https://github.com/teamookla/homebrew-speedtest

owenvoke:
This is what I get when using the official `speedtest` (https://www.speedtest.net/apps/cli).
<details>

Preview of speedtest

</details>

And for the unofficial `speedtest-cli` (https://github.com/sivel/speedtest-cli) (which looks like it aliases to `speedtest` now, which is annoying).
<details>

Preview of speedtest-cli

</details>

marchersimon:
@owenvoke I'm using Linux and `speedtest --help` and `speedtest-cli --help` gives the exact same result. In fact, both point to <https://github.com/sivel/speedtest-cli>.

owenvoke:
Hmm. 🤔 Confusing... I think the unofficial one has been around a lot longer, so perhaps that's why it's conflicting. It looks like in the docs for the official CLI, it mentions that `speedtest-cli` conflicts with their official client. 😕 Not sure which we should document then, probably the official client (for `speedtest`) and the unofficial one for `speedtest-cli` as is currently done. But perhaps have a comment in the description saying something like:
> For the unofficial Speedtest CLI, see `speedtest-cli`.

And vice versa for the official one?

marchersimon:
Sounds good

marchersimon:
Should a also use <https://github.com/sivel/speedtest-cli> for the official page, since even the official command points to it and the official website doesn't give any information on its use.

Waples (approved):

navarroaxel:
> @owenvoke I'm using Linux and `speedtest --help` and `speedtest-cli --help` gives the exact same result. In fact, both point to https://github.com/sivel/speedtest-cli.

@marchersimon I get the same output than @owenvoke.

I checked `apt install speedtest-cli && speedtest-cli --help` in Ubuntu with this output:

<details>
$ speedtest --help
usage: speedtest [-h] [--no-download] [--no-upload] [--single] [--bytes] [--share] [--simple] [--csv] [--csv-delimiter CSV_DELIMITER] [--csv-header] [--json] [--list]
                 [--server SERVER] [--exclude EXCLUDE] [--mini MINI] [--source SOURCE] [--timeout TIMEOUT] [--secure] [--no-pre-allocate] [--version]

Command line interface for testing internet bandwidth using speedtest.net. --------------------------------------------------------------------------
https://github.com/sivel/speedtest-cli

optional arguments:
  -h, --help            show this help message and exit
  --no-download         Do not perform download test
  --no-upload           Do not perform upload test
  --single              Only use a single connection instead of multiple. This simulates a typical file transfer.
  --bytes               Display values in bytes instead of bits. Does not affect the image generated by --share, nor output from --json or --csv
  --share               Generate and provide a URL to the speedtest.net share results image, not displayed with --csv
  --simple              Suppress verbose output, only show basic information
  --csv                 Suppress verbose output, only show basic information in CSV format. Speeds listed in bit/s and not affected by --bytes
  --csv-delimiter CSV_DELIMITER
                        Single character delimiter to use in CSV output. Default ","
  --csv-header          Print CSV headers
  --json                Suppress verbose output, only show basic information in JSON format. Speeds listed in bit/s and not affected by --bytes
  --list                Display a list of speedtest.net servers sorted by distance
  --server SERVER       Specify a server ID to test against. Can be supplied multiple times
  --exclude EXCLUDE     Exclude a server from selection. Can be supplied multiple times
  --mini MINI           URL of the Speedtest Mini server
  --source SOURCE       Source IP address to bind to
  --timeout TIMEOUT     HTTP timeout in seconds. Default 10
  --secure              Use HTTPS instead of HTTP when communicating with speedtest.net operated servers
  --no-pre-allocate     Do not pre allocate upload data. Pre allocation is enabled by default to improve upload performance. To support systems with insufficient
                        memory, use this option to avoid a MemoryError
  --version             Show the version number and exit
</details>

And for the official Ookla command in Ubuntu, I get this output:
<details>
$ speedtest --help
Speedtest by Ookla is the official command line client for testing the speed and performance of your internet connection.

Version: speedtest 1.0.0.2

Usage: speedtest [<options>]
  -h, --help                        Print usage information
  -V, --version                     Print version number
  -L, --servers                     List nearest servers
  -s, --server-id=#                 Specify a server from the server list using its id
  -I, --interface=ARG               Attempt to bind to the specified interface when connecting to servers
  -i, --ip=ARG                      Attempt to bind to the specified IP address when connecting to servers
  -o, --host=ARG                    Specify a server, from the server list, using its host's fully qualified domain name
  -p, --progress=yes|no             Enable or disable progress bar (Note: only available for 'human-readable'
                                    or 'json' and defaults to yes when interactive)
  -P, --precision=#                 Number of decimals to use (0-8, default=2)
  -f, --format=ARG                  Output format (see below for valid formats)
  -u, --unit[=ARG]                  Output unit for displaying speeds (Note: this is only applicable
                                    for ‘human-readable’ output format and the default unit is Mbps)
  -a                                Shortcut for [-u auto-decimal-bits]
  -A                                Shortcut for [-u auto-decimal-bytes]
  -b                                Shortcut for [-u auto-binary-bits]
  -B                                Shortcut for [-u auto-binary-bytes]
      --selection-details           Show server selection details
      --ca-certificate=ARG          CA Certificate bundle path
  -v                                Logging verbosity. Specify multiple times for higher verbosity
      --output-header               Show output header for CSV and TSV formats

 Valid output formats: human-readable (default), csv, tsv, json, jsonl, json-pretty

 Machine readable formats (csv, tsv, json, jsonl, json-pretty) use bytes as the unit of measure with max precision

 Valid units for [-u] flag: 
   Decimal prefix, bits per second:  bps, kbps, Mbps, Gbps
   Decimal prefix, bytes per second: B/s, kB/s, MB/s, GB/s
   Binary prefix, bits per second:   kibps, Mibps, Gibps
   Binary prefix, bytes per second:  kiB/s, MiB/s, GiB/s
   Auto-scaled prefix: auto-binary-bits, auto-binary-bytes, auto-decimal-bits, auto-decimal-bytes
</details>

Also the note from Ookla when you try to install the official client:
```
## Other non-official binaries will conflict with Speedtest CLI
# Example how to remove using apt-get
# sudo apt-get remove speedtest-cli
```

Kick the competition :martial_arts_uniform:

Waples:
on Arch Linux the `speedtest-cli` pkg also points to Sivel's github page. Also, there is not a `speedtest` package on this distribution.. :thinking:

marchersimon:
Hm, maybe @sivel could help clarifying that a bit.

sivel:
speedtest-cli has provided both a `speedtest-cli` and `speedtest` entrypoint since [almost the beginning](https://github.com/sivel/speedtest-cli/commit/acb1049463942c9d727d42cb47cd7a2e511804ac)

However, when Ookla introduced their tool, they also called the CLI `speedtest`, which means that there could be potential conflicts.  I have no plans to remove either entrypoint, so it is up to individual distributions that may package these tools to decide what to provide, or to indicate that they cannot be installed simultaneously.  However, it does seem most flexible to stick with `speedtest-cli` in documentation to avoid any confusion.  While perhaps stating that the `speedtest-cli` package also may contain a `speedtest` entrypoint.

bl-ue:
Wow, thank you @sivel.  That's really clear.

To summarize:

- The official/Ookla package only has `speedtest`, whereas @sivel's contains both `speedtest` and `speedtest-cli`
- @sivel's predates Ookla's

So I propose to do this:
- Have `speedtest-cli` document @sivel's
- Have `speedtest` document Ookla's, but add a note something along the lines of

   ```
   Note: some platforms link `speedtest` to `speedtest-cli`.  If some of the examples in this page don't work, try `tldr speedtest-cli`.
   ```

Waples (changes requested):
As @bl-ue stated on which I agree;
> So I propose to do this:
> - Have `speedtest-cli` document @sivel's
> - Have `speedtest` document Ookla's, but add a note something along the lines of

So IMO either close this PR or revert the changes on `speedtest-cli` and add the documentation for `speedtest`.

owenvoke (approved):
Great work! LGTM. 👍🏻

bl-ue (approved):

pages/common/speedtest.md
```diff
@@ -0,0 +1,33 @@
+# speedtest
+
+> Official command line interface for testing internet bandwidth using https://speedtest.net.
+> Note: some platforms link `speedtest` to `speedtest-cli`.  If some of the examples in this page don't work, try `tldr speedtest-cli`.
```
navarroaxel, line 4:
```suggestion
> Note: some platforms link `speedtest` to `speedtest-cli`. If some of the examples in this page don't work, try `tldr speedtest-cli`.
```

navarroaxel:
I miss the `--json` example for `speedtest-cli` but it's ok.

bl-ue:
Oh yeah, I wonder what happened to it 🤔

marchersimon:
I thought it wasn't too useful and I wanted to keep the number of examples in both pages the same. But I'll add it again.

bl-ue (approved):

pages/common/speedtest.md
```diff
@@ -0,0 +1,33 @@
+# speedtest
+
+> Official command line interface for testing internet bandwidth using https://speedtest.net.
+> Note: some platforms link `speedtest` to `speedtest-cli`. If some of the examples in this page don't work, try `tldr speedtest-cli`.
```
sbrl, line 4: Perhaps this will read better?
```suggestion
> Note: some platforms link `speedtest` to `speedtest-cli`. If some of the examples in this page don't work, see `speedtest-cli`.
```

pages/common/speedtest-cli.md
```diff
@@ -1,32 +1,36 @@
 # speedtest-cli
 
-> Command line interface for testing internet bandwidth using https://speedtest.net.
+> Unofficial command line interface for testing internet bandwidth using https://speedtest.net.
 > More information: <https://github.com/sivel/speedtest-cli>.
```
sbrl, line 4: Perhaps this will read better?
```suggestion
> Unofficial command line interface for testing internet bandwidth using https://speedtest.net.
> See also `speedtest`, the official CLI.
> More information: <https://github.com/sivel/speedtest-cli>.
```

sbrl (changes requested):
Thanks for the edits! I've left some comments below for you to review.

sbrl (approved):
Thanks, @marchersimon!

navarroaxel (approved):

pages/common/speedtest.md
```diff
@@ -0,0 +1,33 @@
+# speedtest
+
+> Official command line interface for testing internet bandwidth using https://speedtest.net.
+> Note: some platforms link `speedtest` to `speedtest-cli`. If some of the examples in this page don't work, see `speedtest-cli`.
```
bl-ue, line 4:
```suggestion
> Note: some platforms link `speedtest` to `speedtest-cli`, which is not the official CLI. If some of the examples in this page don't work, see `speedtest-cli`.
```

pages/common/speedtest-cli.md
```diff
@@ -1,32 +1,37 @@
 # speedtest-cli
 
-> Command line interface for testing internet bandwidth using https://speedtest.net.
+> Unofficial command line interface for testing internet bandwidth using https://speedtest.net.
+> See also `speedtest`, the official CLI.
```
bl-ue, line 4:
```suggestion
> See also `speedtest` for the official CLI.
```

bl-ue (commented):
How about a slight wording update?

pages/common/speedtest.md
navarroaxel: I think is redundant with the first line of the command description.

bl-ue: If you guys think so, that's fine.  A bit of redunancy isn't always a problem, but the section is a little full as it is, so maybe we can skip it.

PR #5805, opened 2021-04-21
cp: sync French translations with English page (remove "--parents")

nicokosi:
Should fix a `fr` warning from https://lukwebsforge.github.io/tldri18n/.

### DIFF
--- a/pages.fr/common/cp.md
+++ b/pages.fr/common/cp.md
@@ -26,7 +26,3 @@
 - Déréférencer les liens symboliques avant de copier :
 
 `cp -L {{link}} {{chemin/vers/répertoire_cible}}`
-
-- Copier en utilisant le chemin complet des fichiers source, en créant les répertoires intermédiaires manquants :
-
-`cp --parents {{source/chemin/vers/fichier}} {{chemin/vers/fichier_cible}}`


marchersimon:
Seems like the `pages.fr/common/cp.md` got translated from `pages/linux/cp.md` instead of `pages/common/cp.md`. The common version uses `-R` and the linux version uses `-r`, but `-R` is also valid in linux. So I think it would be the best to just delete the `linux` page and instead of removing `--parents` from the french translation we should add it to the english version.

marchersimon:
Also all 'cp` pages still use the short form for their arguments, even though there are long forms available for all

bl-ue:
Yeah, a lot of the pages for essential commands still use short options 🤷🏻 they ought to be fixes.

nicokosi:
> Seems like `pages.fr/common/cp.md` got translated from `pages/linux/cp.md` instead of `pages/common/cp.md`. The common version uses `-R` and the linux version uses `-r`, but `-R` is also valid in linux. So I think it would be the best to just delete the `linux` page and instead of removing `--parents` from the french translation we should add it to the english version.

OK for the history but I am not sure it's a good thing to document common options that are not (common). :thought_balloon:

marchersimon:
`--parents` is comon, it just isn't listed in `common/cp.md`. So, the command is the same on all platforms.

nicokosi:
> `--parents` is comon, it just isn't listed in `common/cp.md`. So, the command is the same on all platforms.

Well, not on `macOS`, visibly:
```
$ uname -v
Darwin Kernel Version 20.3.0: Thu Jan 21 00:07:06 PST 2021; root:xnu-7195.81.3~1/RELEASE_X86_64

$ cp --parents a/b/c /tmp/dummyFolder

cp: illegal option -- -
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory
```

marchersimon:
Oh, sorry then. In this case the change seems fine. Thanks.

marchersimon (approved):

nicokosi:
> Also all `cp` pages still use the short form for their arguments, even though there are long forms available for all

You're right. I think it should be done in another pull request (I will *maybe* do it :innocent:).

bl-ue (approved):

PR #5806, opened 2021-04-21
age: add Italian translation

BuriedInTheGround:
- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #5809, opened 2021-04-21
7z, ab, ack: update Italian translations

BuriedInTheGround:

### DIFF
--- a/pages.it/common/7z.md
+++ b/pages.it/common/7z.md
@@ -19,6 +19,10 @@
 
 `7z x {{archivio.7z}} -o{{percorso/alla/directory}}`
 
+- Estrai un archivio sullo standard output:
+
+`7z x {{archivio.7z}} -so`
+
 - Archivia utilizzando uno specifico tipo di archivio:
 
 `7z a -t {{zip|gzip|bzip2|tar}} {{archivio.7z}} {{percorso/a/file_o_directory}}`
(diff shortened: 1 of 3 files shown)


navarroaxel:
These changes where included in the PR #5806 
@BuriedInTheGround you should create a new branch in your fork to avoid add your changes to the `master` branch. You added the changes of both PRs in the same branch.
Let me know if you need help with Git :grin:

bl-ue:
@BuriedInTheGround in order to avoid this (very common) mistake, a good rule of thumb is to never commit directly to `master` on your fork, but to make a new branch from `master` and commit there.  That way you'll keep unrelated commits out of other PRs 👍🏻

BuriedInTheGround:
@navarroaxel @bl-ue 
I'm really sorry! I figured out to have committed to `master` only after pushing, so I then made the new branch. The problem actually is that I didn't know that adding commits to my fork *after* opening the PR would reflect to the PR itself. Can this thing even be prevented in some way?

navarroaxel:
@BuriedInTheGround 
every change pushed or `force pushed` :no_mouth: to a source branch updates the associated PR. Otherwise you can't apply changes requested in a code review.
But if you edit a branch (e.g. `it-updates`), it doesn't affect other PRs from your fork like the another one made from `master`. This happend because you pushed the same commit in both branches.
Don't worry, the next time this will be better.

BuriedInTheGround:
I understand. I thought that new commits wouldn't be added automatically to the PR without some kind of confirmation request. Thanks for the clarification @navarroaxel ! 😄

PR #5140, opened 2021-01-13
multiple pages: normalize "more information" labels

bl-ue:
This PR normalizes the "more information" labels in the all translations of all pages.

I picked each label based on the most common one in the `pages.*` folder. Usually there are 1% or 2% are strange labels and 95%+ are the same, e.g. for German:
```
Mehr Information:                = 2
Mehr Informationen:              = 59
Mehr informationen:              = 2
Weitere Informationen zu Cradle: = 1
Weitere Informationen:           = 21
Weitere informationen:           = 1
```

Labels decided on:

- German (`de`): `Mehr Informationen:`

PR #5814, opened 2021-04-22
German pages: replace `Mehr Informationen` with `Weitere Informationen`

marchersimon:
This is something I wanted to do for quite a while. "Mehr Informationen" as a translation for "More information" is fine, but "Weitere Informationen" feels a bit more natural and professional.

### DIFF
--- a/scripts/set-more-info-link.py
+++ b/scripts/set-more-info-link.py
@@ -8,7 +8,7 @@ labels = {
     'en': 'More information:',
     'bs': 'Više informacija:',
     'da': 'Mere information:',
-    'de': 'Mehr Informationen:',
+    'de': 'Weitere Informationen:',
     'es': 'Más información:',
     'fa': 'اطلاعات بیشتر:',
     'fr': 'Plus d\'informations\xa0:',
--- a/pages.de/common/7z.md
+++ b/pages.de/common/7z.md
@@ -1,7 +1,7 @@
 # 7z
 
 > Ein Dateiarchivierer mit hoher Komprimierungsrate.
-> Mehr Informationen: <https://www.7-zip.org/>.
+> Weitere Informationen: <https://www.7-zip.org/>.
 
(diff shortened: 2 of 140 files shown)


scripts/set-more-info-link.py
```diff
@@ -8,7 +8,7 @@
     'en': 'More information:',
     'bs': 'Više informacija:',
     'da': 'Mere information:',
-    'de': 'Mehr Informationen:',
+    'de': 'Weitere Informationen:',
```
bl-ue, line 11: Good job for remembering 😅

bl-ue (approved):

marchersimon:
> Interesting. thinking
> 
> For reference, here's where I decided to use Mehr: #5140

I asked a few friends of mine and they thought it sounds nicer. However that might differ from area to area. I think I'll ask in some German forum for a definitive answer.

sbrl (approved):
Thanks, @marchersimon!

marchersimon:
I'll convert it to a draft for now to avoid merging until I have a definitive answer.

FantasyCookie17:
I'd say that they both sound equally fine, and "mehr" is shorter…

marchersimon:
My findings so far:
* According to [Google Trends](https://trends.google.com/trends/explore?geo=DE&q=mehr%20Informationen,weitere%20Informationen), `Weitere Informationen` seems to be more popular.
* On [r/German](https://www.reddit.com/r/German/comments/mwr0hh/mehr_informationen_or_weitere_informationen/), I got two interesting answers:

> "Mehr" alludes to missing information. So if you need to know more you need mehr Informationen. If you have already provided sufficient information but want to encourage learning more you can link to weitere Informationen.

> To me, "mehr" implies quantity, and "further" implies more-in-depth.

bl-ue:
That sounds great.  I think we have our answer.  I'll merge now, given that it appears to be done.

PR #5823, opened 2021-04-23
acme.sh: add page

peterbabic:

### DIFF
--- a/pages/common/acme.sh.md
+++ b/pages/common/acme.sh.md
@@ -1,6 +1,6 @@
 # acme.sh
 
-> An ACME Shell script, an acme client alternative to certbot.
+> Shell script implementing ACME client protocol, an alternative to certbot.
 > More information: <https://github.com/acmesh-official/acme.sh>.
 
 - Issue a certificate using webroot mode:
--- a/pages/common/acme.sh.md
+++ b/pages/common/acme.sh.md
@@ -33,4 +33,4 @@
 
 - Install certificate file into location specified by its parameter (not displayed):
 
-`acme.sh --install-cert --domain {{example.com}} --reloadcmd {{reload_commad}}`
+`acme.sh --install-cert --domain {{example.com}} --reloadcmd {{reload_command}}`
--- a/pages/common/acme.sh.md
+++ b/pages/common/acme.sh.md
@@ -1,0 +1,36 @@
+# acme.sh
+
+> An ACME Shell script, an acme client alternative to certbot.
+> More information: <https://github.com/acmesh-official/acme.sh>
+
+- Issue a certificate using webroot mode:
+
+`acme.sh --issue --domain {{domain.tld}} --webroot {{/path/to/webroot}}`
+
+- Issue a certificate using standalone mode using port 80:
+
+`acme.sh --issue --standalone --domain {{domain.tld}}`
+
+- Issue a certificate using standalone TLS mode using port 443:
+
+`acme.sh --issue --alpn --domain {{domain.tld}}`
+
+- Issue a certificate using a working Nginx configuration:
+
+`acme.sh --issue --nginx --domain {{domain.tld}}`
+
+- Issue a certificate using a working Apache configuration:
+
+`acme.sh --issue --apache --domain {{domain.tld}}`
+
+- Issue a wildcard (\*) certificate using a manual DNS mode:
+
+`acme.sh --issue --dns --domain {{*.domain.tld}}`
+
+- Issue a certificate using an automatic DNS API mode:
+
+`acme.sh --issue --dns {{dns_api}} --domain {{domain.tld}}`
+
+- Install certificate file into location specified by its parameter (not displayed):
+
+`acme.sh --install-cert --domain {{domain.tld}} --reloadcmd {{reload_commad}}`
--- a/pages/common/acme.sh.md
+++ b/pages/common/acme.sh.md
@@ -31,6 +31,6 @@
 
 `acme.sh --issue --dns {{dns_cf}} --domain {{example.com}}`
 
-- Install certificate file into location specified by its parameter (not displayed):
+- Install certificate files into specified locations, useful for automatic certificate renewal:
 
 `acme.sh --install-cert -d {{example.com}} --key-file {{/path/to/example.com.key}} --fullchain-file {{/path/to/example.com.cer}} --reloadcmd {{"systemctl force-reload nginx"}}`
--- a/pages/common/acme.sh.md
+++ b/pages/common/acme.sh.md
@@ -1,7 +1,7 @@
 # acme.sh
 
 > An ACME Shell script, an acme client alternative to certbot.
-> More information: <https://github.com/acmesh-official/acme.sh>
+> More information: <https://github.com/acmesh-official/acme.sh>.
 
 - Issue a certificate using webroot mode:
 
--- a/pages/common/acme.sh.md
+++ b/pages/common/acme.sh.md
@@ -5,32 +5,32 @@
 
 - Issue a certificate using webroot mode:
 
-`acme.sh --issue --domain {{domain.tld}} --webroot {{/path/to/webroot}}`
+`acme.sh --issue --domain {{example.com}} --webroot {{/path/to/webroot}}`
 
 - Issue a certificate using standalone mode using port 80:
 
-`acme.sh --issue --standalone --domain {{domain.tld}}`
+`acme.sh --issue --standalone --domain {{example.com}}`
 
 - Issue a certificate using standalone TLS mode using port 443:
 
-`acme.sh --issue --alpn --domain {{domain.tld}}`
+`acme.sh --issue --alpn --domain {{example.com}}`
 
 - Issue a certificate using a working Nginx configuration:
 
-`acme.sh --issue --nginx --domain {{domain.tld}}`
+`acme.sh --issue --nginx --domain {{example.com}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/acme.sh.md
```diff
@@ -0,0 +1,36 @@
+# acme.sh
+
+> An ACME Shell script, an acme client alternative to certbot.
+> More information: <https://github.com/acmesh-official/acme.sh>
+
+- Issue a certificate using webroot mode:
+
+`acme.sh --issue --domain {{domain.tld}} --webroot {{/path/to/webroot}}`
```
marchersimon, line 8: Does this path have to be asolute?

pages/common/acme.sh.md
```diff
@@ -0,0 +1,36 @@
+# acme.sh
+
+> An ACME Shell script, an acme client alternative to certbot.
```
bl-ue, line 3: This is more verbatim from the repository.
```suggestion
> Shell script implementing ACME client protocol, alternative to certbot.
```

bl-ue:
@peterbabic what do you think about using `example.com` instead of `domain.tld`?  While the latter is perfectly good, the former looks more like a url at first glance.

pages/common/acme.sh.md
```diff
@@ -0,0 +1,36 @@
...
+
+`acme.sh --issue --dns --domain {{*.domain.tld}}`
+
+- Issue a certificate using an automatic DNS API mode:
+
+`acme.sh --issue --dns {{dns_api}} --domain {{domain.tld}}`
+
+- Install certificate file into location specified by its parameter (not displayed):
+
+`acme.sh --install-cert --domain {{domain.tld}} --reloadcmd {{reload_commad}}`
```
peterbabic, line 36: This would ideally contain a far longer code snippet, like:

```
acme.sh --install-cert -d {{domain.tld}}  \
  --key-file {{/etc/nginx/ssl/domain.tld.key}} \
  --fullchain-file {{/etc/nginx/ssl/domain.tld.cer}} \
  --reloadcmd {{"systemctl force-reload nginx"}}
```

But I am not sure how to proceed. All the parameters, except the `--domain` are optional. However, there are many parameters in a format `--*-file`handling their respective certificate key files. Without a single *-file parameter, the whole command does literally nothing.

Additionally, `--reloadcmd` is also optional, but it is run without it only rarely (in the situations that the service restart is done completely manually, bypassing the whole point of automatic certificate renewal mechanism).

I am not entirely sure how to proceed, but this single code block is the only one I would've looked for when calling tldr acme.sh.

pages/common/acme.sh.md
peterbabic: @marchersimon  I believe that they have to be absolute, more info https://github.com/acmesh-official/acme.sh/issues/114

pages/common/acme.sh.md
peterbabic: Nice, also what about *an* alternative to certbot?

peterbabic:
> @peterbabic what do you think about using `example.com` instead of `domain.tld`? While the latter is perfectly good, the former looks more like a url at first glance.

```
$ grep -o -i -R "example.com" | wc -l 
216
```

On the contrary:

```
$ grep -o -i -R "domain.tld" | wc -l 
0
```

I agree with `example.com`

pages/common/acme.sh.md
```diff
@@ -0,0 +1,36 @@
...
+
+`acme.sh --issue --apache --domain {{example.com}}`
+
+- Issue a wildcard (\*) certificate using a manual DNS mode:
+
+`acme.sh --issue --dns --domain {{example.com}}`
+
+- Issue a certificate using an automatic DNS API mode:
+
+`acme.sh --issue --dns {{dns_api}} --domain {{example.com}}`
```
sbrl, line 32: Perhaps a concrete example for `dns_api` would aid clarity here?

peterbabic: @sbrl sure.

Instead of `{{dns_api}}` would be one of the many available options specified here https://github.com/acmesh-official/acme.sh/wiki/dnsapi

So, for instance for CloudFlare DNS API it is `{{dns_cf}}`

Also, in example used `{{dns_api}}` is still perfectly valid for custom scripts, if there is a `dns_api.sh` script present in the `.acme.sh/` root folder, more info https://github.com/acmesh-official/acme.sh/issues/1571#issuecomment-384999814

pages/common/acme.sh.md
bl-ue: Updated 👍🏻

pages/common/acme.sh.md
peterbabic: @bl-ue @marchersimon @sbrl please, can anyone provide any kind of feedback?

Should this stay as it is, or is there any way to display long command? I did not find anything "long" related in the contribution guide.

bl-ue: This seems a little long, but it might not be... 🤔 
```markdown
`acme.sh --install-cert -d {{example.com}} --key-file {{/path/to/example.com.key}} --fullchain-file {{/path/to/example.com.cer}} --reloadcmd {{"systemctl force-reload nginx"}}`
```

marchersimon: I also think we should try to stuff it into one line. Mainly because most clients don't support multiline code.

peterbabic: Guys please, I am not sure what to do here. Would appreciate some guidance, if I am expected to commit some changes.

pages/common/acme.sh.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Install certificate file into location specified by its parameter (not displayed):
+
+`acme.sh --install-cert --domain {{example.com}} --reloadcmd {{reload_commad}}`
```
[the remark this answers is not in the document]
peterbabic: I have shortened it especially because it could become too long, as we are still discussing above.
As far as I can tell right now, we have at least three options:
1. drop the example as it is too long (I am strongly opposed, but might be sufficient)
2. include relatively short example with generic values (would serve more as a reminder)
3. include full length example (most valuable, might break some clients without multiline code support)

pages/common/acme.sh.md
peterbabic: 4. add the subcommand-like page, detailing the most used `--install-cert` options, as there are quite a few of them and mention this page in See also section

pages/common/acme.sh.md
peterbabic: Maybe this would also deserved it's separate subcommand page?

bl-ue: But is it actually a subcommand?  If so, then if should just be `dns_api.sh.md` (if we document it, which we probably will).

pages/common/acme.sh.md
bl-ue: Why not just stuff it into one line?  It's a little long, but I don't really see any strong reason why not to.  As for multiline, we can't do that, because that goes against the client spec (or at least the client spec will have to be updated).

pages/common/acme.sh.md
peterbabic: I believe it would be called more like `acme.sh-dns.md`, as the subcommand in question is named `--dns {{ }}`

pages/common/acme.sh.md
peterbabic: I support any idea of documenting the `--install-cert` option, as it has around 5-7 commonly used parameters.

Thinking about it further, the name of the file could probably be called `acme.sh-install-cert.md` similarly as for instance `git-cat-file.md`

Would this somehow break existing conventions?

sbrl: For some commands, it's unavoidable to have long commands - just do `tldr whiptail` and you'll see what I mean. It's not necessarily a _bad_ thing. I say put it all on 1 line.

peterbabic:
So I have added specific examples according to valuable suggestions with the possibility to document subcommands and maybe even open "Lets Document: acme.sh" issue in the future.

pages/common/acme.sh.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Issue a wildcard (\*) certificate using a manual DNS mode:
+
+`acme.sh --issue --dns --domain {{example.com}}`
+
+- Issue a certificate using an automatic DNS API mode:
+
+`acme.sh --issue --dns {{dns_cf}} --domain {{example.com}}`
+
+- Install certificate file into location specified by its parameter (not displayed):
```
peterbabic, line 34: Maybe I should reword this too.

pages/common/acme.sh.md
```diff
@@ -31,6 +31,6 @@
 
 `acme.sh --issue --dns {{dns_cf}} --domain {{example.com}}`
 
-- Install certificate file into location specified by its parameter (not displayed):
+- Install certificate files into specified locations, useful for automatic certificate renewal:
```
marchersimon, line 34:
```suggestion
- Install certificate files into the specified locations (useful for automatic certificate renewal):
```

marchersimon (approved):

Waples (approved):

pages/common/acme.sh.md
sbrl: If we document it as a subcommand, we'd probably do something similar to pacman in #5268, for example. It sounds like the DNS option here is more complex than I thought, so let's leave this particular example as-is.

peterbabic: @sbrl it is already open in https://github.com/tldr-pages/tldr/pull/5852#issuecomment-830626806

PR #5836, opened 2021-04-25
README: add Outfieldr Zig client

ve-nt:

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ both for the command line and for other platforms:
 - [Node.js client](https://github.com/tldr-pages/tldr-node-client):
   `npm install -g tldr`
 - [OCaml client](https://github.com/RosalesJ/tldr-ocaml): `opam install tldr`
+- [Outfieldr](https://gitlab.com/ve-nt/outfieldr): A Zig client
 - [Perl5 client](https://github.com/skaji/perl-tldr):
   `cpanm App::tldr`
 - [PHP client](https://github.com/BrainMaestro/tldr-php):


README.md
```diff
@@ -148,6 +148,7 @@ both for the command line and for other platforms:
     - [DistroWatch](https://distrowatch.com/dwres.php?resource=man-pages)
     - [tldr.ooops.me](https://tldr.ooops.me): web client with multilingual support
     - [TLDR Persian](https://opoet7.github.io/tldr-persian/): Web Client in Persian
+- [Zig client](https://gitlab.com/ve-nt/outfieldr)
```
navarroaxel, line 151: You choose a cool name for your client, I think you should use it
```suggestion
    - [Outfieldr](https://gitlab.com/ve-nt/outfieldr): A Zig client
```

ve-nt: Thanks! I agree with this, although I don't think the whitespace before the checkmark is intentional?

ve-nt:
Woops, just realized with the name change, it should be in a different part of the list to keep the alphabetical order. Fix incoming

bl-ue:
>I've marked the above checklist as done, although they don't apply to this PR (I don't think).

You're certainly correct @ve-nt.  In that case, you can remove them (as I've done).

ve-nt:
Alright, I think everything is fixed and ready to go. Changed the commit title to conform to the repo guidlines as well.

bl-ue (approved):
Awesome!  I'd love to try it out; and it may become my daily client.

I'm pleasantly (very) surprised to read the following from your README:

>It runs ~20-30 times faster than [Tealdeer](https://github.com/dbrgn/tealdeer), which is written in Rust and claims to be faster than the rest. For this, Tealdeer was built with `--release`, and Outfieldr was built with `-Drelease-fast`.  It's worth noting that building with `-Drelease-safe` will give very similar results, only adding 0.1ms extra here or there.

Also, looking at the source code, Zig looks **really** interesting.  I like all of the coding conventions used and the various syntaxes to make common things (such as `try`/`catch`) simpler 🤔

marchersimon (approved):

README.md
bl-ue: No that white space is not intentional ;)

navarroaxel (approved):
Thanks @ve-nt, please remember don't push force to the PR.

ve-nt:
> Thanks @ve-nt, please remember don't push force to the PR.

Oh okay, sorry I didn't realize. I force pushed fixups to keep it in one commit. I'll remember not to in the future :+1:

bl-ue:
Yeah, that's because we auto-squash just before merging.  Maybe we should add this to the contributing guidelines 🤔

ve-nt:
> Awesome! I'd love to try it out; and it may become my daily client.

Awesome! I'd be honored :D. If you have trouble building (particularly boostrapping Gyro) just know that I'm planning on reworking the build system tonight to remove the dependency on Gyro, as other people have had issues with it.

> Also, looking at the source code, Zig looks **really** interesting. I like all of the coding conventions used and the various syntaxes to make common things (such as `try`/`catch`) simpler thinking

Yeah the error handling is great. The `try`/`catch` thing confuses some people because it makes them think of exceptions. Note that Zig doesn't use exceptions, the way it handles errors is very similar to Rust's `Option<T>` and `Result<T, E>`

bl-ue:
Cool @ve-nt; I'll keep that in mind when I study Zig :D

PR #5847, opened 2021-04-27
virsh-undefine: add page

aherst:

### DIFF
--- a/pages/common/virsh-undefine.md
+++ b/pages/common/virsh-undefine.md
@@ -13,4 +13,4 @@
 
 - Delete the configuration file and the specified storage volumes using the target name or the source name (as obtained from the `virsh domblklist` command):
 
-`virsh undefine --domain {{vm_name}} --storage {{target_name1|path/to/storage1,target_name2|path/to/storage2,...}}`
+`virsh undefine --domain {{vm_name}} --storage {{sda,path/to/storage}}`
--- a/pages/common/virsh-undefine.md
+++ b/pages/common/virsh-undefine.md
@@ -11,6 +11,6 @@
 
 `virsh undefine --domain {{vm_name}} --remove-all-storage`
 
-- Delete the configuration file and the specified storage volumes:
+- Delete the configuration file and the specified storage volumes using the target name or the source name (as obtained from the `virsh domblklist` command):
 
 `virsh undefine --domain {{vm_name}} --storage {{target_name1|path/to/storage1,target_name2|path/to/storage2,...}}`
--- a/pages/common/virsh-undefine.md
+++ b/pages/common/virsh-undefine.md
@@ -1,0 +1,16 @@
+# virsh-undefine
+
+> Delete a virtual machine.
+> More information: <https://manned.org/virsh>.
+
+- Delete only the virtual machine configuration file:
+
+`virsh undefine --domain {{vm_name}}`
+
+- Delete the configuration file and all associated storage volumes:
+
+`virsh undefine --domain {{vm_name}} --remove-all-storage`
+
+- Delete the configuration file and the specified storage volumes:
+
+`virsh undefine --domain {{vm_name}} --storage {{target_name1|path/to/storage1,target_name2|path/to/storage2,...}}`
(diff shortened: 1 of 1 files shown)


bl-ue (approved):

pages/common/virsh-undefine.md
```diff
@@ -0,0 +1,16 @@
...
+
+`virsh undefine --domain {{vm_name}}`
+
+- Delete the configuration file and all associated storage volumes:
+
+`virsh undefine --domain {{vm_name}} --remove-all-storage`
+
+- Delete the configuration file and the specified storage volumes:
+
+`virsh undefine --domain {{vm_name}} --storage {{target_name1|path/to/storage1,target_name2|path/to/storage2,...}}`
```
marchersimon, line 16: The argument for `--storage` looks confusing here. How about `target_name_or_path/to/storage1, target_name_or_path/to/storage2, ...`?

Waples: Also the `--domain` flag seems to not be present in the `man` page for it :thinking: rather it's just used as an argument
```   undefine
       Syntax:

          undefine domain [--managed-save] [--snapshots-metadata]
             [--checkpoints-metadata] [--nvram] [--keep-nvram]
             [ {--storage volumes | --remove-all-storage
                [--delete-storage-volume-snapshots]} --wipe-storage]
```

the example from the `man` page states:
```Example:
    --storage vda,/path/to/storage.img
```
so I'm also a bit confused about the `|` in between `{{target_name1|path/to/storage1...}}`

Context:
```bash
  virsh --version
    7.1.0
```

Waples (changes requested):

pages/common/virsh-undefine.md
marchersimon: > I'm also a bit confused about the `|` in between `{{target_name1|path/to/storage1...}}`

You can specifc as many storages as you want and you can specify each as `path` or `target_name`. In the man page, the first one is specified as `target_name`, the second one as `path`.

aherst: The --domain flag is optional. A virsh undefine -- and then two TABs on the command line gives:

```
--checkpoints-metadata             --nvram
--delete-snapshots                 --remove-all-storage
--delete-storage-volume-snapshots  --snapshots-metadata
--domain                           --storage
--keep-nvram                       --wipe-storage
--managed-save 
```                    

While the example will work without the --domain flag, I've included it for completeness.

Waples: @marchersimon so why `|` instead of `,` :thinking:

marchersimon: Because instead of writing the first option for the first argument and the second option for the second argument, we write both options for botth arguments. So it's basically `a|b,a|b`, instead of `a,b`

bl-ue: Yeah @Waples, I agree that it's hard on the eyes.  I had to stare at it myself before I could understand it.

`{{target_name1|path/to/storage1,target_name2|path/to/storage2}}` is saying that it can be any number of `target_name`s **_OR_** `path/to/file`s.  So these are all valid:

```
target_name1,target_name2
path/to/storage1,path/to/storage2
target_name1,path/to/storage2
path/to/storage1,target_name2
...
```
Does that make sense?

bl-ue (approved):

pages/common/virsh-undefine.md
Waples: The man page specifies that it's a comma seperated list, hence why I'm pondering over the `|` instead of `,`; the example must be true to it's form.

aherst: How about if I re-wrote the example as:

```
- Delete the configuration file and the two storage volumes specified using the target name and the source name (as obtained from the `virsh domblklist` command):

`virsh undefine --domain {{vm_name}} --storage {{sda,path/to/source}}`
```

marchersimon: I'd prefer
```
- Delete the configuration file and the specified storage volumes using the target name or the source name (as obtained from the `virsh domblklist` command):

`virsh undefine --domain {{vm_name}} --storage {{sda,path/to/source}}`
```

aherst:
Do we have a consensus on whether or not I should include the --domain flag?

marchersimon:
If it works and makes the command esier to understand, I don't see a reason to remove it.

Waples (approved):

PR #5850, opened 2021-04-28
qutebrowser: add page

Waples:

### DIFF
--- a/pages/common/qutebrowser.md
+++ b/pages/common/qutebrowser.md
@@ -9,7 +9,7 @@
 
 - Open a qutebrowser instance with temporary settings:
 
-`qutebrowser --set {{option}} {{value}}`
+`qutebrowser --set {{content.geolocation}} {{true|false}}`
 
 - Restore a named session of a qutebrowser instance:
 
--- a/pages/common/qutebrowser.md
+++ b/pages/common/qutebrowser.md
@@ -19,6 +19,6 @@
 
 `qutebrowser --target {{auto|tab|tab-bg|tab-silent|tab-bg-silent|window|private-window}}`
 
-- Open qutebrowser with a temporary base directory with stdout logging in json format:
+- Open qutebrowser with a temporary base directory and print logs to stdout as JSON:
 
 `qutebrowser --temp-basedir --json-logging`
--- a/pages/common/qutebrowser.md
+++ b/pages/common/qutebrowser.md
@@ -1,0 +1,24 @@
+# qutebrowser
+
+> A keyboard-driven, vim-like browser based on PyQt5.
+> More information: <https://qutebrowser.org/>.
+
+- Open qutebrowser with a specified (non-default) storage directory:
+
+`qutebrowser --basedir {{base_directory}}`
+
+- Open a qutebrowser instance with temporary settings:
+
+`qutebrowser --set {{option}} {{value}}`
+
+- Restore a named session of a qutebrowser instance:
+
+`qutebrowser --restore {{session_name}}`
+
+- Choose how to open URLs if there is already a qutebrowser instance running:
+
+`qutebrowser --target {{auto|tab|tab-bg|tab-silent|tab-bg-silent|window|private-window}}`
+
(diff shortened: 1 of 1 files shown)


pages/common/qutebrowser.md
```diff
@@ -0,0 +1,24 @@
+# qutebrowser
+
+> A keyboard-driven, vim-like browser based on PyQt5.
+> More information: <https://qutebrowser.org/>.
+
+- Open qutebrowser with a specified (non-default) storage directory:
+
+`qutebrowser --basedir {{base_directory}}`
```
bl-ue, line 8: "non-default" doesn't seem necessary to add here. Is it?
```suggestion
- Open qutebrowser with a specified storage directory:

`qutebrowser --basedir {{path/to/directory}}`
```
Also, `base_directory` in the variable name doesn't seem to give it much more value than with omitting it 🤔

pages/common/qutebrowser.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Open a qutebrowser instance with temporary settings:
+
+`qutebrowser --set {{option}} {{value}}`
```
bl-ue, line 12: Could you add an example of listing the available options?  Without that, this example might not be very useful.

pages/common/qutebrowser.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Restore a named session of a qutebrowser instance:
+
+`qutebrowser --restore {{session_name}}`
```
bl-ue, line 16: Same here.  How can one access a list of valid session names?

pages/common/qutebrowser.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Choose how to open URLs if there is already a qutebrowser instance running:
```
bl-ue, line 18: Is this correct?  I'm trying to avoid using "choose" because that sounds slightly ambiguous.
```suggestion
- Launch qutebrowser, opening all links using the specified method:
```

pages/common/qutebrowser.md
```diff
@@ -0,0 +1,24 @@
...
+
+`qutebrowser --target {{auto|tab|tab-bg|tab-silent|tab-bg-silent|window|private-window}}`
+
+- Open qutebrowser with a temporary base directory with stdout logging in json format:
```
bl-ue, line 22:
```suggestion
- Open qutebrowser with a temporary base directory and print logs to stdout as JSON:
```

pages/common/qutebrowser.md
```diff
@@ -0,0 +1,24 @@
+# qutebrowser
+
+> A keyboard-driven, vim-like browser based on PyQt5.
```
bl-ue, line 3: I can't tell if I like this or not.  At first glance, the sentences seems a bit jumbled without the comma.  Two more seconds of looking at it though and it makes sense.  🤔🤔 
```suggestion
> A keyboard-driven, vim-like browser, based on PyQt5.
```

bl-ue (changes requested):

pages/common/qutebrowser.md
Waples: It's the literal description given by the project, so I just pasted it over. But like you said, read it twice it works better than with the second comma.

pages/common/qutebrowser.md
Waples: Why `links` instead of `URLs`, just a question.

pages/common/qutebrowser.md
Waples: As far as I'm aware of, there is not an argument to list (at time of writing, might open a PR there..) saved named sessions. These can be found in the browser itself (`:session-load {tab}`) via the session loading command.

pages/common/qutebrowser.md
bl-ue: No particular reason.  I didn't even mean to do that.  Changed back.

marchersimon (approved):

bl-ue (approved):

bl-ue:
Nice job @Waples! Looks good to me.

Issue #5791, opened 2021-04-19
Style question: What about variable number of arguments (varargs)?


xeruf:
In some cases, one or multiple (or even zero) of an argument can be passed, should we use any specific syntax here? Didn't find anything on https://github.com/tldr-pages/tldr/blob/master/contributing-guides/style-guide.md#token-syntax.

I was thinking of the classical `...` which I started using in a PR already: https://github.com/xerus2000/tldr/commit/ab754537fee0edb609c6b487ef6c009a5532148a
The question then is whether to use singular or plural, i.e. `files...` or `file...`.

bl-ue:
I made a suggestion here: https://github.com/tldr-pages/tldr/pull/5787#discussion_r615845999

As a matter of fact, I very recently brought this question up: https://github.com/tldr-pages/tldr/pull/5768#discussion_r614397719. My preferred style would be like this:

```
command {{path/to/file1 path/to/file2 ...}}
```

Multiple items being inside one token syntax section like that indicates that the position and count of the arguments is variable.

Waples:
I'm siding beside @bl-ue.
just stuff them in there.

sbrl:
Personally I've used `{{thing1}} {{thing2}} {{thingN}}` myself, but I don't think we need a hard and fast rule here as it's dependent on context. `{{thing1 thing2 ...}}` would also be acceptable, as would `{{thing1|thing2|...}}` for a choice of only 1 option.

bl-ue:
I prefer my suggestion because it looks shorted to me than the 1, 2, N approach.

xeruf:
> I prefer my suggestion because it looks shorted to me than the 1, 2, N approach.

I agree - using N might incline some users to ask what the N stands for

marchersimon:
What do yout think about making an issue or a file, where we collect all conventions we decided upon?

sbrl:
@marchersimon Maybe, but I think it's a bit too dependent on context to make a hard and fast rule about it.

marchersimon:
Seems fair. After all it isn't bad to re-discuss some topics from time to time.

xeruf:
y'all know that there already a file for conventions? ;) https://github.com/tldr-pages/tldr/blob/main/contributing-guides/style-guide.md#token-syntax

sbrl:
Yeah, but we generally speaking try to keep that guide relatively short @xeruf to avoid exhausting new contributors :P

xeruf:
[...]

PR #5852, opened 2021-04-28
acme.sh-dns: add page; acme.sh: remove example

peterbabic:
The wildcard went completely missing somewhere before the merge. In the original design, wildcard certificate usage was about to be referenced with a manual mode example. Running a manual mode with the current example would not produce a wildcard certificate because of the missing \*. subdomain. Furthermore, it would not do anything because a manual DNS mode requires a sort of confirmation command, more info acmesh-official/acme.sh#1029. Because of the many quirks of the various DNS modes, a separate subcommand page `acme.sh dns` was created.

### DIFF
--- a/pages/common/acme.sh-dns.md
+++ b/pages/common/acme.sh-dns.md
@@ -1,0 +1,25 @@
+# acme.sh --dns
+
+> Use a DNS-01 challenge to issue a certificate.
+> DNS API mode documentation: <https://github.com/acmesh-official/acme.sh/wiki/dnsapi>
+> More information: <https://github.com/acmesh-official/acme.sh/wiki/DNS-manual-mode>.
+
+- Issue a certificate using an automatic DNS API mode:
+
+`acme.sh --issue --dns {{gnd_gd}} --domain {{example.com}}`
+
+- Issue a wildcard (\*) certificate using an automatic DNS API mode:
+
+`acme.sh --issue --dns {{dns_namesilo}} --domain {{*.example.com}}`
+
+- Issue a certificate using a DNS alias mode:
+
+`acme.sh --issue --dns {{dns_cf}} --domain {{example.com}} --challenge-alias {{alias-for-example-validation.com}}`
+
+- Issue a certificate while disabling an automatic Cloudflare or Google DNS polling after the DNS record is added by specifying a manual wait time:
+
+`acme.sh --issue --dns {{dns_namecheap}} --domain {{example.com}} --dnssleep 300`
+
+- Issue a certificate using a manual DNS mode:
+
+`acme.sh --issue --dns --domain {{example.com}} --yes-I-know-dns-manual-mode-enough-go-ahead-please`
--- a/pages/common/acme.sh-dns.md
+++ b/pages/common/acme.sh-dns.md
@@ -16,7 +16,7 @@
 
 `acme.sh --issue --dns {{dns_cf}} --domain {{example.com}} --challenge-alias {{alias-for-example-validation.com}}`
 
-- Issue a certificate while disabling an automatic Cloudflare or Google DNS polling after the DNS record is added by specifying a manual wait time:
+- Issue a certificate while disabling automatic Cloudflare / Google DNS polling after the DNS record is added by specifying a custom wait time in seconds:
 
 `acme.sh --issue --dns {{dns_namecheap}} --domain {{example.com}} --dnssleep {{300}}`
 
--- a/pages/common/acme.sh-dns.md
+++ b/pages/common/acme.sh-dns.md
@@ -1,6 +1,6 @@
 # acme.sh --dns
 
-> Use a DNS-01 challenge to issue a certificate.
+> Use a DNS-01 challenge to issue a TLS certificate.
 > DNS API mode documentation: <https://github.com/acmesh-official/acme.sh/wiki/dnsapi>.
 > More information: <https://github.com/acmesh-official/acme.sh/wiki/DNS-manual-mode>.
 
@@ -18,7 +18,7 @@
 
 - Issue a certificate while disabling an automatic Cloudflare or Google DNS polling after the DNS record is added by specifying a manual wait time:
 
-`acme.sh --issue --dns {{dns_namecheap}} --domain {{example.com}} --dnssleep 300`
+`acme.sh --issue --dns {{dns_namecheap}} --domain {{example.com}} --dnssleep {{300}}`
 
 - Issue a certificate using a manual DNS mode:
 
--- a/pages/common/acme.sh-dns.md
+++ b/pages/common/acme.sh-dns.md
@@ -8,7 +8,7 @@
 
 `acme.sh --issue --dns {{gnd_gd}} --domain {{example.com}}`
 
-- Issue a wildcard (\*) certificate using an automatic DNS API mode:
+- Issue a wildcard certificate (denoted by an asterisk) using an automatic DNS API mode:
 
 `acme.sh --issue --dns {{dns_namesilo}} --domain {{*.example.com}}`
 
--- a/pages/common/acme.sh-dns.md
+++ b/pages/common/acme.sh-dns.md
@@ -10,7 +10,7 @@
 
 - Issue a wildcard certificate (denoted by an asterisk) using an automatic DNS API mode:
 
-`acme.sh --issue --dns {{dns_namesilo}} --domain {{*.example.com}}`
+`acme.sh --issue --dns {{dns_namesilo}} --domain  {{example.com}} --domain {{*.example.com}}`
 
 - Issue a certificate using a DNS alias mode:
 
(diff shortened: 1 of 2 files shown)


pages/common/acme.sh.md
```diff
@@ -23,13 +24,9 @@
...
+- Issue a wildcard (\*) certificate using an automatic DNS API mode:
 
-`acme.sh --issue --dns --domain {{example.com}}`
-
-- Issue a certificate using an automatic DNS API mode:
-
-`acme.sh --issue --dns {{dns_cf}} --domain {{example.com}}`
+`acme.sh --issue --dns {{dns_cf}} --domain {{*.example.com}}`
 
 - Install certificate files into the specified locations (useful for automatic certificate renewal):
```
peterbabic, line 31: Btw, shouldn't `the` in `the specified locations` be omitted?

bl-ue: @peterbabic no, I don't see any reason to.  The text can be _slightly_ informal—articles and contractions are usually the most we do.

Waples (approved):

peterbabic:
Guys, additionally, should we also mention custom scripts, I have described in https://github.com/tldr-pages/tldr/pull/5823#discussion_r619343384

It would almost certainly require referencing a path in the example description and I am not sure if we do that. Without the actual path the example does not make much sense.

marchersimon:
I don't think that should be a problem.

sbrl:
In tldr pages we generally try to only document the most common functionality in a page - it's an exercise in determining what's most important. So while custom scripts might be a very useful feature, it sounds like advanced functionality that doesn't necessarily need documenting here. Perhaps it would be more useful in a long-form guide or tutorial rather than a concise set of examples like tldr pages?

pages/common/acme.sh-dns.md
```diff
@@ -0,0 +1,25 @@
+# acme.sh --dns
+
+> Use a DNS-01 challenge to issue a certificate.
```
sbrl, line 3: Perhaps this will read better?
```suggestion
> Use a DNS-01 challenge to issue a TLS certificate.
```

pages/common/acme.sh-dns.md
```diff
@@ -0,0 +1,25 @@
...
+> DNS API mode documentation: <https://github.com/acmesh-official/acme.sh/wiki/dnsapi>.
+> More information: <https://github.com/acmesh-official/acme.sh/wiki/DNS-manual-mode>.
+
+- Issue a certificate using an automatic DNS API mode:
+
+`acme.sh --issue --dns {{gnd_gd}} --domain {{example.com}}`
+
+- Issue a wildcard certificate (denoted by an asterisk) using an automatic DNS API mode:
+
+`acme.sh --issue --dns {{dns_namesilo}} --domain {{*.example.com}}`
```
sbrl, line 13: It's good practice here to also request that the certificate is valid for the root domain. With `certbot`, that would be like so:
```suggestion
`acme.sh --issue --dns {{dns_namesilo}} --domain  {{example.com}} --domain {{*.example.com}}`
```

Does this work for `acme.sh` too?

pages/common/acme.sh-dns.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Issue a certificate using a DNS alias mode:
+
+`acme.sh --issue --dns {{dns_cf}} --domain {{example.com}} --challenge-alias {{alias-for-example-validation.com}}`
+
+- Issue a certificate while disabling an automatic Cloudflare or Google DNS polling after the DNS record is added by specifying a manual wait time:
+
+`acme.sh --issue --dns {{dns_namecheap}} --domain {{example.com}} --dnssleep 300`
```
sbrl, line 21: Missing token syntax?
```suggestion
`acme.sh --issue --dns {{dns_namecheap}} --domain {{example.com}} --dnssleep {{300}}`
```

pages/common/acme.sh-dns.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Issue a wildcard certificate (denoted by an asterisk) using an automatic DNS API mode:
+
+`acme.sh --issue --dns {{dns_namesilo}} --domain {{*.example.com}}`
+
+- Issue a certificate using a DNS alias mode:
+
+`acme.sh --issue --dns {{dns_cf}} --domain {{example.com}} --challenge-alias {{alias-for-example-validation.com}}`
+
+- Issue a certificate while disabling an automatic Cloudflare or Google DNS polling after the DNS record is added by specifying a manual wait time:
```
sbrl, line 19: Perhaps this will read better?
```suggestion
- Issue a certificate while disabling polling after the DNS record is added by specifying a custom wait time in seconds:
```

sbrl (changes requested):
Thanks for the edit and new page! I've left some comments below for you to review.

peterbabic:
@sbrl Agreed.

@marchersimon above + can you provide an existing example/discussion of a good way to insert a path into the page, for a possible future use (only if you do not need to search for it for too long).

pages/common/acme.sh-dns.md
peterbabic: Of course it does. Multiple domains support is available, I just did not find a place to document it nicely on in https://github.com/tldr-pages/tldr/pull/5823

As a side note, I have a wildcard for subdomains that does not share a certificate with the root domain (not residing on the same VPS), but I agree that is more of an edge case. Combining them both into the one certificate is probably fairly common practice.

pages/common/acme.sh-dns.md
peterbabic: It does specifically poll mentioned service providers. Some users consider this behavior a privacy concern. Without mentioning the actual names, it is not really clear what is the purpose. 

What about mentioning the word "privacy" at the very least?

sbrl: Oh, I see. What about this then?
```suggestion
- Issue a certificate while disabling automatic Cloudflare / Google DNS polling after the DNS record is added by specifying a custom wait time in seconds:
```

sbrl:
Generally the convention with file paths is:

`path/to/file` is preferred, with the preferred extension after it if appropriate - e.g. `path/to/file.txt`

If the path *must* be absolute, then prepend a slash (the word `absolute` is suggested but not required): `/absolute/path/to/file.txt`

Generally speaking we drop the `path/to/` bit if we need to specify lots of paths at once - e.g. `tar`

If you need to specify a variable number of paths, see #5791

peterbabic:
@sbrl thank you. Do these convention apply to the actual text? Am I allowed to use backtics in the text to specify the path?

---- example below ----

- Issue a certificate using a custom DNS API script, placed by default at `/root/.acme.sh/dns_custom.sh` (useful when the API is not yet supported upstream):

`acme.sh --issue --dns dns_custom --domain example.com`

pages/common/acme.sh-dns.md
peterbabic: Thinking about it a little bit more, I really like this place to mention multiple domain certificate, because of the introduction of a mentioned good practice. 

However, since multiple domains topic is already open, shouldn't there be a mention in the parent as well?

marchersimon:
@peterbabic we actually use backticks in normal text a lot. Mostly when refering to a command or for URLs.

pages/common/acme.sh-dns.md
```diff
@@ -0,0 +1,25 @@
+# acme.sh --dns
+
+> Use a DNS-01 challenge to issue a TLS certificate.
+> DNS API mode documentation: <https://github.com/acmesh-official/acme.sh/wiki/dnsapi>.
+> More information: <https://github.com/acmesh-official/acme.sh/wiki/DNS-manual-mode>.
```
marchersimon, line 5: I'm not sure if we should use two seperate links for that. What do you think about:
```suggestion
> More information: <https://github.com/acmesh-official/acme.sh/wiki>.
```

pages/common/acme.sh-dns.md
peterbabic: Included via https://github.com/tldr-pages/tldr/pull/5852/commits/5a7e01531e84ae34d62d2bc76271b6fcf3a2807c

pages/common/acme.sh-dns.md
marchersimon: I just saw that `acme.sh.md` uses `https://github.com/acmesh-official/acme.sh`. I'm not sure if we should prefer the wiki or the Readme, but we should at least use the same link for both pages.

peterbabic: But wiki details the DNS mode much better (containing no less than four detailed pages), while the README details the root acme.sh command much better. I am really happy with that as it is.

marchersimon: In that case it's fine of course

sbrl (approved):
Thanks, @peterbabic!

PR #5859, opened 2021-05-01
tar: fix mnemonic for f

tremby:
Before this patch the document suggested that the `f` allows the user to list some files to extract or insert to the archive.

This patch fixes that: the `f` allows an archive file to be specified. Without it the archive would be read from or written to stdout.

### DIFF
--- a/pages/common/tar.md
+++ b/pages/common/tar.md
@@ -4,11 +4,11 @@
 > Often combined with a compression method, such as gzip or bzip2.
 > More information: <https://www.gnu.org/software/tar>.
 
-- [c]reate an archive from [f]iles:
+- [c]reate an archive and write it to a [f]ile:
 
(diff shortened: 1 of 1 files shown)


tremby:
I'm only here making this change, btw, because your demo video shows the tar page, and the misleading mnemonic is the very first thing I noticed. As such it would be nice if when this change (or a similar edit) is accepted that video were regenerated.

bl-ue:
cc @tomadojuice. @tremby that's right, we should regenerate it. FYI it's actually an animated SVG, not a GIF or video.  See https://github.com/tldr-pages/tldr/pull/5507

sbrl (approved):
Thanks, @tremby!

marchersimon (approved):

bl-ue (approved):

PR #5860, opened 2021-05-01
tar: add verbose argument to extract example

tremby:
Before this change, the `v` did not actually do anything since the `t` command already outputs the names of files in the archive.

It is more useful in a `c` or `x` command.

### DIFF
--- a/pages/common/tar.md
+++ b/pages/common/tar.md
@@ -16,9 +16,9 @@
 
 `tar czf {{target.tar.gz}} --directory={{path/to/directory}} .`
 
-- E[x]tract a (compressed) archive [f]ile into the current directory:
+- E[x]tract a (compressed) archive [f]ile into the current directory [v]erbosely:
 
-`tar xf {{source.tar[.gz|.bz2|.xz]}}`
+`tar xvf {{source.tar[.gz|.bz2|.xz]}}`
 
 - E[x]tract a (compressed) archive [f]ile into the target directory:
 


bl-ue (approved):

marchersimon (approved):

sbrl:
Actually, using `-v` with `-t` _does_ actually change the output:

```bash
$ tar -caf archive.tar b/
$ tar -tf archive.tar 
b/
b/a
$ tar -tvf archive.tar 
drwxrwxr-x sbrl/sbrl         0 2021-05-01 19:34 b/
-rw-rw-r-- sbrl/sbrl         0 2021-05-01 19:34 b/a
```

tremby:
I specifically checked this before posting this PR and I could have sworn I didn't see any difference...

I wondered on seeing your comment whether it was the UNIX-style flags you're using rather than the traditional "dashless" ones in the tldr examples that made the difference and so I tried again. But no, output is definitely different with `v` either way.

Questioning my memory and sanity I checked my command history to where I did that check the other day. Turns out I compared `t` with `tz` rather than with `tv` -- oops!

So, as you say, the current example does do something; apologies for my mistake.

I wonder if this change might be useful regardless, or a similar change which keeps `v` in the `t` example but also adds a `v` to an `x` or `c`. My reasoning is that the `v` is useful when creating or extracting. My most common use case is when it'll be a long-running operation, and I want to see some evidence of progress.

I'm going to add a commit which does this, and I'm fine either way if you merge or close.

tremby:
I replaced the commit as mentioned above.

navarroaxel (approved):

bl-ue (approved):

marchersimon (approved):
Thanks @tremby

Waples (approved):

PR #5866, opened 2021-05-01
kdeconnect-cli: add page

marchersimon:
For the link, we have <https://github.com/KDE/kdeconnect-kde/blob/master/cli/kdeconnect-cli.cpp> or <https://manned.org/kdeconnect-cli> and both are not ideal. (`manneg.org`, because it's completely outdated.)

### DIFF
--- a/pages/common/kdeconnect-cli.md
+++ b/pages/common/kdeconnect-cli.md
@@ -31,6 +31,6 @@
 
 `kdeconnect-cli --name {{device_name}} --unlock`
 
-- Send a key to a specific device:
+- Simulate a key press on a specific device:
 
 `kdeconnect-cli --name {{device_name}} --send-keys {{key}}`
--- a/pages/common/kdeconnect-cli.md
+++ b/pages/common/kdeconnect-cli.md
@@ -1,6 +1,7 @@
 # kdeconnect-cli
 
 > KDE Connect CLI.
+> More information: <https://kdeconnect.kde.org>.
 
 - List all devices:
 
--- a/pages/common/kdeconnect-cli.md
+++ b/pages/common/kdeconnect-cli.md
@@ -1,0 +1,35 @@
+# kdeconnect-cli
+
+> KDE Connect CLI.
+
+- List all devices:
+
+`kdeconnect-cli --list-devices`
+
+- List all reachable devices:
+
+`kdeconnect-cli --list-available`
+
+- Request pairing with a specific device, specifying its ID:
+
+`kdeconnect-cli --pair --device {{device_id}}`
+
+- Request pairing with a specific device, specifying its name:
+
+`kdeconnect-cli --pair --name {{device_name}}`
+
+- Ring a specific device:
+
+`kdeconnect-cli --ring --name {{device_name}}`
+
+- Send an SMS with an optional attachment to a specific number:
+
+`kdeconnect-cli --name {{device_name}} --send-sms {{message}} --destination {{phone_number}} --attachment {{path/to/file}}`
+
+- Unlock a specific device:
+
+`kdeconnect-cli --name {{device_name}} --unlock`
+
+- Send a key to a specific device:
+
+`kdeconnect-cli --name {{device_name}} --send-keys {{key}}`


bl-ue:
Would you consider [this](https://manned.org/kdeconnect-cli/904a686c) more update to date?

marchersimon:
They're both the exact same page 🤔

bl-ue:
Yeah, I see that now.  Hmm.

marchersimon:
KDE Connect needs way more attention and development. Most of the commands I listed in the page didn't even work with my phone (Android 11).

pages/common/kdeconnect-cli.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Request pairing with a specific device, specifying its ID:
+
+`kdeconnect-cli --pair --device {{device_id}}`
+
+- Request pairing with a specific device, specifying its name:
+
+`kdeconnect-cli --pair --name {{device_name}}`
+
+- Ring a specific device:
```
sbrl, line 21: Does this cause a device to make a noise, or initiate a call with it?

pages/common/kdeconnect-cli.md
```diff
@@ -0,0 +1,35 @@
+# kdeconnect-cli
+
+> KDE Connect CLI.
```
sbrl, line 3: Let's use that manned.org link without the commit hash. Then it'll stay up to date?

pages/common/kdeconnect-cli.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Send an SMS with an optional attachment to a specific number:
+
+`kdeconnect-cli --name {{device_name}} --send-sms {{message}} --destination {{phone_number}} --attachment {{path/to/file}}`
+
+- Unlock a specific device:
+
+`kdeconnect-cli --name {{device_name}} --unlock`
+
+- Send a key to a specific device:
```
sbrl, line 33: What kind of key is this referring to?

sbrl (changes requested):

pages/common/kdeconnect-cli.md
marchersimon: It makes a noise

pages/common/kdeconnect-cli.md
marchersimon: It's still from 2014 and only has only 6 of 24 flags documented. I wouldn't call that "More" information.

pages/common/kdeconnect-cli.md
marchersimon: A key press

pages/common/kdeconnect-cli.md
sbrl: Ah, ok

pages/common/kdeconnect-cli.md
navarroaxel: So I think the best is the official website:
```suggestion
> KDE Connect CLI.
> More information: <https://kdeconnect.kde.org/>.
```

marchersimon: Probably the best option we have.

pages/common/kdeconnect-cli.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Request pairing with a specific device, specifying its name:
+
+`kdeconnect-cli --pair --name {{device_name}}`
+
+- Ring a specific device:
+
+`kdeconnect-cli --ring --name {{device_name}}`
+
+- Send an SMS with an optional attachment to a specific number:
```
navarroaxel, line 26: If it has an attachment is an MMS, right? Maybe we should let it clear for the cost 🪙

pages/common/kdeconnect-cli.md
marchersimon: Don't they just use man pages from other sources?

pages/common/kdeconnect-cli.md
navarroaxel: I tried it, but I don't understand how this command works.

pages/common/kdeconnect-cli.md
navarroaxel: No, just the official of the command, and the command doesn't have a man page. :man_shrugging: thread closed :joy:

pages/common/kdeconnect-cli.md
marchersimon: It should work that way. However it doesn't work for me either, because a lot of KDE Connect is pretty broken.

pages/common/kdeconnect-cli.md
marchersimon: MMS cost isn't a problem everywhere. Also we already mention that it is optional.

pages/common/kdeconnect-cli.md
bl-ue: Hmm, well maybe we shouldn't even document it then 🤔

marchersimon: I think it's better to have it documented for the cases it does work, than not having it documented because of the cases it doesn't work.

marchersimon: How about
```suggestion
- Simulate a key presses on a specific device:
```

bl-ue: Go for it 👍🏻

bl-ue (approved):

navarroaxel (approved):

sbrl (approved):
Thanks, @marchersimon!

PR #5876, opened 2021-05-03
mmcli: add page

SethFalco:

The 2 example commands are derived from the following documentation:
https://wiki.pine64.org/wiki/PinePhone_FAQ#SMS

Just thought it'd be nice to have quick access to them via tldr.

### DIFF
--- a/pages/linux/mmcli.md
+++ b/pages/linux/mmcli.md
@@ -1,7 +1,7 @@
 # mmcli
 
 > Control and monitor the ModemManager.
-> More information: <https://www.freedesktop.org/software/ModemManager/man/1.0.0/mmcli.8.html>.
+> More information: <https://www.freedesktop.org/software/ModemManager/man/latest/mmcli.1.html>.
 
 - List SMS messages available on the modem:
 
--- a/pages/linux/mmcli.md
+++ b/pages/linux/mmcli.md
@@ -9,4 +9,4 @@
 
 - Delete a message from the modem by path:
 
-`sudo mmcli --modem={{modem}} --messaging-delete-sms={{path}}`
+`sudo mmcli --modem={{modem}} --messaging-delete-sms={{path/to/message_file}}`
--- a/pages/linux/mmcli.md
+++ b/pages/linux/mmcli.md
@@ -7,6 +7,6 @@
 
 `sudo mmcli --modem={{modem}} --messaging-list-sms`
 
-- Delete a message from the modem by path:
+- Delete a message from the modem, specifying its path:
 
 `sudo mmcli --modem={{modem}} --messaging-delete-sms={{path/to/message_file}}`
--- a/pages/linux/mmcli.md
+++ b/pages/linux/mmcli.md
@@ -5,7 +5,7 @@
 
 - List SMS messages available on the modem:
 
-`sudo mmcli -m {{modem}} --messaging-list-sms`
+`sudo mmcli --modem={{modem}} --messaging-list-sms`
 
 - Delete a message from the modem by path:
 
--- a/pages/linux/mmcli.md
+++ b/pages/linux/mmcli.md
@@ -9,4 +9,4 @@
 
 - Delete a message from the modem by path:
 
-`sudo mmcli -m {{modem}} --messaging-delete-sms={{path}}`
+`sudo mmcli --modem={{modem}} --messaging-delete-sms={{path}}`
--- a/pages/linux/mmcli.md
+++ b/pages/linux/mmcli.md
@@ -1,0 +1,12 @@
+# mmcli
+
+> Control and monitor the ModemManager.
+> More information: <https://www.freedesktop.org/software/ModemManager/man/1.0.0/mmcli.8.html>.
+
+- List SMS messages available on the modem:
+
+`sudo mmcli -m {{modem}} --messaging-list-sms`
+
+- Delete a message from the modem by path:
+
+`sudo mmcli -m {{modem}} --messaging-delete-sms={{path}}`


bl-ue (approved):
>Just thought it'd be nice to have quick access to them via tldr.

Definitely!

pages/linux/mmcli.md
```diff
@@ -0,0 +1,12 @@
+# mmcli
+
+> Control and monitor the ModemManager.
+> More information: <https://www.freedesktop.org/software/ModemManager/man/1.0.0/mmcli.8.html>.
+
+- List SMS messages available on the modem:
+
+`sudo mmcli -m {{modem}} --messaging-list-sms`
```
navarroaxel, line 8: We prefer long arguments when there are available, checking the man page we can use:
```suggestion
`sudo mmcli --modem={{modem}} --messaging-list-sms`
```

pages/linux/mmcli.md
```diff
@@ -0,0 +1,12 @@
...
+
+- Delete a message from the modem by path:
+
+`sudo mmcli -m {{modem}} --messaging-delete-sms={{path}}`
```
navarroaxel, line 12: Based on our style guide we prefer `path/to/file`. Could this `path` be `path/to/message_file`? or is it a directory?
```suggestion
`sudo mmcli --modem={{modem}} --messaging-delete-sms={{path}}`
```

pages/linux/mmcli.md
```diff
@@ -0,0 +1,12 @@
+# mmcli
+
+> Control and monitor the ModemManager.
+> More information: <https://www.freedesktop.org/software/ModemManager/man/1.0.0/mmcli.8.html>.
```
nicokosi, line 4: Refer to `latest` instead of `1.0.0` ?
```suggestion
> More information: <https://www.freedesktop.org/software/ModemManager/man/latest/mmcli.1.html>.
```

pages/linux/mmcli.md
SethFalco: I believe it can be, I think its files only so it's a fair suggestion.
Looking at the docs, I can see it accepts an index as well which I didn't realize, not 100% sure if I should document both or just leave it at one for simplicity.

For now, I'll leave it at path only since it's more explicit though, feel free to give your opinion!

pages/linux/mmcli.md
bl-ue: I tried using "latest", but I didn't think to chanege the section number like that—smart!

SethFalco:
Sorry for the commit spam, I wanted to commit your suggestions using the UI since it'd save me from having to co-author manually, but I wasn't sure how to accept multiple suggestions at once. ^-^'

bl-ue:
@SethFalco the CLA isn't working because the email address that you used for the last commit (`sethi@tuta.io`) isn't added to your GitHub account.  To fix this, you can either:
1. Change the email you used for the commit† to [one of the emails attached to your GitHub account](https://github.com/settings/emails),

2. Or, preferably (and much easier), [add the email you used for this commit (`sethi@tuta.io`) to your GitHub account](https://github.com/settings/emails).

†To change the email address of the last commit, run this command:
```
git commit --amend --author="Seth Falco <seth@falco.fun>"
```

SethFalco:
Sorry about that, was getting rid of old email addresses. ^-^'
I'll just add it back to my account and remove it again once this has been merged since multiple of the commits here used sethi@tuta.io; should've updated my local git config sooner.

bl-ue:
You know, I don't think you should do that.  It might cause weird CLA issues in the future with other PRs or something like that.  Instead, run the command I mentioned, in option number 2.

SethFalco:
> You know, I don't think you should do that. It might cause weird CLA issues in the future with other PRs or something like that. Instead, run the command I mentioned, in option number 2.

I can do that if you'd like. If that may indeed cause problems though then I think it'd be a flaw with the bot, since deleting email addresses common enough that people may not even have access to addresses they may have used to commit in the past.

pages/linux/mmcli.md
marchersimon: Even if it is a directory, we could still use `path/to/directory`. Or `path/to/file_or_directory` if both are possible.

pages/linux/mmcli.md
```diff
@@ -0,0 +1,12 @@
+# mmcli
+
+> Control and monitor the ModemManager.
+> More information: <https://www.freedesktop.org/software/ModemManager/man/latest/mmcli.1.html>.
+
+- List SMS messages available on the modem:
+
+`sudo mmcli --modem={{modem}} --messaging-list-sms`
+
+- Delete a message from the modem by path:
```
marchersimon, line 10:
```suggestion
- Delete a message from the modem specifying its path:
```
This wording is more common among the other pages.

SethFalco: I didn't think about that. I searched a bit through the commands to confirm, but I think most tend _not_ to acknowledge their parameters at all. Could it be worth dropping that entirely?

```suggestion
- Delete a message from the modem:
```

image
> It seems most of the description that explicitly say "specify the path" do so because the command is setting an output path or similar.
 
image
> While some mention "specifying its path" or something similar, it's not very common.

image
> Looking at commands like 7z, adb, etc most don't even acknowledge the paramaters, or that a path is required, it's only inferred by the command itself.

marchersimon: I wouldn't remove it entirely since we should try to keep the descriptions as simple as possible. I just think that `by path` sounds a little odd. However, I'm not an native speaker, so let's see what @bl-ue thinks.

bl-ue: It's actually perfectly normal to say something like that.  I'd prefer a slight more verbose version, however:
```suggestion
- Delete a message from the modem by its path:
```
wdyt @SethFalco?

SethFalco: Objectively speaking, I think @marchersimon has a point with "sounds a little odd". Even if mine's grammatically correct, it could be worth being more verbose if it aids in accessibility. (Baring in mind a lot of non-natives will be using the English version anyway.)

As a native speaker, I prefer my version, but ofc I'd say that, I wrote it. 🥇
I'm pretty comfortable with the terse versions, but think it could be worth using the initial suggestion if the explicitness makes things clearer.

navarroaxel (approved):

marchersimon (approved):
Thanks @SethFalco

PR #5881, opened 2021-05-03
more: add examples and link

navarroaxel:
- more: add examples and link
- more: sync Spanish translation

### DIFF
--- a/pages/common/more.md
+++ b/pages/common/more.md
@@ -7,7 +7,7 @@
 
 `more {{path/to/file}}`
 
-- Open a file displaying from a specific line:
+- Open a file at a specific line:
 
 `more {{path/to/file}} +{{line_number}}`
 
--- a/pages/common/more.md
+++ b/pages/common/more.md
@@ -7,9 +7,9 @@
 
 `more {{path/to/file}}`
 
-- Open a file at a specific line:
+- Open a file displaying from a specific line:
 
-`more {{path/to/file}} +{{line_number}}`
+`more +{{line_number}} {{path/to/file}}`
 
 - Display help:
 
--- a/pages/common/more.md
+++ b/pages/common/more.md
@@ -1,12 +1,21 @@
 # more
 
-> Open a file for interactive reading, allowing scrolling and search (in forward direction only).
+> Open a file for interactive reading, allowing scrolling and search.
+> More information: <https://manned.org/more>.
 
 - Open a file:
 
-`more {{source_file}}`
+`more {{path/to/file}}`
 
-- Page down:
+- Open a file displaying from a specific line:
+
+`more {{path/to/file}} +{{line_number}}`
+
+- Display help:
+
+`more --help`
+
+- Go to the next page:
 
 `<Space>`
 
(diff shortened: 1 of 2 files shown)


bl-ue (approved):

pages/common/more.md
```diff
@@ -1,12 +1,21 @@
...
+> Open a file for interactive reading, allowing scrolling and search.
+> More information: <https://manned.org/more>.
 
 - Open a file:
 
-`more {{source_file}}`
+`more {{path/to/file}}`
 
-- Page down:
+- Open a file displaying from a specific line:
```
marchersimon, line 10:
```suggestion
- Open a file at a specific line:
```
Maybe that will read better?

navarroaxel: Because, as a user, you see the full page not just a specific line.
The `--help` says
```
display file beginning from line number
```
The `man more` says
```
+number
    Start displaying each file at line number.
```

bl-ue: This, perhaps?
```suggestion
- Open a file, starting at a specific line:
```

sbrl: `beginning from` for me suggests that it does not show anything before that line - only that line and after, as does `starting at`.

`Open a file at a specific line` suggests to me it opens the entire file and jumps to a preset line to start with, but you can browse the entire file as you like.
`

marchersimon (approved):

pages/common/more.md
```diff
@@ -1,12 +1,21 @@
...
 
 - Open a file:
 
-`more {{source_file}}`
+`more {{path/to/file}}`
 
-- Page down:
+- Open a file at a specific line:
+
+`more {{path/to/file}} +{{line_number}}`
```
sbrl, line 12: Oh, hold on here - I've just looked into it more closely, and the following would be more correct:

```suggestion
- Open a file displaying from a specific line:

`more +{{line_number}} {{path/to/file}}`
```

2 things here:

1. the `+{{line_number}}` *must* be before `{{path/to/file}}`
2. It starts displaying the file at the specified line, cutting everything else before it off, so its more correct to say "displaying from a specific line", rather than "open a file at a specific line", which would imply that you could scroll up to previous lines, which isn't the case with `more` as far as I know (but is with `less`).

marchersimon: I can confirm that it cuts off everything before that line, however `more {{path/to/file}} +{{line_number}}` works for me too (linux, more 2.36.2)

navarroaxel: I checked this in Ubuntu focal (Docker) and `more v2.34` needs the `+10` before the file.

sbrl (approved):
Thanks, @navarroaxel!

bl-ue:
When's this going to be available?  Or rather, what is your release schedule (if any)?

bl-ue:
Oops, sorry!  Wrong issue :)

bl-ue:
In case you're interested, [this](https://github.com/sindresorhus/refined-github/pull/4317#event-4694315814) is the thread I was intended to write to :)

PR #5883, opened 2021-05-03
README: tweak contributors badge

waldyrious:
The shields.io badge for displaying a count of contributors has two variants: `contributors` and `contributors-anon`, as shown in https://shields.io/category/activity (click the demo "contributors" badge there to see the configuration modal).

Apparently the former only counts contributions (i.e. commits) whose author emails can be mapped to GitHub users. Other contributions are considered "anonymous" by GitHub. They are usually the result of either a misconfigured git installation, emails that haven't been added to the user's GitHub profile, or accounts that have been deleted or that changed their email address.

Since the project has always existed on GitHub from its inception, the anonymous count is closer to the real number, since all contributions would have come from GitHub pull requests, which require a GitHub account to create.

There may be a slight overestimation in the contributors-anon count, e.g. from people who authored commits with different author emails; but that is nevertheless a closer estimate than the non-anon count, and besides, it matches what GitHub shows on their own interface.

See more details at <https://github.com/badges/shields/issues/5969>.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 [prs-merged-url]: https://github.com/tldr-pages/tldr/pulls?q=is:pr+is:merged
 [prs-merged-image]: https://img.shields.io/github/issues-pr-closed-raw/tldr-pages/tldr.svg?label=merged+PRs&color=green
 [contributors-url]: https://github.com/tldr-pages/tldr/graphs/contributors
-[contributors-image]: https://img.shields.io/github/contributors/tldr-pages/tldr.svg
+[contributors-image]: https://img.shields.io/github/contributors-anon/tldr-pages/tldr.svg
 [license-url]: https://github.com/tldr-pages/tldr/blob/main/LICENSE.md
 [license-image]: https://img.shields.io/badge/license-CC_BY_4.0-blue.svg
 </div>


waldyrious:
Here's a before/after comparison of the badges:

- `contributors` variant: 
- `contributors-anon` variant:

marchersimon:
Woah, that's a huge difference.

marchersimon (approved):

waldyrious:
Yeah, it's been bugging me for the longest time but I kept postponing looking into it! 😅

Screenshot 2021-05-03 at 19 34 33

marchersimon:
The number of contributions per month would also be interesting.
GitHub commit activity

marchersimon:
What do you think of this badge style? It looks more polished and modern. <https://img.shields.io/github/contributors-anon/tldr-pages/tldr.svg?style=for-the-badge>
<table><tr><td>

</td><td>

</td></table>

bl-ue:
I like the flat badges too!  They're getting increasingly more common, I've observed.

bl-ue (approved):

Waples:
I like the flat version better aswel

Waples (approved):

waldyrious:
Hey @Waples, there was some discussion ongoing in the PR thread. We typically prefer for such discussions to be resolved (or copied to a dedicated issue, for tracking) before merging the PR, otherwise they can get lost/abandoned. In this case, we could have incorporated some of the suggestions into the PR. Next time, please ask if it's OK to merge in case there are open questions.

waldyrious:
I don't feel like changing the badge style is an obvious improvement — that's just a matter of style, and to some degree, design trends. But if most of you would prefer it, I won't object to the change.

I noticed from the screenshot, though, that the logo does not look so good when using the dark theme. We might want to consider a different text color and making the background transparent... any thoughts? 🤔

Also, good idea about the commit activity badge. We currently have a different set of badges in the README and in [CONTRIBTING.md](https://github.com/tldr-pages/tldr/blob/main/CONTRIBUTING.md) (which reminds me, we should update the contributors badge there too). I wonder if we should centralize the badges in the README, since the only difference is swapping the build badge for the CLA badge; all the others are present in both places.

marchersimon:
I found <https://github.community/t/support-theme-context-for-images-in-light-vs-dark-mode/147981>, where they were discussing about a feature to change images depending on the theme. I didn't read through it all, but there seemed to be some workarounds for now. If that doesn't work, removing that background and changing the colors a bit should do fine.

Regarding `Contributing.md`: I don't really see a reason to show the badges there.

waldyrious:
Nice find, but I'd rather make our logo work well in both dark and light backgrounds by default (which I think should be doable), than have two versions of the logo and a GitHub-specific mechanism to swap them.

sbrl:
I'm not convinced that the flat ones are nicer - I'd prefer to stick with the existing style for now.

It would be nice if our logo worked for both the light and dark theme, but the issue there is that if you have enough contrast with 1, you'll have issues with the other.

PR #5887, opened 2021-05-04
stale GitHub action: re-wrap long message

waldyrious:
GitHub's markdown renderer does not wrap single line breaks, so the message the bot leaves when closing a thread ([example](https://github.com/tldr-pages/tldr-lint/pull/64#issuecomment-831610730)) would have an awkward line break. This commit re-wraps the message to avoid that.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -36,5 +36,5 @@ jobs:
         close-pr-message: |
           Hi everyone.
           This thread is being closed as there was no response to the previous prompt.
-          However, please leave a comment whenever you're ready to resume,
-          so the thread can be reopened. Thanks again!
+          However, please leave a comment whenever you're ready to resume, so the thread can be reopened.
+          Thanks again!


waldyrious:
ps — we probably will want to do the same [here](https://github.com/tldr-pages/tldr-lint/blob/126840d26011b99c9a91678c66187a332fa8158c/.github/workflows/stale.yml) and [here](https://github.com/tldr-pages/tldr-c-client/blob/ffcd18b6a7998fc79777189f58711107ea3d8e16/.github/workflows/stale.yml). I wonder if there's a way to set up actions at the organization level... 🤔 perhaps placing them in a `.github` repository would work for that?

navarroaxel (approved):

nicokosi (approved):
Seems OK for me. :innocent:

marchersimon (approved):

bl-ue (approved):

waldyrious:
> @waldyrious do you want to do that in this PR

It can't be done in this PR since those are separate repos. I was kinda hoping someone would offer to do it, but I guess I'll roll up my sleeves and do it myself 😛

bl-ue:
ahem, what about deleting a newline requires rolling up your sleeves??

PR #5888, opened 2021-05-04
venv: add page

Timmmm:
I keep having to look this up! Note that there are other similar commands for the `source bin/activate` on other platforms, but I don't have time to figure out how to do platform-specific tl;drs, sorry.

- [mostly] The page is in the correct platform directory (`common/`, `linux/`, etc.)

### DIFF
--- a/pages/common/venv.md
+++ b/pages/common/venv.md
@@ -1,0 +1,19 @@
+# venv
+
+> Create isolated "virtual" Python environements.
+
+- Create a new venv:
+
+`python3 -m venv <venv_path>`
+
+- Create a new venv in current directory
+
+`python3 -m venv .`
+
+- Activate the venv (Bash/zsh):
+
+`source <venv_path>/bin/activate`
+
+- Install dependencies:
+
+`pip3 install -r requirements.txt`


marchersimon:
Thank you @Timmmm, however every page should be about a command. If there isn't a seperate command called `venv`, it doesn't really fit into tldr's scope.

Timmmm:
> If there isn't a seperate command called venv, it doesn't really fit into tldr's scope.

Seems like a slightly arbitrary restriction. `venv` is a command you just run it via `python3 -m venv`. `pip` is the same, it just happens to have a shortcut installed.

I can understand not wanting to end up with `tldr principal component analysis` or whatever but I think `venv` is pretty clearly "a command".

Your loss though.

navarroaxel:
I wonder if we can document this as `pytthon-venv`. @MasterOdin is it a kind of built-in modules that extends the features of python?

sbrl:
Hey there, @Timmmm! We already have a page about virtualenv here: https://github.com/tldr-pages/tldr/blob/main/pages/common/virtualenv.md

Virtualenv actually does have a dedicated command called `virtualenv`, under which we document it. Are there any improvements you'd like to make to the existing page instead?

Timmmm:
`virtualenv` is not actually the same command:

> virtualenv is a tool to create isolated Python environments. Since Python 3.3, a subset of it has been integrated into the standard library under the venv module. The venv module does not offer all features of this library

> `python-venv`

I mean that kind of makes sense but it's not really discoverable.

sbrl:
Oh, I see, I wasn't aware of that.

Is there a specific instance in which one would prefer `python -m venv` over the main `virtualenv` command?

bl-ue:
I've only ever used the former, and it's always served all my needs.

Timmmm:
> Is there a specific instance in which one would prefer python -m venv over the main virtualenv command?

Yeah if you can't be bothered to install `virtualenv`! (Or didn't know it existed.)

PR #5889, opened 2021-05-04
gradle: add tasks examples

nicokosi:

### DIFF
--- a/pages/common/gradle.md
+++ b/pages/common/gradle.md
@@ -22,3 +22,7 @@
 - Compile and Release package:
 
 `gradle assembleRelease`
+
+- List tasks:
+
+`gradle tasks`
--- a/pages/common/gradle.md
+++ b/pages/common/gradle.md
@@ -23,6 +23,10 @@
 
 `gradle assembleRelease`
 
-- List tasks:
+- List the main tasks:
 
 `gradle tasks`
+
+- List all the tasks:
+
+`gradle tasks --all`


pages/common/gradle.md
```diff
@@ -22,3 +22,7 @@
 - Compile and Release package:
 
 `gradle assembleRelease`
+
+- List tasks:
+
+`gradle tasks`
```
nicokosi, line 28:
```suggestion
`gradle tasks -all`
```
?
See https://docs.gradle.org/current/userguide/command_line_interface.html#sec:listing_tasks

marchersimon: I'd say we should use, what is more helpful / common. If both are commonly used, why not include both?

nicokosi: To be honest, I am not sure to understand well the difference. 😇 As far as I understand, `--all` also lists "tasks called by tasks" so maybe its more simple and useful to document `--all` (instead of giving "wrong"/unclear information)?

nicokosi: As a Gradle user, I don't care, I use `gradle tasks --list` or `gradle tasks --list | grep foo`. 😇

marchersimon: I think it's better to give too much info, than too little.

nicokosi: OK, I followed your advice (I'm not totally convinced :sweat_smile:) in aace0154.

marchersimon: You don't have to make any changes if you're not convinced. At the end, you're the one why knows how to use the tool.

nicokosi: In fact, I am just an average Gradle user and I don't know what's preferable. :innocent:
In doubt, it's OK to document more. :+1: (I am convinced now :blush:).

Waples (approved):

navarroaxel (approved):

marchersimon (approved):
Thanks @nicokosi

PR #5892, opened 2021-05-04
uuid: add page, uuidgen: update description

peterbabic:


pages/linux/uuid.md
```diff
@@ -0,0 +1,32 @@
...
+
+`uuid -v {{3}} ns:URL {{https://example.com}}`
+
+- Generate multiple v1 UUIDs at once by specifying a count:
+
+`uuid -n {{12}}`
+
+- Generate a v1 UUID in a single integer value (SIV) representation format:
+
+`uuid -F {{SIV}}`
```
marchersimon, line 24: What do you think of:
```suggestion
- Generate a v1 UUID and specify the output format:

`uuid -F {{BIN|STR|SIV}}`
```
?

pages/linux/uuid.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Generate a v1 UUID and write the output to a file:
+
+`uuid -o {{path/to/file}}`
+
+- Decode a given UUID:
+
+`uuid -d {{68794df6-5e20-385f-ab08-bb73f8a433cb}}`
```
marchersimon, line 32:
```suggestion
`uuid -d {{uuid}}`
```

pages/linux/uuid.md
peterbabic: have not encountered a pipe | here yet, but definitely resembles well known syntax

pages/linux/uuid.md
```diff
@@ -0,0 +1,32 @@
+# uuid
+
+> Generate and decode an Universally Unique Identifier (UUID).
+> More information: <https://manned.org/uuid.1>.
+
+- Generate a v1 UUID (time and node based):
```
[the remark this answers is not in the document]
peterbabic: v1 is a default type having a *node* linking to a MAC address of the machine that generates it. I also need to learn more about it 😄

pages/linux/uuid.md
bl-ue: We don't use it for actual syntax, we use it to indicate that there are multiple options.
https://sourcegraph.com/search?q=+repo:%5Egithub%5C.com/tldr-pages/tldr%24+%5C%7B%5C%7B%28%5B%5E%7B%7D%5D%2B%3F%5C%7C%29%2B%5B%5E%7B%7D%5D%2B%5C%7D%5C%7D&patternType=regexp

peterbabic: Nice service that sourcegraph, thanks.

pages/linux/uuid.md
```diff
@@ -0,0 +1,32 @@
...
+
+`uuid`
+
+- Generate a v4 UUID (random data based):
+
+`uuid -v {{4}}`
+
+- Generate a v3 UUID (name based) with an URL namespace:
+
+`uuid -v {{3}} ns:URL {{https://example.com}}`
```
peterbabic, line 16: - Generate a v5 UUID (name based) with an URL namespace:

`uuid -v {{5}} ns:URL {{https://example.com}}`

peterbabic: Not sure what I did here yet. I had problems synchronizing because of the switch from `master` to `main` branch that was finalized recently (good job btw).

I wantent to use GitHub UI to add changes. The reason:

UUID v3 uses MD5 and UUID v5 uses SHA1, which is preferred. Our examples here should only include UUID v1, v4 and v5 in the final form (v2 not available and v3 is less secure).

peterbabic: It doesn't look like the suggestions you all leave here, that are ready to be committed. Can you guys give me a hand (replacing number 3 with the number 5) till I figure it out?

marchersimon: @peterbabic you only have to put a

<p>```suggestion</p>

<p>```
around it

peterbabic: @marchersimon that would have never occurred to me by chance

pages/linux/uuid.md
```diff
@@ -0,0 +1,32 @@
+# uuid
+
+> Generate and decode Universally Unique Identifiers (UUID).
+> More information: <https://manned.org/uuid>.
```
peterbabic, line 4: Just a note, the official man page would be at http://www.ossp.org/man/man.cgi/pkg/lib/uuid/uuid_cli.pod

But it is at 500 error for a few days already, I have just written an email there.

The base page is at http://www.ossp.org/pkg/lib/uuid/

Manned.org displays the same man page as installed locally with the command, so it can probably stay as it is.

pages/linux/uuid.md
```diff
@@ -0,0 +1,33 @@
...
+
+`uuid`
+
+- Generate a v4 UUID (random data based):
+
+`uuid -v {{4}}`
+
+- Generate a v5 UUID (name based) with an URL namespace:
+
+`uuid -v {{5}} ns:URL {{https://example.com}}`
```
peterbabic, line 17: I have generalized the example with all my newly acquired knowledge:

```suggestion
- Generate a v5 UUID (name based) with a specified namespace:

`uuid -v {{5}} ns:{{DNS|URL|OID|X500}} {{https://example.com}}`
```

pages/linux/uuidgen.md
```diff
@@ -1,6 +1,8 @@
 # uuidgen
 
 > Generate unique identifiers (UUIDs).
+> See also `uuid`.
+> More information: <https://manned.org/uuidgen>.
 
 - Create a random UUID:
```
peterbabic, line 7: Using `--random` we get an v4 UUID. Similarly below, using `--time` we get a v1 UUID and `--sha1 --namespace` is the v5 UUID (and `--md5 --namespace` would be v3 UUID).

Should there be version mentions here? Referencing to UUID by its version number seems to be very common around the internet.

peterbabic: I could not add suggestions to the whole file as there in the middle are lines which were not modified with this commit (these lines need to be unfolded manually first and then they sadly break the blue vertical line).

peterbabic:
Probably also worth adding [Page edit](https://github.com/tldr-pages/tldr/labels/page%20edit) label.

pages/linux/uuid.md
marchersimon: We can change the link as soon as the website works again, but for now `manned.org` is perfectly fine.

pages/linux/uuid.md
marchersimon: Why does it has to be a URL? You can put any string you want in there. And why is there a URL namespace if you can use a URL with all other namespaces too?

If this is obvious or hard to explain with a few words, then the example is fine as is.

peterbabic: @marchersimon yes, you can put any string there, there is no example of this here. But there are exactly for namepace *prefixes*, mentioned in the example. Does this read better?


```suggestion
- Generate a v5 UUID (name based) with a specified namespace prefix:

`uuid -v {{5}} ns:{{DNS|URL|OID|X500}} {{https://example.com}}`
```

pages/linux/uuidgen.md
marchersimon: > Should there be version mentions here?

Sure, go ahead.

> I could not add suggestions to the whole file as there in the middle are lines which were not modified with this commit (these lines need to be unfolded manually first and then they sadly break the blue vertical line).

Why don't you use the GitHub editor or a local git repo to edit the files?

pages/linux/uuid.md
marchersimon: I think the wording is fine, I'm just confused about the command itself. If any string is possible, why don't we use `uuid -v {{5}} ns:{{DNS|URL|OID|X500}} {{string}}` or something like that?

pages/linux/uuidgen.md
peterbabic: Just wanted to learn the suggestion feature you just taught me, GitHub editor is still something I have not touched yet, one step at the time 👼🏻

marchersimon: Yeah, sadly suggestions can only be made near other suggestions. (Kind of understanable. When working with code it doesn't make much sense suggesting differences to code which wasn't touched).

pages/linux/uuid.md
peterbabic: @marchersimon aah yes, you are right, nice catch! I have edited the suggestion here to match the one in the [uuidgen](https://github.com/tldr-pages/tldr/pull/5892/files#diff-e6eaf9f527253cf5204d3886c8e32d44bb74188debe51af664430522ba68cbfb) file

pages/linux/uuidgen.md
peterbabic: I fully agree, I just did not know that. But then the only way to express my intentions before pushing a change is to do a comment, but here at this specific place a suggestion would work really well, even though outside of the active changes. Nevermind.

marchersimon: While not as aesthetically pleasing, you could use

````
```diff
- before1
- before2
+ after1
+ after2
```
````
which will turn into 
```diff
- before1
- before2
+ after1
+ after2
```
along with the line number of the file you are referencing to.

marchersimon (approved):
Thanks for the great work @peterbabic

pages/linux/uuid.md
```diff
@@ -0,0 +1,33 @@
...
+
+> Generate and decode Universally Unique Identifiers (UUID).
+> See also `uuidgen`.
+> More information: <https://manned.org/uuid>.
+
+- Generate a v1 UUID (based on time and system's hardware address, if present):
+
+`uuid`
+
+- Generate a v4 UUID (random data based):
```
bl-ue, line 11:
```suggestion
- Generate a v4 UUID (based on random data):
```

pages/linux/uuid.md
```diff
@@ -0,0 +1,33 @@
...
+
+- Generate a v1 UUID (based on time and system's hardware address, if present):
+
+`uuid`
+
+- Generate a v4 UUID (based on random data):
+
+`uuid -v {{4}}`
+
+- Generate a v5 UUID (name based) with a specified namespace prefix:
```
peterbabic, line 15: @bl-ue regarding your previous suggestion:

```suggestion
- Generate a v5 UUID (based on the supplied object name) with a specified namespace prefix:
```

marchersimon (approved):

navarroaxel (approved):

marchersimon (approved):

PR #5912, opened 2021-05-07
distccd: add page

tomek-szczesny:
Not a native speaker, corrections are welcome.

### DIFF
--- a/pages/common/distccd.md
+++ b/pages/common/distccd.md
@@ -1,7 +1,7 @@
 # distccd
 
 > Server daemon for distcc distributed compiler.
-> More information: <https://distcc.github.io/>.
+> More information: <https://distcc.github.io>.
 
 - Start a daemon with default settings:
 
--- a/pages/common/distccd.md
+++ b/pages/common/distccd.md
@@ -15,7 +15,7 @@
 
 `distccd --daemon --allow {{ip_address|network_prefix}}`
 
-- Start a daemon with a maximum of 4 tasks at any time and lowered priority (niceness = 5):
+- Start a daemon with lowered priority that can run a maximum of 4 tasks at a time:
 
 `distccd --daemon --jobs {{4}} --nice {{5}}`
 
--- a/pages/common/distccd.md
+++ b/pages/common/distccd.md
@@ -7,7 +7,7 @@
 
 `distccd --daemon`
 
-- Start a daemon, accepting connections from private network IPv4 ranges:
+- Start a daemon, accepting connections from IPv4 private network ranges:
 
 `distccd --daemon --allow-private`
 
--- a/pages/common/distccd.md
+++ b/pages/common/distccd.md
@@ -15,7 +15,7 @@
 
 `distccd --daemon --allow {{IP[/BITS]}}`
 
-- Start a daemon with maximum of 4 tasks at any time and lowered priority (niceness = 5):
+- Start a daemon with a maximum of 4 tasks at any time and lowered priority (niceness = 5):
 
 `distccd --daemon --jobs {{4}} --nice {{5}}`
 
--- a/pages/common/distccd.md
+++ b/pages/common/distccd.md
@@ -3,7 +3,7 @@
 > Server daemon for the distcc distributed compiler.
 > More information: <https://distcc.github.io>.
 
-- Start a daemon with default settings:
+- Start a daemon with the default settings:
 
 `distccd --daemon`
 
--- a/pages/common/distccd.md
+++ b/pages/common/distccd.md
@@ -7,7 +7,7 @@
 
 `distccd --daemon`
 
-- Start a daemon accepting connections from private network ipv4 ranges:
+- Start a daemon, accepting connections from private network IPv4 ranges:
 
 `distccd --daemon --allow-private`
 
--- a/pages/common/distccd.md
+++ b/pages/common/distccd.md
@@ -1,6 +1,6 @@
 # distccd
 
-> Server daemon for distcc distributed compiler.
+> Server daemon for the distcc distributed compiler.
 > More information: <https://distcc.github.io>.
 
 - Start a daemon with default settings:
--- a/pages/common/distccd.md
+++ b/pages/common/distccd.md
@@ -13,7 +13,7 @@
 
 - Start a daemon, accepting connections from a specific network address or address range:
 
-`distccd --daemon --allow {{IP[/BITS]}}`
+`distccd --daemon --allow {{ip_address|network_prefix}}`
 
 - Start a daemon with a maximum of 4 tasks at any time and lowered priority (niceness = 5):
 
--- a/pages/common/distccd.md
+++ b/pages/common/distccd.md
@@ -19,6 +19,6 @@
 
 `distccd --daemon --jobs {{4}} --nice {{5}}`
 
-- Start a daemon and register via mDNS/DNS-SD (Zeroconf):
+- Start a daemon and register it via mDNS/DNS-SD (Zeroconf):
 
 `distccd --daemon --zeroconf`
--- a/pages/common/distccd.md
+++ b/pages/common/distccd.md
@@ -15,7 +15,7 @@
 
 `distccd --daemon --allow {{ip_address|network_prefix}}`
 
-- Start a daemon with lowered priority that can run a maximum of 4 tasks at a time:
+- Start a daemon with a lowered priority that can run a maximum of 4 tasks at a time:
 
 `distccd --daemon --jobs {{4}} --nice {{5}}`
 
--- a/pages/common/distccd.md
+++ b/pages/common/distccd.md
@@ -11,7 +11,7 @@
 
 `distccd --daemon --allow-private`
 
-- Start a daemon accepting connections from specific network address or address range:
+- Start a daemon, accepting connections from a specific network address or address range:
 
 `distccd --daemon --allow {{IP[/BITS]}}`
 
--- a/pages/common/distccd.md
+++ b/pages/common/distccd.md
@@ -1,0 +1,24 @@
+# distccd
+
+> Server daemon for distcc distributed compiler.
+> More information: <https://distcc.github.io/>.
+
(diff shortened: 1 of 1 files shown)


pages/common/distccd.md
```diff
@@ -0,0 +1,24 @@
+# distccd
+
+> Server daemon for distcc distributed compiler.
+> More information: <https://distcc.github.io/>.
```
marchersimon, line 4:
```suggestion
> More information: <https://distcc.github.io>.
```

pages/common/distccd.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Start a daemon with default settings:
```
marchersimon, line 6: Can you run multiple daemons? If not, I think `Start the daemon ...` would be more fitting.

pages/common/distccd.md
```diff
@@ -0,0 +1,24 @@
...
+
+`distccd --daemon`
+
+- Start a daemon accepting connections from private network ipv4 ranges:
```
marchersimon, line 10:
```suggestion
- Start a daemon accepting connections from private network IPv4 ranges:
```

pages/common/distccd.md
```diff
@@ -0,0 +1,24 @@
...
+
+`distccd --daemon --allow-private`
+
+- Start a daemon accepting connections from specific network address or address range:
+
+`distccd --daemon --allow {{IP[/BITS]}}`
+
+- Start a daemon with maximum of 4 tasks at any time and lowered priority (niceness = 5):
```
marchersimon, line 18:
```suggestion
- Start a daemon with a maximum of 4 tasks at any time and lowered priority (niceness = 5):
```

marchersimon (commented):
Looks great. Just a few details below.

marchersimon:
As far as I can tell, `distccd` only works on Linux. If that's the case, we should move it to `linux/`.

pages/common/distccd.md
tomek-szczesny: Technically it is possible to start a few daemons, each listening to a different port. No idea why would anyone want that, but I wouldn't imply it is not possible.

tomek-szczesny:
> As far as I can tell, `distccd` only works on Linux. If that's the case, we should move it to `linux/`.

"distcc is developed on GNU/Linux, but has been reported to work on other systems including FreeBSD, NetBSD, Darwin, Solaris, HP-UX, IRIX, Cygwin and BSD/OS."

bl-ue:
Then I'd say it should definitely stay in `common/` 👍🏻

pages/common/distccd.md
```diff
@@ -0,0 +1,24 @@
...
+
+`distccd --daemon`
+
+- Start a daemon, accepting connections from private network IPv4 ranges:
+
+`distccd --daemon --allow-private`
+
+- Start a daemon, accepting connections from a specific network address or address range:
+
+`distccd --daemon --allow {{IP[/BITS]}}`
```
navarroaxel, line 16: We don´t use this notation `IP[/BITS]`, we use `lower_snake_case` and ORs `|` for argument options:
```suggestion
`distccd --daemon --allow {{ip_address|netmask_address}}`
```

tomek-szczesny: This bit is taken straight from `distccd --help` output.
`    -a, --allow IP[/BITS]      client address access control`
I thought the original output would be clearer, but if it violates tldr guidelines, I am in no opposition.

navarroaxel: :thinking: With your example I think I can enter a specific IP like `192.168.0.4` or a mask like `192.168.0.0/24`. Is this correct?

tomek-szczesny: Yes, that's my understanding as well. Personally my knowledge about networking is quite limited, and perhaps that's the reason why `IP[/BITS]` is easier for me to understand.
We may settle this by splitting this example into two. One example for a specific IP, and another for IP with BITS. ;)

sbrl: Sounds good, @tomek-szczesny!

Indeed, a netmask is not what we're after here - an example of a netmask is `255.255.255.0`.

pages/common/distccd.md
```diff
@@ -0,0 +1,24 @@
...
+
+`distccd --daemon --allow-private`
+
+- Start a daemon, accepting connections from a specific network address or address range:
+
+`distccd --daemon --allow {{ip_address|network_prefix}}`
+
+- Start a daemon with a maximum of 4 tasks at any time and lowered priority (niceness = 5):
+
+`distccd --daemon --jobs {{4}} --nice {{5}}`
```
nicokosi, line 20: I think we should either use hard-coded values:
```
`distccd --daemon --jobs 4 --nice 5`
```
or variables:
```
`distccd --daemon --jobs {{maximum_job_number}} --nice {{no_idea_how_to_name_it_but_I_am_sure_you_know}}`
```

tomek-szczesny: I experimented with a few formats of this example and found this one to be the cleanest for quick reference.
Also, the [style guide](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/style-guide.md) says:

> If the example is clearer with an actual value rather than a generic placeholder, use the actual value. For example, use iostat {{2}} rather than iostat {{interval_in_secs}}.

marchersimon: I think we can keep it like that. One thing we could do is change it to maximum priority. Right now, I don't have a feeling for what the niceness really is. If I knew the maximum value, it would be easier to come up with a lower one. (assuming it goes from 0 to max)

tomek-szczesny, line 20: Suggested modification for a point risen by @marchersimon:

```suggestion
- Start a daemon with a maximum of 4 tasks at any time and the lowest priority (niceness = 19):

`distccd --daemon --jobs {{4}} --nice {{19}}`
```

marchersimon: I didn't know it was that complicated. In that case I think you got a point with 
> For those who don't know how to handle niceness, there is a ready example how to reduce priority of a daemon below "normal". For those who know what niceness is, that's a clear indication that they can put any valid value they want.

Maybe we should get some other opinions (@bl-ue, @navarroaxel, @nicokosi)

tomek-szczesny: I really don't think a novice user needs to know all that. Any priority lower than normal will be fine for most user cases. How often do you adjust priority of the processes anyway? ;)

marchersimon: Since this change turns out to complicate the example quite a bit, I think we should keep the original.

marchersimon (approved):
Looks good, thanks @tomek-szczesny

pages/common/distccd.md
```diff
@@ -0,0 +1,24 @@
+# distccd
+
+> Server daemon for the distcc distributed compiler.
+> More information: <https://distcc.github.io>.
+
+- Start a daemon with default settings:
```
bl-ue, line 6:
```suggestion
- Start a daemon with the default settings:
```

pages/common/distccd.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Start a daemon, accepting connections from a specific network address or address range:
+
+`distccd --daemon --allow {{ip_address|network_prefix}}`
+
+- Start a daemon with a maximum of 4 tasks at any time and lowered priority (niceness = 5):
+
+`distccd --daemon --jobs {{4}} --nice {{5}}`
+
+- Start a daemon and register via mDNS/DNS-SD (Zeroconf):
```
bl-ue, line 22:
```suggestion
- Start a daemon and register it via mDNS/DNS-SD (Zeroconf):
```

pages/common/distccd.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Start a daemon, accepting connections from private network IPv4 ranges:
+
+`distccd --daemon --allow-private`
+
+- Start a daemon, accepting connections from a specific network address or address range:
+
+`distccd --daemon --allow {{ip_address|network_prefix}}`
+
+- Start a daemon with a maximum of 4 tasks at any time and lowered priority (niceness = 5):
```
bl-ue, line 18: `niceness = 5` seems redundant to me, because of `--nice 5`
```suggestion
- Start a daemon with lowered priority that can run a maximum of 4 tasks at a time:
```

pages/common/distccd.md
```diff
@@ -0,0 +1,24 @@
+# distccd
+
+> Server daemon for the distcc distributed compiler.
+> More information: <https://distcc.github.io>.
+
+- Start a daemon with default settings:
+
+`distccd --daemon`
+
+- Start a daemon, accepting connections from private network IPv4 ranges:
```
bl-ue, line 10:
```suggestion
- Start a daemon, accepting connections from private IPv4 ranges:
```

tomek-szczesny: I'm not entirely sure. Here's my [source](https://wiki.archlinux.org/title/Distcc#Volunteers):

> At a minimum, add the --allow-private switch which covers a number of ipv4 private network ranges.

Perhaps that would be the preferred form?

```suggestion
- Start a daemon, accepting connections from IPv4 private network ranges:
```

Waples (approved):

bl-ue (commented):
Thank you @tomek-szczesny!

sbrl (approved):
Thanks, @tomek-szczesny! I fixed a very minor grammatical typo, if that's ok?

tomek-szczesny:
> Thanks, @tomek-szczesny! I fixed a very minor grammatical typo, if that's ok?

Of course. It's surprising that there were still errors in there. :)
Thanks everyone for fixing this page. It does look much better in the end.

tomek-szczesny:
> Thanks, @tomek-szczesny! I fixed a very minor grammatical typo, if that's ok?

Of course. It's surprising that there were still errors in there. :)
Thanks everyone for fixing this page. It does look much better in the end.

marchersimon (approved):

PR #5914, opened 2021-05-08
aria2c clarifying default behaivour and adding usual command-line behaivour setting option.

gurditsbedi:

### DIFF
--- a/pages/common/aria2c.md
+++ b/pages/common/aria2c.md
@@ -12,7 +12,7 @@
 
 `aria2c -o {{filename}} {{url}}`
 
-- Download from multiple sources:
+- Download from multiple sources each URI pointing to the same file:
 
 `aria2c {{url_1}} {{url_2}}`
 
--- a/pages/common/aria2c.md
+++ b/pages/common/aria2c.md
@@ -12,6 +12,10 @@
 
 `aria2c -o {{filename}} {{url}}`
 
+- Download parallel from multiple sources each URI pointing to a different file:
+
+`aria2c -Z {{url_1}} {{url_2}}`
+
 - Download from multiple sources each URI pointing to the same file:
 
 `aria2c {{url_1}} {{url_2}}`


pages/common/aria2c.md
```diff
@@ -12,7 +12,11 @@
 
 `aria2c -o {{filename}} {{url}}`
 
-- Download from multiple sources:
+- Download parallel from multiple sources each URI pointing to a different file:
```
navarroaxel, line 15: I think we can simplify this example description:
```suggestion
- Download multiple files in parallel:
```

pages/common/aria2c.md
```diff
@@ -12,7 +12,11 @@
...
+
+`aria2c -Z {{url_1}} {{url_2}}`
```
navarroaxel, line 17: We prefer long arguments if there are available
```suggestion
`aria2c --force-sequential {{url_1}} {{url_2}}`
```
Could you please expand the other examples?

pages/common/aria2c.md
```diff
@@ -12,7 +12,11 @@
 
 `aria2c -o {{filename}} {{url}}`
```
navarroaxel, line 13:
```suggestion
`aria2c --out={{filename}} {{url}}`
```

navarroaxel (changes requested):

pages/common/aria2c.md
gurditsbedi: I partially agree with you, I had given the longer description so that one can clear understand what happens with and without `--force-sequential` option, which is written just below it. 

Keeping in mind that most people are habituated with using `wget` the above reason makes more sense.

pages/common/aria2c.md
gurditsbedi: I have expanded all the short arguments I saw. Is there a way to specify both short and long arguments in tldr?

navarroaxel: Nop :disappointed:

pages/common/aria2c.md
gurditsbedi: From my personal experience using `aria2c` for 1 year and more. I think keeping the longer variant will be better.

bl-ue: @gurditsbedi I'm not sure why longer description is better, because that's all implied by the short description.  No one would think that it's downloading the same file several times in parallel 😄 — on the contrary, everyone would think that `aria2c` would download each file pointed to by the URLs at the same time.  Is this incorrect?

gurditsbedi: I got what ya saying @bl-ue , the file will be downloaded parallel, but there is limit to parallel downloads by default which is `5` and is controlled `--max-concurrent-downloads` option. 

I don't have any issue with the shorter description.

Should I add a tldr for `--max-concurrent-downloads` also?

bl-ue: Sure! That would be great.

gurditsbedi: I have added `--max-concurrent-downloads`.

navarroaxel: @gurditsbedi The pages can't have more than 8 examples.

gurditsbedi: Done. :)

pages/common/aria2c.md
```diff
@@ -10,19 +10,23 @@
...
-- Download the URIs listed in a file:
+- Download the URIs listed in a file with limited concurrent connections:
 
-`aria2c -i {{filename}}`
+`aria2c --input-file={{filename}} --max-concurrent-downloads={{connections_num}}`
 
 - Download with multiple connections:
 
-`aria2c -s {{connections_num}} {{url}}`
+`aria2c --split={{connections_num}} {{url}}`
```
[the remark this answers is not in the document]
gurditsbedi: Oh, I didn't knew that. I just copied pasted the older placeholder name.

pages/common/aria2c.md
```diff
@@ -10,19 +10,23 @@
...
+- Download multiple files in parallel:
+
+`aria2c --force-sequential {{url_1}} {{url_2}}`
+
+- Download from multiple sources each URI pointing to the same file:
 
 `aria2c {{url_1}} {{url_2}}`
 
-- Download the URIs listed in a file:
+- Download the URIs listed in a file with limited concurrent connections:
```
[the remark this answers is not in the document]
gurditsbedi: Yes, you are correct.

pages/common/aria2c.md
```diff
@@ -10,19 +10,23 @@
...
+- Download multiple files in parallel:
+
+`aria2c --force-sequential {{url_1}} {{url_2}}`
+
+- Download from multiple sources each URI pointing to the same file:
 
 `aria2c {{url_1}} {{url_2}}`
 
-- Download the URIs listed in a file:
+- Download the URIs listed in a file with limited parallel downloads:
```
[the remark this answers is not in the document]
gurditsbedi: The official docs seem to use, "URI" instead of "URL".

PR #5915, opened 2021-05-08
style-guide: clarify that any client may have --render

navarroaxel:

### DIFF
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -34,7 +34,7 @@ tldrl -f {{page.md}}
 For other ways to use `tldrl`, such as linting an entire directory, check out (what else!)
 [`tldr tldrl`](https://github.com/tldr-pages/tldr/blob/main/pages/common/tldrl.md)
 
-If you're using the Node.js or Python clients of tldr-pages, you can preview a page locally using the `--render` argument (aka `-f` in Node.js, and `-r` in Python):
+Your client may be able to preview a page locally using the `--render` flag:
 
 ```
 tldr --render {{page.md}}
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -34,10 +34,10 @@ tldrl -f {{page.md}}
 For other ways to use `tldrl`, such as linting an entire directory, check out (what else!)
 [`tldr tldrl`](https://github.com/tldr-pages/tldr/blob/main/pages/common/tldrl.md)
 
-If you're using the Node.js client of tldr-pages, you can preview a page locally using the `-f` flag (aka `--render`):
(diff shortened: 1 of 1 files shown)


contributing-guides/style-guide.md
```diff
@@ -34,10 +34,10 @@ tldrl -f {{page.md}}
 For other ways to use `tldrl`, such as linting an entire directory, check out (what else!)
 [`tldr tldrl`](https://github.com/tldr-pages/tldr/blob/main/pages/common/tldrl.md)
 
-If you're using the Node.js client of tldr-pages, you can preview a page locally using the `-f` flag (aka `--render`):
+If you're using the Node.js or Python clients of tldr-pages, you can preview a page locally using the `--render` argument (aka `-f` in Node.js, and `-r` in Python):
```
marchersimon, line 37:
```suggestion
If you're using the Node.js or Python client of tldr-pages, you can preview a page locally using the `--render` argument (aka `-f` in Node.js, and `-r` in Python):
```

marchersimon, line 37: Is there a reason you replaced `flag` with `argument`?

navarroaxel: I think `flag` is used for one letter arguments, but the correct term is `command line argument`.

marchersimon: Oh, interesting

sbrl: Another suggestion?

```suggestion
If you're using the Node.js or Python tldr-pages clients, you can preview a page locally using the `--render` argument (aka `-f` in Node.js, and `-r` in Python):
```

marchersimon: Sounds good, however it would still be `client` right?

bl-ue:
The `C` client (which is official) [supports `--render`](https://github.com/tldr-pages/tldr-c-client/blob/ffcd18b6a7998fc79777189f58711107ea3d8e16/man/tldr.1#L34-L35) as well.  I wonder how many other clients do 🤔

contributing-guides/style-guide.md
sbrl: `clients` should be plural, because you're referring to multiple clients theere - both the Node.js and Python clients.

navarroaxel: @sbrl should I add the C client to the list? or maybe a generalization like `try --render in your client.`

bl-ue: I'd write this line like this:
```suggestion
Your client may be able to preview a page locally using the `--render` flag:
```
That way,
- We don't need to list the clients
- We don't need to list the possible short options

marchersimon: Sounds like a good option, especially since most users just install what's in the repos and don't know which one they're using.

bl-ue (approved):

marchersimon (approved):
👍🏽

sbrl (approved):

PR #5933, opened 2021-05-10
age: replace `unencrypted` with `decrypted`

CleanMachine1:

Really small PR.
Feel free to close if it is incorrect

### DIFF
--- a/pages/common/age.md
+++ b/pages/common/age.md
@@ -5,19 +5,19 @@
 
 - Generate an encrypted file that can be decrypted with a passphrase:
 
-`age --passphrase --output {{path/to/encrypted_file}} {{path/to/unencrypted_file}}`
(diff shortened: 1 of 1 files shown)


marchersimon:
I'm not sure if `unencrypted` and `decrypted` are exactly the same. `decrypted` kind of implies that it has been encrypted before.

bl-ue:
@CleanMachine1 I'm not sure about this one, because `decrypted` implies "it used to be encrypted", while "unencrypted" means that it has never been encrypted.  While both technically indicate the same end state, they imply different meaning.  I think `unencrypted` is better.

CleanMachine1:
Yes, thats precisely why I mentioned, "feel free to close if incorrect"

CleanMachine1:
Closing since I also agree it makes not a lot of sense

PR #5790, opened 2021-04-19
multiple pages: use manned.org for manpage links

waldyrious:
As a follow-up to #5528 (where the manned.org page for `ifup` was not used since it defaulted to an older/alternate version of the tool, as opposed to the Debian manpage), [the issue was reported](https://code.blicky.net/yorhel/manned/issues/1) in the manned.org project, and it recently got fixed with the introduction of a bias towards Ubuntu/Debian manpages, in addition (and secondary) to their already existing bias towards ArchLinux pages.

So this PR adds the manned.org link not just to the `ifup.md` page, but also to other pages that linked to Debian or ArchLinux manpages until now (except in pages for commands that are indeed Debian-specific or ArchLinux-specific).

- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #5941, opened 2021-05-11
common/*: page edit

Waples:

Ref: saw a while back we were adding / replacing links, and i found these.

### DIFF
--- a/pages/common/runit.md
+++ b/pages/common/runit.md
@@ -1,7 +1,7 @@
 # runit
 
 > 3-stage init system.
-> More information: <https://wiki.archlinux.org/index.php/Runit>.
+> More information: <https://manned.org/runit>.
 
 - Start runit's 3-stage init scheme:
 
--- a/pages/common/at.md
+++ b/pages/common/at.md
@@ -2,7 +2,7 @@
 
 > Execute commands once at a later time.
 > Service atd (or atrun) should be running for the actual executions.
-> More information: <https://man.archlinux.org/man/at.1>.
+> More information: <https://manned.org/at>.
 
 - Execute commands from standard input in 5 minutes (press `Ctrl + D` when done):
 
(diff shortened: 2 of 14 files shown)


bl-ue:
Does this conflict with #5790?

@CleanMachine1 a lot of reasoning went behind our decision to use manned.org, see https://github.com/tldr-pages/tldr/issues/5510.

Waples:
@bl-ue ah good call, I'll need to run a diff between these two PRs to check.

@CleanMachine1 As @bl-ue said. (I'm also an Arch Linux user and like the arch linux man, but that bias is not approperiat here.)

CleanMachine1:
Good choice

Waples:
@bl-ue i'm already seeing some weird stuff between `main`, `manned-org` and my branch. In main for example; the link for `pages/common/dig.md` is `<https://manpages.debian.org/dnsutils/dig.1.html>.` while in #5790 it was still archlinux man.

pages/common/runit.md
```diff
@@ -1,7 +1,7 @@
 # runit
 
 > 3-stage init system.
-> More information: <https://wiki.archlinux.org/index.php/Runit>.
+> More information: <https://manned.org/runit>.
```
sbrl, line 4: This would be a more appropriate link here, since runit does actually have a homepage.
```suggestion
> More information: <http://smarden.org/runit>.
```

sbrl (changes requested):

pages/common/runit.md
Waples: Still waiting on #5790

bl-ue:
I'm going to close this for right now, because it seems to me like https://github.com/tldr-pages/tldr/pull/5790 covers all of these pages, and if it doesn't, we can reopen it!

PR #5960, opened 2021-05-13
git-archive-file: add page

CleanMachine1:

### DIFF
--- a/pages/common/git-archive-file.md
+++ b/pages/common/git-archive-file.md
@@ -4,6 +4,6 @@
 > Part of `git-extras`.
 > More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-archive-file>.
 
-- Create an archive `zip` copy of the current branch:
+- Pack the currently checked out commit into a zip archive:
 
 `git archive-file`
--- a/pages/common/git-archive-file.md
+++ b/pages/common/git-archive-file.md
@@ -4,6 +4,6 @@
 > Part of `git-extras`.
 > More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-archive-file>.
 
-- Copy HEAD branch into a zip archive:
+- Create an archive `zip` copy of the current branch:
 
 `git archive-file`
--- a/pages/common/git-archive-file.md
+++ b/pages/common/git-archive-file.md
@@ -1,6 +1,6 @@
 # git archive-file
 
-> Copy the current Git branch into a zip archive.
+> Export all the files of the current Git branch into a zip archive.
 > Part of `git-extras`.
 > More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-archive-file>.
 
--- a/pages/common/git-archive-file.md
+++ b/pages/common/git-archive-file.md
@@ -1,0 +1,9 @@
+# git archive-file
+
+> Copy current checked-out branch into a zip archive.
+> Part of `git-extras`.
+> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-archive-file>.
+
+- Copy HEAD branch into a zip archive:
+
+`git archive-file`
--- a/pages/common/git-archive-file.md
+++ b/pages/common/git-archive-file.md
@@ -1,6 +1,6 @@
 # git archive-file
 
-> Copy current checked-out branch into a zip archive.
+> Copy the current Git branch into a zip archive.
 > Part of `git-extras`.
 > More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-archive-file>.
 
--- a/pages.zh/linux/boltctl.md
+++ b/pages.zh/linux/boltctl.md
@@ -1,0 +1,28 @@
+# boltctl
+
+> 控制雷电 (thunderbolt) 设备。
+> 更多信息：<https://manned.org/boltctl>.
+
+- 列出已连接并授权的设备：
+
+`boltctl`
+
+- 列出已连接的设备，且包含未授权的设备：
+
+`boltctl list`
+
+- 临时授权一个设备：
+
+`boltctl authorize {{设备uuid}}`
+
+- 授权并记住一个设备：
+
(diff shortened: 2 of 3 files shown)


pages/common/git-archive-file.md
```diff
@@ -0,0 +1,9 @@
+# git archive-file
+
+> Copy current checked-out branch into a zip archive.
```
navarroaxel, line 3:
```suggestion
> Copy the current Git branch into a zip archive.
```

pages/common/git-archive-file.md
```diff
@@ -0,0 +1,9 @@
...
+> Part of `git-extras`.
+> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-archive-file>.
+
+- Copy HEAD branch into a zip archive:
```
navarroaxel, line 7: `HEAD` is a pointer to the current branch, check: `cat .git/HEAD`.
Do you want to say `last commit` or copy the current working dir with files edited but not committed yet?

CleanMachine1: git archive-file just take a full copy of the current branch you are checked out on (HEAD).

navarroaxel: You can say `the HEAD` but no everyone knows that means `the current branch`.
But in this case is not `the current HEAD` because you can be in a tag, or a detached commit, so:
```suggestion
- Archive the current commit into a zip archive:
```

CleanMachine1: I think that should be 

> Archive the current branch

bl-ue: Yeah, we have issues regarding this Git studff.  I assume it takes a [commit-ish](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefcommit-ishacommit-ishalsocommittish)?

CleanMachine1: https://github.com/tldr-pages/tldr/pull/5960/commits/c3bbf9be5f81a319e1593a01745fa4b6087c5247

sbrl:
Does this archive the current branch, or the currently checked out commit?

CleanMachine1:
can you "checkout" a commit?

If you are on branch main, and you run the git archive-file, you get a .zip with everything on branch main

CleanMachine1:
```

**Comment by navarroaxel** [comment]
> can you "checkout" a commit?
> 
> If you are on branch main, and you run the git archive-file, you get a .zip with everything on branch main

Yes, you can checkout a commit.

@sbrl I tested this command and use the `currently checked out commit`.

File: pages/common/git-archive-file.md (hunk 2 above)
Comment (sbrl) [reply pages/common/git-archive-file.md]: Hrm. I would suggest something like this:

```suggestion
- Pack the currently checked out commit into a zip archive:
```

bl-ue: A branch is really nothing in it of itself, it's just a pointer to a commit, so @sbrl's suggestion is correct. See [this](https://stackoverflow.blog/2021/04/05/a-look-under-the-hood-how-branches-work-in-git/) for a great explanation.

CleanMachine1: https://github.com/tldr-pages/tldr/pull/5960/commits/5284ad6fa74e5e1e6e9c35776c2de5818f794f1f

navarroaxel (approved):

pages/common/git-archive-file.md
```diff
@@ -0,0 +1,9 @@
+# git archive-file
+
+> Copy the current Git branch into a zip archive.
```
bl-ue, line 3: thoughts?
```suggestion
> Export all the files of the current Git branch into a zip archive.
```

bl-ue (commented):
One comment, otherwise looks good!

Note: the page description and the first (or only) example's description _do not_ need to be the same :)

pages/common/git-archive-file.md
CleanMachine1: Seems ok, however copy is easier to read

bl-ue: 'Tis, but export is more descriptive ;)

CleanMachine1: @marchersimon thoughts?

navarroaxel: `Export` is better, also I think it should say `current Git commit` @bl-ue

CleanMachine1:
```suggestion
> Export all the files of the current Git commit into a zip archive.
```

marchersimon: Couldn't we just leave out `all the files of`, so we just have:
`Export the current Git commit into a zup archive` ?

CleanMachine1: I personally say no, since to me, exporting a commit would be just taking the most recent commits changes.

bl-ue: I agree with you @navarroaxel.  @CleanMachine1 branches in Git are really simply pointers to commits, and nothing else.  You can also be checked out on a commit, `e.g. git checkout a1b2c3d`, which is called _detached HEAD._  [This StackOverflow article](https://stackoverflow.blog/2021/04/05/a-look-under-the-hood-how-branches-work-in-git/) I found very explanatory.

bl-ue: I was thinking of `all the files of` because, as @CleanMachine1 pointed out, it may not be obvious to a new user that it's exporting the entire tree.

CleanMachine1: Can I have a suggestion to add, or tell me which

marchersimon: I think @bl-ue's original suggestion is fine.

bl-ue (approved):
I kinda agree with @CleanMachine1's [point](https://github.com/tldr-pages/tldr/pull/5960#discussion_r633124911) about using `commit`, so I think we should keep `branch` in the description. LGTM

marchersimon (approved):
Looks good.

Issue #5962, opened 2021-05-14
 getting the tldr Debian package fixed


patricedenis:
As we discussed in #5937, lots of people are using the `tldr-hs` client that is installed by default when invoking `apt install tldr`.
Unfortunately, the current Debian `tldr` package is not synchronized with the upstream `tldr-hs` client maintained by @psibi.

> More information : <[https://github.com/psibi/tldr-hs](https://github.com/psibi/tldr-hs)>. 

This issue is then here to start a proposal to get this problem fixed.

For information, I'm willing to assign myself to this issue even if I will really like to collaborate with others here as well.
It will be my first attempt to packaging so don't expect this to be fixed in just a day :smiley:

navarroaxel:
Are you thinking about create a bug report into https://bugs.launchpad.net/ubuntu?

I think we should propose to switch to https://github.com/tldr-pages/tldr-python-client like the Arch Linux family uses. Because it is the only client that supports `bash completion` (tab-tab) using the `python-argcomplete` package, as far as I know.

psibi:
> I think we should propose to switch to https://github.com/tldr-pages/tldr-python-client like the Arch Linux family uses. Because it is the only client that supports bash completion (tab-tab) using the python-argcomplete package, as far as I know.

Note that the Haskell one supports bash completion, the steps are here: https://github.com/psibi/tldr-hs/issues/28#issuecomment-686448088 But I'm not sure if the distro's package them.

Also, since it uses optparse-applicative - it supports completion for zsh and fish too IIRC.

bl-ue:
>I think we should propose to switch to https://github.com/tldr-pages/tldr-python-client like the Arch Linux family uses.

What about `tldr-py`?  Should we make it an alias (or the apt equivalent)?  It's currently https://github.com/lord63/tldr.py

https://github.com/tldr-pages/tldr/issues/5937#issuecomment-838880498

navarroaxel:
@psibi why the `completion` is not documented in a more visible place? like the README.md

psibi:
@navarroaxel Yeah, I definitely need to improve the docs. I will see if I can add a mention of it (any PR's welcome!).

CleanMachine1:
Any progress @patricedenis

Looking forward to the day where I see it appear in my apt repo

patricedenis:
[...]

PR #5965, opened 2021-05-14
tldr: refresh

CleanMachine1:

Hi, so just for a test I ran the command `tldr tldr` and got the page.

However the page has some things wrong to the standard.

Watch out for comments on my commit

### DIFF
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -7,14 +7,14 @@
 
 `tldr {{command}}`
 
-- Show the tldr page for `cd`, overriding the default platform:
+- Show the tldr page for `cd`, overriding the default platform(if the client supports -p):
 
 `tldr -p {{windows}} {{cd}}`
 
-- Show the tldr page for a subcommand:
+- Show the tldr page for a sub-command:
 
 `tldr {{git-checkout}}`
 
 - Update local pages (if the client supports caching):
 
-`tldr -u`
+`tldr --update`


pages/common/tldr.md
```diff
@@ -7,14 +7,14 @@
 
 `tldr {{command}}`
 
-- Show the tldr page for `cd`, overriding the default platform:
+- Show the tldr page for `cd`, overriding the default platform(if the client supports -p):
```
CleanMachine1, line 10: The Ubuntu haskell client does not

pages/common/tldr.md
```diff
@@ -7,14 +7,14 @@
...
 
 `tldr -p {{windows}} {{cd}}`
 
-- Show the tldr page for a subcommand:
+- Show the tldr page for a sub-command:
```
CleanMachine1, line 14: small error, feel free to change this back

pages/common/tldr.md
```diff
@@ -7,14 +7,14 @@
...
 
 `tldr {{git-checkout}}`
 
 - Update local pages (if the client supports caching):
 
-`tldr -u`
+`tldr --update`
```
CleanMachine1, line 20: the -u is not included in the haskell platform either, only --update is, so its probably better to show the command for the more used platform

pages/common/tldr.md
bl-ue: I'm not sure about this — if it's really more correct, then we should change it across all of the pages.

pages/common/tldr.md
bl-ue: 0.4 of https://github.com/psibi/tldr-hs doesn't, but 0.9 (the latest version) does, and we're working on that (#5962)

Also, [`-p` is part of the client spec](https://github.com/tldr-pages/tldr/blob/ddf755f35520ab3b128255edd5da32a85f15bd02/CLIENT-SPECIFICATION.md#arguments), so we shouldn't have to do this.

pages/common/tldr.md
bl-ue: This is not the first to make this change, but we can't do this because the long options are optional per the [client spec](https://github.com/tldr-pages/tldr/blob/ddf755f35520ab3b128255edd5da32a85f15bd02/CLIENT-SPECIFICATION.md#arguments):

>Clients MAY choose to only implement the short version of an option, ignoring the long form.

pages/common/tldr.md
CleanMachine1: Ok, will do. 
Is their an easier way to do it than doing a commit?

CleanMachine1:
I am going to close this PR, it is wrong and has no benefits since it conflicts with others.

Thumbs up if I should close

pages/common/tldr.md
bl-ue: Well, you can add a suggestion and apply it right here...
```suggestion
- Show the tldr page for `cd`, overriding the default platform:
```

PR #5967, opened 2021-05-14
dconf, dconf-write: add page

makusu2:

### DIFF
--- a/pages/linux/dconf-write.md
+++ b/pages/linux/dconf-write.md
@@ -17,7 +17,7 @@
 
 - Write an array to a dconf path:
 
-`dconf write {{/example/dconf/path}} '[{{"My First Value", "My Second Value"}}]'`
+`dconf write {{/example/dconf/path}} "[{{'My First Value', 'My Second Value'}}]"`
 
 - Write an empty array to a dconf path:
 
--- a/pages/linux/dconf-write.md
+++ b/pages/linux/dconf-write.md
@@ -9,7 +9,7 @@
 
 - Write a boolean to a dconf path:
 
-`dconf write {{/example/dconf/path}} {{true}}`
+`dconf write {{/example/dconf/path}} {{true|false}}`
 
 - Write an integer to a dconf path:
 
--- a/pages/linux/dconf-write.md
+++ b/pages/linux/dconf-write.md
@@ -1,7 +1,7 @@
 # dconf write
 
 > Write a value to a dconf database path.
-> More information: <https://https://developer.gnome.org/dconf>.
+> More information: <https://developer.gnome.org/dconf>.
 
 - Write a string to a dconf path (note the nested quotes):
 
--- a/pages/linux/dconf-write.md
+++ b/pages/linux/dconf-write.md
@@ -1,0 +1,24 @@
+# dconf write
+
+> Write a value to a dconf database path.
+> More information: <https://https://developer.gnome.org/dconf>.
+
+- Write a string to a dconf path (note the nested quotes):
+
+`dconf write {{/example/dconf/path}} "'{{Example Value}}'"`
+
+- Write a boolean to a dconf path:
+
+`dconf write {{/example/dconf/path}} {{true}}`
+
+- Write an integer to a dconf path:
+
+`dconf write {{/example/dconf/path}} {{16}}`
+
+- Write an array to a dconf path:
+
+`dconf write {{/example/dconf/path}} '[{{"My First Value", "My Second Value"}}]'`
+
+- Write an empty array to a dconf path:
+
+`dconf write {{/example/dconf/path}} "@as []"`
--- a/pages/linux/dconf.md
+++ b/pages/linux/dconf.md
@@ -1,6 +1,7 @@
 # dconf
 
-> Simple tool for manipulating a dconf database.
+> Simple tool for manipulating dconf databases.
+> See also `dconf write`.
 > More information: <https://developer.gnome.org/dconf>.
 
 - Read the value of a dconf path:
--- a/pages/linux/dconf.md
+++ b/pages/linux/dconf.md
@@ -1,0 +1,16 @@
+# dconf
+
+> Simple tool for manipulating a dconf database.
+> More information: <https://https://developer.gnome.org/dconf>.
+
+- Read the value of a dconf path:
+
+`dconf read {{/example/dconf/path}}`
+
+- List contents of a dconf path:
+
+`dconf list {{/example/dconf/path}}`
+
+- Watch for dconf database changes in a path and subpaths:
+
+`dconf watch {{/example/dconf/path}}`
--- a/pages/linux/dconf.md
+++ b/pages/linux/dconf.md
@@ -4,7 +4,7 @@
 > See also `dconf write`.
 > More information: <https://developer.gnome.org/dconf>.
 
-- Read the value of a dconf path:
+- Print the value of a dconf path:
 
 `dconf read {{/example/dconf/path}}`
 
--- a/pages/linux/dconf.md
+++ b/pages/linux/dconf.md
@@ -1,7 +1,7 @@
 # dconf
 
 > Simple tool for manipulating a dconf database.
-> More information: <https://https://developer.gnome.org/dconf>.
+> More information: <https://developer.gnome.org/dconf>.
 
 - Read the value of a dconf path:
 


pages/linux/dconf.md
```diff
@@ -0,0 +1,16 @@
+# dconf
+
+> Simple tool for manipulating a dconf database.
```
CleanMachine1, line 3:
```suggestion
> Simple tool for reading a dconf database.
```

CleanMachine1 (changes requested):

pages/linux/dconf.md
makusu2: It can read, but it can also write and do other things :thinking:

CleanMachine1: So it can, ignore this, I only read the first usage and the filename.
Sorry, new to the job!

CleanMachine1 (approved):

CleanMachine1:
I'd say this is a really good commit, however another maintainer will probably come and suggest some changes, however all a learning experience for me as well

pages/linux/dconf-write.md
```diff
@@ -0,0 +1,24 @@
+# dconf write
+
+> Write a value to a dconf database path.
+> More information: <https://https://developer.gnome.org/dconf>.
```
bl-ue, line 4:
```suggestion
> More information: <https://developer.gnome.org/dconf>.
```

pages/linux/dconf-write.md
```diff
@@ -0,0 +1,24 @@
+# dconf write
+
+> Write a value to a dconf database path.
```
bl-ue, line 3: This is a little confusing to me.  The below commands take a value, which is perfectly intuitive, but they don't take anything like a key.  What's this stuff about "database path"?

pages/linux/dconf-write.md
```diff
@@ -0,0 +1,24 @@
...
+
+`dconf write {{/example/dconf/path}} {{16}}`
+
+- Write an array to a dconf path:
+
+`dconf write {{/example/dconf/path}} '[{{"My First Value", "My Second Value"}}]'`
+
+- Write an empty array to a dconf path:
+
+`dconf write {{/example/dconf/path}} "@as []"`
```
bl-ue, line 24:
```suggestion
`dconf write {{/example/dconf/path}} "{{@as []}}"`
```

pages/linux/dconf-write.md
CleanMachine1: dconf is an internal settings type of thing, you have probably used it, I don't know.

However it is a big database of internal settings

pages/linux/dconf-write.md
makusu2: Whoops

pages/linux/dconf.md
```diff
@@ -0,0 +1,16 @@
...
+> More information: <https://https://developer.gnome.org/dconf>.
```
CleanMachine1, line 4:
```suggestion
> More information: <https://developer.gnome.org/dconf>.
```

pages/linux/dconf-write.md
makusu2: The path is the key. Basically like the registry on Windows

pages/linux/dconf-write.md
CleanMachine1: Done it for the other one as well, I have added the suggestion for it

pages/linux/dconf-write.md
makusu2: Since this example shows how to enter an empty array, this should be entered exactly as written

pages/linux/dconf-write.md
bl-ue: In `/example/dconf/path`, is "path" what I'm thinking the "key" is?

>you have probably used it

🙂 Nope, never.  I've never actually even used Linux as my personal desktop, _ever_! 😄

pages/linux/dconf-write.md
bl-ue: Oh, I see.  Then never mind.

CleanMachine1: not too sure about this one, @makusu2 what do you think

I think it doesn't need to be since its not a "variable" its writing a blank array

pages/linux/dconf-write.md
makusu2: > In /example/dconf/path, is "path" what I'm thinking the "key" is?

Exactly :+1:

CleanMachine1: What OS do you use?
I would recommend Linux, much better for programming

pages/linux/dconf-write.md
CleanMachine1: Feel free to resolve

makusu2:
@CleanMachine1 Did you delete your original comment?

This is just my opinion, but even if you've made a comment with a mistake, you shouldn't delete it. You can edit it to add a strikethrough and explain why you were mistaken, but it's generally not a good idea to just flat out delete it.

CleanMachine1:
Yes, I did, didn't want anyone adding the suggestion, but instead I deleted my comment, didn't delete the reply though, don't know how that happened

pages/linux/dconf-write.md
```diff
@@ -0,0 +1,24 @@
+# dconf write
+
+> Write a value to a dconf database path.
+> More information: <https://developer.gnome.org/dconf>.
+
+- Write a string to a dconf path (note the nested quotes):
```
CleanMachine1, line 6: is that last bit needed?

pages/linux/dconf-write.md
```diff
@@ -0,0 +1,24 @@
...
+
+`dconf write {{/example/dconf/path}} "'{{Example Value}}'"`
+
+- Write a boolean to a dconf path:
+
+`dconf write {{/example/dconf/path}} {{true}}`
```
CleanMachine1, line 12:
```suggestion
`dconf write {{/example/dconf/path}} {{true_or_false}}`
```

Don't commit this until we have had others opinions

pages/linux/dconf-write.md
makusu2: I can remove it if you think I should, I figured I'd point out that the *value* of the string needs to be wrapped in single quotes.

pages/linux/dconf-write.md
makusu2: If anything it'd be `true|false`, but I've seen booleans represented by specifying both or just one

CleanMachine1: from what I have seen we use snake to separate possible commands, some may not understand what the pipe means for usage

makusu2: Actually I only see `true|false` in `pages/common/qutebrowser.md`

makusu2: Can you link an example? I'm getting nothing from `rg -F 'true_or_false'`

CleanMachine1: hmm, yes. Again we need to wait for others opinion

bl-ue:
>This is just my opinion, but even if you've made a comment with a mistake, you shouldn't delete it. You can edit it to add a strikethrough and explain why you were mistaken, but it's generally not a good idea to just flat out delete it.

@CleanMachine1 I agree with @makusu2.  Don't delete comments because it renders the conversion almost completely useless. I was quite confused about that thread before I realized that you had delete a comment.  In addition, don't even hide them — GitHub has a bug where you can't _unhide_ them, and it also really interrupts the reading flow.

pages/linux/dconf-write.md
bl-ue: I'd prefer `true|false` here, because it indicates that you can only use those two values, and nothing else.

makusu2:
```suggestion
`dconf write {{/example/dconf/path}} {{true|false}}`
```

pages/linux/dconf-write.md
bl-ue: I have a Mac, so I use macOS Big Sur 11.3 :)

I know Linux is great, but I've always liked Apple 

I have Windows and Linux VMs, so it's not as if I don't know them.  I'm really familiar with both, to tell the truth, but `dconf` I don't recall using ever. 🤔

pages/linux/dconf-write.md
marchersimon: I think it's safer to keep it, since it's easy to overlook:
image

marchersimon: Also, why is the single quote nested in the double one, but in array example it's the other way around?

pages/linux/dconf.md
marchersimon, line 3:
```suggestion
> Simple tool for manipulating dconf databases.
> See also `dconf write`.
```
I think the plural generally sounds better in descriptions like this. Also, as long as we only have one sub-command page, I think we can cross-reference them.

pages/linux/dconf-write.md
marchersimon, line 3:
```suggestion
> Write a value to a dconf database path.
> See also `dconf`.
```

pages/linux/dconf.md
```diff
@@ -0,0 +1,16 @@
+# dconf
+
+> Simple tool for manipulating a dconf database.
+> More information: <https://developer.gnome.org/dconf>.
+
+- Read the value of a dconf path:
```
marchersimon, line 6:
```suggestion
- Read the value of a dconf path and print it:
```
```suggestion
- Print the value of a dconf path:
```
Is one of those more correct?

CleanMachine1: 2nd is better

bl-ue: I like the second one better too.

pages/linux/dconf.md
bl-ue: >Also, as long as we only have one sub-command page, I think we can cross-reference them.

Sounds like a good idea!

pages/linux/dconf-write.md
bl-ue: I thinks it's fine to keep it too.

CleanMachine1: Ok, however I don't think it makes sense

What do you think about 
```suggestion
- Write a string to a dconf path (within the nested quotes):
```

bl-ue: I don't really see how that adds any value 🤔

CleanMachine1: It makes more sense, in my opinion,

marchersimon: With `note` it sounds like a warning and it can be read as a seperate sentence. With `within` in sounds like a part of the sentence before it and will likely be read as such.

makusu2: > Also, why is the single quote nested in the double one, but in array example it's the other way around?

Good question, it doesn't matter in this example but I'll edit the array example to match :+1:

pages/linux/dconf-write.md
```diff
@@ -0,0 +1,24 @@
...
+
+`dconf write {{/example/dconf/path}} {{true|false}}`
+
+- Write an integer to a dconf path:
+
+`dconf write {{/example/dconf/path}} {{16}}`
+
+- Write an array to a dconf path:
+
+`dconf write {{/example/dconf/path}} '[{{"My First Value", "My Second Value"}}]'`
```
makusu2, line 20: https://github.com/tldr-pages/tldr/pull/5967#discussion_r632982436
```suggestion
`dconf write {{/example/dconf/path}} "[{{'My First Value', 'My Second Value'}}]"`
```

pages/linux/dconf.md
makusu2: > as long as we only have one sub-command page, I think we can cross-reference them.

I was wondering about this when I made the PR, I looked for any other page that referenced the subcommand and couldn't find any, so I figured it wasn't advised. It might be a good idea to add that to the documentation for pages with a single sub-command :rocket:

pages/linux/dconf-write.md
makusu2: I agree with referencing `dconf-write` in `dconf`, but isn't this one implied?

Especially with the rule of:

> You should always add a base page

pages/linux/dconf-write.md
CleanMachine1: push this commit then edit the array

makusu2: > push this commit then edit the array

Push what commit?

But yeah, I've already edited the array example

pages/linux/dconf-write.md
bl-ue: Yeah, it is implied 🤔

bl-ue:
(off topic) Oh, you're in Boston @makusu2? I'm only 77 miles away in northern Connecticut! :D

pages/linux/dconf-write.md
CleanMachine1: The suggested change above

makusu2: I agree with @bl-ue, I don't think the suggested change is more clear.

I also agree with @marchersimon, but I'm not sure whether they're advocating for or against the suggestion :thinking:

makusu2:
> (off topic) Oh, you're in Boston @makusu2? I'm only 77 miles away in northern Connecticut! :D

Oh cool! New England gang :sunglasses:

pages/linux/dconf-write.md
CleanMachine1: Resolve?

pages/linux/dconf-write.md
bl-ue: I'd prefer to keep it how it is (`note`).

pages/linux/dconf-write.md
makusu2: > Resolve?

I'd wait for @marchersimon to respond, but I think so

bl-ue:
Yeah, I don't know of any other tldr maintainer in the US besides @MasterOdin 🤔

bl-ue (approved):

CleanMachine1 (approved):

marchersimon (approved):
Looks good, thanks @makusu2

PR #5969, opened 2021-05-15
stl2gts: add page

marchersimon:


pages/common/stl2gts.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Convert an STL file to a GTS file and do not merge vertices:
+
+`stl2gts --nomerge < {{path/to/file.stl}} > {{path/to/file.gts}}`
+
+- Convert an STL file to a GTS file and display surface statistics:
+
+`stl2gts --verbose < {{path/to/file.stl}} > {{path/to/file.gts}}`
+
+- Print help for `stl2gts`:
```
bl-ue, line 22: I like this better but what do you think?
```suggestion
- Display help:
```

CleanMachine1: Can we have a standard, Print should be for when it just outputs it, while Display is for things which can be interacted with, such as man pages.

pages/common/stl2gts.md
```diff
@@ -0,0 +1,24 @@
+# stl2gts
+
+> Convert STL files to GTS (GNU triangulated surface library) format.
```
CleanMachine1, line 3:
```suggestion
> Convert STL files to GTS (GNU triangulated surface library) file format.
```

marchersimon: I think `GTS format` is legit. It's also what is used in the man page.

pages/common/stl2gts.md
marchersimon: Standard sounds good, however in this case, it just prints it with no possible interaction.

bl-ue: I'm inclined towards standards for everything common such as this, like you, and in the past I've proposed many such trivial normalizations, but now I think it might be better to just suggest differenting wording on a PR-by-PR basis, because—

- Not everyone associates the same meaning with the same words,
- If the pages are more identical terms of wording, they seem more machine generated (which isn't what we want),
- Because contributors themselves might have something to say about the usage of verbs,
- A word might be more appropriate for one page but less fit for another, and
- Lots of standards can intimidate contributors and make contribution more cumbersome than it already is.

pages/common/stl2gts.md
bl-ue: It might be good to also add "the" before "GTS."

marchersimon: While that's technically correct, it sonuds really bad. Are you sure?

CleanMachine1: > It might be good to also add "the" before "GTS."

I disagree, it makes GTS the noun, rather than adjective

bl-ue: Something about the current wording feels like it's missing an article or preposition.  Maybe "Convert STL files into the GTS ..."?

CleanMachine1: I know I said I disagree, yes, after reading a few more times over, it makes more sense

bl-ue (commented):
STL is pretty cool.  I'm dealing with it a lot at work lately :)

bl-ue (approved):

pages/common/stl2gts.md
marchersimon: I neither like `the GTS`, nor `GTS file format`, but in combination they are fine.

marchersimon:
3d printing? It's the only thing I know they are used for.

bl-ue:
You're right — I shouldn't say "a lot" because it's not that much.  I'm working on a program that processes generated STL files...not actually _using_ them for 3D printing.

pages/common/stl2gts.md
CleanMachine1: Resolving since applied.

CleanMachine1 (approved):

CleanMachine1:
Waiting for your opinion Axel, then we can merge

nicokosi (approved):
Looks good to me! 👍

pages/common/stl2gts.md
CleanMachine1: I think it should be fine, its already ok in my opinion, I am ready to merge and resolve this, do you all agree?

Issue #5137, opened 2021-01-11
Let's document: tj/git-extras


bl-ue:
We should add the commands from [tj/git-extras](https://github.com/tj/git-extras) (73):

- [x] [`abort`](https://github.com/tj/git-extras/blob/main/Commands.md#git-abort) → #6243
- [x] [`alias`](https://github.com/tj/git-extras/blob/main/Commands.md#git-alias) → #5979
- [x] [`archive-file`](https://github.com/tj/git-extras/blob/main/Commands.md#git-archive-file) → #5960
- [x] [`authors`](https://github.com/tj/git-extras/blob/main/Commands.md#git-authors) → #5951
- [x] [`browse`](https://github.com/tj/git-extras/blob/main/Commands.md#git-browse) → #5959
- [x] [`browse-ci`](https://github.com/tj/git-extras/blob/main/Commands.md#git-browse-ci) → #8500
- [x] [`brv`](https://github.com/tj/git-extras/blob/main/Commands.md#git-brv) → #5966
- [x] [`bulk`](https://github.com/tj/git-extras/blob/main/Commands.md#git-bulk) → #8946
- [x] [`changelog`](https://github.com/tj/git-extras/blob/main/Commands.md#git-changelog) → #7359
- [x] [`clear-soft`](https://github.com/tj/git-extras/blob/main/Commands.md#git-clear-soft) → #6616 
- [x] [`clear`](https://github.com/tj/git-extras/blob/main/Commands.md#git-clear) → #6614
- [x] [`coauthor`](https://github.com/tj/git-extras/blob/main/Commands.md#git-coauthor) → #6878
- [x] [`commits-since`](https://github.com/tj/git-extras/blob/main/Commands.md#git-commits-since) → #5968
- [ ] [`continue`](https://github.com/tj/git-extras/blob/main/Commands.md#git-continue) →
- [x] [`contrib`](https://github.com/tj/git-extras/blob/main/Commands.md#git-contrib) → #6064
- [x] [`count`](https://github.com/tj/git-extras/blob/main/Commands.md#git-count) → #5970
- [x] [`cp`](https://github.com/tj/git-extras/blob/main/Commands.md#git-cp) → #5947
- [x] [`create-branch`](https://github.com/tj/git-extras/blob/main/Commands.md#git-create-branch) → #5982
- [x] [`delete-branch`](https://github.com/tj/git-extras/blob/main/Commands.md#git-delete-branch) → #6011
[...]

PR #5970, opened 2021-05-15
git-count: add page

CleanMachine1:

For #5137


pages/common/git-count.md
```diff
@@ -0,0 +1,13 @@
+# git count
+
+> Print a commit count.
```
bl-ue, line 3:
```suggestion
> Print the count of all commits.
```

pages/common/git-count.md
```diff
@@ -0,0 +1,13 @@
...
+> Part of `git-extras`.
+> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-count>.
+
+- Print a total commit count:
```
bl-ue, line 7:
```suggestion
- Print the total commit count:
```

pages/common/git-count.md
```diff
@@ -0,0 +1,13 @@
...
+
+`git count`
+
+- Print a total commit count and individuals commit count:
```
bl-ue, line 11:
```suggestion
- Print the total commit count and the commit count of each contributor:
```

bl-ue (changes requested):

pages/common/git-count.md
```diff
@@ -0,0 +1,13 @@
+# git count
+
+> Print the total number of all commits.
+> Part of `git-extras`.
+> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-count>.
+
+- Print the total commit count:
```
bl-ue, line 7:
```suggestion
- Print the total number of commits:
```

pages/common/git-count.md
```diff
@@ -0,0 +1,13 @@
...
+
+`git count`
+
+- Print the total commit count and the commit count of each contributor:
```
bl-ue, line 11:
```suggestion
- Print the total number of commits and the number of commits per contributor:
```

bl-ue (commented):
Ah, I do prefer "number" to "count" @CleanMachine1. 👍🏻

bl-ue (approved):
Very nice looking.

pages/common/git-count.md
```diff
@@ -0,0 +1,13 @@
+# git count
+
+> Print the total number of all commits.
+> Part of `git-extras`.
+> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-count>.
+
+- Print the total number of commits:
```
navarroaxel, line 7: What if we use `display`?
```suggestion
- Display the total number of commits:
```

pages/common/git-count.md
```diff
@@ -0,0 +1,13 @@
...
+
+`git count`
+
+- Print the total number of commits and the number of commits per contributor:
```
navarroaxel, line 11:
```suggestion
- Display the total number of commits and the number of commits per contributor:
```

pages/common/git-count.md
CleanMachine1: No since it doesn't display, it prints

pages/common/git-count.md
CleanMachine1: No since it doesn't display, it prints

pages/common/git-count.md
CleanMachine1: I'd like to really make the standard that we should use 
Display for when it is an interactive log E.G. a man page
and 
Print for when it is just hard printed E.G. echo 1

navarroaxel: We don't have an standard but IMO `display` is more friendly to new users. In the other hand, I prefer `print` when the output is a not for humans text, like [starship](https://github.com/tldr-pages/tldr/blob/main/pages/common/starship.md)

CleanMachine1: @bl-ue Whats your opinion?

marchersimon: I have to reference to @bl-ue's comment [here](https://github.com/tldr-pages/tldr/pull/5969#discussion_r632982394). 

> I'm inclined towards standards for everything common such as this, like you, and in the past I've proposed many such trivial normalizations, but now I think it might be better to just suggest differenting wording on a PR-by-PR basis, because—
> 
> - Not everyone associates the same meaning with the same words,
> - If the pages are more identical terms of wording, they seem more machine generated (which isn't what we want),
> - Because contributors themselves might have something to say about the usage of verbs,
> - A word might be more appropriate for one page but less fit for another, and
> - Lots of standards can intimidate contributors and make contribution more cumbersome than it already is.

I think even if we would find the perfect word in this context, a bit of individuality differentiates our pages from man pages or other documentation. Also, display and print basically mean the same. I get what you mean with display being more interactive, but that's just mere interpretation, and someone else might not agree with that.

pages/common/git-count.md
```diff
@@ -0,0 +1,13 @@
+# git count
+
+> Print the total number of all commits.
```
nicokosi, line 3: `all` is implicit, no? 😇 
```suggestion
> Print the total number of commits.
```

nicokosi: BTW, what are "all commits"? 🤔

CleanMachine1: It was just an error. using the total beforehand makes the usage seem wrong, thanks for finding it

nicokosi (approved):
:+1:

pages/common/git-count.md
```diff
@@ -0,0 +1,13 @@
...
+
+> Print the total number of commits.
+> Part of `git-extras`.
+> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-count>.
+
+- Print the total number of commits:
+
+`git count`
+
+- Print the total number of commits and the number of commits per contributor:
```
marchersimon, line 11:
```suggestion
- Print the number of commits per contributor and the total number of commits:
```
It's really unsatisfying when the order of the description doesn't match the output 😄

marchersimon (approved):

CleanMachine1:
merged?

CleanMachine1:
I'll take that as a yes. Been plenty of time and have had 2 +1 on my merge comment.

PR #5987, opened 2021-05-18
git-authors: fix example descriptions

navarroaxel:

I re-check the command and don't `replace` the file, only append. Also, if you run the command twice, the authors appears twice.

### DIFF
--- a/pages/common/git-authors.md
+++ b/pages/common/git-authors.md
@@ -1,16 +1,16 @@
 # git authors
 
-> Display a list of committers of a Git repository. Everything except `--list` will append the list of commiters to the `AUTHORS` file.
+> Display a list of committers of a Git repository.
 > Part of `git-extras`.
 > More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-authors>.
-
-- Append the list of committers to the the `AUTHORS` file and open it in the default editor:
-
-`git authors`
 
 - Print a full list of committers to stdout instead of to the `AUTHORS` file:
 
 `git authors --list`
+
+- Append the list of committers to the the `AUTHORS` file and open it in the default editor:
+
+`git authors`
 
 - Append the list of committers, excluding emails, to the the `AUTHORS` file and open it in the default editor:
 
--- a/pages/common/git-authors.md
+++ b/pages/common/git-authors.md
@@ -1,6 +1,6 @@
 # git authors
 
-> Display a list of committers of a Git repository.
+> Generate a list of committers of a Git repository.
 > Part of `git-extras`.
 > More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-authors>.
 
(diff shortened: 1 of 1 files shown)


CleanMachine1:
What about refresh?

pages/common/git-authors.md
```diff
@@ -1,17 +1,17 @@
 # git authors
 
-> Display a log or list of committers of a Git repository. Everything except `--list` will create a new file called `AUTHORS`.
+> Display a list of committers of a Git repository. Everything except `--list` will append the list of commiters to the `AUTHORS` file.
```
marchersimon, line 3: Even tough it does append it, I'm not sure if that's an important detail, since in almost all cases it will create it.

bl-ue:
I normally add `refresh` when the page hasn't (or at least, most of what I'm editing) hasn't been touched in a while.  Given that this page was merged into `main` only 15 minutes ago, I think the old description was fine 😄

pages/common/git-authors.md
navarroaxel: If you have a header in the `AUTHORS` file, I think you don't want to recreate a get a mess with the list.

marchersimon (approved):
LGTM, thanks

pages/common/git-authors.md
bl-ue: Let's move this note to the `--list` example.
```suggestion
> Generate a list of committers of a Git repository.
```
"Generate" because this isn't actually reading it out of some secret existing file.  It's generating it based on the commits.

pages/common/git-authors.md
```diff
@@ -1,17 +1,17 @@
...
 > Part of `git-extras`.
 > More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-authors>.
 
-- Create or replace the `AUTHORS` file with a list of committers and open it in the default editor:
+- Append the list of committers to the the `AUTHORS` file and open it in the default editor:
 
 `git authors`
 
 - Print a full list of committers:
```
bl-ue, line 11:
```suggestion
- Print a full list of committers to stdout instead of to the `AUTHORS` file:
```

CleanMachine1:
:rofl: I didn't mean refresh as in the title, I meant inplace of append...

navarroaxel:
> rofl I didn't mean refresh as in the title, I meant inplace of append...

No, because it doesn't replace the list, so there is no a `refreshing` action.

pages/common/git-authors.md
sbrl: Generally speaking I think this detail might be best served in the example description. If we move `--list` to be first, then put this in the description of the 2nd, then reword the first to imply it's similar to the 2nd but is slightly different.

Waples:
Sidenote: In my notifications I see a request to review, but can't find it here 🤔

Waples (approved):

bl-ue (approved):

CleanMachine1 (approved):

sbrl (approved):
Thanks, @navarroaxel!

PR #5992, opened 2021-05-18
shuf: improve usage example and its description

anakimluke:
The command example printed out 10 random numbers, but that wasn't clear to the user.
I added the -n option to the command and modified the description accordingly.

### DIFF
--- a/pages/common/shuf.md
+++ b/pages/common/shuf.md
@@ -15,6 +15,6 @@
 
 `shuf {{filename}} -o {{output_filename}}`
 
-- Generate random numbers in range:
+- Generate 3 random numbers in the range 1-10:
 
-`shuf -i {{1-10}}`
+`shuf -n {{3}} -i {{1-10}}`
--- a/pages/common/shuf.md
+++ b/pages/common/shuf.md
@@ -15,6 +15,6 @@
 
 `shuf {{filename}} -o {{output_filename}}`
 
-- Generate 3 random numbers in the range 1-10:
+- Generate 3 random numbers in the range 1-10 (inclusive):
 
 `shuf -n {{3}} -i {{1-10}}`
--- a/pages/common/shuf.md
+++ b/pages/common/shuf.md
@@ -17,4 +17,4 @@
 
 - Generate 3 random numbers in the range 1-10 (inclusive):
 
-`shuf -n {{3}} -i {{1-10}}`
+`shuf -n {{3}} -i {{1-10}} -r`


pages/common/shuf.md
```diff
@@ -15,6 +15,6 @@
 
 `shuf {{filename}} -o {{output_filename}}`
 
-- Generate random numbers in range:
+- Generate 3 random numbers in the range 1-10:
```
bl-ue, line 18:
```suggestion
- Generate 3 random numbers in the range 1-10 (inclusive):
```

anakimluke:
Actually, this is not right. `-n x` does not guarantee `x` random numbers.

For example, `shuf -n 4 -i 1-3`.

anakimluke:
Well, looks like I misunderstood how the command works, and in doing so I discovered a problem with the original description: It is misleading because the sequence generated is of unique numbers. Adding the `-r` option seems to fix it.

CleanMachine1:
Remember to sign the license agreement otherwise regardless of all contributions, we can not use your commits

CleanMachine1:
Ill review in 12 hours or after more changes

anakimluke:
I'm curious. To accept this PR one would use github's "squash and merge" option? Then you's choose the commit title and description but in the log it'd appear as if I were the author of the commit?

bl-ue:
Exactly. 👍🏻  I don't think with squashing that you can change the commit author — GH makes it the PR author.  Sometimes, if the PR is large and the commits are nicely formatted and logical (e.g. applying a similar change to a selection of pages, one commit per page, or translated a selection of pages, one commit per page) we'd do a rebase, which puts all of the commits into `main`, but that's more rare.

We also get rid of the extended description before merging, except for that which comes from the PR's initial commit itself (which 22cd73440851e26718d69705bad11f88c59d08aa does have one), because it's usually a messy-looking list of commits, mostly `Update pages/.../...` or `Apply suggestions from code review` :)

We _also_ usually get rid of the `Co-authored-by:` lines, because they are automatically added when someone's suggestion gets applied, and they clutteringly show up on the final merged commit (see e64ec920e7752bd5d8141a5cfc91834dc455a555 for an example, which I still shed a tear over 😢), and there's _scarcely_ a non-trivial PR (like this one) that doesn't get one or more suggestions 😅

bl-ue (commented):
Looks good to me!  I'm a bit bothered by the fact that this page uses short options, but I do think it would be better to save that for another PR. Thank you @anakimluke!

bl-ue (approved):

bl-ue:
@anakimluke would you be interested in opening a PR to expand the options?

CleanMachine1:
Why not expand them now, just give suggestions and apply them without using them

bl-ue:
Because, I don't like hijacking PRs (especially a new contributor's first PR) to fix things that are mostly unrelated.

For one, it changes the PR's original intention `shuf: improve usage example and its description` to `shuf: expand short options, improve usage example and its descriptions`.

For two, it will likely make the contributor feel less in power over what changes he gets to make, because he might randomly have to reinvent the page.

For three, it probably will make the contributor more intimidated to contribute further, because he knows that he'll have to go over the same arduous process of revamping the page to fix others' past errors, making the process of contributing a confusing and heavy mess, rather than a fun little break as it ought to be.

For four, it makes the PR more confusing for everyone, contributors and maintainers.  Each PR should encapsulate one logical change.  This PR should be for adding the example, and another PR should deal with the flag expansion.

Those are my thoughts.

anakimluke:
Sure I'll create the PR ;) I don't see in the guidelines a mention of expanding the options. Should it be there? Or it's better to let people figure it out on a per-case basis?

bl-ue:
Seems like you're right, although I'm surprised.   We ought to add it then.  The goal is to have users _understand_ the commands rather than _memorize_ them.  Don't open a PR modifying this page yet until this PR is merged, though.

CleanMachine1 (approved):

PR #5995, opened 2021-05-19
rg: improved fixed string suggestion

DarkArc:
It's better to include a `--` prior to the search string so that strings starting with `--` (e.g. `--foo`) aren't processed as cli options.

https://github.com/BurntSushi/ripgrep/discussions/1560#discussioncomment-6346

### DIFF
--- a/pages/common/rg.md
+++ b/pages/common/rg.md
@@ -34,4 +34,4 @@
 
 - Search a literal string pattern:
 
-`rg --fixed-strings {{string}}`
+`rg --fixed-strings -- {{string}}`


CleanMachine1:
Should it have a space between it?
How does it work

1. --foo
2. -- foo

DarkArc:
Yes, it should have spaces (see the linked issue in the rg repo as well :) ). 

The `--` indicates all subsequent arguments are positional, which ensures `--foo`, `'--foo', etc aren't treated as cli options.

CleanMachine1:
Can't test however, I approve

CleanMachine1 (approved):

bl-ue (approved):
Good catch @DarkArc! I can test, and it works.  Many commands work this way.

CleanMachine1:
Thank you for your contributions, I hope you plan to add and/or fix more.
Help is always wanted and reviews are also wanted. 
If you would like to help you can help and do some of the issues such as #5137 
If not, thanks for your commits regardless

PR #6000, opened 2021-05-20
idnits: add page

bl-ue:


pages/common/idnits.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Show extra information about offending lines:
+
+`idnits --verbose {{path/to/file.txt}}`
+
+- Expect the specified year in the boilerplate instead of the current year:
+
+`idnits --year {{2021}} {{path/to/file.txt}}`
+
+- Assume the document is of the specified status:
```
CleanMachine1, line 23:
```suggestion
- Assume the document is of the specific status:
```
I still don't like this regardless, it doesn't read right

pages/common/idnits.md
```diff
@@ -0,0 +1,25 @@
...
+
+> Check internet-drafts for submission nits.
+> Looks for violations of Section 2.1 and 2.2 of the requirements listed on <https://www.ietf.org/id-info/checklist>.
+> More information: <https://tools.ietf.org/tools/idnits/>.
+
+- Check a file for nits:
+
+`idnits {{path/to/file.txt}}`
+
+- Count the nits without displaying them:
```
marchersimon, line 11:
```suggestion
- Count nits without displaying them:
```
I think that sounds a bit better here.

pages/common/idnits.md
```diff
@@ -0,0 +1,25 @@
+# idnits
+
+> Check internet-drafts for submission nits.
+> Looks for violations of Section 2.1 and 2.2 of the requirements listed on <https://www.ietf.org/id-info/checklist>.
+> More information: <https://tools.ietf.org/tools/idnits/>.
+
+- Check a file for nits:
+
+`idnits {{path/to/file.txt}}`
```
marchersimon, line 9: I don't know what a `nit` is, but can they also occur in other file formats?

pages/common/idnits.md
bl-ue: Ah, that's a lot nicer! I was having trouble with that line, but you really unlocked it 😄

pages/common/idnits.md
bl-ue: Looking at [(a fork of) the source](https://github.com/fno2010/idnits/blob/d38815d5330d80571ba030a6033f35686375c482/idnits), it seems like it prefers `.txt` files, although I believe it will take `.html` files, and _possibly_ `.(txt|pdf|xml|doc|nr)` 🤔

bl-ue: Yeah, it's uncommon for whatever reason.  But the IETF definitely still uses txts mostly, from the 60s 😆

marchersimon (approved):
Looks good, thanks

pages/common/idnits.md
```diff
@@ -0,0 +1,25 @@
...
+
+`idnits --nitcount {{path/to/file.txt}}`
+
+- Show extra information about offending lines:
+
+`idnits --verbose {{path/to/file.txt}}`
+
+- Expect the specified year in the boilerplate instead of the current year:
+
+`idnits --year {{2021}} {{path/to/file.txt}}`
```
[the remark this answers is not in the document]
bl-ue: What about 2k?
```suggestion
`idnits --year {{2000}} {{path/to/file.txt}}`
```

pages/common/idnits.md
bl-ue: The oldest rfc, [rfc1](https://datatracker.ietf.org/doc/html/rfc1), is from 1969 👀

>7 April 1969

marchersimon: I'm always surprised by how long computers actually exist 😅

pages/common/idnits.md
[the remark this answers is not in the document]
marchersimon: I don't think it really matters, but `2021` gives the page a modern touch.

pages/common/idnits.md
bl-ue: Yeah. Off the top of my head, the oldest version of Fortran is from '55 or '58, I think...

pages/common/idnits.md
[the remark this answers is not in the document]
bl-ue: True.  I'll leave it.

pages/common/idnits.md
marchersimon: Wow, that's old.

CleanMachine1 (approved):

pages/common/idnits.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Assume the document is of the specified status:
```
sbrl, line 23: Typo?
```suggestion
- Assume the document is of the specified type:
```

bl-ue: No, not exactly.  I think "type" is also correct here, but they're also statuses of the documents.  The full statuses are—
- O → Obsolete
- U → Unknown state
- E → Experimental
- I → Informational
- B → Best Current Practice
- P → Proposed Standard
- D → Draft Standard
- S → Full Standard
- H → Historic
- N → Not Issued

sbrl: Ah, I see. The example there though only has a single named argument `--doctype` though? I'm confused.

bl-ue: Yes, `--doctype` specifies the status of the document.  The flag itself says `type`, but _everywhere_ else it's referred to as "status".

`idnits file.txt --doctype bcp` would check `file.txt` for nits, assuming `file.txt` is a "Best Current Practice" rfc draft.

sbrl: Oh, that's really weird! Let's keep this as-is though, but I'm explicitly not resolving this conversation for future reference.

sbrl (approved):

PR #6004, opened 2021-05-20
mv: add move multiple source paths to a directory example

jack1142:
I think this is an important feature of `mv` and it's not obvious from the examples that already existed on this page.

### DIFF
--- a/pages/common/mv.md
+++ b/pages/common/mv.md
@@ -3,7 +3,7 @@
 > Move or rename files and directories.
 > More information: <https://www.gnu.org/software/coreutils/mv>.
 
-- Move file in arbitrary location:
+- Move a file to an arbitrary location:
 
 `mv {{source}} {{target}}`
 
--- a/pages/common/mv.md
+++ b/pages/common/mv.md
@@ -9,7 +9,7 @@
 
 - Move files into another directory, keeping the filenames:
 
-`mv {{source1}} {{source2}} {{source3}} {{directory}}`
+`mv {{source1}} {{source2}} {{source3}} {{target_directory}}`
 
 - Do not prompt for confirmation before overwriting existing files:
 
--- a/pages/common/mv.md
+++ b/pages/common/mv.md
@@ -3,9 +3,13 @@
 > Move or rename files and directories.
 > More information: <https://www.gnu.org/software/coreutils/mv>.
 
-- Move files in arbitrary locations:
+- Move file in arbitrary location:
 
 `mv {{source}} {{target}}`
+
+- Move files into another directory, keeping the filenames:
+
+`mv {{source1}} {{source2}} {{source3}} {{directory}}`
 
 - Do not prompt for confirmation before overwriting existing files:
 


CleanMachine1:
Workflow approved, that will iron out the bot bugs.

Thanks for your PR @jack1142

CleanMachine1:
This will need some review, however I am not available for suggestions however your contributions will be merged between 12 hours and a week I would say.

pages/common/mv.md
```diff
@@ -3,10 +3,14 @@
...
 > More information: <https://www.gnu.org/software/coreutils/mv>.
 
-- Move files in arbitrary locations:
+- Move file in arbitrary location:
 
 `mv {{source}} {{target}}`
 
+- Move files into another directory, keeping the filenames:
+
+`mv {{source1}} {{source2}} {{source3}} {{directory}}`
```
CleanMachine1, line 12:
```suggestion
`mv {{/path/to/source1}} {{/path/to/source2}} {{/path/to/source3}} {{/path/to/target}}`
```

CleanMachine1: Dont commit this until we have had others opinions

jack1142: Wouldn't it be better if it were `/path/to/directory` rather than `/path/to/target` considering that it is required to be a directory when multiple files are passed?
image

CleanMachine1: target_directory?

jack1142: Seems fine too, I'll wait to see what others think.

bl-ue: I like `/path/to/target_directory` the best too 👍🏻

marchersimon: Why the absolute paths? It also worrks with relative.

pages/common/mv.md
```diff
@@ -3,10 +3,14 @@
 > Move or rename files and directories.
 > More information: <https://www.gnu.org/software/coreutils/mv>.
 
-- Move files in arbitrary locations:
+- Move file in arbitrary location:
```
marchersimon, line 6:
```suggestion
- Move a file to an arbitrary location:
```

pages/common/mv.md
navarroaxel: yeah, works with relatives.

mfrw: I think `target_directory` would be more clear than just `target`

marchersimon: I think the only change we should make is this:
```suggestion
`mv {{source1}} {{source2}} {{source3}} {{target_directory}}`
```
since the whole page doesn't use `path/to/`. We can of course fix this in a seperate PR, but as @bl-ue pointed out [here](https://github.com/tldr-pages/tldr/pull/5992#issuecomment-844156753), we should not use PRs to fix things unrelated to the original changes.

CleanMachine1: Feel free to commit Simons change!

CleanMachine1 (approved):
Thanks for your first PR!

marchersimon (approved):
Thanks @jack1142

navarroaxel (approved):

PR #6007, opened 2021-05-21
josm: add page

marchersimon:


pages/common/josm.md
```diff
@@ -0,0 +1,32 @@
+# josm
+
+> Extensible OpenStreetMap editor for Java 8+.
+> More information: <https://josm.openstreetmap.de/>.
+
+- Launch JOSM:
```
CleanMachine1, line 6: Should this be `start` or `launch`?

pages/common/josm.md
```diff
@@ -0,0 +1,32 @@
...
+
+`josm --language {{de}}`
+
+- Launch JOSM and reset all preferences to default values:
+
+`josm --reset-preferences`
+
+- Launch JOSM and download a specific bounding box:
+
+`josm --download {{minlat,minlon,maxlat,maxlon}}`
```
CleanMachine1, line 24:
```suggestion
`josm --download {{minlat|minlon|maxlat|maxlon}}`
```

Unaware if this is right

pages/common/josm.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Launch JOSM and download a specefic bounding box as raw GPS:
+
+`josm --downloadgps {{minlat,minlon,maxlat,maxlon}}`
```
CleanMachine1, line 28:
```suggestion
`josm --downloadgps {{minlat|minlon|maxlat|maxlon}}`
```
Same applies

pages/common/josm.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Launch JOSM in maximized mode:
+
+`josm --maximize`
+
+- Launch JOSM and set a specific language:
+
+`josm --language {{de}}`
+
+- Launch JOSM and reset all preferences to default values:
```
CleanMachine1, line 18:
```suggestion
- Launch JOSM and reset all preferences to default:
```

I don't like that usage of values however we'll see what others think

pages/common/josm.md
```diff
@@ -0,0 +1,32 @@
+# josm
+
+> Extensible OpenStreetMap editor for Java 8+.
+> More information: <https://josm.openstreetmap.de/>.
+
+- Launch JOSM:
+
+`josm`
+
+- Launch JOSM in maximized mode:
```
CleanMachine1, line 10:
```suggestion
- Launch JOSM in full-screen mode:
```
Does it just give a full window or is it an F11 type of full
If its not an F11, resolve this although I don't think its right using the same words between command and description.

Its like saying "Describe an egg","An egg shape"

pages/common/josm.md
marchersimon: No, you actually need all four of them, seperated by a comma.

pages/common/josm.md
marchersimon: I don't see a difference.

CleanMachine1: Non-native english speakers are more likely to know the word start rather than launch probably

pages/common/josm.md
marchersimon: I don't think `maximized` is the same as `fullscreen`.

CleanMachine1: When you run it in with --max... is it like pressing F11?

pages/common/josm.md
marchersimon: I'm not so sure about that. Any other opinions?

pages/common/josm.md
marchersimon: I isn't. When using `--maximize` it is like any other regular window with panel and window decoration still visible. If you don't use it, JOSM launches even smaller, so you can drag it around on the screen.

CleanMachine1: Then this is resolved

pages/common/josm.md
bl-ue: We use "launch" in a variety of places, and I'm not certain about intentionally simplifying and restricting our vocabulary to a subset of words just because some people might not know them as well.  There are many complex sentence structures in English—being a rather _unlogical_ language—which are likely to elude those who aren't profficient in it; however, there's no need to use them because that would be _useless_ complexity.

w.r.t. to a choice of words though, I don't think that using more _elaborate_ wording is a problem, because any novice could look it up and learn that _launch_ is a **verb**, and **opens an application**.

Makes sense?

pages/common/josm.md
bl-ue: How about
```suggestion
- Launch JOSM and reset all preferences to their default values:
```

pages/common/josm.md
SethFalco: Just dropping my opinion!

Personally, I'm not against the idea of keeping the English versions accessible to non-native speakers.
The [curb cut effect](https://en.wikipedia.org/wiki/Curb_cut_effect) applies here by being more friendly for:
* Dyslexic users
* Younger demographics
* The tech industry in general, which is dominated by the English language no matter a countries national language

I wouldn't downgrade a sentence for accommodating non-native speakers, though. Launch is a better word (subjective) and basic enough that if someone doesn't know it, they should really learn it.

As a native English speaker, I'd prefer "launch" rather than "start", but in the end it also doesn't make too much difference.

CleanMachine1: I agree, launch definitely is the better word compared to start. Just a suggestion, been discussed and you can resolve if you please

bl-ue: Thank you for dropping in @SethFalco! :)

pages/common/josm.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Launch JOSM and reset all preferences to their default values:
+
+`josm --reset-preferences`
+
+- Launch JOSM and download a specific bounding box:
+
+`josm --download {{minlat,minlon,maxlat,maxlon}}`
+
+- Launch JOSM and download a specefic bounding box as raw GPS:
```
bl-ue, line 26:
```suggestion
- Launch JOSM and download a specific bounding box as raw GPS:
```

bl-ue (changes requested):

bl-ue (approved):

navarroaxel (approved):

CleanMachine1 (approved):

SethFalco (approved):
Looks good to me! 👍🏽

PR #5988, opened 2021-05-18
Add contributing-guides/alias-pages.md

marchersimon:
Ref: #5986 
I hope the page structure is fine.

PR #6011, opened 2021-05-21
git-delete-branch: add page

CleanMachine1:

For #5137

### DIFF
--- a/pages/common/git-delete-branch.md
+++ b/pages/common/git-delete-branch.md
@@ -1,0 +1,13 @@
+# git delete-branch
+
+> Delete local and remote Git branches.
+> Part of `git-extras`. You will get errors if you are checked out on a branch you are deleting.
+> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-delete-branch>.
+
+- Delete a local and remote Git branch:
+
+`git delete-branch {{branch_name}}`
+
+- Delete multiple local and remote Git branches:
+
+`git delete-branch {{branch_name branch_name ...}}`
--- a/pages/common/git-delete-branch.md
+++ b/pages/common/git-delete-branch.md
@@ -10,4 +10,4 @@
 
 - Delete multiple local and remote Git branches:
 
-`git delete-branch {{branch_name branch_name ...}}`
+`git delete-branch {{branch_name1 branch_name2 ...}}`
--- a/pages/common/git-delete-branch.md
+++ b/pages/common/git-delete-branch.md
@@ -1,7 +1,7 @@
 # git delete-branch
 
 > Delete local and remote Git branches.
-> Part of `git-extras`. When deleting a checked out branch, only the remote branch will be deleted.
+> Part of `git-extras`. If deleting the checked out branch, only the remote branch will be deleted.
 > More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-delete-branch>.
 
 - Delete a local and remote Git branch:
--- a/pages/common/git-delete-branch.md
+++ b/pages/common/git-delete-branch.md
@@ -1,7 +1,7 @@
 # git delete-branch
 
 > Delete local and remote Git branches.
-> Part of `git-extras`. You will get errors if you are checked out on a branch you are deleting.
+> Part of `git-extras`. When deleting a checked out branch, only the remote branch will be deleted.
 > More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-delete-branch>.
 
 - Delete a local and remote Git branch:
--- a/pages/common/git-graft.md
+++ b/pages/common/git-graft.md
@@ -1,9 +1,0 @@
-# git graft
-
-> Merge commits from a specific branch into another and delete the source branch.
-> Part of `git-extras`
-> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-graft>.
-
-- Merge all commits not present on destination branch from source branch, and delete source branch:
-
-`git graft {{source_branch}} {{destination_branch}}
--- a/pages/common/git-graft.md
+++ b/pages/common/git-graft.md
@@ -1,0 +1,9 @@
+# git graft
+
+> Merge commits from a specific branch into another and delete the source branch.
+> Part of `git-extras`
+> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-graft>.
+
+- Merge all commits not present on destination branch from source branch, and delete source branch:
+
+`git graft {{source_branch}} {{destination_branch}}


pages/common/git-delete-branch.md
```diff
@@ -0,0 +1,13 @@
+# git delete-branch
+
+> Delete local and remote Git branches.
+> Part of `git-extras`. You will get errors if you are checked out on a branch you are deleting.
```
mfrw, line 4:
```suggestion
> Part of `git-extras`. An error occurs on deleting the current checked out branch.
```

mfrw (changes requested):

pages/common/git-delete-branch.md
CleanMachine1:
```suggestion
> Part of `git-extras`. Errors will occur if you are deleting a checked out branch
```

bl-ue: I agree with the removal of "you", but it seems wrong to put on the same line as "Part of `git-extras`".  I've venture to say that this is note is unnecessary, because it's not likely that a user will think that they _can_ delete the current branch.  Thoughts?

CleanMachine1: A user might make a mistake and not realise they are checked out on a particular branch

marchersimon: Well actually as I've just encountered with #5988, it does *not* give you an error if you try it...

CleanMachine1: error: Cannot delete branch 'testremove' checked out

I have a test repo just for me and while checkedout, I ran this command and got this

> error: Cannot delete branch 'testremove' checked out

CleanMachine1: it does however delete the remote branch since you are not _checked out_ at origin/testremove

marchersimon: Just saw that I got the message too. However it deletes all remote branches. I'd prefer not to include the note above, since it may let the user think it's save to run.

pages/common/git-graft.md
```diff
@@ -0,0 +1,9 @@
+# git graft
+
+> Merge commits from a specific branch into another and delete the source branch.
```
bl-ue:
```suggestion
> Merge commits from a branch into another and delete the source branch.
```

pages/common/git-graft.md
```diff
@@ -0,0 +1,9 @@
...
+> Part of `git-extras`
+> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-graft>.
+
+- Merge all commits not present on destination branch from source branch, and delete source branch:
```
bl-ue:
```suggestion
- Merge all commits not present on the target branch from the source branch, and delete the source branch:
```

pages/common/git-graft.md
```diff
@@ -0,0 +1,9 @@
...
+
+`git graft {{source_branch}} {{destination_branch}}
```
bl-ue:
```suggestion
`git graft {{source_branch}} {{target_branch}}
```

bl-ue (changes requested):
"Target" is more correct, rather than "destination."

bl-ue:
Whoops, that was all incorrect, sorry... 😆

pages/common/git-delete-branch.md
bl-ue: So should we resolve this thread?

bl-ue (approved):

pages/common/git-delete-branch.md
marchersimon: I actually think we should remove the second sentense completely. We also don't include in every Git page, that errors will occur if you're not in a Git repository.

CleanMachine1:
```suggestion
> Part of `git-extras`.
```

CleanMachine1: But it is an easy error to make, many will make it

marchersimon: How about:
```suggestion
> Part of `git-extras`. When deleting the checked out branch, only the remote branch will be deleted.
```

CleanMachine1:
```suggestion
> Part of `git-extras`. When deleting a checked out branch, only the remote branch will be deleted.
```

marchersimon: Can you have multiple checked out branches?

CleanMachine1: Not to my knowledge, however the `the` seems out of place, a fits better and makes it read better

pages/common/git-delete-branch.md
```diff
@@ -0,0 +1,13 @@
...
+> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-delete-branch>.
+
+- Delete a local and remote Git branch:
+
+`git delete-branch {{branch_name}}`
+
+- Delete multiple local and remote Git branches:
+
+`git delete-branch {{branch_name branch_name ...}}`
```
navarroaxel, line 13:
```suggestion
`git delete-branch {{branch_name1 branch_name2 ...}}`
```

pages/common/git-delete-branch.md
```diff
@@ -0,0 +1,13 @@
+# git delete-branch
+
+> Delete local and remote Git branches.
+> Part of `git-extras`. When deleting a checked out branch, only the remote branch will be deleted.
```
bl-ue, line 4: There can only be one checked out branch, so I'd adjust the sentence like so:
```suggestion
> Part of `git-extras`. If deleting the checked out branch, only the remote branch will be deleted.
```

bl-ue (changes requested):

pages/common/git-delete-branch.md
CleanMachine1: I reallly dislike the usage of the `the`, it makes more nouny rather than the verb  status it deserves. Its start to explain :smile:

bl-ue: I'm not really sure what you mean; `a` and `the` are both articles and the only difference between them is the number of the nouns that they're used with.  Using "the" is nothing but more correct here.

CleanMachine1: https://github.com/tldr-pages/tldr/pull/6011#discussion_r637851077 @marchersimon seemed to agree with me. Need a more solid opinion rather than +1

marchersimon: I also think `the` makes more sense here, since there's only *the* one checked out branch. However `a` is fine for me too and it's still your PR.

CleanMachine1: Committedd

pages/common/git-delete-branch.md
sbrl: No, you can't have multiple checked out branches under normal operation to my knowledge.

@bl-ue: On the contrary, I think it's important to keep the `Part of git-extras.`, just in case someone confuses it for an inbuilt git command.

sbrl (approved):
Thanks, @CleanMachine1!

pages/common/git-delete-branch.md
bl-ue: I meant that the note about deleting the checked out branch isn't necessary, @sbrl.

CleanMachine1: Can we reach a result, what should suggestion should  I apply

navarroaxel (approved):

bl-ue (approved):

PR #6019, opened 2021-05-23
make: add Chinese translation

BillLucky:

### DIFF
--- a/pages.zh/common/make.md
+++ b/pages.zh/common/make.md
@@ -4,7 +4,7 @@
 > 通常用于控制源代码中可执行文件的编译。
 > 更多信息见：<https://www.gnu.org/software/make/manual/make.html>.
 
-- 调用 Makefile 中指定的第一个目标（通常命名为 “all”）：
+- 调用 Makefile 中指定的第一个目标（通常命名为 "all"）：
 
 `make`
 
--- a/pages.zh/common/make.md
+++ b/pages.zh/common/make.md
@@ -10,24 +10,24 @@
 
 - 调用指定目标：
 
-`make {{target}}`
+`make {{目标}}`
 
 - 调用一个指定的目标，一次并行执行4个作业：
 
-`make -j{{4}} {{target}}`
+`make -j{{4}} {{目标}}`
 
 - 使用指定的 Makefile 文件：
 
-`make --file {{file}}`
+`make --file {{文件}}`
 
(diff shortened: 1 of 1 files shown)


pages.zh/common/make.md
```diff
@@ -0,0 +1,33 @@
+# make
+
+> Makefile 文件描述目标的任务运行器。
+> 通常用于控制源代码中可执行文件的编译。
+> 更多信息见：<https://www.gnu.org/software/make/manual/make.html>.
+
+- 调用 Makefile 中指定的第一个目标（通常命名为 “all”）：
```
bl-ue, line 7: Is there a need to use this non-ASCII style of quotes @BillLucky?
```suggestion
- 调用 Makefile 中指定的第一个目标（通常命名为 "all"）：
```

bl-ue (approved):
Looks good to me, except for one comment.

pages.zh/common/make.md
BillLucky: I think you are right, it is more natural to use ASCII symbols :) @bl-ue

marchersimon (approved):

pages.zh/common/make.md
```diff
@@ -0,0 +1,33 @@
...
+
+`make {{target}}`
+
+- 调用一个指定的目标，一次并行执行4个作业：
+
+`make -j{{4}} {{target}}`
+
+- 使用指定的 Makefile ：
+
+`make --file {{file}}`
```
navarroaxel, line 21: You can translate the argument option to help non-English speakers.

BillLucky: OK.

Because there is only one parameter here, "Use the specified file" is already clear.

At the same time, we can see in the [English version](https://github.com/tldr-pages/tldr/blob/main/pages/common/make.md), and there is no detailed explanation of the -file parameter.

I can refer to the man manual for a detailed explanation, but that might seem redundant.

```
from `man make`
make -f file

  +-f file, --file=file, --makefile=FILE
             Use file as a makefile.
```

@navarroaxel

navarroaxel: I think you could just translate `file` to Chinese.

BillLucky: Okay

navarroaxel (approved):

bl-ue (approved):

PR #6035, opened 2021-05-23
history: add `!n` and `!?string?` commands

jopefd:

### DIFF
--- a/pages/common/history.md
+++ b/pages/common/history.md
@@ -7,11 +7,11 @@
 
 `history`
 
-- Run the n-th command of the history list
+- Run the n-th command of the history list:
 
 `!{{n}}`
 
-- Run the most recent command of the history list containing the specified string
+- Run the most recent command of the history list containing the specified string:
 
 `!?{{string}}?`
 
--- a/pages/common/history.md
+++ b/pages/common/history.md
@@ -7,9 +7,9 @@
 
 `history`
 
-- Run the n-th command of the history list:
+- Run the Nth command of the history list:
 
-`!{{n}}`
+`!{{N}}`
 
 - Run the most recent command of the history list containing the specified string:
 
--- a/pages/common/history.md
+++ b/pages/common/history.md
@@ -7,6 +7,10 @@
 
 `history`
 
+- Run the n-th command of the history list
+
+`!{{n}}`
+
 - Display the last 20 commands:
 
 `history {{20}}`
--- a/pages/common/history.md
+++ b/pages/common/history.md
@@ -11,6 +11,10 @@
 
 `!{{n}}`
 
+- Run the most recent command of the history list containing the specified string
+
+`!?{{string}}?`
+
 - Display the last 20 commands:
 
 `history {{20}}`


pages/common/history.md
```diff
@@ -7,6 +7,14 @@
 
 `history`
 
+- Run the n-th command of the history list:
```
bl-ue, line 10:
```suggestion
- Run the Nth command of the history list:
```

bl-ue (approved):
Cool! Looks good to me, except I'd put `n-th` differently...

pages/common/history.md
jopefd: Ok! I gonna fix it :)

bl-ue (approved):
Indeed, these work for me.  Interesting!

jopefd:
> Indeed, these work for me. Interesting!

Good! =D

pages/common/history.md
marchersimon: Actually, since all our clients support UTF-8, why not "Nᵗʰ"?

marchersimon (approved):
Thanks @jopefd

CleanMachine1:
Just a heads up, open a branch on your fork with the relevant name, instead of pushing to main branch, just in case you ever become a collaborator and good practise to not push straight to main

CleanMachine1 (approved):

jopefd:
> Just a heads up, open a branch on your fork with the relevant name, instead of pushing to main branch, just in case you ever become a collaborator and good practise to not push straight to main

Ah, understood! Next time I'll do this. Thank you!

pages/common/history.md
jopefd: For me Nᵗʰ is ok too. Should I change to it? @bl-ue @CleanMachine1

marchersimon: I'm not sure anymore. In monospace, which terminals use, it looks like this: `Nᵗʰ`

CleanMachine1: hmmm I like the smaller versions however as mentioned, special characters like that often look awful under most fonts.

navarroaxel (changes requested):
I think `histexpand` is the place for these new examples. @sbrl?

marchersimon:
We should've mentioned the `histexpand` page in `history.md`. `histexpand` already has 8 examples, but there we have `!-n`, which runs the command n lines back (which is in my opinion more useful than `!n`) and 

> \- Run the most recent command starting with string:
    !string

which we could replace with `!?string?`, since it's more universal.

jopefd:
> We should've mentioned the `histexpand` page in `history.md`. `histexpand` already has 8 examples, but there we have `!-n`, which runs the command n lines back (which is in my opinion more useful than `!n`) and
> 
> > - Run the most recent command starting with string:
> > !string
> 
> which we could replace with `!?string?`, since it's more universal.

1. I think it's better to offer options, and let the user choose which is better for its case. In my case I don't use `!-n`, so I'd choose `!n`.
2. Yes, `!?string?` covers `!string`, so it seems me better.

_Note: `!?string?` vs. `!string` and `!-n` vs. `!n` are two distinct cases to me. While `!?string?` covers `!string`, neither `!-n` nor `!n` covers the other. `!n` works for fixed command ids and `!-n` works for relative command ids. That's why I prefer that the user can choose._

marchersimon:
We have to decide for either `!n` or `!-n`, since pages can't have more than 8 examples. Personally I don't use either of them, so I can't judge which would be more helpful. Regarding `!?string?` vs `!string` I'm also not really sure. The first covers the second, but I can imagine that in most cases you want to get the command which starts with a specific string (`!string`).

jopefd:
Alright...

`!?string?` still seems me to be the best option, because, for example, it allows me to search `!?sudo ln -s?`. It wouldn't be uselful search `!sudo`; it'll give me any command runned as sudo...

About `!-n!` to me it's not useful. I don't have any need to run this, and I think _"why does someone need to calculate some command id for execute it? It's just execute `history`, see the id, and execute `!id`"... no calculations_. But I don't know if there is some relevant use for `!-n`...

What do you think? @navarroaxel @CleanMachine1 @bl-ue

CleanMachine1:
I don't really know. However I am watching this PR.

_just  leaving my little notice that im not continuing this on my end_

sbrl:
@navarroaxel: Yeah, I would probably say that `histexpand` is the correct place for these examples. `history` should definitely have a `See also` though pointing to `histexpand`.

navarroaxel:
@jopefd don't close this PR. Just revert `history` and move the changes to `histexpand`.

marchersimon:
I'll close this now, since all changes proposed here are made in #6051

PR #6047, opened 2021-05-25
dd: add backup commands

CleanMachine1:

### DIFF
--- a/pages/common/dd.md
+++ b/pages/common/dd.md
@@ -23,7 +23,7 @@
 
 `dd if=/dev/{{drive_device}} of={{/path/to/file.img}} status=progress`
 
-- Restore a drive from an img file and show the progress:
+- Restore a drive from an IMG file and show the progress:
 
 `dd if={{/path/to/file.img}} of=/dev/{{drive_device}} status=progress`
 
--- a/pages/common/dd.md
+++ b/pages/common/dd.md
@@ -19,7 +19,7 @@
 
 `dd if=/dev/zero of={{file_1GB}} bs=1024 count=1000000`
 
-- Generate a system backup into a img file and show the progress:
+- Generate a system backup into a IMG file and show the progress:
 
 `dd if=/dev/{{drive_device}} of={{/path/to/file.img}} status=progress`
 
--- a/pages/common/dd.md
+++ b/pages/common/dd.md
@@ -19,7 +19,7 @@
 
 `dd if=/dev/zero of={{file_1GB}} bs=1024 count=1000000`
 
-- Generate a system backup into a IMG file and show the progress:
+- Generate a system backup into an IMG file and show the progress:
 
 `dd if=/dev/{{drive_device}} of={{/path/to/file.img}} status=progress`
 
--- a/pages/common/dd.md
+++ b/pages/common/dd.md
@@ -23,7 +23,7 @@
 
 `dd if=/dev/{{drive_device}} of={{/path/to/file.img}} status=progress`
 
-- Flash an img file to a drive and show the progress:
+- Restore a drive from an img file and show the progress:
 
 `dd if={{/path/to/file.img}} of=/dev/{{drive_device}} status=progress`
 
--- a/pages/common/dd.md
+++ b/pages/common/dd.md
@@ -19,6 +19,14 @@
 
 `dd if=/dev/zero of={{file_1GB}} bs=1024 count=1000000`
 
+- Generate a system backup into a img file and show the progress:
+
+`dd if=/dev/{{drive_device}} of={{/path/to/file.img}} status=progress`
+
+- Flash an img file to a drive and show the progress:
+
+`dd if={{/path/to/file.img}} of=/dev/{{drive_device}} status=progress`
+
 - Check progress of an ongoing dd operation (Run this command from another shell):
 
 `kill -USR1 $(pgrep ^dd)`
--- a/pages/common/dd.md
+++ b/pages/common/dd.md
@@ -21,11 +21,11 @@
 
 - Generate a system backup into an IMG file and show the progress:
 
-`dd if=/dev/{{drive_device}} of={{/path/to/file.img}} status=progress`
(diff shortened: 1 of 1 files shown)


pages/common/dd.md
```diff
@@ -19,6 +19,14 @@
 
 `dd if=/dev/zero of={{file_1GB}} bs=1024 count=1000000`
 
+- Generate a system backup into a img file and show the progress:
```
CleanMachine1, line 22:
```suggestion
- Create a system backup into a img file and show the progress:
```
thoughts?

mfrw: `Create` reads better than `Generate` IMO

pages/common/dd.md
```diff
@@ -19,6 +19,14 @@
...
+
+`dd if=/dev/{{drive_device}} of={{/path/to/file.img}} status=progress`
+
+- Flash an img file to a drive and show the progress:
```
mfrw, line 26: should we be consisted with the wording of the above command. Thoughts ?
```suggestion
- Restore a drive from an img file and show the progress:
```

CleanMachine1: But that implies that the drives is being done again

pages/common/dd.md
navarroaxel: Is this an `img` file? I can restore any .img file into a partition?

pages/common/dd.md
navarroaxel: I think `restore` is better.

pages/common/dd.md
CleanMachine1:
```suggestion
- Generate a system backup in the form of an img file and show the progress:
```
```suggestion
- Create a system backup in the form of an img file and show the progress:
```

marchersimon:
I think we should either use `IMG` or `` `.img` ``.

pages/common/dd.md
marchersimon, line 22:
```suggestion
- Generate a system backup into a IMG file and show the progress:
```
`` `img` `` would also be fine.

pages/common/dd.md
```diff
@@ -19,6 +19,14 @@
 
 `dd if=/dev/zero of={{file_1GB}} bs=1024 count=1000000`
 
+- Generate a system backup into a img file and show the progress:
+
+`dd if=/dev/{{drive_device}} of={{/path/to/file.img}} status=progress`
+
+- Restore a drive from an img file and show the progress:
```
marchersimon, line 26:
```suggestion
- Restore a drive from an IMG file and show the progress:
```

marchersimon: @CleanMachine1 what about this one?

mfrw: I think the `IMG` looks more generic than `img`.

mfrw (approved):
LGTM

pages/common/dd.md
```diff
@@ -19,6 +19,14 @@
 
 `dd if=/dev/zero of={{file_1GB}} bs=1024 count=1000000`
 
+- Generate a system backup into a IMG file and show the progress:
+
+`dd if=/dev/{{drive_device}} of={{/path/to/file.img}} status=progress`
+
+- Restore a drive from an IMG file and show the progress:
+
+`dd if={{/path/to/file.img}} of=/dev/{{drive_device}} status=progress`
```
marchersimon, line 28:
```suggestion
`dd if={{path/to/file.img}} of=/dev/{{drive_device}} status=progress`
```
As far as I can remember, this doesn't have to be absolute.

pages/common/dd.md
```diff
@@ -19,6 +19,14 @@
 
 `dd if=/dev/zero of={{file_1GB}} bs=1024 count=1000000`
 
+- Generate a system backup into a IMG file and show the progress:
+
+`dd if=/dev/{{drive_device}} of={{/path/to/file.img}} status=progress`
```
marchersimon, line 24:
```suggestion
`dd if=/dev/{{drive_device}} of={{path/to/file.img}} status=progress`
```
Same here.

pages/common/dd.md
```diff
@@ -19,6 +19,14 @@
 
 `dd if=/dev/zero of={{file_1GB}} bs=1024 count=1000000`
 
+- Generate a system backup into a IMG file and show the progress:
```
marchersimon, line 22:
```suggestion
- Generate a system backup into an IMG file and show the progress:
```
And a little typo.

pages/common/dd.md
CleanMachine1: It doesn't however I would never recommend not. 
This command as you probably know is dangerous. Always use absolute with dd

marchersimon: I can't really think of what could to wrong when using the wrong `img` file.

CleanMachine1: Writing the wrong file comes to mind

marchersimon: Would that be a problem? You could just rewrite with the right one.

bl-ue: Of course, but any command that writes to a file can destroy a file that you don't intend if you specify an incorrect path.

CleanMachine1: Yes however DD is probably one of the easiest commands to mess up on.

I've nearly wiped my entire system before from a mix-up of if and of. 

If we are going to keep it as relative, maybe we should add a small note to say this command is dangerous. 

And although this would apply for a lot of commands, dd is up there with rm /*

bl-ue: It's the users's responsibility to ensure that they don't wreck their system — tldr-pages simply provides examples on how to use the command.

CleanMachine1: Ok, well in that case, we'll use relative. Commits applied in 30 seconds.

marchersimon (approved):
Looks good

CleanMachine1:
11 days and we have received no other activity. Thats why I merged.

PR #6051, opened 2021-05-26
histexpand: add `!N` and `!?string?` commands

jopefd:
**See #6035**

### DIFF
--- a/pages/common/histexpand.md
+++ b/pages/common/histexpand.md
@@ -27,7 +27,7 @@
 
 `!-{{n}}`
 
-- Run the most recent command starting with `string`:
+- Run the most recent command containing `string`:
 
 `!?{{string}}?`
 
--- a/pages/common/histexpand.md
+++ b/pages/common/histexpand.md
@@ -3,7 +3,7 @@
 > Reuse and expand the shell history in `sh`, `bash`, `zsh`, `rbash` and `ksh`.
 > More information: <https://www.gnu.org/software/bash/manual/html_node/History-Interaction>.
 
-- Run the previous command as root (`!!` replaces the previous command):
+- Run the previous command as root (`!!` is replaced by the previous command):
 
 `sudo !!`
 
--- a/pages/common/histexpand.md
+++ b/pages/common/histexpand.md
@@ -19,6 +19,10 @@
 
 `{{command}} !^`
 
+- Run the Nth command of the history:
+
+`!{{N}}`
+
 - Run the command `n` lines back in the history:
 
 `!-{{n}}`
--- a/pages/common/histexpand.md
+++ b/pages/common/histexpand.md
@@ -29,7 +29,6 @@
 
 - Run the most recent command starting with `string`:
 
-`!{{string}}`
 `!?{{string}}?`
 
 - Run the previous command, replacing `string1` with `string2`:
--- a/pages/common/histexpand.md
+++ b/pages/common/histexpand.md
@@ -30,6 +30,7 @@
 - Run the most recent command starting with `string`:
 
 `!{{string}}`
+`!?{{string}}?`
 
 - Run the previous command, replacing `string1` with `string2`:
 
--- a/pages/common/histexpand.md
+++ b/pages/common/histexpand.md
@@ -3,7 +3,7 @@
 > Reuse and expand the shell history in `sh`, `bash`, `zsh`, `rbash` and `ksh`.
 > More information: <https://www.gnu.org/software/bash/manual/html_node/History-Interaction>.
 
-- Run the previous command as root:
+- Run the previous command as root (`!!` replaces the previous command):
 
 `sudo !!`
 
--- a/pages/common/histexpand.md
+++ b/pages/common/histexpand.md
@@ -17,7 +17,7 @@
 
 - Run the Nth command of the history:
 
-`!{{N}}`
+`!{{n}}`
 
 - Run the command `n` lines back in the history:
 
--- a/pages/common/histexpand.md
+++ b/pages/common/histexpand.md
@@ -3,10 +3,6 @@
 > Reuse and expand the shell history in `sh`, `bash`, `zsh`, `rbash` and `ksh`.
 > More information: <https://www.gnu.org/software/bash/manual/html_node/History-Interaction>.
 
-- Run the previous command:
-
-`!!`
-
 - Run the previous command as root:
 
 `sudo !!`


jopefd:
> The [build](https://github.com/tldr-pages/tldr/actions/runs/878466982) for this PR failed with the following error(s):
> 
> ```
> pages/common/histexpand.md:33: TLDR105 There should be only one command per example
> pages/common/histexpand.md:0: TLDR019 Page should only include a maximum of 8 examples
> ```
> 

@navarroaxel and now?

pages/common/histexpand.md
```diff
@@ -19,13 +19,18 @@
...
+`!{{N}}`
+
 - Run the command `n` lines back in the history:
 
 `!-{{n}}`
 
 - Run the most recent command starting with `string`:
 
 `!{{string}}`
+`!?{{string}}?`
```
bl-ue, line 33: One of these has to be removed.  Let's remove the old one
```suggestion
`!?{{string}}?`
```

bl-ue (changes requested):

pages/common/histexpand.md
```diff
@@ -19,13 +19,17 @@
...
 
+- Run the Nth command of the history:
+
+`!{{N}}`
+
 - Run the command `n` lines back in the history:
 
 `!-{{n}}`
 
 - Run the most recent command starting with `string`:
```
bl-ue, line 30: This line should probably update too.  How about
```suggestion
- Run the most recent command containg with `string`:
```

bl-ue (commented):
The `!$` and `!^` examples possibly could be removed. 🤔

pages/common/histexpand.md
navarroaxel: `containing` @bl-ue ?

navarroaxel:
@bl-ue Maybe we can merge 1 and 2, `!!` something like `(preffix with sudo to re-run as root)`

pages/common/histexpand.md
jopefd: yeah, @bl-ue is right

man history:
_!?string[?] Refer to the most recent command preceding the current position in the history list containing string.  The trailing ? may be omitted if string is followed immediately by a newline._

jopefd:
> @bl-ue Maybe we can merge 1 and 2, `!!` something like `(preffix with sudo to re-run as root)`

Yes, I agree! It will be only:

```markdown
 - Run the previous command:

   !!
```

navarroaxel:
> Yes, I agree! It will be only:
> 
> ```gfm
>  - Run the previous command:
> 
>    !!
> ```

@jopefd I think to that example description we can add `(preffix with sudo to re-run as root)` to let it explicitly clear.

jopefd:
> explicitly

Ah, understood now!

marchersimon (approved):
Looks good, thanks @jopefd

pages/common/histexpand.md
```diff
@@ -3,10 +3,6 @@
 > Reuse and expand the shell history in `sh`, `bash`, `zsh`, `rbash` and `ksh`.
 > More information: <https://www.gnu.org/software/bash/manual/html_node/History-Interaction>.
 
-- Run the previous command:
-
-`!!`
```
sbrl, line 8: Despite the fact that this example is simple, I think it's one of the most important on the page, as it introduces the user to the concept of history expansion with something they may already be familiar with:

marchersimon: It's sometimes hard to keep in mind for whom tldr is indented, so I have to agree. I think it would be best if we would make a poll, if and which other command to remove.

(BTW, how do some people have the ability to always find the right xkcd in the right moment?)

sbrl: \* indented → intended

I'd prefer no polls here - generally speaking we try to talk it out instead, so everyone's happy with the outcome. @ waldyrious explained this in a previous issue/PR a while back

marchersimon: Makes sense. So, how if we romevo one of those:
* `!-n` - I know I wanted to keep it at the beginning but I'm really sure that it's really useful anymore
* One of either `!^` or `!$`, since I think they're both rarely used. Maybe we could even combine them somehow.

pages/common/histexpand.md
```diff
@@ -19,13 +15,17 @@
 
 `{{command}} !^`
 
+- Run the Nth command of the history:
+
+`!{{N}}`
```
marchersimon, line 20:
```suggestion
`!{{n}}`
```
We use `n` all over the page and I think it's more universal for that purpuse.

pages/common/histexpand.md
sbrl: What about removing `!^`? `!$` can be useful  e.g. in this situation:

```
nano path/to/file
node !$
```

marchersimon: Any other opinions? @bl-ue, @navarroaxel, @CleanMachine1

navarroaxel: I think we can merge `!!` and `sudo !!` but otherwise remove `!^` is a good option.

navarroaxel: I think we can merge `!!` and `sudo !!` but otherwise remove `!^` is a good option.

marchersimon: That's also an option. If we merged the first to into
```
- Run the previous command as root (`!!` replaces the previous command):

sudo !!
```
we'd solve the problem while still explaining the basic function of `!!` and not deleting any example.

sbrl: Ah, that works @navarroaxel @marchersimon - let's do that.

navarroaxel: @jopefd could you apply this change?

pages/common/histexpand.md
```diff
@@ -3,10 +3,6 @@
 > Reuse and expand the shell history in `sh`, `bash`, `zsh`, `rbash` and `ksh`.
 > More information: <https://www.gnu.org/software/bash/manual/html_node/History-Interaction>.
 
-- Run the previous command:
-
-`!!`
-
 - Run the previous command as root:
 
 `sudo !!`
```
marchersimon, line 8:
```suggestion
- Run the previous command as root (`!!` replaces the previous command):

`sudo !!`
```

pages/common/histexpand.md
marchersimon: I made a [suggestion](https://github.com/tldr-pages/tldr/pull/6051/files#r657176121) for that

navarroaxel (approved):

marchersimon (approved):
👍🏽

pages/common/histexpand.md
```diff
@@ -3,7 +3,7 @@
 > Reuse and expand the shell history in `sh`, `bash`, `zsh`, `rbash` and `ksh`.
 > More information: <https://www.gnu.org/software/bash/manual/html_node/History-Interaction>.
 
-- Run the previous command as root:
+- Run the previous command as root (`!!` replaces the previous command):
```
sbrl, line 6: Perhaps this will read better?
```suggestion
- Run the previous command as root (`!!` is replaced by the previous command):
```

sbrl (approved):
Thanks, everyone! @jopefd - sorry about the length of the review process here haha - it's not usually like this.

PR #6052, opened 2021-05-26
ls: fix typo

mobidyc:
"double curly braces" is an unknown syntax, nor by bash nor ksh or zsh even less by ls.
and currently, no need them to reach the aim.

### DIFF
--- a/pages/common/ls.md
+++ b/pages/common/ls.md
@@ -33,4 +33,4 @@
 
 - Only list directories:
 
-`ls -d {{*/}}`
+`ls -d */`


bl-ue:
Hey there, @mobidyc! Thank you for the PR!

I don't think this change is correct though.  You're definitely right—little or no commands would actually _use_ the double curly braces that we put in pages, but that is special syntax	that the clients are supposed to interpret as formatting and remove.  The official Node.js client (which is what I use) removes those and colors the text enclosed, see below:

Screen Shot 2021-05-26 at 11 14 53 AM

We use the double curly braces whenever a command arguments are variable (can be changed by the user), e.g. a filepath, hostname, port, etc.

CleanMachine1:
This is wrong and should probably be closed, the curly brackets mean something that can be changed like a variable name or name of a file

CleanMachine1:
@bl-ue @marchersimon should we close?

bl-ue:
No, we'll wait until @mobidyc responds.

sbrl:
If they don't reply by the time GitHub Actions comes by and comments, then we can close this.

PR #6053, opened 2021-05-26
ls: add example

mobidyc:
Fastest, non I/O blocking file listing

### DIFF
--- a/pages/common/ls.md
+++ b/pages/common/ls.md
@@ -34,3 +34,7 @@
 - Only list directories:
 
 `ls -d {{*/}}`
+
+- List all files with no stat, sort, color nor any slowdown (useful when listing huge number of files)
+
+`ls -fNu1`


pages/common/ls.md
```diff
@@ -34,3 +34,7 @@
 - Only list directories:
 
 `ls -d {{*/}}`
+
+- List all files with no stat, sort, color nor any slowdown (useful when listing huge number of files)
```
CleanMachine1, line 38:
```suggestion
- List all files with no stat, sort, color nor any slowdown
```

CleanMachine1:
Unless we can merge a command together, this will have to be closed, since the max command in a page is tl:dr.

sbrl (changes requested):
Hey there, @mobidyc! Thanks for the edit. Unfortunately, this brings us above our limit of 8 examples per page. Is there a less useful example we could remove here?

pages/common/ls.md
bl-ue: I don't agree with removing this parenthetical statement.  How about using "large" instead of "huge":
```suggestion
- List all files with no stat, sort, color nor any slowdown (useful when listing large numbers of files)
```

marchersimon:
I think we could either merge `-la` and `-lh` into `-lah` or remove `-F`, since it seems to be the least useful (except for scripting maybe).

sbrl:
I'm reluctant to merge examples - especially on a page like this - because it makes the page more dense/complex and increases cognitive load.

`-F` is useful in order to distinguish files from directories - I have it as a flag in my ls aliases.

marchersimon:
When looking at it that way, tldr is still indented for new command line users, which should get a general idea about a command's usage. Even though `ls -fNu1` surely can be useful, I'm not sure if a typical beginner would need it.

PR #6068, opened 2021-05-29
nvm: add Indonesian translation

reinhart1010:

### DIFF
--- a/pages.id/common/nvm.md
+++ b/pages.id/common/nvm.md
@@ -1,0 +1,33 @@
+# nvm
+
+> Memasang, melepas, atau mengganti versi Node.js yang dipakai.
+> Mendukung nomor versi seperti "0.12" or "v4.2", dan label versi seperti "stable", "system", dsb.
+> Informasi lebih lanjut: <https://github.com/creationix/nvm>.
+
+- Memasang versi Node.js yang ditentukan:
+
+`nvm install {{versi_node_js}}`
+
+- Menggunakan versi Node.js tertentu untuk sesi saat ini:
+
+`nvm use {{versi_node_js}}`
+
+- Menyetel versi Node.js secara default:
+
+`nvm alias default {{versi_node_js}}`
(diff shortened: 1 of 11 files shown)


bl-ue (approved):

pages.id/common/nvm.md
```diff
@@ -0,0 +1,33 @@
+# nvm
+
+> Memasang, melepas, atau mengganti versi Node.js yang dipakai.
```
navarroaxel, line 3: gTranslate says `mengganti` is more similar to `replace` in English. Switch in the main page is used to say that you can go to another node.js version and go back when you want to use another. I just want to check if this is the best verb in Indonesian.

sbrl (approved):

Looks good to me after @navarroaxel's comment is addressed.

pages.id/common/nvm.md
reinhart1010: It's possible to replace `mengganti` to `mengalihkan` which has a more similar meaning to `switch`. However, the phrase `mengganti versi Node.js yang dipakai` is more understandable towards Indonesian users yet still have the closest meaning of `replace the currently used Node.js version` or `use a specific version of Node.js in the current shell` (as in the description of `nvm use node_version`).

I also didn't use the phrase `in the current shell` in the translation since it's difficult to translate `(computer) shell` in Indonesian. The closest word for `shell` is [`cangkang`](https://kbbi.kemdikbud.go.id/entri/cangkang), which means the `shell` as in animal's body (e.g. molluscs). Some people do translate this term into `syel`, which is [**not** a standard Indonesian word](https://kbbi.kemdikbud.go.id/entri/syel) at all.

So my strategy for translating this page to Indonesian is by telling them that `nvm` allows users to replace the currently-used Node.js version, as in current shell session.

navarroaxel (approved):

navarroaxel (approved):

PR #5848, opened 2021-04-27
*: fix typos reported by Hunspell

bl-ue:
Fixes a manually corrected selection of typos reported by [Hunspell](https://hunspell.github.io/) in all of the English pages.

Closes #5945

PR #5921, opened 2021-05-10
anbox: add page

SethFalco:
- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #6069, opened 2021-05-29
*: fix errors reported by languagetool

SethFalco:
Inspired by: https://github.com/tldr-pages/tldr/pull/5848

This only does the English files, LanguageTool does support other languages, but I can't manually review the suggestions.
I dumped only lines starting with `>` or `-` from the English files into a single file to save time.

The following is a non-exhaustive list of the changes made to the repository.

#### Hyphenated Words or Redundant Spaces
* human readable -> human-readable
* Un-select -> Unselect
* time scale -> timescale
* user name -> username

#### Fixed Nouns
* wifi / WiFi -> [Wi-Fi](https://en.wikipedia.org/wiki/Wi-Fi)
* tetris -> Tetris
* coachdb -> CoachDB
* heroku -> Heroku
* monday -> Monday
* friday -> Friday
* vue -> Vue

#### Fixed Acronyms
* url -> URL
* uri -> URI
* usb -> USB
* instances like pdf, png, jpg, xml, css, html, etc.
* instances like cpu, ram, etc

#### Typos
* pathames -> pathnames
* parseable -> parsable

### Misc Errors
* its vs it's
* an vs a
* singular vs plural
* double words (ie and and, or the the)
* inserting the word "the" ie "the largest", or "the highest", etc

#### Redundancies
These technically aren't incorrect, just longer than it needs to be.
* all of the -> all the
* short summary -> summary (short is inferred)
* in order to -> to

### Ignored
The following I ignored on purpose, feel free to drop opinions on them.

#### British vs American English
I also ignored locale-specific reports like British vs American English as discussed on Matrix. ([More Info](https://matrix.to/#/!zXiOpjSkFTvtMpsenJ:gitter.im/$jECymhRKgh-Yf55UbktkX9U7Q2iz8Dt4HYs_9phXacs?via=gitter.im&via=matrix.org&via=matrix.coredump.ch))

### False Positives
It's common to refer to things like npm in lowercase even though it's technically an acronym. This makes LanguageTool mad since it thinks it's an actual word. ie `a npm package` vs `an npm package`. It thinks "a npm package" is correct, but because we _mean_ "a NPM package" it should be "an npm package".

---

I decided to be terrible and fixed another thing on the side in this PR.
There was a command that had extra spaces between arguments, so I also removed that. Nothing to do with LanguageTool.
* pages/common/acme.sh-dns.md

### DIFF
--- a/pages/android/getprop.md
+++ b/pages/android/getprop.md
@@ -27,6 +27,6 @@
 
 `getprop {{ro.oem_unlock_supported}}`
 
-- Display the MAC address of the Android's WiFi card:
+- Display the MAC address of the Android's Wi-Fi card:
 
 `getprop {{ro.boot.wifimacaddr}}`
--- a/pages/common/2to3.md
+++ b/pages/common/2to3.md
@@ -11,7 +11,7 @@
 
 `2to3 --write {{path/to/file.py}}`
 
-- Convert specific Python 2 language features to Python 3:
+- Convert a specific Python 2 language features to Python 3:
 
 `2to3 --write {{path/to/file.py}} --fix={{raw_input}} --fix={{print}}`
 
--- a/pages/common/2to3.md
+++ b/pages/common/2to3.md
@@ -11,7 +11,7 @@
 
 `2to3 --write {{path/to/file.py}}`
 
-- Convert a specific Python 2 language features to Python 3:
+- Convert specific Python 2 language features to Python 3:
 
 `2to3 --write {{path/to/file.py}} --fix={{raw_input}} --fix={{print}}`
 
--- a/pages/common/aria2c.md
+++ b/pages/common/aria2c.md
@@ -16,7 +16,7 @@
 
 `aria2c --force-sequential {{url_1}} {{url_2}}`
 
-- Download from multiple sources, each URI pointing to the same file:
+- Download from multiple sources with each URI pointing to the same file:
 
 `aria2c {{url_1}} {{url_2}}`
 
--- a/pages/common/aria2c.md
+++ b/pages/common/aria2c.md
@@ -16,7 +16,7 @@
 
 `aria2c --force-sequential {{url_1}} {{url_2}}`
 
-- Download from multiple sources each URI pointing to the same file:
+- Download from multiple sources, each URI pointing to the same file:
 
 `aria2c {{url_1}} {{url_2}}`
 
--- a/pages/common/bedtools.md
+++ b/pages/common/bedtools.md
@@ -4,7 +4,7 @@
 > Used to intersect, group, convert and count data in BAM, BED, GFF/GTF, VCF format.
 > More information: <https://bedtools.readthedocs.io/en/latest/>.
 
-- Intersect two files with respect to the sequences' strand and save the result to {{`path/to/output_file`}}:
+- Intersect two files regarding the sequences' strand and save the result to {{`path/to/output_file`}}:
 
 `bedtools intersect -a {{path/to/file_1}} -b {{path/to/file_2}} -s > {{path/to/output_file}}`
 
--- a/pages/common/bedtools.md
+++ b/pages/common/bedtools.md
@@ -4,7 +4,7 @@
 > Used to intersect, group, convert and count data in BAM, BED, GFF/GTF, VCF format.
 > More information: <https://bedtools.readthedocs.io/en/latest/>.
 
-- Intersect two files regarding the sequences' strand and save the result to {{`path/to/output_file`}}:
+- Intersect two files regarding the sequences' strand and save the result to the specified file:
 
 `bedtools intersect -a {{path/to/file_1}} -b {{path/to/file_2}} -s > {{path/to/output_file}}`
 
--- a/pages/common/browser-sync.md
+++ b/pages/common/browser-sync.md
@@ -7,7 +7,7 @@
 
 `browser-sync start --server {{path/to/directory}} --files {{path/to/directory}}`
 
-- Start a server from local directory, watching all css files in some directory:
+- Start a server from local directory, watching all CSS files in some directory:
 
 `browser-sync start --server --files '{{path/to/directory/*.css}}'`
 
--- a/pages/common/browser-sync.md
+++ b/pages/common/browser-sync.md
@@ -7,7 +7,7 @@
 
 `browser-sync start --server {{path/to/directory}} --files {{path/to/directory}}`
 
-- Start a server from local directory, watching all CSS files in some directory:
+- Start a server from local directory, watching all CSS files in a directory:
 
 `browser-sync start --server --files '{{path/to/directory/*.css}}'`
 
--- a/pages/common/chroma.md
+++ b/pages/common/chroma.md
@@ -11,6 +11,6 @@
 
 `chroma --lexer="{{go}}" --formatter="{{html}}" {{source_file}} > {{html_file}}`
 
-- Highlight a source file with C++ lexer and output to an SVG image, using the Monokai style:
+- Highlight a source file with C++ lexer and output to an SVG, using the Monokai style:
 
 `chroma --lexer="{{c++}}" --formatter="{{svg}}" --syle="{{monokai}}" {{source_file}} > {{svg_file}}`
--- a/pages/common/chroma.md
+++ b/pages/common/chroma.md
@@ -7,7 +7,7 @@
 
 `chroma --lexer="{{python}}" {{source_file}}`
 
-- Highlight a source file with Go lexer and output to a HTML file:
+- Highlight a source file with Go lexer and output to an HTML file:
 
 `chroma --lexer="{{go}}" --formatter="{{html}}" {{source_file}} > {{html_file}}`
 
--- a/pages/common/chroma.md
+++ b/pages/common/chroma.md
@@ -7,10 +7,10 @@
 
 `chroma --lexer="{{python}}" {{source_file}}`
 
-- Highlight a source file with Go lexer and output to an HTML file:
+- Highlight a source file with the Go lexer and output to an HTML file:
 
 `chroma --lexer="{{go}}" --formatter="{{html}}" {{source_file}} > {{html_file}}`
 
-- Highlight a source file with C++ lexer and output to an SVG, using the Monokai style:
+- Highlight a source file with the C++ lexer and output to an SVG, using the Monokai style:
 
 `chroma --lexer="{{c++}}" --formatter="{{svg}}" --syle="{{monokai}}" {{source_file}} > {{svg_file}}`
--- a/pages/common/df.md
+++ b/pages/common/df.md
@@ -7,7 +7,7 @@
 
 `df`
 
-- Display all filesystems and their disk usage in human readable form:
+- Display all filesystems and their disk usage in human-readable form:
 
 `df -h`
 
--- a/pages/common/dfc.md
+++ b/pages/common/dfc.md
@@ -3,7 +3,7 @@
 > Gives an overview of the filesystem disk space usage with colours and graphs.
 > More information: <https://projects.gw-computing.net/projects/dfc/wiki>.
 
-- Display filesystems and their disk usage in human readable form with colours and graphs:
+- Display filesystems and their disk usage in human-readable form with colours and graphs:
 
 `dfc`
 
--- a/pages/common/diffoscope.md
+++ b/pages/common/diffoscope.md
@@ -11,7 +11,7 @@
 
 `diffoscope --no-progress {{path/to/file1}} {{path/to/file2}}`
 
-- Compare two files and write a HTML-report to a file (use `-` for stdout):
+- Compare two files and write an HTML-report to a file (use `-` for stdout):
 
 `diffoscope --html {{path/to/outfile|-}} {{path/to/file1}} {{path/to/file2}}`
 
--- a/pages/common/docker-network.md
+++ b/pages/common/docker-network.md
@@ -7,7 +7,7 @@
 
 `docker network ls`
 
-- Create a user defined network:
+- Create a user-defined network:
 
 `docker network create --driver {{driver_name}} {{network_name}}`
 
--- a/pages/common/docker.md
+++ b/pages/common/docker.md
@@ -23,7 +23,7 @@
 
 `docker pull {{image}}`
 
-- Open a shell inside an already running container:
+- Open a shell inside a running container:
 
 `docker exec -it {{container_name}} {{sh}}`
 
--- a/pages/common/docker.md
+++ b/pages/common/docker.md
@@ -23,7 +23,7 @@
 
 `docker pull {{image}}`
 
-- Open a shell inside of an already running container:
+- Open a shell inside an already running container:
 
 `docker exec -it {{container_name}} {{sh}}`
 
--- a/pages/common/dolt.md
+++ b/pages/common/dolt.md
@@ -1,6 +1,6 @@
 # dolt
 
-> Dolt is a SQL database that you can fork, clone, branch, merge, push and pull just like a Git repository.
+> Dolt is an SQL database that you can fork, clone, branch, merge, push and pull just like a Git repository.
 > More information: <https://github.com/dolthub/dolt>.
 
 - Execute a dolt subcommand:
--- a/pages/common/du.md
+++ b/pages/common/du.md
@@ -11,7 +11,7 @@
 
 `du -h {{path/to/directory}}`
 
-- Show the size of a single directory, in human readable units:
+- Show the size of a single directory, in human-readable units:
 
 `du -sh {{path/to/directory}}`
 
--- a/pages/common/fswatch.md
+++ b/pages/common/fswatch.md
@@ -3,7 +3,7 @@
 > A cross-platform file change monitor.
 > More information: <https://emcrisostomo.github.io/fswatch>.
 
-- Run a bash command on file creation, update or deletion:
+- Run a Bash command on file creation, update or deletion:
 
 `fswatch {{path/to/file}} | xargs -n 1 {{bash_command}}`
 
@@ -15,6 +15,6 @@
 
 `fswatch {{path/to/directory}} | xargs -n 1 -I {} echo {}`
 
-- Filter by event type, eg. Updated, Deleted or Created:
+- Filter by event type, e.g. Updated, Deleted or Created:
 
 `fswatch --event {{Updated}} {{path/to/directory}} | xargs -n 1 {{bash_command}}`
--- a/pages/common/fswatch.md
+++ b/pages/common/fswatch.md
@@ -15,6 +15,6 @@
 
 `fswatch {{path/to/directory}} | xargs -n 1 -I {} echo {}`
 
-- Filter by event type, e.g. Updated, Deleted or Created:
+- Filter by event type:
 
-`fswatch --event {{Updated}} {{path/to/directory}} | xargs -n 1 {{bash_command}}`
+`fswatch --event {{Updated|Deleted|Created}} {{path/to/directory}} | xargs -n 1 {{bash_command}}`
--- a/pages/common/git-count-objects.md
+++ b/pages/common/git-count-objects.md
@@ -7,7 +7,7 @@
 
 `git count-objects`
 
-- Display a count of all objects and their total disk usage, displaying sizes in human readable units:
+- Display a count of all objects and their total disk usage, displaying sizes in human-readable units:
 
 `git count-objects --human-readable`
 
@@ -15,6 +15,6 @@
 
 `git count-objects --verbose`
 
-- Display more verbose information, displaying sizes in human readable units:
+- Display more verbose information, displaying sizes in human-readable units:
 
 `git count-objects --human-readable --verbose`
--- a/pages/common/git-describe.md
+++ b/pages/common/git-describe.md
@@ -1,6 +1,6 @@
 # git describe
 
-> Give an object a human readable name based on an available ref.
+> Give an object a human-readable name based on an available ref.
 > More information: <https://git-scm.com/docs/git-describe>.
 
 - Create a unique name for the current commit (the name contains the most recent annotated tag, the number of additional commits, and the abbreviated commit hash):
--- a/pages/common/guetzli.md
+++ b/pages/common/guetzli.md
@@ -7,10 +7,10 @@
 
 `guetzli {{input.jpg}} {{output.jpg}}`
 
-- Create compressed JPEG image from PNG image:
+- Create a compressed JPEG from a PNG:
 
 `guetzli {{input.png}} {{output.jpg}}`
 
-- Compress a JPEG image with desired visual quality (84-100):
+- Compress a JPEG with the desired visual quality (84-100):
 
 `guetzli --quality {{quality_value}} {{input.jpg}} {{output.jpg}}`
--- a/pages/common/hostess.md
+++ b/pages/common/hostess.md
@@ -3,7 +3,7 @@
 > An idempotent command-line utility for managing the `/etc/hosts` file.
 > More information: <https://github.com/cbednarski/hostess>.
 
-- List domains, target IPs and on/off status:
+- List domains, target IP addresses and on/off status:
 
 `hostess list`
 
--- a/pages/common/hostess.md
+++ b/pages/common/hostess.md
@@ -3,7 +3,7 @@
 > An idempotent command-line utility for managing the `/etc/hosts` file.
 > More information: <https://github.com/cbednarski/hostess>.
 
-- List domains, target ips and on/off status:
+- List domains, target IPs and on/off status:
 
 `hostess list`
 
@@ -15,6 +15,6 @@
 
 `hostess del {{local.example.com}}`
 
-- Disable a domain (but don't remove it completely):
+- Disable a domain (but don't remove it):
 
 `hostess off {{local.example.com}}`
--- a/pages/common/identify.md
+++ b/pages/common/identify.md
@@ -3,6 +3,6 @@
 > Command-line utility of Image Magick project to describe the format and characteristics of one or more image files.
 > More information: <https://imagemagick.org/script/identify.php>.
 
-- Collect dimensions of all jpeg files under current directory:
+- Collect dimensions of all JPEG files under current directory:
 
 `identify -format "%f,%w,%h\n" *.{{jpg}} > {{filelist.csv}}`
--- a/pages/common/import.md
+++ b/pages/common/import.md
@@ -7,7 +7,7 @@
 
 `import -window root {{output.postscript}}`
 
-- Capture contents of a remote X server screen in the PNG image format:
+- Capture contents of a remote X server screen in the PNG format:
 
 `import -window root -display {{remote_host}}:{screen}.{display} {{output.png}}`
 
--- a/pages/common/import.md
+++ b/pages/common/import.md
@@ -9,7 +9,7 @@
 
 - Capture contents of a remote X server screen in the PNG format:
 
-`import -window root -display {{remote_host}}:{screen}.{display} {{output.png}}`
+`import -window root -display {{remote_host}}:{{screen}}.{{display}} {{output.png}}`
 
 - Capture a specific window, given its ID as displayed by `xwininfo`, into the JPEG format:
 
--- a/pages/common/install.md
+++ b/pages/common/install.md
@@ -4,22 +4,22 @@
 > Copy files (often executable) to a system location like `/usr/local/bin`, give them the appropriate permissions/ownership.
 > More information: <https://www.gnu.org/software/coreutils/install>.
 
-- Copy files to destination:
+- Copy files to the destination:
 
 `install {{path/to/source}} {{path/to/destination}}`
 
-- Copy files to destination, setting their ownership:
+- Copy files to the destination, setting their ownership:
 
 `install -o {{user}} {{path/to/source}} {{path/to/destination}}`
 
-- Copy files to destination, setting their group ownership:
+- Copy files to the destination, setting their group ownership:
 
 `install -g {{user}} {{path/to/source}} {{path/to/destination}}`
 
-- Copy files to destination, setting their `mode`:
+- Copy files to the destination, setting their `mode`:
 
 `install -m {{+x}} {{path/to/source}} {{path/to/destination}}`
 
-- Copy files and apply access/modification times of source to destination:
+- Copy files and apply access/modification times of source to the destination:
 
 `install -p {{path/to/source}} {{path/to/destination}}`
--- a/pages/common/jhat.md
+++ b/pages/common/jhat.md
@@ -11,6 +11,6 @@
 
 `jhat -p {{port}} {{dump_file.bin}}`
 
-- Analyze a dump letting jhat use up to 8GB RAM (2-4x dump size recommended):
+- Analyze a dump letting jhat use up to 8 GB RAM (2-4x dump size recommended):
 
 `jhat -J-mx8G {{dump_file.bin}}`
--- a/pages/common/jhat.md
+++ b/pages/common/jhat.md
@@ -3,7 +3,7 @@
 > Java Heap Analysis Tool.
 > More information: <https://docs.oracle.com/javase/8/docs/technotes/tools/unix/jhat.html>.
 
-- Analyze a heap dump (from jmap), view via http on port 7000:
+- Analyze a heap dump (from jmap), view via HTTP on port 7000:
 
 `jhat {{dump_file.bin}}`
 
--- a/pages/common/jhat.md
+++ b/pages/common/jhat.md
@@ -3,7 +3,7 @@
 > Java Heap Analysis Tool.
 > More information: <https://docs.oracle.com/javase/8/docs/technotes/tools/unix/jhat.html>.
 
-- Analyze a heap dump (from jmap), view via HTTP on port 7000:
+- Analyze a heap dump (from `jmap`), view via HTTP on port 7000:
 
 `jhat {{dump_file.bin}}`
 
@@ -11,6 +11,6 @@
 
 `jhat -p {{port}} {{dump_file.bin}}`
 
-- Analyze a dump letting jhat use up to 8 GB RAM (2-4x dump size recommended):
+- Analyze a dump letting `jhat` use up to 8 GB RAM (2-4x dump size recommended):
 
 `jhat -J-mx8G {{dump_file.bin}}`
--- a/pages/common/ls.md
+++ b/pages/common/ls.md
@@ -19,7 +19,7 @@
 
 `ls -la`
 
-- Long format list with size displayed using human readable units (KiB, MiB, GiB):
+- Long format list with size displayed using human-readable units (KiB, MiB, GiB):
 
 `ls -lh`
 
--- a/pages/common/mogrify.md
+++ b/pages/common/mogrify.md
@@ -12,7 +12,7 @@
 
 `mogrify -resize {{800x600}} {{DSC*}}`
 
-- Convert all PNG images in the directory to JPEG:
+- Convert all PNGs in the directory to JPEG:
 
 `mogrify -format {{jpg}} {{*.png}}`
 
--- a/pages/common/mogrify.md
+++ b/pages/common/mogrify.md
@@ -16,7 +16,7 @@
 
 `mogrify -format {{jpg}} {{*.png}}`
 
-- Half the saturation of all image files in the current directory:
+- Halve the saturation of all image files in the current directory:
 
 `mogrify -modulate {{100,50}} {{*}}`
 
--- a/pages/common/mogrify.md
+++ b/pages/common/mogrify.md
@@ -16,7 +16,7 @@
 
 `mogrify -format {{jpg}} {{*.png}}`
 
-- Halve the saturation of all image files in the current directory:
+- Half the saturation of all image files in the current directory:
 
 `mogrify -modulate {{100,50}} {{*}}`
 
--- a/pages/common/montage.md
+++ b/pages/common/montage.md
@@ -20,6 +20,6 @@
 
 `montage {{image1.png}} {{image2.jpg}} {{imageN.png}} -geometry +0+0 -tile 2x3 montage_%d.jpg`
 
-- Resize and crop images to completely fill their grid cells before tiling:
+- Resize and crop images to fill their grid cells before tiling:
 
 `montage {{image1.png}} {{image2.jpg}} {{imageN.png}} -geometry +0+0 -resize 640x480^ -gravity center -crop 640x480+0+0 montage.jpg`
--- a/pages/common/pathchk.md
+++ b/pages/common/pathchk.md
@@ -3,7 +3,7 @@
 > Check the validity and portability of one or more pathnames.
 > More information: <https://www.gnu.org/software/coreutils/pathchk>.
 
-- Check pathames for validity in the current system:
+- Check pathnames for validity in the current system:
 
 `pathchk {{path1 path2 …}}`
 
--- a/pages/common/pg_dump.md
+++ b/pages/common/pg_dump.md
@@ -3,7 +3,7 @@
 > Extract a PostgreSQL database into a script file or other archive file.
 > More information: <https://www.postgresql.org/docs/current/app-pgdump.html>.
 
-- Dump database into a SQL-script file:
+- Dump database into an SQL-script file:
 
 `pg_dump {{db_name}} > {{output_file.sql}}`
 
--- a/pages/common/pio-system.md
+++ b/pages/common/pio-system.md
@@ -3,7 +3,7 @@
 > Miscellaneous system commands for PlatformIO.
 > More information: <https://docs.platformio.org/en/latest/core/userguide/system/>.
 
-- Install shell completion for the current shell (supports Bash, fish, zsh and PowerShell):
+- Install shell completion for the current shell (supports Bash, Fish, Zsh and PowerShell):
 
 `pio system completion install`
 
--- a/pages/common/pio-system.md
+++ b/pages/common/pio-system.md
@@ -3,7 +3,7 @@
 > Miscellaneous system commands for PlatformIO.
 > More information: <https://docs.platformio.org/en/latest/core/userguide/system/>.
 
-- Install shell completion for the current shell (supports bash, fish, zsh and powershell):
+- Install shell completion for the current shell (supports Bash, fish, zsh and PowerShell):
 
 `pio system completion install`
 
--- a/pages/common/pretty-bytes.md
+++ b/pages/common/pretty-bytes.md
@@ -1,13 +1,13 @@
 # pretty-bytes
 
-> Convert bytes to a human readable string.
+> Convert bytes to a human-readable string.
 > More information: <https://github.com/sindresorhus/pretty-bytes-cli>.
 
-- Convert numeric bytes value to a human readable string:
+- Convert numeric bytes value to a human-readable string:
 
 `pretty-bytes {{1337}}`
 
-- Convert numeric bytes value from stdin to a human readable string:
+- Convert numeric bytes value from stdin to a human-readable string:
 
 `echo {{1337}} | pretty-bytes`
 
--- a/pages/common/pueue-completions.md
+++ b/pages/common/pueue-completions.md
@@ -1,9 +1,9 @@
 # pueue completions
 
-> Generates shell completion files for bash, elvish, fish, powershell, and zsh.
+> Generates shell completion files for Bash, elvish, fish, PowerShell, and zsh.
 > More information: <https://github.com/Nukesor/pueue>.
 
-- Generate completions for bash:
+- Generate completions for Bash:
 
 `sudo pueue completions bash {{/usr/share/bash-completion/completions/pueue.bash}}`
 
--- a/pages/common/pueue-completions.md
+++ b/pages/common/pueue-completions.md
@@ -1,6 +1,6 @@
 # pueue completions
 
-> Generates shell completion files for Bash, elvish, fish, PowerShell, and zsh.
+> Generates shell completion files for Bash, Elvish, Fish, PowerShell, and Zsh.
 > More information: <https://github.com/Nukesor/pueue>.
 
 - Generate completions for Bash:
--- a/pages/common/qpdf.md
+++ b/pages/common/qpdf.md
@@ -19,7 +19,7 @@
 
 `qpdf --split-pages=n {{input.pdf}} {{out_%d.pdf}}`
 
-- Rotate certain pages of a pdf with a given angle:
+- Rotate certain pages of a PDF with a given angle:
 
 `qpdf --rotate={{90:2,4,6}} --rotate={{180:7-8}} {{input.pdf}} {{output.pdf}}`
 
--- a/pages/common/qpdf.md
+++ b/pages/common/qpdf.md
@@ -23,6 +23,6 @@
 
 `qpdf --rotate={{90:2,4,6}} --rotate={{180:7-8}} {{input.pdf}} {{output.pdf}}`
 
-- Remove the password from a password protected file:
+- Remove the password from a password-protected file:
 
 `qpdf --password={{password}} --decrypt {{input.pdf}} {{output.pdf}}`
--- a/pages/common/quota.md
+++ b/pages/common/quota.md
@@ -2,7 +2,7 @@
 
 > Display users' disk space usage and allocated limits.
 
-- Show disk quotas in human readable units for the current user:
+- Show disk quotas in human-readable units for the current user:
 
 `quota -s`
 
--- a/pages/common/rclone.md
+++ b/pages/common/rclone.md
@@ -3,7 +3,7 @@
 > CLI program to copy/sync/move files and directories to and from many cloud services.
 > More information: <https://rclone.org>.
 
-- List contents of a directory on an rclone remote:
+- List contents of a directory on a rclone remote:
 
 `rclone lsf {{remote_name}}:{{path/to/directory}}`
 
--- a/pages/common/rclone.md
+++ b/pages/common/rclone.md
@@ -3,7 +3,7 @@
 > CLI program to copy/sync/move files and directories to and from many cloud services.
 > More information: <https://rclone.org>.
 
-- List contents of a directory on a rclone remote:
+- List contents of a directory on an rclone remote:
 
 `rclone lsf {{remote_name}}:{{path/to/directory}}`
 
--- a/pages/common/rg.md
+++ b/pages/common/rg.md
@@ -20,7 +20,7 @@
 
 `rg {{regular_expression}} {{set_of_subdirs}}`
 
-- Search for a regular expression in files matching a glob (e.g., `README.*`):
+- Search for a regular expression in files matching a glob (e.g. `README.*`):
 
 `rg {{regular_expression}} --glob {{glob}}`
 
--- a/pages/common/rg.md
+++ b/pages/common/rg.md
@@ -12,7 +12,7 @@
 
 `rg --no-ignore --hidden {{regular_expression}}`
 
-- Search for a regular expression only in a certain filetype (e.g., html, css, etc.):
+- Search for a regular expression only in a certain filetype (e.g., HTML, CSS, etc.):
 
 `rg --type {{filetype}} {{regular_expression}}`
 
--- a/pages/common/rg.md
+++ b/pages/common/rg.md
@@ -12,7 +12,7 @@
 
 `rg --no-ignore --hidden {{regular_expression}}`
 
-- Search for a regular expression only in a certain filetype (e.g., HTML, CSS, etc.):
+- Search for a regular expression only in a certain filetype (e.g. HTML, CSS, etc.):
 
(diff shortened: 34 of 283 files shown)


pages/common/pathchk.md
```diff
@@ -3,7 +3,7 @@
 > Check the validity and portability of one or more pathnames.
 > More information: <https://www.gnu.org/software/coreutils/pathchk>.
 
-- Check pathames for validity in the current system:
+- Check pathnames for validity in the current system:
```
bl-ue, line 6: Cool.  I'm going to add `pathames` to codespell.

pages/common/rclone.md
```diff
@@ -3,7 +3,7 @@
 > CLI program to copy/sync/move files and directories to and from many cloud services.
 > More information: <https://rclone.org>.
 
-- List contents of a directory on an rclone remote:
+- List contents of a directory on a rclone remote:
```
bl-ue, line 6: "a rclone"?

pages/linux/aptitude.md
```diff
@@ -23,7 +23,7 @@
 
 `aptitude remove {{package}}`
 
-- Upgrade installed packages to newest available versions:
+- Upgrade installed packages to the newest available versions:
```
bl-ue, line 26: Possibly, `their` might be better.
```suggestion
- Upgrade installed packages to their newest available versions:
```

pages/linux/faketime.md
```diff
@@ -11,6 +11,6 @@
 
 `faketime '{{yesterday}}' {{bash}}`
 
-- Simulate how any program would act next friday night:
+- Simulate how any program would act next Friday night:
 
 `faketime '{{next Friday 1 am}}' {{path/to/any/program}}`
```
bl-ue, line 16: `Any` sounds weird here — let's fix it now:
```suggestion
- Simulate how a program would act next Friday night:

`faketime '{{next Friday 1 am}}' {{path/to/program}}`
```

pages/linux/sreport.md
```diff
@@ -11,6 +11,6 @@
 
 `sreport job sizes printjobcount`
 
-- Show users with highest cpu time use:
+- Show users with the highest cpu time use:
```
bl-ue, line 14:
```suggestion
- Show users with the highest CPU time use:
```

pages/sunos/prctl.md
```diff
@@ -8,7 +8,7 @@
 
 `prctl {{PID}}`
 
-- Examine process limits and permissions in machine parseable format:
+- Examine process limits and permissions in machine parsable format:
```
bl-ue, line 11: I'll add this to codespell.

bl-ue (commented):
Wow, this is some really superb work — the most ultimate height of typo-fixing that I could have ever imagined 🚀🚀🚀

pages/common/rclone.md
SethFalco: Ahh, sorry about that.
Thanks for catching that, I'll undo that one.

bl-ue:
Here are my opinions regarding your ignored changes (in a nutshell, don't ignore any of them!):

>instances like pdf, psd, png, jpg, xml, css, html, etc. Technically, when used on its own it should be uppercase.

I think we should definitely uppercase these.

>instances like cpu, ram, gpu, psu, etc. (These should definitely be uppercase.)

Uppercase these too.

>all of the -> all the

This has tricked me alot, I think we should use "all the"

>short summary -> summary (short is inferred)

Definitely.

>in order to -> to

Yep, too wordy is beating around the bush.

pages/common/df.md
```diff
@@ -7,7 +7,7 @@
 
 `df`
 
-- Display all filesystems and their disk usage in human readable form:
+- Display all filesystems and their disk usage in human-readable form:
```
CleanMachine1, line 10: I dislike this change.

pages/common/dfc.md
```diff
@@ -3,7 +3,7 @@
 > Gives an overview of the filesystem disk space usage with colours and graphs.
```
CleanMachine1, line 3: I dislike this change.

pages/common/dfc.md
```diff
@@ -3,7 +3,7 @@
...
 > More information: <https://projects.gw-computing.net/projects/dfc/wiki>.
 
-- Display filesystems and their disk usage in human readable form with colours and graphs:
+- Display filesystems and their disk usage in human-readable form with colours and graphs:
```
CleanMachine1, line 6: I dislike this change.

pages/common/diffoscope.md
```diff
@@ -11,7 +11,7 @@
 
 `diffoscope --no-progress {{path/to/file1}} {{path/to/file2}}`
 
-- Compare two files and write a HTML-report to a file (use `-` for stdout):
+- Compare two files and write an HTML-report to a file (use `-` for stdout):
```
CleanMachine1, line 14: I dislike this change. It is wrong IMO and should be left as it was.

pages/common/docker-network.md
```diff
@@ -7,7 +7,7 @@
 
 `docker network ls`
 
-- Create a user defined network:
+- Create a user-defined network:
```
CleanMachine1, line 10: I dislike this change.

pages/common/du.md
```diff
@@ -11,7 +11,7 @@
 
 `du -h {{path/to/directory}}`
 
-- Show the size of a single directory, in human readable units:
+- Show the size of a single directory, in human-readable units:
```
CleanMachine1, line 14: I dislike this change.

pages/common/git-count-objects.md
```diff
@@ -7,14 +7,14 @@
...
 `git count-objects --human-readable`
 
 - Display more verbose information:
 
 `git count-objects --verbose`
 
-- Display more verbose information, displaying sizes in human readable units:
+- Display more verbose information, displaying sizes in human-readable units:
 
 `git count-objects --human-readable --verbose`
```
CleanMachine1, line 20: I dislike this change.

pages/common/git-describe.md
```diff
@@ -1,6 +1,6 @@
 # git describe
 
-> Give an object a human readable name based on an available ref.
+> Give an object a human-readable name based on an available ref.
```
CleanMachine1, line 3: I dislike this change.

pages/common/ls.md
```diff
@@ -19,7 +19,7 @@
 
 `ls -la`
 
-- Long format list with size displayed using human readable units (KiB, MiB, GiB):
+- Long format list with size displayed using human-readable units (KiB, MiB, GiB):
```
CleanMachine1, line 22: I dislike this change.

pages/common/pretty-bytes.md
```diff
@@ -1,13 +1,13 @@
...
+> Convert bytes to a human-readable string.
 > More information: <https://github.com/sindresorhus/pretty-bytes-cli>.
 
-- Convert numeric bytes value to a human readable string:
+- Convert numeric bytes value to a human-readable string:
 
 `pretty-bytes {{1337}}`
 
-- Convert numeric bytes value from stdin to a human readable string:
+- Convert numeric bytes value from stdin to a human-readable string:
```
CleanMachine1, line 11: I dislike this change.

pages/common/qpdf.md
```diff
@@ -23,6 +23,6 @@
 
 `qpdf --rotate={{90:2,4,6}} --rotate={{180:7-8}} {{input.pdf}} {{output.pdf}}`
 
-- Remove the password from a password protected file:
+- Remove the password from a password-protected file:
```
CleanMachine1, line 26: I dislike this change.

pages/common/quota.md
```diff
@@ -2,7 +2,7 @@
 
 > Display users' disk space usage and allocated limits.
 
-- Show disk quotas in human readable units for the current user:
+- Show disk quotas in human-readable units for the current user:
```
CleanMachine1, line 5: I dislike this change.

pages/common/vdir.md
```diff
@@ -8,7 +8,7 @@
 
 `vdir`
 
-- List with sizes displayed in human readable units (KB, MB, GB):
+- List with sizes displayed in human-readable units (KB, MB, GB):
```
CleanMachine1, line 11: I dislike this change.

pages/common/weasyprint.md
```diff
@@ -3,11 +3,11 @@
 > Render HTML to PDF or PNG.
 > More information: <https://weasyprint.org/>.
 
-- Render a HTML file to PDF:
+- Render an HTML file to PDF:
 
 `weasyprint {{path/to/input.html}} {{path/to/output}}.pdf`
 
-- Render a HTML file to PNG, including an additional user stylesheet:
+- Render an HTML file to PNG, including an additional user stylesheet:
```
CleanMachine1, line 11: I dislike this change. It should be a.

Read these 2:

An h t m l 
A h t m l

pages/common/xmllint.md
```diff
@@ -18,7 +18,7 @@
 
 `xmllint --format {{source_file.xml}}`
 
-- Check that a XML file meets the requirements of its DOCTYPE declaration:
+- Check that an XML file meets the requirements of its DOCTYPE declaration:
```
CleanMachine1, line 21: This is wrong as well IMO however this can be desputed

pages/linux/aptitude.md
CleanMachine1: I'd say `the` is better. Why give it personification?

pages/linux/dockerd.md
```diff
@@ -7,7 +7,7 @@
 
 `dockerd`
 
-- Run docker daemon and config it to listen to specific sockets(unix,tcp):
+- Run docker daemon and config it to listen to specific sockets (unix, tcp):
```
CleanMachine1, line 10:
```suggestion
- Run docker daemon and config it to listen to specific sockets (unix and tcp):
```

I don't know about this one.

pages/linux/free.md
```diff
@@ -10,7 +10,7 @@
 
 `free -{{b|k|m|g}}`
 
-- Display memory in human readable units:
+- Display memory in human-readable units:
```
CleanMachine1, line 13: I dislike this change.

pages/linux/lsscsi.md
```diff
@@ -10,6 +10,6 @@
 
 `lsscsi -L`
 
-- List all SCSI devices with human readable disk capacity:
+- List all SCSI devices with human-readable disk capacity:
```
CleanMachine1, line 13: I dislike this change.

pages/osx/du.md
```diff
@@ -10,7 +10,7 @@
 
 `du -h {{path/to/directory}}`
 
-- Show the size of a single directory, in human readable units:
+- Show the size of a single directory, in human-readable units:
```
CleanMachine1, line 13: I dislike this change.

pages/common/diffoscope.md
bl-ue: https://www.gramlee.com/blog/indefinite-articles-a-or-an-before-h-blame-the-confusion-on-the-french/

CleanMachine1: Is that you agreeing with me, from the document, thats a yes

SethFalco: This isn't a subjective matter, "an HTML" is the correct use of a/an.

In English, we say "a" if the next syllable is pronounced starting with the sound of a consonant.
We say "an" if the next syllable starts with the sound of a vowel.

For example:
"an item"
"an element"

"a report"
"a record"

However note it's the _sound_, not the _letter_.

For example, "an HTML" because it's pronounced with an "aye" sound at the start.

pages/common/qpdf.md
SethFalco: This one is up for debate that we'll see what others say.

I can vouch in the Oxford Dictionary it's "password-protected".

https://www.lexico.com/definition/password-protected

pages/common/xmllint.md
marchersimon: Most people read `XML` as `ex em el`, which then would require `an`. Same for `HTML`.

pages/common/qpdf.md
marchersimon: I prefer `password-protected`. A hypen can often improve readability.

SethFalco: Generally, I think the hyphen is better because the word "password" is being used as a modifier.
In this case, it's "a protected file", but we modify the word protected to say how it is protected, hence "password-protected".

Similarly to "human-readable", the word human is used as a modifier of the word "readable".

More on the topic:
https://en.wikipedia.org/wiki/English_compound#Hyphenated_compound_modifiers

Even on the Wikipedia page is does note that different style guides have different recommendations for it. The only real rule is, use a hyphen if the text might be ambiguous without it.

pages/common/browser-sync.md
```diff
@@ -7,7 +7,7 @@
 
 `browser-sync start --server {{path/to/directory}} --files {{path/to/directory}}`
 
-- Start a server from local directory, watching all css files in some directory:
+- Start a server from local directory, watching all CSS files in some directory:
```
bl-ue, line 10:
```suggestion
- Start a server from local directory, watching all CSS files in a directory:
```

pages/common/identify.md
```diff
@@ -3,6 +3,6 @@
 > Command-line utility of Image Magick project to describe the format and characteristics of one or more image files.
 > More information: <https://imagemagick.org/script/identify.php>.
 
-- Collect dimensions of all jpeg files under current directory:
+- Collect dimensions of all JPEG files under current directory:
```
bl-ue, line 6: This could work too, no?
```suggestion
- Collect dimensions of all `.jpg` files under current directory:
```

pages/common/rg.md
```diff
@@ -12,7 +12,7 @@
 
 `rg --no-ignore --hidden {{regular_expression}}`
 
-- Search for a regular expression only in a certain filetype (e.g., html, css, etc.):
+- Search for a regular expression only in a certain filetype (e.g., HTML, CSS, etc.):
```
bl-ue, line 15:
```suggestion
- Search for a regular expression only in a certain filetype (e.g. HTML, CSS, etc.):
```

pages/common/webpack.md
```diff
@@ -7,7 +7,7 @@
 
 `webpack {{app.js}} {{bundle.js}}`
 
-- Load css files too from the js file (this uses the css loader for `.css` files):
+- Load CSS files too from the JS file (this uses the CSS loader for `.css` files):
```
bl-ue, line 10:
```suggestion
- Load CSS files too from the JavaScript file (this uses the CSS loader for `.css` files):
```

pages/linux/dockerd.md
```diff
@@ -7,7 +7,7 @@
 
 `dockerd`
 
-- Run docker daemon and config it to listen to specific sockets(unix,tcp):
+- Run docker daemon and config it to listen to specific sockets (unix and tcp):
```
bl-ue, line 10:
```suggestion
- Run docker daemon and config it to listen to specific sockets (UNIX and TCP):
```

pages/linux/rtcwake.md
```diff
@@ -6,15 +6,15 @@
...
+- Suspend to RAM and wakeup after 10 seconds:
 
 `sudo rtcwake -m mem -s {{10}}`
 
 - Suspend to disk (higher power saving) and wakeup 15 minutes later:
 
 `sudo rtcwake -m disk --date +{{15}}min`
 
-- Freeze the system (more efficient than suspend-to-ram but linux > 3.9 required) and wakeup at a given date and time:
+- Freeze the system (more efficient than suspend-to-RAM but linux > 3.9 required) and wakeup at a given date and time:
```
bl-ue, line 17: We should normalize this too.
```suggestion
- Freeze the system (more efficient than suspend-to-RAM but Linux > 3.9 required) and wakeup at a given date and time:
```

pages/linux/spectre-meltdown-checker.md
```diff
@@ -7,7 +7,7 @@
 
 `sudo spectre-meltdown-checker`
 
-- Check the currently running kernel and show an explanation of the actions to take in order to mitigate a vulnerability:
+- Check the currently running kernel and show an explanation of the actions to take to mitigate a vulnerability:
```
bl-ue, line 10: I still remember when I wrote this line! 😅

bl-ue:
@SethFalco maybe also normalize `eg.` → `e.g.`?

pages/common/identify.md
SethFalco: That's valid, is that input specific to this command, or for other instances or file types as well?

SethFalco:
I'll stop with the commit spam now. I've applied most of the feedback, and some of the things that I remember skipping before.
Later I'll look at LanguageTool again and see what else it was complaining about.

pages/common/identify.md
bl-ue: No, it can use any ImageMagick-supported file type — there are many.  The example is using `.jpg`.

pages/linux/rtcwake.md
```diff
@@ -6,15 +6,15 @@
...
+- Suspend to RAM and wakeup after 10 seconds:
 
 `sudo rtcwake -m mem -s {{10}}`
 
 - Suspend to disk (higher power saving) and wakeup 15 minutes later:
 
 `sudo rtcwake -m disk --date +{{15}}min`
 
-- Freeze the system (more efficient than suspend-to-ram but linux > 3.9 required) and wakeup at a given date and time:
+- Freeze the system (more efficient than suspend-to-RAM but Linux > 3.9 required) and wakeup at a given date and time:
```
bl-ue, line 17: What do you think about this change?
```suggestion
- Freeze the system (more efficient than suspend-to-RAM but Linux >3.9 required) and wakeup at a given date and time:
```

pages/common/identify.md
SethFalco: Oh nono, I'm familiar with ImageMagick.
I meant is your feedback for this command only, or for every instance of `PNG`, `JPG`, `PDF`, `PSD`, etc I changed.

As in do you think they should all be `.jpg`, `.png`, `.whatever`?
Or is the feedback just for this command since the command example uses `.jpg`?

pages/linux/spectre-meltdown-checker.md
SethFalco: Reminds me whenever I'm at work looking at code. 
I see something I hate, do `git blame` to see who wrote it, and look down in shame when it reveals it was me. ^-^'

pages/common/identify.md
bl-ue: Ah, I see.  I guess it's for all commands but that seems like a lot of work — only do it for all them if you really want to.

pages/linux/rtcwake.md
CleanMachine1: Im personally 50/50. The ` > ` fits better mathematically and grammatically, however `>3.9` is overall better for the situation since it highlights it better.

Also what is Linux 3.9... (ik its the kernel). I really dislike that usage, since their are so many distros and to make it seem like a release annoys me

CleanMachine1: Maybe add a reference to Linux being a kernel as well. Maybe Linux kernel ... >

bl-ue: It's a good usage actually — because it was the _Linux kernel_ (i.e. https://github.com/torvalds/linux) that added support for that, not any particular distro (or GUI).

bl-ue:
Now, why'd you call your branch `im-dead` 😆

SethFalco:
> Now why'd you call your branch `im-dead` laughing

Because going through all the problems LanguageTool reported was much more boring than I expected, I wanted to give a subtle clue to my cause of death in case "dying from boredom" could really happen.

SethFalco:
I'm going to be more terse about the changes, but I'll still try to be a bit useful in this comment.

* MBytes -> MB
* PNG image -> PNG 
* PNG images -> PNGs
* Capitalize nouns like minetest -> Mintest or mixxx -> Mixxx
* Fixed more issues with a vs an
* Capitalized more file formats and other acronyms (WAV, FLAC, PID, IP, etc) (may switch some to `.png` or so later)
* Ethernet starts with an uppercase letter? (Never knew)
* Some minor wording adjustments, ie "Remind when it's time to leave." "Set a reminder for when it's time to leave."
* I don't know who Boccaccio is, but LanguageTool does and Wikipedia approves
* More hyphens
* "GUI interface" I make the same mistake too, but "Graphical User Interface interface" isn't what's intended
* aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

pages/common/2to3.md
```diff
@@ -11,7 +11,7 @@
 
 `2to3 --write {{path/to/file.py}}`
 
-- Convert specific Python 2 language features to Python 3:
+- Convert a specific Python 2 language features to Python 3:
```
bl-ue, line 14: `a` shouldn't be here (`a language features`)

SethFalco: Ahh yeah, thanks for that.
No idea how I did that.

```suggestion
- Convert specific Python 2 language features to Python 3:
```

SethFalco:
Forgot to note, some changes may be up for debate.

Namely:
* PNG images -> PNGs

I also did ignore one thing again, opinions welcome:
* 100GB vs 100 GB (it recommends, and I prefer having a space between the value and the units)

CleanMachine1:
I agree with PNGs, however there should probably be a space between the 100 GB

SethFalco:
* Added a space between value and units. (100GB -> 100 GB or 22,000Hz -> 22,000 Hz)
* Fixed some instances where Gigabytes (GB) and Gibibytes (GiB or G) were mixed up.

pages/common/truncate.md
```diff
@@ -7,11 +7,11 @@
 
 `truncate --size {{10G}} {{filename}}`
 
-- Extend the file size by 50M, fill with holes (which reads as zero bytes):
+- Extend the file size by 50 M, fill with holes (which reads as zero bytes):
```
bl-ue, line 10: Why just `M`?

bl-ue (commented):
Adding a space between the value and the unit is much clearer.

pages/common/truncate.md
SethFalco: Valid, I'll update that!

```suggestion
- Extend the file size by 50 MiB, fill with holes (which reads as zero bytes):
```

bl-ue:
Didn't you say somewhere that only 92 files isn't that many to modify in one PR?  281 is 191 more 🤢 😆

pages/android/getprop.md
```diff
@@ -27,6 +27,6 @@
 
 `getprop {{ro.oem_unlock_supported}}`
 
-- Display the MAC address of the Android's WiFi card:
+- Display the MAC address of the Android's Wi-Fi card:
```
sbrl, line 30: This looks very weird to me, but [it's apparently correct](https://www.wi-fi.org/).

pages/common/montage.md
```diff
@@ -20,6 +20,6 @@
 
 `montage {{image1.png}} {{image2.jpg}} {{imageN.png}} -geometry +0+0 -tile 2x3 montage_%d.jpg`
 
-- Resize and crop images to completely fill their grid cells before tiling:
+- Resize and crop images to fill their grid cells before tiling:
```
sbrl, line 23: This change makes this description potentially ambiguous.

pages/common/ssh-keygen.md
```diff
@@ -14,7 +14,7 @@
 
 `ssh-keygen -t ed25519 -a 100`
 
-- Generate an RSA 4096 bit key with email as a comment:
+- Generate an RSA 4096-bit key with email as a comment:
```
sbrl, line 17: Not sure this is really necessary. Personally I'd prefer it with a space, not a dash.

sbrl (commented):
Just took another quick skim through.

pages/common/ssh-keygen.md
SethFalco: Oh damn, peeking around on Wikipedia it does seem space vs hyphen isn't strict.
I thought hyphen was standard.

Also, a snippet from ssh-keygen itself:
```
compact format, taking as little as one bit per certificate if they are
             numbers are 64-bit values, not including zero and may be ex‐
             file using 128-bit AES.  This file is not automatically accessed
```

We'll see what others think, but I can revert this if preferred.

pages/common/montage.md
SethFalco: Could you elaborate?
I believe the word "completely" should be inferred.

**fill**
_intransitive verb_: To put something into (a container, for example) to capacity or to a desired level.
_intransitive verb_: To supply or provide to the fullest extent.

pages/common/ssh-keygen.md
bl-ue: I like it with the hyphen, tbh.

pages/common/montage.md
sbrl: Maybe, but not always in a technical context. For example, does it mean fill as in stretch to fit, or fill as in fill but not changing the aspect ratio?

pages/common/ssh-keygen.md
sbrl: Ah, ok. If the official documentation for the command uses a hyphen, we should too in this specific instance.

pages/common/montage.md
SethFalco: I will concede to the ambiguity you described, but I don't believe the word "completely" has any impact on it.
What you said could be asked regardless of if "completely" is there or not.

CleanMachine1:
You have some things which conflict @SethFalco

pages/common/aria2c.md
```diff
@@ -16,7 +16,7 @@
 
 `aria2c --force-sequential {{url_1}} {{url_2}}`
 
-- Download from multiple sources each URI pointing to the same file:
+- Download from multiple sources, each URI pointing to the same file:
```
bl-ue, line 19:
```suggestion
- Download from multiple sources with each URI pointing to the same file:
```

pages/common/bedtools.md
```diff
@@ -4,7 +4,7 @@
 > Used to intersect, group, convert and count data in BAM, BED, GFF/GTF, VCF format.
 > More information: <https://bedtools.readthedocs.io/en/latest/>.
 
-- Intersect two files with respect to the sequences' strand and save the result to {{`path/to/output_file`}}:
+- Intersect two files regarding the sequences' strand and save the result to {{`path/to/output_file`}}:
```
bl-ue, line 7:
```suggestion
- Intersect two files regarding the sequences' strand and save the result to the specified file:
```

pages/common/chroma.md
```diff
@@ -7,10 +7,10 @@
 
 `chroma --lexer="{{python}}" {{source_file}}`
 
-- Highlight a source file with Go lexer and output to a HTML file:
+- Highlight a source file with Go lexer and output to an HTML file:
```
bl-ue, line 10:
```suggestion
- Highlight a source file with the Go lexer and output to an HTML file:
```

pages/common/chroma.md
```diff
@@ -7,10 +7,10 @@
...
 
 `chroma --lexer="{{go}}" --formatter="{{html}}" {{source_file}} > {{html_file}}`
 
-- Highlight a source file with C++ lexer and output to an SVG image, using the Monokai style:
+- Highlight a source file with C++ lexer and output to an SVG, using the Monokai style:
```
bl-ue, line 14:
```suggestion
- Highlight a source file with the C++ lexer and output to an SVG, using the Monokai style:
```

pages/common/docker.md
```diff
@@ -23,7 +23,7 @@
 
 `docker pull {{image}}`
 
-- Open a shell inside of an already running container:
+- Open a shell inside an already running container:
```
bl-ue, line 26: ?
```suggestion
- Open a shell inside an already-running container:
```

pages/common/dolt.md
```diff
@@ -1,6 +1,6 @@
 # dolt
 
-> Dolt is a SQL database that you can fork, clone, branch, merge, push and pull just like a Git repository.
+> Dolt is an SQL database that you can fork, clone, branch, merge, push and pull just like a Git repository.
```
bl-ue, line 3: This is tricky! What if you read it (as I do) as "sequel," rather than "S-Q-L"?

I wonder if the latter _is_ more common tho 🤔

pages/common/fswatch.md
```diff
@@ -15,6 +15,6 @@
 
 `fswatch {{path/to/directory}} | xargs -n 1 -I {} echo {}`
 
-- Filter by event type, eg. Updated, Deleted or Created:
+- Filter by event type, e.g. Updated, Deleted or Created:
 
 `fswatch --event {{Updated}} {{path/to/directory}} | xargs -n 1 {{bash_command}}`
```
bl-ue, line 20: While we're at it
```suggestion
- Filter by event type:

`fswatch --event {{Updated|Deleted|Created}} {{path/to/directory}} | xargs -n 1 {{bash_command}}`
```

pages/common/guetzli.md
```diff
@@ -7,10 +7,10 @@
 
 `guetzli {{input.jpg}} {{output.jpg}}`
 
-- Create compressed JPEG image from PNG image:
+- Create a compressed JPEG from a PNG:
 
 `guetzli {{input.png}} {{output.jpg}}`
 
-- Compress a JPEG image with desired visual quality (84-100):
+- Compress a JPEG with the desired visual quality (84-100):
```
bl-ue, line 14: ?
```suggestion
- Compress a JPEG with the specified visual quality (between 84 and 100, inclusive):
```

pages/common/hostess.md
```diff
@@ -3,7 +3,7 @@
 > An idempotent command-line utility for managing the `/etc/hosts` file.
 > More information: <https://github.com/cbednarski/hostess>.
 
-- List domains, target ips and on/off status:
+- List domains, target IPs and on/off status:
```
bl-ue, line 6:
```suggestion
- List domains, target IP addresses and on/off status:
```

pages/common/import.md
```diff
@@ -7,7 +7,7 @@
 
 `import -window root {{output.postscript}}`
 
-- Capture contents of a remote X server screen in the PNG image format:
+- Capture contents of a remote X server screen in the PNG format:
 
 `import -window root -display {{remote_host}}:{screen}.{display} {{output.png}}`
```
bl-ue, line 12: Whoa, something strange happened here (not your fault):
```suggestion
`import -window root -display {{remote_host}}:{{screen}}.{{display}} {{output.png}}`
```

pages/common/install.md
```diff
@@ -4,22 +4,22 @@
...
 
 `install -g {{user}} {{path/to/source}} {{path/to/destination}}`
 
-- Copy files to destination, setting their `mode`:
+- Copy files to the destination, setting their `mode`:
 
 `install -m {{+x}} {{path/to/source}} {{path/to/destination}}`
 
-- Copy files and apply access/modification times of source to destination:
+- Copy files and apply access/modification times of source to the destination:
```
bl-ue, line 23: How about `to the specified destination`, instead of `to the destination`?  (I can't submit a suggestion :/)

pages/common/jhat.md
```diff
@@ -2,14 +2,14 @@
...
 `jhat {{dump_file.bin}}`
 
 - Analyze a heap dump, specifying an alternate port for the http server:
 
 `jhat -p {{port}} {{dump_file.bin}}`
 
-- Analyze a dump letting jhat use up to 8GB RAM (2-4x dump size recommended):
+- Analyze a dump letting jhat use up to 8 GB RAM (2-4x dump size recommended):
 
 `jhat -J-mx8G {{dump_file.bin}}`
```
bl-ue, line 15:
```suggestion
`jhat -J-mx{{8G}} {{dump_file.bin}}`
```

pages/common/jhat.md
```diff
@@ -2,14 +2,14 @@
 
 > Java Heap Analysis Tool.
 
-- Analyze a heap dump (from jmap), view via http on port 7000:
+- Analyze a heap dump (from jmap), view via HTTP on port 7000:
```
bl-ue, line 5:
```suggestion
- Analyze a heap dump (from `jmap`), view via HTTP on port 7000:
```

pages/common/jhat.md
```diff
@@ -2,14 +2,14 @@
...
 
 `jhat {{dump_file.bin}}`
 
 - Analyze a heap dump, specifying an alternate port for the http server:
 
 `jhat -p {{port}} {{dump_file.bin}}`
 
-- Analyze a dump letting jhat use up to 8GB RAM (2-4x dump size recommended):
+- Analyze a dump letting jhat use up to 8 GB RAM (2-4x dump size recommended):
```
bl-ue, line 13:
```suggestion
- Analyze a dump letting `jhat` use up to 8 GB RAM (2-4x dump size recommended):
```

pages/common/pg_dump.md
```diff
@@ -3,7 +3,7 @@
 > Extract a PostgreSQL database into a script file or other archive file.
 > More information: <https://www.postgresql.org/docs/current/app-pgdump.html>.
 
-- Dump database into a SQL-script file:
+- Dump database into an SQL-script file:
```
bl-ue, line 6: Same here regarding "sequel" vs. "S-Q-L".

pages/common/pio-system.md
```diff
@@ -3,7 +3,7 @@
 > Miscellaneous system commands for PlatformIO.
 > More information: <https://docs.platformio.org/en/latest/core/userguide/system/>.
 
-- Install shell completion for the current shell (supports bash, fish, zsh and powershell):
+- Install shell completion for the current shell (supports Bash, fish, zsh and PowerShell):
```
bl-ue, line 6:
```suggestion
- Install shell completion for the current shell (supports Bash, Fish, ZSH and PowerShell):
```

pages/common/pueue-completions.md
```diff
@@ -1,9 +1,9 @@
 # pueue completions
 
-> Generates shell completion files for bash, elvish, fish, powershell, and zsh.
+> Generates shell completion files for Bash, elvish, fish, PowerShell, and zsh.
```
bl-ue, line 3:
```suggestion
> Generates shell completion files for Bash, Elvish, Fish, PowerShell, and ZSH.
```

pages/common/scrapy.md
```diff
@@ -19,14 +19,14 @@
 
 `scrapy crawl {{spider_name}}`
 
-- Fetch a webpage as scrapy sees it and print source in stdout:
+- Fetch a webpage as Scrapy sees it and print source in stdout:
```
bl-ue, line 22:
```suggestion
- Fetch a webpage as Scrapy sees it and print the source to stdout:
```

pages/common/scrapy.md
```diff
@@ -19,14 +19,14 @@
...
 
 `scrapy fetch {{url}}`
 
-- Open a webpage in the default browser as scrapy sees it (disable JavaScript for extra fidelity):
+- Open a webpage in the default browser as Scrapy sees it (disable JavaScript for extra fidelity):
 
 `scrapy view {{url}}`
 
-- Open scrapy shell for URL, which allows interaction with the page source in python shell (or ipython if available):
+- Open Scrapy shell for URL, which allows interaction with the page source in python shell (or ipython if available):
```
bl-ue, line 30:
```suggestion
- Open Scrapy shell for URL, which allows interaction with the page source in a Python shell (or IPython if available):
```

pages/common/webpack.md
```diff
@@ -7,11 +7,11 @@
 
 `webpack {{app.js}} {{bundle.js}}`
 
-- Load css files too from the js file (this uses the css loader for `.css` files):
+- Load CSS files too from the JavaScript file (this uses the CSS loader for `.css` files):
 
 `webpack {{app.js}} {{bundle.js}} --module-bind 'css=css'`
```
bl-ue, line 12:
```suggestion
`webpack {{app.js}} {{bundle.js}} --module-bind '{{css}}={{css}}'`
```

pages/common/webtorrent.md
```diff
@@ -1,7 +1,7 @@
 # webtorrent
 
 > The command-line interface for WebTorrent.
-> Supports magnets, urls, info hashes and .torrent files.
+> Supports magnets, URLs, info hashes and .torrent files.
```
bl-ue, line 4:
```suggestion
> Supports magnets, URLs, info hashes and `.torrent` files.
```

pages/common/xz.md
```diff
@@ -11,11 +11,11 @@
...
 
-- Compress a file to the lzma file format:
+- Compress a file to the LZMA file format:
 
 `xz --format=lzma {{file}}`
 
-- Decompress an lzma file:
+- Decompress an LZMA file:
 
 `xz -d --format=lzma {{file.lzma}}`
```
bl-ue, line 20:
```suggestion
`xz -d --format={{lzma}} {{file.lzma}}`
```

pages/common/xz.md
```diff
@@ -11,11 +11,11 @@
 
 `xz -d {{file.xz}}`
 
-- Compress a file to the lzma file format:
+- Compress a file to the LZMA file format:
 
 `xz --format=lzma {{file}}`
```
bl-ue, line 16:
```suggestion
`xz --format={{lzma}} {{file}}`
```

pages/linux/aspell.md
```diff
@@ -14,7 +14,7 @@
 
 `aspell dicts`
 
-- Run aspell with different language (takes two letter ISO 639 language code):
+- Run aspell with different language (takes two-letter ISO 639 language code):
```
bl-ue, line 17:
```suggestion
- Run aspell with a different language (takes two-letter ISO 639 language code):
```

pages/linux/bluetoothctl.md
```diff
@@ -1,6 +1,6 @@
 # bluetoothctl
 
-> Handling bluetooth devices from the shell.
+> Handling Bluetooth devices from the shell.
```
bl-ue, line 3:
```suggestion
> Manage Bluetooth devices from the command-line.
```

pages/linux/kde-inhibit.md
```diff
@@ -10,6 +10,6 @@
 
 `kde-inhibit --screenSaver {{command}} {{command_arguments}}`
 
-- Launch vlc, and inhibit colour correction (night mode) while it's running:
+- Launch VLC, and inhibit colour correction (night mode) while it's running:
```
bl-ue, line 13: Why not while we're here 🙏🏻
```suggestion
- Launch VLC, and inhibit color correction (night mode) while it's running:
```

pages/linux/rtcwake.md
```diff
@@ -1,20 +1,20 @@
...
+- Suspend to RAM and wakeup after 10 seconds:
 
 `sudo rtcwake -m mem -s {{10}}`
 
 - Suspend to disk (higher power saving) and wakeup 15 minutes later:
 
 `sudo rtcwake -m disk --date +{{15}}min`
 
-- Freeze the system (more efficient than suspend-to-ram but linux > 3.9 required) and wakeup at a given date and time:
+- Freeze the system (more efficient than suspend-to-RAM but Linux kernel >3.9 required) and wakeup at a given date and time:
```
bl-ue, line 17: My stance on this change changed — consider _"When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness."_ from [`CONTRIBUTING.md`](https://github.com/tldr-pages/tldr/blob/main/CONTRIBUTING.md).
```suggestion
- Freeze the system (more efficient than suspend-to-RAM but 3.9 or greater of the Linux kernel is required) and wakeup at a given date and time:
```

pages/linux/zramctl.md
```diff
@@ -15,7 +15,7 @@
 
 `sudo modprobe zram num_devices={{2}}`
 
-- Find and initialise the next free zram device to a 2GB virtual drive using LZ4 compression:
+- Find and initialise the next free zram device to a 2 GB virtual drive using LZ4 compression:
 
 `sudo zramctl --find --size {{2GB}} --algorithm {{lz4}}`
```
bl-ue, line 20: This page `linux/zramctl.md` uses `zram` in the descriptions — shouldn't it be `ZRAM`?

pages/osx/sed.md
```diff
@@ -27,7 +27,7 @@
 
 `sed -e 's/{{find}}/{{replace}}/' -e 's/{{find}}/{{replace}}/' {{filename}}`
 
-- Replace separator / by any other character not used in the find or replace patterns, e.g., #:
+- Replace separator / by any other character not used in the find or replace patterns, e.g. #:
```
bl-ue, line 30:
```suggestion
- Replace separator `/ by any other character not used in the find or replace patterns, e.g. `#`:
```

pages/windows/eventcreate.md
```diff
@@ -4,7 +4,7 @@
 > Event IDs can be any number between 1 and 1000.
 > More information: <https://docs.microsoft.com/windows-server/administration/windows-commands/eventcreate>.
 
-- Create a new event with a given id (1-1000) in the log:
+- Create a new event with a given ID (1-1000) in the log:
```
bl-ue, line 7: ?
```suggestion
- Create a new event with a given ID (1 to 1000 inclusive) in the log:
```

pages/windows/robocopy.md
```diff
@@ -20,7 +20,7 @@
 
 `robocopy {{path/to/source}} {{path/to/destination}} /E /XO`
 
-- List all files 50 MBytes or larger in size instead of copying them:
+- List all files 50 MB or larger instead of copying them:
 
 `robocopy {{path/to/source}} {{path/to/destination}} /MIN:52428800 /L`
```
bl-ue, line 25:
```suggestion
`robocopy {{path/to/source}} {{path/to/destination}} /MIN:{{52428800}} /L`
```

bl-ue (changes requested):
31 comments! :D

pages/common/docker.md
SethFalco: Not 100% sure about this one, but I can find examples of "already-running" on Wikipedia when used in a similar context. 

> This results in a delay to the handling of the connection (in comparison to a connection handled by an **already-running** process). 
> \- https://en.wikipedia.org/wiki/Super-server#Performance

However, most of the time it's not done like this. (The example above was only because I specifically searched for it.)

It looks like "already-running" isn't on Merriam-Webster, Lexico, or Collins dictionary either.

To be honest, I feel this should just say "a running container" anyway. The "already" doesn't add anything of value.

```suggestion
- Open a shell inside a running container:
```

Thoughts?

bl-ue: Like it. 👍🏻

pages/common/dolt.md
SethFalco: Valid concern. 🤔 

I'd lean on "an" myself.
LanguageTool reports this should be "An SQL", but that's ofc going to be on the assumption everyone uses S-Q-L.

> The original standard declared that the official pronunciation for "SQL" was an initialism: /ˌɛsˌkjuːˈɛl/ ("ess cue el").[9] Regardless, many English-speaking database professionals (including Donald Chamberlin himself[31]) use the acronym-like pronunciation of /ˈsiːkwəl/ ("sequel"),[32] mirroring the language's prerelease development name, "SEQUEL".[13][14][31]
> \- https://en.wikipedia.org/wiki/SQL#Standardization_history

Based on this, I'd vouch "an SQL" - and people that chose to read it as "sequel" can just auto-correct it their head.

bl-ue: Sure.  I'll go ahead and keep after Mr. Chamberlin and use "seque," but I concede the point here.

pages/common/jhat.md
SethFalco: Looking at the docs, I think it'd be better to do this instead?

```
-J<flag>
             Pass <flag> to the Java virtual machine on which jhat is run. For
             example, -J-Xmx512m to use a maximum heap size of 512MB.
```

```suggestion
`jhat -J{{-mx8G}} {{dump_file.bin}}`
```

The whole `-mx8G` is the user-input, not just `8G`.
I think it accepts any Java VM argument, though.

pages/common/pio-system.md
SethFalco: ZSH should be Zsh I think.

> The Z shell (Zsh) is a Unix shell that can be used as an interactive login shell and as a command interpreter for shell scripting.
> \- https://en.wikipedia.org/wiki/Z_shell

> This site provides an index to Zsh information and archives.
> \- https://www.zsh.org/

> Zsh is a shell with lots of features.
> \- https://sourceforge.net/p/zsh/code/ci/master/tree/

```suggestion
- Install shell completion for the current shell (supports Bash, Fish, Zsh and PowerShell):
```

pages/common/jhat.md
bl-ue: Yeah, you're definitely right about that, but the example itself was only about about changing the max memory.

SethFalco: My understanding is that it's _all_ user-defined input that needs to be put between curly braces?

> Basically, everything which isn't a flag, argument or subcommand should be in {{ }}. Also it makes the page more readable.
> 
> 
> \- https://github.com/tldr-pages/tldr/pull/5921#issuecomment-836462426

In the example in the PR the command was also for a very specific action, however user-defined parameters should still be wrapped, if I understand correctly.

pages/common/pio-system.md
bl-ue: Yeah sorry, I saw it as `ZSH` somewhere but I didn't verify.

pages/common/jhat.md
bl-ue: It's mostly contextual, but I think you're right here.

pages/linux/rtcwake.md
SethFalco: It said it must be newer than 3.9 before, but you wrote or must be 3.9 or newer on your one.
I think we need to verify which it's supposed to be, but overall thing your suggestion is nice.

bl-ue: Hmm, I definitely didn't mean to change that, but I think it's actually correct!
From https://manned.org/rtcwake:
>The processes are frozen, all the devices are suspended and all the processors idled. ... (Available since Linux 3.9.)

pages/common/webpack.md
SethFalco: Note for me I always see the following output when I attempt to use the argument locally on both version 4 and 5. 🤔 

```bash
[webpack-cli] Error: Unknown option '--module-bind'
[webpack-cli] Run 'webpack --help' to see available commands and options
```

It looks to me that this option may have been removed in Webpack 5? But I'm not able to confirm this.

On Webpack v4 it's still listed:
https://v4.webpack.js.org/api/cli/#module-options

On Webpack v5 it's gone:
https://webpack.js.org/api/cli/

Regardless, assuming it's still in version 5:
I think it should wrap the full value, it seems based on this tutorial it's not always going to be `{{...}}={{...}}`.

> ```js
> webpack --module-bind pug-loader --module-bind 'css=style-loader!css-loader'
> ```
> \- https://timhieuvewebpack.blogspot.com/#cli

```suggestion
`webpack {{app.js}} {{bundle.js}} --module-bind '{{css=css}}'`
```

If it was indeed removes, it could be nice to remove from here to make space for more relevant commands.

pages/common/pueue-completions.md
SethFalco:
```suggestion
> Generates shell completion files for Bash, elvish, fish, PowerShell, and Zsh.
```

Don't mind this, just adding the comment, so I can merge the version with Zsh instead of ZSH.

pages/windows/eventcreate.md
SethFalco: I've always been a bit of a fan of the Rust way of denoting this, but I guess based on your Linux comment it might not be favorable to use here. 🤔 

I believe they do `..` for exclusive range (1..1000), and `..=` for inclusive range (1..=1000).

I think it's important to denote if it's inclusive/exclusive, but wondering if there's really any need to replace the "-" with "to".
Any other opinions?

---

Something we should probably think about at some point is the localization of numbers. 🤔
There are some places (though not many) that we use thousand separators in numbers.

Consider the following: 3,141

To someone in France, does this mean 3.141 (three point one four one), or 3,141 (three thousand one hundred and forty-one).

As it's not localized, there's simply no way to tell without context or knowing if we localize strings or not.

SethFalco: Oh wait, I'm an idiot. -_-
Ignore my comment about localizing numbers, I guess that can be handled in said translations as is.

Sorry, I was thinking about how like in software we need to specific set a locale in because numbers can be calculated. I guess that doesn't matter in this case since all numbers are essentially static strings, and can be translated as such.

bl-ue: >To someone in France, does this mean 3.141 (three point one four one), or 3,141 (three thousand one hundred and forty-one).

Good point.  Removing the separator altogether seems like the most expedient solution: 3141

bl-ue: >I've always been a bit of a fan of the Rust way of denoting this, but I guess based on your Linux comment it might not be favorable to use here. 🤔 
>
>I believe they do `..` for exclusive range (1..1000), and `..=` for inclusive range (1..=1000).

I agree, Rust's terseness for numeric ranges awesome, but...

>When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness.

SethFalco: > Good point. Removing the separator altogether seems like the most expedient solution: 3141

I remember seeing a localized number before, so I assumed we had many. Just checked, but it seems we only have the one:

> Benchmark a database with 10 clients, 2 worker threads, and 10,000 transactions per client:
> \- phbench.md

And this one doesn't suffer from the ambiguity described above, so I guess it's not a big deal.

---

Think we should include something about this in the style guide anyway?

Regardless, I believe we need to ensure all numbers have no localization.

From there, a decision can be made:
* Leave them delocalized for all users and that's it.
* Use a syntax to denote localizable numbers, tldr clients would be responsible for if they want to apply the system locale or not, or may choose to make it a setting for users.

I'm a fan of localized numbers myself, as I find them easier to read. It depends on what everyone else thinks, though.

pages/linux/zramctl.md
SethFalco: Just peeked now but I think `zram` is fine?

> Initially, zram had only the latter function, hence the original name "compcache" ("compressed cache"). 
> \- https://en.wikipedia.org/wiki/Zram

> The zram module creates RAM-based block devices named /dev/zram<id> (<id> = 0, 1, …).
\- https://www.kernel.org/doc/html/latest/admin-guide/blockdev/zram.html

pages/windows/eventcreate.md
sbrl: Following [a conversation on Gitter](https://gitter.im/tldr-pages/tldr?at=60cf826b258ebf6d92dced1e), we established that locales are not the same as language. So, we have 2 options:

1. Localise numbers to the language of the translation (but someone can have a localisation setting that isn't the same as the language they are viewing the page in
2. Avoid localising numbers at all

marchersimon: I don't really understand why the client has something to do with this. If I have an English page I write `2021-06-23` and if I translate that to German, I just write `23.06.2021`. Or am I misunderstanding something?

SethFalco: Language and locale are different things, more info at https://github.com/tldr-pages/tldr/issues/6147

SethFalco:
I'll return to this tomorrow, sorry about that!

pages/common/mogrify.md
```diff
@@ -12,11 +12,11 @@
 
 `mogrify -resize {{800x600}} {{DSC*}}`
 
-- Convert all PNG images in the directory to JPEG:
+- Convert all PNGs in the directory to JPEG:
 
 `mogrify -format {{jpg}} {{*.png}}`
 
-- Halve the saturation of all image files in the current directory:
+- Half the saturation of all image files in the current directory:
```
sbrl, line 19: This doesn't make sense. `Halve` is the verb here - `Half` doesn't work the same way.

SethFalco: Aaaa, good catch. Not sure what I was thinking there. 🤔 

```suggestion
- Halve the saturation of all image files in the current directory:
```

pages/common/montage.md
sbrl: I think it's fill as in the CSS keyword `cover` here, but I'm unsure as to how to improve the description.

marchersimon (approved):
Thanks @SethFalco

CleanMachine1:
I seriously don't think I can sit through nearly 300 pages of looking at typos.

As long as at least 1 has done a review its fine by me.

If I am too lazy to review, I don't know how Seth felt going through editing them...

`SethFalco:im-dead` gives a good idea.

marchersimon:
Could you approve then?

CleanMachine1 (approved):

sbrl (approved):
Let's get this merged.

sbrl:
Looks like there's a conflict in `pages/common/paci.md`

SethFalco:
> Looks like there's a conflict in pages/common/paci.md

Thanks for the heads-up, should be good to go!

marchersimon:
Nice, I'll merge

PR #6083, opened 2021-06-01
pacman: use long arguments and replace -Qo example

CleanMachine1:

For me personally the command I have removed seems a little less important compared to the command I have added.

### DIFF
--- a/pages/linux/pacman.md
+++ b/pages/linux/pacman.md
@@ -31,6 +31,6 @@
 
 `pacman --query --unrequired --deps --quiet`
 
-- Empty package cache to free up space:
+- Remove not installed packages and unused repositories from the cache (use two `--clean` flags to clean all packages):
 
 `pacman --sync cc`
--- a/pages/linux/pacman.md
+++ b/pages/linux/pacman.md
@@ -27,7 +27,7 @@
 
 `pacman -Qe`
 
-- List orphan dependencies:
+- List orphan packages (installed as dependencies but not required by any package):
 
 `pacman -Qtdq`
 
--- a/pages/linux/pacman.md
+++ b/pages/linux/pacman.md
@@ -5,27 +5,27 @@
 
 - Synchronize and update all packages:
 
-`pacman -Syu`
+`pacman --sync --refresh --sysupgrade`
 
 - Install a new package:
 
-`pacman -S {{package_name}}`
+`pacman --sync {{package_name}}`
 
 - Remove a package and its dependencies:
 
-`pacman -Rs {{package_name}}`
+`pacman --remove --recursive {{package_name}}`
 
 - Search the package database for a regular expression or keyword:
 
-`pacman -Ss "{{search_pattern}}"`
+`pacman --sync --search "{{search_pattern}}"`
 
 - List installed packages and versions:
 
-`pacman -Q`
+`pacman --query`
 
 - List only the explicitly installed packages and versions:
 
-`pacman -Qe`
+`pacman --query --explicit`
 
 - List orphan packages (installed as dependencies but not required by any package):
 
@@ -33,4 +33,4 @@
 
 - Empty package cache to free up space:
 
-`pacman -Scc`
+`pacman --sync cc`
--- a/pages/linux/pacman.md
+++ b/pages/linux/pacman.md
@@ -27,9 +27,9 @@
 
 `pacman -Qe`
 
-- Find which package owns a certain file:
+- List orphan dependencies:
 
-`pacman -Qo {{filename}}`
+`pacman -Qtdq`
 
 - Empty package cache to free up space:
 
--- a/pages/linux/pacman.md
+++ b/pages/linux/pacman.md
@@ -31,6 +31,6 @@
 
 `pacman --query --unrequired --deps --quiet`
 
-- Remove not installed packages and unused repositories from the cache (use two `--clean` flags to clean all packages):
+- Empty the entire pacman cache:
 
 `pacman --sync --clean --clean`
--- a/pages/linux/pacman.md
+++ b/pages/linux/pacman.md
@@ -33,4 +33,4 @@
 
 - Remove not installed packages and unused repositories from the cache (use two `--clean` flags to clean all packages):
 
-`pacman --sync cc`
+`pacman --sync --clean --clean`
--- a/pages/linux/pacman.md
+++ b/pages/linux/pacman.md
@@ -29,7 +29,7 @@
 
 - List orphan packages (installed as dependencies but not required by any package):
 
-`pacman -Qtdq`
+`pacman --query --unrequired --deps --quiet`
 
 - Empty package cache to free up space:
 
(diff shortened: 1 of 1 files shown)


pages/linux/pacman.md
```diff
@@ -27,9 +27,9 @@
 
 `pacman -Qe`
 
-- Find which package owns a certain file:
+- List orphan dependencies:
 
-`pacman -Qo {{filename}}`
+`pacman -Qtdq`
```
navarroaxel, line 32:
```suggestion
`pacman --query --unrequired --deps --quiet`
```

pages/linux/pacman.md
```diff
@@ -27,9 +27,9 @@
 
 `pacman -Qe`
 
-- Find which package owns a certain file:
+- List orphan dependencies:
```
navarroaxel, line 30:
```suggestion
- List orphan packages (installed as dependencies but not required by any package):
```

navarroaxel (changes requested):
This example is documented in the `pacman-query` page.

pages/linux/pacman.md
CleanMachine1: Shouldn't we change this for all of the commands here?

because that would make the command `sudo pacman -Syu` something which would probably be harder to remember than Syu

navarroaxel: Yeah, I think we should, all the subcommands are in long form.

CleanMachine1:
Note to self. 

Make all the commands long.

CleanMachine1:
@navarroaxel I couldn't find the alternative for the last command. All I could find was cascade and clean.

What's the difference between `sudo pacman -Sc` and `sudo pacman -Scc`

Thanks

marchersimon:
When used with `-R`, it means `--cascade`, but when used with `-S` it means `--clean`. From the manpage:

> Use one --clean switch to only remove packages that are no longer installed; use two to remove all files from the cache.

marchersimon:
Btw, this page has also a `fr`, `ml`, `tr` and `zh` version.

CleanMachine1:
So just 'pacman --sync --clean --clean'

pages/linux/pacman.md
```diff
@@ -5,32 +5,32 @@
...
-`pacman -Qe`
+`pacman --query --explicit`
 
-- Find which package owns a certain file:
+- List orphan packages (installed as dependencies but not required by any package):
 
-`pacman -Qo {{filename}}`
+`pacman --query --unrequired --deps --quiet`
 
 - Empty package cache to free up space:
```
navarroaxel, line 34: You can use the verbose explanation from `pacman-sync` page:
```suggestion
- Remove not installed packages and unused repositories from the cache (use two `--clean` flags to clean all packages):
```

pages/linux/pacman.md
```diff
@@ -5,32 +5,32 @@
...
 
-`pacman -Scc`
+`pacman --sync cc`
```
navarroaxel, line 36:
```suggestion
`pacman --sync --clean`
```

marchersimon: Actually that should be `--clean --clean`

CleanMachine1: Ps this was left as a placeholder!

By no means did I intend to leave it as edited.

It was a mix of a placeholder and me just going to sleep since it was very late and just wanting to commit

CleanMachine1:
```suggestion
`pacman --sync --clean --clean`
```

Once we have received a reply from @navarroaxel we will commit the suggestion. I think it needs to be clean clean however he might have other opinions on what is correct.

navarroaxel: double `clean` is ok for me.

navarroaxel: but with this example description, 1 `clean` sounds better. Because you are saying, use 2 for ....

CleanMachine1: 1 clean signifies just `pacman -Sc` which isn't the same

https://wiki.archlinux.org/title/Pacman#Cleaningthepackagecache

pages/linux/pacman.md
```diff
@@ -5,32 +5,32 @@
...
+`pacman --query --explicit`
 
-- Find which package owns a certain file:
+- List orphan packages (installed as dependencies but not required by any package):
 
-`pacman -Qo {{filename}}`
+`pacman --query --unrequired --deps --quiet`
 
-- Empty package cache to free up space:
+- Remove not installed packages and unused repositories from the cache (use two `--clean` flags to clean all packages):
```
marchersimon, line 34: I don't think this is completely correct. `-c` removes all uninstalled packges and `-cc` removes all files from the cache, regardless if they are a package or something else. So I think the original description is fine.

CleanMachine1: What about
```suggestion
Delete the entire cache to free up space
```

CleanMachine1: Empty implies that some junk is removed. Delete is more extreme which is what clean clean does

sbrl: I think both `Empty` and `Delete` are fine here. Personally I'm inclined to agree with @marchersimon on this one.

Also for reference the `pacman` command has been the subject of frequent changes in the past: https://github.com/tldr-pages/tldr/commits/main/pages/linux/pacman.md

navarroaxel: @marchersimon you don't like the current description? It's the same that the last example in the `pacman-sync` page.

marchersimon: `Remove all not installed packeges` describes what `-Sc` does. `-Scc` however empties the entire cache, regardless if the file belongs to a non-installed package.

marchersimon: The last example of `pacman-sync` is correct, because it does describe `-Sc`.

CleanMachine1:
```suggestion
- Delete the entire pacman cache:
```

marchersimon: I'd preferr `empty the entire pacman cache`, since I think it's used more often in context with cache. "Delete the cache" sounds a little off IMO.

CleanMachine1:
```suggestion
- Empty the entire pacman cache:
```

marchersimon (approved):

navarroaxel (approved):

pages/linux/pacman.md
```diff
@@ -5,32 +5,32 @@
...
-`pacman -Q`
+`pacman --query`
 
 - List only the explicitly installed packages and versions:
 
-`pacman -Qe`
+`pacman --query --explicit`
 
-- Find which package owns a certain file:
+- List orphan packages (installed as dependencies but not required by any package):
```
bl-ue, line 30: I'd add "actually" to help convey that a package installed as a dependency is normally depended upon ;)
```suggestion
- List orphan packages (installed as dependencies but not actually required by any package):
```

bl-ue (approved):
One non-blocking comment :)

PR #6087, opened 2021-06-02
ln: convert paths to absolute and fix typo in Italian page

marchersimon:
When running `ln -s path1/file path2/symlink`, while in `path3` you'd expect it to create a symlink `path2/symlink`, which points to `path3/path1/file`. However the symlink points to `path2/path1/file`, so the base for the relative path of the file is actually the directory of the symlink and not the directory this was executed from. The easiest solution is to just specify an absolute path to the file.

### DIFF
--- a/pages.it/common/ln.md
+++ b/pages.it/common/ln.md
@@ -5,12 +5,12 @@
 
 - Crea un collegamento simbolico a un file (o directory):
 
-`ln -s {{percorso/al/file}} {{percorso/al/collegamento}}`
+`ln -s {{/percorso/al/file}} {{percorso/al/collegamento}}`
 
 - Sovrascrivi un collegamento esistente in modo che punti a un nuovo file:
 
-`ln -sg {{percorso/al/nuovo/file}} {{percorso/al/collegamento}}`
+`ln -sg {{/percorso/al/nuovo/file}} {{percorso/al/collegamento}}`
 
 - Crea un collegamento fisico a un file:
 
-`ln {{percorso/al/file}} {{percorso/al/collegamento}}`
+`ln {{/percorso/al/file}} {{percorso/al/collegamento}}`
--- a/pages.it/common/ln.md
+++ b/pages.it/common/ln.md
@@ -9,7 +9,7 @@
 
 - Sovrascrivi un collegamento esistente in modo che punti a un nuovo file:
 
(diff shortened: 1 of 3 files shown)


bl-ue (approved):

CleanMachine1 (approved):

pages.it/common/ln.md
```diff
@@ -5,12 +5,12 @@
 
 - Crea un collegamento simbolico a un file (o directory):
 
-`ln -s {{percorso/al/file}} {{percorso/al/collegamento}}`
+`ln -s {{/percorso/al/file}} {{percorso/al/collegamento}}`
 
 - Sovrascrivi un collegamento esistente in modo che punti a un nuovo file:
 
-`ln -sg {{percorso/al/nuovo/file}} {{percorso/al/collegamento}}`
+`ln -sg {{/percorso/al/nuovo/file}} {{percorso/al/collegamento}}`
```
mfrw, line 12:
```suggestion
`ln -sf {{/percorso/al/nuovo/file}} {{percorso/al/collegamento}}`
```

I guess we are only adding a `/` so should not be an issue to approve although I do not speak/understand the language.

Also the page for italian does not seem to follow the example from the other pages.

mfrw (changes requested):
I think we might have a mismatch in example w.r.t the italian page.
all the other pages have an example with `-sf` but this one has with `-sg` .

PS: I don't speak the lang.. so please excuse me .. if its a known issue

pages.it/common/ln.md
marchersimon: Good catch, it seems like a typo. I checked with [deepl](https://www.deepl.com/translator#it/en/Sovrascrivi%20un%20collegamento%20esistente%20in%20modo%20che%20punti%20a%20un%20nuovo%20file) and according to the description, it should be `-sf`.

bl-ue:
I think so.  Two commits that effect two completely unrelated changes (but since they're both small it's good for them to be in the same PR) should be rebased.

mfrw (approved):
LGTM :)

pages.it/common/ln.md
mfrw: @marchersimon thank you for pointing out deepl. TIL :)

PR #6097, opened 2021-06-06
git-commit: add example

CleanMachine1:

### DIFF
--- a/pages/common/git-commit.md
+++ b/pages/common/git-commit.md
@@ -23,6 +23,6 @@
 
 `git commit {{path/to/file1}} {{path/to/file2}}`
 
-- Commit staged files to the repository with a message while allowing the commit be empty:
+- Create a commit, even if there are no staged files:
 
 `git commit -m "{{message}}" --allow-empty`
--- a/pages/common/git-commit.md
+++ b/pages/common/git-commit.md
@@ -22,3 +22,7 @@
 - Commit only specific (already staged) files:
 
 `git commit {{path/to/file1}} {{path/to/file2}}`
+
+- Commit staged files to the repository with a message while allowing the commit be empty:
+
+`git commit -m "{{message}}" --allow-empty`


pages/common/git-commit.md
```diff
@@ -22,3 +22,7 @@
 - Commit only specific (already staged) files:
 
 `git commit {{path/to/file1}} {{path/to/file2}}`
+
+- Commit staged files to the repository with a message while allowing the commit be empty:
+
+`git commit -m "{{message}}" --allow-empty`
```
navarroaxel, line 28: Could you expand the arguments?
```suggestion
`git commit --message="{{message}}" --allow-empty`
```

CleanMachine1: I understand however its an incredible annoyance. Also most users would probably remember 

`m for message`

marchersimon: I wouldn't be too sure about that. Newcomers to git prabably don't have to commit that often that typing `--message` would be a problem.

marchersimon: At least I'd remove the `=`:
```suggestion
`git commit --message "{{message}}" --allow-empty`
```

pages/common/git-commit.md
```diff
@@ -22,3 +22,7 @@
 - Commit only specific (already staged) files:
 
 `git commit {{path/to/file1}} {{path/to/file2}}`
+
+- Commit staged files to the repository with a message while allowing the commit be empty:
```
marchersimon, line 26: How about just
```suggestion
- Allow empty commits:
```

pages/common/git-commit.md
bl-ue: Hmm, I wonder if with such a common command as `git commit`, we should maybe keep the short options 🤔

I've never used `--message` before if I recall correctly 😅

marchersimon: It's the same situation as with #6083. Nobody would ever use `--sync --clean --clean`, but it's still good to know what they do, especially for beginners.

pages/common/git-commit.md
CleanMachine1: I don't think so. It doesn't describe it and doesn't fit with the page.

```suggestion
- Create a commit which has the ability to be empty with a message
```

bl-ue: That's too wordy.  How about
```suggestion
- Create an empty commit with no added, deleted, or changed files:
```

CleanMachine1: It isn't mandatory for it to empty, the flag just allows it to be.

CleanMachine1: If there are staged commits, it will put them in.

marchersimon: Another suggestion:
```suggestion
- Create a commit, even if there are no staged files:
```

pages/common/git-commit.md
CleanMachine1: yes but them commands are long and multiple flags, this is `m` if I was thinking for a message I would try -m.

The `S` for sync doesn't mean much to me, I'd immediately think search

pages/common/git-commit.md
CleanMachine1: Yes thats fine

pages/common/git-commit.md
sbrl: I would suggest that this is a rare exception to the rule. Generally speaking you're going to be using this command a *lot*, so I'd suggest keeping the short version in this case, and maybe updating the description like we have on the `tar` page for example.

pages/common/git-commit.md
sbrl: ....but surely you wouldn't use this unless you wanted an empty commit in the first place? Either of @marchersimon's suggestions would be fine.

CleanMachine1: You might use it in a script thats automated.

marchersimon (approved):
👍🏽

bl-ue (approved):
👍🏻

pages/common/git-commit.md
CleanMachine1: > Hmm, I wonder if with such a common command as `git commit`, we should maybe keep the short options thinking
> 
> I've never used `--message` before if I recall correctly sweat_smile

You haven't used message or just always used -m?

navarroaxel (approved):

PR #6103, opened 2021-06-07
docker-build: add example with docker from stdin

SethFalco:

More information: https://docs.docker.com/engine/reference/commandline/build/#text-files
This builds an image from stdin, which allows us to build an image with no build context.

### DIFF
--- a/pages/common/docker-build.md
+++ b/pages/common/docker-build.md
@@ -17,7 +17,7 @@
 
 - Build a docker image with no build context:
 
-`docker build --tag {{name:tag}} - < Dockerfile`
+`docker build --tag {{name:tag}} - < {{Dockerfile}}`
 
 - Do not use the cache when building the image:
 
--- a/pages/common/docker-build.md
+++ b/pages/common/docker-build.md
@@ -17,7 +17,7 @@
 
 - Build a docker image with no build context:
 
-`cat {{Dockerfile}} | docker build --tag {{name:tag}} -`
+`docker build --tag {{name:tag}} - < {{Dockerfile}}`
 
 - Do not use the cache when building the image:
 
--- a/pages/common/docker-build.md
+++ b/pages/common/docker-build.md
@@ -17,7 +17,7 @@
 
 - Build a docker image with no build context:
 
-`docker build --tag {{name:tag}} - < {{Dockerfile}}`
+`cat {{Dockerfile}} | docker build --tag {{name:tag}} -`
 
 - Do not use the cache when building the image:
 
--- a/pages/common/docker-build.md
+++ b/pages/common/docker-build.md
@@ -15,6 +15,10 @@
 
 `docker build --tag {{name:tag}} .`
 
+- Build a docker image with no build context:
+
+`docker build --tag {{name:tag}} - < Dockerfile`
+
 - Do not use the cache when building the image:
 
 `docker build --no-cache --tag {{name:tag}} .`


pages/common/docker-build.md
```diff
@@ -15,6 +15,10 @@
 
 `docker build --tag {{name:tag}} .`
 
+- Build a docker image with no build context:
+
+`docker build --tag {{name:tag}} - < Dockerfile`
```
bl-ue, line 20:
```suggestion
`docker build --tag {{name:tag}} - < {{Dockerfile}}`
```

bl-ue (changes requested):

SethFalco:
@bl-ue Sorry, seems I keep being dumb and not adding labels myself. I'll try to remember in future.

pages/common/docker-build.md
```diff
@@ -15,6 +15,10 @@
 
 `docker build --tag {{name:tag}} .`
 
+- Build a docker image with no build context:
+
+`docker build --tag {{name:tag}} - < {{Dockerfile}}`
```
nicokosi, line 20: Seems specific to *nix shells (it's not a valid on Windows): https://docs.docker.com/engine/reference/commandline/build/#text-files

nicokosi: Should this example be after `docker build {{github.com/creack/docker-firefox}}`? In fact, the 2 first examples are related to context: current dir + URL.

SethFalco: > Seems specific to *nix shells (not valid on Windows):

Ahh shoot, yeah that slipped my mind.
I'll update this later to accommodate that.

> Should this example be after docker build {{github.com/creack/docker-firefox}}?

Sorry, I don't understand what you mean? I put the example 4th just because the 3rd is the standard way to use a Dockerfile with the current directly as the build context, and this command is the standard way to use a Dockerfile to have n build context at all.

pages/common/docker-build.md
```diff
@@ -15,6 +15,10 @@
 
 `docker build --tag {{name:tag}} .`
 
+- Build a docker image with no build context:
+
+`cat {{Dockerfile}} | docker build --tag {{name:tag}} -`
```
sbrl, line 20: Could you use a shell redirect instead?
```suggestion
`docker build --tag {{name:tag}} - < {{Dockerfile}}`
```

SethFalco: That's what I had originally, but Windows (CMD/PowerShell) doesn't support that.

CleanMachine1: A lot of the common pages don't work for windows. But they work for Mac and Linux for example.

SethFalco: Fair enough, I'd still say it's nicer to keep things cross-platform where possible, though.
I'd use the `<` syntax day-to-day, however when I'm writing documentation or tutorials where I think Windows users will consume it, I think it's better to use `cat {} |`.

Otherwise, I think it'd be better to duplicate this page to Windows with the Windows specific syntax.

CleanMachine1: I agree, was just giving the point.

pages/common/docker-build.md
sbrl: Nope @nicokosi, that syntax is not specific to *nix shells - it works on Windows too. Not sure on Powershell, but powershell is not the default command line on Windows anyway.

pages/common/docker-build.md
sbrl: Do you have documentation for that? The `command < path/to/file` syntax *does* actually work on Windows as far as I'm aware. Not sure about Powershell, but it definitely works in CMD I'm pretty sure.

SethFalco: Ahh, I only tested in PowerShell when I tried it on Windows. 🤔 
If I can get confirmation that it works in CMD, then I'll concede.

I'll try later, but if someone else uses Windows and can drop a comment here, that'd be nice!

pages/common/docker-build.md
bl-ue: Windows PowerShell doesn't like it 😞

>The '<' operator is reserved for future use.

But that doesn't matter.

pages/common/docker-build.md
bl-ue: PowerShell ❌ 
cmd.exe ✅

sbrl (approved):
Thanks, @SethFalco!

CleanMachine1 (approved):

sbrl:
Merging, since @bl-ue's comments appear to be addressed and @bl-ue [is stepping away from GH for a while](https://gitter.im/tldr-pages/tldr?at=60d615598a40b117282a96d7).

PR #6104, opened 2021-06-07
readlink: add page

Itai-Nelken:

on MacOS `readlink -f` doesn't work, so I copied `readlink.md` to /pages/osx and removed the `-f` option.
Screen Shot 2021-06-08 at 12 33 16 AM

### DIFF
--- a/pages/osx/readlink.md
+++ b/pages/osx/readlink.md
@@ -3,6 +3,6 @@
 > Follow symlinks and get symlink information.
 > More information: <https://www.gnu.org/software/coreutils/readlink>.
 
-- Print the actual full path to which the symlink points:
+- Print the actual full path which the symlink points to:
 
 `readlink {{path/to/symlink}}`
--- a/pages/osx/readlink.md
+++ b/pages/osx/readlink.md
@@ -5,4 +5,4 @@
 
 - Print the actual full path to which the symlink points:
 
-`readlink {{filename}}`
+`readlink {{path/to/symlink}}`
--- a/pages/osx/readlink.md
+++ b/pages/osx/readlink.md
@@ -3,6 +3,6 @@
 > Follow symlinks and get symlink information.
 > More information: <https://www.gnu.org/software/coreutils/readlink>.
 
-- Get the actual file to which the symlink points:
+- Print the actual full path to which the symlink points:
 
 `readlink {{filename}}`
--- a/pages/osx/readlink.md
+++ b/pages/osx/readlink.md
@@ -1,0 +1,8 @@
+# readlink
+
+> Follow symlinks and get symlink information.
+> More information: <https://www.gnu.org/software/coreutils/readlink>.
+
+- Get the actual file to which the symlink points:
+
+`readlink {{filename}}`
--- a/pages/osx/readlink.md
+++ b/pages/osx/readlink.md
@@ -3,6 +3,6 @@
 > Follow symlinks and get symlink information.
 > More information: <https://www.gnu.org/software/coreutils/readlink>.
 
-- Print the actual full path which the symlink points to:
+- Print the absolute path which the symlink points to:
 
 `readlink {{path/to/symlink}}`


pages/osx/readlink.md
```diff
@@ -0,0 +1,8 @@
+# readlink
+
+> Follow symlinks and get symlink information.
+> More information: <https://www.gnu.org/software/coreutils/readlink>.
+
+- Get the actual file to which the symlink points:
```
bl-ue, line 6:
```suggestion
- Print the actual full path to which the symlink points:
```

bl-ue (approved):
Looks good! I've got one suggestion ;)

CleanMachine1:
Glad to see you contributing!

pages/osx/readlink.md
```diff
@@ -0,0 +1,8 @@
+# readlink
+
+> Follow symlinks and get symlink information.
+> More information: <https://www.gnu.org/software/coreutils/readlink>.
+
+- Print the actual full path to which the symlink points:
+
+`readlink {{filename}}`
```
bl-ue, line 8: I'd maybe say `path/to/symlink` as well, wdyt?
```suggestion
`readlink {{path/to/symlink}}`
```

pages/osx/readlink.md
```diff
@@ -0,0 +1,8 @@
+# readlink
+
+> Follow symlinks and get symlink information.
+> More information: <https://www.gnu.org/software/coreutils/readlink>.
+
+- Print the actual full path to which the symlink points:
```
CleanMachine1, line 6:
```suggestion
- Print the actual full path to which the symlink points to:
```
thoughts?

pages/osx/readlink.md
Itai-Nelken: good idea

pages/osx/readlink.md
Itai-Nelken: sounds good but I;m no english expert..

bl-ue: `to` is already there ("... full path **to** which the ...")

bl-ue: No, it's redundant.  I personally prefer the original because it's more advanced-sounding :) but it's probably less readable for non-native-English-speakers.
```suggestion
- Print the actual full path which the symlink points to:
```

Itai-Nelken: so what sounds better
1) `Print the actual full path to which the symlink points:`
2) `Print the actual full path which the symlink points to:`

I think 2 sounds better

CleanMachine1: Yes, both works and the current version is more concise, however with 2 `to` it reads nicer and releases the formality of it which it doesn't really need.

CleanMachine1:
Does this warrant a page edit and new command tag?

pages/osx/readlink.md
Itai-Nelken: so should I change it to 2?

bl-ue: With two `to`s it's actually grammatically incorrect, or at least inappropriately redundant.

The two are both **100% equivalent**, the only difference is the position.  I like the first because I'm a native speaker so interesting twists appeal to me, but most people reading these pages won't feel the same way and will be confused.  Please, @Itai-Nelken, apply my latest [suggestion](https://github.com/tldr-pages/tldr/pull/6104#discussion_r646960844) above :)

bl-ue:
`new command` is good 👍🏻

CleanMachine1 (approved):
Once the suggestion above is added then I give my full approval

pages/osx/readlink.md
Itai-Nelken: when I said 2 I meant my 2nd suggestion that is the same as yours but I missed it

navarroaxel (approved):

pages/osx/readlink.md
```diff
@@ -0,0 +1,8 @@
+# readlink
+
+> Follow symlinks and get symlink information.
+> More information: <https://www.gnu.org/software/coreutils/readlink>.
+
+- Print the actual full path which the symlink points to:
```
SethFalco, line 6:
```suggestion
- Print the absolute path which the symlink points to:
```

Just my opinion but could we use the word absolute instead of full?
Paths are relative or absolute, while "full" is recognized/valid, it's less common.

https://en.wikipedia.org/wiki/Path_(computing)#Absolute_and_relative_paths

I also feel the use of the word "actual" is redundant here. Usually, when the word "actual" or "actually" like this, it's to add emphasis, but that's not needed here.

Any other opinions?

Itai-Nelken: sounds better to me, I'll replace it.

SethFalco (approved):

CleanMachine1:
4 reviews. Merging.

bl-ue:
Welcome to tldr-pages, @Itai-Nelken! ❤️ :)

PR #6110, opened 2021-06-09
history: add note for zsh

marchersimon:
This page also has an `es` and `pt_PT` translation, which also should be updated now, since we'll probably never see if there's a difference in a command description.

### DIFF
--- a/pages/common/history.md
+++ b/pages/common/history.md
@@ -7,7 +7,7 @@
 
 `history`
 
-- Display the last 20 commands:
+- Display the last 20 commands (in `zsh` it displays all commands starting from the 20th):
 
 `history {{20}}`
 


bl-ue (approved):

pages/common/history.md
```diff
@@ -7,7 +7,7 @@
 
 `history`
 
-- Display the last 20 commands:
+- Display the last 20 commands (in `zsh` it displays all commands starting from the 20th):
```
CleanMachine1, line 10:
```suggestion
- Display the last 20 commands (using `zsh` it displays all commands from history starting from the 20th):
```

PS: I would add either down or up, at the end of that sentence so whichever it is, feel free to edit my suggestion.

nicokosi: That's weird, it [deserves a "wat"](https://www.destroyallsoftware.com/talks/wat)! 🙀

I understand that:
* displays all commands from history starting from the 20th using `zsh`
* display the last 20 commands using all other shells

Am I correct?

CleanMachine1: I think you deserve a "wat" for your message, I don't know what you mean!

nicokosi: Sorry for my weird comment. :innocent:
By "it deserves a "wat", I meant that it is weird that `history -20` has different behavior depending on the shell (`zsh` / others).
And I wanted to share with you this old conference talk that is very funny! (I think) :performing_arts:

marchersimon: > I understand that it:
>
> * displays all commands from history starting from the 20th using zsh
> * displays the last 20 commands using all other shells
>
> Am I correct?

That's how it works on my machine, however it would be nice if someone could verify that / see how other shells behave.

CleanMachine1: for some reason it includes the history command.

sbrl: Generally we've avoided shell-specific examples previously, so if an example is indeed shell specific is there a way we can alter it to make it universal?

CleanMachine1: Doesn't seem to be a way

sbrl: Ah, ok

Let's add this note then, since that  seems to be the best solution to the issue.

@nicokosi Off topic, but that talk was indeed very funny - I remember watching it a few years ago :P

marchersimon: What sould we do about @ CleanMachine1's suggestion above?

sbrl: Hrm, I think the existing change in the PR is fine as-is?

CleanMachine1:
I don't like how it is worded, mainly ZSH being referred to as `it` however if everything thinks its fine, and its not bad then I will concede

CleanMachine1 (approved):

CleanMachine1:
I don't think concede is the right choice of word. But I am keeping it anyway because its a good one :smile:

marchersimon:
> I don't like how it is worded, mainly ZSH being referred to as `it` however if everyone thinks its fine, and its not bad then I will concede

The `it` is not referring to `zsh`. It is this kind of `it` you jut use anywhere, like `it rains` or `it is 16:45`

nicokosi (approved):
👌

sbrl (approved):

PR #6113, opened 2021-06-10
jello: add page

kellyjonbrazil:
add page for jello command (I am the author) https://github.com/kellyjonbrazil/jello

### DIFF
--- a/pages/common/jello.md
+++ b/pages/common/jello.md
@@ -29,4 +29,4 @@
 
 - Output the value of a given key to a string (and disable JSON output):
 
-`cat {{file.json}} | jello -r '"some text: " + _.{{key_name}}'`
+`cat {{file.json}} | jello -r '"{{some text}}: " + _.{{key_name}}'`
--- a/pages/common/jello.md
+++ b/pages/common/jello.md
@@ -3,7 +3,7 @@
 > A command-line JSON processor using Python syntax.
 > More information: <https://github.com/kellyjonbrazil/jello>.
 
-- Output JSON or JSON Lines STDIN data, in pretty-print format:
+- Pretty-print JSON or JSON-Lines data from stdin to stdout:
 
 `cat {{file.json}} | jello`
 
--- a/pages/common/jello.md
+++ b/pages/common/jello.md
@@ -30,14 +30,3 @@
 - Output the value of a given key to a string (and disable JSON output):
 
 `cat {{file.json}} | jello -r '"some text: " + _.{{key_name}}'`
-
-- Output raw string lines or JSON lines using a loop. (Useful for assigning to a BASH variable or array):
-
-```bash
-cat {{file.json}} | jello -lr '\
-result = []
-for i in _:
-    if i == "foo":
-        result.append(i)
-result'
-```
--- a/pages/common/jello.md
+++ b/pages/common/jello.md
@@ -1,0 +1,39 @@
+# jello
+
+> A command-line JSON processor using python syntax.
+> More information: <https://github.com/kellyjonbrazil/jello>.
+
+- Output JSON or JSON Lines STDIN data, in pretty-print format:
+
+`cat {{file.json}} | jello`
+
+- Output a grepable schema from JSON or JSON Lines STDIN data:
+
+`cat {{file.json}} | jello -s`
+
+- Output all elements from arrays (or all the values from objects) in JSON or JSON Lines STDIN data:
+
+`cat {{file.json}} | jello -l`
+
+- Output the first element in JSON or JSON Lines STDIN data:
+
+`cat {{file.json}} | jello _[0]`
+
+- Output the value of a given key of each element in JSON or JSON Lines STDIN data:
+
+`cat {{file.json}} | jello '[i.{{key_name}} for i in _]'`
+
+- Output the value of multiple keys as a new JSON object (assuming the input JSON has the keys `key_name` and `other_key_name`):
+
+`cat {{file.json}} | jello '{{{"my_new_key"}}: _.{{key_name}}, {{"my_other_key"}}: _.{{other_key_name}}}'`
+
+- Output raw string lines or JSON lines using a loop. (Useful for assigning to a BASH variable or array):
+
+```
+cat {{file.json}} | jello -lr '\
+result = []
+for i in _:
+    if i == "foo":
+        result.append(i)
+result'
+```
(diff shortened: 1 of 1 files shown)


pages/common/jello.md
```diff
@@ -0,0 +1,43 @@
...
+
+`cat {{file.json}} | jello '{{{"my_new_key"}}: _.{{key_name}}, {{"my_other_key"}}: _.{{other_key_name}}}'`
+
+- Output the value of a given key to a string (and disable JSON output):
+
+`cat {{file.json}} | jello -r '"some text: " + _.{{key_name}}'`
+
+- Output raw string lines or JSON lines using a loop. (Useful for assigning to a BASH variable or array):
+
+```bash
```
CleanMachine1, line 36: none of this is allowed.

It needs to be 1 line.

pages/common/jello.md
```diff
@@ -0,0 +1,32 @@
+# jello
+
+> A command-line JSON processor using python syntax.
```
marchersimon, line 3:
```suggestion
> A command-line JSON processor using Python syntax.
```

pages/common/jello.md
```diff
@@ -0,0 +1,32 @@
...
+> More information: <https://github.com/kellyjonbrazil/jello>.
+
+- Output JSON or JSON Lines STDIN data, in pretty-print format:
```
marchersimon, line 6:
```suggestion
- Output JSON or JSON-Lines STDIN data, in pretty-print format:
```
`JSON-Lines` is a file format right? If yes I think my suggestion makes it a bit readable. Also I think piping is not the same as reading from STDIN. Maybe we could even rewrite it like that:
```suggestion
- Pretty-print JSON or JSON-Lines input to stdout:
```
The same would apply to all other examples. But maybe let's wait for the other's opinions.

pages/common/jello.md
```diff
@@ -0,0 +1,32 @@
...
+
+`cat {{file.json}} | jello`
+
+- Output a grepable schema from JSON or JSON Lines STDIN data:
```
marchersimon, line 10: I think `grepable` is not a word. I can't really come up with an alternative, since I don't know what it's supposed to mean.

pages/common/jello.md
```diff
@@ -0,0 +1,32 @@
...
+
+`cat {{file.json}} | jello _[0]`
+
+- Output the value of a given key of each element in JSON or JSON Lines STDIN data:
+
+`cat {{file.json}} | jello '[i.{{key_name}} for i in _]'`
+
+- Output the value of multiple keys as a new JSON object (assuming the input JSON has the keys `key_name` and `other_key_name`):
+
+`cat {{file.json}} | jello '{{{"my_new_key"}}: _.{{key_name}}, {{"my_other_key"}}: _.{{other_key_name}}}'`
```
marchersimon, line 28:
```suggestion
`cat {{file.json}} | jello '"{{my_new_key}}": _.{{key_name}}, "{{my_other_key}}": _.{{other_key_name}}}'`
```
We don't really count the `" "` to be part of the argument here.

pages/common/jello.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Output the value of a given key to a string (and disable JSON output):
+
+`cat {{file.json}} | jello -r '"some text: " + _.{{key_name}}'`
```
marchersimon, line 32:
```suggestion
`cat {{file.json}} | jello -r '"{{some text}}: " + _.{{key_name}}'`
```
Since `some text` is something the user has to input it's a variable argument.

marchersimon (changes requested):

pages/common/jello.md
bl-ue: Yeah, it's really unfortunate that they're only one-liners, but... 🤷🏻‍♂️

pages/common/jello.md
kellyjonbrazil: Maybe "Pretty-print JSON or JSON-Lines data from stdin to stdout"?

pages/common/jello.md
kellyjonbrazil: Yeah it is a sort of made-up term. Another way to put it would be grep-able or greppable (https://github.com/tomnomnom/gron).

The idea is that you cannot easily use the `grep` command on JSON data, so this schema view allows you to use `grep` to find what you need more easily in the JSON.

Maybe something like: "Output a schema from JSON or JSON Lines data from stdin to stdout (useful for `grep`):

pages/common/jello.md
kellyjonbrazil: Ah, that makes sense. I still need that extra `{` in the very beginning, though.

pages/common/jello.md
bl-ue: Sounds good to me!
```suggestion
- Pretty-print JSON or JSON-Lines data from stdin to stdout:
```

pages/common/jello.md
kellyjonbrazil: I changed the language. Let me know if that doesn't work.

pages/common/jello.md
```diff
@@ -0,0 +1,32 @@
...
+
+`cat {{file.json}} | jello -l`
+
+- Output the first element in JSON or JSON-Lines data from stdin to stdout:
+
+`cat {{file.json}} | jello _[0]`
+
+- Output the value of a given key of each element in JSON or JSON-Lines data from stdin to stdout:
+
+`cat {{file.json}} | jello '[i.{{key_name}} for i in _]'`
```
bl-ue, line 24: Ah, you can do Python list comprehension. Cool! 🚀

bl-ue (approved):
LGTM!

pages/common/jello.md
kellyjonbrazil: Yep, basically anything you can do in Python. `jello` basically removes the boilerplate code you would need to ingest and output the filtered JSON in different formats. It's very similar to `jq`, which has its own quirky syntax, so I wrote `jello` to work very much like `jq` but with Python syntax.

marchersimon (approved):
Looks good, thanks @kellyjonbrazil

pages/common/jello.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Output the value of multiple keys as a new JSON object (assuming the input JSON has the keys `key_name` and `other_key_name`):
+
+`cat {{file.json}} | jello '{"{{my_new_key}}": _.{{key_name}}, "{{my_other_key}}": _.{{other_key_name}}}'`
```
marchersimon, line 28: @bl-ue do you think the extra curly brace will cause any render errors?

bl-ue: For the node client, it renders beautifully!
Screen Shot 2021-06-11 at 4 23 21 PM

PR #6126, opened 2021-06-14
git-missing: add page

CleanMachine1:

For #5137

### DIFF
--- a/pages/common/git-missing.md
+++ b/pages/common/git-missing.md
@@ -1,0 +1,13 @@
+# git missing 
+
+> Show commits which aren't shared between 2 branches.
+> Part of `git-extras`.
+> More information <https://github.com/tj/git-extras/blob/master/Commands.md#git-missing>.
+
+- Show commits which aren't shared between the current checked-out branch and a specified branch:
+
+`git missing {{branch}}`
+
+- Show commits which aren't shared between 2 specified branches:
+
+`git missing {{branch_1}} {{branch_2}}`
--- a/pages/common/git-missing.md
+++ b/pages/common/git-missing.md
@@ -1,8 +1,8 @@
-# git missing 
+# git missing
 
 > Show commits which aren't shared between 2 branches.
 > Part of `git-extras`.
-> More information <https://github.com/tj/git-extras/blob/master/Commands.md#git-missing>.
+> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-missing>.
 
 - Show commits which aren't shared between the current checked-out branch and a specified branch:
 
--- a/pages/common/git-missing.md
+++ b/pages/common/git-missing.md
@@ -8,6 +8,6 @@
 
 `git missing {{branch}}`
 
-- Show commits which aren't shared between 2 specified branches:
+- Show commits which aren't shared between two branches:
 
 `git missing {{branch_1}} {{branch_2}}`
--- a/pages/common/git-missing.md
+++ b/pages/common/git-missing.md
@@ -1,6 +1,6 @@
 # git missing
 
-> Show commits which aren't shared between 2 branches.
+> Show commits which aren't shared between two branches.
 > Part of `git-extras`.
 > More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-missing>.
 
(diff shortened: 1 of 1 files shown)


pages/common/git-missing.md
```diff
@@ -0,0 +1,13 @@
+# git missing
+
+> Show commits which aren't shared between 2 branches.
```
bl-ue, line 3:
```suggestion
> Show commits which aren't shared between two branches.
```

pages/common/git-missing.md
```diff
@@ -0,0 +1,13 @@
...
+> Part of `git-extras`.
+> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-missing>.
+
+- Show commits which aren't shared between the current checked-out branch and a specified branch:
```
bl-ue, line 7:
```suggestion
- Show commits which aren't shared between the currently checked-out branch and another branch:
```

pages/common/git-missing.md
```diff
@@ -0,0 +1,13 @@
...
+
+`git missing {{branch}}`
+
+- Show commits which aren't shared between 2 specified branches:
```
bl-ue, line 11:
```suggestion
- Show commits which aren't shared between two branches:
```

pages/common/git-missing.md
```diff
@@ -0,0 +1,13 @@
+# git missing
+
+> Show commits which aren't shared between two branches.
```
mfrw, line 3:
```suggestion
> Show commits which are not shared between two branches.
```

`aren't` does not feel formal

pages/common/git-missing.md
```diff
@@ -0,0 +1,13 @@
...
+> Part of `git-extras`.
+> More information: <https://github.com/tj/git-extras/blob/master/Commands.md#git-missing>.
+
+- Show commits which aren't shared between the currently checked-out branch and another branch:
```
mfrw, line 7:
```suggestion
- Show commits which are not shared between the currently checked-out branch and another branch:
```

pages/common/git-missing.md
```diff
@@ -0,0 +1,13 @@
...
+
+`git missing {{branch}}`
+
+- Show commits which aren't shared between two branches:
```
mfrw, line 11:
```suggestion
- Show commits which are not shared between two branches:
```

mfrw (approved):
Minor cosmetic changes :)

pages/common/git-missing.md
CleanMachine1: aren't is more than acceptable, in fact I would argue `are not` is incorrect.
Also although its only a few, `are not` is used 8 files while `aren't` has 11 files

mfrw: Should we be consistent and change are not to aren't in those pages as well ?

CleanMachine1: Whats wrong with aren't, its better, reads nicer

bl-ue: It's merely a matter of preference.  I personally prefer `are not` here because of the formality (`aren't` sounds just a tad too long and informal to me).

>in fact I would argue `are not` is incorrect.

_That_ is incorrect — `aren't` is the contraction of `are` and `not`; thus, `are not` and `aren't` are identical.

bl-ue: >Should we be consistent and change are not to aren't in those pages as well ?

It's not really something that ought to be standardized.  I don't think we should do that.

>Also although its only a few, are not is used 8 files while aren't has 11 files

Those numbers are too small to be meaningful :)

CleanMachine1: aren't definitely reads shorter than are not

mfrw: Yes I agree, it's merely preference. I think this is a nitpick.
Personally this felt informal, but looks like more files have 'aren't'.

bl-ue: >aren't definitely reads shorter than are not

You're right — the was I was pronouncing it in my head sound longer (`arnt` vs `arent`) 🙂

bl-ue: >Personally this felt informal, but looks like more files have 'aren't'.

As I said before, those numbers are _so_ small (and similar) that it's not a good idea to base decisions of them.

CleanMachine1: so what are we doing here

bl-ue: Let's keep it as `aren't`, because I don't really see why we should change it.

owenvoke (approved):

bl-ue (approved):

PR #6128, opened 2021-06-15
vgscan, vgchange: add page

aherst:

### DIFF
--- a/pages/linux/vgchange.md
+++ b/pages/linux/vgchange.md
@@ -4,10 +4,10 @@
 > See also: `lvm`.
 > More information: <https://manned.org/vgchange>.
 
-- Activate or deactivate logical volumes in all volume groups:
+- Change the activation status of logical volumes in all volume groups:
 
 `sudo vgchange --activate {{y|n}}`
 
-- Activate or deactivate logical volumes in the specified volume group (determine with `vgscan`):
+- Change the activation status of logical volumes in the specified volume group (determine with `vgscan`):
 
-`sudo vgchange --activate {{y|n}} {{vg}}}`
+`sudo vgchange --activate {{y|n}} {{volume_group}}}`
--- a/pages/linux/vgchange.md
+++ b/pages/linux/vgchange.md
@@ -1,0 +1,13 @@
+# vgchange
+
+> Change the attributes of a Logical Volume Manager (LVM) volume group.
+> See also: `lvm`.
+> More information: <https://manned.org/vgchange>.
+
+- Activate or deactivate logical volumes in all volume groups:
+
+`sudo vgchange --activate {{y|n}}`
+
+- Activate or deactivate logical volumes in the specified volume group (determine with `vgscan`):
+
+`sudo vgchange --activate {{y|n}} {{vg}}}`
--- a/pages/linux/vgscan.md
+++ b/pages/linux/vgscan.md
@@ -1,0 +1,13 @@
+# vgscan
+
+> Scan for volume groups on all supported Logical Volume Manager (LVM) block devices.
+> See also `lvm`, `vgchange`.
+> More information: <https://manned.org/vgscan>.
+
+- Scan for volume groups and print information about each group found:
+
+`sudo vgscan`
+
+- Scan for volume groups and add or remove the special files in `/dev` needed to access the logical volumes in a group:
+
+`sudo vgscan --mknodes`
--- a/pages/linux/vgscan.md
+++ b/pages/linux/vgscan.md
@@ -8,6 +8,6 @@
 
 `sudo vgscan`
 
-- Scan for volume groups and add or remove the special files in `/dev` needed to access the logical volumes in a group:
+- Scan for volume groups and add the special files in `/dev`, if they don't already exist, needed to access the logical volumes in the found groups :
 
 `sudo vgscan --mknodes`


CleanMachine1:
The description of mass changes contains, multiple changes

and multiple refers to more than 1, this PR contains 2

pages/linux/vgchange.md
```diff
@@ -0,0 +1,13 @@
+# vgchange
+
+> Change the attributes of a Logical Volume Manager (LVM) volume group.
+> See also: `lvm`.
+> More information: <https://manned.org/vgchange>.
+
+- Activate or deactivate logical volumes in all volume groups:
```
CleanMachine1, line 7: I don't like the usage of the activation or deactivate.

Maybe something similar to `change the status of`

pages/linux/vgscan.md
```diff
@@ -0,0 +1,13 @@
+# vgscan
+
+> Scan for volume groups on all supported Logical Volume Manager (LVM) block devices.
+> See also `lvm`, `vgchange`.
+> More information: <https://manned.org/vgscan>.
+
+- Scan for volume groups and print information about each group found:
+
+`sudo vgscan`
```
CleanMachine1, line 9:
```suggestion
`vgscan`
```

CleanMachine1: should sudo be used, even if its needed by the command

aherst: @CleanMachine1, I was following point 2 in the guidelines:

`
When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness. For example, commands that require sudo should include it directly in the examples.
`

pages/linux/vgchange.md
aherst: @CleanMachine1, how about `change the activation status of`?

CleanMachine1: Thats better. Its fine as it is but I am just being picky

pages/linux/vgscan.md
bl-ue: If numbers is what you want @CleanMachine1, try searching 😆
```console
❯ rg '^`sudo ' | wc -l
     849

Comment (CleanMachine1) [reply pages/linux/vgscan.md:9]: I did actually search some results using VSCode since it has a nice little word search thing in its replace tool

Comment (CleanMachine1) [reply pages/linux/vgscan.md:9]: I was going to mention it but other things are inconsistent for example

 `tldr dd`

dd needs root to copy from things like block devices.

Comment (bl-ue) [reply pages/linux/vgscan.md:9]: `dd` could and should be updated.

File: pages/linux/vgscan.md [hunk 3]
```diff
@@ -0,0 +1,13 @@
...
+
+- Scan for volume groups and add or remove the special files in `/dev` needed to access the logical volumes in a group:
+
+`sudo vgscan --mknodes`
```
marchersimon, line 13: So does it add or remove files? Or does that depend on the conext? And since you said `in a group`, does that mean it only does this for one group? If yes, wouldn't you have to specify its name?

marchersimon (changes requested):

pages/linux/vgchange.md
aherst: @CleanMachine1, I think your phrasing is better because it relates back to the synopsis wording so I've made the change.

pages/linux/vgchange.md
```diff
@@ -0,0 +1,13 @@
...
+
+`sudo vgchange --activate {{y|n}}`
+
+- Activate or deactivate logical volumes in the specified volume group (determine with `vgscan`):
+
+`sudo vgchange --activate {{y|n}} {{vg}}}`
```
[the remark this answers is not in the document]
aherst: @marchersimon, change applied.

pages/linux/vgscan.md
```diff
@@ -0,0 +1,13 @@
+# vgscan
+
+> Scan for volume groups on all supported Logical Volume Manager (LVM) block devices.
+> See also `lvm`, `vgchange`.
```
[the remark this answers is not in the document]
aherst: I've cross referenced vgchange because together they address the use case:
1. add a disk formatted for LVM to a running system that doesn't get identified by udev
2. run vgscan to determine the name of the volume_group(s) it contains
3. run vgchange to activate one or more volume group so the logical volume can be accessed

pages/linux/vgscan.md
aherst: It is supposed to do both depending on context. I've only verified the addition. How about if I change the example description to:

```suggestion
- Scan for volume groups and add, if they don't already exist, the special files in `/dev` needed to access the logical volumes in the found groups:

Comment (marchersimon) [reply pages/linux/vgscan.md:13]: Sounds better, however I'm not sure about the grammar here. @bl-ue?

Comment (bl-ue) [reply pages/linux/vgscan.md:13]: Yep that's perfectly good grammar, but it might be confusing to non-native speakers. How about moving `if they don't already exist` to the end, @aherst?
```suggestion
- Scan for volume groups and add the special files in `/dev` needed to access the logical volumes in the found groups, if they don't already exist:

`sudo vgscan --mknodes`
```

marchersimon: Much better

aherst: I want to make sure it's clear that the clause applies to "files" and not to "groups". How about:

```suggestion
- Scan for volume groups and add the special files in `/dev`, if they don't already exist, needed to access the logical volumes in the found groups :

bl-ue: sgtm

pages/linux/vgscan.md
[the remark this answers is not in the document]
aherst: Is there anything else for me to do here?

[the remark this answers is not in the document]
marchersimon: No, thanks. I just forgot to resolve the conversation.

marchersimon (approved):
Thanks @aherst

sbrl (approved):
Thanks, @aherst!

PR #6129, opened 2021-06-15
partprobe: add page

aherst:


bl-ue (approved):
Looks good to me, but perhaps "notify" might be a slightly more accurate verb instead of "inform" here?

aherst:
@bl-ue, suggestion applied.

aherst:
@CleanMachine1, yes it can. I learned how to use partprobe so long ago I don't think it had long options at the time. Change applied.

CleanMachine1 (approved):

Issue #5474, opened 2021-03-19
Let's document: go


bl-ue:
- [x] [`go bug`](https://golang.org/cmd/go/#hdr-Start_a_bug_report) → #4003
- [x] [`go build`](https://golang.org/cmd/go/#hdr-Compile_packages_and_dependencies) → #4003
- [x] [`go clean`](https://golang.org/cmd/go/#hdr-Remove_object_files_and_cached_files) → #4003
- [x] [`go doc`](https://golang.org/cmd/go/#hdr-Show_documentation_for_package_or_symbol) → #4003
- [x] [`go env`](https://golang.org/cmd/go/#hdr-Print_Go_environment_information) → #4003
- [x] [`go fix`](https://golang.org/cmd/go/#hdr-Update_packages_to_use_new_APIs) → #4003
- [x] [`go fmt`](https://golang.org/cmd/go/#hdr-Gofmt__reformat__package_sources) → #6703 
- [x] [`go generate`](https://golang.org/cmd/go/#hdr-Generate_Go_files_by_processing_source) → #4003
- [x] [`go get`](https://golang.org/cmd/go/#hdr-Add_dependencies_to_current_module_and_install_them) → #6491 
- [ ] [`go install`](https://golang.org/cmd/go/#hdr-Compile_and_install_packages_and_dependencies) → #7111 
- [x] [`go list`](https://golang.org/cmd/go/#hdr-List_packages_or_modules) → #4003
- [x] [`go mod`](https://golang.org/cmd/go/#hdr-Module_maintenance) → #4003
- [x] [`go run`](https://golang.org/cmd/go/#hdr-Compile_and_run_Go_program) → #6650 
- [x] [`go test`](https://golang.org/cmd/go/#hdr-Test_packages) → #6130
- [x] [`go tool`](https://golang.org/cmd/go/#hdr-Run_specified_go_tool) → #6998 #7070 
- [x] [`go version`](https://golang.org/cmd/go/#hdr-Print_Go_version) → #4003
- [x] [`go vet`](https://golang.org/cmd/go/#hdr-Report_likely_mistakes_in_packages) → #6832

PR #6130, opened 2021-06-16
go-test: add page

agstrc:

Ref: #5474

### DIFF
--- a/pages/common/go-test.md
+++ b/pages/common/go-test.md
@@ -1,7 +1,7 @@
 # go test
 
 > Tests Go packages (files have to end with `_test.go`).
-> More information: `https://golang.org/cmd/go/#hdr-Testing_flags`
+> More information: <https://golang.org/cmd/go/#hdr-Testing_flags>
 
 - Test the package found at the current directory:
 
@@ -20,4 +20,5 @@
 `go test -v -bench .`
 
 - [v]erbosely test the package at the current directory and run all benchmarks for 50 seconds:
+
 `go test -v -bench . -benchtime 50s`
--- a/pages/common/go-test.md
+++ b/pages/common/go-test.md
@@ -11,7 +11,7 @@
 
 `go test -v`
 
-- Test the packages in the current directory and all subdirectories:
+- Test the packages in the current directory and all subdirectories (note the `...`):
 
 `go test -v ./...`
 
@@ -21,4 +21,4 @@
 
 - Test the package in the current directory and run all benchmarks for 50 seconds:
 
-`go test -v -bench . -benchtime 50s`
+`go test -v -bench . -benchtime {{50s}}`
--- a/pages/common/go-test.md
+++ b/pages/common/go-test.md
@@ -1,24 +1,24 @@
 # go test
 
 > Tests Go packages (files have to end with `_test.go`).
-> More information: <https://golang.org/cmd/go/#hdr-Testing_flags>
+> More information: <https://golang.org/cmd/go/#hdr-Testing_flags>.
 
-- Test the package found at the current directory:
+- Test the package found in the current directory:
 
 `go test`
 
-- [v]erbosely test the package at the current directory:
+- [v]erbosely test the package in the current directory:
 
 `go test -v`
 
-- [v]erbosely test the packages at the current directory and all subdirectories:
+- [v]erbosely test the packages in the current directory and all subdirectories:
 
 `go test -v ./...`
 
-- [v]erbosely test the package at the current directory and run all benchmarks:
+- [v]erbosely test the package in the current directory and run all benchmarks:
 
 `go test -v -bench .`
 
-- [v]erbosely test the package at the current directory and run all benchmarks for 50 seconds:
+- [v]erbosely test the package in the current directory and run all benchmarks for 50 seconds:
 
 `go test -v -bench . -benchtime 50s`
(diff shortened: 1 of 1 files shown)


bl-ue (changes requested):
Will you please change `at the current directory` to `in the current directory`? I think it sounds slightly better.

agstrc:
> Will you please change `at the current directory` to `in the current directory`? I think it sounds slightly better.

Will do.

pages/common/go-test.md
```diff
@@ -0,0 +1,24 @@
...
+
+- [v]erbosely test the packages in the current directory and all subdirectories:
+
+`go test -v ./...`
+
+- [v]erbosely test the package in the current directory and run all benchmarks:
+
+`go test -v -bench .`
+
+- [v]erbosely test the package in the current directory and run all benchmarks for 50 seconds:
```
bl-ue, line 22: Since `-v` is in all of the examples, why not just omit explaining it?  I'd like to keep using `[` as the first character at a minimum.
```suggestion
- Test the packages in the current directory and all subdirectories:

`go test -v ./...`

- Test the package in the current directory and run all benchmarks:

`go test -v -bench .`

- Test the package in the current directory and run all benchmarks for 50 seconds:
```

pages/common/go-test.md
```diff
@@ -0,0 +1,24 @@
...
+
+`go test`
+
+- [v]erbosely test the package in the current directory:
+
+`go test -v`
+
+- Test the packages in the current directory and all subdirectories:
+
+`go test -v ./...`
```
bl-ue, line 16: This example is confusing a bit — is this example basically equivalent (and true)?
```suggestion
- Test the packages in the specified directory and all subdirectories:

`go test -v {{path/to/directory}}`
```

pages/common/go-test.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Test the package in the current directory and run all benchmarks:
+
+`go test -v -bench .`
```
bl-ue, line 20:
```suggestion
`go test -v -bench {{.}}`
```

pages/common/go-test.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Test the package in the current directory and run all benchmarks for 50 seconds:
+
+`go test -v -bench . -benchtime 50s`
```
bl-ue, line 24:
```suggestion
`go test -v -bench {{.}} -benchtime {{50s}}`
```

bl-ue (commented):
A few comments, otherwise LGTM!

pages/common/go-test.md
agstrc: The `go test` tool always takes into consideration the current directory, hence why in the first example no argument was required. The `./` part of the argument specifies that the package in the current directory must be tested, and the `...` part relates to all the other subdirectories, recursively.

pages/common/go-test.md
agstrc: I suppose this reply relates to the the previous comment/suggestion, but the `.` would actually be interpreted as a regular expression if changed. I never mentioned this in the examples as I assumed this would break the 5th guideline mentioned in the contributing guidelines, that being
> Focus on details specific to the command, and avoid explaining general UNIX concepts that could apply to any command (ex: relative/absolute paths, glob patterns/wildcards, special character escaping...).

So I just attempted to keep the examples simple in that regard.

If that really is the case, per my understanding the curly brackets wouldn't be necessary around the `.` character.

pages/common/go-test.md
agstrc: Clarifying, the given examples would not be equivalent. If you pass a specific directory path to the `go test` command, it'll only test the packages found at that directory, therefore ignoring any subdirectory.

bl-ue: Oh, I see. So `...` isn't just a placeholder value, you'd actually pass it to the command?

agstrc: Yes, that is correct.

pages/common/go-test.md
bl-ue: You're right, it's good to avoid introducing general UNIX concepts; however, regular expressions is really a general computer science concept, and use on lots and lots of other pages, so I think it's okay (and advisable) to clarify that here. 

>If that really is the case, per my understanding the curly brackets wouldn't be necessary around the `.` character.

We always put the token syntax (`{{` and `}}`) around user-mutable parameters/arguments.  If the `.` should always be a `.` to make the example correct, then let's leave it as-is.  Otherwise, if it can be changed without making the description inaccurate, put curly braces around it.  Assuming the former:
```suggestion
`go test -v -bench . -benchtime {{50s}}`
```

pages/common/go-test.md
bl-ue: Then maybe
```suggestion
- Test the packages in the current directory and all subdirectories (note the `...`):

`go test -v ./...`
```

agstrc:
@bl-ue I kept the `.` argument as-is in the examples above due to the fact that, for the case described in the examples (run **all** benchmarks), the argument may not be changed. Regarding the regular expressions, I believe that running benchmarks which match a given regular expression would be a slightly more advanced use case, which can be better understood upon accessing the **More information** link.  
If you still believe any changes should be made, let me know.

bl-ue (approved):

mfrw (approved):
LGTM

PR #6136, opened 2021-06-17
README: add python client in How do I use it section

CleanMachine1:


README.md
```diff
@@ -53,6 +53,11 @@ which is supported by the tldr-pages project maintainers:
 
     npm install -g tldr
 
+Or you can use the just as easy to install [python3 client](https://github.com/tldr-pages/tldr-python-client) which is also supported by the tldr-pages project maintainers, however it is available on pip3 which is more likely to be on your computer:
```
sbrl, line 56: This is quite subjective. Specifically `however it is available on pip3 which is more likely to be on your computer` - that all depends on people's setups. Can we make this more objective perhaps?

CleanMachine1: Yes it is subjective, however most linux distributions have it, and Arch has its community repo  package.

And who has NPM and not python PIP.

I imagine that PIP has a lot more downloads compared to NPM

CleanMachine1: I'll find a solution

CleanMachine1: Or you can use the just as easy to install [python3 client](https://github.com/tldr-pages/tldr-python-client) which is also supported by the tldr-pages project maintainers, however it is available on pip which you may have installed over npm

bl-ue: How about this.
```suggestion
You can also use the [official Python client](https://github.com/tldr-pages/tldr-python-client), which is supported by the tldr-pages project maintainers as well.
```

CleanMachine1: Yes that works completely however I feel that the average newb, especially Linux users might have not have NPM but might have pip preinstalled.

And from what I have seen, the python client from pip is better than the haskell client in multiple ways for example customization and command passing, eg `tldr git what-changed` doesn't work

README.md
```diff
@@ -53,6 +53,11 @@ which is supported by the tldr-pages project maintainers:
...
+
+    pip3 install tldr
+
+
```
navarroaxel, line 60: Double enter here:
```suggestion
```

CleanMachine1: I did see that, just oppressed it for an unknown reason

README.md
marchersimon: How about
```suggestion
You can also use the [official Python client](https://github.com/tldr-pages/tldr-python-client), which is customizable and has lots of features and can be installed via `pip3`.
```
I think `which is supported by the tldr-pages project maintainers` is not nessecary, since that's already implied by `official`.

CleanMachine1: Thats fine, however if one has any other good ideas I'll leave this open.

sbrl: Hrm. What about this?
```suggestion
You can also use the [official Python client](https://github.com/tldr-pages/tldr-python-client), which can be installed via `pip3`.
```

marchersimon: I think it's good to know the difference between the python client and the others and `customizable and has lots of features` is objective enough to include it.

sbrl: How does one define `lots of features` though? Isn't the Node.js client customisable too with a config file?

marchersimon: Hm, that wasn't as objective as I thought. How about 
```suggestion
You can also use the fully-featued [official Python client](https://github.com/tldr-pages/tldr-python-client), which can be installed via `pip3`.
```
to at least indicate that it doesn't lack any features?

sbrl (approved):
Thanks, @CleanMachine1!

README.md
```diff
@@ -53,6 +53,10 @@ which is supported by the tldr-pages project maintainers:
 
     npm install -g tldr
 
+You can also use the fully-featued [official Python client](https://github.com/tldr-pages/tldr-python-client), which can be installed via `pip3`.
```
marchersimon, line 56: What do you think?
```suggestion
You can also use the fully-featued, [official Python client](https://github.com/tldr-pages/tldr-python-client), which can be installed via `pip3`.
```

nicokosi:
```suggestion
You can also use the fully-featured [official Python client](https://github.com/tldr-pages/tldr-python-client) which can be installed via `pip3`.
```

CleanMachine1: maybe something like `You can also use, the ....`

marchersimon: @nicokosi nice catch. I also agree on removing the comma before `which`. But what do you think about the comma in `fully-featured, official Python client`? 

@CleanMachine1 I don't think that's correct in any way.

CleanMachine1: hmm, reading it again it doesn't make much sense.

sbrl: @nicokosi's suggestion makes the most grammatical sense to me.

marchersimon (approved):
Thanhs @CleanMachine1

navarroaxel (approved):

CleanMachine1:
Can we merge?

marchersimon:
I think so

PR #5946, opened 2021-05-11
git-verify-commit: add page

CleanMachine1:
- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

For #3953

PR #6139, opened 2021-06-17
git-verify-tag: add page

CleanMachine1:

For #3953

### DIFF
--- a/pages/common/git-verify-tag.md
+++ b/pages/common/git-verify-tag.md
@@ -1,0 +1,17 @@
+# git verify-tag
+
+> Check for GPG verification of tags.
+> If tag isn't signed with -s when made, you will get an error regardless of signing.
+> More information: <https://git-scm.com/docs/git-verify-tag>.
+
+- Check tags for a GPG signature:
+
+`git verify-tag {{tag1 optional_tag2 ...}}`
+
+- Check tags for a GPG signature and show details of each tag:
+
+`git verify-tag {{tag1 optional_tag2 ...}} --verbose`
+
+- Check tags for a GPG signature and print the raw details:
+
+`git verify-tag {{tag1 optional_tag2 ...}} --raw`
--- a/pages/common/git-verify-tag.md
+++ b/pages/common/git-verify-tag.md
@@ -8,7 +8,7 @@
 
 `git verify-tag {{tag1 optional_tag2 ...}}`
 
-- Check tags for a GPG signature and show details of each tag:
+- Check tags for a GPG signature and show details for each tag:
 
 `git verify-tag {{tag1 optional_tag2 ...}} --verbose`
 
--- a/pages/common/git-verify-tag.md
+++ b/pages/common/git-verify-tag.md
@@ -1,7 +1,7 @@
 # git verify-tag
 
 > Check for GPG verification of tags.
-> If a tag wasn't manually signed, an error will occur.
+> If a tag wasn't signed, an error will occur.
 > More information: <https://git-scm.com/docs/git-verify-tag>.
 
 - Check tags for a GPG signature:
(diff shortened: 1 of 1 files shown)


CleanMachine1:
Make sure to take a look at #5946

CleanMachine1:
Also the Gitter section, I have left some more comments

pages/common/git-verify-tag.md
```diff
@@ -0,0 +1,17 @@
+# git verify-tag
+
+> Check for GPG verification of tags.
+> If tag isn't signed with -s when made, you will get an error regardless of signing.
```
marchersimon, line 4: How about a shorter version:
```suggestion
> If a tag wasn't signed, an error will occur.
```

marchersimon, line 4: How about a shorter version:
```suggestion
> If a tag wasn't signed, an error will occur.
```

pages/common/git-verify-tag.md
```diff
@@ -0,0 +1,17 @@
...
+> More information: <https://git-scm.com/docs/git-verify-tag>.
+
+- Check tags for a GPG signature:
+
+`git verify-tag {{tag1 optional_tag2 ...}}`
+
+- Check tags for a GPG signature and show details of each tag:
```
marchersimon, line 11:
```suggestion
- Check tags for a GPG signature and show details for each tag:
```

pages/common/git-verify-tag.md
CleanMachine1: signed manually would be better since it seems that regardless of it being signed, the -s flag seems to flag it up for git, almost as if it is treated different

CleanMachine1:
Thought it was ironic that the verify-tag page wasn't signed, since my key was removed. So I rebased and squashed and signed it

pages/common/git-verify-tag.md
navarroaxel: `with -s` without the full command doesn't have enough sense, also I think should be a `gitconfig` setting to auto sign tags.

bl-ue: I prefer @marchersimon's suggestion — less complexity means less likelyhood of misunderstanding the conversation.

bl-ue:
(Sorry, I just wanted to test the locking feature 😄 — turns out even collaborators can't react/unreact 🤔)

CleanMachine1:
That should probably be changed to give the option

sbrl:
The idea there @CleanMachine1 is that by saying `tldr-pages locked`, it avoids publicly naming the person who locked the thread. This is important in heated situations because otherewise the person doing the locking may be targeted on other platforms and the conversation might just continue anyway. That way they allow everyone some cooling off time.

Thankfully we've never had a need to use that feature haha

sbrl (approved):
Looks ok to me after @marchersimon's comment is addressed.

bl-ue:
>the person doing the locking may be targeted on other platforms

Often though, the one doing the locking says "I'm going to lock this now" or something similar before doing it 😉

bl-ue (approved):

pages/common/git-verify-tag.md
CleanMachine1: @navarroaxel it does sign the tags with `	gpgsign = true` but not flagged for some reason

CleanMachine1:
```suggestion
> If a tag wasn't manually signed, an error will occur.
```

navarroaxel: @CleanMachine1 `manually` or `automatic` is not important here, @marchersimon suggestion is good enough.

CleanMachine1: I disagree but others agree, so OK

marchersimon (approved):
👍🏽

CleanMachine1:
- Bl-ue's approval makes it 3

PR #6144, opened 2021-06-19
mklost+found: add page

CleanMachine1:
- [x] The page description includes a link to documentation or a homepage (if applicable)

Never seen or used the command, and it maybe could be an arch exclusive since I have never seen it on Debian based.

Unaware on how useful it really is since fsck will make one every boot anyway

### DIFF
--- a/pages/linux/mklost+found.md
+++ b/pages/linux/mklost+found.md
@@ -3,6 +3,6 @@
 > Create a lost+found directory.
 > More information: <https://manned.org/mklost+found>.
 
-- Create a lost+found directory:
+- Create a `lost+found` directory in the current directory:
 
 `mklost+found`
--- a/pages/linux/mklost+found.md
+++ b/pages/linux/mklost+found.md
@@ -1,0 +1,8 @@
+# mklost+found
+
+> Create a lost+found directory:
+> More information: <https://manned.org/mklost+found.8>.
+
+- Create a lost+found directory:
+
+`mklost+found`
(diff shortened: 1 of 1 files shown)


CleanMachine1:
Fun fact this is the 2nd page ever to have a + in the name

bl-ue (approved):

SethFalco (approved):

pages/linux/mklost+found.md
```diff
@@ -0,0 +1,8 @@
+# mklost+found
+
+> Create a lost+found directory.
+> More information: <https://manned.org/mklost+found.8>.
```
navarroaxel, line 4:
```suggestion
> More information: <https://manned.org/mklost+found>.
```

pages/linux/mklost+found.md
```diff
@@ -0,0 +1,8 @@
+# mklost+found
+
+> Create a lost+found directory.
+> More information: <https://manned.org/mklost+found>.
+
+- Create a lost+found directory:
```
marchersimon, line 6: @CleanMachine1 as I understand it, `fsck` only creates a `lost+found` folder in the root directory, but this command creates one in the current directory. I don't know what that actually means, but I think we should include that.
```suggestion
- Create a `lost+found` directory in the current directory:
```
(Also I think the backticks make sense here)

CleanMachine1: Ok, just for reference, the lost+found is what fsck outputs, which would normally be corrupted, damaged files which is normally probably junk data.

However I've never used it, I just saw it when making my file system when tabbing

navarroaxel (approved):

PR #6146, opened 2021-06-20
CONTRIBUTING: remove emoticon

bl-ue:
Open for suggestions!

I'm thinking to remove this because it doesn't really feel right any more.  It's also a bit personal, which seems inappropriate because the document is the voice of the whole community (or at least maintenance team).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,7 +44,7 @@ Here are a few guidelines to get started:
 These are all guidelines, not strict rules.
 Use proper judgement, keeping simplicity and user-friendliness as the top priorities.
 
-When in doubt, have a look at a few existing pages :).
+When in doubt, have a look at a few existing pages.
 
 ## Markdown format
 


Stirfry70:
[@]()

Stirfry70:
@bl-ue

CONTRIBUTING.md
```diff
@@ -44,7 +44,7 @@ Here are a few guidelines to get started:
 These are all guidelines, not strict rules.
 Use proper judgement, keeping simplicity and user-friendliness as the top priorities.
 
-When in doubt, have a look at a few existing pages :).
+When in doubt, have a look at a few existing pages.
```
mfrw, line 47: I have a mixed feeling about it. Although, it does feel informal but at the same time it feels more welcoming in my opinion.

bl-ue: You're right — but is that the right place?  Maybe we should adjust wording elsewhere to mimic the feeling.

marchersimon (approved):
👍🏽

CleanMachine1:
Maybe just a simple vote yes or no.

Dislike for no 
Like for yes

(On the first message/title message)

Other than discussing reasons, this should be simply voted for

CONTRIBUTING.md
CleanMachine1: I have mixed opinions, for: its inviting and as a contributor, it removes a takes the edge off contributing to a 33k starred project

against: It seems a bit out of place.

bl-ue:
In tldr-pages, we're supposed to try to avoid voting, and instead discuss things together until we all agree.  This might be to simple though 🤔

bl-ue:
cc @waldyrious

sbrl (changes requested):
I don't agree. While it does speak for the whole community, an emoji like this helps to convey a feeling of friendliness. For this reason, I think it's important to keep it.

bl-ue:
I'm just not really a fan of that type of informality in _formal(-ish)_ documents. For personal messages, they're definitely great, but here it just seems a bit off...a bit more like it was written back when tldr-pages was a much smaller project.

CleanMachine1:
> In tldr-pages, we're supposed to try to avoid voting, and instead discuss things together until we all agree. This might be too simple though thinking

Its not a page edit whereby there is multiple ways of wording it.

Its a:

Yes, I approve of removing the emoticon 
No, I don't approve of removing the emoticon

bl-ue:
Not quite.  There's a "why I want to remove it" and a "why I don't want to remove it." 🙂

CleanMachine1:
> Other than discussing reasons, this should be simply voted for

bl-ue:
Per the [conversation on Gitter](https://gitter.im/tldr-pages/tldr?at=60cfb5e6effd491d8c769d67), I'm going to close this.

@waldyrious:
>Hey @bl-ue, I will comment here for the time being, maybe later I might add more detailed feedback if the discussion is still ongoing. Honestly I don't have strong feelings either way. I like the emoji for the reasons others have pointed out, but it wouldn't bother me much if it were removed. Can you explain better why it bothers you that it's there? I'm not sure I understand the "voice of the community" concern. We do want to convey a friendly, informal, welcoming tone, and the emoji seems to help with that IMO.
>I'd actively support its removal only if some people interpret it as being sarcastic or passive-aggressive.

@bl-ue:
>Of course, but too me emoji are better in message sent by obviously one person, whereas that document is written by a lot of people now. Back in March 2014 when it was written by Romain Prieto it was certainly him speaking for himself, but...

@sbrl:
>That doesn't mean to say that the emoji can't echo the sentiment of the community as a whole

@waldyrious:
>I think that may be just a cultural thing :) the wording of the document itself in no way conveys a personal voice, and I don't feel that using emojis gives that impression.

@waldyrious:
>If others agreed that it did, I'd consider it worthwhile to rework that passage. I agree that the most we can encourage a collective voice in project docs, the better. But I don't think the emoji is working against that.

@bl-ue:
>No, certainly not. It's difficult to put my finger exactly on what bothers me about it but I assure you it does so very little.

PR #6161, opened 2021-06-22
kubectl-run: add page

cristamd:
Added a page with some examples of the `kubectl run` command.


cristamd:
Just saw how to make a sub-page, I'll do that and update the PR.

bl-ue:
@cristamd the usual page PR title format is as follows:

bl-ue:
because you're adding a whole new page here.  If you were just adding an example to `kubectl.md`, the old title would be correct though.

pages/common/kubectl-run.md
```diff
@@ -0,0 +1,24 @@
+# kubectl run
+
+> Run pods in Kubernetes. Specifies pod generator to avoid deprecation error in some K8S versions.
+> More information: <https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#run>.
+
+- Start an nginx pod and expost port 80
```
mfrw, line 6: How about ?
```suggestion
- Spawn an nginx pod and export port 80
```

pages/common/kubectl-run.md
```diff
@@ -0,0 +1,24 @@
...
+
+`kubectl run --generator=run-pod/v1 nginx --image=nginx --port 80`
+
+- Run an nginx pod, setting the TEST_VAR environment variable:
```
mfrw, line 10:
```suggestion
- Sawn an nginx pod and set the TEST_VAR environment variable:
```

pages/common/kubectl-run.md
```diff
@@ -0,0 +1,24 @@
...
+
+`kubectl run --generator=run-pod/v1 nginx --image=nginx --env="TEST_VAR=testing"`
+
+- Show API calls that would be made to create an nginx container:
+
+`kubectl run --generator=run-pod/v1 nginx --image=nginx --dry-run`
+
+- Run a one-off Ubuntu pod interactively, never restart it, and remove it when it exits:
```
mfrw, line 18: I am not sure about the comma here, I might not be correct here .. i am going with my gut.
Please correct me if ....
```suggestion
- Spawn a Ubuntu pod interactively, never restart it and remove it when it exits:
```

pages/common/kubectl-run.md
```diff
@@ -0,0 +1,24 @@
...
+
+`kubectl run --generator=run-pod/v1 -it temp-ubuntu --image=ubuntu:20.04 --restart=Never --rm -- /bin/bash`
+
+- Run a Ubuntu pod overriding the default command with echo and specifying custom arguments:
```
mfrw, line 22:
```suggestion
- Spawn a Ubuntu pod, override the default command with 'echo' and specify custom arguments:
```

mfrw (changes requested):
@cristamd Thank you for your contribution. I have a few review comments :)

pages/common/kubectl-run.md
sbrl: Perhaps this will read better?
```suggestion
- Spawn an nginx pod and expose port 80:
```

pages/common/kubectl-run.md
sbrl: Actually, I think this example is fine as-is.

pages/common/kubectl-run.md
sbrl: I think this example is also fine as-is.

pages/common/kubectl-run.md
sbrl: Again, this is mostly fine as-is - but could probably do with a comma or two:
```suggestion
- Run an Ubuntu pod, overriding the default command with echo, and specifying custom arguments:
```

pages/common/kubectl-run.md
bl-ue: If you remove `one-off` you have to change `a Ubuntu` to `an Ubuntu`.

bl-ue: Also, if you remove `one-off` you have to change `a Ubuntu` to `an Ubuntu`.

cristamd:
Pushed a small change to use `run` instead of `spawn` for consistency with the command name itself, and removed the word one-off from the auto-remove container example. Please let me know if you'd like me to change anything else.

pages/common/kubectl-run.md
mfrw: Sure

pages/common/kubectl-run.md
mfrw: > I think this example is also fine as-is. Any particular reason for `Spawn` over `Run`?

In my understanding a pod is `spawned`. What i am looking for here is like .. we spawn a process and run a command.
Although a nitpick, but my gut disagrees to run a `pod`. I would more like say.. 
Run a Command; Spawn a process .. but still i am open :)

cristamd: The linked documentation page (https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#run) uses the terminology `run` for pods, so I think it would make sense to remain consistent with that in the documentation here.

sbrl: Ah, I see @cristamd. In that case, we should definitely use `Run` instead of `Spawn`.

sbrl (approved):
Thanks, @cristamd!

sbrl:
Merging, since @mfrw's comments appear to be addressed.

PR #6162, opened 2021-06-23
parted: add See also

aherst:
Added a reference to partprobe since it's a good idea to notify the kernel after we manipulate the partition table.

### DIFF
--- a/pages/linux/parted.md
+++ b/pages/linux/parted.md
@@ -1,6 +1,7 @@
 # parted
 
 > A partition manipulation program.
+> See also: `partprobe`.
 > More information: <https://www.gnu.org/software/parted/parted.html>.
 
 - List partitions on all block devices:


bl-ue (approved):

marchersimon (approved):

CleanMachine1:
I don't see its use. So I would personally decline this however I approve on the fact that everyone else approves with it.

CleanMachine1 (approved):

marchersimon:
@CleanMachine1 I don't know either of the commands well, but I trust in @ aherst's experience with that command that you might also want to have a look at `partprobe` if you're using it.

aherst:
Thanks @marchersimon. 

@CleanMachine1, I added the See also to `parted` because it is used to make changes to the partition table on a disk (similar to `fdisk`). If the kernel isn't made aware of the changes before a write, it can corrupt the contents of the disk. For example, after updating a partition table, `fdisk` gives a message similar to:

```
WARNING: Re-reading the partition table failed with error 16: Device or resource busy.
The kernel still uses the old table.
The new table will be used at the next reboot.
Syncing disks.
```

One way to update the kernel is to reboot. Another less intrusive way is to run `partprobe`. Since I had just added a tldr-page for `partprobe`, I thought it would be a good idea to reference it from `parted`. (While preparing this repsonse I discovered that we have an `fdisk` tldr-page. If we agree to this change, I'm going to add the See also to `fdisk` as well.)

CleanMachine1:
After reviewing, I could see its purpose :)

PR #6170, opened 2021-06-24
ufw: add more information link

CleanMachine1:
For #6062

PR #6171, opened 2021-06-25
ufw: use manned as more information link

navarroaxel:

Related discussion: https://github.com/tldr-pages/tldr/pull/6170#discussion_r658667001

### DIFF
--- a/pages/linux/ufw.md
+++ b/pages/linux/ufw.md
@@ -2,7 +2,7 @@
 
 > Uncomplicated Firewall.
 > Frontend for iptables aiming to make configuration of a firewall easier.
-> More information: <https://wiki.ubuntu.com/UncomplicatedFirewall>.
+> More information: <https://manned.org/ufw>.
 
 - Enable ufw:
 


bl-ue (approved):

CleanMachine1:
Why?

the ubuntu page is way better than a man page. It provides examples, modification and what it does feature-wise.

Manned is for last resorts when there is nothing else, correct?

CleanMachine1:
https://github.com/tldr-pages/tldr/pull/6170#discussion_r658709669

Nevermind, I see its use. I hadn't looked at the other PR.

I can't tell which is better however I am willing to have either

navarroaxel:
I believe that a link to a wiki page is better than the same `man page` that the user has in the CLI, but this was the decision that we made in #5510 :man_shrugging:

CleanMachine1 (approved):

sbrl:
Yeah, generally we only go for a man page if there's nothing else that's appropriate, though it's really a case-by-case thing.

sbrl:
Believe it or not, UFW is actually Ubuntu-specific as best I can tell - at least it started that way. To that end, the Ubuntu wiki page is more informative here.

CleanMachine1:
UFW was made for ubuntu to my knowledge however seems to adopted from uncomplicated nature.

marchersimon:
> @CleanMachine1 I believe that a link to a wiki page is better than the same `man page` that the user has in the CLI, but this was the decision that we made in #5510 man_shrugging

@navarroaxel as far as I can remember we only made a decision for which man page side we should use as default. We always preferred Wikis or Homepages over man pages.

PR #6076, opened 2021-05-30
linux/[a-g]: add more information link

Stig124:
Partial fix for #6062 on the linux side of pages, with commits notarized by letters
- [x] The page (if new), does not already exist in the repo.
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #6172, opened 2021-06-25
set-more-info-link.py: add AutoMan option

Stig124:
Fix of bloated PR #6084

### DIFF
--- a/scripts/set-more-info-link.py
+++ b/scripts/set-more-info-link.py
@@ -6,50 +6,49 @@ import subprocess
 import sys
 
 labels = {
-    'en': 'More information:',
-    'bs': 'Više informacija:',
-    'da': 'Mere information:',
-    'de': 'Weitere Informationen:',
-    'es': 'Más información:',
-    'fa': 'اطلاعات بیشتر:',
-    'fr': 'Plus d\'informations\xa0:',
-    'sh': 'Više informacija:',
-    'hi': 'अधिक जानकारी:',
-    'id': 'Informasi lebih lanjut:',
-    'it': 'Maggiori informazioni:',
-    'ja': '詳しくはこちら:',
-    'ko': '더 많은 정보:',
-    'ml': 'കൂടുതൽ വിവരങ്ങൾ:',
-    'nl': 'Meer informatie:',
-    'no': 'Mer informasjon:',
-    'pl': 'Więcej informacji:',
-    'pt_BR': 'Mais informações:',
-    'pt_PT': 'Mais informações:',
-    'ru': 'Больше информации:',
-    'sv': 'Mer information:',
-    'ta': 'மேலும் தகவல்:',
-    'th': 'ดูเพิ่มเติม:',
-    'tr': 'Daha fazla bilgi için:',
-    'zh_TW': '更多資訊：',
-    'zh': '更多信息：',
+    "en": "More information:",
+    "bs": "Više informacija:",
+    "da": "Mere information:",
+    "de": "Weitere Informationen:",
+    "es": "Más información:",
+    "fa": "اطلاعات بیشتر:",
+    "fr": "Plus d'informations\xa0:",
+    "sh": "Više informacija:",
+    "hi": "अधिक जानकारी:",
+    "id": "Informasi lebih lanjut:",
+    "it": "Maggiori informazioni:",
+    "ja": "詳しくはこちら:",
+    "ko": "더 많은 정보:",
+    "ml": "കൂടുതൽ വിവരങ്ങൾ:",
+    "nl": "Meer informatie:",
+    "no": "Mer informasjon:",
+    "pl": "Więcej informacji:",
+    "pt_BR": "Mais informações:",
+    "pt_PT": "Mais informações:",
+    "ru": "Больше информации:",
+    "sv": "Mer information:",
+    "ta": "மேலும் தகவல்:",
+    "th": "ดูเพิ่มเติม:",
+    "tr": "Daha fazla bilgi için:",
+    "zh_TW": "更多資訊：",
+    "zh": "更多信息：",
 }
 
-IGNORE_FILES = (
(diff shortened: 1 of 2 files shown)


Stig124:
Any news?

scripts/set-more-info-link.py
```diff
@@ -6,50 +6,49 @@
...
 
-    if 'TLDR_ROOT' in os.environ:
-        return os.environ['TLDR_ROOT']
+    if "TLDR_ROOT" in os.environ:
+        return os.environ["TLDR_ROOT"]
     else:
         print(
-            '\x1b[31mPlease set TLDR_ROOT to the location of a clone of https://github.com/tldr-pages/tldr.')
+            "\x1b[31mPlease set TLDR_ROOT to the location of a clone of https://github.com/tldr-pages/tldr."
+        )
```
marchersimon, line 51: Why don't we just put this in a single line?

marchersimon (commented):
All the `'` → `"` changes look good, however I can't really review any of the new code since I don't know enough python for that

scripts/set-more-info-link.py
Stig124: > Why don't we just put this in a single line?

`PEP8` (the standard for formatting), kinda push those kind of layouts especially with the line lenght limit (which is 72 char)

sbrl (commented):
Looks ok to me at a quick glance, but haven't tested it.

What's the reasoning behind changing all the quotes in this file?

Stig124:
> Looks ok to me at a quick glance, but haven't tested it.
> 
> What's the reasoning behind changing all the quotes in this file?

`PEP8`, mismatching quotes is not a thing recommended by the standard, so I unified it

scripts/set-more-info-link.py
```diff
@@ -6,50 +6,49 @@
 import sys
```
marchersimon, line 6:
```suggestion
import sys
import requests
```

scripts/set-more-info-link.py
```diff
@@ -62,85 +61,127 @@ def set_link(file, link):
...
-                f'{platform}/{p}' for p in pages if p not in IGNORE_FILES]
+            commands = [f"{platform}/{p}" for p in pages if p not in IGNORE_FILES]
             if args.page in commands:
                 path = os.path.join(pages_dir_path, args.page)
                 target_paths.append(path)
 
     target_paths.sort()
+    if args.man:
+        page_name = os.path.basename(args.page).rstrip(".md")
+        link = manurl + page_name
```
marchersimon, line 163:
```suggestion
        link = manurl + page_name
        request = requests.get(link)
        if not request.status_code == 200:
            print("\x1b[31m" + link + " does not exist")
            sys.exit(1)
```

marchersimon (commented):
I now tested the script and everything seems to work. The only issue I found was that it doesn't detect non-existing links. My suggestion below should take care of that.

scripts/set-more-info-link.py
MasterOdin: Can use `raise SystemExit("\x1b[31m" + link + " does not exist")` instead of the combination of `print` and `sys.exit(1)`. This has the added benefit that the output will go to stderr instead of stdout (though you could do `file=sys.stderr` in the print).

MasterOdin:
> PEP8, mismatching ' and " is not a thing recommended by the standard, so I unified it

If you're going to go through with this change, it should be immediately backed up with a CI workflow that makes sure this trend continues or else it's almost certainly to drift from the standard.

As an aside, I'd very much rather see this kind of formatting change happen in a separate PR, both to make this one easier to review, as well as in the case of ever having to revert or whatever.

scripts/set-more-info-link.py
marchersimon: Sounds good, I can't make a suggestion on the other one, so it's @Stig's turn to apply that if they want.

MasterOdin: Python 3.5.0 was released 09/13/2015 and reached end-of-life in 09/20/2020. Assuming you're not running a very old LTS release (e.g. 16.04), you should be running minimally Python 3.6+.

stig: I think you want @Stig124

scripts/set-more-info-link.py
CleanMachine1: Could this not be merged into one line.
```suggestion
import sys, requests
```

scripts/set-more-info-link.py
CleanMachine1: Wrong one!

scripts/set-more-info-link.py
MasterOdin: Yes, but per [PEP-8]:

> Imports should usually be on separate lines:
> ```
> # Correct:
> import os
> import sys
> ```
> ```
> # Wrong:
> import sys, os
> ```
> It's okay to say this though:
> ```
> # Correct:
> from subprocess import Popen, PIPE
> ```

marchersimon:
Damn over-sensitive touchpad

marchersimon:
I think we can apply the suggestions and merge. @ Stig124 already said [here](https://github.com/tldr-pages/tldr/pull/6076#issuecomment-869921868), that they don't have the time to finish their PRs.

Stig124:
> I think we can apply the suggestions and merge. @ Stig124 already said [here](https://github.com/tldr-pages/tldr/pull/6076#issuecomment-869921868), that they don't have the time to finish their PRs.

Sorry again, I was on vacation, will do those things ASAP

Stig124:
Feel free to takeover my PRs, It's a mess in my life rn

marchersimon:
Sad to hear. I hope it get's better soon.

marchersimon:
@MasterOdin do you think we can merge this as-is or does it need more attention?

PR #6181, opened 2021-06-28
jtbl: add page

kellyjonbrazil:
Add new page for jtbl command

### DIFF
--- a/pages/common/jtbl.md
+++ b/pages/common/jtbl.md
@@ -3,7 +3,7 @@
 > Utility to print JSON and JSON Lines data as a table in the terminal.
 > More information: <https://github.com/kellyjonbrazil/jtbl>.
 
-- Print a table from JSON input:
+- Print a table from JSON or JSON Lines input:
 
 `cat {{file.json}} | jtbl`
 
--- a/pages/common/jtbl.md
+++ b/pages/common/jtbl.md
@@ -15,6 +15,6 @@
 
 `cat {{file.json}} | jtbl -t`
 
-- No wrapping or truncation of rows:
+- Don't wrap or truncate rows:
 
 `cat {{file.json}} | jtbl -n`
--- a/pages/common/jtbl.md
+++ b/pages/common/jtbl.md
@@ -1,0 +1,20 @@
+# jtbl
+
+> Utility to print JSON and JSON Lines data as a table in the terminal.
+> More information: <https://github.com/kellyjonbrazil/jtbl>.
+
+- Print a table from JSON input:
+
+`cat {{file.json}} | jtbl`
+
+- Specify the column width for wrapping:
+
+`cat {{file.json}} | jtbl --cols={{number}}`
+
+- Truncate rows instead of wrapping:
+
+`cat {{file.json}} | jtbl -t`
+
+- No wrapping or truncation of rows:
+
+`cat {{file.json}} | jtbl -n`
--- a/pages/common/jtbl.md
+++ b/pages/common/jtbl.md
@@ -7,14 +7,14 @@
 
 `cat {{file.json}} | jtbl`
 
-- Specify the column width for wrapping:
+- Print a table and specify the column width for wrapping:
 
 `cat {{file.json}} | jtbl --cols={{width}}`
 
-- Truncate rows instead of wrapping:
+- Print a table and truncate rows instead of wrapping:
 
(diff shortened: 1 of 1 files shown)


pages/common/jtbl.md
```diff
@@ -0,0 +1,20 @@
...
+> Utility to print JSON and JSON Lines data as a table in the terminal.
+> More information: <https://github.com/kellyjonbrazil/jtbl>.
+
+- Print a table from JSON input:
+
+`cat {{file.json}} | jtbl`
+
+- Specify the column width for wrapping:
+
+`cat {{file.json}} | jtbl --cols={{number}}`
```
marchersimon, line 12:
```suggestion
- Specify the column width for wrapping:

`cat {{file.json}} | jtbl --cols={{width}}`
```

pages/common/jtbl.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Truncate rows instead of wrapping:
+
+`cat {{file.json}} | jtbl -t`
+
+- No wrapping or truncation of rows:
```
marchersimon, line 18:
```suggestion
- Don't wrap or truncate rows:
```

marchersimon (changes requested):
Just a few suggestion. Also, since the description says you can also print JSON Lines data, does the tool autodetect the format or does it need some additional flag?

kellyjonbrazil:
Thanks for the updates! `jtbl` autodetects JSON vs JSON Lines, so no other options are required.

CleanMachine1:
Since this command works on windows, cat should probably be replaced with echo, since it makes it compatible. However using echo doesn't fit this page at all so I feel we need others opinions

Proof of windows support for anyone 
```
You can install jtbl via pip, via OS Package Repositories, MSI installer for Windows, or by downloading the correct binary for your architecture and running it anywhere on your filesystem.
```

marchersimon (approved):
Thanks @kellyjonbrazil

marchersimon:
@CleanMachine1 I think replacing cat by any other command would probably confuse more than it would help anyone using windows. (Personally I think that we should focus on other platforms, since those are the ones tldr is used on)

kellyjonbrazil:
I think you would use `type` in windows vs. `cat` in linux/etc., but it might get too wordy to include that.

pages/common/jtbl.md
```diff
@@ -0,0 +1,20 @@
...
+
+`cat {{file.json}} | jtbl --cols={{width}}`
+
+- Truncate rows instead of wrapping:
+
+`cat {{file.json}} | jtbl -t`
+
+- Don't wrap or truncate rows:
+
+`cat {{file.json}} | jtbl -n`
```
CleanMachine1, line 20: I like when descriptions tell you exactly what they do, rather than just making the user infer.

So for example command 2 `Print a table from JSON or JSON Lines input and truncate rows instead of wrapping them:

CleanMachine1: Sure.

CleanMachine1 (approved):

nicokosi (approved):
👍
PS: I did not know the "JSON Lines" term (but I knew the concept). 🤓

nicokosi:
> I think you would use `type` in windows vs. `cat` in linux/etc., but it might get too wordy to include that.

Good point. Not sure it's too wordy. What about something like:

> Utility to print JSON and JSON Lines data as a table in the terminal, reading from the standard input (for instance, via `cat` on Linux/macOS or `type` on Windows).
 > More information: <https://github.com/kellyjonbrazil/jtbl>.

mfrw:
> > I think you would use `type` in windows vs. `cat` in linux/etc., but it might get too wordy to include that.
> 
> Good point. Not sure it's too wordy. What about something like:
> 
> > Utility to print JSON and JSON Lines data as a table in the terminal, reading from the standard input (for instance, via `cat` on Linux/macOS or `type` on Windows).
> > More information: https://github.com/kellyjonbrazil/jtbl.

This seems a little wordy to me ... lets see what other folks have to say

CleanMachine1:
do these flags have long versions --

nicokosi:
> do these flags have long versions --

No.

CleanMachine1:
oh cool, you made this! i didn't see that, sorry.

sbrl (approved):
Thanks, @kellyjonbrazil!

PR #6184, opened 2021-06-29
send-to-bot.py: use requests instead of urllib

### LINKED ISSUE #6173: Switch send-to-bot.py to requests from raw urllib (and do some modifications)
I propose that the `send-to-bot.py` file should switch from `urllib` to `requests`, which is a wrapper of urllib3, and would be way more understandable/workable than `urllib.requests` (builtin support for dictionary => JSON conversion...). This script has some issues too, mainly with formatting (indentation and spacing).

A PR is underway to try fixing some of those issues and update that script

Stig124:
Closes #6173

**NOTE** : Might break things as I haven't tested any of this

Update from `urllib.requests` to `requests`, remove `except Exception` (not a good idea as it will introduce bugs [[1]](https://stackoverflow.com/a/18982726/11478363), format (using `black`)

### DIFF
--- a/scripts/send-to-bot.py
+++ b/scripts/send-to-bot.py
@@ -2,12 +2,11 @@
 
 import os
 import sys
-import json
-import urllib.request
+import requests
 
-BOT_URL = 'https://tldr-bot.starbeamrainbowlabs.com'
+BOT_URL = "https://tldr-bot.starbeamrainbowlabs.com"
 
-COMMENT_ERROR="""
+COMMENT_ERROR = """
 The [build](https://github.com/tldr-pages/tldr/actions/runs/{build_id}) for this PR failed with the following error(s):
 
 ```
@@ -17,7 +16,7 @@ The [build](https://github.com/tldr-pages/tldr/actions/runs/{build_id}) for this
 Please fix the error(s) and push again.
 """
 
-COMMENT_CHECK="""
+COMMENT_CHECK = """
 Hello! I've noticed something unusual when checking this PR:
 
 {content}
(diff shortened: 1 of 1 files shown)


navarroaxel:
I think you don't need to change the `.py` file to executable `755`.

scripts/send-to-bot.py
```diff
@@ -27,66 +26,60 @@
...
+    if once:
+        endpoint += "/once"
 
-  if code != 200:
-    print('Error: tldr-bot responded with code', code, file=sys.stderr)
-    print(resp.read(), file=sys.stderr)
-    return False
+    headers = {"Content-Type": "application/json"}
+    data = {"pr_id": pr_id, "body": body}
+    with requests.post(endpoint, data, headers) as r:
```
MasterOdin, line 38:
```suggestion
    data = {"pr_id": pr_id, "body": body}
    with requests.post(endpoint, json=data) as r:
```

Using the `json=` named parameter will automatically handle setting the content-type as well as making sure the data is properly encoded.

MasterOdin (changes requested):
I do agree with moving to `requests` for this as it does make the resulting code simpler, and it's such a ubiquitous package that it's probably installed anyway if you use other python packages.

Will need to modify https://github.com/tldr-pages/tldr/blob/main/.github/workflows/ci.yml to make sure that requests is installed.

scripts/send-to-bot.py
marchersimon: @MasterOdin I think we can apply the changes ourselves by now. Can this PR be merged after this suggestion?

MasterOdin:
I've rolled back to style changes to the file to keep this PR more focused. A style PR (and associated changes to CI workflow) is done in #6454.

MasterOdin (approved):

navarroaxel (approved):

marchersimon:
I think this PR broke something. Here's an example output of the workflow:
image

PR #6188, opened 2021-07-01
git-commit: add gpg-sign example and sync German translation

marchersimon:
If nothing changes, we probably should do a rebase for this one.

### DIFF
--- a/pages/common/git-commit.md
+++ b/pages/common/git-commit.md
@@ -15,6 +15,10 @@
 
 `git commit -a -m "{{message}}"`
 
+- Commit staged files and [S]ign them with the GPG key defined in `~/.gitconfig`:
+
+`git commit -S -m "{{message}}"`
+
 - Update the last commit by adding the currently staged changes, changing the commit's hash:
 
 `git commit --amend`
--- a/pages.de/common/git-commit.md
+++ b/pages.de/common/git-commit.md
@@ -15,6 +15,10 @@
 
 `git commit -a -m "{{nachricht}}"`
 
+- Committe alle gestagten Dateien und [S]igniere sie mit dem in `~/.gitconfig` definierten GPG Schlüssel:
+
+`git commit -S -m "{{nachricht}}"`
+
 - Ersetze den letzten Commit mit den gerade auf dem Stage liegenden Änderungen:
 
 `git commit --amend`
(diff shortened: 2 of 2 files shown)


pages/common/git-commit.md
```diff
@@ -15,6 +15,10 @@
 
 `git commit -a -m "{{message}}"`
 
+- Commit staged files and sign them with the GPG key defined in the config:
```
CleanMachine1, line 18:
```suggestion
- Commit staged files and sign them with the GPG key defined in .gitconfig:
```

marchersimon: What about Mac and Windows users?

CleanMachine1: Ok, what do they have though?

marchersimon: Nevermind, they the file is called `.gitconfig` everywhere. Since Mac also supports it how about if we even used `~/.gitconfig`?

CleanMachine1: Does windows use ~ ?

Then again who is using git for windows anyways?
We could leave it as what you have suggested, or help Linux and Mac users in locating the file for git configuration.

If we are not going to support windows here then it should be ~/.gitconfig

As you have mentioned

navarroaxel: `git-config` page uses `(stored in `~/.gitconfig`)`

pages/common/git-commit.md
```diff
@@ -15,6 +15,10 @@
...
+
+`git commit -S -m "{{message}}"`
```
navarroaxel, line 20: This is not a common command, could we use this?
```suggestion
`git commit --gpg-sign -m "{{message}}"`
```

marchersimon: I think it's implied in the description that `-S` signs with a GPG key. Also, `s` for `sign` is pretty obvious. Also if you're using `-S` instead of auto-signing it's as common as `-m`.

CleanMachine1: Exactly, I'd assume most that use GPG signing use the auto sign in .gitconfig

If I didn't then all of my commands would use -S, making it just as used as -m

CleanMachine1: @navarroaxel anything to add to this or resolve?

CleanMachine1 (approved):

pages/common/git-commit.md
sbrl: I suggest either using the long form or using `[S]` in the description.

marchersimon: I think when you're actually using `-S` it would be very impractical to use the long form, so I'll go with `[S]ign`

SethFalco (approved):

navarroaxel (approved):

sbrl (approved):

PR #6196, opened 2021-07-05
fdisk: refresh

CleanMachine1:
fdisk is a crucial disk editing command, which is key for setting up Arch Linux for example.

However it is not documented very well and is a useful command itself when things like GParted aren't available because someone doesn't have a desktop 

My edits will need a little help

### DIFF
--- a/pages/linux/fdisk.md
+++ b/pages/linux/fdisk.md
@@ -5,11 +5,11 @@
 
 - List partitions:
 
-`fdisk -l`
+`sudo fdisk -l`
 
 - Start the partition manipulator:
 
-`fdisk {{/dev/sdX}}`
+`sudo fdisk {{/dev/sdX}}`
 
 - Once partitioning a disk, create a partition:
 
--- a/pages/linux/fdisk.md
+++ b/pages/linux/fdisk.md
@@ -15,7 +15,7 @@
 
 `n`
 
-- Once partitioning a disk, delete a specific partition:
+- Once partitioning a disk, select a partition to delete:
 
 `d`
 
--- a/pages/linux/fdisk.md
+++ b/pages/linux/fdisk.md
@@ -10,3 +10,27 @@
 - Start the partition manipulator:
 
 `fdisk {{/dev/sdX}}`
+
+- Once partitioning a disk, create a partition:
+
+`n`
+
+- Once partitioning a disk, delete a specific partition:
+
+`d`
+
+- Once partitioning a disk, view the partition table:
+
+`p`
+
+- Once partitioning a disk, write the changes made:
+
+`w`
+
+- Once partitioning a disk, discard the changes made:
+
+`q`
+
+- Once partitioning a disk, open a help menu:
+
+`m`


pages/linux/fdisk.md
```diff
@@ -10,3 +10,27 @@
...
+
+- Once partitioning a disk, write the changes made:
+
+`w`
+
+- Once partitioning a disk, discard the changes made:
+
+`q`
+
+- Once partitioning a disk, open a help menu:
```
navarroaxel, line 34: I think this should be the first interactive mode command.

pages/linux/fdisk.md
```diff
@@ -10,3 +10,27 @@
 - Start the partition manipulator:
 
 `fdisk {{/dev/sdX}}`
+
+- Once partitioning a disk, create a partition:
+
+`n`
+
```
CleanMachine1, line 17: This needs to be looked at, the syntax for adding memory disk allocation is `+{{size}}{{K|M|G|T|P}}`

To a beginner and me, that is odd. I would put 500M or 500MB or 500 MB, the be  confused that it didn't work and have to look it up on my phone, which is annoying enough (when installing arch, unable to use a web browser)

navarroaxel (commented):
Why you didn't add the `g` command?

CleanMachine1:
> Why you didn't add the `g` command?

Probably because I have never used it, I am a BIOS boot person. UEFI booting just annoys me

CleanMachine1:
You could argue that both the MSDOS table and GPT are important, however in comparison to the others, they seem a little less useful since most people formatting their disks will already have a partition table, and just need to partition it correctly, however from a view point of just using the tool instead of GParted, then I could definitely see the use. 

However which is more important can be decided, I'd say p is the least important (still really useful), however all the others are command that everyone using the tool needs to know like discarding if mistakes are made or writing when ready

CleanMachine1:
Another thing that needs to be brought up here, do we make the commands long, i am pretty sure d can be turned into delete. 

Making them long might make it easier for a user, 

g - gpt

gpt makes more sense

pages/linux/fdisk.md
CleanMachine1: Are you referring to the GPT table command or q or m

pages/linux/fdisk.md
CleanMachine1: The only possible solution I can see is maybe putting that in the description

sbrl: `fdisk` is an interactive command, so you don't specify the syntax for the size of the partition here - I'm confused what you're referring to here?

pages/linux/fdisk.md
sbrl: I don't think it matters too much, but @navarroaxel is referring to the help `m` command.

sbrl:
In this case @CleanMachine1 although it would be beneficial to have long options, then man page and help does not mention it, so it's not officially supported.

pages/linux/fdisk.md
CleanMachine1: At the command prompt, it is done during the command, 

EG

$ sudo fdisk /dev/sda 

new 

_Enter size of partition 1_ +500M

pages/linux/fdisk.md
CleanMachine1: I prefer the help message to be at the bottom, and it makes sense anyway because the help command letter is hinted at the bottom always anyway

CleanMachine1:
Open an fdisk (don't write to any disks) and type variations of commands, `ne` counts as `n`, so it seems to cover for the user regardless of typos

pages/linux/fdisk.md
```diff
@@ -10,3 +10,27 @@
 - Start the partition manipulator:
 
 `fdisk {{/dev/sdX}}`
```
CleanMachine1, line 12:
```suggestion
`sudo fdisk {{/dev/sdX}}`
```

pages/linux/fdisk.md
```diff
@@ -10,3 +10,27 @@
...
+
+- Once partitioning a disk, create a partition:
```
marchersimon, line 14: How about
```suggestion
- Create a partition within `fdisk`:
```
Maybe also `inside of` or `in`

CleanMachine1: I don't like that. To me that just means, 

when you use this command, it creates a partition.

or in tldr terms

- Create a partition:

CleanMachine1: It being interactive shows that you need a previous command to reach that stage

marchersimon: That's what the `within fdisk` is for. It tells you that you have to be inside of fdisk (running it) to be able to execute the command.

CleanMachine1: It guess, but I don't think it does as well

CleanMachine1:
```suggestion
- Once manipulating a disk, create a partition:
```
```suggestion
- Once editing a disk, create a partition:
```

sbrl:
> Open an fdisk (don't write to any disks) and type variations of commands, `ne` counts as `n`, so it seems to cover for the user regardless of typos

Yes, while I don't deny it works it's not officially documented, so it's not guaranteed that it's not going to change in the future.

Also, practically all modern operating systems use GPT partition tables - even without UEFI booting (which I might add still allows you to disable secure boot, which for me is what really annoys me)?

navarroaxel:
> Also, practically all modern operating systems use GPT partition tables - even without UEFI booting (which I might add still allows you to disable secure boot, which for me is what really annoys me)?

Yep, with UEFI and GPT you can still disabled `secure boot`.

pages/linux/fdisk.md
```diff
@@ -10,3 +10,27 @@
...
+
+`n`
+
+- Once partitioning a disk, delete a specific partition:
```
marchersimon, line 18: Do you have to select the partition beforehand or do you have to enter the partition in a prompt?

CleanMachine1: I think it gives you a prompt to pick a partition to delete, however if there is only 1 then it is instantly deleted.

pages/linux/fdisk.md
CleanMachine1: Your suggestion?

pages/linux/fdisk.md
CleanMachine1: Any thoughts on this, since it is using block devices it needs root to do anything really

marchersimon: From the guidline:
> When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness. For example, commands that require `sudo` should include it directly in the examples.

pages/linux/fdisk.md
marchersimon: If we're keeping that sentence structure, I'd prefer `Once partitioning a disk, ...` like it is right now.

pages/linux/fdisk.md
marchersimon: How about
```suggestion
- Once partitioning a disk,open a prompt to ask for a partition to delete:
```
With the old description I'd be way to paranoid that it deletes a partition without we knowing which.

pages/linux/fdisk.md
CleanMachine1: `fdisk: cannot open /dev/sda: Permission denied`


Block devices have root permissions attached, so I will add it.

pages/linux/fdisk.md
CleanMachine1:
```suggestion
- Once partitioning a disk, select a partition to delete:
```

marchersimon: Sounds good 👍🏽

marchersimon (approved):
Looks good, thanks

pages/linux/fdisk.md
CleanMachine1: You need to put a +, which IMO is an odd syntax

sbrl (approved):
Thanks, @CleanMachine1!

PR #6200, opened 2021-07-06
style-guide: put token guidelines into sections & add varargs

xeruf:
Fixes: #5791


xeruf:
It was quite unwieldy, now it feels much more manageable to me :)

contributing-guides/style-guide.md
```diff
@@ -47,32 +47,43 @@ in order to allow `tldr` clients to highlight them.
...
+- Use an actual value rather than a generic placeholder where appropriate.
+  For example, use `iostat {{2}}` rather than `iostat {{interval_in_secs}}`.
+
+### Paths
+- Use `{{filename}}` rather than `{{file_name}}`.
+- For any reference to paths to files or directories, use the format `{{path/to/<placeholder>}}`.
+  For example, `ln -s {{path/to/file}} {{path/to/symlink}}`.
+  In case of a possible reference both to a file or a directory, use `{{path/to/file_or_directory}}`
+- Follow the `{{path/to/<placeholder>}}` convention for all path-related commands,
+  except when the file location is implicit.
```
marchersimon, line 63:
```suggestion
- For any reference to paths to files or directories, use the format `{{path/to/<placeholder>}}`, except when the file location is implicit.
  For example, `ln -s {{path/to/file}} {{path/to/symlink}}`.
  In case of a possible reference both to a file or a directory, use `{{path/to/file_or_directory}}`
```
I think we can merge those two

contributing-guides/style-guide.md
```diff
@@ -47,32 +47,43 @@ in order to allow `tldr` clients to highlight them.
...
+#### Extensions
+- If a particular extension is expected for the file, append it.
+  For example, `unrar x {{compressed.rar}}`.
+- In case a generic extension is needed, use `{{.ext}}`, but **only** if an extension is required.
+  For instance, in find.md's example "Find files by extension" (`find {{root_path}} -name '{{*.ext}}'`)
```
marchersimon, line 68:
```suggestion
  For instance, in `find.md`'s example "Find files by extension" (`find {{root_path}} -name '{{*.ext}}'`)
```
I think that's appropriate here

contributing-guides/style-guide.md
```diff
@@ -47,32 +47,43 @@ in order to allow `tldr` clients to highlight them.
...
+  For example, use `iostat {{2}}` rather than `iostat {{interval_in_secs}}`.
+
+### Paths
+- Use `{{filename}}` rather than `{{file_name}}`.
+- For any reference to paths to files or directories, use the format `{{path/to/<placeholder>}}`.
+  For example, `ln -s {{path/to/file}} {{path/to/symlink}}`.
+  In case of a possible reference both to a file or a directory, use `{{path/to/file_or_directory}}`
+- Follow the `{{path/to/<placeholder>}}` convention for all path-related commands,
+  except when the file location is implicit.
+#### Extensions
```
marchersimon, line 64:
```suggestion

### Extensions
```
I don't think this should be a `h4`, because it looks a bit weird if it's just a bit smaller then `h3`.

marchersimon (commented):
Just a few small suggestions, but looks good overall

contributing-guides/style-guide.md
xeruf: I somewhat agree, but semantically it makes sense for me to put it as a subsection of `paths`.

marchersimon: I get your intention, but I don't think it looks like a subsection. It just looks like it's been renderd a little bit too small
image

marchersimon:
@xeruf are you still working on this?

xeruf:
jup, but I haven't been able to keep up with github notifications recently ^^

contributing-guides/style-guide.md
```diff
@@ -47,32 +47,46 @@ in order to allow `tldr` clients to highlight them.
...
+### Paths
+- Use `{{filename}}` rather than `{{file_name}}`.
+- For any reference to paths to files or directories,
+  use the format `{{path/to/<placeholder>}}`,
+  except when the file location is implicit.
+  For example, `ln -s {{path/to/file}} {{path/to/symlink}}`.
+  In case of a possible reference both to a file or a directory,
+  use `{{path/to/file_or_directory}}`
+
+Extensions
```
marchersimon, line 66:
```suggestion
### Extensions
```
I assume you deleted them unintenionally

contributing-guides/style-guide.md
```diff
@@ -47,32 +47,46 @@ in order to allow `tldr` clients to highlight them.
...
+- Use [`snake_case`](https://wikipedia.org/wiki/snake_case) for multi-word tokens.
+- Use an actual value rather than a generic placeholder where appropriate.
+  For example, use `iostat {{2}}` rather than `iostat {{interval_in_secs}}`.
+
+### Paths
+- Use `{{filename}}` rather than `{{file_name}}`.
+- For any reference to paths to files or directories,
+  use the format `{{path/to/<placeholder>}}`,
+  except when the file location is implicit.
+  For example, `ln -s {{path/to/file}} {{path/to/symlink}}`.
```
marchersimon, line 62:
```suggestion
  except when the file location is implicit.
  When the path cannot be relative, but has to start at the root of the filesystem, use `{{/path/to/<placeholder>}}`
  For example, `ln -s {{/path/to/file_or_directory}} {{path/to/symlink}}`.
```
The `ln` page has been updated

marchersimon (commented):
Just two more things I'd like to add

contributing-guides/style-guide.md
xeruf: I deleted them intentionally

contributing-guides/style-guide.md
xeruf: I disagree with that update to the `ln` page though. You can perfectly well provide it a relative path. Actually I can't think of a single example where only absolute paths are okay, doesn't make much sense to me tbh.

marchersimon: While it is possible, I wouldn't recommend it. (I explained why in #6087) But we can still use another command, like `get {{/path/remote_file}}`

xeruf: Well, I prefer relative symlinks most of the time, so I think another example would indeed make more sense. But this `get` has no TLDR page, and since it is a quite generic term I am unsure what program you are referring to?

blueskyson: I think it's more about normalization of all pages. If some pages use relative paths while some pages use absolute path, it would look chaotic. So the current practice in tldr is using absolute path as many as possible.

blueskyson:
LGTM. but I personally wonder that Is there any reason to replace list marks from number to `-`?

contributing-guides/style-guide.md
marchersimon: > Well, I prefer relative symlinks most of the time, so I think another example would indeed make more sense. But this `get` has no TLDR page, and since it is a quite generic term I am unsure what program you are referring to?

My bad, I was talking about the `sftp` page.

@blueskyson what do you mean? We always use relative paths when possible. This part here is just about how to write absolute paths for commands which need it. Some commands can't use relative paths.

marchersimon:
I actually think this would be a benefit, since now we can refer to a specific rule by it's number.

contributing-guides/style-guide.md
blueskyson: I mean when to use of `{{path/to/file}}` or `{{file}}`.

marchersimon: In this guide we advice to use `{{path/to/file}}` instead of `{{file}}`. In case the command get's too long (which isn't the case very often), we can still change it in the PR.

xeruf:
They had no particular order, and it would be annoying to adjust the numbers when a rule is inserted - also the referencing would misalign after a change.

Two ideas to solve this:
- Numbering per section
- Add a shorthand that identifies each rule, e.g. SHO, CAS, VAL for the three naming rules
Latter is my favorite as it is more future-proof.

marchersimon:
I don't think this would be necessary. We don't have to refer to the style-guide that often and we can still refer to the line number or the section

marchersimon (approved):
Looks good, thanks @xeruf

CleanMachine1 (approved):
lgtm, this PR is way overdue

PR #6213, opened 2021-07-10
ssh-keygen: add key format conversion

aduzsardi:

Current versions of openssh generate keys in its own format while many CI/CD, hosted Git platforms do not accept this type of keys  
You can identify the openssh format private key by checking if it has the tags
-----BEGIN OPENSSH PRIVATE KEY-----
-----END OPENSSH PRIVATE KEY-----

### DIFF
--- a/pages.de/common/ssh-keygen.md
+++ b/pages.de/common/ssh-keygen.md
@@ -16,7 +16,7 @@
 
 - Generiere ein 4096 Bit langen RSA Schlüssel-Paar mit der Email im Kommentarfeld:
 
-`ssh-keygen -t {{dsa|ecdsa|ed25519|rsa}} -b {{4096}} -C "{{kommentar|email}}""`
+`ssh-keygen -t {{dsa|ecdsa|ed25519|rsa}} -b {{4096}} -C "{{kommentar|email}}"`
 
 - Entferne den Schlüssel eines Servers aus der `known_hosts` Datei (hilfreich wenn ein Server seinen Schlüssel aktualisiert hat und der alte somit nicht mehr gilt):
 
--- a/pages.de/common/ssh-keygen.md
+++ b/pages.de/common/ssh-keygen.md
@@ -32,4 +32,4 @@
 
 - Ändern Sie den Typ des Schlüssel formats (z. B. vom OPENSSH-Format in PEM), die Datei wird an Ort und Stelle neu geschrieben:
 
-`ssh-keygen -p -N "" -m PEM -f ~/.ssh/{{datei}}`
+`ssh-keygen -p -N "" -m {{PEM}} -f ~/.ssh/{{datei}}`
--- a/pages.de/common/ssh-keygen.md
+++ b/pages.de/common/ssh-keygen.md
@@ -33,3 +33,7 @@
 - Ändere das Passwort eines privaten Schlüssels:
 
 `ssh-keygen -p -f ~/.ssh/{{datei}}`
+
+- Ändern Sie den Typ des Schlüssel formats (z. B. vom OPENSSH-Format in PEM), die Datei wird an Ort und Stelle neu geschrieben:
+
+`ssh-keygen -p -N "" -m PEM -f ~/.ssh/{{datei}}`
--- a/pages.de/common/ssh-keygen.md
+++ b/pages.de/common/ssh-keygen.md
@@ -8,15 +8,15 @@
 
 - Erstelle ein Schlüssel-Paar unter einem bestimmten Dateinamen:
 
-`ssh-keygen -f ~/.ssh/{{datei}}`
+`ssh-keygen -f {{~/.ssh/datei}}`
 
 - Generiere ein ed25519 Schlüssel-Paar mit 100 Schlüssel-Ableitungs-Iterationen:
 
-`ssh-keygen -t ed25519 -a 100`
+`ssh-keygen -t {{ed25519}} -a {{100}}`
 
 - Generiere ein 4096 Bit langen RSA Schlüssel-Paar mit der Email im Kommentarfeld:
 
-`ssh-keygen -t rsa -b 4096 -C "{{email}}"`
+`ssh-keygen -t {{rsa}} -b {{4096}} -C "{{email}}"`
 
 - Entferne den Schlüssel eines Servers aus der `known_hosts` Datei (hilfreich wenn ein Server seinen Schlüssel aktualisiert hat und der alte somit nicht mehr gilt):
 
@@ -24,11 +24,11 @@
 
 - Rufe den Fingerabdrucks eines Schlüssels im MD5 Hex Format ab:
 
-`ssh-keygen -l -E md5 -f ~/.ssh/{{datei}}`
+`ssh-keygen -l -E {{md5}} -f {{~/.ssh/datei}}`
 
 - Ändere das Passwort eines privaten Schlüssels:
 
-`ssh-keygen -p -f ~/.ssh/{{datei}}`
+`ssh-keygen -p -f {{~/.ssh/datei}}`
 
 - Ändern Sie den Typ des Schlüssel formats (z. B. vom OPENSSH-Format in PEM), die Datei wird an Ort und Stelle neu geschrieben:
 
--- a/pages.de/common/ssh-keygen.md
+++ b/pages.de/common/ssh-keygen.md
@@ -18,10 +18,6 @@
 
 `ssh-keygen -t rsa -b 4096 -C "{{email}}"`
 
-- Rufe den Schlüssel-Fingerabdruck von einem Server ab (hilfreich um die Authentizität eines Servers beim ersten Verbinden zu überprüfen):
-
-`ssh-keygen -l -F {{externer_server}}`
-
 - Entferne den Schlüssel eines Servers aus der `known_hosts` Datei (hilfreich wenn ein Server seinen Schlüssel aktualisiert hat und der alte somit nicht mehr gilt):
 
 `ssh-keygen -R {{externer_server}}`
--- a/pages.de/common/ssh-keygen.md
+++ b/pages.de/common/ssh-keygen.md
@@ -16,7 +16,7 @@
 
 - Generiere ein 4096 Bit langen RSA Schlüssel-Paar mit der Email im Kommentarfeld:
 
-`ssh-keygen -t {{rsa}} -b {{4096}} -C "{{email}}"`
+`ssh-keygen -t {{dsa|ecdsa|ed25519|rsa}} -b {{4096}} -C "{{kommentar|email}}""`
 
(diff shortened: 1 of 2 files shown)


aduzsardi:
looks like i miscounted the examples in the file

marchersimon:
@aduzsardi new examples are always welcome. If it's a useful one we can try to replace one or merge two existing. What do you think?

aduzsardi:
if it's acceptable, sure... just which one? :)

marchersimon:
I can't really tell, since I don't have any experience with this command. What would you suggest?

aduzsardi:
I would vote for removing `ssh-keygen -l -F {{remote_host}}` , never seen anybody use this i would say most people  prefer `ssh-keyscan` instead

CleanMachine1:
Can we argue that this is more important than the command being removed, I can't tell my self since the most I have ever used this command was making keys

aduzsardi:
@CleanMachine1 the command being removed has it's uses but i can also use `ssh-keyscan` for the same functionality

`ssh-keygen -l -F {{externer_server}}` will write the content directly to `~/.ssh/known_hosts` 

to achieve the same thing with **ssh-keyscan** you have to use output redirection or pipe it to `tee` , like 
`ssh-keyscan {{externer_server}} >> ~/.ssh/known_hosts`

i don't know of any other way to convert key formats beside the one that i added 

if anybody else has input on this or other point of view, it's more than welcome

marchersimon:
Sounds reasonable. But then, we should probably add the `ssh-keyscan {{externer_server}} >> ~/.ssh/known_hosts` example to `ssh-keyscan`.

pages.de/common/ssh-keygen.md
```diff
@@ -33,3 +29,7 @@
 - Ändere das Passwort eines privaten Schlüssels:
 
 `ssh-keygen -p -f ~/.ssh/{{datei}}`
+
+- Ändern Sie den Typ des Schlüssel formats (z. B. vom OPENSSH-Format in PEM), die Datei wird an Ort und Stelle neu geschrieben:
```
marchersimon, line 33:
```suggestion
- Ändere den Typ des Schlüsselformats (z. B. vom OPENSSH-Format in PEM), die Datei wird an Ort und Stelle neu geschrieben:
```
We generally don't use polite form in German pages.

pages/common/ssh-keygen.md
```diff
@@ -33,3 +29,7 @@
 - Change the password of a key:
 
 `ssh-keygen -p -f ~/.ssh/{{filename}}`
+
+- Change the type of the key format (for example from OPENSSH format to PEM), the file will be rewritten in-place:
```
marchersimon, line 33: I'm not sure the `the file will be rewritten in-place` is neccesary, what do you think?

pages/common/ssh-keygen.md
```diff
@@ -33,3 +29,7 @@
...
+
+`ssh-keygen -p -N "" -m PEM -f ~/.ssh/{{filename}}`
```
marchersimon, line 35:
```suggestion
`ssh-keygen -p -N "" -m {{PEM}} -f ~/.ssh/{{filename}}`
```
Where is the `OPENSSH` part here? If the description mentions it, but it's nowhere in the example, it might be confusing

pages.de/common/ssh-keygen.md
aduzsardi: i used google translate for that 😄

pages/common/ssh-keygen.md
aduzsardi: could be removed , just added it as a _heads up_

pages/common/ssh-keygen.md
aduzsardi: OPENSSH is the format of the private key , like somebody explained it here https://coolaj86.com/articles/the-openssh-private-key-format/

pages.de/common/ssh-keygen.md
marchersimon: Oh, in that case it's not bad 😆

pages/common/ssh-keygen.md
CleanMachine1:
```suggestion
`ssh-keygen -p -N "" -m {{PEM -f {{~/.ssh/OpenSSH_private_key}}`
```

CleanMachine1: Keys can be stored in other places if wanted, and the keys can be chosen in the config file using `IdentityFile /path/to/key`

pages/common/ssh-keygen.md
```diff
@@ -18,10 +18,6 @@
 
 `ssh-keygen -t rsa -b 4096 -C "{{email}}"`
```
navarroaxel, line 19: I think the token syntax is missing here
```suggestion
`ssh-keygen -t {{rsa}} -b {{4096}} -C "{{email}}"`
```

pages/common/ssh-keygen.md
```diff
@@ -33,3 +29,7 @@
 - Change the password of a key:
 
 `ssh-keygen -p -f ~/.ssh/{{filename}}`
```
navarroaxel, line 31: All the path is the argument option?
```suggestion
`ssh-keygen -p -f {{~/.ssh/filename}}`
```

CleanMachine1: Definitely.

pages/common/ssh-keygen.md
aduzsardi: i don't know about that , i'd say it only adds confusion **rsa** is an actual key type and not a place holder although it can be some other type as well, like **ecdsa** or **_dsa_** (not used anymore) same for the bit length 

as for the -C (comment part) it doesn't need to be an email, the option can be omitted entirely , than it will default to _current_user_name@current_host_name_ or can be any string like  _devkey_

pages/common/ssh-keygen.md
aduzsardi: yes, all the path is the argument and it defaults to `~/.ssh/id_rsa` if not specified  

most of the operations that require an -f (file path) default to  `~/.ssh/id_rsa` or  `~/.ssh/id_rsa.pub`

pages/common/ssh-keygen.md
marchersimon: @aduzsardi the token syntax is not just for placeholders, but for anything which can be changed by the user whithout  changing what the command does.

> as for the -C (comment part) it doesn't need to be an email, the option can be omitted entirely , then it will default to _current_user_name@current_host_name_ or can be any string like _devkey_

I think we shouldn't completely remove `email` in this case, since it's used most of the time for this. How about
```suggestion
`ssh-keygen -t rsa -b 4096 -C "{{comment|email}}"`
```

CleanMachine1:
```suggestion
`ssh-keygen -t {{dsa|ecdsa|ed25519|rsa}}-b {{4096}} -C "{{email|email}}"`
```

marchersimon:
@aduzsardi could you please apply [this](https://github.com/tldr-pages/tldr/pull/6213#discussion_r667521150) and [this](https://github.com/tldr-pages/tldr/pull/6213#discussion_r668213108) suggestion?

aduzsardi:
Hey , i had some health issues ... sorry about that
@marchersimon i'll apply those changes , thank you

CleanMachine1 (approved):

pages/common/ssh-keygen.md
```diff
@@ -8,28 +8,28 @@
...
+`ssh-keygen -t {{ed25519}} -a {{100}}`
 
 - Generate an RSA 4096 bit key with email as a comment:
 
-`ssh-keygen -t rsa -b 4096 -C "{{email}}"`
-
-- Retrieve the key fingerprint from a host (useful for confirming the authenticity of the host when first connecting to it via SSH):
-
-`ssh-keygen -l -F {{remote_host}}`
+`ssh-keygen -t {{dsa|ecdsa|ed25519|rsa}} -b {{4096}} -C "{{comment|email}}"`
```
navarroaxel, line 19: This example is not in the DE page.

marchersimon:
> Hey , i had some health issues ... sorry about that
> @marchersimon i'll apply those changes , thank you

No problem at all. I'm glad you're better now

pages/common/ssh-keygen.md
marchersimon: It is there, it's just missing all the other options. I'll make a suggestion for that

pages.de/common/ssh-keygen.md
```diff
@@ -8,28 +8,28 @@
...
+`ssh-keygen -t {{ed25519}} -a {{100}}`
 
 - Generiere ein 4096 Bit langen RSA Schlüssel-Paar mit der Email im Kommentarfeld:
 
-`ssh-keygen -t rsa -b 4096 -C "{{email}}"`
-
-- Rufe den Schlüssel-Fingerabdruck von einem Server ab (hilfreich um die Authentizität eines Servers beim ersten Verbinden zu überprüfen):
-
-`ssh-keygen -l -F {{externer_server}}`
+`ssh-keygen -t {{rsa}} -b {{4096}} -C "{{email}}"`
```
marchersimon, line 19: To keep it in sync with the english page:
```suggestion
`ssh-keygen -t {{dsa|ecdsa|ed25519|rsa}} -b {{4096}} -C "{{kommentar|email}}""`
```

pages.de/common/ssh-keygen.md
```diff
@@ -8,28 +8,28 @@
...
+`ssh-keygen -t {{ed25519}} -a {{100}}`
 
 - Generiere ein 4096 Bit langen RSA Schlüssel-Paar mit der Email im Kommentarfeld:
 
-`ssh-keygen -t rsa -b 4096 -C "{{email}}"`
-
-- Rufe den Schlüssel-Fingerabdruck von einem Server ab (hilfreich um die Authentizität eines Servers beim ersten Verbinden zu überprüfen):
-
-`ssh-keygen -l -F {{externer_server}}`
+`ssh-keygen -t {{dsa|ecdsa|ed25519|rsa}} -b {{4096}} -C "{{kommentar|email}}""`
```
marchersimon, line 19: Oh, sorry, I added an extra quotation mark in my suggestion
```suggestion
`ssh-keygen -t {{dsa|ecdsa|ed25519|rsa}} -b {{4096}} -C "{{kommentar|email}}"`
```

CleanMachine1:
@aduzsardi Could you please revisit this PR, whenever you are available!

aduzsardi:
Hi @CleanMachine1 ,
Updated the PR with latest suggestions/recommendations

CleanMachine1:
@marchersimon @navarroaxel Please review

navarroaxel (approved):

marchersimon (approved):
Thanks @aduzsardi

PR #6215, opened 2021-07-10
scripts: add SPDX-License-Identifier in files

CleanMachine1:


CleanMachine1:
I am unaware if this is correct but in the LICENSE main file it says all scripts are MIT, however if anyone wants to use them, seeing the license here would be quicker, since they may just assume its under creative commons

sbrl (approved):
Don't see any harm in this. If one reads the main LICENSE file it give this information already though.

marchersimon (approved):

marchersimon:
Is there any specifc reason we use the MIT Licence instead of GPL for example?

CleanMachine1:
:shrug: I have no clue, if you are asking me, it just what is in the main license in the root of the repo

waldyrious:
I understand the impetus to put the license information closer to the script files. But instead of duplicating the license into the `scripts/` directory, I'd actually suggest going a step further, and adding a [SPDX license identifier comment](https://spdx.github.io/spdx-spec/appendix-V-using-SPDX-short-identifiers-in-source-files/) in the files themselves. It would be a comment at the top of the file, like this:

```
// SPDX-License-Identifier: MIT
```

With this, we would have the license info as close as it gets to the code, and we wouldn't need to duplicate the license file, which IMO could be a bit confusing.

> Is there any specifc reason we use the MIT Licence instead of GPL for example?

GPL is more protective than MIT, so switching to it would mean adding additional restrictions that the original contributors didn't sign up for. Legally, we could do it for files whose contributors have signed our CLA, but it would be a somewhat fishy thing to do.

CleanMachine1:
If everyone agrees with that, i'll edit the files to include a comment

CleanMachine1:
Please review, I think we are finished here.

scripts/LICENSE
```diff
@@ -0,0 +1,9 @@
+The MIT License (MIT)
```
navarroaxel, line 1: I think the idea of @waldyrious is not add this file, so remove it from the PR.

CleanMachine1:
Can we merge now?

navarroaxel (approved):

waldyrious (approved):
LGTM, thanks for taking the initiative @CleanMachine1!

bl-ue:
In case there comes any future confusion, I hereby make known that the script that I wrote [`set-more-info-link.py`](https://github.com/tldr-pages/tldr/blob/main/scripts/set-more-info-link.py), added in #5654, is part tldr-pages project, and subject to any license restrictions therefrom. (Not that it wasn't already case; it was, but I want to formally acknowledge that I understand this, in case someone becomes paranoid in the future.)

PR #6234, opened 2021-07-15
musescore: add German translation

marchersimon:


navarroaxel (approved):

CleanMachine1 (approved):
What should I even be looking out for?

I cannot translate this without the help of something like google translate.

So apart from more information, what should I look at

marchersimon:
Typically, when I review pages in other languages, I check the link, if all `.` and `:` are there and sometimes if the examples match the English page. 
But the reason I requested your review was just that we need two approvals to merge.

PR #6237, opened 2021-07-15
vim: update description

CleanMachine1:

This PR may be a little controversial, so I don't really care if its closed

### DIFF
--- a/pages/common/vim.md
+++ b/pages/common/vim.md
@@ -1,7 +1,7 @@
 # vim
 
 > Vim (Vi IMproved), a command-line text editor, provides several modes for different kinds of text manipulation.
-> Pressing `i` enters edit mode. `<Esc>` goes back to normal mode, which doesn't allow regular text insertion.
+> Pressing `i` enters insert mode. `<Esc>` enables the use of Vim commands.
 > More information: <https://www.vim.org>.
 
 - Open a file:
--- a/pages/common/vim.md
+++ b/pages/common/vim.md
@@ -1,7 +1,7 @@
 # vim
 
 > Vim (Vi IMproved), a command-line text editor, provides several modes for different kinds of text manipulation.
-> Pressing `i` enters insert mode. `<Esc>` enables the use of Vim commands.
+> Pressing `i` enters insert mode. `<Esc>` enters normal mode, which enables the use of Vim commands.
 > More information: <https://www.vim.org>.
 
 - Open a file:


marchersimon (approved):
👍🏽

pages/common/vim.md
```diff
@@ -1,7 +1,7 @@
 # vim
 
 > Vim (Vi IMproved), a command-line text editor, provides several modes for different kinds of text manipulation.
-> Pressing `i` enters edit mode. `<Esc>` goes back to normal mode, which doesn't allow regular text insertion.
+> Pressing `i` enters insert mode. `<Esc>` enables the use of Vim commands.
```
mfrw, line 4: I think we should mention `normal mode`:
```suggestion
> Pressing `i` enters insert mode. `<Esc>` enters normal mode, which enables the use of Vim commands.
```

CleanMachine1: Is it "normal mode" though, I am not a very experienced VIM user, my knowledge goes as far as knowing what ZZ `:x` and saving commands.

I only know VIM through installing Arch and needing to learn some basics for basic file editing.

Is normal mode correct terminology, I'd call it "command mode"

CleanMachine1: @marchersimon @navarroaxel

sbrl: I would suggest that we should use the official terminology here. Are there any official reference materials that we can look at?

CleanMachine1: Existing as in my edit or before the PR

marchersimon: I think the old description might be a bit problematic, because it sounds like you're still in `insert mode`, but pressing `<ESC>` enables you to use Vim commands too. I don't see a problem with the new one.

CleanMachine1: @mfrw

marchersimon: I'd say we can apply the suggestion above.

CleanMachine1:
```suggestion
> Pressing `i` enters insert mode. `<Esc>` enters command mode.
```

To follow the syntax from the other line before it

marchersimon: I actually prefer the longer variant, because it doesn't sound so monotonous.

mfrw: > Is it "normal mode" though, I am not a very experienced VIM user, my knowledge goes as far as knowing what ZZ `:x` and saving commands.
> 
> I only know VIM through installing Arch and needing to learn some basics for basic file editing.
> 
> Is normal mode correct terminology, I'd call it "command mode"

Yep, its officially [normal](http://vimdoc.sourceforge.net/htmldoc/intro.html#vim-modes-intro) mode

mfrw (approved):
LGTM

mfrw:
Should we hold off merging till the #6250 is also on parity with the current one ?

marchersimon:
I don't think so

PR #6240, opened 2021-07-16
restic: fix restore example

peterbabic:
The restore command, at least as of the current stable version v0.12.0
requires --target to specify where the restored data should go. With the
urrent command here, meaning without the --target, there is this error:

> Fatal: more than one snapshot ID specified

I have also included a latest word that automatically picks the latest
tag without needing the actual snapshot ID.

### DIFF
--- a/pages/common/restic.md
+++ b/pages/common/restic.md
@@ -21,7 +21,7 @@
 
 - Restore a specific path from a specific backup to a target directory:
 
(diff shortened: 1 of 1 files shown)


navarroaxel (approved):

CleanMachine1 (approved):

marchersimon (approved):
Thanks @peterbabic

PR #6249, opened 2021-07-19
pwsh: add page

SethFalco:

### DIFF
--- a/pages/common/pwsh.md
+++ b/pages/common/pwsh.md
@@ -1,7 +1,7 @@
 # pwsh
 
 > PowerShell Core is a cross-platform automation and configuration tool/framework.
-> More information: <https://docs.microsoft.com/en-us/powershell/>.
+> More information: <https://docs.microsoft.com/powershell/>.
 
 - Start an instance of PowerShell:
 
--- a/pages/common/pwsh.md
+++ b/pages/common/pwsh.md
@@ -1,0 +1,20 @@
+# pwsh
+
+> PowerShell Core is a cross-platform automation and configuration tool/framework.
+> More information: <https://docs.microsoft.com/en-us/powershell/>.
+
+- Start an instance of PowerShell:
+
+`pwsh`
+
+- Execute a script and then exit:
+
+`pwsh -File {{path/to/file.ps1}}`
+
+- Set the execution policy for the current session:
+
+`pwsh -ExecutionPolicy {{AllSigned|Bypass|Default|RemoteSigned|Restricted|Undefined|Unrestricted}}`
+
+- Execute a command and then exit:
+
+`pwsh -Command {{command}}`


pages/common/pwsh.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Start an instance of PowerShell:
+
+`pwsh`
+
+- Run the specified script:
+
+`pwsh -File {{path/to/file.ps1}}`
+
+- Set the execution policy for the current session:
```
marchersimon, line 14: Is there maybe an easier description for `execution policy?`

pages/common/pwsh.md
```diff
@@ -0,0 +1,20 @@
...
+
+`pwsh -ExecutionPolicy {{AllSigned|Bypass|Default|RemoteSigned|Restricted|Undefined|Unrestricted}}`
+
+- Run the specified command as though it was typed into PowerShell:
```
marchersimon, line 18: In almost all shell pages, we use `Execute a command and then exit`. Do you think this would fit here as well?

pages/common/pwsh.md
marchersimon: And with `current` do you mean the session you're executing this command in or the session which will be started by this command?

pages/common/pwsh.md
SethFalco: Sounds fair to me, I'll update it.

pages/common/pwsh.md
SethFalco: That's a very good question. 🤔 

```
-ExecutionPolicy
    Sets the default execution policy for the current session and saves it in the $env:PSExecutionPolicyPreference environment variable. This  pa‐
    rameter does not change the PowerShell execution policy that is set in the registry.
```

I'm testing it a bit, but whenever I do `Get-ExecutionPolicy` is always returns `Unrestricted`. 🤔

marchersimon: Well, at least it's not worse than the official docs

marchersimon (approved):
Thanks @SethFalco

pages/common/pwsh.md
```diff
@@ -0,0 +1,20 @@
+# pwsh
+
+> PowerShell Core is a cross-platform automation and configuration tool/framework.
+> More information: <https://docs.microsoft.com/en-us/powershell/>.
```
navarroaxel, line 4: We don't include the i18n for MSFT pages.
```suggestion
> More information: <https://docs.microsoft.com/powershell/>.
```

navarroaxel (changes requested):

pages/common/pwsh.md
sbrl: It's probably best to stick with  the official language here, even if it doesn't make a whole lot of sense I guess?

navarroaxel (approved):

CleanMachine1 (approved):

Waples (approved):

PR #6251, opened 2021-07-19
gsettings: add page

yump:
Examples for common uses of gsettings.  Uncommon uses, involving relocatable schemas and specifying their paths, are not included because I haven't run into them and do not understand them well enough.

### DIFF
--- a/pages/linux/gsettings.md
+++ b/pages/linux/gsettings.md
@@ -19,7 +19,7 @@
 
 `gsettings reset {{org.example.schema}} {{example-key}}`
 
-- Print the value of a key, or schema default if not set (unlike `dconf read`):
+- Print the value of a key, or the schema-provided default if the key has not been set in dconf:
 
 `gsetings get {{org.example.schema}} {{example-key}}`
 
--- a/pages/linux/gsettings.md
+++ b/pages/linux/gsettings.md
@@ -11,7 +11,7 @@
 
 `gsettings list-recursively {{org.example.schema}}`
 
-- Set the value of a key. Fails if key doesn't exist or value is outside range:
+- Set the value of a key. Fails if the key doesn't exist or the value is out of range:
 
 `gsettings set {{org.example.schema}} {{example-key}} {{value}}`
 
@@ -23,7 +23,7 @@
 
 `gsetings get {{org.example.schema}} {{example-key}}`
 
-- Display schema-allowed values for a key (helpful with enum keys, less so for others):
+- Display schema-allowed values for a key (helpful with enum keys):
 
 `gsettings range {{org.example.schema}} {{example-key}}`
 
--- a/pages/linux/gsettings.md
+++ b/pages/linux/gsettings.md
@@ -7,7 +7,7 @@
 
 `gsettings set {{org.example.schema}} {{example-key}} {{value}}`
 
-- Print the value of a key, or the schema-provided default if the key has not been set in dconf:
+- Print the value of a key or the schema-provided default if the key has not been set in `dconf`:
 
 `gsetings get {{org.example.schema}} {{example-key}}`
 
@@ -27,6 +27,6 @@
 
 `gsettings range {{org.example.schema}} {{example-key}}`
 
-- Display the human-readable description for a key:
+- Display the human-readable description of a key:
 
 `gsettings describe {{org.example.schema}} {{example-key}}`
--- a/pages/linux/gsettings.md
+++ b/pages/linux/gsettings.md
@@ -3,6 +3,18 @@
 > Query and modify dconf settings with schema validation.
 > More information: <https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/using_the_desktop_environment_in_rhel_8/configuring-gnome-at-low-level_using-the-desktop-environment-in-rhel-8#using-gsettings-command_configuring-gnome-at-low-level>.
 
+- Set the value of a key. Fails if the key doesn't exist or the value is out of range:
+
+`gsettings set {{org.example.schema}} {{example-key}} {{value}}`
+
+- Print the value of a key, or the schema-provided default if the key has not been set in dconf:
+
+`gsetings get {{org.example.schema}} {{example-key}}`
+
+- Un-set a key, so that its schema default value will be used:
+
+`gsettings reset {{org.example.schema}} {{example-key}}`
+
 - Display all (non-relocatable) schemas, keys, and values:
 
 `gsettings list-recursively`
@@ -10,18 +22,6 @@
 - Display all keys and values (default if not set) from one schema:
 
 `gsettings list-recursively {{org.example.schema}}`
-
-- Set the value of a key. Fails if the key doesn't exist or the value is out of range:
-
-`gsettings set {{org.example.schema}} {{example-key}} {{value}}`
-
-- Un-set a key, so that its schema default value will be used:
-
-`gsettings reset {{org.example.schema}} {{example-key}}`
-
-- Print the value of a key, or the schema-provided default if the key has not been set in dconf:
-
-`gsetings get {{org.example.schema}} {{example-key}}`
 
 - Display schema-allowed values for a key (helpful with enum keys):
 
--- a/pages/linux/gsettings.md
+++ b/pages/linux/gsettings.md
@@ -11,7 +11,7 @@
 
 `gsetings get {{org.example.schema}} {{example-key}}`
 
-- Un-set a key, so that its schema default value will be used:
+- Unset a key, so that its schema default value will be used:
 
 `gsettings reset {{org.example.schema}} {{example-key}}`
 
--- a/pages/linux/gsettings.md
+++ b/pages/linux/gsettings.md
@@ -1,0 +1,32 @@
+# gsettings
+
+> Query and modify dconf settings with schema validation.
+> More information: <https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/using_the_desktop_environment_in_rhel_8/configuring-gnome-at-low-level_using-the-desktop-environment-in-rhel-8#using-gsettings-command_configuring-gnome-at-low-level>.
+
+- Display all (non-relocatable) schemas, keys, and values:
+
+`gsettings list-recursively`
+
+- Display all keys and values (default if not set) from one schema:
+
+`gsettings list-recursively {{org.example.schema}}`
+
+- Set the value of a key. Fails if key doesn't exist or value is outside range:
+
+`gsettings set {{org.example.schema}} {{example-key}} {{value}}`
+
+- Un-set a key, so that its schema default value will be used:
+
+`gsettings reset {{org.example.schema}} {{example-key}}`
+
+- Print the value of a key, or schema default if not set (unlike `dconf read`):
+
+`gsetings get {{org.example.schema}} {{example-key}}`
+
+- Display schema-allowed values for a key (helpful with enum keys, less so for others):
+
+`gsettings range {{org.example.schema}} {{example-key}}`
+
+- Display the human-readable description for a key:
+
+`gsettings describe {{org.example.schema}} {{example-key}}`


CleanMachine1:
I can see quite a lot of basic errors which every first time contributor makes.

I'll try and work through it line by line. The more active you are, the quicker this will be since __you__ apply the edits and changes

pages/linux/gsettings.md
```diff
@@ -0,0 +1,32 @@
+# gsettings
+
+> query and modify dconf settings with schema validation
```
CleanMachine1, line 3: What in the world is a schema validation, other than https://duckduckgo.com/?t=ffab&q=schema+validation 

Is that the average user will understand, because I don't think so.

pages/linux/gsettings.md
```diff
@@ -0,0 +1,32 @@
...
+> More information: <https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/using_the_desktop_environment_in_rhel_8/configuring-gnome-at-low-level_using-the-desktop-environment-in-rhel-8#using-gsettings-command_configuring-gnome-at-low-level>.
```
CleanMachine1, line 4: This link is way to long I feel.

And from what I see, maybe we should use the manned.org link

pages/linux/gsettings.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Dump all (non-relocatable) schemas, keys, and values:
```
CleanMachine1, line 6: By `dump` what do you mean, does it give complete raw data or is it more 'sincere' in how it gives the info

We normally use Display or Print

pages/linux/gsettings.md
yump: `gsettings set` will fail if you try to write a key that shouldn't exist, or supply a value that isn't in the range allowed by the schema. `dconf write` doesn't know about schemas and will let you write anything.

pages/linux/gsettings.md
yump: It is unfortunately very long, but the Redhat documentation does contain examples and links to learn more about the workings of dconf and gsettings.  Also, I feel that anyone who can `tldr gsettings` can just as well `man gsettings` on their own machine.  And may in fact have just *come* from the manpage.

pages/linux/gsettings.md
yump: I have made this change, using "display" where the output is more useful for human consumption, and "print" where the output is directly machine-readable and probably less likely to change and break one's scripts in the future.

yump: It is indeed not entirely complete (doesn't contain the output of `range` or `describe`, for example), and not importable by anything AFAIK. But it is useful for piping to grep if you forget the exact name of a schema or key.

yump:
Do you think it would be appropriate to add a verbatim example? like:

```
- Enable subpixel antialiased text rendering:

`gsettings set org.gnome.settings-daemon.plugins.xsettings antialiasing rgba`
```

or perhaps:

```
- Make caps lock work as an additional ctrl key

`gsettings set org.gnome.desktop.input-sources xkb-options "['ctrl:nocaps']"`
```

7 examples is already pushing it, and it comes close to running afoul of "write every example in a way that they cannot be unmindfully copy-pasted".

But the reason I looked for a tldr for gsettings in the first place was, "what is a VALUE supposed to look like anyway? what's a 'serialized GVariant'?"

pages/linux/gsettings.md
yump: I extended the `gsettings set` a little to mention this.

CleanMachine1: I don't think its required or could just be shortened to `valid input` 

cc @SethFalco @navarroaxel

pages/linux/gsettings.md
CleanMachine1: I'll look for an official site rather than a guide, if you can find one though since you know more about the topic, that would be better

pages/linux/gsettings.md
SethFalco: I think most developers, sysadmins, and power users will probably understand or infer it.
It'd probably be fair to assume others might have to look it up.

I have no opinion on if it should be changed, though. Subjectively, I like it. ^-^'

Both are perfectly valid options, they're just different levels of explicitness:
`valid input` - Input is being validated _somehow_.
`schema validation` - Input is being validated against a schema. (A data model definition, like [JSON Schema](https://json-schema.org/) or [XSD](https://en.wikipedia.org/wiki/XML_Schema_(W3C)).)

pages/linux/gsettings.md
```diff
@@ -0,0 +1,32 @@
+# gsettings
+
+> Query and modify dconf settings with schema validation.
```
marchersimon, line 3: I'm not sure if we even need to be that specific. Reading that it's not obviously clear what it does (in my opinion). How about just:
> GSettings configuration tool

pages/linux/gsettings.md
```diff
@@ -0,0 +1,32 @@
...
+> More information: <https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/using_the_desktop_environment_in_rhel_8/configuring-gnome-at-low-level_using-the-desktop-environment-in-rhel-8#using-gsettings-command_configuring-gnome-at-low-level>.
+
+- Display all (non-relocatable) schemas, keys, and values:
```
marchersimon, line 6: Is there maybe an easier term or a way to describe `non-relocatable`?

pages/linux/gsettings.md
```diff
@@ -0,0 +1,32 @@
...
+
+`gsettings list-recursively`
+
+- Display all keys and values (default if not set) from one schema:
```
marchersimon, line 10: What does the `default` mean here? Is this behaviour default or does it use a default schema?

pages/linux/gsettings.md
```diff
@@ -0,0 +1,32 @@
...
+
+`gsettings list-recursively {{org.example.schema}}`
+
+- Set the value of a key. Fails if key doesn't exist or value is outside range:
```
marchersimon, line 14:
```suggestion
- Set the value of a key. Fails if the key doesn't exist or the value is out of range:
```

pages/linux/gsettings.md
```diff
@@ -0,0 +1,32 @@
...
+
+`gsettings set {{org.example.schema}} {{example-key}} {{value}}`
```
marchersimon, line 16: Asking fellow maintainers: Should we use `example_key` here like we do everywhere or should we use the formatting style of a gsettings key (e.g. `always-show-log-out`)?

pages/linux/gsettings.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Un-set a key, so that its schema default value will be used:
+
+`gsettings reset {{org.example.schema}} {{example-key}}`
+
+- Print the value of a key, or schema default if not set (unlike `dconf read`):
```
marchersimon, line 22: Not sure about this comma here

pages/linux/gsettings.md
```diff
@@ -0,0 +1,32 @@
...
+
+`gsetings get {{org.example.schema}} {{example-key}}`
+
+- Display schema-allowed values for a key (helpful with enum keys, less so for others):
```
marchersimon, line 26:
```suggestion
- Display schema-allowed values for a key (helpful with enum keys):
```
Not sure this is needed. Saying that something's useful for one thing implies that it's not as useful for others.

pages/linux/gsettings.md
marchersimon: It's unfortunate that it's so long, but it seems quite good and far more usable than the official GNOME documentation or the man page.

pages/linux/gsettings.md
marchersimon: (https://github.com/tldr-pages/tldr/pull/6251#discussion_r673337616)

marchersimon:
> Do you think it would be appropriate to add a verbatim example? like:
> 
> ```
> - Enable subpixel antialiased text rendering:
> 
> `gsettings set org.gnome.settings-daemon.plugins.xsettings antialiasing rgba`
> ```
> 
> or perhaps:
> 
> ```
> - Set keyboard to make caps lock work as ctrl, and right alt as compose:
> 
> `gsettings set org.gnome.desktop.input-sources xkb-options "['ctrl:nocaps', 'compose:ralt']"`
> ```
> 
> 7 examples is already pushing it, and it comes close to running afoul of "write every example in a way that they cannot be unmindfully copy-pasted".
> 
> But the reason I looked for a tldr for gsettings in the first place was, "what is a VALUE supposed to look like anyway? what's a 'serialized GVariant'?"

We **could** add one example, but would that bring any benefit? I think the examples we have now pretty much descibe how the command has to look like. If we add another discrete example it should be one you use all of the time.

pages/linux/gsettings.md
yump: One option might be the developer documentation for GSettings itself, which is part of GLib.  The url is not overlong: https://developer.gnome.org/gio/stable/settings.html.  It doesn't really talk about the command line tool `gsettings` at all, but AFAICT, `gsettings` is a fairly thin wrapper around a few of the functions there, and if you do a deep-dive on it you'll probably get a better understanding than the man page.

pages/linux/gsettings.md
yump: The default value for that key.

AFAICT, GSettings uses a dconf database backend that only stores changes from the default values specified in the schema definition.  Schema definitions are found in `/usr/share/glib-2.0/schemas`.

So:

```
$ gsettings reset org.gnome.settings-daemon.plugins.xsettings hinting
$ gsettings get org.gnome.settings-daemon.plugins.xsettings hinting
'slight'
$ dconf read /org/gnome/settings-daemon/plugins/xsettings/hinting
$ # (no output)
```

But if we explicitly set that key, even to one that matches the default...

```
$ gsettings set org.gnome.settings-daemon.plugins.xsettings hinting slight
$ gsettings get org.gnome.settings-daemon.plugins.xsettings hinting
'slight'
$ dconf read /org/gnome/settings-daemon/plugins/xsettings/hinting
'slight'
```

It is stored in the dconf database.

(And incidentally, we have triggered [a very old cairo bug](https://gitlab.freedesktop.org/cairo/cairo/-/issues/23).)

pages/linux/gsettings.md
yump: What I was trying to draw attention to is that `list-recursively` will not actually show *everything*.

What [the documentation](https://developer.gnome.org/gio/stable/GSettings.html) has to say about relocatable schemas is:

>A relocatable schema is one with no path attribute specified on its <schema> element.

[...]

>For example, a relocatable schema could be used to store geometry information for different windows in an application. If the schema ID was org.foo.MyApp.Window, it could be instantiated for paths /org/foo/MyApp/main/, /org/foo/MyApp/document-1/, /org/foo/MyApp/document-2/, etc.

[And elsewhere:](https://developer.gnome.org/gio/stable/gio-GSettingsSchema-GSettingsSchemaSource.html#g-settings-schema-get-path)

>Schemas may be single-instance or relocatable. Single-instance schemas correspond to exactly one set of keys in the backend database: those located at the path returned by this function.

>Relocatable schemas can be referenced by other schemas and can therefore describe multiple sets of keys at different locations. For relocatable schemas, this function will return `NULL`.

The main distinction seems to be that in order to operate on relocatable schemas, you need to know the GSettings schema id AND the dconf path.  `gsettings list-relocatable-schemas` can tell you the ids, but IDK how to get the paths where those schemas might be used.  [According to the discussion here](https://github.com/mate-desktop/mate-panel/issues/675), the `dconf-editor` GUI tool had a bear of a time trying to map the other way (from path to schema id).

It seems like most uses of the `gsettings` tool are, "paste this in your terminal to change such and such setting not exposed in the GUI," or new computer setup automation, which largely don't involve relocatable schemas.

pages/linux/gsettings.md
yump: For what it's worth, [the docs](https://developer.gnome.org/gio/stable/GSettings.html) do specifically say:

>Key names are restricted to lowercase characters, numbers and '-'.

yump:
I looked it up and supposedly the Github UI lets you squash commits on merge if you want it to be a single commit, so I stopped squashing locally because it's error-prone and breaks the discussion history.

pages/linux/gsettings.md
yump: In most recent version, I have re-ordered the examples to put the subcommands with scripting uses, `set`/`get`/`reset`, come first and the "exploratory" ones, including `list-recursively`, are farther down.  This makes the stuff about non-relocatable vs relocatable schemas less prominent, so hopefully users are less likely to get the impression they need to understand what that means to use the tool.

But I do think it should be there somewhere, as a warning that `list-recursively` does not display everything.

pages/linux/gsettings.md
marchersimon: That's why I'm not sure. Independent of what the command uses, we generally use snake_case for arguments.

marchersimon:
We squash PRs before merging by default

pages/linux/gsettings.md
marchersimon: I don't think that page would be helpful for anyone wanting to use the command, but the `redhat` page still links to the official docs.

pages/linux/gsettings.md
```diff
@@ -0,0 +1,32 @@
+# gsettings
+
+> Query and modify dconf settings with schema validation.
+> More information: <https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/using_the_desktop_environment_in_rhel_8/configuring-gnome-at-low-level_using-the-desktop-environment-in-rhel-8#using-gsettings-command_configuring-gnome-at-low-level>.
+
+- Set the value of a key. Fails if the key doesn't exist or the value is out of range:
+
+`gsettings set {{org.example.schema}} {{example-key}} {{value}}`
+
+- Print the value of a key, or the schema-provided default if the key has not been set in dconf:
```
marchersimon, line 10:
```suggestion
- Print the value of a key or the schema-provided default if the key has not been set in `dconf`:
```

pages/linux/gsettings.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Display all keys and values (default if not set) from one schema:
+
+`gsettings list-recursively {{org.example.schema}}`
+
+- Display schema-allowed values for a key (helpful with enum keys):
+
+`gsettings range {{org.example.schema}} {{example-key}}`
+
+- Display the human-readable description for a key:
```
marchersimon, line 30:
```suggestion
- Display the human-readable description of a key:
```

marchersimon (approved):
Looks good, thanks @yump

CleanMachine1 (approved):

navarroaxel (approved):

pages/linux/gsettings.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Set the value of a key. Fails if the key doesn't exist or the value is out of range:
+
+`gsettings set {{org.example.schema}} {{example-key}} {{value}}`
+
+- Print the value of a key or the schema-provided default if the key has not been set in `dconf`:
+
+`gsetings get {{org.example.schema}} {{example-key}}`
+
+- Un-set a key, so that its schema default value will be used:
```
SethFalco, line 14: Sorry to be a pain!

```suggestion
- Unset a key, so that its schema default value will be used:
```

* https://www.collinsdictionary.com/dictionary/english/unset
* https://www.lexico.com/definition/unset
* https://www.merriam-webster.com/dictionary/unset

"Unset" is one word! Other than that I think this PR looks great!

SethFalco (approved):

PR #6254, opened 2021-07-21
nettop: add page

nyameteor:

### DIFF
--- a/pages/osx/nettop.md
+++ b/pages/osx/nettop.md
@@ -1,7 +1,7 @@
 # nettop
 
 > Display updated information about the network.
-More information: <N/A>.
+> More information: <N/A>.
 
 - Monitor TCP and UDP sockets from all interfaces:
 
--- a/pages/osx/nettop.md
+++ b/pages/osx/nettop.md
@@ -14,7 +14,7 @@
 
 `nettop -p {{process_id|process_name}}`
 
-- Display summary only for per-process:
+- Display a per-process summary:
 
 `nettop -P`
 
@@ -22,7 +22,7 @@
 
 `nettop -l {{10}}`
 
-- Monitor in delta mode, updating every 5 seconds:
+- Monitor changes every 5 seconds:
 
 `nettop -d -s {{5}}`
 
--- a/pages/osx/nettop.md
+++ b/pages/osx/nettop.md
@@ -33,4 +33,3 @@
 - Display help message:
 
 `nettop -h`
-
--- a/pages/osx/nettop.md
+++ b/pages/osx/nettop.md
@@ -26,7 +26,7 @@
 
 `nettop -d -s {{5}}`
 
-- List interactive commands while running nettop:
+- While running nettop, list interactive commands:
 
 `h`
 
--- a/pages/osx/nettop.md
+++ b/pages/osx/nettop.md
@@ -1,6 +1,6 @@
 # nettop
 
-> Display dynamic real-time information about the network.
+> Display updated information about the network.
 More information: <N/A>.
 - Monitor TCP and UDP sockets from all interfaces:
 
@@ -18,7 +18,7 @@ More information: <N/A>.
 
 `nettop -P`
 
-- Monitor in logging mode and display 10 samples:
+- Print 10 samples of network information:
 
 `nettop -l {{10}}`
 
--- a/pages/osx/nettop.md
+++ b/pages/osx/nettop.md
@@ -1,0 +1,36 @@
+# nettop
+
+> Display dynamic real-time information about the network.
+
+- Start nettop monitoring all sockets and interfaces:
+
+`nettop`
+
+- Start nettop monitoring TCP sockets and Loopback interfaces:
+
+`nettop -m tcp -t loopback`
+
+- Start nettop monitoring specific process:
+
+`nettop -p {{process_id|process-name}}`
+
+- Start nettop displaying summary only for per-process:
+
+`nettop -P`
+
+- Start nettop in logging mode and display 10 samples:
+
+`nettop -l {{10}}`
+
+- Start nettop in delta mode, updating every 5 seconds:
+
+`nettop -d -s {{5}}`
+
+- See interactive commands while running nettop:
+
+`h`
+
+- Display help message:
+
+`nettop -h`
+
--- a/pages/osx/nettop.md
+++ b/pages/osx/nettop.md
@@ -1,7 +1,7 @@
 # nettop
 
 > Display dynamic real-time information about the network.
-
+More information: <N/A>.
 - Monitor TCP and UDP sockets from all interfaces:
 
 `nettop`
--- a/pages/osx/nettop.md
+++ b/pages/osx/nettop.md
@@ -2,6 +2,7 @@
 
 > Display updated information about the network.
 More information: <N/A>.
+
 - Monitor TCP and UDP sockets from all interfaces:
 
 `nettop`
--- a/pages/osx/nettop.md
+++ b/pages/osx/nettop.md
@@ -1,7 +1,6 @@
 # nettop
 
 > Display updated information about the network.
-> More information: <N/A>.
 
 - Monitor TCP and UDP sockets from all interfaces:
 
--- a/pages/osx/nettop.md
+++ b/pages/osx/nettop.md
@@ -12,7 +12,7 @@
 
 - Start nettop monitoring specific process:
 
-`nettop -p {{process_id|process-name}}`
+`nettop -p {{process_id|process_name}}`
 
 - Start nettop displaying summary only for per-process:
 
--- a/pages/osx/nettop.md
+++ b/pages/osx/nettop.md
@@ -2,11 +2,11 @@
 
 > Display dynamic real-time information about the network.
 
-- Start nettop monitoring all sockets and interfaces:
+- Start nettop monitoring TCP and UDP sockets from all interfaces:
 
 `nettop`
 
-- Start nettop monitoring TCP sockets and Loopback interfaces:
+- Start nettop monitoring TCP sockets from Loopback interfaces:
 
 `nettop -m tcp -t loopback`
 
--- a/pages/osx/nettop.md
+++ b/pages/osx/nettop.md
@@ -2,34 +2,34 @@
 
 > Display dynamic real-time information about the network.
 
-- Start nettop monitoring TCP and UDP sockets from all interfaces:
+- Monitor TCP and UDP sockets from all interfaces:
 
 `nettop`
 
-- Start nettop monitoring TCP sockets from Loopback interfaces:
+- Monitor TCP sockets from Loopback interfaces:
 
-`nettop -m tcp -t loopback`
+`nettop -m {{tcp}} -t {{loopback}}`
 
-- Start nettop monitoring specific process:
+- Monitor a specific process:
 
 `nettop -p {{process_id|process_name}}`
 
-- Start nettop displaying summary only for per-process:
+- Display summary only for per-process:
 
 `nettop -P`
 
-- Start nettop in logging mode and display 10 samples:
+- Monitor in logging mode and display 10 samples:
 
 `nettop -l {{10}}`
 
-- Start nettop in delta mode, updating every 5 seconds:
+- Monitor in delta mode, updating every 5 seconds:
 
 `nettop -d -s {{5}}`
 
-- See interactive commands while running nettop:
+- List interactive commands while running nettop:
 
 `h`
 
-- Display help message:
+- Display help:
 
 `nettop -h`


nyameteor:
Sorry, I can't find a suitable online man page for this command.
These online man pages are probably incomplete or outdated:
- https://www.unix.com/man-page/osx/1/nettop/ 
- https://www.manpagez.com/man/1/nettop/

CleanMachine1:
We use manned.org for our man pages

nyameteor:
> We use manned.org for our man pages

I can't find `nettop` command at manned.org, may be because it's an osx command

marchersimon:
I don't find anything myself except for <https://github.com/cs8425/NetTop>, which doesn't seem to be the same command. I think <https://www.manpagez.com/man/1/nettop/> is the best we can get.

pages/osx/nettop.md
```diff
@@ -0,0 +1,35 @@
+# nettop
+
+> Display dynamic real-time information about the network.
+
+- Start nettop monitoring TCP and UDP sockets from all interfaces:
```
marchersimon, line 5:
```suggestion
- Monitor TCP and UDP sockets from all interfaces:
```
This would make the descriptions a little shorter and less monotonous. That should  be applied everywhere.

pages/osx/nettop.md
```diff
@@ -0,0 +1,35 @@
...
+
+`nettop`
+
+- Start nettop monitoring TCP sockets from Loopback interfaces:
+
+`nettop -m tcp -t loopback`
```
marchersimon, line 11:
```suggestion
`nettop -m {{tcp}} -t {{loopback}}`
```
We use the token syntax not only for placeholders, but for anything which can be changed

pages/osx/nettop.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Start nettop monitoring specific process:
```
marchersimon, line 13:
```suggestion
- Start nettop monitoring a specific process:
```

pages/osx/nettop.md
```diff
@@ -0,0 +1,35 @@
...
+
+`nettop -p {{process_id|process_name}}`
+
+- Start nettop displaying summary only for per-process:
+
+`nettop -P`
```
marchersimon, line 19: Who thought it might be a good idea to use both `-p` and `-P` as arguments? 😆

pages/osx/nettop.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Start nettop in logging mode and display 10 samples:
+
+`nettop -l {{10}}`
+
+- Start nettop in delta mode, updating every 5 seconds:
```
marchersimon, line 25: What are `logging` and `delta` mode exactly?

pages/osx/nettop.md
```diff
@@ -0,0 +1,35 @@
...
+
+`nettop -d -s {{5}}`
+
+- See interactive commands while running nettop:
```
marchersimon, line 29:
```suggestion
- List interactive commands while running nettop:
```
Not obligatory

pages/osx/nettop.md
```diff
@@ -0,0 +1,35 @@
...
+
+`h`
+
+- Display help message:
```
marchersimon, line 33:
```suggestion
- Display help:
```
or
```suggestion
- Display a help message:
```

pages/osx/nettop.md
nyameteor: `delta` mode will show delta value between every update rather than show cumulative value.

`loggin` mode will print the raw form periodically rather than redisplay.

I think man page in https://www.manpagez.com/man/1/nettop/ is outdated so it dones't have these options.
I pasted the result of `man nettop` to here: https://pastebin.com/vqJc1et3, hope you cloud check it if it is convenient :)

pages/osx/nettop.md
CleanMachine1: we opt for the top one normally.

pages/osx/nettop.md
nyameteor: Got it, thanks 😄

pages/osx/nettop.md
CleanMachine1: Does it focus on 1 process, if not then

```suggestion
- Start nettop, monitoring specific processes:
```
would be acceptable

pages/osx/nettop.md
nyameteor: OK, I will change it

pages/osx/nettop.md
nyameteor: Yes, it focus on 1 process, I should add `a` :)

pages/osx/nettop.md
nyameteor: OK, I will choose the top one.

pages/osx/nettop.md
nyameteor: Probably the semantics of list is clearer, I will do as you say.

nyameteor:
I have modified as your suggestions. Looking forward to your next review 😆

pages/osx/nettop.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Monitor in logging mode and display 10 samples:
+
+`nettop -l {{10}}`
+
+- Monitor in delta mode, updating every 5 seconds:
+
+`nettop -d -s {{5}}`
+
+- List interactive commands while running nettop:
```
CleanMachine1, line 29:
```suggestion
- While running nettop, list interactive commands:
```

pages/osx/nettop.md
```diff
@@ -0,0 +1,35 @@
+# nettop
+
+> Display dynamic real-time information about the network.
+
```
CleanMachine1, line 4:
```suggestion
More information <>.

```

Leave this open and we will decide on a link

pages/osx/nettop.md
```diff
@@ -0,0 +1,35 @@
...
+- Monitor TCP sockets from Loopback interfaces:
+
+`nettop -m {{tcp}} -t {{loopback}}`
+
+- Monitor a specific process:
+
+`nettop -p {{process_id|process_name}}`
+
+- Display summary only for per-process:
```
CleanMachine1, line 17:
```suggestion
- Display a summary only for per-process:
```

pages/osx/nettop.md
```diff
@@ -0,0 +1,35 @@
...
+
+`nettop -P`
+
+- Monitor in logging mode and display 10 samples:
+
+`nettop -l {{10}}`
+
+- Monitor in delta mode, updating every 5 seconds:
```
CleanMachine1, line 25:
```suggestion
- Monitor in delta mode and update every 5 seconds:
```
Follow the standard above you used

pages/osx/nettop.md
marchersimon: I'm not sure if that's the case for OSX, but if you can download that package it has to exist somewhere online to browse through. These files probably contain also the manpage. Maybe that's worth a try.

CleanMachine1 (commented):
my turn!

pages/osx/nettop.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Monitor a specific process:
+
+`nettop -p {{process_id|process_name}}`
+
+- Display summary only for per-process:
+
+`nettop -P`
+
+- Monitor in logging mode and display 10 samples:
```
marchersimon, line 21: From how I understood it in the man page, this command doesn't run periodically right? To avoid using `logging mode`, how about
```suggestion
- Print 10 samples of real-time network information: 
```

pages/osx/nettop.md
marchersimon, line 25: And instead of `delta mode` we could use:
```suggestion
- Monitor changes in the network, updating every 5 seconds:
```

pages/osx/nettop.md
nyameteor: Without these two options, this command will show the detail for all processes, a long page will be displayed on screen.
With `-P` option, user can roughly monitor all processes status, while with `-p` option, user can check all connection detail belong to one process, so I think these two options might to be use separately 😄

pages/osx/nettop.md
nyameteor: It's an OSX builtin command. Sorry, according to this answer: https://stackoverflow.com/a/8670402, this command use private code, maybe we can't find the original online man page, just like `launchctl` and `hdiutil`.

pages/osx/nettop.md
nyameteor: OK, it's clearer than before, I will change it.

pages/osx/nettop.md
marchersimon: We tried to introduce 
> More information: N/A.

for pages which didn't have a link a few months ago. But it failed, since tldl-lint wanted to have `< ... >` around the link. Well, then let's introduce:

```suggestion
More information: <N/A>.
```

pages/osx/nettop.md
marchersimon: https://github.com/tldr-pages/tldr/pull/6254/files#r673909261

pages/osx/nettop.md
marchersimon: You could also use `Display` or `Show` but `See` sounds a bit weird in my opinion.

pages/osx/nettop.md
marchersimon: Another suggestion:
```suggestion
- Display a per-process summary only:
```

pages/osx/nettop.md
marchersimon: I wouldn't say that's necessary. A bit diversity in descriptions makes the more interesting.

pages/osx/nettop.md
nyameteor: OK, I see it 😄

pages/osx/nettop.md
CleanMachine1: No, I am not suggesting no link. I am just leaving it here to be discussed and the gap to be filled.

https://www.unix.com/man-page/osx/1/nettop/ 
https://apps.apple.com/us/app/nettop/id1212061421?mt=12

Are these any good?

pages/osx/nettop.md
CleanMachine1: That sounds weird

pages/osx/nettop.md
marchersimon: No, I was talking about that they are very easy to confuse, since they look very similar. 😄

pages/osx/nettop.md
marchersimon: Actually the second one looks pretty good. It doesn't give you information on the link, but after you downloaded it you have access to the man page

pages/osx/nettop.md
nyameteor: Sorry for leaving for a while because of other work...

How about `Monitor relative changes between every 5 seconds` ?
`delta` mode will show relative change, and default mode will show cumulative change.

I feel `delta mode` is a little bit hard to represent in a short words 😂

pages/osx/nettop.md
nyameteor: This `nettop` is an OSX builtin command, and preinstalled in system. OSX  user can use  `man nettop` directly.

I think the second link is another different application, but have the same name `nettop` 😂

CleanMachine1: We should just use the unix link

nyameteor: Compare to `man nettop`, the unix link seems to be outdated or incomplete , this is the result of `man nettop`  in osx: https://pastebin.com/vqJc1et3

pages/osx/nettop.md
CleanMachine1: No its not, but it makes the page a little more uniformed

CleanMachine1: The descriptions can be a little longer, as long as it has a purpose like displaying relevant information, keeping it short is nicer for the user but if it needs to be longer, it should be longer.

I wouldn't instantly correlate what delta mode is, but after thinking for around a minute, I remember it means change.

maybe `Monitor differences/changes between every 5 seconds`

pages/osx/nettop.md
nyameteor: Oh, I misunderstood before. Yes, they are really easy to confuse without man page 😅

pages/osx/nettop.md
nyameteor: Oh, I just found out that GitHub can commit suggestions directly, it's so cool. Looks like I'm still lacking in knowledge about GitHub.

pages/osx/nettop.md
nyameteor: OK, so the description could be longer to make the presentation clearer. 

I think `Monitor changes between every 5 seconds` is good.

CleanMachine1: I don't like `between`

pages/osx/nettop.md
nyameteor: I think `Display a summary for per-process` is good 😆

Maybe we don't need word _only_, I copy the word _only_ from the man page:

```plaintext
     -P       Display per-process summary only, skipping details of open con-
	      nections.  This is equivalent to selecting "Collapse All" in the
	      interactive menu.
```

nyameteor (commented):
Display a summary for per-process

pages/osx/nettop.md
nyameteor: How about `during`?

pages/osx/nettop.md
nyameteor: Yes, this command run only one time, print samples periodically, default Interval time is 1 second. I don't know if this can be called `real-time` 😂

pages/osx/nettop.md
marchersimon: We already have the `--help` example to get more information, so I think we could just use `<N/A>`. An outdated man page is worse than none at all.

pages/osx/nettop.md
marchersimon: I think `per-process` is used as an adjective here, so `a per-process summary` would be the correct way. Maybe @sbrl could help out here.

pages/osx/nettop.md
marchersimon: I jut took this part from the command description:

> Display dynamic real-time information about the network.

But we can of course just remove the `real-time`.

pages/osx/nettop.md
marchersimon: I think that's a case for @sbrl too.

pages/osx/nettop.md
nyameteor: Oh, it's my fault... I copy the words `dynamic` and `real-time` from `top` command to description, my apologies.

Maybe we should use the command description from `man nettop`: 
> Display updated information about the network

pages/osx/nettop.md
nyameteor: Yes, I agree, An outdated man page is a trap :)

pages/osx/nettop.md
nyameteor: So, maybe `Print 10 samples of network information` ?

pages/osx/nettop.md
marchersimon: Then, could you please apply my suggestion above? I'd be interested if `tldr-lint` complains about it.

nyameteor: OK, I have applied your suggestion just now. Leave the next work to `tldr-lint`.

pages/osx/nettop.md
nyameteor: OK, I have applied `Print 10 samples of network information`, 
and change command description to `Display updated information about the network`

CleanMachine1:
If you are working locally, please pull since I have committed to fix a simple error.

pages/osx/nettop.md
```diff
@@ -0,0 +1,36 @@
+# nettop
+
+> Display updated information about the network.
+> More information: <N/A>.
```
CleanMachine1, line 4:
```suggestion
> More information: <N/A>.
```

nyameteor:
> If you are working locally, please pull since I have committed to fix a simple error.

OK, I will pull before making changes. 😄

marchersimon:
> The [build](https://github.com/tldr-pages/tldr/actions/runs/1054458332) for this PR failed with the following error(s):
> 
> ```
> pages/osx/nettop.md:4: TLDR017 Information link should be surrounded with angle brackets
> ```
> 

Well I had a look at the source code and saw that it analyses if it's a valid link. I already tried to change that, but good look messing around with this regex:
``[/^(?:\s+$)/,/^(?:.*?\t+.*)/,/^(?:[^\n]$)/,/^(?:(\s*)#(\s*))/,/^(?:([\>-])(\s*))/,/^(?:([Mm]ore\s+[Ii]nfo(?:rmation)?:?\s*))/,/^(?:(<https?:\/\/[^\s\>]*>))/,/^(?:(.+?)([ ]*)((?:\r\n)|\n|\r))/,/^(?:(.+?)([\.,;!\?]?)([ ]*)((?:\r\n)|\n|\r))/,/^(?:(.+?)([\.:,;]?)([ ]*)((?:\r\n)|\n|\r))/,/^(?:`((?:\r\n)|\n|\r))/,/^(?:`)/,/^(?:\{\{([^\n\`\{\}]*)\}\}(((?:\r\n)|\n|\r)?))/,/^(?:([^\`\n]+?)\{\{)/,/^(?:([^\`\n]+?)(`[ ]*|[ ]*((?:\r\n)|\n|\r)))/,/^(?:[ ]+)/,/^(?:((?:\r\n)|\n|\r)+)/,/^(?:(.+?)[\.:]?((?:\r\n)|\n|\r))/],``

nyameteor:
> > The [build](https://github.com/tldr-pages/tldr/actions/runs/1054458332) for this PR failed with the following error(s):
> > ```
> > pages/osx/nettop.md:4: TLDR017 Information link should be surrounded with angle brackets
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Please fix the error(s) and push again.
> 
> Well I had a look at the source code and saw that it analyses if it's a valid link. I already tried to change that, but good look messing around with this regex:
> `` [/^(?:\s+$)/,/^(?:.*?\t+.*)/,/^(?:[^\n]$)/,/^(?:(\s*)#(\s*))/,/^(?:([\>-])(\s*))/,/^(?:([Mm]ore\s+[Ii]nfo(?:rmation)?:?\s*))/,/^(?:(<https?:\/\/[^\s\>]*>))/,/^(?:(.+?)([ ]*)((?:\r\n)|\n|\r))/,/^(?:(.+?)([\.,;!\?]?)([ ]*)((?:\r\n)|\n|\r))/,/^(?:(.+?)([\.:,;]?)([ ]*)((?:\r\n)|\n|\r))/,/^(?:`((?:\r\n)|\n|\r))/,/^(?:`)/,/^(?:\{\{([^\n\`\{\}]*)\}\}(((?:\r\n)|\n|\r)?))/,/^(?:([^\`\n]+?)\{\{)/,/^(?:([^\`\n]+?)(`[ ]*|[ ]*((?:\r\n)|\n|\r)))/,/^(?:[ ]+)/,/^(?:((?:\r\n)|\n|\r)+)/,/^(?:(.+?)[\.:]?((?:\r\n)|\n|\r))/], ``
> 
> I think the best would be if we'd just remove it completely for now.

Wow, this regex is quite complicated. OK, I will just remove it in next commit.

pages/osx/nettop.md
CleanMachine1: Just remove the between, it still makes complete sense

CleanMachine1:
Sorry if this has been raised, but does this command have long flags, such as -h to --help

nyameteor:
> Sorry if this has been raised, but does this command have long flags, such as -h to --help

This command only have `-h` option, and can't parse long flags.
Also `-h` option is not listed in its man page, which is weird.

pages/osx/nettop.md
nyameteor: Hello, @marchersimon, could you help out here, Do you think `Monitor changes every 5 seconds` can be applied?

CleanMachine1:
> > The [build](https://github.com/tldr-pages/tldr/actions/runs/1054458332) for this PR failed with the following error(s):
> > ```
> > pages/osx/nettop.md:4: TLDR017 Information link should be surrounded with angle brackets
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Please fix the error(s) and push again.
> 
> Well I had a look at the source code and saw that it analyses if it's a valid link. I already tried to change that, but good look messing around with this regex:
> `` [/^(?:\s+$)/,/^(?:.*?\t+.*)/,/^(?:[^\n]$)/,/^(?:(\s*)#(\s*))/,/^(?:([\>-])(\s*))/,/^(?:([Mm]ore\s+[Ii]nfo(?:rmation)?:?\s*))/,/^(?:(<https?:\/\/[^\s\>]*>))/,/^(?:(.+?)([ ]*)((?:\r\n)|\n|\r))/,/^(?:(.+?)([\.,;!\?]?)([ ]*)((?:\r\n)|\n|\r))/,/^(?:(.+?)([\.:,;]?)([ ]*)((?:\r\n)|\n|\r))/,/^(?:`((?:\r\n)|\n|\r))/,/^(?:`)/,/^(?:\{\{([^\n\`\{\}]*)\}\}(((?:\r\n)|\n|\r)?))/,/^(?:([^\`\n]+?)\{\{)/,/^(?:([^\`\n]+?)(`[ ]*|[ ]*((?:\r\n)|\n|\r)))/,/^(?:[ ]+)/,/^(?:((?:\r\n)|\n|\r)+)/,/^(?:(.+?)[\.:]?((?:\r\n)|\n|\r))/], ``
> 
> I think the best would be if we'd just remove it completely for now.

language of the gods :laughing:

nyameteor:
Hello, I just applied your suggestions:

```plaintext
-- Display summary only for per-process:
+- Display a per-process summary:

-- Monitor in delta mode, updating every 5 seconds:
+- Monitor changes every 5 seconds:
```

Looking forward to your next review 😆

pages/osx/nettop.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Monitor TCP sockets from Loopback interfaces:
+
+`nettop -m {{tcp}} -t {{loopback}}`
+
+- Monitor a specific process:
+
+`nettop -p {{process_id|process_name}}`
+
+- Display a per-process summary:
```
nyameteor, line 17: Does `display` sounds like only show once and won't refresh?

This command will monitor processes,  display a per-process summary and refresh result periodically.

CleanMachine1: No, print does though.

nyameteor: OK, I understand, thanks :)

CleanMachine1 (approved):

marchersimon (approved):
Thanks @nyameteor

nyameteor:
Thanks for your review and guidance, hope you have a great day.

CleanMachine1:
Thanks for the kind words!

PR #6261, opened 2021-07-24
ffe: add page

daniel-birket:

I'm submitting this to lint it. Let me fix any lint errors first.

### DIFF
--- a/pages/common/ffe.md
+++ b/pages/common/ffe.md
@@ -6,23 +6,23 @@
 
 - Display all input data using the specified data configuration:
 
-`ffe -c {{path/to/config}} {{path/to/input}}`
+`ffe --configuration={path/to/config.ffe}} {{path/to/input}}`
 
 - Convert an input file to an output file in a new format:
 
-`ffe -o {{path/to/output}} -c {{path/to/config}} {{path/to/input}}`
+`ffe --output={path/to/output}} -c {{path/to/config.ffe}} {{path/to/input}}`
 
-- Interpret input with a selected structure and write with a selected print format:
+- Select input structure and print format from definitions in ~/fferc config file:
 
-`ffe -s {{structure}} -p {{format}} -c {{path/to/config}} {{path/to/input}}`
+`ffe --structure={{structure}} --print={{format}} {{path/to/input}}`
 
 - Write only the selected fields:
 
-`ffe -f "{{FirstName,LastNAme,Age}}" -c {{path/to/config}} {{path/to/input}}`
+`ffe --field-list="{{FirstName,LastName,Age}}" -c {{path/to/config.ffe}} {{path/to/input}}`
 
 - Write only the records that match an expression:
 
-`ffe -e "{{LastName=Smith}}" -c {{path/to/config}} {{path/to/input}}`
+`ffe -e "{{LastName=Smith}}" -c {{path/to/config.ffe}} {{path/to/input}}`
 
 - Display help:
 
--- a/pages/common/ffe.md
+++ b/pages/common/ffe.md
@@ -18,7 +18,7 @@
 
 - Write only the selected fields:
 
-`ffe --field-list="{{First_Name,Last_Name,Age}}" -c {{path/to/config.ffe}} {{path/to/input}}`
+`ffe --field-list="{{FirstName,LastName,Age}}" -c {{path/to/config.ffe}} {{path/to/input}}`
 
 - Write only the records that match an expression:
 
--- a/pages/common/ffe.md
+++ b/pages/common/ffe.md
@@ -12,7 +12,7 @@
 
 `ffe --output={{path/to/output}} -c {{path/to/config.ffe}} {{path/to/input}}`
 
-- Select input structure and print format from definitions in ~/fferc config file:
+- Select input structure and print format from definitions in `~/.fferc` config file:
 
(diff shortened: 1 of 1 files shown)


daniel-birket:
Notes:
- I used single-letter options (-c, -o, -s, -p, -s, -e) instead of long options (--config=, --output=, etc) to reduce the line length. The {{path/to/config}} and {{path/to/output}} names keep the purpose clear, I think.
- There is no way to avoid creating a fairly complex configuration file to interpret the input file and format the output file. I have not attempted to describe that here as it is beyond tldr. Reading will be required.
- The format of the field name list is from an example. I added quotes.
- The format of the expression is one of many. I added quotes.

pages/common/ffe.md
```diff
@@ -0,0 +1,29 @@
...
+
+`ffe -o {{path/to/output}} -c {{path/to/config}} {{path/to/input}}`
+
+- Interpret input with a selected structure and write with a selected print format:
+
+`ffe -s {{structure}} -p {{format}} -c {{path/to/config}} {{path/to/input}}`
+
+- Write only the selected fields:
+
+`ffe -f "{{FirstName,LastNAme,Age}}" -c {{path/to/config}} {{path/to/input}}`
```
CleanMachine1, line 21:
```suggestion
`ffe -f "{{FirstName,LastName,Age}}" -c {{path/to/config}} {{path/to/input}}`
```

pages/common/ffe.md
```diff
@@ -0,0 +1,29 @@
+# ffe
+
+> Extract fields from a flat database file and write to another format.
+> `ffe` needs a text config file to interpret the input and format the output.
```
CleanMachine1, line 4: This line isn't needed IMO, the fact very command has a config file used, kind of gives it away

TimoSavi (commented):
Looks fine!

pages/common/ffe.md
daniel-birket: Disagree. ‘ffe” can also use a default ~/fferc config with a collection of structures and formats that you commonly use and select from. The point is that it needs _some_ config file and there is no avoiding writing one to describe your data. I will remove -c from the -s and -f examples and mention ~/fferc there to illustrate this mode better.

pages/common/ffe.md
daniel-birket: I’ll fix the case. The field names are samples matching the data not file names.

CleanMachine1: we use snake case regardless.

pages/common/ffe.md
sbrl: What about this?
```suggestion
> A configuration file is required to interpret the input and format the output.
```

pages/common/ffe.md
```diff
@@ -0,0 +1,29 @@
...
+
+`ffe --configuration={{path/to/config.ffe}} {{path/to/input}}`
+
+- Convert an input file to an output file in a new format:
+
+`ffe --output={{path/to/output}} -c {{path/to/config.ffe}} {{path/to/input}}`
+
+- Select input structure and print format from definitions in `~/.fferc` config file:
+
+`ffe --structure={{structure}} --print={{format}} {{path/to/input}}`
```
sbrl, line 17: What is "format" in this instance?

daniel-birket: Both {{structure}} and {{format}} are defined in the (not shown) configuration file.

pages/common/ffe.md
navarroaxel: @CleanMachine1 but we should respect the case used as standard in the target database (e.g. a page for MySQL), I think in this case we don't have one. In the `mongoexport` page we use `field1,field2` so the camelCase used in MongoDB is not affected.

Also, we use snake case but in lower case for placeholders. Wdyt @sbrl ?

sbrl: In this case, we should stick with the existing structure since it looks like it has specific technical meaning. There is a typo here though, which we can fix:

```suggestion
`ffe -f "{{FirstName,LastName,Age}}" -c {{path/to/config}} {{path/to/input}}`
```

If it has already been changed  then there's little point in changing it back, but this is what I would do.

pages/common/ffe.md
daniel-birket: I incorporated this.

pages/common/ffe.md
navarroaxel: @daniel-birket what case do you think is a better fit here?

daniel-birket: The sample field names in the --file-list sample were from the author's documentation. I purposely departed from the snake_case used required for tokens in your style guide because they were not tokens but a sample of the format required for the parameter. I have since changed it to Cobra_Case as First_Name,Last_Name as suggested, which will serve as well.

sbrl: In this case, let's match the author's documentation convention. This isn't the same as most other token syntax examples, since it's a specific example and not a descriptive stand-in.

daniel-birket: Ok, I'll change it back to the author's sample format.

navarroaxel (approved):

sbrl (approved):

sbrl:
Thanks, @daniel-birket!

marchersimon (approved):
Thanks @daniel-birket

PR #6263, opened 2021-07-25
yacas: add page

daniel-birket:

The primary function of this command is to start a read-eval-print-loop console session to evaluate math statements.
I have taken a guess at a tldr format for the console session examples by saying "In the console session, ..." and by including the "In>" and "Out>" prompts. Please suggest something better.

The last sample is POSIX plumbing to evaluate a yacas statement at the command line and print the result to stdout. Yacas actually has both an --execute flag and a read-from-stdin flag, -f, but they don't work as advertised and this does.

I have not explained the -p and -c flags (dumb-term and no-prompts). Together as -pc, they are effectively --batch-mode, although one must also give an input file as a parameter to exit without starting the interactive repl.

`yacas --help` isn't a flag and `man yacas` came up dry on MacOS and the read-the-docs site omits command line flags, so the only documentation of the command line flags that I have found is on the abandoned sourceforge site at http://yacas.sourceforge.net/introchapter2.html

yacas --server {{port}}& works, but is perhaps beyond the tldr level?

-- Daniel Birket

### DIFF
--- a/pages/common/yacas.md
+++ b/pages/common/yacas.md
@@ -17,7 +17,7 @@
 
 - Quit from the console session:
 
-`In> {{gquit}}`
+`In> {{quit}}`
 
 - Execute one or more yacas scripts, then exit:
 
--- a/pages/common/yacas.md
+++ b/pages/common/yacas.md
@@ -7,17 +7,17 @@
 
 `yacas`
 
-- In the `yacas` session, execute a statement:
+- While in a `yacas` session, execute a statement:
 
-`In> {{Integrate(x)Cos(x);}}`
+`{{Integrate(x)Cos(x)}};`
 
-- In the `yacas` session, display an example:
+- While in a `yacas` session, display an example:
 
-`In> {{Example();}}`
+`{{Example()}};`
 
-- Quit from the `yacas` session:
+- Quit from a `yacas` session:
 
-`In> {{quit}}`
+`{{quit}}`
 
 - Execute one or more `yacas` scripts (without terminal or prompts), then exit:
 
--- a/pages/common/yacas.md
+++ b/pages/common/yacas.md
@@ -9,15 +9,15 @@
 
 - In the console session, execute a statement:
 
-`In> {{Integrate(x)Cos(x)}};`
+`In> {{Integrate(x)Cos(x);}}`
 
 - In the console session, display an example:
 
-`In> Example();`
+`In> {{Example();}}`
 
 - Quit from the console session:
 
-`In> quit`
+`In> {{gquit}}`
 
 - Execute a yacas script, then exit:
 
--- a/pages/common/yacas.md
+++ b/pages/common/yacas.md
@@ -10,7 +10,6 @@
 - In the console session, execute a statement:
 
 `In> {{Integrate(x)Cos(x)}};`
-`Out> {{Sin(x)}}`
 
 - In the console session, display an example:
 
@@ -19,7 +18,6 @@
 - Quit from the console session:
 
 `In> quit`
-`Quitting...`
 
 - Execute a yacas script, then exit:
 
--- a/pages/common/yacas.md
+++ b/pages/common/yacas.md
@@ -3,23 +3,23 @@
 > Yet Another Computer Algebra System.
 > More information: <http://www.yacas.org>.
 
-- Start an interactive console session:
+- Start an interactive `yacas` session:
 
 `yacas`
 
-- In the console session, execute a statement:
+- In the `yacas` session, execute a statement:
 
 `In> {{Integrate(x)Cos(x);}}`
 
-- In the console session, display an example:
+- In the `yacas` session, display an example:
 
 `In> {{Example();}}`
 
-- Quit from the console session:
+- Quit from the `yacas` session:
 
 `In> {{quit}}`
 
-- Execute one or more yacas scripts (without terminal or prompts), then exit:
+- Execute one or more `yacas` scripts (without terminal or prompts), then exit:
 
 `yacas -p -c {{path/to/script1}} {{path/to/script2}}`
 
--- a/pages/common/yacas.md
+++ b/pages/common/yacas.md
@@ -19,10 +19,10 @@
 
 `In> {{gquit}}`
 
-- Execute a yacas script, then exit:
+- Execute one or more yacas scripts, then exit:
 
-`yacas -pc {{path/to/script}}`
+`yacas -pc {{path/to/script1}} {{path/to/script2}}`
 
-- Execute and print the result of one statement:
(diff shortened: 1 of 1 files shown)


pages/common/yacas.md
```diff
@@ -0,0 +1,28 @@
...
+> Yet Another Computer Algebra System.
+> More information: <http://www.yacas.org>.
+
+- Start an interactive console session:
+
+`yacas`
+
+- In the console session, execute a statement:
+
+`In> {{Integrate(x)Cos(x)}};`
```
CleanMachine1, line 12: What is the In>

is that a part of the command you enter or are you trying to reference the fact that it is when you are in the repl

daniel-birket: Please see the pull-request note. "In>" is the prompt of the repl. I had also included "Out>" and the corresponding sample output but lint rejected it as more than one command. Do you have an example of how tldr represents a repl? I couldn't think of one. I tried "In> `statement;`" with the prompt outside the `` but my tldr client rendered it wrong.

I'm also unsure about the ";" required to terminate the statement. On second thought, it probably belongs inside the {{}} as part of the sample statement.

I've just pushed a commit that uses {{}} to distinguish the prompt from the user's input. It seems clearer to me.

marchersimon: Normally we don't include any prompts, but I don't see any reason why not if you can clearly distinguish between them.

pages/common/yacas.md
```diff
@@ -0,0 +1,28 @@
+# yacas
+
+> Yet Another Computer Algebra System.
+> More information: <http://www.yacas.org>.
+
+- Start an interactive console session:
```
marchersimon, line 6:
```suggestion
- Start an interactive `yacas` session:
```
I think `console` is a little to unspecific. `yacas` isn't like the console you're running it in.

pages/common/yacas.md
```diff
@@ -0,0 +1,28 @@
...
+
+`yacas`
+
+- In the console session, execute a statement:
```
marchersimon, line 10: I've multiple suggestions:
```suggestion
- In the `yacas` session, execute a statement:
```
```suggestion
- While running `yacas`, execute a statement:
```
```suggestion
- Execute a statement within `yacas`:
```
Take the one you like the most (or your own if you have another idea).
That also applies to the next example.

pages/common/yacas.md
```diff
@@ -0,0 +1,28 @@
...
+
+`In> {{Integrate(x)Cos(x);}}`
+
+- In the console session, display an example:
```
marchersimon, line 14: What kind of example is this? And is it random or something specific?

pages/common/yacas.md
```diff
@@ -0,0 +1,28 @@
...
+
+`In> {{Example();}}`
+
+- Quit from the console session:
```
marchersimon, line 18:
```suggestion
- Quit `yacas`:
```

pages/common/yacas.md
```diff
@@ -0,0 +1,28 @@
...
+
+`In> {{quit}}`
+
+- Execute one or more yacas scripts (without terminal or prompts), then exit:
+
+`yacas -p -c {{path/to/script1}} {{path/to/script2}}`
```
marchersimon, line 24: Any reason you didn't use the long options here? Especially in this case I think it would help remember them easier.

pages/common/yacas.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Execute and print the result of one statement, then exit:
```
marchersimon, line 26:
```suggestion
- Calculate and print the result of one statement, then exit:
```
```suggestion
- Compute and print the result of one statement, then exit:
```
I think one of those two would fit better.

pages/common/yacas.md
daniel-birket: The sample above, "Integrate(x)Cos(x);" serves to show the flavor of symbolic math statements accepted by the repl. This particular example computes the integral of the Cosine function and prints "Out> -Sin(x)". I could have used "{{statement}};" but a working example gives the user a better idea of how to use it. Another, simpler example might be "2+2;" but that might give the impression that yacas is a calculator, not a symbolic computer algebra system (like "Mathematica" and "sympy").

The yacas Example() function prints from a list of other example statements that you could try. The first one is "40!" or 40-factorial. There are many. They serve the same function as tldr, so I mentioned them. Both the Example() function and "quit" appear in an introduction paragraph when yacas starts an interactive session.

pages/common/yacas.md
daniel-birket: accepted this suggestion and pushed an update.

pages/common/yacas.md
daniel-birket: yacas is not a calculator. The author's description is that yacas "executes statements", which are very general. For example "Echo()". You would not describe calling the printf function as "calculate" or "compute".

pages/common/yacas.md
daniel-birket: They don't exist.

From the manual:

yacas -c
Inhibit printing of prompts In> and Out>. Useful for non-interactive sessions.

yacas -p
Does not use terminal capabilities, no fancy editing on the command line and no escape sequences printed. Useful for non-interactive sessions.

Although there are some long flags, there are no long versions of the short flags given in this reference.

pages/common/yacas.md
daniel-birket: I have pushed an update using this idea.

pages/common/yacas.md
daniel-birket: I take that back. I actually used "the `yacas` session" like the other examples.

marchersimon (approved):
Looks good, thanks

daniel-birket:
@CleanMachine1 , I updated this one several times since you looked at it. Would you please review again?

pages/common/yacas.md
```diff
@@ -0,0 +1,28 @@
+# yacas
+
+> Yet Another Computer Algebra System.
+> More information: <http://www.yacas.org>.
+
+- Start an interactive `yacas` session:
+
+`yacas`
+
+- In the `yacas` session, execute a statement:
```
CleanMachine1, line 10:
```suggestion
-While in a `yacas` session, execute a statement:
```
Thoughts?

Remember to apply throughout.

CleanMachine1 (approved):

CleanMachine1:
If the Out> is only for output, meaning it cannot take **any** input then, remove the In> since that is basic, its like how bash commands often start with $, but you don't have every command in tldr with a $ before it.

pages/common/yacas.md
daniel-birket: Done

marchersimon (approved):

PR #6276, opened 2021-07-29
ip: add commands

perfecto25:
adding additional IP command examples

### DIFF
--- a/pages/linux/ip.md
+++ b/pages/linux/ip.md
@@ -19,6 +19,16 @@
 
 `ip route`
 
+- show a route for an IP or subnet
+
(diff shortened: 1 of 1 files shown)


marchersimon:
Oh, wow, that's a lot of examples. Unfortunately, to keep pages simple and short, we only allow eight examples per page. What you could do however, is adding those examples to sub-command pages. We already have one for `ip-address`, `ip-link`, `ip-neighbour` and `ip route`. You can also add more subcommand pages if you want to.

marchersimon:
I'll close this now, but feel free to open a new PR when you want to add something in the subcommand-pages. Thanks for you efforts!

PR #6281, opened 2021-07-30
gh-browse: add page

navarroaxel:

### DIFF
--- a/pages/common/gh-browse.md
+++ b/pages/common/gh-browse.md
@@ -3,11 +3,11 @@
 > Open a GitHub repository in the browser or print the URL.
 > More information: <https://cli.github.com/manual/gh_browse>.
 
-- Open the home page of the current repository in the default web browser:
+- Open the homepage of the current repository in the default web browser:
 
 `gh browse`
 
-- Open the home page of a specific repository in the default web browser:
+- Open the homepage of a specific repository in the default web browser:
 
 `gh browse {{owner}}/{{repository}}`
 
@@ -27,10 +27,10 @@
 
 `gh browse --branch {{branch_name}}`
 
-- Open a specific file or directory of the current repository in the web browser (the path should be relative to the repository):
+- Open a specific file or directory of the current repository in the web browser:
 
-`gh browse {{path/to/file_or_folder}}`
+`gh browse {{path_from_root_of_repository}}`
 
-- Display the destination URL without open the web browser:
+- Print the destination URL without open the web browser:
 
 `gh browse --no-browser`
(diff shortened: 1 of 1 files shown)


pages/common/gh-browse.md
```diff
@@ -0,0 +1,36 @@
+# gh browse
+
+> Open a GitHub repository in the browser or print the URL.
+> More information: <https://cli.github.com/manual/gh_browse>.
+
+- Open the home page of the current repository in the default web browser:
```
marchersimon, line 6: I think we should use `homepage` (https://trends.google.com/trends/explore?geo=AT&q=home%20page,homepage)
```suggestion
- Open the homepage of the current repository in the default web browser:
```

pages/common/gh-browse.md
```diff
@@ -0,0 +1,36 @@
...
+
+`gh browse`
+
+- Open the home page of a specific repository in the default web browser:
```
marchersimon, line 10:
```suggestion
- Open the homepage of a specific repository in the default web browser:
```

pages/common/gh-browse.md
```diff
@@ -0,0 +1,36 @@
...
+
+`gh browse {{issue_or_pull_request_number}}`
+
+- Open a specific branch in the web browser:
+
+`gh browse --branch {{branch_name}}`
+
+- Open a specific file or directory of the current repository in the web browser (the path should be relative to the repository):
+
+`gh browse {{path/to/file_or_folder}}`
```
marchersimon, line 32: Do you think we should make a little exception here?
```suggestion
- Open a specific file or directory of the current repository in the web browser:

`gh browse {{path_from_root_of_repository}}`
```

pages/common/gh-browse.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Display the destination URL without open the web browser:
```
CleanMachine1, line 34:
```suggestion
- Print the destination URL without open the web browser:
```

This may be like my newcoming of disliking display but print works better here potentially since it was used in the title description

CleanMachine1 (approved):
After comments are approved :+1:

pages/common/gh-browse.md
CleanMachine1: I don't like this but I can't see an alternative since anything shorter I think of indicates from / root

CleanMachine1: I could maybe have a `the` between `from` and `root`

sbrl: That's right, @CleanMachine1. This is the root of the *repository*, not the system - hence `{{/path/to/file}}` is not appropriate here.

I think this is the best we can do in this situation.

sbrl (approved):

PR #5952, opened 2021-05-13
style-guide: add Chinese translation

noarchwastaken:
a proposal in addition to #5240 to guide new contributors to
unify the formatting of Chinese pages.

Previously, I stated in #5240 that I personally format Chinese differently; for simplicity, I still think it's best to follow a widely used standard, the [Chinese Copywriting Guidelines](https://github.com/sparanoid/chinese-copywriting-guidelines). Sorry for any confusion that might have caused...

- [x] The page (if new), does not already exist in the repo.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message).
- [ ] The page follows the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).

PR #5240, opened 2021-02-05
Chinese pages: correct Chinese punctuation

bl-ue:
Attempt to correct and normalize Chinese punctuation across all Chinese pages.

**Note:** I'm not a Chinese speaker, so there may be mistakes.

cc @einverne @mebeim
cc @gyli @wizarot @starccy @ChungZH @zhouLion @xiaolong-666 @sandylaw @shanoaice @telnetning

Refs:
#2897
#3426
#3442

What I've done so far:
- Change colons before "More information" links to Chinese colon
- Change colons at the end of example descriptsion to English colon per [@mebeim's comment](https://github.com/tldr-pages/tldr/pull/3442#pullrequestreview-303829335)

Questions:
- What should period be at the end of page sentences in page descriptions? for sentences that end with English letters? for sentences that end with Chinese letters?

PR #6285, opened 2021-07-31
style-guide: add Chinese-specific rules

blueskyson:
This is a translation of [#5952](https://github.com/tldr-pages/tldr/pull/5952).
I keep the original meaning with few revisions, so there might exists some sentences that are not fluent.
The words "Chinese-Latin mix" may cause some confusion. It means a sentence that contains both Chinese words and Latin words, but I didn't find a proper term to express such situation.


marchersimon:
Thanks for the traslation @blueskyson. I assume this is a literaly translation of #5952. Personally I don't think this page really does what it should. Things like the layout or token syntax are already explained in our original style guide. I tought it would be the best if this page would just cover the Chinese-specific rules, like half-width vs full-witdh and all the other punctuations. Maybe it's not even worth creating a whole new page for that and only creating a new section in our style guide.

blueskyson:
> Maybe it's not even worth creating a whole new page for that, but only a new section in our style guide.

Sounds like a plan @marchersimon . Shall I close this PR, and append a section with Chinese-specific rules to the style guide in another PR?

marchersimon:
I think #6200 could take a while, so I'll just take care of the marge conficts. And I think you could just rename this PR and do the chnages here, but it doesn'r really matter. My concern is just what we should do with the page @noarchwastaken created in #5952. Do you think a Chinese translation of the Chinese part would be even neccessary?

blueskyson:
> Do you think a Chinese translation of the Chinese part would be even neccessary?
Yes, the Chinese-specific rules need to be mentioned in Chinese translation as well. Not every Chinese writer knows the normalized copywriting of Chinese-Latin mix.

marchersimon:
I'm asking if we should have a Chinese translation of the style guide at all? We don't have that for any other language.

blueskyson:
Hum... I think  Chinese translation is not necessary, but better than nothing. Maybe you could open an issue and vote for having style-guide in other languages or not, then decide whether to add according to voting result.

contributing-guides/style-guide.md
```diff
@@ -91,3 +92,27 @@ The example above does not use a serial comma, so this could mean one of two thi
...
+
+The following guidelines are applied to Chinese (zh) and traditional Chinese (zh_TW):
+
+1. Place one space before/after English words.  
+   For example, use `列出所有 docker 容器` rather than `列出所有docker容器`.
+2. Place one space before/after numbers.  
+   For example, use `宽度为 50 个字` rather than `宽度为50个字`.
+3. Place one space between numbers and units **except** degrees and percentages.  
+   For example, use `容量 50 MB` rather than `容量 50MB`.  
+   For instances of degree and percentage, use `50°C` and `50%` rather than `50 °C` and `50 %`.
```
marchersimon, line 108: I wouldn't say 50MB is really wrong. Also, how does that only apply to Chinese?

contributing-guides/style-guide.md
```diff
@@ -91,3 +92,27 @@ The example above does not use a serial comma, so this could mean one of two thi
...
+## Chinese-Specific Rules
+
+When Chinese words, Latin words and Arabic numerals are written in the same sentence, it takes more attention to copywriting.
+
+The following guidelines are applied to Chinese (zh) and traditional Chinese (zh_TW):
+
+1. Place one space before/after English words.  
+   For example, use `列出所有 docker 容器` rather than `列出所有docker容器`.
+2. Place one space before/after numbers.  
+   For example, use `宽度为 50 个字` rather than `宽度为50个字`.
```
marchersimon, line 105: Could we just combine those two?

marchersimon (commented):
I think the most important rule is kind of missing: When sholudn't you use full-width.

Basically, that would be something like
```
1. Use half-width punctuation after Latin characters.
   For example, use 等效于 `vim -R`. rather than 等效于 `vim -R`。
```

contributing-guides/style-guide.md
blueskyson: It depends on an organization's decision. Internet giants like Google, Microsoft, Wikipedia apply this rule in Chinese regions. But Apple doesn't.

contributing-guides/style-guide.md
blueskyson: Sure!

contributing-guides/style-guide.md
```diff
@@ -91,3 +92,27 @@ The example above does not use a serial comma, so this could mean one of two thi
...
+3. Place one space between numbers and units **except** degrees and percentages.  
+   For example, use `容量 50 MB` rather than `容量 50MB`.  
+   For instances of degree and percentage, use `50°C` and `50%` rather than `50 °C` and `50 %`.
+4. No additional spaces before/after full-width punctuations.  
+   For example, use `开启 shell，进入交互模式` rather than `开启 shell ，进入交互模式`
+5. Use full-width punctuations except for long Latin clauses.  
+   For example, use `嗨，你好。` rather than `嗨, 你好.`
```
blueskyson, line 112: > I think the most important rule is kind of missing: When sholudn't you use full-width.
> 
> Basically, that would be something like
> 
> ```
> 1. Use half-width punctuation after Latin characters.
>    For example, use 等效于 `vim -R`. rather than 等效于 `vim -R`。
> ```

I'll combine it with this rule.

zjuyk:
Sorry, I am late.

Accorrding to [government standard draft](http://www.moe.gov.cn/ewebeditor/uploadfile/2015/01/13/20150113092346124.pdf), Chinese technical documents mainly serve Chinese text, which should be based on Chinese punctuation and supplemented by English punctuation.

Therefore, Chinese sentences should end with Chinese punctuation at the end of the sentence. Only when English sentences are cited in Chinese sentences, we keep them.
For example

- 他说的是“Jane, you can do it.”。
- “I like you.（我喜欢你。）”与“I love you.（我爱你。）”的份量不一样。
- 这个语境下不建议使用 at large。

marchersimon:
Then what about https://github.com/sparanoid/chinese-copywriting-guidelines/blob/master/README.en-US.md? I assume there is no correct form, just one for different purposes.

contributing-guides/style-guide.md
marchersimon: I'm still not sure if we should consider this Chinese-specific.

zjuyk:
You mean this [part](https://github.com/sparanoid/chinese-copywriting-guidelines/blob/master/README.en-US.md#use-punctuation-in-halfwidth-form-for-english-sentences)?

marchersimon:
Yep, for example

blueskyson:
@marchersimon 

The suggestion is the same as rule 5 `Use full-width punctuations except for long Latin clauses` which is the same as @zjuyk 's [comment](https://github.com/tldr-pages/tldr/pull/6285#issuecomment-893471444).

I found some examples of such situation in tldr page:
- > 支持大多数协议，包括 HTTP, FTP, 和 POP3. **in curl.md**
- 在目标模拟器 / 设备实例上开启一个远程 shell: **in adb.md**

If the examples above follows the rule 5, they would be:
- > 支持大多数协议，包括 HTTP, FTP, 和 POP3。
- 在目标模拟器 / 设备实例上开启一个远程 shell：

zjuyk:
@marchersimon 

I think there are no conflicts. One separated english sentence will keep halfwidth form, but actuallly we don't use a separated english sentence in chinese writing. Just cite them with `""` or other punctuations.

zjuyk:
Just like the examples in  https://github.com/sparanoid/chinese-copywriting-guidelines

```
賈伯斯那句話是怎麼說的？「Stay hungry, stay foolish.」

推薦你閱讀《Hackers & Painters: Big Ideas from the Computer Age》，非常的有趣。
```

contributing-guides/style-guide.md
blueskyson: Okay, I cancel this rule. If there would be a unit-related issue, this could be discuss again.

contributing-guides/style-guide.md
blueskyson: `Use half-width punctuation after Latin characters.` conflicts with [this](https://github.com/tldr-pages/tldr/pull/6285#issuecomment-893526569)

marchersimon:
Then what about the punctuation? Is using halfwith after Latin characters wrong? We all agreed on using that.

zjuyk:
> Accorrding to [government standard draft](http://www.moe.gov.cn/ewebeditor/uploadfile/2015/01/13/20150113092346124.pdf), Chinese technical documents mainly serve Chinese text, which should be based on Chinese punctuation and supplemented by English punctuation.
> 
> Therefore, Chinese sentences should end with Chinese punctuation at the end of the sentence. Only when English sentences are cited in Chinese sentences, we keep them.

I respectfully disagree, but this discussion took too much time. I think we need to finish the style guide rules as soon as possible. 

Maybe we should just make a draft style guide. It is easy to correct them by script in the future.

marchersimon (approved):
Thanks @blueskyson, 
maybe we'll refine this in the future, but I think we should get this merged.

contributing-guides/style-guide.md
marchersimon: But that's what I basically did in #5240. Did I misunderstand something? I'm confused.

PR #6289, opened 2021-08-02
arecord: fix name of Chinese page

marchersimon:
There was a typo in the page name of the Chinese page. It was `arecored`, not `arecord`.


mfrw (approved):
Thank you @marchersimon

CleanMachine1 (approved):
should the title not be changed

PR #6306, opened 2021-08-06
lipo: add page

### LINKED ISSUE #6295: page request: lipo
`man lipo` (https://ss64.com/osx/lipo.html)

chrissxYT:
- [x] The page is in the correct platform directory (`osx`, although it's macos)
- [x] The page description includes a link to documentation or a homepage (the only official docs is a guide on how to port to apple silicon).

Closes #6295

### DIFF
--- a/pages/osx/lipo.md
+++ b/pages/osx/lipo.md
@@ -1,6 +1,6 @@
 # lipo
 
-> Tool for handling Mach-O Universal Binaries
+> Tool for handling Mach-O Universal Binaries.
 
 - Create a universal file from two single-architecture files:
 
--- a/pages/osx/lipo.md
+++ b/pages/osx/lipo.md
@@ -1,0 +1,19 @@
+# lipo
+
+> Tool for handling Mach-O Universal Binaries
+
+- Create a universal file from two single-architecture files:
+
+`lipo {{path/to/binary.x86_64}} {{path/to/binary.arm64e}} -create -output {{path/to/binary}}`
+
+- List all architectures contained in a universal file:
+
+`lipo {{path/to/binary}} -archs`
+
+- Display detailed information about a universal file:
+
+`lipo {{path/to/binary}} -detailed_info`
+
+- Extract a single-architecture file from a universal file:
+
+`lipo {{path/to/binary}} -thin arm64e -output {{path/to/binary.arm64e}}`
--- a/pages/osx/lipo.md
+++ b/pages/osx/lipo.md
@@ -1,6 +1,7 @@
 # lipo
 
 > Tool for handling Mach-O Universal Binaries.
+> More information: <https://ss64.com/osx/lipo.html>.
 
 - Create a universal file from two single-architecture files:
 
--- a/pages/osx/lipo.md
+++ b/pages/osx/lipo.md
@@ -16,4 +16,4 @@
 
 - Extract a single-architecture file from a universal file:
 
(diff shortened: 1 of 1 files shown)


CleanMachine1:
Do we have a homepage or anything?

pages/osx/lipo.md
```diff
@@ -0,0 +1,19 @@
+# lipo
+
+> Tool for handling Mach-O Universal Binaries.
+
```
marchersimon, line 4:
```suggestion
> More information: <https://ss64.com/osx/lipo.html>.

```
How about that one?

pages/osx/lipo.md
```diff
@@ -0,0 +1,19 @@
...
+`lipo {{path/to/binary}} -archs`
+
+- Display detailed information about a universal file:
+
+`lipo {{path/to/binary}} -detailed_info`
+
+- Extract a single-architecture file from a universal file:
+
+`lipo {{path/to/binary}} -thin arm64e -output {{path/to/binary.arm64e}}`
```
marchersimon, line 19:
```suggestion
`lipo {{path/to/binary}} -thin {{arm64e}} -output {{path/to/binary.arm64e}}`
```
Are `x86_64` and `arm64e` the only two architectures this command handles?

pages/osx/lipo.md
```diff
@@ -0,0 +1,19 @@
+# lipo
+
+> Tool for handling Mach-O Universal Binaries.
```
marchersimon, line 3: I don't think this describes the command very well. Especially, since it mentions Mach-0, which isn't mentioned again. What do you think about:
```suggestion
> Convert a universal binary to a single architecture file, or vice versa.
```

chrissxYT: I don't think that describes it well, because it doesn't only convert universal ↔ single-arch, it also converts universal ↔ universal, displays information about and "verifies" universal binaries. I think mentioning Mach-O in this way is exactly the right way, because `lipo` only operates on Mach-O, but mentioning that in the description of every example doesn't make sense to me. (it's repetitive)

pages/osx/lipo.md
chrissxYT: Is putting links to online manpage viewers common in tldr? Then it kind of makes sense, that one, however, is really outdated. (look at the supported architectures: x86 and PPC, no ARM)

chrissxYT:
> Do we have a homepage or anything?

Not really. We have [a tutorial on porting to Apple silicon](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary) and [online manpage viewers](https://ss64.com/osx/lipo.html).

pages/osx/lipo.md
marchersimon: We try to put a link to either the homapage or a man page in every tldr page. We`ve had to deal with a lot of outdated manpages already. I think it's still better than nothing.

pages/osx/lipo.md
marchersimon: But what is Mach-0?

chrissxYT: [Mach-O, short for Mach object file format, is a file format for executables, object code, shared libraries, dynamically-loaded code, and core dumps.](https://en.wikipedia.org/wiki/Mach-O) It is Darwin's equivalent to ELF.

marchersimon (approved):
Thanks

CleanMachine1 (approved):

Issue #6167, opened 2021-06-24
Let's document: nmcli-* 


CleanMachine1:
The command is very useful, running a simple command to edit wifi config is both more initiative and often more effective than using a GUI menu for NetworkManager

These pages are poorly documented 

- [ ] general → #8264 
- [ ] networking → #8264
- [x] radio
- [x] connection → #6185 , (don't tick this)
- [x] device
- [x] agent
- [x] monitor → #6168 
- [x] nmcli (forwarding page edit) → #6203 

I haven't checked through these commands however some need a review and some need to be wrote entirely

CleanMachine1:
Closes #6152

PR #6308, opened 2021-08-06
nmcli: add wifi connection commands

lyctw:

### DIFF
--- a/pages/linux/nmcli.md
+++ b/pages/linux/nmcli.md
@@ -18,3 +18,12 @@
 - Display help for a subcommand:
 
 `nmcli {{subcommand}} --help`
+
+- Displays a table of available wifi access points:
+
+`nmcli dev wifi`
+
+- Connect to a WiFi network:
+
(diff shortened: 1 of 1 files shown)


CleanMachine1:
Hi, this doesn't fit our scheme for nmcli

See #6167 

We have separate pages for every sub command.

However I feel that we could add a notice on the top of the page.

marchersimon:
Yep, we already have both of those examples covered in `nmcli device`. (`dev` is just a shortcut for `device`). We could however create an alias page for `nmcli dev`.

> Edit, we'll see what others think of this 6f2c57c

Generally not a bad idea, since otherwise you wouldn't know those exist. But I would write 
```diff
- For in-depth commands, see `nmcli-{{command}}`.
+ For `nmcli` subcommands, see `nmcli-{{subcommand}}`.
```
(Also, please try to avoid commiting to other's branches. That will probably just make troubles on the next push)

lyctw:
Sorry I didn't notice that subcommands.

PR #6309, opened 2021-08-06
vcgencmd: add page

CleanMachine1:

### DIFF
--- a/pages/linux/vcgencmd.md
+++ b/pages/linux/vcgencmd.md
@@ -15,7 +15,7 @@
 
 `vcgencmd measure_volts`
 
-- Print whether the CPU has throttled:
+- Print the throttled state of the system as a bit pattern:
 
 `vcgencmd get_throttled`
 
--- a/pages/linux/vcgencmd.md
+++ b/pages/linux/vcgencmd.md
@@ -1,0 +1,28 @@
+# vcgencmd
+
+> Print system information for a Raspberry Pi.
+> More information: <https://www.raspberrypi.org/documentation/raspbian/applications/vcgencmd.md>.
+
+- List all available commands:
+
+`vcgencmd commands`
+
+- Print the current CPU temperature:
+
+`vcgencmd measure_temp`
+
+- Print the current voltage usage:
+
+`vcgencmd measure_volts`
+
+- Print whether the CPU has throttled:
+
+`vcgencmd get_throttled`
+
+- Print the bootloader config (only available on Raspberry Pi 4 models):
+
+`vcgencmd bootloader_config`
+
+- Display Help:
+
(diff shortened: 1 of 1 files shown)


pages/linux/vcgencmd.md
```diff
@@ -0,0 +1,28 @@
...
+
+`vcgencmd commands`
+
+- Print the current CPU temperature:
+
+`vcgencmd measure_temp`
+
+- Print the current voltage usage:
+
+`vcgencmd measure_volts`
```
marchersimon, line 16: According to the docs you have to specify a specific block here (`{{core|sdram_c|sdram_i|sdram_p}}`)

pages/linux/vcgencmd.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Print whether the CPU has throttled:
```
marchersimon, line 18: I think this is more accurate, since it doesn't return true or false.
```suggestion
- Print the throttled state of the system as a bit pattern:
```

pages/linux/vcgencmd.md
```diff
@@ -0,0 +1,28 @@
...
+
+`vcgencmd get_throttled`
+
+- Print the bootloader config (only available on Raspberry Pi 4 models):
```
marchersimon, line 22: I don't think there's a version 5 yet, but when there is, this page won't be outdated this way.
```suggestion
- Print the bootloader config (only available on Raspberry Pi 4 and above):
```

CleanMachine1: We don't know whether it will have a EEPROM bootloader so I don't know about this one

pages/linux/vcgencmd.md
CleanMachine1: % vcgencmd measure_volts
volt=1.2000V

marchersimon: Then are those maybe optional? Should  we include them?

Also, I'm not really happy with `voltage usage`, since that's technically not really correct. I'd just say `voltage`.

CleanMachine1: Yes, voltage usage is incorrect

pages/linux/vcgencmd.md
```diff
@@ -0,0 +1,28 @@
...
+
+- List all available commands:
+
+`vcgencmd commands`
+
+- Print the current CPU temperature:
+
+`vcgencmd measure_temp`
+
+- Print the current voltage usage:
```
CleanMachine1, line 14:
```suggestion
- Print the current voltage:
```

navarroaxel (approved):

marchersimon (approved):
Splendid

PR #6319, opened 2021-08-10
pip-install: add --find-links example

oryon-dominik:

### DIFF
--- a/pages/common/pip-install.md
+++ b/pages/common/pip-install.md
@@ -15,6 +15,10 @@
 
 `pip install -r {{requirements.txt}}`
 
+- Install packages from an URL or local file archive (.tar.gz | .whl):
+
+`pip install -f {{url | filename}}`
+
 - Install the local package in the current directory in develop (editable) mode:
 
 `pip install -e .`
--- a/pages/common/pip-install.md
+++ b/pages/common/pip-install.md
@@ -17,7 +17,7 @@
 
 - Install packages from an URL or local file archive (.tar.gz | .whl):
 
-`pip install -f {{url | filename}}`
+`pip install -f {{url|path/to/file}}`
 
 - Install the local package in the current directory in develop (editable) mode:
 


pages/common/pip-install.md
```diff
@@ -15,6 +15,10 @@
 
 `pip install -r {{requirements.txt}}`
 
+- Install packages from an URL or local file archive (.tar.gz | .whl):
```
CleanMachine1, line 18:
```suggestion
- Install packages from an URL or local file package:
```

pages/common/pip-install.md
```diff
@@ -15,6 +15,10 @@
...
+
+`pip install -f {{url | filename}}`
```
CleanMachine1, line 20:
```suggestion
`pip install -f {{url|path/to/package_file.tar.gz|whl}}`
```

sbrl: What do you mean by `whl` here?

sbrl: I see. I'm not sure there's really a way to specify that easily. I'd pick the most common filetype and stick with it.

CleanMachine1: I don't know which is more common, however whl is the standard for python packages since its what its designed to do, while tar.gz is just a generic archive  

whl is specific, tar.gz is generic

I'd go with whl

marchersimon: Maybe we should just leave it out, since both options are listed in the description.

pages/common/pip-install.md
oryon-dominik, line 18:
```suggestion
- Install packages from an URL or local file archive:
```

oryon-dominik: In my opinion archive is more precise than package here.

pages/common/pip-install.md
navarroaxel: Maybe this is better
```suggestion
`pip install -f {{url|path/to/file}}`
```

navarroaxel (approved):

marchersimon (approved):
Thanks @oryon-dominik and sorry for the delay

PR #6323, opened 2021-08-11
2to3: add Chinese translation

zjuyk:


marchersimon (approved):
Looks good, thanks @zjuyk

pages.zh/common/2to3.md
```diff
@@ -0,0 +1,32 @@
+# 2to3
+
+> 自动将 Python 2 代码转换成 Python 3。
+> 更多信息：<https://docs.python.org/3/library/2to3.html>.
+
+- 显示将执行的变动但不执行（dry-run）：
+
+`2to3 {{Python 代码路径}}`
```
blueskyson, line 8:
```suggestion
`2to3 {{文件名.py}}`
```

Thank you for contribution. Here is a better way to describe a script as mentioned in rule 6 of [token syntax](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/style-guide.md#token-syntax).

pages.zh/common/2to3.md
```diff
@@ -0,0 +1,32 @@
...
+
+`2to3 --write {{Python 代码路径}} --nofix={{has_key}} --nofix={{isinstance}}`
+
+- 列出 Python 2 所有可转化为 Python 3 的语言特性：
+
+`2to3 --list-fixes`
+
+- 将某一文件夹的所有 Python 2 文件转化为 Python 3：
+
+`2to3 --output-dir={{Python 3 文件夹路径}} --write-unchanged-files --nobackups {{Python 2 文件夹路径}}`
```
blueskyson, line 28:
```suggestion
`2to3 --output-dir={{Python 3 文件夹}} --write-unchanged-files --nobackups {{Python 2 文件夹}}`
```

pages.zh/common/2to3.md
```diff
@@ -0,0 +1,32 @@
+# 2to3
+
+> 自动将 Python 2 代码转换成 Python 3。
```
CleanMachine1, line 3:
```suggestion
> 自动将 Python 2 代码转换成 Python 3.
```

I think this is correct, let other maintainers have a look before committing this since I have never reviewed a Chinese page.

pages.zh/common/2to3.md
```diff
@@ -0,0 +1,32 @@
...
+> 更多信息：<https://docs.python.org/3/library/2to3.html>.
+
+- 显示将执行的变动但不执行（dry-run）：
```
CleanMachine1, line 6: I think these colons used are incorrect. I think it should be `:` rather than `：`

pages.zh/common/2to3.md
blueskyson: Please fix other `{{Python 代码路径}}` as well. 😅

pages.zh/common/2to3.md
blueskyson: Please fix other `{{Python 3 文件夹路径}}` as well.

pages.zh/common/2to3.md
navarroaxel: I think the `。` is correct here. @marchersimon ?

marchersimon: Latin characters after Latin text so it should be `.`.

pages.zh/common/2to3.md
marchersimon: Per our rules it should, but since the parentheses before it is full-width I'm not entirely sure. @blueskyson?

blueskyson: Current way complies the Chinese-specific rules and looks good to me as well. See also the last example in [hexo](https://github.com/tldr-pages/tldr/blob/main/pages.zh/common/hexo.md) use the same way.

pages.zh/common/2to3.md
```diff
@@ -0,0 +1,32 @@
...
+
+`2to3 {{文件.py}}`
+
+- 将 Python 2 文件转化为 Python 3：
```
blueskyson, line 10:
```suggestion
- 将 Python 2 文件转化为 Python 3:
```

Please use half-width when the last character of a sentence is number or Latin character. I'm sorry that I forgot to mention this rule in style guide. I think this is the very last suggestion, the page could be merged after fixing them, thank you.

zjuyk: ok, I will follow the written rules

blueskyson (approved):

PR #6329, opened 2021-08-11
PULL_REQUEST_TEMPLATE: add easy to copy links

CleanMachine1:
Due to the links being a little annoying to copy for the user, unless they read the link and find the page or they press `preview` they wont be able to easily use the links, meaning they are less likely to even give the effort, meaning more bug fixing.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,6 @@
 <!-- Thank you for sending a PR! -->
+<!-- Relevant links - https://github.com/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message -->
+<!-- https://github.com/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines -->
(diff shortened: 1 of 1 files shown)


sbrl (approved):
Fair enough

navarroaxel (approved):

schneiderl (approved):

marchersimon:
I don't really get how that changes anything 🤔

CleanMachine1:
@marchersimon 

Consider a first time contributor, before this PR, the link they had was 

/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message

which does nothing in editor mode, however at least a link can be copied easily.

My first time contributing, I forgot about the preview mode, so the links are active.

I found the pages manually from reading the link, and although it only took me a minute to do so, I would guess a small fraction would not even bother, meaning more potential bug and error fixing for us to do.

marchersimon:
Oh, my fault. I thought you replaced line 2 with line 3, but you added them both 😅

Issue #6062, opened 2021-05-28
Pages that need `More information` links


marchersimon:
This is an updated list of all pages without a `More information` link. For discussion, please use #5510, this is merely a tracking issue.

- [x] android/wm → #6030
- [x] commom/exec → #6536
- [x] common/ftp → #6342
- [x] common/history → #5825
- [x] common/host → #6342
- [x] common/hr → #6342
- [x] common/htop → #6165
- [x] common/iconv → #6342
- [x] common/id3tag → #6342
- [x] common/if → #6342
- [x] common/ifconfig #6218 
- [x] common/import → #6342
- [x] common/ionice → #6342
- [x] common/ipcs → #6342
- [x] common/javac → #6536
- [x] common/jhat → #6266 
- [x] common/jmap → #6404 
- [x] common/jobs → #6545
- [x] common/jstack → #6545
- [x] common/kill → #6545
- [x] common/killall → #6545
- [x] common/last → #6577
- [x] common/ldapsearch → #6577
- [x] common/leave → #6577
- [x] common/lex → #6577
- [x] common/lorem → #6577
- [x] common/lp → #6577
- [x] common/lpstat → #6577
- [x] common/lsof → #6577
- [x] common/mail → #6795
- [x] common/mailx → #6795
- [x] common/mesg → #6795
- [x] common/meshlabserver → #6971
- [x] common/mmv → #6795
- [x] common/monop → #6795
- [x] common/most → #6795
- [x] common/mount → #6788
- [x] common/mscore → #6795
- [x] common/nkf → #6971
- [x] common/nslookup → #6828
- [x] common/objdump → #6828
- [x] common/obs → #6971
- [x] common/odps → #6074
- [x] common/odps-auth → #6074
- [x] common/odps-func → #6074
- [x] common/odps-inst → #6074
- [x] common/odps-resource → #6074
- [x] common/odps-table → #6074
- [x] common/odps-tunnel → #6074
- [x] common/paci → #6333
- [x] common/pactl → #6828
- [x] common/passwd → #6828
- [x] common/patch → #6828
- [x] common/pax → #6971
- [x] common/pdfimages → #6971
- [x] common/pdfjoin → #5787
- [x] common/pdftotext → #6334
- [x] common/ping → #6658
- [x] common/ping6 → #6658
- [x] common/pinky → #6971
- [x] common/platformio → #6335 
- [x] common/ps → #6830
- [x] common/pssh → #6971
- [x] common/pv → #6830
[...]

PR #6335, opened 2021-08-12
platformio: add more information link

Gitleptune:
Official homepage, unsure if it's needed because it's an alias.

### DIFF
--- a/pages/common/platformio.md
+++ b/pages/common/platformio.md
@@ -1,7 +1,7 @@
 # platformio
 
 > This command is an alias of `pio`.
-> More information: <https://docs.platformio.org/en/latest/>.
+> More information: <https://docs.platformio.org/en/latest/core/userguide/>.
 
 - View documentation for the original command:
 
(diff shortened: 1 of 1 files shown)


navarroaxel:
The alias pages shouldn't have a link.

Gitleptune:
Oh I thought it should because it was in #6062 
image

navarroaxel:
> Oh I thought it should because it was in #6062

@Gitleptune Maybe an autogenerated list.
@marchersimon Could you help here?

marchersimon:
Yep, I autogenerated that list and didn't think about aliases. But I don't see a reason not to add it. Especially when we someday implement a system which requires every page to have a more information link, we would have to differentiate between normal pages and alias pages.

sbrl (approved):

marchersimon (approved):
Thanks @Gitleptune

navarroaxel (approved):

PR #6354, opened 2021-08-13
lb: add page

pixelcmtd:

### DIFF
--- a/pages/common/lb.md
+++ b/pages/common/lb.md
@@ -1,0 +1,28 @@
+# lb
+
+> A blog management tool.
+> More information: <https://github.com/LukeSmithxyz/lb>.
+
+- Make a new draft:
+
+`./lb new`
+
+- Edit a draft:
+
+`./lb edit`
+
+- Delete a draft:
+
+`./lb trash`
+
+- Publish a draft:
+
+`./lb publish`
+
+- Delete a published post:
+
+`./lb delete`
+
+- Unpublish a published post to edit it as a draft again:
+
+`./lb revise`
--- a/pages/common/lb.md
+++ b/pages/common/lb.md
@@ -5,24 +5,24 @@
 
 - Make a new draft:
 
-`./lb new`
+`lb new`
 
 - Edit a draft:
 
-`./lb edit`
+`lb edit`
 
 - Delete a draft:
 
-`./lb trash`
+`lb trash`
 
 - Publish a draft:
 
-`./lb publish`
+`lb publish`
 
 - Delete a published post:
 
-`./lb delete`
+`lb delete`
 
 - Unpublish a published post to edit it as a draft again:
 
-`./lb revise`
+`lb revise`
--- a/pages/common/lb.md
+++ b/pages/common/lb.md
@@ -1,6 +1,6 @@
 # lb
 
-> A blog management tool.
+> A tool for managing a blog contained in the current directory.
 > More information: <https://github.com/LukeSmithxyz/lb>.
 
 - Make a new draft:
--- a/pages/common/lb.md
+++ b/pages/common/lb.md
@@ -1,6 +1,7 @@
 # lb
 
 > A tool for managing a blog contained in the current directory.
+> Drafts and posts to operate on are selected interactively when running the commands.
 > More information: <https://github.com/LukeSmithxyz/lb>.
 
 - Make a new draft:


pages/common/lb.md
```diff
@@ -0,0 +1,28 @@
+# lb
+
+> A blog management tool.
+> More information: <https://github.com/LukeSmithxyz/lb>.
+
+- Make a new draft:
+
+`./lb new`
```
marchersimon, line 8: I can't find it in the style guide but I feel like we had a rule to write every command like it was installed on the system.
```diff
- ./lb
+ lb
```

pages/common/lb.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Edit a draft:
```
marchersimon, line 10: Which draft does that edit? Or can you only have one at a time?

pages/common/lb.md
```diff
@@ -0,0 +1,28 @@
...
+
+`./lb edit`
+
+- Delete a draft:
+
+`./lb trash`
+
+- Publish a draft:
```
marchersimon, line 18: Where does that publish? Is there a list of blog sites this supports?

pages/common/lb.md
```diff
@@ -0,0 +1,28 @@
...
+
+`./lb publish`
+
+- Delete a published post:
```
marchersimon, line 22: Here again, which post wil this delete?

pages/common/lb.md
chrissxYT: ok cool, but another rule is to be beginner friendly and literally nobody installs `lb`. the readme doesn't have install instructions, there are no packages in any distribution and i'm not even sure if it works properly when installed.

pages/common/lb.md
chrissxYT: If you have multiple drafts, it has a kind of wizard, like this:
```
1 first-article
2 second-article

Select an article to edit by id.
```

As of [July 2020](https://github.com/LukeSmithxyz/lb/commit/f13cf1a3a642dcaa6eb188b15db4e36450edcabd), if you only have one draft, it automatically selects that one.

pages/common/lb.md
chrissxYT: I don't really get your question. It adds the article to the RSS feed, `index.html`, "blogfile" (`2021.html`) and standalone file (`blog/article-name.html`)

pages/common/lb.md
chrissxYT: All subcommands except `new` use the `listandReturn` function, which displays the prompt I described for `edit`, in this case:
```
1 article

Pick an entry by number to trash, or press Ctrl-C to cancel.
```

pages/common/lb.md
blueskyson:
```suggestion
- Publish a draft to blog directory:
```

Maybe this is clearer?

pages/common/lb.md
marchersimon: @sbrl what do you think?

pages/common/lb.md
marchersimon: Ah, I didn't think the blog would be self-hosted. The suggestion above looks good (if it's correct), but it should be
```suggestion
- Publish a draft to the blog directory:
```

pages/common/lb.md
marchersimon: When deleting something users may be a bit more causious about executing that command if they don't know exactly what will happen. I can't come up with a good formulation for the description, so how about adding ``> All subcommands except `new` will list all blog posts to select from`` to the command description?

pages/common/lb.md
CleanMachine1: Even the documentation shows it just being run from local directory.

I think something like
```diff
./
{{path/to/lb}} ...
```
would be better

marchersimon: Did you mean to **add** `{{path/to/lb}}`? If so, I'm not a fan of it. I'm fine with the way it's now.

pixelcmtd: > I think something like
> 
> ```diff
> + ./
> - {{path/to/lb}} ...
> ```
> 
> would be better

> Did you mean to **add** `{{path/to/lb}}`?

Because then you can just as well assume it is installed in the system. `lb` is supposed to be in the pwd. Some users might want to install it into their systems, so `lb` makes some kind of sense, but I very much don't like `{{path/to/lb}}`, because either it is in the `PWD` or the `PATH`, **not** in another directory, **ever**.

pages/common/lb.md
pixelcmtd: If we put it here, it should also be in `new`'s description. So I would prefer some kind of mentioning it in the general description, like `A tool for managing a blog contained in the current directory.`

marchersimon: Sounds good

pages/common/lb.md
pixelcmtd: Thought about this for...a while and, while I don't like it, I can't come up with a better solution, except a better _formulation_:

Drafts and posts to operate on are selected interactively when running the commands.

marchersimon: Sounds pretty good. I like it.

marchersimon (approved):
👌🏽

blueskyson (approved):

PR #6321, opened 2021-08-11
xetex: add page

chrissxYT:
- [x] The page (if new), does not already exist in the repository.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

basically a copy-paste from `pdflatex`, because the basic usage is the same. closes #6318

PR #6365, opened 2021-08-14
latex, pdflatex, pdftex, tex: replace 'halting' with 'exiting'

chrissxYT:
https://github.com/tldr-pages/tldr/pull/6321#discussion_r687248228

### DIFF
--- a/pages/common/latex.md
+++ b/pages/common/latex.md
@@ -11,6 +11,6 @@
 
 `latex -output-directory={{path/to/directory}} {{source.tex}}`
 
-- Compile a DVI document, halting on each error:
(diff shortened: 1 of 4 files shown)


blueskyson:
It looks like a choosey fix. I personally feel `exiting` and `halting` are the same meaning.

CleanMachine1 (approved):

marchersimon:
I personally don't care either, but I feel like it may not be obvious to what `halting` is to everyone. I originally used it, since it was used in the LaTeX docs.

marchersimon (approved):

blueskyson:
Thanks for your conribution,  @chrissxYT .

Issue #6299, opened 2021-08-04
PDF generation has weird kerning, and examples go off the page


ajs256:
https://tldr.sh/assets/tldr-book.pdf

<img width="469" alt="image" src="https://user-images.githubusercontent.com/67526318/128207124-e38f1cbb-45dc-4ea0-b4a6-114afdf30d6c.png">
<img width="890" alt="image" src="https://user-images.githubusercontent.com/67526318/128207159-cbb15a73-79fa-4933-a956-2105ab0367cc.png">
<img width="645" alt="image" src="https://user-images.githubusercontent.com/67526318/128207213-0b9a23f0-a931-4ee0-aa26-b97e18213c6a.png">

marchersimon:
I noticed that just yesterday and wanted to open an issue about it 😄
Could it maybe have something to do with #6215? Maybe now it's forced to use a different version of Python. (But honestly I have no idea)

ajs256:
> Could it maybe have something to do with #6215?

I'm not sure adding the SPDX license headers would have anything to do with this, did you mean a different PR or commit?

CleanMachine1:
It couldn't, the shebang is at the top of files still

marchersimon:
My fauld, I thought this PR added `#!/usr/bin/env python3` too.

PR #6366, opened 2021-08-14
Add render2.py

blueskyson:
Issue [#6299](https://github.com/tldr-pages/tldr/issues/6299). I fix the bug and make the script faster by processing without saving temporary files. It may stuck for a while in the final stage (convert HTML string to PDF file). The following are some screenshots (test from my ubuntu 20.04 virtual machine).

Screenshot from 2021-08-14 23-37-53
Screenshot from 2021-08-14 23-38-11
Screenshot from 2021-08-14 23-38-32

### DIFF
--- a/scripts/pdf/render2.py
+++ b/scripts/pdf/render2.py
@@ -1,0 +1,90 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: MIT
+
+"""
+A Python script to generate a single PDF document with all the `tldr` pages. It works by generating
+intermediate HTML files from existing md files using Python-markdown, applying desired formatting
+through CSS, and finally rendering them as PDF. There is no LaTeX dependency for generating the PDF.
+"""
+
+import os
+import sys
+import glob
+import re
+import markdown
+import argparse
+from datetime import datetime
+
+from weasyprint import HTML, CSS
+
+def main(loc, colorscheme):
+    oslist = []
+    allmd = []
+
+    # Checking correctness of path
+    if not os.path.isdir(loc):
+        print("Invalid directory. Please try again!", file=sys.stderr)
+        sys.exit(1)
+
+    # Writing names of all directories inside 'pages' to a list
+    for os_dir in os.listdir(loc):
+        oslist.append(os_dir)
+    oslist.sort()
+
+    # A string that stores pages in HTML format
+    html = ''
+
+    # Set up css style sheets
+    csslist = [CSS("basic.css")]
+    if colorscheme != "basic":
+        csslist.append(CSS(colorscheme + ".css"))
+
+    # Required strings to create intermediate HTML files
+    html += '<!doctype html><html><head><meta charset="utf-8">' \
+        +"</head><body><h1 class=title-main>tldr pages</h1>" \
+        + "<h4 class=title-sub>Simplified and community-driven man pages</h4>" \
+        + "<h6 class=title-sub><em><small>Generated on " + datetime.now().strftime("%c") + "</small></em></h6>" \
+        + '<p style="page-break-before: always" ></p>'
+
+    for operating_sys in oslist:
+        # Required string to create directory title pages
+        html += "<h2 class=title-dir>" + operating_sys.capitalize() + "</h2>" \
+            + '<p style="page-break-before: always" ></p>'
+
+        # Creating a list of all md files in the current directory
+        for temp in glob.glob(os.path.join(loc, operating_sys, "*.md")):
+            allmd.append(temp)
+
+        # Sorting all filenames in the directory, to maintain the order of the PDF
+        allmd.sort()
+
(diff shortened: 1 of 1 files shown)


blueskyson:
By the way, I think it would be better to provide the python version in `scripts/pdf/README.md`. I personally use 3.7.9 to debug.

CleanMachine1:
Why not just edit the original, otherwise anything relying on script1 will not get the update.

blueskyson:
@CleanMachine1 I thought you might want to keep the original script.

CleanMachine1:
Its not really up to me, however Git handles going back however it could break backwards compatibility, however I haven't really read the code.

MasterOdin (changes requested):
I would prefer to see the original edited. It's not like the output is being completely changed or anything, and as @CleanMachine1 mentions, if we really want the old file for whatever reason, can just look into git history (not that I think we will).

blueskyson:
Okay, I modified `render.py` directly and deleted `render2.py`. @CleanMachine1 @MasterOdin

marchersimon (approved):
Can't really review the code, but it seems to work, so here's my approval. Thanks!

scripts/pdf/render.py
```diff
@@ -15,106 +15,66 @@
...
-            ap.append(p)
-
-    # Writing the PDF to disk, preserving metadata of first `tldr` page
-    group[2].copy(ap).write_pdf('tldr-pages.pdf')
+    
+    html += "</body></html>"
+    
+    # Writing the PDF to disk
+    print("\nConverting all pages to PDF...")
+    HTML(string=html).write_pdf("tldr-pages.pdf", stylesheets=csslist)
```
MasterOdin, line 73: From the [api reference](https://doc.courtbouillon.org/weasyprint/stable/api_reference.html#weasyprint.HTML.write_pdf) for this function, could `csslist` be a list of filename strings instead of instantiated `CSS` classes?

scripts/pdf/render.py
```diff
@@ -15,106 +15,66 @@
...
-        f.write(header + title_content)
+        + '<p style="page-break-before: always" ></p>'
 
-    group.append(HTML("title.html").render())
+    # Writing names of all directories inside 'pages' to a list
+    for os_dir in os.listdir(loc):
+        oslist.append(os_dir)
+    oslist.sort()
 
     for operating_sys in oslist:
```
MasterOdin, line 46: Can combine these into a single for loop:

```suggestion
    for operating_sys in sorted(os.listdir(loc)):
```

scripts/pdf/render.py
```diff
@@ -15,106 +15,66 @@
...
         # Creating a list of all md files in the current directory
         for temp in glob.glob(os.path.join(loc, operating_sys, "*.md")):
             allmd.append(temp)
-
-        # Sorting all filenames in the directory, to maintain the order of the PDF
         allmd.sort()
 
-        # Conversion of Markdown to HTML
+        # Conversion of Markdown to HTML string
         for page_number, md in enumerate(allmd, start=1):
```
MasterOdin, line 57: Same as above, we can get rid of this double for loop and just do:

```suggestion
        for page_number, md in enumerate(sorted(glob.glob(os.path.join(loc, operating_sys, "*.md"))), start=1):
```

scripts/pdf/render.py
```diff
@@ -15,106 +15,66 @@
...
-
+            with open(md, "r") as inp:
+                text = inp.readlines()
+                for line in text:
+                    if re.match(r'^>', line):
+                        line = line[:0] + '####' + line[1:]
+                    html += markdown.markdown(line)
+            html += '<p style="page-break-before: always" ></p>'
+            print("Rendered page {} of the directory {}".format(
+                str(page_number), operating_sys))
```
MasterOdin, line 66: Given we're using 3.9 to run this script, might as well make this a f-string while we're making changes here:

```suggestion
            print(f"Rendered page {page_number} of the directory {operating_sys}")
```

MasterOdin (changes requested):

scripts/pdf/render.py
blueskyson: Yes, it could. I didn't notice that `stylesheets` accepts string list.

scripts/pdf/render.py
blueskyson: I would prefer to split the for-loop statement into 2 lines:

```python
allmd = sorted(glob.glob(os.path.join(loc, operating_sys, "*.md")))
for page_number, md in enumerate(allmd, start=1):
```

blueskyson:
I've implemented all suggestions and the script still works well.

scripts/pdf/render.py
MasterOdin: I'm fine with splitting it into two lines.

PR #6374, opened 2021-08-16
style-guide: add one Chinese-specific rule

blueskyson:
I forgot to mention this rule last time.

> Use a half-width punctuation to end a sentence when the last character is half-width.

### DIFF
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -109,9 +109,11 @@ The following guidelines are applied to Chinese (zh) and traditional Chinese (zh
    For example, use `开启 shell，进入交互模式` rather than `开启 shell ，进入交互模式`
 4. Use full-width punctuations except for long Latin clauses.  
    For example, use `嗨，你好。` rather than `嗨, 你好.`
-5. Use precise form for technical terms, and do not use unofficial Chinese abbreviations.
+5. Use a half-width period or colon to end a sentence when the last character is half-width.  
+   For example, use `将代码转化为 Python 3.` rather than `将代码转化为 Python 3。`
+6. Use precise form for technical terms, and do not use unofficial Chinese abbreviations.
    For example, use `Facebook` rather than `facebook`, `fb` or `脸书`.
 
-In order to maintain readability and normalization, please comply the 5 rules above as much as possible when translating pages into Chinese.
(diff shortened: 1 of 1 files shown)


marchersimon (approved):
Thanks @blueskyson

contributing-guides/style-guide.md
```diff
@@ -109,9 +109,11 @@ The following guidelines are applied to Chinese (zh) and traditional Chinese (zh
    For example, use `开启 shell，进入交互模式` rather than `开启 shell ，进入交互模式`
 4. Use full-width punctuations except for long Latin clauses.  
    For example, use `嗨，你好。` rather than `嗨, 你好.`
-5. Use precise form for technical terms, and do not use unofficial Chinese abbreviations.
+5. Use a half-width punctuation to end a sentence when the last character is half-width.  
+   For example, use `将代码转化为 Python 3.` rather than `将代码转化为 Python 3。`
```
einverne, line 113: First, thank you for the suggestion, but I'd like to know is any reference to this rule?

blueskyson: Nope, this is the current practice of tldr. Actually this is an unfriendly rule for Chinese.

marchersimon: That's actually my fault for the most part. I wanted to bring this PR forword and all the discussions about it weren't really clear and a bit confusing. I thought anything would be better than nothing and just took what seemed to be the most agreed on. But if it's that bad we should probably change it (again) rather sooner than later.

Unfortunately I don't have a lot of time at the moment, but all it takes is a bit of regex and a lot of time. If we can get a definitive decision on what the rules should look like, I'll get to it as soon as I can.

blueskyson: Anyway, this rule is correct in tldr so far. There is no problem with merging this.

sbrl (approved):
I assume this has been decided previously? No problem with adding this. IIRC there's a Chinese translation of the style guide or something - maybe that could be updated too?

blueskyson:
> I assume this has been decided previously? No problem with adding this. IIRC there's a Chinese translation of the style guide or something - maybe that could be updated too?

@sbrl The Chinese one has this rule. This rule should be added by default but I forgot to do so, therefore I opened this PR.

PR #6379, opened 2021-08-17
svgo: remove deprecated --enable=plugin_name example

1deadpixl:
The `--enable` flag was removed in SVGO 2.0. For now, the only way to enable optional plugins is via config. Setting up a config file is out of scope for tldr so I'm just removing the deprecated flag in this PR.

See: https://github.com/svg/svgo/releases/tag/v2.0.0

### DIFF
--- a/pages/common/svgo.md
+++ b/pages/common/svgo.md
@@ -28,10 +28,6 @@
 
 `svgo {{test.svg}} -o -`
 
-- Optimize a file making sure a given plugin is enabled:
-
-`svgo --enable={{plugin_name}}`
-
 - Show available plugins:
 
 `svgo --show-plugins`


blueskyson (approved):
Thanks for your contribution.

navarroaxel:
@1deadpixl I think you can replace the example with `--config="{multipass:true}"`. Right?

1deadpixl:
@navarroaxel No, that won't work. I just tried it on the latest SVGO version and got an "Cannot find module" error.
The release notes say that "Inlined json config is no longer supported. CLI flags should be used instead."
The `--config` flag is used to load the config from a JS module. It will only take a filepath, not JSON.

navarroaxel (approved):

navarroaxel:
Oh, ok. So I misread the release notes. Thanks @1deadpixl.

PR #6387, opened 2021-08-19
git-branch: replace --show-current with --contains example

navarroaxel:

### DIFF
--- a/pages/common/git-branch.md
+++ b/pages/common/git-branch.md
@@ -11,9 +11,9 @@
 
 `git branch -a`
 
-- Show the name of the current branch:
+- Find what branch a Git commit came from:
 
-`git branch --show-current`
+`git branch --all --contains {{commit_hash}}`
 
 - Create new branch based on the current commit:
 


pages/common/git-branch.md
```diff
@@ -11,9 +11,9 @@
 
 `git branch -a`
 
-- Show the name of the current branch:
+- Find what branch a Git commit came from:
```
CleanMachine1, line 14:
```suggestion
- Find which branch a Git commit came from:
```

marchersimon:
I'm not sure if we should remove `--show-current`. If you're not using a fancy shell theme which shows your branch, you really need that command (Or over `ssh`).

CleanMachine1:
Using git-branch on its own displays all branches and stars the one you are currently on, that is good enough IMO

pages/common/git-branch.md
marchersimon: Another suggetion:
```suggestion
- Show the branch a Git commit came from:
```

navarroaxel:
```suggestion
- List what branches includes a specific Git commit in their history:
```

sbrl: A commit may be part of the history of multiple branches. What about this?

```suggestion
- List which branches include a specific Git commit in their history:
```

PR #6389, opened 2021-08-21
ts: add page

oliver:

### DIFF
--- a/pages/common/ts.md
+++ b/pages/common/ts.md
@@ -1,0 +1,20 @@
+# ts
+
+> Add timestamps to input lines.
+> More information: <https://joeyh.name/code/moreutils/>.
+
+- Add current timestamp to beginning of each line:
+
+`{{some_command}} | ts`
+
+- Add timestamps with higher precision (microseconds):
+
+`{{some_command}} | ts "%b %d %H:%M:%.S"`
+
+- Add [i]ncremental timestamps, and show microseconds:
+
+`{{some_command}} | ts -i "%H:%M:%.S"`
+
+- Convert existing timestamps in log file into [r]elative format:
+
+`cat /var/log/syslog | ts -r`


pages/common/ts.md
```diff
@@ -0,0 +1,20 @@
+# ts
+
+> Add timestamps to input lines.
+> More information: <https://joeyh.name/code/moreutils/>.
+
+- Add current timestamp to beginning of each line:
+
+`{{some_command}} | ts`
+
+- Add timestamps with higher precision (microseconds):
```
mfrw, line 10: How about
```suggestion
- Add timestamps with microsecond precision:
```

pages/common/ts.md
```diff
@@ -0,0 +1,20 @@
...
+
+`{{some_command}} | ts "%b %d %H:%M:%.S"`
+
+- Add [i]ncremental timestamps, and show microseconds:
```
mfrw, line 14:
```suggestion
- Add [i]ncremental timestamps with microsecond precision:
```

pages/common/ts.md
```diff
@@ -0,0 +1,20 @@
+# ts
+
+> Add timestamps to input lines.
```
marchersimon, line 3:
```suggestion
> Add timestamps to every line from standard input.
```
I think that explaines the command a bit better

CleanMachine1:
Hi thanks for your contribution! 

Does this command have use cases outside of piping?

PS. Never seen an account as old as yours + what is probably a very sought after username

oliver:
@CleanMachine1: you're right, this command is probably only useful with pipes. Do you think this should be mentioned somewhere in the tldr page?

Btw. thank _you_ for your work on tldr! Names don't really matter IMO, but actions do :-)

pages/common/ts.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Add current timestamp to beginning of each line:
+
+`{{some_command}} | ts`
+
+- Add timestamps with microsecond precision:
+
+`{{some_command}} | ts "%b %d %H:%M:%.S"`
+
+- Add [i]ncremental timestamps with microsecond precision:
```
marchersimon, line 14:
```suggestion
- Add [i]ncremental timestamps with microsecond precision, starting from zero:
```
I think that better explains what incremental means.

pages/common/ts.md
```diff
@@ -0,0 +1,20 @@
...
+> Add timestamps to input lines.
+> More information: <https://joeyh.name/code/moreutils/>.
+
+- Add current timestamp to beginning of each line:
+
+`{{some_command}} | ts`
+
+- Add timestamps with microsecond precision:
+
+`{{some_command}} | ts "%b %d %H:%M:%.S"`
```
marchersimon, line 12:
```suggestion
`{{some_command}} | ts "{{%b %d %H:%M:%.S}}"`
```
Any user-provided input  should go in token syntax

pages/common/ts.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Add [i]ncremental timestamps with microsecond precision:
+
+`{{some_command}} | ts -i "%H:%M:%.S"`
```
marchersimon, line 16:
```suggestion
`{{some_command}} | ts -i "{{%H:%M:%.S}}"`
```

pages/common/ts.md
oliver: OK, if that's the rule in tldr, I'm fine with the change. I still think this line would be more helpful without brackets, because this specific use case (showing microseconds) is probably quite common, and it would be easier to just copy the line if there were no brackets.

pages/common/ts.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Add timestamps with higher precision (microseconds):
+
+`{{some_command}} | ts "%b %d %H:%M:%.S"`
+
+- Add [i]ncremental timestamps, and show microseconds:
+
+`{{some_command}} | ts -i "%H:%M:%.S"`
+
+- Convert existing timestamps in log file into [r]elative format:
```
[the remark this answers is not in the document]
oliver: I agree that the explanation should not be limited to "log files"; but I also think it would be good to mention that this feature is especially useful for log files. How about this:

```suggestion
- Convert existing timestamps in a text file (eg. a log file) into [r]elative format:
```

Or does that make the sentence too complicated? Or is it self-evident anyway that this is useful for log files?

pages/common/ts.md
marchersimon: Those brackets won't be rendered. This is what this line will look like:
image

pages/common/ts.md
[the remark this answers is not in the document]
marchersimon: Good question. @CleanMachine1 what do you think?

pages/common/ts.md
oliver: Oh, that looks fine indeed! I use the `tldr` command from Ubuntu packages (which is https://github.com/psibi/tldr-hs), and it shows the brackets. With https://github.com/tldr-pages/tldr-python-client this looks much better.

pages/common/ts.md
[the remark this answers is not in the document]
sbrl: Sounds like a good idea!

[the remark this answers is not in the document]
marchersimon: @oliver go for it then

marchersimon (approved):
Looks good, thanks!

PR #6396, opened 2021-08-22
ocrmypdf: updated command for in-place OCR

SwampertX:
Updated the [outdated](https://ocrmypdf.readthedocs.io/en/latest/cookbook.html#modify-a-file-in-place) command.

### DIFF
--- a/pages/common/ocrmypdf.md
+++ b/pages/common/ocrmypdf.md
@@ -9,7 +9,7 @@
 
 - Replace a scanned PDF file with a searchable PDF file:
 
-`ocrmypdf {{path/to/input.pdf}}`
+`ocrmypdf {{path/to/input.pdf}} {{path/to/input.pdf}}`
 
 - Skip pages of a mixed-format input PDF file that already contain text:
 
--- a/pages/common/ocrmypdf.md
+++ b/pages/common/ocrmypdf.md
@@ -9,7 +9,7 @@
 
 - Replace a scanned PDF file with a searchable PDF file:
 
-`ocrmypdf {{path/to/input.pdf}} {{path/to/input.pdf}}`
+`ocrmypdf {{path/to/file.pdf}} {{path/to/file.pdf}}`
 
 - Skip pages of a mixed-format input PDF file that already contain text:
 


pages/common/ocrmypdf.md
```diff
@@ -9,7 +9,7 @@
 
 - Replace a scanned PDF file with a searchable PDF file:
 
-`ocrmypdf {{path/to/input.pdf}}`
+`ocrmypdf {{path/to/input.pdf}} {{path/to/input.pdf}}`
```
marchersimon, line 12:
```suggestion
`ocrmypdf {{path/to/input.pdf}} {{path/to/output.pdf}}`
```
Shouldn't that be outbut.pdf?

SwampertX: This is supposed to be a in-place change, or a "OCR and Replace" function. There is no mistake in the command, but is there perhaps a better way to phrase the description?

marchersimon: So do you have to specify the same filename twice? If those files are not the same, the placeholders also shouldn't be the same.

SwampertX: Yes, you do have to specify this same name twice. You can refer to the docs here: https://ocrmypdf.readthedocs.io/en/latest/cookbook.html#modify-a-file-in-place. We could remove this command altogether as well, since it might be familiar to the user that to replace, you can just replace the "output" with the file of the same name.

daniel-birket: Thank you,  SwapertX. You are correct. Input name file twice to replace the file (if successful). I misread the cookbook example that you cited and I cribbed from.

marchersimon: Ah, I didn't see that the first example already shows how to create a new file. Otherwise I would've said we should make it more generic and use two different file names. 
But since neither of those files are input or output, I'd just say
```suggestion
`ocrmypdf {{path/to/file.pdf}} {{path/to/file.pdf}}`
```

SwampertX: Good suggestion! Sorry for the confusion, seems better now.

marchersimon (approved):
Thanks @SwampertX

navarroaxel (approved):

PR #6398, opened 2021-08-22
readpe: add page

dhondta:
- [X] The page (if new), does not already exist in the repository.

### DIFF
--- a/pages/linux/readpe.md
+++ b/pages/linux/readpe.md
@@ -3,7 +3,7 @@
 > Displays information about PE files.
 > More information: <https://manned.org/readpe>.
 
-- Display all information about the PE file:
+- Display all information about a PE file:
 
 `readpe {{path/to/executable}}`
 
--- a/pages/linux/readpe.md
+++ b/pages/linux/readpe.md
@@ -5,24 +5,24 @@
 
 - Display all information about the PE file:
 
-`readpe {{path/to/binary.exe}}`
+`readpe {{path/to/executable}}`
 
 - Display all the headers present in the PE file:
 
-`readpe --all-headers {{path/to/binary.exe}}`
+`readpe --all-headers {{path/to/executable}}`
 
 - Display all the sections present in the PE file:
 
-`readpe --all-sections {{path/to/binary.exe}}`
+`readpe --all-sections {{path/to/executable}}`
 
 - Display a specific header from the PE file:
 
-`readpe --header {{dos|coff|optional}} {{path/to/binary.exe}}`
+`readpe --header {{dos|coff|optional}} {{path/to/executable}}`
 
 - Display imported functions:
 
-`readpe --imports {{path/to/binary.exe}}`
+`readpe --imports {{path/to/executable}}`
 
 - Display exported functions:
 
-`readpe --exports {{path/to/binary.exe}}`
+`readpe --exports {{path/to/executable}}`
--- a/pages/linux/readpe.md
+++ b/pages/linux/readpe.md
@@ -1,0 +1,28 @@
+# readpe
+
+> Displays information about PE files.
+> More information: https://manpages.debian.org/testing/pev/readpe.1.en.html
+
+- Display all information about the PE file:
+
+`readpe {{path/to/binary.exe}}`
+
+- Display all the headers present in the PE file:
+
+`readpe --all-headers {{path/to/binary.exe}}`
+
+- Display all the sections present in the PE file:
+
+`readpe --all-sections {{path/to/binary.exe}}`
+
+- Display a specific header from the PE file:
+
+`readpe --header {{dos|coff|optional}} {{path/to/binary.exe}}`
+
+- Display imported functions:
+
+`readpe --imports {{path/to/binary.exe}}`
+
+- Display exported functions:
+
+`readpe --exports {{path/to/binary.exe}}`
--- a/pages/linux/readpe.md
+++ b/pages/linux/readpe.md
@@ -15,7 +15,7 @@
 
 `readpe --all-sections {{path/to/executable}}`
 
-- Display a specific header from the PE file:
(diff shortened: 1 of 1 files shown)


pages/linux/readpe.md
```diff
@@ -0,0 +1,28 @@
+# readpe
+
+> Displays information about PE files.
```
marchersimon, line 3: Does PE stand for `Portable Executable`? If yes I'd say we should mention that:
```suggestion
> Displays information about PE (Portable Executable) files.
```

pages/linux/readpe.md
```diff
@@ -0,0 +1,28 @@
...
+> More information: <https://manpages.debian.org/testing/pev/readpe.1.en.html>.
```
marchersimon, line 4:
```suggestion
> More information: <https://manpages.debian.org/latest/pev/readpe.1.html>.
```

pages/linux/readpe.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Display all information about the PE file:
```
marchersimon, line 6:
```suggestion
- Display all information about a PE file:
```

pages/linux/readpe.md
```diff
@@ -0,0 +1,28 @@
...
+
+`readpe {{path/to/binary.exe}}`
```
marchersimon, line 8: Just out of interest: is this the same `.exe` as Windows uses for executables?

pages/linux/readpe.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Display all the headers present in the PE file:
```
marchersimon, line 10:
```suggestion
- Display all the headers present in a PE file:
```

pages/linux/readpe.md
```diff
@@ -0,0 +1,28 @@
...
+
+`readpe --all-headers {{path/to/binary.exe}}`
+
+- Display all the sections present in the PE file:
```
marchersimon, line 14:
```suggestion
- Display all the sections present in a PE file:
```

pages/linux/readpe.md
```diff
@@ -0,0 +1,28 @@
...
+
+`readpe --all-sections {{path/to/binary.exe}}`
+
+- Display a specific header from the PE file:
```
marchersimon, line 18:
```suggestion
- Display a specific header from a PE file:
```

pages/linux/readpe.md
```diff
@@ -0,0 +1,28 @@
...
+
+`readpe --header {{dos|coff|optional}} {{path/to/binary.exe}}`
+
+- Display imported functions:
```
marchersimon, line 22: What exactly are thos imported and exported functions. Would be nice if you could explain that a little more.

einverne (commented):
Looks good to me after @marchersimon 's suggestions.

pages/linux/readpe.md
navarroaxel: Why not?
```suggestion
> More information: <https://manned.org/readpe>.
```

marchersimon: Right, forgot about that. And it doesn't seem like `readpe` comes from debian.

pages/linux/readpe.md
dhondta: Cfr [this page](https://github.com/tldr-pages/tldr/blob/main/pages/linux/readelf.md) ; ELF is not specified.
I think the meaning of PE is straightforward for people using this tool...

pages/linux/readpe.md
dhondta: Same remark as before...

pages/linux/readpe.md
dhondta: Same remark as before...

pages/linux/readpe.md
dhondta: This is something you can isolate using this tool. For more information about the imported or exported functions from a portable executable, you can for instance refer to [this article](https://blog.kowalczyk.info/articles/pefileformat.html).

pages/linux/readpe.md
dhondta: I modified to `readpe {{path/to/executable}}` ; I realized it is not about only `.exe` files, but also `.dll` for instance.

pages/linux/readpe.md
marchersimon: That page is quite old and doesn't really comply to our conventions yet. But when you say `the file` it soudns like there exists this one file, which is somehow special, whereas `a file` could be any file.

pages/linux/readpe.md
marchersimon: Then we should remove the `.exe` part. However if it's only `.exe` and `.dll`, we could add that to the description.

pages/linux/readpe.md
marchersimon: Doesn't sound like there's an easier way to describe it. Also, I don't assume this will print the whole function in binary, but just list the name or something about it. In this case we should use:
```suggestion
- List all imported functions:
```

pages/linux/readpe.md
dhondta: OK, let's do this way then.

pages/linux/readpe.md
dhondta: OK, excellent remark, thank you.

pages/linux/readpe.md
dhondta: Beware that it is not only about EXE and DLL's, but also OCX, SYS, SCR, DRV or EFI.
I suggest to keep it like this : `{{path/to/executable}}`.

CleanMachine1 (approved):

navarroaxel (approved):

marchersimon (approved):
Thanks @dhondta

PR #6432, opened 2021-08-28
adscript: add page

pixelcmtd:

### DIFF
--- a/pages/common/adscript.md
+++ b/pages/common/adscript.md
@@ -9,12 +9,12 @@
 
 - Compile and link a file to a standalone executable:
 
-`adscript --executable --output {{path/to/program}} {{path/to/program.adscript}}`
+`adscript --executable --output {{path/to/file}} {{path/to/input_file.adscript}}`
 
 - Compile a file to LLVM IR instead of native machine code:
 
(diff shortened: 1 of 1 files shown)


CleanMachine1 (approved):

blueskyson (commented):
I think it would be better to replace the word `file` with `source code` or `Adscript`.

pixelcmtd:
> I think it would be better to replace the word `file` with `source code` or `Adscript`.

@blueskyson  I don't think that would sound nice, I would prefer something like `source file`. How do you feel about that one?

blueskyson (approved):

pages/common/adscript.md
```diff
@@ -0,0 +1,20 @@
+# adscript
+
+> Compiler for Adscript files.
+> More information: <https://github.com/Amplus2/Adscript>.
+
+- Compile a file to an object file:
+
+`adscript --output {{path/to/program.o}} {{path/to/program.adscript}}`
```
navarroaxel, line 8: We don't use to use `program` for filenames :thinking:. Could we use this?
```suggestion
`adscript --output {{path/to/file.o}} {{path/to/input_file.adscript}}`
```

PR #6461, opened 2021-09-03
wp: add download and config examples

iSnackyCracky:
Add commands for installing a fresh WordPress System

### DIFF
--- a/pages/common/wp.md
+++ b/pages/common/wp.md
@@ -11,9 +11,9 @@
 
 `wp cli update`
 
-- Download a fresh WordPress installation to current directory:
+- Download a fresh WordPress installation to current directory, optionally specifying the locale:
 
-`wp core download [--locale=<locale>]`
+`wp core download --locale={{locale}}`
 
 - Create basic wpconfig file:
 
--- a/pages/common/wp.md
+++ b/pages/common/wp.md
@@ -15,7 +15,7 @@
 
 `wp core download --locale={{locale}}`
 
-- Create basic wpconfig file:
+- Create basic `wpconfig` file:
 
 `wp config create --dbname=<dbname> --dbuser=<dbuser> [--dbpass=<dbpass>] [--dbhost=<dbhost>]`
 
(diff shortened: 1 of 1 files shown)


pages/common/wp.md
```diff
@@ -11,6 +11,14 @@
 
 `wp cli update`
 
+- Download a fresh WordPress installation to current directory:
+
+`wp core download [--locale=<locale>]`
```
marchersimon, line 16: We don't really have syntax for optional arguments, but we can mention it in the description. Also, we use token syntax for everything which the user has to input:
```suggestion
- Download a fresh WordPress installation to current directory, optionally specifying the locale:

`wp core download --locale={{locale}}`
```

pages/common/wp.md
```diff
@@ -11,6 +11,14 @@
...
+
+- Create basic wpconfig file:
```
marchersimon, line 18: We use backticks for technical terms.
```suggestion
- Create basic `wpconfig` file:
```

pages/common/wp.md
```diff
@@ -11,6 +11,14 @@
...
+
+`wp config create --dbname=<dbname> --dbuser=<dbuser> [--dbpass=<dbpass>] [--dbhost=<dbhost>]`
```
marchersimon, line 20: As already said, we can't use the `[ ]` syntax from man pages. Are those last two flags useful? Either we write them without it or we remove them if they're not that important.

pages/common/wp.md
iSnackyCracky: Ah, okay. Yeah that works fine for me 👍

pages/common/wp.md
iSnackyCracky: I'd guess the --dbpass is important enough to specify it, --dbhost however probably isn't.

navarroaxel: These are very common DB arguments, I think we can left all.

marchersimon (approved):
Thanks @iSnackyCracky

navarroaxel (approved):

PR #6462, opened 2021-09-03
Reorganize translation templates

marchersimon:
Since we're now getting more of those templates we can't just dump them all into `contributing-guides`. I created a directory `contributing-guides/translation-templates`, where we can have one file per template. There was also the idea to put everything into a single file, but that's still open for discussion. Personally I think that would get messy after some time.


blueskyson (approved):
LGTM.

sbrl (approved):

PR #6471, opened 2021-09-03
pages/common/*.md: REPL Normalization

pixelcmtd:
Normalized almost all references to a REPL to `Start a REPL (interactive shell)`.

I think that it is unacceptable for any serious project to have so many conflicting styles:

- `Start a(n) $LANG interactive shell (REPL)`
- `Run interactive REPL`
- `Start a REPL (interactive shell)`
- `Start a REPL`
- `Start the $LANG REPL`
- `Start a(n) $LANG REPL session`
- `Open an Interactive $LANG Shell (REPL)`
- (maybe more)

You might have cases in this diff, where you think that different descriptions might fit better and that is ok. But most of these should be normalized.

### DIFF
--- a/pages/common/cake.md
+++ b/pages/common/cake.md
@@ -31,6 +31,6 @@
 
 `cake server`
 
-- Start a REPL interactive shell instance:
+- Start a REPL (interactive shell):
 
 `cake console`
--- a/pages/common/clj.md
+++ b/pages/common/clj.md
@@ -4,7 +4,7 @@
 > All options can be defined in a `deps.edn` file.
 > More information: <https://clojure.org/guides/deps_and_cli>.
 
-- Start a REPL:
+- Start a REPL (interactive shell):
 
 `clj`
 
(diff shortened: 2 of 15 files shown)


blueskyson:
I personally prefer `Start a(n) $LANG interactive shell (REPL)`. This is the first see the acronym `REPL`. Is this acronym used frequently in programming communities?

marchersimon:
I don't have a strong opinion here. I think it would be good to include what `REPL` stands for (`Read-Evaluete-Print-Loop`), but I'm not sure if that's important. But in this way, is including `REPL` even important? Why do we even include it?

pixelcmtd:
> I think it would be good to include what `REPL` stands for (`Read-Evaluete-Print-Loop`), but I'm not sure if that's important.

I don't think that it is that important, it would really bloat the description.



> But in this way, is including `REPL` even important? Why do we even include it?

Because it is the proper word for it. If I read some weird rambling about `interactive shell`s, I'm gonna be really confused.

CleanMachine1 (approved):

marchersimon (approved):
Thanks @pixelcmtd

PR #6478, opened 2021-09-04
udevadm: add page

aos:
Noticed this wasn't in your database! I use this somewhat frequently :-)

### DIFF
--- a/pages/linux/udevadm.md
+++ b/pages/linux/udevadm.md
@@ -1,11 +1,19 @@
 # udevadm
 
 > Linux `udev` management tool.
-> More information: <https://linux.die.net/man/8/udevadm>.
+> More information: <https://www.freedesktop.org/software/systemd/man/udevadm>.
 
-- Monitor device events:
+- Monitor all device events:
 
-`sudo udevadm monitor [--kernel] [--udev]`
+`sudo udevadm monitor`
+
+- Print `uevents` sent out by the kernel:
+
+`sudo udevadm monitor --kernel`
+
+- Print device events after rule processing by `udev`:
+
+`sudo udevmadm monitor --udev`
 
 - List attributes of a device:
 
--- a/pages/linux/udevadm.md
+++ b/pages/linux/udevadm.md
@@ -11,9 +11,9 @@
 
 `sudo udevadm monitor --kernel`
 
-- Print device events after rule processing by `udev`:
+- Print device events after rules are processed by `udev`:
 
-`sudo udevmadm monitor --udev`
+`sudo udevadm monitor --udev`
 
 - List attributes of a device:
 
--- a/pages/linux/udevadm.md
+++ b/pages/linux/udevadm.md
@@ -11,7 +11,7 @@
 
 `sudo udevadm monitor --kernel`
 
-- Print device events after rules are processed by `udev`:
+- Print device events after being processed by `udev`:
 
 `sudo udevadm monitor --udev`
 
--- a/pages/linux/udevadm.md
+++ b/pages/linux/udevadm.md
@@ -1,0 +1,20 @@
+# udevadm
+
+> Linux `udev` management tool.
+> More information: <https://linux.die.net/man/8/udevadm>.
+
+- Monitor device events:
+
+`sudo udevadm monitor [--kernel] [--udev]`
+
+- List attributes of a device:
+
+`sudo udevadm info --attribute-walk --path {{/dev/sda1}}`
+
+- Reload all `udev` rules:
+
+`sudo udevadm control --reload-rules`
+
+- Trigger all `udev` rules to run:
+
+`sudo udevadm trigger`


pages/linux/udevadm.md
```diff
@@ -0,0 +1,20 @@
+# udevadm
+
+> Linux `udev` management tool.
+> More information: <https://linux.die.net/man/8/udevadm>.
+
+- Monitor device events:
+
+`sudo udevadm monitor [--kernel] [--udev]`
```
navarroaxel, line 8: We don't use `[]` for examples, I think you should split this example to explain the command without params, another for `--kernel` and another for `--udev`.

pages/linux/udevadm.md
```diff
@@ -0,0 +1,20 @@
+# udevadm
+
+> Linux `udev` management tool.
+> More information: <https://linux.die.net/man/8/udevadm>.
```
marchersimon, line 4: This is what we use for man pages:
```suggestion
> More information: <https://manned.org/udevadm>.
```

marchersimon:
@aos there's no need to do force-pushs. We squash all commits before merging.

pages/linux/udevadm.md
navarroaxel: We should use the official website:
```suggestion
> More information: <https://www.freedesktop.org/software/systemd/man/udevadm>.
```

aos:
Thank you for the feedback -- I addressed your comments!

pages/linux/udevadm.md
```diff
@@ -0,0 +1,28 @@
...
+
+`sudo udevadm monitor`
+
+- Print `uevents` sent out by the kernel:
+
+`sudo udevadm monitor --kernel`
+
+- Print device events after rule processing by `udev`:
+
+`sudo udevmadm monitor --udev`
```
blueskyson, line 16:
```suggestion
- Print device events after rules processed by `udev`:

`sudo udevadm monitor --udev`
```

marchersimon: I don't see a problem with `after rule processing`. But maybe we could still rephrase it. What are those rules for? Do they apply on the devices or on the events? Is ``Print device events after being filtered by `udev` `` correct?

aos: I don't see a problem with it either, but I'm happy to take suggestions.

The rules are meant to decide how the device is interfaced with in user-space. You create a rule like this and add it to `/etc/udev/rules.d/`:
```
SUBSYSTEMS=="usb", ATTRS{idVendor}=="1050", ATTRS{idProduct}=="0407", SYMLINK+="/dev/myDevice"
```
Then you plug a device in, the kernel passes up the `uevents`, and `udev` processes them. If they match (`==`), it runs an action. In this case, symlinking to `/dev/myDevice`. Passing in the `--udev` rule, will print any matching rules that **were** processed.

marchersimon: Some suggestions:
> Print device events and all matched `udev` rules

> Print device events along with all matched `udev` rules

> Print device events and `udev` rules in case one matched

I hope I understand this correctly. Or does it only print device events which had a matchnig rule?

aos: It doesn't print the `udev` rule specifically. It prints the device events that `udev` sends out after processing the rule. Here's an example of me unplugging a USB drive:
```
$ sudo udevadm monitor -u
monitor will print the received events for:
UDEV - the event which udev sends out after rule processing

UDEV  [156502.669688] remove   /devices/pci0000:00/0000:00:1d.0/0000:03:00.0/0000:04:01.0/0000:06:00.0/0000:07:01.0/0000:09:00.0/usb7/7-2/7-2.1/7-2.1:1.0
```

marchersimon: Then I think we can either keep the original or change it to 
> Print device events after being processed by `udev`.

aos: Sounds good -- I'll go ahead and use your suggestion

pages/linux/udevadm.md
```diff
@@ -0,0 +1,28 @@
...
+
+- List attributes of a device:
+
+`sudo udevadm info --attribute-walk --path {{/dev/sda1}}`
+
+- Reload all `udev` rules:
+
+`sudo udevadm control --reload-rules`
+
+- Trigger all `udev` rules to run:
```
sbrl, line 26: Hey there! Welcome to tldr-pages, @aos.

What precisely does this do?

aos: Howdy ! This "re-applies" all `udev` rules to run. So here's the general workflow:

1. You write a new rule and add it to `/etc/udev/rules.d/`
2. You reload all rules using `udevadm control --reload-rules` so that `udev` stores it in its in-memory database (this only usually happens on boot)
3. You tell `udev` to re-apply all the rules in its database, or "trigger" them to run

marchersimon (approved):
Thanks @aos

pages/linux/udevadm.md
sbrl: Ah, ok. Thanks for the explanation! This seems ok to me then.

sbrl (approved):

sbrl:
Thanks, @aos!

PR #6481, opened 2021-09-06
transfer.sh: add page

marchersimon:
I'm not sure if we really can take this page, because `transfer.sh` isn't a command.

### DIFF
--- a/pages/common/transfer.sh.md
+++ b/pages/common/transfer.sh.md
@@ -1,0 +1,16 @@
+# transfer.sh
+
+> Command line client for transfer.sh.
+> More information: <https://github.com/dutchcoders/transfer.sh/>.
+
+- Upload a file to transfer.sh and print its URL:
+
+`curl --updload-file {{path/to/file}} https://transfer.sh/{{file_name}}`
+
+- Upload a file to transfer.sh showing a progress bar:
+
+`curl --progress-bar --updload-file {{path/to/file}} https://transfer.sh/{{file_name}}`
+
+- Download a file from transfer.sh:
+
+`curl https://transfer.sh/{{token}}/{{file_name}}`


CleanMachine1:
I have a for and against for this.
For: Why reject anything, it doesn't do any harm
Against: Doesn't fit TLDR at all really, its not a sub-command or command.

marchersimon:
That's what I thought too. Let's get some more opinions.

blueskyson:
How about create a new directory `not-a-command` to store not-a-command pages like `translate.sh`:

```
└── pages
          └──not-a-command
                    └──translate.sh.md
```

marchersimon:
I think that would complicate things too much and would go beyond the scope of tldr.

pages/common/transfer.sh.md
```diff
@@ -0,0 +1,16 @@
+# transfer.sh
+
+> Command line client for transfer.sh.
+> More information: <https://github.com/dutchcoders/transfer.sh/>.
```
navarroaxel, line 4:
```suggestion
> More information: <https://github.com/dutchcoders/transfer.sh>.
```

pages/common/transfer.sh.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Upload a file to transfer.sh and print its URL:
+
+`curl --updload-file {{path/to/file}} https://transfer.sh/{{file_name}}`
```
navarroaxel, line 8:
```suggestion
`curl --upload-file {{path/to/file}} https://transfer.sh/{{filename}}`
```

pages/common/transfer.sh.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Upload a file to transfer.sh showing a progress bar:
+
+`curl --progress-bar --updload-file {{path/to/file}} https://transfer.sh/{{file_name}}`
```
navarroaxel, line 12:
```suggestion
`curl --progress-bar --upload-file {{path/to/file}} https://transfer.sh/{{filename}}`
```

pages/common/transfer.sh.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Download a file from transfer.sh:
+
+`curl https://transfer.sh/{{token}}/{{file_name}}`
```
navarroaxel, line 16:
```suggestion
`curl https://transfer.sh/{{token}}/{{filename}}`
```

navarroaxel (commented):
I think this service should have a wrapper command to simplify the `curl` usage.

sbrl:
Yeah unfortunately, I don't think this page is particularly suitable for tldr. However, if there was a server binary called `transfer.sh`, then it would be in-scope to document that and then tag onto the end some an example or 2 for clients.

marchersimon:
There is a Python client `transfersh`. I'll document this one instead.

PR #6484, opened 2021-09-07
diff: add -Naur create patch file example

mmahmoudian:

`diff` is the most common and easiest way to create patch files apart from using `git`, therefore imho the example was missing from the tldr page.

### DIFF
--- a/pages/common/diff.md
+++ b/pages/common/diff.md
@@ -26,3 +26,7 @@
 - Compare directories, only showing the names of files that differ:
 
 `diff -rq {{old_directory}} {{new_directory}}`
+
+- Create patch file from the differences of two directories or files:
+
+`diff -Naur {{old}} {{new}} > differences.patch`
--- a/pages/common/diff.md
+++ b/pages/common/diff.md
@@ -27,6 +27,6 @@
 
 `diff --recursive --brief {{old_directory}} {{new_directory}}`
 
-- Create a patch file for Git:
+- Create a patch file for Git from the differences of two text files, treating nonexistent files as empty:
 
 `diff --text --unified --new-file {{old_file}} {{new_file}} > {{diff.patch}}`
--- a/pages/common/diff.md
+++ b/pages/common/diff.md
@@ -27,6 +27,6 @@
 
 `diff --recursive --brief {{old_directory}} {{new_directory}}`
 
-- Create patch file from the differences of two directories or files:
+- Create a patch file for Git:
 
-`diff --text --unified --new-file {{old_file_or_directory}} {{new_file_or_directory}} > {{diff.patch}}`
+`diff --text --unified --new-file {{old_file}} {{new_file}} > {{diff.patch}}`
--- a/pages/common/diff.md
+++ b/pages/common/diff.md
@@ -9,24 +9,24 @@
 
 - Compare files, ignoring white spaces:
 
-`diff -w {{old_file}} {{new_file}}`
+`diff --ignore-all-space {{old_file}} {{new_file}}`
 
 - Compare files, showing the differences side by side:
 
-`diff -y {{old_file}} {{new_file}}`
+`diff --side-by-side {{old_file}} {{new_file}}`
 
 - Compare files, showing the differences in unified format (as used by `git diff`):
 
-`diff -u {{old_file}} {{new_file}}`
+`diff --unified {{old_file}} {{new_file}}`
 
 - Compare directories recursively (shows names for differing files/directories as well as changes made to files):
 
-`diff -r {{old_directory}} {{new_directory}}`
+`diff --recursive {{old_directory}} {{new_directory}}`
 
 - Compare directories, only showing the names of files that differ:
 
-`diff -rq {{old_directory}} {{new_directory}}`
+`diff --recursive --brief {{old_directory}} {{new_directory}}`
 
 - Create patch file from the differences of two directories or files:
 
-`diff -Naur {{old}} {{new}} > differences.patch`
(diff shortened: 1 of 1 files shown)


pages/common/diff.md
```diff
@@ -26,3 +26,7 @@
 - Compare directories, only showing the names of files that differ:
 
 `diff -rq {{old_directory}} {{new_directory}}`
+
+- Create patch file from the differences of two directories or files:
+
+`diff -Naur {{old}} {{new}} > differences.patch`
```
navarroaxel, line 32: I'm not a fan of too many short arguments in this way, also I think a patch file could be generated in any of these examples as well.

mmahmoudian: I'm not a big fan either, and it is also partly against tldr's guideline, but every other command in the list had it and I followed their steps.

Regarding your last point though, the simple answer is no it cannot, because the following are typically expected in a conventional patch file:
1. `-a, --text` which will treat all files as text
2.  `-N, --new-file` which will treat absent files as empty
3.  `-u, --unified` which brings in the 3 line context to the diff

mmahmoudian: I will be more than happy to change all the arguments to the long form as that is my preference as well.

marchersimon: I agree that we should use long flags everywhere, so please go ahead.
What do you think about this descripption:
> Create a patch file from the difference of two files, treating them as text files and treating absent files as empty:
>
> \`diff --new-file --text {{path/to/file1}} {{path/to/file2}} > {{path/to/diff.patch}}\`

Could need some improvent tough. I left out the `-r`  and `-u `flag, since both were covered by other examples and this one is already quite long.

mmahmoudian:
I converted all short forms to their long forms. Based on @marchersimon I also tried to reduce the number of arguments in the last item.

In @marchersimon suggestion, there were `path/to/` and `file1`/`file2` which I didn't use to be more similar to the other items in the page.

pages/common/diff.md
```diff
@@ -9,20 +9,24 @@
...
+`diff --recursive {{old_directory}} {{new_directory}}`
 
 - Compare directories, only showing the names of files that differ:
 
-`diff -rq {{old_directory}} {{new_directory}}`
+`diff --recursive --brief {{old_directory}} {{new_directory}}`
+
+- Create patch file from the differences of two directories or files:
+
+`diff --text --unified --new-file {{old_file}} {{new_file}} > {{diff.patch}}`
```
navarroaxel, line 32:
```suggestion
`diff --text --unified --new-file {{old_file_or_directory}} {{new_file_or_directory}} > {{diff.patch}}`
```

sbrl (approved):
Thanks, @mmahmoudian!

navarroaxel (approved):

pages/common/diff.md
```diff
@@ -9,20 +9,24 @@
...
+`diff --recursive {{old_directory}} {{new_directory}}`
 
 - Compare directories, only showing the names of files that differ:
 
-`diff -rq {{old_directory}} {{new_directory}}`
+`diff --recursive --brief {{old_directory}} {{new_directory}}`
+
+- Create patch file from the differences of two directories or files:
+
+`diff --text --unified --new-file {{old_file_or_directory}} {{new_file_or_directory}} > {{diff.patch}}`
```
marchersimon, line 32: Sorry for the delay. Here's my last suggestion:
* Since we left out `--recoursive` we have to remove `directory`
* I think we should also leave out `--unified`, because it was covered in the examples above (except it's a standard flag which is used every time)
* Since we now have more space we can explain `--new-file` better.
```suggestion
- Create a patch file from the differences of two text files, treating nonexistent files as empty:

`diff --text --new-file {{old_file}} {{new_file}} > {{diff.patch}}`
```

mmahmoudian: I believe if `--unified` is not there, then the format is not a conventional patch file as there will be no context lines (making patching much more risky.
About the change of description, I agree. Is there any character limit per line?

marchersimon: Do you think it would help if we write `Create a patch file for Git`? Because that's what `--unified` does right?
Regarding the line lenght. We try to keep it as short as possible, but the above shouldn't be a problem.

sbrl: The limit per line I think is quite high, as the `whiptail` page is still ok.

`Create a patch file for Git` sounds good!

mmahmoudian: So something like this would be satisfactory?

```markdown
Create a patch file for Git
`diff --text --unified --new-file {{old_file}} {{new_file}} > {{diff.patch}}`
```

marchersimon: @sbrl why not

> Create a patch file for Git from the differences of two text files, treating nonexistent files as empty

Otherwise we would just throw flags on the user without explaining them

mmahmoudian: if we don't hit the character limit, I agree with @marchersimon

pages/common/diff.md
```diff
@@ -9,20 +9,24 @@
...
 
-`diff -r {{old_directory}} {{new_directory}}`
+`diff --recursive {{old_directory}} {{new_directory}}`
 
 - Compare directories, only showing the names of files that differ:
 
-`diff -rq {{old_directory}} {{new_directory}}`
+`diff --recursive --brief {{old_directory}} {{new_directory}}`
+
+- Create a patch file for Git:
```
marchersimon, line 30:
```suggestion
- Create a patch file for Git from the differences of two text files, treating nonexistent files as empty:
```

mmahmoudian:
Thank you all for the constructive criticisms and brainstorming.

PR #4460, opened 2020-10-04
Added --montage example

mattryanmtl:
- [ ] The page has 8 or fewer examples.

PR #6487, opened 2021-09-07
feh, hostname, fc, ssh-add: move to common

marchersimon:
In #1891, exact copies of `feh` were added to `linux` and `osx`
In d11523189ffef234004c884ff00d8f0c26d869b6, a Chinese version of the `osx` version was added
In #4460, an example was added to the linux version, which should also apply to the osx version


marchersimon:
Basically the same applies for `hostname`.

marchersimon:
And `fc`

marchersimon:
And `ssh-add`

blueskyson (approved):

pixelcmtd (approved):

PR #6494, opened 2021-09-08
clang: add -c example

pixelcmtd:
the `gcc` page also has it and it makes sense


blueskyson (approved):
LGTM. Thanks.

navarroaxel (approved):

sbrl (approved):
Not sure when this would be useful, but thanks for the contribution anyway!

PR #6496, opened 2021-09-08
llvm-as: add page

pixelcmtd:

### DIFF
--- a/pages/common/llvm-as.md
+++ b/pages/common/llvm-as.md
@@ -11,9 +11,9 @@
 
 `llvm-as --module-hash -o {{path/to/out.bc}} {{path/to/source.ll}}`
 
-- Read IR from `stdin` and assemble it:
+- Read an IR file from `stdin` and assemble it:
 
-`llvm-as -o {{path/to/out.bc}}`
+`cat {{path/to/source.ll}} | llvm-as -o {{path/to/out.bc}}`
 
 - Assemble an IR file to `stdout` (might mess up your terminal):
 
--- a/pages/common/llvm-as.md
+++ b/pages/common/llvm-as.md
@@ -5,16 +5,16 @@
 
 - Assemble an IR file:
 
-`llvm-as -o {{out.bc}} {{source.ll}}`
+`llvm-as -o {{path/to/out.bc}} {{path/to/source.ll}}`
 
 - Assemble an IR file and include a module hash in the produced Bitcode file:
 
-`llvm-as --module-hash -o {{out.bc}} {{source.ll}}`
+`llvm-as --module-hash -o {{path/to/out.bc}} {{path/to/source.ll}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/llvm-as.md
```diff
@@ -0,0 +1,20 @@
+# llvm-as
+
+> LLVM Intermediate Representation (`.ll`) to Bitcode (`.bc`) assembler.
+> More information: <https://llvm.org/docs/CommandGuide/llvm-as.html>.
+
+- Assemble an IR file:
+
+`llvm-as -o {{out.bc}} {{source.ll}}`
```
blueskyson, line 8:
```suggestion
`llvm-as -o {{path/to/out.bc}} {{path/to/source.ll}}`
```

blueskyson (commented):
Format as `path/to/file`.

blueskyson (approved):
Thanks for contribution.

pages/common/llvm-as.md
```diff
@@ -0,0 +1,20 @@
...
+
+`llvm-as -o {{path/to/out.bc}} {{path/to/source.ll}}`
+
+- Assemble an IR file and include a module hash in the produced Bitcode file:
+
+`llvm-as --module-hash -o {{path/to/out.bc}} {{path/to/source.ll}}`
+
+- Read IR from `stdin` and assemble it:
+
+`llvm-as -o {{path/to/out.bc}}`
```
marchersimon, line 16: This is how we usually write such examples
```suggestion
- Read an IR file from `stdin` and assemble it:

`cat {{path/to/source.ll}} | llvm-as -o {{path/to/out.bc}}`
```

pages/common/llvm-as.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Assemble an IR file to `stdout` (might mess up your terminal):
```
marchersimon, line 18: Not sure if printing Bitcode to `stdout` is a good idea. Does this example have any use case?

pixelcmtd: You might want to pipe it to `opt` or something like that.

marchersimon: Then could we turn this into a more useful example? That's actually the purpose of tldr.

sbrl: In this case I don't think it would be a particular problem to include an example command that would be useful. You mention `opt` @pixelcmtd - perhaps in this instance we could include this as an explicit example?

pixelcmtd: If you only like it with a program that it pipes to, I'd instead remove it, because then it doesn't show anything new anymore.

An example of how to make a pipe like `clang | llvm-as | opt | llc` should be in an `llvm` page.

marchersimon (approved):
Looks good, thanks

PR #6513, opened 2021-09-12
tailscale-up: add page

navarroaxel:

### DIFF
--- a/pages/common/tailscale-up.md
+++ b/pages/common/tailscale-up.md
@@ -1,7 +1,7 @@
 # tailscale up
 
 > Connects the client to the Tailscale network.
-> In version 1.8, command line arguments are stored and reused until they're overwritten or `--reset` is called.
+> In version 1.8 and above, command line arguments are stored and reused until they're overwritten or `--reset` is called.
 > More information: <https://tailscale.com/kb/admin/>.
 
 - Connect to Tailscale:
--- a/pages/common/tailscale-up.md
+++ b/pages/common/tailscale-up.md
@@ -19,9 +19,9 @@
 
 `sudo tailscale up --shields-up`
 
-- Connect and accept DNS configuration from the admin panel:
+- Connect and don't accept DNS configuration from the admin panel (defaults to `true`):
 
-`sudo tailscale up --accept-dns={{true|false}}`
+`sudo tailscale up --accept-dns={{false}}`
 
 - Connect and configure Tailscale as a subnet router:
 
--- a/pages/common/tailscale-up.md
+++ b/pages/common/tailscale-up.md
@@ -25,7 +25,7 @@
 
 - Connect and configure Tailscale as a subnet router:
 
-`sudo tailscale up --advertise-routes={{10.0.0.0/24,10.0.1.0/24}}`
+`sudo tailscale up --advertise-routes={{10.0.0.0/24}},{{10.0.1.0/24}}`
 
 - Connect and accept subnet routes from Tailscale:
 
--- a/pages/common/tailscale-up.md
+++ b/pages/common/tailscale-up.md
@@ -1,7 +1,7 @@
 # tailscale up
 
 > Connects the client to the Tailscale network.
-> In version 1.8, command line arguments are stored and reused until they're overwritten or `--reset is called.
+> In version 1.8, command line arguments are stored and reused until they're overwritten or `--reset` is called.
 > More information: <https://tailscale.com/kb/admin/>.
 
 - Connect to Tailscale:
--- a/pages/common/tailscale-up.md
+++ b/pages/common/tailscale-up.md
@@ -7,11 +7,11 @@
 
 `sudo tailscale up`
 
-- Connect and offer the current machine to be an exit node for the internet traffic:
+- Connect and offer the current machine to be an exit node for internet traffic:
 
 `sudo tailscale up --advertise-exit-node`
 
-- Connect using a specific node for the internet traffic:
+- Connect using a specific node for internet traffic:
 
 `sudo tailscale up --exit-node={{exit_node_ip}}`
 
--- a/pages/common/tailscale-up.md
+++ b/pages/common/tailscale-up.md
@@ -1,6 +1,7 @@
 # tailscale up
 
 > Connects the client to the Tailscale network.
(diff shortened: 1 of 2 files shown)


pages/common/tailscale-up.md
```diff
@@ -0,0 +1,36 @@
...
+
+`sudo tailscale up --exit-node={{exit_node_ip}}`
+
+- Connect and block incoming connections to the current node:
+
+`sudo tailscale up --shields-up`
+
+- Connect and accept DNS configuration from the admin panel:
+
+`sudo tailscale up --accept-dns={{true|false}}`
```
marchersimon, line 24: Not sure if we should include the default argument here.

pages/common/tailscale-up.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Connect and configure Tailscale as a subnet router:
+
+`sudo tailscale up --advertise-routes={{10.0.0.0/24,10.0.1.0/24}}`
```
marchersimon, line 28: Are always 2 IP addresses required? If so how about
```suggestion
`sudo tailscale up --advertise-routes={{10.0.0.0/24}},{{10.0.1.0/24}}`
```

pages/common/tailscale-up.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Connect and accept subnet routes from Tailscale:
+
+`sudo tailscale up --accept-routes`
+
+- Reset unspecified settings to their default values and connect:
```
marchersimon, line 34: So when I example use `--accept-dns=true` this value will be saved until I call `--reset`? I think we should mention that in the description.

pages/common/tailscale-up.md
navarroaxel: Then we should add ``(defaults to `true`)`` in the example description:
```suggestion
`sudo tailscale up --accept-dns={{false}}`
```

pages/common/tailscale-up.md
navarroaxel: I'm not sure how the CLI args are saved, but I tried a connection with `--exit-node` and then I had to use `--reset` because without `--exit-node` the connection wasn't working.

pages/common/tailscale-up.md
marchersimon: But then, we should also change the description to

> Connect and don't accept DNS configuration from the admin panel:

pages/common/tailscale-up.md
sbrl: Hmmm, that is very odd behaviour. What about this?
```suggestion
- Reset unspecified settings to their default values and connect (CLI arguments are persisted for future calls):
```

navarroaxel: I found this: https://github.com/tailscale/tailscale/issues/1927#issuecomment-845083085

> in 1.8 the CLI saves the options it has been invoked with, and will only change them if explicitly told to do so. In prior releases if one had used: tailscale up --accept-routes and then later used tailscale up --advertise-exit-node, the --accept-routes would be silently discarded and this led to bad outcomes.
> 
> You should be able to use tailscale up --advertise-exit-node=false to explicitly clear the flag if it is no longer wanted.

pages/common/tailscale-up.md
```diff
@@ -0,0 +1,36 @@
+# tailscale up
+
+> Connects the client to the Tailscale network.
+> More information: <https://tailscale.com/kb/admin/>.
+
+- Connect to Tailscale:
+
+`sudo tailscale up`
+
+- Connect and offer the current machine to be an exit node for the internet traffic:
```
CleanMachine1, line 10:
```suggestion
- Connect and offer the current machine to be an exit node for internet traffic:
```

pages/common/tailscale-up.md
```diff
@@ -0,0 +1,36 @@
...
+
+`sudo tailscale up --advertise-exit-node`
+
+- Connect using a specific node for the internet traffic:
```
CleanMachine1, line 14:
```suggestion
- Connect using a specific node for internet traffic:
```

CleanMachine1 (changes requested):

pages/common/tailscale-up.md
sbrl: Sounds good @marchersimon

pages/common/tailscale-up.md
```diff
@@ -1,6 +1,7 @@
 # tailscale up
 
 > Connects the client to the Tailscale network.
+> In version 1.8, command line arguments are stored and reused until they're overwritten or `--reset is called.
```
marchersimon, line 4:
```suggestion
> In version 1.8, command line arguments are stored and reused until they're overwritten or `--reset` is called.
```

marchersimon (approved):
Thanks, looks good

pages/common/tailscale-up.md
```diff
@@ -0,0 +1,37 @@
+# tailscale up
+
+> Connects the client to the Tailscale network.
+> In version 1.8, command line arguments are stored and reused until they're overwritten or `--reset` is called.
```
sbrl, line 4: Is this only version 1.8, or is it 1.8 and above?

navarroaxel: above

sbrl: Perhaps we could update this to explain that then?

marchersimon:
```suggestion
> In version 1.8 and above, command line arguments are stored and reused until they're overwritten or `--reset` is called.
```

CleanMachine1 (approved):
Interesting command, but I can't see its exact use? Is it a VPN service for standard users or is it a business VPN service for accessing networks, regardless LGTM

navarroaxel:
> Interesting command, but I can't see its exact use? Is it a VPN service for standard users or is it a business VPN service for accessing networks, regardless LGTM

It's service to create connections between your remote machines, like the NAS in your home, or connect to a VPC in AWS, use a remote machine as exit node. Connect to your home network as another a subnet.

sbrl (approved):
* VPC → VPS

CleanMachine1:
Interesting, thanks for the insight

PR #6530, opened 2021-09-15
script/test: ignore python checks if not available

navarroaxel:
Ignore black and flake8 linters if these are not
installed in the dev system.

The git `pre-commit` hook run `scripts/test.sh`, and now the shell script wants `black` and `flake8` commands in the dev computer. I'm not a python dev and I don't use that commands, I think we should check if `black` is available before run it. Otherwise, we are blocking new contributors to contribute because they need to configure python stuff.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -9,6 +9,40 @@
 # NOTE: must be run from the repository root directory to correctly work!
 # NOTE: `set -e` is applied conditionally only if needed.
 
+# check if a command is available to run in the system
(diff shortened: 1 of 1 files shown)


blueskyson (approved):

MasterOdin (approved):

MasterOdin:
Fine having these be ignored if not installed. Given that these are still run on CI, worst case is someone checks in offending code, but the CI blocks it from being merged.

PR #6541, opened 2021-09-16
chage: fix examples and use long arguments

navarroaxel:

### DIFF
--- a/pages.pt_BR/linux/chage.md
+++ b/pages.pt_BR/linux/chage.md
@@ -5,20 +5,20 @@
 
 - Exibir as informações referentes a senha do usuário:
 
-`chage -l {{nome_do_usuario}}`
+`chage --list {{nome_do_usuario}}`
 
 - Habilitar a expiração da senha do usuário em 10 dias:
 
-`sudo chage -M {{10}} {{nome_do_usuario}}`
+`sudo chage --maxdays {{10}} {{nome_do_usuario}}`
 
 - Desabilitar a expiração da senha do usuário:
 
-`sudo chage -M -1 {{nome_do_usuario}}`
+`sudo chage --maxdays {{-1}} {{nome_do_usuario}}`
 
 - Definir a data de expiração da conta do usuário:
 
-`sudo chage -E {{YYYY-MM-DD}}`
+`sudo chage --expiredate {{YYYY-MM-DD}} {{nome_do_usuario}}`
 
 - Obrigar o usuário a alterar sua senha no próximo login:
 
-`sudo chage -d 0`
+`sudo chage --lastday {{0}} {{nome_do_usuario}}`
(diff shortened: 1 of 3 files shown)


SethFalco (approved):

pixelcmtd (approved):

pages.pt_BR/linux/chage.md
```diff
@@ -5,20 +5,20 @@
...
 
 - Habilitar a expiração da senha do usuário em 10 dias:
 
-`sudo chage -M {{10}} {{nome_do_usuario}}`
+`sudo chage --maxdays {{10}} {{nome_do_usuario}}`
 
 - Desabilitar a expiração da senha do usuário:
 
-`sudo chage -M -1 {{nome_do_usuario}}`
+`sudo chage --maxdays {{-1}} {{nome_do_usuario}}`
```
sbrl, line 16: The `-1` is not something that the user can change without changing the meaning of the example, so it shouldn't be in `{{token_syntax}}`.

sbrl (changes requested):

pages.pt_BR/linux/chage.md
SethFalco: This has come up a couple of the times in the repo historically.
Whenever I've seen this commented/discussed, the conclusion has always been arguments should be wrapped in `{{}}`.

> > Sorry about that, I didn't realize we still had to use `{{...}}` if the command literally specifies an argument value.
> 
> No problem. Basically, everything which isn't a flag, argument or subcommand should be in `{{ }}`. Also it makes the page more readable. For comparison:
> image
> image
> 
> — https://github.com/tldr-pages/tldr/pull/5921#issuecomment-836462426

marchersimon: I think we should use token syntax. Token syntax doesn't automatically mean that we value can't be changed. It should just make it easier to differentiate between flags and arguments (In my opinion). In the example above with `anbox` it's the same.

sbrl: Hmmm, interesting. I've always viewed it as something the user can change.

navarroaxel: For me it's always to highlight the argument options like enums, or this int values.

marchersimon: In a case like this I often like to see which one looks more intuitive, but I'm actually not really sure. Maybe this would be worth opening a new issue to discuss what the purpose of the token syntax should be.
image
image

SethFalco: To be honest, I think in this case it's _much_ better to use the token syntax for `-1`.

If we didn't, it'd look to me that:

| Argument | Value
|---|---
|`--maxdays` | N/A
| `-1` | `username`

Without highlighting it, it looks like an argument, not the value to `--maxdays`.

sbrl: Hmm, I do see that there's a  beenfit to the token syntax here.

marchersimon (approved):
👍🏽

Issue #6500, opened 2021-09-09
Period char at the end of links


fgionghi:
I'm on linux and for every command I search I get a link, but at the end of the link there is a period and with CTRL-click it goes to a page that doesn't exist (because of the period).
Can't we just remove the period from all the markdown files?

My workaroud using awk is:
`tldr --theme ocean --random-example | awk '{if ($0 ~ /More information/) {gsub(/.$/,""); print} else {print}}'`
and it does the job but I'm loosing the color theme (and not understanding why, maybe when the output is not redirected to stout it does not use colors?). 

Thank you.

CleanMachine1:
I personally agree however its our syntax. Maybe a space between the link> . So it can be copied easier

CleanMachine1:
You might find it better to just remove all the periods from your list of pages since most tldr clients are just markdown renders + some extras

navarroaxel:
@fgionghi I don't have the `.` issue with `Ctrl + click`, maybe your client has a bug, because the link in closed between `<>` to avoid this situation.

Which client and version you are using? I use the `tldr-python-client` tag `2.0.0`.

sbrl:
The space at  the end of the more information link is required as part of out linter syntax. However, it does sound like the client you are using has a bug, as the angle brackets `<>` are markdown's autolinking syntax - which clearly defines which part is the URL and which part is not - so the ctrl+click should work as expected.

Which client are you using? Said client probably doesn't correctly implement [URLs in terminal emulators](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda).

fgionghi:
>Which client are you using? Said client probably doesn't correctly implement Hyperlinks (a.k.a. HTML-like anchors) in terminal emulators.`

Understood, I'm using tilix (on ubuntu 20, using bash-pages), which seems to support the feature. Don't know what to try, in the settings of tilix I'm not finding anything about that.

sbrl:
I see. I don't get any useful search results when I search for that, do you have a link to your client? I recommend opening an issue against it.

marchersimon:
The issue isn't the client. I know this because in my own client I didn't implement anything to support links and it still works. So the terminal emulator takes care of this. @fgionghi which terminal emulator are you using?

marchersimon:
[...]

PR #6543, opened 2021-09-17
tldr: fix command description in Japanese translation

k1832:
Removed the unnecessary period because periods are not usually used in Japanese sentences.
In addition to this, the period, which is removed in this PR, is confusing when hyperlink doesn't work in the cli as it might be copied with the URL.

This might be a translation-related, so don't hesitate to close this PR if this should not be in PRs.


navarroaxel:
:thinking: your suggestion looks related to <https://github.com/tldr-pages/tldr/issues/6500>. Which client and terminal are you using?

k1832:
I'm on macOS Big Sur (11.4) and using iTerm2.
But this is not about the hyperlink issue, but also about that period in the Japanese sentence being unnatural to me.

k1832:
Maybe the title of PR is not appropriate...

marchersimon:
This link was probably inserted with our [set-more-info-link.py](https://github.com/tldr-pages/tldr/blob/main/scripts/set-more-info-link.py) script. We didn't know about that rule in Japanese. But I'm sure it shouldn't be hard to fix it.

marchersimon:
I found that a lot of Japanese pages use a full-width period at the end of the line (e.g. `pwd` or `sed`). Is this also wrong?

sbrl:
Hmmm, there is a linter issue here because the linter currently requires a full stop at the end of description lines. To resolve this, I would recommend we make the linter language-aware. By doing so, it would enable / disable various linting rules depending on the language selected. I am aware however that this would be a significant undertaking to implement such a system.

marchersimon:
@sbrl I thought the linter only lints English pages? 

@shu-pf what do you think about this? You've contributed some Japanese pages and used a `。` as period.

k1832:
>@sbrl I thought the linter only lints English pages?

I think this is correct.
When I run `npm run lint-tldr-pages`, I got no error even after I removed the period from the Japanese page.
(I also made sure I got an error when I run it after I removed the period on the English page.)

marchersimon:
@k1832 could you provide us a source that in official Japanese there is indeed no period at the end of a sentence? Because a lot of other Japanese contributors used the full-width period (`。`).

sbrl:
@marchersimon that's correct, but it would be nice if it linted other languages to catch issues eventually as well :-)

k1832:
@marchersimon 
Sorry for the confusion.

（First of all, I don't think this period is needed (even on the English page) because this looks like an equation.　`詳しくはこちら: {URL} -> 詳しくはこちら = {URL}`
But I believe that this is not the discussion we want to do here. So I just ignore it and try to explain why I wanted to remove the period in the first place.）

1. Periods in Japanese sentences are unnatural (to me)
In Japanese, periods (Now I'm talking about (half-width) periods, but not full-width periods. sorry for this confusion) can be used in some contexts, which include academic papers and maybe researchers' blogs? , but officially full-width periods are used.
2. A full-width period after a URL is still unnatural (to me)
A full-width period after the English word is weird to me.
If the sentence should end with a full-width period,  I would put it like this.
`詳しくはこちら（<https://tldr.sh>）をご覧ください。`

Sorry for not showing any source with this. Opinions for this PR are always welcome.
I would totally understand if this pull request is closed because of the lack of information.

marchersimon:
Ok, so we should remove all periods after the More information links. But what about regular sentences (like the command description at the beginning)? Some pages have periods and some don't.

k1832:
@marchersimon 
Thank you for your reply!
I believe that regular sentences (like [this one](https://github.com/tldr-pages/tldr/blob/e13e67b1711e4112cca0cc4d07521c0cf901290c/pages/windows/cd.md#L3) right?) should end with period.
Unless you see it as a header.

sanopy:
I agree with removing all periods after links. And I also think that regular sentences should end with a full-width period (`。`).

marchersimon (approved):

navarroaxel (approved):

PR #6555, opened 2021-09-18
chsh: add interactive and --list-shells examples

xeruf:

### DIFF
--- a/pages/common/chsh.md
+++ b/pages/common/chsh.md
@@ -3,13 +3,13 @@
 > Change the user's login shell.
 > More information: <https://manned.org/chsh>.
 
-- Change your own login shell interactively:
+- Change the current user's login shell interactively:
 
 `chsh`
 
 - Change the login shell of `user`:
 
-`chsh --shell {{/usr/bin/shell_binary}} {{user}}`
+`chsh --shell {{/bin/zsh}} {{user}}`
 
 - List available shells:
 
--- a/pages/common/chsh.md
+++ b/pages/common/chsh.md
@@ -7,9 +7,9 @@
 
 `chsh`
 
-- Change the login shell of `user`:
+- Change the login shell for a given user to Zsh:
 
-`chsh --shell {{/bin/zsh}} {{user}}`
+`chsh --shell {{/bin/zsh}} {{username}}`
 
 - List available shells:
 
--- a/pages/common/chsh.md
+++ b/pages/common/chsh.md
@@ -3,6 +3,14 @@
 > Change the user's login shell.
 > More information: <https://manned.org/chsh>.
 
-- Change the user's login shell:
+- Change your own login shell interactively:
 
-`chsh -s {{path/to/shell_binary}} {{username}}`
+`chsh`
+
+- Change the login shell of `user`:
+
+`chsh --shell {{/usr/bin/shell_binary}} {{user}}`
+
+- List available shells:
+
+`chsh --list-shells`


pages/common/chsh.md
```diff
@@ -3,6 +3,14 @@
 > Change the user's login shell.
 > More information: <https://manned.org/chsh>.
 
-- Change the user's login shell:
+- Change your own login shell interactively:
```
CleanMachine1, line 6:
```suggestion
- Change the current user's login shell interactively:
```

pages/common/chsh.md
```diff
@@ -3,6 +3,14 @@
...
 
-`chsh -s {{path/to/shell_binary}} {{username}}`
+`chsh`
+
+- Change the login shell of `user`:
+
+`chsh --shell {{/usr/bin/shell_binary}} {{user}}`
```
CleanMachine1, line 12:
```suggestion
`chsh --shell {{/usr/bin/shell}} {{user}}`
```

navarroaxel: Maybe a specific usage is better here like:
```suggestion
`chsh --shell {{/bin/zsh}} {{user}}`
```
or `chsh --shell $(which zsh)`

CleanMachine1: I think it should be /usr/bin/zsh if we choice that.

pixelcmtd: Why would it be?! Who has their shell in `/usr/bin`??

marchersimon: Why assume that the shell is in `/usr/bin`? It is probably, but maybe not always.
```suggestion
`chsh --shell {{/path/to/shell}} {{username}}`
```

pixelcmtd: It isn't on Darwin:

```
$ ls /usr/bin/zsh /usr/bin/bash /usr/bin/sh
ls: /usr/bin/bash: No such file or directory
ls: /usr/bin/sh: No such file or directory
ls: /usr/bin/zsh: No such file or directory
```

And probably on many Linux distros, too.

`/bin` is the standard directory for the shell, but I actually prefer

```suggestion
`chsh --shell $(which {{zsh}}) {{user}}`
```

CleanMachine1:
```
whereis zsh
zsh: /usr/bin/zsh /usr/lib/zsh /etc/zsh /usr/share/zsh /usr/share/man/man1/zsh.1.gz
```

Same applies for bash, zsh being installed from repos,

this is Arch

navarroaxel: Checking `chsh --list-shells` the path is `/bin/zsh`, also a lot of help posts in SO recommends `/bin/zsh`. Maybe it's the same result for that command.

CleanMachine1: I think marchersimons suggestion is best, generic parameters are often the best. The diff between /usr/bin/zsh and /bin/zsh seems to be nothing major regarding permissions since both are equal

CleanMachine1: And the same SHA hash

xeruf: `/usr/bin` tends to contain all custom shells, while `/bin` is reserved for system stuff:
```sh
❯ chsh --list-shells
/bin/sh
/bin/bash
/bin/zsh
/usr/bin/zsh
/usr/bin/git-shell
```

I would like to provide users a pointer where to find the shell if they are inexperienced, I would also be fine with `/bin` though

pixelcmtd (approved):

marchersimon (approved):
Thanks @xeruf

pages/common/chsh.md
```diff
@@ -3,6 +3,14 @@
...
 
-- Change the user's login shell:
+- Change the current user's login shell interactively:
 
-`chsh -s {{path/to/shell_binary}} {{username}}`
+`chsh`
+
+- Change the login shell of `user`:
+
+`chsh --shell {{/bin/zsh}} {{user}}`
```
navarroaxel, line 12: I think this is the correct example description
```suggestion
- Change the login shell to Zsh of a specific user:

`chsh --shell {{/bin/zsh}} {{username}}`
```

navarroaxel (changes requested):

pages/common/chsh.md
CleanMachine1: it should be `for` instead of `of`

sbrl:
```suggestion
- Change the login shell for a given user:

`chsh --shell {{/bin/zsh}} {{user}}`
```

marchersimon:
Merging, since we have 3 approvals and all comments are addressed.

PR #6557, opened 2021-09-18
raw: add page

CleanMachine1:

### DIFF
--- a/pages/linux/raw.md
+++ b/pages/linux/raw.md
@@ -5,4 +5,4 @@
 
 - Bind a raw character device to a block device:
 
-`raw /dev/raw/raw{{1}} /dev/{{block_device`
+`raw /dev/raw/raw{{1}} {{/dev/block_device}}`
--- a/pages/linux/raw.md
+++ b/pages/linux/raw.md
@@ -1,6 +1,6 @@
 # raw
 
-> Bind a raw character device.
+> Bind a Unix raw character device.
 > More information: <https://manned.org/raw.8>.
 
 - Bind a raw character device to a block device:
--- a/pages/linux/raw.md
+++ b/pages/linux/raw.md
@@ -6,4 +6,3 @@
 - Bind a raw character device to a block device:
 
 `raw /dev/raw/raw{{1}} /dev/{{block_device`
-
--- a/pages/linux/raw.md
+++ b/pages/linux/raw.md
@@ -1,0 +1,9 @@
+# raw
+
+> Bind a raw character device.
+> More information <https://manned.org/raw.8>.
+
+- Bind a raw character device to a block device:
+
+`raw /dev/raw/raw{{1}} /dev/{{block_device`
+
--- a/pages/linux/raw.md
+++ b/pages/linux/raw.md
@@ -1,7 +1,7 @@
 # raw
 
 > Bind a raw character device.
-> More information <https://manned.org/raw.8>.
+> More information: <https://manned.org/raw.8>.
 
 - Bind a raw character device to a block device:
 
--- a/pages/linux/raw.md
+++ b/pages/linux/raw.md
@@ -6,3 +6,11 @@
 - Bind a raw character device to a block device:
 
 `raw /dev/raw/raw{{1}} {{/dev/block_device}}`
+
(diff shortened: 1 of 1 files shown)


pages/linux/raw.md
```diff
@@ -0,0 +1,8 @@
+# raw
+
+> Bind a raw character device.
+> More information: <https://manned.org/raw.8>.
+
+- Bind a raw character device to a block device:
+
+`raw /dev/raw/raw{{1}} /dev/{{block_device`
```
navarroaxel, line 8:
```suggestion
`raw /dev/raw/raw{{1}} {{/dev/block_device}}`
```

CleanMachine1: Not to sure about this, other than the mistake made.

marchersimon: I don't think there are block devices outside of `/dev/`.

pages/linux/raw.md
```diff
@@ -0,0 +1,8 @@
+# raw
+
+> Bind a raw character device.
```
marchersimon, line 3: What exactly is a raw character device?

pages/linux/raw.md
pixelcmtd: I'm pretty sure that you can have block devices in other places. Even if not, you can symlink.

pages/linux/raw.md
pixelcmtd: https://tldp.org/HOWTO/SCSI-2.4-HOWTO/rawdev.html

RTFM.

pages/linux/raw.md
CleanMachine1: Thats almost like saying I can store all of my files in /boot

I don't think its a good idea to encourage it at all.

pages/linux/raw.md
CleanMachine1: No need for that.

pages/linux/raw.md
marchersimon: But when we don't put `/dev/` in token syntax we're indicating that it's required by the tool.

sbrl: > Thats almost like saying I can store all of my files in /boot

Nothing wrong with storing all your documents in `/boot`!

> But when we don't put /dev/ in token syntax we're indicating that it's required by the tool.

Absolutely. In this instance `{{/dev/block_device}}` is most appropriate as this is a specific example (i.e. in place of something like `{{block_device}}`.

pages/linux/raw.md
sbrl: I've now read that @pixelcmtd and I still don't understand.

pixelcmtd: $ tldr raw device

**raw device**

A way of performing raw (i.e. unbuffered) I/O on a block device.

- I don't need no examples.

nicokosi: What about being a little more specific? 
```suggestion
> Bind a Unix raw character device.
```

CleanMachine1: I see no harm in doing so.

pixelcmtd (changes requested):
Why did you not include `-q` and `-a`? (that also requires a description change obviously)

einverne (approved):
This page looks good to me. But I am curious where is this command used?

sbrl (approved):

pixelcmtd:
I still think that `-q` and `-a` should have examples.

pages/linux/raw.md
```diff
@@ -0,0 +1,8 @@
+# raw
+
+> Bind a Unix raw character device.
+> More information: <https://manned.org/raw.8>.
+
+- Bind a raw character device to a block device:
+
+`raw /dev/raw/raw{{1}} {{/dev/block_device}}`
```
marchersimon, line 8:
```suggestion
`raw /dev/raw/raw{{1}} {{/dev/block_device}}`

- Query an existing binding instead of setting a new one:

`raw /dev/raw/raw{{1}}`

- Query all bound raw devices:

`raw -qa`
```

pixelcmtd (approved):

PR #6561, opened 2021-09-19
tlmgr-platform, tlmgr-arch: add page

marchersimon:


pages/common/tlmgr-platform.md
```diff
@@ -0,0 +1,24 @@
...
+> Manage TeX Live platforms.
+> More information: <https://www.tug.org/texlive/tlmgr.html>.
+
+- List all available architectures in the package repository:
+
+`tlmgr platform list`
+
+- Add the executables for a specific architecture:
+
+`sudo tlmgr platform add {{platform}}`
```
navarroaxel, line 12: I feel that this sounds better
```suggestion
`sudo tlmgr platform add {{platform_name}}`
```

marchersimon: Not sure this is necessary. Personally I'm not a fan of it.

pages/common/tlmgr-platform.md
```diff
@@ -0,0 +1,24 @@
+# tldr platform
+
+> Manage TeX Live platforms.
+> More information: <https://www.tug.org/texlive/tlmgr.html>.
+
+- List all available architectures in the package repository:
```
pixelcmtd, line 6: Shouldn't it be `platform(s)` instead of `architecture(s)` everywhere?

marchersimon: It's basically a synonym here. On one side, it's maybe a bit confusing, since the command is called `platform`, but I think architecture is just more fitting here.

pixelcmtd: I think, it is **really** confusing to first `list [...] architectures` and then `switch to [...] platform`, so this **has** to be normalized. And I think that it has to be `platform`, because when you say `architecture`, you mean `platform`. They are not synonyms. An architecture is something like `amd64`, `aarch64`, `riscv64`, ... A platform is something like `x86_64-linux`, `universal-darwin`, ... and that is what you are talking about.

navarroaxel (approved):

pages/common/tlmgr-platform.md
```diff
@@ -0,0 +1,24 @@
...
+
+`sudo tlmgr platform add {{platform}}`
+
+- Remove the executables for a specific platform:
+
+`sudo tlmgr platform remove {{platform}}`
+
+- Auto-detect and switch to the current platform:
+
+`sudo tlmgr platform set {{auto}}`
```
blueskyson, line 20: We can just remove the braces if there is no option other than `auto`. 
```suggestion
`sudo tlmgr platform set auto`
```

marchersimon: But there is. Look at the next example.

blueskyson (approved):

PR #6571, opened 2021-09-20
PULL_REQUEST_TEMPLATE: ask for command version number

marchersimon:
When we get a new page, we don't know the version of the command being documented. This can lead to issues when some examples don't work anymore and we don't know why. I think it would help a lot if we can go back to that PR and look it up.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -10,3 +10,5 @@
 - [ ] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message).
 - [ ] The page follows the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).
 - [ ] The page description includes a link to documentation or a homepage (if applicable).
+
+Version of the command being documented:
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,4 +11,4 @@
 - [ ] The page follows the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).
 - [ ] The page description includes a link to documentation or a homepage (if applicable).
 
-Version of the command being documented:
+ - **Version of the command being documented (if known):**


marchersimon:
I'm not yet really happy with the implementation of my idea, so I'm open for suggestions.

CleanMachine1:
I like this suggestion, maybe a different wording such as 

'Please include the version number of the command you have documented (if applicable) here '

SethFalco:
What about making an optional attribute of each command instead? Pages can have something like [`@since`](https://docs.oracle.com/en/java/javase/17/docs/specs/javadoc/doc-comment-spec.html#since) and [`@deprecated`](https://docs.oracle.com/en/java/javase/17/docs/specs/javadoc/doc-comment-spec.html#deprecated) from [Javadoc](https://docs.oracle.com/en/java/javase/17/javadoc/javadoc.html) that clients can opt to use?

It could be displayed statically or modify the page based on the installed version, though the latter would probably be a lot of work.

```html
- Switch to a specified branch, creating it if neccasary (Since 2.23):

`git switch {{new_branch}}`
```

If it's something that shouldn't be displayed by default, it could just be commented above. It won't be rendered, but it allows a place to keep metadata for commands that clients can choose to display. This way they can check the installed version (or if it's installed at all), or put it behind an argument.

```html
<!-- @since 2.23 -->
- Switch to a specified branch, creating it if neccasary:

`git switch {{new_branch}}`
```

I'm not sure how most clients are implemented, but if they parse Markdown, then it should be fine. If they parse the string parse themselves, then it may break.

marchersimon:
That would probably break a lot of clients and changes to the page itself can be rather problematic due to compatibility. I think it would be good enough if we just could look up the version number in the PR.

navarroaxel (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -10,3 +10,5 @@
 - [ ] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message).
 - [ ] The page follows the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).
 - [ ] The page description includes a link to documentation or a homepage (if applicable).
+
+Version of the command being documented:
```
marchersimon, line 14:
```suggestion
<!-- Please include the version number of the command you have documented (if applicable) here  -->
```
@CleanMachine1 do you mean like this?

sbrl:
```suggestion
Version of the command being documented (if known):
```

marchersimon: Formatted like this I prefer this version, because it's shorter.

CleanMachine1 (approved):

nicokosi (approved):

sbrl (approved):
Thanks, @marchersimon!

PR #6576, opened 2021-09-21
tlmgr-option: add page

marchersimon:


pages/common/tlmgr-option.md
```diff
@@ -0,0 +1,28 @@
...
+
+- List all TeX Live settings:
+
+`tlmgr option showall`
+
+- List all currently set Tex Live settings:
+
+`tlmgr option show`
+
+- List all TeX Live settings as JSON array:
```
navarroaxel, line 14: I prefer `display` or `print` if it's not a list.
```suggestion
- Display all TeX Live settings as JSON array:
```

marchersimon: Makes sense. I prefer `print`tough

navarroaxel (approved):

pages/common/tlmgr-option.md
```diff
@@ -0,0 +1,28 @@
+# tlmgr option
+
+> TeX Live settings manager.
+> More information: <https://www.tug.org/texlive/tlmgr.html>.
+
+- List all TeX Live settings:
```
pixelcmtd, line 6:
```suggestion
- List all available TeX Live settings:
```

pages/common/tlmgr-option.md
```diff
@@ -0,0 +1,28 @@
...
+
+`tlmgr option showall`
+
+- List all currently set Tex Live settings:
+
+`tlmgr option show`
+
+- Print all TeX Live settings as JSON array:
```
pixelcmtd, line 14:
```suggestion
- Print all available TeX Live settings as JSON array:
```

pixelcmtd (commented):
Also: Am I the only one who finds it weird to use the word `setting` everywhere when the command is called `option`?

pages/common/tlmgr-option.md
CleanMachine1: I suppose option would be better, however its like describing the color red as the color red, it doesn't add anything + settings is more common

pages/common/tlmgr-option.md
CleanMachine1: Shouldn't it also be `as **a** JSON array`

pixelcmtd: Maybe. `as JSON` would be the best imo, but if you want to keep the explicit mention of the `array`, I agree.

pages/common/tlmgr-option.md
pixelcmtd: Well, setting doesn't add anything either, so on that basis it doesn't matter either way.

pages/common/tlmgr-option.md
marchersimon: But it's not only `available` ones, but all existing ones. And I prefer `as JSON`.

marchersimon:
> Also: Am I the only one who finds it weird to use the word `setting` everywhere when the command is called `option`?

Because when I hear TeX Live option, I have no idea what it should be, but TeX Live setting is pretty intuitive. `option` does need some context to be the same as `setting`, which we don't really have.

pages/common/tlmgr-option.md
marchersimon: Still, setting is more common and explicit

PR #6577, opened 2021-09-21
common/l*: add more information link

blueskyson:
issue #6062

- Apple doesn't provide pages of `lp`, `lpstat`.
- Repositories of `last`, `lsof`, `lex` don't provide human readable man pages.
- I didn't find an official source of `lorem`.

### DIFF
--- a/pages/common/leave.md
+++ b/pages/common/leave.md
@@ -2,6 +2,7 @@
 
 > Set a reminder for when it's time to leave.
 > To remove reminders use `kill $(pidof leave)`.
+> More information: <https://www.freebsd.org/cgi/man.cgi?query=leave>.
 
 - Set a reminder at a given time:
 
--- a/pages/common/lex.md
+++ b/pages/common/lex.md
@@ -2,6 +2,7 @@
 
 > Lexical analyser generator.
 > Given the specification for a lexical analyser, generates C code implementing it.
+> More information: <https://manned.org/lex.1>.
 
 - Generate an analyser from a Lex file:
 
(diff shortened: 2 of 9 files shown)


pages/common/leave.md
```diff
@@ -2,6 +2,7 @@
 
 > Set a reminder for when it's time to leave.
 > To remove reminders use `kill $(pidof leave)`.
+> More information: <https://www.freebsd.org/cgi/man.cgi?query=leave>.
```
marchersimon, line 5: Why not use <https://manned.org/leave> here?

pages/common/lex.md
```diff
@@ -2,6 +2,7 @@
 
 > Lexical analyser generator.
 > Given the specification for a lexical analyser, generates C code implementing it.
+> More information: <https://manned.org/lex.1>.
```
marchersimon, line 5: `lex.1` refers to the command `flex`. Is this on purpose?

blueskyson: Yes, all `lex` in use nowadays is actually `fast lex`, also known as `flex`.

pages/common/leave.md
blueskyson: Because it's the only official source I found, and its available options does not change at all.

pixelcmtd: FreeBSD is its official source, but it is available on Linux and macOS (and therefore in `common`)? What?

marchersimon (approved):
LGTM

pages/common/leave.md
blueskyson: I couldn't find anyone maintaining this package in Linux. This command originates from BSD and freeBSD's website includes it, so I think freeBSD's man page is a better source than manned.org.

marchersimon: And `manned.org` is getting the man page from the BSD manual.

pixelcmtd (approved):

navarroaxel (approved):

PR #6586, opened 2021-09-23
tlmgr-check: add page

marchersimon:


pages/common/tlmgr-check.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Check the consistency of the whole TeX Live installation:
+
+`tlmgr check all`
+
+- Check the consistency of the whole TeX Live information in verbose mode:
+
+`tlmgr check all -v`
+
+- Check for not installed dependencies:
```
pixelcmtd, line 14:
```suggestion
- Check for missing dependencies:
```

pages/common/tlmgr-check.md
```diff
@@ -0,0 +1,28 @@
+# tlmgr check
+
+> Check the consistency of a TeX Live installation.
+> More information: <https://www.tug.org/texlive/tlmgr.html>.
+
+- Check the consistency of the whole TeX Live installation:
```
pixelcmtd, line 6:
```suggestion
- Check the consistency of the whole installation:
```

pages/common/tlmgr-check.md
```diff
@@ -0,0 +1,28 @@
...
+
+`tlmgr check all`
+
+- Check the consistency of the whole TeX Live information in verbose mode:
```
pixelcmtd, line 10:
```suggestion
- Check the consistency of the whole installation in verbose mode:
```

pages/common/tlmgr-check.md
```diff
@@ -0,0 +1,28 @@
...
+
+`tlmgr check all -v`
+
+- Check for not installed dependencies:
+
+`tlmgr check depends`
+
+- Check if all TeX Live executables are present:
```
pixelcmtd, line 18:
```suggestion
- Check if all executables are present:
```

pages/common/tlmgr-check.md
marchersimon: Why do you think it should be removed? The sentence is quite short and more explicit is always better.

pixelcmtd: > more explicit is always better

Do we have any rule that states that? Because I disagree a lot. There are many horrible ways to make this a lot more explicit about everything. The description explicitly mentions that this operates on a TeX Live installation, so this isn't just explicit, it is redundant.

marchersimon: We do have a rule which states that examples should not depend on the others and can be read on it's own. @CleanMachine1  / @navarroaxel wdyt?

navarroaxel: In this case I think mention `TeX Live` is better.

sbrl: I think keeping this is best, because it's more explicit. Otherwise, one would question which executables it checks.

CleanMachine1: Its fine to me. Same opinion as others thoughts.

pages/common/tlmgr-check.md
CleanMachine1: Explicit here is definitely better.

navarroaxel (approved):

pages/common/tlmgr-check.md
```diff
@@ -0,0 +1,28 @@
...
+
+`tlmgr check executes`
+
+- Check if all files listed in the local TLPDB are present:
+
+`tlmgr check files`
+
+- Check for duplicate filenames in the runfiles sections:
```
CleanMachine1, line 26:
```suggestion
- Check for duplicate filenames in the runfiles section:
```

CleanMachine1: Does this need to be plural?

If not then merge (I've approved)

CleanMachine1 (approved):

pages/common/tlmgr-check.md
marchersimon: That's what the man page says, so I'll just rely on that

PR #6600, opened 2021-09-25
megadl: add page

newtant:
- [X] The page (if new), does not already exist in the repository.


pages/linux/megadl.md
```diff
@@ -0,0 +1,17 @@
...
+
+`megadl {{https://mega.nz/...}}`
+
+- Interactively choose which files to download:
+
+`megadl --choose-files {{https://mega.nz/...}}`
+
+- Additionally limit speed to {{speed}} in kb/s and specify a path:
+
+`megadl --choose-files --limit-speed {{speed}} --path {{/path/to/download/}} {{https://mega.nz/...}}`
```
blueskyson, line 17:
```suggestion
- Optionally limit speed in kb/s and specify a path:

`megadl --choose-files --limit-speed {{speed}} --path {{/path/to/file_or_directory}} {{https://mega.nz/...}}`
```

pages/linux/megadl.md
```diff
@@ -0,0 +1,17 @@
+# megadl
+
+> Download files from mega.nz.
+> Part of the megatools suite.
+> More information: <https://megatools.megous.com/>.
```
blueskyson, line 5: I doesn't see any exact word of `megadl` in this link and its children links. Is there any specific website for this command?

blueskyson (commented):
Thanks for your contribution. Here are some suggestions.

pages/linux/megadl.md
newtant: The official command is `megatools-dl` but, at least on my system, is automatically symlinked as `megadl`. `megadl` is (as far as I know) the command that's actually being used.

blueskyson: Okay, I understand. I think this page should be rename to `megatools-dl.md` if which is the official name, and adding an alias page named `megadl.md` like [this](https://github.com/tldr-pages/tldr/blob/main/pages/common/vi.md).

blueskyson: Also, this website is more suitable https://megatools.megous.com/man/megatools-dl.html for looking up usage .

owenvoke: I think this sounds correct @blueskyson 👍🏻 We should try and use the official name, and if it is always symlinked, add an alias tldr page.

navarroaxel: @newtant Let us know if you need help to create this alias page.

PR #6495, opened 2021-09-08
xrdb: add page

DaxGoon:
- [x] The page (if new), does not already exist in the repository.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

PR #6606, opened 2021-09-28
xrdb: move to linux directory

SethFalco:

This file shouldn't be in the root of the repository, but rather somewhere in `pages/`, I'm assuming `pages/common/`.

### Related
* https://github.com/tldr-pages/tldr/pull/6495


blueskyson (approved):

navarroaxel (changes requested):
This is for X.org, it should be in Linux platform.

SethFalco:
My bad, I assumed it would go under `common` since it says:
> for Unix-like systems

navarroaxel (approved):
Please don't `git push --force` :laughing:

SethFalco:
> Please don't `git push --force` laughing

It's only really a problem to force pushes to significant branches like `main`.
Force pushing to feature branches is fine and makes sense if it's just to fix a mistake in the previous commit.

navarroaxel:
If you always force, we lose track about suggestions and reviews and it's harder to follow the review history. We always squash anyway.

PR #6608, opened 2021-09-28
vlc, ffmpeg, cp: add pt_BR translation

Felipe-noob:

I have 2 questions. 
The cp description is both on /pages/common and /pages/Linux, where should put it? 
The VLC description has a broken YouTube URL and I have no idea what the original video was. Should we change it to some random video, maybe something made by an official FOSS non-profit developer? I think linking to a profit-driven channel would be running free ads for them on TLDR and a broken URL might make the user think VLC isn't working.

### DIFF
--- a/pages.pt_BR/common/ffmpeg.md
+++ b/pages.pt_BR/common/ffmpeg.md
@@ -1,0 +1,36 @@
+# ffmpeg
+
+> Ferramenta de conversão de vídeo.
+> Mais informações: <https://ffmpeg.org>.
+
+- Extrair o som de um vídeo e salvá-lo como MP3:
+
+`ffmpeg -i {{vídeo.mp4}} -vn {{som}}.mp3`
+
+- Converter quadros de um vídeo ou GIF para imagens numeradas individuais:
+
+`ffmpeg -i {{vídeo.mpg|vídeo.gif}} {{quadro_%d.png}}`
+
+- Combinar imagens numeradas (`quadro_1.jpg`, `quadro_2.jpg`, etc) em um vídeo ou GIF:
+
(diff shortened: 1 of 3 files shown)


blueskyson (commented):
@sbrl @marchersimon I found some extensions are outside of token like `{{som}}.mp3` and `{{vídeo_entrada}}.mp4`, which are applied from #1477. But according to style-guide, we should use `{{*.ext}}` if an extension is needed. Should these tokens be changed back?

marchersimon:
> The cp description is both on /pages/common and /pages/Linux, where should put it?

Those are two different pages. Both should be translated and put in the right directory

>  I think linking to a profit-driven channel would be running free ads for them on TLDR and a broken URL might make the user think VLC isn't working.

Sure, sounds good.

Felipe-noob:
As for the video on VLC, I have 3 suggestions:
<https://www.youtube.com/watch?v=cWmURg_rM2o>, from Libreoffice
<https://www.youtube.com/watch?v=HdirtBXW8bI>, from KDE
<https://www.youtube.com/watch?v=V5jP1ym7nhM>, from GNOME
I couldn't find an official VLC channel.

Ok, the two cp pages are slightly different. After the Linux version gets approved, I'll submit the Common version in another PR.

sbrl:
It depends @blueskyson on whether the program requires the file extension to function or not.

> The VLC description has a broken YouTube URL and I have no idea what the original video was. Should we change it to some random video, maybe something made by an official FOSS non-profit developer? I think linking to a profit-driven channel would be running free ads for them on TLDR and a broken URL might make the user think VLC isn't working.

Good idea. I heard that the official channel for Blender 3D doesn't run ads (I think?), so could be an option.

PR #6629, opened 2021-10-01
nodemon: make --inspect a generic example

schwarmco:

If you give `--inspect` as an argument to nodemon, it passes it through to nodejs, which then listens for a Debugger. It is not meant as "inspect the given file for changes".

### DIFF
--- a/pages/common/nodemon.md
+++ b/pages/common/nodemon.md
@@ -5,7 +5,7 @@
 
 - Execute the specified file and watch a specific file for changes:
 
-`nodemon --inspect {{path/to/file.js}}`
+`nodemon {{path/to/file.js}}`
 
 - Manually restart nodemon (note nodemon must already be active for this to work):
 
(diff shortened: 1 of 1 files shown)


navarroaxel:
Hi @schwarmco, welcome and thanks for your submission. I think you can add a new example specific for `--inspect`. What do you think?

schwarmco:
Hi @navarroaxel, thank you for your feedback on that. As `--inspect` is only an example of what argument could be passed down to node here, what do you think of the example i've added?

pages/common/nodemon.md
```diff
@@ -19,6 +19,10 @@
 
 `nodemon {{path/to/file.js}} {{arguments}}`
 
+- Pass arguments to node itself if they're not nodemon arguments already (e.g. --inspect):
```
marchersimon, line 22:
```suggestion
- Pass arguments to node itself if they're not nodemon arguments already (e.g. `--inspect`):
```

marchersimon (approved):
Looks good, I've just left a small suggestion

navarroaxel (approved):

PR #6630, opened 2021-10-01
curl: add --fail in example

janisz:

### DIFF
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -12,9 +12,9 @@
 
 `curl --remote-name {{http://example.com/filename}}`
 
-- Download a file, following location redirects, and automatically continuing (resuming) a previous file transfer:
+- Download a file, following location redirects, and automatically continuing (resuming) a previous file transfer and return error on server error:
 
-`curl --remote-name --location --continue-at - {{http://example.com/filename}}`
+`curl --fail-with-body --remote-name --location --continue-at - {{http://example.com/filename}}`
 
 - Send form-encoded data (POST request of type `application/x-www-form-urlencoded`). Use `--data @file_name` or `--data @'-'` to read from STDIN:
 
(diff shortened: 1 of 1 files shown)


navarroaxel:
Hi @janisz, welcome and thanks for your submission. The pages can have up to 8 examples. Do you thinks we should replace or edit a current example with this new arguments?

janisz:
@navarroaxel Fixed, I edited current line

janisz:
https://daniel.haxx.se/blog/2021/02/11/curl-fail-with-body/

sbrl (commented):
This makes this example rather long.  Furthermore, I can't find `--fail-with-body` in the `curl` man page?

pixelcmtd:
> I can't find `--fail-with-body` in the `curl` man page?

@sbrl Then you're probably running an out-of-date distro.

```
       --fail-with-body
              (HTTP) Return an error on server errors where the HTTP response code is 400 or greater). In normal cases when
              an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also describes
              why and more). This flag will still allow curl to output and save that content but also to return error 22.

              This is an alternative option to -f, --fail which makes curl fail for the same circumstances but without saving
              the content.

              See also -f, --fail. Added in 7.76.0.
```

marchersimon:
I think we could add it to the second example, which is still rather short.

janisz:
Maybe a reason why I think it's important.

Recently I worked withers script that was using curl. Unfortunately when you don't pass -f option curl returns 0 even if server returned error. That takes me some time to debug since it I must check all steps of my script to detect where it fails.

marchersimon:
Still, we should add it to an example, which isn't too long yet. Or we could remove another one if it's not that important.

sbrl:
I'm using Ubuntu 21.04, @janisz

marchersimon:
@janisz why did you close this? I think documenting the flag makes sense, we just have to find a way to fit in into the already full page.

janisz:
Until recent curl versions go into mainstream repositories adding --fail-with-body will only confuse people

pages/common/curl.md
```diff
@@ -12,9 +12,9 @@
 
 `curl --remote-name {{http://example.com/filename}}`
 
-- Download a file, following location redirects, and automatically continuing (resuming) a previous file transfer:
+- Download a file, following location redirects, and automatically continuing (resuming) a previous file transfer and return error on server error:
```
marchersimon, line 15:
```suggestion
- Download a file, following location redirects, and automatically continuing (resuming) a previous file transfer and return an error on server error:
```

marchersimon (approved):
Thanks @janisz

sbrl (approved):
This complicates an already full page, but in this instance I think it might be ok.

PR #6570, opened 2021-09-20
set-more-info-link.py, pages.hi/*: fix punctuation in Hindi translation

marchersimon:
I noticed that Hindi uses a `।` as period. We didn't consider this in `set-more-info-link.py`, so this PR will fix this.
Waiting for #6537

PR #6638, opened 2021-10-01
pwd: add Hindi translation

themayankjha:


themayankjha:
ah okay ill recommit with UNIX line endings

navarroaxel (approved):
Hi @themayankjha, welcome and thanks for your submission.

themayankjha:
thanks :) i recommitted with unix file endings.

blueskyson (approved):
Thanks for contribution.

marchersimon:
Hey @themayankjha, just a quick question about Hindi punctuation. 
* If Hindi uses `|`, why did you (and Hindi contributors before you) use `.` after the more info link?
* Why did you put a space before the `|` in the third line?

I'm currently trying to fix Hindi pages in #6570. Could you maybe have a quick look at it? Thanks!

themayankjha:
Hey @marchersimon !

I just looked into the translations. 

I think the reason for using a  ```.``` instead of the ``` । ``` is that a link is an English character sequence. Since the contributors are probably mostly bilingual, it was just the formatting error that the first guy followed from the English and everyone else adopted to keep the formatting same. So the English full stop after some English characters is what it boils down to.  In my opinion, using the Hindi full stop is the way to go as there is no concept of a period in the Hindi language.

The issue in my commit is an oversight on my part. If you notice ``` | ``` and ``` । ``` are two different symbols and the shorter one is the full stop so there should not have been the space and the longer symbol which I used as I think I copied from a bad website as Hindi keyboards are not widely used.

Tomorrow, I will generate another PR to fix my commit and I will also look into your specified issue.
If you would like I could change all the ```.``` to ``` । ``` in the same commit or I could leave it out for the time being.

marchersimon:
Thanks. #6570 is a PR, not an issue, where I try to fix the existing pages. But I'm not sure if I'm using the correct symbol there.

themayankjha:
oh okay so you already fixed them. Yes, you are using the correct symbol.

themayankjha:
I generated another PR to fix my commit!

PR #6650, opened 2021-10-01
go-run: add page

CleanMachine1:

### DIFF
--- a/pages/common/go-run.md
+++ b/pages/common/go-run.md
@@ -1,0 +1,9 @@
+# go run
+
+> Compile and run Go code.
+> Does not leave a binary.
+> More information <https://pkg.go.dev/cmd/go#hdr-Compile_packages_and_dependencies>.
+
+- Run a Go file:
+
+`go run {{path/to/source_code.go}}`
--- a/pages/common/go-run.md
+++ b/pages/common/go-run.md
@@ -2,7 +2,7 @@
 
 > Compile and run Go code.
 > Does not leave a binary.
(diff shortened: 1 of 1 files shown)


pages/common/go-run.md
```diff
@@ -0,0 +1,9 @@
+# go run
+
+> Compile and run Go code.
+> Does not leave a binary.
+> More information: <https://pkg.go.dev/cmd/go#hdr-Compile_packages_and_dependencies>.
+
+- Run a Go file:
+
+`go run {{path/to/source_code.go}}`
```
navarroaxel, line 9: I think file is enough.
```suggestion
`go run {{path/to/file.go}}`
```

navarroaxel (approved):

pages/common/go-run.md
```diff
@@ -0,0 +1,9 @@
+# go run
+
+> Compile and run Go code.
+> Does not leave a binary on disk.
```
pixelcmtd, line 4: The fact that this is technically very wrong **really** annoys me.

CleanMachine1: Have you got a suggestion?

marchersimon: @pixelcmtd do you mean because it still has to save the executable somewhere? In that case it's probably removed right after being used, so it does indeed leave no binary.

Also, a small suggestion:
```suggestion
> Does not leave a binary on the disk.
```

marchersimon (approved):
👍🏽

PR #6661, opened 2021-10-02
7z, 7za, 7zr: fix Japanese translation

Ry0taK:

Japanese translation of 7z were bit broken (e.g. `道/に/ファイルまたはディレクトリ` means `File or directory on the road`), so I tweaked it a bit.


blueskyson (approved):
Look good to me. Thanks for contribution.

pages.ja/common/7z.md
```diff
@@ -1,36 +1,36 @@
 # 7z
 
-> 7-Zipは高圧縮率のファイルアーカイバ
-> 詳しくはこちら: <https://www.7-zip.org>.
+> 圧縮率の高いファイルアーカイバです。
+> 詳しくはこちら: <https://www.7-zip.org>
```
navarroaxel, line 4: We need a period here
```suggestion
> 詳しくはこちら: <https://www.7-zip.org>.
```

Ry0taK: Thank you for pointing it! Is this period necessary for TLDR to parse markdown correctly?

navarroaxel: It's not required for the markdown, but it's required to keep a format without lints. We don't have a lint for ja so we should do this manually.

navarroaxel (approved):

pages.ja/common/7z.md
Ry0taK: I understand it, I've added period to Japanese translation of `7z*`

PR #6673, opened 2021-10-02
tr: add French translation

noraj:

### DIFF
--- a/pages.fr/common/tr.md
+++ b/pages.fr/common/tr.md
@@ -3,27 +3,27 @@
 > Convertisseur de caractères : exécute des remplacements basés sur des caractères uniques et des jeux de caractères.
 > Plus d'informations : <https://www.gnu.org/software/coreutils/tr>.
 
-- Remplacer toutes les occurrences d'un caractère dans un fichier, et afficher le résultat :
+- Remplace toutes les occurrences d'un caractère dans un fichier, et affiche le résultat :
 
 `tr {{caractère_recherché}} {{caractère_remplacé}} < {{fichier}}`
 
-- Remplacer toutes les occurrences d'un caractère dans la sortie d'une autre commande :
+- Remplace toutes les occurrences d'un caractère dans la sortie d'une autre commande :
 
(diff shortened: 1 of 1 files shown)


navarroaxel (approved):

pages.fr/common/tr.md
```diff
@@ -0,0 +1,32 @@
+# tr
+
+> Convertisseur de caractères : exécute des remplacements basés sur des caractères uniques et des jeux de caractères.
+> Plus d'informations : <https://www.gnu.org/software/coreutils/tr>.
+
+- Remplace toutes les occurrences d'un caractère dans un fichier, et affiche le résultat :
```
marchersimon, line 6: Shouldn't that be a non-breaking space, instead of a regular space?
```suggestion
- Remplace toutes les occurrences d'un caractère dans un fichier, et affiche le résultat :
```

noraj: True. Outside of WYSIWYG editor like MS Word nobody uses them because there is no easy way to insert them from keyboard. I didn't knew we were going into such a level of typography details. Is that possible to add this rule to https://github.com/tldr-pages/tldr/blob/main/CONTRIBUTING.md#translations if it's really required for some languages? And maybe recommend a plugin helping to insert such kind of spaces.
I don't even know how to insert them, the only working shortcut I found for my system us <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>U</kbd><kbd>0</kbd><kbd>0</kbd><kbd>A</kbd><kbd>0</kbd> which is really painful and unusable. Other shortcuts like <kbd>AltGr</kbd>+<kbd>Space</kbd> or r <kbd>Shift</kbd>+<kbd>AltGr</kbd> <kbd>Space</kbd> are not working.

marchersimon: Actually, I thought it was more important, but especially for our use case it doesn't matter at all. Just for the sake of doing it we can do mass PRs where we replace all normal spaces with non-breaking spaces every now and then.

marchersimon (approved):
Thanks @noraj

PR #6703, opened 2021-10-03
go-fmt: add page

morganhoward:

```
go version
go version go1.17.1 darwin/amd64
```


morganhoward:
saw that `go fmt` hadn't been filled out yet here: https://github.com/tldr-pages/tldr/issues/5474

pages/common/go-fmt.md
```diff
@@ -0,0 +1,24 @@
...
+
+`go fmt`
+
+- Format a specific Go package:
+
+`go fmt {{package}}`
+
+- Format the package in the current directory and all subdirectories (note the `...`):
+
+`go fmt ./...`
```
navarroaxel, line 16: This requires token syntax:
```suggestion
`go fmt {{./...}}`
```

pages/common/go-fmt.md
```diff
@@ -0,0 +1,24 @@
...
+> Format Go source files.
+> More information: <https://pkg.go.dev/cmd/go#hdr-Gofmt__reformat__package_sources>.
+
+- Format Go source files in the current directory:
+
+`go fmt`
+
+- Format a specific Go package:
+
+`go fmt {{package}}`
```
navarroaxel, line 12: is this a path?

morganhoward: good question.

according to the docs, `go fmt` takes a package as an argument.

some examples are:
* `go fmt` to format the current package
* `go fmt .` is the same as the above (`.` is the default arg)
* `go fmt ./...` is a go convention (at least I've only seen it used in go) for the current directory and subdirectories 
* `go fmt foo/bar` formats a package in `$GOPATH/src/foo/bar`

`go fmt` calls `gofmt` under the hood and `gofmt` takes a path as an argument.

So, I'd say according to the docs, the argument to `go fmt` is not a path, but it sure looks like a path.

original documentation:
```
go help fmt
usage: go fmt [-n] [-x] [packages]

Fmt runs the command 'gofmt -l -w' on the packages named
by the import paths. It prints the names of the files that are modified.

For more about gofmt, see 'go doc cmd/gofmt'.
For more about specifying packages, see 'go help packages'.

The -n flag prints commands that would be executed.
The -x flag prints commands as they are executed.

The -mod flag's value sets which module download mode
to use: readonly or vendor. See 'go help modules' for more.

To run gofmt with specific options, run gofmt itself.

See also: go fix, go vet.
```

did I answer your question?

morganhoward: related question: with equivalent cmds like `go fmt` and `go fmt .`, is there a convention for how to display those?  These are some ways I came up with; not sure if there is an established convention in this repo or not.

```
- Format Go source files in the current directory:

`go fmt` or `go fmt .`
```
or 

```
- Format Go source files in the current directory:

`go fmt`

- Format Go source files in the current directory (explicit):

`go fmt {{.}}`
```

or 

```
- Format Go source files in the current directory (default argument is .):

`go fmt`
```

pages/common/go-fmt.md
morganhoward: thanks for pointing that out.  I updated the commit with this change.

pages/common/go-fmt.md
```diff
@@ -0,0 +1,26 @@
+# go fmt
+
+> Format Go source files.
+> Prints the filenames that are changed.
+> More information: <https://pkg.go.dev/cmd/go#hdr-Gofmt__reformat__package_sources>.
+
+- Format Go source files in the current directory:
+
+`go fmt`
+`go fmt {{.}}`
```
navarroaxel, line 10: You must pick one
```suggestion
`go fmt`
```

morganhoward: updated.  went with the first command.

pages/common/go-fmt.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Format a specific Go package in your import path (`$GOPATH/src`):
+
+`go fmt {{package}}`
+
+- Format the package in the current directory and all subdirectories (note the `...`):
+
+`go fmt {{./...}}`
+
+- Print what format commands would've been run:
```
marchersimon, line 19:
```suggestion
- Print what format commands would've been run, without modifying anything:
```

PR #6723, opened 2021-10-03
highlight: add page

dmbaturin:
[highlight](http://www.andre-simon.de/doku/highlight/en/highlight.php) is a flexible syntax highlighting tool that supports a variety of formats and can be extended with Lua scripts.

**Version of the command being documented (if known):** 3.60

### DIFF
--- a/pages/common/highlight.md
+++ b/pages/common/highlight.md
@@ -1,0 +1,28 @@
+# highlight
+
+> Outputs syntax-highlighted source code to a variety of formats.
+> More information: <http://www.andre-simon.de/doku/highlight/en/highlight.php>.
+
+- Produce a complete HTML document from a source code file:
+
+`highlight -O html --style {{theme name}} --syntax={{language}} {{source file}}`
+
+- Produce an HTML fragment, suitable for inclusion in a larger document:
+
+`highlight -O html --fragment --syntax={{language}} {{source file}}`
+
+- Inline all CSS styling in every tag:
+
+`highlight -O html --inline-css --syntax={{language}} {{source file}}`
+
+- List all supported languages:
+
+`highlight --list-scripts lang`
+
+- List all installed syntax highlighting themes:
+
+`highlight --list-scripts themes`
+
+- Print a CSS stylesheet for a theme:
+
+`highlight -O html --print-style --style {{theme name}} --syntax {{language}}] --stdout`
--- a/pages/common/highlight.md
+++ b/pages/common/highlight.md
@@ -5,7 +5,7 @@
 
 - Produce a complete HTML document from a source code file:
 
-`highlight -O {{html}} --style {{theme_name}} --syntax={{language}} {{path/to/source_code}}`
+`highlight --out-format={{html}} --style {{theme_name}} --syntax={{language}} {{path/to/source_code}}`
 
 - Produce an HTML fragment, suitable for inclusion in a larger document:
 
--- a/pages/common/highlight.md
+++ b/pages/common/highlight.md
@@ -15,9 +15,9 @@
 
 `highlight -O html --inline-css --syntax={{language}} {{source file}}`
 
-- List all supported languages:
+- List all supported languages, themes, or plugins:
 
-`highlight --list-scripts lang`
+`highlight --list-scripts {{langs|themes|plugins}}`
 
 - List all installed syntax highlighting themes:
 
--- a/pages/common/highlight.md
+++ b/pages/common/highlight.md
@@ -1,17 +1,17 @@
 # highlight
 
 > Outputs syntax-highlighted source code to a variety of formats.
-> More information: <http://www.andre-simon.de/doku/highlight/en/highlight.php>.
+> More information: <http://www.andre-simon.de/doku/highlight/highlight.php>.
 
 - Produce a complete HTML document from a source code file:
 
-`highlight -O html --style {{theme name}} --syntax={{language}} {{source file}}`
+`highlight -O {{html}} --style {{theme_name}} --syntax={{language}} {{path/to/source_code}}`
 
 - Produce an HTML fragment, suitable for inclusion in a larger document:
 
 `highlight -O html --fragment --syntax={{language}} {{source file}}`
 
-- Inline all CSS styling in every tag:
+- Inline the CSS styling in every tag:
 
 `highlight -O html --inline-css --syntax={{language}} {{source file}}`
 
--- a/pages/common/highlight.md
+++ b/pages/common/highlight.md
@@ -21,7 +21,7 @@
 
 - List all installed syntax highlighting themes:
 
-`highlight --list-scripts {{themes}}`
+`highlight --list-scripts themes`
 
 - Print a CSS stylesheet for a theme:
 
--- a/pages/common/highlight.md
+++ b/pages/common/highlight.md
@@ -25,4 +25,4 @@
 
 - Print a CSS stylesheet for a theme:
 
-`highlight -O html --print-style --style {{theme name}} --syntax {{language}}] --stdout`
+`highlight --out-format={{html}} --print-style --style {{theme name}} --syntax {{language}}] --stdout`
--- a/pages/common/highlight.md
+++ b/pages/common/highlight.md
@@ -13,7 +13,7 @@
 
 - Inline the CSS styling in every tag:
 
-`highlight -O html --inline-css --syntax={{language}} {{source file}}`
+`highlight --out-format={{html}} --inline-css --syntax={{language}} {{source file}}`
 
 - List all supported languages, themes, or plugins:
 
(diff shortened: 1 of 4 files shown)


pages/common/highlight.md
```diff
@@ -0,0 +1,28 @@
+# highlight
+
+> Outputs syntax-highlighted source code to a variety of formats.
+> More information: <http://www.andre-simon.de/doku/highlight/en/highlight.php>.
+
+- Produce a complete HTML document from a source code file:
+
+`highlight -O html --style {{theme name}} --syntax={{language}} {{source file}}`
```
marchersimon, line 8:
```suggestion
`highlight -O {{html}} --style {{theme_name}} --syntax={{language}} {{path/to/source_code}}`
```

pages/common/highlight.md
```diff
@@ -0,0 +1,28 @@
+# highlight
+
+> Outputs syntax-highlighted source code to a variety of formats.
+> More information: <http://www.andre-simon.de/doku/highlight/en/highlight.php>.
```
marchersimon, line 4:
```suggestion
> More information: <http://www.andre-simon.de/doku/highlight/highlight.php>.
```

pages/common/highlight.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Produce a complete HTML document from a source code file:
+
+`highlight -O html --style {{theme name}} --syntax={{language}} {{source file}}`
+
+- Produce an HTML fragment, suitable for inclusion in a larger document:
+
+`highlight -O html --fragment --syntax={{language}} {{source file}}`
+
+- Inline all CSS styling in every tag:
```
marchersimon, line 14:
```suggestion
- Inline the CSS styling in every tag:
```
I think this sounds a bit better

pages/common/highlight.md
```diff
@@ -0,0 +1,28 @@
...
+
+`highlight -O html --inline-css --syntax={{language}} {{source file}}`
+
+- List all supported languages:
+
+`highlight --list-scripts lang`
```
marchersimon, line 20:
```suggestion
`highlight --list-scripts {{lang}}`
```

pages/common/highlight.md
```diff
@@ -0,0 +1,28 @@
...
+
+- List all installed syntax highlighting themes:
+
+`highlight --list-scripts themes`
```
marchersimon, line 24:
```suggestion
`highlight --list-scripts {{themes}}`
```

pages/common/highlight.md
dmbaturin: The command to list supported languages is *exactly* `highlight --list-scripts lang`. In this case `lang` is a *keyword*, not a variable.

pages/common/highlight.md
dmbaturin: This is also a wrong suggestion because `themes` is also a keyword here, not a variable.

pages/common/highlight.md
marchersimon: But `lang` is one of multiple arguments for `--list-scripts`. We use token syntax for this too. And it's obvious from the description, that you don't have to put in a specific language.

dmbaturin: `{{lang}}` syntax implies that the character sequence "lang" must be substituted for an actual value. However, in this case, the character sequence "lang" **is** the value.

The `--list-scripts` accepts multiple *fixed* keywords. In the BNF it could be written `--list-scripts <"lang"|"themes"|"plugins">`. I'll reiterate: "lang" (sic), "themes", and "plugins" are keywords, and they are the only keywords that the `--list-scripts` option accepts.

```
$ highlight --list-scripts lang | head

Installed langDefs (located in /usr/share/highlight/langDefs/):

ABAP/4                        : abap4 ( abp )
ABC                           : abc
Advanced Backus-Naur Form     : abnf
ActionScript                  : actionscript ( as )
ADA95                         : ada ( a adb ads gnad )
Agda                          : agda
ALAN Interactive Fiction Language: alan ( alan i )
```

marchersimon: But that's also what we use token syntax for. `{{lang}}` doesn't necessarily have to substituted with anything. Everything, which isn't a flag or subcommand goes into token syntax.

dmbaturin: Please read my previous comment carefully. The command to list all installed language syntaxes (that is, Lua scripts of the type `lang`) is **exactly** `highlight --list-scripts lang`. Writing it `--list-scripts {{lang}}` would be the same as writing `git {{checkout}}`—that would be invalid syntax.

That command is to be copied/pasted or typed unmodified. This is what will happen if the user copies `highlight --list-scripts {{lang}}` and runs it:

```
$ highlight --list-scripts {{lang}}
highlight: Unknown script type '{{lang}}'. Apply one of 'themes', 'langs' or 'plug-ins'.
```

Because the "lang" word in `--list-scripts lang` not only does not need to be substituted with anything, but **MUST NOT** be substituted with anything if your goal is to list scripts of the type "lang", using a meta-variable syntax for it is unnecessary and confusing. 

Here's another example for you: `ping` supports `-M` option whose arguments can be **exactly** `do`, `dont`, or `want`. 

```
       -M pmtudisc_opt
           Select Path MTU Discovery strategy.  pmtudisc_option may be either do (prohibit fragmentation, even local one), want (do PMTU discovery, fragment locally when packet size is large), or dont (do not set DF flag).
```

By your logic the command for sending ICMP echo requests with the DF (Don't Fragment) bit set must be written `ping -M{{do}} {{host}}`. The syntax that does make sense would be `ping -Mdo {{host}}`—the `{{host}}` part must be substituted with an actual host address, but `-Mdo` is exactly the option the user wants to disable fragmentation.

**OR** we would write `ping -M<do|dont|want> {{host}}` to show the user all available options.

dmbaturin: I changed the example `--list-scripts {{langs|themes|plugins}}` and reworded its description accordingly.

pages/common/highlight.md
```diff
@@ -0,0 +1,28 @@
...
+
+`highlight --list-scripts {{langs|themes|plugins}}`
+
+- List all installed syntax highlighting themes:
+
+`highlight --list-scripts themes`
+
+- Print a CSS stylesheet for a theme:
+
+`highlight -O html --print-style --style {{theme name}} --syntax {{language}}] --stdout`
```
navarroaxel, line 28:
```suggestion
`highlight --out-format={{html}} --print-style --style {{theme name}} --syntax {{language}}] --stdout`
```

pages/common/highlight.md
SethFalco: Something worthy of discussion is that it there is ambiguity here though, to be honest.

I do think it could make sense to consider if there is a way to denote that the parameter is intended literally, but I don't know a good way to do that.

pages/common/highlight.md
```diff
@@ -0,0 +1,28 @@
...
+> Outputs syntax-highlighted source code to a variety of formats.
+> More information: <http://www.andre-simon.de/doku/highlight/highlight.php>.
+
+- Produce a complete HTML document from a source code file:
+
+`highlight -O {{html}} --style {{theme_name}} --syntax={{language}} {{path/to/source_code}}`
+
+- Produce an HTML fragment, suitable for inclusion in a larger document:
+
+`highlight -O html --fragment --syntax={{language}} {{source file}}`
```
navarroaxel, line 12:
```suggestion
`highlight --out-format={{html}} --fragment --syntax={{language}} {{source file}}`
```

pages/common/highlight.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Inline the CSS styling in every tag:
+
+`highlight -O html --inline-css --syntax={{language}} {{source file}}`
```
navarroaxel, line 16:
```suggestion
`highlight --out-format={{html}} --inline-css --syntax={{language}} {{source file}}`
```

pages/common/highlight.md
marchersimon: * The `{{ }}` won't be rendered it's just there to hightlight that part. Like this:
image

* I don't think the `|` will get confused with a pipe, because no one writes a pipe without spaces between it.

dmbaturin: Oh, I see. Somehow I never made a connection between curly braces and the output in the client, sorry!

pages/common/highlight.md
```diff
@@ -0,0 +1,24 @@
+# highlight
+
+> Outputs syntax-highlighted source code to a variety of formats.
+> More information: <http://www.andre-simon.de/doku/highlight/highlight.php>.
+
+- Produce a complete HTML document from a source code file:
+
+`highlight --out-format={{html}} --style {{theme_name}} --syntax={{language}} {{path/to/source_code}}`
```
marchersimon, line 8: Why does `--out-format` have an `=` sign, but the other flags don't?

PR #6727, opened 2021-10-03
gpupdate: fix target syntax

koraktor:
This is a fix to the `gpupdate /target` example removing the equal sign. See the linked Microsoft documentation for the correct syntax applied in this patch.


marchersimon (approved):
Thanks @koraktor

PR #6735, opened 2021-10-03
ln: add Hindi translation

2KAbhishek:


pages.hi/common/ln.md
```diff
@@ -0,0 +1,16 @@
+# ln
+
+> फाइलों और निर्देशिकाओं के लिंक बनाता है।
+> अधिक जानकारी: <https://www.gnu.org/software/coreutils/ln>।
```
blueskyson, line 4:
```suggestion
> अधिक जानकारी: <https://www.gnu.org/software/coreutils/ln>.
```

pages.hi/common/ln.md
```diff
@@ -0,0 +1,16 @@
...
+
+- किसी फ़ाइल या निर्देशिका के लिए एक प्रतीकात्मक लिंक बनाएँ:
+
+`ln -s {{फ़ाइल या निर्देशिका / का / पथ}} {{लिंक / का / पथ}}`
```
blueskyson, line 8: I'm not sure whether to put spaces between Hindi characters and slashes. In [chsh](https://github.com/tldr-pages/tldr/blob/main/pages.hi/common/chsh.md), the author didn't use spaces, please let us know which is the correct way in Hindi. In tldr, we tend to not put spaces near slashes.

blueskyson (commented):
Thank you for contribution. Here are some suggestions.

pages.hi/common/ln.md
2KAbhishek: '।' is actually the full stop alternative in hindi, are we sure that we want to use '.' here?

pages.hi/common/ln.md
2KAbhishek: I have removed the spaces, they could be confusing, thanks for the suggestion

pages.hi/common/ln.md
marchersimon: We should keep what's correct. So it should be `!`.

pages.hi/common/ln.md
```diff
@@ -0,0 +1,16 @@
+# ln
+
+> फाइलों और निर्देशिकाओं के लिंक बनाता है।
+> अधिक जानकारी: <https://www.gnu.org/software/coreutils/ln>।
+
+- किसी फ़ाइल या निर्देशिका के लिए एक प्रतीकात्मक लिंक बनाएँ:
+
+`ln -s {{फ़ाइल_या_निर्देशिका/का/पथ}} {{लिंक/का/पथ}}`
```
marchersimon, line 8: This should be an absolute path (same for the other two examples)
```suggestion
`ln -s {{/फ़ाइल_या_निर्देशिका/का/पथ}} {{लिंक/का/पथ}}`
```

marchersimon (approved):
Thanks @2KAbhishek

navarroaxel (approved):

PR #6752, opened 2021-10-03
*: remove unnecesary trailing .1 of manned.org links

pixelcmtd:
As far as i know, we only put extra information in the link if it is needed.


navarroaxel (approved):

CleanMachine1 (approved):
:+1:

PR #6763, opened 2021-10-04
git: add more subcommand examples

Ray6464:
Added notes for using tldr for git subcommands. This note feels more instinctive to me. A CLI user should be able to see how to use tldr without visiting its github repo.
Also elongated line #15 with common subcommands for mnemonic purpose. New users often forget subcommands, and they have no way to recall them with tldr.

### DIFF
--- a/pages/common/git.md
+++ b/pages/common/git.md
@@ -1,7 +1,7 @@
 # git
 
 > Distributed version control system.
-> Some subcommands such as `git commit` have their own usage documentation.
+> Some subcommands such as `commit`, `checkout`, etc. have their own usage documentation. To see tldr for them use `tldr git {{subcommand}}`.
 > More information: <https://git-scm.com/>.
 
 - Check the Git version:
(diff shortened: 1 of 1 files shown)


pages/common/git.md
```diff
@@ -1,7 +1,7 @@
 # git
 
 > Distributed version control system.
-> Some subcommands such as `git commit` have their own usage documentation.
+> Some subcommands such as `commit`, `checkout`, etc. have their own usage documentation. To see tldr for them use `tldr git {{subcommand}}`.
```
sbrl, line 4: Interesting idea, but I feel it's more explicit to say `git commit` instead of just `commit` for example.

marchersimon: I'm not sure if we should change that part at all. I assume (and hope) that most users will know that this means you can run `tldr git commit`.

marchersimon: But token syntax in description line isn't supported right?

sbrl: Oh yeah! We haven't done that by convention, no. Let's apply your suggestion as-is then.

marchersimon (approved):

navarroaxel (approved):

PR #6766, opened 2021-10-04
cgroups: add page

Ray6464:
cgroups are used via cgcreate, cgexec, and cgclassify, and users may look for `tldr cgcreate` or something, so i'll make individual pages for them.

### DIFF
--- a/pages/linux/cgroups.md
+++ b/pages/linux/cgroups.md
@@ -1,5 +1,5 @@
 # cgroups
 
-> cgroups aka control groups is a Linux kernel feature for limiting, mesuring, and controling resource usage by processes.
+> Cgroups aka control groups is a Linux kernel feature for limiting, mesuring, and controling resource usage by processes.
 > See `tldr cgcreate`, `tldr cgexec`, `tldr cgclassify` for more usage info.
-> See more docs at: https://www.kernel.org/doc/Documentation/cgroup-v1/
+> More information: `https://www.kernel.org/doc/Documentation/cgroup-v1/`
--- a/pages/linux/cgroups.md
+++ b/pages/linux/cgroups.md
@@ -2,4 +2,4 @@
 
 > Cgroups aka control groups is a Linux kernel feature for limiting, mesuring, and controling resource usage by processes.
 > See `tldr cgcreate`, `tldr cgexec`, `tldr cgclassify` for more usage info.
-> More information: `https://www.kernel.org/doc/Documentation/cgroup-v1/`
+> More information: <https://www.kernel.org/doc/Documentation/cgroup-v1/>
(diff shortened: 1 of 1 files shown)


marchersimon:
If `cgroup` isn't a command on its own, we can't document it.

SethFalco:
I'm with @marchersimon.

Basically, this suggests we should have a new type of page (we have a regular page, and an alias page already) which behaves as an index page for related commands.

To find commands under `cgroups`, that's where the search feature of clients would probably come in handy. I don't think there needs to be a manually maintained set of index pages.

sbrl:
Hmmm, if people are trying to view it and getting confused then it may be worth making an exception here. This could maybe be an expanded alias page? A page must have at least 1 example I would think, so perhaps 1 example for each command in question would work?

Ray6464:
> Hmmm, if people are trying to view it and getting confused then it may be worth making an exception here. This could maybe be an expanded alias page? A page must have at least 1 example I would think, so perhaps 1 example for each command in question would work?

Yes, something like this should make it better.

marchersimon:
I don't think we can merge this as is, because it's just pointing to other tldr pages. We should add least add a few examples as ＠sbrl said above:

> A page must have at least 1 example I would think, so perhaps 1 example for each command in question would work?

pages/linux/cgroups.md
```diff
@@ -0,0 +1,16 @@
+# cgroups
+
+> Cgroups aka control groups is a Linux kernel feature for limiting, mesuring, and controling resource usage by processes.
+> More information: <https://www.kernel.org/doc/Documentation/cgroup-v2.txt>.
+
+- Show the tldr page for a cgclassify:
```
sbrl, line 6: Perhaps this will read better?
```suggestion
- Show the tldr page for `cgclassify`:
```

pages/linux/cgroups.md
```diff
@@ -0,0 +1,16 @@
...
+
+`tldr cgclassify`
+
+- Show the tldr page for a cgcreate:
```
sbrl, line 10:
```suggestion
- Show the tldr page for `cgcreate`:
```

pages/linux/cgroups.md
```diff
@@ -0,0 +1,16 @@
...
+
+`tldr cgcreate`
+
+- Show the tldr page for a cgexec:
```
sbrl, line 14: Perhaps this will read better?
```suggestion
- Show the tldr page for `cgexec`:
```

sbrl (changes requested):

sbrl (approved):
Thanks, @Ray6464!

navarroaxel (approved):

PR #6812, opened 2021-10-05
*: fix period in Japanese translation

marchersimon:
As discussed in #6543, following rules apply in Japanese pages:

* More information links shouldn't have a period at all
* Regular sentences should end with a full-width period

Thanks to @k1832 and @sanopy for the explanations


scripts/set-more-info-link.py
```diff
@@ -80,6 +80,8 @@ def set_link(file, link):
     # build new line
     if locale == "zh" or locale == "zh_TW":
         new_line = f"> {labels[locale]}<{link}>.\n"
+    elif locale == "ja":
+        new_line = f"> {labels[locale]} <{link}>\n"
     elif locale == "hi":
         new_line = f"> {labels[locale]} <{link}>।\n"
```
navarroaxel, line 86: Could we use alpha order?
```suggestion
    if locale == "hi":
        new_line = f"> {labels[locale]} <{link}>।\n"
    elif locale == "ja":
        new_line = f"> {labels[locale]} <{link}>\n"
    elif locale == "zh" or locale == "zh_TW":
        new_line = f"> {labels[locale]}<{link}>.\n"
```

navarroaxel (approved):

sbrl (approved):
Not having a period on more information links is going to cause a headache for whoever updates the linter to support multiple versions, but this is fine for now.

Note that if the period is becoming part of the URL, then that's a client bug - ref https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

PR #6834, opened 2021-10-06
wc: add example to count characters from text

abeltay:

### DIFF
--- a/pages/common/wc.md
+++ b/pages/common/wc.md
@@ -22,3 +22,7 @@
 - Use standard input to count lines, words and characters (bytes) in that order:
 
 `{{find .}} | wc`
+
+- Count the number of characters from standard input:
+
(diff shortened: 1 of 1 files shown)


marchersimon:
This example doesn't really add anything knew. We already explained the `-c` flag in the third example and `stdin` in the 5th example.

mfrw:
> This example doesn't really add anything new. We already explained the `-c` flag in the third example and `stdin` in the 5th example.


I think having an example on `wc  --max-line-length/-L` (maximum display width). It is the one we are missing as per the man page. (suggestion)

abeltay:
Added the command for `wc -L` from https://www.geeksforgeeks.org/wc-command-linux-examples/ and https://linux.die.net/man/1/wc

PR #6842, opened 2021-10-06
Replace non-breaking space with regular space in all French pages

### LINKED ISSUE #6815: Should we stop using the non-breaking space in French pages?
Until recently we enforced a non-breaking space in front of colons in every French page, because that's apparently the correct syntax. But this [comment](https://github.com/tldr-pages/tldr/pull/6673#discussion_r720878900) from @noraj mentioned an interesting point.

> Outside of WYSIWYG editor like MS Word nobody uses them because there is no easy way to insert them from keyboard.

Almost everyone is viewing tldr-pages in the terminal, so you can't even tell the difference between the two spaces. So should we just stop using them? It would also simplify all this a bit.

Ping @noraj, @Nico385412, @patricedenis, @nicokosi

marchersimon:
Closes #6815 
Ping @noraj @Nico385412

### DIFF
--- a/pages.fr/common/7z.md
+++ b/pages.fr/common/7z.md
@@ -1,36 +1,36 @@
 # 7z
 
 > Un archiveur de fichiers avec un haut taux de compression.
-> Plus d'informations : <https://www.7-zip.org>.
+> Plus d'informations : <https://www.7-zip.org>.
 
-- Compresser un fichier ou un dossier :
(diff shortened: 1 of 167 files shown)


sbrl (approved):
Thanks, @marchersimon!

Let's get this merged quickly to avoid a conflict with other PRs in this busy time.

PR #6844, opened 2021-10-06
common-arguments.md: add file

marchersimon:
I've just had an idea for another translation template. Very commonly used arguments would be great to have in all languages, so that we can make minor changes to translated pages, without speaking the language. I've now only added a few arguments, so please comment below what we should translate too.

### DIFF
--- a/contributing-guides/translation-templates/common-arguments.md
+++ b/contributing-guides/translation-templates/common-arguments.md
@@ -1,0 +1,37 @@
+# Common arguments
+
+This page provides translations of commonly used arguments to simplify maintaining pages in foreign languages.  
+The best way to edit this file is by using [tableconvert.com](https://tableconvert.com/).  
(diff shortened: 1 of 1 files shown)


sbrl (approved):
Nice idea! And tableconveert.com is an awesome tool I did not know about :D

marchersimon:
@SethFalco I only plan on using this for things like `apt install {{package}}`. I hardly think any language has different words for this. If there's still ambiguity, we can of course make a note for that.

@navarroaxel I assume you used tableconvert.com to edit the table:

> Only the left-alignment of the header gets lost and has to be readded again (`|----` → `|:---`).

marchersimon:
Do you think we should change `user` to `username`?

marchersimon:
I'll change it to `username`, because in most cases it's more specific.

PR #6846, opened 2021-10-06
*: sync alias pages in translations

blueskyson:
Sync with `set-alias-page.py` from #6845 .


pages.ja/common/llvm-g++.md
```diff
@@ -0,0 +1,7 @@
+# llvm-g++
+
+> このコマンドは `clang++` のエイリアスです.
```
navarroaxel, line 3:
```suggestion
> このコマンドは `clang++` のエイリアスです。
```

pages.ja/common/llvm-ar.md
```diff
@@ -0,0 +1,7 @@
+# llvm-ar
+
+> このコマンドは `ar` のエイリアスです.
```
navarroaxel, line 3: Should this use this `dot`?
```suggestion
> このコマンドは `ar` のエイリアスです。
```

marchersimon: Yes, the Japanese pages use the full-width period. But this should be fixed in [alias-pages.md](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/translation-templates/alias-pages.md#ja) first.

marchersimon:
There are a lot more Japanese alias pages with the wrong period, but I think we should finally merge this PR and fix this later.

navarroaxel (approved):

PR #6849, opened 2021-10-07
systemctl: add global status example

SomethingGeneric:

**Version of the command being documented (if known):** 249.4-2-arch


SomethingGeneric:
apparently i'm a bonehead and can't count.

anyway, per my second commit, I think that listing `systemctl is-active {{service}}` as a separate entry is less useful than `systemctl status`, as the active status is shown in the `systemctl status {{service}}` information.

PR #6869, opened 2021-10-07
tldr: fix typo

ktonekaboni:
Fixed typo. It seems `-o` is `--os` selection. 

Maybe `--update` instead of `-u` is more convenient, because old tldr-hs client on Ubuntu/Debian doesn't understand -u.

### DIFF
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -9,7 +9,7 @@
 
 - Show the tldr page for `cd`, overriding the default platform:
 
-`tldr -p {{android|linux|osx|sunos|windows}} {{cd}}`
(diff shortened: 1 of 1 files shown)


prometoys:
Hi @navarroaxel,
 
thanks for the link. I checked different tldr clients:

tldr-hs 0.9.1
tldr from nmp 3.3.7
tealdeer v1.4.1 (Linux)
tldr-py 0.7.0

None of them understands the option `-p`. Which client accept `-p` ?

CleanMachine1:
Our official clients - Python

Uses -p

sbrl:
For tldr-node-client, see https://github.com/tldr-pages/tldr-node-client/pull/337, which needs some attention

PR #6897, opened 2021-10-08
sed: add pattern to remove ANSI Color Codes example

kavishgr:

### DIFF
--- a/pages/common/sed.md
+++ b/pages/common/sed.md
@@ -34,3 +34,6 @@
 - Replace separator `/` by any other character not used in the find or replace patterns, e.g. `#`:
 
 `sed 's#{{find}}#{{replace}}#' {{filename}}`
+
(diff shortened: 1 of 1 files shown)


marchersimon:
@CleanMachine1 @navarroaxel any idea how to add this example?

sbrl:
This example looks really useful in specific situations, but unfortunately with pages like `sed` we are particularly limited in what we can show given the wide scope of the command. To this end, I'd rather we demonstrate the basic functionality rather than demonstrating specific use-cases.

However, this example does look like a  great fit for a cheat sheet. I wonder if there's a sed-specific one that this could be contributed to that exists already?

marchersimon:
@kavishgr you could have a look at https://github.com/chubin/cheat.sh.
I'll close this PR now.

PR #6941, opened 2021-10-10
calc, cuyo: add German translation

entensee403:
Translating 'arbitrary-precision' to 'beliebige Genauigkeit' is a little bit of a stretch.
The Wikipedia Page (https://de.wikipedia.org/wiki/Langzahlarithmetik) calls that 'Langzahlarithmetik' which I doubt is more clear.

If you got any suggestions for this, feel free to tell me!


entensee403:
Added the cuyo command

PR #6956, opened 2021-10-11
nodemon: change python example to general example

RedHeadphone:

2.0.13

### DIFF
--- a/pages/common/nodemon.md
+++ b/pages/common/nodemon.md
@@ -23,6 +23,10 @@
 
 `nodemon {{arguments}} {{path/to/file.js}}`
 
-- Run non-node scripts:
+- Run an arbitrary non-node script:
 
-`nodemon --exec "{{command_to_run_script}} {{options}}" {{path/to/script/file}}`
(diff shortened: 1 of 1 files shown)


marchersimon:
Why is this a "fix"? You just converted the Python-specific example into a more general example right?

RedHeadphone:
> Why is this a "fix"? You just converted the Python-specific example into a more general example right?

Yeah you are right I should have wrote better explaining commit.

RedHeadphone:
> No problem. Since we have still room left, what do you think about leaving the Python example and adding the general example below it?

What would you suggest heading of this general example?

marchersimon:
How about
```Markdown
- Run an arbitrary non-node script:

`nodemon --exec "{{command_to_run_script}} {{options}}" {{path/to/script}}`

- Run a Python script:

`nodemon --exec "{{python --verbose}}" {{path/to/file.py}}`
```

RedHeadphone:
@marchersimon instead of `--verbose` how about `options`?

PR #6979, opened 2021-10-12
iptables, nft: add masquerade example

Lodifice:
If find the masquerade (NAT) functionality to be the one I actually use most, and it wasn't documented here.
I have to look it up myself sometimes and I think that other user can also benefit from the example.
For the nft page, it's a pity that you have to issue multiple commands (since a `nat` table or `postrouting` chain do not exist by default), but the TLDR guidelines don't allow multiple commands per example.

### DIFF
--- a/pages/linux/iptables.md
+++ b/pages/linux/iptables.md
@@ -21,7 +21,7 @@
 
 - Add a rule to NAT all traffic from the 192.168.0.0/24 subnet to the host's public IP:
 
-`sudo iptables -t nat -A POSTROUTING -s 192.168.0.0/24 -j MASQUERADE`
+`sudo iptables -t {{nat}} -A {{POSTROUTING}} -s {{192.168.0.0/24}} -j {{MASQUERADE}}`
 
 - Delete chain rule:
 
--- a/pages/linux/iptables.md
+++ b/pages/linux/iptables.md
@@ -19,7 +19,7 @@
 
 `sudo iptables -A {{chain}} -s {{ip}} -p {{protocol}} --dport {{port}} -j {{rule}}`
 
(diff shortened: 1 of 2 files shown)


pages/linux/iptables.md
```diff
@@ -19,6 +19,10 @@
 
 `sudo iptables -A {{chain}} -s {{ip}} -p {{protocol}} --dport {{port}} -j {{rule}}`
 
+- Add a rule to NAT all traffic from the 192.168.0.0/24 subnet to the host's public IP:
```
marchersimon, line 22:
```suggestion
- Add a NAT rule to translate all traffic from the `192.168.0.0/24` subnet to the host's public IP:
```
35531403820b148772e76c2e94e4856aff5ac258 should also be applied here and using Backticks for the IP will make it a different color and make it more readable.

pages/linux/nft.md
```diff
@@ -19,6 +19,10 @@
 
 `sudo nft add rule {{inet}} {{filter}} {{input}} {{tcp}} {{dport \{ telnet, ssh, http, https \} accept}}`
 
+- Add a NAT rule to translate all traffic from the 192.168.0.0/24 subnet to the host's public IP:
```
marchersimon, line 22:
```suggestion
- Add a NAT rule to translate all traffic from the `192.168.0.0/24` subnet to the host's public IP:
```

Lodifice:
Darn, I overlooked the wording for iptables. Thanks a lot for your suggestions!

marchersimon (approved):
Thanks @Lodifice

sbrl (approved):
Great addition. Thanks, @Lodifice!

Issue #6382, opened 2021-08-18
page request: egrep, rgrep, bzgrep, bzegrep, bzfgrep, zgrep, zegrep, zfgrep, ptargrep, zipgrep


pixelcmtd:
- [x] [egrep](https://manned.org/egrep) → #6737
- [x] [rgrep](https://manned.org/rgrep) → #7021 
- [x] [bzgrep](https://manned.org/bzgrep) → #7145
- [x] [bzegrep](https://manned.org/bzegrep) → #7199
- [x] [bzfgrep](https://manned.org/bzfgrep) → #7211
- [x] [zgrep](https://manned.org/zgrep) → #4131
- [x] [zegrep](https://www.unix.com/man-page/freebsd/1/zegrep/) → #7233
- [ ] [zfgrep](https://www.unix.com/man-page/freebsd/1/zfgrep/)→#8708
- [x] [ptargrep](https://manned.org/ptargrep) → #7050
- [x] [zipgrep](https://manned.org/zipgrep) → #7004

antoine-amara:
Hello, I'm working on `egrep` command, I will open a PR soon. I can open an other PR with a french translation too.

CleanMachine1:
Thanks @morevi for the notice, we must have missed it with the flood of hacktoberfest :smile: 

fixed now.

dz-at-tc:
Looking at the task list above, I think:

bzfgrep is complete via PR #7211

zegrep is complete via PR #7233

zfgrep is in progress via my new PR #8708 

So assuming my latest submission gets merged eventually, this issue will finally be complete, yay!

PR #7021, opened 2021-10-15
rgrep: add page

dz-at-tc:
Per issue #6382 

grep (GNU grep) 3.1

### DIFF
--- a/pages/linux/rgrep.md
+++ b/pages/linux/rgrep.md
@@ -1,0 +1,25 @@
+# rgrep
+
+> Recursively find patterns in files using regular expressions.
+> Equivalent to `grep -r`.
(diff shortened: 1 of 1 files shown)


CleanMachine1:
Could you extend the description of each command so it says exactly what the command will do, rather than being dependent on the commands above

dz-at-tc:
> Could you extend the description of each command so it says exactly what the command will do, rather than being dependent on the commands above

But of course...  (in my best Grey Poupon :)

Updated PR coming shortly...

marchersimon:
Is this command an alias to `grep -r`, or does it just behave the same way? If it's an alias, I think we should make an alias page:
```Markdown
# rgrep

> This command is an alias of `grep --recursive`.

`tldr grep`
```

PR #7090, opened 2021-10-18
set: add -e example

Managor:

### DIFF
--- a/pages/common/set.md
+++ b/pages/common/set.md
@@ -17,3 +17,7 @@
 - Set various options, e.g. enable `vi` style line editing:
 
 `set -o {{vi}}`
+
+- Set the shell to exit as soon as the first error is encountered. Mostly used in scripts:
+
+`set -e`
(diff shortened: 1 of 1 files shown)


Managor:
Not sure how to correctly format the additional remark about scripts.

marchersimon (approved):
Thanks @Managor

pages/common/set.md
```diff
@@ -17,3 +17,7 @@
 - Set various options, e.g. enable `vi` style line editing:
 
 `set -o {{vi}}`
+
+- Set the shell to exit as soon as the first error is encountered. Mostly used in scripts:
```
sbrl, line 21: Perhaps this will read better? Just a stylistic choice given existing conventions.
```suggestion
- Set the shell to exit as soon as the first error is encountered (mostly used in scripts):
```

Generally we are wary of documenting things that are mostly used in scripts, but in this case I think it's ok.

navarroaxel (approved):

Issue #6458, opened 2021-09-03
page request: fossil


pixelcmtd:
https://fossil-scm.org/

sbrl:
Thanks for requesting this page! Do you want to have a go yourself?

Pages are best created by people who understand the command they are about.

If you do, take a quick read of [CONTRIBUTING.md](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md), which guides you through the process of creating a new page here. Then, just simply send a pull request to this repository - and if you need any assistance, just ask :-)

pixelcmtd:
@sbrl Could you please stop copy-pasting the same message below every page request? If we want this spammed, I'm sure, we can set up a bot or just change the damn Issue template.

CleanMachine1:
I think its more for presentation of issues for anyone browsing thats new.

marchersimon:
@pixelcmtd I have to agree with @ CleanMachine1  here. Also, in regard of spamming, I don't think we need a page request for every page we're adding. Just opening the PR would be enough.

pixelcmtd:
> I think its more for presentation of issues for anyone browsing thats new.

Then it should be part of the Issue template.

PR #7108, opened 2021-10-20
fossil-add: add page

dz-at-tc:
Per issue #6458 

### DIFF
--- a/pages/linux/fossil-add.md
+++ b/pages/linux/fossil-add.md
@@ -3,10 +3,10 @@
 > Put files or directories under Fossil version control.
 > More information: <https://fossil-scm.org/home/help/add>.
 
-- Add files or directories to the current checkout at the next commit:
+- Stage files or directories for the next commit to the current checkout:
 
 `fossil add {{path/to/file_or_directory}}`
 
-- Reset the add state of a checkout such that newly added but not committed files are no longer added; that is, undo uncommitted add command(s):
+- Undo all add commands issued since the last commit to the current checkout:
 
 `fossil add --reset`
--- a/pages/linux/fossil-add.md
+++ b/pages/linux/fossil-add.md
@@ -3,7 +3,7 @@
 > Put files or directories under Fossil version control.
 > More information: <https://fossil-scm.org/home/help/add>.
 
-- Stage a file or directory for the next commit to the current checkout:
+- Put a file or directory under version control, so it will be in the current checkout:
 
 `fossil add {{path/to/file_or_directory}}`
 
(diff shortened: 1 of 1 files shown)


pages/linux/fossil-add.md
```diff
@@ -0,0 +1,12 @@
+# fossil add
+
+> Put files or directories under Fossil version control.
+> More information: <https://fossil-scm.org/home/help/add>.
+
+- Stage files or directories for the next commit to the current checkout:
+
+`fossil add {{path/to/file_or_directory}}`
+
+- Undo all add commands issued since the last commit to the current checkout:
```
marchersimon, line 10: Sounds a bit too complicated
```suggestion
- Remove all added files from the current checkout:
```

pages/linux/fossil-add.md
```diff
@@ -0,0 +1,12 @@
+# fossil add
+
+> Put files or directories under Fossil version control.
```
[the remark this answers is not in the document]
dz-at-tc: > I'm not sure if that's correct. If you change a file which already is under version control, you still have to add it, right?

No, that's not the case in Fossil.  ie, Fossil does not have the Git concept of the index/staging-area.  When you commit in Fossil, all pending changes are committed.  Thus, add and remove are solely concerned with what is or is not tracked under version control (though with remove you can also optionally also actually remove the file from the local checkout).

That said, if my verbiage is confusing to you, it will likely be so to others.  Given my comments above, would you still like to me to proceed with your suggested change?  I have no objections...  Thoughts @marchersimon ?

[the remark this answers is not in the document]
marchersimon: In this case your wording is perfectly fine. I just thought it was more similar to git.

pages/linux/fossil-add.md
```diff
@@ -0,0 +1,12 @@
...
+> More information: <https://fossil-scm.org/home/help/add>.
+
+- Stage a file or directory for the next commit to the current checkout:
```
marchersimon, line 6: This makes it sound like it works like git. How about this?
```suggestion
- Put a file or directory under version control, so it will be in the current checkout:
```
Or similar. But `stage a file` sounds like it is already under version control.

dz-at-tc: Excellent point, thanks!

pixelcmtd (approved):

sbrl (approved):
Thanks, @dz-at-tc!

marchersimon (approved):
Thanks @dz-at-tc

PR #7115, opened 2021-10-20
fossil-delete, fossil-forget, fossil-rm: add page

dz-at-tc:
Per issue #6458 

This is fossil version 2.5 [a425d84480] 2018-02-14 11:45:01 UTC

### DIFF
--- a/pages/linux/fossil-rm.md
+++ b/pages/linux/fossil-rm.md
@@ -12,6 +12,6 @@
 
 `fossil rm --hard {{path/to/file_or_directory}}`
 
-- Re-add all previously removed and uncommitted files to Fossil version control.
+- Re-add all previously removed and uncommitted files to Fossil version control:
 
 `fossil rm --reset`
--- a/pages/linux/fossil-rm.md
+++ b/pages/linux/fossil-rm.md
@@ -1,7 +1,7 @@
 # fossil rm
 
 > Remove files or directories from Fossil version control.
-> Can alternatively be called as `fossil delete` or `fossil forget`.
+> See also `fossil forget`.
 > More information: <https://fossil-scm.org/home/help>.
 
 - Remove a file or directory from Fossil version control:
--- a/pages/linux/fossil-rm.md
+++ b/pages/linux/fossil-rm.md
@@ -4,14 +4,14 @@
 > Can alternatively be called as `fossil delete` or `fossil forget`.
 > More information: <https://fossil-scm.org/home/help>.
 
-- Remove files or directories from Fossil version control:
+- Remove a file or directory from Fossil version control:
 
 `fossil rm {{path/to/file_or_directory}}`
 
-- Remove files or directories from Fossil version control, also deleting the files or directories from the local checkout:
+- Remove a file or directory from Fossil version control, and also delete it from the disk:
 
 `fossil rm --hard {{path/to/file_or_directory}}`
 
-- Reset the deleted state of a checkout such that newly rm'd but not committed files are no longer deleted; that is, undo uncommitted rm command(s):
+- Re-add all previously removed and uncommitted files to Fossil version control.
 
 `fossil rm --reset`
--- a/pages/linux/fossil-rm.md
+++ b/pages/linux/fossil-rm.md
@@ -1,0 +1,17 @@
+# fossil rm
+
+> Remove files or directories from Fossil version control.
+> Can alternatively be called as `fossil delete` or `fossil forget`.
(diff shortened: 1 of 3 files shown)


pages/linux/fossil-rm.md
```diff
@@ -0,0 +1,17 @@
+# fossil rm
+
+> Remove files or directories from Fossil version control.
+> Can alternatively be called as `fossil delete` or `fossil forget`.
```
marchersimon, line 4: We usually do this with alias pages. Like this:
```markdown
# fossil delete

> This command is an alias of `fossil rm`.
> More information: <https://fossil-scm.org/home/help>.
  
`tldr fossil rm`
```
However the `forget` subcommand is not the exact same:
> If the "--hard" option is specified -OR- the "mv-rm-files" setting is non-zero, files WILL BE removed from disk as well. **This does NOT apply to the 'forget' command**. 

But I think we could still make an alias page like this:
```markdown
# fossil forget

> This command is an alias of `fossil rm`, with the exception, that it never removes files from the disk.
> More information: <https://fossil-scm.org/home/help>.
  
`tldr fossil rm`
```

pages/linux/fossil-rm.md
```diff
@@ -0,0 +1,17 @@
...
+> More information: <https://fossil-scm.org/home/help>.
+
+- Remove files or directories from Fossil version control:
```
marchersimon, line 7:
```suggestion
- Remove a file or directory from Fossil version control:
```

pages/linux/fossil-rm.md
```diff
@@ -0,0 +1,17 @@
...
+
+`fossil rm {{path/to/file_or_directory}}`
+
+- Remove files or directories from Fossil version control, also deleting the files or directories from the local checkout:
```
marchersimon, line 11: How about this? It's a bit shorter.
```suggestion
- Remove a file or directory from Fossil version control, and also delete it from the disk:
```

pages/linux/fossil-rm.md
```diff
@@ -0,0 +1,17 @@
...
+
+`fossil rm --hard {{path/to/file_or_directory}}`
+
+- Reset the deleted state of a checkout such that newly rm'd but not committed files are no longer deleted; that is, undo uncommitted rm command(s):
```
marchersimon, line 15: How about this?
```suggestion
- Re-add all previously removed and uncommitted files to Fossil version control.
```

pages/linux/fossil-rm.md
dz-at-tc: Hi @marchersimon - Ok, thanks for clarifying.

Just so I'm 100% clear, you want me to remove line 4 (ie, Can alternatively be called as `fossil delete` or `fossil forget`)?

I'm happy to add the alias pages that point to fossil rm, but if I do remove line 4 then there is nothing that points to the alternate spellings which, as you point out, perhaps may be interesting, say given the subtle difference of fossil forget.

Again, I'm happy to make the edit, I just want to be clear that losing the two way references is ok, thanks!

marchersimon: We can just replace this line with `` > See also `fossil forget` and `fossil delete` ``, even though I'm not sure if we should even mention `fossil delete`, because it's the exact same.

pages/linux/fossil-rm.md
```diff
@@ -0,0 +1,17 @@
...
+
+- Remove a file or directory from Fossil version control:
+
+`fossil rm {{path/to/file_or_directory}}`
+
+- Remove a file or directory from Fossil version control, and also delete it from the disk:
+
+`fossil rm --hard {{path/to/file_or_directory}}`
+
+- Re-add all previously removed and uncommitted files to Fossil version control.
```
marchersimon, line 15: Sorry, my bad. There should be a colon
```suggestion
- Re-add all previously removed and uncommitted files to Fossil version control:
```

pages/linux/fossil-rm.md
dz-at-tc: Sure, I like that.  Once I add the two alias pages, I'll edit this one to "see also forget" (agreed re delete being the same).  Thanks again @marchersimon !

dz-at-tc: Hey @marchersimon - Ok, I think I have all the changes in.  Just shout if I missed anything, thanks again!

marchersimon (approved):
Thanks @dz-at-tc

navarroaxel (approved):

PR #7132, opened 2021-10-21
pacman: add sudo and use long options

marchersimon:


maverick1872:
Curious why you opted for completely removing the shorthand options instead of keeping them both?

marchersimon:
Because we still couldn't decide whether we should use long or short options for very common pages. Long and short options in one page would probably make the page more confusing. Check #8046, where we discussed this issue.

PR #7153, opened 2021-10-23
lsof: edit -iTCP example with -i6TCP, -n and -P

jerryc05:

### DIFF
--- a/pages/common/lsof.md
+++ b/pages/common/lsof.md
@@ -32,6 +32,6 @@
 
 `lsof +D {{path/to/directory}}`
 
-- Find the process that is listening on a local TCP port:
+- Find the process that is listening on a local IPv4 TCP port, don't convert network numbers or port numbers:
 
-`lsof -iTCP:{{port}} -sTCP:LISTEN`
+`lsof -i4TCP:{{port}} -sTCP:LISTEN -n -P`
--- a/pages/common/lsof.md
+++ b/pages/common/lsof.md
@@ -32,6 +32,6 @@
 
 `lsof +D {{path/to/directory}}`
 
-- Find the process that is listening on a local IPv6 TCP port, don't convert network numbers or port numbers:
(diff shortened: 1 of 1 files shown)


pages/common/lsof.md
```diff
@@ -32,6 +32,6 @@
 
 `lsof +D {{path/to/directory}}`
 
-- Find the process that is listening on a local TCP port:
+- Find the process that is listening on a local IPv4 TCP port, don't convert network numbers or port numbers:
 
-`lsof -iTCP:{{port}} -sTCP:LISTEN`
+`lsof -i4TCP:{{port}} -sTCP:LISTEN -n -P`
```
sbrl, line 37: Not really a fan of explicitly specifying IPv4 here. IPv4 will be going away eventually - we should be promoting the use and uptake of IPv6 instead.

jerryc05: True, but my intention was to inform the users of the -4 (or -6) flag here. 
I believe the users will realize that they can change -4 to -6 for IPv6.

sbrl: True, but could we specify `-6` instead of `-4` here please?

jerryc05: Done.

pages/common/lsof.md
```diff
@@ -32,6 +32,6 @@
 
 `lsof +D {{path/to/directory}}`
 
-- Find the process that is listening on a local TCP port:
+- Find the process that is listening on a local IPv6 TCP port, don't convert network numbers or port numbers:
```
marchersimon, line 35: I think that reads a bit better:
```suggestion
- Find the process that is listening on a local IPv6 TCP port and don't convert network or port numbers:
```

marchersimon (approved):
Thanks @jerryc05 
I just left a small suggestion

navarroaxel (approved):

sbrl (approved):
Thanks, @jerryc05!

PR #7227, opened 2021-10-26
set-more-info-link: fix Turkish translation

navarroaxel:
`Daha fazla bilgi için` means `For more information`. I think this is a more exact translation for Turkish pages.


sbrl (approved):

PR #7230, opened 2021-10-26
common-arguments.md: add zh and zh_TW translation

blueskyson:
See #7102 . This PR was based on main branch. I move the changes into another branch here.


marchersimon (approved):
Thanks @blueskyson

PR #7272, opened 2021-10-27
localectl: add page

Managor:

### DIFF
--- a/pages/linux/localectl.md
+++ b/pages/linux/localectl.md
@@ -3,7 +3,7 @@
 > Control the system locale and keyboard layout settings.
 > More information: <https://www.freedesktop.org/software/systemd/man/localectl.html>.
 
-- Show current settings of the system locale and keyboard mapping.:
+- Show current settings of the system locale and keyboard mapping:
 
 `localectl`
 
--- a/pages/linux/localectl.md
+++ b/pages/linux/localectl.md
@@ -15,6 +15,10 @@
 
 `localectl set-locale {{LANG}}={{en_US.UTF-8}}`
 
+- List available keymaps:
+
+`localectl list-keymaps`
+
 - Set the system keyboard mapping for the console and X11:
 
-`localectl set-keymap {{en}}`
+`localectl set-keymap {{us}}`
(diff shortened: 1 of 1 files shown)


pages/linux/localectl.md
```diff
@@ -0,0 +1,20 @@
+# localectl
+
+> Control the system locale and keyboard layout settings.
+> More information: <https://www.freedesktop.org/software/systemd/man/localectl.html>.
+
+- Show current settings of the system locale and keyboard mapping:
```
marchersimon, line 6:
```suggestion
- Show the current settings of the system locale and keyboard mapping:
```

pages/linux/localectl.md
```diff
@@ -0,0 +1,20 @@
...
+
+`localectl`
+
+- List available locales:
+
+`localectl list-locales`
+
+- Set a system locale variable:
+
+`localectl set-locale {{LANG}}={{en_US.UTF-8}}`
```
marchersimon, line 16:
```suggestion
`sudo localectl set-locale {{LANG}}={{en_US.UTF-8}}`
```

pages/linux/localectl.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Set the system keyboard mapping for the console and X11:
+
+`localectl set-keymap {{en}}`
```
marchersimon, line 20:
```suggestion
`sudo localectl set-keymap {{en}}`
```

pages/linux/localectl.md
Managor: Is the sudo necessary since localectl will prompt you for the password anyways?

marchersimon: In may case it opens the GUI dialog, which is a bit annoying when you're working in the terminal, but that's probably just my setup. Generally, I think we should still include it, just to be safe.

Managor: Timedatectl exhibits the same behavior and it doesn't have sudo listed if we want to go by precedent. 
I personally would like to put as little as possible in the example command as long as it works. The simpler the better.

marchersimon: Sure, then let's leave it out.

marchersimon (approved):
Thanks @Managor

PR #7304, opened 2021-10-30
pw-loopback: refresh

Managor:

0.3.39

### DIFF
--- a/pages/linux/pw-loopback.md
+++ b/pages/linux/pw-loopback.md
@@ -15,7 +15,7 @@
 
 `pw-loopback -m '{{[FL FR]}}' --playback-props='{{media.class=Audio/Source}}'`
 
-- Create a dummy loopback that doesn't automatically connect to anything:
+- Create a dummy loopback device that doesn't automatically connect to anything:
 
 `pw-loopback -m '{{[FL FR]}}' --capture-props='{{media.class=Audio/Sink}}' --playback-props='{{media.class=Audio/Source}}'`
 
--- a/pages/linux/pw-loopback.md
+++ b/pages/linux/pw-loopback.md
@@ -7,14 +7,22 @@
 
 `pw-loopback`
 
-- Create a sink that can be selected in applications that forwards the data to another sink:
+- Create a dummy loopback that doesn't automatically connect to anything
 
-`pw-loopback -m '{{[FL FR]}}' --capture-props='{{media.class=Audio/Sink node.name=my-sink}}'`
+`pw-loopback -m '{{[FL FR]}}' --capture-props='{{media.class=Audio/Sink}}' --playback-props='{{media.class=Audio/Source}}'`
 
-- Create a sink that can be selected in applications that swaps the left and right channels to another sink:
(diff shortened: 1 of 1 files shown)


pages/linux/pw-loopback.md
```diff
@@ -7,14 +7,22 @@
...
-`pw-loopback --capture-props='{{media.class=Audio/Sink node.name=my-sink audio.position=[FL FR]}}' --playback-props='{{audio.position=[FR FL]}}'`
+`pw-loopback -m '{{[FL FR]}}' --playback-props='{{media.class=Audio/Source}}'`
 
-- Create a source that can be selected in applications that swaps the left and right channels from another source:
+- Create a dummy loopback device that doesn't automatically connect to anything:
 
-`pw-loopback --capture-props='{{audio.position=[FR FL]}}' --playback-props='{{media.class=Audio/Source node.name=my-source audio.position=[FL FR]}}'`
+`pw-loopback -m '{{[FL FR]}}' --capture-props='{{media.class=Audio/Sink}}' --playback-props='{{media.class=Audio/Source}}'`
+
+- Create a loopback device that automatically connects to the speakers and swaps the left and right channels between the sink and source:
```
marchersimon, line 22: I don't quite understand the `swaps the left and right channels between the sink and source`. Is there a way to explain this a bit without making it longer?

pages/linux/pw-loopback.md
```diff
@@ -7,14 +7,22 @@
...
+
+`pw-loopback --capture-props='{{media.class=Audio/Sink audio.position=[FL FR]}}' --playback-props='{{audio.position=[FR FL]}}'`
+
+- Create a loopback device that automatically connects to the microphone and swaps the left and right channels between the sink and source:
```
marchersimon, line 26: Same here

pages/linux/pw-loopback.md
Managor: I think that's the correct terminology. The audio enters the sink and exits out of the audio source with the channels stereo channels swapped

sbrl: Hmmm, I'm not sure there's a better way we can word this. Thanks for the clarification here @Managor - I think audio stuff can just be complicated sometimes.

What about adding a 2nd description line something like this?

```markdown
> A sink is ....., and a source is ......
```

marchersimon: Well, after the explanation it makes sense. I assume if someone is working with pipewire it shouldn't be a problem.
@sbrl isn't `sink`/`source` a rather common terminology?

marchersimon (approved):
Thanks @Managor

pages/linux/pw-loopback.md
Managor: If you check the More information site of this page, sink and source are used there extensively

sbrl: Yeah, they are @marchersimon - was just pointing out something that could potentially improve clarity

At any rate, I'm not sure we can really improve on the example description here

sbrl (approved):
Thanks, @Managor!

PR #7330, opened 2021-10-31
tooling: add PR labeler

### LINKED ISSUE #7325: Automate application of labels for PRs
I feel like it would be worth considering automating label application to PRs. In the past two weeks, there's been two new labels added to the organization. Scanning through, there is 24 labels that can be applied to a given PR. Each of these labels are applied manually, which just does not seem tenable to me, and that label application would be missed, especially for newcomers into the tldr organization.

As low hanging fruit, I think the following PRs could be automated without much debate on when to apply them:

- `new command` - PR is for creating a new page
- `page edit` - PR affects only one existing page
- `translation` - PR is for a page not in `pages/` directory

Maybe `mass changes` could be applied as well if there was some agreement on what "multiple pages" means (2+, 10+, etc.).

Once we've automated application of these simple ones, follow-up work could be applied for:

- `review needed` - PR does not yet have 2 reviews (could be applied after X days like stalebot posts)
- `waiting` - PR review requested changes?

An ideal example of this level of label automation would be the https://github.com/DefinitelyTyped/DefinitelyTyped repository.

MasterOdin:
Closes #7325 

Adds a new GH action that acts on any new PR to automatically update its labels. I've added rules for the adding the following:

- new command
- page edit
- translation
- documentation
- tooling

While dealing with these labels could probably be done by using the existing [actions/labeler](https://github.com/actions/labeler), I've gone with writing our own as we may add to it to support things like `mass edit`, `reviews wanted`, etc. which will require more complex logic than simple globing, but will re-use a lot of the logic written here.


navarroaxel:
Sounds like this should be in its own repo.

.github/actions/labeler/src/labeler.ts
```diff
@@ -0,0 +1,122 @@
...
+  const pagePattern = /pages(?:\.[a-z_]+)?/i;
+  const documentationPattern = /[a-z\-]+\.md/i;
+  const toolingPattern = /.*\.(py|[jt]s|sh)/i;
+  for (const file of changedFiles) {
+    if (pagePattern.test(file.filename) || (file.previous_filename && pagePattern.test(file.previous_filename))) {
+      if (translatedPagePattern.test(file.filename) || (file.previous_filename && translatedPagePattern.test(file.previous_filename))) {
+        labels.add('translation');
+      }
+      if (file.status === 'added') {
+        labels.add('new command');
```
marchersimon, line 52: So a translation gets the labels `translation` and `new command`? "new command" should only be applied to new English pages.
Would this fix it?
```suggestion
      } else if (file.status === 'added') {
        labels.add('new command');
```

MasterOdin: What should happen if they add a new page along with a translated version of it? It should have both tags there, right?

marchersimon: In this case, yes. Didn't think of that.

marchersimon:
> Sounds like this should be in its own repo.

@MasterOdin what do you think about this?

MasterOdin:
I do not particularly care. Certainly benefits to isolating the action to its own repo as it has its own unique tooling to work on it.

navarroaxel:
Moves to its own repository https://github.com/tldr-pages/tldr-labeler-action.

PR #7358, opened 2021-11-01
java: add classpath example and update link

erikweibust:

### DIFF
--- a/pages/common/java.md
+++ b/pages/common/java.md
@@ -1,11 +1,15 @@
 # java
 
 > Java Application Launcher.
-> More information: <https://java.com>.
+> More information: <https://www.oracle.com/java/technologies/javase-documentation.html>.
 
 - Execute a java `.class` file that contains a main method by using just the class name:
 
 `java {{classname}}`
+
+- Execute a java program and use additional third-party or user-defined classes
+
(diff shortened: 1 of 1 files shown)


pages/common/java.md
```diff
@@ -1,12 +1,16 @@
 # java
 
 > Java Application Launcher.
-> More information: <https://java.com>.
+> More information: <https://www.oracle.com/java/technologies/javase-documentation.html>.
```
navarroaxel, line 4: I think this is a better reference for the command-line
```suggestion
> More information: <https://docs.oracle.com/en/java/javase/17/docs/specs/man/java.html>.
```

navarroaxel (changes requested):

owenvoke (approved):

pages/common/java.md
erikweibust: I chose the root page of the Java SE documentation over the Java 17 page as Java changes versions every 6 months and that link would be outdated as quick as the PR would be accepted. That said, I'm open to changing it if that is the consensus.

navarroaxel: I understand, and yes it will be outdated. but it's hard to find that with the link that you provide.

navarroaxel (approved):

PR #7379, opened 2021-11-05
rsync: fix --delete example description

dufferzafar:
The page said that the --delete parameter deletes local files that don't exist on remote.

But the actual behaviour is OPPOSITE of that!

From the man page:

```
--delete                delete extraneous files from dest dirs
```

https://www.mankier.com/1/rsync#--delete

### DIFF
--- a/pages/common/rsync.md
+++ b/pages/common/rsync.md
@@ -28,7 +28,7 @@
 
 `rsync -rauL {{remote_host}}:{{path/to/remote_file}} {{path/to/local_directory}}`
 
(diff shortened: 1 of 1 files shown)


mfrw (approved):
Thank you for your contribution @dufferzafar .
nice catch!

dufferzafar:
@mfrw Glad you agree with me, I was actually doubting whether my interpretation is correct 😅 

Anyway, I think we should update docs of other languages as well!

sbrl (approved):
Great catch. Thanks, @dufferzafar!

PR #7401, opened 2021-11-09
i3exit: add page

marchersimon:


pages/linux/i3exit.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Log out of i3:
+
+`i3exit logout`
+
+- Lock i3:
+
+`i3exit lock`
+
+- Shut down the machine:
```
navarroaxel, line 14: We use `system` in `poweroff` and `reboot` pages
```suggestion
- Shut down the system:
```

pages/linux/i3exit.md
```diff
@@ -0,0 +1,32 @@
...
+
+`i3exit shutdown`
+
+- Suspend the machine:
+
+`i3exit suspend`
+
+- Switch the current user:
```
navarroaxel, line 22: Switch to what? :thinking:

marchersimon: I logs you out and prompts you to choose a user. It's the same as in every DE when you click on "switch user".

navarroaxel: It's not the same goes to the select user screens keeping the current session open, than log off and go to the select user to login screen. It's nice to leave that clear here.

pages/linux/i3exit.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Lock i3:
+
+`i3exit lock`
+
+- Shut down the system:
+
+`i3exit shutdown`
+
+- Suspend the machine:
```
navarroaxel, line 18:
```suggestion
- Suspend the system:
```

pages/linux/i3exit.md
```diff
@@ -0,0 +1,32 @@
...
+
+`i3exit suspend`
+
+- Switch the current user:
+
+`i3exit switch_user`
+
+- Hibernate the machine:
```
navarroaxel, line 26:
```suggestion
- Hibernate the system:
```

pages/linux/i3exit.md
```diff
@@ -0,0 +1,32 @@
...
+
+`i3exit hibernate`
+
+- Reboot the machine:
```
navarroaxel, line 30:
```suggestion
- Reboot the system:
```

pages/linux/i3exit.md
marchersimon: Well, I'm not actually sure. I tried it, but it just showed me a blank screen and crashed. But what it does is calling `dm-tool switch-to-greeter`. From the man pages

> Switch to the greeter suitable for logging into a new session.

Sounds like it doesn't log you out. How about we change the description to
```suggestion
- Switch to the login screen to log in as a different user:
```

sbrl: It's sounds like this is essentially a "switch user" feature. @marchersimon's suggestion seems logical to me.

navarroaxel (approved):

CleanMachine1 (approved):

PR #7413, opened 2021-11-12
sed: fix -i example

Pakku:
The empty argument '' causes the command to return an error.

### DIFF
--- a/pages/osx/sed.md
+++ b/pages/osx/sed.md
@@ -13,7 +13,7 @@
 
 - Replace all occurrences of a string [i]n a file, overwriting the file (i.e. in-place):
 
-`sed -i '' 's/{{find}}/{{replace}}/g' {{filename}}`
+`sed -i 's/{{find}}/{{replace}}/g' {{filename}}`
 
 - Replace only on lines matching the line pattern:
 


marchersimon:
Both variants work for me. Could you please provide the exact command you executed and the error you got?

Pakku:
Sure, this is what I'm getting. It appears that with the first variant, the second parameter is treated as the name of the file.

```
{▸} ~ sed -i '' 's/find/replace/g' example.txt
sed: can't read s/find/replace/g: No such file or directory
{▪} ~ sed -i 's/find/replace/g' example.txt   
{▸} ~ 
```

As a reference, [here](https://github.com/tldr-pages/tldr/blob/main/pages/common/sed.md) is without the empty parameter.

marchersimon:
My bad, I wrote `' ''s...` instead of `'' 's...`, which worked.

marchersimon (approved):
Thanks @Pakku

navarroaxel (approved):

navarroaxel (changes requested):
I've tested the example and it's working properly. Are you using `macOS`?

Pakku:
Yes, i'm using macOS. Is it working for you with the empty parameter? For me, it is taking the second parameter as the file path instead of the find/replace string.

marchersimon:
@navarroaxel for me it also didn't work under Linux on bash and zsh with GNU sed 4.8

navarroaxel:
@Pakku Did you install `sed` using `brew install gnu-sed`?
GNU `sed` != macOS `sed`

Pakku:
You are right @navarroaxel ! It seems that i'm using GNU sed 😅 Sorry for the inconveniences!

PR #7414, opened 2021-11-12
locate: move to common

blueskyson:

The function of `locate` is the same on both `linux` and `osx`. But the `linux` one provides more options to use.


pages.zh/common/locate.md
```diff
@@ -13,4 +13,4 @@
 
 - 重新建立文件数据索引数据库。如果要查找最近添加的文件，则需要执行此操作：
 
-`sudo /usr/libexec/locate.updatedb`
```
marchersimon, line 16: Are you sure this isn't necessary on macOS? This seems pretty intentional, especially because it's the only difference to the Linux page.

marchersimon, line 16: Are you sure this isn't necessary on macOS? This seems pretty intentional, especially because it's the only difference to the Linux page.

blueskyson: I just tried `locate/focal,now 4.7.0-1ubuntu1 amd64` on ubuntu, running `sudo updatedb` works fine. Besides, the executable `/usr/libexec/locate.updatedb` does not exist.

marchersimon: > Or use 2 examples to specify the difference on `linux` and `osx`?

That would kind of defeat the purpose of our separated platform directories. We shouldn't let the user have to filter out the useful part more than necessary.

PR #7415, opened 2021-11-12
fish: update page

EmilySeville7cfg:
- short options preferred over long ones
- page made more consistent with elvish page (check my PR)
- commands reordered: interactive commands, non-interactive commands
- set command example removed because it is example of Fish language but not of fish command

**Version of the command being documented (if known):** fish, version 3.3.1

### DIFF
--- a/pages/common/fish.md
+++ b/pages/common/fish.md
@@ -7,26 +7,22 @@
 
 `fish`
 
-- Execute a command and then exit:
+- Start an interactive shell session in [p]rivate mode, where the shell does not access old history or save new history:
+
+`fish -p`
+
+- Execute a [c]ommand and then exit:
 
 `fish -c "{{command}}"`
 
-- Execute a script:
+- Execute a script and then exit:
 
 `fish {{path/to/script.fish}}`
 
-- Check a script for syntax errors:
+- Check a script for syntax errors but [n]ot execute it and then exit:
 
-`fish --no-execute {{path/to/script.fish}}`
+`fish -n {{path/to/script.fish}}`
 
-- Start an interactive shell session in private mode, where the shell does not access old history or save new history:
+- Print the Fish version and then exit:
 
-`fish --private`
-
-- Display version information and exit:
-
-`fish --version`
-
-- Set and export environmental variables that persist across shell restarts (from within the shell only):
-
-`set -Ux {{variable_name}} {{variable_value}}`
+`fish -v`
--- a/pages/common/fish.md
+++ b/pages/common/fish.md
@@ -19,7 +19,7 @@
 
 `fish {{path/to/script.fish}}`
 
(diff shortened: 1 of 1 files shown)


pages/common/fish.md
```diff
@@ -7,26 +7,26 @@
...
 - Execute a script:
 
 `fish {{path/to/script.fish}}`
 
-- Check a script for syntax errors:
-
-`fish --no-execute {{path/to/script.fish}}`
-
-- Start an interactive shell session in private mode, where the shell does not access old history or save new history:
+- Check a script for syntax errors but [n]ot execute it:
```
sbrl, line 22: Perhaps this will read better?
```suggestion
- Check a script for syntax errors but do [n]ot execute it:
```

sbrl (commented):
Thanks for the edit, @EmilySeville7cfg! Is there a reason why you've used the short arguments here instead of the long form? In tldr-pages we prefer the long form.

In any case, this page is looking pretty close to there :D

EmilySeville7cfg:
> Thanks for the edit, @EmilySeville7cfg! Is there a reason why you've used the short arguments here instead of the long form? In tldr-pages we prefer the long form.

The reason [was](https://github.com/tldr-pages/tldr/issues/7416#issuecomment-968088995):

> I think tldr pages can help use and remember short options faster because they are less user friendly than long ones. That's why I prefer to use short options in pages. But the benefit of short options is less typing. ;)

But I fixed it - now only long options appear in .md.

pages/common/fish.md
```diff
@@ -23,10 +27,6 @@
...
 `fish --private`
 
-- Display version information and exit:
+- Print the version:
 
 `fish --version`
-
-- Set and export environmental variables that persist across shell restarts (from within the shell only):
-
-`set -Ux {{variable_name}} {{variable_value}}`
```
marchersimon, line 32: I'm not sure if we should remove this. This specific command is tightly connected to the shell and very useful. We cannot really document it in `set.md`, because it's a shell builtin in this case.

EmilySeville7cfg: I can leave this example here also because there are less than 8 examples now. 😅 I agree, it's useful and I think we can add to another tldr pages builtins useful samples such as this one.

pages/common/fish.md
```diff
@@ -23,10 +27,10 @@
 
 `fish --private`
 
-- Display version information and exit:
+- Define and export environmental variable that persist across shell restarts (builtin):
```
marchersimon, line 30:
```suggestion
- Define and export an environmental variable that persists across shell restarts (builtin):
```

marchersimon (approved):
Just this last suggestion
Thanks @EmilySeville7cfg

sbrl (approved):
Thanks, @EmilySeville7cfg!

PR #7434, opened 2021-11-13
install-tl: add page

marchersimon:


sbrl (approved):
Thanks, @marchersimon!

reinhart1010 (approved):

reinhart1010:
The rest of the content is OK for be, but should https://www.tug.org/texlive/doc/install-tl.html be more appropriate for "More Information"? (similar to the case with my `warp-cli`)?

navarroaxel (approved):

PR #7435, opened 2021-11-14
ngs: add page

EmilySeville7cfg:

**Version of the command being documented (if known):** 0.2.12

### DIFF
--- a/pages/common/ngs.md
+++ b/pages/common/ngs.md
@@ -5,11 +5,11 @@
 
 - Execute a code:
 
-`ngs -e "{{code}}"`
+`ngs -e "echo('ngs is executed')"`
 
 - Execute a script:
 
-`ngs {{path/to/script.ngs}}`
+`ngs "{{path/to/script.ngs}}"`
 
 - Print the version:
 
--- a/pages/common/ngs.md
+++ b/pages/common/ngs.md
@@ -3,9 +3,9 @@
 > Fully featured scripting language created specifically for Ops. REPL is being developed.
 > More information: <https://github.com/ngs-lang/ngs>.
 
-- Execute a command:
+- Execute a code:
 
-`ngs -e "{{command}}"`
+`ngs -e "{{code}}"`
 
 - Execute a script:
 
--- a/pages/common/ngs.md
+++ b/pages/common/ngs.md
@@ -5,7 +5,7 @@
 
 - Execute a code snippet:
 
-`ngs -e "echo('ngs is executed')"`
+`ngs -e "{{echo('ngs is executed')}}"`
 
 - Execute a script:
 
--- a/pages/common/ngs.md
+++ b/pages/common/ngs.md
@@ -1,6 +1,6 @@
 # ngs
 
-> Fully featured scripting language created specifically for Ops. REPL is being developed.
+> Scripting language created specifically for Ops.
 > More information: <https://github.com/ngs-lang/ngs>.
 
 - Execute a code:
--- a/pages/common/ngs.md
+++ b/pages/common/ngs.md
@@ -3,7 +3,7 @@
 > Scripting language created specifically for Ops.
 > More information: <https://github.com/ngs-lang/ngs>.
 
-- Execute a code:
+- Execute a code snippet:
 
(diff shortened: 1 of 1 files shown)


pages/common/ngs.md
```diff
@@ -0,0 +1,16 @@
+# ngs
+
+> Fully featured scripting language created specifically for Ops. REPL is being developed.
+> More information: <https://github.com/ngs-lang/ngs>.
+
+- Execute a command:
```
navarroaxel, line 6: It isn't a command, it's NGS code, or a expression, because you can't use a Bash-like `echo` command, you should run `ngs -e 'echo("hello")'`
```suggestion
- Execute a NGS code:
```

pages/common/ngs.md
```diff
@@ -0,0 +1,16 @@
...
+
+`ngs -e "{{command}}"`
```
navarroaxel, line 8:
```suggestion
`ngs -e "{{code}}"`
```

navarroaxel (approved):

pages/common/ngs.md
```diff
@@ -0,0 +1,16 @@
+# ngs
+
+> Fully featured scripting language created specifically for Ops. REPL is being developed.
```
sbrl, line 3: This information might be outdated in the future. `Fully featured` is subjective and doesn't help more clearly define what the command actually is or does.
```suggestion
> Scripting language created specifically for Ops.
```

pages/common/ngs.md
```diff
@@ -0,0 +1,16 @@
...
+> More information: <https://github.com/ngs-lang/ngs>.
+
+- Execute a code:
```
sbrl, line 6: Perhaps this will read better?
```suggestion
- Execute a snippet of code:
```

sbrl (commented):

pages/common/ngs.md
EmilySeville7cfg: Changed to `Execute a code snippet:`.

pages/common/ngs.md
```diff
@@ -0,0 +1,16 @@
+# ngs
+
+> Scripting language created specifically for Ops.
+> More information: <https://github.com/ngs-lang/ngs>.
+
+- Execute a code snippet:
+
+`ngs -e "echo('ngs is executed')"`
```
CleanMachine1, line 8:
```suggestion
`ngs -e "{{echo('ngs is executed')}}"`
```

pages/common/ngs.md
```diff
@@ -0,0 +1,16 @@
...
+> Scripting language created specifically for Ops.
+> More information: <https://github.com/ngs-lang/ngs>.
+
+- Execute a code snippet:
+
+`ngs -e "{{echo('ngs is executed')}}"`
+
+- Execute a script:
+
+`ngs "{{path/to/script.ngs}}"`
```
navarroaxel, line 12: Wait, we don´t  use quotes for paths
```suggestion
`ngs {{path/to/script.ngs}}`
```

navarroaxel (changes requested):

pages/common/ngs.md
EmilySeville7cfg: But what if script name contains spaces? It will be treated as distinct arguments instead of one file name. I think it's better to use double quotes where we can. ;)

sbrl: The thing to remember here is that we don't want to document shell builtins. The purpose of tldr pages is to document specific commands. To this end, quoting things that contain spaces I think we can assume is common knowledge, so we don't need to unnecessarily complicate pages by including the syntax here.

EmilySeville7cfg: > - When they are required for a command to function correctly
> - When an argument will always contain spaces.

The problem is `ngs` will fail to run script if it's name is unquoted and contains spaces. According to [your message](https://github.com/tldr-pages/tldr/issues/7521#issuecomment-991406076) I have to quote script path.

navarroaxel (approved):

sbrl (approved):
Thanks, @EmilySeville7cfg!

PR #7441, opened 2021-11-15
style-guide: refresh and add shell-specific pages

EmilySeville7cfg:

### DIFF
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -74,6 +74,10 @@ Extensions
   For instance, in `find.md`'s example "Find files by extension" (`find {{root_path}} -name '{{*.ext}}'`)
   using `{{*.ext}}` explains the command without being unnecessarily specific;
   while in `wc -l {{file}}` using `{{file}}` (without extension) is sufficient.
+  
+### Options
+
+- Prefer long options instead of short ones whenever is possible.
 
 ### Special cases
 - If a command performs irreversible changes to a file system or devices,
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -6,7 +6,7 @@ This page lists specific formatting instructions for `tldr` pages.
 
 The basic format of each page should match the following template:
 
-```
+```md
 # command-name
 
 > Short, snappy description.
@@ -26,7 +26,7 @@ There actually is a linter/formatter that enforces the format above.
 It is run automatically on every pull request,
 but you may install it to test your contributions locally before submitting them:
 
-```
+```bash
 npm install tldr-lint
 tldrl -f {{page.md}}
 ```
@@ -36,7 +36,7 @@ For other ways to use `tldrl`, such as linting an entire directory, check out (w
 
 Your client may be able to preview a page locally using the `--render` flag:
 
-```
+```bash
 tldr --render {{page.md}}
 ```
 
@@ -48,6 +48,7 @@ in order to allow `tldr` clients to highlight them.
 Keep the following guidelines in mind when choosing tokens:
 
 ### Naming
+
 - Use short but descriptive tokens,
   such as `{{source_file}}` or `{{wallet.txt}}`.
 - Use [`snake_case`](https://wikipedia.org/wiki/snake_case) for multi-word tokens.
@@ -55,6 +56,7 @@ Keep the following guidelines in mind when choosing tokens:
   For example, use `iostat {{2}}` rather than `iostat {{interval_in_secs}}`.
 
 ### Paths
+
 - Use `{{filename}}` rather than `{{file_name}}`.
 - For any reference to paths of files or directories,
   use the format `{{path/to/<placeholder>}}`,
@@ -80,6 +82,7 @@ Keep the following guidelines in mind when choosing tokens:
 - Prefer long options instead of short ones whenever is possible.
 
 ### Special cases
+
 - If a command performs irreversible changes to a file system or devices,
   write every example in a way that they cannot be thoughtlessly copy-pasted.
   For example, instead of `ddrescue --force --no-scrape /dev/sda /dev/sdb`
@@ -167,8 +170,9 @@ since omitting it can create ambiguity.
 > Delete the Git branches, tags and remotes.
 
 The example above does not use a serial comma, so this could mean one of two things:
-* Delete the Git branches named `tags` and `remotes`.
-* Delete all of the following: Git branches, Git tags, and Git remotes.
+
+- Delete the Git branches named `tags` and `remotes`.
+- Delete all of the following: Git branches, Git tags, and Git remotes.
 
 This can be resolved by inserting a comma before the "and" or "or" in the final element in the list.
 
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -112,7 +112,13 @@ If a page describes some shell (Bash for instance) then when it's possible try f
 - _any other examples_
 - Print the version:
 
-If you want to include shell builtin sample please mark it with `(builtin)` word. For example:
+If you want to include shell builtin sample please place `(builtin)` word at the end of example description such as:
+
+```md
+- Define and export an environmental variable that persists across shell restarts (builtin):
+
+`set --universal --export {{variable_name}} {{variable_value}}`
+```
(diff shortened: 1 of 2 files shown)


contributing-guides/style-guide.md
```diff
@@ -94,6 +98,59 @@ Use backticks on the following:
...
+
+- Define and export environmental variable that persist across shell restarts (builtin):
+
+`set --universal --export {{variable_name}} {{variable_value}}`
+
+- Print the version:
+
+`fish --version`
+```
+
```
marchersimon, line 153: Thanks for the effort, but I think that's really too much for a general style guide. If it becomes too long it's hard to find the thing you're searching for.

contributing-guides/style-guide.md
```diff
@@ -66,14 +66,18 @@ Keep the following guidelines in mind when choosing tokens:
...
 - If a particular extension is expected for the file, append it.
   For example, `unrar x {{compressed.rar}}`.
 - In case a generic extension is needed, use `{{.ext}}`, but **only** if an extension is required.
   For instance, in `find.md`'s example "Find files by extension" (`find {{root_path}} -name '{{*.ext}}'`)
   using `{{*.ext}}` explains the command without being unnecessarily specific;
   while in `wc -l {{file}}` using `{{file}}` (without extension) is sufficient.
+  
+### Options
+
+- Prefer long options instead of short ones whenever is possible.
```
marchersimon, line 80: I think they're mostly referred to as "flags".
```suggestion
### Flags

- Prefer long flags instead of short ones whenever it is possible. E.g. use `--command`, instead of `-c`. This makes it easier for the user to remember and understand them.
```

EmilySeville7cfg: As I know flags are options without value. That's why I used `options` term.

CleanMachine1: Options is definitely not a good fit, if I was new, I would interpret that as being the token syntax at a glance

contributing-guides/style-guide.md
```diff
@@ -94,6 +98,59 @@ Use backticks on the following:
...
+
+- Start an interactive shell session:
+- Start an interactive shell session without loading startup configs:
+- Execute a command:
+- Execute a script:
+- Check a script for syntax errors:
+- _any other examples_
+- Print the version:
+
+If you want to include shell builtin sample please mark it with `(builtin)` word. For example:
```
CleanMachine1, line 113: "What in the world is a shell builtin sample" Me + most others (im guessing)

Even from reading the "For example:" I still have no idea what you are referring to.

contributing-guides/style-guide.md
CleanMachine1: I'd reword this to:

```suggestion
### Flags

- Always use long flags when applicable, for example instead of `bash -c`, use `bash --command`, so the reader can understand the meaning of a flag.
```

EmilySeville7cfg:
I am going resolve all suggestions tomorrow. :)

contributing-guides/style-guide.md
EmilySeville7cfg: Now this sentence sounds like:

> If you want to include shell builtin sample please place `(builtin)` word at the end of example description such as:

contributing-guides/style-guide.md
```diff
@@ -66,16 +67,21 @@ Keep the following guidelines in mind when choosing tokens:
...
 - If a particular extension is expected for the file, append it.
   For example, `unrar x {{compressed.rar}}`.
 - In case a generic extension is needed, use `{{.ext}}`, but **only** if an extension is required.
   For instance, in `find.md`'s example "Find files by extension" (`find {{root_path}} -name '{{*.ext}}'`)
   using `{{*.ext}}` explains the command without being unnecessarily specific;
   while in `wc -l {{file}}` using `{{file}}` (without extension) is sufficient.
+  
+### Flags
+
+- Always use long flags when applicable, for example instead of `fish -c`, use `fish --command`, so the reader can understand the meaning of a flag.
```
sbrl, line 81: This is a convention yes, but it doesn't always hold true. It's actually a case-by case basis. For example, on pages like `ls` or `git commit` which are extremely frequently used we do not actually recommend `--long-arguments`.

contributing-guides/style-guide.md
```diff
@@ -94,6 +100,28 @@ Use backticks on the following:
 - Extensions, ex. `.dll`.
 - Commands, ex. `ls`.
 
+### Shell-specific pages
```
sbrl, line 103: Hrm, I feel like this is overly-specific and isn't a great fit for the style guide. Most pages do not document a shell, so aren't affected by this.

EmilySeville7cfg:
Please reopen pr

contributing-guides/style-guide.md
```diff
@@ -94,6 +100,28 @@ Use backticks on the following:
...
+
+If a page describes some shell (Bash for instance) then when it's possible try follow this structure:
```
pixelcmtd, line 105:
```suggestion
If a page describes a shell, when it's possible try to follow this structure:
```

pixelcmtd (commented):
I'd still love to drastically shorten the new section, might try that later.

contributing-guides/style-guide.md
git-em: Maybe this part could get its own file. There is not that much shells, but there's a lot of builtins.
Even though it's specific, might be good to have this things standardized.

contributing-guides/style-guide.md
```diff
@@ -66,16 +67,23 @@ Keep the following guidelines in mind when choosing tokens:
...
 - If a particular extension is expected for the file, append it.
   For example, `unrar x {{compressed.rar}}`.
 - In case a generic extension is needed, use `{{.ext}}`, but **only** if an extension is required.
   For instance, in `find.md`'s example "Find files by extension" (`find {{root_path}} -name '{{*.ext}}'`)
   using `{{*.ext}}` explains the command without being unnecessarily specific;
   while in `wc -l {{file}}` using `{{file}}` (without extension) is sufficient.
+  
+### Flags
+
+- Always use long flags when available for pages in `linux/` directory. For example instead of `fish -c`, use `fish --command`, so the reader can understand the meaning of a flag.
```
git-em, line 81:
```suggestion
- Use long flags when available for pages in `linux/` directory. For example instead of `fish -c`, use `fish --command`, so the meaning of the flag can be understood.
```

contributing-guides/style-guide.md
EmilySeville7cfg: Please view [this](https://github.com/Console-Utils/tldr/blob/shell-guide/contributing-guides/tool-oriented-guides/shells.md) file.

EmilySeville7cfg:
⚠️ This PR is going to be splitted into several PRs to be faster accepted.

contributing-guides/style-guide.md
```diff
@@ -6,27 +6,27 @@ This page lists specific formatting instructions for `tldr` pages.
...
+# <command-name>
 
 > Short, snappy description.
 > Preferably one line; two are acceptable if necessary.
 > More information: <https://example.com>.
 
 - Example description:
 
-`command -opt1 -opt2 -arg1 {{arg_value}}`
+`<command-name> -opt1 -opt2 -arg1 {{arg_value}}`
```
sbrl, line 18: This is probably best left as-is.

contributing-guides/style-guide.md
```diff
@@ -6,27 +6,27 @@ This page lists specific formatting instructions for `tldr` pages.
...
 
 - Example description:
 
-`command -opt1 -opt2`
+`<command-name> -opt1 -opt2`
```
sbrl, line 22: Same here.

contributing-guides/tool-oriented-guides/package-manager.md
```diff
@@ -0,0 +1,19 @@
+## Package manager page format
+
+If the page describes the package manager, when it's possible try to follow this structure:
+
+```
+- Install the specified packages:
+
+`<command-name>`
```
sbrl, line 8: This `<>` syntax is not a good idea. It confuses things.

sbrl (commented):
Some comments, but this is an *enormous* PR

PR #7472, opened 2021-11-22
labeler.yml: use alias tag v0

navarroaxel:
Automatically use new versions from our labeler actions using the same mechanism that use `actions/checkout@v2` or `actions/setup-node@v2`.

To receive new versions automatically the tag `v0` should be updated when a new tag, e.g. `v0.1.1`, is created.


CleanMachine1 (approved):

mfrw (approved):
LGTM

PR #7493, opened 2021-11-29
git-filter-repo: add page

peterbabic:

**Version of the command being documented (if known):** v2.34.0

### DIFF
--- a/pages/common/git-filter-repo.md
+++ b/pages/common/git-filter-repo.md
@@ -1,12 +1,12 @@
 # git filter-repo
 
-> A versatile tool for rewriting git history.
+> A versatile tool for rewriting Git history.
 > See also: `bfg`.
-> More information: <https://manned.org/git-filter-repo>.
+> More information: <https://github.com/newren/git-filter-repo>.
 
 - Replace sensitive string in all files:
 
-`git filter-repo --replace-text {{<(echo 'my_password==>xxxxxxxx')}}`
+`git filter-repo --replace-text <(echo '{{find}}==>{{replacement}}')`
 
 - Extract a single folder, keeping history:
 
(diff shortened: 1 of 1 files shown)


mfrw (approved):
LGTM.
Thank you for your contribution

pages/common/git-filter-repo.md
```diff
@@ -0,0 +1,21 @@
+# git filter-repo
+
+> A versatile tool for rewriting Git history.
+> See also: `bfg`.
+> More information: <https://github.com/newren/git-filter-repo>.
+
+- Replace sensitive string in all files:
```
mfrw, line 7: nitpick maybe:
```suggestion
- Replace a sensitive string in all files:
```

peterbabic: Since the replacement mechanism reads key-value pairs from the designated file, I would say that a plural form would be far more suitable, as it is even encouraged by design:

```suggestion
- Replace sensitive strings in all files:
```

258204: Would you please add an additional example describing the syntax for multiple key value pairs?

peterbabic: Multiple key-value pairs have be inside a file, can you please point me to some example in `tldr` that describes contents of an input file needed to run a command?

This would enable to add other important `git-filter-repo` commands here, for instance ones that use `.mailcap` file to replace author/email in commit history.

sbrl: Hmmm, describing the format of the file if it's specific would belong in the documentation I think. So to this end it's perhaps best to stick to a single string here in this example instead of trying to expand to multiple strings?

258204: Thanks sbrl.

Unless if there is an easy way to specify multiple strings without specifying an input file, I would recommend keeping 
`- Replace sensitive string in all files:`

navarroaxel (approved):

PR #7499, opened 2021-11-30
makepkg: refresh

CleanMachine1:

**Version of the command being documented (if known):** 6.0.1

### DIFF
--- a/pages/linux/makepkg.md
+++ b/pages/linux/makepkg.md
@@ -1,10 +1,10 @@
 # makepkg
 
-> Creates a package installable with the `pacman` package manager.
-> Runs the commands from a `PKGBUILD` file to build the package.
+> Creates a package which can be used with `pacman`.
+> Uses the `PKGBUILD` file in the current working directory.
 > More information: <https://man.archlinux.org/man/makepkg.8>.
 
-- Make a package (run in the same directory as a `PKGBUILD`):
+- Make a package:
 
 `makepkg`
 
@@ -12,11 +12,11 @@
 
 `makepkg --syncdeps`
 
-- Same as above, but install the package with `pacman` when done:
+- Make a package, install its dependencies then install it to the system:
 
 `makepkg --syncdeps --install`
 
-- Make a package, but skip source checksums:
+- Make a package, but skip checking the source's hashes:
 
 `makepkg --skipchecksums`
 
--- a/pages/linux/makepkg.md
+++ b/pages/linux/makepkg.md
@@ -1,7 +1,7 @@
 # makepkg
 
 > Creates a package which can be used with `pacman`.
-> Uses the `PKGBUILD` file in the current working directory.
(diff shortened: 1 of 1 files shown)


258204 (approved):
Can't find anything I would change.

pages/linux/makepkg.md
```diff
@@ -1,22 +1,22 @@
 # makepkg
 
-> Creates a package installable with the `pacman` package manager.
-> Runs the commands from a `PKGBUILD` file to build the package.
+> Creates a package which can be used with `pacman`.
+> Uses the `PKGBUILD` file in the current working directory.
```
mfrw, line 4: Should we mention `current working dir` here ?
It does take an option `-p` where we can supply any `PKGBUILD` file.
How about:
```suggestion
> Uses the `PKGBUILD` file in the current working directory by default.
```

mfrw (commented):
Should we specify that `PGKBUILD` file is the default action ?

pages/linux/makepkg.md
```diff
@@ -1,22 +1,22 @@
 # makepkg
 
-> Creates a package installable with the `pacman` package manager.
-> Runs the commands from a `PKGBUILD` file to build the package.
+> Creates a package which can be used with `pacman`.
+> Uses the `PKGBUILD` file in the current working directory by default.
```
navarroaxel, line 4: I think this sounds better
```suggestion
> Create a package which can be used with `pacman`.
> Use the `PKGBUILD` file in the current working directory by default.
```

CleanMachine1: I agree with the first line however the second line change seems out of place.

mfrw: Yep, reading it on a new day, it does seem odd :(
We can maybe omit the `by default`, given we start with `Use`.

sbrl: I also agree with @CleanMachine1.

mfrw: Maybe this sounds better ? (thoughts)
```suggestion
> Creates a package which can be used with `pacman`.
> Uses the `PKGBUILD` file in the current working directory to build.
```

CleanMachine1: I think axel's edited suggestion is sufficient

pages/linux/makepkg.md
```diff
@@ -1,22 +1,22 @@
 # makepkg
 
-> Creates a package installable with the `pacman` package manager.
-> Runs the commands from a `PKGBUILD` file to build the package.
+> Create a package which can be used with `pacman`.
+> Uses the `PKGBUILD` file in the current working directory by default.
```
mfrw, line 4: I think this reads fine to me. Thanks @CleanMachine1

mfrw (approved):
LGTM

navarroaxel (approved):

PR #7509, opened 2021-12-05
git-terminal: update instructions to amend author

SethFalco:
Updates the docs for updating an old email address. This way is _much_ better than before, and much simpler to perform.

It might be worth revising the instructions, I wanted them to be verbose so people could maybe take the knowledge away with them, but at the same time explaining things rather than providing flat instructions might be daunting. 🤔 

For example, the fact that an interactive rebase works by rewinding the commit history and replaying commits might be redundant information.

Example of me performing the instructions for a single commit:
Peek 2021-12-05 19-11

### DIFF
--- a/contributing-guides/git-terminal.md
+++ b/contributing-guides/git-terminal.md
@@ -68,24 +68,25 @@ git push --force-with-lease
 
 # Changing the email of any commit(s)
 
-Let's take this commit history as an example:
-
-| Commit Hash | Author Email
-|---|---
-| A | wrong@example.org
-| B | correct@example.org
-| C | correct@example.org
-| D | wrong@example.org
-| E | correct@example.org
-| F (HEAD) | correct@example.org
-
-To change the email of commits A and D, run
+1. Perform an [interactive rebase](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt--i) specifying the reference of the earliest commit to modify as the argument. For example, if the earlist commit with the wrong email address was 6 commits ago, you can specify the commit hash or just `HEAD~6`.
 
 ```bash
-git reset A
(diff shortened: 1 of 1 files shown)


mfrw (approved):
Thank you @SethFalco

mfrw:
> Updates the docs for updating an old email address. This way is _much_ better than before, and much simpler to perform.
> 
> It might be worth revising the instructions, I wanted them to be verbose so people could maybe take the knowledge away with them, but at the same time explaining things rather than providing flat instructions might be daunting. 🤔
> 
> For example, the fact that an interactive rebase works by rewinding the commit history and replaying commits might be redundant information.
> 
> Example of me performing the instructions for a single commit: Peek 2021-12-05 19-11

Since you've recorded this example, I think, we should have it also linked IMO. (Thoughts ?)

SethFalco:
> Since you've recorded this example, I think, we should have it also linked IMO. (Thoughts ?)

If the others think a recording should be added, I'd be happy to use Asciinema to make a visual that's smaller and more appealing. I'm partial on visual instructions myself, but I'm certain that's an unpopular opinion.

SethFalco:
https://asciinema.org/a/fFMZzQOgJyfUf8HTnXyRj0v02

[asciicast](https://asciinema.org/a/fFMZzQOgJyfUf8HTnXyRj0v02)

Seem alright?

mfrw:
This looks awesome. Thank you for the effort :)

sbrl (approved):
Given that this file is dedicated to git tutorials, having a visual option can be really helpful to enhance understanding for some readers. We can also split each git tutorial/guide into it's own file should they become too long :-)

Thanks, @SethFalco!

Issue #7503, opened 2021-12-01
Group commands by shells


EmilySeville7cfg:
There are now some tldr pages related to shell builtins. I suggest group such pages by shell not to be limited by naming conflicts in the same folder. Grouping only by OS is not always enough.

Maybe the following strategy would be enough:

- If command is __not OS specific and is external__ then it goes to `pages/general/general` folder.
- Else if command is __OS specific and is external__ then it goes to `pages/<os_name>/general` folder.
- Else if command is __not OS specific and is shell specific__ then it goes to `pages/general/<shell_name>` folder.
- Else it goes to `pages/<os_name>/<shell_name>` folder.

CleanMachine1:
This makes the structure of tldr a lot more complex.

I can't tell if I like it but this will break a **lot** of clients.

EmilySeville7cfg:
Maybe it's better to have less clients but more robust. ;) Having too many clients is not very convenient - not all of them are up to date for now and moreover not all of them have the same features.

The way we can solve this problem - list tldr clients we want to support officially and try change them so they will support suggested directory structure (maybe not by default but via some flag). I can make PRs for console clients because I don't have much experience in writing GUI apps. ;)

sbrl:
I'm not sure I follow your proposed solution, @EmilySeville7cfg.

While I agree that shell builtins are a sticking point, note that part of what makes tldr-pages really useful and easy to contribute to is it's simplicity - I'm wary of the additional complexity this would introduce.

Perhaps documenting such shell builtins is best left for a cheatsheet-style project rather than tldr-pages itself?

EmilySeville7cfg:
Another solution: describe short list of builtins what can be described here, and accept only them to reduce naming conflicts with external commands similarly named.  For example we accept only POSIX-compatible shell builtins for:

- setting/unsetting variables
- if conditions
- for/while/do-while/until loops
- case patterns

sbrl:
Something like that makes sense, yeah. It's interesting you've brought up POSIX compatibility - this could be a good standard to decide on what can be included in a page about a shell builtin that is present in multiple shells.

Of course, I'm some shells have their own unique builtins that are not found in other shells, and these could be treated as regular commands with a note what shell they are from.

marchersimon:
I'm closing this now, since the original proposal would need a new directory structure.

EmilySeville7cfg:
Okay ;)

Issue #6621, opened 2021-10-01
Let's document: PowerShell


reinhart1010:
Sorry for the bad timing (it's the first day of Hacktoberfest 2021 so yeah), but I think documenting tldr pages for the ~~Windows~~ **cross-platform** PowerShell could be a great idea.

No, not the tldr page for [`windows/powershell`](https://github.com/tldr-pages/tldr/blob/main/pages/windows/powershell.md), but [all the PowerShell cmdlets](https://github.com/PowerShell/PowerShell/tree/658837323599ab1c7a81fe66fcd43f7420e4402b/src/Microsoft.PowerShell.Commands.Utility/commands/utility) like `Get-Help`, `Invoke-WebRequest`, `Get-StartApps` and so on.

These cmdlets uses a combination of `kebab-case` and `(Upper) CamelCase`, while according to the tldr-page guidelines all commands should be lowercase. So this could be a concern for contributors and clients.

Furthermore, some PowerShell cmdlets are only available for Windows, making the page structure slightly complicated. For example, `curl` often redirects to `Invoke-WebRequest` when used in Windows where the actual `curl` is not installed. Note that both `curl` and `Invoke-WebRequest` are completely different commands with different parameters and options.

reinhart1010:
Okay, I've seen that the `Get-Content` PowerShell command is also documented in [`windows/get-content`](https://github.com/tldr-pages/tldr/blob/main/pages/windows/get-content.md), and the command also works in my Linux installation of PowerShell, too!

![image](https://user-images.githubusercontent.com/17312341/135568971-c41047d2-b0a8-4e34-9854-e3b46e1dd72d.png)

sbrl:
Hey there! Thanks for opening this issue. Hmm, I think we want a few opinions on this.

It would be easy if powershell was windows-specific (though I can imagine Windows accounts for the majority of powershell usage  anyway), but as you say powershell is cross-platform.

Perhaps documenting in `common` would work - so long as any commands documented are clearly labelled as powershell commands rather then ordinary commands that work in a regular shell.

Let's wait for some more opinions from people on this before making a decision.

marchersimon:
We probably won't have any name collisions, but even though it's available on all platforms it will be used on Windows most of the time. Not directly comparable, but still: we also don't put Linux commands under common because it's available through WSL.

reinhart1010:
I was thinking to simply put all PowerShell documentation into `powershell/`, and some Windows-only commands will have this notice:

[...]

PR #7512, opened 2021-12-06
nvm.fish: add page

reinhart1010:

**Version of the command being documented (if known):** 2.2.6

This is related to our discussion on #7480. However, this might break some tldr clients as this is the first known file to have more than one dot characters (`.`).

### DIFF
--- a/pages/common/nvm.fish.md
+++ b/pages/common/nvm.fish.md
@@ -1,7 +1,7 @@
-# nvm.fish
+# nvm
 
 > Install, uninstall or switch between Node.js versions under the `fish` shell.
-> Supports version numbers like "0.12" or "v4.2", and labels like "stable", "system", etc.
+> Supports version numbers like "12.8" or "v16.13.1", and labels like "stable", "system", etc.
 > More information: <https://github.com/jorgebucaran/nvm.fish>.
 
 - Install a specific version of Node.js:
--- a/pages/common/nvm.fish.md
+++ b/pages/common/nvm.fish.md
@@ -1,6 +1,6 @@
 # nvm
 
-> Install, uninstall or switch between Node.js versions under the `fish` shell.
+> Install, uninstall, or switch between Node.js versions under the `fish` shell.
 > Supports version numbers like "12.8" or "v16.13.1", and labels like "stable", "system", etc.
 > More information: <https://github.com/jorgebucaran/nvm.fish>.
 
--- a/pages/common/nvm.fish.md
+++ b/pages/common/nvm.fish.md
@@ -1,0 +1,25 @@
+# nvm.fish
+
+> Install, uninstall or switch between Node.js versions under the `fish` shell.
+> Supports version numbers like "0.12" or "v4.2", and labels like "stable", "system", etc.
+> More information: <https://github.com/jorgebucaran/nvm.fish>.
+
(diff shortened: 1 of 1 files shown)


pages/common/nvm.fish.md
```diff
@@ -0,0 +1,25 @@
+# nvm.fish
```
nicokosi, line 1: If I understand correctly, the binary is `nvm` so I think there is a mismatch here.
Moreover, the CLI interface seems similar to https://github.com/nvm-sh/nvm so I am not sure this whole page is needed. 💭

reinhart1010: There are several differences between the original `nvm` with this version. In short, currently the commands between both versions aren't compatible with each other:

1. No `nvm run` or `nvm exec` (though I can just create a PR for that 😀)
2. The default Node.js version is set through `set` instead of a subcommand (again, even though I can just PR about it 😅)

Maybe I should close this PR by sending a couple to https://github.com/jorgebucaran/nvm.fish...

nicokosi: Thanks for the detailed information. :+1: 
I understand that command are different. :ok_hand: 
But I don't understand what to do with the mismatch between the binary name (`nvm`) and the page name (we would need to run `tldr nvm.fish` to find it, if I understand correctly). :thinking:

reinhart1010: There's a bunch of related discussions related to this:

1. #7480 - The one I've referenced here
2. #7503 - Plans to separate commands by shells, if so I can move this into something like `common/fish/nvm`
3. #6621 - PowerShell, related to #7503

sbrl: The title of a page needs to exactly match the command name being executed. This is not the case for the filename.

 - **Filename:** `nvm.fish.md`
 - **Page title:** `nvm`
 - **Command being executed:** `nvm`

Ideally the page title and the filename will match, but this is unfortunately not always possible.

pages/common/nvm.fish.md
```diff
@@ -0,0 +1,25 @@
+# nvm
+
+> Install, uninstall or switch between Node.js versions under the `fish` shell.
```
CleanMachine1, line 3:
```suggestion
> Install, uninstall, or switch between Node.js versions under the `fish` shell.
```

pages/common/nvm.fish.md
```diff
@@ -0,0 +1,25 @@
...
+
+`nvm install {{node_version}}`
+
+- Use a specific version of Node.js in the current shell:
+
+`nvm use {{node_version}}`
+
+- Set the default Node.js version:
+
+`set nvm_default_version {{node_version}}`
```
sbrl, line 17: Is this a typo? Should this be prefixed by `nvm` at all, or is this essentially updating an environment variable that `nvm` (fish) reads?

reinhart1010: Yes it uses the Fish' builtin `set` command to set the default version.

navarroaxel (approved):

sbrl (approved):
Thanks, @reinhart1010!

pages/common/nvm.fish.md
sbrl: For completeness, the command in question where it's also not possible is `rename`.

PR #7514, opened 2021-12-06
CLIENT-SPECIFICATION: add macos recommendation

### LINKED ISSUE #7076: Client spec: `osx` and `macos`
Hi everyone!

Over at [tealdeer](https://github.com/dbrgn/tealdeer) we wanted to allow users to specify `macos` as a platform for filtering pages (see https://github.com/dbrgn/tealdeer/pull/108#discussion_r730431290 and https://github.com/dbrgn/tealdeer/pull/215).

It would of course be great if we could get confirmation from you guys that you won't add a "macos" platform any time soon - or a hard "no" if you _are_ planning to do so. Maybe this alias could even manifest itself in the client spec as a recommendation or option. After all, [the operating system is not called "osx" anymore](https://en.wikipedia.org/wiki/Macintosh_operating_systems#Mac_OS_X_/_OS_X_/_macOS).

Is this alias something the client spec authors would approve or are we shooting ourselves in the foot here? I would love to hear your thoughts on this! 

Slightly related: #5013

MasterOdin:

Closes #7076. This adds to the specification that it's recommended to treat `macos` as an alias for `osx`. At some point it might be good to migrate to `macos` for folder structure, this at least is a step in that direction.

### DIFF
--- a/CLIENT-SPECIFICATION.md
+++ b/CLIENT-SPECIFICATION.md
@@ -226,6 +226,9 @@ the form `vX.Y`) should be done immediately AFTER merging the version bump, as
 the commit hash changes when merging with squash or rebase.
 -->
 
+ - Unreleased
+   -  Add recommendation to support `macos` alias for `osx` ([#7514](https://github.com/tldr-pages/tldr/pull/7514))
+
  - [v1.5, March 17th 2021](https://github.com/tldr-pages/tldr/blob/v1.5/CLIENT-SPECIFICATION.md) ([#5428](https://github.com/tldr-pages/tldr/pull/5428))
    - Add requirement for converting command names to lowercase before running the page resolution algorithm.
    - Use HTTPS for archive links.
--- a/CLIENT-SPECIFICATION.md
+++ b/CLIENT-SPECIFICATION.md
@@ -85,7 +85,9 @@ The main version of every page is stored inside (but not directly) the `pages` d
    - `osx/`
    - ...etc.
 
-Additional platforms MAY be added in the future. Clients MAY NOT support new platforms (though such support is RECOMMENDED), but MUST NOT break if additional platforms are added.
+It is RECOMMENDED that client supports `macos` as an alias for `osx`.
+
+Additional platforms MAY be added in the future. Clients MAY NOT support new platforms (though such support is RECOMMENDED), but MUST NOT break if additional platforms are added. 
 
 The pages themselves reside inside the appropriate platform folder, with the extension `.md`. Here are some example mappings:
 
--- a/CLIENT-SPECIFICATION.md
+++ b/CLIENT-SPECIFICATION.md
@@ -85,7 +85,7 @@ The main version of every page is stored inside (but not directly) the `pages` d
    - `osx/`
    - ...etc.
 
-It is RECOMMENDED that client supports `macos` as an alias for `osx`.
+It is RECOMMENDED that clients support `macos` as an alias for `osx`.
 
 Additional platforms MAY be added in the future. Clients MAY NOT support new platforms (though such support is RECOMMENDED), but MUST NOT break if additional platforms are added. 
 
--- a/CLIENT-SPECIFICATION.md
+++ b/CLIENT-SPECIFICATION.md
@@ -87,7 +87,7 @@ The main version of every page is stored inside (but not directly) the `pages` d
 
 It is RECOMMENDED that clients support `macos` as an alias for `osx`.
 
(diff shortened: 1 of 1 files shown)


sbrl (commented):
Sounds good to me! This will be a long process to rename `osx` to `macos` I'm sure, but it's great to take the first step towards that here.

Could we update the client specification changelog (at the bottom of the file) for this too please?

MasterOdin:
@sbrl I've updated the changelog at the bottom, but did you want me to set the version to 1.6 throughout, or will that happen in its own PR (like other version releases)?

CleanMachine1 (approved):

CLIENT-SPECIFICATION.md
```diff
@@ -85,7 +85,9 @@ The main version of every page is stored inside (but not directly) the `pages` d
    - `osx/`
    - ...etc.
 
-Additional platforms MAY be added in the future. Clients MAY NOT support new platforms (though such support is RECOMMENDED), but MUST NOT break if additional platforms are added.
+It is RECOMMENDED that client supports `macos` as an alias for `osx`.
```
dbrgn, line 88: Wording: "client supports" should be "clients support".

CLIENT-SPECIFICATION.md
```diff
@@ -85,7 +85,9 @@ The main version of every page is stored inside (but not directly) the `pages` d
...
+
+Additional platforms MAY be added in the future. Clients MAY NOT support new platforms (though such support is RECOMMENDED), but MUST NOT break if additional platforms are added.
```
dbrgn, line 90: This is unrelated, but I noticed that the language here is wrong. Instead of `Clients MAY NOT support new platforms` it should be `Clients MAY not support new platforms`. Otherwise it would mean that supporting new platforms is forbidden.

Maybe it could be fixed in this PR as well (I thought about opening a separate PR, but that would conflict with yours).

MasterOdin: I think you want it changed to `SHOULD NOT`? Not sure why lowercasing the `not` here would change the meaning of the sentence.

However, this whole sentence can probably be reworked as I'm not sure why there's also the "though such support is RECOMMENDED" when the behavior is not recommended? What's the sentiment that's being expressed in this sentence?

CLIENT-SPECIFICATION.md
MasterOdin: 702dae2e7415f10487fbf40fc27569c9ce3f8dcf

CLIENT-SPECIFICATION.md
dbrgn: I think it's the other way around? The point is that tldr may add new platform directories (e.g. `android`), and it's possible that a certain client doesn't support it (yet, it's still compliant).

If we use `MAY NOT` or `SHOULD NOT`, then the client is forbidden from adding support for this platform, right? That doesn't make sense at all.

MasterOdin: So should the wording here be `Clients MAY support new platforms (though such support is RECOMMENDED), but MUST NOT break if additional platforms are added to tldr-pages` (drop the `not` altogether, not just lowercase it)?

sbrl: What about `While clients may not support new platforms automatically (though such support is RECOMMENDED), they MUST NOT break if additional platforms are added to tldr-pages.`?

sbrl:
Regarding the version number bump, it's probably worth doing here - though we also have #7561 that should make it into v1.6 too.

I'll write a release when both PRs are merged.

CLIENT-SPECIFICATION.md
MasterOdin: Emphasis added on a proposed tweak:
> While clients __do not need to__ support new platforms automatically (though such support is RECOMMENDED), they MUST NOT break if additional platforms are added to tldr-pages.

using `may not support` implies to me that they should not have this support at all and contradicts the recommendation. My tweak here I think implies what we want?

sbrl: Sounds good, @MasterOdin.

MasterOdin: Done

CleanMachine1 (approved):

CleanMachine1:
whoops, reviewed twice!

MasterOdin:
Friendly bump @sbrl!

blueskyson (approved):

navarroaxel (approved):

PR #7516, opened 2021-12-07
groff: add --RAW-CONTROL-CHARS to pager example

hexagonrecursion:
Without --RAW-CONTROL-CHARS less in the example would mangle the display by converting the control characters into the caret notation.

### DIFF
--- a/pages/common/groff.md
+++ b/pages/common/groff.md
@@ -9,7 +9,7 @@
 
 - Render a man page using the ASCII output device, and display it using a pager:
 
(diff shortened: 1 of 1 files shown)


sbrl (approved):
Fair enough. For reference, `--RAW-CONTROL-CHARACTERS` makes only ansi escape sequences for colours allowed, and everything else is escaped.

CleanMachine1 (approved):

navarroaxel (approved):

PR #7517, opened 2021-12-07
invoke-item: add page

reinhart1010:

**Version of the command being documented (if known):** PowerShell 7.2


reinhart1010:
Quick question: Do you prefer `insode a directory` or `within a directory`?

pages/windows/invoke-item.md
```diff
@@ -0,0 +1,29 @@
...
+
+- Open a file in its default program:
+
+`Invoke-Item -Path "{{path/to/file}}"`
+
+- Open all files inside a directory:
+
+`Invoke-Item -Path "{{path/to/directory}}\*"`
+
+- Open all PNG files inside a directory:
```
SethFalco, line 15:
```suggestion
- Open all PNGs inside a directory:
```

pages/windows/invoke-item.md
```diff
@@ -0,0 +1,29 @@
...
+
+`Invoke-Item -Path "{{path/to/directory}}\*.png"`
```
SethFalco, line 17:
```suggestion
`Invoke-Item -Path "{{path\to\directory\*.png}}"`
```

The entire argument should be wrapped in brackets.

Question for other maintainers:
Since this is Windows-specific, should we be using forward slashes or backslashes? 🤔 At the very least, we probably don't want to mix between both of them in the same argument.

pages/windows/invoke-item.md
```diff
@@ -0,0 +1,29 @@
...
+
+- Open all files inside a directory which contain a specific keyword:
+
+`Invoke-Item -Path "{{path/to/directory}}\* -Include "*{{keyword}}*"`
```
SethFalco, line 21:
```suggestion
`Invoke-Item -Path "{{path\to\directory\*}}" -Include "{{*keyword*}}"`
```

Forgot the `"` at the end of the value for the `-Path` argument.

pages/windows/invoke-item.md
```diff
@@ -0,0 +1,29 @@
...
+
+- Open all files inside a directory except those containing specific keyword:
+
+`Invoke-Item -Path "{{path/to/directory}}\* -Exclude "*{{keyword}}*"`
```
SethFalco, line 25:
```suggestion
`Invoke-Item -Path "{{path\to\directory\*}}" -Exclude "{{*keyword*}}"`
```

pages/windows/invoke-item.md
```diff
@@ -0,0 +1,29 @@
...
+
+- Perform a dry run to determine which files will be opened inside a directory through `Invoke-Item`:
+
+`Invoke-Item -Path "{{path/to/directory}}\*" -WhatIf`
```
SethFalco, line 29:
```suggestion
`Invoke-Item -Path "{{path\to\directory\*}}" -WhatIf`
```

SethFalco (changes requested):

SethFalco:
> Quick question: Do you prefer `inside a directory` or `within a directory`?

I think they should mean the same in practice, but as a reader "within a directory" sounds better to me. \\(^-^)/
Other maintainers may have a different opinion.

pages/windows/invoke-item.md
reinhart1010: In most cases, Windows will automatically convert these slashes into backslashes (either in Command Prompt or PowerShell) so `C:/Users/User/Documents` is still valid, too.

In most DOS and non-PowerShell commands, command parameters (e.g. `/d`) do use forward slashes but only when the character is inserted at the begining of the sentence. For the case of root directory, users may end up to use `C:/` (e.g. `cd C:/`) instead of `/` (e.g. `cd /`) which should be okay.

navarroaxel: @SethFalco the `/` is full compatible with Windows, and the `/` is used in the common platform. To keep the same format `/` should be used in the Windows platform.
```suggestion
`Invoke-Item -Path {{path/to/directory/*.png}}`
```
Also, we don't use `"` for paths.

sbrl: Ref quoting the path: if you want the glob to work IIRC you need to remove the quotes, yeah

pages/windows/invoke-item.md
sbrl: Same here @SethFalco - the quotes around `path/to/directory/*` have to go to make the glob work as far as I'm aware.

pages/windows/invoke-item.md
sbrl: And here
```suggestion
`Invoke-Item -Path {{path/to/directory/*}} -WhatIf`
```

sbrl:
> Quick question: Do you prefer `inside a directory` or `within a directory`?

Both are fine, but I'd prefer `inside a directory` personally. It really depends on context.

pages/windows/invoke-item.md
SethFalco: The point was less about if they should or shouldn't be there, but just if there's an open quote, I'd expect to see a close one.
But I'll keep in mind in future the quotes/globs thing.

pages/windows/invoke-item.md
```diff
@@ -0,0 +1,29 @@
...
+
+`Invoke-Item -Path {{path/to/directory/*}}`
+
+- Open all PNGs inside a directory:
+
+`Invoke-Item -Path {{path/to/directory/*.png}}`
+
+- Open all files inside a directory containing specific keyword:
+
+`Invoke-Item -Path {{path/to/directory/*}} -Include *{{keyword}}*`
```
SethFalco, line 21:
```suggestion
`Invoke-Item -Path {{path/to/directory/*}} -Include {{*keyword*}}`
```

SethFalco (approved):

pages/windows/invoke-item.md
```diff
@@ -0,0 +1,29 @@
...
+
+- Open all files inside a directory:
+
+`Invoke-Item -Path {{path/to/directory/*}}`
+
+- Open all PNGs inside a directory:
+
+`Invoke-Item -Path {{path/to/directory/*.png}}`
+
+- Open all files inside a directory containing specific keyword:
```
navarroaxel, line 19:
```suggestion
- Open all files inside a directory containing a specific keyword:
```

pages/windows/invoke-item.md
```diff
@@ -0,0 +1,29 @@
...
+
+`Invoke-Item -Path {{path/to/directory/*}} -Include {{*keyword*}}`
+
+- Open all files inside a directory except those containing specific keyword:
```
navarroaxel, line 23:
```suggestion
- Open all files inside a directory except those containing a specific keyword:
```

navarroaxel (approved):

PR #7534, opened 2021-12-12
xargs: replace url

mkcn:
https://man7.org is a more correct (maybe even more trusted) reference than https://pubs.opengroup.org for the man page

### DIFF
--- a/pages/common/xargs.md
+++ b/pages/common/xargs.md
@@ -2,7 +2,7 @@
 
 > Execute a command with piped arguments coming from another command, a file, etc.
 > The input is treated as a single block of text and split into separate pieces on spaces, tabs, newlines and end-of-file.
-> More information: <https://pubs.opengroup.org/onlinepubs/9699919799/utilities/xargs.html>.
+> More information: <https://man7.org/linux/man-pages/man1/xargs.1p.html>.
 
 - Run a command using the input data as arguments:
 


pages/common/xargs.md
```diff
@@ -2,7 +2,7 @@
 
 > Execute a command with piped arguments coming from another command, a file, etc.
 > The input is treated as a single block of text and split into separate pieces on spaces, tabs, newlines and end-of-file.
-> More information: <https://pubs.opengroup.org/onlinepubs/9699919799/utilities/xargs.html>.
+> More information: <https://man7.org/linux/man-pages/man1/xargs.1p.html>.
```
CleanMachine1, line 5:
```suggestion
> More information: <https://manned.org/xargs.1pl>.
```

sbrl: Indeed, as per #5143 when referencing man pages we default to manned.org unless of course the man7.org version is better in some way.

sbrl (changes requested):
Waaaaait a moment. The existing URL is a link to the POSIX specification. This link should not be changed.

mkcn:
@sbrl do you mean to leave "https://pubs.opengroup.org/onlinepubs/9699919799/utilities/xargs.html" ? If yes, should I simply delete the request?

sbrl:
Sorry for not getting back to you here. Yeah, we should leave this as-is. I'm closing this PR for now, but we can always re-open if there's more to discuss. Thanks so much for your contribution anyway :D

PR #7544, opened 2021-12-14
chcon: move from common to linux platform

cyqsimon:
AFAIK, `chcon` is a Linux-specific command.

`man chcon`

```man
─────────────────────────────────────────────────────────────────────────────────────────
CHCON(1)                             User Commands                            CHCON(1)

NAME
       chcon - change file security context

SYNOPSIS
       chcon [OPTION]... CONTEXT FILE...
       chcon [OPTION]... [-u USER] [-r ROLE] [-l RANGE] [-t TYPE] FILE...
       chcon [OPTION]... --reference=RFILE FILE...

DESCRIPTION
       Change the SELinux security context of each FILE to CONTEXT.  With --reference,
       change the security context of each FILE to that of RFILE.
```

Unless MacOS has an equivalence to SELinux and also uses the same `chcon` (which I don't believe is the case), then I stand corrected.


CleanMachine1:
Okay, thanks for this.

I have no idea about this command. From what I see its just Linux. I'll approve when another maintainer voices their opinion.

Overall, good observation!

258204:
BRB bricking my mac http://polarhome.com/service/man/?qf=CHCON&tf=2&of=MacOSX&sf=1

258204:
chcon does appear to be installed by default on macos 10.15.7; however, I'm going to check if the functionality has been altered, if so, I will approve this PR.

258204 (approved):
Although Apple ships with chcon, I compared the results running the examples with macos 10.15.7, and rocky 8.5 (with SELinux). The examples didn't work, implying that either chcon is shipped with macos despite not working with it, or the implementation is different between Linux and BSD. 

The page should be moved.

navarroaxel (approved):

PR #7549, opened 2021-12-17
mktemp: refresh page

EmilySeville7cfg:

### DIFF
--- a/pages/common/mktemp.md
+++ b/pages/common/mktemp.md
@@ -1,16 +1,16 @@
 # mktemp
 
 > Create a temporary file or directory.
-> More information: <https://www.gnu.org/software/autogen/mktemp.html>.
+> More information: <https://manned.org/mktemp.1>.
 
-- Create an empty temporary file and return the absolute path to it:
+- Create an empty temporary file and print the absolute path to it:
 
 `mktemp`
 
-- Create a temporary directory and return the absolute path to it:
+- Create an empty temporary file with a given suffix and print the absolute path to file:
 
-`mktemp -d`
+`mktemp --suffix "{{.ext}}"`
 
-- Create a temporary file with a specified suffix:
+- Create a temporary directory and print the absolute path to it:
 
-`mktemp --suffix "{{.txt}}"`
+`mktemp --directory`
--- a/pages/common/mktemp.md
+++ b/pages/common/mktemp.md
@@ -13,4 +13,4 @@
 
 - Create a temporary directory and print the absolute path to it:
 
-`mktemp --directory`
+`mktemp -d`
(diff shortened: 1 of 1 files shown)


pages/common/mktemp.md
```diff
@@ -1,16 +1,16 @@
...
+- Create an empty temporary file with a given suffix and print the absolute path to file:
 
-`mktemp -d`
+`mktemp --suffix "{{.ext}}"`
 
-- Create a temporary file with a specified suffix:
+- Create a temporary directory and print the absolute path to it:
 
-`mktemp --suffix "{{.txt}}"`
+`mktemp --directory`
```
nicokosi, line 16: I suggest not using long options that are visibly not supported on `macOS` (see `*`):
```suggestion
`mktemp -d`
```

\* : 
```sh
$ echo $OSTYPE
darwin20.0
$ mktemp --directory
mktemp: illegal option -- -
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix
```

EmilySeville7cfg: Are you using GNU Coreutils?

nicokosi: > Are you using [GNU Coreutils mktemp](https://www.gnu.org/software/coreutils/manual/coreutils.html#mktemp-invocation)?

No, I am using BSD's `mktemp` (which is the default one on macOS, I guess).

> I think we should prefer by default some specific set of utilities GNU/BSD or anything else not to mix them in an unpredictable way.

Not sure to understand. :innocent:
But other maintainers will probably have an opinion about it. :thought_balloon:

sbrl: Indeed, we do prefer long arguments where possible. However, in this instance they are not portable - so we should use the short ones which do seem to be portable.

EmilySeville7cfg: What to prefer:

- `(non-portable long option: --directory)` in command description as now
- `mktemp -d|--directory` supposing that all portable options are going first

?

pages/common/mktemp.md
```diff
@@ -1,16 +1,16 @@
 # mktemp
 
 > Create a temporary file or directory.
-> More information: <https://www.gnu.org/software/autogen/mktemp.html>.
+> More information: <https://manned.org/mktemp.1>.
```
navarroaxel, line 4: Use the author's web.
```suggestion
> More information: <https://www.gnu.org/software/autogen/mktemp.html>.
```

navarroaxel (approved):

CleanMachine1 (approved):

git-em:
Would it make sense to add a parenthesis saying that the `--suffix` option is "GNU-only"?

Issue #7528, opened 2021-12-11
Client spec: Clarify behavior of special platforms


niklasmohrin:
Hey everyone, I am back with another client spec topic :sweat_smile: 

In our effort to make [tealdeer](https://github.com/dbrgn/tealdeer) behave closer to the client specification, we came to adding support for `tldr --list --platform all` as described in the row for the `--list` flag in the [Arguments table](https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md#arguments). While the behavior for `--list --platform all` is clear to us, [we don't agree](https://github.com/dbrgn/tealdeer/pull/229#discussion_r762630680) on how the platform `all` should behave when used in other contexts like simple page lookup. Mainly, there are two different approaches that came up so far:

1. The platform `all` behaves like the current platform / as if no platform was specified at all, but for `--list` we keep in mind whether it was actually `all` or not.
2. We accept `--platform all` _only_ for `--list` and error if the special platform is passed for other uses.

(some arguments for either side may be found in the linked comment)

I saw that the platform `all` is in the spec since the beginning (#2706) and there have been other issues about the behavior when several (independent) subsystems are used together (#4290). While I don't necessarily agree with all the details in the latter, I think that the specification should specify behavior for the special platform type and its meaning in all contexts in [the platform section](https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md#platform) and explain the desired behavior. In a more general sense, the Arguments table should at least list all the possible values for each flag in the row of that flag itself - it was rather unintuitive to find a mention of `--platform all` for me at first.

Thank you for your work, I hope we can work something out :)

dbrgn:
Thanks @niklasmohrin for creating this issue! I agree, this should be clarified.

I think that treating `all` like `current` - except for the places where considering all platforms makes sense (which is only `--list`, I think) - would be a very logical and easy to implement approach. If someone would specify `-p all` for another command, then it would simply work as it would without any `-p` argument (so it's essentially being ignored).

sbrl:
Ah, great spot here. I wrote the original spec, so this is probably my fault as I probably didn't think of / foresee this problem at the time haha :sweat_smile: This probably means that technically this is undefined behaviour in the context oft he current specification.

Indeed, we should absolutely clarify this. Both of those solutions would work - which one would you prefer? I feel like option 2 would be more semantically correct here since `all` doesn't make sense in the context of the page resolution algorithm?

[...]

PR #7561, opened 2021-12-19
CLIENT-SPECIFICATION: drop the special "all" platform

dbrgn:
As discussed in #7528 and [here](https://github.com/tldr-pages/tldr/pull/7561#issuecomment-997473651):

Drop the special `all` platform.

### DIFF
--- a/CLIENT-SPECIFICATION.md
+++ b/CLIENT-SPECIFICATION.md
@@ -38,7 +38,7 @@ Option             | Required?   | Meaning
 `-v`, `--version`  | Yes         | Shows the current version of the client, and the version of this specification that it implements.
 `-p`, `--platform` | Yes         | Specifies the platform to be used to perform the action (either listing or searching) as an argument. If this option is specified, the selected platform MUST be checked first instead of the current platform as described below.
 `-u`, `--update`   | Conditional | Updates the offline cache of pages. MUST be implemented if cache is supported.
-`-l`, `--list`     | No          | Lists all the pages in the current platform to the standard output. If the special platform `all` is specified a list of all pages in all platforms MUST be displayed.
+`-l`, `--list`     | No          | Lists all the pages in the current platform to the standard output.
 `-L`, `--language` | No          | Specifies the preferred language for the page returned. Overrides other language detection mechanisms. See the [language section](#language) for more information.
 
 Clients MAY choose to only implement the short version of an option, ignoring the long form.


dbrgn:
After some further thought: What we don't specify here (neither for `list` nor for `list-all`) is the output format. Right now tealdeer and other clients as well print the commands to stdout, one per line. That's convenient to process with other tools (e.g. for tab completion). (It wasn't always like this, in an older version, tealdeer listed the pages with comma separation.)

We _could_ include this in the spec. However, with `list-all`, it may make sense to include additional information (e.g. the platform for which this command is defined, and a command may be defined multiple times for different platforms). Is it a good idea to recommend implementing a command, but without specifying what the output should look like? And if no, does it even make sense if the tldr client spec decides how a client should format its output?

In the end, the spec should probably constrain itself to the bare minimum and leave the rest to the implementations. If an implementation thinks a certain non-core feature is useful, it should be free to do so.

So: Maybe we should simply drop the `all` platform from the spec and call it a day.

CLIENT-SPECIFICATION.md
```diff
@@ -38,7 +38,8 @@ Option             | Required?   | Meaning
 `-v`, `--version`  | Yes         | Shows the current version of the client, and the version of this specification that it implements.
 `-p`, `--platform` | Yes         | Specifies the platform to be used to perform the action (either listing or searching) as an argument. If this option is specified, the selected platform MUST be checked first instead of the current platform as described below.
 `-u`, `--update`   | Conditional | Updates the offline cache of pages. MUST be implemented if cache is supported.
-`-l`, `--list`     | No          | Lists all the pages in the current platform to the standard output. If the special platform `all` is specified a list of all pages in all platforms MUST be displayed.
+`-l`, `--list`     | No          | Lists all the pages in the current platform to the standard output. It is RECOMMENDED that the pages are sorted alphanumerically.
+      `--list-all` | No          | Lists all the pages in all platforms to the standard output. It is RECOMMENDED that the pages are sorted alphanumerically.
```
sbrl, line 42: Simply specifying sorting alphabetical ordering isn't always though. [As Jeff Attwood explains](https://blog.codinghorror.com/sorting-for-humans-natural-sort-order/), unicode can cause an issue here. We should find some standard on unicode sorting here.

The other thing we should handle here is the case of when a page exists in multiple platforms. Perhaps we should require the platform name to be specified as well here, and explicitly state that a given page name may exist in multiple categories and that's ok and they `MUST NOT` be deduplicated?

dbrgn: Yeah, you're right. As written in my other comment above, after some thought I'd actually tend towards not adding `--list-all` to the standard at all. Same thing with `--list`, we could leave it to the implementation to choose a useful output format. (Or is there any advantage in having a common output format? Should it be parseable?)

sbrl: Yeah, I think that's a good point. Clients can be for many different platforms and purposes - even if we assume a command line, there's a range of possible purposes that such a client may be implemented for. Let's drop the sorting sentence here.

dbrgn: I pushed a new version that simply removed the `all` platform and doesn't specify anything else. Implementations can still implement `--list-all` if they want, but if the output format isn't standardized, then a standardized command name won't bring any benefits either.

sbrl: Right. Let's add support for `--list-all`. What about specifying 1 command name per line on the standard output? e.g.

```markdown
`--list-all` | No          | Lists all the pages in all platforms to the standard output, with 1 page name per line in the format `<platform>/<pagename>`
```

dbrgn: @sbrl did you take a look at [this comment](https://github.com/tldr-pages/tldr/pull/7561#issuecomment-997473651)? I think adding `--list-all` to the spec isn't a good idea, there are a lot of open questions and a standardized `--list-all` wouldn't really give a lot of benefits while potentially limiting improvements to the output format.

My curent suggestion would be dropping the `all` platform and _not_ standardizing `--list-all`. I renamed the PR to reflect the current state of the proposal. If you agree, this discussion can be marked as resolved.

dbrgn:
github-actions: Please re-open.

dbrgn:
@sbrl @waldyrious any opinions on this? (I updated the PR description to properly reflect the current state of the PR.)

waldyrious:
As I wasn't involved in the previous discussions, I'll let @sbrl take this one, unless there are specific aspects or open questions you'd like my thoughts on.

sbrl (approved):

navarroaxel (approved):

PR #7567, opened 2021-12-21
docker: add docker images example

Raizo62:
Hi

In pages.fr/common/docker.md,  add how to display the images already downloaded : 
`docker images`

### DIFF
--- a/pages.fr/common/docker.md
+++ b/pages.fr/common/docker.md
@@ -6,7 +6,7 @@
 
 - Liste tous les conteneurs Docker (en cours d'exécution ou arrêtés) :
 
-`docker ps -a`
+`docker ps --all`
 
 - Démarre un conteneur à partir d'une image, avec un nom personnalisé :
 
--- a/pages.fr/common/docker.md
+++ b/pages.fr/common/docker.md
@@ -24,6 +24,10 @@
 
 `docker pull {{image}}`
 
+- Affiche les images déjà téléchargées :
+
+`docker images`
+
 - Ouvre un shell dans un conteneur déjà en cours d'exécution :
 
 `docker exec -it {{nom_conteneur}} {{sh}}`
--- a/pages.fr/common/docker.md
+++ b/pages.fr/common/docker.md
@@ -4,10 +4,6 @@
 > Certaines commandes comme `docker run` ont leur propre documentation.
 > Plus d'informations : <https://docs.docker.com/engine/reference/commandline/cli/>.
 
-- Liste les conteneurs Docker en cours d'exécution :
-
-`docker ps`
-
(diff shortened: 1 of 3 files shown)


Raizo62:
I believe that this command is usefull.
I must remove a other command ? Which ? the first (`docker ps`) because it is include in the second (`docker ps -a`)

navarroaxel (changes requested):
Hi @Raizo62, we have 2 issues here:
1. For us, the translation pages are just translation from the main English pages. Because it's harder to maintain, and keep in sync if every language adds editions. So, you should send your edit in the main page first.
2. We have an strict rule that all pages can have up to 8 examples.

I think `docker images` is a basic command, and for this page we can merge `ps` and `ps -a` examples to add the `images` example.

sbrl:
We already have a more detailed page for `docker images` here: <https://github.com/tldr-pages/tldr/blob/main/pages/common/docker-images.md>

The page for the top-level `docker` commands is one of those pages that has *far* too much functionality to completely represent in a single page (similar to `git` for example), so we have sub-pages for each subcommand thereof.

Raizo62:
> We already have a more detailed page for docker images here: https://github.com/tldr-pages/tldr/blob/main/pages/common/docker-images.md

Yes, but we look the page "docker-images.md" only if we already know that we must use the parameter "images" to display the list of downloaded images. This is not obvious.

blueskyson (approved):
Look good to me.

pages/common/docker.md
```diff
@@ -4,10 +4,6 @@
 > Some subcommands such as `docker run` have their own usage documentation.
 > More information: <https://docs.docker.com/engine/reference/commandline/cli/>.
 
-- List currently running docker containers:
-
-`docker ps`
-
 - List all docker containers (running and stopped):
 
 `docker ps -a`
```
navarroaxel, line 9:
```suggestion
`docker ps --all`
```

pages.fr/common/docker.md
```diff
@@ -4,10 +4,6 @@
 > Certaines commandes comme `docker run` ont leur propre documentation.
 > Plus d'informations : <https://docs.docker.com/engine/reference/commandline/cli/>.
 
-- Liste les conteneurs Docker en cours d'exécution :
-
-`docker ps`
-
 - Liste tous les conteneurs Docker (en cours d'exécution ou arrêtés) :
 
 `docker ps -a`
```
navarroaxel, line 9:
```suggestion
`docker ps --all`
```

navarroaxel (approved):

einverne (approved):
Looks good, thanks.

CleanMachine1 (approved):

PR #7578, opened 2021-12-26
lid: add page

cyqsimon:

**Version of the command being documented (if known):** N/A


pages/linux/lid.md
```diff
@@ -0,0 +1,12 @@
+# lid
+
+> Display a user's groups or a group's users.
+> More information: <https://manned.org/lid.8>.
```
CleanMachine1, line 4:
```suggestion
> More information: <https://manned.org/lid>.
```

cyqsimon: There exists `lid.3` which is completely unrelated. I think we should keep the section number.

einverne (commented):
Just tried on Ubuntu 18.04, it shows

```
lid: can't locate `ID': No such file or directory
```
Did I miss something?

cyqsimon:
> Just tried on Ubuntu 18.04, it shows
> 
> ```
> lid: can't locate `ID': No such file or directory
> ```
> 
> Did I miss something?

Maybe Ubuntu has a binary with the same name? Let me grab a VM and test it real quick.

cyqsimon:
Installing Ubuntu on a VM really reminds me hard why I don't use Ubuntu... I'm two hours in and the VM is still struggling on first boot.

einverne:
> Installing Ubuntu on a VM really reminds me hard why I don't use Ubuntu... I'm two hours in and the VM is still struggling on first boot.

Hahaha, just tried on Ubuntu 20.04, not working either. May I ask what OS do you use?

cyqsimon:
Okay, finally got the freakin VM working.

It seems like on Ubuntu the binary name `lid` has been appropriated by the package `id-utils`, and it's a completely irrelevant binary. Therefore the `lid` provided by `libuser` is renamed to `libuser-lid`.

```sh
apt-file list libuser | grep bin | grep lid
libuser: /usr/sbin/libuser-lid
```

I will check Debian to see whether this change comes from Ubuntu's upstream or Ubuntu itself.



> May I ask what OS do you use?

Good thing you asked. I use Arch, btw.

... No not really; I just saw an opportunity to use that line and can't help myself. I use Manjaro (in other words I sacrificed some reputation to help my laziness).

cyqsimon:
Okay, just confirmed with Debian 11 that this rename from `lid` to `libuser-lid` did come from upstream (for Ubuntu).

What should we do in this case, if different Linux distros package the same binary under different names? Is there an existing guideline for this?

blueskyson:
@cyqsimon This situation does not happen often. We should open an issue to discuss how to deal with such situation. I think currently we can create [alias pages](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/translation-templates/alias-pages.md) for these different names.

sbrl:
Wait, so there are 5 different versions of this command that function in 5 different ways? That's wild.

cyqsimon:
This PR is pending resolution of #7596.

cyqsimon:
Pending. Keep alive.

pages/linux/lid.md
```diff
@@ -0,0 +1,12 @@
...
+
+- Show a user's groups (primary and secondary):
+
+`sudo lid {{user}}`
```
EmilySeville7cfg, line 8:
```suggestion
- List primary and secondary groups of a specific user:

`sudo lid {{username}}`
```

There are several similar items (groups) printed, so `list` is better sounds here.

pages/linux/lid.md
```diff
@@ -0,0 +1,12 @@
...
+
+- Show a group's users:
+
+`sudo lid --group {{group}}`
```
EmilySeville7cfg, line 12:
```suggestion
- List users of a specific group:

`sudo lid --group {{group_name}}`
```

EmilySeville7cfg (changes requested):

kbdharun (approved):
LGTM, after @EmilySeville7cfg's suggestions.

Waples:
@kbdharun PR is still marked as WIP; I'll close it end of the week if there are no updates

Waples:
Keeping this open since #7596 is still not 'really' decided..

sbrl:
I believe @MasterOdin had a suggestion that seems to have some agreement in adding a `See also` line to the description.

cyqsimon:
Ah yes this one. I've been constantly putting it off (frankly, because it's annoying and I didn't want to think about it). Yeah I should quit being lazy. Give me a moment.

cyqsimon:
Rebased.

cyqsimon:
So I took a look at all the discussion and still there seems to be a small problem:

If I understand the general consensus correctly, what's proposed is to reserve the file name `lid.md` for the program in the `id-utils` package, and add this one (the program in the `libuser` package) under another name (likely `libuser-lid.md` to align with Debian's binary name).

The problem with this however is that we currently do not have a page for `id-utils`'s `lid`. This means that when a Fedora/Arch user tries `tldr lid`, all they would see is that the page does not exist with no reference to `libuser-lid.md`. This makes `libuser-lid.md` (this PR) basically indiscoverable.

I can add a `lid.md` as a stub with only a "see also", but since I have never used `id-utils` and probably never will, I certainly should not be writing examples for it. If I do so, my guess is that the CI will certainly complain.

pages/linux/lid.md
```diff
@@ -0,0 +1,9 @@
+# lid
+
+> Query ID database and report results.
+> On Fedora and Arch Linux, the binary name `lid` is taken by another program. See `tldr libuser-lid`.
+> More information: <https://example.org/>.
```
cyqsimon, line 5: Also, I couldn't find `id-utils`'s `lid` on [manned.org](https://manned.org/browse/search?q=lid). [die.net](https://linux.die.net/man/1/lid)'s page is out of date, and I really do not want to link [unix.com](https://www.unix.com/man-page/debian/1/LID/).

kbdharun (commented):
@cyqsimon Instead of 2 pages, I would suggest documenting them in just `lid.md` and there we can add a note that the binary name in Fedora and Arch is `libuser-lid` . If the second page is really required then I would suggest using the [generic alias page template](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/translation-templates/alias-pages.md) along with a more information link.

cyqsimon:
> Instead of 2 pages, I would suggest documenting them in just `lid.md` and there we can add a note that the binary name in Fedora and Arch is `libuser-lid` .

No it's the other way around. `libuser-lid` is the binary name used by Debian; on Fedora and Arch it's just `lid`. So adding this page as `lid.md` would not be aligning with Debian naming, thus this quandary.

It seems like from previous discussion that quite a few maintainers prefer aligning with Debian; personally I see both sides of the argument, so I don't really have a strong preference either way.

> The package doesn't seem to be available in Fedora.

The package name is `libuser`. You can get this info with a reverse search: `dnf provides lid`.

pages/linux/lid.md
sbrl:
```suggestion
> More information: <https://www.gnu.org/software/idutils/>.
```
Found the GNU version here: https://www.gnu.org/software/idutils/

This link, given it's the homepage thereof, would be preferable over a man page.

It also has a link to a full manual.

pages/linux/lid.md
```diff
@@ -0,0 +1,9 @@
...
+
+- This page is currently a redirection stub:
```
sbrl, line 7: Maybe, a bit of clarification here to make it clear we're looking for contributors if we're merging this early?

```suggestion
- This page is currently a redirection stub (if you are familiar with this version of this command, please open a pull request!):
```

sbrl:
Oh gosh, I remember this one - it's so confusing! I've added some comments above to hopefully push this one forwards a bit.

sbrl:
You have no idea what you're talking about, @tldr-bot.

pages/linux/lid.md
cyqsimon: Ah thanks for the reminder. I must have missed the notification on @sbrl's reply. Updated.

pages/linux/lid.md
```diff
@@ -0,0 +1,9 @@
+# lid
+
+> Query ID database and report results.
+> On Fedora and Arch Linux, the binary name `lid` is taken by another program. See `tldr libuser-lid`.
+> More information: <https://www.gnu.org/software/idutils/>.
+
+- This page is currently a redirection stub (if you are familiar with this version of this command, please open a pull request!):
```
vitorhcl, line 7:
```suggestion
- View documentation for `libuser-lid` (this page is currently a redirection stub; if you are familiar with this version of this command, please open a pull request!):
```
The tldr-bot is throwing an error because "this" ends with the letter "s"

cyqsimon: I opted not to use the wording "this version of this command", because it can easily be misunderstood. These are two completely different programs that just so happen to have a naming conflict.

cyqsimon:
Bad bot!

pages/linux/lid.md
```diff
@@ -0,0 +1,10 @@
+# lid
+
+> NOTE: This page is currently a redirection stub. If you are familiar with this program, please open a pull request!
```
kbdharun, line 3:
```suggestion
> NOTE: This page is currently a redirection stub. If you are familiar with this program, please open a pull request.
```
This should fix `tldr-bot`'s error.

kbdharun (approved):

pages/linux/lid.md
cyqsimon: I think there is a case to be made for an exception here. If it's up to me, I would like to dismiss this CI error and merge anyway.

cyqsimon: Okay sure. We're pretty much bikeshedding at this point.

cyqsimon: Oh okay, I did not know that. I thought it was just an advisory sort of thing.

PR #7645, opened 2022-01-11
clifm: add page

leo-arch:

**Version of the command being documented (if known):** 1.4

### DIFF
--- a/pages/common/clifm.md
+++ b/pages/common/clifm.md
@@ -3,7 +3,7 @@
 > The command line file manager.
 > More information: <https://github.com/leo-arch/clifm>.
 
-- Change to/open the file whose ELN (entry list number) is 12 (ELN can be replaced by the actual file name, e.g. `/etc`):
+- Change to/open the file whose ELN (entry list number) is 12 (ELN can be replaced by the actual file name, e.g. /etc):
 
 `12`
 
@@ -13,9 +13,9 @@
 
 - Go back and forth in the directory history list:
 
-`Shift-Left` (or `Alt-j`) / `Shift-Right` (or `Alt-k`)
+`Shift-Left (or Alt-j) / Shift-Right (or Alt-k)`
 
-- Select all PNG files in the current directory: 
+- Select all PNG files in the current directory:
 
 `s *.png`
 
@@ -23,7 +23,7 @@
 
 `/*.pdf`
 
-- Remove the file whose ELN is 24 (use `t` to trash the file instead):
+- Remove the file whose ELN is 24 (use t to trash the file instead):
 
 `r 24`
 
@@ -33,4 +33,4 @@
 
 - Open the bookmarks screen:
 
-`bm` (or `Alt-b`)
+`bm (or Alt-b)`
--- a/pages/common/clifm.md
+++ b/pages/common/clifm.md
@@ -23,7 +23,7 @@
 
 `s *.png`
 
-- Remove the previously selected files (use t to send the files to the recycle bin instead):
+- Remove the previously selected files (use `t` to send the files to the recycle bin instead):
 
 `r sel`
 
--- a/pages/common/clifm.md
+++ b/pages/common/clifm.md
@@ -13,7 +13,7 @@
 
 - Go back and forth in the directory history list:
 
-<kbd>Shift-Left</kbd> (or <kbd>Alt-j</kbd>) / <kbd>Shift-Right</kbd> (or <kbd>Alt-k</kbd>)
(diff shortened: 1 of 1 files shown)


pages/common/clifm.md
```diff
@@ -0,0 +1,36 @@
+# clifm
+
+> The command line file manager.
+> More information: <https://github.com/leo-arch/clifm>.
+
+- Change to/open the file whose ELN (entry list number) is 12 (ELN can be replaced by the actual file name, e.g. /etc):
```
sbrl, line 6: These examples are great, but how does one start the program in the first place? tldr pages should tell a story of how the command is used, with the first example on the page being the command users will generally want to use first.

Also, this specific example may be best served as 2 separate examples rather than being rolled into 1.

pages/common/clifm.md
```diff
@@ -0,0 +1,36 @@
...
+
+`12`
+
+- Create a new file and a new directory:
+
+`n file dir/`
+
+- Go back and forth in the directory history list:
```
sbrl, line 14: Perhaps this will read better?
```suggestion
- Navigate backwards and forwards in the directory history list:
```

pages/common/clifm.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Select all PNG files in the current directory:
+
+`s *.png`
+
+- Search for PDF files in the current directory:
+
+`/*.pdf`
+
+- Remove the file whose ELN is 24 (use t to trash the file instead):
```
sbrl, line 26: Perhaps this will read better?
```suggestion
- Remove the file whose ELN is 24 (use t to send the file to the recycle bin instead):
```

pages/common/clifm.md
leo-arch: Hi @sbrl. Thanks for your kind review. 

To start the program one just runs `clifm`, which will open a new command line to run clifm specific commands. I just thought that, having a limited amount of available examples (8), it would be better to use them for less intuitive commands. The same goes for the idea of splitting the example into two. However, I perfectly understand the idea of a commands story. It makes sense in the context of `tldr`. I'll rewrite the examples and push again.

pages/common/clifm.md
```diff
@@ -0,0 +1,36 @@
...
+
+`r 24`
+
+- Copy selected files into the current directory:
```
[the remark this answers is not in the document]
leo-arch: Files selection is exemplified in the fourth example. What I should do, I guess, to make this clearer, is to place this specific example immediately below the files selection one.

pages/common/clifm.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Search for PDF files in the current directory:
+
+`/*.pdf`
+
+- Select all PNG files in the current directory:
+
+`s *.png`
+
+- Remove the previously selected files (use t to send the files to the recycle bin instead):
```
navarroaxel, line 26:
```suggestion
- Remove the previously selected files (use `t` to send the files to the recycle bin instead):
```

pages/common/clifm.md
sbrl: Remember that tldr pages are for both beginners and experts. Space is limited, so we have to balance the need for a page to be useful to both beginners and experts. To this end, larger numbers of advanced examples may be better suited to a dedicated cheatsheet in a different project?

pages/common/clifm.md
```diff
@@ -3,34 +3,34 @@
...
-`/*.pdf`
-
-- Remove the file whose ELN is 24 (use t to trash the file instead):
+- Remove the previously selected files (use `t` to send the files to the recycle bin instead):
 
-`r 24`
+`r sel`
 
-- Copy selected files into the current directory:
+- Learn more:
```
sbrl, line 30: Sorry to nitpick, but would this read better perhaps?
```suggestion
- Display detailed help:
```

sbrl (approved):
Looks good to me after my above comment is addressed :-)

pages/common/clifm.md
leo-arch: Hi @sbrl. I understand your point. Because CliFM is a command line program, it has a lot of available commands. This is why I designed a well documented manpage and a quite big Wiki. However, I really value your work with `tldr`, and I think clifm should be here: a little introduction is always a good thing. I making short intra-program help sections myself: having only a huge manpage/Wiki is far from optimal.

pages/common/clifm.md
leo-arch: Looks good to me.

navarroaxel (approved):

PR #7717, opened 2022-01-27
code: add --new-window example

The-Compiler:
It seems to be VS Code's default behavior (when opened from a terminal) to open files in an existing window, so I end up having to look up --new-window much more than --reuse-window.

1.63.2


The-Compiler:
The webpage to accept the CLA seems... rather broken:

image

Will try again later.

navarroaxel (approved):

blueskyson (approved):

PR #1702, opened 2017-11-28
read: add -r option

pirafrank:
----

- [ ] The page (if new), does not already exist in the repo.

- [ ] The page (if new), has been added to the correct platform folder:  
      `common/` if it's common to all platforms, `linux/` if it's Linux-specific, and so on.

- [ ] The page has 8 or fewer examples.

- [ ] The PR is appropriately titled:  
      `<command name>: add page` for new pages, or `<command name>: <description of changes>` for pages being edited

- [ ] The page follows the [contributing](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md) guidelines

PR #7731, opened 2022-01-31
read: fix missing backslash

Saevyn:


navarroaxel:
@Saevyn welcome. Why this change?

Saevyn:
The current [tldr page](https://tldr.ostera.io/read) for `read` has a missing backslash (\\). This pull request is just a simple fix for that.
I also took the liberty of checking the whole repository for similar errors with `grep -r '\\)'`, but this is the only one that I found.

navarroaxel:
> The current [tldr page](https://tldr.ostera.io/read) for `read` has a missing backslash (\). This pull request is just a simple fix for that. I also took the liberty of checking the whole repository for similar errors with `grep -r '\\)'`, but this is the only one that I found.

@Saevyn I checked the PR #1702 that added this specific example. Looks like the `(\)` was added only to let readers know what `backslash` mean, maybe for help to non-native English speakers.

The escape character is a single backslash `\`, 2 means an already escaped backslash `\\`. Let's wait another reviewer.

navarroaxel:
@bl-ue ?

bl-ue (approved):
@navarroaxel I like this change. Currently, when I run `tldr read` (with the node client), I get this example, among others:
```
- Do not let backslash () act as an escape character:
```
...notice how the backslash isn't rendering.

I tried out this PR, and this is what I got:
```
- Do not let backslash (\) act as an escape character:
```
...which looks a lot better.

einverne:
I just test in the python client, it seems it is a rendering problem of the node client.

image

bl-ue:
Actually, given the possible variance of clients' displaying of the backslash (I believe they're not conforming to the markdown standard, but whatever), how about we remove the ` (\)` entirely?

pages/common/read.md
```diff
@@ -19,7 +19,7 @@
 
 `read -d {{new_delimiter}} {{variable}}`
 
-- Do not let backslash (\) act as an escape character:
+- Do not let backslash (\\) act as an escape character:
```
bl-ue, line 22:
```suggestion
- Do not let backslash act as an escape character:
```

einverne:
Sure, I think the `backslash` word is enough.

Saevyn:
I'd vote for keeping the double backslash `(\\)`, as that should fix the problem for those who are missing it and the Python version would remain unaffected. Alternatively, isolating the backslash in a code block would also fix the issue: `` `(\)` ``

I'm not particularly committed in keeping the backslash, however. I just think it's a nice visual hint.

bl-ue:
`` `(\)` `` does seem to fix it; however, I would prefer it be removed.

My opinion may not as much as it did in the past, though, so I'm willing to be overruled :)

CleanMachine1:
Yeah, I agree with @einverne, remove it, it doesn't really add anything

Waples:
It renders perfectly in the Python client. Fix this outside the `pages`, in the Node client, please.

Saevyn:
> It renders perfectly in the Python client. Fix this outside the `pages`, in the Node client, please.

I don't think this is exclusively a Node client issue. I assume it works in Python because backslashes are printed by default if they're not part of a escape character. However, `print (\)` gives the same result as `print(\\)` in Python, so I see no reason why we should somehow fix this in the Node client when *this* simple fix should suffice. Not to mention that this error is also present in the [live demo](https://tldr.ostera.io), the [Haskell client](https://github.com/psibi/tldr-hs), [the official PDF](https://tldr.sh/assets/tldr-book.pdf) and even [on GitHub itself](https://github.com/tldr-pages/tldr/blob/main/pages/common/read.md) — and those are just the ones I tested.

Honestly, it puzzles me why such a trivial solution for a simple issue as this one was dragged on for so long. I thought it was a no-brainer...

bl-ue:
@Saevyn I agree there, on GitHub and ostera's it looks pretty ugly. What about my previous suggestion to remove the ```(`)``` part entirely? I'm always a bit irked by such "hints" - I really think it's not necessary to indicate what a backslash is...

bl-ue:
How about changing it to `- Do not let the backslash character act as an escape character:` ?

Saevyn:
I think that's a reasonable solution as well! I'd be fine with it, I just don't understand why this PR was closed without even acknowledging the issue.

SethFalco (approved):

pages/common/read.md
SethFalco: If other maintainers prefer this, happy to run with it. I personally have many friends that mix up forward slash (/) and backslash (\\), so I think having the character helps.

kbdharun: > If other maintainers prefer this, happy to run with it. I personally have many friends that mix up forward slash (`/`) and backslash (`\`), so I think having the character helps.

I definitely agree that it might be confusing to some. So, I think we can go with this change for proper rendering.

kbdharun (approved):

SethFalco:
Thanks for your contribution, and sorry again for the hiccups before!

PR #7748, opened 2022-02-05
nm-classic: add page

pixelcmtd:
alias of `nm`


navarroaxel (approved):

EmilySeville7cfg (approved):
I think we can require such descriptions and page body for all aliases via `tldr-lint` not to produce any mistakes. ;)

PR #7758, opened 2022-02-06
nm: update page

EmilySeville7cfg:

## ToDo

- [x] prefer long options
- [x] fix placeholders

### DIFF
--- a/pages/common/nm.md
+++ b/pages/common/nm.md
@@ -5,15 +5,15 @@
 
 - List global (extern) functions in a file (prefixed with T):
 
-`nm --extern-only {{path/to/file.o}}`
+`nm -g {{path/to/file.o}}`
 
 - List only undefined symbols in a file:
 
-`nm --undefined-only {{path/to/file.o}}`
+`nm -u {{path/to/file.o}}`
 
 - List all symbols, even debugging symbols:
 
-`nm --debug-syms {{path/to/file.o}}`
+`nm -a {{path/to/file.o}}`
 
 - Demangle C++ symbols (make them readable):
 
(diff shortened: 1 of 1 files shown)


navarroaxel (approved):

pages/common/nm.md
```diff
@@ -1,19 +1,20 @@
...
 
 - List global (extern) functions in a file (prefixed with T):
 
-`nm -g {{file.o}}`
+`nm --extern-only {{path/to/file.o}}`
 
 - List only undefined symbols in a file:
 
-`nm -u {{file.o}}`
+`nm --undefined-only {{path/to/file.o}}`
```
pixelcmtd, line 12:
```suggestion
`nm -u {{path/to/file.o}}`
```

pages/common/nm.md
```diff
@@ -1,19 +1,20 @@
...
 
 - List all symbols, even debugging symbols:
 
-`nm -a {{file.o}}`
+`nm --debug-syms {{path/to/file.o}}`
```
pixelcmtd, line 16:
```suggestion
`nm -a {{path/to/file.o}}`
```

pixelcmtd (changes requested):

mfrw (commented):
Given these pages are in common, should we create new pages for `osx` which in my understanding would not have the long version of the flags ?

pixelcmtd:
> should we create new pages for `osx` which [...] would not have the long version of the flags ?

if we were to do something like that, i'd instead make new `linux` pages which have the long options, because initially the long options were a gnu-specific thing, for example, early freebsd versions didn't have it either.

EmilySeville7cfg:
> if we were to do something like that, i'd instead make new linux pages which have the long options, because initially the long options were a gnu-specific thing, for example, early freebsd versions didn't have it either.

I think we can use `general/` for pages with most portable options (short/long) and `linux/` for just GNU options.

PR #7772, opened 2022-02-13
PULL_REQUEST_TEMPLATE: refresh

CleanMachine1:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -12,3 +14,11 @@ See also https://github.com/tldr-pages/tldr/blob/main/CONTRIBUTING.md
...
 
 **Version of the command being documented (if known):**
+
+## Follow-up
+
+<!--
+Please tick this box after the CI has reported no issues
+-->
+
+- [ ] Resolved issue(s) reported from tldr-bot (if applicable).
```
navarroaxel, line 24: But we have a CI check for this :thinking:

navarroaxel: But the CI doesn't automatically run for first-time collaborators. The `tldr-lint` is already mentioned in the `style-guide` :thinking: 
Should we re-enforce that tool?

mfrw: I am of the same opinion as @navarroaxel.
We have a CI check for this, in my opinion, it may only increase the cognitive load on the contributors.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -12,3 +14,11 @@ See also https://github.com/tldr-pages/tldr/blob/main/CONTRIBUTING.md
 - [ ] The page description includes a link to documentation or a homepage (if applicable).
```
navarroaxel, line 14: @CleanMachine1 what if we add a check about `tldr-lint` passed?
```markdown
 - [ ] The page passes the `tldr-lint` check.
```

PR #7791, opened 2022-02-17
gcc: fix debug symbols example

jwakely:
The `-Og` option enables optimization, it does **not** add debug info to the output. That's what `-g` does.

### DIFF
--- a/pages/common/gcc.md
+++ b/pages/common/gcc.md
@@ -9,7 +9,7 @@
 
 - Allow warnings, debug symbols in output:
 
-`gcc {{source.c}} -Wall -Og --output {{executable}}`
+`gcc {{source.c}} -Wall -g --output {{executable}}`
 
 - Include libraries from a different path:
 


navarroaxel:
@jwakely could please sign the license to pass the CLA check?

jwakely:
Sorry, no, I don't do CLAs if I can help it.

https://opensource.com/article/19/2/cla-problems
https://sfconservancy.org/blog/2014/jun/09/do-not-need-cla/
https://ben.balter.com/2018/01/02/why-you-probably-shouldnt-add-a-cla-to-your-open-source-project/

I don't believe a single byte edit is copyrightable or needs my permission to distribute it under any licence, so feel free to make the correction yourself.

jwakely:
N.B. if the default branch is now `main` then you should fix the EDIT THIS PAGE ON GITHUB link shown on pages like https://tldr.ostera.io/gcc because that still takes people to the `master` branch.

MasterOdin:
Per the [gcc manual](https://gcc.gnu.org/onlinedocs/gcc-11.2.0/gcc/Debugging-Options.html#Debugging-Options):

> If you are not using some other optimization option, consider using -Og (see [Optimize Options](https://gcc.gnu.org/onlinedocs/gcc-11.2.0/gcc/Optimize-Options.html#Optimize-Options)) with -g. With no -O option at all, some compiler passes that collect information useful for debugging do not run at all, so that -Og may result in a better debugging experience.

Given the limited example nature of tldr, wouldn't it be better to just only include the `-Og` example? Or maybe just amend the description to be something like "Allow warnings, debug symbols in optimized output"?

jwakely:
> Given the limited example nature of tldr, wouldn't it be better to just keep including the -Og example as "best practice",

That current example is wrong though.

> Maybe just amend the description to be something like "Allow warnings, debug symbols in optimized output"?

But the current example doesn't do that. Without `-g` there are no debug symbols. That's the point of this PR. The current example enables warnings and optimizes for a better debugging experience, but doesn't include debug symbols so debugging is practically impossible. That's not helpful.

You make a good point though. It would be a lot better if **every** example included **at least** `-Wall` and `-g` because beginners and people unfamiliar with GCC options should **ALWAYS** have warnings enabled. ***ALWAYS**. And there's no downside to also enabling debug symbols. So include `-Wall -g` in every example. And have an example that adds `-Og` for "optimized for better debugging experience".

MasterOdin:
For clarification, I think I was assuming `gcc` did short flag combining (e.g. `-Og` and `-O -g` were the same), but I realize it doesn't, and that `-Og` is an optimization level. So the example to just add debug symbols would be:

> `gcc {{source.c}} -Wall -Og -g --output {{executable}}`

jwakely:
Correct.

PR #7801, opened 2022-02-18
chsh: Fix `chsh -s SHELL` syntax

### LINKED ISSUE #7800: chsh: bad advice
```
; tldr chsh

chsh

Change the user's login shell.
More information: <https://manned.org/chsh>.

- Change the current user's login shell interactively:
    chsh

- Change the current user's login shell:
    chsh path/to/shell

...
```

The second example there is wrong: on macos and [Linux](https://linux.die.net/man/1/chsh), a non-option argument is always a username, not a shell.

sourcefrog:
Fixes #7800

**Version of the command being documented (if known):** The behavior is the same on recent macOS and Linux (and I think other unix too)

I checked and none of the non-English translations have this bug.

I've used just `-s` because that seems very idiomatic (it's used in all the translations), it may work on older machines where `--shell` does not, and anyhow there's an example with `--shell`.

### DIFF
--- a/pages/common/chsh.md
+++ b/pages/common/chsh.md
@@ -7,7 +7,7 @@
 
 `chsh`
 
-- Change the current user's login shell:
+- Change the login shell of the current user:
 
 `chsh -s {{path/to/shell}}`
 
--- a/pages/common/chsh.md
+++ b/pages/common/chsh.md
@@ -9,7 +9,7 @@
 
 - Change the current user's login shell:
 
-`chsh {{path/to/shell}}`
+`chsh -s {{path/to/shell}}`
 
 - Change the login shell for a given user:
 
--- a/pages/linux/chsh.md
+++ b/pages/linux/chsh.md
@@ -7,13 +7,13 @@
 
 `chsh`
 
-- Change the login shell of the current user:
+- Change the current user's login shell:
 
-`chsh -s {{path/to/shell}}`
+`chsh --shell {{path/to/shell}}`
 
 - Change the login shell for a given user:
 
-`chsh -s {{path/to/shell}} {{username}}`
+`chsh --shell {{path/to/shell}} {{username}}`
 
 - List available shells:
 


blueskyson (approved):

sourcefrog:
> Hi, we tend to try and use longer flags since they are simpler for the user to understand (one valid acceptation is commands that are used often such as git commit -m instead of using --message).

Makes sense in general. That works with git because it's the same everywhere, which will not be true for base-OS commands like chsh.
 
> What OS's aren't working with --shell?

According to <https://man.openbsd.org/chsh>, OpenBSD expects just `-s`, for one.

EmilySeville7cfg:
@CleanMachine1, what about to move `chsh` with long option to `linux/` directory?

pages/common/chsh.md
```diff
@@ -7,9 +7,9 @@
 
 `chsh`
 
-- Change the current user's login shell:
+- Change the login shell of the current user:
 
-`chsh {{path/to/shell}}`
+`chsh -s {{path/to/shell}}`
```
EmilySeville7cfg, line 12: I suggest move copy of this page with long option to `linux/` directory. We can use `common/` as smth middle between `osx/` and `linux/`: with long and short options depending on usage frequency. 🤔

EmilySeville7cfg (changes requested):

pages/common/chsh.md
sourcefrog: Could you just merge this, since the page is simply wrong today, and do that later?

Waples: I'm with Emily on this one.

EmilySeville7cfg: @sourcefrog, please duplicate page with long options to `linux/` and I approve && merge this PR.

sourcefrog: ok

sourcefrog: Actually I noticed macOS also does not accept `--shell`, and that's somewhat more common than OpenBSD 😉 . 

I'll use `-s` generically and `--shell` on Linux.

sourcefrog:
OK, please take another look?

EmilySeville7cfg (approved):

EmilySeville7cfg:
Good job, thanks! 😄

PR #7840, opened 2022-03-02
arpspoof: add page

danbulant:

2.4

### DIFF
--- a/pages/linux/arpspoof.md
+++ b/pages/linux/arpspoof.md
@@ -1,0 +1,15 @@
+# arpspoof
+
+> Intercept packets on a switched LAN.
+> Redirects packets from a target host (or all hosts) on the LAN intended for another host on the LAN by forging ARP replies. This is an effective way of sniffing traffic on a switch.
+> Kernel IP forwarding must be turned on ahead of time.
+> For more information: <https://linux.die.net/man/8/arpspoof>
+
+- Make a simple man-in-the-middle attack between two local hosts (run both at the same time, via `screen` or multiple terminal sessions):
+
+`arpspoof -i {{eth0}} -t {{192.168.0.1}} {{192.168.0.2}}`
+`arpspoof -i {{eth0}} -t {{192.168.0.2}} {{192.168.0.1}}`
+
+- Make a simple man-in-the-middle attack for all local clients:
+
+`arpspoof -i {{eth0}} {{192.168.0.1}}`
--- a/pages/linux/arpspoof.md
+++ b/pages/linux/arpspoof.md
@@ -3,7 +3,7 @@
 > Intercept packets on a switched LAN.
 > Redirects packets from a target host (or all hosts) on the LAN intended for another host on the LAN by forging ARP replies. This is an effective way of sniffing traffic on a switch.
 > Kernel IP forwarding must be turned on ahead of time.
-> For more information: <https://linux.die.net/man/8/arpspoof>
+> More information: <https://manned.org/arpspoof>
 
 - Make a simple man-in-the-middle attack between two local hosts (run both at the same time, via `screen` or multiple terminal sessions):
 
--- a/pages/linux/arpspoof.md
+++ b/pages/linux/arpspoof.md
@@ -9,6 +9,6 @@
 `sudo arpspoof -i {{eth0}} -t {{192.168.0.1}} {{192.168.0.2}}`
 `sudo arpspoof -i {{eth0}} -t {{192.168.0.2}} {{192.168.0.1}}`
 
-- Make a simple man-in-the-middle attack for all local clients:
+- Perform a simple man-in-the-middle attack for all local clients:
 
 `sudo arpspoof -i {{eth0}} {{192.168.0.1}}`
--- a/pages/linux/arpspoof.md
+++ b/pages/linux/arpspoof.md
@@ -1,7 +1,7 @@
 # arpspoof
 
 > Intercept packets on a switched LAN.
-> Redirects packets from a target host (or all hosts) on the LAN intended for another host on the LAN by forging ARP replies. This is an effective way of sniffing traffic on a switch.
+> Redirects packets from a target host (or all hosts) on the LAN intended for another host on the LAN by forging ARP replies.
 > More information: <https://manned.org/arpspoof>
 
 - Make a simple man-in-the-middle attack between two local hosts (run both at the same time, via `screen` or multiple terminal sessions):
--- a/pages/linux/arpspoof.md
+++ b/pages/linux/arpspoof.md
@@ -2,7 +2,6 @@
 
 > Intercept packets on a switched LAN.
 > Redirects packets from a target host (or all hosts) on the LAN intended for another host on the LAN by forging ARP replies. This is an effective way of sniffing traffic on a switch.
-> Kernel IP forwarding must be turned on ahead of time.
 > More information: <https://manned.org/arpspoof>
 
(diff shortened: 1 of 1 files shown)


pages/linux/arpspoof.md
```diff
@@ -0,0 +1,15 @@
+# arpspoof
+
+> Intercept packets on a switched LAN.
+> Redirects packets from a target host (or all hosts) on the LAN intended for another host on the LAN by forging ARP replies. This is an effective way of sniffing traffic on a switch.
+> Kernel IP forwarding must be turned on ahead of time.
+> For more information: <https://linux.die.net/man/8/arpspoof>
```
navarroaxel, line 6: We prefer `manned` for man pages
```suggestion
> More information: <https://manned.org/arpspoof>
```

pages/linux/arpspoof.md
```diff
@@ -0,0 +1,15 @@
+# arpspoof
+
+> Intercept packets on a switched LAN.
+> Redirects packets from a target host (or all hosts) on the LAN intended for another host on the LAN by forging ARP replies. This is an effective way of sniffing traffic on a switch.
+> Kernel IP forwarding must be turned on ahead of time.
```
navarroaxel, line 5: We accept up to 2 lines for command description
```suggestion
```

pages/linux/arpspoof.md
```diff
@@ -0,0 +1,14 @@
+# arpspoof
+
+> Intercept packets on a switched LAN.
+> Redirects packets from a target host (or all hosts) on the LAN intended for another host on the LAN by forging ARP replies. This is an effective way of sniffing traffic on a switch.
```
EmilySeville7cfg, line 4:
```suggestion
> Redirects packets from a target host (or all hosts) on the LAN intended for another host on the LAN by forging ARP replies.
```

Line was too long. ;)

pages/linux/arpspoof.md
```diff
@@ -0,0 +1,14 @@
...
+> More information: <https://manned.org/arpspoof>
```
EmilySeville7cfg, line 5:
```suggestion
> More information: <https://manned.org/arpspoof>.
```

EmilySeville7cfg (changes requested):

pages/linux/arpspoof.md
```diff
@@ -0,0 +1,14 @@
+# arpspoof
+
+> Intercept packets on a switched LAN.
+> Redirects packets from a target host (or all hosts) on the LAN intended for another host on the LAN by forging ARP replies.
+> More information: <https://manned.org/arpspoof>.
+
+- Make a simple man-in-the-middle attack between two local hosts (run both at the same time, via `screen` or multiple terminal sessions):
+
+`sudo arpspoof -i {{eth0}} -t {{192.168.0.1}} {{192.168.0.2}}`
+`sudo arpspoof -i {{eth0}} -t {{192.168.0.2}} {{192.168.0.1}}`
```
EmilySeville7cfg, line 10:
```suggestion
```

Code examples are one-liners.

EmilySeville7cfg (changes requested):

pages/linux/arpspoof.md
```diff
@@ -0,0 +1,14 @@
+# arpspoof
+
+> Intercept packets on a switched LAN.
+> Redirects packets from a target host (or all hosts) on the LAN intended for another host on the LAN by forging ARP replies.
+> More information: <https://manned.org/arpspoof>.
+
+- Make a simple man-in-the-middle attack between two local hosts (run both at the same time, via `screen` or multiple terminal sessions):
```
EmilySeville7cfg, line 7:
```suggestion
- Make a simple man-in-the-middle attack between the specified local hosts (`{{192.168.0.1}} {{192.168.0.2}}`):
```

EmilySeville7cfg (changes requested):

pages/linux/arpspoof.md
danbulant: This needs two lines though to work correctly, and in this case I think it's justified, as it's one of the main purposes of the program - to test against MITM attack possibilities (in environments where you have explicit approval).
I think it's better for it to be two lines rather than one line with `;`.

pages/linux/arpspoof.md
danbulant: I'm not sure - this works only with 2 hosts (well, it might work with more, though I haven't tried that), and I think it's good to mention that both commands need to be run at the same time.

pages/linux/arpspoof.md
EmilySeville7cfg: TlDr pages should have one-line examples. Linter doesn't accept (by default) multiline code examples.

pages/linux/arpspoof.md
sbrl: The `{{token_syntax}}` and code blocks aren't normally used in descriptions?

Either way, I think the original description is pretty much ok? If I were to nitpick I'd say add `e.g.` before the `run both...` bit, but that's not necessary.

pages/linux/arpspoof.md
```diff
@@ -0,0 +1,14 @@
+# arpspoof
+
+> Intercept packets on a switched LAN.
+> Redirects packets from a target host (or all hosts) on the LAN intended for another host on the LAN by forging ARP replies.
```
sbrl, line 4: You mean `to` here?
```suggestion
> Redirects packets from a target host (or all hosts) on the LAN intended to another host on the LAN by forging ARP replies.
```

pages/linux/arpspoof.md
sbrl: Yeah, unfortunately we can't have more than 1 line in each example because of a limitation in our linter (which would cause all subsequent pull requests to fail the linter if we merge 1 PR that doesn't pass the linter).

pages/linux/arpspoof.md
```diff
@@ -0,0 +1,14 @@
...
+> More information: <https://manned.org/arpspoof>.
+
+- Make a simple man-in-the-middle attack between two local hosts (run both at the same time, via `screen` or multiple terminal sessions):
+
+`sudo arpspoof -i {{eth0}} -t {{192.168.0.1}} {{192.168.0.2}}`
+`sudo arpspoof -i {{eth0}} -t {{192.168.0.2}} {{192.168.0.1}}`
+
+- Make a simple man-in-the-middle attack for all local clients:
```
sbrl, line 12: Completely optional.
```suggestion
- Perform a simple man-in-the-middle attack for all local clients:
```

sbrl (commented):
Thanks for the new page! I've left some comments below for you to review. Minor things :-)

CleanMachine1:
If someone wants to assign themselves to this and resolve the issues, that would be great.

Pinging @danbulant just in case.

mhutter:
If it needs to be on one line, this would work:

    `sudo arpspoof -i {{eth0}} -t {{192.168.0.1}} {{192.168.0.2}} & sudo arpspoof -i {{eth0}} -t {{192.168.0.2}} {{192.168.0.1}} &`

johandegn:
I must somehow have missed this PR when i submitted one for this page (#9903). Regarding the discussion about the two lines, i have some input:
The program is in its simplicity just able to forge ARP replies on a network such that the spoofed host will try to connect to your machine instead of the intended. This _can_ be used to facilitate MITM attacks, but i think that the program is more general than that. I do however agree that this is often how it is used used, and there is actually a flag ( `-r`) that essentially collapses the two lines into one by spoofing both the host and the target (i did include an example using that in #9903). So IMHO i think that solves the problem.

danbulant:
@johandegn thanks, didn't know about that flag

PR #7841, opened 2022-03-02
setsid: add page

git-em:

`setsid from util-linux 2.37.4`

### DIFF
--- a/pages/linux/setsid.md
+++ b/pages/linux/setsid.md
@@ -1,0 +1,25 @@
+# setsid
+
+> Run a program in a new session if the calling process is not a process group leader.
+> The created session is by default not controlled by the current terminal.
+> More information: <https://manned.org/setsid>.
+
+- Run a program in a new session:
+
+`setsid {{program}}`
+
+- Run a program in a new session discarding the resulting output and error:
+
+`setsid {{program}} > /dev/null 2>&1`
+
+- Run a program creating a new process:
+
+`setsid --fork {{program}}`
+
+- Return the exit value of a program as the return value of setsid when the program exits:
+
+`setsid --wait {{program}}`
+
+- Run a program in a new session setting the current terminal as the controlling terminal:
+
+`setsid --ctty {{program}}`
--- a/pages/linux/setsid.md
+++ b/pages/linux/setsid.md
@@ -16,7 +16,7 @@
 
 `setsid --fork {{program}}`
 
-- Return the exit value of a program as the return value of setsid when the program exits:
+- Return the exit code of a program as the exit code of setsid when the program exits:
 
 `setsid --wait {{program}}`
 


git-em:
Listed on #2214

pages/linux/setsid.md
```diff
@@ -0,0 +1,25 @@
...
+> The created session is by default not controlled by the current terminal.
+> More information: <https://manned.org/setsid>.
+
+- Run a program in a new session:
+
+`setsid {{program}}`
+
+- Run a program in a new session discarding the resulting output and error:
+
+`setsid {{program}} > /dev/null 2>&1`
```
EmilySeville7cfg, line 13:
```suggestion
`setsid {{program}} > &> /dev/null`
```

EmilySeville7cfg (changes requested):

pages/linux/setsid.md
git-em: I initially written like this, but when i tested it in `dash` it didn't work, so i opted for the most shell agnostic and maybe POSIX compliant way (if i understood [this](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_07_02) well).

git-em: Just as a curiosity, neither seemed to work with `[t]csh`...

pages/linux/setsid.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Run a program creating a new process:
+
+`setsid --fork {{program}}`
+
+- Return the exit value of a program as the return value of setsid when the program exits:
```
pixelcmtd, line 19:
```suggestion
- Return the exit code of a program as the exit code of setsid when the program exits:
```

pages/linux/setsid.md
pixelcmtd, line 13:
```suggestion
`setsid {{program}} >/dev/null 2>&1`
```
I prefer this, I've never seen `&>` and don't know what it does neither how portable it is.

git-em: I believe the space between the "`>`" and the following text is good practice, as seen on this [page](https://www.gnu.org/software/bash/manual/html_node/Redirections.html) about redirection in `bash`. 

As for the `&>`, it's a convenient way to redirect stdout and stderr, present in `bash` ([see previous link](https://www.gnu.org/software/bash/manual/html_node/Redirections.html)) and [`zsh`](https://zsh.sourceforge.io/Doc/Release/Redirection.html), but not in some other shells.

CleanMachine1 (approved):
Once @EmilySeville7cfg's review is done, please merge

pages/linux/setsid.md
EmilySeville7cfg: Yes, maybe it's better to leave your code unchanged because `Tl;Dr` is not restricted to concrete shells.

EmilySeville7cfg (approved):

EmilySeville7cfg:
Merged, thanks! ❤️

PR #7847, opened 2022-03-03
comby: add page

xxchan:

### DIFF
--- a/pages/common/comby.md
+++ b/pages/common/comby.md
@@ -1,20 +1,20 @@
 # comby
 
-> A tool for structural code search and replace that supports ~every language.
+> Tool for structural code search and replace that supports many languages.
 > More information: <https://github.com/comby-tools/comby>.
 
-- Match and rewrite templates, and display any changes on the console:
+- Match and rewrite templates, and print changes:
 
-`comby {{'assert_eq!(:[a], :[b])'}} {{'assert_eq!(:[b], :[a])'}} {{.rs}}`
+`comby '{{assert_eq!(:[a], :[b])}}' '{{assert_eq!(:[b], :[a])}}' {{.rs}}`
 
 - Match and rewrite with rewrite properties:
 
-`comby {{'assert_eq!(:[a], :[b])'}} {{'assert_eq!(:[b].Capitalize, :[a])'}} {{.rs}}`
+`comby '{{assert_eq!(:[a], :[b])}}' '{{assert_eq!(:[b].Capitalize, :[a])}}' {{.rs}}`
 
 - Match and rewrite in-place:
 
 `comby -in-place {{match_pattern}} {{rewrite_pattern}}`
 
-- Only perform matching:
+- Only perform matching and print matches:
 
 `comby -match-only {{match_pattern}} ""`
--- a/pages/common/comby.md
+++ b/pages/common/comby.md
@@ -13,8 +13,8 @@
 
 - Match and rewrite in-place:
 
-`comby -in-place {{match_pattern}} {{rewrite_pattern}}`
+`comby -in-place '{{match_pattern}}' '{{rewrite_pattern}}'`
 
 - Only perform matching and print matches:
 
-`comby -match-only {{match_pattern}} ""`
+`comby -match-only '{{match_pattern}}' ""`
--- a/pages/common/comby.md
+++ b/pages/common/comby.md
@@ -1,0 +1,20 @@
+# comby
+
+> A tool for structural code search and replace that supports ~every language.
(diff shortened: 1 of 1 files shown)


pages/common/comby.md
```diff
@@ -0,0 +1,20 @@
+# comby
+
+> A tool for structural code search and replace that supports ~every language.
+> More information: <https://github.com/comby-tools/comby>.
+
+- Match and rewrite templates, and display any changes on the console:
+
+`comby {{'assert_eq!(:[a], :[b])'}} {{'assert_eq!(:[b], :[a])'}} {{.rs}}`
+
+- Match and rewrite with rewrite properties
```
xxchan, line 10:
```suggestion
- Match and rewrite with rewrite properties:
```

pages/common/comby.md
```diff
@@ -0,0 +1,20 @@
+# comby
+
+> A tool for structural code search and replace that supports ~every language.
+> More information: <https://github.com/comby-tools/comby>.
+
+- Match and rewrite templates, and display any changes on the console:
```
navarroaxel, line 6: We don't use `console`, we could use `stdout`
```suggestion
- Match and rewrite templates, and print changes:
```
which template?

pages/common/comby.md
```diff
@@ -0,0 +1,20 @@
...
+
+`comby {{'assert_eq!(:[a], :[b])'}} {{'assert_eq!(:[b], :[a])'}} {{.rs}}`
+
+- Match and rewrite with rewrite properties:
+
+`comby {{'assert_eq!(:[a], :[b])'}} {{'assert_eq!(:[b].Capitalize, :[a])'}} {{.rs}}`
```
navarroaxel, line 12: Quote outside syntax tokens
```suggestion
`comby '{{assert_eq!(:[a], :[b])}}' '{{assert_eq!(:[b].Capitalize, :[a])}}' {{.rs}}`
```

pages/common/comby.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Match and rewrite in-place:
+
+`comby -in-place {{match_pattern}} {{rewrite_pattern}}`
+
+- Only perform matching:
```
navarroaxel, line 18: match and print something?

xxchan, line 18:
```suggestion
- Only perform matching and print matches:
```

pages/common/comby.md
xxchan: Actually I copied from `tldr rustfmt`: Check a file for formatting and display any changes on the console

> which template?

It's `comby -in-place {{match_pattern}} {{rewrite_pattern}}`, so `'assert_eq!(:[a], :[b])'` is a template. I want to show how this template look like so used an example instead... Especially for the case with `:[b].Capitalize`

pages/common/comby.md
```diff
@@ -0,0 +1,20 @@
...
+
+`comby '{{assert_eq!(:[a], :[b])}}' '{{assert_eq!(:[b], :[a])}}' {{.rs}}`
+
+- Match and rewrite with rewrite properties:
+
+`comby '{{assert_eq!(:[a], :[b])}}' '{{assert_eq!(:[b].Capitalize, :[a])}}' {{.rs}}`
+
+- Match and rewrite in-place:
+
+`comby -in-place {{match_pattern}} {{rewrite_pattern}}`
```
EmilySeville7cfg, line 16:
```suggestion
`comby -in-place '{{match_pattern}}' '{{rewrite_pattern}}'`
```

pages/common/comby.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Only perform matching and print matches:
+
+`comby -match-only {{match_pattern}} ""`
```
EmilySeville7cfg, line 20:
```suggestion
`comby -match-only '{{match_pattern}}' ""`
```

EmilySeville7cfg (changes requested):

pages/common/comby.md
xxchan: since simple examples is already provided above for commands without options, will it be repetitive to give more examples here?

EmilySeville7cfg: Yes it will be repetitive. But on the other hand you can choose one appropriate example which is used when you want to demonstrate some options of command and how they affects command behavior. In case of `sed` previous sentence means - select some sed's command (**s**ubstitute for instance) and use it with some option (`-E` for instance) and without it and show the difference in BRE and ERE syntax on simple replacement. 😁

EmilySeville7cfg (approved):

navarroaxel (approved):

PR #7855, opened 2022-03-06
README: remove build badge

CleanMachine1:


navarroaxel (approved):

owenvoke (commented):
Personally I think this is quite useful, it shows (at a glance) that tests are passing, the PDF is generating successfully, and that the site has been deployed correctly.

Happy either way though. 👍🏻

CleanMachine1:
Ah, ok. 

I didn't think about the PDF and actually thought it was being made [here](https://github.com/tldr-pages/tldr-pages.github.io)
Lets get the opinion of others.

owenvoke:
@CleanMachine1, yeah it gets published as part of [this workflow](https://github.com/tldr-pages/tldr/blob/main/.github/workflows/ci.yml#L44) which is what the badge shows the status of. 👍🏻

agnivade:
I think it's useful and I'd be inclined to keep it.

CleanMachine1:
Agreed, I can see the value of it.

PR #7913, opened 2022-03-20
*: normalize colons

SethFalco:

Normalize colons in each language according to their example.

* For French, a space before the colon.
* For Chinese, using a `：` character instead of a `:` character.
* For everything else, removing the space before the colon.


navarroaxel (approved):

pages.zh/common/alias.md
```diff
@@ -24,6 +24,6 @@
 
 `alias {{rm}}="{{rm -i}}"`
 
-- 创建别名`la`来指代`ls -a`:
+- 创建别名`la`来指代`ls -a`：
```
blueskyson, line 27:
```suggestion
- 创建别名 `la` 来指代 `ls -a`：
```

pages.zh/common/gpg2.md
```diff
@@ -8,7 +8,7 @@
 
 `gpg2 --list-keys`
 
-- 为指定的接收者加密指定的文件，将输出结果写到一个新的文件中，并附加`.gpg`:
+- 为指定的接收者加密指定的文件，将输出结果写到一个新的文件中，并附加`.gpg`：
```
blueskyson, line 11:
```suggestion
- 为指定的接收者加密指定的文件，将输出结果写到一个新的文件中，并附加 `.gpg`：
```

blueskyson (approved):

blueskyson:
I have to delete the 5th rule of Chinese copywriting if this PR merged.
```
5. Use a half-width punctuation to end a sentence when the last character is half-width.
For example, use 将代码转化为 Python 3. rather than 将代码转化为 Python 3。
```

navarroaxel:
> I have to delete the 5th rule of Chinese copywriting in [style-guide](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/style-guide.md) if this PR merged. Because all colons are normalized to full-width.
> 
> > 5. Use a half-width punctuation to end a sentence when the last character is half-width.
> >    For example, use `将代码转化为 Python 3.` rather than `将代码转化为 Python 3。`.

@blueskyson which is the correct grammar for this?

blueskyson:
@navarroaxel Commonly, we use full-width punctuations, so the latter one is correct, which also corresponds to this PR.

When I formulated the copywriting rules, all Chinese pages use half-width punctuations when the last character is Latin, so I kept the wrong rule to avoid revision.

navarroaxel:
> @navarroaxel Commonly, we use full-width punctuations, so the latter one is correct, which also corresponds to this PR.
> 
> When I formulated the copywriting rules, all Chinese pages use half-width punctuations when the last character is Latin, so I kept the wrong rule to avoid revision.

@blueskyson Then, I think we should update the rule

PR #7931, opened 2022-03-27
sed: refresh/add pages

EmilySeville7cfg:


EmilySeville7cfg:
@CleanMachine1, CI fails again.

CleanMachine1:
We can ignore it, i think its because the CI is running version of `black` which are affected with the bug.

blueskyson:
The 4th and the 7th example are the same.

pages/common/sed.md
```diff
@@ -1,36 +1,29 @@
...
 
 > Edit text in a scriptable manner.
+> See also: `awk`, `ed`.
 > More information: <https://www.gnu.org/software/sed/manual/sed.html>.
 
-- Replace the first occurrence of a regular expression in each line of a file, and print the result:
+- Execute a specific expression:
 
-`sed 's/{{regular_expression}}/{{replace}}/' {{filename}}`
+`{{command}} | sed '{{s/apple/mango/g}}'`
```
sbrl, line 10: I think the original was much clearer. Let's revert this change.

sbrl (commented):
This is just one example, but changing some of them to specific examples isn't particularly helpful I don't think.

Can you perhaps explain why you think these changes are necessary?

pages/common/sed.md
EmilySeville7cfg: Done ;)

kbdharun (approved):

blueskyson (approved):

PR #7966, opened 2022-04-06
touch: refresh

EmilySeville7cfg:


EmilySeville7cfg:
ash shell for instance [doesn't support](https://unix.stackexchange.com/questions/429298/how-can-i-expand-a-list-of-letters-numbers-in-ash#:~:text=ash%20does%20not%20support%20brace,better%20way%20of%20doing%20it.) brace expansion.

Waples (approved):
lgtm

pages/common/touch.md
```diff
@@ -21,4 +21,4 @@
 
 - Create multiple files:
 
-`touch {{path/to/file{1,2,3}.txt}}`
+`touch {{path/to/file1 path/to/file2 ...}}`
```
navarroaxel, line 24: :thinking: This is a simple command, maybe we add a new example with `file{1,2,3}` to explain that bash feature too. wdyt?

EmilySeville7cfg: It's a good idea to note about shell specific features. 👍🏻 But there is a problem: we have to decide what features are noted about in tldr pages to make pages more standardized. There are too many pages without notion of brace expansion. It is assumed (as I understand) that user adapt suggested syntax to a specific shell. I suggest orient to a POSIX shell and suggest just some of it's features explicitly (as a separate examples or in braces in code descriptions).

[Here][bashisms] you can find bash-specific features. Brace expansion is one of them: and I am not sure we have to stick to Bash shell in some pages. There are other shells too with their specific syntax so it's not clear why not to note about them too. But in this case we are going to make our descriptions very long. 😄 

[bashisms]: https://mywiki.wooledge.org/Bashism

CleanMachine1: I like our current standard for "multiple files", its clear for users to understand and the people can often almost copy the listed commands.

EmilySeville7cfg: Well, what about using brace expansion everywhere for arbitrary number of arguments? For package names, user names, file names and etc. 🤔

navarroaxel (changes requested):
Long arguments are not available in macOS.

pages/common/touch.md
```diff
@@ -1,24 +1,20 @@
 # touch
 
-> Change a file access and modification times (atime, mtime).
+> Creates files and sets access/modification time.
```
CleanMachine1, line 3:
```suggestion
> Create files and set access/modification time.
```

pages/osx/touch.md
```diff
@@ -0,0 +1,20 @@
+# touch
+
+> Creates files and sets access/modification time.
```
CleanMachine1, line 3:
```suggestion
> Create files and set access/modification time.
```

pages/common/touch.md
```diff
@@ -1,24 +1,20 @@
...
+- Update file access/modification time to the current one:
 
-`touch -t {{YYYYMMDDHHMM.SS}} {{path/to/file}}`
+`touch --no-create --time={{access|modify}} {{path/to/file1 path/to/file2 ...}}`
 
-- Set the time on a file to one hour in the past:
+- Update file [t]ime to a specific one:
 
-`touch -d "{{-1 hour}}" {{path/to/file}}`
+`touch --no-create --time={{mode}} -t {{YYYYMMDDHHMM.SS}} {{path/to/file1 path/to/file2 ...}}`
```
marchersimon, line 16: I think `mode` here is pretty confusing.
```suggestion
`touch --no-create --time={{access|modify}} -t {{YYYYMMDDHHMM.SS}} {{path/to/file1 path/to/file2 ...}}`
```

pages/common/touch.md
```diff
@@ -1,24 +1,20 @@
...
+`touch {{path/to/file1 path/to/file2 ...}}`
 
-- Set the times on a file to a specific date and time:
+- Update file access/modification time to the current one:
 
-`touch -t {{YYYYMMDDHHMM.SS}} {{path/to/file}}`
+`touch --no-create --time={{access|modify}} {{path/to/file1 path/to/file2 ...}}`
 
-- Set the time on a file to one hour in the past:
+- Update file [t]ime to a specific one:
```
marchersimon, line 14:
```suggestion
- Set the file [t]ime to a specific value:
```

pages/common/touch.md
```diff
@@ -1,24 +1,20 @@
...
 
-`touch -d "{{-1 hour}}" {{path/to/file}}`
+`touch --no-create --time={{mode}} -t {{YYYYMMDDHHMM.SS}} {{path/to/file1 path/to/file2 ...}}`
 
-- Use the times from a file to set the times on a second file:
+- Update file time to a specific file time:
```
marchersimon, line 18:
```suggestion
- Set file time of a specific file to the time of another file:
```

marchersimon (approved):
LGTM, thanks!

EmilySeville7cfg:
@navarroaxel, please review this PR, I can't merge it.

pages/common/touch.md
```diff
@@ -1,24 +1,20 @@
...
 
-- Use the times from a file to set the times on a second file:
+- Set file time of a specific file to the time of another file:
 
-`touch -r {{path/to/file1}} {{path/to/file2}}`
-
-- Create multiple files:
-
-`touch {{path/to/file{1,2,3}.txt}}`
+`touch --no-create --time={{mode}} --reference={{path/to/file}} {{path/to/file1 path/to/file2 ...}}`
```
navarroaxel, line 20: This way is more clear
```suggestion
`touch --no-create --time={{access|modify}} --reference={{path/to/file}} {{path/to/file1 path/to/file2 ...}}`
```

pages/osx/touch.md
```diff
@@ -0,0 +1,20 @@
...
+
+`touch {{path/to/file1 path/to/file2 ...}}`
+
+- Update file [a]ccess/[m]odification time to the current one:
+
+`touch -c -{{a|m}} {{path/to/file1 path/to/file2 ...}}`
+
+- Update file [t]ime to a specific one:
+
+`touch -c -{{time_mode}} -t {{YYYYMMDDHHMM.SS}} {{path/to/file1 path/to/file2 ...}}`
```
navarroaxel, line 16: This way is more clear
```suggestion
`touch -c -{{a|m}} -t {{YYYYMMDDHHMM.SS}} {{path/to/file1 path/to/file2 ...}}`
```

pages/osx/touch.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Update file time to a specific file time:
+
+`touch -c -{{time_mode}} -r {{path/to/file}} {{path/to/file1 path/to/file2 ...}}`
```
navarroaxel, line 20:
```suggestion
`touch -c -{{a|m}} -r {{path/to/file}} {{path/to/file1 path/to/file2 ...}}`
```

pages/common/touch.md
EmilySeville7cfg: @navarroaxel, should I always inline all values for options? What about to do it only when value count is small such as here (just two choices)?

CleanMachine1 (approved):

pages/osx/touch.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Update file [a]ccess/[m]odification time to the current one:
+
+`touch -c -{{a|m}} {{path/to/file1 path/to/file2 ...}}`
+
+- Update file [t]ime to a specific one:
+
+`touch -c -{{a|m}} -t {{YYYYMMDDHHMM.SS}} {{path/to/file1 path/to/file2 ...}}`
+
+- Update file time to a specific file time:
```
navarroaxel, line 18: Sounds better
```suggestion
- Set file time of a specific file to the time of another file:
```

pages/common/touch.md
marchersimon: This really depends. There are mostly three factors here: The number of options, how intuitive the options are, and how quickly the options can be found in e.g. the man page.

For few options, we can almost always inline them.
If you have many, but intuitive options (like `read`, `write`, `execute`) which everyone knows, we can be generic.
And if the options, can't be easily found somewhere we should also inline them.

But this is highly dependent on the situation. Just think about how it would be the easiest for the user.

EmilySeville7cfg:
@navarroaxel, can you re-review this PR? 😄

navarroaxel:
@EmilySeville7cfg why you removed the osx page?

EmilySeville7cfg:
Because everything is moved to `general/` page. Sometimes you and others allowed me to duplicate pages to different folders to show long and short options and sometimes - not. I am confused.

navarroaxel:
> Because everything is moved to `general/` page. Our specification [defines](https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md?rgh-link-date=2022-09-13T15%3A09%3A22Z#platform) what I should do in the case when some command is cross-platform:

@EmilySeville7cfg I think you mean `common/` but osx doesn't support long arguments for the `touch` command.

EmilySeville7cfg:
Yeah, I mean it. ;) I think we can use mnemonics with square brackets where possible and not long options here.

marchersimon:
We have multiple cases of pages in the `common` directory with a copy in `osx`, which doesn't support long options. I think removing the long options, just so we can remove the `osx` page isn't worth it. There's no harm in keeping both.

EmilySeville7cfg:
Let's note about it in our specification to prevent any confusions. :)

marchersimon:
@EmilySeville7cfg I like your enthusiasm, but I'm not sure if we have to note every little detail in our specifications. In cases like this ask yourself, what yields the best user experience. This is how most of the specification came to be, but this only works up until a point.

There are so many edge cases that we simply cannot put it all into our specification. And if we did, it would be useless, because the essence of it would be buried beneath minor details.

EmilySeville7cfg:
Before you read it sorry for a wall of text. :)

I wanna specification and guides to be comprehensive just because it makes it easier to remind something when you forgot and make everything standardized. I can make mistakes and I wanna to rely on smth, on you for instance, on our guides. The problem with short and long options still not formalized and it's undefined what to do in each case even we try do the best. Exceptions can exist but they must be explicitly allowed anywhere IMO. I know that not everybody likes my ways and enthusiasm to enhance tldr and for somebody it's easier to edit pages without any strict rules but not for me. 🌝 As in coding here I try to follow just strict standards and practices. But I feel that my attempts to edit smth are not always correct. 

I don't know what to do this case and other similar in such cases. I can't decide what's better. There are page duplicates but are they wrong not when the corresponding commands are cross-platform? According to our specification they must be joined to one page. But maybe fix our specification to make such cases right and permitted? 😉

sbrl:
Ref the short vs long options, as I have explained before we generally follow the convention of the way users would use the command. This means short options for commands that, when users encounter them, are likely to be frequently used.

PR #7982, opened 2022-04-12
ksh: refresh page

EmilySeville7cfg:


blueskyson (approved):

pages/common/tcsh.md
```diff
@@ -23,6 +23,10 @@
 
 `tcsh -n {{path/to/script.tcsh}}`
 
+- Execute specific commands from stdin:
```
marchersimon, line 26: This would be more accurate.
```suggestion
- Execute a specific command from `stdin`:
```

pages/common/ksh.md
```diff
@@ -1,21 +1,25 @@
 # ksh
 
 > Korn Shell, a Bash-compatible command-line interpreter.
-> See also `histexpand` for history expansion.
+> See also: `histexpand` (history expansion).
```
marchersimon, line 4: I don't really like this change. What's wrong with the previous sentence?

pages/common/tcsh.md
EmilySeville7cfg: I am not sure whether the first part of suggestion makes page better. Just one command is demonstrated in placeholder for simplicity but in general many commands can be run.

marchersimon: But the goal is to describe what an example does, not what it could do. I think it's a little confusing, when the description doesn't match the example

navarroaxel:
@EmilySeville7cfg please resolve conflicts.

pages/common/ksh.md
kbdharun: I think we don't need to mention `history expansion` at all or we can keep it as the original version itself.

kbdharun (approved):

pages/common/tcsh.md
EmilySeville7cfg: I guess I refresh all shell pages again soon and address this comment. Possible solution: introduce new generic placeholder and make it recognizable by our TlDr page parsers.

Stay tuned. 😄

PR #8012, opened 2022-04-15
chmod: add/refresh pages

EmilySeville7cfg:


pages/common/chmod.md
```diff
@@ -1,36 +1,24 @@
 # chmod
 
-> Change the access permissions of a file or directory.
+> Change file/directory access permissions.
 > More information: <https://www.gnu.org/software/coreutils/chmod>.
 
-- Give the [u]ser who owns a file the right to e[x]ecute it:
+- Add [r]ead/[w]rite/e[x]ecute permissions (`{{r|w|x}}` is referred later as `{{permissions}}`) for a file/directory owner:
```
blueskyson, line 6: I think using `{{r|w|x}}` is better for reading. When you define `((permissions))`, people must read the first example's description carefully so that they understand what `((permissions))` stands for, and which also makes people read more lines. It's not the goal of "too long don't read".
```suggestion
- Add [r]ead/[w]rite/e[x]ecute permissions for a file/directory owner:
```

pages/common/chmod.md
```diff
@@ -1,36 +1,24 @@
...
 
-`chmod u+x {{file}}`
+`chmod +{{r|w|x}} {{path/to/file_or_directory1 path/to/file_or_directory2 ...}}`
 
-- Give the [u]ser rights to [r]ead and [w]rite to a file/directory:
+- Add specific permissions for an owner/users in [g]roup/[o]ther users/[a]ll at once (`{{u|g|o|a}}` is referred later as `{{users}}`):
```
blueskyson, line 10: The same idea.
```suggestion
- Add specific permissions for an owner/users in [g]roup/[o]ther users/[a]ll at once:
```

pages/common/chmod.md
```diff
@@ -1,36 +1,24 @@
...
 
-`chmod u+rw {{file_or_directory}}`
+`chmod {{u|g|o|a}}+{{permissions}} {{path/to/file_or_directory1 path/to/file_or_directory2 ...}}`
 
-- Remove e[x]ecutable rights from the [g]roup:
+- Add/remove/replace specific permissions:
 
-`chmod g-x {{file}}`
+`chmod {{users}}{{+|-|=}}{{permissions}} {{path/to/file_or_directory1 path/to/file_or_directory2 ...}}`
```
blueskyson, line 16:
```suggestion
`chmod {{u|g|o|a}}{{+|-|=}}{{r|w|x}} {{path/to/file_or_directory1 path/to/file_or_directory2 ...}}`
```

pages/common/chmod.md
EmilySeville7cfg: I suggest use first cases of placeholders with alternatives not to repeat all cases:

```md
- Add [r]ead/[w]rite/e[x]ecute permissions for a file/directory owner:

`chmod +{{r|w|x}} {{path/to/file_or_directory1 path/to/file_or_directory2 ...}}`

- Add specific permissions for an owner/users in [g]roup/[o]ther users/[a]ll at once:

`chmod {{u|g|o|a}}+{{r}} {{path/to/file_or_directory1 path/to/file_or_directory2 ...}}`
```

The problem with complete inlining that it's hard to read smth like this:

```md
- Add/remove/replace specific permissions:

`chmod {{u|g|o|a}}{{+|-|=}}{{r|w|x}} {{path/to/file_or_directory1 path/to/file_or_directory2 ...}}`
```

because several placeholders with alternatives are joined.

einverne (dismissed):

pages/common/chmod.md
```diff
@@ -1,36 +1,24 @@
...
+- Add [r]ead/[w]rite/e[x]ecute permissions for a file/directory owner:
 
-`chmod u+x {{file}}`
+`chmod +{{r|w|x}} {{path/to/file_or_directory1 path/to/file_or_directory2 ...}}`
 
-- Give the [u]ser rights to [r]ead and [w]rite to a file/directory:
+- Add specific permissions for an owner/users in [g]roup/[o]ther users/[a]ll at once:
 
-`chmod u+rw {{file_or_directory}}`
+`chmod {{u|g|o|a}}+{{permissions}} {{path/to/file_or_directory1 path/to/file_or_directory2 ...}}`
```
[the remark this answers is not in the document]
EmilySeville7cfg: I've simplified some examples.

pages/common/chmod.md
EmilySeville7cfg: I've used `chmod {{u}}{{+|-|=}}{{r}} {{path/to/file_or_directory1 path/to/file_or_directory2 ...}}` instead.

pages/common/chmod.md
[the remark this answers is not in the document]
einverne: It's clearer now.

pages/common/chmod.md
```diff
@@ -1,36 +1,24 @@
 # chmod
 
-> Change the access permissions of a file or directory.
+> Change file/directory access permissions.
```
sbrl, line 3: I think the old description is more explicit.

pages/common/chmod.md
```diff
@@ -1,36 +1,24 @@
...
 > More information: <https://www.gnu.org/software/coreutils/chmod>.
 
-- Give the [u]ser who owns a file the right to e[x]ecute it:
+- Add [r]ead/[w]rite/e[x]ecute permissions for a file/directory owner:
 
-`chmod u+x {{file}}`
+`chmod +{{r|w|x}} {{path/to/file_or_directory1 path/to/file_or_directory2 ...}}`
```
sbrl, line 8: This is a dangerous example that's equivalent to assigning 0777 permissions to something. This is not something we want to encourage at all.

pages/common/chmod.md
```diff
@@ -1,36 +1,24 @@
...
+- Add/remove/replace specific permissions:
 
-`chmod g-x {{file}}`
+`chmod {{u}}{{+|-|=}}{{r}} {{path/to/file_or_directory1 path/to/file_or_directory2 ...}}`
 
-- Give [a]ll users rights to [r]ead and e[x]ecute:
+- Remove all permissions:
 
-`chmod a+rx {{file}}`
+`chmod {{u}}= {{path/to/file_or_directory1 path/to/file_or_directory2 ...}}`
```
sbrl, line 20: When would you ever want to execute this? I recommend at least changing `u` to `o`...

pages/common/chmod.md
```diff
@@ -1,36 +1,24 @@
...
-`chmod o= {{file}}`
-
-- Change permissions recursively giving [g]roup and [o]thers the ability to [w]rite:
-
-`chmod -R g+w,o+w {{directory}}`
-
-- Recursively give [a]ll users [r]ead permissions to files and e[X]ecute permissions to sub-directories within a directory:
-
-`chmod -R a+rX {{directory}}`
+`chmod --recursive {{u}}{{+}}{{r}} {{path/to/file_or_directory1 path/to/file_or_directory2 ...}}`
```
sbrl, line 24: Using long form options is good, but in this instance I feel since this is a frequently used command like `ls` we should use short options to reduce typing.

sbrl (commented):
Thanks for the edits! I've left some comments below for you to review.

pages/common/chmod.md
EmilySeville7cfg: I think it's better to split page here into two ones: with short options and long options.

pages/common/chmod.md
EmilySeville7cfg: It is not clear for me what to consider dangerous and what not. I can consider all commands requiring `sudo` as dangerous ones but `chmod` is not a case by this logic.

pages/common/chmod.md
EmilySeville7cfg: Yes, thanks. 👍🏻 Good idea.

pages/common/chmod.md
sbrl: Right. In this case, I mean specifically that `0777` permissions are widely considered bad practice, as it allows *anyone* to write to and change a file. By including it as an example in a tldr page, we encourage it. Resources on this issue:

 - https://pimylifeup.com/chmod-777/
 - https://www.anchor.com.au/blog/2012/09/the-chmod-777-trap-how-and-why-to-avoid-it/

pages/common/chmod.md
sbrl: We don't currently have provision for multiple versions of pages like that. Let's use short options on this page, as `chmod` is often used on a regular basis, similar to `ls`.

To be more specific here, ref the guidelines we don't necessarily need to use long options *all* the time. See also the `tar` command's page for further inspiration of how short options can be demonstrated effectively.

EmilySeville7cfg: But there are already [several PRs](https://github.com/tldr-pages/tldr/pulls/EmilySeville7cfg) doing so with `add` and `refresh` words in the titles. It was not my idea to split pages with long and short options, @pixelcmtd suggested it. But I like it. I am planning to refresh contribution guidelines and explicitly write this rule. There are just general pages which can use both long and short options interchangeably.

pages/common/chmod.md
sbrl: By "more explicit", I mean it has greater clarity.

EmilySeville7cfg: Done 😄

pages/common/chmod.md
sbrl: I'm not sure what having `add` and `refresh` in the title of a PR has anything to do with this issue.

I'm not even sure what you mean by "split pages with long and short options", or how it would even be implemented without either:

1. Breaking clients
2. Causing confusion
3. Resulting in the additional extra work of maintaining multiple copies of a page.

pages/linux/chmod.md
```diff
@@ -0,0 +1,24 @@
+# chmod
```
sbrl, line 1: Wait a moment. I didn't notice this before, but there seems to be a copy of the page in the `chmod` directory. Is this intentional?

sbrl: haha yeah, I do mean the `common` directory lol

Ah, I see. If they are the same except for the last example, it makes sense to merge them and go with the common syntax to reduce maintenance complexity.

pages/common/chmod.md
EmilySeville7cfg: > I'm not even sure what you mean by "split pages with long and short options", or how it would even be implemented without either:
>
>   1. Breaking clients
>   1.  Causing confusion
>   1.  Resulting in the additional extra work of maintaining multiple copies of a page.

When clients can be confused? Current specification mandates clients to have --platform option to point to a specific platform with a particular page.

Yeah, it is an extra work now. We can solve this issue with adding support for switching between long and short options via smth like this: `tldr --long-options` and pages with such placeholders:

```
command {{{-s|--long}}}
```

I have no idea about what syntax is the best. We have to discuss it.

pages/linux/chmod.md
EmilySeville7cfg: > Wait a moment. I didn't notice this before, but there seems to be a copy of the page in the chmod directory. Is this intentional?

Yes, it is intentional. ;)

marchersimon: But I'd assume that osx also accepts long options. So maybe we should move the `common` page with `-R` to `windows` and this one to `common`.

EmilySeville7cfg: It depends on what pages can be placed in `common/`. Pages for commands that are available just on some subset of platforms (Windows, Mac OS, Linux) or commands that are available everywhere?

marchersimon: There are almost no commands, which are available on all platforms (e.g. Android). Typically if a command is available on at least two platforms it goes to `common`.

EmilySeville7cfg: Conversation is being closed as it's okay to duplicate pages to show different short and long options.

[Rules for long options](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/style-guide.md#option-syntax)
[When page duplicates can be created](https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md#platform)

kbdharun (approved):

pages/common/chmod.md
EmilySeville7cfg: > Using long form options is good, but in this instance I feel since this is a frequently used command like ls we should use short options to reduce typing.

Style guide now [requires](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/style-guide.md#option-syntax) long portable options, and it's possible to duplicate pages to show their different options's counterparts, it's valid. Issue is resolved.

EmilySeville7cfg:
Closing PR, to make somebody pay attention to it, review and merge.

PR #8025, opened 2022-04-16
find: add/refresh pages

EmilySeville7cfg:


EmilySeville7cfg:
@navarroaxel, @mfrw, @waldyrious, please review this PR. 😄

pages/common/find.md
```diff
@@ -1,36 +1,29 @@
 # find
 
 > Find files or directories under the given directory tree, recursively.
-> More information: <https://manned.org/find>.
+> See also: `xargs`.
+> More information: <https://manned.org/man/freebsd-13.0/find.1>.
 
-- Find files by extension:
+- Search files by a specific case sensitive glob pattern/case insensitive glob pattern/case sensitive regular expression/case insensitive regular expression:
```
navarroaxel, line 7: Sounds a little hard to understand

EmilySeville7cfg: I think we can limit allowed placeholder count in code examples by linter.

EmilySeville7cfg: 1. Another problem exists here: there are two pages: in `linux/` and in `common/`. I can move the page from `common/` to `osx/`. But this command is also [available](http://gnuwin32.sourceforge.net/packages/findutils.htm) in Windows. I can also write `find.md` in `windows/` but there are too many page duplicates. And it hard to support them all. Another decision is to leave just `common/find.md` with the most used options but then we can't always use long options because `common/` pages prefer the most portable options. Personally I keen on short options for fast typing but long ones for memorizing because it's hard to remember option case.
1. There is another question. MacOS doesn's provide GNU find by default. Should we always prefer preinstalled commands on the latest OS version?
1. And the third question. If some command is not preinstalled by default should we describe it at all for this OS? For instance GNU find is not installed by default on Windows.

CleanMachine1: does
```suggestion
- Search files by a specific search pattern:
```
not work here?

EmilySeville7cfg: Sure it's simpler but the user will not understand what to use of the listed alternatives: `{{name|iname|regex|iregex}}`.

sbrl:
Waaaiiiiit a moment. Before I review the rest of this Pull Request, why is there a duplicate of the `find` command in the `linux` platform here? Are there some specific differences that we can't avoid documenting?

Also, please do not @mention @ waldyrious unless you really really really have to. They are very busy, and do not usually have time to donate to tldr-pages.

EmilySeville7cfg:
> Also, please do not @mention @ waldyrious unless you really really really have to. They are very busy, and do not usually have time to donate to tldr-pages.

Sorry.

EmilySeville7cfg:
> Waaaiiiiit a moment. Before I review the rest of this Pull Request, why is there a duplicate of the find command in the linux platform here? Are there some specific differences that we can't avoid documenting?

`find` in common describes FreeBSD version of this command without for instance `-regextype` option. I think it's my [mistake](https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md#platform):

> The special platform common contains pages for commands that work identically across more than one platform

Note that MacOS doesn's provide `GNU find` by default. But we can install GNU find almost everywhere. Should we always prefer preinstalled commands on the latest OS version? What to do with pages that already exist and describe the same named commands as preinstalled ones? Should we replace such pages with preinstalled command pages?

EmilySeville7cfg:
Closing PR, to make somebody pay attention to it, review and merge.

PR #8068, opened 2022-04-30
ci.yml, stale.yml: update runtime to node16

navarroaxel:
Update stale action to v5.
Update actions to use node16 runtime instead of node12.


mfrw (approved):
LGTM :)
Thanks

EmilySeville7cfg (approved):
👍

PR #8069, opened 2022-05-01
prettier: add --write example for js and ts files

AMATEURTOSS:

Added an example to format only for specific file extensions. I think this might be useful

### DIFF
--- a/pages/common/prettier.md
+++ b/pages/common/prettier.md
@@ -22,3 +22,7 @@
 - Format files or directories recursively using single quotes and no trailing commas:
 
 `prettier --single-quote --trailing-comma {{none}} --write {{path/to/file_or_directory}}`
+
+- Format files with matching extensions:
+
+`prettier --write "**/*.{js,jsx,ts,tsx}"`
--- a/pages/common/prettier.md
+++ b/pages/common/prettier.md
@@ -23,6 +23,6 @@
 
 `prettier --single-quote --trailing-comma {{none}} --write {{path/to/file_or_directory}}`
 
-- Format files with matching extensions:
+- Format JavaScript and TypeScript files recursively, replacing the original:
 
 `prettier --write "**/*.{js,jsx,ts,tsx}"`


pages/common/prettier.md
```diff
@@ -22,3 +22,7 @@
 - Format files or directories recursively using single quotes and no trailing commas:
 
 `prettier --single-quote --trailing-comma {{none}} --write {{path/to/file_or_directory}}`
+
+- Format files with matching extensions:
```
CleanMachine1, line 26: Is this in the current directory.

AMATEURTOSS: `**/*.{js,jsx,ts,tsx}` contain
* `foo.js`
* `bar.ts`
* `foo/bar.js`
* `foo/bar/baz.tsx`

Match whatever is in the current directory

navarroaxel: I think this is an specific usage of the 4th example
```suggestion
- Format JavaScript and TypeScript files recursively, replacing the original:
```

pages/common/prettier.md
```diff
@@ -22,3 +22,7 @@
 - Format files or directories recursively using single quotes and no trailing commas:
 
 `prettier --single-quote --trailing-comma {{none}} --write {{path/to/file_or_directory}}`
+
+- Format JavaScript and TypeScript files recursively, replacing the original:
+
+`prettier --write "**/*.{js,jsx,ts,tsx}"`
```
CleanMachine1, line 28:
```suggestion
`prettier --write "**/*.{{js,jsx,ts,tsx}}"`
```

CleanMachine1 (commented):
Not sure if this is correct.

pages/common/prettier.md
AMATEURTOSS: `prettier --write "**/*.{js,jsx,ts,tsx}"` is whole command. 
if you highlight it, you can highlight `prettier --write "**/*.{{{js,jsx,ts,tsx}}}"` like this

CleanMachine1 (approved):

navarroaxel (approved):

PR #7975, opened 2022-04-11
cpupower: add page

toydotgame:
- [x] The page (if new), does not already exist in the repository.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).
- [x] The page description includes a link to documentation or a homepage (if applicable).

**Version of the command being documented (if known):** 5.17-1

PR #8073, opened 2022-05-02
cpupower: add page

toydotgame:

**Version of the command being documented (if known):** 5.17-1

***

Last PR (#7975) got borked by Git. Starting again.

### DIFF
--- a/pages/linux/cpupower.md
+++ b/pages/linux/cpupower.md
@@ -1,7 +1,7 @@
 # cpupower
 
 > Tools regarding CPU power and tuning options.
-> This command is available as part of the `cpupower` package, or on Fedora as part of `kernel-tools`.
+> This command is available as part of the `cpupower` package, or as part of `kernel-tools` on Fedora.
 > More information: <https://manned.org/cpupower>.
 
 - List CPUs:
--- a/pages/linux/cpupower.md
+++ b/pages/linux/cpupower.md
@@ -6,7 +6,7 @@
 
 - List CPUs:
 
-`sudo cpupower --cpu cpulist`
+`sudo cpupower --cpu {{all}} info`
 
 - Print information about all cores:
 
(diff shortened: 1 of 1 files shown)


pages/linux/cpupower.md
```diff
@@ -0,0 +1,25 @@
+# cpupower
+
+> Tools regarding CPU power and tuning options.
+> This command is available as part of the `cpupower` package, or on Fedora as part of `kernel-tools`.
```
navarroaxel, line 4: This is better for non-native English speakers, I think so.
```suggestion
> This command is available as part of the `cpupower` package, or as part of `kernel-tools` on Fedora.
```

navarroaxel (commented):
It's not a good practice send PRs from your `main`. You should keep your `main` in sync with the `tldr-pages:main`, then create feature branches and send PRs from them. Then you have a clean branch to create more PRs. Let as-is for now. you can change this after merge this PR.

pages/linux/cpupower.md
```diff
@@ -0,0 +1,25 @@
...
+> More information: <https://manned.org/cpupower>.
+
+- List CPUs:
+
+`sudo cpupower --cpu cpulist`
```
marchersimon, line 9: Does this command work for you as is? I get an `Error parsing cpu list`. From the man page it seems that `cpulist` would have to be replaced with something like `all`, but then it needs another subcommand like `info`.

toydotgame: That command is _meant_ to work (going off what the manpages said), but it does seem to break a lot on many installs. I suggested `sudo cpupower -c all info` which indirectly serves the same purpose.
If it's breaking for everyone that's trying it, I'll just go and change it.

Waples (approved):

navarroaxel (approved):

PR #8098, opened 2022-05-19
tcpdump: update

juejun:


mfrw (commented):
@juejun Thank you for your contribution.
We already have 8 examples for this page. Can we please prioritize and have only the most useful 8.
Also, please could you sign the CLA :)

einverne (changes requested):
@juejun thank you for the new example, but there is more than 8 examples here, which one do you think is more important? We can only accept one.

PR #8104, opened 2022-05-22
hydra: expand example description for HTTPS webform credentials

fazlearefin:

### DIFF
--- a/pages/common/hydra.md
+++ b/pages/common/hydra.md
@@ -12,7 +12,7 @@
 
 `hydra -l {{username}} -P {{path/to/wordlist.txt}} {{host_ip}} {{ssh}}`
 
-- Guess HTTPS webform credentials using a list of usernames and a list of passwords:
+- Guess HTTPS webform credentials using a list of usernames and a list of passwords (`{{https_post_request}}` can be like `username=^USER^&password=^PASS^`):
 
 `hydra -L {{path/to/usernames.txt}} -P {{path/to/wordlist.txt}} {{host_ip}} {{https-post-form}} "{{url_without_host}}:{{https_post_request}}:{{login_failed_string}}"`
 
--- a/pages/common/hydra.md
+++ b/pages/common/hydra.md
@@ -12,7 +12,7 @@
 
 `hydra -l {{username}} -P {{path/to/wordlist.txt}} {{host_ip}} {{ssh}}`
 
-- Guess HTTPS webform credentials using two specific lists of usernames and passwords (`{{https_post_request}}` can be like `username=^USER^&password=^PASS^`):
+- Guess HTTPS webform credentials using two specific lists of usernames and passwords ("https_post_request" can be like "username=^USER^&password=^PASS^"):
 
 `hydra -L {{path/to/usernames.txt}} -P {{path/to/wordlist.txt}} {{host_ip}} {{https-post-form}} "{{url_without_host}}:{{https_post_request}}:{{login_failed_string}}"`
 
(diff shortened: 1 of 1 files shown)


pages/common/hydra.md
```diff
@@ -12,7 +12,7 @@
 
 `hydra -l {{username}} -P {{path/to/wordlist.txt}} {{host_ip}} {{ssh}}`
 
-- Guess HTTPS webform credentials using a list of usernames and a list of passwords:
+- Guess HTTPS webform credentials using a list of usernames and a list of passwords (`{{https_post_request}}` can be like `username=^USER^&password=^PASS^`):
```
navarroaxel, line 15: Did you try this in a client with `--render`? I think this token syntax is not valid. It doesn't work in my tldr client made in Rust.

navarroaxel (changes requested):

pages/common/hydra.md
fazlearefin: I use a different client: *Haskell tldr client*. It works perfectly fine with my client. But yes, with the *Rust tealdeer client*, I see formatting issues. Given that the markdown itself is correct, don't you think this is a bug in the rust client itself? 

Let me know what you think and I will try to find an alternative solution.

(I have attached a screenshot comparing the Haskell and Rust clients) 
hydra

fazlearefin: I just tried with the official [tldr client built with python](https://github.com/tldr-pages/tldr-python-client) and I do see the same result as the rust client. Confused now.

EmilySeville7cfg, line 15:
```suggestion
- Guess HTTPS webform credentials using two specific lists of usernames and passwords (`{{https_post_request}}` can be like `username=^USER^&password=^PASS^`):
```

EmilySeville7cfg (approved):
Everything looks nice. Just the possible rewording. ;)

pages/common/hydra.md
navarroaxel: @dbrgn could you please give us your opinion? is it a bug?

dbrgn: In my opinion it is not a bug, because the [original specification](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/style-guide.md) does not indicate that descriptions can contain tokens. In the case of tealdeer, the client does not implement full CommonMark support, because that would be *way* more complex (and could never be fully supported because any HTML is also valid CommonMark, and you cannot render HTML in the console).

The [current version of the client spec](https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md#page-structure) says:

> Although this specification is about the interface that clients must provide, it is also worth noting that pages are written in standard [CommonMark](https://commonmark.org/), which the exception of the non-standard `{{` and `}}` syntax, which surrounds values in an example that users may edit. Clients MUST NOT break if the page format is changed within the _CommonMark_ specification.

Tealdeer does not break when CommonMark is used in the description, but it is simply not rendered. This is in line with the spec.

I would suggest to avoid any type of formatting in descriptions.

navarroaxel: @dbrgn but there are some pages with highlights in the example description (`common/csvcut.md`, `common/strings.md` and `common/supervisorctl.md`) and the italic is not rendered in the terminal.

fazlearefin:
@dbrgn @navarroaxel 
It was my bad. I has using a very old version of Haskell implementation of tldr that comes bundled with Ubuntu. I am now using the latest official python implementation of tldr. I have now fixed the formatting and it should be all good.

navarroaxel (approved):

PR #8141, opened 2022-06-20
rsync: add examples

smzm:


navarroaxel:
Hi @smzm The page can only have up to 8 examples, do you think any old example should be updated? Remember that this command page is a quick start, we don't look for documenting all the options in a such complex command like this.

PR #8146, opened 2022-06-22
warp-cli: add systemctl command to start WARP service.

srt0:
added a new command

### DIFF
--- a/pages/common/warp-cli.md
+++ b/pages/common/warp-cli.md
@@ -3,6 +3,10 @@
 > Official command-line client for Cloudflare's WARP service.
 > More information: <https://developers.cloudflare.com/warp-client/>.
 
+- Start WARP service if it is not already enabled:
+
+`sudo systemctl start warp-svc`
+
 - Register the current device to WARP (must be run before first connection):
 
 `warp-cli register`


pages/common/warp-cli.md
```diff
@@ -3,6 +3,10 @@
 > Official command-line client for Cloudflare's WARP service.
 > More information: <https://developers.cloudflare.com/warp-client/>.
 
+- Start WARP service if it is not already enabled:
+
+`sudo systemctl start warp-svc`
```
navarroaxel, line 8: We document the command itself, we don't use to include instruction about daemon manipulation.

navarroaxel (changes requested):

pages/common/warp-cli.md
srt0: But it is prerequisite. Would help someone who is looking for **warp-cli** command using **tldr**.

CleanMachine1: I don't think this can be added as well. we document the command usage ,not help for a particular tool

navarroaxel: @sbrl?

sbrl: Generally we do not document features of other commands in a tldr page.

reinhart1010: > But it is prerequisite. Would help someone who is looking for **warp-cli** command using **tldr**. Someone who is not familiar with daemon.

`warp-cli` itself is available on both macOS and Linux, as far as I remembered when I wrote this original page. However, `systemctl` is only specific to Linux, and specifically for those who opt for using [Systemd](https://en.wikipedia.org/wiki/Systemd) instead of [SysV init](https://en.wikipedia.org/wiki/Init#SysV-style), [Upstart](https://en.wikipedia.org/wiki/Upstart_(software)) and others.

This means that folks wishing to use `warp-cli` from Linux with a different init system than Systemd (e.g. [Devuan](https://www.devuan.org)) still can't use your command to start WARP in the background. Not to mention those macOS users, too.

PR #8148, opened 2022-06-23
calc: move from `linux` to `common`

marchersimon:
`calc` is available on pretty much anything


mfrw (commented):
I tried running the example:
`calc -p '85 * (36 / 4)'` on windows but it only opened the GUI calc.
Although, this is a nitpick, I wanted to put it out there for thoughts.
IMO it should be self explanatory ..

marchersimon:
Well, at least it works. I looked for something about a GUI in the `README.WINDOWS` file, but couldn't find anything.
However, it probably still behaves the same on OSX, and therefore should still go to `common`

navarroaxel:
@mfrw but we don't have a specific page for the Windows' calculator app, this is for a C program that can be installed on macOS or Windows itself. https://github.com/lcn2/calc/blob/master/HOWTO.INSTALL

mfrw (approved):
Given our discussion, we should move it :)

navarroaxel (approved):

PR #8155, opened 2022-06-26
*: convert en-GB to en-US

SethFalco:
- N/A The page (if new), does not already exist in the repository.
- N/A The page is in the correct platform directory (`common/`, `linux/`, etc.)
- N/A The page has 8 or fewer examples.
- N/A The page description includes a link to documentation or a homepage (if applicable).

N/A

---

Mostly just converts some British English (en-GB) to American English (en-US), I've had these stashed for so long. ^-^' Just thought I'd finally commit it.

There are 2 other minor changes:
* ssh → SSH
* DNS Lookup utility. → DNS lookup utility.


pages/common/stat.md
```diff
@@ -11,7 +11,7 @@
 
 `stat -t {{file}}`
 
-- Show filesystem information:
+- Show file system information:
```
navarroaxel, line 14: We use `filesystem` along the whole doc pages
```suggestion
- Show filesystem information:
```

SethFalco: Ahh that was my bad.
A while back I changed `file system` to `filesystem` thinking `file system` was wrong.
When undoing it, I must've missed that one.

owenvoke (approved):
LGTM, better for consistency I guess. Goodbye my poor, lovely, correct English. 😭 😁

CleanMachine1 (approved):

navarroaxel (approved):

sbrl:
I don't feel this was really necessary - but there's no harm in it since it's already merged I guess.

pages/common/opt.md
```diff
@@ -1,6 +1,6 @@
 # opt
 
-> A tool that takes LLVM source files and runs specified optimizations and/or analyses on them.
+> A tool that takes LLVM source files and runs specified optimizations and/or analysis on them.
```
sbrl, line 3: Nope, I'm not sure this is right. `analyses` is plural, but `analysis` is singular.

pages/linux/dumpe2fs.md
```diff
@@ -12,7 +12,7 @@
 
 `dumpe2fs -b {{/dev/sdXN}}`
 
-- Force display filesystem information even with non-recognisable feature flags:
+- Force display filesystem information even with unrecognizable feature flags:
```
sbrl, line 15: `unrecognised` would fit better here.

sbrl (commented):
Some issues with you find and replace.

pages/common/opt.md
SethFalco: Ahh, actually that one wasn't a find and replace.
I'm just an idiot and read it like "analyze on them" before which, which is wrong.

I'll fix that later!

SethFalco:
> The english language variant we use is standardised to en_US yes, but I'm not sure there's enormous value in strictly enforcing this on existing pages.

I think it's nice to provide consistency.

It's like when you read a book, no one likes when terms are spelled differently throughout the book. Plus the vast majority of tools use American English, so I think it makes sense to people used to a particular spelling.

For example, if someone reads `color` as `colour` a lot, they might get very confused when `command --colour white` isn't working later when they do commands from memory.

Not everyone knows about these differences in language, especially among non-natives, and among people that do know the difference the inconsistency is annoying as hell.

PR #8173, opened 2022-07-01
psql: adds basic Postgres examples

redreceipt:
helpful for a very basic Postgres cheatsheet

### DIFF
--- a/pages/common/psql.md
+++ b/pages/common/psql.md
@@ -15,10 +15,22 @@
 
 `psql -h {{host}} -p {{port}} -U {{username}} -W {{database}}`
 
-- Execute a single SQL query or PostgreSQL command on the given database (useful in shell scripts):
-
-`psql -c '{{query}}' {{database}}`
-
 - Execute commands from a file on the given database:
 
 `psql {{database}} -f {{file.sql}}`
+
+- Select all items in a given table:
+
+`psql -c 'SELECT * FROM {{table}};' {{database}}`
+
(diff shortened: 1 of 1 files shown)


pages/common/psql.md
```diff
@@ -15,10 +15,22 @@
...
 
 `psql {{database}} -f {{file.sql}}`
+
+- Select all items in a given table:
+
+`psql -c 'SELECT * FROM {{table}};' {{database}}`
+
+- Update all items under certain conditions:
+
+`psql -c 'UPDATE {{table}} SET ({{column1}}, {{column2}}) = ({{value1}}, {{value2}}) WHERE {{condition}};' {{database}}`
```
navarroaxel, line 28: But you are explaining SQL not the command options :confused:

redreceipt: I know 😬 I found myself wishing I had a cheat sheet for the Postgres insert statement somewhere and I love TLDR so I figured this might be ok.

blueskyson: I think `psql -c 'SELECT * FROM {{table}};' {{database}}` looks good, but the `UPDATE` one looks too complex. So I would prefer to keep the `SELECT` example and delete the `UPDATE` example.

redreceipt: That's ok. I was trying to solve a different problem than what TLDR is designed for. I'll find a better spot for my Postgres cheat sheet :) thanks guys

PR #8177, opened 2022-07-02
invoke-webrequest: add page

reinhart1010:

**Version of the command being documented (if known):** 7.2.1

After quite a long hiatus I decided to contribute some more on PowerShell (and soon for Indonesian translations for macOS commands).

We have talked a lot about `Invoke-WebRequest` and ["incompatible aliases" when `curl` and `wget` are not installed in Windows](https://github.com/tldr-pages/tldr/issues/6621#issue-1012843699). So I decided to create a special alias page for them, with a command to check which program (`curl`/`wget` or `Invoke-WebRequest`) was used to evaluate the `curl` or `wget` command, despite I'm afraid they won't pass the linter:

> Check whether `wget` is properly installed by printing its version number. If this command evaluates into an error, PowerShell may have substituted this command with `Invoke-WebRequest`


pages/windows/curl.md
```diff
@@ -0,0 +1,15 @@
+# curl
```
sbrl, line 1: This will overwrite the existing `curl` page in the `common` platform.

pages/windows/wget.md
```diff
@@ -0,0 +1,15 @@
+# wget
```
sbrl, line 1: Same here. Let's remove these `wget` and `curl` pages from this pull request.

sbrl (changes requested):

pages/windows/wget.md
reinhart1010: I wrote these two pages for two reasons. First is the fact that in PowerShell for Windows, both commands will be substituted with `Invoke-WebRequest` with incompatible set of accepted parameters when `wget` or `curl` is not visible on `%PATH%`:

image

and similarly, executing `curl` and `wget` without additional subcommand or parameters will still return the same prompt with `Invoke-WebRequest` if both aren't installed on `%PATH%`:

image

As far as I know there are some edge cases where a command documentation is split between `common` and OS-specific versions, e.g. `common/cd` and `windows/cd`. As long as `tldr` clients still can show the common or platform-specific versions of the same command I still prefer to keep `windows/curl` and `windows/wget` this way to avoid confusion among those PowerShell users.

Ah I forgot that the original documentation should be access with `tldr {{curl|wget}} --platform common` instead of `tldr common/{{curl|wget}}`.

marchersimon (approved):
Not entirely sure about the `curl`/`wget`page, but the content looks fine.
It's not the most elegant solution, but I think the only option we have.

pages/windows/wget.md
sbrl: Oh I see, thanks for clarifying this. This is a rather confusing situation, isn't it?

 - `cmd.exe` is the main shell for Windows, which if you have `curl` or `wget` installed works as on Linux
 - PowerShell is also an official shell distributed and enabled by default on Windows, where `curl` and `wget` resolve → `Invoke-WebRequest`

With this in mind, your solution here seems to be the best option. Thanks for explaining - not being a Windows user I didn't understand at first.

sbrl (approved):

PR #8179, opened 2022-07-03
zsh: add noglob example

marchersimon:
I'm not sure if this should be in the `zsh`-page, but adding a new `noglob.md` page didn't feel right either.

I'm also not sure if we should maybe use a more specific example, like `noglob calc 3*sqrt(2)`.

### DIFF
--- a/pages/common/zsh.md
+++ b/pages/common/zsh.md
@@ -24,6 +24,6 @@
 
 `zsh --verbose`
 
-- Execute a command inside `zsh` with disabled glob patterns:
+- Execute a specific command inside `zsh` with disabled glob patterns:
 
 `noglob {{command}}`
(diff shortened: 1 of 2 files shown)


pages/common/zsh.md
```diff
@@ -23,3 +23,7 @@
 - Start an interactive shell session in verbose mode, printing each command before executing it:
 
 `zsh --verbose`
+
+- Execute a command inside `zsh` with disabled glob patterns:
```
mfrw, line 27: Since we've been using `specific` .. how about:
```suggestion
- Execute a specific command inside `zsh` with disabled glob patterns:
```

mfrw (approved):

blueskyson (approved):
Though it looks strange to me, this is acceptable in some cases like `j` and `jo` of `autojump`.

PR #2796, opened 2019-02-22
pages.zh/osx: add Chinese translation

wizarot:
- [x] The page (if new), does not already exist in the repo.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/master/CONTRIBUTING.md#guidelines).

PR #8190, opened 2022-07-06
brew-mas: remove (wrong) Chinese translation

marchersimon:
This page doesn't have an english counterpart. Here's what happened:

- In #2796 `pages.zh/osx/brew-mas.md` was added
- In #3032 `brew-mas` was renamed to `mas` (but not the `zh` version)
- In #7429 a new translation for `mas` was added

So the page I'm deleting here should've been moved in #3032, but has new been replaced anyways.


blueskyson (approved):
Look good to me.

navarroaxel (approved):

Issue #8018, opened 2022-04-16
Page formats for different programs


EmilyGraceSeville7cf:
> ⚠️  These page templates are going to be transferred to style guide after a discussion.

# Description

It is an attempt to standardize as many as possible pages. This issue suggests how to write pages for different programs.

- Any of suggested code examples must be removed with preserved order if the corresponding program doesn't
support suggested syntax.
- Instead of `...` at the end of a page template you can write any examples in any order.

## Screenshot tools

> This template applies for programs such as: `mate-screenshot`, `flameshot`, etc.

Code example descriptions:

1. `Create a fullscreen screenshot`
1. `Create an active window screenshot`
1. `Create a specific area screenshot`
1. `Create a screenshot interactively`
1. Other examples

## Text/graphics/modelling editors

> This template applies for programs such as: `krita`, `pluma`, etc.

Code example descriptions:

1. `Start {{official_editor_name}}`
1. `Open specific files`/`Open specific directories`/`Open specific files/directories:`
1. `Open specific files in a new window`/`Open specific directories in a new window`/`Open specific files/directories in a new window`
1. `Compare two specific files`
1. Other examples

## Shells

> This template applies for programs such as: `bash`, `fish`, etc.

1. `Start an interactive shell session`
1. `Start an interactive shell session without loading startup configs`
1. `Execute specific commands`
1. `Execute a specific script`
1. `Check a specific script for syntax errors`
1. Other examples

Managor:
Closing this as this has seen no interest.

PR #8205, opened 2022-07-13
cot: add page

superatomic:

**Version of the command being documented (if known):** `2.9.2`

### DIFF
--- a/pages/osx/cot.md
+++ b/pages/osx/cot.md
@@ -15,7 +15,7 @@
 
 `cot --new`
 
-- Open a specific file (blocking the terminal until the editor is closed):
+- Open a specific file and block the terminal until it is closed:
 
 `cot --wait {{path/to/file}}`
 
--- a/pages/osx/cot.md
+++ b/pages/osx/cot.md
@@ -7,15 +7,15 @@
 
 `cot`
 
-- Open specific files:
+- Open specific files in the background:
 
 `cot {{path/to/file1 path/to/file2 ...}}`
 
-- Open a new blank document:
+- Open a new blank document in the background:
 
 `cot --new`
 
-- Open a specific file in the foreground (blocking the terminal until the editor is closed):
+- Open a specific file (blocking the terminal until the editor is closed):
 
 `cot --wait {{path/to/file}}`
 
--- a/pages/osx/cot.md
+++ b/pages/osx/cot.md
@@ -7,13 +7,13 @@
 
 `cot`
 
-- Open a new blank document:
-
-`cot --new`
-
 - Open specific files:
 
 `cot {{path/to/file1 path/to/file2 ...}}`
+
+- Open a new blank document:
+
+`cot --new`
 
 - Open a specific file in the foreground (blocking the terminal until the editor is closed):
 
--- a/pages/osx/cot.md
+++ b/pages/osx/cot.md
@@ -15,7 +15,7 @@
 
 `cot {{path/to/file1 path/to/file2 ...}}`
 
-- Open and wait until done editing a specific file:
+- Open and wait until finished editing a specific file:
 
 `cot --wait {{path/to/file}}`
 
--- a/pages/osx/cot.md
+++ b/pages/osx/cot.md
@@ -7,11 +7,11 @@
 
 `cot`
 
-- Open specific files in the background:
+- Open specific files:
 
 `cot {{path/to/file1 path/to/file2 ...}}`
 
-- Open a new blank document in the background:
+- Open a new blank document:
 
 `cot --new`
 
--- a/pages/osx/cot.md
+++ b/pages/osx/cot.md
@@ -1,0 +1,24 @@
+# cot
+
+> The Plain-Text Editor for macOS.
+> More information: <https://coteditor.com/>.
+
+- Start CotEditor:
+
+`cot`
+
+- Open a new blank document:
+
+`cot --new`
+
+- Open specific files:
+
+`cot {{path/to/file1 path/to/file2 ...}}`
+
+- Open and wait until done editing a specific file:
+
+`code --wait {{path/to/file}}`
+
+- Open a file with the cursor at a specific line and column:
+
+`cot --line {{line_number}} --column {{column_number}} {{path/to/file}}`
--- a/pages/osx/cot.md
+++ b/pages/osx/cot.md
@@ -15,7 +15,7 @@
 
 `cot {{path/to/file1 path/to/file2 ...}}`
 
-- Open and wait until finished editing a specific file:
+- Open a specific file in the foreground (blocking the terminal until the editor is closed):
 
 `cot --wait {{path/to/file}}`
 
--- a/pages/osx/cot.md
+++ b/pages/osx/cot.md
@@ -19,6 +19,6 @@
 
 `cot --wait {{path/to/file}}`
 
-- Open a file with the cursor at a specific line and column:
+- Open a specific file with the cursor at a specific line and column:
 
 `cot --line {{line_number}} --column {{column_number}} {{path/to/file}}`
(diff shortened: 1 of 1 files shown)


pages/osx/cot.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Open a new blank document:
+
+`cot --new`
+
+- Open specific files:
+
+`cot {{path/to/file1 path/to/file2 ...}}`
+
+- Open and wait until done editing a specific file:
```
einverne, line 18:
```suggestion
- Open and wait until finishing editing a specific file:
```

einverne (changes requested):
Same here.

pages/osx/cot.md
superatomic: Here's my reply from [the other PR](https://github.com/tldr-pages/tldr/pull/8206) for reference:
> This exact line was taken directly from [the `webstorm` tldr-page](https://github.com/tldr-pages/tldr/blob/da55f621eb9c77d579b4353d78cc4c9bec39c4f3/pages/common/webstorm.md), but I agree that "finishing" is better than "done".
> 
> I think "finished" might be a better form of the word than "finishing" to use here, though.

einverne (approved):

pages/osx/cot.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Open a new blank document:
+
+`cot --new`
+
+- Open specific files:
+
+`cot {{path/to/file1 path/to/file2 ...}}`
+
+- Open and wait until finished editing a specific file:
```
marchersimon, line 18:
```suggestion
- Open a specific file in the foreground (blocking the terminal until the editor is closed):
```

pages/osx/cot.md
```diff
@@ -0,0 +1,24 @@
...
+
+`cot --wait {{path/to/file}}`
+
+- Open a file with the cursor at a specific line and column:
```
marchersimon, line 22:
```suggestion
- Open a specific file with the cursor at a specific line and column:
```

pages/osx/cot.md
superatomic: This line is worded the same way that #8206 was worded. In addition to making this change, should I also open a PR to update the wording for `mate`?

pages/osx/cot.md
superatomic: Same deal here. I can make a new PR to give `mate` the same wording as `cot` once this PR is merged.

pages/osx/cot.md
```diff
@@ -0,0 +1,24 @@
...
+
+`cot`
+
+- Open a new blank document:
+
+`cot --new`
+
+- Open specific files:
+
+`cot {{path/to/file1 path/to/file2 ...}}`
```
EmilySeville7cfg, line 16:
```suggestion
- Open specific files in background:

`cot {{path/to/file1 path/to/file2 ...}}`

- Open a new blank document in background:

`cot --new`
```

- See #8018 why I suggested this. 😄
- Is it right that by default everything is opened in background in this editor? If it's true it's better to specify this explicitly assuming that in general editor opens files in foreground by default (until otherwise noted explicitly).

I can't try this editor, so please check my edits carefully.

pages/osx/cot.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Open a specific file in the foreground (blocking the terminal until the editor is closed):
+
+`cot --wait {{path/to/file}}`
```
EmilySeville7cfg, line 20:
```suggestion
- Open specific files:

`cot --wait {{path/to/file1 path/to/file2 ...}}`
```

EmilySeville7cfg (changes requested):

pages/osx/cot.md
superatomic: After testing the behavior of running `cot`, it appears that the windows will in fact always open in the background, even though the `-g`/`--background` flag is not provided. This might be a bug with `cot`, although I'm not completely sure.

Because of this, I'd prefer to not specifically state that the command opens file in the background, although I can see why we might want to document it anyway, as it is assumed that `cot` would open files in the foreground by default.

(I will make the change to swap the order of `cot --new` and `cot {{path/to/file1 path/to/file2 ...}}` right now)

superatomic: Also, to me, "Open specific files in **the** background" sounds a bit better than "Open specific files in background". If we keep the "in background" wording (see comment above), would that be a better way to word it, or should I keep it without the "the"?

pages/osx/cot.md
superatomic: Should this still keep some information about how this command blocks the terminal until the editor is closed? Maybe "Open specific files (blocking the terminal until the editor is closed)"?

EmilySeville7cfg: I don't think so. Anyway we have to decide whether to write such info everywhere where appropriate or not. 😄

pages/osx/cot.md
EmilySeville7cfg: Let's add `the`.

superatomic: Okay, cool. I've made these changes.

pages/osx/cot.md
marchersimon: I think we should keep the previous version. We somehow have to explain what the `--wait` option does. Otherwise this example would be useless.

pages/osx/cot.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Open specific files in the background:
+
+`cot {{path/to/file1 path/to/file2 ...}}`
+
+- Open a new blank document in the background:
+
+`cot --new`
+
+- Open a specific file (blocking the terminal until the editor is closed):
```
EmilySeville7cfg, line 18:
```suggestion
- Open a specific file in the foreground:
```

EmilySeville7cfg (changes requested):

pages/osx/cot.md
marchersimon: I would be fine with that, but I just have the concern that it's not really clear, what `in the foreground` means here.

EmilySeville7cfg: `in the background` expression also can be confusing. :) Maybe it's better to write what `in the background` actually means such as `without blocking a terminal until the editor is closed`.

marchersimon: That's why I went for this solution, which seemed most natural to me. However, that's just one opinion. I'd say in cases like this we should let the author decide (@superatomic).

superatomic: I was interpreting foreground/background to be whether the window that was opened was sent to the foreground, not whether the terminal blocks.

As it currently stands, either interpretation is correct, although it's probably better to use more descriptive language to avoid confusion.

For this line, I personally like "Open and wait until finished editing a specific file", which was the original text before it was changed. I'd suppose "Open a specific file and wait until it is closed" might be a good way to phrase this as well.

marchersimon: I like the last suggestion. But `wait` is a bit ambiguous. How about `Open a specific file and block the terminal until it is closed`?

superatomic: That sounds good. I chose the word "wait" because it was the name of the option, but I like yours more.

superatomic:
A small comment on the current wording: with the newest changes, do we still need to specify "in the background" on so many of the options, or can we remove it from the end of the lines where it appears?

marchersimon:
I don't think it's very important. You can remove it if you want.

superatomic:
> I don't think it's very important. You can remove it if you want.

I feel like that might make the page clearer. I'll make the changes.

EmilySeville7cfg (approved):

EmilySeville7cfg:
Merged, thanks! 👍🏻

PR #8233, opened 2022-07-20
tldr: add --search example

nebulaa44:

### DIFF
--- a/pages/common/tldr.md
+++ b/pages/common/tldr.md
@@ -19,6 +19,6 @@
 
 `tldr -u`
 
-- Search for a specific tldr-page:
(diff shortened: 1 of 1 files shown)


CleanMachine1:
Hi, thanks for your contribution however I don't think we can add this unfortuately since this command is dependant on the client someone uses. 

Things added to this TLDR Page are only usually things which are a requirement for a tldr client, the spec can be found [here](https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md)

navarroaxel (changes requested):
Sorry, but we can't document non-standard arguments.

PR #8225, opened 2022-07-19
antibody, apg: add French translation

Nico385412:
- [x] The page (if new), does not already exist in the repository.
- [x] The page is in the correct platform directory (`common/`, `linux/`, etc.)
- [x] The page has 8 or fewer examples.
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message).
- [x] The page follows the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).
- [ ] The page description includes a link to documentation or a homepage (if applicable).

**Version of the command being documented (if known):**

PR #8234, opened 2022-07-21
apktool, apm: add French translation, update

Nico385412:

---

Hi everyone 👋 
I translated this 2 pages, but also suggest editing the original english pages.

### apktool

I removed an `s` to framework because we handle [only one framework at a time ](https://forum.xda-developers.com/t/how-to-use-and-install-apktool.1760133/)

Also `framework` does not have a translation in french. (can you confirm @nicokosi ?)

### apm

Same thing as apktool, they say we install `packages` and `themes` but we do not install multiple elements. so i changed the sentence to use singular.

### DIFF
--- a/pages.fr/common/apm.md
+++ b/pages.fr/common/apm.md
@@ -1,0 +1,17 @@
+# apm
+
+> Manageur de paquets pour Atom.
+> Voir `atom`.
+> Plus d'informations : <https://github.com/atom/apm>.
+
+- Installe un paquet depuis http://atom.io/packages ou un thème depuis http://atom.io/themes :
+
+`apm install {{nom_du_paquet}}`
+
+- Supprime un paquet ou un thème :
+
+`apm remove {{nom_du_paquet}}`
+
+- Mets à jour un paquet ou un thème :
+
+`apm upgrade {{nom_du_paquet}}`
--- a/pages.fr/common/apm.md
+++ b/pages.fr/common/apm.md
@@ -1,6 +1,6 @@
 # apm
 
-> Manageur de paquets pour Atom.
+> Gestionnaire de paquets pour Atom.
 > Voir `atom`.
 > Plus d'informations : <https://github.com/atom/apm>.
 
--- a/pages/common/apm.md
+++ b/pages/common/apm.md
@@ -4,14 +4,14 @@
 > See `atom`.
 > More information: <https://github.com/atom/apm>.
 
-- Install packages from http://atom.io/packages and themes from http://atom.io/themes:
+- Install package from http://atom.io/packages or theme from http://atom.io/themes:
 
 `apm install {{package_name}}`
 
-- Remove packages/themes:
+- Remove package/theme:
 
 `apm remove {{package_name}}`
 
-- Upgrade packages/themes:
+- Upgrade package/theme:
 
 `apm upgrade {{package_name}}`
--- a/pages/common/apm.md
+++ b/pages/common/apm.md
@@ -8,7 +8,7 @@
 
 `apm install {{package_name}}`
 
-- Remove package/theme:
+- Remove a package/theme:
 
 `apm remove {{package_name}}`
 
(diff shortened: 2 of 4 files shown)


pages.fr/common/apm.md
```diff
@@ -0,0 +1,17 @@
+# apm
+
+> Manageur de paquets pour Atom.
```
nicokosi, line 3:
```suggestion
> Gestionnaire de paquets pour Atom.
```

pages.fr/common/apm.md
```diff
@@ -0,0 +1,17 @@
...
+
+- Installe un paquet depuis http://atom.io/packages ou un thème depuis http://atom.io/themes :
+
+`apm install {{nom_du_paquet}}`
+
+- Supprime un paquet ou un thème :
+
+`apm remove {{nom_du_paquet}}`
+
+- Mets à jour un paquet ou un thème :
```
nicokosi, line 15:
```suggestion
- Met à jour un paquet ou un thème :
```

pages/common/apm.md
```diff
@@ -4,14 +4,14 @@
 > See `atom`.
 > More information: <https://github.com/atom/apm>.
 
-- Install packages from http://atom.io/packages and themes from http://atom.io/themes:
+- Install package from http://atom.io/packages or theme from http://atom.io/themes:
```
nicokosi, line 7:
```suggestion
- Install a package from http://atom.io/packages or theme from http://atom.io/themes:
```

pages/common/apm.md
```diff
@@ -4,14 +4,14 @@
...
 
 `apm install {{package_name}}`
 
-- Remove packages/themes:
+- Remove package/theme:
```
nicokosi, line 11:
```suggestion
- Remove a package/theme:
```

pages/common/apm.md
```diff
@@ -4,14 +4,14 @@
...
 
 `apm remove {{package_name}}`
 
-- Upgrade packages/themes:
+- Upgrade package/theme:
```
nicokosi, line 15:
```suggestion
- Upgrade a package/theme:
```

pages.fr/common/apm.md
Nico385412: That true, `Gestionnaire` is better than `Manageur`

pages.fr/common/apm.md
Nico385412: same response as this PR ^^" 

https://github.com/tldr-pages/tldr/pull/8225#discussion_r925182444

marchersimon (approved):
Looks good, thanks

pages.fr/common/apm.md
navarroaxel: présent de l'indicatif explains better what a command does, at least in Spanish. I think this apply to French too.

Nico385412: Both are ok in french, but depends on the context and not the actual form that nicokosi ask.

If you want to use the present de l'indicatif you must know that's a tense where you need to identify a subject before using a verb. (e, g j'aime les pommes (I like apples)). And you can't remove the `I` (`j'`) because you loose the `who`

Since we give implicitly an order to a machine (the command) then you don't need to specify the subject, BUT you must use imperative for the verb.

So if we want to use the present de l'indicatif, a correct sentence would be `il met à jour un paquet ou un thème :`
(it upgrades ....)

I hope my explanation are clear ^^" wdyt @nicokosi  ?

nicokosi: Not sure to understand everything. :innocent:

I agree with: :ok_hand: 
> Both are ok in french,

IMHO, the subject is implicitly `apm` and indicative / third person is better. That's why I suggest "_met_" for "_(apm) met_" (indicative present, third person like "it"). :smiley_cat:

nicokosi (approved):
:+1:

navarroaxel (approved):

PR #8238, opened 2022-07-21
pyinfra: add page

CrimsonTome:
- **Version of the command being documented (if known):**  2.3

### DIFF
--- a/pages/common/pyinfra.md
+++ b/pages/common/pyinfra.md
@@ -1,0 +1,28 @@
+# pyinfra
+
+> Automates infrastructure at a large scale.
+> More information: <https://docs.pyinfra.com>.
+
+- Execute a command over ssh:
+
+`pyinfra {{target_ip_address}} exec -- {{command_name_and_arguments}}`
+
+- Execute contents of a deploy file on a list of targets:
+
+`pyinfra {{path/to/inventory.py}} {{path/to/deploy.py}}`
+
+- Execute commands on your local machine:
+
+`pyinfra @local exec -- {{command_name_and_arguemnts}}`
+
+or
+
+`pyinfra @local {{path/to/deploy.py}}`
+
+- Execute commands over Docker:
+
+`pyinfra @docker/{{container}} exec -- {{command_name_and_arguemnts}}`
+
+or
+
+`pyinfra @docker/{{container}} {{path/to/deploy.py}}`
--- a/pages/common/pyinfra.md
+++ b/pages/common/pyinfra.md
@@ -3,26 +3,18 @@
 > Automates infrastructure at a large scale.
 > More information: <https://docs.pyinfra.com>.
 
-- Execute a command over ssh:
+- Execute a command over `SSH`:
 
 `pyinfra {{target_ip_address}} exec -- {{command_name_and_arguments}}`
 
 - Execute contents of a deploy file on a list of targets:
 
-`pyinfra {{path/to/inventory.py}} {{path/to/deploy.py}}`
+`pyinfra {{path/to/target_list.py}} {{path/to/deploy.py}}`
 
-- Execute commands on your local machine:
-
-`pyinfra @local exec -- {{command_name_and_arguemnts}}`
-
-or
+- Execute commands on locally:
 
 `pyinfra @local {{path/to/deploy.py}}`
 
 - Execute commands over Docker:
-
-`pyinfra @docker/{{container}} exec -- {{command_name_and_arguemnts}}`
-
-or
 
 `pyinfra @docker/{{container}} {{path/to/deploy.py}}`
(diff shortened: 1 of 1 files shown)


CrimsonTome:
With regards to the 'or' sections in pyinfra.md, I wasn't sure how to format them because they detail how to do the same operation, just with a slightly different syntax - that has been explained at the top. I just thought it would be a little pointless adding a new section for each way of doing things when it has been explained if that makes sense. Any feedback is much appreciated! :smile:

marchersimon:
To keep our pages maintainable we have to enforce some strict rules, for example that each example is isolated from the others. In this case we should probably decide for one of the two versions - whichever you think is more useful.

pages/common/pyinfra.md
```diff
@@ -0,0 +1,28 @@
+# pyinfra
+
+> Automates infrastructure at a large scale.
+> More information: <https://docs.pyinfra.com>.
+
+- Execute a command over ssh:
```
marchersimon, line 6:
```suggestion
- Execute a command over `ssh`:
```

pages/common/pyinfra.md
```diff
@@ -0,0 +1,28 @@
...
+
+`pyinfra {{target_ip_address}} exec -- {{command_name_and_arguments}}`
+
+- Execute contents of a deploy file on a list of targets:
+
+`pyinfra {{path/to/inventory.py}} {{path/to/deploy.py}}`
```
marchersimon, line 12: I'm not sure if it makes sense here, but it's a bit unclear what `inventory.py` is.
```suggestion
`pyinfra {{path/to/target_list.py}} {{path/to/deploy.py}}`
```

pages/common/pyinfra.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Execute commands on your local machine:
```
marchersimon, line 14:
```suggestion
- Execute commands locally:
```
We try to avoid pronouns like `your`.

CrimsonTome:
Thank you for your review @marchersimon. I will adapt the page later on in the day when I have access to a PC. 

inventory.py is the canonical name of the targets list though I believe you can name it what you like.

pages/common/pyinfra.md
```diff
@@ -0,0 +1,28 @@
...
+
+`pyinfra @local exec -- {{command_name_and_arguemnts}}`
+
+or
```
sbrl, line 18: We have a linter which enforces a strict format, so unfortunately we can't have an `or` branch here. The format is very much
```markdown
- Example description

`example command`
```

pages/common/pyinfra.md
```diff
@@ -0,0 +1,28 @@
...
+
+`pyinfra @local {{path/to/deploy.py}}`
+
+- Execute commands over Docker:
+
+`pyinfra @docker/{{container}} exec -- {{command_name_and_arguemnts}}`
+
+or
```
sbrl, line 26: Same here.

pages/common/pyinfra.md
```diff
@@ -0,0 +1,28 @@
...
+
+`pyinfra @local exec -- {{command_name_and_arguemnts}}`
+
+or
+
+`pyinfra @local {{path/to/deploy.py}}`
+
+- Execute commands over Docker:
+
+`pyinfra @docker/{{container}} exec -- {{command_name_and_arguemnts}}`
```
sbrl, line 24: typo?
```suggestion
`pyinfra @docker/{{container}} exec -- {{command_name_and_arguments}}`
```

sbrl (commented):
Disclaimer: I know @CrimsonTome IRL,

CrimsonTome:
Thank you for the review @sbrl. 

Yeah, I will fix the `or` statements with the preferable option. And oops ty for spotting the typo 😅.
As per my previous comment, I will resolve these later today.

pages/common/pyinfra.md
CleanMachine1:
```suggestion
- Execute a command over `SSH`:
```

sbrl: This doesn't make any sense to have SSH in backticks, because SSH is the name of a protocol and not a command.

sbrl (commented):
Looks ok to me, though as I've commented above SSH does not make sense to be in backticks as it's the name of a protocol and not the name of a command.

sbrl (approved):

CleanMachine1 (approved):
Thanks for your contribution!

pages/common/pyinfra.md
```diff
@@ -0,0 +1,20 @@
+# pyinfra
+
+> Automates infrastructure at a large scale.
+> More information: <https://docs.pyinfra.com>.
+
+- Execute a command over `SSH`:
```
marchersimon, line 6:
```suggestion
- Execute a command over SSH:
```

CrimsonTome:
Okay, all issues *should* be sorted now :)

navarroaxel (approved):

PR #8239, opened 2022-07-21
psql: clarify default host behaviour

medizinmensch:
- **Version of the command being documented (if known):**  12.11

The first example for psql is wrong or at least misleading. It lets users to believe that `psql my_db` is the same as `psql -h 127.0.0.1` which is not. By default (at least in v12.11), it connects via Unix Socket. 

This makes a difference for instance for access configuration in the `pg_hba.conf` file. I could not connect because usually by default (afaik), only the postgres user is allowed to access via socket. 

Alternatively one could say "... connects to the local database..." or something without naming localhost.

### DIFF
--- a/pages/common/psql.md
+++ b/pages/common/psql.md
@@ -3,7 +3,7 @@
 > PostgreSQL command-line client.
 > More information: <https://www.postgresql.org/docs/current/app-psql.html>.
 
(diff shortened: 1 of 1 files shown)


marchersimon (approved):
Looks good, thanks

mfrw (approved):
Thank you @medizinmensch 
LGTM :)

PR #8244, opened 2022-07-23
picom-trans: add page

flipsi:
`picom-trans` doesn't support versions, `picom` itself has version **vgit-7e568**.  

### DIFF
--- a/pages/common/picom-trans.md
+++ b/pages/common/picom-trans.md
@@ -1,0 +1,20 @@
+# picom-trans
+
+> Set window opacity.
+> More information: <https://github.com/yshui/picom>.
+
+- Set opacity of current window to 90 percent:
+
+`picom-trans --current --opacity={{90}}`
+
+- Set opacity of window with name 'Firefox' to 90 percent:
+
+`picom-trans --name={{Firefox}} --opacity={{90}}`
+
+- Select window to set opacity for with mouse cursor:
+
+`picom-trans --select --opacity={{90}}`
+
+- Toggle opacity for window with name 'Firefox':
+
+`picom-trans --name={{Firefox}} --toggle`
--- a/pages/common/picom-trans.md
+++ b/pages/common/picom-trans.md
@@ -3,18 +3,18 @@
 > Set window opacity.
 > More information: <https://github.com/yshui/picom>.
 
-- Set opacity of current window to 90 percent:
+- Set the currently focused window opacity to a certain percentage:
 
 `picom-trans --current --opacity={{90}}`
 
-- Set opacity of window with name 'Firefox' to 90 percent:
+- Set the opacity of a window with a specific name:
 
 `picom-trans --name={{Firefox}} --opacity={{90}}`
 
-- Select window to set opacity for with mouse cursor:
+- Set the opacity of a specific window selected via mouse cursor:
 
 `picom-trans --select --opacity={{90}}`
 
-- Toggle opacity for window with name 'Firefox':
+- Toggle the opacity of a specific window:
 
 `picom-trans --name={{Firefox}} --toggle`
--- a/pages/common/picom-trans.md
+++ b/pages/common/picom-trans.md
@@ -3,7 +3,7 @@
 > Set window opacity.
 > More information: <https://github.com/yshui/picom>.
 
-- Set the currently focused window opacity to a certain percentage:
+- Set the currently focused window opacity to a specific percentage:
 
 `picom-trans --current --opacity={{90}}`
 
--- a/pages/common/picom-trans.md
+++ b/pages/common/picom-trans.md
@@ -5,16 +5,16 @@
 
 - Set the currently focused window opacity to a specific percentage:
 
-`picom-trans --current --opacity={{90}}`
+`picom-trans --current --opacity {{90}}`
 
 - Set the opacity of a window with a specific name:
 
-`picom-trans --name={{Firefox}} --opacity={{90}}`
+`picom-trans --name {{Firefox}} --opacity {{90}}`
 
 - Set the opacity of a specific window selected via mouse cursor:
 
-`picom-trans --select --opacity={{90}}`
+`picom-trans --select --opacity {{90}}`
 
 - Toggle the opacity of a specific window:
 
-`picom-trans --name={{Firefox}} --toggle`
+`picom-trans --name {{Firefox}} --toggle`
--- a/pages/common/picom-trans.md
+++ b/pages/common/picom-trans.md
@@ -1,6 +1,6 @@
 # picom-trans
 
-> Set window opacity.
+> Set the window opacity for the `picom` window compositor.
 > More information: <https://github.com/yshui/picom>.
 
 - Set the currently focused window opacity to a specific percentage:


pages/common/picom-trans.md
```diff
@@ -0,0 +1,20 @@
+# picom-trans
+
+> Set window opacity.
+> More information: <https://github.com/yshui/picom>.
+
+- Set opacity of current window to 90 percent:
```
EmilySeville7cfg, line 6:
```suggestion
- Set the current window opacity to a specific percent:
```

pages/common/picom-trans.md
```diff
@@ -0,0 +1,20 @@
...
+
+`picom-trans --current --opacity={{90}}`
+
+- Set opacity of window with name 'Firefox' to 90 percent:
```
EmilySeville7cfg, line 10:
```suggestion
- Set a specific window opacity:
```

pages/common/picom-trans.md
```diff
@@ -0,0 +1,20 @@
...
+
+`picom-trans --name={{Firefox}} --opacity={{90}}`
+
+- Select window to set opacity for with mouse cursor:
```
EmilySeville7cfg, line 14:
```suggestion
- Set an opacity for a specific window selected via mouse cursor:
```

pages/common/picom-trans.md
```diff
@@ -0,0 +1,20 @@
...
+
+`picom-trans --select --opacity={{90}}`
+
+- Toggle opacity for window with name 'Firefox':
```
EmilySeville7cfg, line 18:
```suggestion
- Toggle a specific window opacity:
```

EmilySeville7cfg (changes requested):

pages/common/picom-trans.md
navarroaxel: Something like this? :thinking: 
```suggestion
- Set an opacity for a window with a specific name:
```

EmilySeville7cfg: Yeah, sounds nicer than my variant.

marchersimon: I'd use the definite article, however: `Set the opacity ...`

pages/common/picom-trans.md
marchersimon: I'm not entirely sure, but I think `Firefox` refers to a window class here, not a single window. Is this correct @flipsi?

pages/common/picom-trans.md
marchersimon: @flipsi I assume `current window` means `currently focused window`, right? It is a bit wordy, but I think we should at least clarify this in the first example.

flipsi: > @flipsi I assume `current window` means `currently focused window`, right? It is a bit wordy, but I think we should at least clarify this in the first example.

Correct, thanks for pointing this out! 

I [adjusted the wording](https://github.com/tldr-pages/tldr/pull/8244/commits/8f97d5e862b4b9a68ba2ebdfd48938e85fe33f7e).

pages/common/picom-trans.md
flipsi: No, as far as I understand, window name is different from window class and this matches the name.
At least I tested this and compared my query with the different values I got from `xprop`. 
Window name is what is most likely shown as a title in window decorations of window managers.

flipsi:
Thank you for your detailed review!

I didn't like the concrete values in descriptions either, but wanted to be consistent with what I found in `pages/common/split.md`. Anyway, I [rephrased the descriptions now].

pages/common/picom-trans.md
```diff
@@ -0,0 +1,20 @@
+# picom-trans
+
+> Set window opacity.
+> More information: <https://github.com/yshui/picom>.
+
+- Set the currently focused window opacity to a certain percentage:
```
EmilySeville7cfg, line 6:
```suggestion
- Set the current window opacity to a specific percentage:
```

AFAIK we prefer `a` instead of `the`, and `specific` instead of `certain`. Pages should be standardized. ;)

pages/common/picom-trans.md
```diff
@@ -0,0 +1,20 @@
...
+
+`picom-trans --current --opacity={{90}}`
+
+- Set the opacity of a window with a specific name:
```
EmilySeville7cfg, line 10:
```suggestion
- Set an opacity of a window with a specific name:
```

pages/common/picom-trans.md
```diff
@@ -0,0 +1,20 @@
...
+
+`picom-trans --name={{Firefox}} --opacity={{90}}`
+
+- Set the opacity of a specific window selected via mouse cursor:
```
EmilySeville7cfg, line 14:
```suggestion
- Set an opacity of a specific window selected via mouse cursor:
```

pages/common/picom-trans.md
```diff
@@ -0,0 +1,20 @@
...
+
+`picom-trans --select --opacity={{90}}`
+
+- Toggle the opacity of a specific window:
```
EmilySeville7cfg, line 18:
```suggestion
- Toggle an opacity of a specific window:
```

EmilySeville7cfg (dismissed):

pages/common/picom-trans.md
marchersimon: I agree with `specific`, but what's wrong with `currently focused`. It's more precise than `current`.

pages/common/picom-trans.md
marchersimon: I think `the` is better in this context, because every window has one opacity. And we're setting this very opacity. `An opacity` sounds like we're setting one of the many opacities a window could have.

pages/common/picom-trans.md
flipsi: 1) What about [Simon's](https://github.com/tldr-pages/tldr/pull/8244#discussion_r928295348) remark about "focussed"? I think we should incorporate this.
2) Okay, but I'm already using `a` (at least in the descriptions below). I think using an indefinite article twice doesn't make much sense here: I'm talking a about **a** window, so it's **the** opacity of this particular window (definite in this context).
3) Okay, I will change `certain` to `specific`. (Wanted to avoid using `specific` twice in the second description)

pages/common/picom-trans.md
```diff
@@ -0,0 +1,20 @@
+# picom-trans
+
+> Set window opacity.
```
marchersimon, line 3:
```suggestion
> Set window opacity for the `picom` window compositor.
```

pages/common/picom-trans.md
```diff
@@ -0,0 +1,20 @@
...
+> Set the window opacity for the `picom` window compositor.
+> More information: <https://github.com/yshui/picom>.
+
+- Set the currently focused window opacity to a specific percentage:
+
+`picom-trans --current --opacity={{90}}`
+
+- Set the opacity of a window with a specific name:
+
+`picom-trans --name={{Firefox}} --opacity={{90}}`
```
marchersimon, line 12: Just one last thing from my side: Could you please test if those flags work without the `=` and remove them if they do?

flipsi: Sure. They do work, so [I removed them as you requested](https://github.com/tldr-pages/tldr/pull/8244/commits/04d516ca26a97fd7bd6bd158dc1dbec86ad1a625). 

The man page however suggest using `=`. So just out of curiosity: Shouldn't we stick to that?

marchersimon: Because it makes it more readable. And it's easier to just never use them instead of remembering where they work and where not. In the end it doesn't make a difference.

marchersimon (approved):
Thanks @flipsi

CleanMachine1 (approved):

CleanMachine1:
Thanks for your contribution!

PR #8245, opened 2022-07-24
ag: fix 小明 to foo.

Brannua:


Brannua:
The page is in the correct platform directory: `common`

pages.zh/common/ag.md
```diff
@@ -9,11 +9,11 @@
 
 - 在指定目录中寻找内容包含 "foo" 的文件：
```
navarroaxel, line 10: what about this non-translated `foo`?

navarroaxel (commented):
Please sign the license to pass the CLA check

pages.zh/common/ag.md
Brannua: > what about this non-translated `foo`?

I just adjust the parameters in the use case to be consistent. Foo is only a literal value and has no special meaning.

PR #8249, opened 2022-07-25
make: add variable override example

adrienthebo:
Makefile variables can be overridden via the command line in the form of
`{{variable}}={{override}}` See the [Make manual section 9.5] for more
information.

[Make manual section 9.5]: https://www.gnu.org/software/make/manual/html_node/Overriding.html

- - -

**Note:** This syntax seems very useful but doesn't seem commonly used or discussed. I see value in adding it but it's a bit further out of the core set of functions that user might need, so I'd like a rationality check on if this seems useful enough for inclusion.

This allows for make invocations such as `make cleanup cloud=aws version=stable`.
 
- - -

- **GNU Make 3.81**  

### DIFF
--- a/pages/common/make.md
+++ b/pages/common/make.md
@@ -32,6 +32,6 @@
 
 `make --environment-overrides {{target}}`
 
-- Override a variable defined in the Makefile via the command line:
+- Override a variable defined in the Makefile:
 
-`make {{target}} {{variable}}={{override}}`
+`make {{target}} {{variable}}={{new_value}}`
--- a/pages/common/make.md
+++ b/pages/common/make.md
@@ -31,3 +31,7 @@
 - Override variables defined in the Makefile by the environment:
 
 `make --environment-overrides {{target}}`
+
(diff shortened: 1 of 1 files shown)


pages/common/make.md
```diff
@@ -31,3 +31,7 @@
 - Override variables defined in the Makefile by the environment:
 
 `make --environment-overrides {{target}}`
+
+- Override a variable defined in the Makefile via the command line:
+
+`make {{target}} {{variable}}={{override}}`
```
marchersimon, line 37:
```suggestion
- Override a variable defined in the Makefile:

`make {{target}} {{variable}}={{new_value}}`
```
I think `via the command line` is a little redundant. I assume you added it for contrast to the previous example. However, I think we could move this one before the `--environment-overrides` anyways, because it's less specific.

marchersimon (approved):
Looks good, thanks!

CleanMachine1 (approved):

CleanMachine1:
thanks for your contribution

PR #8252, opened 2022-07-25
Style guide: refresh placeholder syntax

EmilySeville7cfg:
> ⚠️ This PR is a reference for other smaller PRs. It will be closed after all ideas from here are accepted.

## ToDo

- [Deny](https://github.com/tldr-pages/tldr/pull/8367#discussion_r956662403) substituting concrete usernames and passwords.
- [Prefer](https://github.com/tldr-pages/tldr/pull/8367#discussion_r956662260) `127.0.0.1` as an IP example.
- [Deny](https://github.com/tldr-pages/tldr/pull/9139#discussion_r1004134693) placeholders for concrete values.

## Next PRs

- Describe when and where use short and long options: #9527
- Describe quoting rules inside command descriptions: #9528
- Describe how to write globs and regex (not to over complicate pages)
- Describe how to write `--help` and `--version` options summary
- Add notes according to [this](https://github.com/tldr-pages/tldr/pull/8025#discussion_r948379968) question
- [Deny](https://github.com/tldr-pages/tldr/pull/8367#discussion_r954848116) placeholder usage in descriptions
- Describe how to write [interactive choices affecting command behaviour](https://github.com/tldr-pages/tldr/pull/8474#discussion_r974421653)
- Force writing [descriptive](https://github.com/tldr-pages/tldr/pull/7931#discussion_r906419677) code explanations rather generic. Describe what we want to do rather than how we do it.
- Require explicit type annotations such as in [sed](https://github.com/tldr-pages/tldr/pull/7931) page. Think about rules.
- Require specify where command writes output (stdout, stderr) explicitly. No exceptions.
- [Deny](https://github.com/tldr-pages/tldr/pull/9495#pullrequestreview-1186984875) short->long option conversion when more than 5 options can be converted.

Everything that can be standardized should be standardized ;) Some examples are removed because I want to replace them with smth better when find.

### DIFF
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -121,6 +121,20 @@ in order to allow `tldr` clients to highlight them.
 
 [snake_case]: https://wikipedia.org/wiki/snake_case
 
+### Quoting
+
+- Don't quote path placeholders for page simplicity although shell may require quotes.
+  In other cases quote arguments just when documentation mandates quotes explicitly or
+  you exactly know they are required and the documentation just missing information about this.
+- Use double quotes for placeholders where variable or another substitution occures.
+  In other cases prefer single quotes.
+
+> :bulb: Many shells have [word splitting][word_splitting] enabled by default so unquoted arguments may be splitted to
+several ones. But the main goal of Tl;Dr pages is to provide easy to read and simple in use pages but
+not to strictly follow POSIX shell syntax.
+
+[word_splitting]: https://mywiki.wooledge.org/WordSplitting
+
 ### Paths
 - Use `{{filename}}` rather than `{{file_name}}`.
 - For any reference to paths of files or directories,
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -116,7 +116,7 @@ in order to allow `tldr` clients to highlight them.
 - Use [snake_case][snake_case] for multi-word placeholders.
   Examples: `{{path/to/csharp_solution.sln}}` or `{{language_id}}`.
 - Use an actual value in a placeholder rather than a placeholder
-  with a value description inside it when one of the following conditions are met:
+  with a value description inside it when at least one of the following conditions are met:
   - there is a long (consisting of several letters) option immidiately before
     this placeholder 
     Examples: `pwsh -Command "{{echo 'Hello world!'}}"` or `cut --delimiter="{{,}}" --{{characters}}={{1}}`.
@@ -126,6 +126,18 @@ in order to allow `tldr` clients to highlight them.
     Examples: `bash -c "echo 'Hello world!'"` with `Execute specific [c]ommands:` description.
   - the command or it's the most nested subcommand have just one placeholder mentioned above
     Examples: `wine {{explorer}}` or `iostat {{5}}`.
+- Use the following actual placeholders when they are valid values for a command
+  and at least one of the previous three conditions are met:
+  - `5` for integers:
+    Example: `iostat {{5}}`
+  - `5.2` for floats:
+    Example: `sleep {{5.2}}`
+  - `hello` for strings:
+    Example: `expr length '{{hello}}'`.
+  - `echo "Hello world!"` for commands:
+    Example: `bash -c "{{echo 'Hello world!'}}"`.
+  - `~/.emacs` for files:
+    Example: `touch --no-create --time={{access|modify}} --reference={{~/.emacs}} {{path/to/file1 path/to/file2 ...}}`.
 
 That's why we prefer `expr substr {{string}} {{from}} {{length}}` over `expr substr {{hello}} {{2}} {{3}}`.
 Code sample descriptions must be concise and don't refer to placeholders by their position.
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -112,12 +112,22 @@ in order to allow `tldr` clients to highlight them.
 ### General rules
 
 - Use short but descriptive placeholders,
-  such as `{{path/to/source.cs}}` or `{{wallet.txt}}`.
+  such as `{{path/to/source.cs}}` or `{{drive}}`.
 - Use [snake_case][snake_case] for multi-word tokens.
-- Use an actual value rather than a generic placeholder where appropriate.
-  For example, use `iostat {{2}}` rather than `iostat {{interval_in_secs}}`.
+- Use an actual value in placeholder such as `{{5}}` rather than
+  a placeholder with value description inside it such as `{{seconds}}`
+  when one of the following conditions are met:
+  - there is long (consisting of several letters) option immidiately before
+    this placeholder which allows us understand the meaning of
+    an actual value
+  - there is a short (consisting of one letter) option immidiately before this
+    placeholder with a mnemonic provided via square brackets in a code
+    example description
+  - the command or it's the most nested subcommand performs one simple task such as `sleep` 
+   
+  For example, use `iostat {{2}}` rather than `iostat {{seconds}}`.
 
-> :bulb: Formal rules are described below.
+> :bulb: More formal rules are described below.
 
 [snake_case]: https://wikipedia.org/wiki/snake_case
 
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -100,9 +100,9 @@ Example:
 - View documentation for the original command:
 
 `tldr vim`
+```
 
 Pre-translated alias page templates can be found [here](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/translation-templates/alias-pages.md).
-```
 
 ## Placeholder syntax
 
--- a/contributing-guides/style-guide.md
+++ b/contributing-guides/style-guide.md
@@ -104,22 +104,22 @@ Example:
 Pre-translated alias page templates can be found [here](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/translation-templates/alias-pages.md).
 ```
 
-## Token syntax
+## Placeholder syntax
 
-User-provided values should use the `{{token}}` syntax
+User-provided values should use the `{{placeholder}}` syntax
 in order to allow `tldr` clients to highlight them.
-
-Keep the following guidelines in mind when choosing tokens:
 
 ### General rules
 
-- Use short but descriptive tokens,
+- Use short but descriptive placeholders,
   such as `{{path/to/source.cs}}` or `{{wallet.txt}}`.
-- Use [`snake_case`](https://wikipedia.org/wiki/snake_case) for multi-word tokens.
+- Use [snake_case][snake_case] for multi-word tokens.
 - Use an actual value rather than a generic placeholder where appropriate.
   For example, use `iostat {{2}}` rather than `iostat {{interval_in_secs}}`.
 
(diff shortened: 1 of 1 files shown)


contributing-guides/style-guide.md
```diff
@@ -104,40 +104,60 @@ Example:
 Pre-translated alias page templates can be found [here](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/translation-templates/alias-pages.md).
 ```
 
-## Token syntax
+## Placeholder syntax
 
-User-provided values should use the `{{token}}` syntax
+User-provided values should use the `{{placeholder}}` syntax
```
mfrw, line 109: I actually like the name `placeholder` better than `token`. 👍🏼

marchersimon: It's definitively more descriptive, especially for new contributors. (Even tough `token` just sounds cooler)

contributing-guides/style-guide.md
```diff
@@ -104,40 +104,60 @@ Example:
...
-Keep the following guidelines in mind when choosing tokens:
+### General rules
 
-### Naming
-- Use short but descriptive tokens,
-  such as `{{source_file}}` or `{{wallet.txt}}`.
-- Use [`snake_case`](https://wikipedia.org/wiki/snake_case) for multi-word tokens.
+- Use short but descriptive placeholders,
+  such as `{{path/to/source.cs}}` or `{{wallet.txt}}`.
+- Use [snake_case][snake_case] for multi-word tokens.
```
marchersimon, line 116: Did you remove the URL on purpose? I'm not sure if `[...][...]` is even valid markdown.

mfrw (approved):

contributing-guides/style-guide.md
```diff
@@ -104,40 +104,74 @@ Example:
...
+> :bulb: More formal rules are described below.
+
+[snake_case]: https://wikipedia.org/wiki/snake_case
+
+### Quoting
+
+- Don't quote path placeholders for page simplicity although shell may require quotes.
+  In other cases quote placeholders just when:
+  - documentation mandates quotes explicitly
+  - you exactly know they are required and the documentation just missing information about this
```
sbrl, line 138: Perhaps this will read better?
```suggestion
  - you exactly know they are required and the documentation is just missing information about this.
```

contributing-guides/style-guide.md
```diff
@@ -104,40 +104,74 @@ Example:
...
+
+> :bulb: More formal rules are described below.
+
+[snake_case]: https://wikipedia.org/wiki/snake_case
+
+### Quoting
+
+- Don't quote path placeholders for page simplicity although shell may require quotes.
+  In other cases quote placeholders just when:
+  - documentation mandates quotes explicitly
```
sbrl, line 137:
```suggestion
  - documentation mandates quotes explicitly.
```

contributing-guides/style-guide.md
```diff
@@ -104,40 +104,74 @@ Example:
...
+  - you exactly know they are required and the documentation just missing information about this
+- Use double quotes for placeholders where variable or another substitution occures.
+  In other cases use single quotes.
```
sbrl, line 140: This is overly specific. Let's remove this item from the list.

contributing-guides/style-guide.md
```diff
@@ -104,40 +104,74 @@ Example:
...
+
+> :bulb: Many shells have [word splitting][word_splitting] enabled by default so unquoted arguments may be splitted to
+several ones. But the main goal of Tl;Dr pages is to provide easy to read and simple in use pages but
+not to strictly follow POSIX shell syntax.
```
sbrl, line 144: Erm if you don't provide commands that are compatible with the shell on the target platform, then the examples will not work. Let's remove this.

contributing-guides/style-guide.md
```diff
@@ -104,40 +104,74 @@ Example:
...
+### General rules
 
-### Naming
-- Use short but descriptive tokens,
-  such as `{{source_file}}` or `{{wallet.txt}}`.
-- Use [`snake_case`](https://wikipedia.org/wiki/snake_case) for multi-word tokens.
-- Use an actual value rather than a generic placeholder where appropriate.
-  For example, use `iostat {{2}}` rather than `iostat {{interval_in_secs}}`.
+- Use short but descriptive placeholders,
+  such as `{{path/to/source.cs}}` or `{{drive}}`.
```
sbrl, line 115: Not sure `{{drive}}` makes sense to me.

contributing-guides/style-guide.md
```diff
@@ -104,40 +104,74 @@ Example:
...
+  - there is a short (consisting of one letter) option immidiately before this
+    placeholder with a mnemonic provided via square brackets in a code
+    example description
+  - the command or it's the most nested subcommand performs one simple task such as `sleep` 
+   
+  For example, use `iostat {{2}}` rather than `iostat {{seconds}}`.
+
+> :bulb: More formal rules are described below.
+
+[snake_case]: https://wikipedia.org/wiki/snake_case
```
sbrl, line 131: This is extremely specific. It's difficult to capture a set of rules here, as we generally take it on a case-by-case basis - i.e. where it makes sense. Remember: the style guide is **not** meant to contain a comprehensive list of *all* our styling guidelines. It's meant to give a quick overview for new users.

Let's revert this.

contributing-guides/style-guide.md
```diff
@@ -104,40 +104,74 @@ Example:
...
+  - documentation describes command argument semantics explicitly
+  - you exactly know placeholder meaning (maybe from source code) and the documentation just missing information about this.
+
+> :bulb: If some command is cross-platform then Unix-style path placeholders are preferred.
+
+### Extensions
+
+- Append `.ext` to the end of a file path placeholder just when:
+  - documentation requires extension explicitly
+  - you exactly know it is required (maybe from source code) and the documentation just missing information about this
```
sbrl, line 167: If the extension is required, we prefer `{{path/to/thing}}.txt`

contributing-guides/style-guide.md
```diff
@@ -104,40 +104,74 @@ Example:
...
+  - adding extension is an good practise described anywhere
+  
+  For example: `unrar x {{path/to/compressed.rar}}`
+- Replace `ext` word from placeholder above with one of concreete extensions
```
sbrl, line 171: This does not make grammatical sense, and ref the comment above I've unsure how to correct this.

sbrl (changes requested):

contributing-guides/style-guide.md
EmilySeville7cfg: Where can I put comprehensive set of rules then? I can move it to a separate file. But how to name it?

contributing-guides/style-guide.md
EmilySeville7cfg: Do you mean the default login shell for the target platform?

contributing-guides/style-guide.md
EmilySeville7cfg: I can move all formal rules to a separate file and simplify this style guide. From my perspective it's a bad idea to remove this rule because it let us to know when and what prefer exactly. :)

contributing-guides/style-guide.md
EmilySeville7cfg: Is `{{drive_letter}}` a better choice?

contributing-guides/style-guide.md
EmilySeville7cfg: I've updated (#8366) CONTRIBUTING.md: added info about placeholders. I think it's enough to quickly start writing `tldr` pages. The comprehensive list of formal rules can be put here, in style guide.

contributing-guides/style-guide.md
SethFalco: I don't even know what this is trying to say, whichi s probably why it's hard to correct. 🤔

SethFalco: concreete → concrete

Is "`ext` word" referring to `.rar` in the line above?
Maybe: "Replace the extension in the example above (`rar`)"

No idea what a "concrete extension" is, so a reader won't either.

EmilySeville7cfg: I refer to some extension in placeholder. Not to `rar`.

SethFalco: What does "some extension in placeholder" mean?

EmilySeville7cfg: For instance we have `engrapma {{path/to/archive.ext}}` code sample. It's better to replace `ext` with any concrete extension such as `tar` or `zip` instead of leaving example as is.

SethFalco: Ahh I see, `{{placeholder}}` is what you're renaming token too. 🤔 
I wouldn't say I agree with this change, those are not placeholders, they're just user input.

For many commands, they have defined parameters to match the docs, not a placeholder that users should replace. Token fits better imo.

> It's better to replace ext with any concrete extension such as tar or zip instead of leaving example as is.

Sorry, but I'd also disagree with this. Or at least based on our current logic, I think it's better to keep things agnostic, unless it only supports `tar` files for example, we should probably be omitting the extension entirely. 🤔 

---

Regardless, if we do make this chance to the wording, then my suggestion does make sense. 

> Is "`ext` word" referring to `.rar` in the line above?
> Maybe: "Replace the extension in the example above (`rar`)"

So replace the extension, which in the example above happens to be `rar`. Though if you think it's unclear, then perhaps provide don't reference anything else at all.

Just describe it exactly how you described it me now.

> For instance, for the command `engrapma {{path/to/archive.ext}}`. It's better to replace use a concrete file extension like as `tar` or `zip` rather than something generic like `ext`.

SethFalco:
Some general advice for the guidelines.
Let's try not to get into _too_ much detail about everything.

Making it too long will just make it so less will read it.
Explaining things that should be obvious to most can create more confusion than it solves.

The most common example of this is: "Hey, want to try my vegan apple?" (I'm explaining in more detail that the apple is safe for vegans to eat, but by describing it this way it actually creates confusion and instead people wonder what I meant by that.)

> there is a long (consisting of several letters) option

Or things like this, maybe add a definition section at the top which is under a `<details>` tag or link `long option` to a relevant article.

---

I will review this in more detail later, but overall, I actually disagree with many of the changes here because most of the additions aren't actually helping explain anything. It just confuses users, hinders accessibility, or deters users from reading it.

EmilySeville7cfg:
> Making it too long will just make it so less will read it.
Explaining things that should be obvious to most can create more confusion than it solves.

I can move all details to another file. What about this idea? I can't work in tldr when I don't know how I have to review and what to suggest in particular cases.

SethFalco:
> I can't work in tldr when I don't know how I have to review and what to suggest in particular cases.

There are some things that are alright to be preference, or managed case-by-case.
I'm not saying we can't expand the document, just that you don't seem to put any focus on UX as a writer, which won't make for a very pleasant experience for contributors.

I'd rather we put a focus on them.

For example, some of these changes do nothing but make the document longer and uncomfortable to read, i.e.

- Use short but descriptive tokens, such as `{{source_file}}` or `{{wallet.txt}}`.
- Use short but descriptive placeholders. Examples: `{{path/to/source.cs}}` or `{{drive}}`.

The top one is very pleasant for me to read, no pauses and is very natural.
The second one for seemingly no reason breaks the sentences apart, so I feel more like I'm reading output from a program rather.

PR #8253, opened 2022-07-26
aria2, aria2c: add French translation & page edit

Nico385412:

Hi 👋 
A new day a new translation, and also a page edit.

I noticed the english page has an error in his documentation with the `--force-sequential`

According to the documentation it performs a sequential download, not a parallel download as tldr page says.
I suggest to fix this :p 

image

### DIFF
--- a/pages.fr/common/aria2c.md
+++ b/pages.fr/common/aria2c.md
@@ -14,7 +14,7 @@
 
 - Télécharge plusieurs fichiers (différents) en parallèle :
 
-`aria2c --force-sequential {{url_1}} {{url_2}}`
+`aria2c --force-sequential {{false}} {{url_1}} {{url_2}}`
 
 - Télécharge depuis plusieurs sources avec chaque URI pointant vers le même fichier :
 
--- a/pages.fr/common/aria2c.md
+++ b/pages.fr/common/aria2c.md
@@ -1,0 +1,37 @@
+# aria2c
+
+> Utilitaire de téléchargement.
+> Supporte HTTP(S), FTP, SFTP, BitTorrent, et Metalink.
+> Plus d'informations : <https://aria2.github.io>.
+
+- Télécharge depuis une URI vers un fichier :
+
+`aria2c {{url}}`
+
+- Télécharge un fichier via l'url spécifié en choisissant le nom de ce dernier :
+
+`aria2c --out={{nom_de_fichier}} {{url}}`
+
+- Télécharge plusieurs fichiers en parallèle :
+
+`aria2c --force-sequential {{url_1}} {{url_2}}`
+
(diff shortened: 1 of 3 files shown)


Nico385412:
> This flag description is a little confusing, because of the `Default: false`. Does that mean that `--force-sequential true` sets the option to true, but `--force-sequential` sets it to false?

I aggree it's realy confusing 😅. I'll figure it out this afternoon

Nico385412:
Ok then, after a lot of trouble to install it i figure it out
`--force-sequential` without true/false value **download in parallel** 
However if you set the value to true, it will download in sequential.

By default aria2c download in parallel without the option. you dont need the option `--force-sequential` **as long as the urls point to the same file**

However if the files are =/= you can set `--force-sequential` without parameter do achieve it. (sounds more like a trick ...)

So i have to rollback my changes, the en page is correct ^^"

marchersimon:
The example is still very confusing. How about we change it to:
```markdown
- Download multiple (different) files in parallel:

`aria2c --force-sequential {{false}} {{url_1}} {{url_2}}`

```

Nico385412:
> The example is still very confusing. How about we change it to:
> 
> ```gfm
> - Download multiple (different) files in parallel:
> 
> `aria2c --force-sequential {{false}} {{url_1}} {{url_2}}`
> ```

Yes definitely ! i don't what the others thinks ?

EmilySeville7cfg (approved):

marchersimon:
@Nico385412 I think you can apply the change

Nico385412:
> @Nico385412 I think you can apply the change

Done ! @marchersimon

pages.fr/common/aria2c.md
```diff
@@ -0,0 +1,37 @@
...
+
+`aria2c {{url}}`
+
+- Télécharge un fichier via l'url spécifié en choisissant le nom de ce dernier :
+
+`aria2c --out={{nom_de_fichier}} {{url}}`
+
+- Télécharge plusieurs fichiers (différents) en parallèle :
+
+`aria2c --force-sequential {{url_1}} {{url_2}}`
```
marchersimon, line 17:
```suggestion
`aria2c --force-sequential {{false}} {{url_1}} {{url_2}}`
```

pages/common/aria2c.md
```diff
@@ -12,7 +12,7 @@
 
 `aria2c --out={{filename}} {{url}}`
 
-- Download multiple files in parallel:
+- Download multiple (different) files in parallel:
 
 `aria2c --force-sequential {{url_1}} {{url_2}}`
```
marchersimon, line 17:
```suggestion
`aria2c --force-sequential {{false}} {{url_1}} {{url_2}}`
```

marchersimon (approved):
Thanks!

PR #8275, opened 2022-08-01
Patch 1

GlassGruber:

The second command added is an alternative fallback command in case the server is not correctly responding to HEAD requests, but I don't know if the description should explain this in detail, giving more context and the rationale for it.

A bit of info can be found here also 
https://stackoverflow.com/questions/32488162/curl-suppress-response-body/63867207#63867207

### DIFF
--- a/pages/common/curl.md
+++ b/pages/common/curl.md
@@ -40,6 +40,6 @@
 
 `curl --head {{http://example.com}}`
 
-- Alternatively to above (because some servers may not correctly respond to HEAD requests), send a GET request but then print response code and headers and drop the connection without receiving the response body:
+- Send a GET request and print response code with headers but drop the connection without receiving the response body:
 
 `curl --head -X GET {{http://example.com}}`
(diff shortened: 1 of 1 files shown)


GlassGruber:
Also I hope I've filled correctly the template for the PR in regard of "remove items that do not apply".

A new template specifically for PRs  with corrections/additions of already existing commands could be a good improvement?

GlassGruber:
Oh I'm sorry, I failed to see the "at most 8 examples" request. I understand the reason but isn't the limit a bit low? 
Feel free to close this PR if there is nothing to add.

nicokosi:
> Oh I'm sorry, I failed to see the "at most 8 examples" request. I understand the reason but isn't the limit a bit low? Feel free to close this PR if there is nothing to add.

Hello @GlassGruber; :wave: 

I think that the response is in the [Contributing guidelines](https://github.com/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines): :innocent: 

> Try to keep pages at around 5 examples. Pages can be longer or shorter when appropriate, but don't exceed 8 examples. Remember, it's OK if the page doesn't cover everything; that's what man is for.

GlassGruber:
Oh boy, not only I missed the requirement, but I missed also the very first point in the guidelines, but I've read them I swear! 
I need way more coffee today 🙃 
Sorry for bothering and thank you for answer and your time!

PR #8278, opened 2022-08-02
inotifywait: add --monitor command

CleanMachine1:

### DIFF
--- a/pages/linux/inotifywait.md
+++ b/pages/linux/inotifywait.md
@@ -27,6 +27,6 @@
 
 `while inotifywait --event {{modify}} {{path/to/file}}; do {{command}}; done`
 
-- Continuously watch the file without exiting when an event occurs:
+- Continuously watch a specific file for events without exiting:
 
 `while inotifywait --monitor {{path/to/file}}; do {{command}}; done`
--- a/pages/linux/inotifywait.md
+++ b/pages/linux/inotifywait.md
@@ -26,3 +26,7 @@
 - Only watch for file modification events:
 
 `while inotifywait --event {{modify}} {{path/to/file}}; do {{command}}; done`
+
+- Continuously watch the file without exiting when an event occurs:
+
+`while inotifywait --monitor {{path/to/file}}; do {{command}}; done`


pages/linux/inotifywait.md
```diff
@@ -26,3 +26,7 @@
 - Only watch for file modification events:
 
 `while inotifywait --event {{modify}} {{path/to/file}}; do {{command}}; done`
+
+- Continuously watch the file without exiting when an event occurs:
```
mfrw, line 30: How about:
```suggestion
- Continuously watch a specific file for events without exiting:
```

pages/linux/inotifywait.md
```diff
@@ -26,3 +26,7 @@
...
+
+`while inotifywait --monitor {{path/to/file}}; do {{command}}; done`
```
marchersimon, line 32: I assume `{{command}}` is just some kind of dummy command, right? Could we give a specific example?

CleanMachine1: could be anything

the most generic command I can think of is `echo 1`

marchersimon: We should probably give a specific example here. Otherwise users may be confused what they should use for `{{command}}` (at least I was). How about
```suggestion
`while inotifywait --monitor {{path/to/file}}; do sleep 0; done`
```

CleanMachine1:
```suggestion
`while inotifywait --monitor {{path/to/file}}; do {{sleep 5}}; done`
```

sleep 0 is nothing.

mfrw (approved):
LGTM once we incorporate the `command` suggestion :)

pages/linux/inotifywait.md
marchersimon: But isn't the use of `command` to do exactly nothing? I assumed it was kind of an `noop` command.

navarroaxel: I think `command` is ok, otherwise `echo` with a text is more explicit.

CleanMachine1: I think, in the context of what this command is used for, a command which gives an output or does something to the system would be best, 

echo 1 is my best suggestion.

marchersimon: I now tested this command and it works completely fine without the while loop. The loop doesn't make any difference.

In fact, `{{command}}` is never executed, because `inotifywait --monitor file` never returns a value, because it doesn't exit on its own.

CleanMachine1: Could you provide a suggestion message

CleanMachine1: Im a little unsure what you would like to change, so if we merge this, would you be willing to make the changes and the PR?

marchersimon: Sure!

marchersimon (approved):
Thanks

PR #8283, opened 2022-08-02
cc: add alias page

Ciubix8513:

### DIFF
--- a/pages/linux/cc.md
+++ b/pages/linux/cc.md
@@ -1,0 +1,24 @@
+# cc
+
+> Alias for gcc, preprocess and compile C and C++ source files, then assemble and link them together.
+> More information: <https://gcc.gnu.org>.
+
+- Compile multiple source files into executable:
+
+`cc {{path/to/source1.c path/to/source2.c ...}} --output {{path/to/output_executable}}`
+
+- Allow warnings, debug symbols in output:
+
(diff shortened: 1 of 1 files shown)


pages/linux/cc.md
```diff
@@ -0,0 +1,24 @@
...
+
+`cc {{path/to/source.c}} --output {{path/to/output_executable}} -I{{path/to/header}} -L{{path/to/library}} -l{{library_name}}`
+
+- Compile source code into Assembler instructions:
+
+`cc -S {{path/to/source.c}}`
+
+- Compile source code without linking:
+
+`cc -c {{path/to/source.c}}`
```
marchersimon, line 24: Thanks for the page. However, we do have a specific template to handle binary symlinks (aliases). The page should look like this:
```suggestion
> This command is an alias of `gcc`.
> More information: <https://gcc.gnu.org>.

- View documentation for the original command:

`tldr gcc`
```
Also, since `gcc` is in the `common` directory, this page sholud also be in `common`, not in `linux`. If you want you could contribute your examples to the original `gcc` page. (However, only eight examples are allowed).

marchersimon (approved):
Thanks @Ciubix8513

kbdharun (approved):
Looks good

blueskyson (approved):

Issue #8046, opened 2022-04-23
favour or at least mention single-letter options


blueglyph:
The [guildelines](https://github.com/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines) specify:

> Try to incorporate the spelled-out version of single-letter options in the example's description. The goal is to allow people to understand the syntax of the commands, not just memorize it.

Actually the purpose is to give practical examples of commands that are often used. For example, nobody ever types

```sudo pacman --sync --refresh --sysupgrade```

Instead, people type:

```sudo pacman -Syu```

That's also what you'll see people try to help you out. "_Have you tried pacman -Syu?_". Hmm, what does it do, let's check with tldr... and no luck, it's not there. Back to the man pages then.

So while it is interesting to see there are spelled-out options, it is redundant with the description above the command (_Synchronize and update all packages:_), and not very practical. Sure, I can check the man page for find the equivalent, but then this makes these pages irrelevant.

As a suggestion, it would be best to have at least the practical alternatives for long commands:

```
Synchronize and update all packages:

      sudo pacman --sync --refresh --sysupgrade
      sudo pacman -Syu
```

pixelcmtd:
I have to agree with both the original rationale and your example of it failing.

Our goal is to provide users an alternative for quickly looking up how to use a command in a beginner-friendly way, not how to master it. That's why we should, as we currently do, prefer GNU options.

But yes, nobody types `pacman --sync --refresh --sysupgrade` into their terminal, and if they would, they would waste their time massively. Everyone uses `pacman -Syu` without remembering (or maybe not even having known in the first place) what any of that actually means. I can definitely see, how the latter would be more helpful here (mentioning both like you suggested is also somewhat of a good idea).

I think these sentences at least need a little change: GNU (or long) options are preferred/the default, but Unix (or short) options are fine in certain cases.

Another option, and I think that this would be even better, is a standardized `or` for options. We already use pipes for `or` (i.e. `tldr 7z`) and some pages even already use it for options (i.e. `tldr grep`: `--{{context|before-context|after-context}}={{3}}`). So I'd suggest, whenever there is a GNU style option, but the Unix style option is used by everybody, to use a syntax like this: `{{--sync|-S}}`

CleanMachine1:
From what I can tell, the unwritten rule is `use the short if it is really common`

for example From `tldr git commit`

  - Commit staged files to the repository with a message:
    git commit -m "message"

[...]

PR #8286, opened 2022-08-03
pacman: use short options only

marchersimon:
For #8046 
I felt like the whole issue is mostly about `pacman`.
So we should start with this one and see if there still are other commands where it is necessary to include short options.
I put the short options in token syntax for now, but we can debate about that.


CleanMachine1:
Personally don't like [this current version](https://github.com/tldr-pages/tldr/pull/8286/commits/bbd431c01629241305785d44ed9ff7f4a636cf77).

I think we just pick one, and since its such a common command, maybe we should go with `sudo pacman -Syu`. I've probably said something else in the past as well.

marchersimon:
I just think that it's useful to have the long version, since you directly know what all options do and what would happen if you left out a specific one. But I totally get your point, that it looks weird.

kbdharun:
> I think we just pick one, and since it's such a common command, maybe we should go with `sudo pacman -Syu`. I've probably said something else in the past as well.

Yeah, I think the same too, Even though long options are useful and very informative. I think many would prefer shorter options as it is the one that is widely used. I think the [current implementation](https://github.com/tldr-pages/tldr/pull/8286/commits/bbd431c01629241305785d44ed9ff7f4a636cf77) might confuse users about which command to execute. I went through the issue about it fully and I noticed there were a lot of suggestions and proposed solutions. 

I agree with @CleanMachine1 as this is a special case it is better to use shorter options on the page if users need to refer to longer options they can use the man link provided with `More information:`.

kbdharun (approved):
LGTM, Thanks @marchersimon

EmilySeville7cfg:
I don't like idea to use short options when there are long options supported by command. From my perspective it's better to copy `pacman` page to `common/` and use short options there but not here. Either way it depends on what can be placed in `common/`. And I am going to discuss it in out gitter chat. I'll leave link here for my message.

EmilySeville7cfg (changes requested):
I don't like idea to use short options when there are long options supported by command. From my perspective it's better to copy `pacman` page to `common/` and use short options there but not here. Either way it depends on what can be placed in `common/`. And I am going to [discuss](https://gitter.im/tldr-pages/tldr?at=62f0f457aa09177429231d33) it in out gitter chat.

marchersimon:
I don't see why we should put this into `common`, it's only available on `linux`. Also, if you are on linux and look up the tldr page for `pacman`, you will only see the `linux` version, not the `common` version.

kbdharun:
> I don't see why we should put this into `common`, it's only available on `linux`. Also, if you are on linux and look up the tldr page for `pacman`, you will only see the `linux` version, not the `common` version.

+1, I too think the same.

sbrl (approved):
The `pacman` page should *not* be in `common`. `common` is only for pages about commands that are supported by **more than 1 platform**.

Further, as I have explained in another issue recently, we tldr pages should document usages of commands that are *useful*. Sure, it's helpful to know what the short options stand for, but there are other ways to achieve that than requiring long options everywhere. `sudo pacman --sync --refresh --sysupgrade` is an enormous mouthful to remember and type every time - I'll bet most users use `sudo pacman -Syu` instead.

We never used to require long options on all pages, and that particular guideline was never intended to mean "use long options everywhere". I can dig up multiple references to this if there's interest.

EmilySeville7cfg:
> The pacman page should not be in common. common is only for pages about commands that are supported by more than 1 platform.

@sbrl, what about not to use common at all? It's not clear what platforms on some command is available when it is in `common/`?

> Further, as I have explained in another issue recently, we tldr pages should document usages of commands that are useful. Sure, it's helpful to know what the short options stand for, but there are other ways to achieve that than requiring long options everywhere. sudo pacman --sync --refresh --sysupgrade is an enormous mouthful to remember and type every time - I'll bet most users use sudo pacman -Syu instead.

Let's define subset of pages where we prefer short options. For instance in package manager pages we can prefer short options anywhere. I can't decide whether to approve or decline such PRs without formal rules. I have nothing to refer to. Note that when I block merging by requesting changes it doesn't mean that I don't like changes made in PR at all. I just wanna everything to be standardized.

kbdharun:
> Let's define subset of pages where we prefer short options. For instance in package manager pages we can prefer short options anywhere. I can't decide whether to approve or decline such PRs without formal rules. I have nothing to refer to. Note that when I block merging by requesting changes it doesn't mean that I don't like changes made in PR at all. I just wanna everything to be standardized and that's why I am here.

Yeah, it would be better to specify the use cases for short commands over longer ones for PR's, maybe we can document the edge cases too in `contributing.md`. What is your opinion? @marchersimon @CleanMachine1

pages/linux/pacman.md
```diff
@@ -6,32 +6,32 @@
 
 - Synchronize and update all packages:
 
-`sudo pacman --sync --refresh --sysupgrade`
+`sudo pacman -Syu`
```
EmilySeville7cfg, line 9: All short options should be described via square brackets where possible here and in all translations.

EmilySeville7cfg (commented):
Please describe all short options via square brackets in all pages changed in PR.

navarroaxel (approved):

EmilySeville7cfg:
> Yeah, it would be better to specify the use cases for short commands over longer ones for PR's, maybe we can document the edge cases too in contributing.md. What is your opinion? @marchersimon @CleanMachine1

Style guide maybe the better place to document it.

EmilySeville7cfg (approved):

kbdharun:
> Style guide maybe the better place to document it.

Yeah, That would be good. To place it under special cases under style guide.

PR #8290, opened 2022-08-03
tlmgr-dump-tlpdb, tlmgr-recreate-tlpdb: add page

marchersimon:


pages/common/tlmgr-recreate-tlpdb.md
```diff
@@ -0,0 +1,9 @@
+# tlmgr recreate-tlpdb
+
+> Recreate the TeX Live package database.
+> This command has no documentation and therefore should probably not be used by humans.
```
CleanMachine1, line 4:
```suggestion
> This command should not be invoked by the user. 
```

marchersimon: Well, the thing is, I don't know if it shouldn't. I just assumed it, because there was absolutely no documentation, except for the source code itself.

navarroaxel: I think this command exists to be used under certain circumstances, also there is a `--platform` argument.
 ```suggestion
```

Comment (marchersimon) [reply pages/common/tlmgr-recreate-tlpdb.md:4]: I wouldn't be so sure about that. There are even debug outputs in the source code. Also, this command has to be run as root, which can become risky if it isn't properly documented.

Comment (navarroaxel) [reply pages/common/tlmgr-recreate-tlpdb.md:4]: I just looked around for this command, and I found some threads where this command is the answer to some errors.

Comment (marchersimon) [reply pages/common/tlmgr-recreate-tlpdb.md:4]: How about
```suggestion
> This command should be used with caution, because it is poorly documented.
```

sbrl: That looks like a good suggestion, @marchersimon

navarroaxel:
@CleanMachine1 are we ok with this PR?

pages/common/tlmgr-recreate-tlpdb.md
```diff
@@ -0,0 +1,9 @@
+# tlmgr recreate-tlpdb
+
+> Recreate the TeX Live package database.
+> This command should be used with caution, because it is poorly documented.
```
CleanMachine1, line 4: Not completely sure this line should be here.

CleanMachine1 (approved):
Yes, LGTM other than my 1 suggestion.

pages/common/tlmgr-recreate-tlpdb.md
marchersimon: Well, the command is clearly not meant to be used by end-users. I assume it's for developers only, which means there is no guarantee that it is safe to use in all cases, especially since you have to run it as root.
I'd say we either warn the user or we don't document the command at all.

CleanMachine1: Ok, I think the current wording is a little too negative, how do you feel about

```suggestion
> This command has a lack of documentation and should be used with caution. 
```

navarroaxel: Both sounds good, @marchersimon which do you like more?

navarroaxel (approved):

PR #8292, opened 2022-08-03
cc, g++, gcc: add Tamil translation

kbdharun:
- [x] The page(s) are in the correct platform directories: `common`, `linux`.


pages.ta/common/gcc.md
```diff
@@ -0,0 +1,24 @@
+# gcc
+
+> C மற்றும் C++ மூலக் கோப்புகளை முன் செயலாக்கம் செய்து தொகுத்து, பின்னர் அவற்றைச் சேகரித்து இணைக்கவும்.
+> மேலும் தகவல்: <https://gcc.gnu.org>.
+
+- பல மூல கோப்புகளை இயங்கக்கூடியதாக தொகுக்கவும்:
+
+`gcc {{பாதை /க்கு / மூல1.c பாதை /க்கு / மூல2.c ...}} --output {{பாதை/க்கு/வெளியீடு_இயங்கக்கூடியது}}`
```
marchersimon, line 8: In https://github.com/tldr-pages/tldr/blob/main/contributing-guides/translation-templates/common-arguments.md we have the Tamil translation `கோப்பு` for `path/to/file`. Should we use this or your version?

kbdharun: `கோப்பு/டு/மூல1.c` would become `file/to/source.c` and  but in this case, we are using `path to source`, because path means பாதை, and `கோப்பு` means `file`.  As mentioned in [Link](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/translation-templates/common-arguments.md) , in this case I think `path to source` would better align with the main pages. 

Source means `மூல` in Tamil so I had used `பாதை /க்கு / மூல1.c` , I will update this to `பாதை /டு /மூல.c` for better readability.

marchersimon: So do you think we should update our `common-arguments.md` translation guide?

kbdharun: > So do you think we should update our `common-arguments.md` translation guide?

Yeah, I think it would be better to update it. 
image
image

As `அடைவு` means `directory` only and not `path\to\directory` and `கோப்பு` means `file` only not `path\to\file` . So we might need to update these 3 pages.

marchersimon (approved):
👍🏽

navarroaxel (approved):

PR #8302, opened 2022-08-04
hdparm: add read speed example

marchersimon:
version v9.63


CleanMachine1 (approved):

mfrw (approved):
LGTM. Thanks @marchersimon 

I think this specific example times both the `cache reads` & `device reads` :)

- `-t` -> `device reads`
- `-T` -> `cache reads`

PR #8308, opened 2022-08-05
page-request.yml: reword and remove parts

CleanMachine1:
- [ ] The page(s) are in the correct platform directories: `common`, `linux`, `osx`, `windows`, `sunos`, `android`, etc.

### DIFF
--- a/.github/ISSUE_TEMPLATE/page-request.yml
+++ b/.github/ISSUE_TEMPLATE/page-request.yml
@@ -17,12 +17,13 @@ body:
     attributes:
       label: OS
       description: Select which operating system(s) this command works with.
-      required: true
       options:
-
         - label: Android
         - label: macOS
         - label: Windows
         - label: Linux
         - label: sunOS
+        validations:
+          required: true
+
         
--- a/.github/ISSUE_TEMPLATE/page-request.yml
+++ b/.github/ISSUE_TEMPLATE/page-request.yml
@@ -23,7 +23,5 @@ body:
         - label: macOS
         - label: sunOS
         - label: Windows
-     validations:
-      required: true
 
         
--- a/.github/ISSUE_TEMPLATE/page-request.yml
+++ b/.github/ISSUE_TEMPLATE/page-request.yml
@@ -17,9 +17,9 @@ body:
     attributes:
       label: OS
       description: Select which operating system(s) this command works with
-      validations:
-        required: true
       options:
+        validation:
+          required: true
         - label: Android
         - label: macOS
         - label: Windows
--- a/.github/ISSUE_TEMPLATE/page-request.yml
+++ b/.github/ISSUE_TEMPLATE/page-request.yml
@@ -7,9 +7,7 @@ body:
     attributes:
       label: Command description
       description: Describe a command you want to be summarized.
-      placeholder: Tell us what TLDR page you want to see!
-    validations:
-      required: true
+      placeholder: Tell us about the command!
   - type: input
     attributes:
       label: Command details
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/page-request.yml
```diff
@@ -25,16 +23,7 @@ body:
...
         - Windows
-        - General
         - Other
+        - More than one operating system
     validations:
       required: true
-  - type: input
-    attributes:
-      label: OS details
-      description: Describe any details related to an operating system.
```
navarroaxel, line 35: I think this field should be used to include one example usage of the command. no?

.github/ISSUE_TEMPLATE/page-request.yml
```diff
@@ -25,16 +23,7 @@ body:
         - Android
         - macOS
         - Windows
-        - General
         - Other
+        - More than one operating system
```
EmilySeville7cfg, line 27:
```suggestion
        - More than one operating system
```

What about to use `multiple: true` instead of adding the new choice?

EmilySeville7cfg (changes requested):

.github/ISSUE_TEMPLATE/page-request.yml
mfrw: Yep, I second the idea of check-boxes :)

marchersimon: But it seems a little confusing when you have both. So, when you check `More than one operating system`, what should you select in the dropdown menu?

.github/ISSUE_TEMPLATE/page-request.yml
marchersimon: Sounds like a good idea, but it should be optional

.github/ISSUE_TEMPLATE/page-request.yml
EmilySeville7cfg: @marchersimon, I meant to replace the dropdown with checkboxes.

marchersimon: Oh, my bad. It is just often not easy to find out on which platforms a command works exactly, just that it works on more than one. For example, take `git`. It definitely works on `linux`, `osx`, and `windows`. But does it work on `sunos` and `android`? And is it important to know that? It would go into the `common` platform anyways. 

So I think we should include the `common` platform as an option. (Even tough I prefer radio boxes to a dropdown menu).

EmilySeville7cfg: We can't list all possible platforms. So yeah, it's nice idea to include `common` platform for commands working on all listed platforms.

marchersimon (approved):
There are still some interesting suggestions, but I already approve, since they are not that important.

.github/ISSUE_TEMPLATE/page-request.yml
```diff
@@ -1,40 +1,28 @@
...
-        e.g. OS version obtained via:
-        `uname -a` for Linux
-        `sw_vers` for macOS
-        `ver` for Windows
+        - label: Android
+        - label: macOS
+        - label: Windows
+        - label: Linux
+        - label: sunOS
+
```
navarroaxel, line 28: Alpha order is better, also there is an extra enter there.
```suggestion
        - label: Linux
        - label: macOS
        - label: sunOS
        - label: Windows       
```

CleanMachine1:
I need some help here, I don't have much experience with the syntax here and I don't exactly know what I am doing. Everytime I get a new error after every commit.

mfrw (approved):
LGTM :)
I guess we should wait for other folks as well IMO :)

EmilySeville7cfg (approved):

EmilySeville7cfg:
Merged, thanks! 🚀

Issue #8307, opened 2022-08-05
page request: gum


CleanMachine1:
### Command description

.

### OS

Other

kbdharun:
https://github.com/charmbracelet/gum
Is this the repo for this command or is it a different program .

CleanMachine1:
Yes that is the repo.

PR #8314, opened 2022-08-06
gum: add page

CleanMachine1:
- **Version of the command being documented (if known):** gum version 0.4.0 (131ae3b)

### DIFF
--- a/pages/common/gum.md
+++ b/pages/common/gum.md
@@ -11,7 +11,7 @@
 
 `gum input --placeholder "{{value}}"`
 
-- Interactvely pick an option which will exit if "No" is chosen:
+- Open an interactive confirmation prompt and exit with either `0` or `1`:
 
 `gum confirm "{{Continue?}}" --default=false --affirmative "{{Yes}}" --negative "{{No}}" {{&& echo "Yes selected" || echo "No selected"}}`
 
--- a/pages/common/gum.md
+++ b/pages/common/gum.md
@@ -21,7 +21,7 @@
 
 - Format text to include emojis:
 
-`echo "{{:smile: :heart hello}}" | gum format -t {{emoji}}`
+`echo "{{:smile: :heart: hello}}" | gum format -t {{emoji}}`
 
 - Interactively prompt for multi-line text (CTRL + D to save) and write to `data.txt`:
 
--- a/pages/common/gum.md
+++ b/pages/common/gum.md
@@ -15,7 +15,6 @@
 
 `gum confirm "{{Continue?}}" --default=false --affirmative "{{Yes}}" --negative "{{No}}" {{&& : || exit 1}}`
 
-
 - Show a spinner while a command is taking place with text alongside:
 
 `gum spin --spinner {{dot|line|minidot|jump|pulse|points|globe|moon|monkey|meter|hamburger}} --title "{{loading...}}" -- {{command}}`
--- a/pages/common/gum.md
+++ b/pages/common/gum.md
@@ -21,7 +21,7 @@
 
 - Format text to include emojis:
 
-`echo "{{:smile: :heart: hello}}" | gum format -t {{emoji}}`
+`gum format -t {{emoji}} "{{:smile: :heart: hello}}"`
 
 - Interactively prompt for multi-line text (CTRL + D to save) and write to `data.txt`:
 
--- a/pages/common/gum.md
+++ b/pages/common/gum.md
@@ -7,7 +7,7 @@
 
 `gum choose "{{option_1}}" "{{option_2}}" "{{option_3}}"`
 
-- Open an interactive prompt for the user to input a string with a placeholder:
+- Open an interactive prompt for the user to input a string with a specific placeholder:
 
 `gum input --placeholder "{{placeholder_string}}"`
 
--- a/pages/common/gum.md
+++ b/pages/common/gum.md
@@ -3,7 +3,7 @@
 > A tool for making glamorous shell scripts.
 > More information: <https://github.com/charmbracelet/gum>.
 
-- Interactively pick an option to print to `STDOUT`:
+- Interactively pick a specific option to print to `stdout`:
 
 `gum choose "{{option_1}}" "{{option_2}}" "{{option_3}}"`
 
--- a/pages/common/gum.md
+++ b/pages/common/gum.md
@@ -9,7 +9,7 @@
 
 - Open an interactive prompt for the user to input a string with a specific placeholder:
 
-`gum input --placeholder "{{placeholder_string}}"`
+`gum input --placeholder "{{value}}"`
 
 - Interactvely pick an option which will exit if "No" is chosen:
 
--- a/pages/common/gum.md
+++ b/pages/common/gum.md
@@ -1,0 +1,29 @@
+# gum
+
+> A tool for making glamorous shell scripts.
+> More information: <https://github.com/charmbracelet/gum>.
+
+- Interactively pick an option to print to `STDOUT`:
+
+`gum choose "{{option_1}}" "{{option_2}}" "{{option_3}}"`
+
+- Open an interactive prompt for the user to input a string with a placeholder:
+
+`gum input --placeholder "{{placeholder_string}}"`
+
+- Interactvely pick an option which will exit if "No" is chosen:
+
+`gum confirm "{{Continue?}}" --default=false --affirmative "{{Yes}}" --negative "{{No}}" {{&& : || exit 1}}`
+
+
+- Show a spinner while a command is taking place with text alongside:
+
+`gum spin --spinner {{dot|line|minidot|jump|pulse|points|globe|moon|monkey|meter|hamburger}} --title "{{loading...}}" -- {{command}}`
+
+- Format text to include emojis:
+
+`echo "{{:smile: :heart hello}}" | gum format -t {{emoji}}`
+
+- Interactively prompt for multi-line text (CTRL + D to save) and write to `data.txt`:
+
+`gum write > {{data.txt}}`
--- a/pages/common/gum.md
+++ b/pages/common/gum.md
@@ -13,7 +13,7 @@
 
 - Interactvely pick an option which will exit if "No" is chosen:
 
-`gum confirm "{{Continue?}}" --default=false --affirmative "{{Yes}}" --negative "{{No}}" {{&& : || exit 1}}`
+`gum confirm "{{Continue?}}" --default=false --affirmative "{{Yes}}" --negative "{{No}}" {{&& echo "Yes selected" || echo "No selected"}}`
 
 - Show a spinner while a command is taking place with text alongside:
 


CleanMachine1:
closes #8307

pages/common/gum.md
```diff
@@ -0,0 +1,28 @@
+# gum
+
+> A tool for making glamorous shell scripts.
+> More information: <https://github.com/charmbracelet/gum>.
+
+- Interactively pick an option to print to `STDOUT`:
+
+`gum choose "{{option_1}}" "{{option_2}}" "{{option_3}}"`
```
EmilySeville7cfg, line 8:
```suggestion
`gum choose "{{option1}}" "{{option2}}" "{{option3}}"`
```

pages/common/gum.md
```diff
@@ -0,0 +1,28 @@
+# gum
+
+> A tool for making glamorous shell scripts.
+> More information: <https://github.com/charmbracelet/gum>.
+
+- Interactively pick an option to print to `STDOUT`:
```
EmilySeville7cfg, line 6:
```suggestion
- Interactively pick a specific option to print to `STDOUT`:
```

pages/common/gum.md
```diff
@@ -0,0 +1,28 @@
...
+
+`gum choose "{{option_1}}" "{{option_2}}" "{{option_3}}"`
+
+- Open an interactive prompt for the user to input a string with a placeholder:
```
EmilySeville7cfg, line 10:
```suggestion
- Open an interactive prompt for the user to input a string with a specific placeholder:
```

pages/common/gum.md
```diff
@@ -0,0 +1,28 @@
...
+
+`gum input --placeholder "{{placeholder_string}}"`
+
+- Interactvely pick an option which will exit if "No" is chosen:
+
+`gum confirm "{{Continue?}}" --default=false --affirmative "{{Yes}}" --negative "{{No}}" {{&& : || exit 1}}`
```
EmilySeville7cfg, line 16:
```suggestion
`gum confirm "{{Continue?}}" --default=false --affirmative "{{Yes}}" --negative "{{No}}" {{|| exit 1}}`
```

pages/common/gum.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Show a spinner while a command is taking place with text alongside:
+
+`gum spin --spinner {{dot|line|minidot|jump|pulse|points|globe|moon|monkey|meter|hamburger}} --title "{{loading...}}" -- {{command}}`
```
EmilySeville7cfg, line 20: I don't like too many options listed inside curly braces. I think we should not do that not to introduce extra complexity in tldr pages. Maybe we can limit alternatives up to 5 (inclusive).

EmilySeville7cfg (changes requested):
Here are my first thoughts. Maybe if I use  this command I imagine how this page can be enhanced. ;)

pages/common/gum.md
kbdharun: Is STDOUT in caps necessary?

pages/common/gum.md
kbdharun: Hey, Shouldn't this line be
```suggestion
- Open an interactive prompt for the user to input a commit message:
``` 
[Reference](https://github.com/charmbracelet/gum#readme)
image 
image

pages/common/gum.md
```diff
@@ -0,0 +1,28 @@
...
+> A tool for making glamorous shell scripts.
+> More information: <https://github.com/charmbracelet/gum>.
+
+- Interactively pick an option to print to `STDOUT`:
+
+`gum choose "{{option_1}}" "{{option_2}}" "{{option_3}}"`
+
+- Open an interactive prompt for the user to input a string with a placeholder:
+
+`gum input --placeholder "{{placeholder_string}}"`
```
kbdharun, line 12:
```suggestion
`gum input --placeholder "{{placeholder_value}}"`
```

pages/common/gum.md
kbdharun: Yeah, I too think the same. Maybe we can reduce it to some essential examples as displaying it would be complex for the client. 
(Or)
We can provide only one example as given in the [Github Page](https://github.com/charmbracelet/gum#spin).
```suggestion
`gum spin --spinner {{dot}} --title "{{title}}" -- {{command}}`
```

pages/common/gum.md
EmilySeville7cfg: I like this but our style guide doesn't provide any info about how we have to write stdin, stdout, stderr.

pages/common/gum.md
EmilySeville7cfg, line 12:
```suggestion
`gum input --placeholder "{{value}}"`
```

What about omit `placeholder` term from the placeholder? I don't like this repetition.

pages/common/gum.md
EmilySeville7cfg, line 20:
```suggestion
`gum spin --spinner {{dot|line|minidot|jump|pulse|...}} --title "{{loading...}}" -- {{command}}`
```

Let's use `...` when there are too many alternatives to list.

EmilySeville7cfg (changes requested):

pages/common/gum.md
kbdharun: Yeah, This seems better. It would fix the [issue](https://github.com/tldr-pages/tldr/pull/8314#discussion_r939615534).

pages/common/gum.md
CleanMachine1: > Hey, Shouldn't this line be
> 
> [Reference](https://github.com/charmbracelet/gum#readme) image image

No, since the command doesn't need to be used for a commit message or whatever, that is just an example they provide.

pages/common/gum.md
CleanMachine1: IMO, yes since it gives the user an idea that you can add a command for if `Yes` is chosen.

Maybe we could change it to `{{&& echo "Yes selected" || echo "No selected"}}

pages/common/gum.md
CleanMachine1: see `tldr nmcli`, its better to just show all of the options since otherwise the user may not know that the option exists. The spinner choices aren't well documented from what I can see, I found them by going [here](https://github.com/charmbracelet/gum/blob/main/spin/spinners.go)

CleanMachine1: This could work, although as the user I have no idea of what the other options are and since they aren't obvious like when we use {[file1 file2 ...}}, I'd prefer to keep them.

pages/common/gum.md
CleanMachine1: Looking at what we currently use, lowercase is more used.

pages/common/gum.md
CleanMachine1: There can also be as many options, can we include a ...

pages/common/gum.md
EmilySeville7cfg: > aybe we could change it to {{&& echo "Yes selected" || echo "No selected"}}

I like this variant.

pages/common/gum.md
EmilySeville7cfg: I think we can list all options just when docs don't describe them all and user have to dig into source code to understand what can be used.

EmilySeville7cfg (changes requested):
My suggestions are updated ;)

pages/common/gum.md
CleanMachine1: This is incorrect syntax, 

the each option must be encased with quotation marks.

What do you think of

```suggestion
`gum choose "{{option_1}}" "{{option_2}}" "{{option_3}}" "{{...}}"`
```

EmilySeville7cfg: > the each option must be encased with quotation marks.

But it is [valid](https://github.com/tldr-pages/tldr/blob/main/pages/linux/zip.md) for directories:

```
{{path/to/directory1 path/to/directory2 ...}}
```

🤔

CleanMachine1: $ gum choose "one two three"
> one two three

This is not what we want. We want it to be
> One 
> Two
> Three

since the current output gives one "option"

EmilySeville7cfg: We can put quotes inside placeholder.

mfrw: Let me watch this .. I am the most confused person when it comes to `placeholder` syntax :)
👀

pages/common/gum.md
mfrw: In my opinion, our aim is to help the `user`; and if the options are not absolutely obvious and can fit on an example, we should have them.
Although, there is may be a balance that we folks might want to strike where we do not make it too verbose :)

pages/common/gum.md
marchersimon: @EmilySeville7cfg we always put quotes outside of the placeholders. Because the quote itself is not part of the argument, it is always required. In all our English pages, we have almost 700 matches for `"{{...}}"` and only 7 for `{{"..."}}` (which we should fix).

EmilySeville7cfg: Got it.

EmilySeville7cfg (approved):

pages/common/gum.md
marchersimon: This example with all possible options is exactly as long as the previous one. I think we can keep them all.

pages/common/gum.md
```diff
@@ -0,0 +1,28 @@
...
+
+`gum confirm "{{Continue?}}" --default=false --affirmative "{{Yes}}" --negative "{{No}}" {{&& echo "Yes selected" || echo "No selected"}}`
+
+- Show a spinner while a command is taking place with text alongside:
+
+`gum spin --spinner {{dot|line|minidot|jump|pulse|points|globe|moon|monkey|meter|hamburger}} --title "{{loading...}}" -- {{command}}`
+
+- Format text to include emojis:
+
+`echo "{{:smile: :heart hello}}" | gum format -t {{emoji}}`
```
CleanMachine1, line 24:
```suggestion
`echo "{{:smile: :heart: hello}}" | gum format -t {{emoji}}`
```

pages/common/gum.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Interactively pick a specific option to print to `stdout`:
+
+`gum choose "{{option_1}}" "{{option_2}}" "{{option_3}}"`
+
+- Open an interactive prompt for the user to input a string with a specific placeholder:
+
+`gum input --placeholder "{{value}}"`
+
+- Interactvely pick an option which will exit if "No" is chosen:
```
marchersimon, line 14: Doesn't the command exit in both cases, just with different exit values?
```suggestion
- Open an interactive confirmation prompt and exit with either `0` or `1`:
```
I know, it is not limited to confirmation prompts, but that obviously what the subcommand was meant for

pages/common/gum.md
```diff
@@ -0,0 +1,28 @@
...
+
+`gum confirm "{{Continue?}}" --default=false --affirmative "{{Yes}}" --negative "{{No}}" {{&& echo "Yes selected" || echo "No selected"}}`
+
+- Show a spinner while a command is taking place with text alongside:
+
+`gum spin --spinner {{dot|line|minidot|jump|pulse|points|globe|moon|monkey|meter|hamburger}} --title "{{loading...}}" -- {{command}}`
+
+- Format text to include emojis:
+
+`echo "{{:smile: :heart: hello}}" | gum format -t {{emoji}}`
```
marchersimon, line 24: Why not like this? Not all subcommands accept taking the text from stdin.
```suggestion
`gum format -t {{emoji}} "{{:smile: :heart: hello}}"`
```

CleanMachine1: sure, I see no issue

marchersimon (approved):
Looks good, thanks.

PR #8323, opened 2022-08-08
inotifywait: refresh

marchersimon:
This is mostly to remove the confusing `while` loops. 
I'm not so sure about the last example, but it kind of was there in the previous version, and I didn't want to completely remove it.

### DIFF
--- a/pages/linux/inotifywait.md
+++ b/pages/linux/inotifywait.md
@@ -3,30 +3,34 @@
 > Waits for changes to one or more files.
 > More information: <https://manned.org/inotifywait>.
 
-- Run a command when a file changes:
+- Watch a specific file for events, exiting after the first one:
 
-`while inotifywait {{path/to/file}}; do {{command}}; done`
-
-- Be quiet about watching for changes:
-
-`while inotifywait --quiet {{path/to/file}}; do {{command}}; done`
-
-- Watch a directory recursively for changes:
-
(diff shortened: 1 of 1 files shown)


pages/linux/inotifywait.md
```diff
@@ -3,30 +3,34 @@
...
 
-- Continuously watch a specific file for events without exiting:
+- Watch a file printing only events, and no status messages:
+
+`inotifywait --quiet {{path/to/file}}`
+
+- Run a command when a file is accessed:
 
-`while inotifywait --monitor {{path/to/file}}; do {{command}}; done`
+`inotifywait --event {{access}} {{path/to/file}} && {{command}}`
```
CleanMachine1, line 36: This doesn't seem to do anything, it doesn't execute the command.

marchersimon: Did you test it with an empty file? For some reason reading from an empty file doesn't emit an access event. So this command only works if the file isn't empty.

CleanMachine1: No, the file had some data in it.

CleanMachine1: Never mind, it was because I had --monitor.

CleanMachine1 (approved):

kbdharun (approved):

PR #8333, opened 2022-08-09
Added disclaimer to README.md file

shane-squatch:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+## PLEASE READ THIS DOCUMENT BEFORE CONTINUING
+
 <div align="center">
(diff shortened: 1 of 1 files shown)


kbdharun:
@shane-squatch you need to sign the CLA, check the above comment.

marchersimon (commented):
Hmm, I don't really see the need for this change. With `continue`, do you mean continuing to use `tldr` or continue to contribute to `tldr`? In any case, isn't it kind of clear, that the README has to be read?

kbdharun:
> Hmm, I don't really see the need for this change. With `continue`, do you mean continuing to use `tldr` or continue to contribute to `tldr`? In any case, isn't it kind of clear, that the README has to be read?

+1, I too think the same.

PR #8340, opened 2022-08-11
perl: fix -i option description

izokina:
Description was kind of misleading, since optional parameter for -i provides extension for saving backup, not the edited file.
Basically I've only edited description for one option, so no checklist boxes apply :)

- [ ] Check grammar.

### DIFF
--- a/pages/common/perl.md
+++ b/pages/common/perl.md
@@ -19,7 +19,7 @@
 
 `perl -d {{script.pl}}`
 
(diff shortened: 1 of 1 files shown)


mfrw (approved):
Thanks @izokina.
Great catch!

kbdharun (approved):
LGTM, Thanks for your contribution

PR #8344, opened 2022-08-14
tailscale: add debug example

mfrw:
Signed-off-by: Muhammad Falak R Wani <falakreyaz@gmail.com>

PS: Instead of having `help` subcommand which the tailscale cli anyway suggests, document the hidden `debug` subcommand which is very useful to get more details about a running `tailscale` instance.
Reference: [cmd/tailscale/cli/cli.go](https://github.com/tailscale/tailscale/blob/main/cmd/tailscale/cli/cli.go#L184)

### DIFF
--- a/pages/common/tailscale.md
+++ b/pages/common/tailscale.md
@@ -32,6 +32,6 @@
 
 `tailscale bugreport`
 
-- Display help for a subcommand:
+- Display debug helper commands (undocumented):
 
-`tailscale {{subcommand}} --help`
+`tailscale debug --help`


pages/common/tailscale.md
```diff
@@ -32,6 +32,6 @@
 
 `tailscale bugreport`
 
-- Display help for a subcommand:
+- Display debug helper commands (undocumented):
```
navarroaxel, line 35: it's documented, it has a help page in the terminal.

pages/common/tailscale.md
```diff
@@ -32,6 +32,6 @@
...
 
-`tailscale {{subcommand}} --help`
+`tailscale debug --help`
```
navarroaxel, line 37: Not sure, why not create the `tailscale-debug` page?

mfrw: Its an unstable subcommand. So i thought just to give a heads up to the reader; when the user lands in the main `tailscale` tldr-page.
This may prove as a good starting point to debug.
Personally, I was looking for something like this and given `tailscale --help` does not show its existence, its unlikely anyone would know what they are looking for.
My rationale, nudge the user about its existence - Who knows it might turn out to be useful :)

pages/common/tailscale.md
mfrw: It does but only when one does `tailscale debug --help`. The normal `tailscale --help` does not show it.
IMO, it happens to be pretty useful :)

sbrl: If as you say below the `debug` subcommand is unstable, it does not seem wise to document it here.

mfrw: In my opinion, the reason for adding a `debug` example instead of a page is just to provide the user a hint that it exists :)
It in my opinion is unstable to warrant a separate page but useful enough for debug purposes :)
@sbrl

EmilySeville7cfg (approved):
No unmentioned mandatory fixes found, approved.
Thanks for a contribution! 🚀

pages/common/tailscale.md
Waples: After looking into this, a) yes we can add this but not under the current PR. b) since this is open for close to a year, is it really relevant?

blueskyson (approved):

pages/common/tailscale.md
kbdharun: > After looking into this, a) yes we can add this but not under the current PR. b) since this is open for close to a year, is it really relevant?

Agreed, we can add a dedicated page for this Experimental command if required, but the `debug` command is still unstable. Will close this PR for now, feel free to create a new one for the separate page.

PR #8338, opened 2022-08-10
toolbox*: add Tamil translation

kbdharun:
- [x] The page(s) are in the correct platform directories: `linux`.
- [x] The page(s) have at most 8 examples.
- [x] The page description(s) have links to documentation or a homepage.
- [x] The page(s) follow the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message).
- **Version of the command being documented (if known): `0.0.99.3`**

PR #8347, opened 2022-08-14
aws-cur, aws-ec2, aws-ecr: add French translation

Nico385412:

### DIFF
--- a/pages.fr/common/aws-cur.md
+++ b/pages.fr/common/aws-cur.md
@@ -1,0 +1,16 @@
+# aws cur
+
+> Crée, requête et supprime des rapports de définition d'utilisation AWS.
+> Plus d'informations : <https://awscli.amazonaws.com/v2/documentation/api/latest/reference/cur/index.html>.
+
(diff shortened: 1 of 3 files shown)


kbdharun (approved):
LGTM, Thanks @Nico385412. I think we need to rearrange the commands in `pages.fr/common/aws-ec2.md` and in other `aws-ec2` pages as `help` should come after all subcommands as @EmilySeville7cfg suggested in #8338. What do you think?

EmilySeville7cfg:
> I think we need to rearrange the commands in aws-ec2.md on all pages as help should come after all subcommands as @EmilySeville7cfg suggested in https://github.com/tldr-pages/tldr/pull/8338. What do you think?

I like this idea. 😺

kbdharun:
> > I think we need to rearrange the commands in aws-ec2.md on all pages as help should come after all subcommands as @EmilySeville7cfg suggested in #8338. What do you think?
> 
> I like this idea. smiley_cat But the original pages must be also changed in the same way.

Yeah

CleanMachine1 (approved):

PR #8349, opened 2022-08-14
assoc, attrib: update page

EmilySeville7cfg:
- **Version of the command being documented (if known):** latest

## Changes

- Command examples reordered: from simple to complex. Complexity is measured in arguments count.
- `{{path\to\directory}}` is used instead of `{{path/to/directory}}` to suggest the right path delimiter in Windows.

I think we can apply the same corrections for all other Windows specific pages.

### DIFF
--- a/pages/windows/assoc.md
+++ b/pages/windows/assoc.md
@@ -1,9 +1,9 @@
 # assoc
 
-> Display or change associations of file names.
+> Display or change the default program which handles specific file types 
 > More information: <https://docs.microsoft.com/windows-server/administration/windows-commands/assoc>.
 
-- List all associated file types:
+- List all file types with their default program which uses them:
 
 `assoc`
 
--- a/pages/windows/assoc.md
+++ b/pages/windows/assoc.md
@@ -1,9 +1,9 @@
 # assoc
 
-> Display or change the default program which handles specific file types 
+> Display or change associations between file extensions and file types.
 > More information: <https://docs.microsoft.com/windows-server/administration/windows-commands/assoc>.
 
-- List all file types with their default program which uses them:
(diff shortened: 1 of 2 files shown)


kbdharun (approved):
LGTM
> * `{{path\to\directory}}` is used instead of `{{path/to/directory}}` to suggest the right path delimiter in Windows.

Yeah, I think this would be good. It would prevent people from confusing the path delimiter `\` in Windows with `/`. Let's see what others think.

pages/windows/assoc.md
```diff
@@ -1,16 +1,16 @@
 # assoc
 
-> Display or modify file extension associations.
+> Display or change associations of file names.
```
CleanMachine1, line 3: What is meant by an associations of file names?

EmilySeville7cfg: [How to associate a file with a program in Windows](https://www.computerhope.com/issues/ch000572.htm)

An association is smth that specifies which program should handle which file type.

CleanMachine1: ah ok, so like the default applications each file type

CleanMachine1:
```suggestion
> Display each file type and the associated program which handles them:
```

pages/windows/assoc.md
```diff
@@ -1,16 +1,16 @@
...
 > More information: <https://docs.microsoft.com/windows-server/administration/windows-commands/assoc>.
 
-- Display all associated filetypes:
+- List all associated file types:
```
CleanMachine1, line 6:
```suggestion
- List all file types with their default program which uses them
```

PR #8357, opened 2022-08-15
git-config: add --show-origin example

AriESQ:
git version 2.37.1

`git config --show-origin` is very helpful when troubleshooting git configs that have `include` and/or `includeif` directives. It shows you where a configuration object comes from.

The purpose of this PR was just to introduce the `--show-origin` flag; but I had to find a way to correct the assumptions made by the two previous commands (showing `--local` and `--global`, but obscuring that `--system` is another configuration scope). So, following TLDR tradition, of making successive commands more complex; I added both `--global` and `--show-origin` to my example.

### DIFF
--- a/pages/common/git-config.md
+++ b/pages/common/git-config.md
@@ -12,9 +12,9 @@
 
 `git config --list --global`
 
-- List all configuration entries that have been defined either locally or globally:
+- List only system configuration entries (stored in `/etc/gitconfig`), and show their file location:
 
-`git config --list`
+`git config --list --system --show-origin`
 
 - Get the value of a given configuration entry:
 


pages/common/git-config.md
```diff
@@ -12,9 +12,9 @@
 
 `git config --list --global`
 
-- List all configuration entries that have been defined either locally or globally:
+- List only system configuration entries (stored in `/etc/gitconfig`), and show their file location:
```
CleanMachine1, line 15: I don't completely understand, so its checking /etc/gitconfig 
then checking what settings it has then printing /etc/gitconfig?

AriESQ: I'm trying to keep my example in parallel with the others. Neither of the other two are completely correct; because you can change the location of where git config files are stored. I included the `/etc/gitconfig` example because it is the default, similar to the previous two examples.

See:
https://www.git-scm.com/docs/git-config#Documentation/git-config.txt---global

I acknowledge that `--system --show-origin` is kind of a toy example; `--show-origin` is much more helpful with a broader scope, especially where your git config may have multiple include/includif directives.

AriESQ: It goes a little broader than I understood. `--show-origin` will also show config from other locations, such as arguments passed on a command line. I'm basically of the opinion that for most use cases, you want to use `--show-origin` when you're using `git config --list`

>```Augment the output of all queried config options with the origin type (file, standard input, blob, command line) and the actual origin (config file path, ref, or blob id if applicable).```

https://www.git-scm.com/docs/git-config#Documentation/git-config.txt---show-origin

CleanMachine1 (approved):
Thanks!

kbdharun (approved):
LGTM, Thanks

mfrw (approved):
LGTM :)

PR #8364, opened 2022-08-17
quick-start: add guide

EmilySeville7cfg:
- **Version of the command being documented (if known):** unknown

## Changes

I introduce this quick start for newbies. This quick start should provide a quick introduction for `tldr` pages via examples and not on formal rules.

I am going to standardize (#8252) everything and write formal rules in style guide, that's why we need smth that allow new users quickly edit `tldr` pages.

### DIFF
--- a/contributing-guides/quick-start.md
+++ b/contributing-guides/quick-start.md
@@ -1,0 +1,76 @@
+# Quick start
+
+This page quickly describes `tldr` page syntax with several examples. For strict rules please follow our [style guide](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/style-guide.md).
+
+## Simplest example
+
+Here is an example of page describing `pwd` command:
+
+```md
+# pwd
+
+> Print name of current/working directory.
+> More information: <https://www.gnu.org/software/coreutils/pwd>.
+
+- Print the current directory:
+
+`pwd`
+
+- Print the current directory, and resolve all symlinks (i.e. show the "physical" path):
+
+`pwd --physical`
+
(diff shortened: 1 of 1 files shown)


EmilySeville7cfg:
This PR is replaced with #8366.

sbrl (commented):
This looks cool, but what about the style guide? We really want a single source of truth, so that when we want to update things we don't have to look in multiple places.

contributing-guides/quick-start.md
```diff
@@ -0,0 +1,76 @@
...
+- Print the current directory, and resolve all symlinks (i.e. show the "physical" path):
+
+`pwd --physical`
+
+- Print the current logical directory:
+
+`pwd --logical`
+```
+
+Although `tldr` pages use Markdown syntax we have to note some moments:
```
mfrw, line 28: How about:
```suggestion
Although `tldr` pages use Markdown syntax, we should note some conventions that we follow:
```

EmilySeville7cfg:
> This looks cool, but what about the style guide? We really want a single source of truth, so that when we want to update things we don't have to look in multiple places.

There are already two sources of truth: our CONTRIBUTING.md and the style guide. 😄 Although the first one just summarizes the most important things of the latter one. We need to decide whether to enhance our CONTRIBUTING.md #8366 or whether to leave it unchanged while introducing this quick start.

EmilySeville7cfg:
As #8366 is merged I am not sure whether this quick start is needed. Please reopen this PR is you have another point of view.

PR #8365, opened 2022-08-17
pwd: refresh page

EmilySeville7cfg:
- **Version of the command being documented (if known):** latest

## Changes

- according to [this](https://unixhealthcheck.com/blog?id=157) site `-L` example redundant

I think we can always omit options that are enabled by default unless documentation explicitly tells not to rely on this behavior.

### DIFF
--- a/pages/common/pwd.md
+++ b/pages/common/pwd.md
@@ -10,7 +10,3 @@
 - Print the current directory, and resolve all symlinks (i.e. show the "physical" path):
 
 `pwd -P`
-
-- Print the current logical directory:
-
-`pwd -L`


CleanMachine1:
Mixed opinions on this pr, are we sure there is 0 use-case?

Maybe we could add a note in the command description that it is the default functionality?

EmilySeville7cfg:
> Maybe we could add a note in the command description that it is the default functionality?

There is not point to do that here. Marking smth as a default behavior is useless when there is a shorter command with exactly the same behavior. It's a good idea to rely on defaults when the command author permits it because then our pages become more simple than they could be.

adrienthebo (approved):
This'll be my first review as a collaborator 🥳 , so please bear with me as I get the feel of how tldr-pages reviews pages.

I agree with this change, the `-L` example doesn't provide any value - I can't imagine a case where I'd personally call `pwd -L` over simply `pwd`.

```
❯ pwd
/Users/adrien/.local/sbin
❯ pwd -P
/Users/adrien/.homesick/repos/shell/home/.local/sbin
❯ pwd -L
/Users/adrien/.local/sbin
```

Furthermore, the `-L` flag seems to be ignored when `-P` is passed, which I find surprising as I expected an error. Drawing attention to this somewhat weird flag doesn't provide any value to tldr users that I can see.

```
❯ pwd -P -L
/Users/adrien/.homesick/repos/shell/home/.local/sbin
❯ pwd -L -P
/Users/adrien/.homesick/repos/shell/home/.local/sbin
```

CleanMachine1 (approved):

kbdharun (approved):

PR #8373, opened 2022-08-19
README: add similar project

EmilySeville7cfg:
- **Version of the command being documented (if known):** unknown

## Changes

- Add `learnxinyminutes` to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,11 @@ to see the overall progress of all translations, and which translations are miss
   Rico's cheatsheets are not just focused on the command-line and
   include a plethora of other cheatsheets related to programming.
 
+- [learnxinyminutes](https://learnxinyminutes.com/)
+  Whirlwind tours of (several, hopefully many someday) popular and
+  ought-to-be-more-popular programming languages, presented as valid,
+  commented code and explained as they go.
+
 - [eg](https://github.com/srsudar/eg)
   provides detailed examples with explanations on the command-line.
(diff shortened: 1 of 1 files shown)


CleanMachine1:
This doesn't seem in line with the similar projects IMO.

The other mentioned projects are cheatsheet style repos which are very similar to tldr pages.

The similar projects we have all seem to a common goal of helping with the command-line whether than be remembering or a quick overview of a command. While this project seems to be entire tutorials and guides about learning an entire topic. I think it is a little out of scope.

Let's see what others think before rejecting it though.

 Interesting site regardless!

kbdharun:
> This doesn't seem in line with the similar projects IMO.
> 
> The other mentioned projects are cheatsheet style repos which are very similar to tldr pages.
> 
> The similar projects we have all seem to a common goal of helping with the command-line whether than be remembering or a quick overview of a command. While this project seems to be entire tutorials and guides about learning an entire topic. I think it is a little out of scope.
> 
> Let's see what others think before rejecting it though.
> 
> Interesting site regardless!

Yeah, I agree with @CleanMachine1, even though this is an excellent website, Featuring it readme with other cheetshet resources doesn't fit well. Maybe we can add a section in Wiki like `Additional Resources` or `Recommended Projects` and add it under it.

CleanMachine1:
I personally think after an another review, we should close this PR.

marchersimon:
I too think we should only feature command-line-related projects.

PR #8381, opened 2022-08-20
gcc: mention -c compiles into object file

jxu:


pages/common/gcc.md
```diff
@@ -19,6 +19,6 @@
 
 `gcc -S {{path/to/source.c}}`
 
-- Compile source code without linking:
+- Compile source code without linking, into object file:
```
navarroaxel, line 22: Perhaps this better
```suggestion
- Compile source code into an object file without linking:
```

jxu: I considered that wording as well. As long as it's clear an object file by definition is without linking (I believe)

CleanMachine1: Feel free to commit the change @jxu if it keeps the same meaning

CleanMachine1 (approved):
lgtm after the navarroaxel's comment is resolved.

pages/common/gcc.md
sbrl: @navarroaxel's wording reads better for me

jxu:
I don't know if standard practice is to squash commits together, but since the first commit already got comments I'll leave it

pages/common/gcc.md
```diff
@@ -19,6 +19,6 @@
 
 `gcc -S {{path/to/source.c}}`
 
-- Compile source code without linking:
+- Compile source code into an object file without linking:
```
EmilySeville7cfg, line 22:
```suggestion
- Compile the source code into an object file without linking:
```

Maybe it's better. Anyway everything looks nice. 👍🏻

EmilySeville7cfg (approved):

pages/common/gcc.md
jxu: I don't think it makes a difference. If using "a specific" it would have to be with "source code file".

EmilySeville7cfg (approved):

EmilySeville7cfg:
Merged, thanks! ❤️

PR #8387, opened 2022-08-21
gcc: use short -o option for output 

### LINKED ISSUE #8379: gcc: Use more common output flag -o instead of --output
I know the guidelines say we should prefer full names, but -o is so widespread and standardized that I think it's more appropriate going for -o. Curiously I can't even find --output in the man page. Worth mentioning the g++ tldr and the other options are all abbreviated.

jxu:
Resolves #8379

### DIFF
--- a/pages/common/gcc.md
+++ b/pages/common/gcc.md
@@ -5,15 +5,15 @@
 
 - Compile multiple source files into executable:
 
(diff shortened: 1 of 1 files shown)


mfrw (approved):
LGTM as the discussion on the issue :)

navarroaxel (approved):

PR #8402, opened 2022-08-26
objdump: add intel syntax command

### LINKED ISSUE #8389: objdump: intel syntax
Use -M intel flag

jxu:
Also remove hyphen from disassemble to use more common form.

Maybe make a mention intel syntax applies to x86 asm but that should be obvious.

Resolves #8389

### DIFF
--- a/pages/common/objdump.md
+++ b/pages/common/objdump.md
@@ -7,9 +7,13 @@
 
 `objdump -f {{binary}}`
 
-- Display the dis-assembled output of executable sections:
+- Display the disassembled output of executable sections:
 
 `objdump -d {{binary}}`
+
+- Disaply the disassembled executable sections in intel syntax:
+
+`objdump -M intel -d {{binary}}`
 
 - Display a complete binary hex dump of all sections:
 
--- a/pages/common/objdump.md
+++ b/pages/common/objdump.md
@@ -11,7 +11,7 @@
 
 `objdump -d {{binary}}`
 
-- Disaply the disassembled executable sections in intel syntax:
+- Display the disassembled executable sections in intel syntax:
 
 `objdump -M intel -d {{binary}}`
 


pages/common/objdump.md
```diff
@@ -7,10 +7,14 @@
 
 `objdump -f {{binary}}`
 
-- Display the dis-assembled output of executable sections:
+- Display the disassembled output of executable sections:
 
 `objdump -d {{binary}}`
 
+- Disaply the disassembled executable sections in intel syntax:
```
kbdharun, line 14:
```suggestion
- Display the disassembled executable sections in intel syntax:
```

jxu: Oops! that's embarrassing.

kbdharun (approved):
LGTM, Thanks @jxu . Let's see what others think.

mfrw (approved):
Although, we prefer long options, but given this page uses short options; LGTM :)
Thank you for your contrinution @jxu

pages/common/objdump.md
```diff
@@ -7,10 +7,14 @@
...
 `objdump -f {{binary}}`
 
-- Display the dis-assembled output of executable sections:
+- Display the disassembled output of executable sections:
 
 `objdump -d {{binary}}`
 
+- Display the disassembled executable sections in intel syntax:
+
+`objdump -M intel -d {{binary}}`
```
CleanMachine1, line 16:
```suggestion
`objdump -M {{intel}} -d {{binary}}`
```

Is there multiple options that can go with -M such as amd

jxu: For x86 there is just Intel or AT&T (default for this and GNU tools). AMD processors are also x86.

CleanMachine1: We should the {{ }} around it since it takes arm I assume.

jxu: ARM doesn't have two versions of syntax like x86. In my original post I made a note that maybe this should be mentioned this only applies for x86 binaries

CleanMachine1:
Thanks for your contribution.

PR #8413, opened 2022-08-28
chafa: add page

CairnThePerson:
- **Version of the command being documented (if known):** 1.12.3

### DIFF
--- a/pages/common/chafa.md
+++ b/pages/common/chafa.md
@@ -1,7 +1,7 @@
 # chafa
 
 > Image printing in the terminal.
-> See also: `catimg` and `pixterm`.
+> See also: `catimg`, `pixterm`.
 > More information: <https://hpjansson.org/chafa>.
 
 - Print an image or animated GIF to the terminal:
--- a/pages/common/chafa.md
+++ b/pages/common/chafa.md
@@ -6,20 +6,20 @@
 
 - Print an image or animated GIF to the terminal:
 
-`chafa {{path/to/input}}`
+`chafa {{path/to/file}}`
 
 - Use 24-bit [c]olor:
 
-`chafa -c full {{path/to/input}}`
+`chafa -c full {{path/to/file}}`
 
 - Use dithering to improve output with small color palettes:
 
-`chafa -c 16 --dither ordered {{path/to/input}}`
+`chafa -c 16 --dither ordered {{path/to/file}}`
 
 - Generate output like a low-resolution raster image:
 
-`chafa --symbols vhalf {{path/to/input}}`
+`chafa --symbols vhalf {{path/to/file}}`
 
 - Generate monochrome output with only Braille characters:
 
-`chafa -c none --symbols braille {{path/to/input}}`
+`chafa -c none --symbols braille {{path/to/file}}`
--- a/pages/common/chafa.md
+++ b/pages/common/chafa.md
@@ -16,7 +16,7 @@
 
 `chafa -c 16 --dither ordered {{path/to/file}}`
 
-- Render an image in the terminal to appear pixelated:
+- Render an image, making it appear pixelated:
 
 `chafa --symbols vhalf {{path/to/file}}`
 
--- a/pages/common/chafa.md
+++ b/pages/common/chafa.md
@@ -20,6 +20,6 @@
 
 `chafa --symbols vhalf {{path/to/file}}`
 
-- Render a monochrome image with only Braille characters:
+- Render a monochrome image with only braille characters:
 
 `chafa -c none --symbols braille {{path/to/file}}`
--- a/pages/common/chafa.md
+++ b/pages/common/chafa.md
@@ -1,0 +1,25 @@
+# chafa
+
+> Image printing in the terminal.
+> See also: `catimg` and `pixterm`.
+> More information: <https://hpjansson.org/chafa>.
+
+- Print an image or animated GIF to the terminal:
+
+`chafa {{path/to/input}}`
+
+- Use 24-bit [c]olor:
+
+`chafa -c full {{path/to/input}}`
+
+- Use dithering to improve output with small color palettes:
+
+`chafa -c 16 --dither ordered {{path/to/input}}`
+
+- Generate output like a low-resolution raster image:
+
+`chafa --symbols vhalf {{path/to/input}}`
+
+- Generate monochrome output with only Braille characters:
+
+`chafa -c none --symbols braille {{path/to/input}}`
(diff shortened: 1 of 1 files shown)


pages/common/chafa.md
```diff
@@ -0,0 +1,25 @@
+# chafa
+
+> Image printing in the terminal.
+> See also: `catimg` and `pixterm`.
+> More information: <https://hpjansson.org/chafa>.
+
+- Print an image or animated GIF to the terminal:
```
navarroaxel, line 7: All the tldr examples are for the terminal, should the user use ctrl + C or q to exit?
```suggestion
- Print an image or animated GIF:
```

pages/common/chafa.md
```diff
@@ -0,0 +1,25 @@
...
+
+`chafa {{path/to/input}}`
```
navarroaxel, line 9: We prefer `path/to/file`
```suggestion
`chafa {{path/to/file}}`
```

pages/common/chafa.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Use 24-bit [c]olor:
```
navarroaxel, line 11:
```suggestion
- Print an image using 24-bit [c]olor:
```

pages/common/chafa.md
```diff
@@ -0,0 +1,25 @@
...
+
+`chafa -c full {{path/to/input}}`
+
+- Use dithering to improve output with small color palettes:
```
navarroaxel, line 15:
```suggestion
- Print an image using dithering to improve output with small color palettes:
```

pages/common/chafa.md
```diff
@@ -0,0 +1,25 @@
...
+
+`chafa -c 16 --dither ordered {{path/to/input}}`
+
+- Generate output like a low-resolution raster image:
```
navarroaxel, line 19: Why `generate` and not `print` here?

pages/common/chafa.md
CairnThePerson: It's creating an image out of text characters inside the terminal. And I generally think of "print to the terminal" as a sort of phrase, since just using "print" might make you think of a GUI application or an actual printer.

I already pretty much said this in the description though, so I'd understand if it seems redundant.

pages/common/chafa.md
CairnThePerson: Got it. I'll update that here and in my `pixterm` and `catimg` PRs.

pages/common/chafa.md
CairnThePerson: I think it gets a little dry to say "print" over and over. I already specified that it prints to the terminal in the first example. And since its primary function is to generate images and print them to the terminal, I think "use" should be good enough.

navarroaxel: The example descriptions shouldn't be described using the context. Make them independent

pages/common/chafa.md
CairnThePerson: Printing to terminal is just "spitting out" what it's generated. In the last two examples, I think it's a little more specific to say something like "generate" instead of the verb "use" I'd used before that. So basically, nothing about the printing process is being changed, whereas the process of generation *is*. And it's not simply using a one feature; the whole process is *generating* something similar to a raster image. I hope my reasoning makes sense.

pages/common/chafa.md
CairnThePerson: Hm. I was under the impression that the format is useful for both showing independent examples and giving a brief tutorial as a whole. This is even somewhat supported by the guides, where it says to start with simple examples before going to more complex ones. Maybe this particular point isn't a good example of that, but is it really better to use the same verbage over and over, when the primary use of the tool has already been demonstrated?

pages/common/chafa.md
CairnThePerson: Updated.

CairnThePerson:
@navarroaxel I reworked the descriptions using a common verb ("render") throughout, and I hope I resolved the issues you had with it. Let me know what you think! :slightly_smiling_face:

pages/common/chafa.md
```diff
@@ -0,0 +1,25 @@
+# chafa
+
+> Image printing in the terminal.
+> See also: `catimg`, `pixterm`.
+> More information: <https://hpjansson.org/chafa>.
+
+- Render an image or animated GIF directly in the terminal:
```
CleanMachine1, line 7:
```suggestion
- Render an image or a GIF directly in the terminal:
```

pages/common/chafa.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Render an image with 24-bit [c]olor:
+
+`chafa -c full {{path/to/file}}`
+
+- Improve image rendering with small color palettes using dithering:
+
+`chafa -c 16 --dither ordered {{path/to/file}}`
+
+- Render an image in the terminal to appear pixelated:
```
CleanMachine1, line 19:
```suggestion
- Render an image, making it appear pixelated:
```

pages/common/chafa.md
```diff
@@ -0,0 +1,25 @@
...
+
+`chafa --symbols vhalf {{path/to/file}}`
+
+- Render a monochrome image with only Braille characters:
```
CleanMachine1, line 23:
```suggestion
- Render a monochrome image with only braille characters:
```

Not sure whether this should be capitalized, from a little research, I think it shouldn't be.

pages/common/chafa.md
CairnThePerson: I was thinking it might be an important distinction. `pixterm`, for example, can only render static GIFs. But GIF is synonymous with animation for most people, so maybe that isn't so useful. I might just omit the GIF mention entirely.

```suggestion
- Render an image directly in the terminal:
```

pages/common/chafa.md
CairnThePerson: Nice. I hadn't realized.

pages/common/chafa.md
CairnThePerson: Yeah, I'm just removing it.

CleanMachine1: to me, a gif is a collection of pictures put together (just described a video :smile: ) so in my definition, it is an animation.

CleanMachine1 (approved):
lgtm now, thanks!

pages/common/chafa.md
CairnThePerson: Maybe I'm just being too particular then.

kbdharun (approved):

PR #8425, opened 2022-08-30
ctags: update convenient commands to show function with line number in json format

ramwin:

### DIFF
--- a/pages/common/ctags.md
+++ b/pages/common/ctags.md
@@ -14,3 +14,8 @@
 - Generate tags for all files in the current directory and all subdirectories:
 
 `ctags --recurse`
+
+- Generate tags for a single file, and output them with start line number and end line number in JSON format
+
+`ctags --fields=+ne --output-format=json {{path/to/file}}`
+


pages/common/ctags.md
```diff
@@ -14,3 +14,7 @@
 - Generate tags for all files in the current directory and all subdirectories:
 
 `ctags --recurse`
+
+- Generate tags for a single file, and output them with start line number and end line number in JSON format:
```
CleanMachine1, line 18: By `start line line`, does that mean it will always output 1?

ramwin: output them with start line number and end line number. here them means tags. So the start line number is for each tag. 
For example: the `test.c` looks like them.It contains tag main and fun
```
[#12#wangx@windows:C_lesson] $ cat test.c
#include <stdio.h>
#include <limits.h>

int main() {
    float a = fun();
    printf("a = %f\n", a);
};

char fun() {
    return 'a';
};
```

the ctags will return:
```
[#13#wangx@windows:C_lesson] $ ctags --fields=+ne --output-format=json test.c
{"_type": "tag", "name": "fun", "path": "test.c", "pattern": "/^char fun() {$/", "line": 9, "typeref": "typename:char", "kind": "function", "end": 11}
{"_type": "tag", "name": "main", "path": "test.c", "pattern": "/^int main() {$/", "line": 4, "typeref": "typename:int", "kind": "function", "end": 7}
```

here the line and end line is the start and end line for each function.

navarroaxel: @CleanMachine1 are you ok with this?

CleanMachine1 (approved):

navarroaxel (approved):

PR #8453, opened 2022-09-03
wipefs: add --all example with glob pattern

JPvRiel:
It may often be misconceived that `wipefs --all /dev/sdX` will recurse and wipe sub-level block signatures, i.e. those for filesystems, etc. However `--all` does not recurse the layers. In order to more fully wipe at all layers, the glob pattern can help make the wiping more thorough.

### DIFF
--- a/pages/linux/wipefs.md
+++ b/pages/linux/wipefs.md
@@ -13,7 +13,7 @@
 
 - Wipe all available signature types for the device and partitions using a glob pattern:
 
-`sudo wipefs --all {{/dev/sdX*}}`
+`sudo wipefs --all {{/dev/sdX}}*`
 
 - Perform dry run:
 
--- a/pages/linux/wipefs.md
+++ b/pages/linux/wipefs.md
@@ -7,7 +7,7 @@
 
 `sudo wipefs {{/dev/sdX}}`
 
-- Wipe all available signature types for a specific device (no recursion into partitions):
+- Wipe all available signature types for a specific device with no recursion into partitions:
 
 `sudo wipefs --all {{/dev/sdX}}`
 
--- a/pages/linux/wipefs.md
+++ b/pages/linux/wipefs.md
@@ -7,11 +7,11 @@
 
 `sudo wipefs {{/dev/sdX}}`
 
-- Wipe all available signatures for specified device level (no recursion to sub-level block layers):
+- Wipe all available signature types for the device (no recursion into partitions):
 
 `sudo wipefs --all {{/dev/sdX}}`
 
-- Wipd all available signatures for the specified device at all block levels using a glob pattern:
+- Wipe all available signature types for the device and partitions using a glob pattern:
 
 `sudo wipefs --all {{/dev/sdX*}}`
 
--- a/pages/linux/wipefs.md
+++ b/pages/linux/wipefs.md
@@ -11,7 +11,7 @@
 
 `sudo wipefs --all {{/dev/sdX}}`
 
-- Wipd all available signatures for specified device at all block levels using a glob pattern:
+- Wipd all available signatures for the specified device at all block levels using a glob pattern:
 
 `sudo wipefs --all {{/dev/sdX*}}`
 
--- a/pages/linux/wipefs.md
+++ b/pages/linux/wipefs.md
@@ -7,9 +7,13 @@
 
 `sudo wipefs {{/dev/sdX}}`
 
-- Wipe all available signatures for specified device:
+- Wipe all available signatures for specified device level (no recursion to sub-level block layers):
 
 `sudo wipefs --all {{/dev/sdX}}`
+
+- Wipd all available signatures for specified device at all block levels using a glob pattern:
+
+`sudo wipefs --all {{/dev/sdX*}}
 
 - Perform dry run:
 
--- a/pages/linux/wipefs.md
+++ b/pages/linux/wipefs.md
@@ -7,7 +7,7 @@
 
 `sudo wipefs {{/dev/sdX}}`
 
-- Wipe all available signature types for the device (no recursion into partitions):
+- Wipe all available signature types for a specific device (no recursion into partitions):
 
 `sudo wipefs --all {{/dev/sdX}}`
 
--- a/pages/linux/wipefs.md
+++ b/pages/linux/wipefs.md
@@ -13,7 +13,7 @@
 
 - Wipd all available signatures for specified device at all block levels using a glob pattern:
 
-`sudo wipefs --all {{/dev/sdX*}}
+`sudo wipefs --all {{/dev/sdX*}}`
 
 - Perform dry run:
 


pages/linux/wipefs.md
```diff
@@ -7,10 +7,14 @@
 
 `sudo wipefs {{/dev/sdX}}`
 
-- Wipe all available signatures for specified device:
+- Wipe all available signatures for specified device level (no recursion to sub-level block layers):
 
 `sudo wipefs --all {{/dev/sdX}}`
 
+- Wipd all available signatures for specified device at all block levels using a glob pattern:
```
navarroaxel, line 14:
```suggestion
- Wipd all available signatures for the specified device at all block levels using a glob pattern:
```

pages/linux/wipefs.md
```diff
@@ -7,10 +7,14 @@
 
 `sudo wipefs {{/dev/sdX}}`
 
-- Wipe all available signatures for specified device:
+- Wipe all available signatures for specified device level (no recursion to sub-level block layers):
```
navarroaxel, line 10: Which level? block level?

navarroaxel (commented):
Hi @JPvRiel, welcome and thanks for your submission. I left some comments.

pages/linux/wipefs.md
JPvRiel: "block level" is a bit fuzzy now that I think on it again. My understanding is Linux block layers stack up, such that `/dev/sda` would be the full device, then `/dev/sda1` would be the first partition, etc. But technically, perhaps, we should rather say "device and all partitions", given there's other block stuff, like LVM, which doesn't behave this way.

Also, if one does `wipefs /dev/sda*` and there's a partition with an LVM in the partition, I'm guessing wipefs will only clear up the LVM metadata, but not the filesystem metadata with the logical volumes, etc. So there's only so much wipefs does. To truly recurse to all block layers would require a script.

I'll submit another commit to the pull to word this better. My "block layer" wording is too generic.

(Apologies for a tardy reply.)

pages/linux/wipefs.md
```diff
@@ -7,10 +7,14 @@
 
 `sudo wipefs {{/dev/sdX}}`
 
-- Wipe all available signatures for specified device:
+- Wipe all available signature types for the device (no recursion into partitions):
```
adrienthebo, line 10: **fixup:** When referring to a single entity such as a file or a device, tldr pages use "a specific {{x}}" by convention:

```suggestion
- Wipe all available signature types for a specific device (no recursion into partitions):
```

**Question:** Regarding the change from 'signatures' to 'signature types', I don't understand this change; I would expect that wiping a signature type indicates that we're somehow erasing the concept of a given signature (gpt, mbr etc), and wiping signatures would actually remove the signatures on-disk. Looking at the [wipefs 2.37.85-637cc man page](https://www.man7.org/linux/man-pages/man8/wipefs.8.html) shows that the term "signature type" isn't really used.

Is there something that I'm understanding that makes "signature type" more correct in this context?

adrienthebo (commented):
One small suggestion and one question, looks good otherwise. Thanks for submitting this!

pages/linux/wipefs.md
JPvRiel: > Is there something that I'm understanding that makes "signature type" more correct in this context?

Hi @adrienthebo,  it's a fair question. The reason I made it more verbose from just "all available signatures" to all "available signature types" was that the `--all` means erasing all signatures of any type at the specified block level, but not all, as in recusing into all/deeper block levels and it does not erase "sub-level" block device signatures. I can see how this is unclear. Perhaps it should be worded:

"Wipe all signatures of any type at a specified device block layer (no recursion into partitions)". Does this sound better? Or would you prefer sticking to the original wording given "(no recursion into partitions)" does make it clear enough?

Maybe the man page helps:

> -a, --all
> Erase all available signatures. The set of erased signatures can be restricted with the -t option.

> -t, --types list
> Limit  the  set of printed or erased signatures.  More than one type may be specified in a comma-separated list. The list or individual types can be prefixed with 'no' to specify the types on which no action should be taken.  For more details see mount(8).

My main issue was that I confused all to imply nested/recursive cleanup, but it doesn't. I noticed other people in various forums shared my misconception.

pages/linux/wipefs.md
```diff
@@ -7,10 +7,14 @@
 
 `sudo wipefs {{/dev/sdX}}`
 
-- Wipe all available signatures for specified device:
+- Wipe all available signature types for a specific device (no recursion into partitions):
```
CleanMachine1, line 10:
```suggestion
- Wipe all available signature types for a specific device with no recursion into partitions:
```

pages/linux/wipefs.md
```diff
@@ -7,10 +7,14 @@
...
 
 `sudo wipefs --all {{/dev/sdX}}`
 
+- Wipe all available signature types for the device and partitions using a glob pattern:
+
+`sudo wipefs --all {{/dev/sdX*}}`
```
CleanMachine1, line 16: Why does this need an asterisk, surely just giving the disk is enough if we are wiping all.

Feel free to ignore this, if it is not an error.

navarroaxel: The `*` is required, perhaps this better
```suggestion
`sudo wipefs --all {{/dev/sdX}}*`
```

pages/linux/wipefs.md
JPvRiel: wipefs: cleaner wording about no recursion

pages/linux/wipefs.md
JPvRiel: Seeing no further discussion about why or if the signature type matters, I'll leave this as be and resolve the conversation. It it still seems disagreeable, please open a fresh comment and I'll happily take any suggestions on cleaner wording.

pages/linux/wipefs.md
JPvRiel: @CleanMachine1 

> surely just giving the disk is enough if we are wiping all

Fair question. It depends on intention.

- If you simply want to cleanup the device base signatures, yes, but all means "all signatures or signature types" at the block device level, and does not account for any signatures in other layers.
- If however, by "all" you want every signature across multiple layers, the glob is a partial hack to do that.
- E.g. /dev/sda may have two partions /dev/sda1 and /dev/sda2. If you wanted to wipe not just the partition table / label / signature, but also, filesystem signatures (or LVM, whatever is within the partitions), then the glob helps do that.
- Why? Well, besides a long zero erase, later on, if you use any data loss recovery tools that do a full block scan to find signatures, you ideally don't want any stale not-fully-erased signatures that could float in unallocated blocks read by the recovery program. It would just add to the confusion. E.g. imagine a bad program or something corrupts the partition table and it's backup, then the recovery program might scan the entire drive looking for filesystem signatures.

navarroaxel (approved):

kbdharun (approved):

PR #8466, opened 2022-09-06
git-stash: add -k example and stash specific files example

HUGHNew:

### DIFF
--- a/pages/common/git-stash.md
+++ b/pages/common/git-stash.md
@@ -7,17 +7,17 @@
 
 `git stash [push -m {{optional_stash_message}}]`
 
-- Stash specific file(s), except new (untracked) files:
+- Stash current changes, including new (untracked) files:
 
-`git stash push [-m {{optional_stash_message}}] {{file1}} {{file2}}`
+`git stash -u`
 
 - Stash current changes, [k]eeping index (`push` only)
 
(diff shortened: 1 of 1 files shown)


adrienthebo:
Hi @HUGHNew, thanks for this submission!

As the CI failures indicate, there's a maximum of 8 examples per page to keep each tldr page easy to digest. This change pushes the example count up to 10 so we'll have to pick the 8 most relevant examples to keep. Are these new examples part of the first eight commands that a newcomer might use? Can we combine this into other examples that conveys the same information while not being too information dense?

HUGHNew:
> Hi @HUGHNew, thanks for this submission!
> 
> As the CI failures indicate, there's a maximum of 8 examples per page to keep each tldr page easy to digest. This change pushes the example count up to 10 so we'll have to pick the 8 most relevant examples to keep. Are these new examples part of the first eight commands that a newcomer might use? Can we combine this into other examples that conveys the same information while not being too information dense?

Thanks for your comment!

It seems that we can combine the `push` examples(1st and 4nd) but the combined command may be hard to understand.
On the other hand, I don't know that `-u` or `-k` is more helpful.

So, it's better to leave the PR alone until someone has ideas about it.

Waples:
I think the `-k` example is more useful.

PR #8468, opened 2022-09-06
g++: add multiple file example

Keshavshyper6325:
Added the version command as well as multiple file compile and link commands for g++.

### DIFF
--- a/pages/common/g++.md
+++ b/pages/common/g++.md
@@ -4,10 +4,6 @@
 > Part of GCC (GNU Compiler Collection).
 > More information: <https://gcc.gnu.org>.
 
-- Display version:
-
-`g++ --version`
-
 - Compile a source code file into an executable binary:
 
 `g++ {{path/to/source.cpp}} -o {{path/to/output_executable}}`
@@ -27,3 +23,7 @@
 - Compile and link multiple source code files into executable binary:
 
 `g++ -c {{path/to/source_1.cpp}} {{path/to/source_2.cpp}} && g++ -o {{path/to/output_executable}} {{path/to/source_1.o}} {{path/to/source_2.o}}`
+
+- Display version:
+
+`g++ --version`
--- a/pages/common/g++.md
+++ b/pages/common/g++.md
@@ -22,7 +22,7 @@
 
 - Compile and link multiple source code files into executable binary:
 
-`g++ -c {{path/to/source_1.cpp}} {{path/to/source_2.cpp}} && g++ -o {{path/to/output_executable}} {{path/to/source_1.o}} {{path/to/source_2.o}}`
+`g++ -c {{path/to/source_1.cpp path/to/source_2.cpp ...}} && g++ -o {{path/to/output_executable}} {{path/to/source_1.o path/to/source_2.o ...}}`
 
 - Display version:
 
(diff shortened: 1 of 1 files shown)


kbdharun (approved):
Welcome to tldr @Keshavshyper6325. The changes LGTM, Thanks for your contribution.

pages/common/g++.md
```diff
@@ -4,6 +4,10 @@
 > Part of GCC (GNU Compiler Collection).
 > More information: <https://gcc.gnu.org>.
 
+- Check compiler version information:
```
navarroaxel, line 7: This is not a `check` operation, just displays the version
```suggestion
- Display version:
```

pages/common/g++.md
```diff
@@ -4,6 +4,10 @@
...
+
+`g++ --version`
```
navarroaxel, line 9: This should be the last example

pages/common/g++.md
```diff
@@ -19,3 +19,11 @@
 - Include libraries located at a different path than the source file:
 
 `g++ {{path/to/source.cpp}} -o {{path/to/output_executable}} -I{{path/to/header}} -L{{path/to/library}} -l{{library_name}}`
+
+- Compile and link multiple source code files into executable binary:
+
+`g++ -c {{path/to/source_1.cpp}} {{path/to/source_2.cpp}} && g++ -o {{path/to/output_executable}} {{path/to/source_1.o}} {{path/to/source_2.o}}`
```
marchersimon, line 25: Why would you compile and link the files separately? Wouldn't `g++ {{path/to/source1.cpp}} {{path/to/source2.cpp}} -o {{path/to/executable}}` also work?

EmilySeville7cfg (changes requested):

pages/common/g++.md
Keshavshyper6325: I wrote them seperately just in case someone needs the raw compiled file without linking.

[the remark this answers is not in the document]
Keshavshyper6325: That was bad on my part. Committed!

EmilySeville7cfg (approved):

pages/common/g++.md
mfrw: Honestly I am not opposed to this example as it hints the user about distinct compile/link stages :)

mfrw (approved):

pages/common/g++.md
```diff
@@ -19,3 +19,11 @@
 - Include libraries located at a different path than the source file:
 
 `g++ {{path/to/source.cpp}} -o {{path/to/output_executable}} -I{{path/to/header}} -L{{path/to/library}} -l{{library_name}}`
+
+- Compile and link multiple source code files into executable binary:
```
navarroaxel, line 23: Based on the first example, a connector is missing here
```suggestion
- Compile and link multiple source code files into an executable binary:
```

navarroaxel (approved):

jxu:
This is late but what is the consensus on putting multiple commands into one tldr command? Seems to me this is doing two separate things: 1. compile to object files 2. link object files

mfrw:
> This is late but what is the consensus on putting multiple commands into one tldr command? Seems to me this is doing two separate things: 1. compile to object files 2. link object files

I think we generally refrain from it in order to avoid complicated examples.

There are one-off cases where I reall having used multiple commands like in case of piping input from a file to a filter.

In my opinion, if multiple commands convey the intent clearly; we could maybe use it.

These are my opinions though :)

jxu:
This should be either condensed into one command (don't produce object files) or split into two commands demonstrating compiling to object files and linking.

PR #8474, opened 2022-09-07
pacdiff: add page

isaec:
pacdiff 1.7.0

### DIFF
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -1,6 +1,6 @@
 # pacdiff
 
-> Pacorig, pacnew and pacsave maintenance utility for pacman.
+> Maintenance utility for `.pacorig`, `.pacnew` and `.pacsave` files created by `pacman`.
 > More information: <https://man.archlinux.org/man/pacdiff>.
 
 - Review files that need maintenance in interactive mode:
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -19,7 +19,7 @@
 
 `DIFFPROG={{editor}} pacdiff`
 
-- Scan for files with `locate` instead of searching pacmandb to find untracked files:
+- Scan for configuration files with `locate` instead of using pacman database:
 
 `pacdiff --locate`
 
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -15,7 +15,7 @@
 
 `pacdiff --sudo --backup`
 
-- Override the program for `(M)erge`ing changes (`vim -d` is default):
+- Use a specific editor to view and merge configuration files (default is `vim -d`):
 
 `DIFFPROG={{editor}} pacdiff`
 
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -13,15 +13,11 @@
 
 - Store the original version of any file you `(O)verwrite with` as a `.bak`:
 
-`pacdiff --backup`
+`pacdiff -s --backup`
 
 - Override the program for `(M)erge`ing changes (`vim -d` is default):
 
 `DIFFPROG={{meld}} pacdiff`
-
-- Find files that need maintenance, apply changes with sudo/sudoedit, and keep backups:
-
-`pacdiff --backup --sudo`
 
 - Scan for files with `locate` instead of searching pacmandb to find untracked files:
 
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -11,7 +11,7 @@
 
 `pacdiff --sudo`
 
-- Store the original version of any file you `(O)verwrite with` as a `.bak`
+- Store the original version of any file you `(O)verwrite with` as a `.bak`:
 
 `pacdiff --backup`
 
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -13,7 +13,7 @@
 
 - Store the original version of any file you `(O)verwrite with` as a `.bak`:
 
-`pacdiff -s --backup`
+`pacdiff --sudo --backup`
 
 - Override the program for `(M)erge`ing changes (`vim -d` is default):
 
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -17,7 +17,7 @@
 
 - Override the program for `(M)erge`ing changes (`vim -d` is default):
 
-`DIFFPROG=meld pacdiff`
+`DIFFPROG={{meld}} pacdiff`
 
 - Find files that need maintenance, apply changes with sudo/sudoedit, and keep backups:
 
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -7,7 +7,7 @@
 
 `pacdiff`
 
-- Use sudo and sudoedit to remove and merge files.
+- Use sudo and sudoedit to remove and merge files:
 
 `pacdiff --sudo`
 
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -7,7 +7,7 @@
 
 `pacdiff`
 
-- Use sudoedit to merge or remove files:
+- Use sudo and sudoedit to remove and merge files.
 
 `pacdiff --sudo`
 
@@ -19,7 +19,7 @@
 
 `DIFFPROG=meld pacdiff`
 
-- Find files that need maintenance, apply changes with sudoedit, and keep backups:
+- Find files that need maintenance, apply changes with sudo/sudoedit, and keep backups:
 
 `pacdiff -b -s`
 
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -21,7 +21,7 @@
 
 - Find files that need maintenance, apply changes with sudo/sudoedit, and keep backups:
 
-`pacdiff -b -s`
+`pacdiff --backup --sudo`
 
 - Scan for files with `locate` instead of searching pacmandb to find untracked files:
 
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -17,7 +17,7 @@
 
 - Override the program for `(M)erge`ing changes (`vim -d` is default):
 
-`DIFFPROG={{meld}} pacdiff`
+`DIFFPROG={{editor}} pacdiff`
 
 - Scan for files with `locate` instead of searching pacmandb to find untracked files:
 
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -1,7 +1,7 @@
 # pacdiff
 
 > Pacorig, pacnew and pacsave maintenance utility for pacman.
-> More information: <https://manned.org/pacdiff>.
+> More information: <https://man.archlinux.org/man/pacdiff>.
 
 - Look for files that need maintenance:
 
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -1,0 +1,29 @@
+# pacdiff
+
+> Pacorig, pacnew and pacsave maintenance utility for pacman.
+> If you want to use a different program for diffs, set the `DIFFPROG` enviromental variable to point to your desired binary, with any needed flags. `vim -d` is the default.
+> More information: <https://man.archlinux.org/man/pacdiff.8>.
+
+- Look for files that need maintenance:
+
+`pacdiff`
+
+- Use sudoedit to merge or remove files:
+
+`pacdiff -s`
+
+- Save old versions of modified files as `.bak` files:
+
+`pacdiff -b`
+
+- Find files that need maintenance, apply changes with sudoedit, and keep backups:
+
+`pacdiff -b -s`
+
+- Scan for files with `locate` instead of using pacmandb (much slower, but finds files missing from db):
+
+`pacdiff -l`
+
+- Display help:
+
+`pacdiff -h`
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -3,7 +3,7 @@
 > Pacorig, pacnew and pacsave maintenance utility for pacman.
 > More information: <https://man.archlinux.org/man/pacdiff>.
 
-- Look for files that need maintenance:
+- Review files that need maintenance in interactive mode:
 
 `pacdiff`
 
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -11,7 +11,7 @@
 
 `pacdiff --sudo`
 
-- Review pacfiles, creating `.bak`ups of the original when you `(O)verwrite with`:
+- Review files needing maintenance, creating `.bak`ups of the original if you `(O)verwrite`:
 
 `pacdiff --sudo --backup`
 
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -11,7 +11,7 @@
 
 `pacdiff --sudo`
 
-- Store the original version of any file you `(O)verwrite with` as a `.bak`:
+- Review pacfiles, creating `.bak`ups of the original when you `(O)verwrite with`:
 
 `pacdiff --sudo --backup`
 
--- a/pages/linux/pacdiff.md
+++ b/pages/linux/pacdiff.md
@@ -1,8 +1,7 @@
 # pacdiff
 
 > Pacorig, pacnew and pacsave maintenance utility for pacman.
-> If you want to use a different program for diffs, set the `DIFFPROG` enviromental variable to point to your desired binary, with any needed flags. `vim -d` is the default.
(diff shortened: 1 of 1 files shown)


pages/linux/pacdiff.md
```diff
@@ -0,0 +1,29 @@
+# pacdiff
+
+> Pacorig, pacnew and pacsave maintenance utility for pacman.
+> If you want to use a different program for diffs, set the `DIFFPROG` enviromental variable to point to your desired binary, with any needed flags. `vim -d` is the default.
+> More information: <https://man.archlinux.org/man/pacdiff.8>.
```
CleanMachine1, line 5:
```suggestion
> More information: <https://manned.org/pacdiff>.
```

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,29 @@
+# pacdiff
+
+> Pacorig, pacnew and pacsave maintenance utility for pacman.
+> If you want to use a different program for diffs, set the `DIFFPROG` enviromental variable to point to your desired binary, with any needed flags. `vim -d` is the default.
```
adrienthebo, line 4: The description section should be ["short and snappy"](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/style-guide.md#layout) per the style guide - the intent is to convey what a tool does any why someone might want to use it. Controlling the diff program is technical information; if we want to include that at all then it should probably be an example, not part of the description.

Is it very common to change the DIFFPROG command, or is there a reason to highlight this behavior early on?

adrienthebo, line 4: My understanding is that `pacdiff` manages file backups that pacman creates when potentially overwriting/deleting configuration files. Without searching around I wouldn't be able to determine _why_ I might use this. Can we change this description to indicate that `pacdiff` merges and cleans up these backup files?

pages/linux/pacdiff.md
adrienthebo: I'm okay with using the Arch Linux man pages for this one since this is a pretty Arch-centric command. Would be fine with using `manned.org` as well. 😄

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,29 @@
...
+> More information: <https://man.archlinux.org/man/pacdiff.8>.
+
+- Look for files that need maintenance:
+
+`pacdiff`
+
+- Use sudoedit to merge or remove files:
+
+`pacdiff -s`
```
adrienthebo, line 13:
```suggestion
`pacdiff --sudo`
```

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,29 @@
...
+
+`pacdiff -b`
+
+- Find files that need maintenance, apply changes with sudoedit, and keep backups:
+
+`pacdiff -b -s`
+
+- Scan for files with `locate` instead of using pacmandb (much slower, but finds files missing from db):
+
+`pacdiff -l`
```
adrienthebo, line 25:
```suggestion
`pacdiff --locate`
```

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,29 @@
...
+
+`pacdiff`
+
+- Use sudoedit to merge or remove files:
+
+`pacdiff -s`
+
+- Save old versions of modified files as `.bak` files:
+
+`pacdiff -b`
```
adrienthebo, line 17:
```suggestion
`pacdiff --backup`
```

pages/linux/pacdiff.md
isaec: If manned is prefered that's fine by me -- should I use manned or archlinux?

pages/linux/pacdiff.md
isaec: Unless you know how to use vim for diffs (I certainly don't) `pacdiff` is a really difficult tool to use :laughing:. I'll add an example instead.

isaec: Will do -- passing the `--backup` flag means that the original file is "backed up" if you modify it by overwriting it based on your pacnew.

isaec: Checked the source to be sure. Relevant code is as follows.
```bash
while :; do
			ask "(V)iew, (M)erge, (S)kip, (R)emove %s, (O)verwrite with %s, (Q)uit: [v/m/s/r/o/q] " "$file_type" "$file_type"
			read c || break
			case $c in
				q|Q) exit 0;;
				r|R) $SUDO rm -v "$pacfile"; break ;;
				o|O)
					if (( BACKUP )); then
						$SUDO mv -v "$file" "$file.bak"
					fi
					$SUDO mv -v "$pacfile" "$file"
					break ;;
				v|V)
					if view_diff "$pacfile" "$file"; then
						break
					fi ;;
				m|M)
					if merge_file "$pacfile" "$file"; then
						break
					fi ;;
				s|S) break ;;
				*) msg2 "Invalid answer." ;;
			esac
		done
```
If overwrite is selected, and *only if overwrite is selected*, passing the backup flag will cause the original file to be stored as a ``.bak``. I was under the misconception that the flag would also backup merges. Editing to clarify the behavior of the flag.

isaec:
If this is approved, shall I squash or is that automated?

pages/linux/pacdiff.md
adrienthebo: The example looks great, thanks!

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Look for files that need maintenance:
+
+`pacdiff`
+
+- Use sudo and sudoedit to remove and merge files:
+
+`pacdiff --sudo`
+
+- Store the original version of any file you `(O)verwrite with` as a `.bak`:
```
adrienthebo, line 14: Two points on this one:

1. The `[o]verwrite` syntax is used to indicate that a given concept (overwrite) maps to an option. Since this doesn't have a `-o` option we don't have that mapping. Is there another reason you chose to use the `[o]verwrite` syntax?
2. As you mentioned in a [prior comment](https://github.com/tldr-pages/tldr/pull/8474#discussion_r965517546) overwriting only happens when the `-o` flag is passed; this example doesn't use that. Is this example description correct?

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,32 @@
...
+
+`pacdiff --backup`
+
+- Override the program for `(M)erge`ing changes (`vim -d` is default):
```
adrienthebo, line 18: Similar comment as above, using `\`(M)erge\`` is special syntax to indicate that a concept maps to a CLI flag. Is this your intent?

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,32 @@
...
+
+`DIFFPROG=meld pacdiff`
+
+- Find files that need maintenance, apply changes with sudo/sudoedit, and keep backups:
+
+`pacdiff -b -s`
```
adrienthebo, line 24:
```suggestion
`pacdiff --backup --sudo`
```

pages/linux/pacdiff.md
isaec: 1. That linked code isn't a flag, rather input given during program operation. You are prompted for each file found, and choose what to do with it. Entering `O` to overwrite is one of the options. I put it in graves because it's the exact prompt selection from the code. Is there a better way to specify it?

2. I was trying to clarify with this change that the `--backup` flag was only relevant when the `o` option was selected by using the exact text from the prompt.

CleanMachine1:
> If this is approved, shall I squash or is that automated?

we squash at merge.

pages/linux/pacdiff.md
isaec: @adrienthebo this is the prompt given
```bash
ask "(V)iew, (M)erge, (S)kip, (R)emove %s, (O)verwrite with %s, (Q)uit: [v/m/s/r/o/q] " "$file_type" "$file_type"
```
How should I clarify that the flag `--backup` applies to when the input `O` is given for a file? Would not using a code block, and instead saying "(O)verwrite with" in quotes be better?

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,32 @@
+# pacdiff
+
+> Pacorig, pacnew and pacsave maintenance utility for pacman.
```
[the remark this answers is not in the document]
isaec: Should those be code blocks? They aren't commands. Could `.pacorig` make more sense?

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,32 @@
...
+
+`pacdiff`
+
+- Use sudo and sudoedit to remove and merge files:
+
+`pacdiff --sudo`
+
+- Store the original version of any file you `(O)verwrite with` as a `.bak`:
+
+`pacdiff --backup`
```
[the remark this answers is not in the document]
isaec: Does backup need sudo?

pages/linux/pacdiff.md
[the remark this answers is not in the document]
marchersimon: Oh, I didn't realize those were file types. While `.pacorig` doesn't look very good, it at least explains what they are. But it's up to you. We can also leave it this way.

pages/linux/pacdiff.md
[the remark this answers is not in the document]
marchersimon: I just checked and you need either `sudo` or `-s` when changing files.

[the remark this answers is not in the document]
isaec: Is sudo prefered over `-s`?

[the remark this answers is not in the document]
CleanMachine1: Only use sudo where it could be useful and is required. And in this case since it has the feature built in we should use it, so the `-s` IMO is the best option here for the commands that need root.

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,28 @@
+# pacdiff
+
+> Pacorig, pacnew and pacsave maintenance utility for pacman.
+> More information: <https://manned.org/pacdiff>.
+
+- Look for files that need maintenance:
```
navarroaxel, line 6:
```suggestion
- Review files that need maintenance in interactive mode:
```

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Use sudo and sudoedit to remove and merge files:
+
+`pacdiff --sudo`
+
+- Store the original version of any file you `(O)verwrite with` as a `.bak`:
+
+`pacdiff -s --backup`
+
+- Override the program for `(M)erge`ing changes (`vim -d` is default):
```
navarroaxel, line 18:
```suggestion
- Use a specific editor to view and merge conflicted files:
```

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,28 @@
...
+
+`DIFFPROG={{meld}} pacdiff`
```
navarroaxel, line 20: I don't know about `meld`
```suggestion
`DIFFPROG={{editor}} pacdiff`
```

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Scan for files with `locate` instead of searching pacmandb to find untracked files:
```
navarroaxel, line 22: I'm not sure what this mean.

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,28 @@
+# pacdiff
+
+> Pacorig, pacnew and pacsave maintenance utility for pacman.
```
navarroaxel, line 3: I think this is better
```suggestion
> Maintenance utility for `.pacorig`, `.pacnew` and `.pacsave` files created by `pacman`.
```

pages/linux/pacdiff.md
isaec: Sometimes `.pacnew` files end up on your system without being tracked in pacmandb -- I have personally had this happen. Instead of querying the database, the locate flag will scan your filesystem for matching files. How should the wording be clarified?

pages/linux/pacdiff.md
isaec: @navarroaxel is it not important to clarify the default? Do you think it's valuable to quote the exact prompt of "(M)erge" instead of paraphrasing?

navarroaxel: you're right!

isaec: It should be present tense, "conflicting" right? Do you think the exact prompt should be included?

navarroaxel: which prompt ?

pages/linux/pacdiff.md
navarroaxel:
```suggestion
- Scan for configuration files with `locate` instead of using pacman database:
```

pages/linux/pacdiff.md
```diff
@@ -0,0 +1,28 @@
...
+
+`pacdiff`
+
+- Use sudo and sudoedit to remove and merge files:
+
+`pacdiff --sudo`
+
+- Store the original version of any file you `(O)verwrite with` as a `.bak`:
+
+`pacdiff -s --backup`
```
navarroaxel, line 16: We prefer long arguments if available
```suggestion
`pacdiff --sudo --backup`
```

navarroaxel (commented):
Ok, now I know how to use this utility :grin: awesome!

pages/linux/pacdiff.md
EmilySeville7cfg: > How should I clarify that the flag --backup applies to when the input O is given for a file? Would not using a code block, and instead saying "(O)verwrite with" in quotes be better?

It's undefined in our style guide how you should specify `O` in this case. Until it's formalized use what you and other reviewers prefer. Good catch - this should be described in our style guide. ;)

EmilySeville7cfg (approved):

pages/linux/pacdiff.md
isaec: When you run the command, you get prompted repeatedly to decide what to do with the given file -- I felt it was more clear to include the exact option of "(M)erge" to clarify when the variable matters. The full prompt is
```bash
ask "(V)iew, (M)erge, (S)kip, (R)emove %s, (O)verwrite with %s, (Q)uit: [v/m/s/r/o/q] " "$file_type" "$file_type"
```
I don't know if `DIFFPROG` is also used for `(V)iew`ing? I can't remember... `pacdiff` help says:
```
Environment Variables:
  DIFFPROG        override the merge program (default: 'vim -d')
 ```
 Checking the source code, and V calls view_diff, and M calls merge_file...
Both of those functions use `$diffprog`, so it is used for both viewing and merging files.

Would it be acceptable to update this to:
> - Use a specific editor to `(V)iew` and `(M)erge` configuration files (default is `vim -d`):

or is this unnecessarily clear?

Comment (navarroaxel) [reply pages/linux/pacdiff.md:18]: IMHO the prompt is pretty clear. I focus the example description on which review process the example starts and not how to interact with the `pacdiff` script itself.

File: pages/linux/pacdiff.md (hunk 6 above)
Comment (kbdharun) [reply pages/linux/pacdiff.md:14]: > > How should I clarify that the flag --backup applies to when the input O is given for a file? Would not using a code block, and instead saying "(O)verwrite with" in quotes be better?
> 
> It's undefined in our style guide how you should specify `O` in this case. Until it's formalized use what you and other reviewers prefer. Good catch - this should be described in our style guide. ;)

Yeah, we should document this case, as some other programs might have prompts too.

**Review by kbdharun — APPROVED** [review]

File: pages/linux/pacdiff.md (hunk 6 above)
Comment (navarroaxel) [reply pages/linux/pacdiff.md:14]: For the moment you run this example I think `current` is more accurate.
```suggestion
- Backup the current version of any file you `(O)verwrite` as a `.bak`:
```

navarroaxel: Also, if I'm explaining the full command behavior:
```suggestion
- Review files that need maintenance creating a backup (`.bak`) of the original version of any file you `(O)verwrite`:
```

navarroaxel (changes requested):
I'm blocking the merge until https://github.com/tldr-pages/tldr/pull/8474/files#r967365099 will be resolved.

pages/linux/pacdiff.md
isaec: I think original is more clear then current...

Does the entire command functionality need to be reestablished on each explanation? I don't see this trend followed on other pages.

navarroaxel: I think the example description should explain all the command behavior. It isn't clear for me that this command starts a review process.

isaec: Is the new language clear enough?

navarroaxel: `pacfile` is not domain language, if you google it you don't receive links to Arch Linux.

isaec: Saying `.pac*` doesn't work either...

isaec: is this revision good descriptive enough @navarroaxel?

navarroaxel (approved):

PR #8486, opened 2022-09-12
gdb: add French translation

mcombeau:

### DIFF
--- a/pages.fr/common/gdb.md
+++ b/pages.fr/common/gdb.md
@@ -1,0 +1,24 @@
+# gdb
+
+> Le débogueur GNU.
+> Plus d'informations : <https://www.gnu.org/software/gdb>.
+
+- Débogue un exécutable :
+
+`gdb {{exécutable}}`
+
+- Attache un processus à gdb :
+
+`gdb -p {{identifiant_processus}}`
+
+- Débogue avec un fichier comme image mémoire :
+
+`gdb -c {{fichier}} {{exécutable}}`
+
+- Execute les commandes gdb données au démarrage :
+
+`gdb -ex "{{commandes}}" {{exécutable}}`
+
+- Démarre gdb en passant des arguments à l'exécutable :
+
+`gdb --args {{exécutable}} {{argument1}} {{argument2}}`


pages.fr/common/gdb.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Débogue un exécutable :
+
+`gdb {{exécutable}}`
+
+- Attache un processus à gdb :
+
+`gdb -p {{identifiant_processus}}`
+
+- Débogue avec un fichier comme image mémoire :
```
kbdharun, line 14: I am not sure about this command, the original page has `Debug with a core file:` but when I checked your page with Google Translate it shows `- Debug with a file as dump:`, also from what I remember the command below has fishier which means file. Shouldn't it be core?

mcombeau: Hello, and thank you for your careful review, @kbdharun!

The word "core" is a very difficult word to translate into French. There is no exact equivalent. "Cœur" (meaning "heart") is perhaps the closest but is mainly only used for processor cores. "Noyeau" ("kernel") is another possibility but is typically only used to refer to the OS kernel.

In doubt, I checked the French version man page for gdb (available [here](https://jp.barralis.com/linux-man/man1/gdb.1.php) and [here](https://reposcope.com/man/fr/1/gdb)) which describes the `-c` option as:

```
-core=fichier
-c fichier
        Utiliser le fichier comme image mémoire à examiner.
```

This is as direct a translation of the English gdb man page as possible, which says:

```
--core=file
-c file
        Use file as a core dump to examine.
```

I thought the important concept in these descriptions was the idea of a dump file rather than a "core file", which is why I chose to go with the wording of the man page description, over the exact wording of the English tdlr gdb page.

Please let me know if this is an acceptable compromise or if I should try to incorporate the English word "core" in this translation. Thank you!

kbdharun: Thanks for the explanation 😃, I learnt some new things. I think your current translation is fine.

marchersimon (approved):
Thanks @mcombeau

PR #8493, opened 2022-09-16
clang-format: exhaustive style list

BuonOmo:


kbdharun (approved):
Welcome to tldr @BuonOmo, The changes LGTM. Thanks for your contribution.

CleanMachine1:
While we should definitely add the GNU and the microsoft, I think both links should have the exhaustive list so that the user doesn't have to read the other command examples for context.

pages/common/clang-format.md
```diff
@@ -13,12 +13,12 @@
 
 - Format a file using a predefined coding style:
 
-`clang-format --style={{LLVM|Google|Chromium|Mozilla|WebKit}} {{path/to/file}}`
+`clang-format --style={{LLVM|GNU|Google|Chromium|Microsoft|Mozilla|WebKit}} {{path/to/file}}`
```
EmilySeville7cfg, line 16: There are too many alternatives here. I think we should define the maximum limit of alternative placeholder count not to allow very long commands.

pages/common/clang-format.md
```diff
@@ -13,12 +13,12 @@
 
 - Format a file using a predefined coding style:
```
EmilySeville7cfg, line 14:
```suggestion
- Format a file using a specific predefined coding style:
```

EmilySeville7cfg (changes requested):

pages/common/clang-format.md
BuonOmo: @EmilySeville7cfg those changes could be nice as well but I don't think they are fully necessary, and I am definitely not the author of those. I don't file like it should be done in this PR or even by me, hence I am not sure why you _requested changes_ here?

EmilySeville7cfg: I've requested changes because another my comment requires them.

BuonOmo: @EmilySeville7cfg I guess you mention https://github.com/tldr-pages/tldr/pull/8493#discussion_r973734282.. My answer would be the same: my pull request is about fixing the list and I don't think I should be responsible I should do any other change in this PR.

EmilySeville7cfg: So I am going to resolve this conversation.

pages/common/clang-format.md
EmilySeville7cfg: This change must be done in another PR by somebody else. I approve this PR with keep this note in mind.

EmilySeville7cfg (approved):
Approved, but requires changes in other PRs. See my comments for details.

PR #8503, opened 2022-09-19
cdk: fix typo in example

michalani:
As per https://docs.aws.amazon.com/cdk/v2/guide/cli.html#cli-ref The proper command is `cdk docs`


kbdharun (approved):
LGTM, Welcome to tldr @michalani, Great catch although `cdk doc` is a correct command it is an alias of `cdk docs` so having the main command would be better.

navarroaxel (approved):

PR #8542, opened 2022-09-29
warpd: add page

zim0369:

### DIFF
--- a/pages/linux/warpd.md
+++ b/pages/linux/warpd.md
@@ -7,6 +7,10 @@
 
 `warpd --normal`
 
+- Run warpd in hint mode:
+
+`warpd --hint`
+
 - Move cursor left:
 
 `h`
@@ -26,7 +30,3 @@
 - Emulate left click:
 
 `m`
-
-- Run warpd in hint mode:
-
-`warpd --hint`
--- a/pages/linux/warpd.md
+++ b/pages/linux/warpd.md
@@ -1,27 +1,32 @@
 # warpd
 
 > A modal keyboard driven pointer manipulation program.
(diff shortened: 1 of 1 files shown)


kbdharun (commented):
I think we can reorder the commands a bit. We could move **Line 30-32** (`hint mode`) to the top after `normal mode`.  Or we could move `normal mode` to last before `hint mode`.

CleanMachine1 (approved):

kbdharun (approved):

zim0369:
@kbdharun I assumed that if the user tries these commands and bindings in order, then having them user the `--normal` mode would be a nice intro to how the cursor is used with warpd. Then in the end `--hint` mode would show them a convenient warpd feature.

zim0369:
I would also like to know why listing multiple bindings in a single bullet isn't allowed by the parser. I felt that the movement bindings here would've been better in a single point:
```md
- Movement left/down/up/right:
`h`, `j`, `k` & `l` respectively
```

pages/linux/warpd.md
```diff
@@ -0,0 +1,32 @@
...
+
+`l`
+
+- Emulate left click:
+
+`m`
+
+- Run warpd in hint mode:
+
+`warpd --hint`
```
navarroaxel, line 32: We use to list examples about command arguments, then hot keys for interactive mode. So, IMO this should be the 2nd example in this page.

kbdharun: Agreed that was my suggestion too, as it would better align with our other pages. The PRs Author commented on this above.

PR #8548, opened 2022-09-30
links: add page

suyashbhawsar:

### DIFF
--- a/pages/linux/links.md
+++ b/pages/linux/links.md
@@ -9,7 +9,7 @@
 
 - Apply restrictions for anonymous account:
 
-`links -anonymous {{example.com}}`
+`links -anonymous {{https://example.com}}`
 
 - Enable Cookies (1 to enable):
 
--- a/pages/linux/links.md
+++ b/pages/linux/links.md
@@ -11,7 +11,7 @@
 
 `links -anonymous {{https://example.com}}`
 
-- Enable Cookies (1 to enable):
+- Enable Cookies (`1` to enable):
 
 `links -enable-cookies {{0|1}} {{https://example.com}}`
 
--- a/pages/linux/links.md
+++ b/pages/linux/links.md
@@ -13,7 +13,7 @@
 
 - Enable Cookies (1 to enable):
 
-`links -enable-cookies <0/1> {{example.com}}`
+`links -enable-cookies {{0|1}} {{example.com}}`
 
 - Navigate forwards and backwards through the links on a page:
 
--- a/pages/linux/links.md
+++ b/pages/linux/links.md
@@ -5,7 +5,7 @@
 
 - Visit a website:
 
-`links {{example.com}}`
+`links {{https://example.com}}`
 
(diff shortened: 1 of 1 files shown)


suyashbhawsar:
@kbdharun

pages/linux/links.md
```diff
@@ -0,0 +1,28 @@
+# links
+
+> Command-line web browser.
+> More information: <http://links.twibright.com/>.
+
+- Visit a website:
+
+`links {{example.com}}`
```
CleanMachine1, line 8:
```suggestion
`links {{https://example.com}}`
```

pages/linux/links.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Apply restrictions for anonymous account:
+
+`links -anonymous {{example.com}}`
```
CleanMachine1, line 12:
```suggestion
`links -anonymous {{https://example.com}}`
```

pages/linux/links.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Enable Cookies (1 to enable):
+
+`links -enable-cookies <0/1> {{example.com}}`
```
CleanMachine1, line 16:
```suggestion
`links -enable-cookies {{0|1}} {{example.com}}`
```

pages/linux/links.md
```diff
@@ -0,0 +1,28 @@
...
+
+`Up arrow key, Down arrow key`
+
+- Go forwards and backwards one page:
+
+`Left arrow key, Right arrow key`
+
+- Exit:
+
+`q then y`
```
CleanMachine1, line 28: I don't think we should include these.

suyashbhawsar: Okay

suyashbhawsar: But, it's included in the `lynx` page

suyashbhawsar:
@kbdharun can you review this? Also, what are your suggestions with the navigations, should it be included in the page?

pages/linux/links.md
```diff
@@ -0,0 +1,28 @@
...
```
kbdharun, line 28:
```suggestion
`q then y keys`
```

kbdharun (approved):
> @kbdharun can you review this? 

Sure

>Also, what are your suggestions with the navigations, should they be included on the page?

As @CleanMachine1 suggested we don't generally include those but for special cases like this we would make an exception as these shortcuts would be essential for users to navigate through the application.

pages/linux/links.md
suyashbhawsar: I think it'd be better to keep this page similar to the `lynx` page.

kbdharun: Sure, I too think the same. I will resolve the conversation.

kbdharun (approved):

suyashbhawsar:
@CleanMachine1 do you have any more suggestions for this page?

blueskyson (approved):

PR #8569, opened 2022-10-01
aria2c: Add Indonesian translation

frostg-012:
- [x] The page(s) are in the correct platform directories: `common`.

### DIFF
--- a/pages.id/common/aria2c.md
+++ b/pages.id/common/aria2c.md
@@ -1,0 +1,37 @@
+# aria2c
+
+> Utilitas unduhan cepat.
+> Mendukung HTTP(S), FTP, SFTP, BitTorrent, dan Metalink.
+> Informasi lebih lanjut: <https://aria2.github.io>.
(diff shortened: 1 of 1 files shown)


reinhart1010:
LGTM to me, even though the `{{url}}` part could be replaced into something like `{{http://contoh.com}}`, just like what we did in `curl` and `wget`.

reinhart1010 (approved):
Oh, I see that the English version also uses `url` by default.

kbdharun (approved):

kbdharun:
The translation LGTM.

@reinhart1010 Your suggestion is good, Also the main page refers to both URI and URL. I am confused because 
> URI (Uniform Resource Identifiers) can represent both the URL and the URN of a resource, simultaneously, while URL can only specify the address of the resource on the internet. 

Let me ask somebody else for a review of the original page.

blueskyson (approved):

PR #8577, opened 2022-10-01
php-cs-fixer: add page

owenvoke:
- **Version of the command being documented (if known):** v3.11.0

### DIFF
--- a/pages/common/php-cs-fixer.md
+++ b/pages/common/php-cs-fixer.md
@@ -1,0 +1,36 @@
+# PHP-CS-Fixer
+
+> Automatic coding style fixer for PHP.
+> More information: <https://github.com/FriendsOfPHP/PHP-CS-Fixer>.
+
+- Execute code style fixing:
+
+`php-cs-fixer fix`
+
+- Execute code style fixing for a specific directory:
+
+`php-cs-fixer fix {{path/to/directory}}`
+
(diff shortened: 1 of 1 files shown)


pages/common/php-cs-fixer.md
```diff
@@ -0,0 +1,36 @@
+# PHP-CS-Fixer
+
+> Automatic coding style fixer for PHP.
+> More information: <https://github.com/FriendsOfPHP/PHP-CS-Fixer>.
+
+- Execute code style fixing:
```
CleanMachine1, line 6: I assume this is correct since the next example specifies a directory.
Better to add some clarity if so.
```suggestion
- Execute code style fixing in the current directory:
```

CleanMachine1 (approved):

kbdharun (approved):
LGTM, after [CleanMachine1's](https://github.com/CleanMachine1) suggestion.

pages/common/php-cs-fixer.md
owenvoke: Yeah, sounds good. 👍🏻 It's a bit weird as technically when you don't specify a path it uses directories from a config. So you might have a config that specifies only files in a certain subdirectory, even though you're running the command in the parent directory (where the config is).

kbdharun (approved):
LGTM

PR #8588, opened 2022-10-01
git commit: Use spelled-out version of single letter options 

jay-tau:
Implement point 3 of https://github.com/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines

### DIFF
--- a/pages/common/git-commit.md
+++ b/pages/common/git-commit.md
@@ -5,7 +5,7 @@
 
 - Commit staged files to the repository with a message:
 
-`git commit -m "{{message}}"`
+`git commit --message="{{message}}"`
 
 - Commit staged files with a message read from a file:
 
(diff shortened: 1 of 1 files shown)


kbdharun:
Usually, we prefer longer flags and in very rare instances single letter flags, due to ease of use and many user requests we have used shorter flags. I will drop a message in our chatroom regarding this PR(in gitter/matrix [Link in issues tab]), let's see what others think.

kbdharun:
I am closing this PR, this is an excerpt from our [chatroom](https://gitter.im/tldr-pages/tldr) regarding this PR

IMO if the short option name is easy to remember and explain (e.g. "m" for "message" and "a" for "all"), it's OK to use the short option, as long as we refer the expanded name in the command description. The use of spelled-out options is more useful when the letter does not directly map to a word we'd use to explain what it does, or when the association is awkward (e.g. -x for "expanded" is ok, but -k for "convert links" is not).
I'll also repeat what I said recently in this chat, in response to a similar question 
> We've always focused on clarity rather than conciseness (guideline [#2](https://github.com/tldr-pages/tldr/issues/2) in https://github.com/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines), so we try to explain what a short flag's letter stands for in a way that helps people memorize it.
This is often done in the example's description (for example, using the word "verbose" for a -v option, or "e[x]panded" (with the square brackets) for an -x option).
Sometimes we even make up a "backronym", i.e. come up with a word that explains what a short option means, even if it wasn't the original reason that letter was picked by the program's authors.
But where this is not possible or would result in an awkward description, we resort to listing the spelled-out option in the example (e.g. when the letter doesn't really stand for anything, which is common for commands that have lots of options).

PR #8666, opened 2022-10-02
CONTRIBUTING: fix typo caja to tar

dz-at-tc:
An alternative fix would be to update the link to https://github.com/tldr-pages/tldr/blob/main/pages/linux/caja.md.


kbdharun (approved):

navarroaxel (approved):

PR #8710, opened 2022-10-04
fossil-init: add page

dz-at-tc:
Per issue #6458

### DIFF
--- a/pages/common/fossil-init.md
+++ b/pages/common/fossil-init.md
@@ -1,7 +1,7 @@
 # fossil init
 
 > Initialize a new repository for a project.
-> See also: `fossil clone`
+> See also: `fossil clone`.
 > More information: <https://fossil-scm.org/home/help/init>.
 
 - Create a new repository in a named file:
--- a/pages/common/fossil-init.md
+++ b/pages/common/fossil-init.md
@@ -1,0 +1,9 @@
+# fossil init
+
+> Initialize a new repository for a project.
+> See also: `fossil clone`
+> More information: <https://fossil-scm.org/home/help/init>.
+
+- Create a new repository in a named file:
+
+`fossil init {{path/to/filename}}`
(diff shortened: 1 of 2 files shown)


pages/common/fossil-init.md
```diff
@@ -0,0 +1,9 @@
+# fossil init
+
+> Initialize a new repository for a project.
+> See also: `fossil clone`.
+> More information: <https://fossil-scm.org/home/help/init>.
+
+- Create a new repository in a named file:
+
+`fossil init {{path/to/filename}}`
```
mfrw, line 9: Should we say repository or directory instead ?
```suggestion
`fossil init {{path/to/repository}}`
```

kbdharun: Or would this be better

```suggestion
`fossil init {{path/to/file}}`
```

dz-at-tc: Hmmm, as to filename vs file, I went with filename given the style guide [here](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/style-guide.md#paths), but I'm not wed to it.

As to filename vs repository, just looking at the command alone I see your point...  but I wonder if that makes it clear that repositories are in fact single files in fossil (sqlite files under the hood, vs directories like git and svn).

I think the combination of "Create a new repository in a named file" + "fossil init {{path/to/filename}}" taken together helps to sew together the bigger picture idea that a repository is in fact a file(name).  But I'm happy to change it if you do not agree.

dz-at-tc:
So given my comments above, and given @kbdharun 's thumbs up on them, I'm going to assume that this PR is done / ready to merge unless someone disagrees and says otherwise, thanks!

pages/common/fossil-init.md
mfrw: Today i learnt, fossil does things differently

mfrw (approved):
Thank you :)
And yes, I nitpick a lot :(
I hope you don't mind :)

kbdharun (approved):

PR #8744, opened 2022-10-05
android/*: add Polish translations

kruczyna:

### DIFF
--- a/pages.pl/android/settings.md
+++ b/pages.pl/android/settings.md
@@ -1,0 +1,20 @@
+# settings
+
+> Uzyskaj informacje o systemie operacyjnym Android.
+> Więcej informacji: <https://adbinstaller.com/commands/adb-shell-settings-5b670d5ee7958178a2955536>.
+
+- Listuj ustawienia w przestrzeni `global`:
+
+`settings list {{global}}`
+
+- Wyświetl wartość określonego ustawienia:
+
+`settings get {{global}} {{airplane_mode_on}}`
+
+- Ustaw wartość dla określonego ustawienia:
+
+`settings put {{system}} {{screen_brightness}} {{42}}`
+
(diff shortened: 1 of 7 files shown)


pages.pl/android/settings.md
```diff
@@ -0,0 +1,20 @@
...
+
+`settings get {{global}} {{airplane_mode_on}}`
+
+- Ustaw wartość dla określonego ustawienia:
+
+`settings put {{system}} {{screen_brightness}} {{42}}`
+
+- Delete a specific setting:
+
+`settings delete {{secure}} {{screensaver_enabled}}`
```
kbdharun, line 20:
```suggestion
`settings delete {{bezpieczne}} {{screensaver_enabled}}`
```
If possible, Can you translate the other arguments on this page?
> Note: You can replace the word in my suggestion as I used Google Translate

kruczyna: hi @kbdharun !
I've looked through the variables from the arguments, but TBH it does make much sense from understanding-point of view 😅 
As a Polish person, you will understand them when translated this way, but you'll have to know English to translate them back and make general sense of those commands. Secure is indeed `bezpieczne`, but in this command context, it can also be `zablokowane`, `zabezpieczone`, `choronione`, and so on...
The same goes, unfortunately, for the rest of the arguments 😅 
All in all, if it's necessary, I will translate those, but I just wanted to drop this personal advice, having the advantage of being a native 🙏 
The decision belongs to you, of course, and please let me know what you think 😃

kbdharun: Thanks for the insightful reply, I think the current page is fine.

kbdharun (approved):

navarroaxel (approved):

Issue #8525, opened 2022-09-26
page request: mumble


gutjuri:
### Command description

Mumble is a software for voice-chatting.

I think it would be great to document the subcommand `mumble rpc`.

### OS

- [ ] Android
- [X] Linux
- [x] macOS
- [ ] sunOS
- [X] Windows

EltonAlvess:
i notice that doesn't have any references to a documentation.

The request is about server or client side?

gutjuri:
This request is about the client side

burned42:
I opened https://github.com/tldr-pages/tldr/pull/8750 documenting the mumble client for linux :)

pixelcmtd:
@burned42 Is there any reason you created the page under `linux`? (other than maybe not being familiar with `common` 😁)

burned42:
@pixelcmtd not really, no. I just use the linux version, so that's the only one I'm familiar with and which I was confident enough documenting. If it works identical on other platforms, it could of course also be put into `common`.

PR #8750, opened 2022-10-05
mumble: add page

burned42:
- [x] The page(s) are in the correct platform directories: `linux`
- **Version of the command being documented (if known):** 1.3.4

Related to https://github.com/tldr-pages/tldr/issues/8525 but only covering the linux version.

### DIFF
--- a/pages/linux/mumble.md
+++ b/pages/linux/mumble.md
@@ -1,0 +1,24 @@
+# mumble
+
+> Low-latency, high quality voice chat software.
+> More information: <https://www.mumble.info>.
+
+- Open Mumble:
+
+`mumble`
+
+- Open Mumble and immediately connect to a server:
+
+`mumble mumble://{{username}}@{{example.com}}`
+
+- Open Mumble and immediately connect to a password protected server:
+
+`mumble mumble://{{username}}:{{password}}@{{example.com}}`
+
+- Mute or unmute yourself in running Mumble instance:
+
+`mumble rpc {{mute|unmute}}`
+
+- Mute or unmute yourself and the audio output of Mumble:
+
+`mumble rpc {{deaf|undeaf}}`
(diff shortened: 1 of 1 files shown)


pages/linux/mumble.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Open Mumble and immediately connect to a server:
+
+`mumble mumble://{{username}}@{{example.com}}`
+
+- Open Mumble and immediately connect to a password protected server:
+
+`mumble mumble://{{username}}:{{password}}@{{example.com}}`
+
+- Mute or unmute yourself in running Mumble instance:
```
navarroaxel, line 18: We try to avoid `you`/`your` in the examples, are we talking about the :microphone: ? or we can use `input audio` too.
```suggestion
- Mute/unmute the microphone in a running Mumble instance:
```

pages/linux/mumble.md
```diff
@@ -0,0 +1,24 @@
...
+
+`mumble rpc {{mute|unmute}}`
+
+- Mute or unmute yourself and the audio output of Mumble:
```
navarroaxel, line 22:
```suggestion
- Mute/unmute the microphone and the audio output of Mumble:
```

pages/linux/mumble.md
burned42: @navarroaxel oh ok, I didn't know that you/your should be avoided, must've missed that in the guidelines :see_no_evil: 

So in general yes, it's about whatever is configured as audio input device, which is probably set to a microphone because that's what the software is for, talking to others. I honestly don't know if there are other use cases where you would set your input device to something different than a microphone.
The button you can push in the GUI to do the same is also a microphone BTW

I'm ok with your suggestion and applied it in [1970b4cb26da0bff1fb248224c59b68efcd9ea23](https://github.com/tldr-pages/tldr/pull/8750/commits/1970b4cb26da0bff1fb248224c59b68efcd9ea23)

kbdharun (approved):

navarroaxel (approved):

PR #8789, opened 2022-10-07
qm-create: add German translation

Braintelligence:
- [x] The page(s) are in the correct platform directories: `common`, `linux`, `osx`, `windows`, `sunos`, `android`, etc.
- [x] The page(s) have at most 8 examples.
- [x] The page description(s) have links to documentation or a homepage.
- [x] The page(s) follow the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message).
- **Version of the command being documented (if known):**
version 7.2, Wed 04 May 2022 07:30:00 AM CEST

PR #8806, opened 2022-10-07
qm-create: rephrase example to better reflect command

Faro1991:
version 7.2, Wed 04 May 2022 07:30:00 AM CEST

### DIFF
--- a/pages/linux/qm-create.md
+++ b/pages/linux/qm-create.md
@@ -19,6 +19,6 @@
 
 `qm create {{100}} --archive {{path/to/backup_file.tar}} --force 1`
 
-- Specify a script that is automatically executed depending on the state of the virtual machine:
+- Specify a script that is executed automatically depending on the state of the virtual machine:
 
 `qm create {{100}} --hookscript {{path/to/script.pl}}`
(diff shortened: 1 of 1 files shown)


Faro1991:
related to #8789 - after a short discussion we found that with just four possible states, "specific triggers" is too ambiguous. The execution relies on vm state and only 4 execution points are defined.

pages/linux/qm-create.md
```diff
@@ -19,6 +19,6 @@
 
 `qm create {{100}} --archive {{path/to/backup_file.tar}} --force 1`
 
-- Specify a script that is executed on specific triggers during machine lifetime:
+- Specify a script that is automatically executed depending on the state of the virtual machine:
```
kbdharun, line 22:
```suggestion
- Specify a script that is executed automatically depending on the state of the virtual machine:
```
I think this would be better.

Faro1991: I believe you could write it both ways, but reading it over a few times, I agree. Sounds a little bit more fluent.

navarroaxel (approved):

kbdharun (approved):

PR #8700, opened 2022-10-03
style-guide: add French verb tense guideline

mcombeau:
- [ ] The page(s) are in the correct platform directories: `common`, `linux`, `osx`, `windows`, `sunos`, `android`, etc.
- [ ] The page(s) have at most 8 examples.
- [ ] The page description(s) have links to documentation or a homepage.
- [x] The page(s) follow the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message).
- **Version of the command being documented (if known):**

Hello,

There has been a lot of confusion surrounding the tense that must be used on French pages. (As seen in a recent PR [#8632](https://github.com/tldr-pages/tldr/pull/8632).) The imperative mood used on English pages is not desirable in French because it can seem rude and overly familiar.

Simply using the infinitive would be an option in French and many French-speakers will have that instinct. However, the closest tense to match the English imperative, while remaining respectful, is the third person singular indicative present, which was recommended in the now closed issue [Grammar in French - verbs #4714](https://github.com/tldr-pages/tldr/issues/4714).

To avoid confusion in future contributions, I am proposing adding this French-specific rule in the guidelines.

Please let me know if this is an acceptable change to the guidelines. I'd also love to hear other French-speaking contributors and collaborators' opinions on this (@Nico385412).

Issue #8576, opened 2022-10-01
About mass language style edits on Indonesian translations


reinhart1010:
Hello everyone,

This Hacktoberfest I'll be focusing more on fixing Indonesian translations instead of translating more commands, since I have to fix a fatal styling mistake which I've noticed and even endorsed before.

Our current style guide dictates that each command usage example must be written in imperative form, for example, "install a new package" instead of "installs a new package" or "installing a new package". However, when doing and reviewing Indonesian translations, I always preferred to write these examples in present participle version instead.

For example, see [`adb-install`](https://github.com/tldr-pages/tldr/blob/2e2243e36332cda1495639d8868ee75a128a6633/pages.id/common/adb-install.md)

> **Menginstal** aplikasi Android ke emulator/perangkat

The emphasized word here has the same meaning as "install**ing**", instead of the imperative form of "install" (i.e. "instal" or "pasang"). Yes, the official Great Indonesian Language Dictionary (KBBI) have enlisted this loan word alongside "instalasi" and "instalatur" (installer).

As a result, most of Indonesian translations here are written using this, present participle style. And since they do violate our guide, the best way to fix this is well, rewrite the pages.

That said everyone's still welcome to fix this together, as I'll be splitting this mass change into several pull requests to avoid potential conflicts with new edits throughout this Hacktoberfest.

PR #8822, opened 2022-10-08
style-guide: move language specific rules out of style guide

blueskyson:
As #8700 introduced French specific rules, we have two languages' rules. It's time to have an independent document for language specific rules to make style-guide.md less complex and more focusing on tldr's syntax.


CONTRIBUTING.md
```diff
@@ -115,7 +115,9 @@ Some examples of valid locale tags:
  - Chinese (Singapore): `zh_SG`.
  - Portuguese (Brazil): `pt_BR`.
 
-A list of translated templates for alias pages can be found in [here](contributing-guides/translation-templates/alias-pages.md).
+A list of translated templates for alias pages can be found [here](contributing-guides/translation-templates/alias-pages.md).
+
+Specific rules of Chinese and French can be found [here](contributing-guides/language-specific-rules.md).
```
kbdharun, line 120:
```suggestion
Language specific rules for pages can be found [here](contributing-guides/language-specific-rules.md).
```
I think having this generalised would be good. As many more languages will be added to this list. For example, @reinhart1010 is working on rewriting all Indonesian pages in #8576.

mcombeau:
Great idea, moving the language rules to a separate file! One thing I would suggest is also changing the line under the "Imperative mood" section which currently says :

>This also applies to all translations by default, unless otherwise specified in the language-specific section below.

Including another link to the new language rule file in this line would be useful. (Sorry, I would've suggested this small change in a review directly, but GiHub seems to think that section is too far away from your changes :sweat_smile:)

blueskyson:
@mcombeau Thanks for your reviewing. I didn't noticed that line, and now I fix it.

kbdharun:
@blueskyson I made some conflicting changes to the style guide in #8858. I think we can merge this PR after #8858 is merged.
Once the style guide is updated, You can sync this PR's branch with `main`.

kbdharun:
> @blueskyson I made some conflicting changes to the style guide in #8858. I think we can merge this PR after #8858 is merged. Once the style guide is updated, You can sync this PR's branch with `main`.

Now that #8858 is merged we can work with this PR. Sync your branch with the main. Thanks.

PR #8746, opened 2022-10-05
kubectl-*: add German translation

jnmcfly:
- [x] The page description(s) have links to documentation or a homepage.
- [x] The page(s) follow the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message).

PR #8856, opened 2022-10-09
kubectl-*: move .yaml inside token syntax

kbdharun:
Ref: #8746; In order to better align with other `kubectl` pages having a `.yaml` file extension outside the path, I am updating this page.

### DIFF
--- a/pages/common/kubectl-delete.md
+++ b/pages/common/kubectl-delete.md
@@ -29,4 +29,4 @@
 
 - Delete resources defined in a YAML manifest:
 
-`kubectl delete --filename {{path/to/manifest.yaml}}`
+`kubectl delete --filename {{path/to/manifest}}.yaml`
--- a/pages/common/kubectl-delete.md
+++ b/pages/common/kubectl-delete.md
@@ -29,4 +29,4 @@
 
 - Delete resources defined in a YAML manifest:
 
-`kubectl delete --filename {{path/to/manifest}}.yaml`
+`kubectl delete --filename {{path/to/manifest.yaml}}`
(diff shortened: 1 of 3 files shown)


pages/common/kubectl-delete.md
```diff
@@ -29,4 +29,4 @@
 
 - Delete resources defined in a YAML manifest:
 
-`kubectl delete --filename {{path/to/manifest.yaml}}`
+`kubectl delete --filename {{path/to/manifest}}.yaml`
```
navarroaxel, line 32: The extesion should be outside the token syntax only if it's mandatory. If you can save the YAML in a `.txt` file should be inside the token syntax.

kbdharun: I am not sure about this command other pages had .yaml outside so I made this PR.

navarroaxel: Maybe it's a bug in the page. @bl-ue do you remember?

kbdharun: Adding the pages having `.yaml` outside for reference.

[kubectl-describe](https://github.com/tldr-pages/tldr/blob/main/pages/common/kubectl-describe.md)
[kubectl-get](https://github.com/tldr-pages/tldr/blob/main/pages/common/kubectl-get.md)

bl-ue: @navarroaxel I'm not sure. However, I tested this command invocation and it appears that `kubectl delete --filename` can take any valid filepath. So @kbdharun in this case the `.yaml` extension should actually be inside the `{{ }}` token syntax:
```suggestion
`kubectl delete --filename {{path/to/manifest.yaml}}`
```
And yes, @kbdharun `kubectl-describe` and `kubectl-get` should both be updated as well; I tested and found their behavior to be the same as here. I think just add them to this PR.

navarroaxel (approved):

blueskyson (approved):

PR #8878, opened 2022-10-09
steam, steamcmd: add German translation

LukBukkit:
- **Version of the command being documented (if known):** Not known

### DIFF
--- a/pages.de/common/steam.md
+++ b/pages.de/common/steam.md
@@ -1,0 +1,28 @@
+# steam
+
+> Eine Plattform für Videospiele von Valve.
+> Weitere Informationen: <https://developer.valvesoftware.com/wiki/Command_Line_Options#Steam_.28Windows.29>.
+
+- Starte Steam und gebe Debug-Nachrichten auf die Standardausgabe aus:
+
+`steam`
+
+- Starte Steam und aktiviere die eingebaute Debug-Menüoption:
+
+`steam -console`
+
+- Aktiviere die Menüoption für die Steam-Konsole und öffne diese in einer aktiven Steam-Instanz:
+
+`steam steam://open/console`
+
+- Logge dich in Steam mit den angegebenen Zugangsdaten ein:
+
+`steam -login {{benutzername}} {{passwort}}`
+
+- Starte Steam im Big-Picture-Modus:
+
+`steam -tenfoot`
+
+- Stoppe Steam:
+
+`steam -shutdown`
--- a/pages.de/common/steam.md
+++ b/pages.de/common/steam.md
@@ -1,7 +1,7 @@
 # steam
 
 > Eine Plattform für Videospiele von Valve.
-> Weitere Informationen: <https://developer.valvesoftware.com/wiki/Command_Line_Options#Steam_.28Windows.29>.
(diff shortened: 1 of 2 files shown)


navarroaxel (approved):

pages.de/common/steamcmd.md
```diff
@@ -0,0 +1,16 @@
+# steamcmd
+
+> Ein Kommandozeilenwerkzeug, um über Steam verfügbare Anwendungen zu verwalten.
+> Weitere Informationen: <https://manned.org/steamcmd>.
+
+- Installiere und aktualisiere eine Anwendung ohne dich einzuloggen:
+
+`steamcmd +login anonymous +app_update {{anwendungs_id}} +quit`
```
kbdharun, line 8:
```suggestion
`steamcmd +login {{anonymous}} +app_update {{anwendungs_id}} +quit`
```
Translate `{{anonymous}}` too

pages.de/common/steamcmd.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Installiere oder aktualisiere eine Anwendung unter Angabe deiner Zugangsdaten:
+
+`steamcmd +login {{benutzername}} +app_update {{anwendungs_id}} +quit`
+
+- Installiere eine Anwendung für eine bestimmte Plattform:
+
+`steamcmd +@sSteamCmdForcePlatformType {{windows}} +login {{benutzername}} +app_update {{anwendungs_id}} validate +quit`
```
kbdharun, line 16: Update command argument for login to `anonymous`'s German equivalent.

pages.de/common/steam.md
```diff
@@ -0,0 +1,28 @@
+# steam
+
+> Eine Plattform für Videospiele von Valve.
+> Weitere Informationen: <https://developer.valvesoftware.com/wiki/Command_Line_Options#Steam_.28Windows.29>.
```
kbdharun, line 4:
```suggestion
> Weitere Informationen: <https://developer.valvesoftware.com/wiki/Command_Line_Options>.
```
I will create a new PR to make this change on other pages too.

pages.de/common/steamcmd.md
LukBukkit: The string `anonymous` is a constant required for the anonymous login, thus we can't translate it. Maybe we also should change this in the English version of the page by removing the braces. See https://developer.valvesoftware.com/wiki/SteamCMD#Anonymous

pages.de/common/steamcmd.md
LukBukkit: Here I would argue, that either we use `anonymous` as a constant (without the braces), or we use the concept of a login with an username established with the command before. Once we resolved the first discussion, we can apply the results to this command too 👍

pages.de/common/steamcmd.md
kbdharun: I guess we no need to translate `anonymous` as you suggested. Apply the curly braces in this suggestion alone and the PR is good to go.

kbdharun (approved):
LGTM, after the above suggestion, is resolved.

> Thanks for your contribution.

pages.de/common/steamcmd.md
LukBukkit: Thanks, I've applied your suggestion

PR #8897, opened 2022-10-10
ngrok: add German translation

LukBukkit:
- **Version of the command being documented (if known):** Not known

### DIFF
--- a/pages.de/common/ngrok.md
+++ b/pages.de/common/ngrok.md
@@ -1,0 +1,24 @@
+# ngrok
+
+> Reverse-Proxy, welcher einen sicheren Tunnel von einem öffentlichen Endpunkt zu einem lokal verfügbaren Webservice erstellt.
+> Weitere Informationen: <https://ngrok.com>.
+
+- Veröffentliche einen lokalen HTTP-Service auf dem angegebenen Port:
+
+`ngrok http {{80}}`
+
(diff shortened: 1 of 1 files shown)


pages.de/common/ngrok.md
```diff
@@ -0,0 +1,24 @@
+# ngrok
+
+> Reverse-Proxy, welcher einen sicheren Tunnel von einem öffentlichen Endpunkt zu einem lokal verfügbaren Webservice erstellt.
+> Weitere Informationen: <https://ngrok.com>.
+
+- Veröffentliche einen lokalen HTTP-Service auf dem angegebenen Port:
```
kbdharun, line 6:
```suggestion
- Exponieren einen lokalen HTTP-Service auf dem angegebenen Port:
```
I am not sure if `Veröffentliche` (Publish) would come here. Would this be better?

If so can you update it throughout the page.

Note: I used Google translate to validate this translation.

kbdharun: > Thanks for your review. The word `Exponieren` isn't widely used in German and thus may be confusing to beginners. Therefore, I would prefer to keep it as-it-is. As far as I can tell, the meaning remains the same in both cases as ngrok allows everyone to access the exposed / published ports from the Internet.

Thanks for the information, I guess we can go with your translation itself. Let me ask somebody else's opinion too.

navarroaxel (approved):

PR #8964, opened 2022-10-12
Added cpush windows command

Juane99:

### DIFF
--- a/pages.es/windows/cpush.md
+++ b/pages.es/windows/cpush.md
@@ -1,0 +1,8 @@
+# cpush
+
+> Este comando es un alias de `choco push`.
+> Más información: <https://docs.chocolatey.org/en-us/create/commands/push>.
+
+- Ver documentación para el comando original:
+
+`tldr choco push`


pages.es/windows/cpush.md
```diff
@@ -0,0 +1,8 @@
+# cpush
```
navarroaxel, line 1: This command doesn't exist in the English, please add it in this same PR.

pages.es/windows/cpush.md
```diff
@@ -0,0 +1,8 @@
...
+
+> Este comando es un alias de `choco push`.
+> Más información: <https://docs.chocolatey.org/en-us/create/commands/push>.
+
+- Ver documentación para el comando original:
+
+`tldr choco push`
```
navarroaxel, line 8: I think we should recommend the page with `kebab-case`, both should work in the tldr client.
```suggestion
`tldr choco-push`
```

kbdharun (approved):

PR #8965, opened 2022-10-12
clamav: fix page

navarroaxel:

Fix page for clamav because there is no command called `clamav`
Move freshclam to its own page.
Move clamav page from `linux/` to `common/` platform.


pages/common/clamav.md
```diff
@@ -0,0 +1,17 @@
+# ClamAV
+
+> Open-source anti-virus program.
+> ClamAV isn't a command, but a set of commands. See the pages below.
```
kbdharun, line 4:
```suggestion
> ClamAV isn't a command, but a set of commands.
```
Should we mention `See the pages below separately`, I think it would be self-explanatory.

kbdharun (approved):

einverne (approved):
Looks good to me.

blueskyson (approved):

PR #8975, opened 2022-10-12
dependabot.yml: check for package updates monthly

jlosito:
This will cause dependabot to check all the package dependencies which this project uses for updates in order to keep everything up-to-date.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "monthly"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+
```
navarroaxel, line 8: Why not `npm`?
```suggestion
  - package-ecosystem: "npm"
    directory: "/"
    schedule:
      interval: "weekly"
```

MasterOdin (commented):
Could also add a version check for https://github.com/tldr-pages/tldr/tree/main/scripts/pdf, while also versioning those dependencies.

jlosito:
@MasterOdin The `/scripts/pdf` directory should be added now.

kbdharun (approved):

MasterOdin (approved):

PR #8981, opened 2022-10-12
find: remove redundant -delete from -daystart example

ulysseskan:
I can see people woken up at 4am for some alert they're on-call to resolve, trying to remember the syntax for finding files modified in the last week, and accidentally deleting them because, in a sleep induced haze, they copied and pasted `find root_path -daystart -mtime -7 -delete`, then only modified `root_path` and forgot to take out `-delete`.

This PR makes it a bit less likely they'll do that.

If people want to delete files, tldr already shows them how to do that in the 8th (last) example.  (Edit: To make this clear - there are **two** examples for delete. This PR only modifies one of them to hopefully prevent accidental data loss.)

<sup><sub>As an aside, perhaps suitable for a different PR, `-daystart` is an invalid operator on macOS (and maybe *BSD too?)</sup></sub>

- [ ] The page(s) are in the correct platform directories: `common`, `linux`, `osx`, `windows`, `sunos`, `android`, etc.  [Should this be moved from common to linux due to daystart example?]

### DIFF
--- a/pages/common/find.md
+++ b/pages/common/find.md
@@ -27,9 +27,9 @@
 
 `find {{root_path}} -name '{{*.ext}}' -exec {{wc -l {} }}\;`
 
-- Find files modified in the last 7 days and delete them:
+- Find files modified in the last 7 days:
 
-`find {{root_path}} -daystart -mtime -{{7}} -delete`
+`find {{root_path}} -daystart -mtime -{{7}}`
 
 - Find empty (0 byte) files and delete them:
 


mfrw (commented):
In my opinion, the delete is not redundant as the example does mention `and delete them`.
I was wondering why do you feel we should remove it ?

navarroaxel:
This is a subjective PR, idk if 2 examples for deletion is needed.
wdyt? @kbdharun @bl-ue @CleanMachine1

kbdharun (commented):
> In my opinion, the delete is not redundant as the example does mention `and delete them`. I was wondering why do you feel we should remove it ?

I agree with mfrw, delete here is an essential example here. I understand the point the PRs author make that is that people can copy paste and run this command accidentally which would be rare. Maybe we can add in an disclaimer to the command description. What do you think?

bl-ue (approved):
There already is an example for `-delete`, so removing one doesn't seem to me like it'll cause a problem. Additionally, while admins should be careful copying commands and running them on a production system, given the previous point, I don't see any problems with it. There's my approval, but if you all feel more strongly about it, overrule me. ;)

kbdharun (approved):

mfrw (approved):
I think given the current discussion; I concur with folks here.
It is a subjective PR and I am good with it :)

Thank you @ulysseskan for your contribution

ulysseskan:
Thank you all for your thoughtful reviews.  Who decides when to merge?

bl-ue:
When enough maintainers approve, any maintainer can. ping @navarroaxel

PR #8990, opened 2022-10-13
ical: move german translation to pages.de

marciska:
- [x] The page(s) are in the correct platform directories: `common`, `linux`, `osx`, `windows`, `sunos`, `android`, etc.
- [x] The page(s) follow the [content guidelines](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#guidelines).
- [x] The PR title conforms to the recommended [templates](/tldr-pages/tldr/blob/main/CONTRIBUTING.md#commit-message).

A very recent merge has falsely set the German-translation of the ical-command to the repository root folder of the main branch. This pull request moves it to the correct folder (pages.de/common).

PR #9013, opened 2022-10-13
ical: move German translation to proper place

trolzen:
This German version of the `ical.md` page was accidentally put in the root of the project.


marciska:
There is already a PR addressing this. #8990

trolzen:
> There is already a PR addressing this. #8990

OK, sorry, missed that.

PR #9047, opened 2022-10-15
where: the `/q` parameter doesn't display anything

iridacea:


pages/windows/where.md
```diff
@@ -16,6 +16,6 @@
 
 `where /R {{path/to/directory}} {{file_pattern}}`
 
-- Display only the error code for the location of file pattern:
+- Only return the error code for the location of file pattern:
```
kbdharun, line 19:
```suggestion
- Return the error code for the location of the file pattern:
```
I think this is better. What do you think?

iridacea, line 19:
```suggestion
- Silently return the error code for the location of the file pattern:
```
@kbdharun 
What if we change it this way to reflect the fact that nothing is displayed, and the other command also return error code?

iridacea:
```suggestion
- Silently return the error code for the location of the file pattern:
```
@kbdharun
What if we change it this way to reflect the fact that nothing is displayed, and the other command also return error code?

iridacea:
OK, I decided to go with the second option.

iridacea:
Looks like I accidentally removed review request for @navarroaxel  and can't return it back.

marchersimon (approved):
Looks good, thanks @iridacea

kbdharun (approved):

Issue #8859, opened 2022-10-09
Let's Document: xz utils


gutjuri:
### Command description

xz utils is a suite of tool for handling .xz and .lzma files. Documentation can be found on the following page: https://tukaani.org/xz/

It consists of the following commands:

* [x] xz → #740
* [x] unxz → #9091
* [x] xzcat → #9091
* [x] xzdiff → #8923
* [x] xzcmp → #9704
* [x] xzgrep → #9983
* [x] xzegrep → #10185
* [x] xzfgrep → #10185
* [x] xzless → #9059, #10179
* [x] xzmore → #10186
* [x] lzma → #9091
* [x] unlzma → #9091
* [x] lzcat → #9091
* [x] lzgrep →  #10171, #10185
* [x] lzegrep → #10185
* [x] lzfgrep → #10185
* [x] lzless →  #10172, #10185
* [x] lzmore → #10185

I think some of these are aliases, this however needs further investigation.

### OS

- Linux
- macOS
- sunOS
- Windows

Nikhil-Samant:
@gutjuri , I will do further investigation and add the page

asdsyd:
Is this issue fixed?

gutjuri:
> Is this issue fixed?

No, the commands that have not been marked as completed in the original post are still missing in tldr.

navarroaxel:
> How do I work on this issue ? thanks

Same I answered here: https://github.com/tldr-pages/tldr/issues/7923#issuecomment-1275262574

acuteenvy:
> lzfegrep

There's a typo. It should be `lzfgrep`.

acuteenvy:
I created PRs for all missing pages.

PR #9059, opened 2022-10-15
xzless: add page

TerribleCodes:
- **Version of the command being documented: xz (XZ Utils) 5.2.5**
Issue: #8859

### DIFF
--- a/pages/common/xzless.md
+++ b/pages/common/xzless.md
@@ -1,0 +1,10 @@
+# xzless
+
+> Filters and displays text from `.xz` and `.lzma` files.
+> Uses `less` to present the output.
+> More information: <https://manned.org/xzless.1>.
+
+- Display text:
+
+`xzless {{path/to/archive.xz}}`
+
--- a/pages/common/xzless.md
+++ b/pages/common/xzless.md
@@ -1,7 +1,7 @@
 # xzless
 
 > Filters and displays text from `.xz` and `.lzma` files.
-> Uses `less` to present the output.
+> See also: `less`.
 > More information: <https://manned.org/xzless.1>.
 
 - Display text:
--- a/pages/common/xzless.md
+++ b/pages/common/xzless.md
@@ -2,8 +2,8 @@
 
 > Display text from `.xz` and `.lzma` files.
 > See also: `less`.
-> More information: <https://manned.org/xzless.1>.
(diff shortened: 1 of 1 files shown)


pages/common/xzless.md
```diff
@@ -0,0 +1,9 @@
+# xzless
+
+> Filters and displays text from `.xz` and `.lzma` files.
+> Uses `less` to present the output.
```
navarroaxel, line 4: Perhaps this better
```suggestion
> See also: `less`.
```

pages/common/xzless.md
```diff
@@ -0,0 +1,9 @@
+# xzless
+
+> Filters and displays text from `.xz` and `.lzma` files.
```
navarroaxel, line 3: Maybe simple is better
```suggestion
> Display text from `.xz` and `.lzma` files.
```

pages/common/xzless.md
```diff
@@ -0,0 +1,9 @@
...
+> Uses `less` to present the output.
+> More information: <https://manned.org/xzless.1>.
+
+- Display text:
```
navarroaxel, line 7:
```suggestion
- Open a compressed text file:
```

pages/common/xzless.md
TerribleCodes: I wanted to say "It uses the `less` output format to display the contents". Since the suggestion looks fine I'll commit it.

pages/common/xzless.md
TerribleCodes: Can we say `Compressed text file`? Instead of that we could use "Display text from compressed file"?

navarroaxel: you're right, this works with binaries too
```suggestion
- Open a compressed file:
```

pages/common/xzless.md
```diff
@@ -0,0 +1,9 @@
+# xzless
+
+> Display text from `.xz` and `.lzma` files.
+> See also: `less`.
+> More information: <https://manned.org/xzless.1>.
```
navarroaxel, line 5:
```suggestion
> More information: <https://manned.org/xzless>.
```

kbdharun (approved):

navarroaxel (approved):

PR #9072, opened 2022-10-15
base64: fix German translation

pixelcmtd:
I really don't wanna sound hostile towards the reviewers of #8767, but did you even read it?

### DIFF
--- a/pages.de/common/base64.md
+++ b/pages.de/common/base64.md
@@ -1,20 +1,20 @@
 # base64
 
-> Kodierung oder Dekodierung von Dateien oder Standardeingaben in/aus Base64, zur Standardausgabe.
+> Kodieren oder Dekodieren von Dateien oder Standardeingaben in/aus Base64, zur Standardausgabe.
 > Weitere Informationen: <https://www.gnu.org/software/coreutils/base64>.
 
-- Kodiert den Inhalt einer Datei als base64 und schreibt das Ergebnis nach stdout:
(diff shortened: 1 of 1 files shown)


kbdharun (approved):

kbdharun:
> I really don't wanna sound hostile towards the reviewers of #8767, but did you even read it?

Sorry for the mishap. Will be more careful in German reviews in future. I made comments regarding this in <https://github.com/tldr-pages/tldr/pull/8767#discussion_r989137619> and the author responded both can be used. I go through all the PR pages before giving an approving review. Regarding translations, I could only request the change but when a native speaker opposes it, then it lies outside my scope. I will request reviews from you in future. Thanks for fixing it.

> Me and @navarroaxel also tried requesting reviews from the German team (a team) but to no avail.

pixelcmtd:
> Can you suggest me a tool for validating German translations?

I can suggest you learn German, but I wouldn't recommend it, wasting all of that time and brain space on category theory sounds like a much better (and more fun) idea 😆

I've always heavily disliked the idea of reviewing pages in languages that you do not speak, which is why I don't do it. But I can see that it's sometimes hard to avoid, especially when other collaborators are unreachable. However, I'm really interested in what I need to do to get into the German team.

navarroaxel (approved):

kbdharun:
> > However, I'm really interested in what I need to do to get into the German team.

@navarroaxel can you add @pixelcmtd to German (a team).

PR #9081, opened 2022-10-15
apt: add Japanese translation

Liquid-system:


pages.ja/common/apt.md
```diff
@@ -0,0 +1,37 @@
+# apt
+
+> Debian系ディストリビューションで使われるパッケージ管理システムです。
+> Ubuntuのバージョンが16.04か、それ以降で対話モードを使う場合`apt-get`の代わりとして使用します。
+> 詳しくはこちら: <https://manpages.debian.org/latest/apt/apt.8.html>
+
+- 利用可能なパーケージとバージョンのリストの更新（他の`apt`コマンドの前での実行を推奨）:
+
+`sudo pat update`
```
navarroaxel, line 9:
```suggestion
`sudo apt update`
```

pages.ja/common/apt.md
```diff
@@ -0,0 +1,37 @@
...
+
+- 指定されたパッケージの検索:
+
+`apt search {{package}}`
```
navarroaxel, line 13: Please translate the placeholders for get this easier for non-English speakers.
```suggestion
`apt search {{パッケージ}}`
```

pages.ja/common/apt.md
```diff
@@ -0,0 +1,37 @@
...
+
+- インストールされている全てのパッケージを最新のバージョンにアップグレード:
+
+`sudo apt upgrade`
+
+- インストールできるすべてのパッケージを表示:
+
+`apt list`
+
+- インストールされた全てのパッケージを表示(依存関係も表示):
```
navarroaxel, line 35: Why do you use a different parentheses here? Which is the correct one?
```suggestion
- インストールされた全てのパッケージを表示（依存関係も表示）:
```

Liquid-system:
The parentheses in the last line are the difference between a two-byte character and a half-width character.
I guess we should be unified,  Which should it be?

navarroaxel:
> The parentheses in the last line are the difference between a two-byte character and a half-width character. I guess we should be unified, Which should it be?

I think we should add this rule to the https://github.com/tldr-pages/tldr/blob/main/contributing-guides/style-guide.md in a Japanese-Specific Rules section

Always use Japanese (full-width, double-byte, 全角) parentheses if they contain one or more Japanese character(s). Don't insert any spaces before or after the parentheses.

Use English (half-width, single-byte, 半角) parentheses if they contain only English characters. Should we insert a space before opening parentheses or after closing parentheses?

navarroaxel (approved):

kbdharun:
@Liquid-system can you add Japanese translation to [common-arguments](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/translation-templates/common-arguments.md) too? 
> Thanks in advance

kbdharun (approved):

Liquid-system:
Appreciate the thoughtful review!
> Use English (half-width, single-byte, 半角) parentheses if they contain only English characters. Should we insert a space before opening parentheses or after closing parentheses?

This is not so much a space as it is a letter in itself.
For example, if `( `is made into full-width characters, it becomes `（`, and so on.
And similarly `)` becomes `）`

> @Liquid-system can you add Japanese translation to [common-arguments](https://github.com/tldr-pages/tldr/blob/main/contributing-guides/translation-templates/common-arguments.md) too?

I would like to do it. However , as I wrote earlier, if `/`is made into full-width characters, it becomes `／`
The notation on the translated PATH does not seem to be very common. (Including grammatical issues)
I personally think it would be better to use a notation like `path/to/ファイル名` rather than a complete translation.

PR #9088, opened 2022-10-15
gcc: add -g to -Wall -Og example

osalbahr:
I believe "-g" is the flag for debug symbols. The tldr could do without "-Og" but it is usual to include both.

[-g](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html) Produce debugging information in the operating system’s native format (stabs, COFF, XCOFF, or DWARF). GDB can work with this debugging information.

[-Og](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#Optimize-Options) Optimize debugging experience. -Og should be the optimization level of choice for the standard edit-compile-debug cycle, offering a reasonable level of optimization while maintaining fast compilation and a good debugging experience. It is a better choice than -O0 for producing debuggable code because some compiler passes that collect debug information are disabled at -O0.

### DIFF
--- a/pages/common/gcc.md
+++ b/pages/common/gcc.md
@@ -7,9 +7,9 @@
 
 `gcc {{path/to/source1.c path/to/source2.c ...}} -o {{path/to/output_executable}}`
 
-- Show common warnings, debug symbols in output:
+- Show common warnings, debug symbols in output, and optimize without affecting debugging:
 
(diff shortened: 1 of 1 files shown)


kbdharun (approved):

navarroaxel (approved):

Issue #9042, opened 2022-10-14
page request: rmlint


waldyrious:
### Command description

There are some examples already in the existing documentation: https://rmlint.readthedocs.io/en/latest/rmlint.1.html#examples. The tutorial is also heavily based on (beginner-friendly) examples, which might also be useful: https://rmlint.readthedocs.io/en/latest/tutorial.html

The project repository is at https://github.com/sahib/rmlint.

### OS

- [ ] Android
- [X] Linux
- [X] macOS
- [ ] sunOS
- [ ] Windows

PR #9090, opened 2022-10-15
rmlint: add page

LordUbuntu:
- **Version of the command being documented (if known):** 2.9.0

This contribution fixes [this issue](https://github.com/tldr-pages/tldr/issues/9042)


kbdharun:
@LordUbuntu I think this page should be placed in `common` instead of `linux` as the commands would work with mac too.

pages/linux/rmlint.md
```diff
@@ -0,0 +1,20 @@
+# rmlint
+
+> Extremely fast tool to remove remove duplicates and other lint from your filesystem.
+> More information: <https://rmlint.readthedocs.io/en/latest/index.html>.
```
kbdharun, line 4:
```suggestion
> More information: <https://rmlint.readthedocs.io/en/latest/rmlint.1.html>.
```
My suggestion would be to directly link to the man page.

waldyrious: I'd normally prefer the main website when it does contain actual information (as opposed to many project home pages nowadays that seem to me little more than an interactive advertisement poster), which is the case here; however, the manpage is also very well written, and more to-the-point regarding how to use the tool, so in this case I'll agree with @kbdharun that it's a better target for the More info link.

pages/linux/rmlint.md
```diff
@@ -0,0 +1,20 @@
+# rmlint
+
+> Extremely fast tool to remove remove duplicates and other lint from your filesystem.
```
waldyrious, line 3: It's better to just stick to the facts, and avoid adding assessments of speed, quality, aesthetics, etc. Also, note that you wrote "remove" twice :) Here's a suggestion for this description:

```suggestion
> Identify duplicate files or directories, and other filesystem issues.
```

pages/linux/rmlint.md
```diff
@@ -0,0 +1,20 @@
...
+> More information: <https://rmlint.readthedocs.io/en/latest/index.html>.
+
+- Check directory for duplicates, empty files, and other lint:
```
waldyrious, line 6:
```suggestion
- Check a directory for duplicates, empty files, and other issues:
```

pages/linux/rmlint.md
```diff
@@ -0,0 +1,20 @@
...
+
+`rmlint {{path/to/directory}}`
+
+- Delete lint files found by an rmlint run:
```
waldyrious, line 10:
```suggestion
- Delete duplicate files found by an execution of `rmlint`:
```

pages/linux/rmlint.md
```diff
@@ -0,0 +1,20 @@
...
+
+`./rmlint.sh`
+
+- Find all duplicates with the same base name:
+
+`rmlint --match-basename {{path/to/directory}}`
+
+- Find all duplicates with the same extension:
+
+`rmlint --match-extension {{path/to/directory}}`
```
waldyrious, line 20: I think this may be a little redundant with the previous example. How about an example for locating duplicate directories, which appears to be a unique feature of this tool among other duplicate finder tools?

pages/linux/rmlint.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Check directory for duplicates, empty files, and other lint:
+
+`rmlint {{path/to/directory}}`
+
+- Delete lint files found by an rmlint run:
+
+`./rmlint.sh`
+
+- Find all duplicates with the same base name:
```
waldyrious, line 14: It may be better to be a little more explicit here:

```suggestion
- Find only duplicates that have the same filename in addition to the same contents:
```

waldyrious:
Thanks for the contribution, @LordUbuntu! I have left a few comments inline.

In addition, I'd encourage you to also include the very useful [`-S`/`--rank-by` option](https://rmlint.readthedocs.io/en/latest/tutorial.html#original-detection-selection), which allows controlling which files to consider as the original ones in a group of duplicates. I am particularly fond of the `--rank-by d` configuration, which considers the files at lower path **d**epth to be the originals.

pages/linux/rmlint.md
LordUbuntu: I'll get right to that:
- add `-S`
- add `-D`

pages/linux/rmlint.md
waldyrious: Applied in 40d9eedce1fd7c86defad42ecef67722ff95a17c.

pages/linux/rmlint.md
waldyrious: Applied in [`d53236f`](https://github.com/tldr-pages/tldr/pull/9090/commits/d53236f122d503f3fb8284f6821bfd4707421a15).

pages/linux/rmlint.md
waldyrious: Applied in [`b6375a1`](https://github.com/tldr-pages/tldr/pull/9090/commits/b6375a173de811384fa3df05af3ccadf79c1581e).

pages/linux/rmlint.md
waldyrious: Applied in [`9ad0fcb`](https://github.com/tldr-pages/tldr/pull/9090/commits/9ad0fcb8a887567c297e00bde72677c0c5b5b768).

pages/linux/rmlint.md
waldyrious: Applied in [`c446a4d`](https://github.com/tldr-pages/tldr/pull/9090/commits/c446a4da2b75581568aaaf8f5fd465c2ce5350ec).

pages/linux/rmlint.md
waldyrious: For future reference, these were added in [`cd1db75`](https://github.com/tldr-pages/tldr/pull/9090/commits/cd1db75628e0fa93d84d681769ec83ff5904dbc0).

waldyrious (approved):
I think it's a bit redundant to have two `--rank-by` examples and two very similar options (`--match-basename` / `--match-extension`), but since we're still within the recommended limit for number of examples in a page, I suppose we could leave those to be replaced later if/when further examples are to be added to this page.

LGTM! Thanks for addressing this, @LordUbuntu :)

kbdharun:
> @LordUbuntu I think this page should be placed in `common` instead of `linux` as the commands would work with mac too.

I read more about this command and seems like Linux is the primary focus of the project. Although it would work with Mac too I guess we can leave this in `pages/linux` itself.

kbdharun (approved):

waldyrious:
Oh, right, you had [already pointed this out](https://github.com/tldr-pages/tldr/pull/9090#issuecomment-1279901771). I agree. In fact, the documentation itself [explicitly mentions](https://rmlint.readthedocs.io/en/latest/install.html#build-dependencies) the support for installing on Linux, macOS and FreeBSD, so this should definitely be in the `common/` directory.

pages/linux/rmlint.md
```diff
@@ -0,0 +1,32 @@
+# rmlint
```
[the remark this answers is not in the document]
waldyrious: Applied in [`87757ef`](https://github.com/tldr-pages/tldr/pull/9090/commits/87757ef2d29f6d71fde2043672fb4f73bc4cef37).

pages/linux/rmlint.md
```diff
@@ -0,0 +1,32 @@
...
+
+`rmlint --merge-directories {{path/to/directory}}`
+
+- Mark files at lower path [d]epth as originals:
+
+`rmlint --rank-by=d {{path/to/directory}}`
+
+- Mark files with shortest name [l]ength as originals:
+
+`rmlint --rank-by=l {{path/to/directory}}`
```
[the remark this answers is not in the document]
waldyrious: Applied in [`071cac8`](https://github.com/tldr-pages/tldr/pull/9090/commits/071cac8f532a6c172ede10ede43a726bab89fa88).

pages/linux/rmlint.md
```diff
@@ -0,0 +1,32 @@
...
+
+`./rmlint.sh`
+
+- Find duplicate directory trees:
+
+`rmlint --merge-directories {{path/to/directory}}`
+
+- Mark files at lower path [d]epth as originals:
+
+`rmlint --rank-by=d {{path/to/directory}}`
```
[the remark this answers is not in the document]
waldyrious: Applied in [`e98fb23`](https://github.com/tldr-pages/tldr/pull/9090/commits/e98fb23dd99129fe6ce4ab3392ee436a139d3cb1).

pages/linux/rmlint.md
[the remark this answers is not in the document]
LordUbuntu: > Applied in [`87757ef`](https://github.com/tldr-pages/tldr/pull/9090/commits/87757ef2d29f6d71fde2043672fb4f73bc4cef37).

What did you mean by this?

marchersimon (approved):
Thanks @LordUbuntu

pages/linux/rmlint.md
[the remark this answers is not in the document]
waldyrious: This is just for future reference, so that we can later easily look up what commit resolved each discussion. Sorry if that was confusing :)

